PR #20, opened 2015-02-13
Add simple python bindings

sergey-dryabzhinsky:


sergey-dryabzhinsky:
It's buggy. Please wait till I fix it.

sergey-dryabzhinsky:
Fixed

Cyan4973:
Thanks Sergei

Cyan4973:
By the way Sergey,
note that, usually (if using lz4 situation as an example),
python and other language wrappers tend to have their own project page,
which I link from homepage.

It's not an obligation, just for you to know this solution also exists, should you prefer it.
In my opinion, it has the advantage of providing the author some deserved visibility.

sergey-dryabzhinsky:
Point taken.
Done - https://github.com/sergey-dryabzhinsky/python-zstd

Cyan4973:
Great !

PR #23, opened 2015-02-22
Align streaming compress/decompress FileIO with the improved streaming API

luben:
Hi,

This is just my attempt to understand how the streaming API is intended to work. So if I am totally off the track it is OK, just tell me.

This simplification of FIO_decompressFilename and FIO_compressFilename produces a binary identical compressed file. The decompress function uses much smaller output buffer (blockSize instead of 4*blockSize) and input buffer sized according to compressBound of output buffer. I have tried in the compression function to work with input buffer of blockSize instead if 4*blockSize but it produces larger files (not binary identical to the currently produced ones).


luben:
OK, I think I understand it a little better now - so the blocks are encoded in batches of 4 and each next block of the batch depends on having the plaintext of the previous blocks in the batch.

luben:
Please disregard this pull request.

Cyan4973:
Hi Luben

Yes you are right.
In current source code, blocks are encoded in batches of 4.

For information, note this is just a temporary construction.
The intended plan is for each block to access up to the 3 previous blocks, in a continuous way (round robbin buffer fashion).

This will require some change of code though, and is therefore not ready yet.

luben:
Great. Using ring buffer should improve the compression ratio.

I have created a highly experimental JNI bindings based on the streamingAPI2 branch  - they give access to compress/decompress (from zstd.h) and implement the java InputStream/OutputStream interfaces for transparent compression/decompression using the streaming functions (from zstd_static.h). Currently they are fully interoperable with the files produced by zstd util.

I am not rushing make it public or to use it. So don't fear to break compatibility to make it better.

Regards,
luben

luben:
Hi,

I have pushed my JNI bindings to: https://github.com/luben/Zstd-JNI if you want to put a link to them.

Thanks,
luben

Cyan4973:
Great, I will.
For the time being, github is the "homepage", which puts some restriction on the type of content and even the update method (readme.md).
But I intend to have a separate, more formal, homepage, from which I can link all versions, including yours. I'll drop a note when it's ready.
Regards

luben:
Great, I will update the homepage when it's ready

PR #24, opened 2015-02-25
Removed redefined types to support GCC

technion:
The current build fails with GCC 4.4.7 due to redefined types.

    gcc      -I../lib  -O3 -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Wstrict-prototypes     -DZSTD_VERSION=\"r0.0.1\"  ../lib/zstd.c xxhash.c bench.c fileio.c zstdcli.c -o zstd
    ../lib/zstd.c:108: error: redefinition of typedef âBYTEâ
    ../lib/fse.c:95: note: previous declaration of âBYTEâ was here

..and so on

I've suggested a header for these type definitions. I've tested both GCC and Clang with this configuration.

This is the latest version of GCC available for RedHat/CentOS 6, so is expected to be common.

### DIFF
--- a/lib/fse.c
+++ b/lib/fse.c
@@ -85,29 +85,9 @@
 #include <string.h>     /* memcpy, memset */
 #include <stdio.h>      /* printf (debug) */
 #include "fse_static.h"
+#include "zstdtypes.h"
 
 
-/****************************************************************
-*  Basic Types
-*****************************************************************/
-#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* C99 */
-# include <stdint.h>
-typedef  uint8_t BYTE;
-typedef uint16_t U16;
-typedef  int16_t S16;
-typedef uint32_t U32;
-typedef  int32_t S32;
-typedef uint64_t U64;
-typedef  int64_t S64;
-#else
-typedef unsigned char       BYTE;
-typedef unsigned short      U16;
-typedef   signed short      S16;
-typedef unsigned int        U32;
-typedef   signed int        S32;
-typedef unsigned long long  U64;
-typedef   signed long long  S64;
-#endif
 
 
 /****************************************************************
--- a/lib/zstd.c
+++ b/lib/zstd.c
@@ -100,26 +100,7 @@
 #endif
 
 
-/********************************************************
-*  Basic Types
-*********************************************************/
-#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   // C99
-# include <stdint.h>
-typedef  uint8_t BYTE;
-typedef uint16_t U16;
-typedef  int16_t S16;
-typedef uint32_t U32;
-typedef  int32_t S32;
-typedef uint64_t U64;
-#else
-typedef unsigned char       BYTE;
-typedef unsigned short      U16;
-typedef   signed short      S16;
-typedef unsigned int        U32;
-typedef   signed int        S32;
-typedef unsigned long long  U64;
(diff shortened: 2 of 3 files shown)


Cyan4973:
Right.

Curiously enough, I'm using GCC 4.8 and 4.6 for my tests, and none of them complain about that.
But clang does, and that's why the following lines have been added : 

```
#  ifdef __clang__
#    pragma clang diagnostic ignored "-Wtypedef-redefinition"
#  endif 
```

Maybe something equivalent could be added for gcc too, such as : 
```
#    pragma gcc diagnostic ignored "-Wtypedef-redefinition"
```

Note that the current situation, which is a consequence of including `'fse.c'` rather than `'fse.h'`, is not satisfying. This is due to the fact than gcc/clang look unable, or very limited, in their ability to inline functions from another module, which is a killer for performance.
I'm therefore extremely likely to change this construction in the future, although final solution is not yet known.
At best, I would like to include a normal `.h` file, but on condition of preserving performance for gcc/clang. At worst I plan to include all "memory types and routines" within their own module.
Which solution is most appropriate is not yet settled.

technion:
Yep, I noted after sending the pull that the Travis build runs a more recent GCC and doesn't have a problem with the existing method. Similarly, I get a tonne of warnings about strict aliasing rules which travis doesn't seem to mention.

    ../lib/fse.c:737: warning: dereferencing pointer âCTable.35â does break strict-aliasing rules

As far as I'm aware my PR shouldn't hurt performance - it doesn't go away with fse.c being included directly.

If you want to go in the pragma direction I'm happy to test that out.

Cyan4973:
> If you want to go in the pragma direction I'm happy to test that out.

Yes please,
I'm not planning to add to any file at this stage to lib directory at this stage.
It sounds very little, but every file add complexity for project integration, so I'm trying to avoid that.

What may happen later on is that `fse.c` will simply disappear, being directly integrated into `zstd.c`. At which stage, the question relative to memory access types and functions will be already settled. As a consequence, at the cost of becoming bigger, `zstd.c` will remain the sole source file to include, simplifying integration process.

So yes, even the number of files matters.

For the test, adding 
`#    pragma gcc diagnostic ignored "-Wtypedef-redefinition"`
just after the clang one should do the trick.

technion:
>you said it issued an error

Correct, it is a fatal error and will not build. I have tested the pragma but gcc seems to go out of its way to ignore it:

    $ make
    make[1]: Entering directory `/home/technion/zstd/programs'
    gcc      -I../lib  -O3 -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Wstrict-prototypes   -DZSTD_VERSION=\"r0.0.1\"  ../lib/zstd.c xxhash.c bench.c fileio.c zstdcli.c -o zstd
    ../lib/zstd.c:74: warning: ignoring #pragma gcc diagnostic
    ../lib/zstd.c:110: error: redefinition of typedef ‘BYTE’
    ../lib/fse.c:95: note: previous declaration of ‘BYTE’ was here
    .....
    make[1]: *** [zstd] Error 1 
    make[1]: Leaving directory `/home/technion/zstd/programs'
    make: *** [zstdprograms] Error 2

Cyan4973:
OK, then there is no other way than to change the code.
I'll look into it, and open an issue to track the progresses.

Cyan4973:
As stated within #25 : 

Fix attempt, downloadable at : 
https://github.com/Cyan4973/zstd/tree/streamAPI2

PR #30, opened 2015-03-24
Add `unzstd` program for decompression.

jforberg:
Add **unzstd** symlink to the distribution. Also ensure we use the basename(1) part of the program name, i.e. possible to call the program as ``./unzstd`` or ``/usr/local/bin/zstdcat`` and still get the expected behaviour.


Cyan4973:
Thanks Johan
Sounds like a good idea

PR #32, opened 2015-05-10
ZSTD_getcBlockSize() dropped bits when determining block size

mthiesen:
I am currently experimenting with block sizes larger than 512 kb and noticed crashes when compressed blocks exceeded 512 kb. This was due to the fact that ZSTD_getcBlockSize() only parses the lower 19 bits for the block size whereas it could parse 22 bits (with the top two bits being reserved for the block type).

With this change compressed blocks of up to 4 MB are possible and seem to decompress without any issues.


Cyan4973:
Good point.
Originally, bits 19-21 were reserved for some other use.
As the original need is no longer there, it is tempting to make them useful for block size, although it's also generally good to keep some free room for future extensions.

Note that having larger_than_default block sizes also qualifies as legitimate extension.

Now, there are some reason why I feel it's not necessarily a good idea to authorize very large block sizes : it starts with entropy adaptation.

Within each block, optimal distribution of lengths and distances can be different. Sometimes, drastically different. Sometimes, barely.
Regularly updating these stats is one of the keys to better compression ratio.

With very large blocks, such adaptation is no longer possible : only the "best possible average" will be used, and if statistics vary widely within the block, it's a lost opportunity to compress better.

Note that the opposite is also true, and when statistics are uniform, it's a waste to describe them again. I intend to work on it later this year, to optimize such scenario.

Finding an "optimal block size" is a difficult exercise. In theory, it should vary depending on the source.
In practice though, we just try to find a "good enough" middle ground, and stick with it.

This version of zstd has selected 128 KB as such middle ground, and according to my experience, this value is likely to remain until v1 release.

Now, the format nonetheless allows a bit of room to play with different values. After all, this is an experiment for the time being. Some better alternative can still be found. Up to which point should this possibility be opened ? Up to 4 MB ?

This is, in essence, what your patch propose. I feel it is a bit large, but you could very well have some specific needs to handle too.

For specific needs, I intend to offer a "raw block interface". This interface will not handle headers, hence it will not enforce any block size nor type. Everyone will be free to encapsulate data the way he needs.

For the "frame level interface", which LZ4_getcBlockSize() belongs to, it's a bit different.
Larger block sizes come with increased requirement regarding decoder buffer size, and increased footprint for temporary scratch buffer during compression. Current (limited) frame format is supposed to limit block sizes to 128 KB _also_ for allocation purposes.

So I'm a bit surprised the interface can be used this way. It was unexpected. But I can understand the need to. It's potentially a consequence of the block interface being not yet ready.

I'm currently busy working on xxHash, and intend to work again on zstd during June. Would the availability of a raw block interface answer your need ?

mthiesen:
First of all thank you for your detailed answer and thank you for your work on ZSTD.

I honestly thought that you meant to assign the remaining two bits to the block length as well and simply forgot. Especially as block lengths of up to 512 KB were already working, which is different from the default 128 KB.

Please note that I'm not talking about compression here at all but about decompression. Frankly I don't see how the maximum block size reduces the memory requirement for the decompressor as matches are not limited to the current block. If I read the code correctly the maximum offset should be ```1 << MaxOff``` with ```MaxOff``` being 31 at the moment. At least there is nothing in the decompressor code that enforces the ```g_maxDistance``` limit. There will need to be something of that sort in the future of course, once you support some kind of streaming decompression interface. But I think the choice of dictionary size should be left up to the user (within reasonable limits).

I want to explain a little bit what I am doing and how I discovered this limit:

I was really excited when you announced ZSTD as I always wanted something to beat Deflate in both compression ratio and decompression speed.

However I was a little bit disappointed that there was no HC encoder available and ZSTD could not quite match Deflates compression ratio. Looking at the decompressor I couldn't see any reason why ZSTD should not beat Deflate, especially as it uses FSE instead of Huffman for entropy coding.

Thats why I concocted a frankensteinian creation that I like to call __Deflate 2 ZSTD__. It dissects a Deflate stream into its individual tokens and converts it to a ZSTD stream. The Deflate streams have to be modified just a little bit to achieve this:
* ZSTD expects the last 16 bytes to be literals, Deflate does not care
* ZSTD has a minimum match length of 4 not 3 as Deflate. I convert all matches of 3 to literal runs.
* Deflate has unbounded block sizes, ZSTD has a limit. I split long blocks to make them fit.

To achieve maximum compression I use https://github.com/google/zopfli to create the Deflate streams. Furthermore I use the Selesia Corpus for my test data.

I am very well aware of the block splitting problem. As are the makers of Zopfli, which occasionally produces blocks that are multiple megabytes in size. And this is how I ran into the block size problem.

As it turned out my initial hunch was correct: ZSTD could beat Deflate for all but three files, not by much though. I suspected that the three files (mozilla, ooffice and x-ray) were larger because for those Deflate used a lot of 3 byte matches. I then proceeded to reduce MINMATCH to 3 and saw ZSTD not only beat Deflate on every single file of the Corpus but again improved compression of every single file compared to the MINMATCH 4 version.

This has to be taken with a grain of salt of course. The stream I used was produced by a compressor that thought that 3 byte matches were viable. Maybe the 4 byte match length is an advantage for a compressor that is tailored to ZSTD.

This is a summary of my preliminary results:

| Compressor | Compressed Size |
| ---------------- | ------------------------: |
| ZSTD          | 73.8 MB | 
| Deflate (zopfli) | 65.5 MB |
| Deflate 2 ZSTD | 64.7 MB |
| Deflate 2 ZSTD (MINMATCH 3) | 64.2 MB |

As you can see the results are very promising. One can only dream of what a HC compressor tailored specifically to ZSTD can achieve, one that makes proper use of ```lastOffset```and the larger dictionary size.

As it turns out the block sizes that Zopfli uses are not always advantagous to compression. Most of the time smaller blocks improve compression and in the few cases where larger blocks are advantagous the wins are small (< 1%).

That being said, I am perfectly fine if you refuse this pull request. I don't really need the larger blocks as the possible wins are small at best. On the other hand I don't see the point of arbitrarily restricting the block size as it doesn't reduce the memory requirements for decompression anyway. It's your call.

Keep up the great work, I am excited to see how this project evolves.

Cyan4973:
I'm very impressed by the amount of work and skill you show on this project. Your results are very precious.

Zstd HC is indeed an objective, however it's unlikely to be available soon, as the amount of work needed to complete it is fairly large. In order to reduce this delay, I'm likely to cut the work into smaller pieces and start releasing as soon as something decent is available, keeping a lot of other improvements for the future.

The approach you selected is eye-opening. Indeed, Zopfli already realized an excellent parser for zip, could it be re-used for Zstd ?

Well, for an early release, the answer is yet. The shortcomings are real, but you already identified them. It's possible to live with them. It's even probably possible to progressively modify it to take advantage of Zstd's features.

Among them, the most important one for compression ratio is larger window size. Mechanically, the gain to expect are quite large. The downside is fairly significant too : compression speed is killed fast by searching more and farther. This, in turn, requires developing new techniques to complete some sub-tasks more efficiently. These are the kind of incremental improvements that can be brought progressively.

The second important feature is "rep-code". It make a sensible difference for "regular structures", such as a table of something, with something being "fixed size". It is a quite common setup beyond text files. This a new concept compared to zip, so maybe it could be complex to integrate. But once it's done, there is no other drawback : speed should not be affected (in fact, it can even be improved), and compression ratio is improved a measurable little.

Once that's done, there are the "breaking changes". One of them is to move from 4 to 3 bytes minimum matches, as you suggest.
For this proposition, it requires carefully evaluating pro and con :
Pros : better compression at HC setting
Cons : useless at Fast setting, small increase in compression cost, 
Cons : slower decompression

Probably the most important point is the impact on decompression speed. If it's small, then it's a fine proposition. If it's large, then it's a problem, it goes against the trade-off selected by zstd.

Note that increased compression from zstd is expected to come from increased window size, rather than smaller matches. But that's nonetheless the kind of choices which are worthwhile to evaluate during this first year. Once it's settled, it will be difficult if not impossible to change such choice. So better carefully evaluate it now.


> That being said, I am perfectly fine if you refuse this pull request. I don't really need the larger blocks as the possible wins are small at best. 

I understand the point, and yes, I agree in some circumstances, there is a small gain in increasing block size. Most likely, the gains come from using a single header instead of multiple ones to describe the statistics. It's only really worthwhile when the statistics don't change much between consecutive blocks. And the gain are small because statistics costs are typically around 200 bytes range.

Nonetheless, I can increase the maximum size of blocks.
Just I wish I could keep at least one bit in reserve, for potential future expansion.
Hence, the number of bits for block size would be expanded from 19 to 21 bits.
This would allow maximum block sizes of __2MB -1 bytes__.
Note the __-1__, it's important : a non compressible block of exactly 2 MB would need the 22nd byte for a correct description. Hence, if the maximum block size must be a power of 2, the maximum allowed will be 1 MB.

Would that help for your use case ?

> Frankly I don't see how the maximum block size reduces the memory requirement for the decompressor as matches are not limited to the current block.

You are correct.
The most important value is the window size.
Knowing this value, the "buffer manager" will have to create a buffer large enough to include, at most "WindowSize + MaxBlockSize - 1" bytes. In general, WindowSize is larger than MaxBlockSize, so it dominates the equation. Moreover, if there is some kind of implicit guarantee that "WindowSize == N x MaxBlockSize" *exactly*, then the extra room is not really necessary, and bufferSize == WindowSize.

Nonetheless, with large MaxBlockSizes, and without guarantee to precise sizes, the first equation must be used ("WindowSize + MaxBlockSize - 1" bytes), and MaxBlockSize could well dominate, or at least take a sizable chunk of it.

It's up to the buffer manager to decide that.

Currently, the only frame format available is fairly limited : WindowSize can only be 512 KB, maxBlockSize can only be 128 KB. Hence the behavior you observe.

But it's just early days. In the future, these values will be parameters, and will be able to change.


> I think the choice of dictionary size should be left up to the user (within reasonable limits).

Yes correct.


> If I read the code correctly the maximum offset should be 1 << MaxOff with MaxOff being 31 at the moment

It's early days for such limitations, but just for the sake of completion,
the current maximum limit is __64 MB - 1__, 
because the 32-bits version of FSE can only guarantee reading up to 25 bits in a single row.

I expect to keep this limitation for the foreseeable future, since there's already plenty of job to do with 64 MB. But if it becomes a problem, this limitation will be updated too. There should be no problem to use larger offset values. The code will just have to adapt to it.

mthiesen:
Thank you for your answer and for your explanations. You have clarified a lot of things, I understand your choices better now.

I suspected that the choice of 4 bytes minimum match length was due to decompression speed. I am also wondering how decompression speed will decrease once the decompressor is properly secured against buffer overruns. It's not exactly fair to compare ZSTD and zlib in its current state.

I close this pull request. I no longer think it's worth the trouble. Especially because the larger block sizes don't have a meaningful impact on compressed size.

I will of course continue experimenting with ZSTD. I don't really need bigger blocks for this.

PR #36, opened 2015-06-26
+ Squash Benchmark URL

tomByrer:


Cyan4973:
Yes, I plan to link it, when squash-benchmark is ready for general usage.
Which in my definition means there's a release, an installable package, that everyone can download and install to test on its machine.
It's a complex task, so let's wait for the author to get it done.

The reason is there are a few strange results which I suspect could be related to how the API is used, rather than the underlying algorithm itself. But without a testable prototype, it's difficult to check that and pinpoint where potentials problems may lie (if there are).

tomByrer:
Thanks, does look like [you 2 are friends](https://github.com/quixdb/squash-benchmark/issues/17).
Do you think Squash (unified API) itself is good?

(You can close if you don't want this issue reminder.)

Cyan4973:
From a "general" perspective, I'm not sure. It looks difficult to have a unified API for multiple algorithms with vastly different capabilities. This was attempted and failed before on several occasions.

But nonetheless, it seems an interesting objective, worthwhile to try, so I wish Evan all the success. More importantly for our topic, it's a founding block of Squash benchmark, so even if it would serve only this purpose, it's useful enough.

PR #46, opened 2015-10-10
Remove one malloc/free from compression

xaphier:
By making the buffer part of the zstd context structure, one malloc/free
can be removed from the compression.


Cyan4973:
Good catch Xavier.
This was supposed to be introduced into next version of zstd, but since your patch is ready, let's use it now ;)

PR #56, opened 2015-10-30
HC modes

Cyan4973:


README.md
```diff
@@ -7,42 +7,40 @@ It is provided as a BSD-license package, hosted on Github.
...
+|Name             | Ratio | C.speed | D.speed |
+|-----------------|-------|--------:|--------:|
+|                 |       |   MB/s  |  MB/s   |
+| brotli -9       | 3.729 |     4   |   340   |
+| **zstd 0.3 -9** |**3.447**|**30** | **640** |
+| [zlib 1.2.8] -9 | 3.133 |    10   |   300   | 
+| LZO 2.06 -999   | 2.790 |     1   |   560   |
+| [LZ4] r131 -9   | 2.720 |    25   |  2100   |
 
+[lzma]:http://www.7-zip.org/
```
tobijdc: Link to lzma given, but not mentioned anywhere

Cyan4973: Good point.
Well, the table for "high compression" scenario is a mess, did it too fast, questionable selection of compression levels. I should probably scrap it and build a more complete new one.

tomByrer:
Curious why you did 23 levels, & not 5 or 9 like some others?  Most people use only 3-4 of the 5-9 levels anyhow.  The only time a wide range is really helpful is lossless compression, like JPEG, where you can sometimes tolerate losing info.

tobijdc:
I think it is a good first table, but it can be improved

Cyan4973:
@tomByrer  : 
Well, first it's an experiment, it may prove a bad idea after all.

Second, I was asked to work more on the "high speed" part, because that's the one reason to use Zstandard today. Initially, my first table was only 9 levels, but it provided too little choices for the "high speed" variants. So I spaced the levels more closely to each other. In the end, it required 25 levels to cover the same range as the initial 9-levels table.

Third, my focus is less on human selecting manually a level than on automatic processes trying to find some optimal setting for their configuration (and eventually updating them "live"). More fine-tuned levels give more opportunity to match speed/compression capacity.

Arguably, today compression levels beyond 10 are not interesting yet, they are mostly there for me to experiment. I would suggest to stick with lower compression levels for now (1-9), to get an interesting speed / compression trade off.

tomByrer:
> automatic processes trying to find some optimal setting for their configuration (and eventually updating them "live")

Oh, that is interesting!
Perhaps 'levels' would be one way do indicate compression attempts.  But I'm more interested in either setting a time limit (compressing live or a huge batch job that I need to email in 1 minute), or seeing how much compression is possible without regards to time (sometimes do this with PNG, using highest zopfli, then 3 more PNG optimizers, can take a few minutes per image).

So really setting time limit is more interesting for me to configure, but knowing how many rounds of compression completed within that time can be a helpful metric.

PR #67, opened 2015-11-06
Dev

Cyan4973:
Faster fast clevels


tomByrer:
I see by the graph a more smoother downward slope, so now it looks less like a ski jump.  Good job!

PR #84, opened 2015-12-01
Fix compression on Big Endian CPUs

peterh:
This fixes a typo that leads to incorrect compressed data on Big Endian CPUs.

### DIFF
--- a/lib/zstd_compress.c
+++ b/lib/zstd_compress.c
@@ -620,7 +620,7 @@ static unsigned ZSTD_NbCommonBytes (register size_t val)
     }
     else   /* Big Endian CPU */
     {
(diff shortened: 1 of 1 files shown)


Cyan4973:
Indeed, serious problem and fairly obvious fix. Thanks @peterh !

minor downside : the fix should be redirected to "dev" branch, to follow branch update policy.
I can take care of it.

On a related note : 
Big-endian targets are not automatically tested within CI suite.
I can test it, but only manually, and it takes quite a lot of time. Consequently, it's not run regularly.

It would be great if it was possible to run even basics tests automatically at each update for big endian targets.

Cyan4973:
moved to #85

PR #90, opened 2015-12-07
Add __declspec(dllexport) via optional ZSTDLIB_API macro when building a DLL for Windows

KrzysFR:
I'm building a .NET wrapper for Zstandard, and for that I need to build zstd as a windows DLL with a `__declspec(dllexport)` prefix on all the public methods defined in the header files. This allows me to use `LoadLibrary` and `GetProcAddress` at runtime to bind to the native library.

I looked at other projects and how they were adressing this, and used the same pattern as, for example, SQLite which defines a `SQLITE_API` prefix. Other libraries also have a similar LIBFOO_API.

This PR is about adding a `ZSTDLIB_API` macro as a prefix to all public methods in the `zstd_*.h` headers, which is by default not defined. Only when building with `ZSTD_DLL_EXPORT=1` then this macro would be defined as `__declspec(dllexport)` on Win32.

Questions:
- Are the names `ZSTDLIB_API` and `ZSTD_DLL_EXPORT` OK to you?
- How do you document these new compile options? I'm using a custom vcproj for my lib, so I may have missed things in the Makefile
- I'm not exactly sure what to use for non-Win32 platforms. I have currently used `extern` but this may not be the proper keyword? If this needed at all on linux?
- I currently have only added a `ZSTDLIB_API` for `zstd*.h`. I'm not directly using fse or huff0, but maybe a similar technique could be used with a `FSE_API` and `HUFF0_LIB` prefix?

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -43,6 +43,24 @@ extern "C" {
 #include <stddef.h>   /* size_t */
 
 
+/* ***************************************************************
+*  Tuning parameters
+*****************************************************************/
+/*!
+*  ZSTD_DLL_EXPORT :
+*  Enable exporting of functions when building a DLL
+*/
+#if defined(ZSTD_DLL_EXPORT) && (ZSTD_DLL_EXPORT==1)
+#  if _WIN32
+#    define ZSTDLIB_API __declspec(dllexport)
+#  else
+#    define ZSTDLIB_API extern
+#  endif
+#else
+#  define ZSTDLIB_API
+#endif
+
+
 /* *************************************
 *  Version
 ***************************************/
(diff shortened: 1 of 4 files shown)


KrzysFR:
The CI build failed with gcc because it seems that `zstd_buffered.h` does not include `zstd.h`. This did not fail in my VisualStudio project for some reason.

I tried duplicating the definition of ZSTDLIB_API in both files, but I'm not sure how to best solve this, without introducting a common.h that would be included by everyone...

KrzysFR:
aaand I created the PR against master instead of dev >_<

I've reopened a new PR in #91

PR #91, opened 2015-12-07
Add __declspec(dllexport) via optional ZSTDLIB_API macro when building a DLL for Windows

KrzysFR:
I'm building a .NET wrapper for Zstandard, and for that I need to build zstd as a windows DLL with a `__declspec(dllexport)` prefix on all the public methods defined in the header files. This allows me to use `LoadLibrary` and `GetProcAddress` at runtime to bind to the native library.

I looked at other projects and how they were adressing this, and used the same pattern as, for example, SQLite which defines a `SQLITE_API` prefix. Other libraries also have a similar LIBFOO_API.

This PR is about adding a `ZSTDLIB_API` macro as a prefix to all public methods in the `zstd_*.h` headers, which is by default not defined. Only when building with `ZSTD_DLL_EXPORT=1` then this macro would be defined as `__declspec(dllexport)` on Win32.

Questions:
- Are the names `ZSTDLIB_API` and `ZSTD_DLL_EXPORT` OK to you?
- How do you document these new compile options? I'm using a custom vcproj for my lib, so I may have missed things in the Makefile
- ~~I'm not exactly sure what to use for non-Win32 platforms. I have currently used `extern` but this may not be the proper keyword? If this needed at all on linux?~~ (not needed on linux)
- I currently have only added a `ZSTDLIB_API` for `zstd*.h`. I'm not directly using fse or huff0, but maybe a similar technique could be used with a `FSE_API` and `HUFF0_LIB` prefix?

note: The CI build initially failed with gcc because it seems that `zstd_buffered.h` does not include `zstd.h`. This did not fail in my VisualStudio project for some reason.

I tried duplicating the definition of ZSTDLIB_API in both files, but I'm not sure how to best solve this, without introducting a common.h that would be included by everyone...

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -43,6 +43,24 @@ extern "C" {
 #include <stddef.h>   /* size_t */
 
 
+/* ***************************************************************
+*  Tuning parameters
+*****************************************************************/
+/*!
+*  ZSTD_DLL_EXPORT :
+*  Enable exporting of functions when building a DLL
+*/
+#if defined(ZSTD_DLL_EXPORT) && (ZSTD_DLL_EXPORT==1)
+#  if _WIN32
+#    define ZSTDLIB_API __declspec(dllexport)
+#  else
+#    define ZSTDLIB_API extern
+#  endif
+#else
+#  define ZSTDLIB_API
+#endif
+
+
 /* *************************************
 *  Version
 ***************************************/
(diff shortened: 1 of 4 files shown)


Cyan4973:
Wow, it seems my earlier comment to this PR has just totally disappeared ... 8O

OK, let's try to summarize here what I already wrote.

> Are the names `ZSTDLIB_API` and `ZSTD_DLL_EXPORT` OK to you ?

I guess so. The name must help the reader to understand intention.
Regarding `ZSTD_DLL_EXPORT`, since it seems specific to windows, maybe some kind of `WIN` somewhere in the name would help.

> How do you document these new compile options?

No idea. It's probably better to enable them automatically, typically on detecting Visual Studio environment, especially if it doesn't harm outside of DLL compilation context.

> I'm not exactly sure what to use for non-Win32 platforms. 

As far as I know, the convention in C is that any symbol not explicitly private with a `static` keyword is considered public, hence it can be invoked when compiled within a DLL.
Well, apparently, .NET has additional requirements ....

>  I'm not directly using fse or huff0, but maybe a similar technique could be used with a FSE_API and HUFF0_LIB prefix?

Huff0 and FSE are private API used within Zstd. I believe they should not be exposed.

KrzysFR:
> Regarding `ZSTD_DLL_EXPORT`, since it seems specific to windows, maybe some kind of WIN somewhere in the name would help.

I believe `dllexport` and `dllimport` are already windows specific, so maybe the current name is enough? Linux developpers will probably not look for it.

> No idea. It's probably better to enable them automatically, typically on detecting Visual Studio environment, especially if it doesn't harm outside of DLL compilation context.

Sorry, I meant how exactly do you want new options documented in the zstdh.h file ? I added a `Tunning options` banner and use the `/*! ... */` comment. See https://github.com/Cyan4973/zstd/pull/91/files#diff-a7634981d7010e4420d7ab0274a30503R50 

For VS itself, I simply have a custom vcproj with only the things I need and I set the ZSTD_DLL_EXPORT variable to 1. People may want to use VS to build a non-dll version or cli, so I think it should be opt-in.

> As far as I know, the convention in C is that any symbol not explicitly private with a static keyword is considered public, hence it can be invoked when compiled within a DLL.
Apparently, .NET has additional requirements ....

This is not .NET per se, only the way `LoadLibrary`/`GetProcAddress` seems to work. Since it seems that only VS+win32 requires the `__declspec(dllexport)` keyword, I can simplify the #ifs to only deal with win32. For all other platforms, the `ZSTDLIB_API` would be a no-op.

I will update the PR with the changes shortly.

Also, do you have any idea how I could solve the issue with `zstdb_buffered.h` not including `zstd.h`, forcing me to duplicate the #define of ZSTDLIB_API in both files?

KrzysFR:
I'm having some issues making it build with gcc. Once the CI passes, I'll squash the commits to cleanup the PR...

Cyan4973:
Looks good to me

PR #93, opened 2015-12-08
Move tests from zstd-playTests target to separate shell script.

annulen:
This patch allows me to run tests on MIPS board which lacks make.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -169,7 +169,7 @@ test32: test-zstd32 test-fullbench32 test-fuzzer32 test-zbuff32
 test-all: test test32 valgrindTest
 
 zstd-playTests: datagen
-	ZSTD=$(ZSTD) ./playTests.sh
+	ZSTD=$(ZSTD) ./playTests.sh --test-large-data
 
 test-zstd: ZSTD = ./zstd
 test-zstd: zstd zstd-playTests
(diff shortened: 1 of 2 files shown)


Cyan4973:
make-less platforms ? a new area for me ... ;)

The only minor issue is that I was considering to cut this test script into 2 parts,
one for "quick essential tests",
and the other one for longer tests, such as checking each compression level and very large file support.

Maybe this modification should be done first, so it's easier to see how to adapt it to make-less platforms ?

annulen:
Well, I can build make for my platform, I just don't have it readily available, as I've never needed to have it on target before.
Also, I believe putting non-trivial shell constructs into makefile is extremely bad practice, all those $$ escapes and prohibition of empty lines hurt readability a lot

As an alternative to cutting script, we could add command line parameters to script

Cyan4973:
Yes, it could be done this way

annulen:
Which round-trip tests do you consider non-essential?

Cyan4973:
Everything below and including 
`@./datagen -g270000000 | md5sum > tmp1`

annulen:
Done

Cyan4973:
Almost good !

For some reason, some of the tests are not present within `playTests.sh` : compression levels > 12 and huge file test are not in this list.

That being said, I can add them manually, the script design is simple enough

annulen:
My bad

Cyan4973:
All good !

PR #95, opened 2015-12-10
Add a Visual Studio project for building zstdlib.dll on Windows

KrzysFR:
This PR adds a new project to the Visual Studio solution, in order to build a Windows dll that exports all methods marked with the ZSTDLIB_API prefix. This library can be used directly by wrappers written in other languages (like .NET in my case).

The commit adds or changes:
- add exclusions to .gitignore to remove all the cruft left by VS when it builds the project.
- updates the .sln so that it loads with last version of VS installed (VS 2015)
- add stdlib.vcproj that builds `zstdlib_x64.dll` and `zstdlib_x86.dll`, configured with `ZSTD_DLL_EXPORT=1` and `ZSTD_LEGACY_SUPPORT=0` by default (no legacy support)
- changes the Platform Toolset of all projects to Visual Studio 2013 (v120) which may be controversial
- misc cleanup of others vcproj (include paths, stacksize 25000 vs 19000)

notes:
- I also use `ZSTD_HEAP_MODE=0` but there does not seem to be any code that use this #define anymore?
- The Appveryor CI server builds using MSBuild v14.0 anyway (VS2015) so changing the version of the sln and maybe platform toolset should not break the build

__VC++ Runtime version__

The most controversial change may be the swtich to VS 2013 toolset, which requires installing the VC++ Runtime 12.0 (instead of 11.0 previously). I'm not sure if people currently using the .sln are all targetting v110 (which is now somewhat old), or if everyone is changing the settings or simply including the files in their own projects? We build and redist for v120 so that's why I had to set this parameter...

__Library file name__

Another point may be the name of the dll produced. Currently it is `zstdlib_{ARCH}.dll` (`zstdlib_x64.dll` and `zstdlib_x86.dll`) because it helps solve some issues with having .NET applications compiled with `AnyCPU`, and which must load a runtime the appropriate library (some binary can run in 32 or 64 bit depending on the OS and JIT). Another solution is have the same name for all variants, but in different subfolders (so `./x86/zstdlib.dll` and `./x64/zstdlib.dll`).


Cyan4973:
A pretty solid PR as usual @KrzysFR .

> changes the Platform Toolset of all projects to Visual Studio 2013 (v120) which may be controversial

As we are nearing 2016, I believe the change to VS2013 runtime v120  isn't excessive. So it's okay.

>  I'm not sure if people currently using the .sln are all targetting v110 (which is now somewhat old), or if everyone is changing the settings or simply including the files in their own projects?

The initial assumption was that programmers would directly include files into their own project. A Visual solution was only provided as a shortcut and example. However, the assumption could prove wrong, especially as the number of files to include is fairly large, hence not obvious.

> I also use ZSTD_HEAP_MODE=0 but there does not seem to be any code that use this #define anymore?

Correct. Heap mode is one of those minor design elements that were left unfinished. I'll have to look into it later on.

> Library file name

The name you selected (`zstdlib_x64.dll` and `zstdlib_x86.dll`) look good to me.

PR #99, opened 2015-12-18
Update version to 0.4.5 in zstdlib resources

KrzysFR:
I forgot to say in #85 that, in order for the windows dll to have the proper file versions attributes (visible in the Details tab of the dll properties in explorer), you need to specify both `FILEVERSION` and `PRODUCTVERSION` attributes in the `VERSIONINFO` structure must be updated in the project resources (in `zstdlib.rc`). 

image

These come from:
image

Also, you are currently updating the `ZSTD_VERSION_RELEASE` defines in `zstd.h` from within the `Makefile` which will not be seen by Visual Studio, meaning that `ZSTD_versionNumber()` will still return 0.4.3 currently.

You can always edit them by hand (from VS), but it is easy to forget before publishing a new version.

I haven't found any elegant solution from within VS to update both the `zstd.h` and `zstd.rc`- using the export line in `Makefile` as a source - at build time.

Do you know if it is possible to tweak your Makefile to patch both `lib/zstd.h` and `visual/2013/zstdlib/zstdlib.rc`, and commit the changes everytime you publish a new version? Then anyone pulling the new version and rebuilding with VS would have the correct versions.


Cyan4973:
Mmmh, that's not good.
We need to have this identifier respect the [single point of truth](https://en.wikipedia.org/wiki/Single_Source_of_Truth).

In theory, the source value should be within `zstd.h`.
Everything else should relate to it.

This has been achieved within `lib/Makefile`.
`sed` scripts extract values from `zstd.h`, which can then be assembled to create versioned library names

This script version unfortunately doesn't work outside of `lib/Makefile` because the `sed` scripts remain "as is" until their evaluation in a command, and many commands wouldn't accept that. Fortunately, I recently found a way to force their evaluation. This would allow to spread the modified method to other Makefiles.

It still leaves `cmake` and visual projects out of the loop.
I'm not sure how to proceed, but indeed, it's important that they follow the same principle : the single point of truth is within `zstd.h`.

As a good news, note that `zstdcli.c` can already learn its version number from `zstd.h`, so if no `ZSTD_VERSION` is specified, it will find the right version by itself. This only impacts the command line utility though.

KrzysFR:
I did not merge from the tip of the dev branch when I did the original PR, so the problem is in fact only with `zstdlib.rc`. Also, I see that I did push another commit to my dev branch to fix a missing `ZSTDLIB_API` that was also included in the PR. Sorry about that.

Cyan4973:
Is this patch ready for merge, or should another one be submitted ?

KrzysFR:
It should be okay to merge, though it changes two things: the version in zstdlib.rc, and adds missing ZSTDLIB_API prefix for two new methods in zstdlib.h.

PR #103, opened 2015-12-23
Updated gitignore and added vs 2015 project

smasherprog:


smasherprog:
Wrong branch.. Will redo

PR #105, opened 2016-01-07
CMake build (Linux and Windows MSVC tested only) integration support …

mailagentrus:
CMake build (Linux and Windows MSVC tested only) integration support added;


mailagentrus:
You can reopen issue: https://github.com/Cyan4973/zstd/issues/66

Cyan4973:
Thanks @mailagentrus, it's an excellent PR.
The content of patch looks very good to me, so we just have to discuss a small detail :

- I worry about having a `CMakeLists.txt` files within the same directory as `Makefile`.
Anyone like me who would blindly tape `cmake .` ends up with `Makefile` overwritten in the process.
For this reason, I tend to prefer having a separate directory for `CMakeLists.txt`.

Since you have 2 such files, plus a few dependency files, it could prove non-trivial to do.
Does it make sense to regroup them into a dedicated directory ? 
Does it require to merge both `CMakeLists.txt` first ?

mailagentrus:
Hmm, I think, that it's possible to reorganize CMakeLists.txt file structure to build lib and programs in specified output directory. Now we have next directory hierarchy:
root
|
..lib (with Makefile)
..programs (with Makefile)
..Makefile

What hierarchy do you prefer to split Makefiles and CMakeLists? :) I'll fix that at this weekends..

Cyan4973:
I would like to suggest :  

root
|
lib (with Makefile)
programs (with Makefile)
contrib / cmake (with CMakeLists.txt and dependencies)

The "contrib" emphasizes that the maintainer of `CMakeLists.txt` isn't me, since I have no skill in CMake files. In which case, it would also be appropriate for your name to be provided as the author of the files.

mailagentrus:
Added contrib folder with CMakeFiles. Added /analyze options for MSVC projects, set O3 by default optimization flag in Release mode...

Cyan4973:
minor typo : you named the directory `contib` instead of `contrib`.

mailagentrus:
You are right, sorry. Typo is fixed.

mailagentrus:
New line before EOF added (for each file in PR)...

t-mat:
Thank you @mailagentrus !  These changes will help future commits.

mailagentrus:
Thank you @t-mat ! Yes, for example, user can build project with MSVC 2015 with these changes (automatic conversion MSVC solution file from 2013 to 2015 version was broken for me)...

Cyan4973:
Looks like a great PR to merge,
just ... well, if you look at directory name, although the change made it closer, it has not yet reached `contrib` yet ....

mailagentrus:
Contrib directory name successfully changed...

Cyan4973:
Thanks @mailagentrus, it's a great PR.

I tested CMake generation of Makefile on Linux (Mint 17 x64) using gcc 4.8.4,
compilation fails on `paramgrill` target, which is a test tool.

It seems the linking stage fails, looking for symbol `log`.
So I suspect link instruction to math library `-lm` is missing.

mailagentrus:
Will check this issue and fix tomorrow morning... I think so, look like missed -lm linker flag for paramgrill tool.

PR #107, opened 2016-01-10
Paramgrill math library

mailagentrus:
Please, test PR cmake build for paramgrill target with gcc 4.8.4. I have only 5.3 and 4.9.3 on my archlinux...

### DIFF
--- a/contrib/cmake/programs/CMakeLists.txt
+++ b/contrib/cmake/programs/CMakeLists.txt
@@ -67,8 +67,7 @@ IF (UNIX)
     TARGET_LINK_LIBRARIES(zbufftest libzstd_static)
 
     ADD_EXECUTABLE(paramgrill ${PROGRAMS_DIR}/datagen.c ${PROGRAMS_DIR}/xxhash.c ${PROGRAMS_DIR}/paramgrill.c)
-    TARGET_LINK_LIBRARIES(paramgrill libzstd_static)
-    SET_TARGET_PROPERTIES(paramgrill PROPERTIES LINK_FLAGS "-lm")
+    TARGET_LINK_LIBRARIES(paramgrill libzstd_static m) #m is math library
 
     ADD_EXECUTABLE(datagen ${PROGRAMS_DIR}/datagen.c ${PROGRAMS_DIR}/datagencli.c)
     TARGET_LINK_LIBRARIES(datagen libzstd_static)


Cyan4973:
Seems to work fine !

Cyan4973:
Just added : CI tests to ensure cmake script remain valid in the future

Cyan4973:
minor but annoying issue : 

the `cmake` script requires CMake 2.8.12 or higher,
but Travis CI only provides CMake 2.8.7.

https://travis-ci.org/Cyan4973/zstd/jobs/101451216

Is there anything in the script which requires 2.8.12+ and would not work with 2.8.7 ?

mailagentrus:
No, it is copy and paste 2.8.12 as minimal version from my other projects :) I will check cmake building project with 2.8.7 version. Is it possible to upgrade Travis CI cmake version?

Cyan4973:
Could be, this is something to look for and test since it could be the repo used by Travis has 2.8.7 as latest version. Travis is not exactly "up to date " : `gcc` is 4.6, `clang` is 3.4, to name a few.

But my main concern is not just about Travis. This condition increases pressure on cmake minimum version requirement, and maybe it's not necessary. This situation, with a user having a good enough cmake which doesn't satisfy manually-specified minimum version requirement, will reproduce in the future.

So better have a lower version requirement if it works.

As a sidenote, considering that `CMAKE_MINIMUM_REQUIRED` is a mandatory statement within `CMakeLists.txt`, and that it can be a serious headache to know which statement depends on which minimum version, I'm a bit surprised there is no tool to select the right minimum version automatically.

Cyan4973:
For your information : 

I tried to arbitrarily decrease `CMAKE_MINIMUM_REQUIRED` to `2.8.7` within latest "dev" update.
It doesn't work : 
https://travis-ci.org/Cyan4973/zstd/jobs/101567333

`Unknown CMake command "TARGET_INCLUDE_DIRECTORIES".`

It's really difficult to find minimum version required on the cmake.org website, which is a pity. Fortunately, googling finds [a few people with same issue](http://stackoverflow.com/questions/20643368/cmake-different-include-directories-for-different-targets/20643487#20643487).

Bottom line : `2.8.12` seems necessary to support this keyword.

Which leaves us with the following remaining options : 
- Try to update cmake on Travis
- Give up CI tests for cmake
- Do not use `TARGET_INCLUDE_DIRECTORIES` within `CMakeLists.txt`

mailagentrus:
Ok, i will try to find solution.

PR #108, opened 2016-01-11
travis and CMake 2.8.8 installation test;

mailagentrus:


mailagentrus:
Hmm, look like final commit added CMake 2.8.8 support for travis...

mailagentrus:
I think, that the best way to integrate this changes is merge all tests into last one commit. Tests before last commit failed...

Cyan4973:
This PR looks great imho !

I guess I only have a few minor questions or comments, which doesn't jeopardize this PR.

- It's a minor point : this commit adds a few more actions to be done `before_install` and `install`, which are applied to all tests, irrespective of the fact they are needed or not.
  It increases the workload of all tests.
  Note that it's not specific to `cmake`. `valgrind` is also installed for all tests while being useful to only one of them.
  It would be good to find a way to restrict preparation actions to what is necessary just for the current test.

- Travis already had cmake 2.8.7. This PR improves the version to 2.8.8.
  What is specific to 2.8.8 that couldn't have been done with 2.8.7 ?

- Now that you created a method to update cmake to some more recent version, why 2.8.8 and not 2.8.12 (or newer) while at it ?

mailagentrus:
Oh, it was not final commit :)) Last commit only add Cmake 2.8.8 integration support... I plan to downgrade minimal CMake version to 2.8.7

mailagentrus:
I use INCLUDE_DIRECTORIES for more than one target in one project (please, see matrix: https://cmake.org/Wiki/CMake_Version_Compatibility_Matrix/Properties_on_Targets)

Imho, I can add specific code to support CMake 2.8.7 "from the box" (in travis for example).

mailagentrus:
COMPARE_VERSION_STRINGS is used to compare versions. Cmake can compare version, for example: try to compare 2.8.8 and 3.2. Which version is greate, which is less? So, this macros was temporary solution to compare minimal support version (to define USE_DEPRECATED_CODE_STYLE).

I plan to compare ${CMAKE_MAJOR_VERSION}, ${CMAKE_MINOR_VERSION} and etc instead of COMPARE_VERSION_STRINGS.

Cyan4973:
Thanks for these very detailed answers @mailagentrus .

I'm supposed to publish a release soon. Should I wait for another commit from you ?

PR #115, opened 2016-01-30
Added VS2013 project for dictBuilder

KrzysFR:
Added a .vcxproj for the new `dictBuilder` to the VS2013 solution.

I had to fiddle with a few things to make it build, with my limited knowledge of C++ projects in VS:
- included `string.h` instead of `strings.h`
- copy/pasted a bunch of defines to make VS shut up about `INLINE` / `FORCE_INLINE` from other files.
- disabled "treat warning as errors" because there are a bunch of warnings in VS that I do not understand

With these hacks, I can produce a working dictBuilder.exe that seems to output the same dictionary as the version I build previously on linux.

Also note I added the new `error_private.h` and `error_public.h` to the `zstdlib.vcxproj`, for completness (not required but easier to load these files from the IDE if they are included)

### DIFF
--- a/dictBuilder/divsufsort_private.h
+++ b/dictBuilder/divsufsort_private.h
@@ -50,7 +50,11 @@ extern "C" {
 # include <stddef.h>
 #endif
 #if HAVE_STRINGS_H
-# include <strings.h>
+# ifdef _WIN32
+#  include <string.h>
+# else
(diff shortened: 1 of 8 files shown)


Cyan4973:
Looks all good.
I'll take care of the Visual warnings in a later update.

PR #136, opened 2016-02-25
merging mml3 mode 

Cyan4973:
as suggested by @inikep


Cyan4973:
@inikep : there are a number of issues when trying to merge your dev branch,
likely related to the removal of `zstd_opt` strategy and its ripple effects on the rest of the code base.

See : https://travis-ci.org/Cyan4973/zstd/builds/111780392

You might want to fix that on your branch, or let me do it after the merge, as you prefer.

inikep:
I fixed problems with paramgrill:
https://github.com/inikep/zstd/commit/59493e8669832b56624c1e1f844d5915a64bc147

Please look at these errors, I'm not sure if it's related with my changes:
Starting zstd tester (64-bits, 0.6.0)
Seed = 4916
Error detected in Unit tests !

Cyan4973:
`./fuzzer -v`

result : 

`test  8 : compress with duplicated context : Error detected in Unit tests !`

Likely explanation : 
some new variables were probably added into compression context,
these new variables are probably not taken care of within existing `ZSTD_copyCCtx()` function.

inikep:
I've found a bug in ZSTD_copyCCtx() but there is another problem. The error is reported here:
        /* offset beyond prefix */
        if (sequence.offset > (size_t)(oLitEnd - vBase))
            return ERROR(corruption_detected);
Do you have any idea what can be a problem?

Cyan4973:
` offset beyond prefix`

Decoder received an offset which is too large (goes "beyond prefix"). So it aborts operation to avoid reading outside its buffer.

This could be a case of the decoder badly decoding offset, or the encoder providing a bad offset value.

inikep:
The problem is ZSTD_decompressBlock - Block functions produce and decode raw zstd blocks, without frame metadata.
There is no place to signal MINMATCH=3 or MINMATCH=4 as ZSTD_decompressBegin does not have compressionLevel.
What are your thoughts on this?

inikep:
There is one more bug found by Travis. So far I've found it's not related to a new changes.
When you change zstd v0.5.1 in zstd_opt.h:126 from:
    return ((litLength<<3) + ZSTD_highbit((U32)matchLength+1) + Offbits + ZSTD_highbit((U32)offset+1);
to:
    return (litLength<<3) + ZSTD_highbit((U32)matchLength+1) + ZSTD_highbit((U32)offset+1);
then fuzzer.exe generates the same error:
Seed = 7536
 11646      Error => streaming decompressed data corrupted : byte 2617994 / 3810755  (31!=30) (seed 7536, test nb 11646)

Cyan4973:
> The problem is ZSTD_decompressBlock - Block functions produce and decode raw zstd blocks, without frame metadata. There is no place to signal MINMATCH=3 or MINMATCH=4 as ZSTD_decompressBegin does not have compressionLevel.

Indeed, it's a difficult point.
`ZSTD_decompressBegin()` is a generic context initializer, which is supposed to learn frame parameters from frame header. There's no point to force any parameter, it's not up to the decoder to decide if the frame use mml3 or mml4, it's up to the frame to tell it.

But in the case of block functions, there is by definition no frame, hence no header to read and know this information.

A few ideas to move forward : 
- Add a new method `ZSTD_decompressBlocksBegin()` where mml3/4 can be selected.
  + Pros : mml3/4 selection becomes possible
  + Cons : API becomes a bit more complex.
                 Caller needs to be aware of which proper setting is required to properly decode the block.
                 This is not trivial : mml3/4 is hidden behind "compression level", so the caller might be unaware of the proper setting, and require either look into `zstd_compress.c` source code, or access `_advanced()` functions within `zstd_static.h`, or have some trial & error process. None of these perspectives look acceptable.

- Forbid mml3 in block mode. Automatically change it into mml4.
  + Pros : Fairly simple solution, low impact
  + Cons : obviously, no mml3 available for block mode. Could be detrimental for small blocks (which is the primary scenario for block compression mode, mostly intended to save Frame header space).

- move mml3/4 into block header.
  + Pros : automatic and compatible with block mode. Could allow later optimizations where mml3/4 is selected per block (though gain perspective is pretty low).
  + Cons : repeated information. Increase header cost (once per block vs once per frame)

inikep:
Current version seems to be ready to merge.

Cyan4973:
:+1:

PR #144, opened 2016-03-10
removing of HashLog3 parameter

inikep:
removing of HashLog3 parameter


Cyan4973:
`clang` complains of a small issue with `paramgrill` : 

https://travis-ci.org/Cyan4973/zstd/jobs/115071895

PR #146, opened 2016-03-14
propagate errors when decompressing; also unlink dstfile if we create

chipturner:
… it but fail to successfully compress or decompress

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -428,6 +428,10 @@ static int FIO_compressFilename_extRess(cRess_t ress,
 
     result = FIO_compressFilename_internal(ress, dstFileName, srcFileName, cLevel);
 
+    if (result != 0) {
+      unlink(dstFileName);
+    }
+
     fclose(ress.srcFile);   /* no pb to expect : only reading */
     if (fclose(ress.dstFile)) EXM_THROW(28, "Write error : cannot properly close %s", dstFileName);
     return result;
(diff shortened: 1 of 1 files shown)


Cyan4973:
Minor error from Visual : 

```
warning C4996: 'unlink': The POSIX name for this item is deprecated. Instead, use the ISO C++ conformant name: _unlink. See online help for details. [C:\projects\zstd\visual\2013\zstd\zstd.vcxproj]
```

which gives : 
https://msdn.microsoft.com/en-us/en-en/library/ttcz0bys(v=vs.110).aspx

Cyan4973:
minor note :

There is a subtlety with gzip decompression : 
using `gzip -dcf file` (or `zcat -f file` which is equivalent)
on a file _not_ compressed by `gzip`,
`gzip` is supposed to output verbatim the content of `file` instead of failing decompression.

I was told this was done to avoid breaking pipes containing both gzip and non-gzip content.

This is not yet supported by `zstd`, but when it will,
it will also be necessary to output an error code of zero for this specific case.

Cyan4973:
Second note : 

I will also add a new test into the test suite
to cover this case (as it obviously wasn't up to now).

chipturner:
Apparently `_unlink` doesn't exist in Linux.  Looks like `remove` will be sufficient and (hopefully) work on both platforms.

Cyan4973:
It's also possible to special-case Visual with a `#define`, notably if `_unlink` on Windows has the same interface as `unlink` on Linux.

Cyan4973:
Well, if that works fine with `remove`, let's use that :)

PR #152, opened 2016-03-18
Update VS2013 vcxprojs to all build to the same folders

KrzysFR:
Currently, when VS 2013 builds the various projects (zstd.exe, tools, zstlib, ...), they all end up in various folders all other the place, making it difficult to find the correct location:
- `zstd.exe` ends up `../vs2013/Release/zstd.exe' for both x86 and x64 build
- the other projects end up in '.../vs2013/Win32/Release/` or `.../vs2013/x64/Release/`

This PR changes all the vcxprojs to output to `./bin/{ARCH}/{CONFIG}` by default (ex: `bin/Win32/Release` or `./bin/x64/Release`). This way, all the outputs are in a `bin` subfolder, and are easy to find.

This would also work if in the future there are additional release types (like a ReleaseStatic, or with custom settings, ...)

If people are running a CI server to run tests after the build, on Windows, they would need to update the paths to `zstd.exe` or `fuzzer.exe` from `.../Release/zstd.exe` to `.../bin/x64/Release/zstd.exe` (for x64) or `.../bin/Win32/Release/zstd.exe` (for x86).

### DIFF
--- a/visual/2013/fullbench/fullbench.vcxproj
+++ b/visual/2013/fullbench/fullbench.vcxproj
@@ -22,6 +22,7 @@
     <ProjectGuid>{61ABD629-1CC8-4FD7-9281-6B8DBB9D3DF8}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>fullbench</RootNamespace>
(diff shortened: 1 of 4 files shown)


Cyan4973:
:+1: Good initiative

PR #156, opened 2016-03-31
Win32: Support for stdcall linkage and legacy VC6 compiler

wqweto:
MSVC 6.0 has couple of troubles with current codebase
1. Has much less intrinsics (no `__BitScanForward/Reverse`)
2. Doesn't support for `long long` and uses proprietary `__int64` type instead
3. Doesn't support `ULL` and `LL` suffixes for 64-bit literals and uses proprietary `ui64` and `i64` suffixes
4. Doesn't support variadic macros
5. Is fussy about param constness on function pointer types

Legacy MSVC support is based on `_MSC_VER` being 1400 for Visual Studio 2005. Anything below is considered legacy version.

This also introduces `ZSTDLIB_STDCALL` constant on functions included in the public interface to be easily forced to stdcall calling convention (for x86 builds only) and adds `visual/stdcall` directory with a short batch file for building library DLLs with stdcall calling convention (for x86) both release and debug builds. `build.bat` can build x64 versions too.


visual/stdcall/build.bat
```diff
@@ -0,0 +1,41 @@
+@echo off
+setlocal
+:: call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\vcvars32.bat"
+:: call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\vcvars64.bat"
+set lib_name=zstdlib-x86
+set cl_exe=cl.exe /DZSTDLIB_STDCALL=__stdcall
+set cl_flags=/MT /LD /O2 /Ox
+set link_flags=/link /base:0x3f400000 /version:1.0 /incremental:no /opt:ref /merge:.rdata=.text /ignore:4078
+set src_dir=..\..\lib
+set bir_dir=Release
```
wqweto: `bir_dir` is a funny misspelling, will change it to `bin_din` (kidding:-))

KrzysFR:
Could it be possible to merge both `ZTDBLIB_STDCALL` and `ZSTDLIB_API` macros into a single one? They seem to be always present together (except for the few cases where probably `ZSTDLIB_API` was missing). This could reduce the clutter in the method declaration that only concerns Windows platform, and also reduce risk of adding one but forgetting the other.

wqweto:
Could be done but wrapping retval type in macro always looks funny to me
```
ZSTDLIB_API(size_t) ZSTD_compress(void* dst, size_t maxDstSize, const void* src, size_t srcSize, int compressionLevel)
```

Cyan4973:
Not sure why this patch was removed.
It's not a bad idea to attempt vc6 compatibility, although it seems to introduce quite some complexity throughout the code, making overall more difficult to read.

Also : we are in the middle of a big code update, I would advise to wait for it before submitting such changes, as otherwise, we will have a lot of difficulties merging it.

PR #157, opened 2016-04-01
Win32: Support for stdcall linkage and legacy VC6 compiler (second attempt)

wqweto:
MSVC 6.0 has couple of troubles with current codebase
1. Has much less intrinsics (no `__BitScanForward`/`Reverse`)
2. Doesn't support `long long` and uses proprietary `__int64` type instead
3. Doesn't support `ULL` and `LL` suffixes for 64-bit literals and uses proprietary `ui64` and `i64` suffixes
4. Doesn't support variadic macros
5. Is fussy about param constness on function pointer types

Legacy MSVC support is based on `_MSC_VER` being 1400 for Visual Studio 2005. Anything below is considered legacy version.

This also introduces `ZSTD_DLL_STDCALL` constant for functions included in the public interface to be easily forced to stdcall calling convention (for x86 builds only) and adds visual/stdcall directory with a short batch file for building library DLLs with stdcall calling convention (for x86) both release and debug builds. `build.bat` can compile x64 versions of the DLLs too.


lib/mem.h
```diff
@@ -63,7 +63,17 @@ extern "C" {
...
-#if defined (__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */)
+#if defined(_MSC_VER) && (_MSC_VER < 1400)     /* For VC6 and other legacy versions */
+  typedef unsigned char       BYTE;
+  typedef unsigned short      U16;
+  typedef   signed short      S16;
+  typedef unsigned int        U32;
+  typedef   signed int        S32;
+  typedef unsigned __int64    U64;
+  typedef   signed __int64    S64;
+  #define LIT_S64(x) x##i64
```
Cyan4973: minor detail : why did you selected `LIT_` name ?

Cyan4973:
Woof, that's a large code base change !
Although most of the changes look pretty well located, so I don't see large issues.

The real pb though is that we are going to merge `alternate2` into `dev060` and then into `dev`, 
and this patch will become incompatible.
Please wait a bit for this change to happen, you'll be able to re-submit your PR right afterwards.

lib/mem.h
wqweto: `LIT` is from "literal". Naming is hard so I'm always open for suggestions.

wqweto:
Sure, I'll rebase it then. Do you have ETA for this merge?

lib/mem.h
Cyan4973: In LZ compression, literals designate bytes which won't be regenerated using an LZ match.
Since it's not the point here, you might refer to another concept I'm not aware of.
Anyway, in the context of LZ compression, it will be confusing.

The objective of these macros is to write 64-bits explicit constants, that are currently written `12345ULL`, in a format which VC6 is able to understand.
So I would suggest as names `CONST_U64()` and `CONST_S64()`.

Cyan4973:
early next week

lib/mem.h
wqweto: Makes sense and will take care of this. Although "literals" is the precise language term, in context of ZSDT it's really kind of confusing.

PR #161, opened 2016-04-04
Dev

inikep:
opt_parser: price estimation adapted to the new length format
bench.c: possibility of testing levels from -bX to -rY
bench.c: better timer for Windows


lib/zstd_stats.h
```diff
@@ -0,0 +1,164 @@
+/*
+    zstd - standard compression library
```
Cyan4973: If I do understand, this is a new file, for debug-only operations ?

lib/zstd_opt.h
```diff
@@ -154,23 +162,39 @@ MEM_STATIC void ZSTD_updatePrice(seqStore_t* seqStorePtr, U32 litLength, const B
...
+        const BYTE LL_deltaCode = 19;
+        const BYTE llCode = (litLength>63) ? (BYTE)ZSTD_highbit(litLength) + LL_deltaCode : LL_Code[litLength];
+        seqStorePtr->litLengthFreq[llCode]++;
+        seqStorePtr->litLengthSum++;
+    }
 
     /* match offset */
     seqStorePtr->offCodeSum++;
-    BYTE offCode = offset ? (BYTE)ZSTD_highbit(offset+1) + 1 : 0;
+    BYTE offCode = ZSTD_highbit(offset+1);
```
Cyan4973: minor `U32`->`BYTE` implicit conversion warning here.

lib/zstd_stats.h
inikep: Yes, by default it shouldn't be included at all:

#if ZSTD_OPT_DEBUG == 3
    #include "zstd_stats.h"
#else

Cyan4973: Clear enough. No pb on compilation.
I'm just wondering about source tree.

Currently, legacy support is within its own directory, because we don't want to include that in the general case.
I'm wondering if we should do the same for debug-only files.

inikep: I used this file for generation of stats (separately for literals, matches, offsets, ...). It might be useful in the future, but:
1. The file "zstd_stats.h" may not be released to public at all (more files = more confusion).
2. It's good to have the "zstd_stats.h" file in github (in case of e.g. HDD crash).

Cyan4973: I fully agree.
So, how to combine 1. and 2. ?
What do you think of having a `lib/debug` directory ?

inikep: "debug" is ignored in .gitignore and can give problems with Visual Studio (debug/release)

Cyan4973: ok, any other suggestion ?

Cyan4973: `lib/debugOnly` ?

inikep: _debug or .debug

Cyan4973: good point, 
ok for `.debug`

inikep: zstd_stats.h moved to .debug/

programs/bench.c
```diff
@@ -44,32 +44,58 @@
...
+#elif defined(_WIN32)
+   typedef LARGE_INTEGER BMK_time_t;
+#  define BMK_initTimer(x) if (!QueryPerformanceFrequency(&x)) { fprintf(stderr, "ERROR: QueryPerformance not present\n"); }
+#  define BMK_getTime(x) QueryPerformanceCounter(&x)
+#  define BMK_getSpanTimeMicro(ticksPerSecond, clockStart, clockEnd) (1000000ULL*(clockEnd.QuadPart - clockStart.QuadPart)/ticksPerSecond.QuadPart)
+#  define BMK_getSpanTimeNano(ticksPerSecond, clockStart, clockEnd) (1000000000ULL*(clockEnd.QuadPart - clockStart.QuadPart)/ticksPerSecond.QuadPart)
+#else
+   typedef int BMK_time_t;
+#  define BMK_initTimer(ticksPerSecond) ticksPerSecond=0
+#  define BMK_getTimeMicro(clockStart) clockStart=1
```
Cyan4973: For some reason, this path seems to declare `BMK_getTimeMicro(x)` instead of `BMK_getTime(x)` like the other ones.

Are we sure this path work ?
Is there a way to test it ?

inikep: A good catch, when you will change BMK_getTimeMicro(x) to BMK_getTime(x) it compiles without any warnings. But the problem is that when you run it there will be "WARNING !!! Invalid Checksum".

It can be fixed with changing 2 "for" loops to "do-while" loops but the results without timer have no sense unless we will print only ratio without ctime/dtime:
 1#remix_tar         : 101679616 ->  43214701 (2.353), 101.7 MB/s , 101.7 MB/s
 2#remix_tar         : 101679616 ->  41813113 (2.432), 101.7 MB/s , 101.7 MB/s
 3#remix_tar         : 101679616 ->  40671617 (2.500), 101.7 MB/s , 101.7 MB/s

What do you think about it?

Cyan4973: I believe all 3 paths should behave the same.

`clock_t` methodology worked well and is portable.
It should be default, when nothing else applies.

inikep: clock() works only for POSIX. Do you want to ignore non-POSIX and non-Windows operating systems? I know it's hard to find one but there are some.

Cyan4973:
`clock()` and `clock_t` are part of `<time.h>` which is itself part of the [C standard libraries](https://en.wikipedia.org/wiki/C_date_and_time_functions).
So it's supposed to be supported beyond POSIX.

programs/bench.c
Cyan4973: `clock()` and `clock_t` are part of `<time.h>` which is itself part of the [C standard libraries](https://en.wikipedia.org/wiki/C_date_and_time_functions).
So it's supposed to be supported beyond POSIX.

inikep: OK, this is a fix: clock() is default timer for all platforms except Windows:
https://github.com/inikep/zstd/commit/1eeddde6257b8d8156c5a4e2fd94548245e63b29

Cyan4973: LGTM (Looks Good To Me) !

Issue #160, opened 2016-04-04
v0.6.0 Streaming API


luben:
I know it is work in progress but just to point out that the streaming API (from zstd_static.h) is incompatible with the results of the CLI zstd program.

- when compressing it creates result that is 8 bytes smaller then the CLI
- when decompressing a file compressed by the CLI util I get "Src size incorrect" on the first ZSTD_decompressContinue with the size returned by the first invocation of ZSTD_nextSrcSizeToDecompress.

The streaming API is compatible in itself though - it roundtrips the compression/decompression.

Cyan4973:
Thanks for the notification.
Indeed, `dev060` branch was not yet fully stabilized, but we are getting close to this point.
We'll look into this issue too.

Cyan4973:
The 8 bytes of difference come from one version storing the frame content length, and the other version not.

Presumably, the streaming version does not store this information, because it doesn't know it.
The cli version can know it if the source is provided as a file : `zstd filename` .
But if it is provided as pipe data, it will not : `cat filename | zstd -c > filename.zst`

I presume the problem come from the presence of the frame content field, making the header larger.
There is indeed a difficulty now, because the size of the frame header is basically variable.

It can be made to work, since `zbuff` is just using this API successfully, but maybe it's too complex or tortured.
I will have a closer look into it.

luben:
I will take a look how is done in zbuff.c or may be switch directly to it.

luben:
I made it work but it is sub-optimal IMHO, here is how it works in my case:

1. Without size:

- ZSTD_createDCtx + ZSTD_decompressBegin
- do the ZSTD_decodeParams, reading the first 5 bytes (the frame header)
- allocate the buffers (using sizes in the params)
- call ZSTD_decompressContinue on the frame header
- continue with alternating ZSTD_nextSrcSizeToDecompress and ZSTD_decompressContinue

2. With size:

- ZSTD_createDCtx + ZSTD_decompressBegin
- do the ZSTD_decodeParams, reading the first 13 bytes (the frame header 5 + 8)
- allocate the buffers (using sizes in the params)
- call ZSTD_decompressContinue on the first 5 bytes of the frame header
- copy the next 8 bytes of the frame header in the inputBuffer
- call ZSTD_decompressContinue on it
- continue with alternating ZSTD_nextSrcSizeToDecompress and ZSTD_decompressContinue

[...]

PR #162, opened 2016-04-05
Consume the large frame header in one go if we have the data

luben:
My proposal explained in #160


luben:
discarding it, details and discussion in the issue.

PR #163, opened 2016-04-07
3 repcodes and MINMATCH=3

inikep:


programs/Makefile
```diff
@@ -84,6 +84,10 @@ default: zstd
 
 all: zstd zstd32 fullbench fullbench32 fuzzer fuzzer32 zbufftest zbufftest32 paramgrill datagen
 
+zstd_clang: $(ZSTD_FILES) $(ZSTD_FILES_LEGACY) $(ZSTDDIR)/zbuff.c $(ZSTDDIR)/zdict.c $(ZSTDDIR)/divsufsort.c \
+        zstdcli.c fileio.c bench.c xxhash.c datagen.c dibio.c
+	clang.exe --target=x86_64-w64-mingw32 -Werror -Wconversion -Wno-sign-conversion $(FLAGS) -DZSTD_LEGACY_SUPPORT=$(ZSTD_LEGACY_SUPPORT) $^ -o $@$(EXT)
```
Cyan4973: `clang.exe` is Windows specific.
Prefer something more portable, either `clang$(EXT)` or even simply `clang` as I guess it will work on Windows too.

programs/Makefile
```diff
@@ -53,7 +53,7 @@ BINDIR  = $(PREFIX)/bin
 MANDIR  = $(PREFIX)/share/man/man1
 ZSTDDIR = ../lib
 
-ZSTD_FILES := $(ZSTDDIR)/huff0.c $(ZSTDDIR)/fse.c $(ZSTDDIR)/zstd_compress.c $(ZSTDDIR)/zstd_decompress.c
+ZSTD_FILES := $(ZSTDDIR)/huff0.c $(ZSTDDIR)/fse.c $(ZSTDDIR)/zstd_decompress.c $(ZSTDDIR)/zstd_compress.c
```
Cyan4973: Is the order change intentional ?

inikep: Yes, currently it gives the same speed on fizzle and seems to be safer.

programs/Makefile
inikep: I've installed clang on Windows and this is intented just for me.
The "--target=x86_64-w64-mingw32" also depends on mingw version.
Do you have any ideas how to hide it better?

Cyan4973: I don't think you need to put that in the `Makefile` then.
You could have your own local batch file, and launch the compilation process like this : 
`make zstd CC="clang.exe --target=x86_64-w64-mingw32" `

You can also use MOREFLAGS the same way if you want to add specific flags not present by default in the `Makefile`

lib/zstd_compress.c
```diff
@@ -1504,118 +1689,195 @@ static size_t ZSTD_BtFindBestMatch_selectMLS_extDict (
...
 
-    zc->nextToUpdate = target;
-    return hashTable[ZSTD_hashPtr(ip, hashLog, mls)];
-}
+    const U32 maxSearches = 1 << ctx->params.cParams.searchLog;
+    const U32 mls = ctx->params.cParams.searchLength;
 
+    /* init */
+    U32 rep[ZSTD_REP_INIT];
+    for (U32 i=0; i<ZSTD_REP_INIT; i++)
```
Cyan4973: I found this construction in a number of places :
`for (U32 i=0; ...`

While I really do like it, and would like to use it,
it's unfortunately incompatible with non-C99 compilers.

Yes, the year is 2016, but it's nonetheless too soon to drop support for these compilers,
the goal being wide portability.

Therefore prefer : 
`{ U32 i; for (i=0; ... `
which is C89 compatible.

It requires one more bracket. Still, portability is more important.

_Note_ : I made the associated modifications in ea63bb7b5e41db6143905e7c8a0ff2385a2ff677 .

_Note 2_ : maybe we need a test to ensure this compatibility is respected. Currently, all compilations in `Makefile` are done with `-std=c99` compilation flag.

inikep: If we want zstd to work with -std=c89 we need many more changes than this.
I checked and there are plenty of errors. Some are as simple as:
error: C++ style comments (//) are not allowed in ISO C90

Cyan4973: While it's a good idea to converge towards C90,
hence to convert all `//` into `/* ... */`, _at least in the library itself_,
it's not a mandatory target either.

Reason is, strict C90 is very limited, especially regarding 64-bits types.

But the world is not divided between "strict c90" and "strict c99".
This would be too easy.
In fact, most compilers which are not completely c99 do support a subset of it,
among which the `//` comment tag is the most common.

It's pretty hard to find a compiler which doesn't support `//`,
but unfortunately it's very common for non-c99 compilers to not support `if (U32 i=0; ...`
Hence I put more pressure on avoiding this construction.

Bottom line : yes, try to stick to c90 whenever possible, including regarding comments tags.

inikep: So we should target -std=gnu89.

Currently my compilation with -std=gnu89 shows only the following error:
./lib/zstd_opt.h:566:12: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
            for (U32 i=0; i<ZSTD_REP_NUM; i++)

Cyan4973: ok

PR #166, opened 2016-04-10
Win32: Support for stdcall linkage and legacy VC6 compiler (3rd attempt)

wqweto:
This PR is rebased to the newest commits in `dev` branch.


Cyan4973:
Thanks, this is a great patch.

As we are hours away from v0.6, we'll refrain from introducing such large change, in order to avoid last minute regression.

We will therefore integrate your patch for v0.6.1 or v0.7.0.

lib/zbuff.h
```diff
@@ -45,29 +45,40 @@ extern "C" {
 *  Compiler specifics
 *****************************************************************/
 /*!
+*  ZSTD_DLL_STDCALL :
+*  Provide the ability to use stdcall linkage on public interface when building a Windows DLL
```
Cyan4973: I'm not an expert of this domain,
but shouldn't `stdcall` be the default calling convention when compiling for Windows anyway ?

lib/zbuff.h
```diff
@@ -45,29 +45,40 @@ extern "C" {
...
 #endif
 
-
 /* *************************************
 *  Streaming functions
 ***************************************/
 typedef struct ZBUFF_CCtx_s ZBUFF_CCtx;
-ZSTDLIB_API ZBUFF_CCtx* ZBUFF_createCCtx(void);
-ZSTDLIB_API size_t      ZBUFF_freeCCtx(ZBUFF_CCtx* cctx);
+ZSTDLIB_API(ZBUFF_CCtx*) ZBUFF_createCCtx(void);
```
Cyan4973: It's indeed pretty strange to see such construction, with the type inside parenthesis,
but if I understand correctly, `__stdcall` must be written after the function return type,
in which case, I don't see any better solution.

Cyan4973:
I'm a bit embarrassed.
This is a great patch, really, but unfortunately it's also a bit invasive.

In general, adding support for an additional platform tends to require a few macros and definitions changes in some well located place. Here, there are really a lot of changes, in many places and files.

This would be less of a problem if the purpose is to add a major platform, but this is also not the case : VC6, while arguably still in use, is by now a fairly old environment, with only diminishing perspectives.

So I'm a puzzled. Is it a good idea to introduce so many changes in the main reference trunk only for this platform ?

I suspect the answer is no.
In some circumstances, forks are a better answer. This is probably one of those circumstances.

This is really an embarrassing conclusion, as I appreciate the great work you made to create this patch.

I also wonder if a lighter patch would pass the bar to become acceptable in reference trunk.
My feeling is that the `stdcall` part is likely the major contributor to patch's invasiveness.
Maybe without it, the amount of change would become more localized, hence more acceptable.
But I'm not sure though if it makes sense.

wqweto:
No problem. Let's leave it out of main trunk for now.

I'll just rebase my stdcall fork on latest releases as time permits.

PR #170, opened 2016-04-15
Change 'Warning as Error' to false and fix vcxproj

KrzysFR:
This PR changes the VC projects of fuzzer and fullBench to not treat warning as errors. It also removes the old `zstd_buffered*.h` files from the project which do not exist anymore;

Also, I had to change `.gitattributes` to treat vcxproj as text files with windows style line ending, in order to be able to see changes in difftools. For some reason, the very first commit after that is a wall of pink, not sure how to work around that... anyone has an idea?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,5 +11,7 @@
 # Visual Studio
 *.sln binary
 *.suo binary
-*.vcxproj* binary
+*.vcxproj* text eol=crlf
+
+
 
--- a/visual/2013/fullbench/fullbench.vcxproj
+++ b/visual/2013/fullbench/fullbench.vcxproj
@@ -94,7 +94,7 @@
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
       <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <TreatWarningAsError>true</TreatWarningAsError>
+      <TreatWarningAsError>false</TreatWarningAsError>
       <EnablePREfast>true</EnablePREfast>
       <AdditionalOptions>/analyze:stacksize25000 %(AdditionalOptions)</AdditionalOptions>
     </ClCompile>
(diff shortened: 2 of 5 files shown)


Cyan4973:
Thanks for notification @KrzysFR .
There are several improvements proposed in your patch.

Regarding the warnings in x64 mode, the better solution is to fix them.
I think they slipped through because our automated Visual tests on AppVeyor are too light, and only run the first configuration, which is debug 32-bits mode. We would probably benefit from a better AppVeyor script to test more than that and catch such warning before release.

`.gitattributes` changes are good too.
They were treated as binary so far to avoid automatic conversion of eol, but your proposed solution is better.

Cleaning the project files is good too.
Note however that the `lib` directory just received a major change, with direct impact on project files.

KrzysFR:
I'll try to extract the gitattributes and cleanup in a separate PR.

For the warning as error, though, when they happen, you are not able to build for release straight after a checkout, and most people might not be able to understand and fix the offending warning.

Are you proposing that by adding a CI build step that build for Release, warnings introduced in commits would be caught and fixed immediately?

Cyan4973:
> Are you proposing that by adding a CI build step that build for Release, warnings introduced in commits would be caught and fixed immediately?

Yes. Maybe we should also consider 2 different configurations.
One for CI, which would be very unforgiving, treat warnings as errors, impose a high warning level (4), testing both 32 and 64 bits paths (which can generate different warnings), so that warnings get caught and fixed early (at each commit),

and another configuration for users, with lighter conditions, such as default warning level (3) and authorized warnings, so they could make modifications if necessary, with less burden,

That will highly depend on what we can do with our Windows CI environment, AppVeyor, which currently is very limited. This is an ongoing topic, led by @inikep .

inikep:
1. In the https://github.com/inikep/zstd/commits/dev branch the following patches are already implemented:
- Treat vcxproj as text files with Windows style line endings
- Remove old zstsd_buffered_*.h from zstd.vcxproj

2. The above-mentioned branch compiles without a problem with Visual Studio 2012 and 2015 with Win32 and x64 (TreatWarningsAsErrors="true"). 

3. We don't need two projects for users and AppVeyor. I can set TreatWarningsAsErrors="false" in the VS project and enable in only for AppVeyor (msbuild /p:TreatWarningsAsErrors="true"), but currently TreatWarningsAsErrors="true" should not cause any problems.

KrzysFR:
This looks the ideal solution to me, thanks!

Can close this issue for me? (only have access the email right now)

Cyan4973:
Latest merge b3a9938d0c78f213136ba841aef3240f46785bb6 should provide a few improvements on this topic

PR #171, opened 2016-04-15
Add missing ZSTDLIB_API to new API

KrzysFR:
There were a few missing ZSTDLIB_API in zstd_static.h for the new 0.6 API.

Also fixed new name for ZSTD_getCParams in a comment.

### DIFF
--- a/lib/zstd_compress.c
+++ b/lib/zstd_compress.c
@@ -2552,7 +2552,7 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
 },
 };
 
-/*! ZSTD_getParams() :
+/*! ZSTD_getCParams() :
 *   @return ZSTD_parameters structure for a selected compression level and srcSize.
 *   `srcSize` value is optional, select 0 if not known */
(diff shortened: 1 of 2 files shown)


Cyan4973:
Good catch ! Thanks @KrzysFR !

Cyan4973:
The directory structure of the `lib` directory has been changed quite a bit, in order to allow separate compilation of compression and decompression code.
As a consequence, your patch no longer merges.

The proposed changed is still valid though.
I guess it will require a rebase before re-submit.

KrzysFR:
Will try to rebase this to the latest branch. Though, if you are planning to add more APIs, maybe we should wait for that too?

Cyan4973:
There is a zlib API wrapper ongoing, but it will be separated.
Not sure it's worth waiting for it.

Alternatively, I could also add your changes manually. They are simple enough.

KrzysFR:
If you can add them yourself, please do.

What do you mean by zlib wrapper?

Cyan4973:
> What do you mean by zlib wrapper?

An API which looks like zlib, but handle both zstd and zlib formats transparently.

Cyan4973:
Your proposed modifications have been integrated into "dev" branch

Issue #173, opened 2016-04-19
compression/decompression error


ghood:
When compressing and then decompressing a test file, I don't get the original file back:
$ ./zstd -V
*** zstd command line interface 64-bits 0.6.0, by Yann Collet ***
$ ./zstd test -o test.zst
Compressed 10688440 bytes into 4209313 bytes ==> 39.38%                        
$ ./zstd -d test.zst -o test.decoded
Successfully decoded 10688440 bytes                                            
$ cmp test test.decoded
test test.decoded differ: byte 2588022, line 3678
$ ls -l test*
-rw-r--r-- 1 ghood staff 10688440 Apr 19 14:12 test
-rw-r--r-- 1 ghood staff 10688440 Apr 19 14:25 test.decoded
-rw-r--r-- 1 ghood staff  4209313 Apr 19 14:24 test.zst
$ md5sum test*
21c75c134b09e734bd4d5b78e749b1fd  test
6e5dcc872da801fd07103fba6a7ee097  test.decoded
83dad083093ae38d3838de71666ff024  test.zst
$ uname -a
Linux axon 3.13.0-43-generic #72 SMP Mon Jan 12 10:50:11 EST 2015 x86_64 x86_64 x86_64 GNU/Linux

The zipped test file is attached.
[test.zip](https://github.com/Cyan4973/zstd/files/226584/test.zip)

Anyone know what is going wrong?
Thanks.
--Greg

luben:
Looks the but was introduced with v0.4.6, it works as expected with v0.4.5

Cyan4973:
I confirm both reports : the test file fails round-trip test at `-1` setting for all versions starting v0.4.6.
Now onto bug chasing ...

Cyan4973:
Preliminary assessment :

This looks like an error in the huffman decoder.

The guilty byte is right in the middle of a long literal section. So the problem seems unrelated to match finding and decoding. It could happen with any compression mode, and it just randomly happens here at `-1` setting and after v0.4.6 changes "by chance".

huff0 has an internal fuzzer to experience and test as many strange situations as possible, but somehow it did not found this one. I believe a first task would be to create a test which reproduce the problem. This would make it possible to both control its correction and continuously check in future versions that it remains safe.

Cyan4973:
The guilty code is the quad-symbol huffman decoder.
[...]

PR #174, opened 2016-04-20
Revert "minor fast mode improvement"

luben:
This reverts commit 96ffa424ebc4025744aaf95f3430a8142d94ec97.

Fixes issue #173


luben:
... but it makes the compression ratio worse on the fast levels (compared 1 and 3)

Cyan4973:
okay,
one of the issues I see with this patch, even though it works, is that `MINMATCH` used to mean `4` in the original version, while it means `3` in 0.6+ version. So the formula `(ip-(mlCode+MINMATCH)+2` has a different result.

Let's try to understand what's happening there...

luben:
Completely agree. I just bisected the commit that has introduced this particular compression corruption and verified that reverting it fixes the issue. I agree that we should have better understanding why it happens and create a proper fix.

luben:
Closing it as we agree this is not the right fix. At least we now know which change introduced the break.

PR #179, opened 2016-05-04
Remove extraneous semicolons in zstd_internal.h

thatsafunnyname:
Remove extraneous semicolons in zstd_internal.h to avoid compiler informational %CC-I-EXTRASEMI, and also to avoid the null statements, the null statements are not causing a problem on these lines, but the EXTRASEMI informational can be useful at detecting places where null statements can cause unexpected behaviour, so it is good to have compilations without this information being generated (these were the only lines that generated it in zstd).

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -63,7 +63,7 @@ extern "C" {
 /*-**************************************************************
 *  Basic Types
 *****************************************************************/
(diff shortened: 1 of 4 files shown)


thatsafunnyname:
Extraneous semicolons in zstd_internal.h removed with:
https://github.com/Cyan4973/zstd/commit/ddb8ebd5b306240915d734403c808ff481243f7b
"Stricter gnu90 tests"
Thanks.

PR #180, opened 2016-05-04
No stdint.h on OpenVMS.

thatsafunnyname:
No stdint.h on OpenVMS, it is inttypes.h instead, some projects use HAVE_STDINT_H.

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -63,7 +63,7 @@ extern "C" {
 /*-**************************************************************
 *  Basic Types
 *****************************************************************/
-#if defined (__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */)
+#if !defined (__VMS) && ( defined (__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */) )
 # include <stdint.h>
   typedef  uint8_t BYTE;
   typedef uint16_t U16;
--- a/programs/datagen.c
+++ b/programs/datagen.c
@@ -34,7 +34,7 @@
 /*-************************************
 *  Basic Types
 **************************************/
-#if defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)   /* C99 */
+#if !defined (__VMS) && ( defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)   /* C99 */ )
 # include <stdint.h>
   typedef  uint8_t BYTE;
   typedef uint16_t U16;
--- a/programs/xxhash.c
+++ b/programs/xxhash.c
@@ -123,7 +123,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size) { return memcp
 ***************************************/
 #ifndef MEM_MODULE
 # define MEM_MODULE
-# if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* C99 */
+# if !defined (__VMS) && ( defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* C99 */ )
 #   include <stdint.h>
     typedef uint8_t  BYTE;
     typedef uint16_t U16;


Cyan4973:
Thanks, it's a simple enough change.

Just for my understanding : I'm surprised of this issue, because `stdint.h` is supposed to be [part of the C99 standard library](https://en.wikibooks.org/wiki/C_Programming/C_Reference/stdint.h), and the test is supposed to check for `C99` compliance.

Does that mean OpenVMS claims C99 compliance but does not provide `stdint.h` ?

Related : my understanding is that `inttypes.h` is a superset of `stdint.h`, which it is supposed to include. See https://en.wikibooks.org/wiki/C_Programming/C_Reference/inttypes.h : 
> The inttypes.h file is a C header file that is part of the C standard library and API. It was added with the 1999 version of the ISO C standard (known as C99). It includes the stdint.h header

thatsafunnyname:
Yes, the OpenVMS HP C compiler claims C99 compliance but stdint.h is not provided.

A mention in HP OpenVMS Systems, UNIX, Linux to OpenVMS Porting Guidelines:

http://h71000.www7.hp.com/portability/portingguidelines.html

    4.1. Data Structures and Data types
      4.1.1.   int64_t and uint64_t
        In UNIX or Linux, int64_t and uint64_t data types are defined in the stdint.h, sys.h or types.h
        header files. However, in OpenVMS, these data types are defined in the inttypes.h header file.
        Hence, the header file inttypes.h must be included when porting to OpenVMS systems.
 
Some detailed discussion:
 
https://groups.google.com/d/msg/comp.os.vms/Bnh3tIOc7bo/flrc8d626WgJ
 

>   Craig A. Berry 12/24/14 Re: compiler language support
>  
>   On 12/22/14 3:11 PM, John Reagan wrote: 
> > On Monday, December 22, 2014 9:58:04 AM UTC-5, Stephen Hoffman wrote: 
> > 
> > 
> >> Interesting.   Haven't heard that one.   AFAIK, C99 was just something 
> >> that never made the list of stuff that got (completely) implemented and 
> >> shipped. 
> > 
> > I haven't heard that either and I was there.  The compiler and RTL were under 
> > the same management for years. 
> 
> I no longer remember exactly how I came to this conclusion. Maybe it's 
> an impression I formed so long ago that it feels like a memory. I've 
> long tried to reconcile the repeated assertions by members of the 
> compiler team that the compiler is C99-compliant but the CRTL is not. 
> That's a complete non sequitur as far as the standard goes since the 
> standard governs the language, not the compiler nor the library by 
> itself, and I've never seen another implementation where the standards 
> compliance for the compiler was thought to be separate from the state of 
> the library. 
> 
> And this isn't just being pedantic. When the compiler defines: 
> 
> __STDC_VERSION__=199901L 
> 
> but the library and headers don't implement everything mandatory from 
> C99, then various packages fail to compile because they test for C99 and 
> assume the required features are available. In some cases they have 
> perfectly adequate fall-back code that they don't make use of since 
> they've been told by the compiler that they don't need it. 
> 
> >> The C99 headers would have been solely part of the C kit, after all. 
> >> Those headers are not shipped with VMS. 
> 
> The library ships with the operating system. The headers are really part 
> of the library whether they ship with it or not. 
> 
>  >> And there are those C99 
> >> headers that are missing.   Ugh.  Now which C99 header is the one I'm 
> >> that always running into?   stdint.h?  Yeah.  stdint.h is missing. 
> > 
> > Yeah, that one always surprises me.  stdint.h and fenv.h are both missing. 
> > [fenv.h is an edge condition that we wouldn't support anyway.] 
> 
> 
> stdint.h is puzzling since most of it would be so easy to implement: 
> just a bunch of typedef statements based on built-in types. 
> 
> inttypes.h is also missing a number of things required by the standard, 
> mostly format specifiers. It's not clear to me in this case what the 
> compiler knows, what the headers know, and what the relevant library 
> functions (scanf, printf, etc.) know. 
 
Just FYI, gcc on OpenVMS will provide stdint.h from:

https://github.com/gcc-mirror/gcc/commit/e55b581dfa231b70aa01ea2b85758369f98386be#diff-b6372309633c7131af243fa2773f7f1b
https://gcc.gnu.org/ml/gcc-patches/2012-03/msg01133.html

(I am using the HP C compiler).
Thanks.

Bloody2:
This issue can be solved by providing users with exotic compilers with an independent stdint.h header implementation. From my codebase:

//
// stdint.h:
//
// I'm no longer willing to define my own types for everything just because
// certain compilers refuse to supply even the most fundamental stuff.
// If someone wants to use my sourcecode, then that someone will have to keep
// up with me, not the other way around.
//
// Download an implementation for your compiler here:
//
//   C (portable, including MSC):
//     https://github.com/James-Jones/HLSLCrossCompiler/blob/master/include/pstdint.h
//
//   C (MSC only):
//     https://code.google.com/p/msinttypes/downloads/list
//
//   C++ (for old compilers):
//     http://www.boost.org/doc/libs/release/boost/cstdint.hpp
//
#include <stdint.h> // ***** see this header file for download locations *****

PR #183, opened 2016-05-12
added zlibWrapper - zstd wrapper for zlib

inikep:
bench.c: experimental -r (operate recursively on directories) for Windows and _POSIX_C_SOURCE >= 200112L

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -56,11 +56,17 @@ extern "C" {
 /*-****************************************
 *  Dependencies
 ******************************************/
-#include <stdlib.h>     /* _POSIX_C_SOURCE, malloc */
+#include <stdlib.h>     /* features.h with _POSIX_C_SOURCE, malloc */
 #include <stdio.h>      /* fprintf */
 #include <sys/types.h>  /* stat */
 #include <sys/stat.h>   /* stat */
 #include "mem.h"        /* U32, U64 */
+
+
+/* *************************************
+*  Constants
+***************************************/
+#define LIST_SIZE_INCREASE   (8*1024)
 
 
 /*-****************************************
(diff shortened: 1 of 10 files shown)


programs/util.h
```diff
@@ -302,36 +315,47 @@ UTIL_STATIC int UTIL_prepareFileList(const char *dirName, char** bufStart, char*
...
+    (void)bufStart; (void)bufEnd; (void)pos;
+    fprintf(stderr, "Directory %s ignored (zstd compiled without _WIN32 or _POSIX_C_SOURCE)\n", dirName);
     return 0;
 }
 
 #endif // #ifdef _WIN32
 
 
-UTIL_STATIC int UTIL_createFileList(const char **inputNames, unsigned nbNames, unsigned maxListSize, const char*** filenameTable, char** allocatedBuffer)
+UTIL_STATIC int UTIL_createFileList(const char **inputNames, unsigned nbNames, const char*** filenameTable, char** allocatedBuffer)
```
Cyan4973: Any reason why the return type is an `int` ?

In general, for a `createThing()` method, I recommend a `thing*` return type (which is `NULL` if allocation fails).

Here, it seems you need to create 2 buffers, so that could a reason.
In which case, maybe a `struct` can be used to combine both into a single object.

Cyan4973:
Excellent `readme` for `zlibWrapper`

programs/util.h
inikep: I've started with "int" because it returns number of files in a list.
Maybe now "const char*** filenameTable" should become return value and nbFiles should be a parameter.
Especially when filenameTable==NULL means error.

Cyan4973: Yes, I agree.
What about `allocatedBuffer` ?

inikep: allocatedBuffer is a buffer for the list of files
filenameTable are only pointers to allocatedBuffer

Actually allocatedBuffer==filenameTable[0] but declaration as "const char**" (inherited from bench.c) do not allow to free(filenameTable[0]).

Cyan4973: Yes.
What I meant is that when you have a `createThing()`, you expect to have a reverse `freeThing()`.

And this is also the case here.

What is less common is that you create and free 2 objects.
It translates into secondary issues, such as which one should become the return type of `createThing()`, why not the other one.

It's not really a problem, it works, so it's more a question of API consistency.

PR #189, opened 2016-05-21
Minor changes needed for OpenVMS

thatsafunnyname:
Define util sleep functions on VMS.
If PRIO_PROCESS is not defined then do not attempt to use it in the definition of SET_HIGH_PRIORITY.


Cyan4973:
All good !

PR #191, opened 2016-05-24
AppVeyor tests

inikep:


lib/common/zbuff.h
```diff
@@ -38,7 +38,8 @@ extern "C" {
 /* *************************************
 *  Dependencies
 ***************************************/
-#include <stddef.h>   /* size_t */
+#include <stddef.h>      /* size_t */
+#include "zstd_static.h" /* ZSTD_customMem */
```
Cyan4973: There is an issue here if `zbuff.h` needs to include `zstd_static.h`.
`zbuff.h` is supposed to be stable and DLL compatible. `zstd_static.h` is clearly not.

On the other hand, `zbuff_static.h` can include `zstd_static.h`.
So I would suggest to move `ZBUFF_createCCtx_advanced()` and `ZBUFF_createDCtx_advanced()` declarations there.

programs/tests/playTestsMSYS.sh
```diff
@@ -0,0 +1,190 @@
+#!/bin/sh -e
```
Cyan4973: What's the difference between `playTests.sh` and `playTestsMSYS.sh` ?

inikep: playTests.sh doesn't work with MSYS on Windows because of "diff" and some minor things. playTestsMSYS.sh is slightly changed playTests.sh which works with MSYS.

Cyan4973: Does it sounds plausible to have the few necessary changes (like `diff`) inside run-time variables, in order to have a single script ?

inikep: It is possible but will take some time. Working on it.

inikep:
The last commit (fixed Visual Studio warning) should work fine. It works in my branch:
https://ci.appveyor.com/project/inikep/zstd/build/1.0.203

Cyan4973:
Indeed, latest version was not update on this repository's Appveyor, for some reason...

Cyan4973:
Looks good to me !
:tada:

PR #193, opened 2016-05-31
Expose function to add entropy tables to pre-built dictionary.

ot:
In some cases a custom dictionary building algorithm tailored for a specific
input can be more effective than the one produced by `ZDICT_trainFromBuffer`,
but with the current API it's not possible encode the entropy tables into the
custom-built dictionary.

This commit extracts the logic to add entropy tables to a dictionary from
`ZDICT_trainFromBuffer` and exposes it as a function
`ZDICT_addEntropyTablesFromBuffer`.


Cyan4973:
Great idea Giuseppe !

Also, note that the project doesn't take PR directly into master.

It should go first through "dev" branch, or even its own feature branch.
Note that the currently active feature branch is dev070.
It features some little changes within `zdict` which will lightly affect your patch (`dictID` header field).

lib/dictBuilder/zdict.c
```diff
@@ -810,6 +810,31 @@ static size_t ZDICT_fastSampling(void* dictBuffer, size_t dictSize,
...
 }
 
+size_t ZDICT_addEntropyTablesFromBuffer_advanced(void* dictBuffer, size_t dictContentSize, size_t dictBufferCapacity,
+                                                 const void* samplesBuffer, const size_t* samplesSizes, unsigned nbSamples,
+                                                 ZDICT_params_t params)
+{
+    unsigned compressionLevel = params.compressionLevel;
+    if (compressionLevel==0) compressionLevel = g_compressionLevel_default;
+
+    size_t hSize;
```
Cyan4973: Unfortunately, we only use variable declaration at beginning of scope, for maximum portability.
You can either declare `hSize` higher in the block,
or change `compressionLevel` declaration to become a one-liner : 

`unsigned const compressionLevel = params.compressionLevel ? params.compressionLevel : g_compressionLevel_default; `

which I would favor to grab the `const` property. 
But that's just a coding taste...

ot: Oh yeah sorry, my C++ habits :) I'll switch to the ternary operator.

ot:
> The currently active feature branch is dev070.

Sure, I'll rebase on that branch.

ot:
Looks like GitHub does not support changing the base branch of a PR, I'll open a new one.

Cyan4973:
Indeed, I believe it's the number one feature request since a few years already ...

https://github.com/isaacs/github/issues?q=is%3Aissue+is%3Aopen+sort%3Acomments-desc

PR #196, opened 2016-05-31
improved Travis tests

inikep:


Cyan4973:
You selected "dev" as a target rather than "dev070".
Any specific reason ?

inikep:
It is still compatible with 0.6. Do you want me to make a PR to dev070?

Cyan4973:
It's indeed a good reason.
That being said, I'm also a bit worried about the merge of `dev` and `dev070` later on, since none of them will be the ancestor of the other. 

A rebase is likely an answer. I think it would have been a no brainer without the "skippable frames" changes. But with them, we have a few files in common. Could prove a headache. 

Suggest to use `dev070` as a target to avoid such issue. Besides, I believe we won't need to release a 0.6.2, so we'll go straight to next major version.

PR #197, opened 2016-06-01
improved Travis tests, skippable frames

inikep:


inikep:
There was quite a lot of collisions to fix. I've also added xxhash.c to fullbench (VS 2008 and VS 2010).

lib/common/zstd_internal.h
```diff
@@ -253,6 +253,6 @@ typedef struct {
 
 const seqStore_t* ZSTD_getSeqStore(const ZSTD_CCtx* ctx);
 void ZSTD_seqToCodes(const seqStore_t* seqStorePtr, size_t const nbSeq);
-
+int ZSTD_isSkipFrame(ZSTD_DCtx* dctx);
```
Cyan4973: Why do we need it declared within `zstd_internal.h` ?
It seems only used within `zstd_decompress.c`. Maybe it could remain local to this file ?

Cyan4973:
Looks good to me

PR #198, opened 2016-06-02
Dev070

inikep:


lib/compress/zbuff_compress.c
```diff
@@ -113,20 +112,18 @@ ZBUFF_CCtx* ZBUFF_createCCtx_advanced(ZSTD_customMem customMem)
...
+        memcpy(&zbc->customMem, &defaultCustomMem, sizeof(ZSTD_customMem));
         zbc->zc = ZSTD_createCCtx();
         return zbc;
     }
 
     if (!customMem.customAlloc || !customMem.customFree)
         return NULL;
 
-    zbc = (ZBUFF_CCtx*)customMem.customAlloc(sizeof(ZBUFF_CCtx));
+    zbc = (ZBUFF_CCtx*)customMem.customAlloc(customMem.opaque, sizeof(ZBUFF_CCtx));
```
Cyan4973: There is now this new `customMem.opaque` in many places.
What is it for ? In which way does it help ?

inikep: This is required to make custom memory allocation compatible with zlib:
    alloc_func zalloc;  /* used to allocate the internal state */
    free_func  zfree;   /* used to free the internal state */
    voidpf     opaque;  /* private data object passed to zalloc and zfree */

lib/common/zstd_static.h
```diff
@@ -97,9 +97,14 @@ typedef struct {
...
 
-typedef void* (*ZSTD_allocFunction) (size_t size);
-typedef void  (*ZSTD_freeFunction) (void* address);
-typedef struct { ZSTD_allocFunction customAlloc; ZSTD_freeFunction customFree; } ZSTD_customMem;
+/* custom memory allocation functions */
+typedef void* (*ZSTD_allocFunction) (void* opaque, size_t size);
+typedef void  (*ZSTD_freeFunction) (void* opaque, void* address);
+typedef struct { ZSTD_allocFunction customAlloc; ZSTD_freeFunction customFree; void* opaque; } ZSTD_customMem;
+
+void* ZSTD_defaultAllocFunction(void* opaque, size_t size);
```
Cyan4973: Do we need these prototypes (`ZSTD_defaultAllocFunction()`, `ZSTD_defaultFreeFunction()`, `defaultCustomMem`) exposed in public area `zstd_static.h` ?

What about `zstd_internal.h` ?

inikep: OK, I will move them

lib/common/zstd_internal.h
```diff
@@ -255,4 +255,9 @@ const seqStore_t* ZSTD_getSeqStore(const ZSTD_CCtx* ctx);
 void ZSTD_seqToCodes(const seqStore_t* seqStorePtr, size_t const nbSeq);
 int ZSTD_isSkipFrame(ZSTD_DCtx* dctx);
 
+/* custom memory allocation functions */
+void* ZSTD_defaultAllocFunction(void* opaque, size_t size);
+void ZSTD_defaultFreeFunction(void* opaque, void* address);
+MEM_STATIC ZSTD_customMem const defaultCustomMem = { ZSTD_defaultAllocFunction, ZSTD_defaultFreeFunction, NULL };
```
Cyan4973: `static` instead of `MEM_STATIC` here : this is a constant variable, not a function.

programs/zbufftest.c
```diff
@@ -512,8 +514,8 @@ int main(int argc, const char** argv)
     int result=0;
     U32 mainPause = 0;
     const char* programName = argv[0];
-    ZSTD_customMem customMem = { ZBUFF_allocFunction, ZBUFF_freeFunction };
-    ZSTD_customMem customNULL = { NULL, NULL };
+    ZSTD_customMem customMem = { ZBUFF_allocFunction, ZBUFF_freeFunction, (void*)777 };
```
Cyan4973: 777 ?

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -74,22 +76,45 @@ const char * zstdVersion() { return ZSTD_VERSION_STRING; }
 ZEXTERN const char * ZEXPORT z_zlibVersion OF((void)) { return zlibVersion();  }
 
 
+void* ZWRAP_allocFunction(void* opaque, size_t size)
```
Cyan4973: This function prototype is not `static`, hence it is accessible from outside.
Is it intended ?

programs/zbufftest.c
inikep: The "opaque" is ignored so value doesn't matter. But I think it should be != 0 for better debugging. Any thoughts?

Cyan4973: I believe `NULL` is a more common value for such usage.

"Better debugging" is fine, it must just be temporary. The shipping code would be better with a more standard `NULL` value.

zlibWrapper/zstd_zlibwrapper.c
inikep: There is no declaration of ZWRAP_allocFunction so it's not accessible from outside.
It will be not inlined as it used as:
```
ZSTD_customMem ZWRAP_customMem = { ZWRAP_allocFunction, ZWRAP_freeFunction, &zwd->allocFunc };
```
Using static can be only helpful to remove it from symbols list.

Cyan4973: > Using "static" can be only helpful to remove it from symbols list.

Yes, that's exactly the point.
If it's not intended to be called from outside its own unit, let's `static` that.
In my view, `static` should have been "default", but it's not, so we have to add it manually.

zlibWrapper/examples/example.c
```diff
@@ -62,11 +62,14 @@ void *myalloc(q, n, m)
     unsigned n, m;
 {
     q = Z_NULL;
-    return calloc(n, m);
+    void *buf = calloc(n, m);
```
Cyan4973: A declaration after a statement.
gcc/clang will run fine, but some compilers will not like it.

Since it's only an example file, it's not a high priority item.
That being said, I recommend the usage of compilation flag `-Wdeclaration-after-statement` to detect such cases early.

zlibWrapper/Makefile
```diff
@@ -7,8 +7,9 @@
 
 # Paths to static and dynamic zlib and zstd libraries
 ifneq (,$(filter Windows%,$(OS)))
-STATICLIB = ../../zlib/libz.a ../lib/libzstd.a
-IMPLIB    = ../../zlib/libz.dll.a ../lib/libzstd.a
+ZLIBDIR = ../../zlib1.2.8
```
Cyan4973: `ZLIBDIR` : how is this path determined ?
If it must be set manually, and if it is dependent on local settings,
could there be a message and/or a comment to attract user attention on the need to define this variable (potentially directly on the command line) ?

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -67,29 +69,52 @@ static int g_useZSTD = ZWRAP_USE_ZSTD;   /* 0 = don't use ZSTD */
...
+
+
 
 /* *** Compression *** */
 
 typedef struct {
     ZBUFF_CCtx* zbc;
     size_t bytesLeft;
     int compressionLevel;
+    z_stream allocFunc; /* copy of zalloc, zfree, opaque */
```
Cyan4973: Is it a kind of circular reference,
where `z_stream` reference `opaque` which is in fact `ZWRAP_CCtx` type which reference the same `z_stream` ?

inikep: zalloc, zfree, opaque are supplied with z_stream during deflateInit().
ZWRAP_CCtx->opaque keeps pointer to copy of z_stream (in case it's changed, zlib.h says only "The application must initialize zalloc, zfree and opaque before calling the init function.").
z_stream->opaque is a parameter supplied by user.

zlibWrapper/Makefile
inikep: It must be set manually. I need to check if MSYS supports -lz

There is more issues with Windows. For example the last separate Windows binary is zlib 1.2.3:
http://gnuwin32.sourceforge.net/packages/zlib.htm

zlibWrapper/zstd_zlibwrapper.c
Cyan4973: OK, so it's a copy, not a reference to parent.

If `zalloc`, `zfree` or `opaque` are changed after `deflateInit()`, the changes will not be taken in consideration.
No allocation is needed outside of `deflateInit()` anyway,  but `zfree` will be needed at the end. Hopefully, it seems logical to not "change" `zfree` nor `opaque` between `deflateInit()` and `deflateEnd()`. Maybe a documentation warning on this fact would help.

It also means that usage of `ZWRAP_CCtx.allocFun` is fairly limited. Just once, when calling `ZWRAP_createCCtx()`.

My own choice would have been to create a structure which copy only the needed elements, `zalloc`, `zfree` and `opaque`. But I guess it's mostly a matter of coding preference.

zlibWrapper/Makefile
Cyan4973: Fine. So I guess an `echo` warning message will help.

zlibWrapper/zstd_zlibwrapper.c
inikep: `ZWRAP_CCtx.allocFun` is used here
`ZSTD_customMem ZWRAP_customMem = { ZWRAP_allocFunction, ZWRAP_freeFunction, &zwc->allocFunc };`
It means it's also needed for `ZWRAP_freeCCtx()`.

Cyan4973: Right.
By the way, `ZWRAP_freeCCtx()` still uses `free()` directly, so I guess a fix is needed there.

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -31,11 +31,13 @@
 
 #include <stdio.h>           /* fprintf */
 #include <stdlib.h>          /* malloc */
```
Cyan4973: If no function call `malloc()` nor `free()` directly, it should be possible to remove this `#include`
(well, if `stdlib` is not included either in any other include. But that should be the case too).
It will also help to spot if any invocation of `free()` is forgotten somewhere.

zlibWrapper/Makefile
inikep: I checked and MinGW/MSYS2 supports -lz so there is no need for "ifneq (,$(filter Windows%,$(OS))".

BTW, do we support CygWin compilation of zstd? It's installed on Travis so we use it.

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -31,11 +31,13 @@
 
 #include <stdio.h>           /* fprintf */
```
Cyan4973: `stdio` is needed for debug,
but also for `FINISH_WITH_ERR` macros.

See comments below

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -500,6 +531,7 @@ ZEXTERN int ZEXPORT z_deflatePending OF((z_streamp strm,
         return deflatePending(strm, pending, bits);
     FINISH_WITH_ERR("deflatePending is not supported!");
```
Cyan4973: Is that an expected behavior from zlib to write to `stderr` by default ?
It looks like a CLI functionality, rather than a system library one.

zlibWrapper/Makefile
Cyan4973: I think it's fair to support a limited number of systems.
After all, it's only an example.

Just be clear when a system is not supported, so that it looks like a known limitation rather than an error.

inikep: I was asking about whole zstd package. Do we support CygWin compilation? I didn't found any code that detects CygWin so it may throw compilation errors.

Cyan4973: Fair point.
The initial intention was to support it, but the goal was postponed because there used to be no way to test it properly.

Maybe that action could be resurrected thanks to the great improvements you brought to CI test suite. If CygWin target can be selected, then it becomes possible to test compilation and run.

In any case, don't consider it a primary objective. It's more a "nice to have" target for the time being.

zlibWrapper/zstd_zlibwrapper.c
inikep: Indeed zlib does something like this:
`strm->msg = (char *)"incorrect length check";`

from zlib.h:
In the error case, msg may be set but then points to a static string (which must not be deallocated).

What should do zlibWrapper in case of error? Which error code should be returned? The list:
```
#define Z_OK            0
#define Z_STREAM_END    1
#define Z_NEED_DICT     2
#define Z_ERRNO        (-1)
#define Z_STREAM_ERROR (-2)
#define Z_DATA_ERROR   (-3)
#define Z_MEM_ERROR    (-4)
#define Z_BUF_ERROR    (-5)
#define Z_VERSION_ERROR (-6)
```

Cyan4973: Well, the precise error code to use is obviously situation dependent.

Currently, it seems you selected `Z_MEM_ERROR` for all errors within `FINISH_WITH_ERR` macros.

I suspect `Z_ERRNO` to be the "default error" when not specific,
`Z_MEM_ERROR` more likely related to memory management issues,
`Z_DATA_ERROR` more likely related to corruption issues,
and `Z_VERSION_ERROR` potentially related to unauthorized header fields or values.
No idea for `Z_BUF_ERROR` or `Z_STREAM_ERROR`.

In any case, the precise error number is at this stage less important than removing the `fprintf` .

inikep: OK I will remove fprintf.

Z_ERRNO = If an error occurred in the file system and not in the compression library, errnum is set to Z_ERRNO and the application may consult errno to get the exact error code.

Z_STREAM_ERROR = if the stream state was inconsistent

Z_BUF_ERROR if no progress is possible (for example avail_in or avail_out was zero) or if there was not enough room in the output buffer when Z_FINISH is used.

zlibWrapper/zstd_zlibwrapper.c
inikep: I think I fixed all problems. This takes 10 commits. When I will publish them they will be in tihs (PR #198). Do you prefer them in a separate PR?

Cyan4973: No, let's review them here.
I think the point is to completely finish one feature.
So we keep every correction commit into the same PR.

Only unrelated commits get into a separate PR.

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -29,13 +29,14 @@
     - zstd source repository : https://github.com/Cyan4973/zstd
 */
 
-#include <stdio.h>           /* fprintf */
-#include <stdlib.h>          /* malloc */
+#include <stdarg.h>          /* va_list */
```
Cyan4973: minor point :
I guess `stdarg` is required because of `LOG_WRAPPER`,
which is itself only necessary during debug.

So it would be better to group them together and activated on condition of some specific debug flag.

The goal here is to reduce external dependencies to a minimum, to improve portability perspectives.

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -630,9 +692,9 @@ ZEXTERN int ZEXPORT z_compress OF((Bytef *dest,   uLongf *destLen,
     if (!g_useZSTD)
         return compress(dest, destLen, source, sourceLen);
 
-    size_t dstCapacity = *destLen; 
-    LOG_WRAPPER("z_compress sourceLen=%d dstCapacity=%d\n", (int)sourceLen, (int)dstCapacity);
-    { size_t const errorCode = ZSTD_compress(dest, dstCapacity, source, sourceLen, -1);
+    { size_t dstCapacity = *destLen; 
+      size_t const errorCode = ZSTD_compress(dest, dstCapacity, source, sourceLen, -1);
```
Cyan4973: Why a compression level of `-1` ?
Why not `ZWRAP_DEFAULT_CLEVEL` ?

Cyan4973:
Almost there :clap: 
One minor point on an include, and a question on a default value.

zlibWrapper/zstd_zlibwrapper.c
inikep: va_list is used by z_gzprintf not by LOG_WRAPPER

zlibWrapper/zstd_zlibwrapper.c
inikep: I missed that :)

Cyan4973:
merged

Issue #201, opened 2016-06-06
[0.7.0] No more VS2013 project makes it difficult to target a MS VC Runtime above 10.0


KrzysFR:
I need to build the Windows library targetting the MS VC Runtime 12.0 (the one that came out with VS2013, which maps to MSVCR120.dll), but in the latest dev 0.7.0 branch, there is only two projects left: one for Visual Studio 2008 and one for Visual Studio 2010 (targetting MSVCR100.dll). The project for VS2013 is gone. I'm not sure how to use the CMake stuff, but after a quick look, it does not seem to have code that specifically deal with the version of the VC runtime in it.

Usually, each developper will use version X of Visual Studio (2010, 2013, 2015, 15/vNext, ....), while targetting a version Y of the Microsoft VC Runtime (10.0, 11.0, 12.0, 14.0, ..., with Y <= X). I guess most dev may use a more recent version of VS than the version of the runtime they target, because retargetting all your dependencies to the latest VC runtime can be a lot of work, or even not possible at all. This means that the version of the VC runtime may not be the same as the version of Visual Studio project, and will be different for each person.

Currently, if you have VS 2015 Update 2, and open the VS2010 project in the dev branch, VS attempts to convert the project to the latest version possible (14.0, which will probably break because of some includes that are different). And if you don't perform the conversion, then the build will probably fail because you don't have the 10.0 SDK installed on your dev machine or CI build server (unless you have also installed VS2010 before). Plus, if you want a different version (12.0 for me), you then need to update each project (for all the projects, times two for Release/Debug, times two again for Win32/x64). And then you are left with a change in the .vcproj that will cause trouble when you update from git again (forcing you to maintain a custom branch).

I can see that having to maitain one set of projects for each versions of Visual Studio can be too much work, but what do you think would be the best way to be able to specify which version of the VC runtime to target when building? Maybe a settings for CMake that would then create a set of VS projects specific to you, and not checked in the repo? (this would be ideal).

KrzysFR:
To be more precise, I'm talking about the `PlatformToolset` field in the `.vcxproj`, which I had to add to all configurations of all projects.

```XML
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
    <ConfigurationType>DynamicLibrary</ConfigurationType>
    <UseDebugLibraries>true</UseDebugLibraries>
    <CharacterSet>MultiByte</CharacterSet>
[...]

PR #202, opened 2016-06-07
Added build commands for various versions of Visual Studio

KrzysFR:
This PR adds a set of build commands for Windows, intended for people who want to quickly build `zstd.exe` or `zstdlib_*.dll` without needing to open Visual Studio and fight with the project conversion process (see Issue #201 for the context), and is inspired by the build scripts used in [GitExtensions](https://github.com/gitextensions/gitextensions/blob/aefed53bb6879658e3eb22322dcb9cae93a05547/Build/build.generic.cmd)

- `build.generic.cmd` is the main build script that runs msbuild on the Visual Studio project, for a specific configuration (ex: Release, x64, v120).
- `build.VSxxx.cmd` are helper scripts that call `build.generic.cmd` with a set of default settings for this particular version of Visual Studio.

For example, somone with Visual Studio 2013 installed, would run `build.VS2013.cmd` and get a version of `zstd.exe` compiled with msbuild v12.0 and targeting MSCVR120.dll. Likewise, `build.VS2015.cmd` would build the project using msbuild v14.0 / MSCVR140.dll.

People checking out the repository, and just wanting to build `zstd.exe` or `zstdlib_xxx.dll` to play with it, will only need to run the corresponding build.VSxxx.cmd (matching the version they have installed), and be done.

People who need fine grained settings, or wanting to integrate this with a build server, would probably `build.generic.cmd` with the exact set of arguments they want...or more probably call msbuild themselves...

__TO BE DECIDED / NOT TESTED YET__

- The result of the build is stored in `build/bin/{VS_VERSION}/{ARCH}/` so for example `build/bin/VS2013/x64/`, instead of under each of the `projects/VS2010/{PROJECT}/x64/Release` folders. This makes it simpler to access imho because you are already in the build folder.

- I call msbuild with `Clean,Build` target, which has the side effect of clearing everything under `build/bin` for some reason. This wipes previous builds (especially if you build x86 then x64, then the x86 build gets wiped, which is unfortunate). Don't know how to prevent that.

- Currently, `build.VSxxx.cmd` build both Win32 and x64 build at the same time. If this is not desirable, I would need to duplicate all scripts, and have `build.VS2013.Win32.cmd` and `build.VS2013.x64.cmd`. I don't know what is best.

- There is no way to choose between statically linked vs dynamically linked. We would need to add another argument to `build.generic.cmd` for that.

- I was not able to test the VS2010 and VS2012 build scripts because I don't have them installed anywhere, so I don't know if these scripts work.

- No integration with CMake... is this needed? If CMake would create version of the projects specific to VS20xx, then maybe the `build.VS20xx.cmd` count use the path to that project, instead of defaulting to the VS2010 project?

### DIFF
--- a/build/build.generic.cmd
+++ b/build/build.generic.cmd
@@ -1,33 +1,46 @@
 @echo off
 
-rem
-rem Parameters
-rem   %1: vs_solution_version: VS solution version: VS2012, VS2013, VS2015
-rem   %2: vs_configuration:    VS configuration:    Debug, Release
-rem   %3: vs_target:           target:              Build or Rebuild
-rem   %4: vs_platform:         platform:            x64 or Win32
-rem   %5: vs_toolset:          toolset:             v100, v110, v120, v140
-rem
+IF "%1%" == "" GOTO display_help
 
-set vs_solution_version=%1
-set vs_configuration=%2
-set vs_target=%3
-set vs_platform=%4
-set vs_toolset=%5
+SET vs_version=%1
 
-set msbuild="%windir%\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe"
-IF %vs_solution_version% == VS2013 SET msbuild="C:\Program Files (x86)\MSBuild\12.0\Bin\MSBuild.exe"
-IF %vs_solution_version% == VS2015 SET msbuild="C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe"
+SET vs_platform=%2
+IF "%vs_platform%" == "" SET vs_platform=x64
+
+SET vs_configuration=%3
+IF "%vs_configuration%" == "" SET vs_configuration=Release
+
+SET vs_toolset=%4
+
+GOTO build
+
+:display_help
+
+echo Syntax: build.generic.cmd vs_version vs_platform vs_configuration vs_toolset
+echo   vs_version:          VS installed version (VS2012, VS2013, VS2015, ...)
+echo   vs_platform:         Platform (x64 or Win32)
+echo   vs_configuration:    VS configuration (Release or Debug)
+echo   vs_toolset:          Platform Toolset (v100, v110, v120, v140)
+
+EXIT /B 1
+
+:build
+
+SET msbuild="%windir%\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe"
+IF %vs_version% == VS2013 SET msbuild="C:\Program Files (x86)\MSBuild\12.0\Bin\MSBuild.exe"
+IF %vs_version% == VS2015 SET msbuild="C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe"
 rem TODO: Visual Studio "15" (vNext) will use MSBuild 15.0 ?
 
-set project="%~p0\..\projects\VS2010\zstd.sln"
+SET project="%~p0\..\projects\VS2010\zstd.sln"
 
-set output=%~p0%bin
-set output="%output%/%vs_solution_version%/%vs_platform%/"
+SET msbuildparams=/verbosity:minimal /nologo /t:Clean,Build /p:Platform=%vs_platform% /p:Configuration=%vs_configuration%
+IF NOT "%vs_toolset%" == "" SET msbuildparams=%msbuildparams% /p:PlatformToolset=%vs_toolset%
 
-set msbuildparams=/verbosity:minimal /p:Platform="%vs_platform%" /p:Configuration="%vs_configuration%" /p:PlatformToolset="%vs_toolset%" /t:Clean,Build /nologo /p:OutDir=%output%
+SET output=%~p0%bin
+SET output="%output%/%vs_configuration%/%vs_platform%/"
+SET msbuildparams=%msbuildparams% /p:OutDir=%output%
 
-echo ### Building %vs_solution_version% project for %vs_configuration% %vs_platform% (%vs_toolset%)...
+echo ### Building %vs_version% project for %vs_configuration% %vs_platform% (%vs_toolset%)...
 echo ### Build Params: %msbuildparams%
 
 %msbuild% %project% %msbuildparams%
--- a/build/build.generic.cmd
+++ b/build/build.generic.cmd
@@ -1,0 +1,38 @@
+@echo off
+
+rem
+rem Parameters
+rem   %1: vs_solution_version: VS solution version: VS2012, VS2013, VS2015
+rem   %2: vs_configuration:    VS configuration:    Debug, Release
+rem   %3: vs_target:           target:              Build or Rebuild
+rem   %4: vs_platform:         platform:            x64 or Win32
+rem   %5: vs_toolset:          toolset:             v100, v110, v120, v140
+rem
+
+set vs_solution_version=%1
+set vs_configuration=%2
+set vs_target=%3
+set vs_platform=%4
+set vs_toolset=%5
+
+set msbuild="%windir%\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe"
+IF %vs_solution_version% == VS2013 SET msbuild="C:\Program Files (x86)\MSBuild\12.0\Bin\MSBuild.exe"
+IF %vs_solution_version% == VS2015 SET msbuild="C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe"
+rem TODO: Visual Studio "15" (vNext) will use MSBuild 15.0 ?
+
+set project="%~p0\..\projects\VS2010\zstd.sln"
+
+set output=%~p0%bin
+set output="%output%/%vs_solution_version%/%vs_platform%/"
+
+set msbuildparams=/verbosity:minimal /p:Platform="%vs_platform%" /p:Configuration="%vs_configuration%" /p:PlatformToolset="%vs_toolset%" /t:Clean,Build /nologo /p:OutDir=%output%
+
+echo ### Building %vs_solution_version% project for %vs_configuration% %vs_platform% (%vs_toolset%)...
+echo ### Build Params: %msbuildparams%
+
+%msbuild% %project% %msbuildparams%
+IF ERRORLEVEL 1 EXIT /B 1
+echo # Success
+echo # OutDir: %output%
+echo #
+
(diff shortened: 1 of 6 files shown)


Cyan4973:
> The result of the build is stored in build/bin/{VS_VERSION}/{ARCH}/ 

Looks good to me

> I call msbuild with `Clean,Build` target (...) This wipes previous builds

This one is probably worth fixing.
Not sure if it makes sense, but what about just `Build` ?

> Currently, build.VSxxx.cmd build both Win32 and x64 build at the same time

Looks good enough to me.
Maybe later someone will complain about targeting ARM instead.
No urgency on this item.

> There is no way to choose between statically linked vs dynamically linked.

This is good enough for a first version.

> I was not able to test the VS2010 and VS2012 build scripts 

It should be possible to download and install VS2012 and VS2010 express to test.

> No integration with CMake... is this needed?

Not for a first version.
It's already a significant step forward.
Let's wait and see if this new capability generates feature requests.

build/build.generic.cmd
```diff
@@ -0,0 +1,38 @@
...
+rem   %5: vs_toolset:          toolset:             v100, v110, v120, v140
+rem
+
+set vs_solution_version=%1
+set vs_configuration=%2
+set vs_target=%3
+set vs_platform=%4
+set vs_toolset=%5
+
+set msbuild="%windir%\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe"
```
Cyan4973: The path to `msbuild` is hard-coded, which feels strange.
Could there be a better way ?
http://stackoverflow.com/questions/328017/path-to-msbuild

KrzysFR:
> > I call msbuild with Clean,Build target (...) This wipes previous builds
> This one is probably worth fixing.
> Not sure if it makes sense, but what about just Build ?

The only issue I could see, is that `Build` compares the last modified date of the source files vs the the files in the build folder, to decide to rebuild the project or not: When changing branch, or reverting to previous commits, you sometimes get source files that jump back into the past, while the existing .exe/.dll files remaining unchanged. The next time msbuild runs, it will think that they have already been build and skip it. Maybe `Rebuild` target could help here? It's not clear to me what exactly Build and Rebuild do differently in this case. I got into the habit of always cleaning first, to prevent any mistakes like that.

Another issue is if you sudenly decide to rename `zstd.exe` to `zorglub.exe` (because why not!), then without a `Clean`, people will end up with both exe in their existing build folder. I had issues where build scripts would copy all files to the final build artefact, and got all files included as part of the new version.

build/build.generic.cmd
KrzysFR: I copied and adapted what was done in GitExtensions.

I'm not exactly sure how I could read a value from the registry and insert it into a variable, from a `.cmd` script...

build/build.generic.cmd
```diff
@@ -0,0 +1,38 @@
...
+IF %vs_solution_version% == VS2013 SET msbuild="C:\Program Files (x86)\MSBuild\12.0\Bin\MSBuild.exe"
+IF %vs_solution_version% == VS2015 SET msbuild="C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe"
```
Cyan4973: There are specific `SET` for `VS2013` and `VS2015`.
What about `VS2012` and `VS2010` ?

build/build.generic.cmd
```diff
@@ -0,0 +1,38 @@
+@echo off
+
+rem
+rem Parameters
```
Cyan4973: If someone calls `build.generic.cmd` with no argument, or the wrong nb of arguments, or some other easily detectable error,
could it be possible to display a help text ?

I feel your comments below are pretty good, they could just be `echo` to the console

KrzysFR:
> It should be possible to download and install VS2012 and VS2010 express to test.

The real test would to be have a fresh install of Windows, install ONLY Visual Studio 2012, and verify that `build.VS2012.cmd` runs fine. Repeating this test for all versions of VS from 2008 to 2015 and even 15.0/vNext...Not a trivial amount of work :)

> I call msbuild with `Clean,Build` target (...) This wipes previous builds

I replaced `Clean,Rebuild` by `Rebuild`, and it does not change much (which is [expected](http://stackoverflow.com/questions/1247457/difference-between-rebuild-and-clean-build-in-visual-studio-2008))

But what is strange, is that building the VS2013 for Win32 will clean everything under `build\bin\**\Win32`, but keep intact everything under `build\bin\**\x64\`, which means that there must be some pattern that msbuild recognized with `.../bin/{SOMETHING}/{ARCH}` while building.

Maybe, building each version in a different `VSxxx` subfolder is overkill, and not required? I could simplify and have them build in `build\bin\(Release|Debug)\(Win32|x64)` instead, and get rid of the `VS20xx` subfolder?

build/build.generic.cmd
KrzysFR: I *think* that before VS2013, msbuild was part of the .NET framework, and thus installed under the .NET Framework build folder. Starting from VS2013, they shipped it as part of VS , under its own folder (I remember TeamCity was having problems with VS2013 initially, because of that change, and I had invoke msbuild myself for a while)

See https://blogs.msdn.microsoft.com/visualstudio/2013/07/24/msbuild-is-now-part-of-visual-studio/

build/build.generic.cmd
KrzysFR: Since there are 5 (possibly more) arguments, I could at least `echo` the commands if there are NO arguments, but maybe replace the others with default values? I need to brush up on my batch skills, it's been a long time :)

Cyan4973:
> Maybe, building each version in a different VSxxx subfolder is overkill, and not required? I could simplify and have them build in build\bin\(Release|Debug)\(Win32|x64) instead, and get rid of the VS20xx subfolder?

I liked the initial proposition of `build/bin/{VS_VERSION}/{ARCH}/`,

The hidden assumption was that building for a selected `VS_VERSION1` would have no impact on a different `VS_VERSION2` directory. It's surprising to learn that building into `build/bin/{VS_VERSION1}` wipes out `build/bin/{VS_VERSION2}` content.

To be fair, I'm not sure to understand anymore what is being wiped when calling the batch.

KrzysFR:
If you have build VS2013, you have both `build/bin/VS2013/Win32/*` and `build/bin/VS2013/x64` with exe/dll.

Now, you build VS2015. First when it run the Win32 builds, it creates an empty `build/bin/VS2015/Win32` folder, and at the same time deletes everything in `build/bin/VS2013/Win32/' folder, but does not touch `build/bin/VS2013/x64`.

Anyway, I made some changes to output in `build/bin/Release/{ARCH}/` which solves the problem since there's only one folder per target.

Maybe people who want to build for multiple version of VS *at the same time* are not the most common scenario?

Cyan4973:
OK, let's try that

KrzysFR:
I pushed the following changes:
- add message with syntax if no arguments
- except the VS Version, all other arguments are optional (with default version)
- remove target which is always `Clean,Rebuild`
- changed output path to `build/bin/{CONFIG}/{ARCH}` (which is closer to what VS would do)

What is still to be decided is if we keep the script that build woth x64+Win32 at the same time, or duplicate them into x64 and Win32 scripts for each VS version.

Cyan4973:
Let's keep the simpler proposition, and build x64 + win32 at the same time.

KrzysFR:
While I am it, I found the following differences while targetting v120 or v140 (I don't have v100 so can't test it)

- for `v120`, zstd.exe (x64) takes **378 KB** and only depends on KERNEL32.dll. PDB size is **2.1 MB**
- for `v140`, zstd.exe (x64) takes **395 KB** and depends on KERNEL32.dll and ADVAPI32.dll for only a single mysterious `SystemFunction036` method (0x031A), whatever that is. PDB size is **4.5 MB** (!!). There are also more files in the output folders (`.iobj` and `.ipdb` files)

The v14.0 or MSCVRT seems to be a lot bigger than the previous ones... This could be a problem if you check in both dll and pdb files in your git repository (like I do)...

The result of building with VS2015 but targetting for `v120` produces the same result as building with VS2013, which means that the version of the compiler is less important than the version of the target runtime.

When attempting to build with a version of VS2010 when it is not installed produces the following explicit error message:
```
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cpp.Platform.targets(42,5): error MSB8020:
The builds tools for Visual Studio 2010 (Platform Toolset = 'v100') cannot be found.
To build using the v100 build tools, either click the Project menu or right-click the solution, and then select "Update VC++ Projects...".
Install Visual Studio 2010 to build using the Visual Studio 2010 build tools.
[C:\Data\Git\GitHub\zstd\projects\VS2010\zstd\zstd.vcxproj]
```

But attempting to build for VS2012 which is not installed either fails at LINK time with:
```
LINK : fatal error LNK1181: cannot open input file 'gdi32.lib' [C:\Data\Git\GitHub\zstd\projects\VS2010\zstd\zstd.vcxproj]
```

Attempting to target v140 but with v2013 produces the following error messages:
```
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.Cpp.Platform.targets(64,5): error MSB8020:
The build tools for v140 (Platform Toolset = 'v140') cannot be found.
To build using the v140 build tools, please install v140 build tools. 
Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Upgrade Solution...".
[C:\Data\Git\GitHub\zstd\projects\VS2010\zstd\zstd.vcxproj]
```

Cyan4973:
No binary shall be checked in the git repository.

Also :
It seems that `SystemFunction036` == `RtlGenRandom`
https://msdn.microsoft.com/fr-fr/library/windows/desktop/aa387694(v=vs.85).aspx

KrzysFR:
> No binary shall be checked in the git repository.
I was not talking about the zstd repository :)

> It seems that SystemFunction036 == RtlGenRandom
Ok, my version of depends.exe may not be up to date. I guess they fixed a potential flow in the current random implementation in v140.

The only bad news is the huge size of the pdb produces with the latest version of MSVCRT: 4.5 MB for zstd.exe, and 3.0 MB for zstdlib.

But of course, if we run `zstd -9` on the library pdb, it gets down to 452 KB :)

KrzysFR:
I couldn't resist:
```
C:\Data\Git\GitHub\zstd\build\bin\Release\x64>zstd.exe zstd.exe
Compressed 404480 bytes into 215293 bytes ==> 53.23%
```

PR #207, opened 2016-06-12
Fix hardcoded path in msbuild scripts

KrzysFR:
This address feedback from #201 by fixing the hardcoded path to `Program Files (X86)`, as well as fixing a leak of the variables defined in the script by adding `SETLOCAL`

- Whether running on 32-bit, 64-bit or WoW64, `%programfiles(x86)%` returns `C:\Program Files (x86)` by default.
- Without `SETLOCAL`, the script would leak all variables defined via `SET` in the CMD session, even after the scripts exists.

### DIFF
--- a/build/build.generic.cmd
+++ b/build/build.generic.cmd
@@ -2,48 +2,50 @@
 
 IF "%1%" == "" GOTO display_help
 
-SET vs_version=%1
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good to me

PR #209, opened 2016-06-14
Removed old xxx_static.h header references from VS2010 project

KrzysFR:
Removed reference to the old `..._static.h` files from the `.vcxproj` of the VS2012 solution.

VS does not seem to complain when a reference .h is missing (only when you click on the file in the Solution Explorer to open it), but I figured it would be better if they are removed.

Note that I updated the project with VS2015, which also updated the `.vcxproj.filters` files to add some stuff (maybe they were out of sync?). I think the `.vcxproj.filters` files are not used by msbuild, so it should not impact the command line build.

### DIFF
--- a/projects/VS2010/fullbench/fullbench.vcxproj
+++ b/projects/VS2010/fullbench/fullbench.vcxproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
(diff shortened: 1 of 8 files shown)


Cyan4973:
> Removed old xxx_static.h header references

That's a good point !

> Note that I updated the project with VS2015, which also updated the .vcxproj.filters files to add some stuff (maybe they were out of sync?). 

The objective is to have project files compatible from VS2010 to VS2015 (through automatic upgrade process). It seems also necessary for appveyor tests, which check compatibility with all these compilers.

Let's see if it still works as intended...

Cyan4973:
I could check that Visual projects can still be opened and upgraded using Visual 2012.

Appveyor tests seem to work fine too.

So it looks fine from my perspective

PR #210, opened 2016-06-15
zlibWrapper: improved memory deallocation in case of error

inikep:


Cyan4973:
Looks all good

Cyan4973:
Apparently this PR triggered an unrelated corruption error inside `fuzzer` tests.

To be investigated ...

PR #211, opened 2016-06-15
Fix Max Compression Level in zstd.1

tobijdc:
This is also wrong in master.

The file seems outdated, the "--ultra" parameter is also not listed.
Is this man-page still used?


Cyan4973:
Fair enough.

It's always a problem to manually keep updated multiple separated documentation.
`man` page format (troff) is also pretty poor. It's far from the simplicity of markdown.
At least, within `zstd -h`, it's possible to use program constants to have automatic update of parameters such as max compression level.

Thanks for the update Tobias.

PR #214, opened 2016-06-16
fixing a segfault

velavokr:
The commit fixes a crash in the following code path:

The crash was caused by an out-of-memory access here:
```c
void ZSTD_seqToCodes(const seqStore_t* seqStorePtr, size_t const nbSeq)
{
...
        size_t u;
        for (u=0; u<nbSeq; u++) {
            U32 const  ll = llTable[u];
            llCodeTable[u] = (ll>63) ? (BYTE)ZSTD_highbit(ll) + LL_deltaCode : LL_Code[ll];
        }
```

The gdb showed a really huge nbSeq originating from here:
```c
    const U32*  const offsetTable = seqStorePtr->offsetStart;
    const U32*  const offsetTableEnd = seqStorePtr->offset;
...
    size_t const nbSeq = offsetTableEnd - offsetTable;
```

The `offsetTableEnd` was zero. The AddressSanitizer showed that the buffer was malloc'ed in `ZSTD_resetCCtx_advanced`. The initialization code seemed unfinished and after a brief examination of the surroundings I found `ZSTD_resetSeqStore` function which I believe does the right thing (and fixes the issue for me).

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -325,7 +325,7 @@ static size_t ZSTD_resetCCtx_advanced (ZSTD_CCtx* zc,
     zc->seqStore.mlCodeStart = zc->seqStore.llCodeStart + maxNbSeq;
     zc->seqStore.offCodeStart = zc->seqStore.mlCodeStart + maxNbSeq;
     zc->seqStore.litStart = zc->seqStore.offCodeStart + maxNbSeq;
-
+    ZSTD_resetSeqStore(&zc->seqStore);
     zc->stage = 1;
     zc->dictID = 0;
     zc->loadedDictEnd = 0;


Cyan4973:
Thanks for notification and fix @velavokr .
Indeed, looks like an initialization issue.

`ZSTD_resetSeqStore()` is supposed to be invoked at the beginning of every block.
See : 
https://github.com/Cyan4973/zstd/blob/master/lib/compress/zstd_compress.c#L1131
https://github.com/Cyan4973/zstd/blob/master/lib/compress/zstd_compress.c#L1242
https://github.com/Cyan4973/zstd/blob/master/lib/compress/zstd_compress.c#L1746
https://github.com/Cyan4973/zstd/blob/master/lib/compress/zstd_compress.c#L1916
https://github.com/Cyan4973/zstd/blob/master/lib/compress/zstd_opt.h#L468
https://github.com/Cyan4973/zstd/blob/master/lib/compress/zstd_opt.h#L760

So it's weird that it was missed in your case, since it seems to be invoked on all paths.
Maybe there is a specific combination that misses it ?
What's the compression level used and the size of the sample to compress ?
Is there any sample that could help observe the issue ?


I'm impressed you could find such a bug so fast
while our fuzzer tools where unable to find it in several days of testing.
Would you mind sharing how you found it ?

Cyan4973:
I made a small change in the "dev" branch :
https://github.com/Cyan4973/zstd/tree/dev

Rather that initializing at the beginning of the compression function, initialization happens at dispatch point. So it's declared once instead of 6 times, making it easier to ensure initialization wasn't forgotten.

That being said, it's unclear if it will help your situation, as I still don't know how it evaded initialization to begin with. At this stage, I'm unfortunately unable to reproduce the bug.

velavokr:
Thanks a lot, I'll check it. As for the dataset to reproduce the issue, it will take me some time to dump it in a usable form. And it was indeed difficult to get into trouble, I got there only by shuffling the otherwise benigh input.What I was trying to do is to use your "external dictionary" mode with "precalculated contexts" on some internal protobuf messages to save us some cpu and traffic. The compression level was 1 and I did some preprocessing to improve compression while not losing in the decompression speed (basically, encoding frequent strings as varint8 numbers).BTW, thanks for lz4 and zstd, we use them heavily here. The compression/speed ratio is indeed amasing and I was happy to abandon my own coding library (although the dictionary part of it proved useful as a preprocessing step to zstd). Great job!17.06.2016, 14:00, "Yann Collet" <notifications@github.com>:I made a small change in the "dev" branch :
https://github.com/Cyan4973/zstd/tree/dev

Rather that initializing at the beginning of the compression function, initialization happens at dispatch point. So it's declared once instead of 6 times, making it easier to ensure initialization wasn't forgotten.

That being said, it's unclear if it will help your situation, as I still don't know how it evaded initialization to begin with. At this stage, I'm unfortunately unable to reproduce the bug.

Cyan4973:
You're very welcomed @velavokr

>  I was trying to do is to use your "external dictionary" mode with "precalculated contexts" on some internal protobuf messages

It seems your use case is indeed quite advanced (aka, it's not a simple call to `ZSTD_compress()`),

As a consequence, I suspect that access to bug sample is just one part.
Another important part is the sequence of invocations used in your code.
There's a fair chance that the specific sequence you use wasn't tested properly, in which case it would deserve an update to our test suite.

Cyan4973:
Coming back to this :

Do you have any more information on how to reproduce this bug ?

velavokr:
+ reply@- notifications@Hi! It is indeed hard to reproduce, yesterday I changed slightly the sampling code (swapped mersenne for pcg) and the effect disappeared completely, so I had to revert some code to get back to the crashing version. Hope these details are enough to reproduce it: https://yadi.sk/d/yUucdADdsgxNg 21.06.2016, 01:07, "Yann Collet" <notifications@github.com>:Coming back to this :Do you have any more information on how to reproduce this bug ?—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.

velavokr:
Oh, I forgot the most important detail: we are using 0.6.0.

So I couldn't check your fix against it.

Cyan4973:
> Oh, I forgot the most important detail: we are using 0.6.0.

Ah, yes, indeed, it's an important point.

At the very least, I would suggest an update to 0.6.1, which should prove friction-less :
format is exactly the same, interfaces are preserved,
and it solves quite a bunch of bugs, making it overall a more solid release.


> Hope these details are enough to reproduce it

My understanding is that you are using dictionary compression,
and that the reported bug happens during dictionary builder phase.

Indeed, it's a very different code path than compression / decompression itself.
The dictionary builder is unfortunately still in "beta" shape,
so it's not surprising to find issues there.

To solve the problem, it is necessary to reproduce and observe the issue.
To do that, I would need a way to rebuild the dictionary the same way as you do.

It seems the sample pool consists of 3712 files, of total size 1,432,121 bytes.
But unfortunately, in order to reproduce the bug, it's necessary to have access to the 3712 separated files, and in the same order.
A `tar` archive would be ideal for that goal.

velavokr:
> But unfortunately, in order to reproduce the bug, it's necessary to have access to the 3712 separated files, and in the same order.
There are lengths serialized as 64bitLE numbers in the file named ZDICT_trainFromBuffer_samplesSizes.64bitLE.bin

This is the exact input I gave to ZDICT_trainFromBuffer.

Cyan4973:
> There are lengths serialized as 64bitLE numbers in the file named ZDICT_trainFromBuffer_samplesSizes.64bitLE.bin

OK, got it.

I'm surprised as the gdb trace seems to show console result from `zstd --train`,
but it could also happen if you raised the notification level.

velavokr:
Here is the minimal working example to reproduce the crash on the supplied data

https://gist.github.com/velavokr/6dbb7add70a5051184309f9848b50cef

velavokr:
Yes, I found that knob recently :)

Cyan4973:
Thanks to all your elements, I can perfectly reproduce the bug.
On to fixing it now ....

Cyan4973:
Latest update in the "dev" branch should solve this issue :
https://github.com/Cyan4973/zstd/commit/23042929daf5fb207bfadaffdd85325bfbd16e24

The problem was on dictBuilder side.

The fix will be part of v0.7.1

Cyan4973:
Fix pushed to master

PR #216, opened 2016-06-21
v0.7.1

Cyan4973:
hotfix


KrzysFR:
Version in `common/zstd.h` is still `0.7.0`, is this normal?

Cyan4973:
Good point !

PR #234, opened 2016-07-03
Redundant entry for options in the man page.

juanfra684:
".SH OPTIONS" is enough.


Cyan4973:
Thanks @juanfra684 .
You are correct.

Could you please redirect your patch into the "dev" branch instead ?
Direct PR to master are not permitted by our [Branch Policy](https://github.com/Cyan4973/zstd#branch-policy).

Alternatively, I can also apply your patch manually if you prefer.

PR #240, opened 2016-07-08
Changed : Move fse_decompress.c from common/ to decompress/

gymdis:
For consistency.


Cyan4973:
ah, unfortunately, it was not a mistake.

"fse_decompress.c" is needed for compression,
since it's part of the dictionary module.

So it's really "common", from a `zstd` perspective.
"common" regroup files required for both compression and decompression.

gymdis:
Oh no, my bad, I didn't realize that. :)

Could the common parts be extracted to a new "fse_common.c" or would there not be anything left?

Cyan4973:
There would be nothing left.
The entire decompression module is required.

Maybe we could rename `fse_decompress.c` into `fse_common.c`,
but we also try to keep in sync with [FSE directory](https://github.com/Cyan4973/FiniteStateEntropy)
and from within FSE context, it's really a decompression module, so it makes sense to keep this name.

PR #247, opened 2016-07-14
Adding brew formula for easy install

cadedaniel:
I am not sure how popular zstd is yet, thus we can have this Brew formula live here until the brew team accepts it in their repo.

Feedback welcome.

### DIFF
--- a/zstd.rb
+++ b/zstd.rb
@@ -1,0 +1,18 @@
+class Zstd < Formula
+  desc "Zstandard - Fast real-time compression algorithm"
+  homepage "http://www.zstd.net/"
+  url "https://github.com/Cyan4973/zstd/archive/v0.7.3.tar.gz"
+  sha256 "767da2a321b70d57a0f0776c39192a6c235c8f1fd7f1268eafde94a8869c3c71"
+
+  def install
+    system "make", "install", "PREFIX=#{prefix}"
+  end
+
+  test do
+    (testpath/"input.txt").write("Hello, world." * 10)
+    system "#{bin}/zstd", "input.txt", "-o", "compressed.zst"
+    system "#{bin}/zstd", "--test", "compressed.zst"
+    system "#{bin}/zstd", "-d", "compressed.zst", "-o", "decompressed.txt"
+    system "cmp", "input.txt", "decompressed.txt"
+  end
+end


chipturner:
Tested on my mac with:

```
brew tap cadedaniel/zstd zstd
brew install zstd
```

it worked!  once merged, the 'tap' command would change to point to the main repo, but otherwise should work identically.

@cadedaniel how do we update/maintain this file?  I'm glad to update it as we make releases; is there a command to sync it from the tarball?

cadedaniel:
Thanks @chipturner for validating that it works on a machine that's not my own :)

The idea is that this gets updated with every `zstd` release. To update the formula you need do two things; update the URL and the SHA hash.

To point the url to the new tarball release:
`url "https://github.com/Cyan4973/zstd/archive/v0.7.3.tar.gz"`

To see the new sha hash:
`brew fetch https://raw.githubusercontent.com/cadedaniel/zstd/master/zstd.rb` (that url will be the new `zstd.rb`), and it will output the SHA256.

Another idea that we could implement is to add a HEAD option. That would just look up the latest commit on zstd MASTER, and pull the code from there.

I don't know of any command to automate this, luckily the changes are pretty simple for every release.

Cyan4973:
I'm rather favorable to the idea of a brew install option.

It would better if it could be automated. Having to update the file with each release is yet another action which slows down release process.

It's also a bit unusual to have to create the release first, and then update master branch to put relevant information, thus making `main` and `latest release` out of sync by a few lines. Probably not a big deal, but a change compared to current process.

chipturner:
In the future, the recipe would live in the official homebrew repository; there it would definitely require process after a release was made.  Once it is upstream (they want a certain installation level before admitting new packages to the main repo), this becomes a lot less clunky.

We could put the file in another repo (new?), though it would be a bit disjointed and hard to discover.

I can take care of updating this file when we make releases and until such a time as we get it upstream.

cadedaniel:
Well, after reading through some of the [Homebrew docs](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md) I get the feeling that they want a formula maintainer to update the formula each time there's a new release. So like @chipturner says once `zstd` gets to the point where it's in the brew repo, it'll definitely need that release process.

Until then, there's another solution. If there were a new tag in `zstd` called `latest_release` that always pointed to the latest release, you can set up the formula to download that using `--HEAD`

It would look like this:
`head "https://github.com/Cyan4973/zstd.git", :tag => "latest_release"`
That way anybody who ran:
`brew install <zstd.rb url> --HEAD` would get the `latest_release` tag, which points to the latest release.

This would require a dev on this repo to update that tag every time you do a release, but I find that much easier/painless than editing this file every time.

Another idea is to just change the file every time you do a version bump. Obviously there's the drawback of actually doing that, but at least you won't have the issue where `main` and `latest_release` are out of sync, as @Cyan4973 mentioned.

cadedaniel:
Last thought: the `--HEAD` option would make it easier each release, but it's not the conventional `brew` way of doing things (since `--HEAD` is usually for the ["development cutting edge"](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#head).

chipturner:
My gut says we should just do it like this for now; it's a minor upkeep and is a little odd that it's in the master branch, but my hope is we can get it into mainstream brew pretty quickly, and having a brew tap already available will be worth it -- many more people can play with it from the first moment a release is available.

chipturner:
Looks like `brew tap` wants it to be on the master branch (or whatever the github branch is); can you merge it there too?  Alternatively the default branch github uses and that `git clone` returns can be changed to `dev` globally, but that might be disruptive.

Cyan4973:
I typically update the `master` branch before creating a release.
Next release is `v0.7.4`, and will be ready in a few days.
If you need it faster than that, I can make an exception to update master branch immediately.

chipturner:
No rush at all -- I can send a pull request, but since it has to have the sha of the release tarball, it would be separate from the tagging of 0.7.4, I think.

Cyan4973:
Yes. Indeed, v0.7.4 brew must be completed _after_ v0.7.4 release.
The only point of merging into `main` early 
is to make brew of `v0.7.3` available from `main`.

Cyan4973:
I initially intended to release `v0.7.4` for the new streaming API,
but it could be a stretch.

Maybe better to release `v0.7.4` now with its share of improvements,
and keep some more time for discussion and feedback on the streaming API.

Cyan4973:
v0.7.4 is out.
Could be a good time to update brew file ...

PR #250, opened 2016-07-16
Add legacy support for the low-level streaming API

luben:
Couple of functions and declarations were missing in order to be able to use the low-level streaming API with the legacy decoders.


Cyan4973:
Good point, thanks @luben !

PR #253, opened 2016-07-19
Fix compile issue with ZSTD_LEGACY_SUPPORT=1 and ZSTD_HEAPMODE=0

gymdis:
ZSTD_decompressDCtx was undefined.


Cyan4973:
Thanks for the fix @gymdis !
Indeed, that's a good catch !

gymdis:
Glad I could be of use. :)

PR #259, opened 2016-07-25
legacy decoder for v0.7 format

inikep:


Cyan4973:
Travis seems to have problems, since the test last long, taking a substantial time budget.

Final result is not perfect : 
https://travis-ci.org/Cyan4973/zstd/jobs/147276812

Still some issue with inlining on ppc platform.

I'm quite surprised though. These functions did not changed substantially between 0.6 and 0.7, so there is very little reason why they would pass without problem on 0.6, and fail with a inline warning on 0.7.

Suggest you remove the `-Werror` flag, because if you have to find them one by one, it will take a while.

inikep:
I removed some FORCE_INLINE for v0.6 as well.

It's better to remove -Werror as we will face the same problems with v0.8 and newer legacy decoders.

Cyan4973:
:+1: 

Still, puzzled why `inline` becomes a problem for `ppc` target specifically, while it was working well so far.
Maybe we could remove this warning specifically on `ppc` target, keeping the ability to detect other kind of warnings.

inikep:
A good idea. "-Wno-attributes" should be enough.

PR #262, opened 2016-07-26
Miscellaneous updates

ebiggers:


lib/compress/zstd_compress.c
```diff
@@ -2311,19 +2311,19 @@ static size_t ZSTD_writeFrameHeader(void* dst, size_t dstCapacity,
     U32 const dictIDSizeCode = (dictID>0) + (dictID>=256) + (dictID>=65536);   /* 0-3 */
     U32 const checksumFlag = params.fParams.checksumFlag>0;
     U32 const windowSize = 1U << params.cParams.windowLog;
-    U32 const directModeFlag = params.fParams.contentSizeFlag && (windowSize > (pledgedSrcSize-1));
+    U32 const singleSegment = params.fParams.contentSizeFlag && (windowSize > (pledgedSrcSize-1));
```
Cyan4973: :+1:

lib/common/zstd_internal.h
```diff
@@ -137,7 +137,10 @@ static const U32 OF_defaultNormLog = 5;
 /*-*******************************************
 *  Shared functions to include for inlining
 *********************************************/
-static void ZSTD_copy8(void* dst, const void* src) { memcpy(dst, src, 8); }
+static void ZSTD_copy8(void* dst, const void* src)
+{
+    MEM_write64(dst, MEM_read64(src));
```
Cyan4973: While I understand it does better follow "mem.h",
I'm wondering in which case this construction is better ?

lib/decompress/zstd_decompress.c
```diff
@@ -863,9 +863,11 @@ static size_t ZSTD_decompressFrame(ZSTD_DCtx* dctx,
 
     /* Frame Header */
     {   size_t const frameHeaderSize = ZSTD_frameHeaderSize(src, ZSTD_frameHeaderSize_min);
+        size_t result;
         if (ZSTD_isError(frameHeaderSize)) return frameHeaderSize;
         if (srcSize < frameHeaderSize+ZSTD_blockHeaderSize) return ERROR(srcSize_wrong);
-        if (ZSTD_decodeFrameHeader(dctx, src, frameHeaderSize)) return ERROR(corruption_detected);
+        result = ZSTD_decodeFrameHeader(dctx, src, frameHeaderSize);
```
Cyan4973: :+1:

lib/common/zstd_internal.h
ebiggers: I did not identify a specific case.  I just felt that if someone has disabled the `memcpy()` version, then it shouldn't be used here either.

If you are concerned about the load/store potentially being suboptimal then one option is to only do it when `memcpy()` is disabled:

```
static void ZSTD_copy8(void* dst, const void* src)
{
#if defined(MEM_FORCE_MEMORY_ACCESS) && (MEM_FORCE_MEMORY_ACCESS>=1)
    MEM_write64(dst, MEM_read64(src));
#else
    memcpy(dst, src, 8);
#endif
}
```

I would also consider moving the "copy" functions into `mem.h`.

Cyan4973: Something I did not took time to describe last year while redacting 
http://fastcompression.blogspot.fr/2015/08/accessing-unaligned-memory.html

is that `memcpy(dst, src, 8);` would always work correctly, on all architectures and compiler tested.

Problems were limited to `U64 val; memcpy(&val, src, 8);` which, in some cases, generated some completely useless store + retrieve intermediate operations. Hence the different tricks (`__packed`, direct `U64*` ) now present in `mem.h` to avoid these inefficiencies.

As an example :
https://godbolt.org/g/h1ODwa

ebiggers: If you are confident about that I can drop this patch.  But I still think it is inconsistent to always use `memcpy()` here, despite it potentially being disabled in `mem.h`.

Cyan4973: Yes, I feel confident.
I also don't consider `memcpy()` to be disabled :
it's just displaced by some more efficient method (under conditions) for the specific usage of `MEM_read32()` and equivalents.
But `memcpy()` is still used in many other places in the code, so it's all but disabled.

ebiggers: Okay, I've removed the commit for now.

Cyan4973:
:+1:

PR #265, opened 2016-07-31
Enable build on FreeBSD ports (includes DragonFly BSD)

jrmarino:
Zstd has been introduced to FreeBSD ports
(http://www.freshports.org/archivers/zstd/) which DragonFly BSD also
uses.  FreeBSD and DragonFly use the install targets (albeit modified in
some cases) so they must be added to the associated Makefile filters.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,10 +70,10 @@ clean:
 	@echo Cleaning completed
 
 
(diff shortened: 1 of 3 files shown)


jrmarino:
sure, I seem to have missed that branch.

PR #266, opened 2016-07-31
Enable build on FreeBSD ports (includes DragonFly BSD) [dev branch]

jrmarino:
Zstd has been introduced to FreeBSD ports
(http://www.freshports.org/archivers/zstd/) which DragonFly BSD also
uses.  FreeBSD and DragonFly use the install targets (albeit modified in
some cases) so they must be added to the associated Makefile filters.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,10 +70,10 @@ clean:
 	@echo Cleaning completed
 
 
-#------------------------------------------------------------------------
-#make install is validated only for Linux, OSX, kFreeBSD and Hurd targets
-#------------------------------------------------------------------------
-ifneq (,$(filter $(shell uname),Linux Darwin GNU/kFreeBSD GNU))
+#----------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Cyan4973:
:+1:

Cyan4973:
This looks interesting.
The idea of replacing `install` by a variable which can be tuned by environment looks good, at least on paper.

Now, in practice, non-BSD users must still be able to install `libzstd`, without any added complexity compared to today. Variable names should also be more "neutral".
It definitely looks doable.

jrmarino:
right.  Sorry, I didn't mean to suggest the name is "BSD_INSTALL_X".  those are predefined variable specific to FreeBSD ports.  Let's say the variable was called "INSTALL_LIBRARY=install -m 755" then I would just predefine INSTALL_LIBRARY (to the value of BSD_INSTALL_LIB).

I just wanted to illustate wan I meant and I think you understand that just fine.   Thanks!

PR #276, opened 2016-08-02
Add Ada bindings to list of "known ports"

jrmarino:


jrmarino:
Hi Yann,
You might consider arrange these "ports" in alphabetical order.
Also, when I first saw these described as "ports", I thought they were the reference library reimplemented in the local language, but they all seem to be just bindings.   The description as "ports" was misleading to me, so you might want to think about describing them as bindings.

Both of these are just suggestions/observations so take them or leave them as you wish!.

Cyan4973:
There are very good points @jrmarino .

I tend to classify language binding : 
1 - on "freshness" level
2 - on "language diffusion" level, as measured by [Tiobe index](http://www.tiobe.com/tiobe-index/).

though I may as well remove point 2 and replace it by alphabetical order.

Good point regarding "ports".
This chapter currently only contains "bindings", but there is no intention to create a separate "ports" category. Both will be present there.
So a question is, what could be a good name for this chapter ?

jrmarino:
I see this as similar to libsodium (https://github.com/jedisct1/libsodium)

Denis just called it "Bindings for other languages". 
https://download.libsodium.org/libsodium/content/bindings_for_other_languages/

I think that's fine.  If somebody actually reimplements zstd, and it's a full and current implementation (which can't be done until zstd API is fully stable), then maybe a new chapter could be added.  It would be a very short chapter.  I think that approach would be more accurate and easier to grasp quickly too.

PR #279, opened 2016-08-10
Portability

inikep:


lib/common/mem.h
```diff
@@ -110,7 +110,7 @@ MEM_STATIC void MEM_check(void) { MEM_STATIC_ASSERT((sizeof(size_t)==4) || (size
 #ifndef MEM_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
 #  if defined(__GNUC__) && ( defined(__ARM_ARCH_6__) || defined(__ARM_ARCH_6J__) || defined(__ARM_ARCH_6K__) || defined(__ARM_ARCH_6Z__) || defined(__ARM_ARCH_6ZK__) || defined(__ARM_ARCH_6T2__) )
 #    define MEM_FORCE_MEMORY_ACCESS 2
-#  elif defined(__INTEL_COMPILER) || \
+#  elif defined(__INTEL_COMPILER) /*|| defined(_MSC_VER)*/ || \
```
Cyan4973: Looks like an aborted test case.
Does it bring anything to keep it in the code in commented form ?

lib/common/xxhash.c
```diff
@@ -52,7 +52,7 @@
 #ifndef XXH_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
 #  if defined(__GNUC__) && ( defined(__ARM_ARCH_6__) || defined(__ARM_ARCH_6J__) || defined(__ARM_ARCH_6K__) || defined(__ARM_ARCH_6Z__) || defined(__ARM_ARCH_6ZK__) || defined(__ARM_ARCH_6T2__) )
 #    define XXH_FORCE_MEMORY_ACCESS 2
-#  elif defined(__INTEL_COMPILER) || \
+#  elif (defined(__INTEL_COMPILER) && !defined(WIN32)) || \
```
Cyan4973: do you mean Intel Compiler, outside of Windows (Linux?), behaves differently ?

Cyan4973:
Some remaining minor questions around Intel Compiler.
As agreed, I'll accept "Portability" package before "Testing" one.

lib/common/xxhash.c
inikep: On Windows
__attribute__((packed)) unalign
is not supported. 
Therefore we need:
a) XXH_FORCE_MEMORY_ACCESS = 0
b) __pragma( pack(push, 1) )

lib/common/mem.h
inikep: I marked that VC++ was tested but as I wrote there was no improvement, but now "_MSC_VER" is able to work because of `__pragma( pack(push, 1) )`

lib/common/xxhash.c
Cyan4973: OK,
that sounds really weird from Intel to support `__attribute__` depending on running on Linux or Windows,
as if it was somewhat reliant on some 3rd party compiler system,
but well, having no experience of `icc`, I can't comment ...

inikep: from Asana: icc is a strange compiler as it is partially compatible with gcc (with Linux/Unix only) and Visual C++ (with Windows only)

lib/common/mem.h
Cyan4973: So, it works, but brings no improvement ?

Maybe we don't need it then.
Default setting, using `memcpy()`, is best, whenever possible.

inikep: Correct: it works, but brings no improvement.

Cyan4973: I'll removed it manually after the merge

PR #280, opened 2016-08-10
Testing

inikep:


Cyan4973:
Is this PR still valid, or do you need to create another one ?

inikep:
it's valid but also contains the same files from #279 Portability

PR #283, opened 2016-08-11
Testing2

inikep:


Cyan4973:
OK,
very little comments here,
just it feels preferable that "regrouped tests" on Travis have a sense of "common topic",
which is on average well respected.
For example, `gpptest` `gnu90test` `c99test` etc. look pretty good together.

On the other hand, `test-zstd_nolegacy` `zlibwrapper` and `cmaketest` are a strange mix.

We can improve that later on.

inikep:
This was not a random regrouping. We have:
a) 4 tests - Ubuntu 12.04 container
b) 6 tests - Ubuntu 12.04
c) 1 test - Ubuntu 14.04
d) 1 test - Mac OS

For "Ubuntu 12.04 container"
a1) make asan = 13 min
a2) make usan = 13 min
a3) make test && travis-install = 12 min
a4) the rest = 3 min

`test-zstd_nolegacy zlibwrapper and cmaketest` are a strange mix because they are `the rest` for "Ubuntu 12.04 container"

Issue #287, opened 2016-08-12
zstd CLI log lines truncate file names to last 20 characters


david-y-lam:
Any reason why we don't print out the full file name?

```
davidlam-mbp:zstd davidlam$ touch my_very_long_file_name_that_is_over_20_characters_long
davidlam-mbp:zstd davidlam$ echo my_very_long_file_name_that_is_over_20_characters_long  | wc -c
      55
davidlam-mbp:zstd davidlam$ ./zstd ./my_very_long_file_name_that_is_over_20_characters_long -o /dev/null
r_20_characters_long :1300.00%   (     0 =>     13 bytes, /dev/null)
```

For comparison, `gzip` prints out full file name in `verbose` mode. If this is a bug, I can send up a pull request.

```
davidlam-mbp:zstd davidlam$ gzip --verbose my_very_long_file_name_that_is_over_20_characters_long
my_very_long_file_name_that_is_over_20_characters_long:   -99.9% -- replaced with my_very_long_file_name_that_is_over_20_characters_long.gz
```

Cyan4973:
This was a UI choice,
to avoid mangled output when some files end up being very long.
This is more useful when there are a lot of files of variable name length.

Though this can be changed if there are better choices.

david-y-lam:
@Cyan4973, I was confused by this truncation and personally would like to see the full file name (I don't know of any Unix CLI that shortens file params in output like this FWIW); I think that if the user has long file names already, then they would not be bothered by very long log lines. I'll send a pull request.

david-y-lam:
Fix got merged into `dev`, closing for you.

PR #290, opened 2016-08-13
Remove file name truncation in zstd CLI output

david-y-lam:
This is to to fix the issue in #287.

Test Plan: 

Tested that file names exceeding 20 characters are printed in full and that short file names are left-justified with padding on right.

```
davidlam-mbp:zstd davidlam$ ./zstd ./my_very_long_file_name_that_is_over_20_characters_long -o /dev/null
./my_very_long_file_name_that_is_over_20_characters_long :1300.00%   (     0 =>     13 bytes, /dev/null)
davidlam-mbp:zstd davidlam$ ./zstd ./README.md -o /dev/null
./README.md          : 45.69%   (  5529 =>   2526 bytes, /dev/null)
```


david-y-lam:
How do we feel about this one? :)

Cyan4973:
It's okay.
I'm waiting a bit before the merge as I wonder about a potential collision issue.
But we'll solve it anyway.

david-y-lam:
thanks!

PR #296, opened 2016-08-18
Other

inikep:


inikep:
Fixes issue described here:
http://encode.ru/threads/2119-Zstandard?p=49504&viewfull=1#post49504

inikep:
Strange: Disabling icc warning (using the "-W177:false" compilation flag) brokens MSVC++ compilation. Microsoft and Intel should work together on this.

Cyan4973:
A lot of code in "legacy" codec is removed.
Is it because it's never referenced ?

inikep:
icc was right with warnings and the code was never used. I decided to remove it for "legacy" and set pragma for ordinary code:
```
if defined(__INTEL_COMPILER)
 #  pragma warning(disable : 177)    /* disable: message #177: function was declared but never referenced */
#endif
```

PR #300, opened 2016-08-22
Add a link to C# binding

borzunov:
I suggest adding a link to C# binding. The link is inserted so as to maintain binding order mentioned [here](https://github.com/Cyan4973/zstd/pull/276#issuecomment-236909368).


Cyan4973:
:+1:

PR #302, opened 2016-08-25
improved documentation

inikep:


inikep:
source code was not changed at all in this PR

zstd_compression_format.md
```diff
@@ -97,6 +97,42 @@ to decode all concatenated frames in their sequential order,
 delivering the final decompressed result as if it was a single content.
 
 
+Skippable Frames
```
Cyan4973: Skippable Frames are an oddity,
not really the main objective of the specification.

Is there any gain in putting it at the beginning of the document, rather than the end ?

Cyan4973:
These are pretty good modifications.
I agree with most of them,
just a last remaining question on skippable frames.

zstd_compression_format.md
inikep: I moved "skippable frames" from the middle where they did not fit. For me they can be at the beginning because the section is short and easy to read and it's related to "Frame Concatenation". I can move them to the end if you wish.

Cyan4973: I've read the modified document, 
and I agree this section it not too long,
so it should not confuse the reader before they reach `General Structure of Zstandard Frame format`

PR #306, opened 2016-08-29
Cleanup some errors in typedef comments and remove duplicated HOWTO f…

david-y-lam:
Was confused by some of the comments which referenced no longer existent files, so I fixed those.

Also, `zbuff_decompress.c` contains a comment block that is similar to `zbuff.h` (but is slightly out of date) so I removed it.

```
davidlam-mbp:zstd davidlam$ diff ./header_doc src_doc
14,17c14,16
< *  The content of `dst` will be overwritten (up to *dstCapacityPtr) at each function call, so save its content if it matters, or change `dst`.
< *  @return : 0 when a frame is completely decoded and fully flushed,
< *            1 when there is still some data left within internal buffer to flush,
< *            >1 when more data is expected, with value being a suggested next input size (it's just a hint, which helps latency),
---
> *  The content of @dst will be overwritten (up to *dstCapacityPtr) at each function call, so save its content if it matters, or change @dst.
> *  @return : a hint to preferred nb of bytes to use as input for next function call (it's only a hint, to help latency),
> *            or 0 when a frame is completely decoded,
21c20
< *  output : ZBUFF_recommendedDOutSize== 128 KB block size is the internal unit, it ensures it's always possible to write a full block when decoded.
---
> *  output : ZBUFF_recommendedDOutSize==128 KB block size is the internal unit, it ensures it's always possible to write a full block when decoded.
```


david-y-lam:
Should we/I also add comment to `zbuff.h` saying that functions there will be deprecated by equivalent ones in `zstd.h`?

Cyan4973:
These are good points @david-y-lam .

> Should we/I also add comment to zbuff.h saying that functions there will be deprecated by equivalent ones in zstd.h?

Yes

david-y-lam:
updated

PR #310, opened 2016-08-31
Change type of ZSTD_freeCStream to void

bos:
There is no way this function can fail, so having its return type
suggest that it can is misleading. On top of which almost no uses
of it in the code base actually checked its result anyway :-)

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2806,14 +2806,13 @@ ZSTD_CStream* ZSTD_createCStream_advanced(ZSTD_customMem customMem)
     return zcs;
 }
 
-size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
+void ZSTD_freeCStream(ZSTD_CStream* zcs)
 {
-    if (zcs==NULL) return 0;   /* support free on NULL */
+    if (zcs==NULL) return;   /* support free on NULL */
     ZSTD_freeCCtx(zcs->zc);
     if (zcs->inBuff) zcs->customMem.customFree(zcs->customMem.opaque, zcs->inBuff);
     if (zcs->outBuff) zcs->customMem.customFree(zcs->customMem.opaque, zcs->outBuff);
     zcs->customMem.customFree(zcs->customMem.opaque, zcs);
-    return 0;
 }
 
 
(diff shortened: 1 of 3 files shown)


nabijaczleweli:
:+1: can't fail

lib/compress/zstd_compress.c
```diff
@@ -2806,14 +2806,13 @@ ZSTD_CStream* ZSTD_createCStream_advanced(ZSTD_customMem customMem)
     return zcs;
 }
 
-size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
+void ZSTD_freeCStream(ZSTD_CStream* zcs)
 {
-    if (zcs==NULL) return 0;   /* support free on NULL */
+    if (zcs==NULL) return;   /* support free on NULL */
```
intdel: If the return type is void, what purpose does "return" still have? Does it kind of work like `break`?

david-y-lam: @intdel, similar to what it was doing before - it will exit the function, but not return a value to the caller.

terrelln:
The rest of the free functions, {ZSTD_free*, ZBUFF_free*}, all always succeed, but return size_t.  If we are going to change one, we should change them all.

MarkLavrynenko:
you should also delete errorCode local variable and merge master branch

Cyan4973:
Sorry, I did not take the time to answer this one correctly.

I would have liked to have this discussion many months ago, when this item was on the "design" list, since both solutions looked plausible, and I was looking for some external point of view.

But as we were marching towards `v1.0`, a decision had to be made, and the selected one included a `size_t` return type. 
The reasoning was : maybe at some point in the future there will be some possible errors during `free`. In which case, we will appreciate to have a way to provide such errors.
`size_t` allows that, `void` doesn't.

Many months have passed since, and the "error-during-free" scenario never materialized.
Maybe it will in the future, or maybe not.

An additional problem now is that these prototypes are part of the ___stable API___, since multiple releases. By definition, they are not supposed to be changed, for API / ABI compatibility.
During v0.x development, some exception would have been possible, but now that we have reached v1.0, I don't see how to do that.

So that makes a lot of argument to keep it as it is, considering the relatively small benefit (it's always possible to _not_ check the function return value).

We could introduce new prototypes, with `void` return type, and gradually (slowly) deprecate older ones. But function names cannot be overloaded. Current name is pretty clear, and I don't see how to introduce a second name which would be as good as this one.

PR #313, opened 2016-08-31
Changed URLs

94b1:
Changed URLs to reflect change in ownership from and reduce page load times of zstd.net due to multiple redirects.

I have signed the CLA at code.facebook.com/cla


Cyan4973:
Good point, thanks !

PR #318, opened 2016-09-01
Fix typo in index page title

camwiegert:
Caught a typo in the documentation home page title.


Cyan4973:
Good point !
Impressive it remained unnoticed  ...

PR #324, opened 2016-09-01
Added Python 2.7 and 3.x bindings

dimkr:


gyscos:
Does this belong in this repo? Other bindings are linked to from the [zstd website](http://facebook.github.io/zstd/) (like this [python binding](https://pypi.python.org/pypi/zstd) - though perharps not up to date.)

PR #325, opened 2016-09-01
Add Parallel Zstandard tool to contrib/

terrelln:
Parallel Zstandard is a Pigz-like tool for Zstandard.


Cyan4973:
As a minor detail,
it's probably better for the graph to scale to the useful region, from 2.5 to 4.0,
rather starting from 0.
This way, differences will be more visible.

PR #326, opened 2016-09-02
Update scale for compression speed graph

terrelln:
Focus in on interesting range.


Cyan4973:
OK, sorry to bother you, some possible minor improvement, while at it ... :)

On the X axis, you can probably start from 4, rather than 1, in order to better center the useful part.
Also, a light vertical scale could be useful to properly follow which point is at which speed

Cyan4973:
Some other, less cosmetics, issues to look at : 
https://travis-ci.org/facebook/zstd/jobs/156968020

PR #327, opened 2016-09-02
Pzstd should now compile with gcc

terrelln:
There were small issues with flags and missing includes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,9 @@ matrix:
       env: PLATFORM="Ubuntu 12.04 container" CMD="make test && make clean && make travis-install"
     - os: linux
       sudo: false
+      language: cpp
       install:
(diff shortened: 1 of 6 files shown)


terrelln:
Still an issue with travis's g++ version, it doesn't support C++ 11.

Cyan4973:
Still some issues apparently : 

```
g++ -I../../lib -I../../lib/common -I../../lib/dictBuilder -I../../programs -I. -O3 -std=c++11   -c Pzstd.cpp -o Pzstd.o
cc1plus: error: unrecognized command line option ‘-std=c++11’
```

maybe a problem of g++ version.
default compiler is quite old in Travis.

terrelln:
Yup, I'm updating the default g++ to 4.8 for CXX, but I'll leave CC alone.

Cyan4973:
Considering latest test result,
https://travis-ci.org/facebook/zstd/jobs/156989526
I suspect a missing dependency.

terrelln:
It finally works 😄

PR #339, opened 2016-09-04
Add Erlang binding

mururu:
I've released the Erlang binding. Could you add it to the list?

https://hex.pm/packages/zstd


Cyan4973:
sure !

PR #349, opened 2016-09-06
[pzstd] Windows build and default number of threads

### LINKED ISSUE #331: pzstd
Ran into a few bugs using pzstd, but wow is it ever blazingly fast.  Apologies, C isn't a language I'm particularly familiar with or I'd try to fix these myself.

1) To build on Linux (g++ only?) you need to pass the -pthreads flag.  Easy to fix so I could get the code to compile, but I'm not sure the appropriate Makefile voodoo to make it work on multiple platforms.

2) decompression over stdin seems to still require an input file.  compression over stdin works fine.

3) on 32bit environment I could manage to get
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted
```
if I tried to pass in large content over stdin, or read large block devices.

terrelln:
Pzstd now builds on windows with minGW-64 and VS2015.
This resolves #331, as the last issue of setting a reasonable default for the number of threads has been fixed.

It doesn't build on minGW-32, because by default it doesn't support pthreads, and thus doesn't support `std::thread`.  This is fixable.  I think a version of minGW-64 supports a posix threads implementation for minGW-32.  There are also libraries that provide this functionality if necessary.

### DIFF
--- a/contrib/pzstd/Pzstd.cpp
+++ b/contrib/pzstd/Pzstd.cpp
@@ -41,7 +41,7 @@ size_t pzstdMain(const Options& options, ErrorHolder& errorHolder) {
       return 0;
     }
     std::error_code ec;
-    inputSize = file_size(options.inputFile, ec);
+    inputSize = static_cast<size_t>(file_size(options.inputFile, ec));
     if (ec) {
       inputSize = 0;
     }
@@ -155,7 +155,7 @@ static void compress(
     ZSTD_parameters parameters) {
   auto guard = makeScopeGuard([&] { out->finish(); });
   // Initialize the CCtx
-  std::unique_ptr<ZSTD_CStream, size_t (&)(ZSTD_CStream*)> ctx(
+  std::unique_ptr<ZSTD_CStream, size_t (*)(ZSTD_CStream*)> ctx(
       ZSTD_createCStream(), ZSTD_freeCStream);
   if (!errorHolder.check(ctx != nullptr, "Failed to allocate ZSTD_CStream")) {
     return;
@@ -311,7 +311,7 @@ static void decompress(
     std::shared_ptr<BufferWorkQueue> out) {
   auto guard = makeScopeGuard([&] { out->finish(); });
   // Initialize the DCtx
-  std::unique_ptr<ZSTD_DStream, size_t (&)(ZSTD_DStream*)> ctx(
+  std::unique_ptr<ZSTD_DStream, size_t (*)(ZSTD_DStream*)> ctx(
       ZSTD_createDStream(), ZSTD_freeDStream);
   if (!errorHolder.check(ctx != nullptr, "Failed to allocate ZSTD_DStream")) {
     return;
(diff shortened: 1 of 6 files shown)


contrib/pzstd/Pzstd.cpp
```diff
@@ -41,7 +41,7 @@ size_t pzstdMain(const Options& options, ErrorHolder& errorHolder) {
       return 0;
     }
     std::error_code ec;
-    inputSize = file_size(options.inputFile, ec);
+    inputSize = static_cast<size_t>(file_size(options.inputFile, ec));
```
Cyan4973: Prefer using 64-bits integer for file sizes,
as even on 32-bits platforms, you may have to handle files > 4 GB.

I'm not sure how it would behave  on a 32-bits system if file size was 4GB+1 bytes.
Would `inputSize` value become `1` ?
That could be a use case interesting to add in the test suite by the way.

Conversion to `size_t` can happen at a later stage, when you need to allocate a buffer for example.

terrelln:
The C and C++ standards guarantee that unsigned overflow works as a mod by 2^(# bits), but integer overflow is undefined behavior, so 4GB + 1 = 1 for 32 bit unsigned.  I'll add a test case to check for this in a 32 bit build.

contrib/pzstd/Pzstd.cpp
```diff
@@ -217,14 +217,17 @@ static void compress(
  * @param numThreads The number of threads available to run compression jobs on
  * @param params     The zstd parameters to be used for compression
  */
-static size_t
-calculateStep(size_t size, size_t numThreads, const ZSTD_parameters& params) {
+static size_t calculateStep(
+    std::uintmax_t size,
+    size_t numThreads,
+    const ZSTD_parameters &params) {
   size_t step = 1ul << (params.cParams.windowLog + 2);
```
Cyan4973: This is just a minor question here, as I don't foresee any problem.

I read `1ul` as `(unsigned long) 1`, which may, or may not, be equivalent to `size_t`.
Any reason you prefer this to `(size_t)1` ?

The only issue I can think of is a potential conversion warning from some over-zealous compiler.

Cyan4973:
:+1:

terrelln:
Thanks @djw432067.  The narrowing should be fixed in https://github.com/facebook/zstd/pull/349/commits/823bf3d08de6aff1b0fd10f384bf72e727412d07

terrelln:
The pzstd argument parsing is going to be updated, as it is not POSIX compliant and differs from zstd's argument parsing.

I don't really want to have an environment variable control the behavior of the program.

The options I can think of are:

- Revert back to having no default
- Allow users to `#define PZSTD_NUM_THREADS` at compile time, with a default to `std::thread::harware_concurrency()` (maybe this is what you meant by environment variable)
- Print the default number of threads in the help message (will happen if we keep a default)
- Print the number of threads used to stderr every run when it isn't specified
- Maybe there is a better cross platform way to get this information.

Cyan4973:
I believe you can change argument parsing
and keep the ability to have some default number of threads.

I also like your second bullet point.

PR #352, opened 2016-09-07
Specify that dictionary ID is little-endian

terrelln:
I noticed that the wiki and `zstd_compression_format.md` are out of sync.  Is there a clean way to have the wiki track this file, or should we just have that part of the wiki link to this file?


Cyan4973:
Good point.
The wiki page was more a test-ground, and is no longer up to date.
It should be removed.

PR #354, opened 2016-09-08
Add nodejs native module binding

zwb-ict:
This package is different from **node-zstandard**. 
It's nodejs native module binding ZSTD's API.


Cyan4973:
erf, I'll need a bit more precise and explicit differentiation between your version and node-zstandard.
"native" doesn't tell enough.

From a high-level look at the code, it seems to me that :
- node-zstandard works on files and streams
- node-zstd works on buffers
- node-zstd supports dictionary compression (on buffers)

As a loosely connected issue, I also note the exposition of a "fast dictionary compression API".
However, it seems to miss the point.
It can only be "fast" if the digested dictionary (CDict, or DDict) is prepared in advance, preserved, and then re-used for multiple compression / decompression. Rebuilding the digested dictionary for each file is no better, and in fact slower, than using the more straighforward `*_usingDict()` variant.

zwb-ict:
@Cyan4973 Thank you for your feedback. I'll think carefully, then give you a more precise info.

PR #363, opened 2016-09-12
Dev

inikep:


Cyan4973:
Great, this patch fixes many issues.

Seems there is a last remaining one, regarding clang compilation in 32-bits mode.
Probably a missing package, like multilib.

inikep:
This is a bug in Debian with 32-bit asan on 64-bit machine:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=772122

I read that it's fixed for clang 3.8 and 3.9 (both in experimental branch) and I'm working on it.

Cyan4973:
:+1:

PR #366, opened 2016-09-14
zstd.exe has FileVersion and ProductVersion with 32-bit gcc (MinGW)

inikep:


programs/Makefile
```diff
@@ -57,7 +57,13 @@ endif
 ifneq (,$(filter Windows%,$(OS)))
 EXT =.exe
 VOID = nul
-RES_FILE = ..\projects\VS2010\zstd\zstd.res
+RES64_FILE = ..\projects\VS2010\zstd\generate_res\zstd64.res
```
Cyan4973: @inikep : 
I did not realize that with this change, `programs` is now dependent on `projects` (recently renamed `build`).
which does not feel correct : `programs` should not depend on `build`.

So, would it be possible to have `zstd??.res` stored within `programs` instead ?
I guess it would change the corresponding Visual projects ?

inikep:
The zstdXX.res is a Windows compiled object file. I think it shouldn't go directly to `programs/` but we can move `projects\VS2010\zstd\generate_res` to `programs\generate_res`. The issue remain similar as generating a new zstdXX.res (we do it manually) is dependent on a VS2010 project.

Cyan4973:
ok

PR #367, opened 2016-09-14
Add FSE decoding tables for pred. distributions to spec for reference + clarify bitstream reading

jrudolph:
I found the decoding table algorithm a bit hard to understand (but it was possible if you somewhat know how the tables are going to look like in the end), so I think it would be nice to have the actual tables for the predefined tables in the spec to test your implementation against.

Tbh. I think the description of how padding is added to the end of the stream was just wrong. I spent some time (on the beach ;)) with my hand-written notes trying to find out what's going on. This didn't work out (not least because of the sun). I then added debug output to zstd to find out which values to expect at the start of the reversed bitstream which I then tried to find in the bits. It was only when I eventually looked into the zstd compressor code where the `FlushXXX` calls caught my attention that I understood what was going on... 

WDYT?


Cyan4973:
These are very good points @jrudolph .
I fully agree with your edits

jrudolph:
Cool, thanks for merging.

PR #368, opened 2016-09-14
[programs] fix the memory leak

bryongloden:
Passing one pointer into `realloc()` and assigning the result directly into that same pointer variable can cause a memory leak if the reallocation fails, because the original allocation will still exist. The correct way to do this is to use a temporary pointer variable.

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -344,7 +344,7 @@ UTIL_STATIC const char** UTIL_createFileList(const char **inputNames, unsigned i
 {
     size_t pos;
     unsigned i, nbFiles;
-    char *bufend, *buf;
+    char *bufend, *buf, *buftmp;
     const char** fileTable;
 
     buf = (char*)malloc(LIST_SIZE_INCREASE);
@@ -356,9 +356,15 @@ UTIL_STATIC const char** UTIL_createFileList(const char **inputNames, unsigned i
             size_t len = strlen(inputNames[i]);
             if (buf + pos + len >= bufend) {
                 ptrdiff_t newListSize = (bufend - buf) + LIST_SIZE_INCREASE;
-                buf = (char*)realloc(buf, newListSize);
-                bufend = buf + newListSize;
-                if (!buf) return NULL;
+                buftmp = (char*)realloc(buf, newListSize);
+                if (buftmp != NULL) {
+                    buf = buftmp;
+                    bufend = buf + newListSize;
+                    if (!buf) return NULL;
+                } else {
+                    free(buf);
+                    /* And handle error */
+                }
             }
             if (buf + pos + len < bufend) {
                 strncpy(buf + pos, inputNames[i], bufend - (buf + pos));


programs/util.h
```diff
@@ -356,9 +356,15 @@ UTIL_STATIC const char** UTIL_createFileList(const char **inputNames, unsigned i
             size_t len = strlen(inputNames[i]);
             if (buf + pos + len >= bufend) {
                 ptrdiff_t newListSize = (bufend - buf) + LIST_SIZE_INCREASE;
-                buf = (char*)realloc(buf, newListSize);
-                bufend = buf + newListSize;
-                if (!buf) return NULL;
+                buftmp = (char*)realloc(buf, newListSize);
```
Cyan4973: prefer declaring variable in narrowest scope.
buftmp should be declared here, instead of line 347.
It can be made `const` too.
It also does not need to be `char*` since it's not accessed anyway.
So I would lean for : 
`void* const buftmp = realloc(buf, newListSize);`
and corresponding cast at line 361

programs/util.h
```diff
@@ -356,9 +356,15 @@ UTIL_STATIC const char** UTIL_createFileList(const char **inputNames, unsigned i
...
+                if (buftmp != NULL) {
+                    buf = buftmp;
+                    bufend = buf + newListSize;
+                    if (!buf) return NULL;
+                } else {
+                    free(buf);
+                    /* And handle error */
```
Cyan4973: Which error should be handled here ?
`return NULL;` ? ( ping @inikep)
The patch seems incomplete here.

programs/util.h
```diff
@@ -356,9 +356,15 @@ UTIL_STATIC const char** UTIL_createFileList(const char **inputNames, unsigned i
...
             if (buf + pos + len >= bufend) {
                 ptrdiff_t newListSize = (bufend - buf) + LIST_SIZE_INCREASE;
-                buf = (char*)realloc(buf, newListSize);
-                bufend = buf + newListSize;
-                if (!buf) return NULL;
+                buftmp = (char*)realloc(buf, newListSize);
+                if (buftmp != NULL) {
+                    buf = buftmp;
+                    bufend = buf + newListSize;
+                    if (!buf) return NULL;
```
Cyan4973: if `buf = buftmp` and `buftmp != NULL`, then `!buf` is never true.
This test should be removed

Cyan4973:
Good catch @bryongloden !
see comments in code

inikep:
Yes, a good catch. I forgot about `If realloc() fails the original block is left untouched; it is not freed or moved.`. But I would prefer a solution with a new function e.g. `UTIL_realloc()` that will free an original buffer even if new it's not allocated by `realloc()`. This function can be useful for the future and will fix 3 `realloc()` calls in `util.h`.

inikep:
The path is implemented at https://github.com/inikep/zstd/commit/6173931868adc4e7c91aaf65aa8a7c3ef8d57b93

Cyan4973:
Fixed by #370

PR #380, opened 2016-09-22
[pzstd] CLI + memory usage + 32 bit mode

terrelln:
Make the pzstd CLI the same as the zstd CLI, except with an extra option "-p" or "--processes" to specify the number of threads.  The default number of threads is `std::thread::hardware_concurrency()` if `PZSTD_NUM_THREADS` is undefined, else it is `PZSTD_NUM_THREADS`.

Improve memory usage.  pzstd now uses between 60% and 75% of the memory it used to use, with the same performance.

Add 32 bit pzstd build and test to Makefile.


Cyan4973:
There is a minor warning on gcc tests :

```
utils/FileSystem.h:27:30: warning: declaration ‘struct stat’ does not declare anything [enabled by default]
 using file_status = struct ::stat;
```

I recommend setting a "debug build" mode with `-Werror` so that warnings get notified as errors and get caught during tests.
(note : the normal release build mode should not have `-Werror` enabled !)

Also : I may have missed it, but I don't see pzstd 32-bits mode tests, neither on Appveyor nor travis.

terrelln:
The warning is a false positive, I might be able to quiet it by using a typedef instead.

I'll add the 32-bit tests to the contbuild.

Cyan4973:
I guess I will need a message to know when you are done with all the improvements ...

terrelln:
I'll just close the pull request and start another.  Its taking some trial and error to get the CI environments set up.

PR #382, opened 2016-09-23
[pzstd] CLI + memory usage + 32 bit mode

terrelln:
* Make the pzstd CLI the same as the zstd CLI, except with an extra option `-p` or `--processes` to specify the number of threads. The default number of threads is `std::thread::hardware_concurrency()` if `PZSTD_NUM_THREADS` is undefined, else it is `PZSTD_NUM_THREADS`.
* Improve memory usage. pzstd now uses between 60% and 75% of the memory it used to use, with the same performance.
* Fix 32 bit pzstd build and add 32 bit tests to travis.
* Add tests to MinGW64 appveyor, and modify `WorkQueue` unit tests to not trigger extremely degraded performance.
* Make pzstd compile without warnings with `-Wall -Wextra -pedantic -Werror`

Note: I changed `programs/datagen.h` and `programs/util.h`.  `util.h` was triggering a macro redefinition warning, and `programs/datagen.h` didn't have include guards.  It also doesn't have `extern "C"`, but it seems that `make gpptest` relies on that, so I added it at the include sites.


Cyan4973:
👍

inikep:
```
>zstd -22f --ultra d:\GitHub-testy\silesia_tar
d:\GitHub-testy\silesia_tar : 24.90%   (211947520 => 52770647 bytes, d:\GitHub-testy\silesia_tar.zst)

>pzstd -p2 -22f --ultra d:\GitHub-testy\silesia_tar

```

I miss the line from zstd with results.

terrelln:
I'll work on a results line.  Having the line updated live will be a little bit tricky, since it means coordinating some amount of state between the threads, so I think I will leave that off until a future version.

However, adding the final results line can be done easily.

Cyan4973:
Alternatively, you could have a simple animation, such as a rotating char, just to show user that something is happening and process is not dead. This is of course only useful if compression takes a long time.

Issue #379, opened 2016-09-21
Support for Legacy format missing or incomplete when building the VS2008 and VS2010 projects


KrzysFR:
I just realized the hard way that the default build of zstdlib.dll on Windows using msbuild or the Visual Studio projects, does not include support for legacy formats out of the box (fails with `Unknown frame descriptor` when decoding older files).

This is due to the following reasons:
- `ZSTD_LEGACY_SUPPORT` is set to 0 by default in the .csproj settings.
- VS2008 project does not include `zstd_v07.h` so would only support v0.1 to 0.6?
- VS2010 projects includes neither `zstd_legacy.h` nor any of the various `zstd_v0x.h/.c` files (setting `ZSTD_LEGACY_SUPPORT=1` will do nothing at all)

I tried adding the files and changing `ZSTD_LEGACY_SUPPORT` to 1, and now it builds fine, BUT the library file size went from 256 KB (for x64/Release without support), to 360 KB with full support of v0.1 to v0.7.

In my particular case, I only need support for v0.6, but the current mechanism only allow ON or OFF and all the files must be included (or else msbuild complains with lots of undefined symbols). It would be nice to be able to define, for example, `ZSTD_LEGACY_MIN_VERSION = 060` (or equivalent) to say "I only want 0.6 and 0.7". Another approach would be to have individual switches to enable specific versions? (ex: I have old 0.3 files, a lot of 0.6 files, so would only need support for both..)

Questions:
- What should the default be for the VS2008 and VS2010 projects regarding legacy support? compiled by default or not? (current is not)
- Should the v1.0.x library include support for all legacy versions back to v0.1? or should the default only maybe include support back to 0.6 ? or maybe 0.3 or 0.4 ?

Below is a comparison of the size of the dll (release, x64, build with VS2013, with MSVCR 12.0) produced by manually removing support for various legacy versions:
- no legacy support: **256 KB**
- v0.7 only: **279 KB** (+23 KB)
- v0.6. to v0.7: **297 KB** (+41 KB)
- v0.3 to v0.7: **339 KB** (+83 KB)
- v0.1 to v0.7: **360 KB** (+104 KB)

It seems that each version adds ~20 KB of code.

inikep:
When I updated VS projects I didn't change behavior of ZSTD_LEGACY_SUPPORT:
[...]

PR #383, opened 2016-09-23
Add support for Legacy format to Visual Studios solutions

KrzysFR:
As discussed in #379, legacy support is missing from the VS solutions with default build settings (missing preprocessor definition, and all files not included).

This PR changes:
- define `ZSTD_LEGACY_SUPPORT` to 1 for zstdlib and zstd.exe (I'm not sure if they should be defined for the other tests projects?)
- do not define `ZSTD_HEAPMODE` to 0 for the zstdlib (since the default should be 1)
- add various missing zstd_v0x.c and .h files

Notes:
- I'm not able to test on VS2008/2005, so hopefully AppVeyor will be able to help use here?
- as is, I'm getting a 'warning treated as error' with the Debug build (not for Release). Don't know what's going on here, but this does not prevent the release build.

```C
c:\data\git\github\zstd\lib\legacy\zstd_v06.h(33): error C2220: warning treated as error - no 'object' file generated
c:\data\git\github\zstd\lib\legacy\zstd_v06.h(33): warning C4005: 'ZSTDLIB_API': macro redefinition
  c:\data\git\github\zstd\lib\zstd.h(27): note: see previous definition of 'ZSTDLIB_API'
```


facebook-github-bot-2:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

Cyan4973:
Still some issues apparently : 

```
*** 
*** Building Visual Studio 2008 x64\Debug in C:\projects\zstd-p0yf0 
*** 
Microsoft (R) Build Engine version 14.0.25420.1
Copyright (C) Microsoft Corporation. All rights reserved.
c:\projects\zstd-p0yf0\lib\legacy\zstd_v06.h(33): error C2220: warning treated as error - no 'object' file generated [C:\projects\zstd-p0yf0\build\VS2008\zstd.sln]
c:\projects\zstd-p0yf0\lib\legacy\zstd_v06.h(33): warning C4005: 'ZSTDLIB_API' : macro redefinition [C:\projects\zstd-p0yf0\build\VS2008\zstd.sln]
Command exited with code 1
```

Also :
note that signing CLA is actually important before merging any source code change.

KrzysFR:
That's the same issue I have, only in Debug (Release still running at the moment). I think this was not seen because the zstd project only included up to zstd_v05.c and so the problem never happend before...

note: I have already signed the CLA with our corporate account a few days ago, but it does not seem to have worked (received no notification). Will retry again when the person in charge of the corp account here is avail..

KrzysFR:
Looks there is two ways to solve this issue:

**Method 1:**

Adding a `#ifndef .... #endif` around the ZSTDLIB_API defines in both `zstd_v06.h` and `zstd_v07.h` seems to solve the debug build issue:

Ex: for zstd_v06.h
```C
/*!
*  ZSTDv06_DLL_EXPORT :
*  Enable exporting of functions when building a Windows DLL
*/
#ifndef ZSTDLIB_API
#  if defined(_WIN32) && defined(ZSTDv06_DLL_EXPORT) && (ZSTDv06_DLL_EXPORT==1)
#    define ZSTDLIB_API __declspec(dllexport)
#  else
#    define ZSTDLIB_API
#  endif
#endif

..

ZSTDLIB_API size_t ZSTDv06_decompress( void* dst, size_t dstCapacity,
                              const void* src, size_t compressedSize);
```
Does this look like the typical way to solve issues like this? This seems a bit hacky to me...

**Method 2:**

Use a different name for the macro to conditionally export methods for v6 and/or v7.

- The test does look specifically for `ZSTDv06_DLL_EXPORT=1`  so it seems that the intent here is to only export methods starting with `ZSTDv06_methodName()` if this is the case.
- ZSTD_LEGACY_SUPPORT seems to mean "I wand to decompress older files", which may not be the same as "I want to also be able to call `ZSTDv06_foo()` from my custom code".

So it seems like the .h should instead define a `ZSTDLIBv06_API` (and likewise `ZSTDLIBv07_API`) which will not overlap with the main `ZSTDLIB_API` used for non-legacy method.

```C
/*!
*  ZSTDv06_DLL_EXPORT :
*  Enable exporting of functions when building a Windows DLL
*/
#if defined(_WIN32) && defined(ZSTDv06_DLL_EXPORT) && (ZSTDv06_DLL_EXPORT==1)
#  define ZSTDLIBv06_API __declspec(dllexport)
#else
#  define ZSTDLIBv06_API
#endif

..

ZSTDLIBv06_API size_t ZSTDv06_decompress( void* dst, size_t dstCapacity,
                              const void* src, size_t compressedSize);
```

I think method 2 is better, but I'd like to have your opinion on this.

KrzysFR:
Pushed commit with method 2, will see if AppVeyor like this better. (6 x 18 minutes total build time? that's a bit slow :) )

also: signing the corporate cla (link in first paragraph of https://code.facebook.com/cla) does not seem to work (page says "Félicitations. Vous avez signé numériquement le contrat de licence de contributeur. Signé le : 23 septembre 2016 Veuillez imprimer cette page pour vos dossiers."), but we didn't receive any email nor feedback. I signed the CLA with my own account, and seems like this worked. Not sure what's going on here.

Cyan4973:
I suspect that's because the bot looks into a list of approved account, 
maybe sometimes this list takes time to update ...

KrzysFR:
Seems that the AppVeyor Debug build is fixed.

There's a `No artifacts found matching 'bin\zstd.exe' path` warning at the end of the build (https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.26/job/b1h5j203wid4s682#L920), not sure what that's about.

KrzysFR:
I found an old VM with VS2008 and verified that I can build the VS2008 solution (Debug and Release).

Cyan4973:
I checked the CLA list, you are in, indeed.
I wonder if it could be a temporary inactive bot bug...

inikep:
`No artifacts found matching 'bin\zstd.exe' path` is connected with automatic build of executables.

PR #384, opened 2016-09-23
Zlibwrapper

inikep:


zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -14,50 +14,48 @@
...
-#define ZBUFF_STATIC_LINKING_ONLY  /* ZBUFF_createCCtx_advanced */
-#include "zbuff.h"
 #include "zstd_internal.h"         /* defaultCustomMem */
 
 
 #define Z_INFLATE_SYNC              8
-#define ZWRAP_HEADERSIZE            4
+#define ZLIB_HEADERSIZE             4
+#define ZSTD_HEADERSIZE             ZSTD_frameHeaderSize_min
 #define ZWRAP_DEFAULT_CLEVEL        5   /* Z_DEFAULT_COMPRESSION is translated to ZWRAP_DEFAULT_CLEVEL for zstd */
```
Cyan4973: minor note :
default `cLevel` is `3` within `zstd` cli.
It looks like a logical choice for `ZWRAP_DEFAULT_CLEVEL`, except if you have better reasons to select otherwise.

inikep: Z_DEFAULT_COMPRESSION is 6 in zlib so I'm not sure if ZWRAP_DEFAULT_CLEVEL shouldn't also be 6

Cyan4973: Your comment makes it pretty clear that `Z_DEFAULT_COMPRESSION` is still default for zlib,
and `ZWRAP_DEFAULT_CLEVEL` is only default when `zstd` mode is active.

I find it fine to have different default values for different algorithms. After all, the scales are completely different.

inikep: It works like this:
- the compression level 0 for zlib is translated into Z_DEFAULT_COMPRESSION
- the compression level 0 for zstd is translated into ZWRAP_DEFAULT_CLEVEL
Since we are using zlibWrapper people might expect compression level 0 to get similar results from zstd as from zlib. I didn't make comparison but some time ago you've written that level 5 of zstd is similar to 6 of zlib.

Cyan4973: Nowadays, I tend to oppose zstd level 3 to zlib level 6.
On large sources, zstd has a little edge in compression ratio.
On small sources, zlib 6 tends to compress more.
In both cases, zstd 3 is much faster, which is its signature.

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -283,9 +366,12 @@ ZEXTERN int ZEXPORT z_deflateParams OF((z_streamp strm,
 /* *** Decompression *** */
 
 typedef struct {
-    ZBUFF_DCtx* zbd;
-    char headerBuf[ZWRAP_HEADERSIZE];
+    ZSTD_DStream* zbd;
+    char headerBuf[16]; /* should be equal or bigger than ZSTD_frameHeaderSize_min */
```
Cyan4973: you could use a static assert to ensure this condition at compile time

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -181,74 +246,92 @@ ZEXTERN int ZEXPORT z_deflate OF((z_streamp strm, int flush))
...
+        LOG_WRAPPERC("- deflate1 flush=%d avail_in=%d avail_out=%d total_in=%d total_out=%d\n", (int)flush, (int)strm->avail_in, (int)strm->avail_out, (int)strm->total_in, (int)strm->total_out);
+        res = deflate(strm, flush);
         return res;
     }
 
     zwc = (ZWRAP_CCtx*) strm->state;
+    if (zwc == NULL) { LOG_WRAPPERC("zwc == NULL\n"); return Z_STREAM_ERROR; }
+
+    if (zwc->zbc == NULL) {
+        int res = ZWRAP_initializeCStream(zwc, (flush == Z_FINISH) ? strm->avail_in : 0);
```
Cyan4973: My understanding here is that `ZWRAP_initializeCStream()` is invoked with the size of first source _if available_ `(flush == Z_FINISH)`.

So I guess that's where start the discussion : 
what happens if first source is very small, resulting in very small compression parameters, but later sources using the same context are much larger, resulting in poor compression ratios for these later sources ?

I guess the impact will be more clearly visible if there are some tests to show it.

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -181,74 +246,92 @@ ZEXTERN int ZEXPORT z_deflate OF((z_streamp strm, int flush))
...
+        if (res != Z_OK) return ZWRAPC_finishWithError(zwc, strm, res);
+    } else {
+        if (strm->total_in == 0) {
+            size_t const errorCode = ZSTD_resetCStream(zwc->zbc, (flush == Z_FINISH) ? strm->avail_in : zwc->pledgedSrcSize);
```
Cyan4973: Potential idea to avoid compression parameters adaptation issue in prior comment :
use `ZSTD_initCStream_advanced()` here
when new source to compress is larger, or 2x larger, than source size used to init existing `CStream`, or even always,
_when there is no dictionary_. `ZSTD_resetCStream()` is definitely very important for performance when there is a dictionary.

Difference between using `ZSTD_resetCStream()` sometimes or never should be measured with a benchmark.
Performance differences can only become visible if successive source sizes feature non negligible variations.

zlibWrapper/zstd_zlibwrapper.c
inikep: OK, changed to 3

zlibWrapper/zstd_zlibwrapper.c
inikep: added `MEM_STATIC_ASSERT(sizeof(zwd->headerBuf) >= ZSTD_frameHeaderSize_min);`

zlibWrapper/zstd_zlibwrapper.c
inikep: This was exactly my issue with `ZSTD_resetCStream()`

zlibWrapper/zstd_zlibwrapper.c
```diff
@@ -153,24 +198,44 @@ ZEXTERN int ZEXPORT z_deflateInit2_ OF((z_streamp strm, int level, int method,
...
-        LOG_WRAPPER("- deflateSetDictionary level=%d\n", (int)strm->data_type);
-        { size_t const errorCode = ZBUFF_compressInitDictionary(zwc->zbc, dictionary, dictLength, zwc->compressionLevel);
-          if (ZSTD_isError(errorCode)) return Z_MEM_ERROR; }
+        LOG_WRAPPERC("- deflateSetDictionary level=%d\n", (int)zwc->compressionLevel);
+        if (!zwc) return Z_STREAM_ERROR;
+        if (zwc->zbc == NULL) {
+            int res = ZWRAP_initializeCStream(zwc, 0);
+            if (res != Z_OK) return ZWRAPC_finishWithError(zwc, strm, res);
+        }
+        { size_t const errorCode = ZSTD_initCStream_usingDict(zwc->zbc, dictionary, dictLength, zwc->compressionLevel);
```
inikep: fixed

zlibWrapper/zstd_zlibwrapper.c
inikep: Now wrapper calls `ZSTD_initCStream_advanced` when there is no dictionary and `ZSTD_resetCStream` with a dictionary.

Cyan4973:
👍

PR #385, opened 2016-09-23
Update .gitignore for new location of msbuild projects

KrzysFR:
It seems that when the projects folder was moved to the new path in cfe5fe45819804b6ef148dc8524fcec1fcd1fc43, the `build/bin` was changed to `build/` instead of `bin/` and building makes a lot of stuff show up in git.


Cyan4973:
Good point !

PR #387, opened 2016-09-23
[pzstd] Print (de)compression results

terrelln:
At the end of (de)compression with verbosity > 1, print the same result line that zstd prints.  The result line should be printed in exactly the same conditions as zstd prints it.

Live updates of the (de)compression status is not currently included, as the writer thread doesn't know how much input corresponds to the output it receives.  This is doable, but I think it should be put off to a future version.  The change involves adding extra shared state, and I don't want to make changes to the multithreaded part of the code without having enough time to test them.

### DIFF
--- a/contrib/pzstd/Pzstd.cpp
+++ b/contrib/pzstd/Pzstd.cpp
@@ -14,6 +14,7 @@
 #include "utils/ThreadPool.h"
 #include "utils/WorkQueue.h"
 
+#include <chrono>
 #include <cstddef>
 #include <cstdio>
 #include <memory>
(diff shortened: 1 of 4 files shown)


contrib/pzstd/Pzstd.cpp
```diff
@@ -14,6 +14,7 @@
 #include "utils/ThreadPool.h"
 #include "utils/WorkQueue.h"
 
+#include <chrono>
```
Cyan4973: As a minor detail,
I like when a reason to add an include is provided in comment.
Typically, which function / type / constant is used.

PR #404, opened 2016-10-06
Build separation

pixelb:
This is used in the rpm package to allow building the libs and programs in the %build stage,
and installation separated in the %install stage

### DIFF
--- a/contrib/pzstd/Makefile
+++ b/contrib/pzstd/Makefile
@@ -37,9 +37,8 @@ default: pzstd
 all: pzstd
 
 
-libzstd.a: $(ZSTD_FILES)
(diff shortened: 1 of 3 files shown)


terrelln:
I'm rewriting the pzstd Makefile right now.  I have a draft on my [dev branch](https://github.com/terrelln/zstd/commit/91c5b9256a513dc86b6b1fcdf6d27e9401392733).  If you have any thoughts, please let me know.

Cyan4973:
Should we drop this PR, in favor of @terrelln one ?

pixelb:
Definitely drop the pzstd one.
Something similar to the first commit would be good though.
I'll close this as that's trivial enough

PR #406, opened 2016-10-06
pzstd: fix test failure on headless build

pixelb:
If I/O is not connected to a terminal
(like in a package build system for example),
a portion of the test expected to fail would pass.
This removes the system dependent test.


terrelln:
This is just a bad test, since its testing implementation details, and should be deleted.

PR #417, opened 2016-10-13
Fix ubsan failures (pass NULL to memcpy)

terrelln:
Each of the places fixed passed `NULL` to `memcpy()` during the run of `./contrib/pzstd/test/PzstdTest`.  The ubsan failures only showed up when testing with `qemu-aarch64`.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2235,7 +2235,7 @@ static size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
     BYTE* op = ostart;
     U32 const maxDist = 1 << cctx->params.cParams.windowLog;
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
OK,
I'm just a bit surprised that the warning only shows up with `qemu-aarch64`, since undefined is supposed to be undefined, whatever the platform,
and we already have undefined behavior check setup on Travis CI : https://travis-ci.org/facebook/zstd/jobs/167199851#L140

So a related question could be : how come this issue did not show up before in usan tests ? Do we have to improve this test ?

terrelln:
I was able to catch the same errors with gcc-5's ubsan.  However, all previous ubsan runs I did were with gcc-4.9, which doesn't catch them.

PR #421, opened 2016-10-19
Fix uninitialized memory access

spaskob:
if result is non-zero fParams field is not guaranteed to be initialized.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -306,6 +306,7 @@ unsigned long long ZSTD_getDecompressedSize(const void* src, size_t srcSize)
 static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t srcSize)
 {
     size_t const result = ZSTD_getFrameParams(&(dctx->fParams), src, srcSize);
+    if (result > 0) return ERROR(srcSize_wrong);  /* Don't access fParams if call to ZSTD_getFrameParams is not successful */
     if (dctx->fParams.dictID && (dctx->dictID != dctx->fParams.dictID)) return ERROR(dictionary_wrong);
     if (dctx->fParams.checksumFlag) XXH64_reset(&dctx->xxhState, 0);
     return result;


lib/decompress/zstd_decompress.c
```diff
@@ -306,6 +306,7 @@ unsigned long long ZSTD_getDecompressedSize(const void* src, size_t srcSize)
 static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t srcSize)
 {
     size_t const result = ZSTD_getFrameParams(&(dctx->fParams), src, srcSize);
+    if (result > 0) return ERROR(srcSize_wrong);  /* Don't access fParams if call to ZSTD_getFrameParams is not successful */
```
terrelln: It would be better to use `CHECK_F(ZSTD_getFrameParams(&(dctx->fParams), src, srcSize));`

lib/decompress/zstd_decompress.c
```diff
@@ -306,6 +306,7 @@ unsigned long long ZSTD_getDecompressedSize(const void* src, size_t srcSize)
...
     if (dctx->fParams.dictID && (dctx->dictID != dctx->fParams.dictID)) return ERROR(dictionary_wrong);
     if (dctx->fParams.checksumFlag) XXH64_reset(&dctx->xxhState, 0);
     return result;
```
terrelln: and `return 0;`

lib/decompress/zstd_decompress.c
Cyan4973: It's not equivalent.
Here, the test returns an error code when `result > 0`, which is different from `ZSTD_isError(result)`.

I will have a closer look into the failure scenario, to see when it could cause troubles.

Cyan4973:
`ZSTD_decodeFrameHeader()` is a `static` function, hence cannot be called directly.

It is used within only 2 places :
 
- `ZSTD_decompressContinue()` : in this context, the function is necessarily called with the right number of bytes (the header size) as it is checked with `dctx->expected`. 
So it cannot trigger the case where `ZSTD_getFrameParams() > 0`, which is only possible when `srcSize` is too small. It protects all variants of streaming functions.

- `ZSTD_decompressFrame()` : this one is called by "one-pass" decompression functions (such as `ZSTD_decompress()`), and expects the full frame size as a single input. The call to `ZSTD_decodeFrameHeader()` is preceded by 2 tests : 
  + First it is checked that the frame has the minimum size possible (6+3 = 9 bytes), which guarantee there is enough size to get the header size (5 bytes).
  + Second, the frame size is checked to ensure it is valid, and at a minimum it is `headerSize + 3` which is obviously `> headerSize`. Consequently, the call to `ZSTD_decodeFrameHeader()` can be issued with the correct size, hence cannot return `ZSTD_getFrameParams() > 0` (expect for invalid header content, which returns an error code).

Consequently, the error mentioned cannot happen in current version of the code. @spaskob , if you have experienced otherwise, please tell us.

A more complex question is : ok, say it works fine with current code, but could a later careless usage of this function trigger this code path, leading to wrong execution ?

And I would agree with that.
Even if the bug cannot be triggered today, it's nonetheless better to check for it, in case a future code modification would accidentally trigger it.

So I'll add it to `dev` branch.

PR #424, opened 2016-10-22
Dictionary Builder: offcodeMax is too small for 3 dictionary sizes

terrelln:
If `2^N - 4 < dictContentSize + 128 KB < 2^N` then the dictionaries
`offcodeMax` will be too small to represent a offset from the end
of the first block back to the beginning of the dictionary content.
Not that this offset is useful, since the offset is more bits than
the match could possibly be.


Cyan4973:
I suggest to drop this patch.

The presence of this `offcode` in the stats is not free, it (slightly) decreases compression ratio for all other symbols.

But this new `offcode` will never be used.

The condition to need such `offcode` is to generate a match within the last 7 bytes of input data.
But the reference encoder never generates such a match.
So it will never need it.

PR #432, opened 2016-10-28
windres

inikep:


Cyan4973:
minor comment on default `timeout` (see asana),
all good for the patch at large

PR #439, opened 2016-11-03
ZSTD_compress_usingDict() specify dict == NULL

terrelln:
Specify that when `dict == NULL || dictSize < 8` no dictionary
gets loaded.
Also add some periods.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -1135,7 +1135,8 @@ When `Repeated_Offset2` is used, it's swapped with `Repeated_Offset1`.
 Dictionary format
 -----------------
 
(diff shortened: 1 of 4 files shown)


Cyan4973:
Thanks, that's a good point to clarify @terrelln .

The initial intention was that if _either_ `dict == NULL || dictSize == 0`, then no dictionary is loaded.

A point opened for discussion is that maybe that's too permissive.
That is, if someone specify `dict == NULL` and `dictSize == 12345`,
is it more logical to consider it as a runtime error, instead of just transparently not loading anything ?

Looks so. But then, all tests should reflect that.
Since the code was initially created with the permissive spec in mind, I'm not sure if it currently checks this double condition.

PR #441, opened 2016-11-03
preserve owner, file permission and modification time

inikep:


Makefile
```diff
@@ -14,10 +14,12 @@ ZWRAPDIR = zlibWrapper
 TESTDIR  = tests
 
 # Define nul output
+VOID = /dev/null
```
Cyan4973: Does that mean `VOID` is never used in combination with Windows ?

inikep: /dev/null works fine with MSYS

Cyan4973:
There were a few problem in `dev` branch when this patch was submitted.
Fixing them unfortunately created a conflict...

Worth a rebase.

zlibWrapper/examples/zwrapbench.c
```diff
@@ -130,7 +130,7 @@ void BMK_SetBlockSize(size_t blockSize)
 **********************************************************/
 typedef struct
 {
-    const char* srcPtr;
+    char* srcPtr;
```
Cyan4973: A minor question : 
it's not a big deal since `zwrapbench` is a test program,
but is there any specific reason you need `srcPtr` to be non-`const` ?

Cyan4973: I see the warning now on OS-X `clang` : 
`warning: assigning to 'z_Bytef *' (aka 'unsigned char *') from 'const void *' discards qualifiers`

That's strange, because `next_in` is supposed to be `z_const Bytef *`, no just `Bytef *`.

So, for some reason, `z_const` is somewhat translated into nothing.

`z_const` is defined into `zconf.h` (which is included by `zlib.h`) : 
```
#if defined(ZLIB_CONST) && !defined(z_const)
#  define z_const const
#else
#  define z_const
#endif
```

A solution could be to `#define ZLIB_CONST` 
or even directly `#define z_const const`
before including `zlib.h` .

It's unclear why this design choice was made.
Looks like a design error to me.
A `const` property should not be optional.

Cyan4973: made a test by adding `#define ZLIB_CONST` before `#include <zlib.h>` within `zstd_zlibwrapper.h`, fixed the `const` issue within `zwrapbench`.

inikep: The issue here is that `z_const` and `ZLIB_CONST` were introduced in zlib 1.2.6 and we want to support 1.2.3+.

inikep:
I think I found a solution.
`zwrapbench` will use `z_const` that is `const` for zlib 1.2.6+ with `#define ZLIB_CONST` and ` ` for older.

inikep:
It seems to be Travis error. It works fine in my branch:
https://travis-ci.org/inikep/zstd/builds/173314467

Cyan4973:
Yes, looks so, test restarted

PR #442, opened 2016-11-03
The static library was moved to libzstd.a

terrelln:
This should fix the build, but we'll see what CI says.


Cyan4973:
Travis is still sluggish to take off,
but Appveyor is already working better with this patch.

PR #450, opened 2016-11-14
Dev11

inikep:


Cyan4973:
What does bring this patch ?

I noticed the usage of `QEMU_SYS` which looks fine.

I also noticed the new Visual project `libzstd`. But there is already an existing `zstdlib` project (which has apparently been renamed `libzstd-dll`).
How do they compare ?
What's the impact on `build/VS_scripts` ?

inikep:
This patch has:
- added Travis CI tests: `aarch64test` and `ppc64test`
- `libzstd` (static) and `libzstd-dll` (dynamic) libraries

`build/VS_scripts` will build now also: 
- `libzstd` as `libzstd_static_x64.lib` or `libzstd_static_x86.lib`
- `libzstd-dll` as `libzstd_x64.dll` or `libzstd_x86.dll`

KrzysFR:
I think you could probably merge both `libzstd` and `libstd-dll` into a single `.csproj`, by introduction two new targets, like maybe `ReleaseStatic` and `DebugStatic` in addition to the existing `Release` and `Debug` (maybe need to double that for 32 vs 64, I guess..?)

By the way, I initially named the dlls with the `_x64.dll'  and `_x86.dll` suffixes to simplify a bit the scenario of redistributing both versions in setups, when all dlls are in the same folder.

There is also the other approach of having both versions with the same file name, but in diffrent subfolders (usually named `x64` and `x86`). I'm not sure how you deal with this issues on *nixes, but since you are changing the name of the file, and want to have the same naming conventions on all platforms, it would be the ideal time to update this (`libzstd_x64.dll` vs `x64/libzstd.dll`) ?

A bit more context on why this would be important, for people using "native" libraries in .NET (historically Windows only, but now also applies to linux and *nixes with .NET Core):
- If all version of the library have the same file name (in different folders), it simplifies a bit the way you do P/Invoke on .NET: you must specify the name of the dll as a constant in the `[DllImport("foo.dll")]` declaration in your code. Then you can "help" the CLR load the correct file by manually loading the native dll in the process using `LoadLibrary("path/to/your/x64/foo.dll")` and the .NET CLR will magically see that it's already loaded, and use it directly, otherwise it would try to find it by itself, and fail (it will looks in the current dir, not in subfolders). You can't do this when the dll does not have the same filename for x64 vs x86, but it MUST match the constant string in the code :(
- On the other hand, if you don't use `[DllImport(...)]` but instead load the library yourself, and extract method pointers and create delegates by hand, you can name the native library anything you want, and have more flexibility at runtime (like only loading part of the API, or probing for the existence of specific exported methods).

I don't know if the same restrictions apply to Java on Windows, maybe someone with some experience on this could weight on this?

inikep:
1. I like the naming `_x64.dll` and `_x86.dll` but I think the endings should be removed to allow `[DllImport("foo.dll")]` to work.
2. `libzstd.dll` has the import library `libzstd.lib` so there was a collision with a static library. What do you think about the name `libzstd_static.lib`?

KrzysFR:
For 1. the developer will still need to "do" something, either at setup time by copying the correct file to the application folder depending on the arch of the system (makes things more complex, and disallow easy "XCOPY" deployment), OR need to make sure to manually `LoadLibrary(...)` the file *before* the first call to a P/Invoked method. So either way, it is not magical.

For 2. I'm not sure what you mean by `.lib`?  If my memory is correct (last time I used a .lib with .NET was probably 10years+ ago when COM was still not dead yet...), you only need the `.dll` at runtime when using "native" libraries (build from regular C or C++, with no COM in it), while the `.lib` would be needed at compile time for COM objects.

Anyway, the `static` suffix looks fine to me. Though up to now, I've only used `dll` that used dynamic linking (hence the need to compile against the correct version of the MSVC runtime), while zstd was the first one statically linked.

Be aware though, that if you change the name of the library to be the same for all platforms, people already using it will either need to change the way they load the library in .NET, OR more probably, need to rename the files before importing them in their projects (what I'll do probably), and still redist them with the previous names. Probably not that big of a deal.

inikep:
Thanks KrzysFR.

The Import Library (https://msdn.microsoft.com/en-us/library/f0z8kac4.aspx) is required when compiling with Visual C/C++ and linking executable with DLL.

KrzysFR:
Oh ok, I thought you were talking about using it with .NET (which only requires the dll).

PR #453, opened 2016-11-15
fullbench-dll

inikep:


Cyan4973:
A description of the change is welcomed.
My understanding is that this patch adds `ZSTD_DLL_IMPORT`

inikep:
The main change is the `fullbench-dll` project for Visual Studio that shows how to use `libzstd-dll`.

inikep:
`ZSTD_decodeSeqHeaders` and `ZSTD_decodeLiteralsBlock` are not available from DLL therefore they are disabled with `ZSTD_DLL_IMPORT`.

PR #460, opened 2016-11-23
Dev11

inikep:


programs/bench.c
```diff
@@ -483,12 +483,18 @@ static void BMK_syntheticTest(int cLevel, int cLevelLast, double compressibility
     free(srcBuffer);
 }
 
+#define ZSTD_MAX_CLEVEL     22
```
Cyan4973: It's possible to use `ZSTD_maxCLevel()` to get this value instead.
`ZSTD_maxCLevel()` is runtime, so it eliminates potential future synchronisation problems (compared to a local `#define`).

PR #461, opened 2016-11-29
libzstd.pc.in: Change to use variables for libdir and includedir

obache:
It is safe to use such variables for future maintainance.

### DIFF
--- a/lib/libzstd.pc.in
+++ b/lib/libzstd.pc.in
@@ -10,5 +10,5 @@ Name: zstd
 Description: lossless compression algorithm library
 URL: https://github.com/facebook/zstd
 Version: @VERSION@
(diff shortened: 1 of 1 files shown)


Cyan4973:
Could you describe what is the gain achieved by changing these variables ?

obache:
One thing should be defined once.
It is usual way used as examples in pkg-config document.

You can just change one line, `libdir=@LIBDIR@` , if you want to change stub string `@LIBDIR@`, its  meaning, and so on. 
Users can retrieve `libdir` from pc file safely with `pkg-config --variable=libdir`, because it is guaranteed to be matched in `Libs`. And vice versa, users can change such variables temporary with `pkg-config --define-variable=libdir=/foo/bar` safely for testing and so on.

PR #464, opened 2016-11-29
Fix ZSTD_STATIC_LINKING_ONLY with double include

terrelln:
Allow static linking section to be included even if zstd has already been included without it.
Minimal repro:
    
    #include <zstd.h>
    #define ZSTD_STATIC_LINKING_ONLY
    #include <zstd.h>

    ZSTD_frameParams params; /* Compiler error: ZSTD_frameParams not defined */


Cyan4973:
Very good point

PR #467, opened 2016-12-02
gzip decompression from CLI

inikep:


programs/fileio.c
```diff
@@ -647,6 +651,53 @@ static unsigned FIO_passThrough(FILE* foutput, FILE* finput, void* buffer, size_
 }
 
 
+#ifdef ZSTD_GZDECOMPRESS
+static size_t FIO_decompressGzFile(dRess_t ress, size_t headBufSize, const char* srcFileName, FILE* srcFile)
```
Cyan4973: Just read your comment on Asana.
https://app.asana.com/0/91843330278060/155387480861121/f

So I guess it means it works at Frame level, and therefore is compatible with `stdin`, and mixed zstd/gzip streams.

inikep: I started with gz files and went to frame so now `decompressGzFrame` will be even better name.

PR #474, opened 2016-12-10
Fix compression segfault

terrelln:
The bug is exposed by [test.c](https://gist.github.com/terrelln/6b7c01740fc05b0bcf5b3572056c9869), and the test case is fixed after the patch.  I'm still unfamiliar with the compression code, so there maybe be consequences that I don't see, but I think this patch works.

When the overflow protection kicks in, it makes sure that `ip - ctx->base` isn't too large. However, it didn't ensure that saved offsets are still valid. This change ensures that any valid offsets (<= windowLog) are still representable after the update.

The bug would shop up on line 1056, when `offset_1 > current + 1`, which causes an underflow. This in turn, would cause a segfault on line 1063.

The input must necessarily be longer than 1 GB for this issue to occur. Even then, it only occurs if one of the last 3 matches is larger than the chain size and block size.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2277,7 +2277,7 @@ static size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
         if (cctx->lowLimit > (1<<30)) {
             U32 const btplus = (cctx->params.cParams.strategy == ZSTD_btlazy2) | (cctx->params.cParams.strategy == ZSTD_btopt) | (cctx->params.cParams.strategy == ZSTD_btopt2);
             U32 const chainMask = (1 << (cctx->params.cParams.chainLog - btplus)) - 1;
-            U32 const supLog = MAX(MAX(cctx->params.cParams.chainLog, 17 /* blockSize */), cctx->params.cParams.windowLog);
+            U32 const supLog = MAX(cctx->params.cParams.windowLog, 17 /* blockSize */);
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -2277,7 +2277,7 @@ static size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
         if (cctx->lowLimit > (1<<30)) {
             U32 const btplus = (cctx->params.cParams.strategy == ZSTD_btlazy2) | (cctx->params.cParams.strategy == ZSTD_btopt) | (cctx->params.cParams.strategy == ZSTD_btopt2);
             U32 const chainMask = (1 << (cctx->params.cParams.chainLog - btplus)) - 1;
-            U32 const supLog = MAX(cctx->params.cParams.chainLog, 17 /* blockSize */);
+            U32 const supLog = MAX(MAX(cctx->params.cParams.chainLog, 17 /* blockSize */), cctx->params.cParams.windowLog);
```
Cyan4973: Since `windowLog >= chainLog`, we can probably simplify a bit this expression.

Cyan4973:
Great @terrelln ! I believe you tacked the root cause and solution.
Does that sound possible to add the test case into the CI test suite ?

lib/compress/zstd_compress.c
terrelln: The raw chain log can be up to `windowLog + 1`, but in the pervious line that extra 1 was subtracted off for computing the `chainMask`.  Is it safe to subtract it here as well?

PR #475, opened 2016-12-11
Make zstd build out of the box properly on FreeBSD

DimitryAndric:
* Add `-D_POSIX_C_SOURCE=200112L` to enable POSIX functionality
* Remove last bashism from `tests/playTests.sh`
* Use `gmd5sum` from the sysutils/coreutils port

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -98,7 +98,7 @@ extern "C" {
 #    define SET_HIGH_PRIORITY /* disabled */
 #  endif
 #  define UTIL_sleep(s) sleep(s)
-#  if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L)
+#  if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || (defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L))
 #      define UTIL_sleepMilli(milli) { struct timespec t; t.tv_sec=0; t.tv_nsec=milli*1000000ULL; nanosleep(&t, NULL); }
 #  else
 #      define UTIL_sleepMilli(milli) /* disabled */
--- a/contrib/pzstd/Options.cpp
+++ b/contrib/pzstd/Options.cpp
@@ -22,18 +22,12 @@
     defined(__CYGWIN__)
 #include <io.h> /* _isatty */
 #define IS_CONSOLE(stdStream) _isatty(_fileno(stdStream))
-#else
-#if defined(_POSIX_C_SOURCE) || defined(_XOPEN_SOURCE) ||                      \
-    defined(_POSIX_SOURCE) ||                                                  \
-    (defined(__APPLE__) &&                                                     \
-     defined(                                                                  \
-         __MACH__)) /* https://sourceforge.net/p/predef/wiki/OperatingSystems/ \
-                       */
+#elif defined(_POSIX_C_SOURCE) || defined(_XOPEN_SOURCE) || defined(_POSIX_SOURCE) || (defined(__APPLE__) && defined(__MACH__)) || \
+      defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)  /* https://sourceforge.net/p/predef/wiki/OperatingSystems/ */
 #include <unistd.h> /* isatty */
 #define IS_CONSOLE(stdStream) isatty(fileno(stdStream))
 #else
 #define IS_CONSOLE(stdStream) 0
-#endif
 #endif
 
 namespace pzstd {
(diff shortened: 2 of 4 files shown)


lib/Makefile
```diff
@@ -41,6 +41,10 @@ CPPFLAGS  += -I./legacy -DZSTD_LEGACY_SUPPORT=1
 ZSTD_FILES+= $(wildcard legacy/*.c)
 endif
 
+ifeq ($(shell uname), FreeBSD)
+CPPFLAGS  += -D_POSIX_C_SOURCE=200112L
```
Cyan4973: It seems this detection macro is not useful in `/lib`.
Feel free to explain if you believe differently.

programs/Makefile
```diff
@@ -52,6 +52,10 @@ CPPFLAGS  += -I$(ZSTDDIR)/legacy
 ZSTDLEGACY_FILES:= $(ZSTDDIR)/legacy/*.c
 endif
 
+ifeq ($(shell uname), FreeBSD)
+CPPFLAGS  += -D_POSIX_C_SOURCE=200112L
```
Cyan4973: I tend to prefer such detection macro to be set from source code itself. My understanding is that this macro should be enabled from `utils.h`. Is there a safe way to do so ?
cc @inikep , he has some experience on this topic.

Cyan4973:
Totally agree on the objective.
Let's work out the last details.

programs/Makefile
inikep: 1. The issue concerns the availability of `-r` (recursive) option that uses `opendir/closedir` functions (http://man.he.net/man3/closedir) conforming to SVr4, POSIX.1-2001, 4.3BSD

2. The `_POSIX_C_SOURCE=200112L` macro is used to detect availability of `opendir/closedir` functions on Linux. My understanding is that `_POSIX_C_SOURCE` shouldn't be defined by user at all because it may not work e.g. with older versions of an operating system and `glibc`. It's better to use `--std=gnu99` or `-D_GNU_SOURCE` which should be more portable as they define a proper version `_POSIX_C_SOURCE`.

3. The proposed changes force the C parser to turn on '-r' on FreeBSD. But it's better to do auto detection of FreeBSD only for relevant file that is `util.h`. It seems that all BSD-forks are based on 4.4BSD therefore we need only to check availability of one of:
__DragonFly__
__FreeBSD__
__NetBSD__
__OpenBSD__

inikep:
This PR should turn on the '-r' option for *BSD and Solaris:
https://github.com/facebook/zstd/pull/476
I don't have access to *BSD or Solaris to test it.

Cyan4973:
With latest patch by @inikep , recursive mode `-r` should work properly on FreeBSD.
There should be no need to add `_POSIX_C_SOURCE` to `CPPFLAGS`.
Could you please remove this part ?
This will allow us to integrate the rest of the patch, which looks good.

DimitryAndric:
@Cyan4973, I have rebased, and split this up in two separate commits:
* Changes to `tests/playTests.sh` to make it run correctly on FreeBSD
* Changes to `contrib/pzstd/Options.cpp`, `programs/util.h` and `programs/zstdcli.c` to enable use of `isatty()` and `nanosleep()` on the BSDs.
Let me know if you'd rather squash this, or put it in separate pull requests.

programs/util.h
```diff
@@ -98,7 +98,7 @@ extern "C" {
 #    define SET_HIGH_PRIORITY /* disabled */
 #  endif
 #  define UTIL_sleep(s) sleep(s)
-#  if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L)
+#  if defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || (defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L))
```
inikep: http://man.openbsd.org/nanosleep.2 says:
The nanosleep() system call has been available since NetBSD 1.3 and was ported to OpenBSD 2.1 and FreeBSD 3.0.

It means that this change will not work e.g. with FreeBSD older than 3.0 (Fri, 16 Oct 1998).

DimitryAndric: Well, in practice nobody runs these old versions anymore.  The oldest supported version of FreeBSD is 9.3, and even that will reach end-of-life on 2016-12-31.  In practice, there may be a few equipment vendors that run older forked versions of FreeBSD, but certainly not older than FreeBSD 7.x or 8.x.

inikep: I'm sure you are right but I wanted to make it clear for @Cyan4973.
There are some OSes (e.g. Windows, CentOS/Red Hat Enterprise Linux) with 10-years maintenance updates but FreeBSD 3.0 is 18-years old.

Cyan4973: You would be surprised how old some systems can be out there...

Nevertheless, there are portability risks that are reasonable to assume, and not supporting FreeBSD < 3.0 is one of them. We'll learn from user feedbacks if there is any need to take care of this use case.

PR #477, opened 2016-12-12
ZStd-compression:  cParams.strategy validity check

ssakash:
**Summary of changes**:

* Improve check for unsupported compression parameter.


Cyan4973:
Compilation fails because `enum` are not naturally ordered, then must be casted beforehand.

Note that the added test doesn't look very useful, since `enum` start at zero by default. Therefore, `ZSTD_fast` gets value `0`, so this side of clamping is always correct.

ssakash:
Well, I could still update it by explicit casting and the compilation will succeed. why was this PR closed? I don't see how clamping on both sides is not better instead of only throwing unsupported error for values over `` ZSTD_btopt2``.

ssakash:
Oh, nevermind. I see that you're casting it to an unsigned integer which means that there's no possibility for a negative value. So don't bother re-opening the pull request, it's useless.

PR #483, opened 2016-12-14
fullbench: added test of ZSTD_compressContinue_extDict

inikep:


Cyan4973:
keeping this PR for next version

lib/compress/zstd_compress.c
```diff
@@ -2240,7 +2245,7 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
 
 static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc, void* dst, size_t dstCapacity, const void* src, size_t srcSize)
 {
-    ZSTD_blockCompressor const blockCompressor = ZSTD_selectBlockCompressor(zc->params.cParams.strategy, zc->lowLimit < zc->dictLimit);
+    ZSTD_blockCompressor const blockCompressor = ZSTD_selectBlockCompressor(zc->params.cParams.strategy, zc->seqStore.forceExtDict || (zc->lowLimit < zc->dictLimit));
```
Cyan4973: minor point : this creates an additional run-time test for a feature which is essentially a test.

inikep: I thought also about:
1) creating hidden `ZSTD_compressBlock_internal_extDict` with `extDict=1` but it forces to create plenty of other hidden functions.
2) creating a new function only to test `ZSTD_selectBlockCompressor` with `extDict=1`

Cyan4973:
I totally agree on the objective, but would favor a method avoiding any impact for normal runtime operations.

Does that sound possible to create a separate `ZSTD_compressBlock_internal()` version which specifically only select `*_extDict()` encoders ?

tests/fullbench.c
```diff
@@ -268,6 +271,13 @@ static size_t benchMem(const void* src, size_t srcSize, U32 benchNb)
         if (g_zcc==NULL) g_zcc = ZSTD_createCCtx();
         break;
     case 12 :
+        if (g_zcc==NULL) g_zcc = ZSTD_createCCtx();
+        if (g_zcc) {
+            seqStore_t* seqStorePtr = ZSTD_getSeqStore(g_zcc);
+            if (seqStorePtr) seqStorePtr->forceExtDict = 1;
```
inikep: Is it a problem? We have hidden functions so I think we can have hidden parameters. `seqStore_t` is not exposed in `zstd.h`. When we unlikely change `seqStore_t` we can change our testing programs.

lib/compress/zstd_compress.c
Cyan4973: > it forces to create plenty of other hidden functions.

Indeed, that can be a deal breaker if too large.
Do we have a precise idea of how many of them would be required ?

inikep:
The issue was that we should also create `extDict` versions of `ZSTD_compress_generic`, `ZSTD_compressContinue_internal`, `ZSTD_compressEnd`.

tests/fullbench.c
Cyan4973: It's a source of future maintenance difficulties (though hopefully its impact is likely going to be limited to `fullbench`). Nothing dramatic though, it's part of the trade-off to find.

lib/compress/zstd_compress.c
inikep: I thought also about:
3. Forcing `zc->lowLimit < zc->dictLimit` to `1` but I was afraid of side-effects.

Cyan4973: Good idea, but it's indeed a bit too risky.
A first obvious side effect is when `zc->dictLimit == 0`.
There are potentially others.
Having a dedicated variable feels a wiser choice.

Cyan4973:
If `fullbench` is the only target, there is a way to achieve the same objective without impacting `zstd` : 
start by compressing 8 bytes from somewhere in memory,
then proceed with compressing the real target data.

This will create a separate memory segment, hence trigger the `extDict` variant.

PR #487, opened 2016-12-16
Export ZBUFF_isError() and ZBUFF_getErrorName()

terrelln:
In `zstd_common.c`, since `zbuff.h` wasn't included, the symbols are by default
hidden in the shared library, and the exported symbols aren't defined.

I think I got changed every Makefile, but CI will confirm.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -27,7 +27,7 @@ else
 ALIGN_LOOP =
 endif
 
-CPPFLAGS+= -I$(ZSTDDIR) -I$(ZSTDDIR)/common -I$(ZSTDDIR)/dictBuilder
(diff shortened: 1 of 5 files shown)


programs/Makefile
```diff
@@ -27,7 +27,7 @@ else
 ALIGN_LOOP =
 endif
 
-CPPFLAGS+= -I$(ZSTDDIR) -I$(ZSTDDIR)/common -I$(ZSTDDIR)/dictBuilder
+CPPFLAGS+= -I$(ZSTDDIR) -I$(ZSTDDIR)/common -I$(ZSTDDIR)/dictBuilder -I./deprecated
```
Cyan4973: Is it necessary there ?
`ZBUFF_*` is never used from `zstd` nor any of its variants.

PR #488, opened 2016-12-16
Export ZBUFF_isError() and ZBUFF_getErrorName()

terrelln:
In `zstd_common.c`, since `zbuff.h` wasn't included, the symbols are by default
hidden in the shared library, and the exported symbols aren't defined.

### DIFF
--- a/lib/deprecated/zbuff_common.c
+++ b/lib/deprecated/zbuff_common.c
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) 2016-present, Yann Collet, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/*-*************************************
+*  Dependencies
+***************************************/
+#include "error_private.h"
+#include "zbuff.h"
+
+/*-****************************************
+*  ZBUFF Error Management  (deprecated)
+******************************************/
+
+/*! ZBUFF_isError() :
+*   tells if a return value is an error code */
+unsigned ZBUFF_isError(size_t errorCode) { return ERR_isError(errorCode); }
+/*! ZBUFF_getErrorName() :
+*   provides error code string from function result (useful for debugging) */
+const char* ZBUFF_getErrorName(size_t errorCode) { return ERR_getErrorName(errorCode); }
+
--- a/lib/dictBuilder/zdict.h
+++ b/lib/dictBuilder/zdict.h
@@ -19,15 +19,18 @@ extern "C" {
 #include <stddef.h>  /* size_t */
 
 
-/*======  Export for Windows  ======*/
-/*!
-*  ZSTD_DLL_EXPORT :
-*  Enable exporting of functions when building a Windows DLL
-*/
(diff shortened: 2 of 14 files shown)


lib/deprecated/zbuff_common.c
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+/*-*************************************
+*  Dependencies
+***************************************/
+#include "error_private.h"
```
Cyan4973: no need to include `zbuff.h` ?

Cyan4973:
Good point.

Could we have a test in CI environment which shows the shortcomings of pre-patched source code ?
This to ensure both that current patch improve it, and that similar issue will be checked in the future.

terrelln:
Should the functions `ZSTD_getErrorCode()` and `ZSTD_getErrorString()` be exported?

Cyan4973:
> Should the functions ZSTD_getErrorCode() and ZSTD_getErrorString() be exported?

Yes

terrelln:
I also exported all of the unexported advanced functions.

tests/symbols.c
```diff
@@ -0,0 +1,144 @@
...
+#include "zstd_errors.h"
+#define ZSTD_STATIC_LINKING_ONLY
+#include "zstd.h"
+#define ZBUFF_DISABLE_DEPRECATE_WARNINGS
+#define ZBUFF_STATIC_LINKING_ONLY
+#include "zbuff.h"
+#define ZDICT_STATIC_LINKING_ONLY
+#include "zdict.h"
+
+static const void *symbols[] = {
```
Cyan4973: This is a great way to test presence of symbols.

Obviously, it features the maintenance shortcoming that it needs to be manually edited whenever a new prototype is added.
As it stands, it's very easy to create a new prototype and forget to add it to `symbols.c`, and then test passes correctly...

Unfortunately, I also have no really better alternative to propose.

I was vaguely considering to create a build mode for `tests` programs, that would use dynamic `libzstd` instead of compiling and linking directly `lib` source code.
Of course, there's a bit of the same problem, a new prototype needs to be invoked from at least one of these test programs to check its presence. But it's supposed to be the case, any prototype should have at least one basic unit test associated.

terrelln: I think unit tests will only get us protection from regressions for this failure, not protection against failures in new functions.

To solve this problem I think you would need a linter that runs on commits.

lib/dictBuilder/zdict.h
```diff
@@ -19,13 +19,13 @@ extern "C" {
...
 
 
-/*======  Export for Windows  ======*/
-/*!
-*  ZSTD_DLL_EXPORT :
-*  Enable exporting of functions when building a Windows DLL
-*/
-#if defined(_WIN32) && defined(ZSTD_DLL_EXPORT) && (ZSTD_DLL_EXPORT==1)
+/* =====   ZDICTLIB_API : control library symbols visibility   ===== */
+#if defined(__GNUC__) && (__GNUC__ >= 4)
```
inikep: This breaks DLL compiled with gcc/MinGW. A fix for zstd.h is available here:
https://github.com/inikep/zstd/commit/60f10aab6c26dda2224583de84fe64ab7a649f55

PR #490, opened 2016-12-18
Fix cmake build (MSVC)

jacquesg:


Cyan4973:
cc @inikep , as I'm unsure if the proposed change is correct.
It seems to me it should better be `libzstd`.

I also note that we have currently no test to check this.
Maybe we need to add one in Appveyor.

jacquesg:
This addresses a regression introduced in the 1.1.2 branch. There seems to be 2 rc files, ```build/VS2010/libzstd-dll/libzstd-dll.rc``` and ```build/VS2010/zstd/zstd.rc```. The latter seems to be more up to date.

inikep:
A good catch. The patch however is not good because `cmake/lib/CMakeLists.txt` generates a static and a dynamic library but `zstd.rc` is for `zstd.exe` as it contains e.g.
`VALUE "OriginalFilename", "zstd.exe"`
We should use `build/VS2010/libzstd-dll/libzstd-dll.rc`.

jacquesg:
Ok. Should I also bring the content of the 2 rc files in line with each other in terms of the ```LegalCopyright``` etc?

inikep:
Thanks, but I already updated it in my branch.

jacquesg:
I've updated the PR. I've also added an extra commit which doesn't add the rc file to the static library.

inikep:
You're right, the rc file is not required for the static library. But the line you've removed caused some errors detected with Travis CI:
`CMake Error at lib/CMakeLists.txt:147 (SET_TARGET_PROPERTIES):
  set_target_properties Can not find target to add properties to:
  libzstd_static
the variable PREFIX=~/install_test_dir
CMake Error at lib/CMakeLists.txt:169 (INSTALL):
  install TARGETS given target "libzstd_static" which does not exist in this
  directory.`

inikep:
I think that
`ADD_LIBRARY(libzstd_static STATIC ${Sources} ${Headers} ${PlatformDependResources})`
should be replaced with
`ADD_LIBRARY(libzstd_static STATIC ${Sources} ${Headers})`

jacquesg:
Indeed, fixed.

PR #502, opened 2016-12-26
Move -std=c++11 cxxflag to PZSTD_CXXFLAGS

Chocobo1:
Fixes the problem that the compiler doesn't enable c++11 mode by default and the package build system has its own CXXFLAGS.


terrelln:
LGTM

PR #504, opened 2016-12-30
[zstdmt] Add thread pool

terrelln:
Adds a thread pool in `common/pool.{c, h}`.
Adds simple tests for the thread pool.

If pthreads isn't available, which is determined by `#ifdef ZSTD_PTHREAD`, it falls back to executing the functions inline.

I'm merging it into zstdmt .  I will need it, or something similar, eventually for the dictionary builder.  I can either merge the dictionary builder into the zstdmt branch, or I can remove the multithreaded functionality until the zstdmt branch is merged into dev.  Which do you think would be better?

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -1,0 +1,190 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include "pool.h"
+#include <stddef.h>  /* size_t */
+#include <stdlib.h>  /* malloc, calloc, free */
+
+#ifdef ZSTD_PTHREAD
+
+#include <pthread.h>
+
+/* A job is a function and an opaque argument */
+typedef struct POOL_job_s {
+  POOL_function function;
+  void *opaque;
+} POOL_job;
+
+struct POOL_ctx_s {
+    /* Keep track of the threads */
+    pthread_t *threads;
+    size_t numThreads;
+
+    /* The queue is a circular buffer */
+    POOL_job *queue;
+    size_t queueHead;
+    size_t queueTail;
+    size_t queueSize;
+    /* The mutex protects the queue */
+    pthread_mutex_t queueMutex;
+    /* Condition variable for pushers to wait on when the queue is full */
+    pthread_cond_t queuePushCond;
+    /* Condition variables for poppers to wait on when the queue is empty */
+    pthread_cond_t queuePopCond;
+    /* Indicates if the queue is shutting down */
+    int shutdown;
+};
+
+/* POOL_thread() :
+   Work thread for the thread pool.
+   Waits for jobs and executes them.
+   @returns : NULL on failure else non-null.
+*/
+static void *POOL_thread(void *opaque) {
+    POOL_ctx *ctx = (POOL_ctx *)opaque;
+    if (!ctx) { return NULL; }
+    for (;;) {
+        /* Lock the mutex and wait for a non-empty queue or until shutdown */
+        if (pthread_mutex_lock(&ctx->queueMutex)) { return NULL; }
+        while (ctx->queueHead == ctx->queueTail && !ctx->shutdown) {
+            if (pthread_cond_wait(&ctx->queuePopCond, &ctx->queueMutex)) { return NULL; }
+        }
+        /* empty => shutting down: so stop */
+        if (ctx->queueHead == ctx->queueTail) {
+            if (pthread_mutex_unlock(&ctx->queueMutex)) { return NULL; }
+            return opaque;
+        }
+        {
+            /* Pop a job off the queue */
+            POOL_job job = ctx->queue[ctx->queueHead];
+            ctx->queueHead = (ctx->queueHead + 1) % ctx->queueSize;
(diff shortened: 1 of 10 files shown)


terrelln:
The [travis](https://travis-ci.org/terrelln/zstd/builds/187640575) and [appveyor](https://ci.appveyor.com/project/terrelln/zstd/build/1.0.209) tests when based on top of dev.

Cyan4973:
If we want zstdmt to converge to use the generic `threadPool`, then better merge it into the `zstdmt branch.
I'll work on the adaptation

lib/common/pool.h
```diff
@@ -0,0 +1,45 @@
...
+void POOL_free(POOL_ctx *ctx);
+
+/*! POOL_function :
+    The function type that can be added to a thread pool.
+*/
+typedef void (*POOL_function)(void *);
+/*! POOL_add_function :
+    The function type for a generic thread pool add function.
+*/
+typedef void (*POOL_add_function)(void *, POOL_function, void *);
```
Cyan4973: What's the purpose of this prototype definition ?

lib/common/pool.h
```diff
@@ -0,0 +1,45 @@
...
+
+/*! POOL_add() :
+    Add the job `function(opaque)` to the thread pool.
+    Possibly blocks until there is room in the queue.
+*/
+void POOL_add(void *ctx, POOL_function function, void *opaque);
```
Cyan4973: You've selected the `void*` version for `ctx`, instead of `POOL_ctx*` which would offer stronger type guarantee. I suspect you have a scenario in mind to make this preference. Which one ?

Cyan4973: A comment : 
when invoking `POOL_add()`, one has to make sure that `opaque` argument will outlive the function call.

It may feel obvious, but it's also easy to make this mistake : 
define `void* opaque` as a kind of `structJob* jobDescription` which is defined on the fly into a `POOL_add()` loop, hence becomes invalid on reaching the end of the loop.

This lifetime issue is an important reason why previous version of zstdmt would use a non-generic job post / consume mechanism : it allowed passing the job description "by value", eliminating lifetime issue. 
Unfortunately, by definition, a generic threadPool cannot do that, since it cannot guess the content of `opaque`.

Ensuring that `void* opaque` outlive its thread requires a more complex strategy which typically collects back `opaque` on job completion.
Nothing earth shattering, but nonetheless something to worry about.
Might be worth a code comment.

lib/common/pool.h
terrelln: It is the type of `POOL_add()`, and a function taking a generic thread pool would take a function pointer of this type.  However, since this isn't a public header, it can't be used in public interfaces, so maybe it isn't that useful.

lib/common/pool.h
terrelln: I selected the `void*` version because it can be used as an argument to a function with a signature like:

    /* Function takes a thread pool to run jobs on */
    void fn(void *threadPool, POOL_add_function addJob) {
      addJob(threadPool, &someFn, opaque);
    }

I originally had `POOL_add()` and `POOL_add_generic()`.  I combined them because the linker on the `make aarch64test` [segfaulted](https://travis-ci.org/terrelln/zstd/jobs/187623333) on `POOL_add_generic()`.  I picked the more generic one so I could get the tests passing.  If they can both be added without segfaulting the linker, that would be great.

terrelln: Good point, I will add a comment.  I nearly made the same mistake in the dictionary builder, and I wrote it.

lib/common/pool.c
```diff
@@ -0,0 +1,190 @@
...
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#include "pool.h"
+#include <stddef.h>  /* size_t */
+#include <stdlib.h>  /* malloc, calloc, free */
+
+#ifdef ZSTD_PTHREAD
+
+#include <pthread.h>
```
Cyan4973: In latest `zstdmt` branch update,
`threading.h` from Tino's project has been added.
Its objective is to provide a wrapper `pthread` interface for non-POSIX systems. Typically, it needs to support `_WIN32` and non-pthreads systems.

The number of `pthreads` symbols it supports is currently pretty low.
See [`threading.h`](https://github.com/facebook/zstd/blob/zstdmt/lib/common/threading.h) file for more details.

lib/common/pool.c
```diff
@@ -0,0 +1,190 @@
...
+
+    /* The queue is a circular buffer */
+    POOL_job *queue;
+    size_t queueHead;
+    size_t queueTail;
+    size_t queueSize;
+    /* The mutex protects the queue */
+    pthread_mutex_t queueMutex;
+    /* Condition variable for pushers to wait on when the queue is full */
+    pthread_cond_t queuePushCond;
```
Cyan4973: For example, `pthread_cond_t` is not supported currently by `threading.h` (outside of posix systems).
Which means, there is a need for one of these solutions : 
- extend `threading.h` to support `pthread_cont_t`
- rewrite `pool.c` to not need `pthread_cont_t`
- write a specific `_WIN32` version of `pool.c`

lib/common/pool.h
Cyan4973: For which use case is it beneficial to create and use a generic `POOL_add` version,
thus requiring `void*` type for `threadPool` argument ?

terrelln:
I will add `pthread_cond_t` to `threading.h`.

Cyan4973:
> mutexes must be unlocked by the same thread that locks them

Great catch !
I updated zstdmt to use `pthread_cond_t` to send signals between threads.

PR #505, opened 2017-01-01
pthread_cond_t

terrelln:
Add support for condition variables to `threading.h`.  This initial implementation only works for Windows Vista and above.  Supporting earlier Windows versions will require a custom implementation of condition variables.  However, adding support for older Windows only requires changing the threading files, so it can be done transparently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
       os: linux
       sudo: false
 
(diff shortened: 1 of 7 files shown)


terrelln:
Tests based off of the dev branch for [appveyor](https://ci.appveyor.com/project/terrelln/zstd/build/1.0.224) and [travis](https://travis-ci.org/terrelln/zstd/builds/187977085).

PR #507, opened 2017-01-03
Add new dictionary builder

terrelln:
Adds dictionary builder based off of the heuristic in [Effective Construction of Relative Lempel-Ziv Dictionaries](http://www2016.net/proceedings/proceedings/p807.pdf).  It divides the input equally into epochs.  Then it looks at each segment of length `k` in the epoch and picks the one with the best score.  The score of a segment is the sum of the frequencies of the segments of length `d` it covers.  `k` is called the segment size, and `d` is called the dmer size.  It continues this until the dictionary is full.

1.  `COVER_trainFromBuffer()` builds a dictionary with a set segment size and dmer size.  This functionality can be used on the CLI as `zstd --train --cover=k=#,d=# ...`.
2.  `COVER_optimizeTrainFromBuffer()` tries a bunch of different segment and dmer sizes.  Optional either `k` or `d` can be fixed.  `steps` controls how many different segment sizes are tried.  This functionality can be used on the CLI as `zstd --train --optimize-cover`, with optional parameters passed like `--optimize-cover=steps=#`.  This function is ready to be parallelized once the `zstdmt` branch has been merged into `dev`.  All of the threading specific code is protected by `#ifdef ZSTD_PTHREAD` and uses `threading.h`.
3.  Minor change to `dibio.c`: Shuffle all of the file names before loading them to avoid bias.  In my tests this caused up to around 10% compression ratio swings.
4.  Add tests to `playTests.sh` and `fuzzer.c`.

Tests so far have shown relative compression ratio improvements over the current dictionary builder ranging from 7-20%.

### DIFF
--- a/build/VS2005/fuzzer/fuzzer.vcproj
+++ b/build/VS2005/fuzzer/fuzzer.vcproj
@@ -332,6 +332,10 @@
 				>
 			</File>
 			<File
+				RelativePath="..\..\..\lib\dictBuilder\cover.c"
+				>
+			</File>
+			<File
 				RelativePath="..\..\..\lib\dictBuilder\divsufsort.c"
 				>
 			</File>
--- a/build/VS2005/zstd/zstd.vcproj
+++ b/build/VS2005/zstd/zstd.vcproj
@@ -344,6 +344,10 @@
 				>
 			</File>
 			<File
+				RelativePath="..\..\..\lib\dictBuilder\cover.c"
+				>
+			</File>
+			<File
 				RelativePath="..\..\..\lib\dictBuilder\divsufsort.c"
 				>
 			</File>
--- a/build/VS2005/zstdlib/zstdlib.vcproj
+++ b/build/VS2005/zstdlib/zstdlib.vcproj
@@ -328,6 +328,10 @@
 			UniqueIdentifier="{4FC737F1-C7A5-4376-A066-2A32D752A2FF}"
 			>
 			<File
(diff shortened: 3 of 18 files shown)


Cyan4973:
Great PR Nick !
A few questions : 
- I had in mind you wanted `POOL` to be generic enough to be used by `COVER` to improve speed.
But `COVER` is merged into `dev`, while `POOL` is merged into `zstdmt`.
Did the objective changed ?
- Speaking of which : I suspect there will be some conflicts when `zstdmt` is merged into `dev` after `COVER`, typically in files like `zstdcli.c`.
- What is the _speed_ performance of `COVER` ? I remember Giuseppe saying it's faster than the current dictionary builder, is that correct ?

terrelln:
* In COVER_optimizeTrainFromBuffer(), each call to COVER_tryParameters() with the same value of d can be called in parallel.  I tore out the POOL code, but a 3 line diff can add it back in once zstdmt is merged.  This could also be merged into zstdmt instead, depending on the release schedule.
* The conflicts should be easily resolved, I don't think I modified the same lines.
* I'll measure the speed more precisely when I have access to a computer, but it runs at around the same speed with fixed parameters.  When optimizing the parameters it runs much slower (by default 5-10 mins on a single thread, controlled by the "steps" parameter) since it just builds N dictionaries and picks the parameters that gives the best compression ratio, but can be parallelized to make it run much faster when more cores are available.

PR #509, opened 2017-01-10
Handle cover dictionary builder maximum input size for 32-bit mode

terrelln:
* The cover algorithm requires a single buffer of 4x the input size, so in 32-bit mode, the maximum input size is 1 GB.
* Document the memory requirements for the cover dictionary builder.
* Make error messages on OOM or too large input clearer.


Cyan4973:
All good !

PR #511, opened 2017-01-15
Set dictionary ID in ZSTD_initCStream_usingCDict()

indygreg:
When porting python-zstandard to use ZSTD_initCStream_usingCDict()
so compression dictionaries could be reused, an automated test
failed due to compressed content changing.

I tracked this down to ZSTD_initCStream_usingCDict() not
setting the dictID field of the ZSTD_CCtx attached to the
ZSTD_CStream instance.

I'm not 100% convinced this is the correct or full solution,
as I'm still seeing one automated test failing with this change.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2970,6 +2970,7 @@ size_t ZSTD_initCStream_usingCDict(ZSTD_CStream* zcs, const ZSTD_CDict* cdict)
     ZSTD_parameters const params = ZSTD_getParamsFromCDict(cdict);
     size_t const initError =  ZSTD_initCStream_advanced(zcs, NULL, 0, params, 0);
     zcs->cdict = cdict;
+    zcs->cctx->dictID = params.fParams.noDictIDFlag ? 0 : cdict->refContext->dictID;
     return initError;
 }
 


Cyan4973:
Thanks for the patch @indygreg .
Looks like a genuine bug you discovered and fixed.

It will also be great to add a relevant test case, as this bug should have already been exposed. I'll take care of this part.

Our rules require CLA to be completed. This is a one-time process, you won't need to do it again for future patches.

Alternatively, should you prefer it, I can also manually push the same fix from my own clone.

indygreg:
The Facebook CLA tool is confused: when I go to complete the form, it tells me I've already registered my email address. I sent an email to Facebook, but my guess is I won't hear back until after the USA holiday today.

Sorry for no test case - I wasn't sure where/how to code that.

Go ahead and push the same fix from your clone if you want.

Cyan4973:
Sorry for your bad experience with the CLA process.
I'll try to get some help for you.

In parallel, I'll start developing a test case for this issue.

JoelMarcey:
@indygreg Ok. The CLA issue should be resolved now. Sorry about that.

Cyan4973:
I've added the test checking `dictID` in combination with `ZSTD_initCStream_usingCDict()`.
It was non trivial, so I can not expect external contributor to sort this out.
As expected, the test fails.

I'll push it to github, and then merge your patch.
Hopefully, it will show that your patch solves the issue.

Cyan4973:
works great !

PR #515, opened 2017-01-18
Don't create dict in streaming apis if dictSize == 0

iburinoc:


Cyan4973:
Looks good to me !

lib/zstd.h
```diff
@@ -510,7 +510,7 @@ ZSTDLIB_API unsigned ZSTD_getDictID_fromFrame(const void* src, size_t srcSize);
 /*=====   Advanced Streaming compression functions  =====*/
 ZSTDLIB_API ZSTD_CStream* ZSTD_createCStream_advanced(ZSTD_customMem customMem);
 ZSTDLIB_API size_t ZSTD_initCStream_srcSize(ZSTD_CStream* zcs, int compressionLevel, unsigned long long pledgedSrcSize);   /**< pledgedSrcSize must be correct */
-ZSTDLIB_API size_t ZSTD_initCStream_usingDict(ZSTD_CStream* zcs, const void* dict, size_t dictSize, int compressionLevel);
+ZSTDLIB_API size_t ZSTD_initCStream_usingDict(ZSTD_CStream* zcs, const void* dict, size_t dictSize, int compressionLevel); /* note: a dict will not be used if dict == NULL or dictSize < 8 */
```
terrelln: Instead of `/*` use `/**<`. That way it will be picked up in the [manual](https://github.com/facebook/zstd/blob/dev/doc/zstd_manual.html).

PR #516, opened 2017-01-19
fix for zlibWrapper

inikep:
added the totalInBytes flag - we need it as strm->total_in can be reset by user

### DIFF
--- a/zlibWrapper/.gitignore
+++ b/zlibWrapper/.gitignore
@@ -20,3 +20,6 @@ zwrapbench
 *.bat
 *.zip
 *.txt
+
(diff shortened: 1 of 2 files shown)


zlibWrapper/.gitignore
```diff
@@ -20,3 +20,6 @@ zwrapbench
 *.bat
 *.zip
 *.txt
+
+# Directories
+minizip/
```
Cyan4973: Is this `/` on purpose ?

inikep: yes

PR #518, opened 2017-01-21
zstd with multi-threading

Cyan4973:
This adds the Multi-Threading variant of Zstandard
to both the API (`zstdmt_compress.h`) and cli (`zstmt`).

The cli can be built with `make zstdmt`.
Use the new `-T#` command to select number of threads (`1` by default).

MT is still considered an experimental mode.
For the time being, single-thread remains default mode for `zstd`.


programs/bench.c
```diff
@@ -212,34 +231,36 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
...
+        UTIL_time_t coolTime, coolTick;
         U64 const maxTime = (g_nbSeconds * TIMELOOP_MICROSEC) + 1;
         U64 totalCTime=0, totalDTime=0;
         U32 cCompleted=g_decodeOnly, dCompleted=0;
 #       define NB_MARKS 4
         const char* const marks[NB_MARKS] = { " |", " /", " =",  "\\" };
         U32 markNb = 0;
 
-        UTIL_getTime(&coolTime);
+		UTIL_initTimer(&coolTick);
```
terrelln: Can you fix the indentation in this file?

Cyan4973: done !

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+
+    /* fill input buffer */
+    {   size_t const toLoad = MIN(input->size - input->pos, zcs->inBuffSize - zcs->inBuff.filled);
+        memcpy((char*)zcs->inBuff.buffer.start + zcs->inBuff.filled, input->src, toLoad);
+        input->pos += toLoad;
+        zcs->inBuff.filled += toLoad;
+    }
+
+    if ( (zcs->inBuff.filled == zcs->inBuffSize)  /* filled enough : let's compress */
+        && (zcs->nextJobID <= zcs->doneJobID + zcs->jobIDMask) ) {   /* avoid overwriting job round buffer */
```
terrelln: Is the second part of the condition possible? IIUC it shouldn't be, and the condition should be deleted or put in an assert statement.

If it is, it could cause busy waiting in the users loop, so it needs to be avoided.

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+        memcpy(zcs->inBuff.buffer.start, (const char*)zcs->jobs[jobID].srcStart + zcs->targetSectionSize, zcs->inBuff.filled);
+
+        DEBUGLOG(3, "posting job %u   (%u bytes)  (note : doneJob = %u=>%u)", zcs->nextJobID, (U32)zcs->jobs[jobID].srcSize, zcs->doneJobID, zcs->doneJobID & zcs->jobIDMask);
+        POOL_add(zcs->factory, ZSTDMT_compressChunk, &zcs->jobs[jobID]);   /* This call is blocking if all workers are busy */
+        zcs->nextJobID++;
+    }
+
+    /* check if there is any data available to flush */
+    {   unsigned const jobID = zcs->doneJobID & zcs->jobIDMask;
+        ZSTDMT_jobDescription job = zcs->jobs[jobID];
```
terrelln: This read needs to be protected by the mutex, since the job might not be done yet.

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+}
+
+
+static size_t ZSTDMT_flushStream_internal(ZSTDMT_CCtx* zcs, ZSTD_outBuffer* output, unsigned endFrame)
+{
+    size_t const srcSize = zcs->inBuff.filled;
+
+    DEBUGLOG(1, "flushing : %u bytes to compress", (U32)srcSize);
+    if ( ((srcSize > 0) || (endFrame && !zcs->frameEnded))
+       && (zcs->nextJobID <= zcs->doneJobID + zcs->jobIDMask) ) {
```
terrelln: again, this shouldn't be possible, and if it is it needs to be fixed.

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+                ZSTDMT_waitForAllJobsCompleted(zcs);
+                ZSTDMT_releaseAllJobResources(zcs);
+                return job.cSize;
+            }
+            memcpy((char*)output->dst + output->pos, (const char*)job.dstBuff.start + job.dstFlushed, toWrite);
+            output->pos += toWrite;
+            job.dstFlushed += toWrite;
+            if (job.dstFlushed == job.cSize) {   /* output buffer fully flushed => next one */
+                ZSTDMT_releaseBuffer(zcs->buffPool, job.dstBuff); zcs->jobs[wJobID].dstBuff = g_nullBuffer;
+                zcs->jobs[wJobID].jobCompleted = 0;
```
terrelln: In `ZSTDMT_compressStream()` the equivalent if statement has the code:

```c++
zcs->jobs[jobID].dstBuff = g_nullBuffer;
zcs->jobs[jobID].jobCompleted = 0;
```
Should the same be here?

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+    unsigned nbBuffers;
+    buffer_t bTable[1];   /* variable size */
+} ZSTDMT_bufferPool;
+
+static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(unsigned nbThreads)
+{
+    unsigned const maxNbBuffers = 2*nbThreads + 2;
+    ZSTDMT_bufferPool* const bufPool = (ZSTDMT_bufferPool*)calloc(1, sizeof(ZSTDMT_bufferPool) + maxNbBuffers * sizeof(buffer_t));
+    if (bufPool==NULL) return NULL;
+    bufPool->totalBuffers = maxNbBuffers;
```
Cyan4973: It's implied by `calloc()`. But well, let's nonetheless do it for clarity.

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+ZSTDMT_CCtx *ZSTDMT_createCCtx(unsigned nbThreads)
+{
+    ZSTDMT_CCtx* cctx;
+    U32 const minNbJobs = nbThreads + 2;
+    U32 const nbJobsLog2 = ZSTD_highbit32(minNbJobs) + 1;
+    U32 const nbJobs = 1 << nbJobsLog2;
+    DEBUGLOG(4, "nbThreads : %u  ; minNbJobs : %u ;  nbJobsLog2 : %u ;  nbJobs : %u  \n",
+            nbThreads, minNbJobs, nbJobsLog2, nbJobs);
+    if ((nbThreads < 1) | (nbThreads > ZSTDMT_NBTHREADS_MAX)) return NULL;
+    cctx = (ZSTDMT_CCtx*) calloc(1, sizeof(ZSTDMT_CCtx) + nbJobs*sizeof(ZSTDMT_jobDescription));
```
Cyan4973: Yes it does. Could be optimized by one.

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,681 @@
...
+}
+
+
+size_t ZSTDMT_compressCCtx(ZSTDMT_CCtx* mtctx,
+                           void* dst, size_t dstCapacity,
+                     const void* src, size_t srcSize,
+                           int compressionLevel)
+{
+    ZSTD_parameters params = ZSTD_getParams(compressionLevel, srcSize, 0);
+    size_t const chunkTargetSize = (size_t)1 << (params.cParams.windowLog + 2);
```
Cyan4973: It's not useful _currently_, because `ZSTDMT_compress()` does not allow custom parameters.

`ZSTDMT_SECTION_LOGSIZE_MIN` was created in case someone uses very small `windowLog` values, resulting in very small job sizes (too small to be worth the context-switching cost).

Of course, if we create `ZSTDMT_compress_advanced()`, the answer would be different.

lib/compress/zstdmt_compress.c
Cyan4973: The second part of the condition is possible.
It requires specific conditions, but a fuzzer using random length and random flushes can statistically get there from time to time.

When we are in this situation, it basically means we have a lot of completed jobs waiting to be flushed. It could be because an intermediate job is not finished, but that's highly unlikely.
The most common reason to reach this situation is that the output buffer is so small that it needs several trips to flush the content of finished jobs.

In this situation, a call to `ZSTDMT_compressStream()` does not take any more input, but can still flush data into output buffer. After a few round-trip, one job get completely flushed, freeing space to start a new one.

This is a corner case, it's not expected to happen very often. There are some good "security margins" allowing the encoder to have a few jobs in advance before getting there.

lib/compress/zstdmt_compress.c
Cyan4973: If job is not done yet, then `job.jobCompleted == 0`, since it's the last operation done for a job.
In which case, the flushing section is just skipped, which does not seem incorrect ?

lib/compress/zstdmt_compress.c
Cyan4973: same answer

lib/compress/zstdmt_compress.c
Cyan4973: It's the same.
Maybe `zcs->jobs[wJobID].dstBuff = g_nullBuffer;` should be on its own line to be more visible.

lib/compress/zstdmt_compress.c
terrelln: Isn't it possible that the job is setting `job.jobCompleted = 1` at the same time this is executing, since this doesn't acquire the mutex?

Cyan4973: Yes, it is possible. But what is the impact ?
If the read happens before, then `job.jobCompleted == 0` and the flush is skipped (will be found on next call).
If the read happens after, then `job.jobCompleted == 1`, and the flush is triggered, which is correct since it means compression is completed.

Not sure to see what could go wrong ...

lib/compress/zstdmt_compress.c
terrelln: Ah, I missed the second statement on the line.

lib/compress/zstdmt_compress.c
terrelln: The write isn't guaranteed to be atomic. Additionally, by not protecting the write the compiler may do strange things that make the generated code incorrect.

lib/compress/zstdmt_compress.c
terrelln: Does `zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask` imply that at least once job is done then?

lib/compress/zstdmt_compress.c
terrelln: Protecting the write with the mutex is actually not enough, since writes to other parts of the job data structure could happen. The main thread must check `jobCompleted == 1` with the mutex held before reading any other part of the job (and using the result, I believe it is okay to throw away the read).

One bad case is that the compiler decides to reorder the read of `zcs->jobs[jobID].cSize` before the read of `zcs->jobs[jobID].jobCompleted`, and the job sets `cSize` and `jobCompleted = 1` between the reads. In this case we have read the wrong `cSize`.

I think the following code should work:

```c++
int jobCompleted;
pthread_mutex_lock(&zcs->jobCompleted_mutex);
jobCompleted = zcs->jobs[jobID].jobCompleted;
pthread_mutex_unlock(&zcs->jobCompleted_mutex);
if (jobCompleted) {
    ZSTDMT_jobDescription job = zcs->jobs[jobID];
    /* ... */
}
```

lib/compress/zstdmt_compress.c
Cyan4973: Mostly yes.
`jodIDMask` is necessarily `> nbThreads`.
Therefore, for `zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask`, it needs at least a few jobs to be completed. 

In some (rare) cases, it _could_ be that the first job is really much longer than the other ones, and therefore is the one we are waiting for. For example, the first job could be "large", while all the following ones could be "short", due to usage of `flushStream()`.

But now, it depends on what is meant by "completed".
From a worker thread perspective, a job is finished when compression is completed.
That's fine, but from a main thread perspective, a job is only finished when the entire output has been flushed. Only then it is released, and the slot made available again for a new job.

Therefore, in some cases, it could be that multiple input jobs were completed, but the space provided to flush them was simply not large enough.
In which case, we would have the first job "finished" from a worker perspective, but still "unfinished" from a main thread perspective, because first job output is still not entirely flushed.

lib/compress/zstdmt_compress.c
Cyan4973: All right, modified it as you suggest

lib/compress/zstdmt_compress.c
terrelln: Okay, I think I understand now. This scenario should trigger it, right?

There are 4 threads.
Lets assume that flush is never called for simplicity.
The first compression worker hangs.
Some of the rest of the jobs finish and more start.
They aren't blocking on `POOL_add()` because there are enough workers, since some jobs have finished.
Enough of the rest jobs finish compression and move on to cause `zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask`.
However, if `zcs->inBuff.filled == zcs->inBuffSize` no forward progress can be made until the first job finishes.

Cyan4973: Yes, that's a scenario where it can happen.
If the first compression job is still unfinished, and `inBuff` is filled, then indeed, no further progress can be made.

Note that, in fuzzer tests, this is not the scenario I observed.
Rather, I observed the first compression job as completed, but its result was still not completely flushed. For this reason, it was impossible to re-use the `jobDescription` slot to create a new job.
In this case, there _are_ progresses when calling `ZSTD_compressStream()` : it does not accept any more input if `inbuff` is filled, but it continues flushing whatever it can into `dst` buffer.

terrelln: Ah yeah, that would be the case exposed when calling `flush()` since it waits for the earliest job to be done.

I'm worried about busy waiting in the case I described, which would only be encountered if `flush()` isn't called. I think that `ZSTDMT_compressStream()` always needs to make forwards progress, because the user has no wait of externally knowing when to wait or not.

In the case when the input buffer is full and no jobs can be added, I think we should to check to see if the first job is completed, and if it isn't we wait until it is.

lib/compress/zstdmt_compress.c
```diff
@@ -0,0 +1,686 @@
...
+        DEBUGLOG(3, "posting job %u   (%u bytes)  (note : doneJob = %u=>%u)", zcs->nextJobID, (U32)zcs->jobs[jobID].srcSize, zcs->doneJobID, zcs->doneJobID & zcs->jobIDMask);
+        POOL_add(zcs->factory, ZSTDMT_compressChunk, &zcs->jobs[jobID]);   /* This call is blocking if all workers are busy */
+        zcs->nextJobID++;
+    }
+
+    /* check if there is any data available to flush */
+    {   unsigned const jobID = zcs->doneJobID & zcs->jobIDMask;
+        unsigned jobCompleted;
+        pthread_mutex_lock(&zcs->jobCompleted_mutex);
+        jobCompleted = zcs->jobs[jobID].jobCompleted;
```
terrelln: I think replacing `jobCompleted = zcs->jobs[jobID].jobCompleted` with something like this will work. We make sure that if we can't make forward progress by passing more input, we can make forward progress by flushing output.

```c
while (zcs->jobs[jobID].jobCompleted == 0 && zcs->inBuff.filled == zcs->inBuffSize && zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask) {
    pthread_cond_wait(&zcs->jobCompleted_cond, &zcs->jobCompleted_mutex);
}
jobCompleted = zcs->jobs[jobID].jobCompleted;
```

Cyan4973: Well, to be complete, the result `zcs->inBuff.filled == zcs->inBuffSize` can be the consequence of current invocation filling the input buffer entirely,
which can be considered as a forward progress in itself.

A second point is that if, at this place in the code, `zcs->inBuff.filled == zcs->inBuffSize`, then necessarily `zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask`, because if it wasn't, `inBuff` would have been transferred into a job description, and as a consequence, the new `inBuff.filled` would be `== 0`.

Cyan4973: block on this specific scenario (no new job created and no flush possible) is implemented

Cyan4973:
> Have you run the fuzzer and the stream test with TSAN? 

I've ran some new TSAN tests today, and they behave correctly this time.
I don't witness the flow of warnings that was initially experienced.
It could be there were really some race conditions that were solved by later patches.

Anyway, the test signal seems now good enough to be considered part of our C.I suite.


> Can TSAN tests also be added to travis, either in this PR or a future one.

_possibly_, as a shorter test, like ASAN and USAN we currently run there.
Though my primary intention is to run a 24/24 fuzzer process with TSAN enabled within FB dev infrastructure.

PR #524, opened 2017-01-28
Fix segfault in zstreamtest MT

terrelln:
It was reading beyond the end of the input buffer because no errors were
detected. Once that was fixed, it wasn't making forward progress because
no errors were detected and it was waiting for input.


Cyan4973:
great catch !

PR #526, opened 2017-01-30
Add educational decoder to /contrib

iburinoc:
Add a self-contained decoder that implements the Zstandard format specification and can be used as the base for alternate implementations, or to gain a better understanding of how the format works without the tricks and optimizations of the reference implementation.

### DIFF
--- a/contrib/educational_decoder/README.md
+++ b/contrib/educational_decoder/README.md
@@ -1,15 +1,16 @@
 Educational Decoder
 ===================
 
-`zstd_decompress.c` is a self-contained implementation of a decoder according
-to the Zstandard format specification written in C99.
+`zstd_decompress.c` is a self-contained implementation in C99 of a decoder,
+according to the [Zstandard format specification].
 While it does not implement as many features as the reference decoder,
 such as the streaming API or content checksums, it is written to be easy to
 follow and understand, to help understand how the Zstandard format works.
 It's laid out to match the [format specification],
-so it can be used to understand how confusing segments could be implemented.
+so it can be used to understand how complex segments could be implemented.
 It also contains implementations of Huffman and FSE table decoding.
 
+[Zstandard format specification]: https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md
 [format specification]: https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md
 
 `harness.c` provides a simple test harness around the decoder:
--- a/contrib/educational_decoder/README.md
+++ b/contrib/educational_decoder/README.md
@@ -1,0 +1,18 @@
+Educational Decoder
+===================
+
+`zstd_decompress.c` is a self-contained implementation of a decoder according
+to the Zstandard format specification written in C99.
+While it does not implement as many features as the reference decoder,
+such as the streaming API or content checksums, it is written to be easy to
+follow and understand, to help understand how the Zstandard format works.
+It's laid out to match the [format specification],
+so it can be used to understand how confusing segments could be implemented.
+It also contains implementations of Huffman and FSE table decoding.
+
+[format specification]: https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md
+
+`harness.c` provides a simple test harness around the decoder:
+
+    harness <input-file> <output-file> [dictionary]
+
--- a/contrib/educational_decoder/harness.c
+++ b/contrib/educational_decoder/harness.c
@@ -1,0 +1,93 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "zstd_decompress.h"
+
+typedef unsigned char u8;
+
+// There's no good way to determine output size without decompressing
+// For this example assume we'll never decompress at a ratio larger than 16
+#define MAX_COMPRESSION_RATIO (16)
+
+u8 *input;
+u8 *output;
+u8 *dict;
+
+size_t read_file(const char *path, u8 **ptr) {
+  FILE *f = fopen(path, "rb");
+  if (!f) {
+    fprintf(stderr, "failed to open file %s\n", path);
+    exit(1);
+  }
+
+  fseek(f, 0L, SEEK_END);
+  size_t size = ftell(f);
+  rewind(f);
+
+  *ptr = malloc(size);
+  if (!ptr) {
+    fprintf(stderr, "failed to allocate memory to hold %s\n", path);
+    exit(1);
+  }
+
+  size_t pos = 0;
+  while (!feof(f)) {
+    size_t read = fread(&(*ptr)[pos], 1, size, f);
+    if (ferror(f)) {
+      fprintf(stderr, "error while reading file %s\n", path);
+      exit(1);
+    }
+    pos += read;
+  }
+
+  fclose(f);
+
+  return pos;
+}
+
+void write_file(const char *path, const u8 *ptr, size_t size) {
+  FILE *f = fopen(path, "wb");
+
+  size_t written = 0;
+  while (written < size) {
+    written += fwrite(&ptr[written], 1, size, f);
+    if (ferror(f)) {
+      fprintf(stderr, "error while writing file %s\n", path);
+      exit(1);
+    }
+  }
+
+  fclose(f);
+}
+
+int main(int argc, char **argv) {
+  if (argc < 3) {
+    fprintf(stderr, "usage: %s <file.zst> <out_path> [dictionary]\n", argv[0]);
+
+    return 1;
+  }
+
+  size_t input_size = read_file(argv[1], &input);
+  size_t dict_size = 0;
+  if (argc >= 4) {
+    dict_size = read_file(argv[3], &dict);
+  }
+
+  output = malloc(MAX_COMPRESSION_RATIO * input_size);
+  if (!output) {
+    fprintf(stderr, "failed to allocate memory\n");
+    return 1;
+  }
+
+  size_t decompressed =
+      ZSTD_decompress_with_dict(output, input_size * MAX_COMPRESSION_RATIO,
+                                input, input_size, dict, dict_size);
+
+  write_file(argv[2], output, decompressed);
+
+  free(input);
+  free(output);
+  free(dict);
+  input = output = dict = NULL;
+}
+
--- a/contrib/educational_decoder/harness.c
+++ b/contrib/educational_decoder/harness.c
@@ -1,3 +1,12 @@
+/*
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
 #include <stdio.h>
 #include <stdlib.h>
 
--- a/contrib/educational_decoder/harness.c
+++ b/contrib/educational_decoder/harness.c
@@ -5,8 +5,8 @@
 
 typedef unsigned char u8;
 
-// There's no good way to determine output size without decompressing
-// For this example assume we'll never decompress at a ratio larger than 16
+// If the data doesn't have decompressed size with it, fallback on assuming the
+// compression ratio is at most 16
 #define MAX_COMPRESSION_RATIO (16)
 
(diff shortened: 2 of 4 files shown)


contrib/educational_decoder/README.md
```diff
@@ -0,0 +1,18 @@
+Educational Decoder
+===================
+
+`zstd_decompress.c` is a self-contained implementation of a decoder according
+to the Zstandard format specification written in C99.
```
Cyan4973: The reference implementation is written in C,
but the format specification is just plain markdown.
Better link to the specification, as you do below.

contrib/educational_decoder/harness.c
```diff
@@ -0,0 +1,93 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "zstd_decompress.h"
+
+typedef unsigned char u8;
+
+// There's no good way to determine output size without decompressing
+// For this example assume we'll never decompress at a ratio larger than 16
+#define MAX_COMPRESSION_RATIO (16)
```
Cyan4973: I would recommend usage of `ZSTD_getDecompressedSize()` to determine output buffer size
(or any other equivalent of your liking).

You can still have a fallback using `MAX_COMPRESSION_RATIO`, since in some cases (streaming pipe), decompressed size is unknown (not stored in the header).
In which case, use a relevant warning message to inform users about what's happening.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2096 @@
+/// Zstandard educational decoder implementation
```
Cyan4973: Add some Facebook license boilerplate, similar to existing source files

contrib/educational_decoder/README.md
```diff
@@ -0,0 +1,18 @@
...
+While it does not implement as many features as the reference decoder,
+such as the streaming API or content checksums, it is written to be easy to
+follow and understand, to help understand how the Zstandard format works.
+It's laid out to match the [format specification],
+so it can be used to understand how confusing segments could be implemented.
```
Cyan4973: prefer difficult / complex rather than "confusing".
If a specification sentence is confusing, we update it to no longer be.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2096 @@
...
+size_t ZSTD_decompress(void *dst, size_t dst_len, const void *src,
+                       size_t src_len);
+/// If `dict != NULL` and `dict_len >= 8`, does the same thing as
+/// `ZSTD_decompress` but uses the provided dict
+size_t ZSTD_decompress_with_dict(void *dst, size_t dst_len, const void *src,
+                                 size_t src_len, const void *dict,
+                                 size_t dict_len);
+
+/******* UTILITY MACROS AND TYPES *********************************************/
+#define MAX_WINDOW_SIZE ((size_t)512 << 20)
```
Cyan4973: Specification _recommends_ supporting up to 8 MB window size

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2096 @@
...
+// Max block size decompressed size is 128 KB and literal blocks must be smaller
+// than that
+#define MAX_LITERALS_SIZE ((size_t)(1024 * 128))
```
Cyan4973: minor detail : try to remain consistent when expressing sizes 
512<<20 then 128<<10
or
1024*128, therefore 1024*1024*512

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2096 @@
...
+    // Since we decode in reverse order, overflowing the stream is offset going
+    // negative
+    while (1) {
+        if (dst > dst_max - 2) {
+            OUT_SIZE();
+        }
+        *dst++ = FSE_decode_symbol(dtable, &state1, src, &offset);
+        if (offset < 0) {
+            // There's still a symbol to decode in state2
+            *dst++ = FSE_decode_symbol(dtable, &state2, src, &offse
```
Cyan4973: Always check the result of `malloc()`.
You need to have a way to express "init failed".

As a simpler educational decoder, it's possible to simplify this issue by simply aborting the program
(`exit(errorNb)`).
A more "production-level" decoder will have to return some kind of "error code" that the library user can trap and use.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+    *literals = malloc(regenerated_size);
+    if (!*literals) {
+        BAD_ALLOC();
+    }
+
+    if (block_type == 2) {
+        // Decode provided Huffman table
+
+        HUF_free_dtable(&ctx->literals_dtable);
+        const size_t size = decode_huf_table(streams->src, compressed_size,
```
terrelln: I'm not sure if it makes sense to pass `streams` to `HUF` and `FSE` or not. Thats up to you, but if you pass a raw pointer, you should do something like this (modulo function names).

```c
const u8 *const src = streams_get_src_bytes(streams, compressed_size); // does not update state but does bounds checks
size_t const size = decode_huf_tables();
streams_advance_src(streams, size); // updates state
```

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+    // Decode the weights
+    *num_symbs = FSE_decompress_interleaved2(
+        &dtable, weights, HUF_MAX_SYMBS, src + read, compressed_size - read);
+
+    FSE_free_dtable(&dtable);
+
+    return compressed_size;
+}
+/******* END LITERALS DECODING ************************************************/
+
```
terrelln: End of review.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+    if (!dict || !dict->content)
+        return;
+
+    if (ctx->header.dictionary_id == 0 && dict->dictionary_id != 0) {
+        // The dictionary is unneeded, and shouldn't be used as it may interfere
+        // with the default offset history
+        return;
+    }
+
+    // If the dictionary id is 0, it doesn't matter if we provide the wrong raw
```
iburinoc: So the only case to do any early returns here is if the header has a non-zero id and the provided dictionary doesn't match it, right?

terrelln: Yeah, that is how the zstd decompressor behaves.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+    HUF_free_dtable(&context->literals_dtable);
+
+    FSE_free_dtable(&context->ll_dtable);
+    FSE_free_dtable(&context->ml_dtable);
+    FSE_free_dtable(&context->of_dtable);
+
+    memset(context, 0, sizeof(frame_context_t));
+}
+
+static void parse_frame_header(frame_header_t *const header,
```
Cyan4973: I second @terrelln statement

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+static size_t decode_sequences(frame_context_t *const ctx, const u8 *src,
+                               size_t src_len,
+                               sequence_command_t **const sequences) {
+    size_t num_sequences;
+
+    // Decode the sequence header and allocate space for the output
+    if (src_len < 1) {
+        INP_SIZE();
+    }
+    if (src[0] == 0) {
```
Cyan4973: A quick reminder of the different choices defined in the specification would be welcomed here

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+                match_length -= dict_copy;
+            }
+        } else if (offset > ctx->header.window_size) {
+            CORRUPTION();
+        }
+
+        // We must copy byte by byte because the match length might be larger
+        // than the offset
+        // ex: if the output so far was "abc", a command with offset=3 and
+        // match_length=6 would produce "abcabcabc" as the new output
```
Cyan4973: 👍

Cyan4973:
The comments added in later patched make a positive difference.
Most code blocks now have a short introduction explaining their intention, which is great.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+        }
+        const u8 symb = src[0];
+        src++;
+        src_len--;
+        FSE_init_dtable_rle(table, symb);
+
+        return 1;
+    }
+    case seq_fse: {
+        size_t read =
```
terrelln:
```c
size_t const read
```

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+        src_len -= read;
+
+        return read;
+    }
+    case seq_repeat:
+        // Don't have to do anything here as we're not changing the table
+        return 0;
+    default:
+        // Impossible, as mode is from 0-3
+        return -1;
```
terrelln: Don't return an error code and not check it. You should abort.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+    int last_block = 0;
+
+    do {
+        if (src + 3 > src_end) {
+            INP_SIZE();
+        }
+        // Parse the block header
+        last_block = src[0] & 1;
+        const int block_type = (src[0] >> 1) & 3;
+        const size_t block_len = read_bits_LE(src, 21, 3);
```
terrelln: Can you add a function for this, since this is copy/pasted from above?

Also, make an enum for `block_type`.

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+
+        src += 3;
+        switch (block_type) {
```
terrelln:
```c
switch (block_type) {
case 0:
case 2:
    // ...
    break;
case 1:
    // ...
    break;
case 3:
default:
    CORRUPTION();
```

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2168 @@
...
+    if (!dict->content) {
+        BAD_ALLOC();
+    }
+
+    dict->content_size = src_len;
+    memcpy(dict->content, src, src_len);
+}
+
+/// If parse_dictionary is given a raw content dictionary, it delegates here
+static void init_raw_content_dict(dictionary_t *const dict, const u8 *const src,
```
terrelln: Change this to `init_dictionary_content()`, that way it can be used both for raw content dictionaries and zstd dictionary content.

```c
if (/* raw content dictionary */) {
  dict->dictionary_id = 0;
  init_dictionary_content();
}
// ...
init_dictionary_content();
return;

File: contrib/educational_decoder/zstd_decompress.c [hunk 19]
```diff
@@ -0,0 +1,2168 @@
...
+    FSE_free_dtable(&dict->ll_dtable);
+    FSE_free_dtable(&dict->of_dtable);
+    FSE_free_dtable(&dict->ml_dtable);
+
+    free(dict->content);
+
+    memset(dict, 0, sizeof(dictionary_t));
+}
+/******* END DICTIONARY PARSING ***********************************************/
+
```
terrelln: End of review

Cyan4973:
Looks good to me.
Do you feel it ready for merge @iburinoc ?

contrib/educational_decoder/zstd_decompress.c
```diff
@@ -0,0 +1,2337 @@
...
+    }
+
+    out->ptr[0] = symb;
+    out->ptr++;
+    out->len--;
+}
+
+/// Returns the number of bytes left to be read in this stream.  The stream must
+/// be byte aligned.
+static inline size_t IO_istream_l
```
terrelln: Do you handle the `-1` return value in every case you call it, or ensure that `num > 0`?

iburinoc:
@Cyan4973 Yeah I think it's ready

PR #528, opened 2017-02-03
improved Makefile

inikep:
Autodetection of zlib is not executed for every target.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -65,14 +65,7 @@ else
 EXT =
 endif
 
-# zlib detection
-HAVE_ZLIB := $(shell echo "int main(){}" | $(CC) -o have_zlib -x c - -lz && echo 1 || echo 0)
-ifeq ($(HAVE_ZLIB), 1)
-ZLIBCPP = -DZSTD_GZDECOMPRESS
-ZLIBLD = -lz
-endif
-
-.PHONY: default all clean clean_decomp_o install uninstall generate_res
+.PHONY: default zstd all clean clean_decomp_o install uninstall generate_res
 
 default: zstd-release
 
@@ -80,7 +73,6 @@ all: zstd
 
 $(ZSTDDECOMP_O): CFLAGS += $(ALIGN_LOOP)
 
-zstd-internal : CPPFLAGS += -DZSTD_LEGACY_SUPPORT=$(ZSTD_LEGACY_SUPPORT)
 zstd-internal : $(ZSTDLIB_OBJ) zstdcli.o fileio.o bench.o datagen.o dibio.o
 ifeq ($(HAVE_ZLIB), 1)
 	@echo "==> building zstd with .gz decompression support "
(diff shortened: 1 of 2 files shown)


programs/Makefile
```diff
@@ -138,7 +137,9 @@ zstd-compress: $(ZSTDCOMMON_FILES) $(ZSTDCOMP_FILES) zstdcli.c fileio.c
 	$(CC) $(FLAGS) -DZSTD_NOBENCH -DZSTD_NODICT -DZSTD_NODECOMPRESS $^ -o $@$(EXT)
 
 zstdmt: CPPFLAGS += -DZSTD_MULTITHREAD
+ifeq (,$(filter Windows%,$(OS)))
```
Cyan4973: does that work with MSYS2 environment too ?

Cyan4973:
I'm not that much concerned by zlib detection run for every target,
as the `programs` directory is only in charge of `zstd` cli and derivatives.
The "extra work" is not that large, just running `make` session creates a lot of overhead.

Leaving an artefact in the directory though is a valid concern. Better remove it.

programs/Makefile
inikep: Yes, it works for gcc/MSYS2 and gcc/MinGW64. It doesn't work with `-lpthread`.

programs/Makefile
```diff
@@ -92,12 +84,19 @@ ifneq (,$(filter Windows%,$(OS)))
...
 
-zstd : CPPFLAGS += $(ZLIBCPP)
-zstd : LDFLAGS += $(ZLIBLD)
-zstd : zstd-internal
+zstd : CPPFLAGS += -DZSTD_LEGACY_SUPPORT=$(ZSTD_LEGACY_SUPPORT)
+zstd :
+	@echo "int main(){}" | $(CC) -o have_zlib -x c - -lz && echo zlib found || echo zlib not found
+	@if [ -s have_zlib ]; then \
+         $(RM) have_zlib$(EXT) fileio.o \
+         && $(MAKE) zstd-internal HAVE_ZLIB=1 CFLAGS="$(CFLAGS)" CPPFLAGS="$(CPPFLAGS) -DZSTD_GZDECOMPRESS " LDFLAGS="$(LDFLAGS) -lz" ; \
```
Cyan4973: This solution calls `make` inside `make`. While it works, I prefer to avoid it when possible.

I think you managed to control the implied limitation (second `make` cannot modify environment variables), so I believe it does work correctly.

The issue is more about how evolutive the construction is, notably with nested targets, considering here the possibility of additional cli variants.

inikep: I checked all targets and it seems to work fine. 
My solution is only a proposal. If you prefer not to call `make` from `make` I don't see any option besides a global HAVE_ZLIB variable like in your solution.

Cyan4973:
Latest version of `gzstd_default` branch redirects the result of compilation test towards `/dev/null`, eliminating test artefact.
In my view, this is good enough, as I don't feel concerned by the cpu cost of the test itself.

The patch for `zstdmt` Windows compilation is good, so I will add it.

PR #532, opened 2017-02-06
Change zlib include to be a system include

terrelln:
I'm not sure how to test this. I tried adding the flag `-nostdinc` and them adding back in the standard includes with `-isystem`, which supposedly doesn't add paths to the quotes include path, but gcc and clang both still allowed `"zlib.h"` to be included.

However, we can ensure that `#include <zlib.h>` works in the `HAVE_ZLIB` test.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -67,7 +67,7 @@ endif
 
 # zlib detection
 VOID = /dev/null
-HAVE_ZLIB := $(shell echo "int main(){}" | $(CC) -o $(VOID) -x c - -lz 2> $(VOID) && echo 1 || echo 0)
+HAVE_ZLIB := $(shell echo "\#include <zlib.h>\nint main(){}" | $(CC) -o $(VOID) -x c - -lz 2> $(VOID) && echo 1 || echo 0)
 ifeq ($(HAVE_ZLIB), 1)
 ZLIBCPP = -DZSTD_GZDECOMPRESS
 ZLIBLD = -lz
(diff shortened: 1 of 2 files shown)


programs/Makefile
```diff
@@ -67,7 +67,7 @@ endif
 
 # zlib detection
 VOID = /dev/null
-HAVE_ZLIB := $(shell echo "int main(){}" | $(CC) -o $(VOID) -x c - -lz 2> $(VOID) && echo 1 || echo 0)
+HAVE_ZLIB := $(shell echo "\#include <zlib.h>\nint main(){}" | $(CC) -o $(VOID) -x c - -lz 2> $(VOID) && echo 1 || echo 0)
```
inikep: Doesn't work with MinGW:
```
sh-4.3$ echo "int main(){}" | gcc -o have_zlib -x c - -lz && echo 1 || echo 0
1
```
```
sh-4.3$ echo "#include <zlib.h>\nint main(){}" | gcc -o have_zlib -x c - -lz && echo 1 || echo 0
<stdin>:1:18: warning: extra tokens at end of #include directive
C:/Aplikacje/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.1.0/../../../../x86_64-w64-mingw32/lib/../lib/libmingw32.a(lib64_libmingw32_a-crt0_c.o): In function `main':
C:/repo/mingw-w64-crt-git/src/mingw-w64/mingw-w64-crt/crt/crt0_c.c:18: undefined reference to `WinMain'
collect2.exe: error: ld returned 1 exit status
0
```

inikep: I fixed it with `echo -e` at https://github.com/inikep/zstd/commit/ca20edd96024e71011f83e526d831028c877e30f

PR #533, opened 2017-02-07
--priority=rt

inikep:


Cyan4973:
Could this new setting be documented ? `zstd -h`, `man zstd` ?

inikep:
Done

programs/Makefile
```diff
@@ -67,7 +67,10 @@ endif
 
 # zlib detection
 VOID = /dev/null
-HAVE_ZLIB := $(shell echo "\#include <zlib.h>\nint main(){}" | $(CC) -o $(VOID) -x c - -lz 2> $(VOID) && echo 1 || echo 0)
+HAVE_ZLIB := $(shell echo -e "\#include <zlib.h>\nint main(){}" | $(CC) -o have_zlib -x c - -lz 2> $(VOID) && echo 1 || echo 0)
+ifeq ($(HAVE_ZLIB), 1)
```
Cyan4973: minor detail : why having a separate `ifeq ($(HAVE_ZLIB), 1)` for `$(shell rm` ?
Could it be merged with below one ?

inikep: Will not work for some targets e.g. `zstd-decompress`, `zstd-frugal`.

Cyan4973: I'm not sure to understand why.
The second `ifeq` seems to have the same result as the first one,
irrespective of the target (`zstd`, `zstd-frugal`, etc.).

inikep: Will not second `if` be run only for `zstd-internal`? We have to use something like this:
```
ifeq ($(HAVE_ZLIB), 1)
       TEMP := $(shell rm have_zlib$(EXT))
	@echo "==> building zstd with .gz decompression support "
else
	@echo "==> no zlib, building zstd with .zst support only (no .gz support) "
endif
```

inikep: I checked also:
```
ifeq ($(HAVE_ZLIB), 1)
TEMP := $(shell rm have_zlib$(EXT))
	@echo "==> building zstd with .gz decompression support "
else
	@echo "==> no zlib, building zstd with .zst support only (no .gz support) "
endif
```
and the results is
`Makefile:88: *** recipe commences before first target.  Stop.`

Cyan4973: I'm pretty sure the second `ifeq ($(HAVE_ZLIB), 1)` is played like the first, so there seems to be no reason to separate them.
Only the 3rd one is triggered from within `zstd-internal`.

It's not possible to define or update variables from within the 3rd one.
It's not possible to start commands from the 1st and 2nd one
(but you can define and evaluate variable, hence `TMP := $(shell echo ...` would work)

inikep: But the issue here is that
```
ifeq ($(HAVE_ZLIB), 1)
TEMP := $(shell rm have_zlib$(EXT))
	@echo "==> building zstd with .gz decompression support "
else
	@echo "==> no zlib, building zstd with .zst support only (no .gz support) "
endif
```
breaks the `zstd-internal` target. I'm not sure what is your solution.

terrelln: `echo -e` isn't available on all platforms, including my Mac, but I'm not sure what it can be replaced with.

Cyan4973: > I'm not sure what is your solution.

```
ifeq ($(HAVE_ZLIB), 1)
TEMP := $(shell rm have_zlib$(EXT))
endif
ifeq ($(HAVE_ZLIB), 1)
ZLIBCPP = -DZSTD_GZDECOMPRESS
ZLIBLD = -lz
endif
```

> echo -e isn't available on all platforms,

Is the `-e` required because of `#include <zlib.h>` ?

terrelln: It is required because of the newline, so yeah.

inikep: @Cyan4973 Sorry, I didn't see second `ifeq ($(HAVE_ZLIB)`

Cyan4973: Do we need the `#include <zlib.h>` ?
If yes, can we generate a newline differently ?
double `echo` into a temporary file maybe ?

inikep: > echo -e isn't available on all platforms, including my Mac, but I'm not sure what it can be replaced with.

What shell are you using? With Linux and MSYS `-e` is required and can be selected as beginning of a script:
`playTests.sh:1` `#!/bin/sh -e`
It seems to work with macOS. Maybe `-e` is ignored.

inikep: > Do we need the #include <zlib.h> 

I think Nick added it to make sure if <zlib.h> is always available. It makes sense as <zlib.h> is sometimes in a separate package: http://packages.ubuntu.com/trusty/zlib1g-dev

inikep:
> echo -e isn't available on all platforms, including my Mac, but I'm not sure what it can be replaced with.

What shell are you using? The following post claims that is available with Bash shell:
http://superuser.com/questions/33914/why-doesnt-echo-support-e-escape-when-using-the-e-argument-in-macosx

We are also using it in `playTests.sh:1` where `-e` is set as default with `#!/bin/sh -e`. Works fine with macOS on Travis.

terrelln:
I use zsh, but it works in zsh. It just fails in the Makefile invocation of `$(shell echo -e)`, but I'm not sure why that is.

inikep:
I found it on http://stackoverflow.com:
`this is what happens in OS X (Mountain Lion):`
```
$ bash
bash-4.2$ echo -e foo
foo
bash-4.2$ sh
sh-3.2$ echo -e foo
-e foo
```

inikep:
Second, make uses `/bin/sh` as a shell but the functionality of `/bin/sh` is implemented differently on Mac OS X and Linux:
- On Mac OS X, the functionality of `/bin/sh` is implemented by `bash`. Also on Mac OS X, bash is compiled with `--enable-strict-posix-default`. One consequence of this flag is that the echo command does not understand the `-e` flag.
- On Linux, the functionality of `/bin/sh` is implemented by `dash` which is less strict with respect to POSIX specification. Therefore the flag `-e` is implemented in the echo command.

inikep:
I think I found a solution:
```
inikep@fizzle:~$ echo '#include <zlib.h>\nint main(){}'
#include <zlib.h>\nint main(){}
inikep@fizzle:~$ echo $'#include <zlib.h>\nint main(){}'
#include <zlib.h>
int main(){}
```

inikep:
`echo $'#include'` works only with `bash` (macOS, MSYS) but not with `dash` (Linux).

1. I decided to use `printf` as it seems to be more portable:
http://www.in-ulm.de/~mascheck/various/echo+printf/
The only worry if `printf` is available in the default installation.
For Linux `printf` belongs to GNU coreutils (https://en.wikipedia.org/wiki/List_of_GNU_packages).

2. I have also tried to use a temporary file and `cat`. Works fine but `cat` also belongs to GNU coreutils.

inikep:
The newest idea:
```
HAVE_ZLIB_H := $(shell echo '\#include <zlib.h>' | $(CC) -E -H -o $(VOID) - 2>$(VOID) && echo 1 || echo 0)
ifeq ($(HAVE_ZLIB_H), 1)
HAVE_ZLIB := $(shell echo 'int main(){}' | $(CC) -o have_zlib -x c - -lz 2> $(VOID) && echo 1 || echo 0)
endif
```
It uses a preprocessor from gcc to check for `zlib.h`. Then it checks linking with `-lz`. Works fine with `gcc` and `clang`. I'm not sure how it works with icc and Sun Studio (Solaris).

inikep:
Ready to be accepted.

PR #535, opened 2017-02-08
Remove test-longmatch from test target and only run it once

terrelln:
It was timing out in `qemu` runs.

The travis file should be ok, but it doesn't run the modified test on my branch, so the PR tests will determine if it is ok.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
 
 
     # Standard Ubuntu 12.04 LTS Server Edition 64 bit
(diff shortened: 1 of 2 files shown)


Cyan4973:
The `pool` test seems to have the same issue : 
https://travis-ci.org/facebook/zstd/jobs/199358041#L1100

inikep:
I already fixed it in my PR disabling `test-pool` for ppc64:
https://github.com/facebook/zstd/pull/533/commits/4b4f8c2d71ddd22df117b1330495acaa1100d95d

Issue #506, opened 2017-01-01
Empty source inputs have ambiguous frame headers when writing content size


indygreg:
(This issue is derived from indygreg/python-zstandard/issues/11.)

Calling `ZSTD_compress_advanced()` with a `srcSize=0` and `ZSTD_frameParams.frameContentSize=1` leads to an ambiguous frame header.

From `ZSTD_writeFrameHeader()`:

```
    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize > (pledgedSrcSize-1));
    U32   const fcsCode = params.fParams.contentSizeFlag ?
                     (pledgedSrcSize>=256) + (pledgedSrcSize>=65536+256) + (pledgedSrcSize>=0xFFFFFFFFU) :   /* 0-3 */
                      0;
```

If `pledgedSrcSize==0`, `singleSegment` will be `0` if `contentSizeFlag=1` because `windowSize` and `pledgedSrcSize` are `U32` and the `>` comparison fails because `-1` is wrapping. **This is possibly a bug.**

Anyway, in the current behavior, compressing an empty input while told to write content size results in `Frame_Content_Size_flag=0` and `Single_Segment_flag=0`. This is effectively the same as not writing the content size and means that the simple decompress API can't be used and we must go to streaming mode because the output size is unknown. **This defeats the purpose of writing the content size.**

If `Single_Segment_flag=1` (as I /think/ it should be for the empty input because the decompressed output most certainly fits in a single window size because it is length 0), there is still an ambiguous case as to how to decompress this. This is because `ZSTD_getDecompressedSize()` returns `0`, which can mean *content size unknown* or a content size of `0`. If you call `ZSTD_getFrameParams()` as the docs for `ZSTD_getDecompressedSize()` suggest, you don't get anything useful, as `frameContentSize=0`. So you still have an ambiguous scenario: you don't know if the content size is unknown or if the content size is 0. This means you have to use streaming decompression, which again undermines the purpose of writing the content size.

My main question in this issue is how is the empty input supposed to be handled by the simple decompression API? How are we supposed to disambiguate between content size not being written and a `0` content size?

Cyan4973:
This is a known weakness of current API.

Indeed, on the decoding side, there is no way currently to make a distinction between `decompressed size unknown` and `decompressed size is known and is zero` (empty frame).

[...]

PR #536, opened 2017-02-09
Simple API multiframe decoding

iburinoc:
- `ZSTD_decompress` and similar now handle inputs with multiple frames (including skippable frames), decompressing them in order
- `ZSTD_findDecompressedSize` replaces `ZSTD_getDecompressedSize`
- Handles multiple frames
  - Returns `ZSTD_CONTENTSIZE_UNKNOWN` instead of 0 if the size can't be determined
  - Returns `ZSTD_CONTENTSIZE_ERROR` if an error occurred instead of 0
    - #506 is no longer ambiguous
- `ZSTD_getFrameContentSize` gets the size of a single ZSTD frame

- For the advanced compression API, if `contentSizeFlag == 1`, then `pledgedSrcSize` is interpreted as accurate
  - Note: if `pledgedSrcSize == 0` and `contentSizeFlag == 1`, this now means the content size is exactly 0 bytes.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2362,7 +2362,7 @@ static size_t ZSTD_writeFrameHeader(void* dst, size_t dstCapacity,
     U32   const dictIDSizeCode = (dictID>0) + (dictID>=256) + (dictID>=65536);   /* 0-3 */
     U32   const checksumFlag = params.fParams.checksumFlag>0;
     U32   const windowSize = 1U << params.cParams.windowLog;
-    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize > (pledgedSrcSize-1));
+    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize > pledgedSrcSize);
     BYTE  const windowLogByte = (BYTE)((params.cParams.windowLog - ZSTD_WINDOWLOG_ABSOLUTEMIN) << 3);
     U32   const fcsCode = params.fParams.contentSizeFlag ?
                      (pledgedSrcSize>=256) + (pledgedSrcSize>=65536+256) + (pledgedSrcSize>=0xFFFFFFFFU) :   /* 0-3 */
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -344,8 +344,12 @@ size_t ZSTD_copyCCtx(ZSTD_CCtx* dstCCtx, const ZSTD_CCtx* srcCCtx, unsigned long
 {
     if (srcCCtx->stage!=ZSTDcs_init) return ERROR(stage_wrong);
 
+
     memcpy(&dstCCtx->customMem, &srcCCtx->customMem, sizeof(ZSTD_customMem));
-    ZSTD_resetCCtx_advanced(dstCCtx, srcCCtx->params, pledgedSrcSize, ZSTDcrp_noMemset);
+    {   ZSTD_parameters params = srcCCtx->params;
+        params.fParams.contentSizeFlag = (pledgedSrcSize > 0);
+        ZSTD_resetCCtx_advanced(dstCCtx, params, pledgedSrcSize, ZSTDcrp_noMemset);
+    }
 
     /* copy tables */
     {   size_t const chainSize = (srcCCtx->params.cParams.strategy == ZSTD_fast) ? 0 : (1 << srcCCtx->params.cParams.chainLog);
@@ -2362,7 +2366,7 @@ static size_t ZSTD_writeFrameHeader(void* dst, size_t dstCapacity,
     U32   const dictIDSizeCode = (dictID>0) + (dictID>=256) + (dictID>=65536);   /* 0-3 */
     U32   const checksumFlag = params.fParams.checksumFlag>0;
     U32   const windowSize = 1U << params.cParams.windowLog;
-    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize > pledgedSrcSize);
+    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize >= pledgedSrcSize);
     BYTE  const windowLogByte = (BYTE)((params.cParams.windowLog - ZSTD_WINDOWLOG_ABSOLUTEMIN) << 3);
     U32   const fcsCode = params.fParams.contentSizeFlag ?
                      (pledgedSrcSize>=256) + (pledgedSrcSize>=65536+256) + (pledgedSrcSize>=0xFFFFFFFFU) :   /* 0-3 */
@@ -2845,7 +2849,11 @@ static ZSTD_parameters ZSTD_getParamsFromCDict(const ZSTD_CDict* cdict) {
 size_t ZSTD_compressBegin_usingCDict(ZSTD_CCtx* cctx, const ZSTD_CDict* cdict, unsigned long long pledgedSrcSize)
 {
     if (cdict->dictContentSize) CHECK_F(ZSTD_copyCCtx(cctx, cdict->refContext, pledgedSrcSize))
-    else CHECK_F(ZSTD_compressBegin_advanced(cctx, NULL, 0, cdict->refContext->params, pledgedSrcSize));
+    else {
+        ZSTD_parameters params = cdict->refContext->params;
+        params.fParams.contentSizeFlag = (pledgedSrcSize > 0);
+        CHECK_F(ZSTD_compressBegin_advanced(cctx, NULL, 0, params, pledgedSrcSize));
+    }
     return 0;
 }
 
@@ -2939,7 +2947,7 @@ size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
 size_t ZSTD_CStreamInSize(void)  { return ZSTD_BLOCKSIZE_ABSOLUTEMAX; }
 size_t ZSTD_CStreamOutSize(void) { return ZSTD_compressBound(ZSTD_BLOCKSIZE_ABSOLUTEMAX) + ZSTD_blockHeaderSize + 4 /* 32-bits hash */ ; }
 
-size_t ZSTD_resetCStream(ZSTD_CStream* zcs, unsigned long long pledgedSrcSize)
+static size_t ZSTD_resetCStream_internal(ZSTD_CStream* zcs, unsigned long long pledgedSrcSize)
 {
     if (zcs->inBuffSize==0) return ERROR(stage_wrong);   /* zcs has not been init at least once => can't reset */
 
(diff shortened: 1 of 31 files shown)


lib/zstd.h
```diff
@@ -80,15 +80,53 @@ ZSTDLIB_API size_t ZSTD_compress( void* dst, size_t dstCapacity,
...
-    `compressedSize` : must be the _exact_ size of a single compressed frame.
+    `compressedSize` : must be the _exact_ size of some number of compressed and/or skippable frames.
     `dstCapacity` is an upper bound of originalSize.
     If user cannot imply a maximum upper bound, it's better to use streaming mode to decompress data.
     @return : the number of bytes decompressed into `dst` (<= `dstCapacity`),
               or an errorCode if it fails (which can be tested using ZSTD_isError()). */
 ZSTDLIB_API size_t ZSTD_decompress( void* dst, size_t dstCapacity,
                               const void* src, size_t compressedSize);
 
+#define ZSTD_CONTENTSIZE_UNKNOWN (0ULL - 1)
```
Cyan4973: It's too soon to integrate these new prototypes into `stable`.
They need first to stage through `experimental`.
There is always a chance that we may have missed something, that we'll find out through usages and feedbacks.

So, declare these new definitions behind `ZSTD_STATIC_LINKING_ONLY`

lib/zstd.h
```diff
@@ -80,15 +80,53 @@ ZSTDLIB_API size_t ZSTD_compress( void* dst, size_t dstCapacity,
...
+*    note 4 : If source is untrusted, decompressed size could be wrong or intentionally modified.
+*             Always ensure result fits within application's authorized limits.
+*             Each application can set its own limits.
+*    note 5 : ZSTD_findDecompressedSize handles multiple frames, and so it must traverse the input to
+*             read each contained frame header.  This is efficient as most of the data is skipped,
+*             however it does mean that all frame data must be present and valid. */
+ZSTDLIB_API unsigned long long ZSTD_findDecompressedSize(const void* src, size_t srcSize);
+
 /*! ZSTD_getDecompressedSize() :
+*   WARNING: This function is now obsolete.  ZSTD_findDecompressedSize should be used,
```
Cyan4973: It's still good announce the intention to obsolete this prototype in the future, explain why and provide guidance for replacement.
Just, for now, candidate replacements are still in `experimental`, so `ZSTD_getDecompressedSize()` is not obsolete _yet_.

lib/compress/zstd_compress.c
```diff
@@ -2362,7 +2362,7 @@ static size_t ZSTD_writeFrameHeader(void* dst, size_t dstCapacity,
     U32   const dictIDSizeCode = (dictID>0) + (dictID>=256) + (dictID>=65536);   /* 0-3 */
     U32   const checksumFlag = params.fParams.checksumFlag>0;
     U32   const windowSize = 1U << params.cParams.windowLog;
-    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize > (pledgedSrcSize-1));
+    U32   const singleSegment = params.fParams.contentSizeFlag && (windowSize > pledgedSrcSize);
```
Cyan4973: minor : `>=` would be fine too

lib/zstd.h
```diff
@@ -520,7 +558,7 @@ ZSTDLIB_API ZSTD_CStream* ZSTD_createCStream_advanced(ZSTD_customMem customMem);
 ZSTDLIB_API size_t ZSTD_initCStream_srcSize(ZSTD_CStream* zcs, int compressionLevel, unsigned long long pledgedSrcSize);   /**< pledgedSrcSize must be correct */
```
Cyan4973: Here, we have to choose if `pledgedSrcSize==0` really means `0`, or if it means `unknown`.
Both have pros and cons.

Up to now `pledgedSrcSize==0` meant `unknown`, so preserving this behavior would allow existing users to continue working as intended.
But it would make this API unable to generate a `zero-size` in frame header.
To do so, one would need to implement `ZSTD_initCStream_advanced()`.

Changing it to mean `0` could break a few programs, so it needs to be properly documented.
Work-around to mean `unknown` would be to use `ZSTD_initCStream()`.
Existing programs would need to create a branch for this usage.

Due to the risk of breaking existing applications, I would rather lean for the first solution.
To be debated.

lib/zstd.h
```diff
@@ -520,7 +558,7 @@ ZSTDLIB_API ZSTD_CStream* ZSTD_createCStream_advanced(ZSTD_customMem customMem);
...
 ZSTDLIB_API size_t ZSTD_initCStream_usingDict(ZSTD_CStream* zcs, const void* dict, size_t dictSize, int compressionLevel); /**< note: a dict will not be used if dict == NULL or dictSize < 8 */
 ZSTDLIB_API size_t ZSTD_initCStream_advanced(ZSTD_CStream* zcs, const void* dict, size_t dictSize,
-                                             ZSTD_parameters params, unsigned long long pledgedSrcSize);  /**< pledgedSrcSize is optional and can be zero == unknown */
+                                             ZSTD_parameters params, unsigned long long pledgedSrcSize);  /**< pledgedSrcSize is optional and can be 0 (meaning unknown). note: if the contentSizeFlag is set, pledgedSrcSize == 0 means the source size is actually 0 */
 ZSTDLIB_API size_t ZSTD_initCStream_usingCDict(ZSTD_CStream* zcs, const ZSTD_CDict* cdict);  /**< note : cdict will just be referenced, and must outlive compression session */
 ZSTDLIB_API size_t ZSTD_resetCStream(ZSTD_CStream* zcs, unsigned long long pledgedSrcSize);  /**< re-use compression parameters from previous init; skip dictionary loading stage; zcs must be init at least once before */
```
Cyan4973: `ZSTD_resetCStream()` carries the same risks as `ZSTD_initCStream_srcSize()` regarding meaning of `pledgedSrcSize==0`.

lib/zstd.h
```diff
@@ -578,7 +616,7 @@ ZSTDLIB_API size_t ZSTD_sizeof_DStream(const ZSTD_DStream* zds);
 /*=====   Buffer-less streaming compression functions  =====*/
 ZSTDLIB_API size_t ZSTD_compressBegin(ZSTD_CCtx* cctx, int compressionLevel);
 ZSTDLIB_API size_t ZSTD_compressBegin_usingDict(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, int compressionLevel);
-ZSTDLIB_API size_t ZSTD_compressBegin_advanced(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, ZSTD_parameters params, unsigned long long pledgedSrcSize);
+ZSTDLIB_API size_t ZSTD_compressBegin_advanced(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, ZSTD_parameters params, unsigned long long pledgedSrcSize); /**< pledgedSrcSize is optional and can be 0 (meaning unknown). note: if the contentSizeFlag is set, pledgedSrcSize == 0 means the source size is actually 0 */
 ZSTDLIB_API size_t ZSTD_copyCCtx(ZSTD_CCtx* cctx, const ZSTD_CCtx* preparedCCtx, unsigned long long pledgedSrcSize);
```
Cyan4973: same issue for `ZSTD_copyCCtx()` and `ZSTD_compressBegin_usingCDict()` : what does mean `pledgedSrcSize==0` ? It has to be clearly specified.

tests/Makefile
```diff
@@ -27,7 +27,7 @@ TESTARTEFACT := versionsTest namespaceTest
 
 
 CPPFLAGS+= -I$(ZSTDDIR) -I$(ZSTDDIR)/common -I$(ZSTDDIR)/compress -I$(ZSTDDIR)/dictBuilder -I$(ZSTDDIR)/deprecated -I$(PRGDIR)
-CFLAGS  ?= -O3
+CFLAGS  ?= -g
```
iburinoc: Oops, this was not intended to be committed.

lib/zstd.h
```diff
@@ -88,44 +88,12 @@ ZSTDLIB_API size_t ZSTD_compress( void* dst, size_t dstCapacity,
...
-*             Each application can set its own limits.
-*    note 5 : ZSTD_findDecompressedSize handles multiple frames, and so it must traverse the input to
-*             read each contained frame header.  This is efficient as most of the data is skipped,
-*             however it does mean that all frame data must be present and valid. */
-ZSTDLIB_API unsigned long long ZSTD_findDecompressedSize(const void* src, size_t srcSize);
-
 /*! ZSTD_getDecompressedSize() :
-*   WARNING: This function is now obsolete.  ZSTD_findDecompressedSize should be used,
-*   or if only exactly one ZSTD frame is needed, ZSTD_getFrameContentSize can be used.
+*   NOTE: This function is planned to be obsolete, in favour of ZSTD_findDecompressedSize,
```
Cyan4973: The direct replacement of `ZSTD_getDecompressedSize()` will be `ZSTD_getFrameContentSize()`,
so channel attention in this direction.

`ZSTD_findDecompressedSize()` is more powerful and has stronger requirements (needs the entire frames, not just header), it answer a (slightly) different question.

lib/zstd.h
```diff
@@ -422,6 +393,41 @@ typedef void* (*ZSTD_allocFunction) (void* opaque, size_t size);
 typedef void  (*ZSTD_freeFunction) (void* opaque, void* address);
 typedef struct { ZSTD_allocFunction customAlloc; ZSTD_freeFunction customFree; void* opaque; } ZSTD_customMem;
 
+/***************************************
+*  Decompressed size functions
+***************************************/
+/*! ZSTD_getFrameContentSize() :
+*   `src` should point to the start of a ZSTD encoded frame
```
Cyan4973: give some direction on what should be `srcSize` for better success (`ZSTD_frameHeaderSize_max`)

tests/zstreamtest.c
```diff
@@ -438,6 +438,34 @@ static int basicUnitTests(U32 seed, double compressibility, ZSTD_customMem custo
       if (!ZSTD_isError(r)) goto _output_error;  /* must fail : frame requires > 100 bytes */
       DISPLAYLEVEL(3, "OK (%s)\n", ZSTD_getErrorName(r)); }
 
+    /* Unknown srcSize */
```
Cyan4973: 👍

lib/decompress/zstd_decompress.c
```diff
@@ -306,6 +306,98 @@ size_t ZSTD_getFrameParams(ZSTD_frameParams* fparamsPtr, const void* src, size_t
     return 0;
 }
 
+static size_t ZSTD_frameSrcSize(const void* src, size_t srcSize);
```
terrelln: @cyan What do you think about exposing this function in the experimental API?

Cyan4973: Good point ! Sounds like a good idea !

Cyan4973:
There's still an error on Travis : 
https://travis-ci.org/facebook/zstd/jobs/200471309#L1050

I may not be related to your patch, we just have to be sure.
Investigate `zstreamtest` with seed 3874 and multi-threading in order to identify the issue.
It might be `aarch64` specific (that would require `qemu-aarch64-static` and `aarch64-linux-gnu-gcc` to check).

PR #537, opened 2017-02-09
Fix small bugs

terrelln:
* Clang doesn't allow C++14 attributes to go after `__attribute__(())`, so reorder them.
* A compiler was complaining about redefinition of `_POSIX_C_SOURCE`, @inikep is this change correct?

### DIFF
--- a/programs/platform.h
+++ b/programs/platform.h
@@ -77,10 +77,9 @@ extern "C" {
 #    define PLATFORM_POSIX_VERSION 200112L
 #  else
 #    if defined(__linux__) || defined(__linux)
-#      ifdef _POSIX_C_SOURCE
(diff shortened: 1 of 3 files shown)


programs/platform.h
```diff
@@ -77,6 +77,9 @@ extern "C" {
 #    define PLATFORM_POSIX_VERSION 200112L
 #  else
 #    if defined(__linux__) || defined(__linux)
+#      ifdef _POSIX_C_SOURCE
+#        undef _POSIX_C_SOURCE
+#      endif
```
inikep: This should be better:
```
#      ifndef _POSIX_C_SOURCE
#        define _POSIX_C_SOURCE 200112L  /* use feature test macro */
#      endif
```

PR #540, opened 2017-02-09
zstdcli: Fix reporting incorrect sizes of large flies on MinGW

ds77:
On MinGW, use _stat64() and struct _stat64 instead of stat() and struct stat_t. This fixes reporting incorrect sizes for large files.

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -141,7 +141,7 @@ UTIL_STATIC void UTIL_waitForNextTick(UTIL_time_t ticksPerSecond)
 /*-****************************************
 *  File functions
 ******************************************/
-#if defined(_MSC_VER) || defined(__MINGW32__) && defined(__MSVCRT__) && _WIN32_WINNT >= _WIN32_WINNT_WIN2K
+#if defined(_MSC_VER)
 	#define chmod _chmod
 	typedef struct __stat64 stat_t;
 #else
@@ -172,7 +172,7 @@ UTIL_STATIC int UTIL_setFileStat(const char *filename, stat_t *statbuf)
 UTIL_STATIC int UTIL_getFileStat(const char* infilename, stat_t *statbuf)
 {
     int r;
-#if defined(_MSC_VER) || defined(__MINGW32__) && defined(__MSVCRT__) && _WIN32_WINNT >= _WIN32_WINNT_WIN2K
+#if defined(_MSC_VER)
     r = _stat64(infilename, statbuf);
     if (r || !(statbuf->st_mode & S_IFREG)) return 0;   /* No good... */
 #else
@@ -186,9 +186,13 @@ UTIL_STATIC int UTIL_getFileStat(const char* infilename, stat_t *statbuf)
 UTIL_STATIC U64 UTIL_getFileSize(const char* infilename)
 {
     int r;
-#if defined(_MSC_VER) || defined(__MINGW32__) && defined(__MSVCRT__) && _WIN32_WINNT >= _WIN32_WINNT_WIN2K
-    struct __stat64 statbuf;
+#if defined(_MSC_VER)
+    struct _stat64 statbuf;
     r = _stat64(infilename, &statbuf);
+    if (r || !(statbuf.st_mode & S_IFREG)) return 0;   /* No good... */
+#elif defined(__MINGW32__) && defined (__MSVCRT__)
+    struct _stati64 statbuf;
+    r = _stati64(infilename, &statbuf);
     if (r || !(statbuf.st_mode & S_IFREG)) return 0;   /* No good... */
 #else
     struct stat statbuf;
@@ -212,7 +216,7 @@ UTIL_STATIC U64 UTIL_getTotalFileSize(const char** fileNamesTable, unsigned nbFi
 UTIL_STATIC int UTIL_doesFileExists(const char* infilename)
 {
     int r;
-#if defined(_MSC_VER) || defined(__MINGW32__) && defined(__MSVCRT__) && _WIN32_WINNT >= _WIN32_WINNT_WIN2K
+#if defined(_MSC_VER)
     struct __stat64 statbuf;
     r = _stat64(infilename, &statbuf);
     if (r || !(statbuf.st_mode & S_IFREG)) return 0;   /* No good... */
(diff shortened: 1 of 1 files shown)


Cyan4973:
cc @inikep 
I guess the patch works,
I'm just surprised that the existing solution would "not work" with minGW target, I had the feeling this was already tried and tested

ds77:
Well, bug is in zstd tool, which seems to get less care than library, and affects only MinGW build handling 2GB+ files (stdio is fine).

Existing solution (both official release binary and my build) works fine up to 2147483647 bytes (st_size is int32).

Cyan4973:
Ideally, it also means we need a new test case to control this issue.

We have automatic minGW tests setup on Appveyor : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.512/job/2awk9qexd3x9ws1o
but currently, they only care about compilation correctness,
and do not attempt to compress very large files > 2 GB.

So that could be another point to change.
And ideally, we would need the test to fail without the patch.

Note that your proposed fix currently fails to compile on Appveyor : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/job/qpem3xuuo63g4u98

ds77:
Hmm, I was not expecting that. Tested on x86_64-w64 and i686-w64 GCC 6.3.0 (msys2).
Appveyor version is probably missing _stat64 definition. Not sure if this is due to older MinGE version (4.9.3), Linux host or something else...

inikep:
So far I can confirm that this is a regression introduced in v1.1.3.
```
d:\GitHub\Cyan4973>zstd112_64.exe -1f Ubuntu1504
Ubuntu1504           : 35.86%   (4404019200 => 1579069985 bytes, Ubuntu1504.zst)

d:\GitHub\Cyan4973>zstd113_64.exe -1f Ubuntu1504
Read : 4196 / 104 MB ==> 35.85%Error 26 : Compression error during frame end : Src size incorrect
```

ds77:
I think I found the problem with my patch breaking build. MinGW on Appveryor seems to target Windows NT, and _stat64 was added to MSVCRT in Win2k. Relevant MinGW code: https://sourceforge.net/p/mingw/mingw-org-wsl/ci/5.0-active/tree/mingwrt/include/sys/stat.h#l190

inikep:
Moreover the issue was not detected with AppVeyor while 64-bit MinGW tests were successfully performed with 6GB stdio/stdin:
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.512/job/rc5kq1b41ur486u5#L1221

ds77:
For stdin getFileSize returns 0, so bug will not show up. For file I/O, wrong getFileSize result gets passed as pledgedSrcSize and ZSTD_endStream fails.

Cyan4973:
The fix proposed by @inikep looks good (simple and well located)
but it seems we also have a test problem to look at.

ds77:
Yes, it should work well on mingw-w64 (for both i686 and x86_64) but I think it will compile, but won't fix the bug on MinGW (not w64), because _FILE_OFFSET_BITS has no effect there.
My last commit is ugly, but should work on both mingw and mingw-w64, and affected functions reuires cleanup of copy-and-paste anywany.

Cyan4973:
As far as I can tell, binaries generated by MinGW32 seem unaffected by this problem.
For reference, there's a pre-compiled version of zstd for win32 available in "Release" tab.
It works fine with large files.
```
$ ./zstd -v loong
*** zstd command line interface 32-bits v1.1.3, by Yann Collet ***
loong                :  9.80%   (6442450944 => 631337587 bytes, loong.zst)
```

inikep:
The original version worked fine for 32-bit MinGW because of:
```
+#if !defined(__64BIT__)
  #  if !defined(_FILE_OFFSET_BITS)   
  #    define _FILE_OFFSET_BITS 64                      /* turn off_t into a 64-bit type for ftello, fseeko */
  #  endif
```

ds77:
Is is fine on 32-bit releases of mingw-w64 project (mingw-w64.org). Won't work od classic mingw (www.mingw.org) because _FILE_OFFSET_BITS is ignored in its headers. I tested inikep's fix on current mingw (_not_ mingw-w64!). Compiles, but does not fix sile size issue.

ds77:
AppVeyor build fails because older versions of mingwrt check __MSVCRT_VERSION__ instead of _WIN32_WINNT. With __MSVCRT_VERSION__ = 0x600 _stat64() is not available.

_stati64() works with both mingw-w64 and non-w64 mingw, regardless of mingwrt version - tested with 3.21.x (used on AppVeror) and 3.22.x (current).

Latest commit should finally work everywhere, and it's also simpler.

Cyan4973:
So it seems we different _flavors_ of MinGW to care about ... nice ...

inikep:
There is only mingw (32-bit) and mingw-w64 (32-bit and 64-bit) :)

mingw seems to be almost dead, but I confirmed that there was an issue with mingw and large files.
The idea of @ds77 with `struct __stat64` is good. I tested a simpler version with `_stat64()` (64-bit time and file size) instead of `_stati64()` (32-time and 64-bit file size) and it works with all version of mingw:
```
#if defined(_MSC_VER) || defined(__MINGW32__)
      r = _stat64(infilename, statbuf);
      if (r || !(statbuf->st_mode & S_IFREG)) return 0;   /* No good... */
#else
```

ds77:
@inikep Approximately this was also my first idea, but it caused AppVeyor build to fail, since version of mingw library (mingwrt) used there wont't enable _stat64 unless `__MSVCRT_VERSION__` >= 0x601 (it's 0x600 by default).
New version (mingwrt 3.22.x) checks for `__MSVCRT_VERSION__ >= __MSVCR61_DLL || _WIN32_WINNT >= _WIN32_WINNT_WIN2K`, which I believe is true on default settings.

Which version you used for your test?

inikep:
You're right, I have mingwrt 3.22 and there is a problem with AppVeyor. This PR seems good to be merged.

Is there any reason to prefer `__stat64` over `_stat64`? Now these changes concern only `#if defined(_MSC_VER)`.

ds77:
There was in previous version, due to non-w64 mingw. I have left the change for cosmetic reasons - this struct seems to be "officially" called __stat64 - see MS docs: https://msdn.microsoft.com/en-us/library/14h5k7ff.aspx Both versions shoud work.

Maybe such change woud be better left for some cosmetic cleanup of util.h some time in future (or never). Should I push another commit reverting this?

inikep:
No, it's fine. I think `struct __stat64` is better distinguished from `_stat64()` but it seems that mingw-w64 prefers `struct _stat64`.

PR #541, opened 2017-02-10
fixed "mingw32" AppVeyor target

inikep:
improved AppVeyor tests


Cyan4973:
Does this fix also contain #540 ?
Or are they separate / complementary ?

inikep:
The fixes are now separate and both ready to be merged.

PR #542, opened 2017-02-12
zstdcli: fix writing 2GB+ sparse files under Windows

ds77:
There is another bug related to large files on Windows (MinGW, likely also VC++).

### How to reproduce this problem
1. Generate zero-filled file larger than 2048 MB
`dd if=/dev/zero of=testfile bs=1M count=2049`
(see [other solutions](http://stackoverflow.com/questions/982659/quickly-create-large-file-on-a-windows-system) if you don't have dd)
2. Create compressed test file
`zstd -1 testfile -o testfile.zst`
3. Attempt to decompress or test it
`zstd --test testfile.zst`
sample result:
`Decoded : 2005 MB...     Error 71 : 1 GB skip error (sparse file support)`

### Cause
`fseek()` fails if file position indicator is over 2GB (even for small relative offset). This leads to skip errors in `FIO_fwriteSparse()`/`FIO_fwriteSparseEnd()` during decompression of large sparse files.
`_FILE_OFFSET_BITS=64` does not help there, even for MinGW versions supporting this option.

### Fix
Replace `fseek()` in `FIO_fwriteSparse()` and `FIO_fwriteSparseEnd()` with macro expanding to 64-bit fseek version provided by current platform (includes  fallback workaround using Win32 API).

### Note
While I suspect that problem directly relates to MSVCRT (thus affects also Visual C++ biulds) and have roughly verified all conditionally enabled solutions under MinGW (`_fseeki64()` is available in MinGW-w64), **this still needs proper confirmation under Visual C++**.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -109,9 +109,9 @@ static clock_t g_time = 0;
             method = FILE_BEGIN;
 
         if (SetFilePointerEx((HANDLE) _get_osfhandle(_fileno(file)), off, NULL, method))
-                return 0;
+            return 0;
         else
-                return -1;
+            return -1;
     }
 #else
 #   define LONG_SEEK fseek
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -91,6 +91,32 @@ static clock_t g_time = 0;
 
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 
+#if defined(_MSC_VER) && _MSC_VER >= 1400
+#   define LONG_SEEK _fseeki64
+#elif defined(__MINGW32__) && !defined(__STRICT_ANSI__) && !defined(__NO_MINGW_LFS) && defined(__MSVCRT__)
+#   define LONG_SEEK fseeko64
+#elif defined(_WIN32) && !defined(__DJGPP__)
+#   include <windows.h>
+    static int LONG_SEEK(FILE* file, __int64 offset, int origin) {
+        LARGE_INTEGER off;
+        DWORD method;
+        off.QuadPart = offset;
+        if (origin == SEEK_END)
+            method = FILE_END;
+        else if (origin == SEEK_CUR)
+            method = FILE_CURRENT;
+        else
+            method = FILE_BEGIN;
+
+        if (SetFilePointerEx((HANDLE) _get_osfhandle(_fileno(file)), off, NULL, method))
+                return 0;
+        else
+                return -1;
+    }
+#else
+#   define LONG_SEEK fseek
+#endif
+
 
 /*-*************************************
 *  Local Parameters - Not thread safe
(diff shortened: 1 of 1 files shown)


ds77:
Looks like I accidentally merged with upstream master instead of dev when preparing this branch.
I don't want to make mess in history of this repo, but. I can't rebase properly this without forced push closing this PR.
I'll sort this out and open new PR.

Cyan4973:
Fortunately, Visual compliance is exactly what our automated CI tests check,
and your patch seems to pass it.
cc @inikep for review.

ds77:
I was fairly sure that this patch won't break anything. My concern is whether any change is really needed for Visual C++ - i have not dug deeply enough to check if new MSVCRTxx versions used by VC++ do not contain improved fseek() with different behavior (ie. work fine for small relative offsets even in large files).

inikep:
I confirm the issue for MinGW.

This is the same error we have seen with LZ4 for MacOS and FreeBSD that `fseek` doesn't work with large files and should be replaced with `fseeko`.

programs/fileio.c
```diff
@@ -91,6 +91,32 @@ static clock_t g_time = 0;
 
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 
+#if defined(_MSC_VER) && _MSC_VER >= 1400
+#   define LONG_SEEK _fseeki64
+#elif defined(__MINGW32__) && !defined(__STRICT_ANSI__) && !defined(__NO_MINGW_LFS) && defined(__MSVCRT__)
+#   define LONG_SEEK fseeko64
+#elif defined(_WIN32) && !defined(__DJGPP__)
```
inikep: What compilers are you targeting here? Visual with _MSC_VER >= 1400? MinGW with __STRICT_ANSI__?

ds77: Yes, this is fallback solution created with this cases in mind. Now after looking into [MSDN docs](https://msdn.microsoft.com/en-us/library/s3f49ktz(v=vs.80).aspx) I see that VC++ 2003 will fail on other code anyway (due to lack of unsigned long long), but MinGW in standards mode is an important configuration.

inikep:
I tested it with MinGW and MinGW-w64 and it works fine. Looks good to be merged.

programs/fileio.c
Cyan4973: Pay attention to maintenance cost. It is important too, even if only maintainers pay it with their time. Any line of code added must be worth it.

Dropping a non-negligible amount of lines to be maintained for the benefit of a rare (dying) configuration which doesn't work anyway due to other reasons, this feels like a bad investment.

Now, it seems your patch also benefits MinGW in `__STRICT_ANSI__` mode, which is a more palatable target. 
For this reason, I'll merge it.

ds77: Good point about mainteance.
I was writing this with MinGW in mind and added VC++ version check after realizing that `_fseeki64` was introduced in VS 2005 ("*I am not breaking the build*" type of thing).

PR #544, opened 2017-02-14
Portable way to avoid empty unit warning in threading.c

ds77:
When ZSTD_MULTITHREAD is not defined, `lib/common/threading.c ` triggers empty translation unit warning, which is currently suppressed only under MSVC++.

It may be worth to consider building lib in strict standards mode, and this is currently the only warning generated by `-pedantic` under GCC and Clang. It could be suppressed for GCC/Clang another `#pragma`, but proposed commit avoids this problem in a portable way by including one of standard C headers.

### DIFF
--- a/lib/common/threading.c
+++ b/lib/common/threading.c
@@ -15,10 +15,11 @@
  * This file will hold wrapper for systems, which do not support pthreads
  */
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
excellent point !

PR #546, opened 2017-02-14
fix memory leak

zefanxu:
there was a small issue of memory leak for output file name in the example, simple free() should fix it


Cyan4973:
good catch !

PR #547, opened 2017-02-14
Avoid fseek()'s 2GiB barrier with MacOS and *BSD

inikep:


inikep:
MacOS and *BSD needs `fseeko` to work with 2GB+ files.
This is the same issue that we recently fixed for MinGW with https://github.com/facebook/zstd/pull/542.
The patch is based on the method used in LZ4. I don't have MacOS to tests if it's working.

inikep:
All is fine except `Segmentation fault (core dumped)` with `pool` on arm.

Cyan4973:
cc @terrelln 
We are used to time out for `pool` test on ARM, but segfault is something new.
There might be something to look for after all : 
https://travis-ci.org/facebook/zstd/jobs/201452950#L1064

PR #550, opened 2017-02-16
execSequence copies up to 2*WILDCOPY_OVERLENGTH extra

iburinoc:
#549 didn't extend the buffers enough for all cases.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -2260,7 +2260,7 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
 
             /* Adapt buffer sizes to frame header instructions */
             {   size_t const blockSize = MIN(zds->fParams.windowSize, ZSTD_BLOCKSIZE_ABSOLUTEMAX);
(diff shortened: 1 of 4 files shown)


terrelln:
Why wasn't `WILDCOPY_OVERLENGTH` enough, and why is `2 * WILDCOPY_OVERLENGTH` enough?

Cyan4973:
Indeed, I guess it comes from the "match copy" part of the decoder
(in which case, the overcopy is at most `2*WILDCOPY_OVERLENGTH - MINMATCH`, 
but that's pedantic, `2*WILDCOPY_OVERLENGTH` will do just fine).

Did you update the test case ?

iburinoc:
`ZSTD_execSequence` from lines [1049-1075](https://github.com/facebook/zstd/blob/3ebec8383bed7d5f744488eaebf4a86d4cdf02c1/lib/decompress/zstd_decompress.c#L1049) can copy up to `2*WILDCOPY_OVERLENGTH` regardless of the actual match length, so one `WILDCOPY_OVERLENGTH` isn't enough.  This should be enough as I haven't found anywhere overwriting by more, and the program that found these bugs (WIP) didn't find any more errors after this fix.

PR #551, opened 2017-02-17
Added a Meson-based build system

dimkr:


Cyan4973:
Interesting, I wasn't aware of `meson` build system until now.

We can certainly accept a PR for a new build system, but we also have to consider the associated maintenance.
We have no plan to maintain ourselves the meson build script.
This should be made clear.

In general, codes we do not guarantee to maintain end up into `contrib` directory.
This way, it's clearer for users that this code is just made available with no guarantee, and is not maintained by upstream.

If you plan on becoming the maintainer of this meson script, I would invite you to add a `README.md` file describing your project and appointing yourself as maintainer for future questions to come.

Cyan4973:
looks good to me !

contrib/meson/meson.build
```diff
@@ -0,0 +1,79 @@
...
+                       link_with: libzstd)
+test('fullbench', fullbench)
+
+fuzzer = executable('fuzzer',
+                    datagen_c, join_paths(tests_dir, 'fuzzer.c'),
+                    include_directories: test_includes,
+                    link_with: libzstd)
+test('fuzzer', fuzzer)
+
+if target_machine.system() != 'windows'
```
TingPing: This should be `host_system` btw: https://github.com/mesonbuild/meson/wiki/Cross-compilation

contrib/meson/meson.build
```diff
@@ -0,0 +1,79 @@
+project('zstd', 'c', license: 'BSD')
+
+libm = meson.get_compiler('c').find_library('m', required: true)
```
TingPing: required should be false, plenty of platforms don't have a separate libm.

PR #556, opened 2017-02-21
original size unknown

sjnam:
In the directory zstd/examples,

After the command, "./streaming_compression README.md"

the following command always fails. ( malloc: can not allocate... )
./simple_decompression README.md.zst

because
ZSTD_findDecompressedSize(cBuff, cSize);
always return "ZSTD_CONTENTSIZE_UNKNOWN" not "0"

By the way, 
is that OK? the file compressed by streaming can not be simple decompressed 
because I can not get decomprssed size?

### DIFF
--- a/examples/dictionary_decompression.c
+++ b/examples/dictionary_decompression.c
@@ -78,10 +78,14 @@ static void decompress(const char* fname, const ZSTD_DDict* ddict)
     size_t cSize;
     void* const cBuff = loadFile_orDie(fname, &cSize);
     unsigned long long const rSize = ZSTD_findDecompressedSize(cBuff, cSize);
-    if (rSize==ZSTD_CONTENTSIZE_UNKNOWN) {
+    if (rSize==ZSTD_CONTENTSIZE_ERROR) {
+        fprintf(stderr, "%s : it was not compressed by zstd.\n", fname);
+        exit(5);
+    } else if (rSize==ZSTD_CONTENTSIZE_UNKNOWN) {
         fprintf(stderr, "%s : original size unknown \n", fname);
         exit(6);
     }
+
     void* const rBuff = malloc_orDie((size_t)rSize);
 
     ZSTD_DCtx* const dctx = ZSTD_createDCtx();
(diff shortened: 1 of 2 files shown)


Cyan4973:
That's a very good catch @sjnam .
These examples were recently changed to use `ZSTD_findDecompressedSize()` instead of previous `ZSTD_getDecompressedSize()`, but without adjusting the test case for it.
Your patch fix the issue.

Nonetheless, the test should also be extended to `ZSTD_CONTENTSIZE_ERROR`, as this is also a potential outcome, notably when trying to decompress an invalid file (for example, because it was not compressed by `zstd`).

All these use cases were previously coalesced to `0`, but now, they are identified.

> the file compressed by streaming can not be simple decompressed because I can not get decomprssed size?

Right.
The simple decompression needs an upper bound of the decompressed size.
If this information is not in the header, it can only "guess it".
In some applications, it's possible, because there are some "implicit" application limits.
But for a completely generic decoder such as `simple_decompression`, there is no such assumption. Therefore, no content size information ==> no idea how much memory will be needed ==> end of decoding.

Cyan4973:
Thanks for adding the test. It makes the example more complete !

I believe there is still one corner case left (but I can also take care of it if you get bored patching these examples).

What would happen now if `rSize ==0` ?
This is now a legitimate value : it means it's an empty frame.
It can be tested by compressing a `0` size file, but as we'll see, the resulting behaviour is compiler dependent.

Problem is, `malloc()` _may_ return `NULL` on `0` input ( [depending on compiler](http://www.cplusplus.com/reference/cstdlib/malloc/)), which will be interpreted as `error` by `malloc_orDie()`.

But as stated, this is not an error : it's just an empty frame.

Proposing a quick fix to avoid this issue : don't allocate 0 bytes.
Typically, within `malloc_orDie()` : `malloc(size + !size)`.

sjnam:
@Cyan4973 
Yes, I would appreciate it if you could take care of this patch.
Thank you.

Cyan4973:
Sure,
I'll use the opportunity to add relevant test cases too.
Thanks for the patch !

PR #557, opened 2017-02-21
travis.yml: remove tests that overlap with Circle CI

inikep:


inikep:
There was an issue with asan32 because of Ubuntu 12.04 PPA. I switched asan32 to Ubuntu 14.04 and I also had to upgrade clang so I selected clang-4.0.

PR #560, opened 2017-02-22
Change name to to findFrameCompressedSize and add skippable support

iburinoc:


lib/decompress/zstd_decompress.c
```diff
@@ -1437,17 +1437,20 @@ size_t ZSTD_generateNxBytes(void* dst, size_t dstCapacity, BYTE byte, size_t len
     return length;
 }
 
-/** ZSTD_getFrameCompressedSize() :
+/** ZSTD_findFrameCompressedSize() :
  *  compatible with legacy mode
  *  `src` must point to the start of a ZSTD or ZSTD legacy frame
```
terrelln: Update the comment

lib/decompress/zstd_decompress.c
```diff
@@ -1437,17 +1437,20 @@ size_t ZSTD_generateNxBytes(void* dst, size_t dstCapacity, BYTE byte, size_t len
...
 {
 #if defined(ZSTD_LEGACY_SUPPORT) && (ZSTD_LEGACY_SUPPORT==1)
-    if (ZSTD_isLegacy(src, srcSize)) return ZSTD_getFrameCompressedSizeLegacy(src, srcSize);
+    if (ZSTD_isLegacy(src, srcSize)) return ZSTD_findFrameCompressedSizeLegacy(src, srcSize);
 #endif
-    {
+    if (srcSize >= ZSTD_skippableHeaderSize &&
+            (MEM_readLE32(src) & 0xFFFFFFFF0U) == ZSTD_MAGIC_SKIPPABLE_START) {
+        return ZSTD_skippableHeaderSize + MEM_readLE32((const BYTE*)src + 4);
+    } else {
```
terrelln: Instead, I think this is simpler. But it would require moving `ZSTD_getFrameParams()` before `ZSTD_frameHeaderSize()`.

```c
if (fParams.windowSize == 0) {
  return ZSTD_skippableHeaderSize + fParams.frameContentSize;
}
```

iburinoc: `fParams.windowSize == 0 && fParams.frameContentSize == 0` is an ambiguous case that can happen in either an empty skippable frame or an empty Zstd frame with `singleSegment` set, so that wouldn't work.

terrelln: Good point. Would you mind updating the documentation in `zstd.h:702` to note the ambiguity?

PR #561, opened 2017-02-22
Fix bug in FSE distribution normalization

iburinoc:
A bug in FSE_normalizeM2 could cause a divide by 0 in some corner cases.  Tests added in Cyan4973/FiniteStateEntropy#79.


Cyan4973:
Good catch !

PR #563, opened 2017-02-23
Host zstd_manual.html on gh-pages

iburinoc:
Adds the current (`master`) version of the manual to the zstd gh-pages so that it can be linked to and easily viewed without having to download and open a local copy.


Cyan4973:
This file will not be automatically updated when new versions are published.
Could it be a link or an iframe to `master/doc/zstd_manual.html` instead ?
(FYI, raw file content can be downloaded from https://raw.githubusercontent.com/facebook/zstd/dev/doc/zstd_manual.html)

iburinoc:
Yeah, the problem with the raw.githubusercontent.com is the content type is raw instead of html, but I'll look at trying to make that work, maybe that's overridable with an iframe?

iburinoc:
New version loads content from master branch using AJAX and replace insert it into the page.  Not sure if this approach would work with a page that had to load any other resources (i.e. js, css), but the manual is pretty simple.

See [https://iburinoc.github.io/zstd/zstd_manual.html](https://iburinoc.github.io/zstd/zstd_manual.html) for an example.

Cyan4973:
That's great !

PR #564, opened 2017-02-23
Move educational_decoder to doc/ and add doc README

iburinoc:
- Also make some minor bugfixes to educational decoder


Cyan4973:
some conflict needs to be solved

PR #571, opened 2017-02-27
gzip tests

inikep:


tests/gzip/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+
+.PHONY: all
+all: test-gzip-env test-helin-segv test-hufts test-keep test-list test-memcpy-abuse test-mixed test-null-suffix-clobber test-stdin test-trailing-nul test-unpack-invalid test-zdiff test-zgrep-context test-zgrep-f test-zgrep-signal test-znew-k test-z-suffix 
+	@echo Testing completed
+
+.PHONY: zstd
+zstd:
+	$(MAKE) -C $(PRGDIR) zstd
+	#alias gzip='$(PRGDIR)/zstd --format=gzip'
```
Cyan4973: if both `alias` and `ln` are commented, it means the `gzip` version invoked is the regular OS one ?

tests/gzip/gzip-env.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+# Test the obsolescent GZIP environment variable.
```
Cyan4973: I don't remember exactly, but it seems this variable resulted in security troubles for `gzip` (the variable could be abused to make `gzip` do unwanted silent actions).
The consequence of which was to either disable or limit the scope of this variable.
Could be worth investigating, to know which level of support is actually desirable.

tests/gzip/Makefile
inikep: 1. `alias` doesn't work because `make` runs a new `sh`
2. `ln` works fine but a path depends on OS and I think that `ln` should be set manually (and one can remove it manually)

Cyan4973: as far as I understand, this is just for test,
so what about : 
`ln -sf $(PRGDIR)/zstd gzip` ?
It will just create a local link to just built zstd, so nothing disruptive.

Then, I guess it requires a way to force tests to use this local link instead of `/usr/local/bin/gzip`.
Either a variable, or a way to force `./` directory precedence over `/usr/local/bin`

inikep: Ubuntu 16.04:
a) has `/bin/gzip` (1.7)
b) `make install` (from gzip 1.8) installs additional `/usr/local/bin/gzip`
```
>echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
To force local path firs we need something like `PATH=.:$PATH`

Cyan4973: > To force local path first we need something like PATH=.:$PATH

Yes, I think that's the correct route to run `gzip` compatibility tests with `zstd`

inikep:
I forced scripts to use "./gzip" for gzip tests. You may accept this PR without waiting for CIs because gzip tests are not checked :)

PR #575, opened 2017-02-28
Cmake improvement

Majlen:
Added support for few more targets in cmake build system.

The pull request also includes some small fixes (better handling of library version symlinks, stopped using -Wstrict-prototypes for C++).

I do have an idea for further improvements (especially better handling of compile flags, detecting C/C++ standards, etc.), so please let me know if it is something you would be interested in.


Cyan4973:
Thanks @Majlen ,
these are good improvements to the `cmake` script.

FYI, we "just" maintain the `cmake` script : we make sure it stays functional, but we do not "improve" it. Feel free to evolve it further. As long as it does not increase maintenance, it's a good step forward.

As an additional point :
you will want to have a look at `cmaketest` : 
https://github.com/facebook/zstd/blob/dev/Makefile#L201

This is the script we use to test `cmake` automatically, keeping it functional.
Depending on your modifications, you may want to edit or expand it, to make sure it tests new paths.

PR #577, opened 2017-02-28
XZ/LZMA compression and decompression

inikep:


Cyan4973:
Great addition to the cli @inikep !
We'll make sure to test it for next `zstd` release

PR #588, opened 2017-03-06
support -Werror=format-security

pixelb:
Fedora now enables this option by default, resulting
in the following build failure:

Logging.h: In instantiation of
'void pzstd::Logger::operator()(int, const char*, Args ...)
Pzstd.cpp:413:48:   required from here
Logging.h:46:17: error: format not a string literal and no format arguments
[-Werror=format-security]
     std::fprintf(out_, fmt, args...);
     ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~


Cyan4973:
cc @terrelln 

What's the behavior of this flag with `clang++` ?

terrelln:
Clang supports -Wformat-security.

PR #590, opened 2017-03-06
Set up "short" tests on CircleCI

iburinoc:


Cyan4973:
Quick comment : 

I was looking at the test results on CircleCI.

https://circleci.com/gh/iburinoc/zstd/63?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

While reading them, I was surprise to notice tests are being played in the `dependency` category.
Dependency starts by a bunch of `apt-get` stuff, which feels right, but then continue with tests themselves.
By contrast, the `test` category is empty.

So maybe it's normal, or is a side-effect of the method selected to  to separate tests into multiple VM.
Or maybe it could be modified to be organised as intended.

iburinoc:
This was originally done by @inikep, I believe the reason was because if an individual command fails in the `dependencies` section, it stops the build immediately, but if done in the `test` section it continues running.  I could move it to the `test` section if we're ok with that behaviour.

Cyan4973:
Good point @iburinoc, I forgot about this behavior.
Is there a benefit in having tests stopping immediately after a failure ?

I think this approach would be more beneficial if tests were long,
because it frees up resources.

But since tests on CircleCI are supposed to be fast,
I believe it's better to let them run,
so that we have a more complete error map at the end.

iburinoc:
This should work, although the yaml syntax to get `parallel: true` is a bit of a pain.

inikep:
I did it in `dependencies` section because it has default `parallel: true` (look at comments).  yaml syntax for `test` is not very readable.

Cyan4973:
Thanks for details @inikep

PR #593, opened 2017-03-08
Reduce point at which we reduce offsets to protect against UB

iburinoc:
If we wait until `(2U<<30)` we risk doing some pointer subtractions where the difference is greater than or equal to `1<<31`, which is undefined behaviour on 32-bit architectures.  Found by running `playTests.sh --test-large-data` with ubsan.


Cyan4973:
great catch !

PR #594, opened 2017-03-08
Small fixes

terrelln:
* Take `sizeof(*zcs)` instead of `sizeof(zcs)`.
* Avoid defining `inline`, and use the `FORCE_INLINE` definition from `zstd_internal.h` instead.


Cyan4973:
good welcomed little fixes

PR #596, opened 2017-03-09
programs/Makefile: remove zstd-internal target

ldv-alt:
zstd-internal was intended to be a helper target, but it doesn't help
at all, what it does in practice is a useless rebuild of zstd every time
"make zstd" is invoked.

Fixes: 030ac243a0f3 ("Changed Makefile to generate zstd with .gz support by default")


Cyan4973:
Good point !
No need to rebuild `zstd` if it's already built !

I also wasn't aware of this multi-target notation, which is nice for readability

PR #597, opened 2017-03-10
Add medium tests on TravisCI

iburinoc:
Also turn off appveyor tests on feature branches.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,44 @@
+Testing
+=======
+
+Zstandard CI testing is split up into three sections:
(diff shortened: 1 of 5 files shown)


Cyan4973:
Looks good to me !
We will discuss the part were appveyor tests are simply disabled for feature branches,
we may need some minimal tests there if possible,
but there is no hurry, so we can merge this patch now

TESTING.md
```diff
@@ -0,0 +1,44 @@
+Testing
```
Cyan4973: 👍

PR #599, opened 2017-03-10
Update .travis.yml for long tests (master branch only)

iburinoc:
Introduce the long tests onto the master branch.  If merged before 1.1.4 release, these tests will fail on master until the release (as they test behaviour that's been fixed since 1.1.3), but the crons should pass.  The config is set up so that when this branch is built as a cron, it actually builds `dev` branch, using the long test config (to get things such as OS X testing, etc.), so after merging a cron should be set up on TravisCI on master branch.

This will cause merge conflicts for release merges (`dev -> master`).  These should be resolved in favour of the version in master, and never merged back into `dev` branch.
Alternatively, the `.gitattributes` change should make this easy, if
```
git config --local merge.ours.driver true
```
is run within a `zstd` clone, then merging `dev` into `master` on that clone will always take `master`s version of `.travis.yml`.

Future updates to this file will have to be done by pull requests to `master`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,3 +19,6 @@
 # Windows
 *.bat text eol=crlf
 *.cmd text eol=crlf
+
+# .travis.yml merging
+.travis.yml merge=ours
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,175 +1,52 @@
+# Long tests: run on commits to master branch/cron builds
+
 language: c
+sudo: required
+dist: trusty
 matrix:
-  fast_finish: true
   include:
-    # OS X Mavericks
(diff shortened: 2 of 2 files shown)


Cyan4973:
I will wait a bit to merge this PR.
Typically just before merging `dev` into `master`

iburinoc:
Alright, sounds good.  Also, when setting up the cron on TravisCI, it seems like the daily crons run every 24hrs, starting from the time the cron was added, so it might be beneficial to add it at night-time so that it doesn't run during the day.

Cyan4973:
> it seems like the daily crons run every 24hrs, starting from the time the cron was added, 

arf ...
a pity we don't have more control on the launch time.
Ok, I'll try to add it at a "late" hour

PR #600, opened 2017-03-13
Add ability to control level of legacy support, and drop support for versions v01-v03

iburinoc:


lib/legacy/zstd_legacy.h
```diff
@@ -28,6 +28,31 @@ extern "C" {
 #include "zstd_v06.h"
 #include "zstd_v07.h"
 
+#ifndef ZSTD_LEGACY_SUPPORT
```
Cyan4973: Another use case to properly support is if `ZSTD_LEGACY_SUPPORT` is defined with value `0`.
This is up to now supported, and means "no legacy support".

But in proposed code, it seems it would mean the otherwise, supporting _all_ legacy versions.

iburinoc: Right, good point.  In practice this wouldn't affect much, as all the locations that include `zstd_legacy.h` check if `ZSTD_LEGACY_SUPPORT == 0` first, but I'll add a check here just to be sure.

Cyan4973:
great patch !

PR #601, opened 2017-03-14
Add xz and lzma support.

terrelln:
Finish feature started by @inikep.

* Add xz and lzma compression and decompression support to target `xzstd`.
* Fix bug in gzip decompression that silently accepted truncated files.
* Add gzip frame composition tests.
* Add xz/lzma compatibility tests.
* Add xz/lzma frame composition tests.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -44,6 +44,9 @@
 #    define z_const
 #  endif
 #endif
+#if defined(ZSTD_LZMACOMPRESS) || defined(ZSTD_LZMADECOMPRESS)
+#  include <lzma.h>
+#endif
 
 
 /*-*************************************
@@ -71,7 +74,6 @@
 #define MAX_DICT_SIZE (8 MB)   /* protection against large input (attack scenario) */
 
 #define FNSPACE 30
-#define GZ_EXTENSION ".gz"
 
 
 /*-*************************************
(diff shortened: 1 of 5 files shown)


Cyan4973:
Excellent !
The fix for gzip frames with associated test set is very welcomed !

programs/fileio.c
```diff
@@ -866,11 +940,12 @@ static unsigned long long FIO_decompressGzFrame(dRess_t* ress, FILE* srcFile, co
...
         if (strm.avail_in == 0) {
             ress->srcBufferLoaded = fread(ress->srcBuffer, 1, ress->srcBufferSize, srcFile);
-            if (ress->srcBufferLoaded == 0) break;
+            if (ress->srcBufferLoaded == 0) flush = Z_FINISH;
             strm.next_in = (z_const unsigned char*)ress->srcBuffer;
             strm.avail_in = (uInt)ress->srcBufferLoaded;
         }
-        ret = inflate(&strm, Z_NO_FLUSH);
+        ret = inflate(&strm, flush);
+        if (ret == Z_BUF_ERROR) EXM_THROW(39, "zstd: %s: premature end", srcFileName);
```
inikep: Now zstd can decompress `*.*` (.gz/.xz/.lzma/.zst) and unrecognized files will be skipped.
But here EXM_THROW finishes a program with exit(). 
I think that a better policy will be to use `return 0` which is later translated to a skipped file and subsequent files can be decompressed.

Cyan4973: I agree,
there are several places in `fileio.c` where an error detection leads to an abrupt end (`EXM_THROW`). We would be better off counting that as a simple error and move on instead.
This will have to wait though, as we have not enough time to integrate that improvement into next release.

inikep: EXM_THROW usually makes sense for compression but `return 0` should be better for decompression.

PR #602, opened 2017-03-14
Add functions missing from manual, and fix parameter alignment

iburinoc:


Cyan4973:
`tsan` test flagged an issue,
but it's unlikely related to this PR.

Note that I'm currently unable to reproduce the same issue ...

PR #603, opened 2017-03-15
Prevent fuzz testers from combining large dicts with high clevel

iburinoc:
Fixes https://travis-ci.org/facebook/zstd/jobs/211257353


Cyan4973:
Good ! this should help speed, especially `qemu` tests

PR #604, opened 2017-03-16
Convert manpage to Markdown, convertible with ronn

iburinoc:


Cyan4973:
Thanks !
It will make the maintenance of the man page *much* easier from now on.

PR #605, opened 2017-03-16
Fix zstd not erroring on compressing to terminal without forceStdout

iburinoc:
Not sure I understand why the condition was the way it was before, but I don't think this should break any case that worked properly before.

Fixes this case writing compressed data to term:
```
echo bob | zstd
```

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -635,7 +635,7 @@ int main(int argCount, const char* argv[])
 
     /* Check if input/output defined as console; trigger an error in this case */
     if (!strcmp(filenameTable[0], stdinmark) && IS_CONSOLE(stdin) ) CLEAN_RETURN(badusage(programName));
-    if (outFileName && !strcmp(outFileName, stdoutmark) && IS_CONSOLE(stdout) && strcmp(filenameTable[0], stdinmark) && !(forceStdout && (operation==zom_decompress)))
+    if (outFileName && !strcmp(outFileName, stdoutmark) && IS_CONSOLE(stdout) && !strcmp(filenameTable[0], stdinmark) && !forceStdout && operation!=zom_decompress)
         CLEAN_RETURN(badusage(programName));
 
     /* user-selected output filename, only possible with a single file */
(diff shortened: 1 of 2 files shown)


Cyan4973:
We would need to add a test case to ensure such issue is detected in the future

Cyan4973:
Almost there

Apparently, something doesn't work on Windows. Maybe the `IS_CONSOLE()` detection routine.
To be investigated...

iburinoc:
Can't reproduce locally, it works fine in my Windows VM.  I suspect the problem is that appveyor doesn't pretend to be a terminal to the scripts so IS_CONSOLE returns 0, I guess the best solution here is to just not run that test on windows.

Cyan4973:
Indeed, maybe `stdin` and `stdout` are not detected as console within Appveyor.
Could it be similar to : 
https://github.com/lz4/lz4/issues/308

In which case some similar solution may apply : 
https://github.com/lz4/lz4/issues/308#issuecomment-275030972

iburinoc:
Hmm, doesn't seem to work, tried [https://github.com/iburinoc/zstd/commit/69ead66018812a6cd996d40db0be732a3366b6a2](https://github.com/iburinoc/zstd/commit/69ead66018812a6cd996d40db0be732a3366b6a2).  Passes locally, fails on appveyor.  Seems like that solution only works when `/dev/tty`/`CON` actually points to a real TTY.  It worked for LZ4 because the case that was tested was `make -C tests test | cat`, so `/dev/tty` still existed.  This doesn't seem to work in a truly tty-less environment, such as by detaching it from the terminal with setsid (`./playTests.sh: line 78: /dev/tty: No such device or address`).

Cyan4973:
OK, so I guess the fix is to disable the test for Windows ?

iburinoc:
I guess, yeah.  It still feels fragile to have that test there on other platforms and fail if playTests.sh isn't run from a terminal, I'll look at disabling it if `stdin`/`stdout` aren't TTYs.

Cyan4973:
good fix @iburinoc !

PR #608, opened 2017-03-18
Add test for creating ZZTD_CDict/ZSTD_DDict

luben:
Currently it's failing but it is passing with 1.1.3

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -167,6 +167,9 @@ longmatch  : $(ZSTD_FILES) longmatch.c
 invalidDictionaries  : $(ZSTD_FILES) invalidDictionaries.c
 	$(CC)      $(FLAGS) $^ -o $@$(EXT)
 
+validDictionaries  : $(ZSTD_FILES) $(ZDICT_FILES) validDictionaries.c
+	$(CC)      $(FLAGS) $^ -o $@$(EXT)
+
 legacy : CFLAGS+= -DZSTD_LEGACY_SUPPORT=4
 legacy : CPPFLAGS+= -I$(ZSTDDIR)/legacy
 legacy : $(ZSTD_FILES) $(wildcard $(ZSTDDIR)/legacy/*.c) legacy.c
--- a/tests/validDictionaries.c
+++ b/tests/validDictionaries.c
@@ -3,40 +3,94 @@
 #include "zstd.h"
 #include "dictBuilder/zdict.h"
 
+#define SIZE 64*1024
+#define DICT_SIZE 32*1024
+
 int main(int argc, const char** argv) {
   ZSTD_CDict *cdict;
   ZSTD_DDict *ddict;
-  size_t dict_size, i;
-  unsigned char *samples = malloc(64*1024);
-  size_t sample_sizes[1] =  { 64*1024 };
-  void *dict = malloc(32*1024);
+  size_t size, dict_size, i;
+  unsigned char *samples = malloc(SIZE);
+  size_t sample_sizes[1] =  { SIZE };
+  void *dict = malloc(DICT_SIZE);
+  size_t compress_bound = ZSTD_compressBound(SIZE);
+  void *compressed = malloc(compress_bound);
+  void *decompressed = malloc(SIZE);
+  ZSTD_CCtx* cctx = ZSTD_createCCtx();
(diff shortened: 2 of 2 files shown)


luben:
In more details:ZSTD_createCDict succeeds but ZSTD_createDDict fails. I can still decompress it using  ZSTD_decompress_usingDict() providing the dictionary buffer

luben:
Actually I also have problems with using the byte buffer dictionaries. I am not sure it's directly related, will update the test case

luben:
Added tests for roundtrip with buffer dicts - the test also fail but all pass on 1.1.3

Cyan4973:
Thanks for the reference test @luben .

I've started playing with it.

First thing I notice : 

`Created dictionary of size 122`

This is suppose to be way to small for a valid dictionary.
I'm surprised the dictionary builder does not simply output an error.
Maybe something to look at.

Now, what I look at what data was provided to create the dictionary : 
https://github.com/luben/zstd/blob/dev/tests/validDictionaries.c#L23

I'm not surprised that the dictionary builder can't find anything useful : it's just a pure noise, no redundancy.

So my expectation is that the dictionary builder should fail on this sample.

Now, this does not tell why the test was working with 1.1.3 and does no longer work with 1.1.4.
More to investigate ...

luben:
Yes, the sample is not the best opportunity for finding redundancy. It came from fuzzy testing the JVM bindings. If you play with the modulus, with 5 it start to pass all the tests as the dictionary is bugger:  817 bytes.

Cyan4973:
As suggested I changed modulo to `% 5`, which generated enough redundancy (due to the limited number of combinations) to create a dictionary.
Indeed, all tests were apparently passed, and without error : 
```
Created dictionary of size 699
Compressed with dict
Decompressed with dict
Created CDict
Created DDict
Compressed with CDict
Decompressed with DDict
```

I guess the title of #609 is appropriate.

In the first case (`% 32`), it does not even reach `ZSTD_createDDict()` : it fails at `ZSTD_decompress_usingDict()`.

In the second case, all tests pass fine.

In my view, it's a bug in the dictionary builder, which should have errored out in the first situation, instead of creating a (bogus) dictionary.

luben:
On the other hand, the first version of the test had the CDict/DDict creation as the first steps, before doing any compression. What is strange that it succeeds creating a CDict and is able to compress using it but fails on creation of DDict. So we are left with compressed data that we can't decompress using the same dictionary we used to compress it.

Cyan4973:
Indeed, irrespective of the dictionary builder issue (after all, a dictionary can be created with a different tool),
compression and decompression should apply the same set of rules regarding acceptance or refusal of data as dictionary. It's a discrepancy to solve.

Cyan4973:
Following investigation with the help of @terrelln , it seems the issue is the expected one : 
- The dictionary is indeed invalid : it contains `repcodes` which are larger than the dictionary content size (which is `0`). The dictionary builder must be updated to not create such dictionary.
- The invalid dictionary was nonetheless accepted by compressor and decompressor. It would just have no impact.
- While developing v1.1.4, a validation rule has been strengthened on the decoder side, with the result of rejecting this kind of dictionary.
- The compressor has not been changed, hence it still accepts the invalid dictionary, and is just doing nothing useful with it.

The current mood is that the new validation rule in the decoder is correct. Which means it should be replicated on the compression side.
And then dictionary builder must also be updated to not produce such dictionaries.

We believe it can break a few 3rd-party tests which have been triggering this use case and expect it to pass. We nonetheless believe it's better to break these tests, and force them to be adjusted to create valid dictionaries.

Patches to follow.

luben:
I agree with your decision. Currently these test are broken anyway if they test the end-to-end roundtrip.

Cyan4973:
There are 2 patchs available in 2 branches to fix this issue : 

- branch `loadCDict` : it ensures that the rules to load dictionary are the same for CDict and DDict. When playing this test case with `loadCDict` branch, the test shall fail at compression stage.

- branch `stricterDictBuilder` : it ensures that the dictionary created has some minimum condition, such as a minimum size. When playing this test case with `stricterDictBuilder` branch, the test shall fail at dictionary builder stage.

Both patches will be merged into `dev`. They can be observed separately in their own branches.

luben:
Great. This will solve the issue.

luben:
Closing it as the issue is addressed

PR #610, opened 2017-03-20
Strongly limit compression level when source is large for tests running on QEMU

iburinoc:
Aim to fix test timeouts on Travis, e.g. https://travis-ci.org/facebook/zstd/jobs/212897352


Cyan4973:
Looks good !

PR #611, opened 2017-03-20
Further cmake improvement

Majlen:
I've added more missing targets (zstdcat, unzstd symlinks, manpage, generated documentation). The build system is also more modular, you can choose to build only libraries, etc.

Installation of targets is using standard cmake variables and should respect FHS (more or less) now. Library version symlinks are generated automatically during install phase. Binaries (with the exception of tests) are using shared libraries instead of static ones.

There's also a big cleanup in AddExtraCompilationFlags, which should be much easier to read now, and small changes to Makefile to test all the branches in CircleCI.

The only downside might be bumping required cmake version to 3.1 due to easier handling of C/C++ standards.


Cyan4973:
That's a giant update you're bringing on the `cmake` front @Majlen !

It looks all good, and am supportive of the improved modularity.

I'm also slightly worried about the new version requirement.
Last time we tried it, we discovered there were many systems out there still using rather old versions of `cmake`, so we targeted `v2.8.7`.

I have no idea how much of a jump `v3.1` is.
`cmake` tests are now run on CircleCI, which delivers `v3.2.2`, which is good enough to pass this test.
I've also checked my own dev environments, and they use `v3.7.2`, so they are pretty safe. 
But that's not the point : question is, how many people will be left out by the increased version requirement ?

That's a difficult one, plus the answer obviously changes with time, as we can expect more and more systems to be updated.

As a general way of dealing with this, a question is : 
which feature(s) relies on `v3.1` availability ?
Is it essential, or could it be traded for another construction with lower version requirement ?

iburinoc:
Notably, Ubuntu 12.04 LTS ships `cmake 2.8.7` and Ubuntu 14.04 LTS ships `cmake 2.8.12`

Cyan4973:
Thanks @iburinoc ! These are pretty good reference points.

With Ubuntu 12.04 LTS nearing its end of life, we might get away with `2.8.7`,
but I also believe it would be reasonable to not move too fast beyond `2.8.12`.
Ubuntu 14.04 LTS is still with us for quite some time,
and there are a few other distros which will be based on it, or provide a similar level of support.

Majlen:
Well, I think, that the only thing depending on 3.1 is the C++11 and C99 handling, which should be pretty straightforward to change. I am currently testing, if that's the only case (I am not sure, I consulted mainly 3.1 documentation, which is easier to read for me).

I would just like to ask, do you know a way to tell MSVC to enforce a given C/C++ standard (if there is a way)?

Majlen:
There was one other thing depending on cmake >3.1, which was the way I used FindThreads module. It wasn't a difficult fix, so the change is in there.

Testing CMake with MSVC on AppVeyor should now be fixed too (the build system does not try to support multithreading on non UNIX platforms).

Cyan4973:
Looks good to me !

So it seems `2.8.7` remains good enough for the new script ? That would be great.
I can merge this PR.

Majlen:
Yes, it should be sufficient.

PR #612, opened 2017-03-20
Emscripten port, first pass

cordovapolymer:
Merged emscripten port with master


Cyan4973:
Thanks for the proposed patch @cordovapolymer !

I would suggest to put the full emscripten project under its own `contrib/emscripten` directory.
Try to make it as self-contained as possible,
that is it can rely on other "stable" directories, such as `lib` for example,
but not the other way round.

Also, try to make a few tests that "prove" that it works, at least in some controlled case, document it in the directory's `README`.

Should you need to modify `cmake` build system, note that there is a big patch ongoing : #611 . Maybe wait for it to be completed before merging your modifications, to avoid conflicts.

And finally : no PR to `master` please.
We only take PR to `dev` or to their own feature branch (I can create one for you if you wish).

PR #625, opened 2017-03-24
limited CDict acceptation criteria to be the same as DDict

Cyan4973:


lib/compress/zstd_compress.c
```diff
@@ -2628,15 +2629,20 @@ static size_t ZSTD_loadDictEntropyStats(ZSTD_CCtx* cctx, const void* dict, size_
     cctx->rep[2] = MEM_readLE32(dictPtr+8); if (cctx->rep[2] == 0 || cctx->rep[2] >= dictSize) return ERROR(dictionary_corrupted);
```
terrelln: We can delete the checks here, since we perform them below.

Cyan4973: Good point !

Issue #624, opened 2017-03-24
Updated build 1.1.4 crashes.


sebbellob:
Build 1.1.4 x64 crashes on Win7 x64 SP1.
Even the parameters cannot be seen,instant crash after running the.exe.
ATM i cant tell more details,because im from mobile.

inikep:
It seems that the issue is dependency of 64-bit build on zlib1.dll
Is the 32-bit executable OK on Win7? It uses a static zlib library.

sebbellob:
Checking ASAP.

PR #627, opened 2017-03-24
Statically build and link zlib on appveyor

iburinoc:
With this, `zstd_x86.exe` and `zstd_x64.exe` artifacts on appveyor will be statically linked to `zlib`, and can be bundled and released with the windows packages.


Cyan4973:
Great ! will probably help #624

PR #635, opened 2017-03-29
Remove extra 'F' from skippable magic mask

iburinoc:
Didn't cause correctness issues, but technically was too large for its type.


Cyan4973:
Good point @iburinoc ! 
Silent truncation.
I'm surprised it did not generated any warning so far...

PR #647, opened 2017-04-05
Add linux kernel modules to contrib/

terrelln:
The header file is in `include/linux/zstd.h` and the sources are in `lib/zstd/`.
There are two independent kernel modules, `zstd_compress` and `zstd_decompress`.
* Spaces were converted to tabs.
* Unnecessary files were deleted.
* `zstd_compress.c` and `zstd_decompress.c` were renamed to `compress.c` and `decompress.c` because they clashed with the kernel module name.
* Cross platform macros were deleted.
* Unused functions were deleted.
* Standard library headers were replaced with kernel headers.
* `mem.h` and `xxhash.c` were changed to use native kernel unaligned access functions.
* Module macros were added to export the symbols.
* All allocating functions were changed to take a `workspace` and `workspaceSize` instead.
   Which means we can delete all freeing functions, since there is nothing to do.
   In this process, `ZSTD_create{C,D}Stream()` and `ZSTD_init{C,D}Stream()` were merged.


Cyan4973:
That's a great patch @terrelln !

Quick question regarding the userland tests :
does it make any sense to have them part of CI "long test" suite ?

Issue #395, opened 2016-09-28
Is zstd splitabble in hadoop/spark/etc? 


trixpan:
I know the 4mc container adds splitability to zstd but is the standard by itself able to do so without the help of 3rd party implementations? 

Note: I am aware zstd still not officially implemented in Hadoop so this is more like a theoretical question.

Cyan4973:
We have plan to re-use and extend the skip format based on skippable frames implemented in `pzstd` for multi-threaded decompression. 4mc's @carlomedas is aware of this context. We may spend some time in the future to properly define it and make it part of the official format.

advancedxy:
Hi, @trixpan, I talked to the 4mc's author, It would be possible to write metadata in zstd skippable frames and make the whole file splittable by hadoop(need a special InputFormat).

But the idea would be same with 4mc in the essential.

trixpan:
@advancedxy I saw 4mc previously. 

I think people's concern about 4mc is the limited community (although I must confess compression algorithms don't strike me as pieces of code with hundreds of contributors)

churrodog:
Is the format in which pzstd uses splittable frames official?  If it is then I can update the hadoop project to utilize anything compressed with zstd-cli then copied into hadoop.  Have you guys decided or finalized the format of pzstd files?  Thank you

Cyan4973:
We don't have yet a splittable format.

This is an active topic, we are working on it, and expect to make progresses throughout March for this objective.
Btw, it's probably the best time to be involved in this process, as we will work on this specification with several projects interested in a splittable format. So don't hesitate if you want to be part of it.

The final format will likely be different from current `pzstd` one.
Not that there is any problem with `pzstd` format, it's just that we can do better.
`pzstd` format will likely be discontinued, though it's not a concern for existing compressed data : `pzstd` format has been designed to be compatible with `zstd` decoder, and it is perfectly decodable by any `zstd` compliant decoder. That will remain true.
We merely are going to use different ways to generate splittable and multi-threaded data.

churrodog:
@Cyan4973 thanks for the update, I think having zstd splittable in hadoop would be very attractive to a lot of users.  I would like to be involved if possible discussing the proposed splittable format.  Thank you.

Cyan4973:
cc @iburinoc

carlomedas:
I definitely agree with this approach: 4mc is nice but we need this as mainstream in zstd.
[...]

PR #654, opened 2017-04-11
[RFC] Splittable Format and API

iburinoc:
This is an initial implementation of a zstd splittable format and API, as per #395.  Comments and feedback are appreciated.


Cyan4973:
For clarity of repository, we want to move the seekable format and demo into `contrib`,
it's not validated as part of Zstandard yet (it will come at a later stage).

contrib/seekable_format/zstd_seekable_compression_format.md
```diff
@@ -0,0 +1,107 @@
...
+### Seek Table Format
+The structure of the seek table frame is as follows:
+
+|`Skippable_Magic_Number`|`Frame_Size`|`[Seek_Table_Entries]`|`Seek_Table_Footer`|
+|------------------------|------------|----------------------|-------------------|
+| 4 bytes                | 4 bytes    | 8-12 bytes each      | 9 bytes           |
+
+__`Skippable_Magic_Number`__
+
+Value : 0x184D2A5?, which means any value from 0x184D2A50 to 0x184D2A5F.
```
Cyan4973: All 16 values are fine for "skippable frames" in general,
but for this specification in particular,
should we pin down a specific number instead ?
Maybe one which is less likely to be used, such as *E ?
cc @terrelln for reference

iburinoc: *E sounds reasonable

contrib/seekable_format/zstd_seekable_compression_format.md
```diff
@@ -0,0 +1,107 @@
...
+__`Frame_Size`__
+
+The total size of the skippable frame, not including the `Skippable_Magic_Number` or `Frame_Size`.  This is for compatibility with [Zstandard skippable frames].
+
+[Zstandard skippable frames]: https://github.com/facebook/zstd/blob/master/doc/zstd_compression_format.md#skippable-frames
+
+#### `Seek_Table_Footer`
+The seek table footer format is as follows:
+
+|`Number_Of_Chunks`|`Seek_Table_Descriptor`|`Seekable_Magic_Number`|
```
Cyan4973: It seems `Seekable_Magic_Number` is not defined afterwards,
maybe it's implied that it is the same as the `Skippable_Magic_Number` in previous paragraph ?
Even if that's so, it needs to be described.

Also interesting : explain the reasoning for this construction.

For reference : https://github.com/thejoshwolfe/yauzl/issues/48#issuecomment-266587526
showing that there is nothing obvious and even commonly distributed formats can have it wrong.

contrib/seekable_format/zstd_seekable_compression_format.md
```diff
@@ -0,0 +1,107 @@
...
+| 1-0        | `Unused_Bits`             |
+
+While only `Checksum_Flag` currently exists, there are 7 other bits in this field that can be used for future changes to the format,
+for example the addition of inline dictionaries.
+
+__`Checksum_Flag`__
+
+If the checksum flag is set, each of the seek table entries contains a 4 byte checksum of the uncompressed data contained in its chunk.
+
+`Reserved_Bits` are not currently used but may be used in the future for breaking changes, so a compliant decoder should ensure they are set to 0.  `Unused_Bits` may be used in the future for non-breaking changes, so a compliant decoder should not interpret these bits.
```
Cyan4973: 👍

contrib/seekable_format/zstd_seekable_compression_format.md
iburinoc: Oops, it does have a value that I forgot to include.

contrib/seekable_format/zstd_seekable_compression_format.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### __`Seek_Table_Entries`__
+
+`Seek_Table_Entries` consists of `Number_Of_Chunks` (one for each chunk in the data, not including the seek table frame) entries of the following form, in sequence:
+
+|`Compressed_Size`|`Decompressed_Size`|`[Checksum]`|
```
Cyan4973: Just a question : what about the initial idea of working with offsets ?

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+ *  Performs a binary search to find the last chunk with a decompressed offset
+ *  <= pos
+ *  @return : the chunk's index */
+static U32 ZSTD_seekable_offsetToChunk(const seekTable_t* table, U64 pos)
+{
+    U32 lo = 0;
+    U32 hi = table->tableLen;
+
+    while (lo + 1 < hi) {
+        U32 mid = lo + ((hi - lo) >> 1);
```
Cyan4973: (minor) `const` whenever you can

contrib/seekable_format/zstd_seekable_compression_format.md
iburinoc: I decided on size because the main upside to offset is not having to parse the entire seek table, but I can't think of scenarios where that's too useful (in large files we have to do a lot of work/reading anyway, in small files the seek table is small), and using offsets would either make the table bigger (8 bytes per field), or complicate the format.  As is a max chunk size of 4 GB seems reasonable.

We could change this however if it turns out the ability to seek to the middle of the jump table is desirable.

contrib/seekable_format/zstd_seekable.h
```diff
@@ -0,0 +1,140 @@
...
+#endif
+
+static const unsigned ZSTD_seekTableFooterSize = 9;
+
+#define ZSTD_SEEKABLE_MAGICNUMBER 0x8F92EAB1
+
+#define ZSTD_SEEKABLE_MAXCHUNKS 0x8000000U
+
+/* 0xFE03F607 is the largest number x such that ZSTD_compressBound(x) fits in a 32-bit integer */
+#define ZSTD_SEEKABLE_MAX_CHUNK_DECOMPRESSED_SIZE 0xFE03F607
```
Cyan4973: Could we get a static assert on this value ?

`ZSTD_compressBound()` was recently changed, it may be changed again in the future.
In case of a future change, it's necessary to ensure that `ZSTD_compressBound(ZSTD_SEEKABLE_MAX_CHUNK_DECOMPRESSED_SIZE)` doesn't overflow.

Note that it may require to expose a macro version of `ZSTD_compressBound()`

contrib/seekable_format/zstd_seekable.h
```diff
@@ -0,0 +1,140 @@
...
+
+/*-****************************************************************************
+*  Seekable Format
+*
+*  The seekable format splits the compressed data into a series of "chunks",
```
Cyan4973: naming : "chunks" or "frames" ?

iburinoc: I guess the meaning lines up well with regular zstd frames, I can change chunks to frames.

contrib/seekable_format/zstd_seekable.h
```diff
@@ -0,0 +1,140 @@
...
+*  a different position is required to continue decompression.
+*  Note that ZSTD_seekable_updateOffset will error if given an offset other
+*  than the one requested from ZSTD_seekable_getSeekOffset().
+******************************************************************************/
+
+/*===== Seekable decompressor management =====*/
+ZSTDLIB_API ZSTD_seekable_DStream* ZSTD_seekable_createDStream(void);
+ZSTDLIB_API size_t ZSTD_seekable_freeDStream(ZSTD_seekable_DStream* zds);
+
+/*===== Seekable decompression functions =====*/
```
Cyan4973: The API feels generally slightly awkward.
I suspect it's designed to look like the generic streaming API, which works from buffer to buffer.

But it could be that `FILE*` operations are in fact a better fit for the job.
Now that could be 2 different set of operations.

These are difficult questions.
We will probably need to discuss this a bit together to converge on a design.
A generic tool for this objective is to develop sample user programs to get a "feel" of how usable the API is.

iburinoc: Yeah, it would be much cleaner with `FILE*` objects, however I wasn't sure if that would be too restrictive for layers built on top.

contrib/seekable_format/zstd_seekable_compression_format.md
Cyan4973: I agree.
The main benefit to using offsets is to keep the table on-disk, to preserve memory usage, only reading the required offsets when useful.
I can think of a few scenarios which would benefit from this construction, but they all involve specific situations (lowmem) we should not worry about at this stage. There's ample time to decide if it's a worthwhile scenario in some distant future.

contrib/seekable_format/zstd_seekable.h
Cyan4973: It certainly is restrictive, and that's why we should have a brainstorm session about it.

contrib/seekable_format/zstdseek_compress.c
```diff
@@ -0,0 +1,298 @@
...
+     * zcs->frameCSize: the amount of the table written so far
+     *
+     * This function is written this way so that if it has to return early
+     * because of a small buffer, it can keep going where it left off.
+     */
+
+    size_t const sizePerFrame = 8 + (zcs->checksumFlag?4:0);
+    size_t const seekTableLen = ZSTD_seekable_seekTableSize(zcs);
+
+#define st_write32(x, o)                                                       \
```
terrelln: Can you make this a static function instead, and check if the result is non-zero.

contrib/seekable_format/zstdseek_compress.c
```diff
@@ -0,0 +1,298 @@
...
+                       ZSTD_skippableHeaderSize + sizePerFrame * zcs->frameDSize + 8);
+        }
+
+        zcs->frameDSize++;
+    }
+
+    st_write32(zcs->framelog.size, seekTableLen - ZSTD_seekTableFooterSize);
+
+    if (output->size - output->pos < 1) return seekTableLen - zcs->frameCSize;
+    if (zcs->frameCSize < seekTableLen - 4) {
```
terrelln: Why is this if statement here?

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+        checksumFlag = sfd >> 7;
+
+        /* check reserved bits */
+        if ((checksumFlag >> 2) & 0x1f) {
+            return ERROR(corruption_detected);
+        }
+    }
+
+    numFrames = MEM_readLE32(ip-9);
+    sizePerEntry = 8 + (checksumFlag?4:0);
```
terrelln: These can be dropped down to the next scope and made `const`.

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+                free(entries);
+                return ERROR(memory_allocation);
+            }
+
+            /* compute cumulative positions */
+            for (idx = 0, pos = 0; idx < numFrames; idx++) {
+                entries[idx].cOffset = cOffset;
+                entries[idx].dOffset = dOffset;
+
+                cOffset += MEM_readLE32(tableBase + pos); pos += 4;
```
terrelln: Can you put each statement on its own line.

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+        }
+    }
+}
+
+size_t ZSTD_seekable_initDStream(ZSTD_seekable_DStream* zds, U64 rangeStart, U64 rangeEnd)
+{
+    /* restrict range to the end of the file, of non-negative size */
+    rangeStart = MIN(rangeStart, zds->seekTable.entries[zds->seekTable.tableLen].dOffset);
+    rangeEnd = MIN(rangeEnd, zds->seekTable.entries[zds->seekTable.tableLen].dOffset);
+    rangeEnd = MAX(rangeEnd, rangeStart);
```
terrelln:
```c
rangeEnd = MIN(rangeEnd, zds->seekTable.entries[zds->seekTable.tableLen].dOffset);
rangeStart = MIN(rangeStart, rangeEnd);
```

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+        if (MEM_readLE32(base) != (ZSTD_MAGIC_SKIPPABLE_START | 0xE)) {
+            return ERROR(prefix_unknown);
+        }
+        if (MEM_readLE32(base+4) + ZSTD_skippableHeaderSize != frameSize) {
+            return ERROR(prefix_unknown);
+        }
+
+        {   /* Allocate an extra entry at the end so that we can do size
+             * computations on the last element without special case */
+            seekEntry_t* entries = malloc(sizeof(seekEntry_t) * (numFrames + 1));
```
terrelln: I think some compilers will complain that you don't cast to a `seekEntry_t*`.

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+            return ERROR(init_missing); /* ZSTD_seekable_initDStream should be called first */
+        case zsds_decompress: {
+            BYTE* const outBase = (BYTE*)output->dst + output->pos;
+            size_t const outLen = output->size - output->pos;
+            while (zds->decompressedOffset < zds->targetStart) {
+                U64 const toDecompress =
+                        zds->targetStart - zds->decompressedOffset;
+                size_t const prevInputPos = input->pos;
+
+                ZSTD_outBuffer outTmp = {
```
terrelln: Same question here about C99.

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+    rangeStart = MIN(rangeStart, zds->seekTable.entries[zds->seekTable.tableLen].dOffset);
+    rangeEnd = MIN(rangeEnd, zds->seekTable.entries[zds->seekTable.tableLen].dOffset);
+    rangeEnd = MAX(rangeEnd, rangeStart);
+
+    zds->targetStart = rangeStart;
+    zds->targetEnd = rangeEnd;
+    zds->stage = zsds_seek;
+
+    /* force a seek first */
+    zds->curFrame = (U32) -1;
```
terrelln: No space between cast and value

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -0,0 +1,367 @@
...
+                if (ret == 0) {
+                    /* frame is done */
+                    /* make sure this lines up with the expected frame border */
+                    if (zds->decompressedOffset !=
+                                jt->entries[zds->curFrame + 1].dOffset ||
+                        zds->compressedOffset !=
+                                jt->entries[zds->curFrame + 1].cOffset)
+                        return ERROR(corruption_detected);
+                    ZSTD_resetDStream(zds->dstream);
+                    zds->stage = zsds_seek;
```
terrelln: Why do you need to seek after this? You've already asserted you are in the correct spot, right?

contrib/seekable_format/zstdseek_compress.c
iburinoc: It's the same logic that's present in `st_write32` to avoid writing data we've written in a previous call.  Without it, the seekable frame descriptor could be written multiple times, instead of writing the magic number.

contrib/seekable_format/zstdseek_decompress.c
iburinoc: The next frame to read might not be the frame immediately after, e.g. in the case of interleaved skippable frames.  This will seek over them.

terrelln: Right, since `dOffset == 0` for skippable frames. Can you add a comment that explains that?

iburinoc: Not quite, `dSize == 0` for skippable frames, which translates to `entries[i].dOffset == entries[i+1].dOffset` if `i` is the index of a skippable frame.  Using offset in the decoder even though the format uses size is a bit confusing, but necessary for the decoder to be efficient.  I'll add a comment explaining why the seek is necessary.

contrib/seekable_format/zstd_seekable.h
```diff
@@ -0,0 +1,164 @@
...
+ZSTDLIB_API size_t ZSTD_seekable_decompressFrame(ZSTD_seekable* zs, void* dst, size_t dstSize, unsigned frameIndex);
+
+/*===== Seek Table access functions =====*/
+ZSTDLIB_API unsigned ZSTD_seekable_getNumFrames(ZSTD_seekable* const zs);
+ZSTDLIB_API unsigned long long ZSTD_seekable_getFrameCompressedOffset(ZSTD_seekable* const zs, unsigned frameIndex);
+ZSTDLIB_API unsigned long long ZSTD_seekable_getFrameDecompressedOffset(ZSTD_seekable* const zs, unsigned frameIndex);
+ZSTDLIB_API size_t ZSTD_seekable_getFrameCompressedSize(ZSTD_seekable* const zs, unsigned frameIndex);
+ZSTDLIB_API size_t ZSTD_seekable_getFrameDecompressedSize(ZSTD_seekable* const zs, unsigned frameIndex);
+
+ZSTDLIB_API unsigned ZSTD_seekable_offsetToFrame(ZSTD_seekable* const zs, unsigned long long offset);
```
Cyan4973: `ZSTD_seekable_offsetToFrameIndex` for consistent naming

contrib/seekable_format/zstd_seekable.h
```diff
@@ -0,0 +1,164 @@
...
+static const unsigned ZSTD_seekTableFooterSize = 9;
+
+#define ZSTD_SEEKABLE_MAGICNUMBER 0x8F92EAB1
+
+#define ZSTD_SEEKABLE_MAXFRAMES 0x8000000U
+
+/* Limit the maximum size to avoid any potential issues storing the compressed size */
+#define ZSTD_SEEKABLE_MAX_FRAME_DECOMPRESSED_SIZE 0x80000000U
+
+#define ZSTD_SEEKABLE_FRAMEINDEX_TOOLARGE (0ULL-2)
```
Cyan4973: This macro value is of different nature.
I would rather `#define` it closer to the family of functions which need it.

iburinoc:
Demo using `parallel_processing.c`, which sums up the values of all the bytes in seekable compressed file by processing frames in parallel using a thread pool:

```
# create test file
devvm27563 examples [splittable] $ ../../../tests/datagen -g4096M > test.dat
devvm27563 examples [splittable ?] $ gcc -x c - -O3 <<EOF
> #include <stdio.h>
> int main() {
> char c;
> unsigned long long total = 0;
> while ((c = fgetc(stdin)) != EOF) { total += c; }
> printf("Sum: %llu\n", total);
> return 0;
> }
> EOF
# compute the expected answer with a test program
devvm27563 examples [splittable ?] $ cat test.dat | ./a.out
Sum: 240342218668
# create seekable file (compression level 5)
devvm27563 examples [splittable ?] $ time ./seekable_compression test.dat $((16 * 1024 * 1024))

real    1m29.590s
user    1m27.569s
sys     0m2.005s
# test with 1, 2, 4, 8, 16 threads (on a 24 core machine)
devvm27563 examples [splittable ?] $ time ./parallel_processing test.dat.zst 1
Sum: 240342218668

real    0m9.352s
user    0m8.946s
sys     0m0.399s
devvm27563 examples [splittable ?] $ time ./parallel_processing test.dat.zst 2
Sum: 240342218668

real    0m4.701s
user    0m9.004s
sys     0m0.383s
devvm27563 examples [splittable ?] $ time ./parallel_processing test.dat.zst 4
Sum: 240342218668

real    0m2.485s
user    0m9.426s
sys     0m0.421s
devvm27563 examples [splittable ?] $ time ./parallel_processing test.dat.zst 8
Sum: 240342218668

real    0m1.289s
user    0m9.443s
sys     0m0.499s
devvm27563 examples [splittable ?] $ time ./parallel_processing test.dat.zst 16
Sum: 240342218668

real    0m0.768s
user    0m10.493s
sys     0m0.709s
```
As seen from the timing results, the decompression and processing speed parallelizes very efficiently, gaining a ~2x speed up each time the number of cores is doubled.

Cyan4973:
Yes, that's a _great_ example, very clear.
Hopefully, it also helps to get a user-side feeling of the API for such a use case.

Obviously, now that you successfully completed parallel processing,
there will be one to wonder, why not compression in parallel too ...

iburinoc:
Demo using parallel_compression.c, which does compression manually and uses the seekable API to construct a seek table from it. (./parallel_compression FILE BLOCK_SIZE NB_THREADS).
```
devvm25806 examples [splittable ?] $ ./datagen -g4192M > test.dat
devvm25806 examples [splittable !?] $ time ./parallel_compression test.dat 1048576 1

real    1m28.619s
user    1m25.908s
sys     0m5.726s
devvm25806 examples [splittable ?] $ time ./parallel_compression test.dat 1048576 2

real    0m45.397s
user    1m25.755s
sys     0m5.411s
devvm25806 examples [splittable ?] $ time ./parallel_compression test.dat 1048576 4

real    0m24.319s
user    1m27.992s
sys     0m5.293s
devvm25806 examples [splittable ?] $ time ./parallel_compression test.dat 1048576 8

real    0m13.489s
user    1m28.168s
sys     0m5.592s
devvm25806 examples [splittable ?] $ time ./parallel_compression test.dat 1048576 16

real    0m8.563s
user    1m30.531s
sys     0m5.454s
```

The files can be decompressed using `zstd` and are compatible with the seekable capabilities of `parallel_processing` and `seekable_decompression`.

Cyan4973:
To be merged after release `v1.2.0`

PR #656, opened 2017-04-13
Make -T0 mean "number of physical cores"

iburinoc:
Need separate core counting routines for windows, OS X, and Linux.

Tested on (HT: HyperThreading):
Mac, HT
Linux, HT
Linux, no HT
Windows, no HT
Linux, Aarch64, no HT

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -147,12 +147,10 @@ typedef struct {
 
 
 
-#ifndef MIN
-#   define MIN(a,b)    ((a) < (b) ? (a) : (b))
-#endif
-#ifndef MAX
-#   define MAX(a,b)    ((a) > (b) ? (a) : (b))
-#endif
+#undef MIN
+#undef MAX
+#define MIN(a,b)    ((a) < (b) ? (a) : (b))
+#define MAX(a,b)    ((a) > (b) ? (a) : (b))
 
 static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
                         const char* displayName, int cLevel,
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -146,8 +146,13 @@ typedef struct {
 } blockParam_t;
 
 
-#define MIN(a,b) ((a)<(b) ? (a) : (b))
-#define MAX(a,b) ((a)>(b) ? (a) : (b))
+
+#ifndef MIN
+#   define MIN(a,b)    ((a) < (b) ? (a) : (b))
+#endif
+#ifndef MAX
+#   define MAX(a,b)    ((a) > (b) ? (a) : (b))
+#endif
 
 static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
                         const char* displayName, int cLevel,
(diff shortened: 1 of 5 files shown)


programs/bench.c
```diff
@@ -146,8 +146,13 @@ typedef struct {
 } blockParam_t;
 
 
-#define MIN(a,b) ((a)<(b) ? (a) : (b))
-#define MAX(a,b) ((a)>(b) ? (a) : (b))
+
+#ifndef MIN
```
Cyan4973: What happens if `MIN` or `MAX` are already defined ?

iburinoc: We get a compiler warning on some compilers

programs/util.h
```diff
@@ -488,6 +489,183 @@ UTIL_STATIC void UTIL_freeFileList(const char** filenameTable, char* allocatedBu
...
+
+#elif defined(__APPLE__)
+
+#include <sys/sysctl.h>
+
+/* Use apple-provided syscall
+ * see: man 3 sysctl */
+UTIL_STATIC int UTIL_countPhysicalCores(void)
+{
+    static S32 numPhysicalCores; /* apple specifies int32_t */
```
Cyan4973: Why is it not set to `0` ? Is that implied ?

programs/util.h
```diff
@@ -488,6 +489,183 @@ UTIL_STATIC void UTIL_freeFileList(const char** filenameTable, char* allocatedBu
...
+
+    if (numPhysicalCores != 0) return numPhysicalCores;
+
+    numPhysicalCores = (int)sysconf(_SC_NPROCESSORS_ONLN);
+    if (numPhysicalCores == -1) {
+        /* value not queryable, fall back on 1 */
+        return numPhysicalCores = 1;
+    }
+
+    /* try to determine if there's hyperthreading */
```
Cyan4973: How long last this step ?

programs/util.h
iburinoc: It's implied for static variables, which keep their value across calls (otherwise we wouldn't be able to tell which call was the first time).  This way we don't have to recompute it if it's called twice.

iburinoc: Nevermind apparently `static S32 numPhysicalCores = 0;` doesn't get reset to 0 on every call, only the first.  Didn't know that.  Should I make the initial value of 0 explicit?

Cyan4973: I tend to do that. It makes it clear what the initial value is.

Without it, it relies entirely on implicit assumption that it should be initialised to zero by default. Not very difficult, just a bit confusing when compared with non-static variables, which are _not_ initialised at all.

programs/util.h
iburinoc: I tested exec'ing `zstd -T0` 1000 times in a loop and timing the total time of that with and without the call to UTIL_countPhysicalCores and it seems like without the call, the average running time is 1.136s, and with it it's 1.137s (so probably negligible, given variance).

programs/util.h
iburinoc: Good point, although I think I'll always find reading code that looks like this a bit weird...
```
static int i = 0;
if (i != 0) return i;
```

PR #659, opened 2017-04-15
Enable multithreading on BSD

bapt:


bapt:
Note that actually Apple could use the same code

Cyan4973:
Thanks @bapt , looks good !

We'll see if it can be coalesced with OS-X too

PR #660, opened 2017-04-17
Add multithreading tests to playTests.sh

iburinoc:


tests/playTests.sh
```diff
@@ -461,6 +466,16 @@ roundTripTest -g516K 19   # btopt
 
 fileRoundTripTest -g500K
 
+if [ -n "$hasMT" ]
+then
+    $ECHO "\n**** zstdmt round-trip tests **** "
+    roundTripTest -g516K "16 -T0"
```
Cyan4973: More data is necessary to trigger multi-threading at these levels.
For levels >= 16 MB, expect about ~32 MB per job (by default).
Level 1 is 2 MB, Level 3 (default) is 4 MB.

This can be overriden by command `-B#`, allowing to select a custom job size.
But even there, there is a minimum 1 MB applied for job size.

iburinoc: Ah ok, I'll fix that then.

PR #661, opened 2017-04-17
Make zstd compile with mt by default

iburinoc:


programs/Makefile
```diff
@@ -231,6 +246,9 @@ install: zstd
 	@$(INSTALL_PROGRAM) zstd $(DESTDIR)$(BINDIR)/zstd
 	@ln -sf zstd $(DESTDIR)$(BINDIR)/zstdcat
 	@ln -sf zstd $(DESTDIR)$(BINDIR)/unzstd
+ifeq ($(HAVE_THREAD),1)
+	@ln -sf zstd $(DESTDIR)$(BINDIR)/zstdmt
```
Cyan4973: Maybe we should keep `zstdmt` link as part of installation even when multi-threading is not enabled.
The goal would be to have a predictable installation scope.

Consider a scenario where some scripts are deployed on many different machines :
Invoking `zstdmt` on a non-pthread system will basically get `zstd` single-threaded with a warning message that multi-threading is not enabled. This is probably better than reaching an error.

Cyan4973:
Big :+1 for taking the time to also update `cmake` and `Visual` build scripts

PR #662, opened 2017-04-17
Add zstdmt and -T0 to man page

iburinoc:


programs/zstd.1.md
```diff
@@ -101,6 +103,8 @@ the last one takes effect.
     Note that decompression will also require more memory when using these levels.
 * `-T#`:
     Compress using # threads (default: 1).
+    If `#` is 0, attempt to detect the number of physical CPU cores and compress with
+    that many threads.
     This modifier is only available if `zstd` was compiled with multithreading support.
```
Cyan4973: What happens if `zstd` is compiled without multithreading support ?

I guess `-T#` still "works", in the sense that the CLI doesn't exit with an error code,
it just outputs a warning message about the fact it's single threaded.

In which case, the modifier is "available", it just doesn't allow to use anything else than 1.
(The difference can be important for scripts which hard-code the use of `-T#` on multiple machines)

PR #666, opened 2017-04-18
Add MT enabled targets for libzstd

iburinoc:


Cyan4973:
Is it possible to add some instructions in the README ?

Which targets exist (`lib-mt` may not be obvious)
what are the side effects (calling `ZSTDMT_*` functions from a non-mt library)
and build impact (`-pthread` flag when linking against the `-mt` version)

PR #668, opened 2017-04-19
Add zlib, lzma, pkg-config support to cmake build system

Majlen:
New round of cmake improvements:

- added lzma, zlib support
- cmake creates pkg-config file from prepared template
- files are copied to working copy during build not during configure phase anymore (should be less confusing this way)
- fixed appending compile definitions

This PR bumps the required cmake version to 2.8.9 to support finding lzma library. I think we agreed on sticking to versions < 2.8.12 in earlier iteration, but feel free to correct me.


Cyan4973:
Great patch for `cmake` users @Majlen !

> I think we agreed on sticking to versions < 2.8.12

Correct.
The previous restriction of using < 2.8.7 was related to Ubuntu Precise (LTS 12.04) which is reaching its end of life. Our new compatibility target is now Ubuntu Trusty (LTS 14.04), which provides `cmake` 2.8.12

Cyan4973:
note : `make cmakebuild` currently fails on my test systems.

I suspect that's because `Makefile` is not created in `build/cmake/build`, for some reason.

All tested systems have a rather recent version of `cmake` (v3+).
Maybe there is a difference with v2.8.x in the way directories are managed ?

Majlen:
I don't think there are differences in that. On my system with cmake-3.7 (on Debian stretch) I have no issues. I can take a look on that if you provide me with some output.

Maybe you're missing zlib and lzma headers? Callling `make cmakebuild` always builds the zstd binary with support for those formats. It is enforced to check all the possible options in Circle CI.

Cyan4973:
I most likely miss lzma headers.
But it doesn't seem to be the problem (or at least it does not mention it).

Command is just `make cmakebuild`. Here is the result on my laptop : 
```
make cmakebuild
cmake --version
cmake version 3.7.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
rm -f -r build/cmake/build
mkdir build/cmake/build
cd build/cmake/build ; cmake -DCMAKE_INSTALL_PREFIX:PATH=~/install_test_dir -DZSTD_BUILD_CONTRIB:BOOL=ON -DZSTD_BUILD_STATIC:BOOL=ON -DZSTD_BUILD_TESTS:BOOL=ON -DZSTD_ZLIB_SUPPORT:BOOL=ON -DZSTD_LZMA_SUPPORT:BOOL=ON .. ; /Library/Developer/CommandLineTools/usr/bin/make install ; /Library/Developer/CommandLineTools/usr/bin/make uninstall
-- ZSTD_LEGACY_SUPPORT not defined!
ZSTD VERSION 1.1.5
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/cyan/dev/zstd/github/build/cmake
make[1]: *** No rule to make target `install'.  Stop.
make[1]: *** No rule to make target `uninstall'.  Stop.
make: *** [cmakebuild] Error 2
```

Note the `Build files` line, it does not seem to written in the expected directory

Cyan4973:
Note 2 :
it also makes our daily chron job fail, due to absence of lzma : 
https://travis-ci.org/facebook/zstd/jobs/223865114

We will have to find a way to make this test work,
(build without lzma ? install lzma ?)
as we don't want to keep a bad signal on CI results.

iburinoc:
We can fix CI with `brew install lzma`, I can do that.  It's interesting that your local build fails, `make cmakebuild` works for me (both on local (`cmake version 3.7.2`) and on Centos7 server).

PR #669, opened 2017-04-20
CMake improvements when embedding in another project

ligfx:


Cyan4973:
Thanks for your patch@ligfx ,
my `cmake` skill is a bit limited, so I'm not sure if I can help to solve the current issue :
https://circleci.com/gh/facebook/zstd/221#tests/containers/1

Keep in mind we want to keep `cmake` version requirement at <= 2.8.12.
This is a strong requirement for reasonable portability.
It may be related to the error reported on Circle CI, or may not, I'm not sure.

cc @Majlen, who contributed the most to this `cmake` script.

build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
```diff
@@ -2,24 +2,24 @@ include(CheckCXXCompilerFlag)
 include(CheckCCompilerFlag)
 
 function(EnableCompilerFlag _flag _C _CXX)
-    message("Checking flag ${_flag}")
+    string(REGEX REPLACE "[^A-Za-z0-9+]+" "_" varname "${_flag}")
```
iburinoc: This is the part that causes the fail, at least with `2.8.12`.  It's because CMake tries to compile the var name as a regex inside of `CHECK_CXX_COMPILER_FLAG` (no clue why...) so "`..._STD_C++11`" fails because it's not valid.  Easiest way to fix it I think would just be to take the `+` out of the match group here so it's converted to a `_`.

ligfx: Good catch, I pushed a fix.

PR #673, opened 2017-04-24
Add LZ4 compress/decompress support to CLI

iburinoc:


terrelln:
This only works with lz4-1.7.3 and newer, because of the use of `LZ4F_cctx` and `LZ4F_dctx`.  It needs to work with at least `lz4-r131`, since that is the newest version available in many package managers. Using the obsolete enums, I think we can support down to `lz4-r123`.

1. We need to decide which version of lz4 we want to support. The two obvious choices are `lz4-r123`, the version where lz4 frame is introduced, and `lz4-r129`, where the enums are deprecated.
2. We need to add a version check in the Makefile, checking both `LZ4_VERSION_NUMBER` and `LZ4_versionNumber()`.

iburinoc:
Built and tested (playTests.sh) the new version against r123, r131, and latest.

Cyan4973:
Great patch @iburinoc , and great comment from @terrelln !
Glad you spotted this versioning issue

programs/fileio.c
```diff
@@ -514,6 +520,79 @@ static unsigned long long FIO_compressLzmaFrame(cRess_t* ress, const char* srcFi
...
+    LZ4F_compressionContext_t ctx;
+
+    LZ4F_errorCode_t const errorCode = LZ4F_createCompressionContext(&ctx, LZ4F_VERSION);
+    if (LZ4F_isError(errorCode)) EXM_THROW(31, "zstd: failed to create lz4 compression context");
+
+    memset(&prefs, 0, sizeof(prefs));
+
+    prefs.autoFlush = 1;
+    prefs.compressionLevel = compressionLevel;
+    prefs.frameInfo.blockMode = blockIndependent; /* stick to defaults for lz4 cli */
```
terrelln: `make clean zstd4 CC=gcc-6` emits deprecation warnings.

```c
#if LZ4_VERSION_NUMBER <= 10600
#define LZ4F_blockIndependent blockIndependent
#define LZ4F_max4MB max4MB
/* ... */
#endif


/* ... */
prefs.frameInfo.blockMode = LZ4F_blockIndependent;
```

Cyan4973: Great work-around !

Indeed, using the old enums on newer versions of lz4frame is supposed to trigger deprecation warnings.
This is by design, though it could be disabled by using `-Wno-deprecated` with `gcc`
(note : there is no custom macro to do that).

However, there is a long term intention to remove the legacy enums from lz4frame at some point in the future, since they clobber namespace. Hence, we'd better comply with this long-term plan, and use new enums instead.

The version-conditioned `#define` replacement is a great solution which makes it possible to remain compatible with older versions.

PR #680, opened 2017-04-27
SquashFS linux kernel patch

iburinoc:
Depends on https://github.com/facebook/zstd/pull/679 and existing zstd kernel patch.

Tested locally using https://github.com/iburinoc/squashfs-tools on 4.6-rc7 kernel.

### DIFF
--- a/contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
+++ b/contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
@@ -1,0 +1,146 @@
+/*
+ * Squashfs - a compressed read only filesystem for Linux
+ *
+ * Copyright (c) 2017 Facebook
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2,
+ * or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ * zstd_wrapper.c
+ */
+
+#include <linux/mutex.h>
+#include <linux/buffer_head.h>
+#include <linux/slab.h>
+#include <linux/zstd.h>
+#include <linux/vmalloc.h>
+
+#include "squashfs_fs.h"
+#include "squashfs_fs_sb.h"
+#include "squashfs.h"
+#include "decompressor.h"
+#include "page_actor.h"
+
+struct workspace {
+	void *mem;
+	size_t mem_size;
+};
+
+static void *zstd_init(struct squashfs_sb_info *msblk, void *buff)
+{
+	struct workspace *wksp = kmalloc(sizeof(*wksp), GFP_KERNEL);
+	if (wksp == NULL)
+		goto failed;
+	wksp->mem_size = ZSTD_DStreamWorkspaceBound(max_t(size_t,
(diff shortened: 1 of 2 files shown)


contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
```diff
@@ -0,0 +1,146 @@
...
+			in_buf.pos = 0;
+			offset = 0;
+		}
+
+		if (out_buf.pos == out_buf.size) {
+			out_buf.dst = squashfs_next_page(output);
+			out_buf.pos = 0;
+			if (out_buf.dst != NULL) {
+				out_buf.size = PAGE_SIZE;
+			} else {
```
terrelln: When does `squashfs_next_page()` return `NULL`? Could it return non-null later?

Does `zlib_inflate()` fail when `avail_out == 0`, or does it just no-op? If it fails, does zstd also fail? If zstd doesn't fail, could it spin forever?

iburinoc: It looks like it has a fixed number of pages and returns NULL when it reaches the end of that.  And yeah it looks like `zlib_inflate` fails if `avail_out == 0`, while `ZSTD_decompressStream` doesn't.  Looking at the LZ4 and LZO wrappers, it looks like in this case they just stop and return (don't return an error), so I'll update to match that behaviour.

iburinoc: Nevermind, they do output an error if there's more data to write than space in the pages given.

contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
```diff
@@ -0,0 +1,147 @@
...
+	ZSTD_DStream *stream;
+	size_t total_out = 0;
+	size_t zstd_err;
+	int k = 0;
+	ZSTD_inBuffer in_buf = { NULL, 0, 0 };
+	ZSTD_outBuffer out_buf = { NULL, 0, 0 };
+
+	stream = ZSTD_initDStream(wksp->mem_size, wksp->mem, wksp->mem_size);
+
+	if (!stream) {
```
terrelln: Looking at `zlib_wrapper.c`, they are careful to always call `squashfs_first_page()` and `squashfs_finish_page()`. Is this necessary, or can you exit without calling any of them?

contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
```diff
@@ -0,0 +1,147 @@
...
+		zstd_err = ZSTD_decompressStream(stream, &out_buf, &in_buf);
+		total_out += out_buf.pos; /* add the additional data produced */
+
+		if (in_buf.pos == in_buf.size && k < b)
+			put_bh(bh[k++]);
+	} while (zstd_err != 0 && !ZSTD_isError(zstd_err));
+
+	if (ZSTD_isError(zstd_err)) {
+		ERROR("zstd decompression error: %d\n",
+				(int)ZSTD_getErrorCode(zstd_err));
```
terrelln: Do you need a `goto out;` here?

contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
```diff
@@ -0,0 +1,147 @@
...
+			in_buf.src = bh[k]->b_data + offset;
+			in_buf.size = avail;
+			in_buf.pos = 0;
+			offset = 0;
+		}
+
+		if (out_buf.pos == out_buf.size) {
+			out_buf.dst = squashfs_next_page(output);
+			if (out_buf.dst == NULL) {
+				/* shouldn't run out of pages before stream is
```
terrelln: Do you need to call `squashfs_finish_page()`?

contrib/linux-kernel/fs/squashfs/zstd_wrapper.c
iburinoc: Ok it seems like you definitely need to call squashfs_finish_page if you call squashfs_first_page, but looking at the LZO wrapper and looking at page_actor.c, it seems that it's not necessary if you never call squashfs_first_page.

PR #683, opened 2017-05-02
[CLI] Make cover the default dictionary builder

terrelln:
* `--train` now uses `--train-cover=d=8,steps=4`.
* `--cover` and `--optimize-cover` are now written as `--train-cover`.
* The legacy dictionary builder is accessed through `--train-legacy`.
   The help and manual docs for `-s` were removed, but it is still supported.
* Make `COVER_optimizeTrainFromBuffer()` faster by checking only 2 values for `d`, and slightly modify `k` values.
* Optimize `COVER_[strict_]cmp()` for the common case where `d <= 8`.
* Fix `LZ4_MSG` in `xzstd` target.


Cyan4973:
Looks good !

PR #690, opened 2017-05-09
lib/common: warning: this statement may fall through

joscollin:
The following warning appears during the build:
```
zstd/lib/common/bitstream.h: In function ‘BIT_initDStream’:
zstd/lib/common/bitstream.h:277:33: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 7: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[6]) << (sizeof(bitD->bitContainer)*8 - 16);
              
zstd/lib/common/bitstream.h:278:6: note: here
      case 6: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[5]) << (sizeof(bitD->bitContainer)*8 - 24);

zstd/lib/common/bitstream.h:278:33: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 6: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[5]) << (sizeof(bitD->bitContainer)*8 - 24);

zstd/lib/common/bitstream.h:279:6: note: here
      case 5: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[4]) << (sizeof(bitD->bitContainer)*8 - 32);

zstd/lib/common/bitstream.h:279:33: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 5: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[4]) << (sizeof(bitD->bitContainer)*8 - 32);

zstd/lib/common/bitstream.h:280:6: note: here
      case 4: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[3]) << 24;

zstd/lib/common/bitstream.h:280:33: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 4: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[3]) << 24;

zstd/lib/common/bitstream.h:281:6: note: here
      case 3: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[2]) << 16;

zstd/lib/common/bitstream.h:281:33: warning: this statement may fall through [-Wimplicit-fallthrough=]
      case 3: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[2]) << 16;

zstd/lib/common/bitstream.h:282:6: note: here
      case 2: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[1]) <<  8;
```
Signed-off-by: Jos Collin <jcollin@redhat.com>

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -39,7 +39,6 @@
 extern "C" {
 #endif
 
-
 /*
 *  This API consists of small unitary functions, which must be inlined for best performance.
 *  Since link-time-optimization is not available for all compilers,
@@ -303,13 +302,25 @@ MEM_STATIC size_t BIT_initDStream(BIT_DStream_t* bitD, const void* srcBuffer, si
         bitD->bitContainer = *(const BYTE*)(bitD->start);
         switch(srcSize)
         {
-            case 7: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[6]) << (sizeof(bitD->bitContainer)*8 - 16);
-            case 6: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[5]) << (sizeof(bitD->bitContainer)*8 - 24);
-            case 5: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[4]) << (sizeof(bitD->bitContainer)*8 - 32);
-            case 4: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[3]) << 24;
-            case 3: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[2]) << 16;
-            case 2: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[1]) <<  8;
-            default:;
+	    case 7: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[6]) << (sizeof(bitD->bitContainer)*8 - 16);
+	            /* fall-through */
+
+	    case 6: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[5]) << (sizeof(bitD->bitContainer)*8 - 24);
+	            /* fall-through */
+	
+	    case 5: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[4]) << (sizeof(bitD->bitContainer)*8 - 32);
+	            /* fall-through */
+	
+	    case 4: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[3]) << 24;
+	            /* fall-through */
+	
+	    case 3: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[2]) << 16;
+	            /* fall-through */
+	
+	    case 2: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[1]) <<  8;
+	            /* fall-through */
+	
+            default: break;
         }
         { BYTE const lastByte = ((const BYTE*)srcBuffer)[srcSize-1];
           bitD->bitsConsumed = lastByte ? 8 - BIT_highbit32(lastByte) : 0;


joscollin:
Retest this please

terrelln:
Thanks for the PR @joscollin! In this case the fall-through is intentional. For case N, each of the cases 2 through N-1 need to be called.

To quiet this we can add a cross-platform macro `FALLTHROUGH`, similar to `FORCE_INLINE` in `zstd_internal.h` that expands to `__attribute__ ((fallthrough))` on clang and gcc versions that support it.

```c
switch (srcSize)
{
    case 7:
         /* ... */
         FALLTHROUGH;
    case 6:
        /* ... */
}
```

joscollin:
@terrelln Done, Please review now. Thanks.

lib/common/bitstream.h
```diff
@@ -39,6 +39,17 @@
 extern "C" {
 #endif
 
+#  if defined (__cplusplus) || defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* C99 */
+#    ifdef __GNUC__
+#      define FALLTHROUGH static inline __attribute__((fallthrough))
```
terrelln: The define should be
```c
#define FALLTHROUGH __attribute__((fallthrough))
```

lib/common/bitstream.h
```diff
@@ -39,6 +39,15 @@
 extern "C" {
 #endif
 
+#if defined (__cplusplus) || defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* C99 */
+# if __GNUC__ >= 7
+#  define FALLTHROUGH __attribute__((fallthrough))
+# else
+#  define FALLTHROUGH
+# endif
+#endif /* __STDC_VERSION__ */
```
terrelln: The build is still failing for Appveyor. The C99 check isn't necessary since we don't use the `inline` keyword. This should do the trick:

```c
#if __GNUC__ >= 7
#  define FALLTHROUGH __attribute__((fallthrough))
#else
#  define FALLTHROUGH
#endif
```

lib/common/bitstream.h
```diff
@@ -274,13 +283,32 @@ MEM_STATIC size_t BIT_initDStream(BIT_DStream_t* bitD, const void* srcBuffer, si
...
         {
-            case 7: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[6]) << (sizeof(bitD->bitContainer)*8 - 16);
-            case 6: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[5]) << (sizeof(bitD->bitContainer)*8 - 24);
-            case 5: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[4]) << (sizeof(bitD->bitContainer)*8 - 32);
-            case 4: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[3]) << 24;
-            case 3: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[2]) << 16;
-            case 2: bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[1]) <<  8;
-            default:;
+	case 7: {
+	  bitD->bitContainer += (size_t)(((const BYTE*)(srcBuffer))[6]) << (sizeof(bitD->bitContainer)*8 - 16);
```
terrelln: nit: Use 4 spaces for indentation on zstd (and lines below).

terrelln:
Can you please base your PR off of a newer commit. Your branch is 951 commits behind dev, so our CircleCI tests aren't running.

terrelln:
Since we don't have a gcc-7 compiler in our test suite yet, I've tested the `FALLTHROUGH` macro on godbolt https://godbolt.org/g/bGe6C1.

Cyan4973:
This code paragraph is a [Duff's device](https://en.wikipedia.org/wiki/Duff%27s_device).
It is a very common construction, so if `gcc` wants to flag them, it's going to produce a lot of noises.

According to [this article](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/), the `-Wimplicit-fallthrough` compilation flag is very new, and only works with latest `gcc-7`, which only version I'm aware of is [7.1 from May 2nd 2017](https://gcc.gnu.org/gcc-7/).
It's _very_ recent. As a consequence, it's also hard to test (availability of `gcc-7` is limited right now), and we cannot even target the flag itself as it's unknown in prior version of `gcc`.

According to [gcc documentation](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html), now `-Wextra` automatically enables `-Wimplicit-fallthrough=3`, and that's why this flag ends up being used when compiling `zstd` with `gcc-7`.

Fall-throughs are regularly used in several places within our code base, but apparently only a few of them are being flagged. For example, this one is not :
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3353 .
This suggests that these pass-through are either undetected, or filtered out.

According to the blog article, they are likely filtered out.
Indeed, there is a `/* pass-through */` comment nearby,
which seems to be all it takes to be filtered out.

That would be my preferred solution, rather than adding another target-specific macro for such a restricted use case.

joscollin:
@Cyan4973 I have changed the code to use just the comment /* fall-through */ after removing the FALLTHROUGH macro. But the warnings came back in my environment now. Please suggest. 

This is my gcc version.
```
$ gcc --version
gcc (GCC) 7.0.1 20170309 (Red Hat 7.0.1-0.12)
```

Cyan4973:
We don't have access to this compiler version, so we cannot test your exact configuration.
`gcc 7.0.1` seems to be a development version, which will probably not last long.
This version is supposed to be superceded by `gcc 7.1`, which is an official release.
On `gcc 7.1`, the `/* fall-through */` is expected to work, according to tests on godbolt platform.

> the warnings came back in my environment now. Please suggest

If the question is "how can I compile `zstd` on _my_ platform _without warnings_", 
then it's possible to turn-off this specific warning.
Try : `make zstd MOREFLAGS=-Wno-implicit-fallthrough`

joscollin:
@terrelln @Cyan4973 I have upgraded my gcc to 7.1.1 and tested the new changes. Please review now. Thanks.

Cyan4973:
looks fine !

PR #705, opened 2017-05-20
[libzstd] Allow users to define custom visibility

terrelln:
Useful if users want to hide all of zstd symbols.


terrelln:
The `zstreamtest` failure looks unrelated, but it is unclear if it is a qemu problem or zstd problem.

PR #711, opened 2017-05-25
Fix typo in libzstd.a-mt make rules

ldv-alt:
The macro name is ZSTD_MULTITHREAD, not ZSTD_MULTHREAD.

Fixes: ca6fae78080d ("Add MT enabled targets for libzstd")


Cyan4973:
nice catch ! thanks @ldv-alt

PR #723, opened 2017-06-15
Adding zstd -l

paulcruz74:
- zstd -l filename.zst prints file info
- can use -v for verbose
- afl tested 2 hours with no recognized timeouts or crashes


Cyan4973:
Could you also update the relevant `man` file, at:
https://github.com/facebook/zstd/blob/dev/programs/zstd.1.md

Note : you only need to modify the markdown version. It will be automatically translated into troff format with `make man`.

programs/zstdcli.c
```diff
@@ -672,7 +673,23 @@ int main(int argCount, const char* argv[])
         }
     }
 #endif
-
+    if(operation==zom_list){
+        if(filenameIdx==0){
+            DISPLAY("No files given\n");
+            CLEAN_RETURN(0);
+        }
+        DISPLAY("===========================================\n");
```
Cyan4973: By default, `DISPLAY` writes into `stderr`.
That's fine for warning and error messages.
But when it's an intentional product, it's better to write into `stdout`.

The way to achieve this redirection within `zstdcli`
is to set `g_displayOut = stdout;` before `DISPLAY`.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    unsigned long long compressedSize;
+    int usesCheck;
+} fileInfo_t;
+
+
+int calcFrameHeaderSize(BYTE frameHeaderDescriptor){
+    int frameContentSizeBytes = 0;
+    int windowDescriptorBytes;
+    int dictionaryIDBytes;
+    int frameContentSizeFlag = frameHeaderDescriptor >> 6;
```
Cyan4973: `const` whenever you can

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    int numActualFrames;
+    int numSkippableFrames;
+    unsigned long long decompressedSize;
+    int canComputeDecompSize;
+    unsigned long long compressedSize;
+    int usesCheck;
+} fileInfo_t;
+
+
+int calcFrameHeaderSize(BYTE frameHeaderDescriptor){
```
Cyan4973: `static` if the function is not meant to be called outside of its module.
It reduces number of entry points to protect.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    dictionaryIDBytes = dictionaryIDFlag ? 1 << (dictionaryIDFlag - 1): 0;
+    return 4 + 1 + windowDescriptorBytes + frameContentSizeBytes + dictionaryIDBytes;
+}
+
+/*
+ * Reads information from file, stores in *info
+ * if successful, returns 0, otherwise returns 1
+ */
+int getFileInfo(fileInfo_t* info, const char* inFileName){
+    FILE* srcFile = FIO_openSrcFile(inFileName);
```
Cyan4973: 1) `FILE* const srcFile =` , it gives to reviewer the important signal that `srcFile` is not supposed to be redirected to another file after this initialization.

2) I don't see the corresponding `fclose()`. It should be at the same level as `FIO_openSrcFile()` (in this case, at the end of the function).

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    if(srcFile==NULL){
+        return 1;
+    }
+    info->compressedSize = (unsigned long long)UTIL_getFileSize(inFileName);
+    info->decompressedSize = 0;
+    info->numActualFrames = 0;
+    info-> numSkippableFrames = 0;
+    info->canComputeDecompSize = 1;
+    /* begin analyzing frame */
+    while(1){
```
Cyan4973: Small risk to trigger a warning on Visual because condition is always `true`.
If that's the case, a synonymous expression could be `for ( ; ; )`

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+            }
+        }
+        else if(magicNumber==ZSTD_MAGIC_SKIPPABLE_START){
+            BYTE frameSizeBuffer[4];
+            long frameSize;
+            size_t readBytes = fread(frameSizeBuffer, 1, 4, srcFile);
+            info->numSkippableFrames++;
+            if(readBytes != 4){
+                DISPLAY("There was an error reading skippable frame size");
+                exit(1);
```
Cyan4973: Irrecoverable `exit()`.

In the case of a list of files to `--list`, it would be better if such a case was flagged as an `error`, but the analysis would continue onto next file.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+            }
+            frameSize = MEM_readLE32(frameSizeBuffer);
+            fseek(srcFile, frameSize, SEEK_CUR);
```
Cyan4973: Check `fseek()` return value.
It could fail, for any reason
(one that comes to mind is `frameSize > 2 GB` on a 32-bits system).

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+            if(readBytes != 1){
+                DISPLAY("There was an error with reading frame header descriptor\n");
+                exit(1);
+            }
+            /* calculate actual frame header size */
+            totalFrameHeaderBytes = calcFrameHeaderSize(frameHeaderDescriptor);
+
+            /* reset to beginning of from and read entire header */
+            fseek(srcFile, -5, SEEK_CUR);
+            frameHeader = (BYTE*)malloc(totalFrameHeaderBytes);
```
Cyan4973: Check `frameHeader` value,
could be `NULL` is `malloc()` fails.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+            }
+            frameSize = MEM_readLE32(frameSizeBuffer);
+            fseek(srcFile, frameSize, SEEK_CUR);
+        }
+
+    }
+    return 0;
+}
+void displayInfo(const char* inFileName, fileInfo_t* info, int displayLevel){
+    double compressedSizeMB = (double)info->compressedSize/(1 MB);
```
Cyan4973: `double const`

programs/fileio.c
terrelln: We can probably reuse `ZSTD_frameHeaderSize()`, and expose it in either `zstd_internal.h` or `zstd.h`.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    double decompressedSizeMB = (double)info->decompressedSize/(1 MB);
+
+    if(displayLevel<=2){
```
Cyan4973: minor detail : you may also consider `DISPLAYLEVEL(2,...)`

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        if(info->usesCheck && info->canComputeDecompSize){
+            DISPLAY("Skippable  Non-Skippable  Compressed  Uncompressed  Ratio  Check  Filename\n");
+            DISPLAY("%9d  %13d  %7.2f MB    %7.2f MB  %5.3f  XXH64  %s\n",
+                    info->numSkippableFrames, info->numActualFrames, compressedSizeMB, decompressedSizeMB,
+                    compressedSizeMB/decompressedSizeMB, inFileName);
+        }
+        else if(!info->usesCheck){
+            DISPLAY("Skippable  Non-Skippable  Compressed  Uncompressed  Ratio  Check  Filename\n");
+            DISPLAY("%9d  %13d  %7.2f MB    %7.2f MB  %5.3f  %s\n",
```
Cyan4973: No "space" for "Check". Is it still correctly aligned ?

By the way, consider adding test cases within `playTests.sh` to check your code paths.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                    info->numSkippableFrames, info->numActualFrames, compressedSizeMB, decompressedSizeMB,
+                    compressedSizeMB/decompressedSizeMB, inFileName);
+        }
+        else if(!info->canComputeDecompSize){
+            DISPLAY("Skippable  Non-Skippable  Compressed  Uncompressed  Ratio  Check  Filename\n");
+            DISPLAY("%9d  %13d  %7.2f MB      XXH64  %s\n",
```
Cyan4973: It will probably simplify your branching if you make `XXH64` a `%s` argument.
Something like : 
`const char checkString[] = info->usesCheck ? " XXH64 " : "  None ";`

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    }
+    info->compressedSize = (unsigned long long)UTIL_getFileSize(inFileName);
+    info->decompressedSize = 0;
+    info->numActualFrames = 0;
+    info-> numSkippableFrames = 0;
+    info->canComputeDecompSize = 1;
+    /* begin analyzing frame */
+    while(1){
+        BYTE magicNumberBuffer[4];
+        size_t numBytesRead = fread(magicNumberBuffer, 1, 4, srcFile);
```
terrelln: It might be simpler to unconditionally try to read `ZSTD_FRAMEHEADERSIZE_MAX` bytes.

```c
BYTE headerBuffer[ZSTD_FRAMEHEADERSIZE_MAX];
size_t const numBytesRead = fread(headerBuffer, 1, sizeof(headerBuffer), srcFile);
if (numBytesRead < ZSTD_frameHeaderSize_min) {
    /* handle error */
}
U32 const magicNumber = MEM_readLE32(headerBuffer);

if (magicNumber == ZSTD_MAGICNUMBER) {
    U64 const frameContentSize = ZSTD_getFrameContentSize(headerBuffer, numBytesRead);
    if (frameContentSize == ZSTD_CONTENTSIZE_ERROR || ZSTD_CONTENTSIZE_UNKNOWN) {
        /* handle error */
    }
    size_t const headerSize = ZSTD_frameHeaderSize(headerBuffer, numBytesRead);
    assert(!ZSTD_isError(headerSize));

    int const ret = fseek(srcFile, headerSize, SEEK_SET);
    /* skip the rest of the blocks in the frame */

} else if (magicNumber == ZSTD_MAGIC_SKIPPABLE_START) {
    /* Skippable frames */
}
```

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    }
+    else{
+        DISPLAY("# Zstandard Frames: %d\n", info->numActualFrames);
+        DISPLAY("# Skippable Frames: %d\n", info->numSkippableFrames);
+        DISPLAY("Compressed Size: %.2f MB (%llu B)\n", compressedSizeMB, info->compressedSize);
+        if(info->canComputeDecompSize){
+            DISPLAY("Decompressed Size: %.2f MB (%llu B)\n", decompressedSizeMB, info->decompressedSize);
+            DISPLAY("Ratio: %.4f\n", compressedSizeMB/decompressedSizeMB);
+        }
+        if(info->usesCheck){
```
Cyan4973: minor : code style preference : I'd prefer `if (info->usesCheck) {`.
Adding spaces for clarity.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+            DISPLAY("Check: XXH64\n");
+        }
+    }
+
+}
+
+int FIO_listFile(const char* inFileName, int displayLevel){
+    const char* const suffixPtr = strrchr(inFileName, '.');
+    DISPLAY("File: %s\n", inFileName);
+    if(!suffixPtr || strcmp(suffixPtr, ZSTD_EXTENSION)){
```
Cyan4973: `xz` doesn't limit itself to `.xz` suffix extension files.
I don't see a reason to do it.
We should be able to `zstd -l tmp`, and get a correct answer if it's `zstd` compressed.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        DISPLAYLEVEL(1, "file %s was not compressed with zstd -- ignoring\n\n", inFileName);
+        return 1;
+    }
+    else{
+        fileInfo_t* info = (fileInfo_t*)malloc(sizeof(fileInfo_t));
```
Cyan4973: check `malloc()` result, and then `free()`,
__or__, (better in this case)
statically allocated `fileInfo_t` on stack.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        int error = getFileInfo(info, inFileName);
```
Cyan4973: `int const`

programs/fileio.c
terrelln: You need to `free()` anything you `malloc()`.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        if(error==1){
+            DISPLAY("An error occurred with getting file info\n");
+            exit(1);
```
Cyan4973: Better signal an error, but continue processing.
Ex : 
```
xz -l LICENSE  xxhsum.c.xz
xz: LICENSE: File format not recognized
Strms  Blocks   Compressed Uncompressed  Ratio  Check   Filename
    1       1       9988 B     38.8 KiB  0.251  CRC64   xxhsum.c.xz
echo $?
1
```

The program end result is still an error if one file fails.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        }
+        displayInfo(inFileName, info, displayLevel);
+    }
+    DISPLAY("\n");
+    return 0;
```
Cyan4973: If the program result depends on having all files successfully listed,
it must be a variable.

programs/fileio.c
```diff
@@ -862,6 +862,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+
+
+int calcFrameHeaderSize(BYTE frameHeaderDescriptor){
+    int frameContentSizeBytes = 0;
+    int windowDescriptorBytes;
+    int dictionaryIDBytes;
+    int frameContentSizeFlag = frameHeaderDescriptor >> 6;
+    int singleSegmentFlag = (frameHeaderDescriptor & (1 << 5)) >> 5;
+    int dictionaryIDFlag = frameHeaderDescriptor & 3;
+    if(frameContentSizeFlag!=0){
```
terrelln: Style nit, this should be

```c
if (frameContentSizeFlag != 0) {
```

here and elsewhere.

lib/zstd.h
```diff
@@ -461,6 +461,11 @@ ZSTDLIB_API unsigned long long ZSTD_getFrameContentSize(const void *src, size_t
  *            however it does mean that all frame data must be present and valid. */
 ZSTDLIB_API unsigned long long ZSTD_findDecompressedSize(const void* src, size_t srcSize);
 
+/*! ZSTD_frameHeaderSize() :
+*   `src` should point to the start of a ZSTD frame
+*   `srcSize` must be >= ZSTD_frameHeaderSize_prefix.
```
Cyan4973: I suspect `ZSTD_frameHeaderSize_prefix` is not exposed.
So, either expose it, or translate it.

programs/fileio.c
```diff
@@ -862,6 +863,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    info->decompressedSize = 0;
+    info->numActualFrames = 0;
+    info->numSkippableFrames = 0;
+    info->canComputeDecompSize = 1;
+    /* begin analyzing frame */
+    for( ; ; ){
+        BYTE headerBuffer[ZSTD_FRAMEHEADERSIZE_MAX];
+        size_t const numBytesRead = fread(headerBuffer, 1, sizeof(headerBuffer), srcFile);
+        if (numBytesRead < ZSTD_frameHeaderSize_min) {
+            if(feof(srcFile)){
```
Cyan4973: This properly covers `numBytesRead==0`,
but what if `numBytesRead>0` ?

programs/fileio.c
```diff
@@ -862,6 +863,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                            detectError = 1;
+                            break;
+                        }
+                    }
+                }
+
+                /* skip the rest of the blocks in the frame */
+                {
+                    int lastBlock = 0;
+                    size_t readBytes = 0;
```
Cyan4973: `readBytes` is only necessary inside the `do {` loop,
so better declare it here.

programs/fileio.c
```diff
@@ -862,6 +863,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                            detectError = 1;
+                            break;
+                        }
+                    }
+                }
+                info->numActualFrames++;
+            }
+            else if (magicNumber == ZSTD_MAGIC_SKIPPABLE_START) {
+                BYTE frameSizeBuffer[4];
+                size_t readBytes = fread(frameSizeBuffer, 1, 4, srcFile);
```
Cyan4973: `const`

programs/fileio.c
```diff
@@ -862,6 +863,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                if (readBytes != 4) {
+                    DISPLAY("There was an error reading skippable frame size");
+                    detectError = 1;
+                    break;
+                }
+                {
+                    long const frameSize = MEM_readLE32(frameSizeBuffer);
```
Cyan4973: prefer an `U32` type, which is the return type of `MEM_readLE32()`.
`long` is signed, which will spell troubles if value returned is >= 2 GB.

programs/fileio.c
```diff
@@ -862,6 +863,193 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                    int const ret = fseek(srcFile, frameSize, SEEK_CUR);
```
Cyan4973: A consequence is that you have to pay attention when using `U32 frameSize` with `fseek()`, since `fseek()` requires a `long` argument.

programs/zstdcli.c
paulcruz74: does `g_displayOut` need to be set each time that `DISPLAY()` is called? I set it equal to `stdout` at the beginning of the block where `zom_list` is detected.

Cyan4973: Once is enough. It's a toggle.

programs/fileio.c
paulcruz74: changed `long` to `U32`, but what kind of errors should I look out for regarding the `fseek()`?

Cyan4973: http://www.cplusplus.com/reference/cstdio/fseek/

> Return Value
> If successful, the function returns zero.
> Otherwise, it returns non-zero value.
> If a read or write error occurs, the error indicator (ferror) is set.

Basically, if it's `!=0`, then something bad happened.
You don't necessarily need something more accurate.

programs/fileio.c
```diff
@@ -862,6 +863,205 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        }
+        {
+            U32 const magicNumber = MEM_readLE32(headerBuffer);
+            if (magicNumber == ZSTD_MAGICNUMBER) {
+                U64 const frameContentSize = ZSTD_getFrameContentSize(headerBuffer, numBytesRead);
+                if (frameContentSize == ZSTD_CONTENTSIZE_ERROR || frameContentSize == ZSTD_CONTENTSIZE_UNKNOWN) {
+                    info->canComputeDecompSize = 0;
+                }
+                else {
+                    info->decompressedSize += frameContentSize;
```
Cyan4973: `canComputeDecompSize` is set to `0` in the other branch.
I feel it would be better to set to `1` here, rather than relying on some long-distance initialisation.

programs/fileio.c
```diff
@@ -862,6 +863,205 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        }
+    }
+
+    DISPLAYOUT("\n");
+}
+
+int FIO_listFile(const char* inFileName, int displayLevel){
+    fileInfo_t info;
+
+    /* initialize info to avoid warnings */
```
Cyan4973: I would recommend to just `memset()` the structure.

As a general rule, try to simplify initialisation to a simple `memset(,0,)`.
It makes much simpler to follow for an external reviewer.
A more complex scheme will require to keep in mind the state of the structure, and track its changes throughout its lifetime. It's more complex, hence lend itself to errors, especially at later stages (when the code is modified again, eventually by someone else).

programs/zstdcli.c
```diff
@@ -672,7 +675,25 @@ int main(int argCount, const char* argv[])
...
+        DISPLAY("Printing information about compressed files\n");
+        DISPLAY("===========================================\n");
+        DISPLAY("Number of files listed: %d\n", filenameIdx);
+        {
+            int error = 0;
+            unsigned u;
+            for(u=0; u<filenameIdx;u++){
+                error = FIO_listFile(filenameTable[u],g_displayLevel);
```
Cyan4973: This will give you the last `error` of the last file.
If you want a non-zero value if "any" error happen, prefer `error +=` or `error |=`.

programs/fileio.c
```diff
@@ -862,6 +863,205 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    info.usesCheck = 0;
+    DISPLAYOUT("File: %s\n", inFileName);
+    {
+        int const error = getFileInfo(&info, inFileName);
+        if (error == 1) {
+            /* display error, but provide output */
+            DISPLAY("An error occurred with getting file info\n");
+        }
+        else if (error == 2) {
+            DISPLAYOUT("File %s not compressed with zstd\n\n", inFileName);
```
paulcruz74: At the moment, --list prints a space between consecutive files when outputting information, so the double `\n\n` is so that when a file is not compressed with zstd it still prints a line between that file and the next. Should I just get rid of the line in between files?

Cyan4973: `xz` doesn't seem to output a line separation between files : 

```
xz -l *
xz: CONTRIBUTING.md: File format not recognized
xz: LICENSE: File format not recognized
xz: LICENSE-examples: File format not recognized
xz: Makefile: File format not recognized
```

programs/zstdcli.c
```diff
@@ -672,7 +675,25 @@ int main(int argCount, const char* argv[])
...
+        g_displayOut = stdout;
+        if(filenameIdx==0){
+            DISPLAY("No files given\n");
+            CLEAN_RETURN(0);
+        }
+        DISPLAY("===========================================\n");
+        DISPLAY("Printing information about compressed files\n");
+        DISPLAY("===========================================\n");
+        DISPLAY("Number of files listed: %d\n", filenameIdx);
+        {
```
Cyan4973: As software architecture guideline,
try to limit the amount of "intelligent" work done inside `zstdcli.c`.
`zstdcli` should be just about managing input, and redirecting them to appropriate module.

Here, you could have a function `FIO_listMultipleFiles()` that would do the loop and printing inside `fileio.c`.
It's not much, but it puts the workload in the expected location.

tests/playTests.sh
```diff
@@ -537,6 +537,29 @@ fi
...
+./datagen > tmp1
+./datagen > tmp2
+./datagen > tmp3
+./datagen > tmp4
+$ZSTD tmp*
+$ZSTD -l *.zst
+$ZSTD -lv *.zst
+$ZSTD --list *.zst
+$ZSTD --list -v *.zst
+
```
Cyan4973: Add a test that check the detected error regarding return code on mixed `zst` and `non-zst` input.
Typically something like : 
`!$ZSTD -l tmp tmp.zst`.

The test should fail with older code `error = FIO_listFile()`
and work with `error |= FIO_listFile()`.

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+        }
+    }
+    fclose(srcFile);
+    return detectError;
+}
+
+static void displayInfo(const char* inFileName, fileInfo_t* info, int displayLevel){
+    double const compressedSizeMB = (double)info->compressedSize/(1 MB);
+    double const decompressedSizeMB = (double)info->decompressedSize/(1 MB);
+    const char* checkString = (info->usesCheck ? "XXH64" : "None");
```
Cyan4973: `const char* const`
`checkString` is not re-assigned after this assignment.

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    if (displayLevel <= 2) {
+        if (!info->decompUnavailable) {
+            double const ratio = (info->decompressedSize == 0) ? 0.0 : compressedSizeMB/decompressedSizeMB;
```
Cyan4973: Add a test case in `playTests` for null-size compressed file.

programs/fileio.c
Cyan4973: Correction : it depends on display level.
There is a one-line separation between files in `--verbose` mode : 

```
xz -l -v *
CONTRIBUTING.md (1/22)
xz: CONTRIBUTING.md: File format not recognized

LICENSE (2/22)
xz: LICENSE: File format not recognized

LICENSE-examples (3/22)
xz: LICENSE-examples: File format not recognized

Makefile (4/22)
xz: Makefile: File format not recognized
```

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+
+                /* skip the rest of the blocks in the frame */
+                {
+                    int lastBlock = 0;
+                    do {
+                        BYTE blockHeaderBuffer[3];
+                        U32 blockHeader;
+                        int blockSize;
+                        size_t readBytes = fread(blockHeaderBuffer, 1, 3, srcFile);
+                        if (readBytes != 3) {
```
Cyan4973: `size_t const` ?

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                }
+
+                /* skip the rest of the blocks in the frame */
+                {
+                    int lastBlock = 0;
+                    do {
+                        BYTE blockHeaderBuffer[3];
+                        U32 blockHeader;
+                        int blockSize;
+                        size_t readBytes = fread(blockHeaderBuffer, 1, 3, srcFile);
```
Cyan4973: `size_t const` ?

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                        if (readBytes != 3) {
+                            DISPLAY("There was a problem reading the block header\n");
+                            detectError = 1;
+                            break;
+                        }
+                        blockHeader = MEM_readLE24(blockHeaderBuffer);
+                        lastBlock = blockHeader & 1;
+                        blockSize = (blockHeader - (blockHeader & 7)) >> 3;
```
Cyan4973: minor detail : 
This formulation works,
but I believe the `- (blockHeader & 7)` part is not useful,
as this part of data will be automatically rounded down to zero by `>> 3`.

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                    }
+                }
+                {
+                    /* check if checksum is used */
+                    BYTE const frameHeaderDescriptor = headerBuffer[4];
+                    int const contentChecksumFlag = (frameHeaderDescriptor & (1 << 2)) >> 2;
+                    if (contentChecksumFlag) {
+                        info->usesCheck = 1;
+                    }
+                    if (contentChecksumFlag) {
```
Cyan4973: minor : same `if` condition, could be a single branch ?

programs/fileio.c
```diff
@@ -862,6 +863,199 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+                BYTE frameSizeBuffer[4];
+                size_t const readBytes = fread(frameSizeBuffer, 1, 4, srcFile);
+                if (readBytes != 4) {
+                    DISPLAY("There was an error reading skippable frame size");
+                    detectError = 1;
+                    break;
+                }
+                {
+                    U32 const frameSize = MEM_readLE32(frameSizeBuffer);
+                    int const ret = fseek(srcFile, (long)frameSize, SEEK_CUR);
```
Cyan4973: What happens if `frameSize >= 2 GB` ?
`(long) frameSize` will likely be converted to a negative number,
so the outcome of `fseek()` will not be the expected one.

You should check this condition.

Cyan4973: Additional notes : 
- Consider usage of `LONG_SEEK`.
   I don't like it much, but it may be the only way to work properly on large files (> 4GB) on 32-bits systems
- Consider adding a "large file" (> 4 GB) test.

paulcruz74: are there performance costs with `LONG_SEEK`? should I check if it's needed before hand (`(long)frameSize < 0`) or just default to `LONG_SEEK` in this case?

tests/playTests.sh
```diff
@@ -537,6 +537,43 @@ fi
...
+$ECHO "\n**** zstd --list/-l error detection tests ****"
+! $ZSTD -l tmp1 tmp1.zst
+! $ZSTD --list tmp*
+! $ZSTD -lv tmp1*
+! $ZSTD --list -v tmp2 tmp23.zst
+
+$ECHO "\n**** zstd --list/-l test with null files ****"
+./datagen -g0 > tmp5
+$ZSTD tmp5
+! $ZSTD -l tmp5*
```
Cyan4973: you also need a test for NULL-content (but valid!) frames.

programs/fileio.c
```diff
@@ -862,6 +863,222 @@ int FIO_compressFilename(const char* dstFileName, const char* srcFileName,
...
+    return detectError;
+}
+
+static void displayInfo(const char* inFileName, fileInfo_t* info, int displayLevel){
+    double const compressedSizeMB = (double)info->compressedSize/(1 MB);
+    double const decompressedSizeMB = (double)info->decompressedSize/(1 MB);
+    const char* const checkString = (info->usesCheck ? "XXH64" : "None");
+    if (displayLevel <= 2) {
+        if (!info->decompUnavailable) {
+            double const ratio = (info->decompressedSize == 0) ? 0.0 : compressedSizeMB/decompressedSizeMB;
```
Cyan4973: That line is the reason why I believe it is useful to have a NULL-content frame test.

Cyan4973:
Almost there !
One last test to implement,
then we'll let run CI,
and then merge  🏁

PR #724, opened 2017-06-17
INSTALL_DATA instead of INSTALL_LIB for libzstd.a

yurivict:
INSTALL_LIB can be passed -s flag to strip symbols. Static libraries should not be stripped, only dynamic ones should be stripped.


Cyan4973:
Good point !

yurivict:
Thanks for merging.

This is the problem why static build affecting tor can fail that I reported earlier.

Are you able to make a release with this so that the tor package can use zstd?

Cyan4973:
There will be a new release probably early July.
It will include this fix.

PR #725, opened 2017-06-19
New Advanced API

Cyan4973:
In this proposal, parameters are pushed one by one into an existing CCtx,
   and then applied on all subsequent compression jobs.
   When no parameter is ever provided, CCtx is created with compression level ZSTD_CLEVEL_DEFAULT.

   This API is intended to replace all others experimental API.
   It can basically do all other use cases, and even new ones.
   It stands a good chance to become "stable",
   after a reasonable testing period.


lib/zstd.h
```diff
@@ -412,6 +421,9 @@ typedef struct {
 typedef void* (*ZSTD_allocFunction) (void* opaque, size_t size);
 typedef void  (*ZSTD_freeFunction) (void* opaque, void* address);
 typedef struct { ZSTD_allocFunction customAlloc; ZSTD_freeFunction customFree; void* opaque; } ZSTD_customMem;
+/* use this constant to defer to stdlib's functions */
+static const ZSTD_customMem ZSTD_defaultCMem = { NULL, NULL, NULL};
```
terrelln: nit: Space before `}`.

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+ */
+
+typedef enum {
+    /* compression parameters */
+    ZSTD_p_compressionLevel=100, /* Update all compression parameters according to pre-defined cLevel table
+                              * Default level is ZSTD_CLEVEL_DEFAULT==3.
+                              * Special: value 0 means "do not change cLevel". */
+    ZSTD_p_windowLog,        /* Maximum allowed back-reference distance, expressed as power of 2.
+                              * Must be clamped between ZSTD_WINDOWLOG_MIN and ZSTD_WINDOWLOG_MAX.
+                              * default value : set through compressionLevel.
```
terrelln: Delete the default value line, since none of the other compression parameters mention it.

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+                             /* Not ready yet ! */
+#endif
+
+    /* frame parameters */
+    ZSTD_p_contentSizeFlag=200, /* Content size is written into frame header _whenever known_ (default:1) */
+    ZSTD_p_checksumFlag,     /* A 32-bits checksum of content is written at end of frame (default:0) */
+    ZSTD_p_dictIDFlag,       /* When applicable, dictID of dictionary is provided in frame header (default:1) */
+
+    /* dictionary parameters */
+    ZSTD_p_refDictContent=300, /* Content of dictionary content will be referenced, instead of copied (default:0).
```
terrelln: Can we `#if 0` all the parameters that aren't ready yet, so the comments don't get out of date.

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+                              * This avoids duplicating dictionary content.
+                              * But it also requires that dictionary buffer outlives its user (CDict) */
+                             /* Not ready yet ! */
+    ZSTD_p_rawContentDict,   /* load dictionary in "content-only" mode (no header analysis) (default:0) */
+                             /* question : should there be an option to load dictionary only in zstd format, rejecting others with an error code ? */
```
terrelln: Yes, I think so.

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+ *  Reference a prefix (content-only dictionary) to bootstrap next compression job.
+ *  Decompression will have to use same prefix.
+ *  Prefix is only used once. Tables are discarded at end of compression job.
+ *  If there is a need to use same prefix multiple times, consider embedding it into a ZSTD_CDict.
+ * @result : 0, or an error code (which can be tested with ZSTD_isError()).
+ *  Special : Adding a NULL (or 0-size) dictionary invalidates any previous prefix, meaning "return to no-dictionary mode".
+ *  Note 1 : Prefix buffer is referenced. It must outlive compression job.
+ *  Note 2 : Referencing a prefix involves building tables, which are dependent on compression parameters.
+ *           It's a CPU-heavy operation, with non-negligible impact on latency. */
+ZSTDLIB_API size_t ZSTD_CCtx_refPrefix(ZSTD_CCtx* cctx, const void* prefix, size_t prefixSize);   /* Not ready yet ! */
```
terrelln: Can we `#if 0` it out?

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+ *           Note that ZSTD_CONTENTSIZE_UNKNOWN is default value for new compression jobs.
+ *  Note 2 : If all data is provided and consumed in a single round,
+ *           this value is overriden by srcSize instead. */
+ZSTDLIB_API size_t ZSTD_CCtx_setPledgedSrcSize(ZSTD_CCtx* cctx, unsigned long long pledgedSrcSize);
+
+/*! ZSTD_CCtx_loadDictionary() :
+ *  Create an internal CDict from dict buffer.
+ *  Decompression will have to use same buffer.
+ * @result : 0, or an error code (which can be tested with ZSTD_isError()).
+ *  Special : Adding a NULL (or 0-size) dictionary invalidates any previous prefix,
```
terrelln: Should it be "dictionary or prefix"?

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+    ZSTD_e_continue=0, /* collect more data, encoder transparently decides when to output result, for optimal conditions */
+    ZSTD_e_flush,      /* flush any data provided so far - frame will continue, future data can still reference previous data for better compression */
+    ZSTD_e_end         /* flush any remaining data and ends current frame. Any future compression starts a new frame. */
+} ZSTD_EndDirective;
+
+/*! ZSTD_compressStream_generic() :
+ *  Behave about the same as ZSTD_compressStream. To note :
+ *  - Compression parameters are pushed into CCtx before starting compression, using ZSTD_setCCtxParameter()
+ *  - Compression parameters cannot be changed once compression is started.
+ *  - *dstPos must be <= dstCapacity, *srcPos must be <= srcSize
```
terrelln: These comments are out of date.

lib/zstd.h
```diff
@@ -765,6 +849,239 @@ ZSTDLIB_API size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t ds
...
+ *  which have troubles handling structures containing memory pointers.
+ */
+size_t ZSTD_compress_generic_simpleArgs (
+                            ZSTD_CCtx* cctx,
+                            void* dst, size_t dstCapacity, size_t* dstPos,
+                      const void* src, size_t srcSize, size_t* srcPos,
+                            ZSTD_EndDirective endOp);
+
+
+/* note : this part if completely experimental, not ready yet.
```
terrelln: `#if 0`

lib/compress/zstdmt_compress.h
```diff
@@ -15,25 +15,32 @@
...
+/* ===   Memory management   === */
 typedef struct ZSTDMT_CCtx_s ZSTDMT_CCtx;
 ZSTDLIB_API ZSTDMT_CCtx* ZSTDMT_createCCtx(unsigned nbThreads);
-ZSTDLIB_API size_t ZSTDMT_freeCCtx(ZSTDMT_CCtx* cctx);
+ZSTDLIB_API ZSTDMT_CCtx* ZSTDMT_createCCtx_advanced(unsigned nbThreads,
+                                                    ZSTD_customMem cMem);
+ZSTDLIB_API size_t ZSTDMT_freeCCtx(ZSTDMT_CCtx* mtctx);
+
+ZSTDLIB_API size_t ZSTDMT_sizeof_CCtx(ZSTDMT_CCtx* mtctx);
+ZSTDLIB_API size_t ZSTDMT_estimateCCtxSize(ZSTD_compressionParameters cParams, unsigned nbThreads);   /* not ready yet */
```
Cyan4973: nope, no longer 😛

Cyan4973: Yes, it is, I forgot about this one.

The initial intention was to make it compatible with `ZSTD_initStaticCCtx()`.
But since then, I've removed that possibility : `ZSTD_initStaticCCtx()` only works for single-threaded (for now).

I might as well remove the prototype, since it's confusing.

programs/bench.c
```diff
@@ -262,42 +260,72 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
...
-                    ZSTD_CDict* cdict;
+                    ZSTD_customMem const cmem = { NULL, NULL, NULL };
                     if (comprParams->windowLog) zparams.cParams.windowLog = comprParams->windowLog;
                     if (comprParams->chainLog) zparams.cParams.chainLog = comprParams->chainLog;
                     if (comprParams->hashLog) zparams.cParams.hashLog = comprParams->hashLog;
                     if (comprParams->searchLog) zparams.cParams.searchLog = comprParams->searchLog;
                     if (comprParams->searchLength) zparams.cParams.searchLength = comprParams->searchLength;
                     if (comprParams->targetLength) zparams.cParams.targetLength = comprParams->targetLength;
-                    if (comprParams->strategy) zparams.cParams.strategy = (ZSTD_strategy)(comprParams->strategy - 1);
+                    if (comprParams->strategy) zparams.cParams.strategy = comprParams->strategy;
```
Cyan4973: `ZSTD_strategy` enum type has been slightly changed, 
with `ZSTD_fast`, the first strategy, not assigned the value `=1` (it used to be `=0`).
It makes it possible to keep value `0` to mean "do not change strategy".

This order used to be simulated by manually adding, and then removing (line above) `1` to `ZSTD_strategy` value (added by Przemyslaw when creating the advanced compression arguments command line).

lib/zstd.h
Cyan4973: I intend to have this parameter enabled before the release.
It's associated with `ZSTD_CCtx_refPrefix()`, which I also consider to enable.

lib/zstd.h
Cyan4973: I guess the better way to give this instruction
is to create an enum,
so that we can't have both "rawContent" mode and "fullDict" mode enabled at the same.

Which means I need to change this parameter name, and create a new `enum`.

lib/zstd.h
Cyan4973: Good catch, it must be `dictionary`.

`prefix` is a pretty new concept which I consider using for the limited "single-usage content-only dictionary" use case.

lib/zstd.h
Cyan4973: Same as above, I intend to enable it before the release.

lib/zstd.h
Cyan4973: I think I will simply throw this part away.
I won't implement it for next release,
and I believe next work on immutable dictionary compression
will make these functions irrelevant.

lib/compress/zstd_compress.c
```diff
@@ -3000,11 +3334,39 @@ static ZSTD_parameters ZSTD_makeParams(ZSTD_compressionParameters cParams, ZSTD_
...
+                    ZSTD_CDict* cdict,
+              const void* dictBuffer, size_t dictSize, unsigned byReference,
+                    ZSTD_compressionParameters cParams)
+{
+    if ((byReference) || (!dictBuffer) || (!dictSize)) {
+        cdict->dictBuffer = NULL;
+        cdict->dictContent = dictBuffer;
+    } else {
+        void* const internalBuffer = ZSTD_malloc(dictSize, cdict->refContext->customMem);
+        if (!internalBuffer) return ERROR(memory_allocation);
```
terrelln: If this condition happens, `cdict->dictBuffer` is uninitialized, so in `ZSTD_freeCDict()` we will `free()` an uninitialized pointer.

lib/compress/zstd_compress.c
```diff
@@ -3000,11 +3334,39 @@ static ZSTD_parameters ZSTD_makeParams(ZSTD_compressionParameters cParams, ZSTD_
...
 ZSTD_CDict* ZSTD_createCDict_advanced(const void* dictBuffer, size_t dictSize, unsigned byReference,
                                       ZSTD_compressionParameters cParams, ZSTD_customMem customMem)
 {
-    if (!customMem.customAlloc && !customMem.customFree) customMem = defaultCustomMem;
-    if (!customMem.customAlloc || !customMem.customFree) return NULL;
+    DEBUGLOG(5, "ZSTD_createCDict_advanced");
+    if (!customMem.customAlloc ^ !customMem.customFree) return NULL;
 
     {   ZSTD_CDict* const cdict = (ZSTD_CDict*) ZSTD_malloc(sizeof(ZSTD_CDict), customMem);
         ZSTD_CCtx* const cctx = ZSTD_createCCtx_advanced(customMem);
```
terrelln: I think we have the same problem here if `cdict && !cctx`, I think using `ZSTD_calloc()` would fix both problems.

lib/compress/zstd_compress.c
```diff
@@ -3144,251 +3532,260 @@ size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
...
-        zcs->outBuffSize = 0;
-        ZSTD_free(zcs->outBuff, zcs->customMem);
-        zcs->outBuff = (char*)ZSTD_malloc(outBuffSize, zcs->customMem);
-        if (zcs->outBuff == NULL) return ERROR(memory_allocation);
-        zcs->outBuffSize = outBuffSize;
+        ZSTD_freeCDict(zcs->cdictLocal);
+        zcs->cdict = NULL;
+        zcs->cdictLocal = ZSTD_createCDict_advanced(dict, dictSize, 0 /* copy */, params.cParams, zcs->customMem);
+        if (zcs->cdictLocal == NULL) return ERROR(memory_allocation);
+        zcs->cdict = zcs->cdictLocal;
```
terrelln: Can you move this line above the `NULL` check, and delete `zcs->cdict = NULL;` to be consistent with other CDict creation constructs?

lib/compress/zstd_compress.c
```diff
@@ -3144,251 +3532,260 @@ size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
...
+    } else {
+        if (cdict) {
```
terrelln: Here we probably want to `ZSTD_freeCDict(zcs->cdictLocal)`. I don't imagine it is common, but if someone compresses using a `dict`, and then a `cdict`, it looks like we will keep the `cdictLocal` around until `ZSTD_free()` or another compression using a `dict`, when we don't need it.

Does the same thing happen for non-streaming mode?

lib/compress/zstd_compress.c
```diff
@@ -3144,251 +3532,260 @@ size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
...
 
-typedef enum { zsf_gather, zsf_flush, zsf_end } ZSTD_flush_e;
-
-MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src, size_t srcSize)
+MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity,
+                           const void* src, size_t srcSize)
 {
     size_t const length = MIN(dstCapacity, srcSize);
-    memcpy(dst, src, length);
+    if (length) memcpy(dst, src, length);
```
terrelln: Nice

lib/compress/zstd_compress.c
```diff
@@ -3144,251 +3532,260 @@ size_t ZSTD_freeCStream(ZSTD_CStream* zcs)
...
+                    DEBUGLOG(5, "Frame completed on flush");
+                    someMoreWork = 0;
+                    ZSTD_startNewCompression(zcs);
+                    break;
+                }
                 zcs->streamStage = zcss_load;
                 break;
             }
 
         case zcss_final:
```
terrelln: This case isn't used anymore, right?

lib/compress/zstd_compress.c
```diff
@@ -3584,30 +4014,24 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
...
     if (compressionLevel > ZSTD_MAX_CLEVEL) compressionLevel = ZSTD_MAX_CLEVEL;
     cp = ZSTD_defaultCParameters[tableID][compressionLevel];
-    if (MEM_32bits()) {   /* auto-correction, for 32-bits mode */
-        if (cp.windowLog > ZSTD_WINDOWLOG_MAX) cp.windowLog = ZSTD_WINDOWLOG_MAX;
-        if (cp.chainLog > ZSTD_CHAINLOG_MAX) cp.chainLog = ZSTD_CHAINLOG_MAX;
-        if (cp.hashLog > ZSTD_HASHLOG_MAX) cp.hashLog = ZSTD_HASHLOG_MAX;
-    }
-    cp = ZSTD_adjustCParams(cp, srcSize, dictSize);
-    return cp;
+    return ZSTD_adjustCParams_internal(cp, srcSizeHint, dictSize);   /* no need to ensure initial CParams validity */
```
terrelln: Does this still work for 32-bit mode?

lib/compress/zstd_compress.c
```diff
@@ -275,23 +546,43 @@ size_t ZSTD_estimateCCtxSize(ZSTD_compressionParameters cParams)
     return sizeof(ZSTD_CCtx) + neededSpace;
 }
 
+size_t ZSTD_estimateCStreamSize(ZSTD_compressionParameters cParams)
+{
+    size_t const CCtxSize = ZSTD_estimateCCtxSize(cParams);
+    size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << cParams.windowLog);
+    size_t const inBuffSize = ((size_t)1 << cParams.windowLog) + blockSize;
+    size_t const outBuffSize = ZSTD_compressBound(blockSize) + 1;
```
Cyan4973: No idea.
I just copied it "as is" from current `ZSTD_initCStream()` :
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3220

This could be to ensure some kind of `<` condition.

And maybe the extra byte is actually superfluous.

lib/compress/zstd_compress.c
Cyan4973: nice catch.
I will move initialisation before the test.
I'm also considering changing some variable names, because it feels somewhat confusing...

lib/compress/zstd_compress.c
Cyan4973: No, I don't think so.
In this case, we just `free(cdict)`.
We do not dereference nor read its content.

lib/compress/zstd_compress.c
Cyan4973: Yes, you are right.
Internal `CDict` is kept until its replacement, or final removal of host `CCtx`.
Indeed, it means memory is still reserved, so it seems better to free it before the final `ZSTD_freeCCtx()` (at least in theory).

lib/compress/zstd_compress.c
Cyan4973: Right. On reaching the end, the stage is automatically set to `zcss_init`, so `zcss_final` is no longer used.
Can be removed.

lib/compress/zstd_compress.c
terrelln: Oh yeah, I mistook it for a `ZSTD_createCDict()`, but that wouldn't make much sense.

lib/compress/zstd_compress.c
Cyan4973: Since Sean introduced "long range matches" for 32-bits mode,
there is no more any difference in `windowLog` support between 32-bits vs 64-bits,
https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L356
hence no need for a second pass adjust for 32-bits.

lib/zstd.h
```diff
@@ -1018,17 +1017,17 @@ typedef enum {
     ZSTD_e_end         /* flush any remaining data and ends current frame. Any future compression starts a new frame. */
 } ZSTD_EndDirective;
 
-/*! ZSTD_compressStream_generic() :
+/*! ZSTD_compress_generic() :
  *  Behave about the same as ZSTD_compressStream. To note :
- *  - Compression parameters are pushed into CCtx before starting compression, using ZSTD_setCCtxParameter()
+ *  - Compression parameters are pushed into CCtx before starting compression, using ZSTD_CCtx_setParameter()
  *  - Compression parameters cannot be changed once compression is started.
  *  - *dstPos must be <= dstCapacity, *srcPos must be <= srcSize
```
terrelln: It should be `input` and `output`, not `dstPos` and `srcPos`.

lib/compress/zstdmt_compress.c
```diff
@@ -505,33 +559,46 @@ static void ZSTDMT_waitForAllJobsCompleted(ZSTDMT_CCtx* zcs) {
...
     zcs->frameContentSize = pledgedSrcSize;
+    if (dict) {
+        ZSTD_freeCDict(zcs->cdictLocal);
+        zcs->cdict = NULL;
+        zcs->cdictLocal = ZSTD_createCDict_advanced(dict, dictSize, 0 /* byRef */,
+                                                params.cParams, zcs->cMem);
+        if (zcs->cdictLocal == NULL) return ERROR(memory_allocation);
+        zcs->cdict = zcs->cdictLocal;
+    } else {
+        zcs->cdict = cdict;
```
terrelln: We should probably free `zcs->cdictLocal` here as well.

lib/compress/zstdmt_compress.c
```diff
@@ -546,24 +613,38 @@ static size_t ZSTDMT_initCStream_internal(ZSTDMT_CCtx* zcs,
...
+    return ZSTDMT_initCStream_internal(mtctx, dict, dictSize, NULL, params, pledgedSrcSize);
 }
 
+size_t ZSTDMT_initCStream_usingCDict(ZSTDMT_CCtx* mtctx,
+                               const ZSTD_CDict* cdict,
+                                     ZSTD_frameParameters fParams,
+                                     unsigned long long pledgedSrcSize)
+{
+    ZSTD_parameters params = ZSTD_getParamsFromCDict(cdict);
+    if (cdict==NULL) return ERROR(GENERIC);   /* method incompatible with NULL cdict */
```
terrelln: `ZSTD_initCStream_usingCDict()` returns `ERROR(dictionary_wrong)` in this case.

lib/compress/zstdmt_compress.c
```diff
@@ -699,23 +795,29 @@ static size_t ZSTDMT_flushNextJob(ZSTDMT_CCtx* zcs, ZSTD_outBuffer* output, unsi
...
+        return ERROR(stage_wrong);
+    }
+    if (zcs->nbThreads==1) {
+        return ZSTD_compressStream(zcs->cctxPool->cctx[0], output, input);
+    }
 
     /* fill input buffer */
     {   size_t const toLoad = MIN(input->size - input->pos, zcs->inBuffSize - zcs->inBuff.filled);
-        memcpy((char*)zcs->inBuff.buffer.start + zcs->inBuff.filled, input->src, toLoad);
+        memcpy((char*)zcs->inBuff.buffer.start + zcs->inBuff.filled, (const char*)input->src + input->pos, toLoad);
```
terrelln: We should check that `toLoad > 0` here.

lib/compress/zstdmt_compress.c
```diff
@@ -611,22 +696,33 @@ static size_t ZSTDMT_createCompressionJob(ZSTDMT_CCtx* zcs, size_t srcSize, unsi
...
         }
-        DEBUGLOG(5, "inBuff filled to %u", (U32)zcs->inBuff.filled);
+        DEBUGLOG(5, "inBuff currently filled to %u", (U32)zcs->inBuff.filled);
         zcs->inBuff.filled -= srcSize + zcs->dictSize - newDictSize;
-        DEBUGLOG(5, "new job : filled to %u, with %u dict and %u src", (U32)zcs->inBuff.filled, (U32)newDictSize, (U32)(zcs->inBuff.filled - newDictSize));
-        memmove(zcs->inBuff.buffer.start, (const char*)zcs->jobs[jobID].srcStart + zcs->dictSize + srcSize - newDictSize, zcs->inBuff.filled);
+        DEBUGLOG(5, "new job : inBuff filled to %u, with %u dict and %u src",
+                    (U32)zcs->inBuff.filled, (U32)newDictSize,
+                    (U32)(zcs->inBuff.filled - newDictSize));
+        memmove(zcs->inBuff.buffer.start,
```
terrelln: I think that `zcs->inBuff.filled` can be `0` here, so we should check for that.

lib/compress/zstdmt_compress.c
Cyan4973: Yes, indeed.

Note that, in the long run, this `cdictLocal` only exists because it's currently allowed to invoke `ZSTDMT_*` functions directly.
In the future, when multi-threading will be only accessible through `ZSTD_compress_generic()`, this `cdictLocal` will no longer be useful, as cdict will be stored directly within host `cctx`.

lib/compress/zstdmt_compress.c
Cyan4973: Yes this is possible (though the conditions are pretty strong),
but I don't think this would be a problem anyway.

As far as I can tell, it's allowed to `memmove(,,0)`.
What's undefined is to `memmove(dst,src,0)`  _with `dst` or `src` invalid, including `NULL`_.

But here both `src` and `dst` pointers are well initialized and valid.

lib/compress/zstdmt_compress.c
Cyan4973: same comment

lib/compress/zstd_compress.c
Cyan4973: This test is done because, in some cases (`*flush()`, `*end()`), `{src,srcSize}=={NULL,0}`.
In which case, according to the standard, copying `0` bytes from `NULL` is UB.
https://stackoverflow.com/a/5243068/646947

`if (length)` catches that case, but `if (src)` may be even proper.

lib/compress/zstdmt_compress.c
terrelln: Good point, I got my UB conditions mixed up.

lib/decompress/zstd_decompress.c
```diff
@@ -2252,82 +2289,87 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
...
                 if (neededInSize==0) {  /* end of frame */
-                    zds->stage = zdss_init;
+                    zds->streamStage = zdss_init;
                     someMoreWork = 0;
                     break;
                 }
                 if ((size_t)(iend-ip) >= neededInSize) {  /* decode directly from src */
-                    const int isSkipFrame = ZSTD_isSkipFrame(zds->dctx);
-                    size_t const decodedSize = ZSTD_decompressContinue(zds->dctx,
+                    const int isSkipFrame = ZSTD_isSkipFrame(zds);
```
terrelln: nit: `const` on the right.

programs/fileio.c
```diff
@@ -102,10 +99,35 @@ void FIO_setNotificationLevel(unsigned level) { g_displayLevel=level; }
...
 
-#undef MIN
+#undef MIN  /* in case it would be already defined */
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 
 
+/*-*************************************
+*  Errors
+***************************************/
+#ifndef DEBUG
```
terrelln: Should it also use `ZSTD_DEBUG`?

tests/zstreamtest.c
```diff
@@ -711,8 +729,16 @@ static int fuzzerTests(U32 seed, U32 nbTests, unsigned startTest, double compres
 
         /* states full reset (deliberately not synchronized) */
         /* some issues can only happen when reusing states */
-        if ((FUZ_rand(&lseed) & 0xFF) == 131) { ZSTD_freeCStream(zc); zc = ZSTD_createCStream(); resetAllowed=0; }
-        if ((FUZ_rand(&lseed) & 0xFF) == 132) { ZSTD_freeDStream(zd); zd = ZSTD_createDStream(); ZSTD_initDStream_usingDict(zd, NULL, 0);  /* ensure at least one init */ }
+        if ((FUZ_rand(&lseed) & 0xFF) == 131) {
+            ZSTD_freeCStream(zc);
+            zc = ZSTD_createCStream();
```
terrelln: We should probably check for `NULL`.

tests/zstreamtest.c
```diff
@@ -711,8 +729,16 @@ static int fuzzerTests(U32 seed, U32 nbTests, unsigned startTest, double compres
...
+            resetAllowed=0;
+        }
+        if ((FUZ_rand(&lseed) & 0xFF) == 132) {
+            ZSTD_freeDStream(zd);
+            zd = ZSTD_createDStream();
```
terrelln: Same here.

tests/zstreamtest.c
```diff
@@ -1294,8 +1597,22 @@ int main(int argc, const char** argv)
             result = basicUnitTests(0, ((double)proba) / 100, customMem);  /* use custom memory allocation functions */
     }   }
 
-    if (!result && !mtOnly) result = fuzzerTests(seed, nbTests, testNb, ((double)proba) / 100, bigTests);
-    if (!result) result = fuzzerTests_MT(seed, nbTests, testNb, ((double)proba) / 100, bigTests);
+    if (!result) {
```
terrelln: If no flags are set, do we only want to run `simple_api`, or do we want to run them all?

lib/decompress/zstd_decompress.c
```diff
@@ -145,36 +176,72 @@ size_t ZSTD_decompressBegin(ZSTD_DCtx* dctx)
...
+#if defined(ZSTD_LEGACY_SUPPORT) && (ZSTD_LEGACY_SUPPORT >= 1)
+        if (dctx->legacyContext)
+            ZSTD_freeLegacyStreamContext(dctx->legacyContext, dctx->previousLegacyVersion);
+#endif
+        ZSTD_free(dctx, cMem);
+        return 0;
+    }
 }
 
+/* no longer appropriate */
```
Cyan4973: `ZSTD_copyDCtx()` used to be the starting method within `ZSTD_decompress_usingDDict()`.
But that's no longer the case, and `ZSTD_copyDCtx()` has no more usage anywhere.
The only reason it's still there is that it's part of the public API (though experimental section).

I'm wondering if it's still useful to keep it.
Maybe there are external use cases which could benefit from ability to clone a `DCtx` ?
I don't know.

programs/fileio.c
Cyan4973: Yes, probably, it would make more sense

tests/zstreamtest.c
Cyan4973: I changed the logic to only run `simple_api`.
In most cases, I know which API I need to test.

Cyan4973:
qemu ARM fuzzer tests still fail in multithreading mode.
The latest version of `--no-big-tests` has highly reduced the amount of memory needed during tests.
According to my measurement, using `top`, `zstreamtest --mt --no-big-tests` process may request up to 200 MB at peak, but rarely ever use more than 120 MB. This should be good enough for qemu, even on a Travis VM.

terrelln:
Looks good to me once the tests pass.

Cyan4973:
OK, I've spent a lot of time into this (too much really),
so let's write down some findings : 

- `qemu-arm-static` tests fail very frequently on reaching `zstreamtest --newapi --no-big-tests`
- I tried to reproduce it on a native linux box, with 4 GB of RAM. It never fails.
- I tried to reproduce it on a Linux VM in a Mac host, with 4 GB of assigned RAM. It never fails.
- Same linux VM, but with only 2 GB of assigned RAM : 
   + It almost always fail, it's just a matter of time.
   + Failure seems to be a `qemu-arm-static` segfault, but it's hard to be sure
   + It never fails at the same place. No way to reliably reproduce the same failure.
   + The first run needs some time before failing. Subsequent runs fail faster.
- `qemu-arm-static` allocates 4 GB upfront, whatever the real RAM size
   + I have not found any way to reduce that amount

_What I suspect_ : 
- `zstreamtest --no-big-tests` never needs more than 200 MB. So these 4 GB are generally not used. The linux host is clever enough to _not_ attribute 4 GB of _physical_ memory. It waits for the memory to be really invoked before actually providing it. So the initial "4 GB allocation" is fake.
- As long as tests are single-threaded, there is no problem, because memory does not move too much : some memory is `free()` and then `malloc()` again, there's a good chance that all these buffers remain more or less in the same region. So the total amount of physical RAM touched never goes too far, probably < 500 MB, or even less.
- When multi-threading is enabled, it becomes messy : buffers are allocated and free in a much more chaotic ways. Consequently, address space starts to become fragmented. The Linux host has to provide more and more "real" RAM to serve its guest. At some point, on 2 GB systems, there is not enough RAM available (swap included), and the system crashes.

Note that : 
I'm not completely happy with this explanation, because `zstreamtest --mt` should fail as badly as `zstreamtest --newapi` does. That's not the case in my tests : `zstreamtest --mt` does fail, but much less often. So there is something fishy. Maybe the `--mt` test is different for some reason, for example maybe it moves buffers around less often, so it takes more time for memory fragmentation to ramp up.

Anyway, this leaves us with few potential solutions : 
- I could not find a way to reduce amount of memory allocated by `qemu-arm-static`, so I gave up that route.
- Simplest solution : do not run `--newapi` tests with `qemu-user-static` on Travis CI. Feels a bit too strong.
- Less drastic solution : disable `multi-threading` with `--newapi` on `qemu-user-static`. At least the single-threaded version would still be tested. This test doesn't fail on 2 GB VM.

For now, I'm considering solution 3.

Cyan4973:
Solution 3 implemented. Not entirely satisfactory (solution 1 would have been better) but tests now pass.
Maybe in the future, if it's possible to reduce the amount of free/create inside `zstreamtest --newapi`, it will be possible to enable again the full test in TravisCI VM.

PR #727, opened 2017-06-19
[libzstd] Fix UBSAN failure

terrelln:
Detected by a check added to UBSAN in trunk clang.


Cyan4973:
While I can certainly accept the fix, which looks totally fine,
I'm nonetheless surprised by the diagnosis : 

> (-offset+ZSTD_REP_MOVE-1) is unsigned, and is added to start

If that's so, it should have failed *all the time*, not only on some very rare circumstances...

PR #729, opened 2017-06-20
Corpus

paulcruz74:
- added dictionary support through `--use-dict #` parameter (takes size of dictionary)
- added a test to ensure decompressing with dictionary is possible

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -342,21 +342,36 @@ test-decodecorpus: decodecorpus
 
 test-decodecorpus-cli: decodecorpus
 	@echo "\n ---- decodecorpus basic cli tests ----"
-	mkdir tests
-	./decodecorpus -n5 -otests -ptests
-	cd testdir
-	zstd -d z000000.zst -D dictionary -o tmp0
-	zstd -d z000001.zst -D dictionary -o tmp1
-	zstd -d z000002.zst -D dictionary -o tmp2
-	zstd -d z000003.zst -D dictionary -o tmp3
-	zstd -d z000004.zst -D dictionary -o tmp4
-	diff z000000 tmp0
-	diff z000001 tmp1
-	diff z000002 tmp2
-	diff z000003 tmp3
-	diff z000004 tmp4
-	cd ..
-	rm -rf testdir
+	@mkdir testdir
+	./decodecorpus -n5 -otestdir -ptestdir
+	@cd testdir \
+	zstd -d z000000.zst -o tmp0 \
+	zstd -d z000001.zst -o tmp1 \
+	zstd -d z000002.zst -o tmp2 \
+	zstd -d z000003.zst -o tmp3 \
+	zstd -d z000004.zst -o tmp4 \
+	diff z000000 tmp0 \
+	diff z000001 tmp1 \
+	diff z000002 tmp2 \
+	diff z000003 tmp3 \
+	diff z000004 tmp4 \
+	@rm ./* \
+	@cd ..
+
+	./decodecorpus -n5 -otestdir -ptestdir --use-dict=1MB
+	@cd testdir \
+	zstd -d z000000.zst -D dictionary -o tmp0 \
+	zstd -d z000001.zst -D dictionary -o tmp1 \
+	zstd -d z000002.zst -D dictionary -o tmp2 \
+	zstd -d z000003.zst -D dictionary -o tmp3 \
+	zstd -d z000004.zst -D dictionary -o tmp4 \
+	diff z000000 tmp0 \
+	diff z000001 tmp1 \
+	diff z000002 tmp2 \
+	diff z000003 tmp3 \
+	diff z000004 tmp4 \
+	@cd .. 
+	@rm -rf testdir
 
 test-pool: pool
 	$(QEMU_SYS) ./pool
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -344,33 +344,32 @@ test-decodecorpus-cli: decodecorpus
 	@echo "\n ---- decodecorpus basic cli tests ----"
 	@mkdir testdir
 	./decodecorpus -n5 -otestdir -ptestdir
-	@cd testdir \
-	zstd -d z000000.zst -o tmp0 \
-	zstd -d z000001.zst -o tmp1 \
-	zstd -d z000002.zst -o tmp2 \
-	zstd -d z000003.zst -o tmp3 \
-	zstd -d z000004.zst -o tmp4 \
-	diff z000000 tmp0 \
-	diff z000001 tmp1 \
-	diff z000002 tmp2 \
-	diff z000003 tmp3 \
-	diff z000004 tmp4 \
-	@rm ./* \
-	@cd ..
-
+	@cd testdir && \
+	$(ZSTD) -d z000000.zst -o tmp0 && \
+	$(ZSTD) -d z000001.zst -o tmp1 && \
+	$(ZSTD) -d z000002.zst -o tmp2 && \
+	$(ZSTD) -d z000003.zst -o tmp3 && \
+	$(ZSTD) -d z000004.zst -o tmp4 && \
+	diff z000000 tmp0 && \
+	diff z000001 tmp1 && \
+	diff z000002 tmp2 && \
+	diff z000003 tmp3 && \
+	diff z000004 tmp4 && \
+	rm ./* && \
+	cd ..
 	./decodecorpus -n5 -otestdir -ptestdir --use-dict=1MB
-	@cd testdir \
-	zstd -d z000000.zst -D dictionary -o tmp0 \
-	zstd -d z000001.zst -D dictionary -o tmp1 \
-	zstd -d z000002.zst -D dictionary -o tmp2 \
-	zstd -d z000003.zst -D dictionary -o tmp3 \
-	zstd -d z000004.zst -D dictionary -o tmp4 \
-	diff z000000 tmp0 \
-	diff z000001 tmp1 \
-	diff z000002 tmp2 \
-	diff z000003 tmp3 \
-	diff z000004 tmp4 \
-	@cd .. 
+	@cd testdir && \
+	$(ZSTD) -d z000000.zst -D dictionary -o tmp0 && \
+	$(ZSTD) -d z000001.zst -D dictionary -o tmp1 && \
+	$(ZSTD) -d z000002.zst -D dictionary -o tmp2 && \
+	$(ZSTD) -d z000003.zst -D dictionary -o tmp3 && \
+	$(ZSTD) -d z000004.zst -D dictionary -o tmp4 && \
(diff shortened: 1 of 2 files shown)


tests/Makefile
```diff
@@ -180,7 +180,7 @@ legacy : CPPFLAGS+= -I$(ZSTDDIR)/legacy
 legacy : $(ZSTD_FILES) $(wildcard $(ZSTDDIR)/legacy/*.c) legacy.c
 	$(CC)      $(FLAGS) $^ -o $@$(EXT)
 
-decodecorpus	: $(filter-out $(ZSTDDIR)/compress/zstd_compress.c, $(wildcard $(ZSTD_FILES))) decodecorpus.c
+decodecorpus	: $(filter-out $(ZSTDDIR)/compress/zstd_compress.c, $(wildcard $(ZSTD_FILES))) $(ZDICT_FILES) decodecorpus.c
 	$(CC)      $(FLAGS) $^ -o $@$(EXT) -lm
```
Cyan4973: Internal tests, using ZSTD API, is fine.
I was thinking it could be augmented with CLI-based tests.
They use different entry points, and may be subject to different limitations.

tests/decodecorpus.c
```diff
@@ -306,15 +314,19 @@ static void writeFrameHeader(U32* seed, frame_t* frame)
     pos += 4;
 
     {
+        int const dictBits = info.useDict ? 3 : 0;
         BYTE const frameHeaderDescriptor =
-                (BYTE) ((fcsCode << 6) | (singleSegment << 5) | (1 << 2));
+                (BYTE) ((fcsCode << 6) | (singleSegment << 5) | (1 << 2) | dictBits);
```
Cyan4973: Comment or link to the specification
so that we understand which bit does what.

tests/decodecorpus.c
```diff
@@ -662,6 +674,12 @@ static U32 generateSequences(U32* seed, frame_t* frame, seqStore_t* seqStore,
                           MIN(frame->header.windowSize,
                               (size_t)((BYTE*)srcPtr - (BYTE*)frame->srcStart))) +
                          1;
+                if (info.useDict && (RAND(seed) & 1)) {
+                    /* need to occasionally generate offsets that go past the start */
+                    /* we still need to be within the windowSize however */
```
Cyan4973: minor clarification : you don't need to respect `windowSize` as a max distance for a match within the dictionary.
In fact, there is no limit, save the size of dictionary itself.
However, the dictionary cease being accessible after regenerating `windowSize` bytes.

tests/decodecorpus.c
```diff
@@ -677,11 +695,20 @@ static U32 generateSequences(U32* seed, frame_t* frame, seqStore_t* seqStore,
...
-        } while (offset > (size_t)((BYTE*)srcPtr - (BYTE*)frame->srcStart) || offset == 0);
+        } while (((!info.useDict) && (offset > (size_t)((BYTE*)srcPtr - (BYTE*)frame->srcStart))) || offset == 0);
 
-        {   size_t j;
+        {
+            size_t j;
+            BYTE* const dictEnd = info.dictContent + info.dictContentSize;
             for (j = 0; j < matchLen; j++) {
-                *srcPtr = *(srcPtr-offset);
+                if ((void*)(srcPtr - offset) < (void*)frame->srcStart) {
```
Cyan4973: how do you handle matches which overlap both the dictionary and the current buffer ?

tests/decodecorpus.c
```diff
@@ -1121,20 +1148,94 @@ static void initFrame(frame_t* fr)
...
 }
 
+/*_*******************************************************
+*  Dictionary Helper Functions
+*********************************************************/
+/* returns 0 if successful, otherwise returns 1 upon error */
+static int genRandomDict(U32 dictID, U32 seed, size_t dictSize, BYTE* fullDict){
+    /* allocate space for samples */
+    unsigned const numSamples = 4;
+    BYTE samples[5000];
```
Cyan4973: minor : 5000 bytes is a "relatively" large value for stack allocation.
Granted, on PC/Server, it should not cause troubles,
although we have micro-thread tasks which can be affected.

But just as a good practice, try to avoid that.
If you need a large buffer, `malloc()` it.
And then `free()` it ...

tests/decodecorpus.c
```diff
@@ -1121,20 +1148,94 @@ static void initFrame(frame_t* fr)
...
+    size_t sampleSizes[4];
+
+    /* generate samples */
+    {
+        unsigned i = 1;
```
Cyan4973: anything more expressive than `i`, `j` ?
Or describe what does this paragraph ?

tests/decodecorpus.c
```diff
@@ -1121,20 +1148,94 @@ static void initFrame(frame_t* fr)
...
+        }
+    }
+
+    return 0;
+}
+
+static dictInfo initDictInfo(int useDict, size_t dictContentSize, BYTE* dictContent, U32 dictID){
+    /* allocate space statically */
+    dictInfo dictOp;
+    memset((void*)(&dictOp), 0, sizeof(dictOp));
```
Cyan4973: minor : I think you can get away with `(void*)`.
Any pointer can be `void*`, that's basically the point of this type, to not need a cast.

tests/decodecorpus.c
```diff
@@ -1196,6 +1297,63 @@ static size_t testDecodeStreaming(frame_t* fr)
...
+    U32 const dictID = RAND(&seed);
+    size_t errorDetected = 0;
+    BYTE* const fullDict = malloc(dictSize);
+    if (fullDict == NULL) {
+        return ERROR(GENERIC);
+    }
+
+    /* generate random dictionary */
+    {
+        int ret = genRandomDict(dictID, seed, dictSize, fullDict);
```
Cyan4973: `const`

tests/decodecorpus.c
```diff
@@ -1272,7 +1445,10 @@ static int generateCorpus(U32 seed, unsigned numFiles, const char* const path,
 
         DISPLAYUPDATE("\r%u/%u        ", fnum, numFiles);
 
-        seed = generateFrame(seed, &fr);
+        {
+            dictInfo const info = initDictInfo(0, 0, NULL, 0);
+            generateFrame(seed, &fr, info);
```
Cyan4973: you no longer update `seed` after each `generateFrame()`. Is that intentional ?

tests/decodecorpus.c
```diff
@@ -1294,6 +1470,92 @@ static int generateCorpus(U32 seed, unsigned numFiles, const char* const path,
...
+    /* allocate space for the dictionary */
+    fullDict = malloc(dictSize);
+    if (fullDict == NULL) {
+        DISPLAY("Error: could not allocate space for full dictionary.\n");
+        return 1;
+    }
+
+    /* randomly generate the dictionary */
+    {
+        int ret = genRandomDict(dictID, seed, dictSize, fullDict);
```
Cyan4973: `const`

tests/decodecorpus.c
paulcruz74: The loop copies a single byte at a time into `srcPtr` and then increments `srcPtr` for a total of `matchLen` times. If a match has overlap with both the dictionary and the current buffer, I believe the `if` condition would be triggered (offset goes past the start of the source), in which case we copy the byte from the dictionary. Once the loop gets to the part of the match that overlaps with the current buffer, the condition is no longer satisfied and it  copies the byte from the source buffer.

tests/decodecorpus.c
```diff
@@ -1339,6 +1601,7 @@ static void advancedUsage(const char* programName)
     DISPLAY( "\n");
     DISPLAY( "Advanced arguments :\n");
     DISPLAY( " --content-size    : always include the content size in the frame header\n");
+    DISPLAY( " --use-dict #      : include a dictionary used to decompress the corpus\n");
```
Cyan4973: Prefer `--use-dict=#`, for consistency with other tools within `zstd` repository.
Look into `zstdcli.c` for guidance if need be.

tests/Makefile
paulcruz74: Should I write a separate script for testing with CLI? something similar to `playTests.sh`?

tests/decodecorpus.c
paulcruz74: I must have misunderstood that. So only `windowSize` bytes can be regenerated from the dictionary. Does that mean that we can also regenerate additional bytes from the current buffer? For example, can we have a match that overlaps with the dictionary for `windowSize` bytes and then overlaps with the current buffer for additional length? Are there any restrictions on that?

tests/Makefile
Cyan4973: Actually, I was thinking the tests script could be located directly in `Makefile`.
You can have a look at `lz4` project, in `tests/Makefile`, for guidance.

The reason we have `playTests.sh` in `zstd` is that some (rare) platforms don't have `make` pre-installed, but their owner nonetheless want to run some tests.

tests/decodecorpus.c
Cyan4973: Indeed, if the dictionary is larger than `windowSize`, and the encoder uses this opportunity to create offsets larger than `windowSize`, the match must nonetheless end at of before dictionary limit. It cannot overlap.

If the offset is smaller than `windowSize`, then it can overlap with current buffer.

tests/decodecorpus.c
```diff
@@ -654,14 +672,31 @@ static U32 generateSequences(U32* seed, frame_t* frame, seqStore_t* seqStore,
...
+                            /* make sure lenPastStart does not go past dictionary start though */
+                            lenPastStart = MIN(lenPastStart+MIN_SEQ_LEN, info.dictContentSize);
+                            offset = ((BYTE*)srcPtr - (BYTE*)frame->srcStart) + lenPastStart;
+                        }
+                        {
+                            U32 const matchLenBound = MIN(frame->header.windowSize, lenPastStart);
+                            matchLen = MIN(matchLen, matchLenBound);
+                        }
+                    }
+                }
```
paulcruz74: Currently still testing for bugs, it seems that even though the bus error doesn't occur anymore, there are instances where the manually decompressed version and the versions obtained through CLI/API don't match. I suspect it still has something to do with the offsets/match lengths I'm using

paulcruz74: Fixed the bug detected by the API test (checksum mismatch during file generation). The last sequence is specifically generated to adhere to the content size that was predetermined. Currently investigating a bug with the CLI test where corruption is detected.

paulcruz74: Fixed corruption problem by ensuring generation of long offsets can't occur once we've generated `windowSize` worth of decompressed data. Should I add `test-decodecorpus-cli` to `test-all` in the Makefile so that it gets run during CI?

Cyan4973: Yes.
Also, ensure the test does not cost too much time.

PR #734, opened 2017-06-23
Assume

Cyan4973:
Convert `assert()` into `assume()` in release mode

### DIFF
--- a/doc/zstd_manual.html
+++ b/doc/zstd_manual.html
@@ -389,6 +389,12 @@ unsigned long long ZSTD_getFrameContentSize(const void *src, size_t srcSize);
             however it does mean that all frame data must be present and valid. 
 </p></pre><BR>
 
+<pre><b>size_t ZSTD_frameHeaderSize(const void* src, size_t srcSize);
+</b><p>   `src` should point to the start of a ZSTD frame
+   `srcSize` must be >= ZSTD_frameHeaderSize_prefix.
+   @return : size of the Frame Header 
+</p></pre><BR>
+
 <a name="Chapter13"></a><h2>Context memory usage</h2><pre></pre>
 
 <pre><b>size_t ZSTD_sizeof_CCtx(const ZSTD_CCtx* cctx);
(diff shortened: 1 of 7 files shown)


terrelln:
We can do slightly better for clang, see https://github.com/facebook/folly/blob/master/folly/Assume.h.

Cyan4973:
Good catch !

Cyan4973:
mmmh, now I've got a ton of 
`warning: the argument to '__builtin_assume' has side effects that will be discarded [-Wassume]`
on `clang`.

Maybe related to `complex_condition` thing.

So maybe it will be necessary to have 2 types of `assert()` after all.

terrelln:
We can probably write a function called `assume()` or `ZSTD_assume()`, and have `assert()` call that. I guess the problem is that `__builtin_assume()` doesn't necessarily evaluate its arguments.

Alternatively, I think marking functions that are called as arguments to `assert()` as `__attribute__((pure))` or `__attribute__((const))`, assuming they are, would work.

terrelln:
Oh, but that executes the function

Cyan4973:
Yes, function get invoked, which produces an associated performance drop (acceptable in debug mode).

This is more likely to be the case for `gcc` only, since its `assume()` is not a `__builtin`, but a user function. See : 
https://stackoverflow.com/questions/25667901/assume-clause-in-gcc#comment64879855_26195434

Cyan4973:
`__builtin_assume()` seems to fail on CircleCI, using `clang` v3.4.

Is there a minimum version required for this `__builtin` to exist ?

Cyan4973:
I've been benchmarking the `assume` version this week-end,
difference with `dev` branch are very small, well within margin error (< 1 %).
Overall, it does not look positive, the `assume` version seems to be a tiny bit slower, which is not the expectation.

Anyway, even if this is considered a tie, since there is no perceptible gain, and this patch just increases code complexity, I think it's better to pass (at least for now).

PR #735, opened 2017-06-24
[zstdcli] Support xz by default when liblzma is available

terrelln:
* Support lzma by default
* Fix conversion warnings
* Review lzma compression/decompression code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,10 +224,10 @@ asan-%: clean
 	LDFLAGS=-fuse-ld=gold MOREFLAGS="-g -fno-sanitize-recover=all -fsanitize=address" $(MAKE) -C $(TESTDIR) $*
 
 msan: clean
-	$(MAKE) test CC=clang MOREFLAGS="-g -fsanitize=memory -fno-omit-frame-pointer"   # datagen.c fails this test for no obvious reason
+	$(MAKE) test CC=clang MOREFLAGS="-g -fsanitize=memory -fno-omit-frame-pointer HAVE_LZMA=0"   # datagen.c fails this test for no obvious reason
 
 msan-%: clean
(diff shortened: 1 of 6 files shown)


Cyan4973:
We should indicate how to build `zstd` without `xz`/`lzma` support, for whoever wants to control the processus.
It typically means adding a paragraph in `readme.md`, regarding `HAVE_LZMA` : 
https://github.com/facebook/zstd/tree/dev/programs#compilation-variables

Note : forcing the values of these variables is useful both ways : 
`== 0` disable the selected support
`== 1` forces it, which means linking stage will fail later if dependency is not fulfilled. This can be useful if a distribution wants to enforce a certain level of service when compiling `zstd`.

Cyan4973:
Interesting, `msan` found an issue _inside liblzma.so_ :
https://travis-ci.org/facebook/zstd/jobs/246417043#L1183

terrelln:
Good point, I didn't realize that the readme existed.

I'll look into the MSAN failure, but it might be a false positive, since MSAN doesn't work well with non-instrumented binaries.

terrelln:
I just built `liblzma` with MSAN, and there is no failure. When I linked against the uninstrumented `liblzma` MSAN fails. Its failing because unlike ASAN/TSAN/UBSAN, MSAN needs all linked libraries to be instrumented.

Cyan4973:
Looks good to me !
Waiting for CI test results before merge

Cyan4973:
minor issue : 
https://travis-ci.org/facebook/zstd/jobs/247212007#L578

I guess `HAVE_LZMA=0` should be its own argument, it's not part of `LDFLAGS`

Cyan4973:
Note : `pzstd` tests failed on Appveyor while testing this merge : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.1213/job/4ddtl5t5eenyeght

This is more than likely unrelated to this patch.
But it reminds us that `pzstd` tests have always been a bit flacky (I would suspect `gtests` framework here) and take some time. So maybe it's better to simply disable them in Appveyor CI ?

PR #736, opened 2017-06-24
[zdict] Make COVER the default algorithm

terrelln:
* Prefix legacy dictionary builder functions with `LEGACY` instead of `ZDICT`.
* Extract common parameters into `ZDICT_params_t`.
* The stable API keeps the `ZDICT` prefix, and users of the advanced API
  explicitly choose an algorithm `COVER` or `LEGACY`.

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -398,7 +398,8 @@ typedef struct {
  */
 static COVER_segment_t COVER_selectSegment(const COVER_ctx_t *ctx, U32 *freqs,
                                            COVER_map_t *activeDmers, U32 begin,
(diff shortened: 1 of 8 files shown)


Cyan4973:
It's not very clear what's the appropriate prefix.
Sometime we've got `ZDICT_*`, sometimes `COVER_*`, and sometimes `LEGACY_*`.

I believe it would be preferable to reduce the nb of prefixes (typically sticking to `ZDICT_`), and distinguish versions by suffix. For example `LEGACY_trainFromBuffer()` could become `ZDICT_trainFromBuffer_legacy()` ?

terrelln:
Yeah, I wasn't sure what was best. That naming scheme seems less confusing.

Issue #21, opened 2015-02-17
Zstd library should be compatible with static allocation


Cyan4973:
Requested by Dimitri
Original discussion : http://fastcompression.blogspot.fr/2015/01/zstd-stronger-compression-algorithm.html?showComment=1424173050454#c7703504284913974280

DCNMS:
Thanks Yann to take the static allocation in consideration.

This compatibily would allow embedded system use. I target use on STM32.

Cyan4973:
This objective looks today a bit difficult.

For decompression, it looks remotely possible, but difficult.
`ZSTD_Dctx` currently has a static size, yes, but a fairly large one.

Should this amount of memory be considered too large for embedded systems, reducing this memory is possible, especially when working with small data blocks. But then, it makes the required Dctx memory variable, which is exactly what must be avoided.

Over the long term, it's possible to reduce sharply the size of Dctx, but that would require a "dedicated" version, handling compressed data differently. Nothing technically impossible, just a different version, with the charge to maintain it. So my feeling is that it must be carefully pondered if it needs to be part of the reference library.

For compression, it starts by being worse : the amount of memory to reserve depends on the size of data to compress, and selected compression level. So it is variable, by construction.

The only solution I can think of would be to allow maximum memory size enforcement. In which case, the compression algorithm would refuse to start if it requires more than allowed memory. Rather than selecting a compression level, the caller should better use `_advanced()` variants, which allow control over individual parameters, in order to precisely control the amount of memory spent.

That being said, using current source code, the final amount of memory reserved can still be quite high even for small blocks, from an embedded system perspective. I haven't made a precise calculation, but I suspect that it requires a minimum close to 100 KB. Going lower is possible, but once again, it would require a specific version, designed to be frugal on memory space. Technically achievable, but a specific effort, with dedicated maintenance. Same question than the decompression part.

For discussion

Bulat-Ziganshin:
LZMA api allows to pass alloc/free functions provided by the library user. On embedded systems, this custom alloc function then should just return pointers to preallocated static buffer.

Cyan4973:
That's a good point @Bulat-Ziganshin .

Also, as a noticeable progress to note on this front, v0.6 now makes it possible to select fairly small memory sizes for decompression. Decoding tables are a mere 5 KB, and work buffers are now guaranteed to be <= block size, so a block size of 4 KB for example induces a pretty small memory budget for the decompressor.

DCNMS:
[...]

PR #737, opened 2017-06-27
simplified API for *initStatic*()

Cyan4973:
After redacting https://github.com/facebook/zstd/issues/21#issuecomment-310737366, it felt that the previous `*_initStatic*()` had shortcomings making it a bit unnecessary complex to use.

This proposal improves usability on 2 points : 
- Simplified `*_estimate*()` functions : In particular, the compression estimations were using `ZSTD_compressParameters` type as argument, which is an advanced struct. The new version now requests a simpler `int compressionLevel` for basic prototypes, while still allowing the richer `cParams` in `_advanced()` version. `*_estimateDStreamSize*()` has been updated too, requesting a `size_t windowLog` rather than more peculiar `ZSTD_frameHeader` type, to be easier to invoke.
- Added `ZSTD_initStatic{C,D}Stream()`, which are merely redirections towards `ZSTD_initStatic{C,D}Ctx`, since they are now the same type. The new naming makes the allocation / initialization flow more logical  :  `ZSTD_estimateCStreamSize()` => `ZSTD_initStaticCStream()` => `ZSDT_initCStream()`, etc.) .

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -495,11 +495,14 @@ ZSTDLIB_API size_t ZSTD_sizeof_DDict(const ZSTD_DDict* ddict);
 
 /*! ZSTD_estimate*() :
  *  These functions make it possible to estimate memory usage
- *  of a future target object, before its allocation,
- *  given a set of parameters, which vary depending on target object.
+ *  of a future {D,C}Ctx, before its creation.
  *  The objective is to guide decision before allocation.
+ *  ZSTD_estimateCCtxSize() will consider src size to be arbitrarily "large".
+ *  If srcSize is known to be small, ZSTD_estimateCCtxSize_advanced() will provide a better (smaller) estimation.
+ *  ZSTD_estimateCCtxSize_advanced() can be used in tandem with ZSTD_getCParams() to create cParams from compressionLevel.
  *  Note : CCtx estimation is only correct for single-threaded compression */
-ZSTDLIB_API size_t ZSTD_estimateCCtxSize(ZSTD_compressionParameters cParams);
+ZSTDLIB_API size_t ZSTD_estimateCCtxSize(int compressionLevel);
+ZSTDLIB_API size_t ZSTD_estimateCCtxSize_advanced(ZSTD_compressionParameters cParams);
 ZSTDLIB_API size_t ZSTD_estimateDCtxSize(void);
 
 /*! ZSTD_estimate?StreamSize() :
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -412,7 +412,7 @@ typedef struct {
 
 typedef struct {
     unsigned long long frameContentSize;
-    unsigned windowSize;
+    size_t windowSize;
     unsigned dictID;
     unsigned checksumFlag;
 } ZSTD_frameHeader;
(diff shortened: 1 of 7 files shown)


Cyan4973:
Any comment on this patch @terrelln ?

lib/zstd.h
```diff
@@ -495,23 +495,38 @@ ZSTDLIB_API size_t ZSTD_sizeof_DDict(const ZSTD_DDict* ddict);
...
- *  of a future target object, before its allocation,
- *  given a set of parameters, which vary depending on target object.
+ *  of a future {D,C}Ctx, before its creation.
  *  The objective is to guide decision before allocation.
+ *  ZSTD_estimateCCtxSize() will consider src size to be arbitrarily "large".
+ *  If srcSize is known to be small, ZSTD_estimateCCtxSize_advanced() can provide a tighter estimation.
+ *  ZSTD_estimateCCtxSize_advanced() can be used in tandem with ZSTD_getCParams() to create cParams from compressionLevel.
  *  Note : CCtx estimation is only correct for single-threaded compression */
-ZSTDLIB_API size_t ZSTD_estimateCCtxSize(ZSTD_compressionParameters cParams);
+ZSTDLIB_API size_t ZSTD_estimateCCtxSize(int compressionLevel);
```
terrelln: Is the memory usage monotonically increasing with compression level? Its probably worth adding a test, since it would be surprising if it wasn't.

Cyan4973: Good point !
It's not, indeed, as there is one exception (that I'm aware of) : 
level 4 uses 2 MB of tables, while level 5 "only" uses 1.25 MB.

Indeed, this can be a surprising outcome.

One way to solve this could be to change definition of compression level 4 and/or 5.

Another one would be to provide the "max memory usage of every compression level before and including selected one".

Cyan4973: OK, I've tested a modified compression level table to guarantee that memory budget is monotonically increasing (I found a few more cases besides level 4->5).
The new compression level table still guarantees that higher compression levels compress more, at least on `silesia.tar`, and that lower compression levels compress faster, at least on target host (Chip's Entropy).
What is slightly impacted is the regularity of the speed slowdown. And still, the impact is quite moderate.

As a plus side, I believe the regularity of speed decrease was really too specific to target host, since `paramgrill` calculated it on this machine, and it would be different on another host (especially with different cache size). I actually expect the monotonically increase in memory budget to smooth things for _other_ hosts. So I'm inclined to adopt the new table.

In itself, it's a good enough quick fix for `ZSTD_estimateCCtxSize()`, where this property _feels implied_.

Now, thing is, a future change _might_ impact this property.
So it could be documented.
It could also be enforced within `paramgrill` so that it respects this property while generating a new compression levels table.

And lastly, `ZSTD_estimateCCtxSize()` could internally implement a "max of all lower levels memory usage" policy, even if it's not useful with the new compression level table, just to be on the safe side regarding future changes.

terrelln: Either way sounds reasonable to me, but I would like a test that makes sure `ZSTD_estimateCCtxSize()` is always increasing, so we can't modify the parameters and accidentally forget about it.

Cyan4973: ok `static_assert` doesn't work, since a `static const` variable is not a `constant expression`,
so let's try a more runtime `assert()` method ...

Cyan4973:
monotonic property and associated test added

PR #738, opened 2017-06-27
[linux-kernel] Update patches for v2

terrelln:
- Use div_u64 for division of u64s (2/4)
- Reduce stack usage of FSE_buildCTable_wksp(), FSE_compress_wksp, and HUF_readDTableX4 (2/4)

### DIFF
--- a/contrib/linux-kernel/lib/zstd/compress.c
+++ b/contrib/linux-kernel/lib/zstd/compress.c
@@ -84,7 +84,7 @@ struct ZSTD_CCtx_s {
 	FSE_CTable offcodeCTable[FSE_CTABLE_SIZE_U32(OffFSELog, MaxOff)];
 	FSE_CTable matchlengthCTable[FSE_CTABLE_SIZE_U32(MLFSELog, MaxML)];
 	FSE_CTable litlengthCTable[FSE_CTABLE_SIZE_U32(LLFSELog, MaxLL)];
-	unsigned tmpCounters[HUF_WORKSPACE_SIZE_U32];
+	unsigned tmpCounters[HUF_COMPRESS_WORKSPACE_SIZE_U32];
 };
 
 size_t ZSTD_CCtxWorkspaceBound(ZSTD_compressionParameters cParams)
@@ -587,8 +587,6 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 {
 	const int longOffsets = zc->params.cParams.windowLog > STREAM_ACCUMULATOR_MIN;
 	const seqStore_t *seqStorePtr = &(zc->seqStore);
-	U32 count[MaxSeq + 1];
-	S16 norm[MaxSeq + 1];
 	FSE_CTable *CTable_LitLength = zc->litlengthCTable;
 	FSE_CTable *CTable_OffsetBits = zc->offcodeCTable;
 	FSE_CTable *CTable_MatchLength = zc->matchlengthCTable;
@@ -602,7 +600,21 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 	BYTE *op = ostart;
 	size_t const nbSeq = seqStorePtr->sequences - seqStorePtr->sequencesStart;
 	BYTE *seqHead;
-	BYTE scratchBuffer[1 << MAX(MLFSELog, LLFSELog)];
+
+	U32 *count;
+	S16 *norm;
+	U32 *workspace;
+	size_t workspaceSize = sizeof(zc->tmpCounters);
+	{
+		size_t spaceUsed32 = 0;
+		count = (U32 *)zc->tmpCounters + spaceUsed32;
+		spaceUsed32 += MaxSeq + 1;
+		norm = (S16 *)((U32 *)zc->tmpCounters + spaceUsed32);
+		spaceUsed32 += ALIGN(sizeof(S16) * (MaxSeq + 1), sizeof(U32)) >> 2;
+
+		workspace = (U32 *)zc->tmpCounters + spaceUsed32;
+		workspaceSize -= (spaceUsed32 << 2);
+	}
 
 	/* Compress literals */
 	{
@@ -638,7 +650,7 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 	/* CTable for Literal Lengths */
 	{
 		U32 max = MaxLL;
-		size_t const mostFrequent = FSE_countFast_wksp(count, &max, llCodeTable, nbSeq, zc->tmpCounters);
+		size_t const mostFrequent = FSE_countFast_wksp(count, &max, llCodeTable, nbSeq, workspace);
 		if ((mostFrequent == nbSeq) && (nbSeq > 2)) {
 			*op++ = llCodeTable[0];
 			FSE_buildCTable_rle(CTable_LitLength, (BYTE)max);
@@ -646,7 +658,7 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 		} else if ((zc->flagStaticTables) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
 			LLtype = set_repeat;
 		} else if ((nbSeq < MIN_SEQ_FOR_DYNAMIC_FSE) || (mostFrequent < (nbSeq >> (LL_defaultNormLog - 1)))) {
-			FSE_buildCTable_wksp(CTable_LitLength, LL_defaultNorm, MaxLL, LL_defaultNormLog, scratchBuffer, sizeof(scratchBuffer));
+			FSE_buildCTable_wksp(CTable_LitLength, LL_defaultNorm, MaxLL, LL_defaultNormLog, workspace, workspaceSize);
 			LLtype = set_basic;
 		} else {
 			size_t nbSeq_1 = nbSeq;
@@ -662,7 +674,7 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 					return NCountSize;
 				op += NCountSize;
 			}
-			FSE_buildCTable_wksp(CTable_LitLength, norm, max, tableLog, scratchBuffer, sizeof(scratchBuffer));
+			FSE_buildCTable_wksp(CTable_LitLength, norm, max, tableLog, workspace, workspaceSize);
 			LLtype = set_compressed;
 		}
 	}
@@ -670,7 +682,7 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 	/* CTable for Offsets */
 	{
 		U32 max = MaxOff;
-		size_t const mostFrequent = FSE_countFast_wksp(count, &max, ofCodeTable, nbSeq, zc->tmpCounters);
+		size_t const mostFrequent = FSE_countFast_wksp(count, &max, ofCodeTable, nbSeq, workspace);
 		if ((mostFrequent == nbSeq) && (nbSeq > 2)) {
 			*op++ = ofCodeTable[0];
 			FSE_buildCTable_rle(CTable_OffsetBits, (BYTE)max);
@@ -678,7 +690,7 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 		} else if ((zc->flagStaticTables) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
 			Offtype = set_repeat;
 		} else if ((nbSeq < MIN_SEQ_FOR_DYNAMIC_FSE) || (mostFrequent < (nbSeq >> (OF_defaultNormLog - 1)))) {
-			FSE_buildCTable_wksp(CTable_OffsetBits, OF_defaultNorm, MaxOff, OF_defaultNormLog, scratchBuffer, sizeof(scratchBuffer));
+			FSE_buildCTable_wksp(CTable_OffsetBits, OF_defaultNorm, MaxOff, OF_defaultNormLog, workspace, workspaceSize);
 			Offtype = set_basic;
 		} else {
 			size_t nbSeq_1 = nbSeq;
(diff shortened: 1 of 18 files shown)


terrelln:
I'm going to reduce the stack usage of a few more functions.

contrib/linux-kernel/lib/zstd/huf_compress.c
```diff
@@ -174,15 +186,23 @@ size_t HUF_writeCTable(void *dst, size_t maxDstSize, const HUF_CElt *CTable, U32
...
 {
-	BYTE huffWeight[HUF_SYMBOLVALUE_MAX + 1];  /* init not required, even though some static analyzer may complain */
-	U32 rankVal[HUF_TABLELOG_ABSOLUTEMAX + 1]; /* large enough for values from 0 to 16 */
+	BYTE *huffWeight = (BYTE *)workspace;  /* init not required, even though some static analyzer may complain */
+	U32 *rankVal = (U32 *)(huffWeight + HUF_SYMBOLVALUE_MAX + 1); /* large enough for values from 0 to 16 */
 	U32 tableLog = 0;
 	U32 nbSymbols = 0;
+	size_t readSize;
+
+	if ((HUF_SYMBOLVALUE_MAX + 1) * (sizeof(U32) + sizeof(BYTE)) > workspaceSize)
```
terrelln: This is wrong, fixing it.

contrib/linux-kernel/lib/zstd/compress.c
```diff
@@ -587,8 +587,10 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
 {
 	const int longOffsets = zc->params.cParams.windowLog > STREAM_ACCUMULATOR_MIN;
 	const seqStore_t *seqStorePtr = &(zc->seqStore);
-	U32 count[MaxSeq + 1];
-	S16 norm[MaxSeq + 1];
+	U32 *count = (U32 *)zc->tmpCounters;
+	S16 *norm = (S16 *)(count + MaxSeq + 1);
```
Cyan4973: I would recommend to re-order space
to have `U32` first, followed by `U16`, then `BYTE`.
It naturally solves alignment risks.

contrib/linux-kernel/lib/zstd/huf_decompress.c
```diff
@@ -483,9 +491,15 @@ size_t HUF_readDTableX4(HUF_DTable *DTable, const void *src, size_t srcSize)
 	HUF_STATIC_ASSERT(sizeof(HUF_DEltX4) == sizeof(HUF_DTable)); /* if compiler fails here, assertion is wrong */
 	if (maxTableLog > HUF_TABLELOG_MAX)
 		return ERROR(tableLog_tooLarge);
+
+	workspace = rankStart0 + HUF_TABLELOG_MAX + 2;
+	if ((BYTE *)workspace > wkspEnd)
+		return ERROR(tableLog_tooLarge);
+  workspaceSize = wkspEnd - (BYTE *)workspace;
```
terrelln: Fixing formatting.

contrib/linux-kernel/lib/zstd/fse_decompress.c
```diff
@@ -288,16 +290,24 @@ size_t FSE_decompress_usingDTable(void *dst, size_t originalSize, const void *cS
...
-	short counting[FSE_MAX_SYMBOL_VALUE + 1];
+	FSE_DTable *dt = (FSE_DTable *)workspace;
+	short *counting = (short *)(dt + FSE_DTABLE_SIZE_U32(maxLog));
+	BYTE *wkspEnd = (BYTE *)workspace + workspaceSize;
 	unsigned tableLog;
 	unsigned maxSymbolValue = FSE_MAX_SYMBOL_VALUE;
+	size_t NCountLength;
+
+	workspace = counting + FSE_MAX_SYMBOL_VALUE + 1;
+	if ((BYTE *)workspace > wkspEnd)
```
Cyan4973: In theory, such comparison is UB, because to be true, `workspace` must be outside of valid memory object.
Would it be possible to convert this test
into a size comparison ?
Something like `if ( ((FSE_MAX_SYMBOL_VALUE+1)*sizeof(short)) > workspaceSize )` ?

contrib/linux-kernel/lib/zstd/huf_compress.c
```diff
@@ -87,11 +87,15 @@ size_t HUF_compressWeights(void *dst, size_t dstSize, const void *weightTable, s
...
-	BYTE scratchBuffer[1 << MAX_FSE_TABLELOG_FOR_HUFF_HEADER];
+	BYTE *wkspEnd = (BYTE *)workspace + workspaceSize;
+	FSE_CTable *CTable = (FSE_CTable *)workspace;
+	U32 *count = (U32 *)(CTable + FSE_CTABLE_SIZE_U32(MAX_FSE_TABLELOG_FOR_HUFF_HEADER, HUF_TABLELOG_MAX));
+	S16 *norm = (S16 *)(count + HUF_TABLELOG_MAX + 1);
 
-	U32 count[HUF_TABLELOG_MAX + 1];
-	S16 norm[HUF_TABLELOG_MAX + 1];
+	workspace = norm + HUF_TABLELOG_MAX + 1 + ((HUF_TABLELOG_MAX + 1) & 1);
+	if ((BYTE *)workspace > wkspEnd)
```
Cyan4973: same comment here, about size comparison

contrib/linux-kernel/lib/zstd/huf_decompress.c
```diff
@@ -86,20 +86,26 @@ typedef struct {
...
+	U32 *rankVal = (U32 *)(huffWeight + HUF_SYMBOLVALUE_MAX + 1); /* large enough for values from 0 to 16 */
+	BYTE *wkspEnd = (BYTE *)workspace + workspaceSize;
 	U32 tableLog = 0;
 	U32 nbSymbols = 0;
 	size_t iSize;
 	void *const dtPtr = DTable + 1;
 	HUF_DEltX2 *const dt = (HUF_DEltX2 *)dtPtr;
 
+	workspace = rankVal + HUF_TABLELOG_ABSOLUTEMAX + 1;
+	if ((BYTE *)workspace > wkspEnd)
```
Cyan4973: and same here

contrib/linux-kernel/lib/zstd/huf_decompress.c
```diff
@@ -486,11 +491,15 @@ size_t HUF_readDTableX4_wksp(HUF_DTable *DTable, const void *src, size_t srcSize
 	HUF_STATIC_ASSERT(sizeof(HUF_DEltX4) == sizeof(HUF_DTable)); /* if compiler fails here, assertion is wrong */
 	if (maxTableLog > HUF_TABLELOG_MAX)
 		return ERROR(tableLog_tooLarge);
-	if ((BYTE *)(rankVal + HUF_TABLELOG_MAX) > wkspEnd)
+
+	workspace = rankStart0 + HUF_TABLELOG_MAX + 2;
+	if ((BYTE *)workspace > wkspEnd)
```
Cyan4973: well...

contrib/linux-kernel/test/Makefile
```diff
@@ -5,21 +5,21 @@ SOURCES := $(wildcard ../lib/zstd/*.c)
 OBJECTS := $(patsubst %.c,%.o,$(SOURCES))
 
 ARFLAGS := rcs
-CXXFLAGS += -std=c++11
-CFLAGS += -g -O0
+CXXFLAGS += -std=c++11 -g -O3
+CFLAGS += -g -O3 -Wframe-larger-than=400
```
Cyan4973: why 400 ?
I expected linux kernel robot to test 1024 ?

contrib/linux-kernel/lib/zstd/compress.c
terrelln: Since the next usage is inside of `FSE_.*()` functions, to make that work here I would have to put `norm` at the end of the workspace. Do you think its worth it?

I'll definitely update the other places though.

Cyan4973: `-Wcast-align` is supposed to trigger a warning if one upgrade a `U16` pointer into a `U32` one,
even if you manually take good care of address alignment,
which is what happens below when `norm` get used as a base for `workspace`.

The order in memory can be different.
You can keep `norm` right after `count` if it's better,
but `workspace` will have to be based from `count`, because it is also a `U32`.

contrib/linux-kernel/test/Makefile
terrelln: Yeah it does, but the reviewer was concerned about stack usage, since the kernel stack is 16 KB at most. There were a few functions between 512 and 1024 bytes, so I fixed them up as well. After that, they were all below 400.

contrib/linux-kernel/lib/zstd/compress.c
```diff
@@ -587,8 +587,10 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
...
+	U32 *const workspace = PTR_ALIGN((U32 *)(norm + MaxSeq + 1), sizeof(U32));
```
Cyan4973: I know this is supposed to work, since `PTR_ALIGN` makes a good job at properly aligning the resulting ptr,
but still, from a compiler perspective, promoting a `U16*` into a `U32*` is ground for a `-Wcast-align` warning :
https://godbolt.org/g/vy7fR3

contrib/linux-kernel/lib/zstd/huf_decompress.c
```diff
@@ -86,20 +87,33 @@ typedef struct {
 	BYTE nbBits;
 } HUF_DEltX2; /* single-symbol decoding */
 
-size_t HUF_readDTableX2(HUF_DTable *DTable, const void *src, size_t srcSize)
+size_t HUF_readDTableX2_wksp(HUF_DTable *DTable, const void *src, size_t srcSize, void *workspace, size_t workspaceSize)
 {
-	BYTE huffWeight[HUF_SYMBOLVALUE_MAX + 1];
-	U32 rankVal[HUF_TABLELOG_ABSOLUTEMAX + 1]; /* large enough for values from 0 to 16 */
```
Cyan4973: minor : `HUF_TABLELOG_ABSOLUTEMAX` is quite small, it shouldn't worry stack size

contrib/linux-kernel/lib/zstd/huf_decompress.c
```diff
@@ -465,27 +480,49 @@ static void HUF_fillDTableX4(HUF_DEltX4 *DTable, const U32 targetLog, const sort
...
+	spaceUsed += sizeof(rankValCol_t) * HUF_TABLELOG_MAX;
+	rankStats = (U32 *)((BYTE *)workspace + spaceUsed);
+	spaceUsed += sizeof(U32) * HUF_TABLELOG_MAX + 1;
+	rankStart0 = (U32 *)((BYTE *)workspace + spaceUsed);
+	spaceUsed += sizeof(U32) * HUF_TABLELOG_MAX + 2;
+	sortedSymbol = (sortedSymbol_t *)((BYTE *)workspace + spaceUsed);
+	spaceUsed += sizeof(sortedSymbol_t) * (HUF_SYMBOLVALUE_MAX + 1);
+	weightList = (BYTE *)workspace + spaceUsed;
+	spaceUsed += HUF_SYMBOLVALUE_MAX + 1;
+	spaceUsed = ALIGN(spaceUsed, sizeof(U32));
```
Cyan4973: 👍

contrib/linux-kernel/lib/zstd/huf_compress.c
```diff
@@ -174,15 +200,29 @@ size_t HUF_writeCTable(void *dst, size_t maxDstSize, const HUF_CElt *CTable, U32
 	return ((maxSymbolValue + 1) / 2) + 1;
 }
 
-size_t HUF_readCTable(HUF_CElt *CTable, U32 maxSymbolValue, const void *src, size_t srcSize)
+size_t HUF_readCTable_wksp(HUF_CElt *CTable, U32 maxSymbolValue, const void *src, size_t srcSize, void *workspace, size_t workspaceSize)
 {
-	BYTE huffWeight[HUF_SYMBOLVALUE_MAX + 1];  /* init not required, even though some static analyzer may complain */
-	U32 rankVal[HUF_TABLELOG_ABSOLUTEMAX + 1]; /* large enough for values from 0 to 16 */
```
Cyan4973: minor : `HUF_TABLELOG_ABSOLUTEMAX` is quite small, it shouldn't worry stack size

contrib/linux-kernel/lib/zstd/compress.c
terrelln: Yeah, wrote this before I saw your second comment. I've fixed it so that it doesn't use the `U16 *` to derive the `U32 *`.

contrib/linux-kernel/lib/zstd/huf_compress.c
terrelln: Yeah, but since we are already passing the workspace for `huffWeight` and `HUF_readStats_wksp()`, we might as well take advantage of it to reduce the stack a bit further.

terrelln:
The patch is ready for review again. I've fixed the bugs in the stack computations that my fuzzers revealed. The fuzzers are still running.

contrib/linux-kernel/lib/zstd/compress.c
```diff
@@ -602,7 +600,22 @@ ZSTD_STATIC size_t ZSTD_compressSequences(ZSTD_CCtx *zc, void *dst, size_t dstCa
...
+	size_t workspaceSize = sizeof(zc->tmpCounters);
+	{
+		size_t spaceUsed = 0;
+		count = (U32 *)((BYTE *)zc->tmpCounters + spaceUsed);
+		spaceUsed += sizeof(U32) * (MaxSeq + 1);
+		norm = (S16 *)((BYTE *)zc->tmpCounters + spaceUsed);
+		spaceUsed += sizeof(S16) * (MaxSeq + 1);
+		spaceUsed = ALIGN(spaceUsed, sizeof(U32));
+
+		workspace = (U32 *)((BYTE *)zc->tmpCounters + spaceUsed);
```
Cyan4973: If a `U32*` ptr is created from a `BYTE*` ptr,
even if the initial source was a `U32*`,
it's still a concern for `-Wcast-align`.

See "doubleCast" example :
https://godbolt.org/g/qrWbLq

Solution : express `spaceUsed` in `U32` increment.

Cyan4973:
There are still some alignment promotion issues in many places in the code.
I've only commented one of them, but the same concern and solution is applicable to the rest of the code.

PR #739, opened 2017-06-28
ZSTD_refPrefix

Cyan4973:
This patch implements `ZSTD_refPrefix()` and `ZSTD_p_refDictContent`,
completing the advanced API to its intended target for release.

It's built on top of `simplified *initStatic*()`


lib/compress/zstd_compress.c
```diff
@@ -3902,7 +3937,10 @@ size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
         } else
 #endif
         {
-            CHECK_F( ZSTD_resetCStream_internal(cctx, params, cctx->pledgedSrcSizePlusOne-1) );
+            const void* const prefix = cctx->prefix;
+            size_t const prefixSize = cctx->prefixSize;
+            cctx->prefix = NULL; cctx->prefixSize = 0;   /* single usage */
```
terrelln: Is this comment accurate? It looks like `ZSTD_resetCStream_internal()` will reuse the prefix.

tests/zstreamtest.c
```diff
@@ -210,9 +210,9 @@ static int basicUnitTests(U32 seed, double compressibility, ZSTD_customMem custo
     /* context size functions */
```
terrelln: Do we have a tests that make sure the prefix is only used once for block/streaming compression?

Cyan4973: not yet.
The current fuzzer test always reset dictionary to `{NULL,0}` before each new frame, so we can't verify this property.

It's easier to add such test in unit tests.
One way to do it would be 
to make a new compression after a `ZSTD_refPrefix()` one,
and check that decompression works without loading a dictionary.

lib/compress/zstd_compress.c
```diff
@@ -3924,29 +3924,30 @@ size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
...
     if (cctx->streamStage == zcss_init) {
         /* transparent reset */
+        const void* const prefix = cctx->prefix;
+        size_t const prefixSize = cctx->prefixSize;
         ZSTD_parameters params = cctx->requestedParams;
         if (cctx->compressionLevel != ZSTD_CLEVEL_CUSTOM)
             params.cParams = ZSTD_getCParams(cctx->compressionLevel,
-                                    cctx->pledgedSrcSizePlusOne-1, 0 /* dictSize */);
+                                cctx->pledgedSrcSizePlusOne-1, 0 /*dictSize*/);
+        cctx->prefix = NULL; cctx->prefixSize = 0;   /* single usage */
```
terrelln: Oh I understand how it works now.

lib/compress/zstd_compress.c
Cyan4973: yes it is

PR #740, opened 2017-06-28
decodecorpus.c patch

paulcruz74:
- fixed generation of test cases where dictionary header is overriding dictionary content
- fixed `DISPLAY` statements for errors to provide useful `seed` and `fnum`


tests/decodecorpus.c
```diff
@@ -1392,7 +1394,7 @@ static int runTestMode(U32 seed, unsigned numFiles, unsigned const testDurationS
 
     for (fnum = 0; fnum < numFiles || clockSpan(startClock) < maxClockSpan; fnum++) {
         frame_t fr;
-
+        U32 const seedCopy = seed;
```
Cyan4973: OK, looks good.
So, if I understand properly,
`seed` is used (and modified) only once per round,
to generate a frame.

By displaying the current round's `seed`,
it's possible to jump directly to said `seed` in future tests, using `-s#` command,
in order to reproduce an error case.

That would make `fnum` non-essential, since it's not a useful information to jump directly to the wanted test case.

tests/decodecorpus.c
```diff
@@ -1216,8 +1216,8 @@ static int genRandomDict(U32 dictID, U32 seed, size_t dictSize, BYTE* fullDict){
         /* create variables */
         size_t dictWriteSize = 0;
         ZDICT_params_t zdictParams;
-        size_t const headerSize = dictSize/4;
-        size_t const dictContentSize = dictSize - dictSize/4;
+        size_t const headerSize = MAX(dictSize/4, 256);
```
Cyan4973: 👍

Cyan4973:
Good fix !

PR #742, opened 2017-06-30
Reduce stack usage of HUF_readDTableX4 and HUF_readDTableX2

stellamplau:


lib/decompress/huf_decompress.c
```diff
@@ -67,6 +67,12 @@
 #define HUF_STATIC_ASSERT(c) { enum { HUF_static_assert = 1/(int)(!!(c)) }; }   /* use only *after* variable declarations */
 
 
+/* **************************************************************
+*  Byte alignment for workSpace management
+****************************************************************/
+#define ALIGN(x, a) ALIGN_MASK((x), (a) - 1)
```
Cyan4973: namespace : `ALIGN` name risks being too generic, it may collide with another macro using same name into a 3rd-party project.
Prefer `HUF_ALIGN` and `HUF_ALIGN_MASK`.

lib/common/huf.h
```diff
@@ -111,6 +111,9 @@ HUF_PUBLIC_API size_t HUF_compress2 (void* dst, size_t dstCapacity, const void*
 #define HUF_WORKSPACE_SIZE_U32 (HUF_WORKSPACE_SIZE / sizeof(U32))
 HUF_PUBLIC_API size_t HUF_compress4X_wksp (void* dst, size_t dstCapacity, const void* src, size_t srcSize, unsigned maxSymbolValue, unsigned tableLog, void* workSpace, size_t wkspSize);
 
+/* The workspace must have alignment at least 4 and be at least this large */
+#define HUF_DECOMPRESS_WORKSPACE_SIZE (2 << 10)
```
Cyan4973: In which case are 2 KB necessary ?
Is that an over-estimate ?

stellamplau: I think Nick calculated this to be ~1500, @terrelln?

lib/decompress/huf_decompress.c
```diff
@@ -404,6 +450,7 @@ static void HUF_fillDTableX4Level2(HUF_DEltX4* DTable, U32 sizeLog, const U32 co
 }
 
 typedef U32 rankVal_t[HUF_TABLELOG_MAX][HUF_TABLELOG_MAX + 1];
```
Cyan4973: Maybe we should redefine `rankVal_t` as a table of `rankValCol_t`.
That would make the relation more obvious.

lib/common/huf.h
Cyan4973: Ah, I guess it comes from Linux's kbuild robot : 
https://lkml.org/lkml/2017/6/25/65

There, `HUF_readDTableX4()` is estimated at 1636 bytes, which is largest stack usage.
So 2KB seems to be an overestimate.

Are we comfortable with ~400 bytes overestimate ?
We could reduce that, by using `7<<8` instead, which would reduce size by 256 bytes.
To control if it's good enough, one should increase `HUF_TABLELOG_MAX` value, up to its maximum `HUF_TABLELOG_ABSOLUTEMAX`, as some tables will increase size as a result, in order to control if everything still fits.

Of course, the best solution would be to have the actual formula calculating `HUF_DECOMPRESS_WORKSPACE_SIZE` from `HUF_TABLELOG_MAX` (which is the only constant value that may change).

stellamplau: From some quick calculations, with `HUF_TABLELOG_MAX` set to `HUF_TABLELOG_ABSOLUTE_MAX`, ~1860 bytes is needed (so 7<<8 would not be enough).  with its current value, ~1500 bytes is needed. 

Do you think it's worth having the actual formula?

Cyan4973: As long as the code remains stable, the approximation feels good enough.

Thing is, if, in the future, the code is changed, by another programmer,
it might break the limit, unintentionally.

This should be correctly detected by the runtime test, so I don't worry about buffer overflow.
Just, the error message is not so clear. So debugging may take a while.

Maybe some additional comments regarding the size of workspace, where it's used, how much is actually spent, and potential risks if code is changed, would be welcomed, and be good enough for the time being.

Complementary note : 
I believe the proper way to deal with this situation would be to completely define these workspaces, using one structure per usage context, defining all necessary tables (since they have fixed size), and then join them into a union. 
Then, workspace size would be automatically derived from union size. And there would be no need to actually play with a ptr to define multiple areas : just go straight using the appropriate structure.

If it's too much work, we could keep that action for another day.

stellamplau: Your note makes sense. For now, I have added some additional comments.

Cyan4973:
No more comments on my side.
Ready to merge if it's good enough for @terrelln , and after CI tests completion

terrelln:
Looks good to me!

Cyan4973:
First merge for 👑  @stellamplau 👑  
🎉

PR #743, opened 2017-07-01
compress_generic() automatic optimization opportunities

Cyan4973:
`ZSTD_compress_generic()` can automatically detect it is in "single pass" scenario when provided with right set of conditions. This improves performance (~+15%) as well as reduces memory requirements.

The optimization has also been ported to Multi-threading mode, where it works very well with "small" sources (a few MB). `ZSTDMT_compress()` has also been updated to perform better with "large" sources (several hundreds MB), as previous method produced too large jobs, and overall speed was limited by slowest job. The new strategy is providing a nice boost to compression speed (~+15%), and `ZSTD_compress_generic()` benefits it.

Also : created `ZSTDMT_compress_advanced()` which provides a more complete support for advanced scenarios, such as custom compression parameters and dictionary.

There is one (potential) downside :
In single-thread mode, API is always "blocking", it gives back control to caller when it's done.
By contrast, in multi-thread mode, it depends : streaming API is asynchronous, meaning it gives back control to caller almost immediately while working in a separate thread, but single-pass mode `ZSTDMT_compress()` is blocking.
As a consequence, when `ZSTD_compress_generic()` uses automatic optimisation opportunity, it silently moves from asynchronous (streaming) to blocking mode.
These 2 behaviors are very different.
I suspect it would be better to make them "explicit".
That's probably a battle for another day.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2954,6 +2954,8 @@ static size_t ZSTD_compressContinue_internal (ZSTD_CCtx* cctx,
     const BYTE* const ip = (const BYTE*) src;
     size_t fhSize = 0;
 
+    DEBUGLOG(5, "ZSTD_compressContinue_internal");
+    DEBUGLOG(5, "stage: %u", cctx->stage);
     if (cctx->stage==ZSTDcs_created) return ERROR(stage_wrong);   /* missing init (ZSTD_compressBegin) */
 
     if (frame && (cctx->stage==ZSTDcs_init)) {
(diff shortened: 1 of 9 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -526,8 +550,10 @@ size_t ZSTDMT_compressCCtx(ZSTDMT_CCtx* mtctx,
                 if (chunkID) {   /* note : chunk 0 is already written directly into dst */
                     if (!error)
                         memmove((char*)dst + dstPos, mtctx->jobs[chunkID].dstBuff.start, cSize);  /* may overlap if chunk decompressed within dst */
-                    if (chunkID >= compressWithinDst)   /* otherwise, it decompresses within dst */
+                    if (chunkID >= compressWithinDst) {  /* otherwise, it decompresses within dst */
```
terrelln: minor: Why do the comments say decompress?

lib/compress/zstd_compress.c
```diff
@@ -3796,6 +3801,20 @@ static size_t ZSTD_compressStream_generic(ZSTD_CStream* zcs,
             return ERROR(init_missing);
 
         case zcss_load:
+            if ( (flushMode == ZSTD_e_end)
+              && ((size_t)(oend-op) >= ZSTD_compressBound(iend-ip))
+              && (zcs->inBuffPos == 0) ) {
```
Cyan4973: Nope. If `inbuffPos == inToCompress`, it merely means we have no data stored within internal buffer to compress another block. But previous data can still span 2 segments, which is something we want to avoid : in such case, the compressor would automatically discard the oldest segment, keeping only the newest one in memory. This would negatively impact compression ratio.
And we want this optimization to have no impact on compression ratio (it's merely an _opportunity_).

With `inbuffPos==0`, we are guaranteed that, either there is nothing yet within internal buffer (most expected scenario), or, we have reached the top of internal round buffer and restart from the beginning, in which case past history is necessarily a single segment, which can be referenced by `ZSTD_compressEnd()`.

There is another case which can be valid, though it's not done here and is more complex to describe :
- next block must lay in a single segment (`inbuffPos == inToCompress`)
- history must be a single segment (`lowLimit == dictLimit`)

Technically, it can only happen while filling internal round buffer during the first pass (srcSize < windowSize).
After that point, history is always separated into 2 segments, except when `inbuffPos==0`.

Not sure yet if this use case is common enough to be worth implementing.

Cyan4973:
All points answered and fixed

PR #745, opened 2017-07-01
[fuzz] Add libFuzzer targets

terrelln:
* Add 4 fuzz targets, `simple_round_trip`, `simple_decompress`, `stream_round_trip`, and `stream_decompress`. I don't yet have coverage for dictionaries.
* The regression driver serves both as a regression test, and as a binary for afl-fuzz.
* Next steps is to check in corpora for each target, and add regression tests to Travis or Circle CI.


Cyan4973:
This is a great addition to the repo @terrelln .
I've got 2 questions : 
- Proposed directory is `/fuzz`, directly under root. Could it be under `/tests/fuzz` instead ?
- Mentioned the objective to run fuzzer tests on Travis / CircleCI. Fuzzer testing tends to be long, aka incompatible with these systems time-out policy. So, how do you see this part playing ?

terrelln:
* Sounds good
* We won't run the fuzzer tests on Travis. We will run the fuzz target on each input, without fuzzing, to make sure that none of the checked in inputs cause it to fail. We can check in all the inputs that caused failures in the past, as well as a minimal corpus that gets all the coverage we have found so far, so we get more regression testing.

Cyan4973:
Sounds all good !
So, just a directory change, and it's a merge

PR #747, opened 2017-07-03
No exit() on decoding errors

Cyan4973:
`fileio.c` used to `exit()` on decoding errors, which is a problem when decoding is ordered on a bunch of files (for example `zstd -t *`).

A previous patch added the ability to continue processing the list of files in case of `zstd` decoding errors. This patch extends this capability to `gzip`, `lzma` and `lz4` formats.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1344,8 +1344,11 @@ static unsigned FIO_passThrough(FILE* foutput, FILE* finput, void* buffer, size_
     unsigned storedSkips = 0;
 
     /* assumption : ress->srcBufferLoaded bytes already loaded and stored within buffer */
-    { size_t const sizeCheck = fwrite(buffer, 1, alreadyLoaded, foutput);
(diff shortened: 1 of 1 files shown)


programs/fileio.c
```diff
@@ -1463,9 +1483,11 @@ static unsigned long long FIO_decompressGzFrame(dRess_t* ress,
 
     if (strm.avail_in > 0) memmove(ress->srcBuffer, strm.next_in, strm.avail_in);
     ress->srcBufferLoaded = strm.avail_in;
-    ret = inflateEnd(&strm);
-    if (ret != Z_OK) EXM_THROW(32, "zstd: %s: inflateEnd error %d", srcFileName, ret);
-    return outFileSize;
+    if (inflateEnd(&strm) != Z_OK) {
+        DISPLAYLEVEL(1, "zstd: %s: inflateEnd error \n", srcFileName);
```
terrelln: We should probably only display this if `decodingError == 0`, since `inflateEnd()` will return failure if the stream is incomplete.

PR #753, opened 2017-07-10
adaptive compression v1

paulcruz74:
- compression level adapts based on which thread is waiting
- 3 separate threads in charge of creating jobs, compression, and writing

### DIFF
--- a/contrib/adaptive-compression/Makefile
+++ b/contrib/adaptive-compression/Makefile
@@ -24,6 +24,9 @@ all: adapt datagen
 adapt: $(ZSTD_FILES) adapt.c
 	$(CC) $(FLAGS) $^ -o $@
 
+adapt-debug: $(ZSTD_FILES) adapt.c
+	$(CC) $(FLAGS) -DDEBUG_MODE=2 $^ -o adapt
+
 datagen : $(PRGDIR)/datagen.c datagencli.c
 	$(CC)      $(FLAGS) $^ -o $@$(EXT)
 
--- a/contrib/adaptive-compression/adapt.c
+++ b/contrib/adaptive-compression/adapt.c
@@ -924,6 +924,7 @@ static void help()
     PRINT("  -i#    : provide initial compression level\n");
     PRINT("  -h     : display help/information\n");
     PRINT("  -f     : force the compression level to stay constant\n");
+    PRINT("  -c     : force write to stdout\n");
     PRINT("  -p     : hide progress bar\n");
     PRINT("  -q     : quiet mode -- do not show progress bar or other information\n");
 }
--- a/contrib/adaptive-compression/adapt.c
+++ b/contrib/adaptive-compression/adapt.c
@@ -434,44 +434,66 @@ static void* compressionThread(void* arg)
 
         /* adapt compression level */
         adaptCompressionLevel(ctx);
-        
+
         /* compress the data */
         {
+            size_t const compressionBlockSize = 4 << 17; /* 128 KB */
             unsigned const cLevel = ctx->compressionLevel;
+            unsigned blockNum = 0;
+            size_t remaining = job->src.size;
+            size_t srcPos = 0;
+            size_t dstPos = 0;
+            size_t dictPos = 0;
             DEBUG(3, "cLevel used: %u\n", cLevel);
             DEBUG(3, "compression level used: %u\n", cLevel);
-            /* begin compression */
-            {
-                size_t const useDictSize = MIN(getUseableDictSize(cLevel), job->dictSize);
-                DEBUG(3, "useDictSize: %zu, job->dictSize: %zu\n", useDictSize, job->dictSize);
-                size_t const dictModeError = ZSTD_setCCtxParameter(ctx->cctx, ZSTD_p_forceRawDict, 1);
-                size_t const initError = ZSTD_compressBegin_usingDict(ctx->cctx, job->src.start + job->dictSize - useDictSize, useDictSize, cLevel);
-                size_t const windowSizeError = ZSTD_setCCtxParameter(ctx->cctx, ZSTD_p_forceWindow, 1);
-                if (ZSTD_isError(dictModeError) || ZSTD_isError(initError) || ZSTD_isError(windowSizeError)) {
-                    DISPLAY("Error: something went wrong while starting compression\n");
-                    signalErrorToThreads(ctx);
-                    return arg;
-                }
-            }
 
-            /* continue compression */
-            if (currJob != 0) { /* not first job flush/overwrite the frame header */
-                size_t const hSize = ZSTD_compressContinue(ctx->cctx, job->dst.start, job->dst.capacity, job->src.start + job->dictSize, 0);
-                if (ZSTD_isError(hSize)) {
-                    DISPLAY("Error: something went wrong while continuing compression\n");
-                    job->compressedSize = hSize;
-                    signalErrorToThreads(ctx);
-                    return arg;
+            /* reset compressed size */
+            job->compressedSize = 0;
+
+            while (remaining != 0) {
+                size_t const actualBlockSize = MIN(remaining, compressionBlockSize);
+                /* begin compression */
+                {
+                    size_t const useDictSize = MIN(getUseableDictSize(cLevel), job->dictSize);
+                    DEBUG(3, "useDictSize: %zu, job->dictSize: %zu\n", useDictSize, job->dictSize);
+                    size_t const dictModeError = ZSTD_setCCtxParameter(ctx->cctx, ZSTD_p_forceRawDict, 1);
+                    size_t const initError = ZSTD_compressBegin_usingDict(ctx->cctx, job->src.start + job->dictSize - useDictSize + dictPos, useDictSize, cLevel);
+                    size_t const windowSizeError = ZSTD_setCCtxParameter(ctx->cctx, ZSTD_p_forceWindow, 1);
+                    if (ZSTD_isError(dictModeError) || ZSTD_isError(initError) || ZSTD_isError(windowSizeError)) {
+                        DISPLAY("Error: something went wrong while starting compression\n");
+                        signalErrorToThreads(ctx);
+                        return arg;
+                    }
                 }
-                ZSTD_invalidateRepCodes(ctx->cctx);
-            }
-            job->compressedSize = (job->lastJob) ?
-                                    ZSTD_compressEnd     (ctx->cctx, job->dst.start, job->dst.capacity, job->src.start + job->dictSize, job->src.size) :
-                                    ZSTD_compressContinue(ctx->cctx, job->dst.start, job->dst.capacity, job->src.start + job->dictSize, job->src.size);
-            if (ZSTD_isError(job->compressedSize)) {
-                DISPLAY("Error: something went wrong during compression: %s\n", ZSTD_getErrorName(job->compressedSize));
-                signalErrorToThreads(ctx);
-                return arg;
+
+                /* continue compression */
+                if (currJob != 0 || blockNum != 0) { /* not first block of first job flush/overwrite the frame header */
+                    size_t const hSize = ZSTD_compressContinue(ctx->cctx, job->dst.start + dstPos, job->dst.capacity - dstPos, job->src.start + job->dictSize + srcPos, 0);
+                    if (ZSTD_isError(hSize)) {
+                        DISPLAY("Error: something went wrong while continuing compression\n");
+                        job->compressedSize = hSize;
+                        signalErrorToThreads(ctx);
+                        return arg;
+                    }
+                    ZSTD_invalidateRepCodes(ctx->cctx);
+                }
+                {
+
+                    size_t const ret = (job->lastJob && remaining <= compressionBlockSize) ?
+                                            ZSTD_compressEnd     (ctx->cctx, job->dst.start + dstPos, job->dst.capacity - dstPos, job->src.start + job->dictSize + srcPos, actualBlockSize) :
+                                            ZSTD_compressContinue(ctx->cctx, job->dst.start + dstPos, job->dst.capacity - dstPos, job->src.start + job->dictSize + srcPos, actualBlockSize);
+                    if (ZSTD_isError(ret)) {
+                        DISPLAY("Error: something went wrong during compression: %s\n", ZSTD_getErrorName(ret));
+                        signalErrorToThreads(ctx);
+                        return arg;
+                    }
+                    job->compressedSize += ret;
+                    remaining -= actualBlockSize;
+                    srcPos += actualBlockSize;
+                    dstPos += ret;
+                    dictPos += actualBlockSize;
+                    blockNum++;
+                }
             }
             job->dst.size = job->compressedSize;
         }
--- a/contrib/adaptive-compression/adapt.c
+++ b/contrib/adaptive-compression/adapt.c
@@ -88,7 +88,9 @@ typedef struct {
     unsigned allJobsCompleted;
     unsigned adaptParam;
     unsigned compressionCompletionMeasured;
+    unsigned writeCompletionMeasured;
     double compressionCompletion;
+    double writeCompletion;
     mutex_t jobCompressed_mutex;
     cond_t jobCompressed_cond;
     mutex_t jobReady_mutex;
@@ -97,6 +99,8 @@ typedef struct {
     cond_t allJobsCompleted_cond;
     mutex_t jobWrite_mutex;
     cond_t jobWrite_cond;
+    mutex_t completion_mutex;
+    mutex_t stats_mutex;
     size_t lastDictSize;
     inBuff_t input;
     cStat_t stats;
@@ -156,6 +160,8 @@ static int freeCCtx(adaptCCtx* ctx)
         error |= destroyCond(&ctx->allJobsCompleted_cond);
         error |= destroyMutex(&ctx->jobWrite_mutex);
         error |= destroyCond(&ctx->jobWrite_cond);
+        error |= destroyMutex(&ctx->completion_mutex);
+        error |= destroyMutex(&ctx->stats_mutex);
         error |= ZSTD_isError(ZSTD_freeCCtx(ctx->cctx));
         free(ctx->input.buffer.start);
         if (ctx->jobs){
@@ -200,6 +206,8 @@ static adaptCCtx* createCCtx(unsigned numJobs)
         pthreadError |= initCond(&ctx->allJobsCompleted_cond);
         pthreadError |= initMutex(&ctx->jobWrite_mutex);
         pthreadError |= initCond(&ctx->jobWrite_cond);
+        pthreadError |= initMutex(&ctx->completion_mutex);
+        pthreadError |= initMutex(&ctx->stats_mutex);
         if (pthreadError) return NULL;
     }
     ctx->numJobs = numJobs;
@@ -315,24 +323,44 @@ static void adaptCompressionLevel(adaptCCtx* ctx)
         }
         else if ((writeSlow || createSlow) && ctx->compressionLevel < (unsigned)ZSTD_maxCLevel()) {
             DEBUG(3, "increasing compression level %u\n", ctx->compressionLevel);
-            ctx->compressionLevel++;
-            reset = 1;
+            double completion;
+            pthread_mutex_lock(&ctx->completion_mutex.pMutex);
+            completion = ctx->writeCompletion;
+            pthread_mutex_unlock(&ctx->completion_mutex.pMutex);
+            {
+                unsigned const maxChange = (unsigned)((1-completion) * (MAX_COMPRESSION_LEVEL_CHANGE - 1)) + 1;
+                unsigned const change = writeSlow ? MIN(maxChange, ZSTD_maxCLevel() - ctx->compressionLevel) : 1;
+                DEBUG(2, "writeSlow: %u, change: %u\n", writeSlow, change);
+                DEBUG(2, "write completion: %f\n", completion);
+                ctx->compressionLevel += change;
+                reset = 1;
+            }
         }
         else if (compressSlow && ctx->compressionLevel > 1) {
-            double const completion = ctx->compressionCompletion;
-            unsigned const maxChange = (unsigned)((1-completion) * (MAX_COMPRESSION_LEVEL_CHANGE-1)) + 1;
-            unsigned const change = MIN(maxChange, ctx->compressionLevel - 1);
-            DEBUG(3, "decreasing compression level %u\n", ctx->compressionLevel);
-            DEBUG(3, "completion: %f\n", completion);
-            ctx->compressionLevel -= change;
-            reset = 1;
+            double completion;
+            pthread_mutex_lock(&ctx->completion_mutex.pMutex);
+            completion = ctx->compressionCompletion;
+            pthread_mutex_unlock(&ctx->completion_mutex.pMutex);
+            {
+                unsigned const maxChange = (unsigned)((1-completion) * (MAX_COMPRESSION_LEVEL_CHANGE-1)) + 1;
+                unsigned const change = MIN(maxChange, ctx->compressionLevel - 1);
+                DEBUG(3, "decreasing compression level %u\n", ctx->compressionLevel);
+                DEBUG(3, "completion: %f\n", completion);
+                ctx->compressionLevel -= change;
+                reset = 1;
+            }
         }
         if (reset) {
             ctx->stats.readyCounter = 0;
             ctx->stats.writeCounter = 0;
             ctx->stats.compressedCounter = 0;
+
+            pthread_mutex_lock(&ctx->completion_mutex.pMutex);
             ctx->compressionCompletion = 1;
             ctx->compressionCompletionMeasured = 0;
+            ctx->writeCompletion = 1;
+            ctx->writeCompletionMeasured = 0;
+            pthread_mutex_unlock(&ctx->completion_mutex.pMutex);
         }
     }
 }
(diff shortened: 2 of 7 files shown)


contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+
+static void displayProgress(unsigned jobDoneID, unsigned cLevel, unsigned last)
+{
+    if (!g_useProgressBar) return;
+    UTIL_time_t currTime;
+    UTIL_getTime(&currTime);
+    double const timeElapsed = (double)(UTIL_getSpanTimeMicro(g_ticksPerSecond, g_startTime, currTime) / 1000.0);
+    double const sizeMB = (double)g_streamedSize / (1 << 20);
+    double const avgCompRate = sizeMB * 1000 / timeElapsed;
+    fprintf(stdout, "\r| %4u jobs completed | Current Compresion Level: %2u | Time Elapsed: %5.0f ms | Data Size: %7.1f MB | Avg Compression Rate: %6.2f MB/s |", jobDoneID, cLevel, timeElapsed, sizeMB, avgCompRate);
```
paulcruz74: I'm writing to `stdout` for the progress bar, but in cases where the user is also writing to `stdout` this creates a problem.

Cyan4973: Yes,
always write accessory notifications to `stderr`,
`stdout` is here reserved for the compressed data itself.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+static void help()
+{
+    PRINT("Usage:\n");
+    PRINT("  ./multi [options] [file(s)]\n");
+    PRINT("\n");
+    PRINT("Options:\n");
+    PRINT("  -oFILE : specify the output file name\n");
+    PRINT("  -v     : display debug information\n");
+    PRINT("  -i#    : provide initial compression level\n");
+    PRINT("  -s     : display information stats\n");
```
Cyan4973: I would recommend `-v` for information stats, and `-vv` for debug

contrib/adaptive-compression/Makefile
```diff
@@ -0,0 +1,33 @@
...
+CFLAGS   += -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow                 \
+            -Wstrict-aliasing=1 -Wswitch-enum -Wdeclaration-after-statement \
+            -Wstrict-prototypes -Wundef -Wformat-security                   \
+            -Wvla -Wformat=2 -Winit-self -Wfloat-equal -Wwrite-strings      \
+            -Wredundant-decls
+CFLAGS   += $(DEBUGFLAGS)
+CFLAGS   += $(MOREFLAGS)
+FLAGS     = $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
+
+all: clean adapt
```
terrelln: minor: Remove `clean`.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
+#define DISPLAY(...) fprintf(stderr, __VA_ARGS__)
```
terrelln: You should put the `include`s first, and you need to add the copyright/license header.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+#define DEBUGLOG(l, ...) { if (g_displayLevel>=l) { DISPLAY(__VA_ARGS__); } }
+#define FILE_CHUNK_SIZE 4 << 20
+#define MAX_NUM_JOBS 2;
+#define stdinmark  "/*stdin*\\"
+#define stdoutmark "/*stdout*\\"
+#define MAX_PATH 256
+#define DEFAULT_DISPLAY_LEVEL 1
+#define DEFAULT_COMPRESSION_LEVEL 6
+#define DEFAULT_ADAPT_PARAM 1
+typedef unsigned char BYTE;
```
terrelln: You should `#include "mem.h"` to get `BYTE`.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+        int const allJobsCondError = pthread_cond_destroy(&ctx->allJobsCompleted_cond);
+        int const jobWriteMutexError = pthread_mutex_destroy(&ctx->jobWrite_mutex);
+        int const jobWriteCondError = pthread_cond_destroy(&ctx->jobWrite_cond);
+        int const fileCloseError =  (ctx->dstFile != NULL && ctx->dstFile != stdout) ? fclose(ctx->dstFile) : 0;
+        int const cctxError = ZSTD_isError(ZSTD_freeCCtx(ctx->cctx)) ? 1 : 0;
+        if (ctx->jobs){
+            freeCompressionJobs(ctx);
+            free(ctx->jobs);
+        }
+        return compressedMutexError | compressedCondError | readyMutexError | readyCondError | fileCloseError | allJobsMutexError | allJobsCondError | jobWriteMutexError | jobWriteCondError | cctxError;
```
terrelln: Break this line up, it is very long

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+        int const compressedMutexError = pthread_mutex_destroy(&ctx->jobCompressed_mutex);
+        int const compressedCondError = pthread_cond_destroy(&ctx->jobCompressed_cond);
+        int const readyMutexError = pthread_mutex_destroy(&ctx->jobReady_mutex);
+        int const readyCondError = pthread_cond_destroy(&ctx->jobReady_cond);
+        int const allJobsMutexError = pthread_mutex_destroy(&ctx->allJobsCompleted_mutex);
+        int const allJobsCondError = pthread_cond_destroy(&ctx->allJobsCompleted_cond);
+        int const jobWriteMutexError = pthread_mutex_destroy(&ctx->jobWrite_mutex);
+        int const jobWriteCondError = pthread_cond_destroy(&ctx->jobWrite_cond);
+        int const fileCloseError =  (ctx->dstFile != NULL && ctx->dstFile != stdout) ? fclose(ctx->dstFile) : 0;
+        int const cctxError = ZSTD_isError(ZSTD_freeCCtx(ctx->cctx)) ? 1 : 0;
```
terrelln:
```c
int const cctxError = ZSTD_isError(ZSTD_freeCCtx(ctx->cctx));
```

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+            free(ctx->jobs);
+        }
+        return compressedMutexError | compressedCondError | readyMutexError | readyCondError | fileCloseError | allJobsMutexError | allJobsCondError | jobWriteMutexError | jobWriteCondError | cctxError;
+    }
+}
+
+static adaptCCtx* createCCtx(unsigned numJobs, const char* const outFilename)
+{
+
+    adaptCCtx* ctx = malloc(sizeof(adaptCCtx));
```
terrelln: Use `calloc()` instead, it is more efficient and standard than `malloc()` followed by `memset()`.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+    pthread_mutex_lock(&ctx->allJobsCompleted_mutex);
+    while (ctx->allJobsCompleted == 0) {
+        pthread_cond_wait(&ctx->allJobsCompleted_cond, &ctx->allJobsCompleted_mutex);
+    }
+    pthread_mutex_unlock(&ctx->allJobsCompleted_mutex);
+}
+
+static unsigned adaptCompressionLevel(adaptCCtx* ctx)
+{
+    unsigned reset = 0;
```
terrelln: Could you add a high level comment block explaining how you adapt the compression level.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+    unsigned const allSlow = ctx->adaptParam < ctx->stats.compressedCounter && ctx->adaptParam < ctx->stats.writeCounter && ctx->adaptParam < ctx->stats.readyCounter ? 1 : 0;
```
terrelln: If the ternary operator results are `1` and `0`, you can just use the condition.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+                ctx->threadError = 1;
+                DISPLAY("Error: something went wrong during compression: %s\n", ZSTD_getErrorName(compressedSize));
+                return arg;
+            }
+            job->compressedSize = compressedSize;
+        }
+        pthread_mutex_lock(&ctx->jobCompressed_mutex);
+        ctx->jobCompressedID++;
+        DEBUGLOG(2, "signaling for job %u\n", currJob);
+        pthread_cond_signal(&ctx->jobCompressed_cond);
```
terrelln: You can move the signal outside of the lock.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+        /* problem occurred, free things then return */
+        DISPLAY("Error: problem occurred during job creation\n");
+        free(job->src.start);
+        free(job->dst.start);
+        return 1;
+    }
+    memcpy(job->src.start, data, srcSize);
+    pthread_mutex_lock(&ctx->jobReady_mutex);
+    ctx->jobReadyID++;
+    pthread_cond_signal(&ctx->jobReady_cond);
```
terrelln: Signal can happen without holding the lock.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+    ctx = createCCtx(numJobs, outFilename);
+    if (ctx == NULL) {
+        ret = 1;
+        goto cleanup;
+    }
+
+    /* create output thread */
+    {
+        pthread_t out;
+        if (pthread_create(&out, NULL, &outputThread, ctx)) {
```
terrelln: I think you will sleep forever in `waitUntilAllJobsCompleted()` if this fails (and same for the compression thread). I think that applies for a lot of the failure conditions.

contrib/adaptive-compression/adapt.c
Cyan4973: Are you sure ? 
https://stackoverflow.com/questions/4544234/calling-pthread-cond-signal-without-locking-mutex

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+    waitUntilAllJobsCompleted(ctx);
+    if (g_displayStats) printStats(ctx->stats);
+    /* file compression completed */
+    ret  |= (srcFile != NULL) ? fclose(srcFile) : 0;
+    ret |= (ctx != NULL) ? freeCCtx(ctx) : 0;
+    free(src);
+    return ret;
+}
+
+static int compressFilenames(const char** filenameTable, unsigned numFiles, unsigned forceStdout)
```
Cyan4973: Nothing bad here,
just keep in mind that your primary use case is to read from `stdin` and/or write to `stdout`.
Reading/writing to file means being limited by filesystem speed, which is generally local, fast, and difficult to measure.
Pipe scenarios, involving typically network channels, look like a much better fit for this project.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+            }
+        }
+        if (feof(srcFile)) {
+            DEBUGLOG(2, "THE STREAM OF DATA ENDED %u\n", ctx->nextJobID);
+            ctx->lastJobID = ctx->nextJobID;
+            break;
+        }
+    }
+
+cleanup:
```
Cyan4973: One way to avoid `goto`,
is to create 2 nested functions,
the first one being in charge of creating / freeing resources,
while the second one just uses them.

This way, you can easily have multiple exit points within function 2,
and still get the guarantee that resources will be properly freed.

contrib/adaptive-compression/adapt.c
terrelln: In that example, if `condition = true` is put inside the lock, the signal can be put outside, since it piggybacks on the synchronization of `condition`. The situation presented in the answer can't happen because we hold the lock when we check `condition == false`, and the condition variable guarantees that it will behave as if it atomically releases the mutex and starts waiting.

But you're right that we should probably leave it like this, since it almost certainly doesn't matter performance-wise, and the correctness tricky.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,591 @@
...
+            ctx->stats.readyCounter++;
+            DEBUGLOG(2, "waiting on job ready, nextJob: %u\n", currJob);
+            pthread_cond_wait(&ctx->jobReady_cond, &ctx->jobReady_mutex);
+        }
+        pthread_mutex_unlock(&ctx->jobReady_mutex);
+        DEBUGLOG(2, "compressionThread(): continuing after job ready\n");
+        /* compress the data */
+        {
+            unsigned const cLevel = adaptCompressionLevel(ctx);
+            DEBUGLOG(2, "cLevel used: %u\n", cLevel);
```
Cyan4973: minor detail : 
this `DEBUGLOG()` is incompatible with `-Wdeclaration-after-statement`, since a `size_t` variable is declared just after.
It works on Mac, because the default compiler basically ignore this compilation flag.
If you test it with gcc, or Linux, the compilation is likely to trigger a warning.

Note that respecting this compilation flag is not very important as long as your project remains in `/contrib`.

contrib/adaptive-compression/adapt.c
terrelln: Yeah the performance of these things is very complex, especially with regards to fairness. I say we leave the signal inside the critical section because it is simpler, and as an added benefit it might perform better, though it probably doesn't matter.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,712 @@
...
+        }
+        free(ctx);
+        return error;
+    }
+}
+
+static adaptCCtx* createCCtx(unsigned numJobs, const char* const outFilename)
+{
+
+    adaptCCtx* ctx = calloc(1, sizeof(adaptCCtx));
```
Cyan4973: `const` whenever you can : 
`adaptCCtx* const ctx =`

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,712 @@
...
+    if (ctx == NULL) {
+        DISPLAY("Error: could not allocate space for context\n");
+        return NULL;
+    }
+    ctx->compressionLevel = g_compressionLevel;
+    pthread_mutex_init(&ctx->jobCompressed_mutex, NULL);
```
Cyan4973: pedantic : 
`pthread_mutex_init()` seems to be able to fail : 
http://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_init.html

I don't foresee a lot of problems here, but might be worth adding a check at some point.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,712 @@
...
+        freeCCtx(ctx);
+        return NULL;
+    }
+    if (!ctx->jobs) {
+        DISPLAY("Error: could not allocate space for jobs during context creation\n");
+        freeCCtx(ctx);
+        return NULL;
+    }
+    {
+        unsigned const stdoutUsed = !strcmp(outFilename, stdoutmark);
```
Cyan4973: General note : it feels a bit strange to have a `FILE*` object opened inside `adaptCCtx`.
It seems to imply both lifetimes are tied.
I suspect that's because you have a thread dedicated to write into `dstFile`.

I believe `adaptCCtx` can be more general, as the speed adaptation method doesn't depend on any specific signal from `FILE* dstFile`.

That being said, for a demo, it's not yet a big deal.
Just food for thought, to keep in mind for an improved version with the aim to become part of `zstd` cli or library.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,712 @@
...
+        DEBUG(3, "DICTIONARY ENDED\n");
+        DEBUG(3, "%.*s", (int)job->src.size, (char*)job->src.start);
+        /* compress the data */
+        {
+            unsigned const cLevel = adaptCompressionLevel(ctx);
+            DEBUG(3, "cLevel used: %u\n", cLevel);
+            DEBUG(3, "compression level used: %u\n", cLevel);
+            /* begin compression */
+            {
+                size_t useDictSize = MIN(getUseableDictSize(cLevel), job->dictSize);
```
Cyan4973: `const` ?

contrib/adaptive-compression/test-performance.sh
```diff
@@ -0,0 +1,59 @@
+echo "testing time -- no limits set"
```
Cyan4973: Most of these tests are redundant. They don't bring a lot of additional value.
As a general rule, find a way to reduce time spent in tests. The faster, the better.

Also : here would be a good place to introduce rate-limited tests

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,894 @@
...
+        pthreadError |= initCond(&ctx->jobCompressed_cond);
+        pthreadError |= initMutex(&ctx->jobReady_mutex);
+        pthreadError |= initCond(&ctx->jobReady_cond);
+        pthreadError |= initMutex(&ctx->allJobsCompleted_mutex);
+        pthreadError |= initCond(&ctx->allJobsCompleted_cond);
+        pthreadError |= initMutex(&ctx->jobWrite_mutex);
+        pthreadError |= initCond(&ctx->jobWrite_cond);
+        pthreadError |= initMutex(&ctx->completion_mutex);
+        pthreadError |= initMutex(&ctx->stats_mutex);
+        if (pthreadError) return NULL;
```
Cyan4973: You really need to free allocated resources before leaving the function.
Here, `ctx` is allocated, but not freed, so it will leak.

I suspect the situation is a bit complex because you have many exit points.
This is likely a consequence of having both allocate and init in the same function.

One potential solution : you could nest `init` inside `create`. This way, you could exit `init` from multiple points, and still have a single place to check `init` result and proceed with freeing allocated resources if need be.

lib/compress/zstd_compress.c
```diff
@@ -2997,6 +3002,10 @@ static size_t ZSTD_compressContinue_internal (ZSTD_CCtx* cctx,
         return fhSize;
 }
 
+ZSTDLIB_API double ZSTD_getCompletion(ZSTD_CCtx* cctx)
+{
+    return cctx->completion;
```
Cyan4973: Risk of race condition if you read this variable while it's being modified

lib/compress/zstd_compress.c
```diff
@@ -2895,6 +2899,7 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
         }
 
         remaining -= blockSize;
+        cctx->completion = 1 - (double)remaining/srcSize;
```
Cyan4973: I prefer to avoid floating point operations from the library itself.

lib/compress/zstd_compress.c
```diff
@@ -140,6 +140,9 @@ struct ZSTD_CCtx_s {
     /* Multi-threading */
     U32 nbThreads;
     ZSTDMT_CCtx* mtctx;
+
+    /* adaptive compression */
+    double completion;
```
Cyan4973: Can you find another method which doesn't involve tracking a new variable inside `CCtx` ?

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1002 @@
...
+        free(job.dst.start);
+        free(job.src.start);
+    }
+}
+
+static int destroyMutex(mutex_t* mutex)
+{
+    if (mutex->noError) {
+        int const ret = pthread_mutex_destroy(&mutex->pMutex);
+        return ret;
```
Cyan4973: (minor) : you could `return` directly

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1006 @@
...
+    while (ctx->allJobsCompleted == 0 && !ctx->threadError) {
+        pthread_cond_wait(&ctx->allJobsCompleted_cond.pCond, &ctx->allJobsCompleted_mutex.pMutex);
+    }
+    pthread_mutex_unlock(&ctx->allJobsCompleted_mutex.pMutex);
+}
+
+/* map completion percentages to values for changing compression level */
+static unsigned convertCompletionToChange(double completion)
+{
+    if (completion < 0.1) {
```
Cyan4973: Prefer using some constant, or even some `#define`, over hard-coded values.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1006 @@
...
+            DEBUG(2, "compress waiting on write or create, tried to increase compression level by %u\n\n", boundChange);
+        }
+
+    }
+
+    if (ctx->compressionLevel == prevCompressionLevel) {
+        ctx->convergenceCounter++;
+    }
+
+    if (g_forceCompressionLevel) {
```
Cyan4973: (minor) you could even shortcut that at the beginning, avoiding to go through the resolution.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1006 @@
...
+            pthread_mutex_lock(&ctx->jobReady_mutex.pMutex);
+            if (currJob + 1 > ctx->jobReadyID) willWaitForCreate = 1;
+            pthread_mutex_unlock(&ctx->jobReady_mutex.pMutex);
+
+            pthread_mutex_lock(&ctx->jobWrite_mutex.pMutex);
+            if (currJob - ctx->jobWriteID >= ctx->numJobs) willWaitForWrite = 1;
+            pthread_mutex_unlock(&ctx->jobWrite_mutex.pMutex);
+
+
+            if (willWaitForCreate) {
```
Cyan4973: What happens if `willWaitForCreate == 0` ?
It seems that `ctx->compressWaitCreateCompletion` will not be updated,
so it will keep its previous value.
And this value will be used in next `adaptCompressionLevel` decision ...

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1006 @@
...
+                DEBUG(2, "compression will wait for create on job %u\n", currJob);
+                pthread_mutex_lock(&ctx->createCompletion_mutex.pMutex);
+                ctx->compressWaitCreateCompletion = ctx->createCompletion;
+                DEBUG(2, "create completion %f\n", ctx->compressWaitCreateCompletion);
+                pthread_mutex_unlock(&ctx->createCompletion_mutex.pMutex);
+
+            }
+
+            if (willWaitForWrite) {
```
Cyan4973: same comment

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1002 @@
...
+        ctx->compressionCompletion = 0;
+        pthread_mutex_unlock(&ctx->compressionCompletion_mutex.pMutex);
+
+        /* adapt compression level */
+        if (currJob) adaptCompressionLevel(ctx);
+
+        DEBUG(2, "job %u compressed with level %u\n", currJob, ctx->compressionLevel);
+        /* compress the data */
+        {
+            size_t const compressionBlockSize = 1 << 17; /* 128 KB */
```
Cyan4973: or `ZSTD_BLOCKSIZE_MAX`

contrib/adaptive-compression/adapt.c
paulcruz74: for each completion value (`compressWaitCreateCompletion`, `compressWaitWriteCompletion`, etc.) whenever the value is read during `adaptCompressionLevel()` it is reset to 1. Therefore, if the respective thread does not wait in between successive calls to `adaptCompressionLevel()`, the value will remain at 1

Cyan4973: OK, this works.
But if I'm not mistaken, an `else ... = 1.0` would work the same ?

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1011 @@
...
+        {
+            size_t const compressedSize = job->compressedSize;
+            size_t remaining = compressedSize;
+            if (ZSTD_isError(compressedSize)) {
+                DISPLAY("Error: an error occurred during compression\n");
+                signalErrorToThreads(ctx);
+                return arg;
+            }
+            {
+                size_t const blockSize = MAX(compressedSize >> 7, 1 << 10);
```
Cyan4973: why this choice of write unit size ?

contrib/adaptive-compression/adapt.c
paulcruz74: ah yes that would have worked as well

Cyan4973: OK, so if they work the same, is there a difference then ?
The answer is yes.

If _another part_ of the code needs to deal with it, this is a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction). 
It makes re-using / re-factoring the code more complex.

contrib/adaptive-compression/adapt.c
```diff
@@ -0,0 +1,1011 @@
...
+    unsigned nextJobID;
+    unsigned threadError;
+    unsigned jobReadyID;
+    unsigned jobCompressedID;
+    unsigned jobWriteID;
+    unsigned allJobsCompleted;
+    unsigned adaptParam;
+    unsigned convergenceCounter;
+    unsigned cooldown;
+    double createWaitCompressionCompletion;
```
Cyan4973: Document what these variables mean, their ranges

contrib/adaptive-compression/adapt.c
paulcruz74: for this write unit size, as well as `readBlockSize` I wasn't sure how granular I would need to make the mini-chunks in order to get useful readings for the completion %s. I eventually decided to make them approximately equal to 1/128th of the job size (or compressed size for write) since then each mini-chunk would be approximately 1% of the total job. `1 << 10` is simply a kilobyte, which I arbitrarily chose to avoid the case where `compressedSize >> 10 = 0`

contrib/adaptive-compression/adapt.c
paulcruz74: I'm not quite sure I understand this one,

Cyan4973: `return pthread_mutex_destroy(&mutex->pMutex);`

PR #756, opened 2017-07-12
Memory optimisations for ZSTDMT

Cyan4973:
This patches substantially improves memory usage of `ZSTDMT_*()` API
by making sure resources are allocated "just in time" and released "as soon as possible",
typically directly from within workers.

The new test `fuzzer --memtest` make it possible to measure memory usage of different variants.
For streaming with 2 threads and compression level 1, memory usage drops from ~30 MB to ~18 MB.

More savings are possible by reducing `queueSize` in the thread pool,
so that we have less jobs, hence less buffers, prepared in advanced of workers.

The patch also fixes some issues found during investigation : 
- custom allocator is properly propagated into `ZSTDMT_CCtx*` and child `ZSTD_CCtx*`. Only the thread pool itself now directly calls `stdlib.h`.
- `ZSTDMT_compress_advanced()` (single pass function) correctly generates frame checksum when it's required. Relevant test added in `fuzzer.c`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,7 @@
 v1.3.1
+perf: substantially decreased memory usage in Multi-threading mode, thanks to reports by Tino Reichardt
 build: fix Visual compilation for non x86/x64 targets, reported by Greg Slazinski (#718)
-API exp : breaking change : ZSTD_getframeHeader() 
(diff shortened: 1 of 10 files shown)


mcmilk:
Nice to see ;-)

PR #758, opened 2017-07-12
Add working prototype of a long distance matcher

stellamplau:


terrelln:
Can we deduplicate `ldm-main.c`, `util.[hc]`, `LDM_decompress()`, and other common functions between the versions? Can you also add a README explaining the compression strategy each version uses, so readers know what to expect going in?

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,660 @@
...
+}
+
+/**
+ * Computes a checksum based on rsync's checksum.
+ *
+ * a(k,l) = \sum_{i = k}^l x_i (mod M)
+ * b(k,l) = \sum_{i = k}^l ((l - i + 1) * x_i) (mod M)
+ * checksum(k,l) = a(k,l) + 2^{16} * b(k,l)
+  */
+static U32 getChecksum(const char *data, U32 len) {
```
terrelln: You should also check out the rolling hash that rzip uses https://github.com/ckolivas/lrzip/blob/master/rzip.c#L403.

stellamplau: I didn't use rzip's rolling hash since it seems to use XORs (correct me if I'm wrong here), which means that reordering bytes results in the same hash. Maybe it doesn't matter and a faster hash function is more important. I can experiment.

Cyan4973: A simple rolling hash : 
adding an element : `hash += element * CONSTANT;`
removing an element : `hash -= element * pow(CONSTANT,length);`

Updating a hash of known length : `pow(CONSTANT,length)` becomes itself a constant.
So it's extremely fast.

A variant I like is : `element = UCharValue + 1;`
The idea is to avoid having `element * CONSTANT==0` when `element==0` (which is quite common).

Element could be made a 2 or 4 bytes value instead of a `UChar` in order to progress faster,
but then, in order to check hashes at any position,
you would need to maintain 2 or 4 hashes interlaced, which wouldn't be any faster.
So I tend to prefer the `UChar` variant.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+#define ML_MASK ((1U<<ML_BITS)-1)
+#define RUN_BITS (8-ML_BITS)
+#define RUN_MASK ((1U<<RUN_BITS)-1)
+
+#define COMPUTE_STATS
+#define CHECKSUM_CHAR_OFFSET 2
+//#define RUN_CHECKS
+//#define LDM_DEBUG
+
+typedef  uint8_t BYTE;
```
terrelln: These typedefs can be moved to a common header.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+  unsigned step;                      // ip step, should be 1.
+
+  // DEBUG
+  const BYTE *DEBUG_setNextHash;
+} LDM_CCtx;
+
+/**
+ * Outputs compression statistics.
+ */
+static void printCompressStats(const LDM_CCtx *cctx) {
```
terrelln: This function can probably be shared, if it takes a `LDM_compressStats const*` and a hash table pointer/size instead.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+
+/**
+ * Checks whether the MINMATCH bytes from p are the same as the MINMATCH
+ * bytes from match.
+ *
+ * This assumes MINMATCH is a multiple of four.
+ *
+ * Return 1 if valid, 0 otherwise.
+ */
+static int LDM_isValidMatch(const BYTE *p, const BYTE *match) {
```
terrelln: This function can be shared.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+ * a(k,l) = \sum_{i = k}^l x_i (mod M)
+ * b(k,l) = \sum_{i = k}^l ((l - i + 1) * x_i) (mod M)
+ * checksum(k,l) = a(k,l) + 2^{16} * b(k,l)
+  */
+static U32 getChecksum(const char *data, U32 len) {
+  U32 i;
+  U32 s1, s2;
+  const schar *buf = (const schar *)data;
+
+  s1 = s2 = 0;
```
terrelln: Why are you using a `schar` instead of a `BYTE`?

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+    if (offset != 0 && !LDM_isValidMatch(cctx->ip, offset + cctx->ibase)) {
+      cctx->stats.numCollisions++;
+    }
+  }
+#endif
+
+  // Hash only every HASH_ONLY_EVERY times, based on cctx->ip.
+  // Note: this works only when cctx->step is 1.
+  if (((cctx->ip - cctx->ibase) & HASH_ONLY_EVERY) == HASH_ONLY_EVERY) {
+    (cctx->hashTable)[hash] = (hashEntry){ (offset_t)(cctx->ip - cctx->ibase) };
```
terrelln: This is a bit strange, I'd prefer

```c
cctx->hashTable[hash].offset = cctx->ip - cctx->ibase;
```

or

```c
hashEntry const e = { cctx->ip - cctx->ibase);
cctx->hashTable[hash] = e;
```

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+            cctx->ip - cctx->ibase);
+  }
+#endif
+}
+
+static void putHashOfCurrentPositionFromHash(
+    LDM_CCtx *cctx, hash_t hash, U32 sum) {
+#ifdef COMPUTE_STATS
+  if (cctx->stats.numHashInserts < LDM_HASHTABLESIZE_U32) {
+    offset_t offset = (cctx->hashTable)[hash].offset;
```
terrelln: No need for the parens around `cctx->hashTable`. Same for all array accesses above.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+  return cctx->hashTable[hash].offset + cctx->ibase;
+}
+
+/**
+ *  Counts the number of bytes that match from pIn and pMatch,
+ *  up to pInLimit.
+ *
+ *  TODO: make more efficient.
+ */
+static unsigned countMatchLength(const BYTE *pIn, const BYTE *pMatch,
```
terrelln: This function can be shared.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+    }
+    return (unsigned)(pIn - pStart);
+  }
+  return (unsigned)(pIn - pStart);
+}
+
+void LDM_readHeader(const void *src, U64 *compressSize,
+                    U64 *decompressSize) {
+  const U64 *ip = (const U64 *)src;
+  *compressSize = *ip++;
```
terrelln: This function can be shared.

This is undefined behavior, since `src` might not be 8 byte aligned, and it also isn't cross platform (which doesn't really matter). You should use

```c
BYTE const* ip = (BYTE const*)src;
*compressedSize = LDM_readLE64(ip);
ip += sizeof(U64);
*decompressedSize = LDM_readLE64(ip);
ip += sizeof(U64);
```

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+  return 0;
+}
+
+/**
+ * Write current block (literals, literal length, match offset,
+ * match length).
+ *
+ * Update input pointer, inserting hashes into hash table along the way.
+ */
+static void outputBlock(LDM_CCtx *cctx,
```
terrelln: This function can be shared if you only pass in the parts of `CCtx` you need.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+      }
+    }
+
+    // Set start of next block to current input pointer.
+    cctx.anchor = cctx.ip;
+    LDM_updateLastHashFromNextHash(&cctx);
+  }
+_last_literals:
+  /* Encode the last literals (no more matches). */
+  {
```
terrelln: This can be separated into its own function and shared.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+  }
+
+#ifdef COMPUTE_STATS
+  printCompressStats(&cctx);
+#endif
+
+  return (cctx.op - (const BYTE *)cctx.obase);
+}
+
+typedef struct LDM_DCtx {
```
terrelln: All the decompression functions can be shared, right?

contrib/long_distance_matching/ldm.h
```diff
@@ -0,0 +1,55 @@
+#ifndef LDM_H
```
terrelln: You only need one copy of the header, since they expose the same set of functions.

contrib/long_distance_matching/main-ldm.c
```diff
@@ -0,0 +1,269 @@
+// TODO: file size must fit into a U32
```
terrelln: You only need one copy of main, since the versions expose the same functions.

Another option, which might help testing, is to make a struct like

```c
typedef struct {
    size_t (*compress)(void const* src, size_t srcSize, void* dst, size_t maxDstSize);
    size_t (*decompress)(void const* src, size_t srcSize, void* dst, size_t maxDstSize);
    char const* name;
} compressor;
```

Then each version implements a `static` compression function, and uses a shared decompression function, and at the bottom of the source file they write

```c
compressor const ldmv1_compressor = {
    .compress = LDMv1_compress,
    .decompress = LDM_decompress,
    .name = "LDMv1",
};
```

Then in a common header, have an array of compressors

```c
extern const compressor ldmv1_compressor;
extern const compressor ldmv2_compressor;
const compressor compressors[] = {
  ldmv1_compressor,
  ldmv2_compressor,
};
```

Then you can accept a "name" argument on the command line, and loop through the compressors until you find one that matches the name, and use that one. You could also benchmark them all in one command.

contrib/long_distance_matching/main.c
```diff
@@ -0,0 +1,240 @@
+#include <stdint.h>
```
terrelln: Whats the difference between `ldm-main.c` and `main.c`?

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+  // DEBUG
+  const BYTE *DEBUG_setNextHash;
+} LDM_CCtx;
+
+/**
+ * Outputs compression statistics.
+ */
+static void printCompressStats(const LDM_CCtx *cctx) {
+  const LDM_compressStats *stats = &(cctx->stats);
+#ifdef COMPUTE_STATS
```
Cyan4973: At this stage, computing stats is pretty important, as this is the primary tool to guide decisions.
The below module is already pretty good.
What I believe is important : 
- nb matches (done)
- Total Match Length (stored, to be displayed)
- Average Match Length (done)
- Average Offset (done)
- Min/Max Offset (todo)
- If possible : offset histogram, classified by power of 2 (more difficult)
- Memory usage, comparison to source size
- When relevant / possible : scanning speed

Feel free to add relevant indicators.

Ultimately, we want to merge this algorithm with the rest of zstd,
so these indicators will help judge, between different variants,
which one is more efficient doing complementary work,
anticipating future combination with regular zstd match finder.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,663 @@
...
+  dctx->maxDecompressSize = maxDecompressSize;
+
+  dctx->ibase = src;
+  dctx->ip = (const BYTE *)src;
+  dctx->iend = dctx->ip + dctx->compressSize;
+  dctx->op = dst;
+  dctx->oend = dctx->op + dctx->maxDecompressSize;
+}
+
+size_t LDM_decompress(const void *src, size_t compressSize,
```
Cyan4973: Decompression itself is not important at this stage,
but if the goal is to check that data generated from match-finding stage is correct,
it's indeed a relevant test tool.

contrib/long_distance_matching/util.h
```diff
@@ -0,0 +1,27 @@
+#ifndef LDM_UTIL_H
```
Cyan4973: You might be interested in using `lib/common/mem.h` for an equivalent set of functions,
since, ultimately, this will be the module to use when integrating your work into `zstd`.

Cyan4973:
I still have to test the compiled program,
but initial read of source code looks fairly good.

As a general direction, focus on generating relevant stats,
as this is information we will need for arbitration between variants.

contrib/long_distance_matching/ldm.c
stellamplau: It's not needed -- I will switch to `BYTE`.

contrib/long_distance_matching/circular_buffer_table.c
```diff
@@ -0,0 +1,107 @@
...
+  table->bucketOffsets = calloc(size, sizeof(BYTE));
+  table->offsetBase = offsetBase;
+  return table;
+}
+
+static LDM_hashEntry *getBucket(const LDM_hashTable *table, const hash_t hash) {
+  return table->entries + (hash << HASH_BUCKET_SIZE_LOG);
+}
+
+LDM_hashEntry *HASH_getValidEntry(const LDM_hashTable *table,
```
terrelln: We don't just want to resolve collisions with the buckets, we also want to keep multiple candidates that all start with the same `LDM_MIN_MATCH_LENGTH` bytes, so we can look at all of them and select the best match (including reverse match).

contrib/long_distance_matching/circular_buffer_table.c
```diff
@@ -0,0 +1,107 @@
...
+                                  const hash_t hash,
+                                  const U32 checksum,
+                                  const BYTE *pIn,
+                                  int (*isValid)(const BYTE *pIn, const BYTE *pMatch)) {
+  LDM_hashEntry *bucket = getBucket(table, hash);
+  LDM_hashEntry *cur = bucket;
+  // TODO: in order of recency?
+  for (; cur < bucket + HASH_BUCKET_SIZE; ++cur) {
```
terrelln: We will want to check every match in the bucket (if that is too slow, we could try to sample the matches).

First check the checksum, if it matches, then count the match length, and return the match with the longest match length. Since we don't want to repeat this work, we will have to communicate the match length, and match starting position to the caller.

If we need more speed, we can try fancier strategies, but lets do the easy stuff first.

contrib/long_distance_matching/circular_buffer_table.c
stellamplau: I did some experimenting with this. There is a large (> 2x?) speed hit in checking the backwards/reverse match every time (even when comparing checksums first). I approximated by checking backwards only if the forward match is long enough.  

To give a brief idea of what the speeds look like in terms of bucket size and checking all the entries (this is from 4 versions of clang--3.9.0, 3.9.1, 4.0.0, 4.0.1--tarred together):

Bucket size, speed, check first or check all?
8, 106 MB/s, first only
8, 91 MB/s, all
16, 74 MB/s, first only
16, 65 MB/s, all

I don't really notice an improvement in compression ratio (and even see it worsening slightly but that might be file dependent).

contrib/long_distance_matching/circular_buffer_table.c
stellamplau: Sounds good. 

Minor side design question: Implementing the change made me realize the "hash table" is probably too tightly coupled with the ldm.c, since without function pointers the "getValidEntry" function requires passing in a lot of specific parameters located in LDM_CCtx. Do you think decoupling the hash table really makes sense?

contrib/long_distance_matching/circular_buffer_table.c
terrelln: What compression ratio are you getting (and file size)?

contrib/long_distance_matching/circular_buffer_table.c
terrelln: I imagine the parameters you need are the base pointer, the anchor, the current pointer, and the end. It would need to return the match offset/length (if any) and where it started (in case of reverse match). That sounds reasonable to me, and I think the `getValidEntry()` should remain its own function. But you should structure the code in a way that makes sense to you.

contrib/long_distance_matching/circular_buffer_table.c
terrelln: Also, what parameters are you using?

stellamplau: Here are some stats from compressing the file. The ldm example was run with
`#define LDM_MIN_MATCH_LENGTH 64`
`#define HASH_BUCKET_SIZE_LOG 3`
`#define LDM_WINDOW_SIZE_LOG 28`
`#define LDM_MEMORY_USAGE 23`

^ I made some changes so memory usage of 2^23 and bucket size of 2^3 implies 2^16 buckets. 

File size: 568645152

Method |  Size | Ratio
-------- | -----| -----
lrzip -p 32 -n -w 1        |	      181887800	|      3.13
ldm 			|			      153828356	|      3.70
lz4                                   | 	      104075804	|      5.46
lrzip -p 32 -l -w 1          |           	       88496017	|      6.43
lrzip -p 32 -n                 |         	       80074779	|      7.10
zstd -1                          |       	       69203650	|      8.22
lrzip -p 32 -n -w 1 & zstd -1          | 	       66116676	 |     8.60
gzip -9                           |      	       64613027 |	      8.80
lrzip -p 32 -g -L 9 -w 1        |        	       58757173	|      9.68
zstd -T16 -9                           | 	       54564744	|     10.42
lrzip -p 32 -n -w 1 & zstd -T32 -19   |  	       49499696	|     11.49
zstd -T32 -19                         |  	       47919263	|     11.87
ldm & zstd -1 				|	       26363651   |   21.57	
lrzip -p 32 -n & zstd -1             |   	       21549633	|     26.39
lrzip -p 32 -g -L 9                    | 	       19035126	   |  29.87
lrzip -o 32 -n & zstd -T32 -19     |    	       16350887	 |    34.78

terrelln: Awesome! It looks like you boost the compression ratio quite a bit, what changed?

Did you make `HASH_ONLY_EVERY` larger?

terrelln: I think we should check everything in the bucket, but maybe we want `HASH_BUCKET_SIZE_LOG == 0`. It seems to me that only checking the first thing that matches in a bucket is very similar to `HASH_BUCKET_SIZE_LOG == 0`, since there will be less collisions. The difference would be in the collisions of the rolling hash before it is masked to the table size.

stellamplau: I calculate `HASH_ONLY_EVERY` as a function of the other parameters for an attempt at more even window coverage. I pushed an update (though I need to clean it up).

stellamplau: Btw, the stats above were calculated from checking everything in the bucket, though I didn't really see significant changes (from a cursory check) from not checking everything.

contrib/long_distance_matching/ldm_with_table.c
```diff
@@ -0,0 +1,973 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "ldm.h"
+
+#define LDM_HASHTABLESIZE (1 << (LDM_MEMORY_USAGE))
+//#define LDM_HASH_ENTRY_SIZE 4
+#define LDM_HASHTABLESIZE_U32 ((LDM_HASHTABLESIZE) >> 2)
+#define LDM_HASHTABLESIZE_U64 ((LDM_HASHTABLESIZE) >> 4)
```
terrelln: I think this is off by 1, it should be 3.

contrib/long_distance_matching/ldm_with_table.c
```diff
@@ -0,0 +1,973 @@
...
+
+// Insert every (HASH_ONLY_EVERY + 1) into the hash table.
+#define HASH_ONLY_EVERY_LOG (LDM_WINDOW_SIZE_LOG-((LDM_MEMORY_USAGE) - 4))
```
terrelln: Do you want 3 or 4 here?

contrib/long_distance_matching/ldm_with_table.c
```diff
@@ -0,0 +1,973 @@
...
+#define HASH_ONLY_EVERY ((1 << HASH_ONLY_EVERY_LOG) - 1)
+
+/* Hash table stuff. */
+#define HASH_BUCKET_SIZE (1 << (HASH_BUCKET_SIZE_LOG))
+#define LDM_HASHLOG ((LDM_MEMORY_USAGE)-4-HASH_BUCKET_SIZE_LOG)
```
terrelln: 3 or 4?

contrib/long_distance_matching/ldm_with_table.c
stellamplau: Yes agreed, silly me.

contrib/long_distance_matching/ldm_common.c
```diff
@@ -0,0 +1,99 @@
...
+
+#include "ldm.h"
+
+void LDM_readHeader(const void *src, U64 *compressedSize,
+                    U64 *decompressedSize) {
+  const BYTE *ip = (const BYTE *)src;
+  *compressedSize = MEM_readLE64(ip);
+  ip += sizeof(U64);
+  *decompressedSize = MEM_readLE64(ip);
+  // ip += sizeof(U64);
```
Cyan4973: Probably a line worth removing

contrib/long_distance_matching/ldm_common.c
```diff
@@ -0,0 +1,99 @@
+#include <stdio.h>
+
+#include "ldm.h"
+
```
Cyan4973: Here would be a good place to document your header format

contrib/long_distance_matching/main.c
```diff
@@ -0,0 +1,255 @@
...
+
+    if (0 == result) result = memcmp(b0, b1, r0);
+  }
+  return result;
+}
+
+/* Verify the input file is the same as the decompressed file. */
+static void verify(const char *inpFilename, const char *decFilename) {
+  FILE *inpFp = fopen(inpFilename, "rb");
+  FILE *decFp = fopen(decFilename, "rb");
```
Cyan4973: ok, it's just a demo,
but for better robustness, you should check if `inpFp` and `decFp` are not `NULL`, signalling a problem with `fopen()`.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,836 @@
...
+ * Create a hash table that can contain size elements.
+ * The number of buckets is determined by size >> HASH_BUCKET_SIZE_LOG.
+ */
+static LDM_hashTable *HASH_createTable(U32 size) {
+  LDM_hashTable *table = malloc(sizeof(LDM_hashTable));
+  table->numBuckets = size >> HASH_BUCKET_SIZE_LOG;
+  table->numEntries = size;
+  table->entries = calloc(size, sizeof(LDM_hashEntry));
+  table->bucketOffsets = calloc(size >> HASH_BUCKET_SIZE_LOG, sizeof(BYTE));
+  return table;
```
Cyan4973: Here also, it's just a demo,
but for better robustness, you should check that `table`, `entries` and `bucketOffsets` are not `NULL`,
and return a `NULL` pointer object if one of them is (after releasing allocated resources).

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,836 @@
...
+  size_t matchLength = 0;
+  while (pIn > pAnchor && pMatch > pBase && pIn[-1] == pMatch[-1]) {
+    pIn--;
+    pMatch--;
+    matchLength++;
+  }
+  return matchLength;
+}
+
+/**
```
Cyan4973: 👍 
I really like your clear explanations on what each function do

Cyan4973:
Looks good to me.
@terrelln, do you want to have a look before a merge ?

terrelln:
Taking a look now

contrib/long_distance_matching/main.c
```diff
@@ -0,0 +1,265 @@
...
+        printf("Decompress error");
+        return 1;
+    }
+    gettimeofday(&tv2, NULL);
+    printf("Total decompress time = %f seconds\n",
+          (double) (tv2.tv_usec - tv1.tv_usec) / 1000000 +
+          (double) (tv2.tv_sec - tv1.tv_sec));
+  }
+  /* verify */
+  verify(inpFilename, decFilename);
```
terrelln: nit: Return non-zero from `verify()` and `main()` if verification fails.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,855 @@
...
+#define LDM_HASHTABLESIZE_U64 ((LDM_HASHTABLESIZE) >> 3)
+
+#if USE_CHECKSUM
+  #define LDM_HASH_ENTRY_SIZE_LOG 3
+#else
+  #define LDM_HASH_ENTRY_SIZE_LOG 2
+#endif
+
+// Entries are inserted into the table HASH_ONLY_EVERY + 1 times "on average".
+#ifndef HASH_ONLY_EVERY_LOG
```
terrelln: Note: We may want to allow non-power of 2 values for this. That way we could aim to make the table say 80-90% full, in the hope that we keep a bit more long matches.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,855 @@
...
+                                        U64 *pBackwardMatchLength) {
+  LDM_hashTable *table = cctx->hashTable;
+  LDM_hashEntry *bucket = getBucket(table, hash);
+  LDM_hashEntry *cur = bucket;
+  LDM_hashEntry *bestEntry = NULL;
+  U64 bestMatchLength = 0;
+#if !(USE_CHECKSUM)
+  (void)checksum;
+#endif
+  for (; cur < bucket + HASH_BUCKET_SIZE; ++cur) {
```
terrelln: nit: I generally like to put `cur = bucket` here instead of above, since it makes it very clear where it starts.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,855 @@
...
+  LDM_hashTable *table = malloc(sizeof(LDM_hashTable));
+  if (!table) return NULL;
+
+  table->numBuckets = size >> HASH_BUCKET_SIZE_LOG;
+  table->numEntries = size;
+  table->entries = calloc(size, sizeof(LDM_hashEntry));
+  table->bucketOffsets = calloc(size >> HASH_BUCKET_SIZE_LOG, sizeof(BYTE));
+
+  if (!table->entries || !table->bucketOffsets) {
+    free(table->bucketOffsets);
```
terrelln: This could be

```c
HASH_destroyTable(table);
```

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,855 @@
...
+  printf("Num buckets, bucket size: %d (2^%d), %d\n",
+         table->numBuckets, NUM_HASH_BUCKETS_LOG, HASH_BUCKET_SIZE);
+  printf("Hash table size, empty slots, %% empty: %u, %u, %.3f\n",
+         table->numEntries, ctr,
+         100.0 * (double)(ctr) / table->numEntries);
+}
+
+// TODO: This can be done more efficiently (but it is not that important as it
+// is only used for computing stats).
+static int intLog2(U64 x) {
```
terrelln: You don't need to change this, but gcc/clang provide a builtin `__bultin_clzll()` that can be used to compute this efficiently.

contrib/long_distance_matching/ldm.c
```diff
@@ -0,0 +1,855 @@
...
+  printf("avg offset length: %.1f\n",
+         ((double)stats->totalOffset) / (double)stats->numMatches);
+  printf("min offset, max offset: %u, %u\n",
+         stats->minOffset, stats->maxOffset);
+
+  printf("\n");
+  printf("offset histogram | match length histogram\n");
+  printf("offset/ML, num matches, %% of matches | num matches, %% of matches\n");
+
+  for (; i <= intLog2(stats->maxOffset); i++) {
```
terrelln: FYI, you generally only want to put constant expressions in the loop condition check. The compiler will need to recompute it every time if it can't prove it  doesn't change during the loop. It can probably prove that for simple expressions/functions it can inline, but better safe than sorry.

PR #770, opened 2017-07-18
[zstdcli] Fix -t in streaming mode

terrelln:
It would fail in streaming mode because stdout is a console, but it is actually `/dev/null`, since it got overwritten later.


Cyan4973:
Good catch,
that test case was definitely missing !

PR #778, opened 2017-07-28
[libzstd] Fix bug in Huffman decompresser

terrelln:
The zstd format specification doesn't enforce that Huffman compressed
literals (including the table) have to be smaller than the uncompressed
literals. The compressor will never Huffman compress literals if the
compressed size is larger than the uncompressed size. The decompresser
doesn't accept Huffman compressed literals with 4 streams whose compressed
size is at least as large as the uncompressed size.

* Make the decompresser accept Huffman compressed literals whose size
  increases.
* Add a test case that exposes the bug. The compressed file has to be
  statically generated, since the compressor won't normally produce files
  that expose the bug.


lib/decompress/huf_decompress.c
```diff
@@ -921,7 +921,7 @@ static const algo_time_t algoTime[16 /* Quantization */][3 /* single, double, qu
 U32 HUF_selectDecoder (size_t dstSize, size_t cSrcSize)
 {
     /* decoder timing evaluation */
-    U32 const Q = (U32)(cSrcSize * 16 / dstSize);   /* Q < 16 since dstSize > cSrcSize */
+    U32 const Q = cSrcSize >= dstSize ? 15 : (U32)(cSrcSize * 16 / dstSize);   /* Q < 16 since dstSize > cSrcSize */
```
Cyan4973: nit: comment is no longer valid

lib/decompress/huf_decompress.c
```diff
@@ -977,7 +977,6 @@ size_t HUF_decompress4X_hufOnly_wksp(HUF_DTable* dctx, void* dst,
 {
     /* validation checks */
     if (dstSize == 0) return ERROR(dstSize_tooSmall);
-    if ((cSrcSize >= dstSize) || (cSrcSize <= 1)) return ERROR(corruption_detected);   /* invalid */
```
Cyan4973: What about `cSrcSize==0` ?

tests/playTests.sh
```diff
@@ -386,6 +386,13 @@ $ZSTD -t tmpSplit.* && die "bad file not detected !"
 ./datagen | $ZSTD -c | $ZSTD -t
 
 
+
+$ECHO "\n**** golden files tests **** "
```
Cyan4973: 👍

lib/decompress/huf_decompress.c
```diff
@@ -977,7 +977,7 @@ size_t HUF_decompress4X_hufOnly_wksp(HUF_DTable* dctx, void* dst,
...
+    if (cSrcSize == 0) return ERROR(dstSize_tooSmall);
```
Cyan4973: I don't think that's the right error code.
I would rather expect `ERROR(corruption_detected)` here.

Cyan4973:
This patch makes it possible for `HUF_selectDecoder()` to accept `cSrcSize >= dstSize`.

This capability is then only accessible through `HUF_decompress4X_hufOnly_wksp()`, since it's the only version needed by `ZSTD_decompress()`. 

(note : other entry points using `HUF_selectDecoder()` have additional checks making it impossible for `cSrcSize > dstSize` to exist, and hardwire `cSrcSize == dstSize` to "raw uncompressed mode").

OK, so that will work.
Opening this new possibility into `zstd` decoder is harmless.

Exploiting it from the encoder side though is problematic, because all `zstd` reference decoder versions, up to `1.3.0`, enforce this size restriction. Any compressed data featuring `cLiteralSize >= dstLiteralSize` will fail on these decoders. So we have to refrain using it for some time, while waiting for the world to update.

There is a way out though : the single-stream variant, `HUF_decompress1X2_DCtx_wksp()`, is unaffected by this restriction. So it's still possible to insert huffman headers in (very) small blocks through the single-stream variant, it will have no negative impact on existing decoders.

PR #781, opened 2017-08-01
Allow queueSize=0 in pool.c

stellamplau:
When queueSize=0, wait until there is an available thread before allowing a job to be added.

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -66,6 +66,7 @@ static void* POOL_thread(void* opaque) {
     for (;;) {
         /* Lock the mutex and wait for a non-empty queue or until shutdown */
         pthread_mutex_lock(&ctx->queueMutex);
+
         while (ctx->queueEmpty && !ctx->shutdown) {
             pthread_cond_wait(&ctx->queuePopCond, &ctx->queueMutex);
         }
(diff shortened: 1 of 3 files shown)


lib/common/pool.c
```diff
@@ -159,14 +174,21 @@ void POOL_add(void* ctxVoid, POOL_function function, void *opaque) {
...
     {   POOL_job const job = {function, opaque};
-        /* Wait until there is space in the queue for the new job */
+
+        // Wait until there is space in the queue for the new job.
+        // If the ctx->queueSize is 1 (the pool was created with an
+        // intended queueSize of 0) and there is no job already waiting,
+        // wait until there is a thread free for the new job.
         size_t newTail = (ctx->queueTail + 1) % ctx->queueSize;
-        while (ctx->queueHead == newTail && !ctx->shutdown) {
+        while (ctx->queueHead == newTail && !ctx->shutdown &&
```
terrelln: Can we have a separate `static int isQueueFull(POOL_ctx const*)` function, that handles the special cases?

lib/common/pool.c
```diff
@@ -159,14 +174,21 @@ void POOL_add(void* ctxVoid, POOL_function function, void *opaque) {
 
     pthread_mutex_lock(&ctx->queueMutex);
     {   POOL_job const job = {function, opaque};
-        /* Wait until there is space in the queue for the new job */
+
+        // Wait until there is space in the queue for the new job.
```
terrelln: You have to use C-style comments (`/* */`) in zstd code, because some compiles don't accept C++/C99 style comments (`//`).

tests/poolTests.c
```diff
@@ -50,21 +51,45 @@ int testOrder(size_t numThreads, size_t queueSize) {
   return 0;
 }
 
+void waitFn(void *opaque) {
+  (void)opaque;
+  usleep(100);
```
Cyan4973: Note that we also have `UTIL__sleep()` and `UTIL_sleepMilli()` within `programs/util.h`
with the benefit of being portable on Windows and Posix.

PR #782, opened 2017-08-03
Fix compression failure on incompressible data

### LINKED ISSUE #780: V1.3.0 x64 - Windows binary bug
Hi !

Today i faced the first bug in V1.3.0 x64 windows binary.
See attached picture : 

bug

The third try is only a cosmetic bug ?! This POSIX-text thing...

terrelln:
If the destination buffer is the minimum allowed size in
`ZSTD_compressSequences()` (2^17), then if the block isn't compressible
compression might fail with `dstSize_tooSmall`, when it should instead emit
a raw uncompressed block.

Additionally, `ZSTD_compressLiterals()` implicitly called
`ZSTD_noCompressLiterals()` if Huffman compression failed. Make that
explicit.

Fixes #780.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -942,7 +942,7 @@ static size_t ZSTD_compressLiterals (ZSTD_entropyCTables_t * entropy,
         else { entropy->hufCTable_repeatMode = HUF_repeat_check; }       /* now have a table to reuse */
     }
 
-    if ((cLitSize==0) | (cLitSize >= srcSize - minGain)) {
+    if ((cLitSize==0) | (cLitSize >= srcSize - minGain) | ERR_isError(cLitSize)) {
         entropy->hufCTable_repeatMode = HUF_repeat_none;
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1261,6 +1266,20 @@ MEM_STATIC size_t ZSTD_compressSequences (seqStore_t* seqStorePtr,
     { int i; for (i=0; i<ZSTD_REP_NUM; i++) seqStorePtr->rep[i] = seqStorePtr->repToConfirm[i]; }
 
     return op - ostart;
+
+_dst_size_too_small:
```
Cyan4973: My understanding is that this part is always triggered by ` ERROR(dstSize_tooSmall)`.

Therefore, could we nest it instead ?
`ZSTD_compressSequences()` would become `ZSTD_compressSequences_internal()` (for example),
it would be called by a new `ZSTD_compressSequences()`
which only role is to read the return code, and trigger raw uncompressed block when it is `== ERROR(dstSize_tooSmall)`.

It's basically the same, but feels cleaner than a `goto` statement.
What do you think ?

PR #783, opened 2017-08-07
Add long range matcher

stellamplau:


Cyan4973:
Just for code review, not meant to be merged

lib/compress/zstd_compress.c
```diff
@@ -646,10 +697,10 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
                                     || (params.cParams.strategy == ZSTD_btultra)) ?
                                     optPotentialSpace : 0;
             size_t const bufferSpace = buffInSize + buffOutSize;
-            size_t const neededSpace = entropySpace + optSpace + tableSpace
-                                     + tokenSpace + bufferSpace;
+            size_t const neededSpace = entropySpace + optSpace + + ldmSpace
```
Cyan4973: minor : 
optSpace __+ +__ ldmSpace ?

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2883,714 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+ *
+ *  hBits is the length of the small hash, which is the most significant hBits
+ *  of rollingHash. The checksum is the next 32 most significant bits, followed
+ *  by LDM_HASH_EVERY_LOG bits that make up the tag. */
+static void ZSTD_ldm_makeEntryAndInsertByTag(ldmEntry_t* table,
+                                             BYTE* bucketOffsets,
+                                             U64 rollingHash, U32 hBits,
+                                             U32 offset)
+{
+    U32 tag = ZSTD_ldm_getTag(rollingHash, hBits);
```
Cyan4973: minor : coding style : 
try to `const` whenever possible,
it's a great hint for reviewers that the value is set for the rest of the scope.

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2883,714 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+
+            ip -= backwardMatchLength;
+            match -= backwardMatchLength;
+            mLength += backwardMatchLength;
+
+            /* Overwrite rep codes */
+            for (i = 0; i < ZSTD_REP_NUM; i++)
+                seqStorePtr->rep[i] = repToConfirm[i];
+
+            ZSTD_ldm_fillTables(cctx, anchor, base);
```
Cyan4973: `base` is a `const` ptr.
It does not seem correct.
You want to add new segments only,
this is adding the same segment multiple times, just enlarging it at each attempt.

terrelln:
Let me know when the code is ready for review and I'll take an in-depth look.

stellamplau:
Apart from adding the long distance parameters to the compression context, the code can be reviewed

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2884,684 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+        {
+            ldmEntry_t* const bucket =
+                ZSTD_ldm_getBucket(hashTable,
+                                   ZSTD_ldm_getSmallHash(rollingHash, hBits));
+            ldmEntry_t* cur;
+            size_t bestMatchLength = 0;
+            U32 const checksum = ZSTD_ldm_getChecksum(rollingHash, hBits);
+
+            for (cur = bucket; cur < bucket + LDM_BUCKET_SIZE; ++cur) {
+                const BYTE*const  pMatch = cur->offset + base;
```
terrelln: Spacing.

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2884,684 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+                if (cur->checksum != checksum || cur->offset <= lowestIndex) {
+                    continue;
+                } else {
```
terrelln: I'd delete the `else` to reduce the indentation a bit, and just declare the match lengths at the top of the loop.

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2884,684 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+        rollingHash = ZSTD_ldm_fillLdmHashTable(hashTable, bucketOffsets,
+                                                rollingHash, lastHashed,
+                                                ip + mLength,
+                                                base, hBits);
+        lastHashed = ip + mLength - 1;
+        ip += mLength;
+        anchor = ip;
+
+        /* Check immediate repcode */
+        while ( (ip <= ilimit)
```
terrelln: Does this help compression ratio, or can we should we just axe it?

terrelln:
Can you squash all of your commits into one, and rebase on top of the `facebook:longRangeMatcher` branch instead of merging it in? There is a bunch of unrelated changes showing up in the PR.

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2884,684 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+    case ZSTD_lazy2:
+        if (iend - HASH_READ_SIZE >= zc->base) {
+            ZSTD_insertAndFindFirstIndex(
+                    zc, iend - HASH_READ_SIZE, mls);
+            break;
+        }
+        return 0;
+    case ZSTD_btlazy2:
+    case ZSTD_btopt:
+    case ZSTD_btultra:
```
stellamplau: I think the bt variants all call updateTree within the block compressors. Actually, it seems like the lazy variants update the hash tables themselves, and only fast/dfast need the tables to be filled.

terrelln: Aren't you setting `nextToUpdate` after this though?

terrelln: Oh, I missed the return statement.

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2884,684 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+        return 0;
+    default:
+        assert(0);  /* not possible : not a valid strategy id */
+    }
+
+    zc->nextToUpdate = (U32)(iend - zc->base);
```
terrelln: Can you delete this, since it is never called?

lib/compress/zstd_compress.c
stellamplau: I kept it because the compression ratio improved and it didn't seem to have any negative effects on speed. Although only very slightly (order of ~2-3% maybe). It could be removed if it's better for code clarity. Thoughts?

terrelln: If it helps compression ratio I think we should keep it.

lib/compress/zstd_compress.c
```diff
@@ -2803,18 +2884,684 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+    const BYTE* lastHashed;
+    ZSTD_blockCompressor blockCompressor =
+        ZSTD_selectBlockCompressor(cctx->appliedParams.cParams.strategy, 0);
+    size_t i, lastLiterals;
+
+    /* Save seqStorePtr->rep and copy repToConfirm */
+    for (i = 0; i < ZSTD_REP_NUM; i++)
+        savedRep[i] = repToConfirm[i] = seqStorePtr->rep[i];
+
+    rollingHash = ZSTD_ldm_getRollingHash(ip, LDM_MIN_MATCH_LENGTH);
```
terrelln: What if `srcSize < LDM_MIN_MATCH_LENGTH`?

lib/compress/zstd_compress.c
```diff
@@ -623,6 +644,14 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
             return ZSTD_continueCCtx(zc, params, pledgedSrcSize);
     }   }
 
+    {
+        zc->ldmState.hashLog = LDM_HASH_LOG;
+        zc->ldmState.bucketLog =
+            MIN(LDM_BUCKET_SIZE_LOG, LDM_BUCKET_SIZE_LOG_MAX);
+        zc->ldmState.hashEveryLog =
```
terrelln: What if `cParams.windowLog < ldmState.hashLog` (caught by the simple_round_trip fuzzer with UBSAN)?

stellamplau:
This is ready for review. Some notes:
- handling of `ldmHashEveryLog` is different to the other parameters to maintain a "dynamic" default value that's a function of two other values. Maybe there's a better way?
- max value for `minMatch` is currently somewhat arbitrary
- I'm not sure we want to expose all the ldm parameters? I held off on adding the bucketLog...

Cyan4973:
Having dynamic parameter values is not an issue I believe.
We already have such behaviour : provide a compression level and a `srcSize`, and compression parameters will be adjusted on the fly.

This is still early days for this new capability.
It's unclear which parameters can make an impact, and why to change them.
That means, the more flexible it is to set them, the better for later analysis and tuning.

At some point, we will have to decide which parameters are really worthwhile for an API user, and which ones should not be changed. It will reduce the list of front-facing parameters.

lib/common/zstd_internal.h
```diff
@@ -249,6 +249,26 @@ typedef struct {
...
+    ldmEntry_t* hashTable;
+    BYTE* bucketOffsets;    /* Next position in bucket to insert entry */
+    U64 hashPower;          /* Used to compute the rolling hash.
+                             * Depends on ldmParams.minMatchLength */
+} ldmState_t;
+
+typedef struct {
+    U32 enableLdm;          /* 1 if enable long distance matching */
+    U32 hashLog;            /* Log size of hashTable */
+    U32 bucketLog;          /* Log number of buckets, at most 4 */
```
Cyan4973: For my understanding, what's the reason for `bucketLog <= 4` ?

lib/common/zstd_internal.h
```diff
@@ -249,6 +249,26 @@ typedef struct {
...
+} ldmEntry_t;
+
+typedef struct {
+    ldmEntry_t* hashTable;
+    BYTE* bucketOffsets;    /* Next position in bucket to insert entry */
+    U64 hashPower;          /* Used to compute the rolling hash.
+                             * Depends on ldmParams.minMatchLength */
+} ldmState_t;
+
+typedef struct {
```
Cyan4973: 👍 
Clear structure description

lib/common/zstd_internal.h
stellamplau: Actually currently it should be at most 3 since a BYTE array is used to store the insertion position per bucket. It can be expanded at the expense of memory?

stellamplau: Nevermind about the previous comment -- it can be larger.

lib/compress/zstd_compress.c
```diff
@@ -354,6 +374,23 @@ size_t ZSTD_CCtx_setParameter(ZSTD_CCtx* cctx, ZSTD_cParameter param, unsigned v
         DEBUGLOG(5, " setting overlap with nbThreads == %u", cctx->requestedParams.nbThreads);
         return ZSTD_CCtxParam_setParameter(&cctx->requestedParams, param, value);
 
+    case ZSTD_p_longDistanceMatching:
+        if (cctx->cdict) return ERROR(stage_wrong);
```
Cyan4973: I understand that long range match is not meant to work in combination with dictionary,

My question is, how bad is it if it happens ?
Does it crashes ? does it produces corrupted data ?
Or is it just inefficient ?

Also : 
what about "one time dictionaries" declared with `ZSTD_refPrefix()` ?

lib/zstd.h
```diff
@@ -976,6 +978,22 @@ typedef enum {
     /* advanced parameters - may not remain available after API update */
     ZSTD_p_forceMaxWindow=1100, /* Force back-reference distances to remain < windowSize,
                               * even when referencing into Dictionary content (default:0) */
+    ZSTD_p_longDistanceMatching,  /* Enable long distance matching. This
```
Cyan4973: So, it's a `0 / 1` switch ?
Maybe `enableLongDistanceMatching` in this case ?

Also : try to describe what happens if expected setting order is not respected,
i.e. if `compressionLevel` is set after `enableLongDistanceMatching`, what happens ?

stellamplau: I'm currently going with the "long distance matching has to be toggled on/off explicitly" semantics, though let me know if you think it makes more sense to be reset with `compressionLevel`. I'll add comments.

lib/compress/zstd_compress.c
```diff
@@ -488,6 +551,9 @@ size_t ZSTD_CCtx_setParametersUsingCCtxParams(
                     cctx, ZSTD_p_overlapSizeLog, params->overlapSizeLog) );
     }
```
Cyan4973: Unrelated : I forgot to mention, I like the new function name.
It respects very well repository's naming convention.

lib/zstd.h
Cyan4973: I think that's fine.
As long as the logic is consistent, and clearly described, it's a good choice.

Cyan4973:
I believe it also deserves a description paragraph (README ?), where the benefits and trade-off of the solution are explained, and benched.

programs/zstdcli.c
```diff
@@ -123,6 +127,7 @@ static int usage_advanced(const char* programName)
     DISPLAY( " -l     : print information about zstd compressed files \n");
 #ifndef ZSTD_NOCOMPRESS
     DISPLAY( "--ultra : enable levels beyond %i, up to %i (requires more memory)\n", ZSTDCLI_CLEVEL_MAX, ZSTD_maxCLevel());
+    DISPLAY( "--long  : enable long distance matching\n");
```
Cyan4973: Should these new commands be also documented in `man zstd` ?

lib/compress/zstd_compress.c
```diff
@@ -305,6 +314,17 @@ size_t ZSTDMT_CCtxParam_setMTCtxParameter(
     ZSTD_CCtx_params* params, ZSTDMT_parameter parameter, unsigned value);
 size_t ZSTDMT_initializeCCtxParameters(ZSTD_CCtx_params* params, unsigned nbThreads);
 
+static size_t ZSTD_ldm_initializeParameters(ldmParams_t* params, U32 enableLdm)
+{
+    assert(LDM_BUCKET_SIZE_LOG <= LDM_BUCKET_SIZE_LOG_MAX);
```
terrelln: nit: This can be a `ZSTD_STATIC_ASSERT`.

lib/compress/zstd_compress.c
```diff
@@ -36,6 +36,14 @@ static const U32 g_searchStrength = 8;   /* control skip over incompressible dat
 #define HASH_READ_SIZE 8
 typedef enum { ZSTDcs_created=0, ZSTDcs_init, ZSTDcs_ongoing, ZSTDcs_ending } ZSTD_compressionStage_e;
 
+#define LDM_BUCKET_SIZE_LOG 3
+#define LDM_BUCKET_SIZE_LOG_MAX 4
```
terrelln: Is this still 4?

lib/compress/zstd_compress.c
stellamplau: It won't crash/produce corrupted data (from some quick testing) with `refPrefix` or cdict. Long distance matching doesn't have any effect except in the compression stage. However, it does modify the window size (which I believe isn't used in loading a dictionary?), and to keep with the semantics of "not modifying compression parameters after loading a dictionary", I didn't allow it to be set if a cdict was loaded.

FYI: There's currently a bug in parameter setting with cdict and `CCtx_loadDictionary` due to the dual use of `copyCCtx` as a function to copy a cctx and to copy a dictionary `refContext`. I believe the `refContext` will only ever contain compression parameters in its `appliedParameters`, so after `ZSTD_CCtx_loadDictionary`, `ZSTD_copyCCtx_internal` is called in `ZSTD_compressBegin_internal` and all other parameters are discarded. I'll send in a fix.

lib/compress/zstdmt_compress.c
```diff
@@ -447,8 +449,6 @@ ZSTDMT_CCtx* ZSTDMT_createCCtx_advanced(unsigned nbThreads, ZSTD_customMem cMem)
     ZSTDMT_initializeCCtxParameters(&mtctx->params, nbThreads);
     mtctx->cMem = cMem;
     mtctx->allJobsCompleted = 1;
-    mtctx->sectionSize = 0;
```
terrelln: This seems like an unrelated change, why do we need it?

lib/zstd.h
terrelln: Should we set `ZSTD_p_longDistanceMatching=1200`?

programs/zstdcli.c
```diff
@@ -720,6 +730,10 @@ int main(int argCount, const char* argv[])
         BMK_setBlockSize(blockSize);
         BMK_setNbThreads(nbThreads);
         BMK_setNbSeconds(bench_nbSeconds);
+        BMK_setLdmFlag(ldmFlag);
+        BMK_setLdmMinMatch(g_ldmMinMatch);
+        BMK_setLdmHashLog(g_ldmHashLog);
+        BMK_setLdmHashEveryLog(g_ldmHashEveryLog);
```
terrelln: Add `if (g_ldmHashEveryLog != LDM_HASHEVERYLOG_DEFAULT)` so we don't depend on `LDM_HASHEVERYLOG_DEFAULT` being the same across modules.

programs/zstdcli.c
```diff
@@ -304,6 +309,9 @@ static unsigned parseCompressionParameters(const char* stringPtr, ZSTD_compressi
         if (longCommandWArg(&stringPtr, "targetLength=") || longCommandWArg(&stringPtr, "tlen=")) { params->targetLength = readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
         if (longCommandWArg(&stringPtr, "strategy=") || longCommandWArg(&stringPtr, "strat=")) { params->strategy = (ZSTD_strategy)(readU32FromChar(&stringPtr)); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
         if (longCommandWArg(&stringPtr, "overlapLog=") || longCommandWArg(&stringPtr, "ovlog=")) { g_overlapLog = readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
+        if (longCommandWArg(&stringPtr, "ldmHashLog=") || longCommandWArg(&stringPtr, "ldmHlog=")) { g_ldmHashLog = readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
```
terrelln: `"ldmhlog"` not `"ldmHlog"`.

programs/zstdcli.c
```diff
@@ -304,6 +309,9 @@ static unsigned parseCompressionParameters(const char* stringPtr, ZSTD_compressi
...
+        if (longCommandWArg(&stringPtr, "ldmSearchLength=") || longCommandWArg(&stringPtr, "ldmSlen=")) { g_ldmMinMatch = readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
```
terrelln: `"ldmslen"`

programs/zstdcli.c
```diff
@@ -304,6 +309,9 @@ static unsigned parseCompressionParameters(const char* stringPtr, ZSTD_compressi
...
+        if (longCommandWArg(&stringPtr, "ldmHashEveryLog=")) { g_ldmHashEveryLog = readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
```
terrelln: Should we have a "short" version for this as well?

lib/zstd.h
Cyan4973: The enum will need a pinned value before reaching "stable" status,
but at this stage, this is still pretty experimental, so it's not too important.
Better expect to change it sometimes in the future.

lib/compress/zstdmt_compress.c
stellamplau: This was to fix the merging with the parameters code. The mtctx->sectionSize field no longer exists and was moved to `CCtx_params`.

lib/compress/zstd_compress.c
```diff
@@ -2984,18 +3115,707 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+            if (ip + rLength < ilimit) {
+                rollingHash = ZSTD_ldm_fillLdmHashTable(
+                                ldmState, rollingHash, lastHashed,
+                                ip + rLength, base, hBits, ldmParams);
+                lastHashed = ip + rLength - 1;
+            }
+            ip += rLength;
+            anchor = ip;
+
+            continue;   /* faster when present ... (?) */
```
terrelln: Can you delete the `continue` and comment unless it is true in this case.

lib/compress/zstd_compress.c
```diff
@@ -2984,18 +3115,707 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
+    }
+
+    /* Overwrite rep */
+    for (i = 0; i < ZSTD_REP_NUM; i++)
+        seqStorePtr->rep[i] = repToConfirm[i];
+
+    ZSTD_ldm_limitTableUpdate(cctx, anchor);
+    ZSTD_ldm_fillFastTables(cctx, anchor);
+
+    lastLiterals = blockCompressor(cctx, anchor, iend - anchor);
```
terrelln: There is something wrong with this call with the optimal parser (the fuzz crashes).

lib/compress/zstd_compress.c
```diff
@@ -2984,18 +3115,707 @@ static ZSTD_blockCompressor ZSTD_selectBlockCompressor(ZSTD_strategy strat, int
...
 {
-    ZSTD_blockCompressor const blockCompressor = ZSTD_selectBlockCompressor(zc->appliedParams.cParams.strategy, zc->lowLimit < zc->dictLimit);
     const BYTE* const base = zc->base;
     const BYTE* const istart = (const BYTE*)src;
     const U32 current = (U32)(istart-base);
+    size_t lastLLSize;
+    const BYTE* anchor;
+    const ZSTD_blockCompressor blockCompressor =
+        zc->appliedParams.ldmParams.enableLdm ?
+            (zc->lowLimit < zc->dictLimit ? ZSTD_compressBlock_ldm_extDict :
```
terrelln: `U32 const extDict = zc->lowLimit < zc->dictLimit;`

stellamplau:
I don't have any changes planned for this PR at the moment. I'm planning on fixing the merge conflict with the dev branch and moving the ldm functions to their own file in a separate PR so code review is easier.

lib/compress/zstd_compress.c
```diff
@@ -1003,6 +1159,12 @@ static void ZSTD_reduceIndex (ZSTD_CCtx* zc, const U32 reducerValue)
 
     { U32 const h3Size = (zc->hashLog3) ? 1 << zc->hashLog3 : 0;
       ZSTD_reduceTable(zc->hashTable3, h3Size, reducerValue); }
+
+    { if (zc->appliedParams.ldmParams.enableLdm) {
+          U32 const ldmHSize = 1 << zc->appliedParams.ldmParams.hashLog;
```
terrelln: nit: `(U32)1`, its a good habit to cast the `1` to whatever type you need. In C++ I write `auto const x = int_type(1) << y`.

Cyan4973:
One last conversion warning to fix before the merge : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.1589/job/wnsvp6riugpcofh0#L38

PR #787, opened 2017-08-10
Bugfixes for Valgrind and TSAN

paulcruz74:
- detach `pthread_mutex_t` variables so that valgrind doesn't complain about accessing memory out of scope
- fix TSAN warnings about potential data race
- change name of the executable from `multi` to `zstd-adaptive` in the help menu


contrib/adaptive-compression/adapt.c
```diff
@@ -982,7 +1018,7 @@ static unsigned readU32FromChar(const char** stringPtr)
 static void help(void)
 {
     PRINT("Usage:\n");
-    PRINT("  ./multi [options] [file(s)]\n");
+    PRINT("  zstd-adaptive [options] [file(s)]\n");
```
Cyan4973: minor detail : 
another way to get the name of executable is to probe `argv[0]`.
Advantage is, it's dynamic.
Though it provides the full path, not just the binary name.

Cyan4973:
Looks good to me

PR #789, opened 2017-08-12
Educational decoder 2

paulcruz74:
- Added Makefile
- fixed some bugs regarding warnings from clang
- simplified to single frame decoding
- created separate API for dictionary creation/initialization/freeing
- modularized sequence execution
- renamed `IO_read_bytes` and `IO_write_bytes`

### DIFF
--- a/doc/educational_decoder/harness.c
+++ b/doc/educational_decoder/harness.c
@@ -106,7 +106,7 @@ int main(int argc, char **argv) {
         return 1;
     }
 
-    dictionary_t* parsed_dict = create_dictionary();
+    dictionary_t* const parsed_dict = create_dictionary();
     if (dict) {
         parse_dictionary(parsed_dict, dict, dict_size);
     }
--- a/doc/educational_decoder/harness.c
+++ b/doc/educational_decoder/harness.c
@@ -106,9 +106,15 @@ int main(int argc, char **argv) {
         return 1;
     }
 
+    dictionary_t* parsed_dict = create_dictionary();
+    if (dict) {
+        parse_dictionary(parsed_dict, dict, dict_size);
+    }
     size_t decompressed =
         ZSTD_decompress_with_dict(output, decompressed_size,
-                                  input, input_size, dict, dict_size);
+                                  input, input_size, parsed_dict);
+
+    free_dictionary(parsed_dict);
 
     write_file(argv[2], output, decompressed);
 
--- a/doc/educational_decoder/harness.c
+++ b/doc/educational_decoder/harness.c
@@ -87,7 +87,7 @@ int main(int argc, char **argv) {
     }
 
     size_t decompressed_size = ZSTD_get_decompressed_size(input, input_size);
-    if (decompressed_size == -1) {
+    if (decompressed_size == (size_t)-1) {
         decompressed_size = MAX_COMPRESSION_RATIO * input_size;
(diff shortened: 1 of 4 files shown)


doc/educational_decoder/harness.c
```diff
@@ -106,9 +106,15 @@ int main(int argc, char **argv) {
         return 1;
     }
 
+    dictionary_t* parsed_dict = create_dictionary();
```
Cyan4973: minor : `dictionary_t* const` since `parsed_dict` is not going to be changed later.

doc/educational_decoder/zstd_decompress.c
```diff
@@ -1303,145 +1272,148 @@ static void execute_sequences(frame_context_t *const ctx, ostream_t *const out,
...
         {
-            // If the sequence asks for more literals than are left, the
-            // sequence must be corrupted
-            if (seq.literal_length > IO_istream_len(&litstream)) {
-                CORRUPTION();
-            }
+            const u32 literals_size = copy_literals(seq, &litstream, out);
+            total_output += literals_size;
+        }
+        size_t offset = compute_offset(seq, offset_hist);
```
Cyan4973: minor : `const` whenever you can

doc/educational_decoder/zstd_decompress.c
```diff
@@ -1303,145 +1272,148 @@ static void execute_sequences(frame_context_t *const ctx, ostream_t *const out,
...
-                // "There is an exception though, when current sequence's
-                // literals length is 0. In this case, repeated offsets are
-                // shifted by one, so Repeated_Offset1 becomes Repeated_Offset2,
-                // Repeated_Offset2 becomes Repeated_Offset3, and
-                // Repeated_Offset3 becomes Repeated_Offset1 - 1_byte."
-                idx++;
-            }
+    // Copy any leftover literals
+    {
+        size_t len = IO_istream_len(&litstream);
```
Cyan4973: minor, but let's detail since this is a learning opportunity : 

This part is slightly more complex than necessary.
I would expect here to find something like `copy_literals()`.

`copy_literals()` cannot be used "as is" because it requires a `sequence_command_t` input type, and you only have a `size_t len`.
However, `copy_literals()` effectively only needs `seq.literal_length`.
It could be refactored to only use this length, instead of the full `sequence_command_t`.
In which case, it becomes usable for leftover literals.

If you believe it's valuable to not get into the details of `sequence_command_t`, you could nest it under some `copy_literals_from_seq()`, which just extracts `seq.literal_length` and pass it to the new `copy_literals()`.
However, I see just above: `size_t match_length = seq.match_length;`
which means the structure of `sequence_command_t` is already a dependency of `execute_sequences()`. So it doesn't seem worthwhile to avoid it.

doc/educational_decoder/zstd_decompress.c
```diff
@@ -1303,145 +1272,148 @@ static void execute_sequences(frame_context_t *const ctx, ostream_t *const out,
...
-                memcpy(write_ptr, ctx->dict_content + dict_offset, dict_copy);
-                write_ptr += dict_copy;
-                match_length -= dict_copy;
-            }
-        } else if (offset > ctx->header.window_size) {
-            CORRUPTION();
+    return seq.literal_length;
+}
+
+static size_t compute_offset(sequence_command_t seq, u64 *const offset_hist) {
```
Cyan4973: 👍

PR #793, opened 2017-08-14
fix typo in lz4 support code

codicodi:


Cyan4973:
Good catch, thanks @codicodi !

PR #805, opened 2017-08-19
fix #803 : wrong example in huffman bitstream section

### LINKED ISSUE #803: Question about Huffman encoding example in Zstd compression format documentation
The excellent compression format documentation contains following example in the section 'Huffman-coded Streams':

> For example, if the literal sequence "0145" was encoded using the prefix codes above,
> it would be encoded as:
> ```
> 00000001 01110000
> ```
> 
> |Symbol  |   5  |   4  |  1 | 0 | Padding |
> |--------|------|------|----|---|---------|
> |Encoding|`0000`|`0001`|`01`|`1`| `10000` |

It appears that the example intends to represent a byte from lowest to highest bit. My understanding is however that the encodings or codes are represented from highest bit to lowest bit and would need to be reversed in the example, which they aren't. In the example it is unclear how the decoder after reading `1` should decode `01` instead of another `1`.

Does it make sense to represent the bytes from highest to lowest bit as the codes? In that case the example encoding will be:
```
00010000 00001101
```
A byte is represented with the highest bit first and it is also obvious that the decoder must read `01` after having read `1`.

Cyan4973:
reported by @ulikunitz

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -16,7 +16,7 @@ Distribution of this document is unlimited.
 
 ### Version
 
-0.2.5 (31/03/17)
+0.2.6 (19/08/17)
 
 
 Introduction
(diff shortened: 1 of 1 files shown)


ulikunitz:
I suggest to change the statement in front of the alternative representation to:

Here is an alternative representation with the symbol codes separated by underscore.

Otherwise it looks good to me.

PR #810, opened 2017-08-21
Move parameters into a single (opaque) structure

stellamplau:
This is for code review only and is not ready to be merged (more testing, function renaming, and completing TODOs needs to be done).

This allows compression parameters (other than just the ZSTD_parameters) to be added/passed around.


lib/compress/zstd_compress.c
```diff
@@ -3224,28 +3424,41 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
...
+                                                pledgedSrcSize);
 }
 
 
 size_t ZSTD_compressBegin_usingDict(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, int compressionLevel)
 {
     ZSTD_parameters const params = ZSTD_getParams(compressionLevel, 0, dictSize);
-    return ZSTD_compressBegin_internal(cctx, dict, dictSize, ZSTD_dm_auto, NULL,
-                                       params, 0, ZSTDb_not_buffered);
+    return ZSTD_compressBegin_advanced(cctx, dict, dictSize, params, 0);
```
terrelln: I'm not sure that this should be calling `ZSTD_compressBegin_advanced()`, cc @Cyan4973.

lib/compress/zstd_compress.c
```diff
@@ -3418,44 +3645,46 @@ static size_t ZSTD_initCDict_internal(
     }
     cdict->dictContentSize = dictSize;
 
-    {   ZSTD_frameParameters const fParams = { 0 /* contentSizeFlag */,
+    {   ZSTD_frameParameters const fParams = { 0 /*contentSizeFlag */,
```
terrelln: White space.

lib/compress/zstd_compress.c
```diff
@@ -3525,17 +3755,16 @@ ZSTD_CDict* ZSTD_initStaticCDict(void* workspace, size_t workspaceSize,
     }
     cdict->refContext = ZSTD_initStaticCCtx(ptr, cctxSize);
 
-    if (ZSTD_isError( ZSTD_initCDict_internal(cdict,
-                                              dict, dictSize,
-                                              1 /* byReference */, dictMode,
-                                              cParams) ))
+    if (ZSTD_isError( ZSTD_initCDict_internal(cdict, dict, dictSize,
```
terrelln: This diff is already pretty large, can we avoid all of the formatting-only changes?

lib/compress/zstd_compress.c
```diff
@@ -3681,24 +3914,27 @@ size_t ZSTD_initCStream_internal(ZSTD_CStream* zcs,
...
+            ZSTD_CCtx_params const cdictParams = ZSTD_getCCtxParamsFromCDict(cdict);
             params.cParams = cdictParams.cParams;  /* cParams are enforced from cdict */
         }
         ZSTD_freeCDict(zcs->cdictLocal);
         zcs->cdictLocal = NULL;
         zcs->cdict = cdict;
     }
-
     zcs->requestedParams = params;
-    zcs->compressionLevel = ZSTD_CLEVEL_CUSTOM;
```
terrelln: We still need a line like this, right? Otherwise we will lose the dictionary when a user calls `ZSTD_resetCStream()`. Can we add a test that will catch this?

lib/compress/zstd_compress.c
```diff
@@ -3728,30 +3965,32 @@ size_t ZSTD_initCStream_advanced(ZSTD_CStream* zcs,
...
+    cctxParams.compressionLevel = compressionLevel;
+    return ZSTD_initCStream_internal(zcs, dict, dictSize, NULL, cctxParams, 0);
 }
 
 size_t ZSTD_initCStream_srcSize(ZSTD_CStream* zcs, int compressionLevel, unsigned long long pledgedSrcSize)
 {
     ZSTD_parameters params = ZSTD_getParams(compressionLevel, pledgedSrcSize, 0);
     params.fParams.contentSizeFlag = (pledgedSrcSize>0);
-    return ZSTD_initCStream_internal(zcs, NULL, 0, NULL, params, pledgedSrcSize);
+    return ZSTD_initCStream_advanced(zcs, NULL, 0, params, pledgedSrcSize);
```
terrelln: This should still call `ZSTD_initCStream_internal()`, since we don't need to check the parameters.

lib/compress/zstd_compress.c
stellamplau: Yeah I wasn't sure why `ZSTD_initCStream_usingDict()` set the compression level only for it to be reset in `ZSTD_initCStream_internal()`.

lib/compress/zstd_compress.c
stellamplau: As far as I can tell, this just causes an unnecessary `ZSTD_checkCParams`. I can call `compressBegin_internal` directly.

lib/common/zstd_internal.h
```diff
@@ -286,6 +286,28 @@ typedef struct {
...
 
+struct ZSTD_CCtx_params_s {
+    ZSTD_compressionParameters cParams;
+    ZSTD_frameParameters fParams;
+
+    int compressionLevel;
+    U32 forceWindow;           /* force back-references to respect limit of
+                                * 1<<wLog, even for dictionary */
+    /* Dictionary */
+    U32 dictContentByRef;
```
Cyan4973: As mentioned, it feels strange to have this property "separated" from the pointer it's supposed to impact.
It feels better to keep it as a property of the `loadDictionary` command.

That means, either creating a new `ZSTD_CCtx_loadDictionary_byReference()`,
or changing `ZSTD_CCtx_loadDictionary()` prototype to add a loading parameter (`byCopy, byRef`),
or add a `ZSTD_CCtx_loadDictionary_advanced()` variant with said loading parameter.

For reference, our current `ZSTD_createCDict*()` API contains all 3 variants.

lib/compress/zstdmt_compress.c
```diff
@@ -329,10 +341,18 @@ void ZSTDMT_compressChunk(void* jobDescription)
...
     } else {  /* srcStart points at reloaded section */
         if (!job->firstChunk) job->params.fParams.contentSizeFlag = 0;  /* ensure no srcSize control */
-        {   size_t const dictModeError = ZSTD_setCCtxParameter(cctx, ZSTD_p_forceRawDict, 1);  /* Force loading dictionary in "content-only" mode (no header analysis) */
-            size_t const initError = ZSTD_compressBegin_advanced(cctx, job->srcStart, job->dictSize, job->params, job->fullFrameSize);
-            if (ZSTD_isError(initError) || ZSTD_isError(dictModeError)) { job->cSize = initError; goto _endJob; }
-            ZSTD_setCCtxParameter(cctx, ZSTD_p_forceWindow, 1);
+        { ZSTD_CCtx_params jobParams = job->params;
+          /* Force loading dictionary in "content-only" mode (no header analysis) */
+          size_t const dictModeError =
+              ZSTD_CCtxParam_setParameter(&jobParams, ZSTD_p_dictMode, 1);
```
Cyan4973: I would prefer the use of relevant enum type `ZSTD_dictMode_e` (value `ZSTD_dm_rawContent` for this case).
It's easier to understand than `1`.

lib/compress/zstdmt_compress.c
```diff
@@ -329,10 +341,18 @@ void ZSTDMT_compressChunk(void* jobDescription)
...
+          size_t const forceWindowError =
+              ZSTD_CCtxParam_setParameter(&jobParams, ZSTD_p_forceMaxWindow, !job->firstChunk);
+          /* Note: ZSTD_setCCtxParameter() should not be used here.
+           * ZSTD_compressBegin_advanced_internal() copies the ZSTD_CCtx_params
+           * directly to appliedParams. */
+          size_t const initError = ZSTD_compressBegin_advanced_internal(cctx, job->srcStart, job->dictSize, jobParams, job->fullFrameSize);
```
Cyan4973: Yes, this construction is clearer than previous hack.
All parameters are present in `jobParams`, instead of being partially in `params` and partially in `cctx`.
That's an improvement.

lib/common/zstd_internal.h
stellamplau: If I understand correctly, we want to remove `ZSTD_p_refDictContent` completely, as well as `dictContentByRef` from both `ZSTD_CCtx_params` and `ZSTD_CCtx`?

Cyan4973: Yes

lib/compress/zstdmt_compress.c
```diff
@@ -528,17 +548,18 @@ static unsigned computeNbChunks(size_t srcSize, unsigned windowLog, unsigned nbT
...
-                               unsigned overlapLog)
+static size_t ZSTDMT_compress_advanced_internal(
+                ZSTDMT_CCtx* mtctx,
+                void* dst, size_t dstCapacity,
+          const void* src, size_t srcSize,
+          const ZSTD_CDict* cdict,
+                ZSTD_CCtx_params const cctxParams,
+                unsigned overlapLog)
 {
+    ZSTD_CCtx_params const requestedParams = ZSTDMT_makeJobCCtxParams(cctxParams);
```
Cyan4973: OK, this is where it's a bit confusing. I believe we can clarify this.

If I do understand correctly, `ZSTDMT_makeJobCCtxParams()` creates a "souped down" version of `cctxParams`, to be used by workers. Basically it throws away anything that's not useful or misleading for workers, such as `nbThreads`, since each worker is a single thread.
So let's call it this way, this variable should be called `jobParams`, and should only used to provide parameters to worker through the `jobs` table.
For all other uses, we should continue reading values from `cctxParams`, which is our "source of truth".

lib/compress/zstdmt_compress.c
```diff
@@ -528,17 +548,18 @@ static unsigned computeNbChunks(size_t srcSize, unsigned windowLog, unsigned nbT
...
-                         const ZSTD_CDict* cdict,
-                               ZSTD_parameters const params,
-                               unsigned overlapLog)
+static size_t ZSTDMT_compress_advanced_internal(
+                ZSTDMT_CCtx* mtctx,
+                void* dst, size_t dstCapacity,
+          const void* src, size_t srcSize,
+          const ZSTD_CDict* cdict,
+                ZSTD_CCtx_params const cctxParams,
+                unsigned overlapLog)
```
Cyan4973: I think we could go "all in", and use `cctxParams` to transmit _all_ parameters.
This includes `overlapLog`, and also `nbThreads` and `jobSize`.
Prototype and body should be adapted to reflect it.

lib/compress/zstdmt_compress.c
```diff
@@ -683,23 +720,22 @@ static void ZSTDMT_waitForAllJobsCompleted(ZSTDMT_CCtx* zcs)
...
-
-/** ZSTDMT_initCStream_internal() :
- *  internal usage only */
-size_t ZSTDMT_initCStream_internal(ZSTDMT_CCtx* zcs,
-                    const void* dict, size_t dictSize, const ZSTD_CDict* cdict,
-                    ZSTD_parameters params, unsigned long long pledgedSrcSize)
+size_t ZSTDMT_initCStream_internal(
+        ZSTDMT_CCtx* zcs, const void* dict, size_t dictSize,
+        const ZSTD_CDict* cdict, ZSTD_CCtx_params cctxParams,
+        unsigned long long pledgedSrcSize)
```
Cyan4973: This leads me to a question : to we want `pledgedSrcSize` outside or within `cctxParams` ?

lib/compress/zstdmt_compress.c
stellamplau: I was considering moving `ZSTDMT_makeJobCCtxParams` to `zstd_compress.c` to make it the responsibility of `zstd_compress` to pass only relevant parameters to `zstdmt`. Do you think that would make more sense?

lib/compress/zstdmt_compress.c
```diff
@@ -708,14 +744,14 @@ size_t ZSTDMT_initCStream_internal(ZSTDMT_CCtx* zcs,
         zcs->allJobsCompleted = 1;
     }
 
-    zcs->params = params;
+    zcs->params = requestedParams;
```
Cyan4973: Except if I miss something, our "source of truth" should be `cctxParams`

lib/zstd.h
```diff
@@ -1096,6 +1102,46 @@ size_t ZSTD_compress_generic_simpleArgs (
...
+ZSTDLIB_API ZSTD_CCtx_params* ZSTD_createCCtxParams(void);
+
+/*! ZSTD_resetCCtxParams() :
+ *  Reset params to default, with the default compression level. */
+ZSTDLIB_API size_t ZSTD_resetCCtxParams(ZSTD_CCtx_params* params);
+
+/*! ZSTD_initCCtxParams() :
+ *  Set the compression and frame parameters of cctxParams according to params.
+ *  All other parameters are reset to their default values. */
+ZSTDLIB_API size_t ZSTD_initCCtxParams(ZSTD_CCtx_params* cctxParams, ZSTD_parameters params);
```
Cyan4973: Consider also a variant initializing `cctxParams` from `cLevel`. It's probably going to be useful for a user-facing prototype.

Also : if the transformation is guaranteed to succeed, we don't need an error code, so we should probably use this opportunity to return the target type directly.

So it would give something like this : 
```
ZSTD_CCtx_params ZSTD_initCCtxParams(int compressionLevel);
ZSTD_CCtx_params ZSTD_initCCtxParams_advanced(ZSTD_parameters params);
```

lib/compress/zstd_compress.c
```diff
@@ -202,29 +197,109 @@ size_t ZSTD_sizeof_CStream(const ZSTD_CStream* zcs)
 /* private API call, for dictBuilder only */
 const seqStore_t* ZSTD_getSeqStore(const ZSTD_CCtx* ctx) { return &(ctx->seqStore); }
 
-static ZSTD_parameters ZSTD_getParamsFromCCtx(const ZSTD_CCtx* cctx) { return cctx->appliedParams; }
-
 /* older variant; will be deprecated */
+/* Both requested and applied params need to be set as this function can be
```
Cyan4973: Question : is it time to consider retiring it ?

lib/compress/zstdmt_compress.c
Cyan4973: Not sure about this one.
I believe `zstdmt` should have all cards in hands,
the way it distributes jobs across threads, with resulting consequences on which parameters are valid or not, is its own responsibility.

lib/compress/zstd_compress.c
stellamplau: Did it only still exist to support zstdmt? It's the only way to set forceWindow/forceRawDict with the old api (I think)

lib/compress/zstdmt_compress.c
stellamplau: It is a parameter, albeit maybe a "single use" one. When calling `ZSTD_CCtx_reset`, we would probably still want to reset it, which might be strange since all the other parameters persist?

lib/compress/zstd_compress.c
Cyan4973: Yes, it was created for `zstdmt`.
You may now be able to transmit these parameters with your new `ZSTD_CCtx_params` structure.

lib/compress/zstdmt_compress.c
Cyan4973: Yes, that's a good point.
It's indeed only valid for the current session.
So let's keep it outside of `ZSTD_CCtx_params`, since its lifetime is different.

lib/common/zstd_internal.h
```diff
@@ -286,6 +286,27 @@ typedef struct {
...
 
+struct ZSTD_CCtx_params_s {
+    ZSTD_compressionParameters cParams;
+    ZSTD_frameParameters fParams;
+
+    int compressionLevel;
+    U32 forceWindow;           /* force back-references to respect limit of
+                                * 1<<wLog, even for dictionary */
+    /* Dictionary */
+    ZSTD_dictMode_e dictMode;   /* select restricting dictionary to "rawContent"
```
Cyan4973: The same question applies here :
Is it a property of the loading operation ?
In which case, would it be better to make it an argument of the load operation ?

lib/zstd.h
```diff
@@ -496,16 +498,19 @@ ZSTDLIB_API size_t ZSTD_sizeof_DDict(const ZSTD_DDict* ddict);
  *  It will also consider src size to be arbitrarily "large", which is worst case.
  *  If srcSize is known to always be small, ZSTD_estimateCCtxSize_advanced() can provide a tighter estimation.
  *  ZSTD_estimateCCtxSize_advanced() can be used in tandem with ZSTD_getCParams() to create cParams from compressionLevel.
+ *  ZSTD_estimateCCtxSize_advanced_opaque() can be used in tandem with ZSTD_CCtxParam_setParameter(). Only single-threaded compression is supported. This function will return 0 if ZSTD_p_nbThreads is set to a value > 1.
  *  Note : CCtx estimation is only correct for single-threaded compression */
 ZSTDLIB_API size_t ZSTD_estimateCCtxSize(int compressionLevel);
 ZSTDLIB_API size_t ZSTD_estimateCCtxSize_advanced(ZSTD_compressionParameters cParams);
+ZSTDLIB_API size_t ZSTD_estimateCCtxSize_advanced_opaque(const ZSTD_CCtx_params* params);
```
Cyan4973: We could change these 2 function names with something like `_usingCParams` and `_usingCCtxParams`.
It will help clarity.

lib/zstd.h
```diff
@@ -515,6 +520,7 @@ ZSTDLIB_API size_t ZSTD_estimateDCtxSize(void);
  *         In this case, get total size by adding ZSTD_estimate?DictSize */
 ZSTDLIB_API size_t ZSTD_estimateCStreamSize(int compressionLevel);
 ZSTDLIB_API size_t ZSTD_estimateCStreamSize_advanced(ZSTD_compressionParameters cParams);
```
Cyan4973: Same comment here

lib/common/zstd_internal.h
Cyan4973: As discussed : 
with now 2 possible parameters, loading dictionary is an operation featuring 6 different variants.
So it's no longer practical to use function name to distinguish them

Consider a new prototype `ZSTD_CCtx_loadDictionary_advanced()` with these 2 parameters as arguments.
You can keep existing `ZSTD_CCtx_loadDictionary_byReference()`, which presumes a standard value for `dictMode`.

Prefer using `enum` as argument type, as opposed to integer values, since they are more readable.
The tricky point about `enum` is selecting a correct prefix, so that the naming doesn't pollute namespace.
An example : `{ ZSTD_dl_byRef, ZSTD_dl_byCopy } `.

You can then retrofit the relatively similar `ZSTD_createCDict_advanced()` to ensure it uses the same argument types, in order to keep the API consistent.

lib/compress/zstd_compress.c
```diff
@@ -202,178 +197,325 @@ size_t ZSTD_sizeof_CStream(const ZSTD_CStream* zcs)
...
+            cctx->requestedParams.nbThreads = 1;
+            cctx->mtctx = ZSTDMT_createCCtx_advanced(value, cctx->customMem);
+            if (cctx->mtctx == NULL) return ERROR(memory_allocation);
+        }
+
+        /* Need to initialize overlapSizeLog */
+        return ZSTDMT_initializeCCtxParameters(&cctx->requestedParams, value);
+
+    case ZSTD_p_jobSize:
+        if (cctx->requestedParams.nbThreads <= 1) return ERROR(parameter_unsupported);
```
Cyan4973: do we need to make this test here if it's already taken care of inside `ZSTD_CCtxParam_setParameter()` ?

lib/compress/zstd_compress.c
```diff
@@ -202,178 +197,325 @@ size_t ZSTD_sizeof_CStream(const ZSTD_CStream* zcs)
...
+        DEBUGLOG(5, " setting nbThreads : %u", value);
+#ifndef ZSTD_MULTITHREAD
+        if (value > 1) return ERROR(parameter_unsupported);
+#endif
+        if ((value>1) && (cctx->requestedParams.nbThreads != value)) {
+            if (cctx->staticSize)  /* MT not compatible with static alloc */
+                return ERROR(parameter_unsupported);
+            ZSTDMT_freeCCtx(cctx->mtctx);
+            cctx->requestedParams.nbThreads = 1;
+            cctx->mtctx = ZSTDMT_createCCtx_advanced(value, cctx->customMem);
```
Cyan4973: Changing `nbThreads` is effectively applied immediately here.
The inherited implementation did it this way because there used to be a single `cctx->nbThreads` variable.

But now, there is a stronger separation between `requestedParams` and `appliedParams`.
So I suspect it would be better to follow the same logic for `nbThreads` as for all other parameters,
aka, "deferred evaluation". Setting the value is just that, setting the value into `requestedParams`.

Which means, translating `nbThreads` into a working `ZSTDMT_CCtx*` context should be done when the compression job actually starts.

lib/compress/zstd_compress.c
```diff
@@ -202,178 +197,325 @@ size_t ZSTD_sizeof_CStream(const ZSTD_CStream* zcs)
...
 
+/**
+ * This function should be updated whenever ZSTD_CCtx_params is updated.
+ * Parameters are copied manually before the dictionary is loaded.
+ * The multithreading parameters jobSize and overlapSizeLog are set only if
+ * nbThreads > 1.
+ *
+ * Pledged srcSize is treated as unknown.
+ */
+size_t ZSTD_CCtx_applyCCtxParams(ZSTD_CCtx* cctx, const ZSTD_CCtx_params* params)
```
Cyan4973: comment about naming : 
within `CCtx`, we have now `requestedParams` and `appliedParams`.
From an API user perspective, this distinction is invisible.
So `applyCCtxParams` looks fine : just applies parameters from `ZSTD_CCtx_params*` into `ZSTD_CCtx*`. Clear.

From a Zstandard developer perspective though, `cctx->appliedParams` exists, and `ZSTD_CCtx_applyCCtxParams()` seems to refer to it, due to name similarity.
But that's not the case : `ZSTD_CCtx_applyCCtxParams()` only sets parameters inside `cctx->requestedParams`, as it should.

I was wondering if naming could be improved to please both audiences.

lib/compress/zstd_compress.c
```diff
@@ -116,15 +114,12 @@ struct ZSTD_CCtx_s {
     U32    frameEnded;
 
     /* Dictionary */
-    ZSTD_dictMode_e dictMode; /* select restricting dictionary to "rawContent" or "fullDict" only */
-    U32 dictContentByRef;
     ZSTD_CDict* cdictLocal;
     const ZSTD_CDict* cdict;
     const void* prefix;
```
Cyan4973: minor detail : 
I just realised that `prefix`, `prefixSize` and `pledgedSrcSizePlusOne` share the same property : 
they are valid only once, for the next compression job.
After that, they must be reset to `0`.

In the future, it might be useful to join then into some identical "category".

lib/common/zstd_internal.h
stellamplau: Re `dictMode`: 
- `ZSTD_CCtx_refPrefix` allows the dict mode to be set via `ZSTD_p_dictMode`, which is relevant in `ZSTD_compress_generic` for the non-multithreaded version. The multithreaded version seems to just treat the mode as `ZSTD_dm_auto`. Was this the intended functionality? Allowing the dict mode option here effectively implies adding a "single-use" dictMode parameter when `ZSTD_CCtx_refPrefix` is used.
- The old functionality in `zstdmt_compress` with setting `ZSTD_p_forceRawDict` in `ZSTDMT_compressChunk` seems like it was a no-op since `ZSTD_compressBegin_advanced` just used `ZSTD_dm_auto`. Am I missing something? If not, is it important to set to raw mode?

Cyan4973: These are very good points @stellamplau , and it's pretty complex to track.

`ZSTD_CCtx_refPrefix()` main objective is to make it possible to compress a content `C` given another content `P`. This content is typically a very similar copy of `C` such as a previous revision, or simply the previous content in a sequential order controlled by the application. This second use case is representative of `ZSTDMT` methodology.

The primary intention is that such content is "anything", meaning not a formatted dictionary. Hence, it should be loaded using mode `ZSTD_dm_rawContent`. Following the code flow, we can see it ends up in `ZSTD_compress_insertDictionary()`, which accepts `ZSTD_dictMode_e dictMode` as argument, transmitted from `ZSTD_resetCStream_internal()`. Meaning it's not necessarily `ZSTD_dm_rawContent`.

With the disappearance of `dictMode` from `CCtx`, it becomes a problem. Which parameter to provide ?
Well, for `ZSTD_CCtx_refPrefix()`, a quickfix could be to always load in `ZSTD_dm_rawContent` mode. It closes the other mode, but I still have to find a use case which actually needs that.

A complexity is that `ZSTD_compress_usingDict()` also uses a "throw-away" dictionary, but is supposed to load it in `ZSTD_dm_auto` mode. And they currently share the same loading pipeline. Something to look at.

Cyan4973: Part 2 : It seems it was a broken behavior.
`ZSTD_dm_rawContent` mode was created with `zstdmt` in mind.
`zstdmt` loads a part of previous job as "dictionary", but we want to load it as pure content, with no specific metadata. That's what happen with default `ZSTD_dm_auto`, as long as the content doesn't start with the same magic number as a dictionary built with `zstd`.
The chances of this happening are very small (1 in 4 Billions if values are evenly distributed, but since they are not, and since the magic number is selected to avoid any kind of pattern, it's likely even less). So in practice `ZDSTD_dm_auto` is a very close cousin.
But still, what we want is `ZSTD_dm_rawContent`.

note : we should have a test to dictionary loading modes.
Check that, when dictionary loading is set with `ZSTD_dm_rawContent` mode, when presented with a content starting with the same magic number as a dictionary, it nonetheless load it as raw content.
Method : take a random content, overwrite first bytes with magic number, load it with `ZSTD_dm_auto`, it should fail. Now with `ZSTD_dm_rawContent`, it should work.

lib/compress/zstd_compress.c
```diff
@@ -3227,28 +3393,42 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
...
+    return ZSTD_compressBegin_advanced_internal(cctx, dict, dictSize, ZSTD_dm_auto,
+                                                cctxParams,
+                                                pledgedSrcSize);
 }
 
-
 size_t ZSTD_compressBegin_usingDict(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, int compressionLevel)
 {
     ZSTD_parameters const params = ZSTD_getParams(compressionLevel, 0, dictSize);
+    ZSTD_CCtx_params cctxParams =
```
terrelln: nit: `const`

lib/compress/zstd_compress.c
```diff
@@ -3325,9 +3505,13 @@ static size_t ZSTD_compress_internal (ZSTD_CCtx* cctx,
                          const void* dict,size_t dictSize,
                                ZSTD_parameters params)
 {
-    CHECK_F( ZSTD_compressBegin_internal(cctx, dict, dictSize, ZSTD_dm_auto, NULL,
-                                         params, srcSize, ZSTDb_not_buffered) );
-    return ZSTD_compressEnd(cctx, dst,  dstCapacity, src, srcSize);
+    ZSTD_CCtx_params cctxParams =
```
terrelln: nit: `const`

lib/compress/zstd_compress.c
```diff
@@ -3420,19 +3613,22 @@ static size_t ZSTD_initCDict_internal(
 
     {   ZSTD_frameParameters const fParams = { 0 /* contentSizeFlag */,
                     0 /* checksumFlag */, 0 /* noDictIDFlag */ };  /* dummy */
-        ZSTD_parameters const params = ZSTD_makeParams(cParams, fParams);
+        ZSTD_CCtx_params cctxParams = cdict->refContext->requestedParams;
+        cctxParams.fParams = fParams;
```
terrelln: Do you need to set `fParams` here?

lib/compress/zstd_compress.c
```diff
@@ -3682,23 +3876,23 @@ size_t ZSTD_initCStream_internal(ZSTD_CStream* zcs,
         }
         ZSTD_freeCDict(zcs->cdictLocal);
         zcs->cdictLocal = ZSTD_createCDict_advanced(dict, dictSize,
-                                            zcs->dictContentByRef, zcs->dictMode,
+                                            ZSTD_dlm_byCopy, ZSTD_dm_auto,
```
terrelln: Previously `ZSTD_initCStream()` respected the dict loading parameters set by the new API, but now it doesn't. Is this okay? Without mixing the new API and the old CStream API there is no way to set these parameters, but if someone mixes the two APIs, the behavior changes.

lib/compress/zstd_compress.c
```diff
@@ -3728,30 +3923,32 @@ size_t ZSTD_initCStream_advanced(ZSTD_CStream* zcs,
                                  const void* dict, size_t dictSize,
                                  ZSTD_parameters params, unsigned long long pledgedSrcSize)
 {
+    ZSTD_CCtx_params cctxParams =
```
terrelln: nit `const`

lib/compress/zstdmt_compress.c
```diff
@@ -186,6 +186,18 @@ static void ZSTDMT_releaseBuffer(ZSTDMT_bufferPool* bufPool, buffer_t buf)
     ZSTD_free(buf.start, bufPool->cMem);
 }
 
+/* Sets parameters relevant to the compression job, initializing others to
+ * default values. Notably, nbThreads should probably be zero. */
+static ZSTD_CCtx_params ZSTDMT_makeJobCCtxParams(ZSTD_CCtx_params const params)
+{
+    ZSTD_CCtx_params jobParams;
+    memset(&jobParams, 0, sizeof(ZSTD_CCtx_params));
```
terrelln: nit: `memset()`

lib/compress/zstdmt_compress.c
```diff
@@ -714,7 +768,7 @@ size_t ZSTDMT_initCStream_internal(ZSTDMT_CCtx* zcs,
         DEBUGLOG(4,"cdictLocal: %08X", (U32)(size_t)zcs->cdictLocal);
         ZSTD_freeCDict(zcs->cdictLocal);
         zcs->cdictLocal = ZSTD_createCDict_advanced(dict, dictSize,
-                                                    0 /* byRef */, ZSTD_dm_auto,   /* note : a loadPrefix becomes an internal CDict */
+                                                    ZSTD_dlm_byCopy, ZSTD_dm_rawContent, /* note : a loadPrefix becomes an internal CDict */
```
terrelln: Why are we forcing `rawContent` now?

Cyan4973: This might be related to the discussion @stellamplau and I had yesterday : 
https://github.com/facebook/zstd/pull/810#discussion_r135933032
In short, `ZSTD_refPrefix()` now assumes `ZSTD_dm_rawContent`.

Still I don't think it's the right place to enforce this mode.
It should be enforced from the calling side, aka `ZSTD_compress_generic()`, which means `ZSTDMT_initCStream_internal()` should have another argument, `dictMode` (only useful when `dict` is defined)

The reason is, `ZSTDMT_initCStream_internal()` is also accessible through another entry point, `ZSTDMT_initCStream_advanced()`, which makes it possible to select a `dict`, but for which the implied mode is `ZSTD_dm_auto`.

Passing the `dictMode` is the only way to make the distinction between these 2 entry points.

stellamplau: Yeah I was just about to comment that forcing `rawContent` here was a mistake and the function should probably contain a `dictMode` parameter. I neglected the `ZSTDMT_initCStream_advanced` caller.

lib/compress/zstd_compress.c
Cyan4973: I think `ZSTD_initCStream_internal()` is never invoked from the new advanced API,
which goes straight to `ZSTD_resetCStream_internal()` instead.

`ZSTD_initCStream_internal()` is called by : 
- `ZSTD_initCStream_usingCDict_advanced()` : which doesn't use a `dict`, since it refers to a `cdict`
- `ZSTD_initCStream_advanced()` : which can use `dict`, and implies `ZSTD_dm_auto` mode. `ZSTD_dlm_byCopy` loading mode is implied too, just because it's safer and therefore doesn't require the user to pay attention to its `dict` buffer lifetime.
- `ZSTD_initCStream_usingDict()` : which also uses `dict` and implies the `ZSTD_dm_auto` and `ZSTD_dlm_byCopy`.
- `ZSTD_initCStream_srcSize()` : which doesn't use a `dict`.

So, in the 2 valid cases, these parameters are correct.

Still, that's the situation _today_. What if it changes in the future ?
Someone may try to use this interface for some future feature we still don't know about.
So, at a minimum, these implied parameters should be documented.

As forward thinking, they could also become arguments to the function, so they could be modified in the future, but it introduces additional complexity for something which is not useful immediately. So is it worth it ?

lib/compress/zstd_compress.c
```diff
@@ -516,56 +654,79 @@ ZSTD_compressionParameters ZSTD_adjustCParams(ZSTD_compressionParameters cPar, u
...
-size_t ZSTD_estimateCCtxSize_advanced(ZSTD_compressionParameters cParams)
+size_t ZSTD_estimateCCtxSize_advanced_usingCCtxParams(const ZSTD_CCtx_params* params)
 {
-    size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << cParams.windowLog);
-    U32    const divider = (cParams.searchLength==3) ? 3 : 4;
-    size_t const maxNbSeq = blockSize / divider;
-    size_t const tokenSpace = blockSize + 11*maxNbSeq;
+    /* Estimate CCtx size is supported for single-threaded compression only. */
+    if (params->nbThreads > 1) {
+      return 0;
```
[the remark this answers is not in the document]
Cyan4973: Good catch ! It should be an error code.

lib/compress/zstdmt_compress.c
terrelln: Lets add a test for `ZSTDMT_initCStream_advanced()` with a dictionary, since it looks like we don't have one already.

PR #812, opened 2017-08-22
fixed extraordinary scenario where all fields use maximum nbBits

Cyan4973:
can only happen if windowLog>=27  (level 22 --ultra)


terrelln:
I've confirmed that this fixes the problem.

PR #814, opened 2017-08-23
Fix displayLevel overflow

dscheg:
When `ZDICT_trainFromBuffer` is called `notificationLevel` set to `0` which leads to int overflow. As a result there is some stderr-garbage on building dict with libzstd.


terrelln:
Looks good to me!

Ideally we would thread through the display level to every subroutine that needs it, and I would definitely accept a PR that does it, but changing the global has the same effect, since it is local to `cover.c`. Since it is working as is, I'll defer fixing it until the next time I change cover.

PR #816, opened 2017-08-24
[dictBuilder] Set default compression level to 3

terrelln:
I'm not sure why it was 6, but shouldn't it match the default compression level in the library/CLI?

### DIFF
--- a/lib/dictBuilder/zdict.c
+++ b/lib/dictBuilder/zdict.c
@@ -60,7 +60,7 @@
 
 #define NOISELENGTH 32
 
-static const int g_compressionLevel_default = 6;
+static const int g_compressionLevel_default = 3;
 static const U32 g_selectivity_default = 9;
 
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Any test/benchmark showing the impact of modifying this setting ?

terrelln:
I'll run some tests and post the results here.

terrelln:
Since `ZDICT_trainFromBuffer()` doesn't take a compression level, maybe we should leave the level it selects alone. However, I think the other functions should default to the same level as zstd does.

It doesn't effect the command line, only the library. I ran on the github data set.

```
> ./zstd -6 -T16 --train -r github && ./zstd -3 -D dictionary -r github -c | wc -c
767339
> ./zstd -3 -T16 --train -r github && ./zstd -3 -D dictionary -r github -c | wc -c
756872
```

PR #817, opened 2017-08-25
[pool] Accept custom allocators

terrelln:
* pthreads functions might still allocate.

### DIFF
--- a/lib/common/error_private.h
+++ b/lib/common/error_private.h
@@ -48,9 +48,7 @@ typedef ZSTD_ErrorCode ERR_enum;
 /*-****************************************
 *  Error codes handling
 ******************************************/
-#ifdef ERROR
-#  undef ERROR   /* reported already defined on VS 2015 (Rich Geldreich) */
-#endif
+#undef ERROR   /* reported already defined on VS 2015 (Rich Geldreich) */
 #define ERROR(name) ZSTD_ERROR(name)
 #define ZSTD_ERROR(name) ((size_t)-PREFIX(name))
 
--- a/lib/common/error_private.h
+++ b/lib/common/error_private.h
@@ -51,7 +51,8 @@ typedef ZSTD_ErrorCode ERR_enum;
 #ifdef ERROR
 #  undef ERROR   /* reported already defined on VS 2015 (Rich Geldreich) */
 #endif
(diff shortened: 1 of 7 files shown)


Cyan4973:
Could be a good idea to retrofit `zstdmt` to use it

terrelln:
Add `extern "C"` to `zstd_internal.h`.

lib/common/error_private.h
```diff
@@ -51,7 +51,8 @@ typedef ZSTD_ErrorCode ERR_enum;
 #ifdef ERROR
 #  undef ERROR   /* reported already defined on VS 2015 (Rich Geldreich) */
```
Cyan4973: very minor : 
it's allowed to `#undef` an identifier without previous definition.
So the `#ifdef` guard is not useful.

lib/common/pool.h
```diff
@@ -16,6 +16,7 @@ extern "C" {
 
 
 #include <stddef.h>   /* size_t */
+#include "zstd_internal.h"   /* ZSTD_customMem */
```
Cyan4973: Are you comfortable with having `pool` depending on `zstd_internal` now ?

lib/common/threading.h
```diff
@@ -37,7 +37,15 @@ extern "C" {
...
 #endif
 
+#ifdef ERROR
+#  undef ERROR   /* reported already defined on VS 2015 (Rich Geldreich) */
+#endif
 #include <windows.h>
+#ifdef ERROR
+#  undef ERROR   /* reported already defined on VS 2015 (Rich Geldreich) */
+#endif
+#define ERROR(name) ZSTD_ERROR(name)
```
Cyan4973: In which case is the `ERROR()` defined here useful ?

lib/common/pool.h
terrelln: Its not a problem for me, if we ever need to extract it, we just need to delete the advanced versions.

lib/common/threading.h
terrelln: Yeah, if `zstd_internal.h` is included after, then it isn't useful. But if it is included before, we've just undefined `ERROR`, so we need to redefine it.

PR #818, opened 2017-08-26
Sort input file list

bmwiedemann:
in order to make builds reproducible
in spite of indeterministic filesystem readdir order.
See https://reproducible-builds.org/ for why this is good.


bmwiedemann:
signed CLA

Cyan4973:
Yes, that's a good point @bmwiedemann . Thanks for the patch !

PR #820, opened 2017-08-29
Fix build with GCC 3.x.x (3.4.6 tested) on 32-bit x86

NWilson:
It seems that old versions of GCC fail with an Internal Compiler Error when they try to generate the code for __builtin_ctzll() in 32-bit mode. We can simply avoid emitting this intrinsic in 32-bit code - since that branch should be eliminated in any case after inlining and optimising.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1333,7 +1333,7 @@ static unsigned ZSTD_NbCommonBytes (register size_t val)
             unsigned long r = 0;
             _BitScanForward64( &r, (U64)val );
             return (unsigned)(r>>3);
-#       elif defined(__GNUC__) && (__GNUC__ >= 3)
+#       elif defined(__GNUC__) && (__GNUC__ >= 3) && (__LONG_MAX__ != 2147483647L)
             return (__builtin_ctzll((U64)val) >> 3);
 #       else
             static const int DeBruijnBytePos[64] = { 0, 0, 0, 0, 0, 1, 1, 2,
@@ -1367,7 +1367,7 @@ static unsigned ZSTD_NbCommonBytes (register size_t val)
             unsigned long r = 0;
             _BitScanReverse64( &r, val );
             return (unsigned)(r>>3);
-#       elif defined(__GNUC__) && (__GNUC__ >= 3)
+#       elif defined(__GNUC__) && (__GNUC__ >= 3) && (__LONG_MAX__ != 2147483647L)
             return (__builtin_clzll(val) >> 3);
 #       else
             unsigned r;


Cyan4973:
Another solution could be to require `__GNUC__ >= 4` for these 2 intrinsics.
Older versions would use the DeBruijn path.

NWilson:
That's true - but for all I know, there's someone out there with GCC 3 on a 64-bit machine, who'd appreciate the ctz intrinsic. I don't know how critical these are for perf, and I didn't want to risk a perf regression for someone just for the sake of build fix.

Cyan4973:
Optimization for a specific target is a narrower issue than correct build.
We can add some more logic later on to make it work, based on some real use case, which is  easier than trying to guess them beforehand. In the meantime, it's just an optimization issue, at least the library still builds and runs correctly.

`__GNUC__` is always defined when compiling with `gcc`, and even by some gcc compatible compilers (like `clang`). It's a very reliable identifier.
`__LONG_MAX__` on the other hand is more problematic : it's generally defined within `limits.h`, which may or may not be present depending on compiler, version, etc.. This is a more risky dependency.

zeux:
MinGW *does* define `__GNUC__`

NWilson:
There may be some confusion between `__LONG_MAX__` and `LONG_MAX`. The latter is defined in `limits.h` (which isn't included in zstd_compress.c), and comes from the C standard library, which will typically be bundled along with the compiler and provide the standard C definitions based on knowledge of the compile target. However, `__LONG_MAX__` is a compiler-specific thing (like `__GNUC__`), one of GCC's built-in predefined macros that's not dependent on any headers being included. I used it because its format is known/guaranteed (given the existing `__GNUC__` check), whereas `LONG_MAX` could be a string with casts etc. There's not much to choose between them though.

Thanks for helping with this!

Cyan4973:
Good point @NWilson, I wasn't aware of this.

Relevant documentation : 
https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

One place where I would be cautious about relying on `__LONG_MAX__` value is `x32` compilation mode.
It's a strange mode, where registers are still 64-bits wide, but pointers and object sizes `size_t` are 32-bits. I don't know about `long` width in this case.
`lz4` currently works pretty well in `x32` mode, extracting +10% performance from this particular setting.

(_Edit_ : I just checked `x32` mode, `__LONG_MAX__` has a 64-bits value there, so it's fine)

Another one is OpenVMS 64-bits, which has a strange setting too, where registers and pointers are 64-bits, but `size_t` is 32-bits only. Not sure about `long` either. And in contrast with `x32`, I can't test this one directly, since OpenVMS is everything but open, and requires a specific license.

That's 2 I know. There might be other cases too.

NWilson:
Happy to change it to use `LONG_MAX`, which I suppose is a bit more standards-compliant.

If you don't have an OpenVMS test system, then you can't really support it - unless you know it compiles there currently? At my company here we support around 10 OSs/UNIX platforms, which feels like a lot, and even we consider OpenVMS far too small to worry about!

Cyan4973:
I prefer to reduce dependencies to a minimum, they always add troubles.
So, in this respect, `__LONG_MAX__` is better than `LONG_MAX`.

I'm just slightly worried if some specific version of `gcc` do not support it (`Some of these macros may not be defined on particular systems`), or if a non-gcc but compatible compiler defines `__GNUC__` but not `__LONG_MAX__`.

In any case, `__GNUC__` macro is always needed, it's a preliminary test for `__LONG_MAX__`.
So the minimum dependency we could have is to rely only on `__GNUC__`.
It that's enough to solve this issue, then we could use `__GNUC__ >= 4`, and so be it.

If there are known use cases which use gcc v3.x for 64-bits compilation, then indeed `__LONG_MAX__` can be considered, to better optimize binary for these platforms. But I would prefer that such use case are identified first. Otherwise, it's a blind optimisation, adding its own shares of risks.

Cyan4973:
Proposed fix pushed into `dev` by 3306bcb0e623ccd51f9f1049dba8e88c55e1db0d.
Keeping the simpler `__GNUC__` only solution for now.
Keeping a note about `__LONG_MAX__`, in case it could prove useful in the future.

PR #829, opened 2017-09-01
fixed decompression bug reported by @Etsukata (#828)

Cyan4973:


Etsukata:
90 minutes fuzzing(the same as I tried at https://github.com/facebook/zstd/pull/828)  found no crash case on this branch :-)

PR #831, opened 2017-09-02
Split parsers out of zstd_compress.c

terrelln:
Splits each parser into its own file, and moves the `ZSTD_CCtx` to the `zstd_compress.h`. There are no other changes.

I benchmarked each strategy on 2 computers on Silesia with many different compilers. The results are neutral to a 2% improvement, depending on the compiler (and noise).

https://gist.github.com/terrelln/0ffe81b35b883d5ca163555cbb5f4caa
https://gist.github.com/terrelln/c99cb5a8b219fde62078450ebe71651e


terrelln:
I've made a final change marking `zstd_compress.h` as extern "C".

The only conflicts with the long range matcher patch are changes to the `ZSTD_CCtx_s` struct, and the last literal changes to each match finder. Since this patch is just cut and pasting, it will be extremely easy to merge. Either patch could come first.

lib/compress/zstd_compress.h
```diff
@@ -0,0 +1,341 @@
+#ifndef ZSTD_COMPRESS_H
```
Cyan4973: Each file should have a proper header

Cyan4973:
Looks good to me (just a few headers).

It's a welcomed refactoring, it makes the role of each part clearer.
It's great that the compiler can take advantage of it.

PR #833, opened 2017-09-05
Fix parameter handling with ZSTD_copyCCtx

stellamplau:
`ZSTD_copyCCtx` currently is used in the following ways:
- Initializing a `cctx` using a `cdict` in `compressBegin_internal`: a `cdict` currently only contains compression parameters. By using only the cdict's applied parameters, the other parameters are now "lost". Before the parameter restructuring, only the compression parameters were copied over with `copyCCtx`. This fix preserves this behavior. Though future use-cases may want different behaviour?
- As a standalone function in the buffer-less streaming API called in the `ZSTDcs_init` stage. Again, old behavior is preserved and fields that can be changed in/before the ZSTDcs_init stage are copied.


Cyan4973:
Looks good to me !

PR #838, opened 2017-09-11
Add long distance matcher

stellamplau:


tests/playTests.sh
```diff
@@ -553,13 +553,25 @@ roundTripTest -g516K 19   # btopt
...
 if [ -n "$hasMT" ]
 then
     $ECHO "\n**** zstdmt round-trip tests **** "
     roundTripTest -g4M "1 -T0"
     roundTripTest -g8M "3 -T2"
     roundTripTest -g8000K "2 --threads=2"
     fileRoundTripTest -g4M "19 -T2 -B1M"
+
+    $ECHO "\n**** zstdmt long distance matching round-trip tests **** "
+    roundTripTest -g8M "3 --long -T2"
```
Cyan4973: Aren't "long mode" and "multithreading" currently incompatible ?
What's the expected result of this test ?

lib/compress/zstd_ldm.c
```diff
@@ -0,0 +1,703 @@
...
+                              ldmState, rollingHash, lastHashed,
+                              ip + mLength, base, hBits,
+                              ldmParams);
+            lastHashed = ip + mLength - 1;
+        }
+        ip += mLength;
+        anchor = ip;
+
+        /* check immediate repcode */
+        while (ip < ilimit) {
```
Cyan4973: Just a question : what's the benefit of checking repcode directly within the long match finder ?

Cyan4973:
Code quality is really great.
I only have a couple of questions to go.

lib/compress/zstd_ldm.c
stellamplau: It seemed to improve the compression ratio very slightly without really affecting speed

tests/playTests.sh
stellamplau: I might be missing something, but is there a reason they should be incompatible in terms of correctly compressing/decompressing data? The test should succeed

Cyan4973: My understanding is that `ldm` calls `blockCompressor` directly within its own thread.
So it's effectively not multi-threaded.

It can nonetheless work (aka. compress correctly).
It's just that, given 2 orders, `--long -T2`, it's unclear that only one of them will be effectively applied, and which one will be forgotten.

I assume `ldm` currently has priority : https://github.com/facebook/zstd/pull/838/files#diff-62f99f3a22fe685ec153abb360cca5d5R1535
The behaviour should just be documented clearly (or forbidden).

stellamplau: Everything goes through ZSTD_compress_generic(), right? That's where multithreading/not is decided. If nbthreads > 1, then we take the multithreading path which, for each of its jobs, calls ZSTD_compressContinue(). The ldm parameters are passed to the jobs. This is where the block compressors are called from ZSTD_compressBlock_internal, so doesn't that imply it's multithreaded with ldm?

I'm less familiar with the multithreading code though...

terrelln: AFAIK long range matching and multithreading are "compatible", it will just consume a lot of memory. Something like 4 * 128 MB * nb threads just for the round buffers.

Cyan4973: Right, so it works, a bit like `-22 --ultra` does today, with a fairly large default job size of 512 MB.
Inter-jobs reloading is also by default 32 MB, which is both not large for ldm, but too much for the regular match finder.
I guess we'll have time to reconsider how it works later as an optimisation round.

terrelln:
Awesome! Congrats @stellamplau!

PR #839, opened 2017-09-12
[bench] Use higher resolution timer on POSIX

terrelln:
The timer used was only accurate up to 0.01 seconds. This timer is accurate up to 1 ns.
It is a monotonic timer that measures the real time difference, not on CPU time.

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -118,35 +118,70 @@ static int g_utilDisplayLevel;
 *  Time functions
 ******************************************/
 #if defined(_WIN32)   /* Windows */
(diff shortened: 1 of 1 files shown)


programs/util.h
```diff
@@ -5,7 +5,6 @@
  * This source code is licensed under both the BSD-style license (found in the
  * LICENSE file in the root directory of this source tree) and the GPLv2 (found
  * in the COPYING file in the root directory of this source tree).
- * You may select, at your option, one of the above-listed licenses.
```
Cyan4973: This line should remain

PR #840, opened 2017-09-13
Make decodecorpus generate raw compressed blocks

stellamplau:
Adds a flag to generate raw zstd compressed blocks (with no block/frame header)


Cyan4973:
Looks good to me

PR #841, opened 2017-09-13
modified util::time API (T19505791)

Cyan4973:
for easier invocation.
- no longer expose frequency timer :
it's either useless, or stored internally in a static variable (init is only necessary once).
- UTIL_getTime() provides result by function return.

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -118,38 +118,64 @@ static int g_utilDisplayLevel;
 *  Time functions
 ******************************************/
 #if defined(_WIN32)   /* Windows */
-    typedef LARGE_INTEGER UTIL_freq_t;
     typedef LARGE_INTEGER UTIL_time_t;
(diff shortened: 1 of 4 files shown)


programs/util.h
```diff
@@ -118,38 +118,68 @@ static int g_utilDisplayLevel;
...
     typedef struct timespec UTIL_freq_t;
     typedef struct timespec UTIL_time_t;
-    UTIL_STATIC void UTIL_initTimer(UTIL_freq_t* res)
+    UTIL_STATIC UTIL_time_t UTIL_getTime(void)
     {
-        if (clock_getres(CLOCK_MONOTONIC, res))
-            UTIL_DISPLAYLEVEL(1, "ERROR: Failed to init clock\n");
+        UTIL_time_t time;
+        if (clock_gettime(CLOCK_MONOTONIC, &time))
+            UTIL_DISPLAYLEVEL(1, "ERROR: Failed to get time\n");   /* we could also exit() */
```
terrelln: Either way is good for me.

Cyan4973: I'll keep the "display" behaviour for the time being because it's less disruptive.

The comment is left in the code just in case such error ever happen (which is doubtful),
in which case, we may start a discussion on what could be the better behaviour.

PR #842, opened 2017-09-13
Add flag to limit max decompressed size in decodeCorpus

stellamplau:


tests/decodecorpus.c
```diff
@@ -1778,10 +1786,13 @@ static void advancedUsage(const char* programName)
...
-    DISPLAY( "Advanced arguments :\n");
-    DISPLAY( " --content-size    : always include the content size in the frame header\n");
-    DISPLAY( " --use-dict=#      : include a dictionary used to decompress the corpus\n");
-    DISPLAY( " --gen-blocks      : generate raw compressed blocks without block/frame headers\n");
+    DISPLAY( "Advanced arguments        :\n");
+    DISPLAY( " --content-size           : always include the content size in the frame header\n");
+    DISPLAY( " --use-dict=#             : include a dictionary used to decompress the corpus\n");
+    DISPLAY( " --gen-blocks             : generate raw compressed blocks without block/frame headers\n");
+    DISPLAY( " --max-block-size-log=#   : max block size log, must be in range [2, 17]\n");
+    DISPLAY( " --max-content-size-log=# : max content size log, must be <= 20\n");
```
Cyan4973: Is there a reason why it should be limited <= 20 in all circumstances ?

stellamplau: Only that buffers are currently statically allocated with a max size of 1<<20. Do we want to increase it?

Cyan4973: Indeed, if buffers are statically allocated, they need a limit, and a reasonable one.

I did not paid attention to this implementation detail before.
My (incorrect) assumption was that buffers were allocated with `malloc()`.

Which reason makes it desirable to statically allocate these buffers ? Convenience ?

stellamplau: My guess is it was just convenience and there wasn't an immediate/anticipated need to fuzz test data larger than 1 MB here.

Cyan4973:
👍

PR #846, opened 2017-09-14
Split dict

Cyan4973:
added ability to split input files for dictionary training using command `-B#`

This is a similar behaviour as benchmark module,
which can also also split input into independent blocks using `-B#`.

### DIFF
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -14,7 +14,7 @@
 *  Compiler Warnings
 ****************************************/
 #ifdef _MSC_VER
-#  pragma warning(disable : 4127)                /* disable: C4127: conditional expression is constant */
+#  pragma warning(disable : 4127)    /* disable: C4127: conditional expression is constant */
 #endif
 
 
@@ -44,7 +44,7 @@
 #define SAMPLESIZE_MAX (128 KB)
 #define MEMMULT 11    /* rough estimation : memory cost to analyze 1 byte of sample */
 #define COVER_MEMMULT 9    /* rough estimation : memory cost to analyze 1 byte of sample */
-static const size_t maxMemory = (sizeof(size_t) == 4) ? (2 GB - 64 MB) : ((size_t)(512 MB) << sizeof(size_t));
+static const size_t g_maxMemory = (sizeof(size_t) == 4) ? (2 GB - 64 MB) : ((size_t)(512 MB) << sizeof(size_t));
 
 #define NOISELENGTH 32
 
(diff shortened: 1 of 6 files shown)


programs/dibio.c
```diff
@@ -98,32 +97,46 @@ const char* DiB_getErrorName(size_t errorCode) { return ERR_getErrorName(errorCo
...
+        FILE* const f = fopen(fileName, "rb");
+        if (f==NULL) EXM_THROW(10, "zstd: dictBuilder: %s %s ", fileName, strerror(errno));
+        DISPLAYUPDATE(2, "Loading %s...       \r", fileName);
+        for (cnb=0; cnb<nbChunks; cnb++) {
+            size_t const toLoad = (size_t)MIN(maxChunkSize, remainingToLoad);
+            if (toLoad > *bufferSizePtr-pos) break;
+            {   size_t const readSize = fread(buff+pos, 1, toLoad, f);
+                if (readSize != toLoad) EXM_THROW(11, "Pb reading %s", fileName);
+                pos += readSize;
+                chunkSizes[nbLoadedChunks++] = toLoad;
```
terrelln: Do we need to check that `chunkSizes` is large enough?

Edit: I see that the call-site sizes it, can we add a comment saying that you need to call `DiB_fileStats()` to get the size.

programs/dibio.c
```diff
@@ -230,64 +256,65 @@ size_t ZDICT_trainFromBuffer_unsafe_legacy(void* dictBuffer, size_t dictBufferCa
...
+        DISPLAYLEVEL(2, "!  As a consequence, only the first %u bytes of each sample are loaded \n", SAMPLESIZE_MAX);
     }
-    if ((nbFiles < 5) || (totalSizeToLoad < 9 * (unsigned long long)maxDictSize)) {
+    if (fs.nbChunks < 5) {
         DISPLAYLEVEL(2, "!  Warning : nb of samples too low for proper processing ! \n");
         DISPLAYLEVEL(2, "!  Please provide _one file per sample_. \n");
-        DISPLAYLEVEL(2, "!  Do not concatenate samples together into a single file, \n");
-        DISPLAYLEVEL(2, "!  as dictBuilder will be unable to find the beginning of each sample, \n");
-        DISPLAYLEVEL(2, "!  resulting in poor dictionary quality. \n");
+        EXM_THROW(14, "nb of samples too low");   /* we now clearly forbid this case */
```
terrelln: We'll probably get some issues because of this, but then hopefully we won't have to deal with it again.

Cyan4973: I now prefer to forbid it clearly, rather than having some user attempting dictionary generation with a single large archive, generate a s*** dictionary as a consequence, then jump to conclusion that dictionary compression doesn't help.
This just happened once again this week, proving this is a long-enduring pattern, with no end in sight (and I'm not counting the larger number of silent failures). Forbidding it should put an end to such scenario.

What I'm wondering here is
if it would be useful to advertise the `-B#` splitting feature in the warning message.

programs/dibio.c
Cyan4973: You are right.
`chunkSizes` is properly sized thanks to `DiB_fileStats()`,
but since it's a separate action,
it's nonetheless good practice to ensure that table size is sufficient.

I'll add both the comment and the test.

programs/dibio.c
terrelln: I think it couldn't hurt. That way the people who need the `-B#` command know it exists, and can tell us if they need more than splitting every `#` bytes.

PR #848, opened 2017-09-16
[block] Don't use fParams in ZSTD_decompressBlock()

terrelln:
* `fParams` are initialized.
* Caught by the `block_decompress` fuzzer. It now doesn't crash with MSAN:
```
CC=clang CXX=clang++ python fuzz.py build all --enable-msan --enable-coverage --lib-fuzzing-engine ~/build/libfuzzer/msan/libFuzzer.a
./fuzz.py libfuzzer block_decompress
```

This command requires building `libFuzzer.a` with MSAN, which is a pain. Once we have the coverage corpora set up, it will be easy to run regression tests like this without any external dependencies/C++, so MSAN testing will be easy.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -860,7 +860,10 @@ size_t ZSTD_execSequenceLast7(BYTE* op,
 }
 
 
-static seq_t ZSTD_decodeSequence(seqState_t* seqState)
+typedef enum { ZSTD_lo_isRegularOffset, ZSTD_lo_isLongOffset=1 } ZSTD_longOffset_e;
+
+
+static seq_t ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets)
 {
     seq_t seq;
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Another possibility to determine long mode is to check if `offsets > threshold` actually exists, based on the `offsetCodes` distribution : the largest offset code is extracted during the header decoding process.

Problem is, it happens at another place in the code, and the decision to select `ZSTD_decompressSequencesLong()` is supposed to be done before this point. 
So it would require to change the logic to decode fse tables first, and then decide which decoder variant to select.

Code modification is indeed more complex.

terrelln:
I think that is what we'll need to do if offsets become large enough to require special handling in 64-bit mode. Since it only effect of getting it wrong is slowing down 32-bit mode, the change probably isn't worth the effort. I'll create a task for the change.

Cyan4973:
Code looks good to me.
Any benchmark to check there is no performance regression ?

terrelln:
```
gcc-5
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 444.1 MB/s ,1103.5 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 442.1 MB/s ,1101.5 MB/s
gcc-6
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 433.1 MB/s ,1105.9 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 434.3 MB/s ,1106.1 MB/s
gcc-7
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 440.9 MB/s ,1109.6 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 440.7 MB/s ,1115.3 MB/s
clang-3.8
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 436.6 MB/s , 879.8 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 436.9 MB/s , 887.3 MB/s
clang-3.9
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 427.3 MB/s ,1026.9 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 426.9 MB/s ,1017.1 MB/s
```

These numbers (before/after on Silesia) are pretty consistent. It looks like everything but gcc-7 and clang are about neutral. gcc-7 and clang-3.8 consistently speed up, and clang-3.9 consistently slows down. The change is < 1% either way.

PR #849, opened 2017-09-16
[bitstream] Allow adding 31 bits at a time

terrelln:
One change needed to allow larger window sizes. I picked 31 because that allows offsets up to 2^32-4, which is large enough so this won't be the limiting factor again (unless significant changes are made to the rest of the code).

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -194,11 +194,14 @@ MEM_STATIC unsigned BIT_highbit32 (register U32 val)
 }
 
 /*=====    Local Constants   =====*/
-static const unsigned BIT_mask[] = { 0, 1, 3, 7, 0xF, 0x1F, 0x3F, 0x7F,
-                                    0xFF, 0x1FF, 0x3FF, 0x7FF, 0xFFF, 0x1FFF, 0x3FFF, 0x7FFF,
-                                    0xFFFF, 0x1FFFF, 0x3FFFF, 0x7FFFF, 0xFFFFF, 0x1FFFFF, 0x3FFFFF, 0x7FFFFF,
-                                    0xFFFFFF, 0x1FFFFFF, 0x3FFFFFF };   /* up to 26 bits */
-
+static const unsigned BIT_mask[] = {
+    0,          1,         3,         7,         0xF,       0x1F,
+    0x3F,       0x7F,      0xFF,      0x1FF,     0x3FF,     0x7FF,
+    0xFFF,      0x1FFF,    0x3FFF,    0x7FFF,    0xFFFF,    0x1FFFF,
+    0x3FFFF,    0x7FFFF,   0xFFFFF,   0x1FFFFF,  0x3FFFFF,  0x7FFFFF,
+    0xFFFFFF,   0x1FFFFFF, 0x3FFFFFF, 0x7FFFFFF, 0xFFFFFFF, 0x1FFFFFFF,
+    0x3FFFFFFF, 0x7FFFFFFF}; /* up to 31 bits */
+#define BIT_MASK_SIZE (sizeof(BIT_mask) / sizeof(BIT_mask[0]))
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good !
Note that in 32-bits mode, limit should be 24-bits for a single `add`.
I guess currently the limit is entirely driven by documentation ?

terrelln:
Good point, I'll add an assert for that case.

lib/common/bitstream.h
```diff
@@ -220,11 +222,14 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
 }
 
 /*! BIT_addBits() :
- *  can add up to 26 bits into `bitC`.
+ *  can add up to 31 bits into `bitC`.
  *  Note : does not check for register overflow ! */
 MEM_STATIC void BIT_addBits(BIT_CStream_t* bitC,
                             size_t value, unsigned nbBits)
 {
+    MEM_STATIC_ASSERT(sizeof(BIT_mask) / sizeof(BIT_mask[0]) == 32);
```
Cyan4973: Since the constant `sizeof(BIT_mask) / sizeof(BIT_mask[0]` comes back often, you could hide it in a `#define` ?
It would also help to understand what it means.

lib/common/bitstream.h
```diff
@@ -220,11 +223,14 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
...
 
 /*! BIT_addBits() :
- *  can add up to 26 bits into `bitC`.
+ *  can add up to 31 bits into `bitC`.
  *  Note : does not check for register overflow ! */
 MEM_STATIC void BIT_addBits(BIT_CStream_t* bitC,
                             size_t value, unsigned nbBits)
 {
+    MEM_STATIC_ASSERT(BIT_MAX_NBBITS == 32);
+    assert(nbBits <= BIT_MAX_NBBITS);
```
Cyan4973: should it be `<` in this `assert()` ?
It's supposed to add up to 31 bits, but `BIT_MAX_NBBITS == 32`

lib/common/bitstream.h
```diff
@@ -220,11 +223,14 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
...
 /*! BIT_addBits() :
- *  can add up to 26 bits into `bitC`.
+ *  can add up to 31 bits into `bitC`.
  *  Note : does not check for register overflow ! */
 MEM_STATIC void BIT_addBits(BIT_CStream_t* bitC,
                             size_t value, unsigned nbBits)
 {
+    MEM_STATIC_ASSERT(BIT_MASK_SIZE == 32);
+    assert(nbBits < BIT_MASK_SIZE);
+    assert(nbBits + bitC->bitPos <= sizeof(bitC->bitContainer) * 8);
```
Cyan4973: I believe it should be `<` too.
After this function, `bitC->bitPos += nbBits;`,
so if it ends up being `== 64`, next shift operation will fail.

lib/common/bitstream.h
```diff
@@ -235,6 +241,7 @@ MEM_STATIC void BIT_addBitsFast(BIT_CStream_t* bitC,
                                 size_t value, unsigned nbBits)
 {
     assert((value>>nbBits) == 0);
+    assert(nbBits + bitC->bitPos <= sizeof(bitC->bitContainer) * 8);
```
Cyan4973: same comment here

Cyan4973:
Any intention to add an `assert()` for 32-bits mode ?

terrelln:
`assert(nbBits + bitC->bitPos < sizeof(bitC->bitContainer) * 8)` covers both modes. I don't think we can assert anything stronger, since in special cases it is okay to write more than 25/57 bits (you know the bitstream is empty).

PR #850, opened 2017-09-16
[fse] Fix FSE_optimalTableLog() for srcSize==1

terrelln:
* Normally when nbSeq == 1, we use the predefined tables, but if the ofcode > 28 we can't, so we run into this issue.

### DIFF
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -461,6 +461,7 @@ static unsigned FSE_minTableLog(size_t srcSize, unsigned maxSymbolValue)
     U32 minBitsSrc = BIT_highbit32((U32)(srcSize - 1)) + 1;
     U32 minBitsSymbols = BIT_highbit32(maxSymbolValue) + 2;
     U32 minBits = minBitsSrc < minBitsSymbols ? minBitsSrc : minBitsSymbols;
+    assert(srcSize > 1); /* Not supported, RLE should be used instead */
     return minBits;
 }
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
If `nbSeq==1`, it means the statistics are 100% concentrated in a single symbol, so it should be encoded using RLE mode : https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#sequences_section_header .

RLE mode doesn't need `FSE_minTableLog()`, so it does not trigger that issue.
More importantly, it's significantly cheaper than encoding a FSE header for a single symbol.

I think the current limitation lies in this line : 
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1260
which reserves RLE for `nbSeq>=2`, because it bets that if `nbSeq==1`, then default distribution is better (because RLE costs 1 byte, while default distribution costs 5-6 bits per symbol).
Well, that could be something to change, either conditionally or unconditionally, as preferred.

Another alternative could be to re-use previous FSE table (_if_ it supports the requested `ofcode`). But it's more difficult to implement. Could be considered at a later stage, when several different symbols must be encoded (in which case, RLE is not a good fit).

terrelln:
I will use to RLE if the basic tables can't be used. I think we should either fix this, or return an error code if `srcSize == 1`, since `FSE_optimalTableLog()` is an exposed function for FSE. I'd prefer the fix, since it isn't performance sensitive.

terrelln:
Ok, looking into the code more when `nbSeq == 1`, normalization and compression both return `0`, to signal RLE mode. I think we could just `assert(nbSeq > 1)` in `FSE_optimalTableLog()` and add it to the documentation.

Cyan4973:
Agreed

PR #855, opened 2017-09-19
[libzstd] Increase MaxOff

terrelln:
* Allow ofcode up to 31
* Initial benchmarks look good, I will paste the results here when they're complete

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -123,7 +123,8 @@ typedef enum { set_basic, set_rle, set_compressed, set_repeat } symbolEncodingTy
 #define MaxLit ((1<<Litbits) - 1)
 #define MaxML  52
 #define MaxLL  35
-#define MaxOff 28
+#define DefaultMaxOff 28
+#define MaxOff 31
 #define MaxSeq MAX(MaxLL, MaxML)   /* Assumption : MaxOff < MaxLL,MaxML */
 #define MLFSELog    9
 #define LLFSELog    9
@@ -149,8 +150,8 @@ static const S16 ML_defaultNorm[MaxML+1] = { 1, 4, 3, 2, 2, 2, 2, 2, 2, 1, 1, 1,
 #define ML_DEFAULTNORMLOG 6  /* for static allocation */
 static const U32 ML_defaultNormLog = ML_DEFAULTNORMLOG;
 
-static const S16 OF_defaultNorm[MaxOff+1] = { 1, 1, 1, 1, 1, 1, 2, 2, 2, 1, 1, 1, 1, 1, 1, 1,
-                                              1, 1, 1, 1, 1, 1, 1, 1,-1,-1,-1,-1,-1 };
+static const S16 OF_defaultNorm[DefaultMaxOff+1] = { 1, 1, 1, 1, 1, 1, 2, 2, 2, 1, 1, 1, 1, 1, 1, 1,
+                                                     1, 1, 1, 1, 1, 1, 1, 1,-1,-1,-1,-1,-1 };
 #define OF_DEFAULTNORMLOG 5  /* for static allocation */
 static const U32 OF_defaultNormLog = OF_DEFAULTNORMLOG;
 
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1250,20 +1250,30 @@ void ZSTD_seqToCodes(const seqStore_t* seqStorePtr)
         mlCodeTable[seqStorePtr->longLengthPos] = MaxML;
 }
 
-MEM_STATIC symbolEncodingType_e ZSTD_selectEncodingType(FSE_repeat* repeatMode,
-        size_t const mostFrequent, size_t nbSeq, U32 defaultNormLog)
+typedef enum {
+    ZSTD_defaultDisallowed = 0,
+    ZSTD_defaultAllowed = 1
+} ZSTD_defaultPolicy_e;
+
+MEM_STATIC symbolEncodingType_e ZSTD_selectEncodingType(
+        FSE_repeat* repeatMode, size_t const mostFrequent, size_t nbSeq,
+        U32 defaultNormLog, ZSTD_defaultPolicy_e const isDefaultAllowed)
 {
 #define MIN_SEQ_FOR_DYNAMIC_FSE   64
 #define MAX_SEQ_FOR_STATIC_FSE  1000
-
-    if ((mostFrequent == nbSeq) && (nbSeq > 2)) {
+    ZSTD_STATIC_ASSERT(ZSTD_defaultDisallowed == 0 && ZSTD_defaultAllowed != 0);
+    if ((mostFrequent == nbSeq) && (!isDefaultAllowed || nbSeq > 2)) {
+        /* Prefer set_basic over set_rle when there are 2 or less symbols,
+         * since RLE uses 1 byte, but set_basic uses 5-6 bits per symbol.
+         * If basic encoding isn't possible, always choose RLE.
+         */
         *repeatMode = FSE_repeat_check;
         return set_rle;
     }
-    if ((*repeatMode == FSE_repeat_valid) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
+    if (isDefaultAllowed && (*repeatMode == FSE_repeat_valid) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
         return set_repeat;
     }
-    if ((nbSeq < MIN_SEQ_FOR_DYNAMIC_FSE) || (mostFrequent < (nbSeq >> (defaultNormLog-1)))) {
+    if (isDefaultAllowed && ((nbSeq < MIN_SEQ_FOR_DYNAMIC_FSE) || (mostFrequent < (nbSeq >> (defaultNormLog-1))))) {
         *repeatMode = FSE_repeat_valid;
         return set_basic;
     }
@@ -1299,6 +1309,7 @@ MEM_STATIC size_t ZSTD_buildCTable(void* dst, size_t dstCapacity,
             count[codeTable[nbSeq-1]]--;
             nbSeq_1--;
         }
+        assert(nbSeq_1 > 1);
         CHECK_F(FSE_normalizeCount(norm, tableLog, count, nbSeq_1, max));
         {   size_t const NCountSize = FSE_writeNCount(op, oend - op, norm, max, tableLog);   /* overflow protected */
             if (FSE_isError(NCountSize)) return NCountSize;
@@ -1436,7 +1447,7 @@ MEM_STATIC size_t ZSTD_compressSequences_internal(seqStore_t* seqStorePtr,
     /* CTable for Literal Lengths */
     {   U32 max = MaxLL;
         size_t const mostFrequent = FSE_countFast_wksp(count, &max, llCodeTable, nbSeq, entropy->workspace);
-        LLtype = ZSTD_selectEncodingType(&entropy->litlength_repeatMode, mostFrequent, nbSeq, LL_defaultNormLog);
+        LLtype = ZSTD_selectEncodingType(&entropy->litlength_repeatMode, mostFrequent, nbSeq, LL_defaultNormLog, ZSTD_defaultAllowed);
         {   size_t const countSize = ZSTD_buildCTable(op, oend - op, CTable_LitLength, LLFSELog, (symbolEncodingType_e)LLtype,
                     count, max, llCodeTable, nbSeq, LL_defaultNorm, LL_defaultNormLog, MaxLL,
                     entropy->workspace, sizeof(entropy->workspace));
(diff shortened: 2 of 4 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -936,13 +941,16 @@ static seq_t ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e l
 
     seq.matchLength = ML_base[mlCode]
                     + ((mlCode>31) ? BIT_readBitsFast(&seqState->DStream, mlBits) : 0);  /* <=  16 bits */
-    if (MEM_32bits() && (mlBits+llBits>24)) BIT_reloadDStream(&seqState->DStream);
+    if (MEM_32bits() && (mlBits+llBits>(STREAM_ACCUMULATOR_MIN_32-kMaxExtraBits32)))
+        BIT_reloadDStream(&seqState->DStream);
+    if (MEM_64bits() && (totalBits > STREAM_ACCUMULATOR_MIN_64 - (LLFSELog+MLFSELog+OffFSELog)))
+        BIT_reloadDStream(&seqState->DStream);
+    assert(16 + LLFSELog + MLFSELog + OffFSELog < STREAM_ACCUMULATOR_MIN_64);
```
Cyan4973: Is it a `STATIC_ASSERT()` ?

lib/decompress/zstd_decompress.c
```diff
@@ -862,20 +862,24 @@ size_t ZSTD_execSequenceLast7(BYTE* op,
...
     U32 const mlCode = FSE_peekSymbol(&seqState->stateML);
-    U32 const ofCode = FSE_peekSymbol(&seqState->stateOffb);   /* <= maxOff, by table construction */
+    U32 const ofCode = FSE_peekSymbol(&seqState->stateOffb);   /* <= MaxOff, by table construction */
 
     U32 const llBits = LL_bits[llCode];
     U32 const mlBits = ML_bits[mlCode];
     U32 const ofBits = ofCode;
     U32 const totalBits = llBits+mlBits+ofBits;
 
+    U32 const kMaxExtraBits32 =
```
Cyan4973: Is that a `static const` value ?

terrelln: I'd prefer not to make it static in the function, since that might make the compiler emit the code for static initialization. It could moved to static outside the function scope though.

lib/decompress/zstd_decompress.c
```diff
@@ -936,13 +941,16 @@ static seq_t ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e l
 
     seq.matchLength = ML_base[mlCode]
                     + ((mlCode>31) ? BIT_readBitsFast(&seqState->DStream, mlBits) : 0);  /* <=  16 bits */
-    if (MEM_32bits() && (mlBits+llBits>24)) BIT_reloadDStream(&seqState->DStream);
+    if (MEM_32bits() && (mlBits+llBits>(STREAM_ACCUMULATOR_MIN_32-kMaxExtraBits32)))
+        BIT_reloadDStream(&seqState->DStream);
+    if (MEM_64bits() && (totalBits > STREAM_ACCUMULATOR_MIN_64 - (LLFSELog+MLFSELog+OffFSELog)))
```
Cyan4973: I now realise that we have a single value `STREAM_ACCUMULATOR_MIN_64` (and other equivalents) applied to both read and write side.
But it's slightly misleading.

On the write side, it's not possible to write more than 56 bits, otherwise the following shift operation may fail.
No such problem on the read side, which means  it's possible to read up to 57 bits.
Not a large difference, but still.

`STREAM_ACCUMULATOR_MIN_64` is set to the larger one, 57 bits.
It's possible to make up for it by using `<` for write side, and `<=` for read side.
But it risks being confusing.

Other solutions : 
- have 2 different macro/variables, depending on read/write side
- settle for the smaller value (56 bits for above example)

What do you think ?

lib/decompress/zstd_decompress.c
```diff
@@ -862,20 +862,24 @@ size_t ZSTD_execSequenceLast7(BYTE* op,
...
     U32 const llBits = LL_bits[llCode];
     U32 const mlBits = ML_bits[mlCode];
     U32 const ofBits = ofCode;
     U32 const totalBits = llBits+mlBits+ofBits;
 
+    U32 const kMaxExtraBits32 =
+        ZSTD_WINDOWLOG_MAX_32 > STREAM_ACCUMULATOR_MIN_32
+            ? (ZSTD_WINDOWLOG_MAX_32 - 1) - STREAM_ACCUMULATOR_MIN_32
+            : 0;
+
```
Cyan4973: Do we want to assert a condition on `kMaxExtraBits32` ?

lib/decompress/zstd_decompress.c
terrelln: If there isn't a penalty for using the smaller value (with the current value of MaxOff), I think we should use it, and maybe note that it could be one larger for read so we don't forget. Otherwise, I think we should have two different macros, since it is too easy to miss that it is different.

terrelln:
Before/after

```
gcc-5
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 441.4 MB/s ,1103.2 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 439.8 MB/s ,1101.9 MB/s
gcc-6
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 435.3 MB/s ,1107.3 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 436.0 MB/s ,1109.2 MB/s
gcc-7
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 441.7 MB/s ,1110.7 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 441.2 MB/s ,1114.7 MB/s
clang-3.8
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 438.6 MB/s , 876.0 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 437.8 MB/s , 876.3 MB/s
clang-3.9
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 426.9 MB/s ,1017.1 MB/s
 1#silesia.tar       : 211988480 ->  73652051 (2.878), 428.3 MB/s ,1002.1 MB/s
```

I ran against trunk clang on my devserver and there wasn't a significant difference

```
clang-trunk
 1#silesia.tar       : 211988480 ->  73651692 (2.878), 209.8 MB/s , 490.2 MB/s
 1#silesia.tar       : 211988480 ->  73651692 (2.878), 210.6 MB/s , 489.5 MB/s
```

lib/decompress/zstd_decompress.c
Cyan4973: That's fine as is

lib/decompress/zstd_decompress.c
Cyan4973: Agreed.
Let's go for the smaller value.

Let's also check where it's used.
It's quite possible that current tests take care of this difference alternating `<` and `<=` when needed.

Cyan4973:
I guess this PR depends on https://github.com/facebook/zstd/pull/849 ?

terrelln:
I think they should be independent, but you'll need both to work with 1 GB offsets.

lib/compress/zstd_compress.c
```diff
@@ -1251,19 +1251,23 @@ void ZSTD_seqToCodes(const seqStore_t* seqStorePtr)
 }
 
 MEM_STATIC symbolEncodingType_e ZSTD_selectEncodingType(FSE_repeat* repeatMode,
-        size_t const mostFrequent, size_t nbSeq, U32 defaultNormLog)
+        size_t const mostFrequent, size_t nbSeq, U32 defaultNormLog, U32 basic)
```
Cyan4973: Is `basic` a good name ?
I find it a bit difficult to associate what it is about.
Of course, since we have been talking about it, I can follow. But it's easy to lose this background context.

lib/common/zstd_internal.h
```diff
@@ -123,7 +123,8 @@ typedef enum { set_basic, set_rle, set_compressed, set_repeat } symbolEncodingTy
 #define MaxLit ((1<<Litbits) - 1)
 #define MaxML  52
 #define MaxLL  35
-#define MaxOff 28
+#define BasicMaxOff 28
```
Cyan4973: I believe I would prefer `DefaultMaxOff` here,
or anything which helps remember it is the limit for default distribution of Offset Codes.

terrelln:
Ah, I only searched for `MaxOff` in `lib/`, but `decodecorpus.c` also uses it. I'm searching the other directories for uses now.

terrelln:
Fixed invalid `assert(ofBits < MAX_WINDOW_SIZE)`, which is only true for valid data. Change to `assert(ofBits <= MaxOff)` which should be true for all data. Also caught by my fuzzer tests, which now pass.

lib/compress/zstd_compress.c
```diff
@@ -1250,20 +1250,30 @@ void ZSTD_seqToCodes(const seqStore_t* seqStorePtr)
...
+    if ((mostFrequent == nbSeq) && (!defaultAllowed || nbSeq > 2)) {
+        /* Prefer set_basic over set_rle when there are 2 or less symbols,
+         * since RLE uses 1 byte, but set_basic uses 5-6 bits per symbol.
+         * If basic encoding isn't possible, always choose RLE.
+         */
         *repeatMode = FSE_repeat_check;
         return set_rle;
     }
-    if ((*repeatMode == FSE_repeat_valid) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
+    if (defaultAllowed && (*repeatMode == FSE_repeat_valid) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
```
Cyan4973: Why is it necessary to check `defaultAllowed` for `set_repeat` mode ?
Isn't table validity already checked in `*repeatMode` ?

lib/compress/zstd_compress.c
```diff
@@ -1250,20 +1250,30 @@ void ZSTD_seqToCodes(const seqStore_t* seqStorePtr)
...
+
+MEM_STATIC symbolEncodingType_e ZSTD_selectEncodingType(
+        FSE_repeat* repeatMode, size_t const mostFrequent, size_t nbSeq,
+        U32 defaultNormLog, ZSTD_defaultPolicy_e const defaultAllowed)
 {
 #define MIN_SEQ_FOR_DYNAMIC_FSE   64
 #define MAX_SEQ_FOR_STATIC_FSE  1000
 
-    if ((mostFrequent == nbSeq) && (nbSeq > 2)) {
+    if ((mostFrequent == nbSeq) && (!defaultAllowed || nbSeq > 2)) {
```
Cyan4973: This test result is tied to `ZSTD_defaultAllowed==1`. 
Is it worthwhile to make it a static assert ?

terrelln: I'm not sure I understand what you mean. `defaultAllowed` is dynamic for offsets, since we check if `max <= DefaultMaxOff` value.

lib/compress/zstd_compress.c
terrelln: `FSE_repeat_valid` is only for dictionaries, or the default distribution, which we assume is valid. We don't yet use repeated FSE tables from a previous block, which would be `FSE_repeat_check`. When I add that capability, I will handle checking for large offsets.

lib/compress/zstd_compress.c
Cyan4973: I meant `ZSTD_defaultAllowed` as an `enum` value,
not `defaultAllowed` as a variable name.
Both are indeed spelled pretty close.

You are testing `!defaultAllowed`, which implies that `ZSTD_defaultAllowed==1`.
Since implicit assumption could be broken in the future, or just not clearly understood by a future reviewer, you could go being explicit with `defaultAllowed==ZSTD_defaultAllowed` instead.
(btw, I think I would prefer something like `isDefaultAllowed` as the variable name.)

I understand it's technically the same.
It's just about being more readable and reducing side effects in case of future changes.

terrelln:
Added static assert and changed name to `isDefaultAllowed`.

Cyan4973:
Some remaining issues on 32-bits platforms : 
https://travis-ci.org/facebook/zstd/jobs/279634639#L767

terrelln:
Good point, I think the assert is too strict. In the case where the offset is too large `ofBits >= MAX_WINDOW_LOG_32`, which is when `extraBits` might fail the assert, we don't care if we read too many bytes. We'll get garbage data, but won't cause any undefined behavior.

We're left with the single assert `assert(ofBits <= MaxOff);`, which must be correct, because otherwise we will read beyond the end of `OF_base`.

terrelln:
Looks like there is another failure in 32-bit mode on ARM.

terrelln:
I haven't been able to reproduce the failure on my Mac in 32-bit mode with or without ASAN/UBSAN. I also ran in a Ubuntu VM with `qemu-arm-static` and couldn't reproduce it. I'll continue running the tests in case.

Is `qemu-arm-static` the test that occasionally segfaults?

Cyan4973:
Yes, indeed `qemu-arm-static` is known to sometimes `segfault` in multi-threading mode.

One identified culprit has been the lack of memory : this an emulator running inside a VM, with no clear guarantee on the amount of memory that can be allocated.
For this reason was created the `--no-big-tests` mode.
Since then, the nb of segfaults has been seriously reduced, but not completely eliminated.

It might be possible to reproduce the issue in a memory-constrained VM (typically limited to 1 GB RAM). If the issue can be reproduced, that would help to design a better `--no-big-tests` mode that no longer fails on travisCI.

In the meantime, I'll restart the test, and should it work fine as expected, consider this last error as a test fluke.

terrelln:
I've also run the test 10+ times on my Mac and no failures so far.

Cyan4973:
I'm not surprised, our laptop are pretty powerful and have plenty of RAM compared to Travis VM.

PR #859, opened 2017-09-22
Prepare for ZSTD_WINDOWLOG_MAX == 31

terrelln:
* Fix `cycleMask` to be computed using `chainLog` instead of `hashLog`. This will improve the compression ratio a bit for levels where `chainLog > hashLog`, the chain table or binary tree is being used, and the file triggers overflow protection.
* Fix some shifts to make sure they don't overflow a signed `1`.
* Enforce upper bounds on `ZSTD_WINDOWLOG_MAX_32`, `ZSTD_WINDOWLOG_MAX`, and `ZSTD_CHAINLOG_MAX` to ensure that overflow protection works.

See the comments to see why `ZSTD_CHAINLOG_MAX <= 30` is necessary. I think I've enumerated all the possible error conditions, but it is entirely possible I missed one.

I'm not sure if `base` or `dictBase` could overflow? I guess we already have pointer underflows during delta calculations, so if it already works, it should continue to work.


lib/compress/zstd_compress.c
```diff
@@ -1599,20 +1599,41 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
             return ERROR(dstSize_tooSmall);   /* not enough space to store compressed block */
         if (remaining < blockSize) blockSize = remaining;
 
-        /* preemptive overflow correction */
+        /* preemptive overflow correction:
+         * 1. correction is large enough:
+         *    lowLimit > (3<<29) ==> current > 3<<29 + 1<<windowLog - blockSize
+         *    newCurrent < 1<<chainLog + 1<<windowLog
```
Cyan4973: Slightly more precise definition : 
`windowSize <= newCurrent < windowSize + (1<<cycleLog)`

lib/compress/zstd_compress.c
```diff
@@ -1599,20 +1599,41 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
             return ERROR(dstSize_tooSmall);   /* not enough space to store compressed block */
         if (remaining < blockSize) blockSize = remaining;
 
-        /* preemptive overflow correction */
+        /* preemptive overflow correction:
+         * 1. correction is large enough:
+         *    lowLimit > (3<<29) ==> current > 3<<29 + 1<<windowLog - blockSize
```
Cyan4973: It shows that overflow correction is driven by the value of `lowLimit`.
This is true because currently `lowLimit` is constantly updated to ensure that no offset beyond `windowSize` can be valid, effectively acting as a "free" offset checker.

In the future, if we implement a "real" offset checker, notably to better serve small window sizes, it will no longer be necessary to update `lowLimit` to do the "offset validation". 
But it will still be necessary to update it to make sure we trigger overflow correction.

Nothing to be worried about at this stage, just something to remember in the future.

Cyan4973:
It looks good.
Could we have a test that actually triggers this limit ?

We already have tests with more than 4 GB of input, triggering overflow protection.
What we miss is a test which does that, while also using max possible window size (1 GB in 32 bits, 2 GB in 64 bits).

Another question is about decodability : 
I understand 32-bits cannot encode a window size > 1 GB,
I'm not sure if it means it cannot _decode_ a window size == 2 GB.

terrelln:
* I'll look into adding that test when I change the maximum window size.
* I'll look into the decoder limit. It would be nice to support 2 GB is possible.
* I am not going to change the maximum window size until I clean up the window size too large logic / interface.

PR #860, opened 2017-09-22
Add Default LZ4 Support When Available

felixhandte:


programs/fileio.c
```diff
@@ -672,7 +672,7 @@ static unsigned long long FIO_compressLz4Frame(cRess_t* ress,
     prefs.compressionLevel = compressionLevel;
     prefs.frameInfo.blockMode = LZ4F_blockIndependent; /* stick to defaults for lz4 cli */
     prefs.frameInfo.blockSizeID = LZ4F_max4MB;
-    prefs.frameInfo.contentChecksumFlag = (contentChecksum_t)g_checksumFlag;
+    prefs.frameInfo.contentChecksumFlag = (LZ4F_contentChecksum_t)g_checksumFlag;
```
terrelln: Is `LZ4F_contentChecksum_t` available in all versions of lz4 that have frame support? I seem to remember using `contentChecksum_t` because that isn't the case.

felixhandte: The `LZ4F_` prefix dates from [this commit](https://github.com/lz4/lz4/commit/081bcca33bba4f6ac3e913b3cf2bc2ba58b6b697) in mid-2015. So no, I guess not. When I built, though, `LZ4F_ENABLE_OBSOLETE_ENUMS` wasn't set, so the build failed. The alternative is to set `LZ4F_ENABLE_OBSOLETE_ENUMS` before I include `lz4frame.h`, I guess.

terrelln: Yeah, we should define that. It looks like it was changed in the last release https://github.com/lz4/lz4/commit/4531637ecdc4b12154c66cc69cfaade185039e4c.

Cyan4973:
The first version with `lz4frame.h` support is [r123](https://github.com/lz4/lz4/releases/tag/r123) .

For information, the library version of `r123` is [v1.3.1](https://github.com/lz4/lz4/blob/r123/lz4.h#L52).
Both numbering systems used to be independent.
This policy was changed later on, starting with `v1.7.3`.

Makefile
```diff
@@ -96,7 +96,7 @@ clean:
 	@$(MAKE) -C $(ZWRAPDIR) $@ > $(VOID)
 	@$(MAKE) -C examples/ $@ > $(VOID)
 	@$(MAKE) -C contrib/gen_html $@ > $(VOID)
-	@$(RM) zstd$(EXT) zstdmt$(EXT) tmp*
+	@$(RM) -rf zstd$(EXT) zstdmt$(EXT) tmp* lz4
```
Cyan4973: `-f` is implied by `$(RM)` : https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html
`-r` is not, and is only useful for directory.

For clarity, I would prefer to separate deleting files from deleting directories.
Hence, it would keep existing line, and add another one : 
`@$(RM) -r lz4`
This makes it easier to understand that `lz4` is a directory in this context.

Cyan4973:
This looks good !
After the code, and the tests,
comes the doc.

Please update https://github.com/facebook/zstd/blob/dev/programs/README.md , describe the build variable `HAVE_LZ4`.

`zstd -h` is already correctly done : https://github.com/facebook/zstd/blob/dev/programs/zstdcli.c#L151

But we have apparently completely missed the `man` page : https://github.com/facebook/zstd/blob/dev/programs/zstd.1.md
All options regarding `zstd --format=#` should be explained.

Note : the way we create `man` page for `zstd` is uncommon : rather than editing directly `zstd.1` using roff format, which is painful, we edit `zstd.1.md` in markdown format. `zstd.1.md` is then translated into `zstd.1` using [`make man`](https://github.com/facebook/zstd/blob/dev/programs/Makefile#L236), which requires a few dependencies. If you have troubles with these dependencies, we can help you install them, or run the transformation stage on our dev rigs.

programs/README.md
```diff
@@ -40,6 +40,16 @@ There are however other Makefile targets that create different variations of CLI
   In which case, linking stage will fail if `lzma` library cannot be found.
   This might be useful to prevent silent feature disabling.
 
+- __HAVE_LZ4__ : `zstd` can compress and decompress files in `.lz4` formats.
+  This is ordered through commands `--format=lz4`.
+  Alternatively, symlinks named `lz4`, or `unlz4` will mimic intended behavior.
+  `.lz4` support is automatically enabled when `lz4` library is detected at build time.
+  It's possible to disable `.lz4` support, by setting HAVE_LZ4=0 .
+  Example : make zstd HAVE_LZ4=0
+  It's also possible to force compilation with lzma support, using HAVE_LZ4=1.
```
Cyan4973: force compilation with ***lz4*** support

PR #862, opened 2017-09-25
[zstd] Backport kernel patch from @ColinIanKing

terrelln:
* Make the U32 table in `FSE_normalizeCount()` static.
* Patch from https://lkml.kernel.org/r/20170922145946.14316-1-colin.king@canonical.com.
* Clang makes non-static tables static anyways. gcc however, does [weird things](https://godbolt.org/g/fvTcED).
* Benchmarks showed no difference in speed.


Cyan4973:
Looks good to me !

Issue #591, opened 2017-03-07
Ability to make magic number optional


indygreg:
The frame header's 4 byte magic number header can be a substantial size of a full frame when doing dictionary compression on small data. In the case of the Firefox Mercurial repository, there are ~3M "chunks" that need compressed. That would produce ~12MB of zstd magic numbers in what is currently ~1800 MB of zlib-compressed data. That's ~0.67%, which feels high. In some cases, the 4 byte magic number is the difference between zstd (without dictionary compression) being smaller than zlib.

I know there is a block API that can be used to bypass the zstd frame header. But that requires a bit too much wheel reinvention for me. The frame header is generally useful and aside from the magic number I think it is reasonably designed in terms of not wasting space.

I'd like to float the idea for making the magic number optional in both the compressor and decompressor so advanced consumers (who care about size) can omit it while having what is a near fully-functional frame. If this were implemented, I imagine it would be built on top of `ZSTD_setCCtxParameter()` and a to-be-invented `ZSTD_setDDctxParameter()`. I anticipate the cost of checking this flag to have a negligible impact on performance.

Are you receptive to this feature request?

Cyan4973:
I'm mostly worried about the complexity.
Not on the implementation, it's simple enough, but on the presentation side.
Up to now, adding a new feature required to add a new prototype, or a new field, etc.
This is exploding complexity.

We want to keep the interface manageable.
As you have already inferred, we are preparing a new `setParameter` logic.
The new logic is not yet settled, let alone done. But we are working on that.

Once it's in place, new feature, such as this request, will be easier to present.
So it will be time to consider its implementation.

indygreg:
I'm tentatively fine waiting until a `setParameter` API settles before this is implemented. Whatever happens, Mercurial support for compression engines is extensible. So worst case we initially support the canonical frame with the magic number then introduce a one-off compression engine later that has it optional. That's slightly annoying, but not a deal breaker.

Cyan4973:
There's some news on this front.

With latest `dev` branch update, we have a new "advanced API", which is supposed to take over all other ones, to reach "stable" status one day, after sufficient testing time : 
https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L863

This new API features [`ZSTD_CCtx_setParameter()`](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L971) , which is the generic mechanism by which we will introduce the "noMagic" mode requested here.

[...]

PR #863, opened 2017-09-26
magicless frames (#591)

Cyan4973:
This patch make it possible to generate and decode magic-less Zstandard frames, as requested in #591.
To do this, it's required to use the new advanced API to pass relevant parameters.
For this purpose, decoding side has received an advanced API similar to the compression side.

The new format is tested in `fuzzer.c`.

Some other formats are also listed, but they will be removed if no clear way to decode them is implemented before release.


lib/decompress/zstd_decompress.c
```diff
@@ -264,53 +268,79 @@ unsigned ZSTD_isFrame(const void* buffer, size_t size)
...
+ *  decode Frame Header, or require larger `srcSize`.
+ *  note : only works for formats ZSTD_f_zstd1 and ZSTD_f_zstd1_magicless
+ * @return : 0, `zfhPtr` is correctly filled,
+ *          >0, `srcSize` is too small, value is wanted `srcSize` amount,
+ *           or an error code, which can be tested using ZSTD_isError() */
+static size_t ZSTD_getFrameHeader_internal(ZSTD_frameHeader* zfhPtr, const void* src, size_t srcSize, ZSTD_format_e format)
 {
     const BYTE* ip = (const BYTE*)src;
-    if (srcSize < ZSTD_frameHeaderSize_prefix) return ZSTD_frameHeaderSize_prefix;
+    size_t const minInputSize = (format==ZSTD_f_zstd1_magicless) ?
```
terrelln: This is repeated 3+ times, can it be a small function which also contains the (static) asserts?

lib/decompress/zstd_decompress.c
```diff
@@ -264,53 +268,79 @@ unsigned ZSTD_isFrame(const void* buffer, size_t size)
...
+                    ZSTD_frameHeaderSize_prefix - ZSTD_frameIdSize :
+                    ZSTD_frameHeaderSize_prefix;
 
+    ZSTD_STATIC_ASSERT((unsigned)ZSTD_f_zstd1 < (unsigned)ZSTD_f_zstd1_magicless);
+    assert((unsigned)format <= ZSTD_f_zstd1_magicless);  /* only supports formats ZSTD_f_zstd1 and ZSTD_f_zstd1_magicless */
+    if (srcSize < minInputSize) return minInputSize;
+
+    if (format != ZSTD_f_zstd1_magicless)
```
terrelln: Can this be written as `&&` instead?

lib/decompress/zstd_decompress.c
Cyan4973: Yes, some kind of stage check function should do it.
It will be necessary for any new parameter anyway.

PR #864, opened 2017-09-26
Increase maximum window size

terrelln:
* Maximum window size in 32-bit mode is 1GB, since allocations for 2GB fail
  on my Mac.
* Maximum window size in 64-bit mode is 2GB, since that is the largest
  power of 2 that works with the overflow prevention.
* Allow `--long=windowLog` to set the window log, along with
  `--zstd=wlog=#`. These options also set the window size during
  decompression, but don't override `--memory=#` if it is set.
* Present a helpful error message when the window size is too large during
  decompression.
* The long range matcher defaults to a hash log 7 less than the window log,
  which keeps it at 20 for window log 27.
* Keep the default long range matcher window size and the default maximum
  window size at 2^27 for the API and CLI.
* Add tests that use the maximum window size and hash size for compression
  and decompression.

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -102,6 +102,7 @@ static const U32 repStartValue[ZSTD_REP_NUM] = { 1, 4, 8 };
 #define BIT0   1
 
 #define ZSTD_WINDOWLOG_ABSOLUTEMIN 10
+#define ZSTD_WINDOWLOG_DEFAULTMAX 27 /* Default maximum allowed window log */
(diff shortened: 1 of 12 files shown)


terrelln:
I fixed the "real" test failures that showed up so far. The other test failures so far were maybe allocating too much memory, so I will let them run again and see what happens. We might have to skip `windowLog==31` tests on Travis.

Cyan4973:
> We might have to skip windowLog==31 tests on Travis.

Most probably !

PR #870, opened 2017-09-27
Change c++ test to use CXX and CXXFLAGS environment variables

Cyan4973:
Fix OS-X warning on compiling `.c` files with `clang++`
Also changed test name from `gpptest` to `cpptest`, since it's no longer `g++` specific

note : I've also been considering `c++test` as the new name for this test, since it avoids potential confusion with `CPPFLAGS`, but ultimately avoided it in case `+` character could cause problem on some platform such as Windows, or any other intermediate parsing scheme. 
`cxxtest` is another potential candidate.


Cyan4973:
Switched name to `cxxtest` as suggested.

Also to follow : `.travis.yml` in `master` branch will be changed to use this new name.
There is no test in `dev` branch using `cxxtest` (`gpptest`)

PR #871, opened 2017-09-27
[libzstd][opt] Simplify repcode logic

terrelln:
Simplify optimal parser repcode logic so that static analyzers (and people) don't think that it is possible for `opt[cur].off == ZSTD_REP_MOVE_OPT && mlen == 1` to be true and cause a buffer overflow.


Cyan4973:
Looks good to me !
Has it solved static analyser warning too ?

terrelln:
Yeah, it solves the warning.

PR #873, opened 2017-09-28
Leaner tests

Cyan4973:
Adjustments to tests to make them compatible with a memory-starved VM
(tested on Ubuntu Xenial VM with 768 MB RAM).
Works also for `qemu` related tests, notably ARM ones which have been quite flaky.

Longer tests are also tamed, to run faster, hence avoid triggering time-out on travisCI.

Incidentally, this patch also improves behaviour of parameter adjustment in presence of weird compression parameters, and behaviour of `zstdmt` in situation of memory allocation error.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -634,26 +634,31 @@ static U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
     mostly downsizing to reduce memory consumption and initialization.
     Both `srcSize` and `dictSize` are optional (use 0 if unknown),
     but if both are 0, no optimization can be done.
-    Note : cPar is considered validated at this stage. Use ZSTD_checkParams() to ensure that. */
+    Note : cPar is considered validated at this stage. Use ZSTD_checkCParams() to ensure that condition. */
 ZSTD_compressionParameters ZSTD_adjustCParams_internal(ZSTD_compressionParameters cPar, unsigned long long srcSize, size_t dictSize)
 {
     assert(ZSTD_checkCParams(cPar)==0);
-    if (srcSize+dictSize == 0) return cPar;   /* no size information available : no adjustment */
 
-    /* resize params, to use less memory when necessary */
-    {   U32 const minSrcSize = (srcSize==0) ? 500 : 0;
+    /* resize windowLog if src is small, to use less memory when necessary */
+    ZSTD_STATIC_ASSERT(ZSTD_CONTENTSIZE_UNKNOWN == -1ULL);
+    if ( (dictSize || (srcSize+1 > 1))  /* srcSize test depends on static assert condition */
+      && (srcSize-1 < (1ULL<<ZSTD_WINDOWLOG_MAX)) ) /* no correction is srcSize is large enough */ {
+        U32 const minSrcSize = (srcSize==0) ? 513 : 0;
         U64 const rSize = srcSize + dictSize + minSrcSize;
-        if (rSize < ((U64)1<<ZSTD_WINDOWLOG_MAX)) {
+        if (rSize < (1ULL<<ZSTD_WINDOWLOG_MAX)) {
             U32 const srcLog =
-                    MAX(ZSTD_HASHLOG_MIN, (rSize==1) ? 1 : ZSTD_highbit32((U32)(rSize)-1) + 1);
+                    MAX(ZSTD_HASHLOG_MIN,
+                        (rSize==1) ? 1 : ZSTD_highbit32((U32)(rSize)-1) + 1);
             if (cPar.windowLog > srcLog) cPar.windowLog = srcLog;
     }   }
     if (cPar.hashLog > cPar.windowLog) cPar.hashLog = cPar.windowLog;
     {   U32 const cycleLog = ZSTD_cycleLog(cPar.chainLog, cPar.strategy);
-        if (cycleLog > cPar.windowLog) cPar.chainLog -= (cycleLog - cPar.windowLog);
+        if (cycleLog > cPar.windowLog)
+            cPar.chainLog -= (cycleLog - cPar.windowLog);
     }
 
-    if (cPar.windowLog < ZSTD_WINDOWLOG_ABSOLUTEMIN) cPar.windowLog = ZSTD_WINDOWLOG_ABSOLUTEMIN;  /* required for frame header */
+    if (cPar.windowLog < ZSTD_WINDOWLOG_ABSOLUTEMIN)
+        cPar.windowLog = ZSTD_WINDOWLOG_ABSOLUTEMIN;  /* required for frame header */
 
     return cPar;
 }
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -640,7 +640,7 @@ ZSTD_compressionParameters ZSTD_adjustCParams_internal(ZSTD_compressionParameter
     assert(ZSTD_checkCParams(cPar)==0);
 
     /* resize windowLog if src is small, to use less memory when necessary */
-    ZSTD_STATIC_ASSERT(ZSTD_CONTENTSIZE_UNKNOWN == -1ULL);
+    ZSTD_STATIC_ASSERT(ZSTD_CONTENTSIZE_UNKNOWN == (0ULL - 1));
     if ( (dictSize || (srcSize+1 > 1))  /* srcSize test depends on static assert condition */
       && (srcSize-1 < (1ULL<<ZSTD_WINDOWLOG_MAX)) ) /* no correction is srcSize is large enough */ {
         U32 const minSrcSize = (srcSize==0) ? 513 : 0;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -642,7 +642,7 @@ ZSTD_compressionParameters ZSTD_adjustCParams_internal(ZSTD_compressionParameter
     /* resize windowLog if src is small, to use less memory when necessary */
     ZSTD_STATIC_ASSERT(ZSTD_CONTENTSIZE_UNKNOWN == (0ULL - 1));
     if ( (dictSize || (srcSize+1 > 1))  /* srcSize test depends on static assert condition */
-      && (srcSize-1 < (1ULL<<ZSTD_WINDOWLOG_MAX)) ) /* no correction is srcSize is large enough */ {
+      && (srcSize-1 < (1ULL<<ZSTD_WINDOWLOG_MAX)) ) /* no correction when srcSize is large enough */ {
         U32 const minSrcSize = (srcSize==0) ? 513 : 0;
         U64 const rSize = srcSize + dictSize + minSrcSize;
         if (rSize < (1ULL<<ZSTD_WINDOWLOG_MAX)) {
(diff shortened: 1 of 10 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -496,11 +496,29 @@ static void ZSTDMT_releaseAllJobResources(ZSTDMT_CCtx* mtctx)
...
+    }
+}
+
 size_t ZSTDMT_freeCCtx(ZSTDMT_CCtx* mtctx)
 {
     if (mtctx==NULL) return 0;   /* compatible with free on NULL */
     POOL_free(mtctx->factory);
-    if (!mtctx->allJobsCompleted) ZSTDMT_releaseAllJobResources(mtctx); /* stop workers first */
+    if (!mtctx->allJobsCompleted) {
+        ZSTDMT_waitForAllJobsCompleted(mtctx);
```
terrelln: Is this necessary? All the workers should be stopped because we called `POOL_free()`.

lib/compress/zstd_compress.c
```diff
@@ -643,26 +643,31 @@ static U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
...
 {
     assert(ZSTD_checkCParams(cPar)==0);
-    if (srcSize+dictSize == 0) return cPar;   /* no size information available : no adjustment */
 
-    /* resize params, to use less memory when necessary */
-    {   U32 const minSrcSize = (srcSize==0) ? 500 : 0;
+    /* resize windowLog if src is small, to use less memory when necessary */
+    ZSTD_STATIC_ASSERT(ZSTD_CONTENTSIZE_UNKNOWN == (0ULL - 1));
+    if ( (dictSize || (srcSize+1 > 1))  /* srcSize test depends on static assert condition */
+      && (srcSize-1 < (1ULL<<ZSTD_WINDOWLOG_MAX)) ) /* no correction when srcSize is large enough */ {
```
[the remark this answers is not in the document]
Cyan4973: > Doesn't rsize < (1ULL<<ZSTD_WINDOWLOG_MAX) already cover this?

It avoids a potential overflow for extraordinarily large values of `srcSize`.
Presuming that `srcSize == 0ULL-2`, it passes `(srcSize+1 > 1)` test (which is only there to filter `srcSize==0` and `srcSize=ZSTD_CONTENTSIZE_UNKNOWN`)
but now `rSize = srcSize + dictSize + minSrcSize;` can be `< (1ULL<<ZSTD_WINDOWLOG_MAX)`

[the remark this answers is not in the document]
Cyan4973: > Also, if srcSize == 0 && dictSize > 0 we won't adjust parameters now.

Good point.
This formula interprets `srcSize == 0` as "`srcSize` is unknown" and potentially very large.
But previous formula would interpret "unknown" as "probably small if there is a dictionary",
and basically evaluate it to `minSrcSize`.

I'll update the formula to recover this behaviour.

lib/compress/zstdmt_compress.c
Cyan4973: Good point.
I understand the code the same way, but here was my experience : 

while testing on the memory-limited VM, I got a few "memory allocation" errors.
The more difficult test was when using `zstreamtest --newapi`.
After taking care of the more important ones (they would break on trying to allocate large `workplace`), I started to experience rarer and more random allocation errors on input buffer allocation : 
https://github.com/facebook/zstd/pull/873/files/e4ec42772066fb0c30b2d43ab6e75108815bb86a#diff-92cfacb89db0f858d089c8021c329941L1019
(note : this error is no longer possible in this version, since the code just move on, trying to flush some data to free some memory, instead of issuing an error).

Thing is, it would not just issue an error (which was correctly logged on console) : `zstreamtest` would crash immediately after. The error message wasn't very clear (double free, but no trace nor error position).
I interpreted that scenario as a potential error in `ZSTDMT_freeCCtx()`, which is called immediately after detecting the error.
So I thought : maybe some worker are still running, let's be sure they are no longer working on the memory we are trying to free.
So I added this line.
And the crash stopped.

Retrospectively, it feels strange, as indeed, `POOL_free()` should have already taken care of this situation.

The bug is hard to reproduce because it's a bit random, depending on memory pressure. It does not reliably happen at exactly the same place, though some configurations are more likely to trigger it because they require more memory.

It cannot be reproduced with current version of `shortests` branch.
One would need to go back in time, finding an intermediate version which still feature the issue.
Likely this one : d6770f80af091f208c2bfc2f49281984b00be52b
Even then, the memory pressure must be "just right", to not fail too soon, during context creation, but still fail while requesting an input buffer.

Cyan4973: I could reproduce the crash, using above version.
Here is the corresponding trace : 
```
  3220/ 10000    Error => ZSTD_compress_generic(zc, &outBuff, &inBuff, flush) : Allocation error : not enough memory  (seed 6328, test nb 3220)        [12/4776]
*** Error in `./zstreamtest': double free or corruption (top): 0x0000000001d8d770 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f4c770077e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7f4c7701037a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7f4c7701453c]
./zstreamtest[0x44bfc5]
./zstreamtest[0x43a9bd]
./zstreamtest[0x4656f7]
./zstreamtest[0x4012ed]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f4c76fb0830]
./zstreamtest[0x4041a9]
======= Memory map: ========
00400000-00479000 r-xp 00000000 08:01 1585860                            /home/cyan/dev/zstd/github/tests/zstreamtest
00678000-00679000 r--p 00078000 08:01 1585860                            /home/cyan/dev/zstd/github/tests/zstreamtest
00679000-0067a000 rw-p 00079000 08:01 1585860                            /home/cyan/dev/zstd/github/tests/zstreamtest
01d8c000-01dae000 rw-p 00000000 00:00 0                                  [heap]
7f4c2bb4e000-7f4c35d44000 rw-p 00000000 00:00 0 
7f4c48000000-7f4c4927b000 rw-p 00000000 00:00 0 
7f4c4927b000-7f4c4c000000 ---p 00000000 00:00 0 
7f4c50000000-7f4c52001000 rw-p 00000000 00:00 0 
7f4c52001000-7f4c54000000 ---p 00000000 00:00 0 
7f4c54000000-7f4c55678000 rw-p 00000000 00:00 0 
7f4c55678000-7f4c58000000 ---p 00000000 00:00 0 
7f4c58000000-7f4c5947e000 rw-p 00000000 00:00 0 
7f4c5947e000-7f4c5c000000 ---p 00000000 00:00 0 
7f4c5d538000-7f4c5d539000 ---p 00000000 00:00 0 
7f4c5d539000-7f4c5dd39000 rw-p 00000000 00:00 0 
7f4c5dd39000-7f4c5dd3a000 ---p 00000000 00:00 0 
7f4c5dd3a000-7f4c5e53a000 rw-p 00000000 00:00 0 
7f4c5e53a000-7f4c5e53b000 ---p 00000000 00:00 0 
7f4c5e53b000-7f4c66d7c000 rw-p 00000000 00:00 0 
7f4c693e6000-7f4c693fc000 r-xp 00000000 08:01 791755                     /lib/x86_64-linux-gnu/libgcc_s.so.1
7f4c693fc000-7f4c695fb000 ---p 00016000 08:01 791755                     /lib/x86_64-linux-gnu/libgcc_s.so.1
7f4c695fb000-7f4c695fc000 rw-p 00015000 08:01 791755                     /lib/x86_64-linux-gnu/libgcc_s.so.1
7f4c695fc000-7f4c6e6f7000 rw-p 00000000 00:00 0 
7f4c6e6f7000-7f4c6e6f8000 ---p 00000000 00:00 0 
7f4c6e6f8000-7f4c76f90000 rw-p 00000000 00:00 0 
7f4c76f90000-7f4c77150000 r-xp 00000000 08:01 791717                     /lib/x86_64-linux-gnu/libc-2.23.so
7f4c77150000-7f4c77350000 ---p 001c0000 08:01 791717                     /lib/x86_64-linux-gnu/libc-2.23.so
7f4c77350000-7f4c77354000 r--p 001c0000 08:01 791717                     /lib/x86_64-linux-gnu/libc-2.23.so
7f4c77354000-7f4c77356000 rw-p 001c4000 08:01 791717                     /lib/x86_64-linux-gnu/libc-2.23.so
7f4c77356000-7f4c7735a000 rw-p 00000000 00:00 0 
7f4c7735a000-7f4c77372000 r-xp 00000000 08:01 791863                     /lib/x86_64-linux-gnu/libpthread-2.23.so
7f4c77372000-7f4c77571000 ---p 00018000 08:01 791863                     /lib/x86_64-linux-gnu/libpthread-2.23.so
7f4c77571000-7f4c77572000 r--p 00017000 08:01 791863                     /lib/x86_64-linux-gnu/libpthread-2.23.so
7f4c77572000-7f4c77573000 rw-p 00018000 08:01 791863                     /lib/x86_64-linux-gnu/libpthread-2.23.so
```

Cyan4973: Actually, any kind of modification, even unrelated ones such as merely enabling log traces, would make the crashing bug disappear.

Cyan4973: OK, so since there is no relation between the "fix" and the issue, let's remove this invocation.

terrelln: Do you have a trace with debug symbols?

Cyan4973: `-g` flag was used on the binary which created this trace.

Though if you mean "debug symbols within standard lib", I'll need some help to trigger those ones.

terrelln: Thats strange, I wonder why it wasn't symbolized then. If you still have the binary that was running (or maybe just the compiler options), I think `llvm-symbolizer -obj=zstreamtest -pretty-print < $BACKTRACE_FILE` will give you the symbols.

Cyan4973: Binary is no longer there, but I can try to generate the issue again.
It was just the regular `make zstreamtest` using d6770f80af091f208c2bfc2f49281984b00be52b commit

terrelln: I wasn't able to build the same binary, but if you can, this command should work in the test directory

```
< backtrace.txt llvm-symbolizer
```

where `backtrace.txt` is

```
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5) 0x7f4c770077e5
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a) 0x7f4c7701037a
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c) 0x7f4c7701453c
./zstreamtest 0x44bfc5
./zstreamtest 0x43a9bd
./zstreamtest 0x4656f7
./zstreamtest 0x4012ed
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0) 0x7f4c76fb0830
./zstreamtest 0x4041a9
```

Cyan4973: I tried to use `llvm-symbolizer`, but it just re-prints the exact content of the trace.

Cyan4973: tried without brackets too, no better ...

Cyan4973: OK, found it and fixed it.
It's an even more complex scenario than anticipated, where a buffer from the pool could be freed twice.
Sent an additional patch into PR which makes this situation impossible now.

PR #876, opened 2017-09-30
CLI Fix : srcSize written in frame headers when compressing multiple files

Cyan4973:
It used to be disabled, likely to circumvent a bug that this patch fix 
(bad usage of the new api by `fileio.c`).

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -423,7 +423,7 @@ static cRess_t FIO_createCResources(const char* dictFileName, int cLevel,
             CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_p_contentSizeFlag, srcIsRegularFile) );
             CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_p_dictIDFlag, g_dictIDFlag) );
             CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_p_checksumFlag, g_checksumFlag) );
-            CHECK( ZSTD_CCtx_setPledgedSrcSize(ress.cctx, srcSize) );
+            (void)srcSize;
             /* compression level */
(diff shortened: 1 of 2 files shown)


programs/fileio.c
```diff
@@ -974,7 +979,7 @@ int FIO_compressMultipleFilenames(const char** inFileNamesTable, unsigned nbFile
     char*  dstFileName = (char*)malloc(FNSPACE);
     size_t const suffixSize = suffix ? strlen(suffix) : 0;
     U64 const srcSize = (nbFiles != 1) ? 0 : UTIL_getFileSize(inFileNamesTable[0]) ;
-    int const isRegularFile = (nbFiles > 1) ? 0 : UTIL_isRegularFile(inFileNamesTable[0]);  /* won't write frame content size when nbFiles > 1 */
+    int const isRegularFile = (nbFiles > 1) ? 1 : UTIL_isRegularFile(inFileNamesTable[0]);  /* if nbFiles > 1, it's not stdin */
```
terrelln: I don't think thats quite true, `zstd < file1 - file2 -c`

Cyan4973: `isRegularFile` is actually no longer required, 
so it will be removed.

terrelln:
I believe Circle CI is failing because, unlike Travis, it runs based off the commit the PR is based off of, while Travis runs based off of the result of merging it into `dev`. So the fix for the `block_round_trip` regression test hasn't been picked up yet.

Cyan4973:
It's also the only environment where `make -j regressiontest` is run

terrelln:
Its also run on Travis with ASAN/UBSAN/MSAN (`uasanregressiontest`, `msanregressiontest`). Circle CI only runs it without sanitizers, so it finishes quicker.

PR #877, opened 2017-09-30
added ZSTD_COMPRESSBOUND() as a macro

Cyan4973:
ZSTD_compressBound() works fine, but is only useful for dynamic allocation.
For static allocation purposes, a macro is needed to provide a value at compilation time.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -131,10 +131,11 @@ ZSTDLIB_API unsigned long long ZSTD_getDecompressedSize(const void* src, size_t 
 
 
 /*======  Helper functions  ======*/
-ZSTDLIB_API int         ZSTD_maxCLevel(void);               /*!< maximum compression level available */
+#define ZSTD_COMPRESSBOUND(srcSize)   ((srcSize) + ((srcSize)>>9) + (((srcSize) < 128 KB) ? ((128 KB - (srcSize)) >> 11) /* margin, from 64 to 0 */ : 0))
 ZSTDLIB_API size_t      ZSTD_compressBound(size_t srcSize); /*!< maximum compressed size in worst case scenario */
 ZSTDLIB_API unsigned    ZSTD_isError(size_t code);          /*!< tells if a `size_t` function result is an error code */
 ZSTDLIB_API const char* ZSTD_getErrorName(size_t code);     /*!< provides readable string from an error code */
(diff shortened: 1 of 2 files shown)


lib/zstd.h
```diff
@@ -131,10 +131,11 @@ ZSTDLIB_API unsigned long long ZSTD_getDecompressedSize(const void* src, size_t
 
 
 /*======  Helper functions  ======*/
-ZSTDLIB_API int         ZSTD_maxCLevel(void);               /*!< maximum compression level available */
+#define ZSTD_COMPRESSBOUND(srcSize)   ((srcSize) + ((srcSize)>>9) + (((srcSize) < 128 KB) ? ((128 KB - (srcSize)) >> 11) /* margin, from 64 to 0 */ : 0))
```
terrelln: I don't think this works with `windowSize == 1024`.

terrelln: https://godbolt.org/g/T28RHB

Cyan4973: where does `extra` comes from ?

terrelln: 18 bytes for the header, 3 bytes per block of 1024 bytes, and 4 bytes for a checksum.

Cyan4973: That's pretty good.
So this calculation is for a scenario where `windowSize` is set with minimum allowed size.

I will restore the `srcSize >> 8` formula then.
Using `(srcSize >> 10) * 3` , or even `extra` formula, is tempting, but unfortunately, it requires to add a constant (header and footer).
As soon as a constant is present, it's no longer possible to guarantee that the sum of 2 limits <= limit of the sum, which is important for `zstdmt`, as it makes it possible to run `ZSTDMT_compress()` with no extra memory (using only the output buffer, as long as `dstCapacity >= ZSTD_compressBound(srcSize)`).

Alternatively, we could make `zstdmt` _not_ depend on `ZSTD_compressBound()` and use its own internal formula, guaranteeing the wanted property. I instinctively preferred preserving this correlation with `ZSTD_compressBound()`.

terrelln: Can you add a comment explaining that? I had forgotten that we need the sum of 2 limits <= limit of sum for zstdmt.

Cyan4973: Done.
To be fair, I don't think this property matters to users.
It's mostly important for us (dev)

PR #884, opened 2017-10-08
Guard uses of __builtin_clz and __builtin_clzll

allanjude:
Prevents undefined reference to `__clzdi2' on FreeBSD sparc64 compiled with gcc 4.2

Fallback to the software implementation by making the preprocessor guard better

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -175,7 +175,7 @@ MEM_STATIC unsigned BIT_highbit32 (register U32 val)
         unsigned long r=0;
         _BitScanReverse ( &r, val );
         return (unsigned) r;
-#   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* Use GCC Intrinsic */
+#   elif defined(__GNUC__) && (__GNUC__ >= 3) && __has_builtin(__builtin_clz)   /* Use GCC Intrinsic */
         return 31 - __builtin_clz (val);
 #   else   /* Software version */
         static const unsigned DeBruijnClz[32] = { 0,  9,  1, 10, 13, 21,  2, 29,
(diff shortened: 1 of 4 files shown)


allanjude:
I see that this only works in clang, and does not fail in gcc on FreeBSD because of a compatibility macro.

Will need to revisit this

Cyan4973:
Maybe using [`__GNUC_MINOR__`](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html#Common-Predefined-Macros) ?

Also, is there a difference to do between `__builtin_clz` and `__builtin_clzll`, the 32-bits version being more "commonly" supported ?

cemeyer:
We could provide a similar compatibility macro in Zstd, right Allan?

As it is, the lack of (something like) this revision in Zstandard makes backporting new versions to FreeBSD more painful than necessary.  (Some of our architectures do not provide `__builtin_clz` or `__builtin_clzll`.)

Cyan4973:
note that `__has_builtin` is `clang` specific, it will not work on `gcc`.

allanjude:
Yeah, the problem with the FreeBSD compatibility macro, is that is just returns false, which would be a large pessimization on modern GCC that does have it.

cemeyer:
Sure.  We could do `HAVE_BUILTIN_CLZ(LL)` instead of the specific checks for GCC versions.  There is precedent in the tree, e.g., `DYNAMIC_BMI2`.

terrelln:
Yeah, we can definitely make a macro for the test in `compiler.h` so updates only have to change one place.

tsoome:
There is also an problem that even in (recent) gcc, the __has_builtin() itself may be missing; meaning that lib/common/compiler.h should have something like:

#ifdef __has_builtin
#define HAS_BUILTIN(x)  __has_builtin(x)
#else
#define HAS_BUILTIN(x) 0
#endif

and then you can use HAS_BUILTIN() ...

cemeyer:
@tsoome Yeah, that is basically the compatibility macro that just returns false that @allanjude is describing.

Instead we are suggesting specific `HAVE_BUILTIN_FOO` macros for the feature in compiler.h, that can be override by distro/architectures in one place.  Then code that uses `__builtin_clz` or whatever is guarded by one of those macros.

PR #889, opened 2017-10-13
Fix invalid use of dictionary offcode table

### LINKED ISSUE #888: Compression fails when using custom dict in ZSTD_compress_advanced
Reproducible only at compression levels 17 and higher. Reproducible with different dictionaries.

I attach to this ticket all required data to reproduce this bug.

ZSTD versions: 1.3.1, 1.3.2

Compilers:
GCC 6.4.1
GCC 7.2.1

Optimization levels: -O0 through -O3

Attachment content:
SimplePacker.c - code snippet for compression/decompression
input_data - file to encode
first.dict, second.dict - dictionaries with which it can be reproduced

```Bash
> SimplePacker c first.dict input_file temp.zstd
> SimplePacker x first.dict temp.zstd output_file
> diff input_file output_file
Binary files input_file and output_file differ

> SimplePacker c second.dict input_file temp.zstd
> SimplePacker x second.dict temp.zstd output_file
Decompression error!
> diff input_file output_file
Binary files input_file and output_file differ
```

When compiling with `-fsanitize=undefined -fsanitize=address` following is reported during compression:

```
common/bitstream.h:234:44: runtime error: index 48830 out of bounds for type 'unsigned int [32]'
common/bitstream.h:234:44: runtime error: load of address 0x0000006fb298 with insufficient space for an object of type 'const unsigned int'
0x0000006fb298: note: pointer points here
<memory cannot be printed>
ASAN:DEADLYSIGNAL
=================================================================
==28892==ERROR: AddressSanitizer: SEGV on unknown address 0x0000006fb298 (pc 0x00000042228b bp 0x7ffc3cddad90 sp 0x7ffc3cdda710 T0)
==28892==The signal is caused by a READ memory access.
    #0 0x42228a in ZSTD_compressBlock_internal (/home/umogslayer/zenith/zenith-engine/Src/TcPacker/src/1.3.2/SimplePacker+0x42228a)
    #1 0x42ea61 in ZSTD_compressEnd (/home/umogslayer/zenith/zenith-engine/Src/TcPacker/src/1.3.2/SimplePacker+0x42ea61)
    #2 0x4318c5 in ZSTD_compress_advanced (/home/umogslayer/zenith/zenith-engine/Src/TcPacker/src/1.3.2/SimplePacker+0x4318c5)
    #3 0x4032ac in main (/home/umogslayer/zenith/zenith-engine/Src/TcPacker/src/1.3.2/SimplePacker+0x4032ac)
    #4 0x7f63f1b1cf49 in __libc_start_main (/lib64/libc.so.6+0x20f49)
    #5 0x403739 in _start (/home/umogslayer/zenith/zenith-engine/Src/TcPacker/src/1.3.2/SimplePacker+0x403739)
```
[report_bug.tar.gz](https://github.com/facebook/zstd/files/1381521/report_bug.tar.gz)

terrelln:
Fixes #888.

Improvements to the test suite coming next.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1282,6 +1282,7 @@ symbolEncodingType_e ZSTD_selectEncodingType(
 #define MAX_SEQ_FOR_STATIC_FSE  1000
     ZSTD_STATIC_ASSERT(ZSTD_defaultDisallowed == 0 && ZSTD_defaultAllowed != 0);
     if ((mostFrequent == nbSeq) && (!isDefaultAllowed || nbSeq > 2)) {
+        DEBUGLOG(5, "Selected set_rle");
         /* Prefer set_basic over set_rle when there are 2 or less symbols,
          * since RLE uses 1 byte, but set_basic uses 5-6 bits per symbol.
          * If basic encoding isn't possible, always choose RLE.
@@ -1291,13 +1292,16 @@ symbolEncodingType_e ZSTD_selectEncodingType(
     }
     if ( isDefaultAllowed
       && (*repeatMode == FSE_repeat_valid) && (nbSeq < MAX_SEQ_FOR_STATIC_FSE)) {
+        DEBUGLOG(5, "Selected set_repeat");
         return set_repeat;
     }
     if ( isDefaultAllowed
       && ((nbSeq < MIN_SEQ_FOR_DYNAMIC_FSE) || (mostFrequent < (nbSeq >> (defaultNormLog-1)))) ) {
+        DEBUGLOG(5, "Selected set_basic");
         *repeatMode = FSE_repeat_valid;
         return set_basic;
     }
+    DEBUGLOG(5, "Selected set_compressed");
     *repeatMode = FSE_repeat_check;
     return set_compressed;
 }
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1536,6 +1536,12 @@ MEM_STATIC size_t ZSTD_compressSequences(seqStore_t* seqStorePtr,
     int const uncompressibleError = (cSize == ERROR(dstSize_tooSmall)) && (srcSize <= dstCapacity);
     if (ZSTD_isError(cSize) && !uncompressibleError)
         return cSize;
+    /* We check that dictionaries have offset codes available for the first
+     * block. After the first block, the offcode table might not have large
+     * enough codes to represent the offsets in the data.
+     */
+    if (entropy->offcode_repeatMode == FSE_repeat_valid)
+        entropy->offcode_repeatMode = FSE_repeat_check;
```
Cyan4973: I'm not completely sure to understand what this line does.
What's the meaning of `FSE_repeat_check` ?
It seems to make "repeat" mode selection impossible. Is that correct ?

If yes, what about the first block ? Can it still benefit from dictionary's offset table ?

terrelln: This line is the equivalent of [this line](https://github.com/facebook/zstd/commit/830ef4152a8b7568916ecb600a0aac1a5212c08f#diff-62f99f3a22fe685ec153abb360cca5d5L1136), but only for offcodes. Its called after `ZSTD_compressSequences_internal()` finishes, so the first block can still use the dictionaries table (I just doubled checked that it is used).

Right now, we don't do anything with `check`, it is the same as `none`. The `valid` repeat mode means you can assume the table is valid. `check` means a table exists, but you need to ensure that it has non-zero probabilities for all the symbols you require. `none` means we don't have a table to repeat (first block, or previous block was incompressible).

PR #890, opened 2017-10-13
Don't allow default tables to repeat

terrelln:
It isn't useful in any case to repeat default tables.
Saves a few bytes on Silesia, since we don't trigger the dictionary heuristic.

Before: 211988480 => 73651998 bytes
After:  211988480 => 73651721 bytes

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1298,7 +1298,13 @@ symbolEncodingType_e ZSTD_selectEncodingType(
     if ( isDefaultAllowed
       && ((nbSeq < MIN_SEQ_FOR_DYNAMIC_FSE) || (mostFrequent < (nbSeq >> (defaultNormLog-1)))) ) {
         DEBUGLOG(5, "Selected set_basic");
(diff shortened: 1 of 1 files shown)


Cyan4973:
Feels good to separate both logics.

For completeness, the only minor advantage I could see for repeated tables is that they are already built, hence re-used in place.

For decompression, default distribution tables are already pre-baked in binary, so there is no build time either. It's just a different memory segment, so maybe a small impact on cpu cache (unlikely to be measurable).

Compression side rebuilds default tables for each block, so in this case, there could be a small benefit.
But it's probably very small, maybe not even measurable : default tables are tiny, making build time very fast; this operation is a very small fraction of global compression time, even for fast compression levels.
Worst case, if it ends up being measurable, these tables could be pre-baked in binary, like decompression tables.

So yes, better separate the logic, it's not only more efficient : it's also simpler to maintain.

terrelln:
We could have a `FSE_repeat_default` and `FSE_repeat_dictionary`

Cyan4973:
some minor conflict to resolve

PR #891, opened 2017-10-14
Content size

Cyan4973:
This patch enables 3 things : 
- when using `zstd` cli : ensure that `srcSize` is stored in frame header when known 
- when using `API` : ensure that `contentSizeFlag` is set to `1` by default
- when compressing one file using a dictionary : use `srcSize` to better estimate compression parameters.

What it does not change : 
- when compressing _multiple files_ using a dictionary : it still "bets" that future `srcSize` are (likely) going to be small.


lib/compress/zstd_compress.c
```diff
@@ -2568,13 +2574,19 @@ size_t ZSTD_initCStream_usingCDict(ZSTD_CStream* zcs, const ZSTD_CDict* cdict)
     return ZSTD_initCStream_usingCDict_advanced(zcs, cdict, fParams, 0);  /* note : will check that cdict != NULL */
 }
 
+/* ZSTD_initCStream_advanced() :
+ * pledgedSrcSize is optional: it can be 0, which means "unknown" if contentSizeFlag is set, or "unknown" if contentSizeFlag is not set. */
```
terrelln: Doesn't it mean zero if contentSizeFlag is set?

Cyan4973: Yes totally.
The comment has been changed in [`zstd.h`](https://github.com/facebook/zstd/pull/891/files#diff-71b1d36582455b38213ddb321e85c65bR737).
Maybe I should change it into `zstd_compress.c` as well.

Cyan4973:
Issue with Visual build fixed.

It was multiple reasons combined : 
- Visual uses a different code path, because `ZSTD_NEWAPI` is not defined in the project
- In this path, `fileio.c` sets `pledgedSrcSize=0` to mean "unknown". It used to be correct, and should still be correct according to function documentation. However, it's better to use `ZSTD_CONTENTSIZE_UNKNOWN` now.
- `ZSTDMT_resetCCtx()` interprets `pledgedSrcSize==0` as "empty", instead of "unknown", in contrast with its own documentation in `zstdmt_compress.h`.

Fixing a single of these reasons is enough avoid the issue.
The patch 32c9f715ae1532d4ea06ee5f811ed61c46e48d31 goes a bit further : 
- `zstdmt_compress.h` : `ZSTDMT_resetCCtx()` documentation now firmly invites using `ZSTD_CONTENTSIZE_UNKNOWN` to mean "unknown".
- `ZSTDMT_resetCCtx()` honor its legacy commitment that `0` means "unknown".
- `fileio.c` sends `pledgedSrcSize=ZSTD_CONTENTSIZE_UNKNOWN` to mean "unknown".

Sidenote : `fileio.c` still confuses an empty file with an unknown size file,
because [`UTIL_getFileSize`](https://github.com/facebook/zstd/blob/dev/programs/util.h#L316) will return `0` for both cases. It's not new, it has always been the case.

There might be a way to distinguish both situations, by combining `UTIL_getFileSize()` with `UTIL_isRegularFile()`. What is not yet clear is if this result should be provided directly by `UTIL_getFileSize()` or be recombined externally. `ZSTD_CONTENTSIZE_UNKNOWN` is very Zstandard specific, `UTIL_getFileSize()` should not be tied to it.

Cyan4973:
A follow up action could be to reduce code complexity by keeping a single code path in `fileio.c` and `bench.c`, the `ZSTD_NEWAPI` one, which is now stable enough.

I believe it's better to make it a separate PR though.

Cyan4973:
A change in `UTIL_getFileSize()` to report an error instead `0` when it cannot determine the file size, translates into a lot of other changes in the code base.
So it will have to be its own separate patch.

PR #893, opened 2017-10-17
Fix LZ4 Compression Bug

felixhandte:
Fixes issue where, when `zstd --format=lz4` is fed an input larger than 128KB,
the read overruns the input buffer.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -616,11 +616,16 @@ static unsigned long long FIO_compressLzmaFrame(cRess_t* ress,
 #endif
 
 #ifdef ZSTD_LZ4COMPRESS
+#if LZ4_VERSION_NUMBER <= 10600
+#define LZ4F_blockLinked blockLinked
+#define LZ4F_max64KB max64KB
+#endif
 static int FIO_LZ4_GetBlockSize_FromBlockId (int id) { return (1 << (8 + (2 * id))); }
 static unsigned long long FIO_compressLz4Frame(cRess_t* ress,
                             const char* srcFileName, U64 const srcFileSize,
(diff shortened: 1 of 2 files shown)


felixhandte:
This is basically the minimal patch that fixes the issue. Not sure if this is actually what we want to commit.

This all is a little sketchy anyways. For example, the destination buffer is sized to ZSTD_compressBound(ZSTD_BLOCKSIZE_MAX), which doesn't intrinsically have anything to do with LZ4_compressBound().

Cyan4973:
Indeed, `cRess_t* ress` was created for Zstandard usage,
so I don't think it's a good idea to use these ressources for LZ4.
I'd prefer `FIO_compressLz4Frame()` creating and using its own buffers.

terrelln:
Currently the rest of the algorithms also use `cRess_t* ress`. But their APIs don't depend on the size of the buffers.

Cyan4973:
Maybe we could extract useful things from `ress` to provide as arguments to 3rd-party algorithm, when useful.
One such example could be the dictionary.

But `ress` itself does not feel suitable for something else than `zstd`.

Cyan4973:
Could we add a test to this PR, that shows that the fix works (i.e. it fails with previous version) ?

Cyan4973:
Is this patch going to be cancelled, in favor of a new one ?

felixhandte:
Ah, no, I've updated this PR to include the fix we discussed--switching zstd's LZ4 impl to 64KB chained blocks. I will add a test, though.

felixhandte:
I've added a simple test that fails without this changeset, but passes with it.

PR #894, opened 2017-10-17
Add `check` Target to Makefile

felixhandte:


Cyan4973:
Is there a specific reason to add this target ?

Cyan4973:
https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets

PR #895, opened 2017-10-18
Distinguish 0-size from size-unavailable

Cyan4973:
This patch introduces an error code for `UTIL_getFileSize()`, called `UTIL_FILESIZE_UNKNOWN`.

Up to now, `UTIL_getFileSize()` would answer `0` to mean "error", making it impossible to separate from a genuine empty file.

Most users of `UTIL_getFileSize()` were hard-wired to consider `0` as "error", so this change introduces many corrections throughout the code base.

0-size files are now correctly reported (header includes the `0` content size), while non-files (pipes) continue having an unknown size.

Is this functionality worth its complexity ?
Not sure, though I like the fact that checking for a file size error is now more explicit.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -25,7 +25,7 @@
 *  Includes
 ***************************************/
 #include "platform.h"   /* Large Files support, SET_BINARY_MODE */
-#include "util.h"       /* UTIL_getFileSize */
+#include "util.h"       /* UTIL_getFileSize, UTIL_isRegularFile */
 #include <stdio.h>      /* fprintf, fopen, fread, _fileno, stdin, stdout */
 #include <stdlib.h>     /* malloc, free */
 #include <string.h>     /* strcmp, strlen */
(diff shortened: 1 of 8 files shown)


Cyan4973:
Note : the visual studio test cannot be successful, as it currently uses the `ZSTD_MULTITHREAD` code path, which is [unable to express `srcSize==0`](https://github.com/facebook/zstd/blob/dev/lib/compress/zstdmt_compress.h#L53).

programs/fileio.c
```diff
@@ -564,7 +564,7 @@ static unsigned long long FIO_compressGzFrame(cRess_t* ress,
                 strm.avail_out = (uInt)ress->dstBufferSize;
             }
         }
-        if (!srcFileSize)
+        if (srcFileSize != UTIL_FILESIZE_UNKNOWN)
```
terrelln: `srcFileSize == UTIL_FILESIZE_UNKNOWN` here and below.

programs/fileio.c
```diff
@@ -851,13 +850,13 @@ static int FIO_compressFilename_internal(cRess_t ress,
                 compressedfilesize += outBuff.pos;
         }   }
         if (g_nbThreads > 1) {
-            if (!fileSize)
+            if (fileSize != UTIL_FILESIZE_UNKNOWN)
```
terrelln: `==`

programs/fileio.c
Cyan4973: Nice catch !

PR #898, opened 2017-10-20
Update link to Java decoder from Presto to Airlift

trixpan:
@martint seems to have use the Presto Zstandard code as part of airlift.

Given airlift aircompressor is on Maven Central it makes more sense to use it as the reference of his implementation of Zstd in pure Java.


trixpan:
@facebook-github-bot CLA signed, thanks

PR #905, opened 2017-11-01
Allow skippable frames of any size

### LINKED ISSUE #904: zstd error out on skippable frames
Hi,
Zstd framing format is compatible with Lz4 for skippable frames (per @Cyan4973 and docs)
lz4 utility is skipping skippable frames, but zstd does not.
I created 48-byte file with one skippable frame:

 hexdump -C /tmp/test.zst
00000000  5e 2a 4d 18 28 00 00 00  00 00 00 00 00 00 00 00  |^*M.(...........|
00000010  00 00 00 00 16 03 01 00  c3 ff 0f 00 03 00 00 00  |................|
00000020  00 00 00 00 00 00 00 00  00 00 00 00 00 00 e3 ce  |................|
00000030

lz4 handles it:

lz4 -t /tmp/test.zst
/tmp/test.zst        : decoded 0 bytes  

zstd failing:

zstd -t /tmp/test.zst
/tmp/test.zst : Decoding error (36) : Corrupted block detected 

zstd tested on today's version of repo.

Please modify zstd utility to silently skip skippable frames (that is exactly what skippable frames for).
I am creating files with skippable frames that contains transport info, and I want to be able to decompress it using standard tools.

Also, I do have impression that originally zstd was handling skippable frames correctly - means silently skipping htem, like lz4? Am I wrong?

Thank you

terrelln:
Previously we were only able to decode skippable frames that are smaller than `zds->inBuffSize`.

Fixes #904.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -2555,17 +2555,21 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
             /* fall-through */
         case zdss_load:
             {   size_t const neededInSize = ZSTD_nextSrcSizeToDecompress(zds);
-                size_t const toLoad = neededInSize - zds->inPos;   /* should always be <= remaining space within inBuff */
+                size_t const toLoad = neededInSize - zds->inPos;
+                int const isSkipFrame = ZSTD_isSkipFrame(zds);
                 size_t loadedSize;
-                if (toLoad > zds->inBuffSize - zds->inPos) return ERROR(corruption_detected);   /* should never happen */
-                loadedSize = ZSTD_limitCopy(zds->inBuff + zds->inPos, toLoad, ip, iend-ip);
+                if (isSkipFrame) {
+                    loadedSize = MIN(toLoad, (size_t)(iend-ip));
+                } else {
+                    if (toLoad > zds->inBuffSize - zds->inPos) return ERROR(corruption_detected);   /* should never happen */
+                    loadedSize = ZSTD_limitCopy(zds->inBuff + zds->inPos, toLoad, ip, iend-ip);
+                }
                 ip += loadedSize;
                 zds->inPos += loadedSize;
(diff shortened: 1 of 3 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -2557,8 +2557,12 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
             {   size_t const neededInSize = ZSTD_nextSrcSizeToDecompress(zds);
                 size_t const toLoad = neededInSize - zds->inPos;   /* should always be <= remaining space within inBuff */
```
Cyan4973: The comment is no longer correct

lib/decompress/zstd_decompress.c
```diff
@@ -2557,8 +2557,12 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
...
                 size_t loadedSize;
-                if (toLoad > zds->inBuffSize - zds->inPos) return ERROR(corruption_detected);   /* should never happen */
-                loadedSize = ZSTD_limitCopy(zds->inBuff + zds->inPos, toLoad, ip, iend-ip);
+                if (ZSTD_isSkipFrame(zds)) {
```
Cyan4973: very minor : 
we have `int const isSkipFrame = ZSTD_isSkipFrame(zds);` a few lines below.
Maybe it could be upped here.
The idea is to use the same `isSkipFrame` value.
This is a guarantee that it remains the same value between the 2 places where `ZSTD_isSkipFrame(zds)` is currently invoked.

Cyan4973:
Fix looks good to me.

Let's add a test case with a large skippable frame.

There is one in `zstreamtest` : https://github.com/facebook/zstd/blob/dev/tests/zstreamtest.c#L216

The skippable frame size is small (`11`), maybe it needs to be larger to trigger the issue ?

terrelln:
I've verified that the modified tests fail before the patch, and pass after.

PR #906, opened 2017-11-01
fix : ZSTD_compress_generic(,,,ZSTD_e_end) automatically sets pledgedSrcSize

Cyan4973:
as per documentation, on `ZSTD_setPledgedSrcSize()` :
> If all data is provided and consumed in a single round,
> this value (pledgedSrcSize) is overriden by srcSize instead.

This wasn't applied before compression level is transformed into compression parameters.
As a consequence, small input missed compression parameters adaptation.

It works fine now : compression was compared with `ZSTD_compress_advanced()`,
results were the same.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -286,7 +286,6 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
                     U32 nbLoops = 0;
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_nbThreads, g_nbThreads);
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionLevel, cLevel);
-                    //ZSTD_CCtx_setPledgedSrcSize(ctx, blockTable[0].srcSize);  /* estimation, only to help cLevel conversion before applying advanced compression parameters */
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_enableLongDistanceMatching, g_ldmFlag);
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_ldmMinMatch, g_ldmMinMatch);
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_ldmHashLog, g_ldmHashLog);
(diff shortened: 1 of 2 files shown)


programs/bench.c
```diff
@@ -286,6 +286,7 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
                     U32 nbLoops = 0;
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_nbThreads, g_nbThreads);
                     ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionLevel, cLevel);
+                    //ZSTD_CCtx_setPledgedSrcSize(ctx, blockTable[0].srcSize);  /* estimation, only to help cLevel conversion before applying advanced compression parameters */
```
terrelln: Some compiler won't like the `//` comments right?

Cyan4973: Yes, some old ones since `//` is not part of `C90`

Cyan4973: The choice here is between suppressing the line and enabling it.

Enabling the line should improve compression adaptation, notably for small files, when dictionary or advanced parameters are used.

But I'm concerned that this adaptation behaviour could be "specific" to benchmark mode, hence misleading.

The behavior should be replicated on the command line side.
As this is a different topic, I delayed the decision with the `//` comment.
It requires reading and potentially modifying `fileio.c`, so a bit more work.
It should be part of another patch.

If `//` is problematic, I can suppress the line in this patch.

terrelln: IIRC Appveyor tests will reject `//`, but I don't remember if it is MinGW or Visual.

Cyan4973: I believe it requires a much older compiler to fail the `//` comment test.

One way to fail it is to compile with `gcc` and `-std=c90 -pedantic`.
But in this case, a ton of other things fail too (typically, this setup does not support `unsigned long long` type).

Cyan4973:
Removed the `//` line

programs/bench.c
terrelln: I must be misremembering the failure then.

PR #908, opened 2017-11-06
Modified one pointer arithmetic expression to a more conformant way.

Cyan4973:


lib/compress/zstdmt_compress.c
```diff
@@ -142,36 +142,42 @@ static size_t ZSTDMT_sizeof_bufferPool(ZSTDMT_bufferPool* bufPool)
 
 static void ZSTDMT_setBufferSize(ZSTDMT_bufferPool* bufPool, size_t bSize)
 {
+    ZSTD_pthread_mutex_lock(&bufPool->poolMutex);
```
terrelln: I'm okay with acquiring the lock, but it isn't currently necessary right?

Cyan4973: Yes, right.
It's a bit overcautious, but should not impact performance.

I briefly suspected `bSize` initialisation as a possible source for a bug I was hunting, adding the lock helped removing that possibility.

PR #912, opened 2017-11-10
Optimal parser refactoring

Cyan4973:
This patch doesn't change functionality nor performance.
Its objective is mostly to rewrite the optimal parser in a way which makes it easier to understand and modify. "Aggressive scoping" is the most relevant rule applied in this patch.

On the noticeable changes : 
- `zstd_compress.h` becomes `zstd_compress_internal.h`, to emphasise that its definitions are strictly internal, meant to be used solely by `/lib/compress` modules.
- Many definitions were moved from `zstd_internal.h` into `zstd_compress_internal.h`. `zstd_internal.h` is reserved for definitions that need to be shared across `/lib` sub-directories, namely `/compress`, `/decompress` and `/dictBuilder`.

The next important change will be the modification of `zstd_opt`'s match finder to transparently manage rep codes, so that it's possible to have a single parser for both single-segment and double-segments memory configuration.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -8,8 +8,9 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
+#include "zstd_compress_internal.h"
 #include "zstd_opt.h"
-#include "zstd_lazy.h"
+#include "zstd_lazy.h"   /* ZSTD_updateTree, ZSTD_updateTree_extDict */
 
 
 #define ZSTD_LITFREQ_ADD    2
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -100,7 +100,7 @@ static U32 ZSTD_getLiteralPrice(optState_t* optPtr, U32 litLength, const BYTE* l
     U32 price;
 
     if (optPtr->staticPrices)
-        return ZSTD_highbit32((U32)litLength+1) + (litLength*6);  /* 6 bit per literal - no real estimation */
+        return ZSTD_highbit32((U32)litLength+1) + (litLength*6);  /* 6 bit per literal - no statistic used */
 
     if (litLength == 0)
         return optPtr->log2litLengthSum - ZSTD_highbit32(optPtr->litLengthFreq[0]+1);
@@ -414,8 +414,8 @@ FORCE_INLINE_TEMPLATE
 size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
                                       const void* src, size_t srcSize, const int ultra)
 {
-    seqStore_t* seqStorePtr = &(ctx->seqStore);
-    optState_t* optStatePtr = &(ctx->optState);
+    seqStore_t* const seqStorePtr = &(ctx->seqStore);
+    optState_t* const optStatePtr = &(ctx->optState);
     const BYTE* const istart = (const BYTE*)src;
     const BYTE* ip = istart;
     const BYTE* anchor = istart;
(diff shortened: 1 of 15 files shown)


lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
-                if (mlen > last_pos || price < opt[mlen].price)
-                    SET_PRICE(mlen, mlen, matches[u].off, litlen, price);   /* note : macro modifies last_pos */
-                mlen++;
-        }   }
+            /* set prices for first matches from position == 0 */
+            {   U32 matchNb;
+                U32 pos = last_pos /*some repCode (assumed cheaper)*/ ? last_pos : minMatch;
+                for (matchNb = 0; matchNb < nb_matches; matchNb++) {
+                    U32 const end = matches[matchNb].len;
+                    while (pos <= end) {
```
terrelln: We are losing some optimality here, since matches aren't fully sorted by length (also different offsets). I guess we should leave it for another diff, since this one should have no functionality changes.

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
                     price = ZSTD_getLiteralPrice(optStatePtr, litlen, anchor);
-           } else {
-                litlen = 1;
-                price = opt[cur - 1].price + ZSTD_getLiteralPrice(optStatePtr, litlen, inr-1);
-           }
-
-           if (cur > last_pos || price <= opt[cur].price)
-                SET_PRICE(cur, 1, 0, litlen, price);
+                }
+                if (cur > last_pos || price <= opt[cur].price)
```
terrelln: All the previous checks have been `price < opt[cur].price`, is there a reason for this to be different? If so can you add a comment.

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
-
-                       if (mlen > sufficient_len || cur + mlen >= ZSTD_OPT_NUM) {
-                            best_mlen = mlen; best_off = i; last_pos = cur + 1;
-                            goto _storeSequence;
-                       }
+            {   U32 const ll0 = (opt[cur].mlen != 1);
+                U32 const lastR = ZSTD_REP_CHECK + ll0;
+                U32 repCode4;  /* universal referential */
+                for (repCode4=ll0; repCode4<lastR; repCode4++) {  /* check rep */
+                    const S32 repCur = (repCode4==ZSTD_REP_MOVE_OPT) ? (opt[cur].rep[0] - 1) : opt[cur].rep[repCode4];
```
terrelln: You renamed this to `repOffset` above, do you want to do the same here?

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
-                const S32 repCur = (i==ZSTD_REP_MOVE_OPT) ? (rep[0] - 1) : rep[i];
-                if ( (repCur > 0) && (repCur < (S32)(ip-prefixStart))
-                    && (ZSTD_readMINMATCH(ip, minMatch) == ZSTD_readMINMATCH(ip - repCur, minMatch))) {
-                    mlen = (U32)ZSTD_count(ip+minMatch, ip+minMatch-repCur, iend) + minMatch;
-                    if (mlen > sufficient_len || mlen >= ZSTD_OPT_NUM) {
-                        best_mlen = mlen; best_off = i; cur = 0; last_pos = 1;
-                        goto _storeSequence;
+        {   U32 const ll0 = !initLL;
+            U32 const lastR = ZSTD_REP_CHECK + ll0;
+            U32 repCode;
```
terrelln: Can you use the names `repCode4` and `repCode3` here as well?

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
+                        assert(repCode3 < 3);
+
+                        if (matchLength > sufficient_len || cur + matchLength >= ZSTD_OPT_NUM) {
+                            best_mlen = matchLength;
+                            best_off = repCode4;
+                            last_pos = cur + 1;
+                            goto _shortestPath;
+                        }
 
-                       best_off = i - (opt[cur].mlen != 1);
```
terrelln: Why can you delete this `best_off` store? I see that you use `repCode3` below when getting/setting the price, but does anyone else depend on this store?

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
+                    goto _shortestPath;
+                }
 
-            /* set prices using matches at position = cur */
-            for (u = 0; u < match_num; u++) {
-                mlen = (u>0) ? matches[u-1].len+1 : best_mlen;
-                best_mlen = matches[u].len;
+                /* set prices using matches at position = cur */
+                for (matchNb = 0; matchNb < nb_matches; matchNb++) {
+                    U32 mlen = (matchNb>0) ? matches[matchNb-1].len+1 : best_mlen;
```
terrelln: You renamed this to `matchLength` elsewhere, can you do the same here?

lib/compress/zstd_opt.c
terrelln: I believe no one else reads it, but I just want to call it out.

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
+                    best_mlen = matches[nb_matches-1].len;
+                    best_off = matches[nb_matches-1].off;
+                    last_pos = cur + 1;
+                    goto _shortestPath;
+                }
 
-            /* set prices using matches at position = cur */
-            for (u = 0; u < match_num; u++) {
-                mlen = (u>0) ? matches[u-1].len+1 : best_mlen;
-                best_mlen = matches[u].len;
```
terrelln: Same with deleting this store here as with `best_off`.

lib/compress/zstd_opt.c
```diff
@@ -423,237 +424,241 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
-                offset--;
-            } else {
-                if (offset != 0) {
-                    best_off = (offset==ZSTD_REP_MOVE_OPT) ? (rep[0] - 1) : (rep[offset]);
-                    if (offset != 1) rep[2] = rep[1];
+        /* save sequences */
+        {   U32 pos;
+            for (pos=0; pos < last_pos; ) {
+                U32 const litLength = (U32)(ip - anchor);
+                U32 const mlen = opt[pos].mlen;
```
terrelln: Can you rename this to `matchLength` like above?

lib/compress/zstd_opt.c
Cyan4973: yeah, that would make sense, 
actually, this part completely disappears in next patch (including the `repCode4` / `repCode3` logic, which I found inconvenient).
All repcode-related source code is rewritten.

lib/compress/zstd_opt.c
Cyan4973: Not sure to follow this one.
Matches are sorted by the matchfinder.
They should be sorted in both gradually longer length and longer offset.

In next patch, since repcodes and normal offsets match search are fused, normal offsets match lengths are necessarily larger than repcode ones.

lib/compress/zstd_opt.c
Cyan4973: Good point !
Not sure about this one.
It's not a normal match here, it's a literal replacement.
So this `<=` favor literals over matches.

Not sure of the impact.
First, equality is less common in `zstd` compared to `lz4`, as price is expressed in bits, instead of bytes.
Second, all costs are "evaluated", which basically means they are wrong by a small amount. I noticed in the code many heuristics that try to compensate for that in bizarre ways. That could be one of them.

A simple answer to this `<` vs `<=` is : try both, keep the better one.

lib/compress/zstd_opt.c
Cyan4973: That's a relevant point.
Thing is, this code will disappear in next update.

lib/compress/zstd_opt.c
Cyan4973: `best_off` was improperly used as a temporary variable to store the same value as `repCode3`.
It was not necessary, as the scope of the storage does not outlive this inner loop, hence it was replaced by a `repCode`, which has a narrower, and more accurate, scope.

lib/compress/zstd_opt.c
Cyan4973: same idea : `best_mlen` was used as a temporary store, confusing the lifetime of its value. Replaced by scope-local `lastML`.

lib/compress/zstd_opt.c
Cyan4973: Good point.
I prefer `mlen`, but I had to juggle a bit with variable names during transition (initially, `mlen` was a function-level variable, used in several places for different unrelated usages).
I'll ensure consistency of these names.

lib/compress/zstd_opt.c
Cyan4973: I believe they are still sorted.

For `minmatch==3`, it seems there is only a single attempt, using a dedicated hash table. So there can be only a single match, though its length might be accidentally larger than `3`.
Either way, such match modifies `bestLength`, which means future matches will only be added to the list if they are [larger than `bestLength`](https://github.com/facebook/zstd/blob/72c0f4574cfef1d5d5f35fc7277ee353793956ab/lib/compress/zstd_opt.c#L318).

So I believe they are still sorted in both gradually larger length and offset.

PR #916, opened 2017-11-15
Optimal parser part 2 : Merge normal and _extDict parsers

Cyan4973:
There is now a single optimal parser for both normal and _extDict modes.

The enabler was to isolate _extDict specificities into the match finder.

The new parser code is highly simplified compared to earlier versions.
This should help reviews, and future modifications.

This new version does not compress identically compared with `dev` nor `v1.3.1`,
but it's within the same range. Speed is about the same. Compression ratio is similar, though exact compressed size vary.

```
calgary.tar : -b19 :  903756 (v1.3.1) ==>  903635 (patch)
enwik7      : -b19 : 2830032 (v1.3.1) ==> 2830756 (patch)
```

The difference is likely because search used to stop immediately after repcode when finding `sufficient_len`, while it will continue searching now. Searching more is supposed to be better, or so I believed, but we have here the proof it's not always the case. This is likely because statistics are only roughly evaluated, hence parsing decisions are based on flawed data.

The next modification will attempt to keep the table clean after each match insertion.
Currently, it's not, and rely on the mechanical study at each position to fix left over.
This prevent ability to skip positions, and flexible insertions.
Both capabilities are important to consider writing faster variants.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -264,14 +264,10 @@ static U32 ZSTD_insertBtAndGetAllMatches (
     size_t bestLength = minMatchLen-1;
 
 #if defined(ZSTD_DEBUG) && (ZSTD_DEBUG >= 8)
-    static const BYTE* g_start = NULL;
-    if (g_start==NULL) g_start = base;  /* note : index only works for compression within a single segment */
-    {   U32 const pos = (U32)(ip - g_start);
-        g_debuglog_enable = (pos==3673728);
-    }
-    if (current == 8793162) g_debuglog_enable = 1;
+    g_debuglog_enable = (current ==  5202593);
+    //g_debuglog_enable = (current ==  8845622);
+    //if (current == 12193408) g_debuglog_enable = 1;
 #endif
-
 
     /* check repCode */
 #if 0
@@ -293,7 +289,7 @@ static U32 ZSTD_insertBtAndGetAllMatches (
                     matches[mnum].len = (U32)repLen;
                     mnum++;
                     if ( (repLen > ZSTD_OPT_NUM)
-                      || (ip+repLen == iLimit) ) {  /* best possible */
+                       | (ip+repLen == iLimit) ) {  /* best possible */
                         return mnum;
         }   }   }   }
     } else {   /* extDict */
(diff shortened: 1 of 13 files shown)


lib/compress/zstd_opt.c
```diff
@@ -263,47 +258,80 @@ static U32 ZSTD_insertBtAndGetAllMatches (
...
-    if (minMatch == 3) { /* HC3 match finder */
+    /* check repCode */
+    {   U32 const lastR = ZSTD_REP_NUM + ll0;
+        U32 repCode;
+        for (repCode = ll0; repCode < lastR; repCode++) {
+            U32 const repOffset = (repCode==ZSTD_REP_NUM) ? (rep[0] - 1) : rep[repCode];
+            U32 const repIndex = current - repOffset;
+            U32 repLen = 0;
+            assert(current >= dictLimit);
+            if (!extDict /*static*/ || (repIndex>=dictLimit)) {
```
terrelln: Can you add an assert here like you added earlier `assert(repIndex>=dictLimit)`? I think it will be generally useful, and will add clarity to unfamiliar readers.

Cyan4973: Actually, the next test `(repOffset-1 < current-dictLimit)` is equivalent to `current > repIndex >= dictLimit`. It's doing 2 tests in one comparison thanks to the overflow effect.

Which means, I don't need the preliminary `(repIndex>=dictLimit)`, since it's actually included in next test. I'll update the code (and comments) accordingly.

terrelln:
I was looking at `ZSTD_insertAndFindFirstIndexHash3()`, and wondering if `nextToUpdate3` can get out of date, for instance if there is always a repcode match of length >= 3, and fall into the `extDict`.

At the beginning of `ZSTD_compressBlock_opt_generic` we set `nextToUpdate3 = nextToUpdate`. This is necessary because `nextToUpdate3` isn't updated when we reduce tables (but does it hurt compression ratio?). I think this solves the problem, since `nextToUpdate` must not be in the `extDict`, do you agree?

Cyan4973:
Yes, I agree, due to initialisation at beginning of block, `nextToUpdate3` cannot be within `extDict`.

PR #917, opened 2017-11-16
When invoked as *cat, imply --keep, do NOT remove source files

allanjude:
if you 'zstdcat file.zstd', you expect the same behaviour as gzcat and friends, and zstdcat deletes the source file. 'cat' should not modify the source file, ever.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -421,10 +421,10 @@ int main(int argCount, const char* argv[])
     /* preset behaviors */
     if (exeNameMatch(programName, ZSTD_ZSTDMT)) nbThreads=0;
     if (exeNameMatch(programName, ZSTD_UNZSTD)) operation=zom_decompress;
-    if (exeNameMatch(programName, ZSTD_CAT)) { operation=zom_decompress; forceStdout=1; FIO_overwriteMode(); outFileName=stdoutmark; g_displayLevel=1; }
+    if (exeNameMatch(programName, ZSTD_CAT)) { operation=zom_decompress; FIO_setRemoveSrcFile(0); forceStdout=1; FIO_overwriteMode(); outFileName=stdoutmark; g_displayLevel=1; }
     if (exeNameMatch(programName, ZSTD_GZ)) { suffix = GZ_EXTENSION; FIO_setCompressionType(FIO_gzipCompression); FIO_setRemoveSrcFile(1); }    /* behave like gzip */
     if (exeNameMatch(programName, ZSTD_GUNZIP)) { operation=zom_decompress; FIO_setRemoveSrcFile(1); }                                          /* behave like gunzip */
-    if (exeNameMatch(programName, ZSTD_GZCAT)) { operation=zom_decompress; forceStdout=1; FIO_overwriteMode(); outFileName=stdoutmark; g_displayLevel=1; }  /* behave like gzcat */
+    if (exeNameMatch(programName, ZSTD_GZCAT)) { operation=zom_decompress; FIO_setRemoveSrcFile(0); forceStdout=1; FIO_overwriteMode(); outFileName=stdoutmark; g_displayLevel=1; }  /* behave like gzcat */
     if (exeNameMatch(programName, ZSTD_LZMA)) { suffix = LZMA_EXTENSION; FIO_setCompressionType(FIO_lzmaCompression); FIO_setRemoveSrcFile(1); }    /* behave like lzma */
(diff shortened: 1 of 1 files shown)


terrelln:
Unlike `gzip`, `zstd` keeps the source file by default. `zstdcat file.zst` should only remove the source file if the `--rm` flag was passed. In that case I don't think your patch will work, because `FIO_setRemoveSrcFile(1)` gets called after you set it to `0`.

I'm not sure if we want to disallow `--rm` for the `*cat` programs, but if we do, it will have to be [here](https://github.com/facebook/zstd/blob/dev/programs/zstdcli.c#L478). It doesn't sound unreasonable to me. What do you think @Cyan4973?

cemeyer:
@terrelln Respectfully, empirically that is not true:

```
$ zstd --version
*** zstd command line interface 64-bits v1.3.2, by Yann Collet ***
$ zstdcat --version
*** zstd command line interface 64-bits v1.3.2, by Yann Collet ***

$ chmod ugo-w testfile
$ zstd -6 ./testfile
$ chmod ugo-w testfile
chmod: testfile: No such file or directory
$ chmod ugo-w testfile.zst
$ zstdcat testfile.zst >/dev/null
$ ls testfile.zst
ls: testfile.zst: No such file or directory
```

Truss shows:
```
$ echo "abc"  |zstd -6 > ./testfile.zstd
$ sudo chown root:wheel ./testfile.zstd
$ sudo chmod ugo-w ./testfile.zstd
$ sudo chmod ugo+r ./testfile.zstd
$ ls -l ./testfile.zstd
-r--r--r--  1 root  wheel  17 Nov 15 17:24 ./testfile.zstd
$ truss zstdcat ./testfile.zstd
...
fstatat(AT_FDCWD,"./testfile.zstd",{ mode=-r--r--r-- ,inode=35232595,size=17,blksize=32768 },AT_SYMLINK_NOFOLLOW) = 0 (0x0)
unlink("./testfile.zstd")			 = 0 (0x0)
```

Clearly `zstdcat` is unlinking source files.

terrelln:
@cemeyer Are using zstd-1.3.2 from the master branch? If not, which revision are you testing on? I tested with zstd-1.3.2 before responding, but wasn't able to reproduce it. What system are you testing on?

My test was:

```
echo hello | zstd > hello.zst
zstdcat hello.zst
zstdcat hello.zst > /dev/null
ls hello
# still there
```

I'll look into it more.

Cyan4973:
I agree that `zstdcat` should not remove source file.

And I'm actually surprised it does.
I made a quick verification on my system, it doesn't erase source file.
So I guess there's something more.

cemeyer:
This is `zstdcat` etc as integrated on FreeBSD CURRENT: https://github.com/freebsd/freebsd/tree/master/sys/contrib/zstd

Built like so: 

https://github.com/freebsd/freebsd/blob/master/lib/libzstd/Makefile

https://github.com/freebsd/freebsd/blob/master/usr.bin/zstd/Makefile

@bapt Updated to 1.3.2 here: https://github.com/freebsd/freebsd/commit/41e12ee91f1de0c5c4eb8a3066a3c06229dcccb8#diff-5390a031dc4b49debf00ae8512c0f92d

cemeyer:
I'd like to add — thanks for your patience in helping us track this down.  It seems like the mistake was on the FreeBSD side.

terrelln:
The freebsd version of zstd defaults `g_removeSrcFile` to 1 https://github.com/freebsd/freebsd/blob/master/sys/contrib/zstd/programs/fileio.c#L206, where it normally defaults to 0 https://github.com/facebook/zstd/blob/dev/programs/fileio.c#L191. So your proposed fix works for the patched version.

Would you like us to add a macro like `ZSTD_REMOVE_SOURCE_DEFAULT` that defaults to `0`, and can be overridden on the command line? That way we can support this feature and add tests. The only thing I'm worried about is inconsistent behavior of the zstd tool across platforms.

cemeyer:
I think we just want to match upstream zstd behavior :-).  I'll go ahead and try to figure out what we need to do to get onto actual 1.3.2.

terrelln:
Great!

cemeyer:
Some left hand not talking to the right, unfortunately.  For anyone following along with historical interest, @bapt introduced this in FreeBSD r316980.  [I've asked for a revert of that change](https://lists.freebsd.org/pipermail/svn-src-all/2017-November/153819.html).

allanjude:
My fault again, sorry for the noise. I really am just trying to be helpful ;)

PR #919, opened 2017-11-17
Change default behaviour to match gzip(1), bzip2(1), xz(1)

bapt:


bapt:
The intent of the pull request is to be able to start a discussion about the default behaviour of zstd.

I think it would be a good idea that the default behaviour is the same as gzip, bzip2 or xz counter part. This would allow zstd to be a dropin replacement (some tools could use it without intrusive modifications.)

The modifications needed are pretty simple: the verbosity level has to be reduced and the default behaviour should be to remove the source for all cases but zstdcat.

An example on freebsd that depends on that behaviour to not have to change too much the program, is our newsyslog (log rotation program).

What do you think?

bapt:
I do personally prefer the current default behaviour, my point was mainly to allow it to be a dropped in replacement for tools already aware of gzip(1) xz(1) etc. I will adapt programs in FreeBSD for that purpose

Cyan4973:
What do you think of the idea of having a dedicated link name which automatically trigger `--rm -q` ? Would it help, or would it be superfluous ?

allanjude:
The dedicated name was also my suggestion to bapt, something like 'zzip' and 'zunzip' that are gzip/xz like, and keep zstd doing what it does now

Cyan4973:
Sure, whatever works best for FreeBSD !

I guess `zzip` naming suggestion comes from swapping the first letter within `gzip`.
It works. The only minor downside I can think of is that the name seems to imply [`zip` as an archive format](https://en.wikipedia.org/wiki/Zip_(file_format)), which `zstd` is not (neither is `gzip`).

A relative to this naming suggestion could be `zz`,`unzz`, which mimics `xz` naming scheme (though it might sound a bit sleepy...)

PR #924, opened 2017-11-20
faster btopt variant

Cyan4973:
`zstd_opt.c` now supports 3 variants, through `optLevel` parameter.
- btultra -> (2)
- old btopt -> (1)
- new btopt -> (0)

btopt(0) is about +15% faster compared to btopt(1), at the cost of -0.02 ratio. This is merely an "average" : variance can be large depending on file content. Direction though is always the same.

I initially considered exposing the 3 different variants, by my feeling is that the performance gap between (0) and (1) is too small to deserve creating a new strategy. So this patch makes `btopt` adopt the faster setting.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -240,9 +240,9 @@ FORCE_INLINE_TEMPLATE
 U32 ZSTD_insertBtAndGetAllMatches (
                         ZSTD_CCtx* zc,
                         const BYTE* const ip, const BYTE* const iLimit, const int extDict,
-                        U32 nbCompares, U32 const mls,
+                        U32 nbCompares, U32 const mls, U32 const sufficient_len,
                         U32 rep[ZSTD_REP_NUM], U32 const ll0,
-                        ZSTD_match_t* matches, const U32 minMatchLen)
+                        ZSTD_match_t* matches, const U32 lengthToBeat)
 {
     const BYTE* const base = zc->base;
     U32 const current = (U32)(ip-base);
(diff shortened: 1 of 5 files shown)


lib/compress/zstd_opt.c
```diff
@@ -646,10 +656,11 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
                     }
                 }
 
-                ZSTD_updatePrice(optStatePtr, llen, anchor, offset, mlen);
+                ZSTD_updateStats(optStatePtr, llen, anchor, offset, mlen);
                 ZSTD_storeSeq(seqStorePtr, llen, anchor, offset, mlen-MINMATCH);
                 anchor = ip;
         }   }
+        ZSTD_setLog2Prices(optStatePtr);
     }   /* for (cur=0; cur < last_pos; ) */
```
terrelln: This comment isn't accurate, right?

Cyan4973: correct ! fixed

Cyan4973:
> I want to confirm where the extra speed is coming from
Yes, correct for both

lib/compress/zstd_opt.c
```diff
@@ -646,11 +657,12 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
                 }
 
-                ZSTD_updatePrice(optStatePtr, llen, anchor, offset, mlen);
+                ZSTD_updateStats(optStatePtr, llen, anchor, offset, mlen);
                 ZSTD_storeSeq(seqStorePtr, llen, anchor, offset, mlen-MINMATCH);
                 anchor = ip;
         }   }
-    }   /* for (cur=0; cur < last_pos; ) */
+        ZSTD_setLog2Prices(optStatePtr);
+    }   while (ip < ilimit)
```
gyscos: Is this supposed to be a comment? Currently it seems to apply to the next block after the comment, which may be confusing?

Cyan4973: great catch @gyscos !
Yes, it's supposed to be a comment ! fixed

Issue #927, opened 2017-11-20
pkgconfig file should go to $LIBDIR/pkgconfig


rworkman:
The pkg-config file has arch-specific linker flags and thus should be in the arch-specific pkgconfig directory

Cyan4973:
- Which arch-specific linker flag is it ?
- Installation directory for `libzstd.pc` is defined here : https://github.com/facebook/zstd/blob/dev/lib/Makefile#L137. Its definition depends on target operating system. It is `$(LIBDIR)/pkgconfig` for non-BSD systems.

rworkman:
I guess it's not fair to say that it matters right now, but this:  'Libs: -L${libdir} -lzstd' 
I'll concede that my understanding may be flawed, but it was/is essentially this: if anything in the .pc file will have an arch-specific path in it, then it should be installed to $libdir/pkgconfig -- this would make it so a multilib system could have e.g. /usr/lib/pkgconfig/libzstd.pc and a /usr/lib64/pkgconfig/libzstd.pc which would not conflict with each other (whether that's a good idea is another discussion :))

Issue #926, opened 2017-11-20
Build: libraries install to /usr/lib on x86_64


rworkman:
Please consider adding a cmake flag for e.g. -DLIBDIR:STRING="/usr/lib64"

Cyan4973:
patch welcomed !

krab:
Was fixed here https://github.com/facebook/zstd/pull/990 now installs to correct lib64 on x86_64, and directory always can be changed with `-DCMAKE_INSTALL_LIBDIR=my_awesome_64bit_dir` relative to `-DCMAKE_INSTALL_PREFIX=/usr`

PR #929, opened 2017-11-21
Build system: support alternate LIBDIR, MANDIR, DOCDIR

rworkman:
This addresses #926 and #927


programs/Makefile
```diff
@@ -262,9 +262,7 @@ mandir      ?= $(datarootdir)/man
 man1dir     ?= $(mandir)/man1
 
 ifneq (,$(filter $(shell uname),OpenBSD FreeBSD NetBSD DragonFly SunOS))
-MANDIR  ?= $(PREFIX)/man/man1
-else
-MANDIR  ?= $(man1dir)
+MANDIR  ?= $(PREFIX)/man
```
Cyan4973: This construction prevents using `mandir` nor `man1dir` as an input of this Makefile.
These variables are recommended by the [GNU Makefile coding standards](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables).

It also changes the destination directory for non-BSD targets, 
from `$(prefix)/share/man/man1` to `$(prefix)/man/man1`.

programs/Makefile
```diff
@@ -281,17 +279,17 @@ INSTALL_MAN     ?= $(INSTALL_DATA)
 .PHONY: install
 install: zstd
 	@echo Installing binaries
-	@$(INSTALL) -d -m 755 $(DESTDIR)$(BINDIR)/ $(DESTDIR)$(MANDIR)/
+	@$(INSTALL) -d -m 755 $(DESTDIR)$(BINDIR)/ $(DESTDIR)$(MANDIR)/man1
```
Cyan4973: Is there an issue inserting `/man1` directly into `$MANDIR` ?
Would you prefer such variable to be called `$MAN1DIR` instead ?

build/cmake/programs/CMakeLists.txt
```diff
@@ -45,9 +45,9 @@ IF (UNIX)
         COMMENT "Copying manpage zstd.1")
     ADD_CUSTOM_TARGET(zstdcat.1 ALL ${CMAKE_COMMAND} -E create_symlink zstd.1 zstdcat.1 DEPENDS zstd.1 COMMENT "Creating zstdcat.1 symlink")
     ADD_CUSTOM_TARGET(unzstd.1 ALL ${CMAKE_COMMAND} -E create_symlink zstd.1 unzstd.1 DEPENDS zstd.1 COMMENT "Creating unzstd.1 symlink")
-    INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/zstd.1 DESTINATION "share/man/man1")
-    INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/zstdcat.1 DESTINATION "share/man/man1")
-    INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/unzstd.1 DESTINATION "share/man/man1")
+    INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/zstd.1 DESTINATION "${CMAKE_INSTALL_MANDIR}/man1")
```
Cyan4973: why removing `share` from the man install path ?

Makefile
```diff
@@ -14,6 +14,11 @@ ZWRAPDIR = zlibWrapper
 TESTDIR  = tests
 FUZZDIR  = $(TESTDIR)/fuzz
 
+# Define some install locations
```
Cyan4973: Why defining `cmake` variables within the root `Makefile` ?

Cyan4973: If these variables are only used by `cmake`,
maybe it's better for code maintenance to define them close to the place where they are used.

Makefile
```diff
@@ -14,6 +14,11 @@ ZWRAPDIR = zlibWrapper
...
+LIBDIR   = $(CMAKE_INSTALL_PREFIX)/lib
+MANDIR   = $(CMAKE_INSTALL_PREFIX)/share/man
```
Cyan4973: My understanding is that `MANDIR` is different depending on Linux / BSD targets.
Here it seems there is no distinction.

build/cmake/programs/CMakeLists.txt
```diff
@@ -45,9 +45,9 @@ IF (UNIX)
...
+    INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/zstdcat.1 DESTINATION "${CMAKE_INSTALL_MANDIR}/man/man1")
```
Cyan4973: I guess you meant `"${CMAKE_INSTALL_MANDIR}/man1"` ?

Cyan4973:
How do you test the result of your installation script ?

programs/Makefile
rworkman: Yeah, this was unintentional.

programs/Makefile
rworkman: Disclaimer: this is my first time messing with a cmake-ish (but it's really not even that) build system; from what I recall from messing with autotools and even much older Makefiles, I've never seen MANDIR have the man? subdir in it -- it's always been $(MANDIR)/man?/

build/cmake/programs/CMakeLists.txt
rworkman: Indeed.

rworkman:
The build process here is essentially this:

    patch -p1 < $CWD/0001-Build-system-support-alternate-LIBDIR-MANDIR-DOCDIR.patch
    
    make \
      PREFIX=/usr \
      LIBDIR=/usr/lib${LIBDIRSUFFIX} \
      MANDIR=/usr/man \
      DOCDIR=/usr/doc/$PKGNAM-$VERSION
    
    make install \
      PREFIX=/usr \
      LIBDIR=/usr/lib${LIBDIRSUFFIX} \
      MANDIR=/usr/man \
      DOCDIR=/usr/doc/$PKGNAM-$VERSION \
      DESTDIR=$PKG

About to push a new commit. I'm not convinced that it addresses the comment about where the variables are defined, but maybe it's acceptable now since they aren't specific to cmake any more.

rworkman:
This iteration drops DOCDIR (it wasn't used - I had planned to use it, but meh - too much trouble), add PREFIX, and addresses the review points, I think.

Cyan4973:
I'm not comfortable applying this patch.
I do believe it does improve situation on your system,
but its impact seems too wide and affect different systems in hard to guess ways.

Please note that creating and maintaining a build/install script for a large number of systems is a fairly difficult exercise. While nothing is perfect, it can certainly be improved. When visibility is poor, it's important to progress cautiously by tiny steps, with limited / manageable scope.

I would be ok to start with a patch impacting only `cmake` to solve #926.

Issue #930, opened 2017-11-22
Need clarification on the benchmarking


scottchiefbaker:
I apologize for a silly question, but can someone explain what the output of the benchmark stuff is:

```
 8#test.bin          :1872691200 -> 872584382 (2.146),  22.3 MB/s ,1166.8 MB/s
```

I looked in the man page, and documentation and couldn't figure out what each field was. My best guess is:

compression level#filename     input size -> compress_size (compression ratio), compression speed, decompression speed? Assuming that last one is decompression speed is the entire input compressed and then decompressed to measure that speed or how does that work?

Cyan4973:
Yes, you are correct in your interpretation of the fields.
Indeed, it would probably be better to document it in the `man` page.

For both compression and decompression speed, the entire input is compressed / decompressed in-memory to measure speed. A compression / decompression run lasts at least 1 sec, so when files are small, they are compressed / decompressed several times per run, in order to improve measurement accuracy.  Your test file is ~1.8GB though, which is pretty large, so it most likely needs only one decompression per run.

scottchiefbaker:
@Cyan4973 I've been looking for an excuse to submit a PR. Would you be willing to take a PR to update the man page with that information.

Cyan4973:
👍  sure !

scottchiefbaker:
Should I be sending a PR for `programs/zstd.1` or for `programs/zstd.1.md`? I'm guessing the markdown one is generated from the raw man page?

Cyan4973:
It's the other way round :)
We create `zstd.1` from `zstd.1.md`, which is much easier to edit

PR #931, opened 2017-11-23
Include information about the benchmark output/methodology

scottchiefbaker:
Addresses #930


Cyan4973:
Just for your information,
it's possible to generate the `man.1` man page from `man.1.md` markdown source by using `make man` in `/programs`. This command needs a few dependencies to succeed though.

PR #934, opened 2017-11-29
 zstd_opt: changed cost formula

Cyan4973:
There was a flaw in the formula
which compared literal cost with match cost at a given position :

a non-null literal suite is going to be part of next sequence,
while if position ends a previous match, to immediately start another match,
next sequence will have a litlength of zero.
A litlength of zero has a non-null cost.
It follows that literals cost should be compared to match cost + litlength==0.

Not doing so gave a structural advantage to matches, which would be selected more often.
I believe that's what led to the creation of the strange `factor` heuristic which adds a fixed (but complex to calculate) cost to each match.
The heuristic was actually compensating.
It was probably created through multiple trials, settling for best outcome on a given scenario (I suspect `silesia.tar`), which means it's likely compensating for several unidentified side effects.
The problem with this heuristic is that it's hard to understand,
and unfortunately, any change in the parser impacts the way it should be calculated and its effects.

Using the "proper" cost formula makes it possible to remove `factor` heuristic, since the bias favoring matches was likely the strongest.

Unfortunately, in a head to head comparison, the new formula is sometimes better, sometimes worse.
Note that all differences are very small (< 0.01 ratio).
In general, the newer formula is better for smaller files (for example, `calgary.tar` and `enwik7`).
I suspect that's because starting statistics are poor.
However, for `silesia.tar` specifically, it's worse at level 22 (while being better at level 17, so even compression level changes comparison result ...).

It's a pity that `zstd -22` gets worse on `silesia.tar`.
That being said, I like that the new code gets rid of some strange variables,
which were introducing complexity for any future evolution (faster variants being in mind).
Therefore, in spite of this important detrimental side effect, I tend to be in favor of this patch.

The patch also introduces a few simplifications and changes which prepare its capability to accept different parsing strategies.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -32,10 +32,9 @@ static void ZSTD_setLog2Prices(optState_t* optPtr)
 }
 
 
-static void ZSTD_rescaleFreqs(optState_t* optPtr, const BYTE* src, size_t srcSize)
+static void ZSTD_rescaleFreqs(optState_t* const optPtr,
+                              const BYTE* const src, size_t const srcSize)
 {
-    optPtr->cachedLiterals = NULL;
-    optPtr->cachedPrice = optPtr->cachedLitLength = 0;
     optPtr->staticPrices = 0;
 
     if (optPtr->litLengthSum == 0) {  /* first init */
@@ -97,7 +96,7 @@ static void ZSTD_rescaleFreqs(optState_t* optPtr, const BYTE* src, size_t srcSiz
 
 
 static U32 ZSTD_rawLiteralsCost(const optState_t* const optPtr,
-                                 const BYTE* const literals, U32 const litLength)
+                                const BYTE* const literals, U32 const litLength)
 {
     if (optPtr->staticPrices) return (litLength*6);  /* 6 bit per literal - no statistic used */
     if (litLength == 0) return 0;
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_opt.c
```diff
@@ -161,13 +171,14 @@ FORCE_INLINE_TEMPLATE U32 ZSTD_getPrice(optState_t* optPtr,
         price += ML_bits[mlCode] + optPtr->log2matchLengthSum - ZSTD_highbit32(optPtr->matchLengthFreq[mlCode]+1);
     }
 
-    price += ZSTD_getLiteralPrice(optPtr, litLength, literals) + optPtr->factor;
-    DEBUGLOG(8, "ZSTD_getPrice(ll:%u, ml:%u) = %u", litLength, matchLength, price);
+    //price += optPtr->factor;
```
terrelln: Delete

lib/compress/zstd_opt.c
```diff
@@ -454,6 +465,59 @@ repcodes_t ZSTD_updateRep(U32 const rep[3], U32 const offset, U32 const ll0)
         memcpy(opt[pos].rep, &rep_, sizeof(rep_));            \
     }
 
+typedef struct {
+    const BYTE* anchor;
+    U32 litlen;
+    U32 rawLitCost;
+} cachedLiteralPrice_t;
+
+static U32 ZSTD_rawLiteralsCost_cached(
```
terrelln: Does caching this actually make a difference? I saw that in silesia it triggers < 1% of the time in the dev branch.

lib/compress/zstd_opt.c
```diff
@@ -96,53 +89,70 @@ static void ZSTD_rescaleFreqs(optState_t* optPtr, const BYTE* src, size_t srcSiz
...
+#if 1
+        return contribution;
+#else
+        return MAX(0, contribution); /* sometimes better, sometimes not ... */
+#endif
     }
+}
 
-    return price;
+static int ZSTD_literalsContribution(const BYTE* const literals, U32 const litLength,
```
terrelln: Can you add documentation to each of these functions? Especially `ZSTD_litLengthContribution()`.

lib/compress/zstd_opt.c
```diff
@@ -96,53 +89,70 @@ static void ZSTD_rescaleFreqs(optState_t* optPtr, const BYTE* src, size_t srcSiz
...
+}
+
+static U32 ZSTD_fullLiteralsCost(const BYTE* const literals, U32 const litLength,
+                                 const optState_t* const optPtr)
+{
+    return ZSTD_rawLiteralsCost(literals, litLength, optPtr)
+         + ZSTD_litLengthPrice(litLength, optPtr);
+}
+
+static int ZSTD_litLengthContribution(U32 const litLength, const optState_t* const optPtr)
```
terrelln: If I understand correctly, this is the cost of encoding `litLength` literals, less the cost of encoding empty literals.

lib/compress/zstd_opt.c
```diff
@@ -541,16 +615,21 @@ size_t ZSTD_compressBlock_opt_generic(ZSTD_CCtx* ctx,
...
                 }
+                assert(price < 1000000000); /* overflow check */
                 if (price <= opt[cur].price) {
                     DEBUGLOG(7, "rPos:%u : better price (%u<%u) using literal",
                                 cur, price, opt[cur].price);
-                    SET_PRICE(cur, 1/*mlen*/, 0/*offset*/, litlen, price, opt[cur-1].rep);
+                    opt[cur].mlen = 1;
+                    opt[cur].off = 0;
+                    opt[cur].litlen = litlen;
+                    opt[cur].price = price;
```
terrelln: IIUC anything that ends in a literal, we store the price less the cost of empty literals, correct? That way it gets a fair comparison against anything that ends in a match.

lib/compress/zstd_opt.c
```diff
@@ -454,6 +465,59 @@ repcodes_t ZSTD_updateRep(U32 const rep[3], U32 const offset, U32 const ll0)
         memcpy(opt[pos].rep, &rep_, sizeof(rep_));            \
```
terrelln: Is `SET_PRICE()` used anywhere anymore, or can it be deleted?

lib/compress/zstd_opt.c
Cyan4973: Yes,
so that it can be directly compared to the cost of a match ending at the same position

lib/compress/zstd_opt.c
Cyan4973: It can be deleted.

lib/compress/zstd_opt.c
Cyan4973: As discussed, it likely matters for some corner cases, when `litlen` becomes very big, requiring a match found after a long section without any match. In which case, the literal section will be scanned several times (at least 3 times, more likely 4-5 times).
In most situations though, I expect the size of a literals section to remain small, so the speed difference would not be visible.

Observing the difference will likely require a data source that can produce such corner case.

lib/compress/zstd_opt.c
Cyan4973: correct

Cyan4973:
Changes applied

PR #936, opened 2017-11-30
Reduce memory stress for qemu tests

Cyan4973:
`zstreamtest --newapi` tends to fail on qemu-arm and qemu-ppc targets (32-bits).
I suspect that's because some of these tests require too much resource.
A complementary directive `--no-big-tests`, is supposed to avoid that.
However, the test still fails from time to time on travis CI.

Unfortunately, I've been unable to reproduce it, so I'm left speculating.
Considering the issue might be a too large resource requirement, I tried to find which case could trigger such requirement, and found that `ZSTD_p_enableLongDistanceMatching` could be enabled, and even associated with multithreading ! Combined with an undisclosed input length, it leads to the creation of several 256 MB buffers. That sounds like a possible reason.

This patch just turns off `ZSTD_p_enableLongDistanceMatching` tests when directive `--no-big-tests` is triggered.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -470,7 +470,7 @@ size_t ZSTD_CCtxParam_setParameter(
  * pledgedSrcSize is considered unknown
  */
 size_t ZSTD_CCtx_setParametersUsingCCtxParams(
-        ZSTD_CCtx* const cctx, const ZSTD_CCtx_params* const params)
+        ZSTD_CCtx* cctx, const ZSTD_CCtx_params* params)
 {
     if (cctx->streamStage != zcss_init) return ERROR(stage_wrong);
     if (cctx->cdict) return ERROR(stage_wrong);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -271,11 +271,11 @@ size_t ZSTD_CCtx_setParameter(ZSTD_CCtx* cctx, ZSTD_cParameter param, unsigned v
     case ZSTD_p_forceMaxWindow :  /* Force back-references to remain < windowSize,
                                    * even when referencing into Dictionary content
                                    * default : 0 when using a CDict, 1 when using a Prefix */
-        cctx->loadedDictEnd = 0;
+        cctx->loadedDictEnd = 0;  /* ? */
         return ZSTD_CCtxParam_setParameter(&cctx->requestedParams, param, value);
 
     case ZSTD_p_nbThreads:
-        if (value > 1 && cctx->staticSize) {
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -465,14 +467,17 @@ size_t ZSTD_CCtxParam_setParameter(
...
- * Pledged srcSize is treated as unknown.
+ * pledgedSrcSize is considered unknown
  */
 size_t ZSTD_CCtx_setParametersUsingCCtxParams(
         ZSTD_CCtx* cctx, const ZSTD_CCtx_params* params)
 {
     if (cctx->streamStage != zcss_init) return ERROR(stage_wrong);
     if (cctx->cdict) return ERROR(stage_wrong);
 
+#if 1
```
terrelln: Is this intentional, or accidentally left in from debugging?

Cyan4973: Indeed, I just changed this behavior, and kept the old code in, in case the new path would trigger some error in future tests (most likely in `zstreamtest --opaqueapi`).

I've seen no error so far, but was willing to wait a bit longer before removing the old path.

terrelln: I remember there being a subtle reason why it is there, but I don't remember exactly what it is. I'll take a deeper look.

terrelln: Not setting `loadedDictEnd` would nullify the effect of `ZSTD_p_forceMaxWindow` right?

Cyan4973:
Well, bad news, it hasn't totally solved the issue ...
(though it goes farther than previous test)
https://travis-ci.org/facebook/zstd/jobs/309282908#L773

Cyan4973:
removing `-ftrapv` seems to help

lib/compress/zstd_compress.c
Cyan4973: To be checked

Cyan4973: I believe setting `loadedDictEnd=0` has no practical impact.

The impact of `forceWindow` happens during dictionary loading, as it should : 
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1890
There, `loadedDictEnd` is set unconditionally, irrespective of its previous value.

Just to be on the safe side, since the main user of `forceWindow` is ZSTDMT, I tested it at different compression settings : compression result is strictly identical, at all levels, wether `loadedDictEnd` is set to `0` or not. Also ran fuzzer tests, no impact either.

As a consequence, changing `cctx->loadedDictEnd` will be removed. 
With this change, setting a compression parameter doesn't produce any side effect.

Cyan4973:
The last test failure is fixed in #939

PR #943, opened 2017-12-07
Fix #942

### LINKED ISSUE #942: ZSTD_compressStream always returns 1?
Just going by the example program streaming_compression.c, ZSTD_compressStream is called as follows:
`    toRead = ZSTD_compressStream(cstream, &output , &input);`
and then the next file read reads toRead number of bytes:
`    while( (read = fread_orDie(buffIn, toRead, fin)) ) {`
However, at some point between the last time I pulled zstd (probably almost a year at this point) and today, it seems like ZSTD_compressStream basically always returns 1. Which means we're reading 1 byte at a time. My compression speed dropped to a crawl as a result, since I was modelling after examples/streaming_compression.c. On a fresh compile on latest repo, streaming_compression.c seems to itself have this same problem. Or is there something I'm doing wrong here???

Cyan4973:
streaming interface does not compress after `ZSTD_initCStream()`

While the final result is still, technically, a frame,
the resulting frame expands initial data instead of compressing it.
This is because the `ZSTD_CStream` creates a tiny 1-byte buffer for input,
because it believes input is empty (0-bytes),
because it was told so by `ZSTD_initCStream()`,
because in the past `0` used to mean "unknown".

This patch fixes the issue and adds a test case.


terrelln:
I believe we also need a fix for `ZSTD_initCStream_srcSize()`, since the documentation says it accepts 0 as unknown for now.

Cyan4973:
Right, `ZSTD_initCStream()` relied on this behaviour, and that's why it failed.

I'll fix it so that `0` gets interpreted as "unknown" for the time being,
although the documentation shall make it clear that no one should rely on this behaviour anymore.

terrelln:
Should `ZSTD_initCStream_usingCDict_advanced()` set `pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN` when `pledgedSrcSize == 0 && fParams.contentSizeFlag == 0`, like `ZSTD_initCStream_advanced()`?

Other than that, the `ZSTD_initCStream_X()` functions look good.

Cyan4973:
Good point. I don't think it should.

The trick used in `ZSTD_initCStream_advanced()` is just to preserve behaviour of (potentially) existing applications that may rely on this mechanism. 
I'm not sure if there are users of `ZSTD_initCStream_advanced()` relying on `pledgedSrcSize == 0 && fParams.contentSizeFlag == 0` to mean "unknown", there might be none, but just in case, it is mapped to `ZSTD_CONTENTSIZE_UNKNOWN`.
Note that relying on this behaviour must be avoided, as it will be removed in the future.

In contrast, I don't expect anyone to use `ZSTD_initCStream_usingCDict_advanced()` with the assumption that `pledgedSrcSize == 0 && fParams.contentSizeFlag == 0` means "unknown". This function was introduced much later, within `v1.2.0`, at the same time as constant `ZSTD_CONTENTSIZE_UNKNOWN`. So better enforce the new behaviour there.

What I can do though is to update documentation to make this point very clear.

terrelln:
folly currently depends on that behavior of `ZSTD_initCStream_advanced()`, but it can be fixed.

PR #945, opened 2017-12-13
Fix cdict compressor repcodes

terrelln:
The compressor wasn't using the repcodes from the cdict. If the repcodes aren't the default, which is never the case right now, and the repcode is used, the data will be corrupt.


Cyan4973:
Great find !

Cyan4973:
This patch opens new issues for PPC platforms.
It's probably an endian-related issue.

terrelln:
The test was using `MEM_read32()` instead of `MEM_readLE32()` when searching for the repcodes in the dictionary. Fixed now.

PR #946, opened 2017-12-13
Allow -o with multiple files

terrelln:
Allow `-o` with multiple files, which will work like `-c`.

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -173,6 +173,22 @@ rm tmp
 $ZSTD -f tmp && die "tmp not present : should have failed"
 test ! -f tmp.zst  # tmp.zst should not be created
 
+$ECHO "test : compress multiple files"
+$ECHO hello > tmp1
+$ECHO world > tmp2
+$ZSTD tmp1 tmp2 -o "$INTOVOID"
+$ZSTD tmp1 tmp2 -c | $ZSTD -t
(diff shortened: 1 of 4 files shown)


Cyan4973:
tests seems to be hitting an unrelated limitation on Windows.

Also, unrelated to this change but present in test traces,
the following error message is likely incorrect : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.1994/job/7d127hm57g57x7lc#L1617

tests/playTests.sh
```diff
@@ -165,6 +165,22 @@ rm tmp
 $ZSTD -f tmp && die "tmp not present : should have failed"
 test ! -f tmp.zst  # tmp.zst should not be created
 
+$ECHO "test : compress multiple files"
+echo hello > tmp1
+echo world > tmp2
+$ZSTD tmp1 tmp2 -o "$INTOVOID"
```
Cyan4973: Is this test supposed to work ?
I was expecting it to fail, because there are 2 input files associated with `-o` command.

terrelln: It only works with `/dev/null` or `-c`.

terrelln:
Rebase on top of PR #949 to fix the Windows test, since `$INTOVOID == NUL != nulmark == nul`.

PR #952, opened 2017-12-14
[fileio] Merge end loop for small optimization

terrelln:
When the entire file fits in `ress.srcBuffer`, setting `ZSTD_e_end` allows us to use the single-pass API.


Cyan4973:
The code simplification introduced by this patch is very welcomed.
Unfortunately, changes in `fileio` conflict with earlier patch ...

terrelln:
The conflict has been resolved.

PR #956, opened 2017-12-16
Split ZSTD_CCtx into smaller sub-structures

terrelln:
* Split out `ZSTD_matchState_t`.
* Move repcodes to `ZSTD_blockState_t` together with `ZSTD_entropyCTables_t`. Now there are two `ZSTD_blockState_t` objects, `prev` and `next`. When the block is confirmed we swap `prev` and `next`.
* Pass the match finders and dictionary builders only the pieces of the context that they need.

Notes to review the patch:
* The changes to the match finders are mechanical, and I don't expect issues (though still take a look!). The biggest change is that it reads/writes to the parameter `rep` instead of reading from `rep` and writing to `repToConfirm`.
* The biggest change is switching to `ZSTD_blockState_t`, and making the confirmation a pointer swap. This gives us the ability to always have valid entropy tables, instead of losing history when we reach an incompressible block. The places where this requires non-trivial changes are initialization/memory allocation since we change how much memory we need, and `ZSTD_compressBlock_internal()`, `ZSTD_compressSequences_internal()`, and `ZSTD_compressLiterals()`.

Follow up patches:
* Shrink the `ZSTD_CDict` to only contain `ZSTD_blockState_t` and `ZSTD_matchState_t`.
* Clean up `ZSTD_compressLiterals()`. It's currently copying the Huffman table an extra time.

Benchmarks:
Benchmarked on `silesia.tar` using one level for each strategy (levels 1, 3, 5, 7, 9, 13, 16). For each level I benchmark before the patch, then after.

```
level 1
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 249.7 MB/s , 647.0 MB/s
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 252.2 MB/s , 659.2 MB/s
level 3
 3#silesia.tar       : 211988480 ->  66993817 (3.164), 144.6 MB/s , 502.5 MB/s
 3#silesia.tar       : 211988480 ->  66993817 (3.164), 143.8 MB/s , 518.7 MB/s
level 5
 5#silesia.tar       : 211988480 ->  64757856 (3.274),  79.2 MB/s , 478.4 MB/s
 5#silesia.tar       : 211988480 ->  64757856 (3.274),  79.1 MB/s , 475.4 MB/s
level 7
 7#silesia.tar       : 211988480 ->  61762589 (3.432),  46.4 MB/s , 498.6 MB/s
 7#silesia.tar       : 211988480 ->  61762589 (3.432),  49.6 MB/s , 522.5 MB/s
level 9
 9#silesia.tar       : 211988480 ->  60692437 (3.493),  30.9 MB/s , 525.2 MB/s
 9#silesia.tar       : 211988480 ->  60692437 (3.493),  31.2 MB/s , 518.0 MB/s
level 13
13#silesia.tar       : 211988480 ->  58443973 (3.627),  9.62 MB/s , 504.7 MB/s
13#silesia.tar       : 211988480 ->  58443973 (3.627),  9.82 MB/s , 508.6 MB/s
level 16
16#silesia.tar       : 211988480 ->  56346349 (3.762),  4.76 MB/s , 510.7 MB/s
16#silesia.tar       : 211988480 ->  56346349 (3.762),  4.97 MB/s , 507.7 MB/s

```

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -51,12 +51,16 @@ typedef struct {
     FSE_CTable offcodeCTable[FSE_CTABLE_SIZE_U32(OffFSELog, MaxOff)];
     FSE_CTable matchlengthCTable[FSE_CTABLE_SIZE_U32(MLFSELog, MaxML)];
     FSE_CTable litlengthCTable[FSE_CTABLE_SIZE_U32(LLFSELog, MaxLL)];
-    U32 workspace[HUF_WORKSPACE_SIZE_U32];
     HUF_repeat hufCTable_repeatMode;
     FSE_repeat offcode_repeatMode;
     FSE_repeat matchlength_repeatMode;
     FSE_repeat litlength_repeatMode;
 } ZSTD_entropyCTables_t;
+
+typedef struct {
+  ZSTD_entropyCTables_t entropy;
+  U32 rep[ZSTD_REP_NUM];
+} ZSTD_blockState_t;
 
 typedef struct {
     U32 off;
@@ -92,6 +96,22 @@ typedef struct {
     /* end : updated by ZSTD_setLog2Prices */
     U32  staticPrices;           /* prices follow a pre-defined cost structure, statistics are irrelevant */
 } optState_t;
+
+typedef struct {
+    BYTE const* nextSrc;    /* next block here to continue on current prefix */ /* TODO: Does this belong here? */
+    BYTE const* base;       /* All regular indexes relative to this position */
+    BYTE const* dictBase;   /* extDict indexes relative to this position */
+    U32 dictLimit;          /* below that point, need extDict */
+    U32 lowLimit;           /* below that point, no more data */
+    U32 nextToUpdate;       /* index from which to continue table update */
+    U32 nextToUpdate3;      /* index from which to continue table update */
+    U32 hashLog3;           /* dispatch table : larger == faster, more memory */
+    U32 loadedDictEnd;      /* index of end of dictionary */                    /* TODO: Does this belong here? */
+    U32* hashTable;
+    U32* hashTable3;
+    U32* chainTable;
+    optState_t opt;         /* optimal parser state */
+} ZSTD_matchState_t;
 
 typedef struct {
     U32 offset;
@@ -136,15 +156,6 @@ struct ZSTD_CCtx_params_s {
 };  /* typedef'd to ZSTD_CCtx_params within "zstd.h" */
 
 struct ZSTD_CCtx_s {
-    const BYTE* nextSrc;    /* next block here to continue on current prefix */
-    const BYTE* base;       /* All regular indexes relative to this position */
-    const BYTE* dictBase;   /* extDict indexes relative to this position */
-    U32   dictLimit;        /* below that point, need extDict */
-    U32   lowLimit;         /* below that point, no more data */
-    U32   nextToUpdate;     /* index from which to continue dictionary update */
-    U32   nextToUpdate3;    /* index from which to continue dictionary update */
-    U32   hashLog3;         /* dispatch table : larger == faster, more memory */
-    U32   loadedDictEnd;    /* index of end of dictionary */
     ZSTD_compressionStage_e stage;
     U32   dictID;
     ZSTD_CCtx_params requestedParams;
@@ -159,12 +170,11 @@ struct ZSTD_CCtx_s {
     size_t staticSize;
 
     seqStore_t seqStore;    /* sequences storage ptrs */
-    optState_t optState;
     ldmState_t ldmState;    /* long distance matching state */
-    U32* hashTable;
-    U32* hashTable3;
-    U32* chainTable;
-    ZSTD_entropyCTables_t* entropy;
+    ZSTD_matchState_t matchState;
+    ZSTD_blockState_t* prevBlock;
+    ZSTD_blockState_t* nextBlock;
+    U32* entropyWorkspace;  /* entropy workspace of HUF_WORKSPACE_SIZE bytes */
 
     /* streaming */
     char*  inBuff;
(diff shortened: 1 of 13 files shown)


lib/compress/zstd_compress_internal.h
```diff
@@ -93,6 +97,22 @@ typedef struct {
     U32  staticPrices;           /* prices follow a pre-defined cost structure, statistics are irrelevant */
 } optState_t;
 
+typedef struct {
+    BYTE const* nextSrc;    /* next block here to continue on current prefix */ /* TODO: Does this belong here? */
```
Cyan4973: probably not.
This information is never used by matchfinder.

It's my understanding that `ZSTD_matchState_t` is only for core match finder / parser capabilities,
while `ZSTD_blockState_t` preserves additional data that needs to be transferred between blocks.
`nextSrc` seems to belong there.

lib/compress/zstd_compress_internal.h
```diff
@@ -93,6 +97,22 @@ typedef struct {
...
+    BYTE const* base;       /* All regular indexes relative to this position */
+    BYTE const* dictBase;   /* extDict indexes relative to this position */
+    U32 dictLimit;          /* below that point, need extDict */
+    U32 lowLimit;           /* below that point, no more data */
+    U32 nextToUpdate;       /* index from which to continue table update */
+    U32 nextToUpdate3;      /* index from which to continue table update */
+    U32 hashLog3;           /* dispatch table : larger == faster, more memory */
+    U32 loadedDictEnd;      /* index of end of dictionary */                    /* TODO: Does this belong here? */
```
Cyan4973: To review.
I don't remember exactly what `loadedDictEnd` does.
My first thought was "it should be `== dictLimit`", but since it exists, that's probably not the case.
Maybe `loadedDictEnd` is really for dictionary loading, while `dictLimit` is a more generic limit for `_extDict` ?
We may have a naming issue here, with different concepts having too similar names.

lib/compress/zstd_compress_internal.h
terrelln: Currently `ZSTD_blockState_t` transfers state from the last **compressed** block (or dictionary) to the next compressed block. Specifically, there are two copies `prevBlock` and `nextBlock`, and we only set `prevBlock = nextBlock` after we confirm that we are compressing the block. That isn't the case for `nextSrc`, since it needs to be updated for uncompressed blocks as well.

The logic for placing it here is that it is only used when `base`, `dictBase`, and the other variables here are updated. Given `ip` and `nextSrc`, we can update the rest of the variables. It could also go in the block state, which will then have to have a substructure for the part that needs "confirmation".

lib/compress/zstd_compress_internal.h
terrelln: I believe `loadedDictEnd` is for `ZSTD_p_forceWindow`. It is used [here](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1716) to either always allow references into the dictionary, or not (if `ZSTD_p_forceWindow == 1`).

lib/compress/zstd_compress_internal.h
Cyan4973: Maybe there is a need to distinguish inter-blocks state, from inter-compressed-blocks state.
I guess the inter-compressed-blocks state would become a part of inter-block state, and could use the swapping strategy proposed here.

lib/compress/zstd_compress_internal.h
Cyan4973: Right, so this is something set once, on loading dictionary with `ZSTD_p_forceWindow` (or `0` otherwise), and never updated again afterwards.
It's nonetheless useful during compression, to appropriately decide when to update `lowLimit`. So it's part of the compression context.
But it's not part of the "matchfinder" context, since it's never used in matchfinder / parser (while `lowlimit` is).

lib/compress/zstd_compress_internal.h
```diff
@@ -191,6 +201,12 @@ struct ZSTD_CCtx_s {
 };
 
 
+typedef size_t (*ZSTD_blockCompressor) (
+        ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
```
Cyan4973: Seeing this, I'm wondering if `rep` should be part of `ZSTD_matchState`.
I guess the issue is, `rep` is only updated if the current block is actually compressible, a result which will be discovered only after the block compressor is done. Hence it's included into `blockState` instead.
Maybe there is a need to underline the reasoning behind `blockState` existence directly into the type name.

Cyan4973:
This looks good.
I'm wondering if we should consider the following alternative structure (everything else remaining equal) : 

- `blockState` : fields always transferred between successive blocks
- `blockState.interCBlock` (or equivalent) : fields only updated if previous block was compressible (possible using "swapping" strategy, hence 2 copies)
- `blockState.matchState` : fields required by the matchfinder and parser (aka `blockCompressor`) (sidenote: maybe we should rename it `blockParser`, since compression itself only happens in the second entropy stage) 

For discussion

terrelln:
I like that structure. I'll update the patch.

terrelln:
I've rebased on top of the other changes made, but I still have to make the changes.

Cyan4973:
The patch is large, so I don't guarantee to have followed everything, but from what I could read, it looks pretty good.

Did you checked the potential impact on performance ?

Also : do you consider the patch "ready to merge", or do you want to add other topics (as listed in the description) ?

terrelln:
The only thing left to do is performance testing. I plan on following up with those items in subsequent PRs, since this one is already large enough.

terrelln:
I ran the benchmarks, and they look neutral, as expected. See the PR description.

Cyan4973:
Looks good to me !
It's a nice refactoring patch, which will help code evolution.

Just a little conflict to fix, due to prior merging of #984

PR #957, opened 2017-12-16
zstdmt via compress_generic: reduce opportunity to free/create mtctx

Cyan4973:
`zstreamtest --newapi` (and `--opaqueapi`) creates and destroys way too many threads
resulting in failure of tsan tests.
This issue is potentially connected to the `qemu` flaky tests on travis CI.

This is because, in `--newapi` mode, the nb of threads can be changed at each test.

The `--no-big-tests` directive reduces this choice to 1/2 threads,
in order to limit memory usage, especially for qemu and 32-bits builds.
Unfortunately, swapping between 1 and 2 threads is enough to constantly create/destroy new mtctx.

This patch takes advantage of the following property :
via compress_generic, no internal mtctx is needed for nbThreads < 2.
As a consequence, when nbThreads == 2, the currently active mtctx is fine, and can be re-used.
The strategy is generalized through control of current `nbThreads`.

It dramatically reduces the nb of thread creations when invoking `zstreamtest --newapi --no-big-tests`. It now only happens when parent cctx itself is created, which is randomized to 1/256 tests.

Expected outcome :
- at a minimum : tsan tests should now work without exploding the thread counter nearly as fast as they used to.
- at best : flaky qemu tests on `zstreamtest --newapi --no-big-tests` may stop being flaky, due to less stress from constant thread creation/destruction

Real world impact :
minimal, I don't expect users to constantly modify `nbThreads` at every invocation.
When `nbThreads` remains stable, existing implementation already re-uses existing mtctx.

Also : `zstreamtest --newapi` but without `--no-big-tests` doesn't benefit as much,
since this setting can select a random `nbThreads` value between 1 and 4.
The current patch reduces opportunity to free/create mtctx (for example : 3->1->3 doesn't need a new mtctx), but doesn't completely eliminate it, since `nbThreads` can still change between 2/3/4.
A more complete solution could be to separate thread usage from thread creation. Then, it becomes possible to downsize a pool by using 2 out of 4 allocated threads for example.
This would require a larger change to `POOL_*` api though.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2851,7 +2851,9 @@ size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
         if ((cctx->pledgedSrcSizePlusOne-1) <= ZSTDMT_JOBSIZE_MIN)
             params.nbThreads = 1; /* do not invoke multi-threading when src size is too small */
         if (params.nbThreads > 1) {
-            if (cctx->mtctx == NULL || cctx->appliedParams.nbThreads != params.nbThreads) {
+            if (cctx->mtctx == NULL || (params.nbThreads != ZSTDMT_getNbThreads(cctx->mtctx))) {
+                DEBUGLOG(4, "ZSTD_compress_generic: creating new mtctx for nbThreads=%u (previous: %u)",
+                            params.nbThreads, ZSTDMT_getNbThreads(cctx->mtctx));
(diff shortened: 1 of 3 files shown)


Cyan4973:
some good news : 
- `qemu` multithreading tests have completed without issue on this PR (though it could be pure luck)
- tsan fuzzer tests are running since several hours on both devserver and a low-mem VM, no issue so far (to be more complete, I don't remember at which nb of tests it used to stop due to thread counter exhaustion)

PR #961, opened 2017-12-22
fix unbounded range

shawnl:
I think you meant 8 MiB or smaller, instead of an unbounded (and illogical) range


Cyan4973:
Indeed !

PR #965, opened 2017-12-27
%.o objects files in /tests

Cyan4973:
Recipes in `/tests` rebuild everything from source for each target.
`zstd` is still a "small" project, so it's not prohibitive,
yet, rebuilding same files over and over represents substantial redundant work.

This patch replaces `*.c` files from `/lib` by their corresponding `*.o` object files.
They cannot be compiled and stored directly within `/lib`,
since `/tests` triggers additional debug capabilities unwelcome in release binary.
So the resulting `*.o` object files are stored within `/tests`.

It turns out, it's not common for several targets to use *exactly* the same rules.
Using only the default rules (debug enabled, multi-threading disabled, no legacy)
a surprisingly small amount of targets share their work.

It's because, in many cases there are additional modifications requested :
some targets are 32-bits, some enable multi-threading, some enable legacy support,
some disable asserts, some want different kind of sanitizer, etc.

I created 2 sets of object files : with and without multithreading.
Several targets now share their object files, saving compilation time on running `make all`.
For these targets, when modifying one source file, only this one needs rebuilding.

For all other targets requiring different settings, build from source `*.c` remains the default rule.

The new rules have been successfully tested with `-j` parallel compilation.


tests/Makefile
```diff
@@ -87,54 +98,92 @@ zstd-nolegacy:
...
+zstdm_%.o : $(ZSTDDIR)/common/%.c
+	$(CC) -c $(CPPFLAGS) $(CFLAGS) $< -o $@
+
+zstdc_%.o : $(ZSTDDIR)/compress/%.c
+	$(CC) -c $(CPPFLAGS) $(CFLAGS) $< -o $@
+
+zstdd_%.o : $(ZSTDDIR)/decompress/%.c
+	$(CC) -c $(CPPFLAGS) $(CFLAGS) $< -o $@
+
+zstdmt%.o : CPPFLAGS += -DZSTD_MULTITHREAD=1
```
terrelln: Should you use `$(MULTITHREAD_CPP)` here?

Cyan4973: Yes @terrelln , I guess it would feel more logical

PR #966, opened 2017-12-28
installing libraries to correct paths on x86_64

udaykkumar:
[#926](https://github.com/facebook/zstd/issues/926)


Cyan4973:
Thanks for the patch @udaykkumar !

Could you please point to a documentation which explains how this `/lib` vs `/lib64` configuration works ?

Currently, `Makefile` installs `libzstd` into `/usr/local/lib` by default. It has always worked fine so far, no complaint despite multiple releases.
With this patch, `Makefile` and `cmake` would feature a different behavior, which is probably not great. I don't oppose a policy change, but it's better be well documented.

build/cmake/lib/CMakeLists.txt
```diff
@@ -143,10 +143,18 @@ IF (ZSTD_BUILD_STATIC)
             OUTPUT_NAME ${STATIC_LIBRARY_BASE_NAME})
 ENDIF (ZSTD_BUILD_STATIC)
 
+
+GET_PROPERTY( LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS )
```
Cyan4973: Apparently, `FIND_LIBRARY_USE_LIB64_PATHS ` requires cmake >= v3.0.2.
https://cmake.org/cmake/help/v3.0/prop_gbl/FIND_LIBRARY_USE_LIB64_PATHS.html

Note that `CMakeLists.txt` is currently compatible with >= v2.8.9 : 
https://github.com/facebook/zstd/blob/dev/build/cmake/CMakeLists.txt#L11
As a general rule, increasing the minimum version number is not desirable, as it correspondingly decreases the number of distributions which can satisfy the dependency.

We currently try to keep `cmake` script compatible with versions <= 2.8.12, as it is the version shipped within Ubuntu Trusty, an actively supported LTS release, used by several CI systems such as TravisCI and CircleCI.

udaykkumar: This is the one I was referring to https://cmake.org/cmake/help/v2.8.12/cmake.html v2.8.12, and Probably  my mistake I should have approached for make file as well

udaykkumar: The intention was to let `cmake` or `make` decide install location based on arch, I believe the `cmake` property `FIND_LIBRARY_USE_LIB64_PATHS` should help in this. If you feel this is worth an effort, specially considering there will be a change in install location for future releases. I'll go ahead and do a bit of work on `Makefile` as well.

Cyan4973:
Looking around, I found this documentation, called [Filesystem Hierarchy Standard](http://www.pathname.com/fhs/).

If I read it correctly, it states that [`/usr/lib64` is completely optional](http://www.pathname.com/fhs/pub/fhs-2.3.html#USRLIBLIBRARIESFORPROGRAMMINGANDPA), only `/usr/lib` is really necessary. (note: logic is mirrored for `/usr/local/lib`).

Given this background, I'm failing to see a benefit in installing within `/usr/local/lib64` instead of `/usr/local/lib`. I'm actually worried of potential impact on path management.

The only use case that comes to mind is if someone wants several binary variants : then the 32-bits version can be installed in `/usr/local/lib32` for example.
But that's a fairly advanced scenario.

Cyan4973:
Another note, reading #926 : 

My understanding of this request is
to create a flag that user can trigger, in order to install `libzstd` within `/usr/lib64` (or `/usr/local/lib64` ?) instead of the current default `usr/local/lib`.

It doesn't state that the new default should become `/usr/local/lib64`.

udaykkumar:
I suppose this doesn't seem to address the actual requirement in anyway and seems to be diverting away for what is required. I'll close this PR if there is nothing to discuss anymore

Cyan4973:
Yes, it's a difficult topic, due to the variety of platforms that can be impacted.

I think the "opt-in" approach requested in #926 is fine, 
it keeps user in charge, without introducing new risks for default install.

PR #968, opened 2017-12-29
meson: get soversion right

shawnl:
this is necessary to link correctly

### DIFF
--- a/contrib/meson/meson.build
+++ b/contrib/meson/meson.build
@@ -84,7 +84,9 @@ libzstd = library('zstd',
                   include_directories: libzstd_includes,
                   c_args: libzstd_cflags,
                   dependencies: libzstd_deps,
-                  install: true)
+                  install: true,
+                  soversion: '1',
+                 )
 
 programs_dir = join_paths('..', '..', 'programs')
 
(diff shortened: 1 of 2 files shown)


contrib/meson/meson.build
```diff
@@ -40,6 +40,9 @@ libzstd_includes = [include_directories(common_dir, dictbuilder_dir, compress_di
 
 legacy = get_option('legacy_support')
 if legacy != 'false'
+    if legacy == 'true'
+        legacy = '0'
```
Cyan4973: Could you describe what this statement is expected to achieve ?

shawnl: it allos you to do "meson configure -Dlegacy=true" and have it equivalent to make -DZSTD_LEGACY_SUPPORT=0 (which compiles in all legacy options). That 0 does that is not so clear.

shawnl: Not that I see the documentation I see this is totally wrong, I will update the patch

Cyan4973: You probably want a positive value >= 1 here

shawnl:
fixed.

PR #970, opened 2017-12-30
Delayed update Binary Tree for Lazy strategy

Cyan4973:
This is a pretty nice speed win.

The new strategy consists in stacking new candidates as if it was a hash chain.
Then, only if there is a need to actually consult the chain, they are batch-updated,
before starting the match search itself.
This is supposed to be beneficial when skipping positions,
which happens a lot when using lazy strategy.

The baseline performance for `btlazy2` on my laptop is :
```
15#calgary.tar       :   3265536 ->    955985 (3.416),  7.06 MB/s , 618.0 MB/s
15#enwik7            :  10000000 ->   3067341 (3.260),  4.65 MB/s , 521.2 MB/s
15#silesia.tar       : 211984896 ->  58095131 (3.649),  6.20 MB/s , 682.4 MB/s
```
(only level 15 remains for `btlazy2`, as this strategy is squeezed between `lazy2` and `btopt`)

After this patch, and keeping all parameters identical,
speed is increased by a pretty good margin (+30-50%),
but compression ratio suffers a bit :
```
15#calgary.tar       :   3265536 ->    958060 (3.408),  9.12 MB/s , 621.1 MB/s
15#enwik7            :  10000000 ->   3078318 (3.249),  6.37 MB/s , 525.1 MB/s
15#silesia.tar       : 211984896 ->  58444111 (3.627),  9.89 MB/s , 680.4 MB/s
```

That's because it keeps `1<<searchLog` as a maximum number of unsorted candidates to update.
For a hash chain, it represents the total number of candidates in the chain,
while for binary tree, it represents the maximum depth of searches.
Keep in mind that a lot of candidates won't even be visited in the binary tree,
since they are filtered out by the sort operation.

As a consequence, in the new implementation,
the effective depth of the binary tree is substantially shorter.

To compensate, it's enough to increase `searchLog` value.
Here is the result after adding just +1 to searchLog :
```
15#calgary.tar       :   3265536 ->    956311 (3.415),  8.32 MB/s , 611.4 MB/s
15#enwik7            :  10000000 ->   3067655 (3.260),  5.43 MB/s , 535.5 MB/s
15#silesia.tar       : 211984896 ->  58113144 (3.648),  8.35 MB/s , 679.3 MB/s
```

aka, almost the same compression ratio as before,
but with a noticeable speed increase (+20-30%).

This modification makes `btlazy2` more competitive.
A new paramgrill round determined it could also be applied to level 13 and 14, so compression level table has been adapted.

Not that this patch will conflict with #964, since they work on same files.
#964 usefulness is less clear though, since it merely moves around speed/ratio trade off,
whereas this patch produces a net gain, albeit on a limited scope (`btlazy2`).


lib/compress/zstd_compress.c
```diff
@@ -1112,19 +1142,30 @@ static void ZSTD_ldm_reduceTable(ldmEntry_t* const table, U32 const size,
 *   rescale all indexes to avoid future overflow (indexes are U32) */
 static void ZSTD_reduceIndex (ZSTD_CCtx* zc, const U32 reducerValue)
 {
-    { U32 const hSize = (U32)1 << zc->appliedParams.cParams.hashLog;
-      ZSTD_reduceTable(zc->hashTable, hSize, reducerValue); }
+    {   U32 const hSize = (U32)1 << zc->appliedParams.cParams.hashLog;
+        ZSTD_reduceTable(zc->hashTable, hSize, reducerValue);
+    }
+
+    if (zc->appliedParams.cParams.strategy != ZSTD_btlazy2) {
```
terrelln: Should this be deleted? It looks like for strategy that isn't `ZSTD_btlazy2` or `ZSTD_fast`, the chain table will be reduced twice.

lib/compress/zstd_compress.c
```diff
@@ -1112,19 +1142,30 @@ static void ZSTD_ldm_reduceTable(ldmEntry_t* const table, U32 const size,
...
+        U32 const chainSize = (U32)1 << zc->appliedParams.cParams.chainLog;
+        ZSTD_reduceTable(zc->chainTable, chainSize, reducerValue);
+    }
 
-    { U32 const chainSize = (zc->appliedParams.cParams.strategy == ZSTD_fast) ? 0 : ((U32)1 << zc->appliedParams.cParams.chainLog);
-      ZSTD_reduceTable(zc->chainTable, chainSize, reducerValue); }
+    if (zc->appliedParams.cParams.strategy != ZSTD_fast) {
+        U32 const chainSize = (U32)1 << zc->appliedParams.cParams.chainLog;
+        if (zc->appliedParams.cParams.strategy != ZSTD_btlazy2)
```
terrelln: Should this be `== ZSTD_btlazy2`?

Cyan4973: Yes, indeed, it should.
And now I'm wondering why the associated test has not crashed ...

lib/compress/zstd_compress.c
Cyan4973: You are right, this is messed up.
As a mitigation, I suspect that the second `reduce` is a no-op, since it will discover it can't reduce indexes much further.
Still, it's useless cpu cycles spent.

lib/compress/zstd_compress.c
Cyan4973: OK, both issues must be fixed.
Though indeed, it's not as impactful as I initially feared.

The situation was worse when `ZSTD_DUBT_UNSORTED_MARK` was defined as `0xFFFFFFFF`. In which case, after a reduce, the resulting index was pointing nowhere.

Now, since it's defined as `1`, a reduce operation squash it to `0`, which means "no further candidate". It's a much smaller impact, and I could not create a corruption or crash issue with it. But that could just be due to lack of samples. `ZSTD_preserveUnsortedMark()` just ensures there is no unforeseen side-effect after a reduce.

lib/compress/zstd_lazy.c
```diff
@@ -15,76 +15,117 @@
...
-            continue;
-        }
-#endif
-
-        if ((!extDict) || (matchIndex+matchLength >= dictLimit)) {
-            assert(matchIndex+matchLength >= dictLimit);   /* might be wrong if extDict is incorrectly set to 0 */
-            match = base + matchIndex;
+        if ( (!extDict)
+          || (matchIndex+matchLength >= dictLimit)  /* both in current segment*/
+          || (current < dictLimit) /* both in extDict */) {
```
terrelln: Could `matchLength` be potentially cut to the end of the extDict in this case, where it really extends further?

lib/compress/zstd_opt.c
```diff
@@ -265,6 +264,147 @@ static U32 ZSTD_insertAndFindFirstIndexHash3 (ZSTD_CCtx* const cctx, const BYTE*
 /*-*************************************
 *  Binary Tree search
 ***************************************/
+/** ZSTD_insertBt1() : add one or multiple positions to tree.
+ *  ip : assumed <= iend-8 .
+ * @return : nb of positions added */
+static U32 ZSTD_insertBt1(ZSTD_CCtx* zc,
```
terrelln: Just to make sure, this is just copied from the old `ZSTD_lazy.c` with no other changes, right?

lib/compress/zstd_lazy.c
```diff
@@ -108,132 +146,154 @@ static U32 ZSTD_insertBt1(ZSTD_CCtx* zc,
...
-            if (matchLength > matchEndIdx - matchIndex)
-                matchEndIdx = matchIndex + (U32)matchLength;
-            if ( (4*(int)(matchLength-bestLength)) > (int)(ZSTD_highbit32(current-matchIndex+1) - ZSTD_highbit32((U32)offsetPtr[0]+1)) )
-                bestLength = matchLength, *offsetPtr = ZSTD_REP_MOVE + current - matchIndex;
-            if (ip+matchLength == iend) {   /* equal : no way to know if inf or sup */
-                break;   /* drop, to guarantee consistency (miss a little bit of compression) */
+    /* batch sort stacked candidates */
+    matchIndex = previousCandidate;
+    while (matchIndex) {  /* will end on matchIndex == 0 */
+        U32* const nextCandidateIdxPtr = bt + 2*(matchIndex&btMask) + 1;
```
terrelln: Any reason to have this variable instead of just `*(bt + 2*(matchIndex&btMask) + 1)`?

lib/compress/zstd_lazy.c
Cyan4973: Yes.
This is a simplification. It avoids creating some new counting logic, which would be quite complex to handle due to buffer changes, while the expected improvement is fairly limited (negligible).

lib/compress/zstd_lazy.c
Cyan4973: Not really. It was just easier to think it this way.
It can be coalesced into `nextCandidateIdx` if that matters.

lib/compress/zstd_opt.c
Cyan4973: Yes. Since `ZSTD_insertBt1()` is no longer provided by `zstd_lazy.c`, it needs to be implemented directly within `zstd_opt.c`.

Cyan4973:
Yes, this is a correct description.

lib/compress/zstd_lazy.c
```diff
@@ -108,132 +146,154 @@ static U32 ZSTD_insertBt1(ZSTD_CCtx* zc,
...
-
-void ZSTD_updateTree_extDict(ZSTD_CCtx* zc,
-                const BYTE* const ip, const BYTE* const iend,
-                const U32 nbCompares, const U32 mls)
-{
-    ZSTD_updateTree_internal(zc, ip, iend, nbCompares, mls, 1 /*extDict*/);
 }
 
 
 static size_t ZSTD_insertBtAndFindBestMatch (
```
terrelln: Should this also be renamed to `DUBT`?

PR #972, opened 2018-01-02
zstd: fix crash when not overwriting existing files

pixelb:
This fixes the following crash:
  $ touch exists
  $ programs/zstd -r examples/ -o exists
  zstd: exists already exists; not overwritten
  Segmentation fault (core dumped)

* programs/fileio.c (FIO_compressMultipleFilenames):
Handle the case where we're not overwriting the destination.

Reported at https://bugzilla.redhat.com/1530049


Cyan4973:
Thanks for this straightforward patch @pixelb  !

Just waiting for the new corresponding test case to be confirmed and added before merging the fix.

PR #973, opened 2018-01-03
Add test case for PR #972

terrelln:
Will fail before PR #972 is landed.


Cyan4973:
Thanks  @terrelln !

I completely forgot (or did not notice) that `-o` behavior was changed to allow multiple source files.
I guess I shall update the `v1.3.3` release note to advertise the change.

I also believe the man page has not been updated, as it still pretends that `-o` only works for one source.

Cyan4973:
Wasn't this test supposed to fail ?

I suspect it does not make a difference between a "normal" failure (which is expected) and a segfault.

terrelln:
I messed up initially and it passed. I fixed it for my machine, though it looks like it didn't work on Linux. Looking into it.

terrelln:
Looks like a problem with my shell syntax https://travis-ci.org/facebook/zstd/jobs/324831773#L646.

terrelln:
The tests fail in the expected place now.

PR #988, opened 2018-01-15
CMake: added option `INSTALL_LIB_DIR`

krab:
Libraries installation directory name now can be specified by using `-DINSTALL_LIB_DIR=<dirname>`

For example for x86_64 Fedora/openSUSE: `cmake -DCMAKE_INSTALL_PREFIX=/usr -DINSTALL_LIB_DIR=lib64`
Debian/Ubuntu: `cmake -DCMAKE_INSTALL_PREFIX=/usr -DINSTALL_LIB_DIR=lib/x86_64-linux-gnu`


Conan-Kudo:
@krab Why not just use `GNUInstallDirs`? https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

krab:
Because zstd author wants cmake 2.8.9 support w/o changing how install works by default, plus cmake only patch to start with.

https://github.com/facebook/zstd/pull/929#issuecomment-347334719
https://github.com/facebook/zstd/pull/966#issuecomment-354285246

Conan-Kudo:
@krab `GNUInstallDirs` has been present since CMake 2.8.5, so insofar as a CMake-only improvement, that should work just fine.

krab:
@Conan-Kudo i guess you are right. I should rewrite new pull request against GNUInstallDirs to avoid potential problems with ldconfig.
 https://github.com/Kitware/CMake/commit/620939e4e6f5a61cd5c0fac2704de4bfda0eb7ef

Conan-Kudo:
💯

PR #1000, opened 2018-01-24
Progressive flush

Cyan4973:
This patch was more difficult than anticipated to get correct.
I'm thankful we have fuzzer tools which are unforgiving and produce crazy combinations to test.
Nonetheless, please review this patch carefully.

In the end, the only visible feature this patch enables is to receive regular feedback on the compression process, ensuring smooth display on console during compression with large window sizes.
Try it on some large file with `zstd --long`, the difference with current `dev` branch will be obvious.

It also flushes jobs immediately, before it ends, which also improves latency when window is large, hence when job is large, though it requires a rate-limited I/O to see its impacts (by default, files are cached by OS, so flush speed is too fast to notice a difference).

If it was the only benefit, its complexity would be questionable.
But the real price will be seen in future patches, when it becomes possible to dynamically probe compression speed to decide automatically to increase / decrease compression level depending on I/O condition, as demoed in `adaptive-compression`.

### DIFF
--- a/lib/common/pool.h
+++ b/lib/common/pool.h
@@ -17,7 +17,8 @@ extern "C" {
 
 
 #include <stddef.h>   /* size_t */
-#include "zstd_internal.h"   /* ZSTD_customMem */
+#define ZSTD_STATIC_LINKING_ONLY   /* ZSTD_customMem */
+#include "zstd.h"
 
 typedef struct POOL_ctx_s POOL_ctx;
 
@@ -27,35 +28,43 @@ typedef struct POOL_ctx_s POOL_ctx;
  *  The maximum number of queued jobs before blocking is `queueSize`.
  * @return : POOL_ctx pointer on success, else NULL.
 */
-POOL_ctx *POOL_create(size_t numThreads, size_t queueSize);
+POOL_ctx* POOL_create(size_t numThreads, size_t queueSize);
 
-POOL_ctx *POOL_create_advanced(size_t numThreads, size_t queueSize, ZSTD_customMem customMem);
+POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize, ZSTD_customMem customMem);
 
 /*! POOL_free() :
     Free a thread pool returned by POOL_create().
 */
-void POOL_free(POOL_ctx *ctx);
+void POOL_free(POOL_ctx* ctx);
 
 /*! POOL_sizeof() :
     return memory usage of pool returned by POOL_create().
 */
-size_t POOL_sizeof(POOL_ctx *ctx);
+size_t POOL_sizeof(POOL_ctx* ctx);
 
 /*! POOL_function :
     The function type that can be added to a thread pool.
 */
-typedef void (*POOL_function)(void *);
+typedef void (*POOL_function)(void*);
 /*! POOL_add_function :
     The function type for a generic thread pool add function.
 */
-typedef void (*POOL_add_function)(void *, POOL_function, void *);
+typedef void (*POOL_add_function)(void*, POOL_function, void*);
 
 /*! POOL_add() :
-    Add the job `function(opaque)` to the thread pool.
+    Add the job `function(opaque)` to the thread pool. `ctx` must be valid.
     Possibly blocks until there is room in the queue.
     Note : The function may be executed asynchronously, so `opaque` must live until the function has been completed.
 */
-void POOL_add(void *ctx, POOL_function function, void *opaque);
+void POOL_add(POOL_ctx* ctx, POOL_function function, void* opaque);
+
+
+/*! POOL_tryAdd() :
+    Add the job `function(opaque)` to the thread pool if a worker is available.
+    return immediately otherwise.
+   @return : 1 if successful, 0 if not.
+*/
+int POOL_tryAdd(POOL_ctx* ctx, POOL_function function, void* opaque);
 
 
 #if defined (__cplusplus)
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1249,7 +1249,7 @@ static size_t ZSTDMT_flushStream_internal(ZSTDMT_CCtx* mtctx, ZSTD_outBuffer* ou
     size_t const srcSize = mtctx->inBuff.filled - mtctx->prefixSize;
     DEBUGLOG(5, "ZSTDMT_flushStream_internal");
 
-    if ( ((srcSize > 0) || (endFrame && !mtctx->frameEnded))
+    if ( (mtctx->jobReady || (srcSize > 0) || (endFrame && !mtctx->frameEnded))
        && (mtctx->nextJobID <= mtctx->doneJobID + mtctx->jobIDMask) ) {
            DEBUGLOG(5, "ZSTDMT_flushStream_internal : create a new job");
         CHECK_F( ZSTDMT_createCompressionJob(mtctx, srcSize, endFrame) );
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -300,28 +300,28 @@ static void ZSTDMT_releaseCCtx(ZSTDMT_CCtxPool* pool, ZSTD_CCtx* cctx)
 
 
 /* ------------------------------------------ */
-/* =====          Thread worker         ===== */
+/* =====          Worker thread         ===== */
 /* ------------------------------------------ */
 
 typedef struct {
-    buffer_t src;
-    const void* srcStart;
-    size_t   prefixSize;
-    size_t   srcSize;
-    size_t   consumed;
-    buffer_t dstBuff;
-    size_t   cSize;
-    size_t   dstFlushed;
-    unsigned firstChunk;
-    unsigned lastChunk;
-    unsigned frameChecksumNeeded;
-    ZSTD_pthread_mutex_t* mtctx_mutex;
-    ZSTD_pthread_cond_t* mtctx_cond;
-    ZSTD_CCtx_params params;
-    const ZSTD_CDict* cdict;
-    ZSTDMT_CCtxPool* cctxPool;
-    ZSTDMT_bufferPool* bufPool;
-    unsigned long long fullFrameSize;
+    size_t   consumed;       /* SHARED - init0 by mtctx, then modified by worker AND read by mtctx */
+    size_t   cSize;          /* SHARED - init0 by mtctx, then modified by worker AND read by mtctx */
+    ZSTD_pthread_mutex_t* mtctx_mutex;   /* Thread-safe - used by mtctx and (all) worker */
+    ZSTD_pthread_cond_t* mtctx_cond;     /* Thread-safe - used by mtctx and (all) worker */
+    ZSTDMT_CCtxPool* cctxPool;           /* Thread-safe - used by mtctx and (all) worker */
+    ZSTDMT_bufferPool* bufPool;          /* Thread-safe - used by mtctx and (all) worker */
+    buffer_t dstBuff;        /* set by worker (or mtctx), then read by worker, then modified by mtctx => no barrier */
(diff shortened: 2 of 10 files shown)


lib/common/pool.h
```diff
@@ -27,35 +28,43 @@ typedef struct POOL_ctx_s POOL_ctx;
...
 /*! POOL_function :
     The function type that can be added to a thread pool.
 */
-typedef void (*POOL_function)(void *);
+typedef void (*POOL_function)(void*);
 /*! POOL_add_function :
     The function type for a generic thread pool add function.
 */
-typedef void (*POOL_add_function)(void *, POOL_function, void *);
+typedef void (*POOL_add_function)(void*, POOL_function, void*);
```
terrelln: Ah right, this is why `POOL_add()` took a `void*` instead of a `POOL_ctx*`. Its so that work could be scheduled on an external thread pool, like:

```c
ZSTD_CCtx_setThreadPool(void* ctx, POOL_add_function add);
```

Its something that could be interesting in the future, but we can change it back then if we end up needing it.

lib/compress/zstdmt_compress.c
```diff
@@ -993,134 +1001,162 @@ size_t ZSTDMT_initCStream(ZSTDMT_CCtx* zcs, int compressionLevel) {
...
-            /* single chunk exception : checksum is calculated directly within worker thread */
-            zcs->params.fParams.checksumFlag = 0;
-    }   }
+    if (zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask) {
+        DEBUGLOG(5, "ZSTDMT_createCompressionJob: will not create new job : table is full");
+        assert((zcs->nextJobID & zcs->jobIDMask) == (zcs->doneJobID & zcs->jobIDMask));
+        return 0;
+    }
+
+    if (!zcs->jobReady) {
```
terrelln: I wish github would mark whitespace only changes as no-change lines...

lib/compress/zstdmt_compress.c
```diff
@@ -993,134 +1001,162 @@ size_t ZSTDMT_initCStream(ZSTDMT_CCtx* zcs, int compressionLevel) {
...
-        zcs->inBuff.buffer = g_nullBuffer;
-        zcs->inBuff.filled = 0;
-        zcs->prefixSize = 0;
-        zcs->frameEnded = 1;
-        if (zcs->nextJobID == 0) {
-            /* single chunk exception : checksum is calculated directly within worker thread */
-            zcs->params.fParams.checksumFlag = 0;
-    }   }
+    if (zcs->nextJobID > zcs->doneJobID + zcs->jobIDMask) {
+        DEBUGLOG(5, "ZSTDMT_createCompressionJob: will not create new job : table is full");
```
terrelln: Is it possible for `jobReady == 1` here? I don't think it is.

lib/compress/zstdmt_compress.c
```diff
@@ -993,134 +1001,162 @@ size_t ZSTDMT_initCStream(ZSTDMT_CCtx* zcs, int compressionLevel) {
...
+                DEBUGLOG(5, "Job %u completed (%u bytes), moving to next one",
+                        zcs->doneJobID, (U32)job.dstFlushed);
+                ZSTDMT_releaseBuffer(zcs->bufPool, job.dstBuff);
+                zcs->jobs[wJobID].dstBuff = g_nullBuffer;
+                zcs->jobs[wJobID].jobCompleted = 0;
+                zcs->consumed += job.srcSize;
+                zcs->produced += job.cSize;
+                zcs->doneJobID++;
+            } else {
+                zcs->jobs[wJobID].dstFlushed = job.dstFlushed;   /* remember how much was flushed for next attempt */
```
terrelln: Since this is updated without a lock, and with `job.jobCompleted == 0`, we have to be careful to never read/write `dstFlushed` in the worker thread.

lib/compress/zstdmt_compress.c
```diff
@@ -993,134 +1001,162 @@ size_t ZSTDMT_initCStream(ZSTDMT_CCtx* zcs, int compressionLevel) {
...
+        }   }
+
+        /* return value : how many bytes left in buffer ; fake it to 1 when unknown but >0 */
         if (job.cSize > job.dstFlushed) return (job.cSize - job.dstFlushed);
-        if (zcs->doneJobID < zcs->nextJobID) return 1;   /* still some buffer to flush */
-        zcs->allJobsCompleted = zcs->frameEnded;   /* frame completed and entirely flushed */
-        return 0;   /* everything flushed */
-}   }
+        if (job.srcSize > job.consumed) return 1;   /* current job not completely compressed */
```
terrelln: Do we need `if (!job.jobCompleted) return 1;`? Otherwise, we could have consumed all the input, read all the output, but the job still has to produce one more chunk of output. Without it, setting `allJobsCompleted = zcs->frameEnded;` and returning `0` could be wrong.

lib/compress/zstdmt_compress.c
Cyan4973: I don't think it is, since `nextJobID` is updated only after the job is accepted by `POOL_tryAdd()` (see line 1077)

lib/compress/zstdmt_compress.c
Cyan4973: `dstFlushed` is indeed never used / read in the worker thread.
Maybe I should refactor `ZSTDMT_jobDescription` to clearly label which member is shared or not.

lib/compress/zstdmt_compress.c
Cyan4973: I've considered removing `job.jobCompleted` because it's (currently) the same signal as `job.srcSize == job.consumed`.

The worker thread uses the synchronous bufferless API, so there is no "data retention" inside context : any data read immediately produces a compressed block. When `job.srcSize == job.consumed`, we have reached the end of job, and produced all the compressed content.

terrelln: Ahh, consumed is only updated after the compression is completed. Can you add either add an assert, or go ahead and remove `jobCompleted`?

Cyan4973: Removing `jobCompleted` introduces a new problem : 
in some (rare) cases, it is necessary to create a last 0-size job just to finish the frame.
In which case, `job.srcSize == 0 == job.consumed` immediately, even though the job itself (creating an empty block with "end of frame" marker) is not yet done.

Potential mitigations : 
1) keep `jobCompleted` signal
2) add some special code to create a last 0-size block without going through the job creation process.
    + sounds like an obvious optimisation.
        However, it adds complexity, for a very rare use case. 
        More importantly, `zstdmt` is up to now agnostic of the frame format. Creating such a `0-size` last block directly inside `zstdmt_compress.c` makes the frame format a new dependency of this module. Alternatively, the `0-size` last block could be done inside `zstd_compress.c` and invoked from `zstdmt_compress.c`, though this is a very specialised entry point without another user for the time being.

Cyan4973: finally decided to remove `job.jobCompleted`
and use `job.consumed==job.srcSize` instead,
complemented with `ZSTD_writeLastEmptyBlock()` for the special case `job.srcSize==0`.

Cyan4973:
Checking performance : 
I'm testing a large ~10GB file on `entropy`.
While the file seems "large", `entropy` has enough memory to cache it all, so speed is really high.
Anyway, the objective is to test potential speed regression due to usage of a single mutex.
Here are some results : 
```
time zstd -1 -T12 10GBfile -c > /dev/null
v1.3.3 : 5.640s
dev    : 5.680s
pflush : 5.720 s
```

There is definitely a loss of performance compared to `v1.3.3`, though it's moderate. 
It's too easy to attribute it entirely to the lock. I don't know the root cause at this stage.
But this provides a starting point towards experimenting one-lock-per-job strategy.

Cyan4973:
Latest update adds one mutex per job (instead of a single mutex by context).
This sharply reduces risks of lock contention when a lot of threads are active, since only one worker and the main context share a mutex.

Tested on the same file and system as before, performance improves a bit : 
```
before : 5.720s
after  : 5.660s
v1.3.3 : 5.640s
```

This is still a bit slower than v1.3.3, though the difference feels low enough.
I've repeated the result many times though, so it's more than a measurement fluke.
I have no idea here where the remaining difference may come from. Increased complexity maybe.

lib/compress/zstdmt_compress.c
```diff
@@ -361,76 +364,70 @@ void ZSTDMT_compressChunk(void* jobDescription)
...
         for (blockNb = 1; blockNb < nbBlocks; blockNb++) {
-            size_t const cSize = ZSTD_compressContinue(cctx, op, oend-op, ip, ZSTD_BLOCKSIZE_MAX);
+            size_t const cSize = ZSTD_compressContinue(cctx, op, oend-op, ip, blockSize);
             if (ZSTD_isError(cSize)) { job->cSize = cSize; goto _endJob; }
-            ip += ZSTD_BLOCKSIZE_MAX;
+            ip += blockSize;
             op += cSize; assert(op < oend);
             /* stats */
-            ZSTD_PTHREAD_MUTEX_LOCK(job->jobCompleted_mutex);   /* note : it's a mtctx mutex */
+            ZSTD_PTHREAD_MUTEX_LOCK(&job->job_mutex);   /* note : it's a mtctx mutex */
```
terrelln: The comment is out of date now.

lib/compress/zstdmt_compress.c
```diff
@@ -361,76 +364,70 @@ void ZSTDMT_compressChunk(void* jobDescription)
...
 
     /* compress */
-#if 0
-    job->cSize = (job->lastChunk) ?
-                 ZSTD_compressEnd     (cctx, dstBuff.start, dstBuff.size, src, job->srcSize) :
-                 ZSTD_compressContinue(cctx, dstBuff.start, dstBuff.size, src, job->srcSize);
-#else
-    if (sizeof(size_t) > sizeof(int)) assert(job->srcSize < ((size_t)INT_MAX) * ZSTD_BLOCKSIZE_MAX);   /* check overflow */
-    {   int const nbBlocks = (int)((job->srcSize + (ZSTD_BLOCKSIZE_MAX-1)) / ZSTD_BLOCKSIZE_MAX);
+    {   size_t const blockSize = ZSTD_BLOCKSIZE_MAX;
```
terrelln: Is it possible for the user to pass in a window size < `ZSTD_BLOCKSIZE_MAX`?

Cyan4973: Let's see,
A `windowSize < ZSTD_BLOCKSIZE_MAX` presumes a `srcSize < ZSTD_BLOCKSIZE_MAX`.
In such situation, when `windowSize` is automatically set by `zstd`, it always ensures that `windowSize >= srcSize`. So the algorithm skips the loop and goes straight to "last block".

However, we could be in a less common situation,
where `windowSize` has been set manually to a low value, 
while `srcSize` could be very large.
In which case, the real `blockSize` used in the frame will be `== windowSize < ZSTD_BLOCKSIZE_MAX`,
while `ZSTDMT_compressChunk()` will continue to feed `ZSTD_compressContinue()` with inputs of `ZSTD_BLOCKSIZE_MAX` bytes.

That being, said, it's not necessarily a problem.
`ZSTD_compressContinue()` will, internally, cut this input into as many blocks as necessary.
It knows the size of blocks : it's stored as `cctx->blockSize`, and was set during initialisation.
`cctx->blockSize` has taken into consideration `windowSize`.

All is well as long as `blockSize` is a clean power of 2, so that `inputSize == ZSTD_BLOCKSIZE_MAX` will be divided into an integer number of blocks. The specification allows non-power-of-2 window sizes, and by extension, block sizes, but nothing in the code make it possible right now.
Besides, in such a case, the drawback will be that `ZSTD_compressContinue()` will create a "small last block", but that's a minor setback, everything will continue to work and respect the specification.

I guess I'd better rename this variable, so that it's not confused with internal block size.

PR #1001, opened 2018-01-24
cli: fix align of defaults

ne-sted:
Also `-b#` value changed to 3 as it is by fact


Cyan4973:
Thanks for the corrections @ne-sted !

PR #1005, opened 2018-01-29
Update zstd.1

systemcrash:
bump version # here too


Cyan4973:
Sorry @systemcrash , we can't proceed with a merge if you don't sign the CLA.

If you can't or don't want to sign in, it's okay, don't worry,
these files are going to be automatically updated in a future update anyway,
as they are part of artefacts produced by `make all`.

systemcrash:
signed, but I figured it was NBD.

PR #1010, opened 2018-02-03
Updatable compression parameters

Cyan4973:
This patch makes it possible to change compression parameters *during* compression.

Only compression level and compression parameters (except `windowLog`) can be updated mid-course.
Any other parameter will trigger an error.

How it works : just invoke `ZSTD_CCtx_setParameter()` as usual. If the updated parameter belongs to the authorised list, it will be pushed to zstdmt at next invocation of `ZSTD_compress_generic()`, and then applied on next job.
Note that there is some latency while waiting for next job, especially at slower compression level.

Note also that this capability has no impact on single-threaded compression : compression parameters remain the same up to the end (`ZSTD_e_end`) of current frame.

On CLI, the compression level is now displayed at the beginning of the status line.

The capability was tested with a silly scenario that was randomly changing compression level inside `fileio.c`, to visualize its impact. The "silly scenario" is not included in this patch.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -10,7 +10,7 @@
 
 
 /* ======   Tuning parameters   ====== */
-#define ZSTDMT_NBTHREADS_MAX 200
+#define ZSTDMT_NBWORKERS_MAX 200
 #define ZSTDMT_JOBSIZE_MAX  (MEM_32bits() ? (512 MB) : (2 GB))  /* note : limited by `jobSize` type, which is `unsigned` */
 #define ZSTDMT_OVERLAPLOG_DEFAULT 6
 
@@ -97,9 +97,9 @@ typedef struct ZSTDMT_bufferPool_s {
     buffer_t bTable[1];   /* variable size */
 } ZSTDMT_bufferPool;
 
-static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(unsigned nbThreads, ZSTD_customMem cMem)
+static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(unsigned nbWorkers, ZSTD_customMem cMem)
 {
-    unsigned const maxNbBuffers = 2*nbThreads + 3;
+    unsigned const maxNbBuffers = 2*nbWorkers + 3;
     ZSTDMT_bufferPool* const bufPool = (ZSTDMT_bufferPool*)ZSTD_calloc(
         sizeof(ZSTDMT_bufferPool) + (maxNbBuffers-1) * sizeof(buffer_t), cMem);
     if (bufPool==NULL) return NULL;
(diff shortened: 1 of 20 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -664,13 +666,29 @@ static ZSTD_CCtx_params ZSTDMT_initJobCCtxParams(ZSTD_CCtx_params const params)
...
+ *  New parameters will be applied to next compression job. */
+void ZSTDMT_updateCParams_whileCompressing(ZSTDMT_CCtx* mtctx, int compressionLevel, ZSTD_compressionParameters cParams)
+{
+    U32 const saved_wlog = mtctx->params.cParams.windowLog;   /* Do not modify windowLog while compressing */
+    DEBUGLOG(5, "ZSTDMT_updateCParams_whileCompressing (level:%i)",
+                compressionLevel);
+    mtctx->params.compressionLevel = compressionLevel;
+    if (compressionLevel != ZSTD_CLEVEL_CUSTOM)
+        cParams = ZSTD_getCParams(compressionLevel, mtctx->frameContentSize, 0 /* should be dictSize */ );
+    cParams. windowLog = saved_wlog;
```
terrelln: nit: delete space after `.`.

lib/compress/zstdmt_compress.c
```diff
@@ -664,13 +666,29 @@ static ZSTD_CCtx_params ZSTDMT_initJobCCtxParams(ZSTD_CCtx_params const params)
...
+ *  while compression is ongoing.
+ *  New parameters will be applied to next compression job. */
+void ZSTDMT_updateCParams_whileCompressing(ZSTDMT_CCtx* mtctx, int compressionLevel, ZSTD_compressionParameters cParams)
+{
+    U32 const saved_wlog = mtctx->params.cParams.windowLog;   /* Do not modify windowLog while compressing */
+    DEBUGLOG(5, "ZSTDMT_updateCParams_whileCompressing (level:%i)",
+                compressionLevel);
+    mtctx->params.compressionLevel = compressionLevel;
+    if (compressionLevel != ZSTD_CLEVEL_CUSTOM)
+        cParams = ZSTD_getCParams(compressionLevel, mtctx->frameContentSize, 0 /* should be dictSize */ );
```
terrelln: What do you mean `should be dictSize`? Are you labeling the parameter, or is it a TODO?

programs/zstdcli.c
```diff
@@ -716,10 +716,10 @@ int main(int argCount, const char* argv[])
     /* Welcome message (if verbose) */
     DISPLAYLEVEL(3, WELCOME_MESSAGE);
 
-    if (nbThreads == 0) {
-        /* try to guess */
-        nbThreads = UTIL_countPhysicalCores();
-        DISPLAYLEVEL(3, "Note: %d physical core(s) detected \n", nbThreads);
+    if (nbWorkers == 0) {
```
terrelln: It looks like there isn't any way to disable async mode, given that zstd is compiled with threading support. Do we want to provide a way to do that or not?

lib/compress/zstdmt_compress.c
Cyan4973: It's a reminder that the strange corner case scenario of multi-threaded compression started from `ZSTD_compress_generic()` using a `CDict` is not very clear to me regarding compression parameters.

When there is a dictionary, it's only active for first job.
Job 1 compression parameters are enforced from `CDict`.
But since Job 1 writes the frame header, it's necessary to preserve compression parameters across all jobs, so how are they transferred ?

When invoking ZSTDMT interface directly, situation is clearer : compression parameters are extracted from `CDict`, and then passed as parameter to `ZSTD_initCStream_internal()` : 
https://github.com/facebook/zstd/blob/dev/lib/compress/zstdmt_compress.c#L975

When invoked from `ZSTD_compress_generic()`, this is less clear to me : 
initialisation invokes `ZSTD_initCStream_internal()`, but `params` seem to come from `requestedParams`, and I'm not sure if this is correctly synchronised with `CDict`.
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3177

So still some clarifications to work out.

programs/zstdcli.c
Cyan4973: That's a good point.
I'm not too sure if "disabling async mode" is a useful feature though.
I fail to find a good use case to justify another command.

lib/compress/zstdmt_compress.c
Cyan4973: OK, the way it's setup currently : 
when there is a `CDict` set, either through `ZSTD_CCtx_refCDict()` or `ZSTD_CCtx_loadDictionary()`, it's no longer possible to change any compression parameter.

So, currently, changing compression parameters "on the fly" is only possible for compression without dictionary.

That's a limitation, but not a strong one : dictionary and multi-threading have no reason to be triggered together. Data flow which can benefit from automatic compression level adaptation are large, so they don't benefit from dictionary.

programs/zstdcli.c
Cyan4973: Actually, there is a good enough reason : memory usage.
The single-thread mode currently uses much less memory than `zstdmt`.
So that could be an argument.

Cyan4973: Still looking for a good name for such command.
I looked into `pigz` and `pxz` command list, just in case there would be something similar, but found nothing.

So now, inventing yet another command, it could be : 
- `--lowmem` : to emphasize memory usage reduction
- `--single-thread` : as the name implies
- `--synchronous` : to make it opposed to "asynchronous"

The second option looks the clearer one.

Cyan4973:
added command `--single-thread`

programs/zstdcli.c
terrelln: I like `--single-thread`.

PR #1011, opened 2018-02-03
[decompress] Support BMI2

terrelln:
Initial support for BMI2 in the decompression code. Initial benchmarks show this gets all or nearly all the benefits from `-mbmi2`, more details to come later. This patch is to show what the implementation looks like.

It seems clang only benefits from the HUF BMI2, and gcc benefits from HUF and FSE.

## Benchmarks

Comparing zstd decompression on Silesia before, with `-mbmi2`, and after on levels 1 and 2 with different compilers.

### MacBook Pro

```
zstd.before.apple
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 295.0 MB/s , 678.0 MB/s
 2#silesia.tar       : 211988480 ->  70158789 (3.022), 225.7 MB/s , 623.6 MB/s
zstd.after.apple
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 295.9 MB/s , 716.7 MB/s
 2#silesia.tar       : 211988480 ->  70158789 (3.022), 218.6 MB/s , 660.7 MB/s
zstd.mbmi2.apple
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 326.5 MB/s , 691.7 MB/s
 2#silesia.tar       : 211988480 ->  70158789 (3.022), 237.1 MB/s , 640.7 MB/s

zstd.before.gcc-6
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 285.3 MB/s , 714.1 MB/s
 2#silesia.tar       : 211988480 ->  70158789 (3.022), 216.5 MB/s , 677.7 MB/s
zstd.after.gcc-6
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 288.8 MB/s , 768.3 MB/s
 2#silesia.tar       : 211988480 ->  70158789 (3.022), 216.5 MB/s , 707.9 MB/s
zstd.mbmi2.gcc-6
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651689 (2.878), 301.3 MB/s , 773.6 MB/s
 2#silesia.tar       : 211988480 ->  70158789 (3.022), 228.8 MB/s , 699.6 MB/s
```

## Haswell Devserver (with bmi2)

```
zstd.before.clang
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 249.5 MB/s , 601.8 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 186.4 MB/s , 521.9 MB/s
zstd.after.clang
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 252.8 MB/s , 638.7 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 188.9 MB/s , 548.1 MB/s
zstd.mbmi2.clang
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 282.9 MB/s , 660.7 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 203.3 MB/s , 557.8 MB/s

zstd.before.gcc-7.2
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 262.5 MB/s , 723.8 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 193.6 MB/s , 633.8 MB/s
zstd.after.gcc-7.2
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 260.6 MB/s , 843.4 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 191.2 MB/s , 716.9 MB/s
zstd.mbmi2.gcc-7.2
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 273.4 MB/s , 816.6 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 200.8 MB/s , 712.7 MB/s
```

## Sandy Bridge Devserver (without bmi2)

```
zstd.before.clang
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 208.1 MB/s , 514.1 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 168.5 MB/s , 452.5 MB/s
zstd.after.clang
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 208.4 MB/s , 503.6 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 167.6 MB/s , 443.6 MB/s

zstd.before.gcc-7.2
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 214.0 MB/s , 623.2 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 167.4 MB/s , 554.3 MB/s
zstd.after.gcc-7.2
Benchmarking levels from 1 to 2
 1#silesia.tar       : 211988480 ->  73651330 (2.878), 214.1 MB/s , 622.5 MB/s
 2#silesia.tar       : 211988480 ->  70158493 (3.022), 165.1 MB/s , 554.1 MB/s
```

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -63,6 +63,24 @@
 #  endif
 #endif
 
+/* target attribute */
+#if defined(__GNUC__)
+#  define TARGET_ATTRIBUTE(target) __attribute__((__target__(target)))
+#else
+#  define TARGET_ATTRIBUTE(target)
+#endif
+
+/* Enable runtime BMI2 dispatch based on the CPU.
+ * Enabled for clang/gcc on x86 when BMI2 isn't enabled by default.
+ */
+#ifndef DYNAMIC_BMI2
+  #if defined(__GNUC__) && (defined(__x86_64__) || defined(_M_X86)) && !defined(__BMI2__)
+  #  define DYNAMIC_BMI2 1
+  #else
+  #  define DYNAMIC_BMI2 0
+  #endif
+#endif
+
 /* prefetch */
 #if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_I86))  /* _mm_prefetch() is not defined outside of x86/x64 */
 #  include <mmintrin.h>   /* https://msdn.microsoft.com/fr-fr/library/84szxsww(v=vs.90).aspx */
--- a/lib/decompress/huf_decompress.c
+++ b/lib/decompress/huf_decompress.c
@@ -144,72 +144,54 @@ size_t HUF_readDTableX2(HUF_DTable* DTable, const void* src, size_t srcSize)
                                  workSpace, sizeof(workSpace));
 }
 
+typedef struct { U16 sequence; BYTE nbBits; BYTE length; } HUF_DEltX4;  /* double-symbols decoding */
 
-static BYTE HUF_decodeSymbolX2(BIT_DStream_t* Dstream, const HUF_DEltX2* dt, const U32 dtLog)
-{
-    size_t const val = BIT_lookBitsFast(Dstream, dtLog); /* note : dtLog >= 1 */
-    BYTE const c = dt[val].byte;
-    BIT_skipBits(Dstream, dt[val].nbBits);
-    return c;
-}
+#define FUNCTION(fn) fn##_default
+#define TARGET
+#include "huf_decompress_impl.h"
+#undef TARGET
+#undef FUNCTION
 
-#define HUF_DECODE_SYMBOLX2_0(ptr, DStreamPtr) \
-    *ptr++ = HUF_decodeSymbolX2(DStreamPtr, dt, dtLog)
+#if DYNAMIC_BMI2
 
-#define HUF_DECODE_SYMBOLX2_1(ptr, DStreamPtr) \
-    if (MEM_64bits() || (HUF_TABLELOG_MAX<=12)) \
-        HUF_DECODE_SYMBOLX2_0(ptr, DStreamPtr)
+#define FUNCTION(fn) fn##_bmi2
+#define TARGET TARGET_ATTRIBUTE("bmi2")
+#include "huf_decompress_impl.h"
+#undef TARGET
+#undef FUNCTION
 
-#define HUF_DECODE_SYMBOLX2_2(ptr, DStreamPtr) \
-    if (MEM_64bits()) \
-        HUF_DECODE_SYMBOLX2_0(ptr, DStreamPtr)
+#endif
 
-HINT_INLINE size_t HUF_decodeStreamX2(BYTE* p, BIT_DStream_t* const bitDPtr, BYTE* const pEnd, const HUF_DEltX2* const dt, const U32 dtLog)
-{
-    BYTE* const pStart = p;
-
-    /* up to 4 symbols at a time */
-    while ((BIT_reloadDStream(bitDPtr) == BIT_DStream_unfinished) && (p <= pEnd-4)) {
-        HUF_DECODE_SYMBOLX2_2(p, bitDPtr);
-        HUF_DECODE_SYMBOLX2_1(p, bitDPtr);
-        HUF_DECODE_SYMBOLX2_2(p, bitDPtr);
-        HUF_DECODE_SYMBOLX2_0(p, bitDPtr);
+typedef size_t (*HUF_decompress_usingDTable_t)(void *dst, size_t dstSize,
+                                               const void *cSrc,
+                                               size_t cSrcSize,
+                                               const HUF_DTable *DTable);
+#if DYNAMIC_BMI2
+#define X(fn)                                                                  \
+    static size_t fn(void* dst, size_t dstSize, void const* cSrc,              \
+                     size_t cSrcSize, HUF_DTable const* DTable, int bmi2)      \
+    {                                                                          \
+        if (bmi2) {                                                            \
+            return fn##_bmi2(dst, dstSize, cSrc, cSrcSize, DTable);            \
+        }                                                                      \
+        return fn##_default(dst, dstSize, cSrc, cSrcSize, DTable);             \
     }
+#else
+#define X(fn)                                                                  \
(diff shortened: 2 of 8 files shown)


lib/decompress/huf_decompress_impl.h
```diff
@@ -0,0 +1,365 @@
...
+#ifndef FUNCTION_NAME
+#  error "FUNCTION_NAME(name) must be defined"
+#endif
+
+#ifndef TARGET
+#  error "TARGET must be defined"
+#endif
+
+
+static TARGET BYTE FUNCTION_NAME(HUF_decodeSymbolX2)(BIT_DStream_t* Dstream, const HUF_DEltX2* dt, const U32 dtLog)
```
Cyan4973: One suggestion : 
would that produce the same optimization with a construction like that : 

```
static type HUF_decodeX() {  (...) }   // normal function, default compilation flags

#ifdef DYNAMIC_BMI2
static TARGET_ATTRIBUTE("bmi2") type HUF_decodeX_bmi2() {
   return HUF_decodeX();   // Ensure it's inlined
}
#endif
```

It would avoid creation of an additional file to include several times, which is also harder to parse for editors.
But it's unclear if it would produce the same gains.

terrelln: I can try it

terrelln: Sadly I didn't get the same gains using that method.

terrelln:
I've added benchmarks, and changed from returning function pointers to just taking an extra `const int bmi2` parameter. I'll make the compression changes in a separate PR.

The only regression is clang compiled zstd on a server without bmi2 support. I'm not too worried, since clang compiles the Huffman code pretty terribly anyways. However, if you think it is worth pursuing, I'll look into it further.

lib/common/compiler.h
```diff
@@ -63,6 +63,21 @@
...
 
+/* target attribute */
+#if defined(__GNUC__)
+#  define TARGET_ATTRIBUTE(target) __attribute__((__target__(target)))
+#else
+#  define TARGET_ATTRIBUTE(target)
+#endif
+
+#ifndef DYNAMIC_BMI2
+  #if defined(__GNUC__) && (defined(__x86_64__) || defined(_M_X86))
```
Cyan4973: probably add `&& !defined(__BMI2__)`
as in this case, the base code path is already bmi2, so there is no need for a dispatch

Cyan4973:
Some questions : 
- Is the scope of this patch limited to Huffman decoder ?
  + if yes, is there a plan regarding FSE decoder ?
- Regarding benchmark : do I read correctly that your patch produces higher decoding performance than blindly enabling `-mbmi2` on some targets, namely `apple-clang`, and possibly `apple-gcc` and `haswell-gcc` ? If yes, do you have an explanation ?

terrelln:
* The patch changes the Huffman decoder and `ZSTD_decompressSequences(Long)` and functions it calls, which covers the FSE decoder.
* The difference in gcc is noise, but `apple-clang` is faster with the patch than with `-mbmi2`. I'm not sure why...

Cyan4973:
I would have expected the `BIT_*` functions to also benefit from `bmi2`.
I see there are no changes to these functions.

Is that because , since they are all `inline` functions, they benefit from the `bmi2` instruction set just by virtue of being inlined ?

lib/decompress/huf_decompress.c
```diff
@@ -218,7 +200,7 @@ size_t HUF_decompress1X2_usingDTable(
 {
     DTableDesc dtd = HUF_getDTableDesc(DTable);
     if (dtd.tableType != 0) return ERROR(GENERIC);
-    return HUF_decompress1X2_usingDTable_internal(dst, dstSize, cSrc, cSrcSize, DTable);
+    return HUF_decompress1X2_usingDTable_internal(dst, dstSize, cSrc, cSrcSize, DTable, /* bmi2 */ 0);
 }
 
 size_t HUF_decompress1X2_DCtx_wksp(HUF_DTable* DCtx, void* dst, size_t dstSize,
```
Cyan4973: I guess this version is no longer used within `zstd` source code, replaced by the newer `HUF_decompress1X2_DCtx_wksp_bmi2()` ?

terrelln: Yeah, we could delete it. I'm not exactly sure what we want to keep around as part of the Huffman module, since a lot of functions are unused in zstd.

terrelln:
If a function is inlined into the bmi2 target then it will also be compiled with bmi2 support. Since we are getting nearly all the benefit, the `BIT_*` functions must be inlined, which makes sense.

terrelln:
Clang is still a bit slower, but the gap is now ~10 MB/s instead of 30 MB/s.

lib/decompress/huf_decompress.c
Cyan4973: That's a good point.
The initial "vision" is that `zstd` would be a client program of `fse` and `hufff0`, and would therefore use the files from these projects unmodified.

This is less and less the case. While `zstd` is actively maintained, `fse` is falling behind, due to lack of attention. Versions start to diverge, we add too many specialized capabilities which are mostly useful for `zstd`, and less for a "generic" entropy interface.

So maybe at some point, it will be better to bite the bullet, and accept that they are now 2 separate code bases, with no coming back.
At this stage, a lot of cleaning will be possible.

PR #1014, opened 2018-02-11
Faster decoding speed

Cyan4973:
A little patch written under pain killer medicine. Hopefully, not too insane.
This one offers a generic +5-10% decoding speed improvement to most sources I benchmarked under `gcc`.

The cost is that fse decoding tables are now twice larger, inflating from 5kb to 10kb.
I believe it's a reasonable trade-off.
All related fse decoding function had to be rewritten to adopt the new data structure.
They are included directly within `zstd_decompress.c`.

A few numbers from "entropy" : 
before : 
```
./zstd -b silesia.tar enwik8 calgary.tar -S
 3#silesia.tar       : 211948032 ->  66994707 (3.164), 262.3 MB/s ,1029.8 MB/s
 3#enwik8            : 100000000 ->  35687378 (2.802), 192.6 MB/s , 832.2 MB/s
 3#calgary.tar       :   3265536 ->   1065161 (3.066), 214.4 MB/s , 921.7 MB/s
```
after :
```
./zstd -b silesia.tar enwik8 calgary.tar -S
 3#silesia.tar       : 211948032 ->  66994707 (3.164), 259.8 MB/s ,1110.8 MB/s
 3#enwik8            : 100000000 ->  35687378 (2.802), 190.7 MB/s , 923.0 MB/s
 3#calgary.tar       :   3265536 ->   1065161 (3.066), 212.7 MB/s ,1019.5 MB/s
```

Gains are cumulative with other improvements, such as `-mbmi2`.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1277,11 +1277,8 @@ static size_t ZSTD_decompressSequences(
         dctx->fseEntropy = 1;
         { U32 i; for (i=0; i<ZSTD_REP_NUM; i++) seqState.prevOffset[i] = dctx->entropy.rep[i]; }
         CHECK_E(BIT_initDStream(&seqState.DStream, ip, iend-ip), corruption_detected);
-        DEBUGLOG(6, "ZSTD_decompressSequences: init stateLL");
         ZSTD_initFseState(&seqState.stateLL, &seqState.DStream, dctx->LLTptr);
-        DEBUGLOG(6, "ZSTD_decompressSequences: init stateOffb");
         ZSTD_initFseState(&seqState.stateOffb, &seqState.DStream, dctx->OFTptr);
-        DEBUGLOG(6, "ZSTD_decompressSequences: init stateML");
         ZSTD_initFseState(&seqState.stateML, &seqState.DStream, dctx->MLTptr);
 
         for ( ; (BIT_reloadDStream(&(seqState.DStream)) <= BIT_DStream_completed) && nbSeq ; ) {
@@ -1294,7 +1291,7 @@ static size_t ZSTD_decompressSequences(
         }   }
 
         /* check if reached exact end */
-        DEBUGLOG(5, "after decode loop, remaining nbSeq : %i", nbSeq);
+        DEBUGLOG(5, "ZSTD_decompressSequences: after decode loop, remaining nbSeq : %i", nbSeq);
         if (nbSeq) return ERROR(corruption_detected);
         /* save reps for next block */
         { U32 i; for (i=0; i<ZSTD_REP_NUM; i++) dctx->entropy.rep[i] = (U32)(seqState.prevOffset[i]); }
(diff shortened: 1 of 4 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -80,20 +80,33 @@ typedef enum { ZSTDds_getFrameHeaderSize, ZSTDds_decodeFrameHeader,
...
+    BYTE nbAdditionalBits;
+    BYTE nbBits;
+    U32  baseValue;
+} ZSTD_seqSymbol;
+
 typedef struct {
-    FSE_DTable LLTable[FSE_DTABLE_SIZE_U32(LLFSELog)];
-    FSE_DTable OFTable[FSE_DTABLE_SIZE_U32(OffFSELog)];
-    FSE_DTable MLTable[FSE_DTABLE_SIZE_U32(MLFSELog)];
+    ZSTD_seqSymbol LLTable[FSE_DTABLE_SIZE_U32(LLFSELog)];
```
terrelln: Is `FSE_DTABLE_SIZE_U32(LLFSELog)` still the correct size? We'll probably want a new macro

```c
#define ZSTD_SEQTABLE_SIZE(FSELog) FSE_DTABLE_SIZE_U32((FSELog))
```

terrelln:
If I understand correctly, you are just including the base value in the table, right?

Cyan4973:
right : `symbol -> additionalBits, baseValue`.
`additionalBits` takes the seat of `symbol`, so the increased size comes from adding `baseValue` to the table.

There might be some way to reduce this increase by `+2 bytes` instead of `+4 bytes`, but I was worried of increased complexity, and figured that 10KB remained a reasonable budget for tables.

lib/decompress/zstd_decompress.c
Cyan4973: Yes, it's the correct size.
Yes, adding a new macro is a nonetheless cleaner way to do this.

lib/decompress/zstd_decompress.c
```diff
@@ -652,98 +667,210 @@ size_t ZSTD_decodeLiteralsBlock(ZSTD_DCtx* dctx,
...
-    { {  0,  8,  5 } }, { { 32, 10,  5 } }, { {  0, 11,  5 } }, { {  0, 13,  6 } },
-    { { 32, 16,  5 } }, { {  0, 17,  5 } }, { { 32, 19,  5 } }, { {  0, 20,  5 } },
-    { { 32, 22,  5 } }, { {  0, 23,  5 } }, { {  0, 25,  4 } }, { { 16, 25,  4 } },
-    { { 32, 26,  5 } }, { {  0, 28,  6 } }, { {  0, 30,  6 } }, { { 48,  0,  4 } },
-    { { 16,  1,  4 } }, { { 32,  2,  5 } }, { { 32,  3,  5 } }, { { 32,  5,  5 } },
-    { { 32,  6,  5 } }, { { 32,  8,  5 } }, { { 32,  9,  5 } }, { { 32, 11,  5 } },
-    { { 32, 12,  5 } }, { {  0, 15,  6 } }, { { 32, 17,  5 } }, { { 32, 18,  5 } },
-    { { 32, 20,  5 } }, { { 32, 21,  5 } }, { { 32, 23,  5 } }, { { 32, 24,  5 } },
-    { {  0, 35,  6 } }, { {  0, 34,  6 } }, { {  0, 33,  6 } }, { {  0, 32,  6 } },
+static const ZSTD_seqSymbol LL_defaultDTable[(1<<LL_DEFAULTNORMLOG)+1] = {
```
terrelln: Do we have tests that will find bugs/typos in these arrays? If I change a random entry here, are our existing tests likely to find the bug?

Cyan4973: These tables were generated by a program from default distribution stats, precisely to avoid typo.

> If I change a random entry here, are our existing tests likely to find the bug?

Good point, I haven't checked.
I suspect that it will be caught indirectly by the fuzzer tests, as the round trip test using the damaged table will likely fail. But not sure how much time will be required to generate the failing test. 

Furthermore, debugging the source of the problem (being the damaged table) is going to be even harder. But I guess this is a secondary issue.

terrelln: Can you check in the program that created them, and comment here about how to invoke it, in case someone else needs to change it in the future?

Cyan4973: The generating program is no more : it was just a temporary thing.
But I could explain how to create it, it's not very difficult.

Cyan4973:
Added a code comment explaining how the default distribution tables are generated.

Also : tested multiple manual modifications of the default tables.
Good news : `fuzzer` catch these errors in short order : often directly during unit tests, those that pass are caught after a mere dozens of rounds. 

note: I believe it's possible to manufacture some changes in a way that `fuzzer` doesn't test, typically by modifying a table entry corresponding to very large offsets, which will never be triggered by `fuzzer`.
But for the purpose of this code modification, it answer most of our concerns : a wrong table is unlikely to survive `fuzzer` tests. And then we also have `playtests.sh`, which checks very large files, and then `roundtripcrash`, for in-depth `afl` tests. So that's a lot of checkpoints.

PR #1015, opened 2018-02-14
Add the packaging metadata to build the zstd snap

elopio:
This package will let you publish the latest zstd in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and Linux distributions. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,28 @@
+name: zstd
+version: git
+summary: Zstandard - Fast real-time compression algorithm
+description: |
+  Zstandard, or zstd as short version, is a fast lossless compression
+  algorithm, targeting real-time compression scenarios at zlib-level and better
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for the suggestion and patch @elopio !

Does it need to be in its own `snap/` directory at root, 
or could it be moved to another directory, like `contrib/snap/` ?

elopio:
Hey, thanks for looking at this pr so quickly :)
If you want to use build.snapcraft.io for the continuous build and delivery, the metadata must be in snap/snapcraft.yaml.
If you want to put it somewhere else, you will have to use travis, circle ci or some other system to build and push the snap to the store.

Cyan4973:
I can accept the patch, but will need to move the new directory into `/contrib`.
For clarity, it's preferable to limit the nb of directories and files at root.

PR #1021, opened 2018-02-22
Split block compresser out of long range matcher

terrelln:
* `ZSTD_ldm_generateSequences()` generates the LDM sequences and
  stores them in a table. It should work with any chunk size, but
  is currently only called one block at a time.
* `ZSTD_ldm_blockCompress()` emits the pre-defined sequences, and
  instead of encoding the literals directly, it passes them to a
  secondary block compressor. The code to handle chunk sizes greater
  than the block size is currently commented out, since it is unused.
  The next PR will uncomment exercise this code.
* During optimal parsing, ensure LDM `minMatchLength` is at least
  `targetLength`. Also don't emit repcode matches in the LDM block
  compressor. Enabling the LDM with the optimal parser now actually improves
  the compression ratio.
* The compression ratio is about 0.01% worse than before, because I removed the
  immediate repcode checking. With the immediate repcode checking it is very
  slightly different, because the repcode handling is slightly different. If I remove
  immediate repcode checking in both branches the compressed size is exactly
  the same. I removed the repcode checking in the 2nd commit, so I can add it back
  in easily if we want to.
* The speed looks to be the same or better than before.

Up Next (in a separate PR)
--------------------------

Allow sequence generation to happen prior to compression, and produce more
than a block worth of sequences. Expose some API for zstdmt to consume.
This will test out some currently untested code in
`ZSTD_ldm_blockCompress()`.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -718,12 +718,11 @@ size_t ZSTD_estimateCCtxSize_usingCCtxParams(const ZSTD_CCtx_params* params)
         size_t const blockStateSpace = 2 * sizeof(ZSTD_compressedBlockState_t);
         size_t const matchStateSize = ZSTD_sizeof_matchState(&params->cParams, /* forCCtx */ 1);
 
-        size_t const ldmSpace = params->ldmParams.enableLdm ?
-            ZSTD_ldm_getTableSize(params->ldmParams.hashLog,
-                                  params->ldmParams.bucketSizeLog) : 0;
+        size_t const ldmSpace = ZSTD_ldm_getTableSize(params->ldmParams);
+        size_t const ldmSeqSpace = ZSTD_ldm_getMaxNbSeq(params->ldmParams, blockSize) * sizeof(rawSeq);
 
         size_t const neededSpace = entropySpace + blockStateSpace + tokenSpace +
-                                   matchStateSize + ldmSpace;
+                                   matchStateSize + ldmSpace + ldmSeqSpace;
 
         DEBUGLOG(5, "sizeof(ZSTD_CCtx) : %u", (U32)sizeof(ZSTD_CCtx));
         DEBUGLOG(5, "estimate workSpace : %u", (U32)neededSpace);
@@ -990,7 +989,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
 
     if (params.ldmParams.enableLdm) {
         /* Adjust long distance matching parameters */
-        ZSTD_ldm_adjustParameters(&params.ldmParams, params.cParams.windowLog);
+        ZSTD_ldm_adjustParameters(&params.ldmParams, &params.cParams);
         assert(params.ldmParams.hashLog >= params.ldmParams.bucketSizeLog);
         assert(params.ldmParams.hashEveryLog < 32);
         zc->ldmState.hashPower =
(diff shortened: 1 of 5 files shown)


lib/compress/zstd_ldm.h
```diff
@@ -37,21 +37,58 @@ extern "C" {
...
+    ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
+    ZSTD_compressionParameters const* cParams, void const* src, size_t srcSize,
+    int const extDict);
+
 
 /** ZSTD_ldm_initializeParameters() :
  *  Initialize the long distance matching parameters to their default values. */
 size_t ZSTD_ldm_initializeParameters(ldmParams_t* params, U32 enableLdm);
 
 /** ZSTD_ldm_getTableSize() :
```
Cyan4973: Could you document that `ZSTD_ldm_getTableSize()` is able to detect if Long Range mode is triggered or not, and appropriately reports a size of `0` when it is not ? 
It seems to be a change of scope.

lib/compress/zstd_ldm.h
```diff
@@ -37,21 +37,58 @@ extern "C" {
...
-size_t ZSTD_ldm_getTableSize(U32 hashLog, U32 bucketSizeLog);
+size_t ZSTD_ldm_getTableSize(ldmParams_t params, size_t maxChunkSize);
+
+/** ZSTD_ldm_getMaxNbSeq() :
+ *  Return an upper bound on the number of sequences that can be produced by
+ *  the long distance matcher.
+ */
+size_t ZSTD_ldm_getMaxNbSeq(size_t maxChunkSize, U32 minMatchLength);
 
 /** ZSTD_ldm_getTableSize() :
```
Cyan4973: The name of this function looks incorrect.

lib/compress/zstd_compress.c
```diff
@@ -1820,34 +1820,41 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
...
+                                           ms, &zc->appliedParams.ldmParams,
+                                           src, srcSize, extDict);
+            size_t nbSeq = nbStoredSeq;
+            lastLLSize =
+                ZSTD_ldm_blockCompress(zc->ldmSequences, &nbSeq,
+                                       ms, &zc->seqStore,
+                                       zc->blockState.nextCBlock->rep,
+                                       &zc->appliedParams.cParams,
+                                       src, srcSize, extDict);
+            assert(nbSeq == nbStoredSeq /* Every stored sequence should be executed */);
```
Cyan4973: This check is surprising at this place.

While the documentation states that "`*nbSeq` is set to the number of sequences consumed", `ZSTD_ldm_blockCompress()` doesn't seem to modify this value anyway.

Is that related to "handle chunk sizes greater than the block size" ?

With current scope, it seems clearer to just pass `nbSeq` as a normal (read-copy) parameter, and make this check (if it's useful) inside `ZSTD_ldm_blockCompress()` instead. Thoughts ?

terrelln: Yeah, I can do that, and then revert it in the next PR.

The next thing I need to do on the LDM side is allow `zstdmt` to generate the LDM sequences for the whole chunk while holding a lock, and then compress blocks after that.

lib/compress/zstd_ldm.c
```diff
@@ -375,280 +420,151 @@ size_t ZSTD_compressBlock_ldm(
...
-        ldmState_t* ldmState, ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
-        ZSTD_CCtx_params const* params, void const* src, size_t srcSize)
+/**
+ * If the sequence length is longer than remaining then the sequence is split
+ * between this block and the next.
+ *
+ * Returns the current sequence to handle, or if the rest of the block should
+ * be literals, it returns a sequence with offset == 0.
+ */
+static rawSeq maybeSplitSequence(rawSeq* sequences, size_t* nbSeq,
```
Cyan4973: I guess this is for chunk sizes > block size ?
So this code is not useful yet ?

lib/compress/zstd_ldm.c
```diff
@@ -375,280 +420,151 @@ size_t ZSTD_compressBlock_ldm(
...
+    return sequences - sequencesStart;
 }
 
-size_t ZSTD_compressBlock_ldm_extDict(
-        ldmState_t* ldmState, ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
-        ZSTD_CCtx_params const* params, void const* src, size_t srcSize)
+/**
+ * If the sequence length is longer than remaining then the sequence is split
+ * between this block and the next.
+ *
```
Cyan4973: Another strategy could be to reduce block size, ending where the "long match" starts ?
This avoids splitting a match between 2 blocks, and would likely compress a little bit better,
though if `matchSize` is larger `blockSize`, it will still be necessary to split it .

lib/compress/zstd_ldm.c
```diff
@@ -41,12 +50,22 @@ void ZSTD_ldm_adjustParameters(ldmParams_t* params, U32 windowLog)
...
+    size_t const maxNbSeq =
+        ZSTD_ldm_getMaxNbSeq(maxChunkSize, params.minMatchLength);
+    size_t const totalSize = ldmBucketSize + ldmHSize * sizeof(ldmEntry_t) +
+                             maxNbSeq * sizeof(rawSeq);
+    return params.enableLdm ? totalSize : 0;
+}
+
+size_t ZSTD_ldm_getMaxNbSeq(size_t maxChunkSize, U32 minMatchLength)
+{
+    return maxChunkSize / minMatchLength;
```
Cyan4973: What's the value of `minMatchLength` when `params.enableLdm==0` ?
Maybe add `assert(minMatchLength > 0)` too ?

Cyan4973:
The new strategy, uncoupling sequence generation from block compression, looks pretty good.
It will be easier to combine for future evolutions.
There are still a few tests to pass, but I'm confident it can be done.

That LDM now works "in addition to" optimal parser, and improves, rather than degrade, total compression ratio, is very good news. Compression enthusiasts will like it.

Compression ratio is worse by 0.01% for levels below optimal parser, but that's a region where we can more easily compromise on ratio in favor of speed, modularity or performance. So it feels acceptable.
I'm just slightly surprised that the block compressor's own repcode detector doesn't produce the same result. Maybe it finds shorter matches that interfere with repcode matches ? I just want to be sure that it can actually find the repcode when it's the only match to find.

lib/compress/zstd_ldm.c
terrelln: That could be possible, though then that would have to be handled by whoever calls `ZSTD_compressBlock_internal()`, or we'll have to bubble up the information that we didn't actually compress the last few bytes through a few layers.

Since we have to handle the splitting when `matchSize > blockSize` anyways, I'd rather keep the complexity internal to the function.

We can look at it again once the code is actually needed, until then I'll `#if 0` it out.

terrelln:
Once this is merged I am ready to put up a PR refactoring the window management code.

PR #1022, opened 2018-02-22
Implemented BMI2 functions directly within huf_decompress.c

Cyan4973:
This makes it easier to edit for maintenance and evolutions
(I plan to experiment modifications in huffman decompression functions).

The methology followed seems broadly applicable to other BMI2 modules.

Performance was tracked rigorously at each step,
there is no noticeable loss (nor win) of performance compared to `#include` version.

Note however that 4X decoder variants tend to be extremely sensitive to code alignment.
This source code resulted in pretty good performance for gcc 7.2 and 7.3,
but future changes (even in other parts of the code) might trigger the issue again.


Cyan4973:
- External changes within `huf_decompress.c` file : 
   This happened a few times with `HUF_decompress4X4_usingDTable()` : adding or removing dummy functions before the main function body could produce swings of up to 30%. 
   This particular issue should be fixable with `-falign-functions=#`, 
   though the current version of source code does not need it.

 - Internal changes within function.
    I'm currently trying to optimize with `HUF_decompress4X2_usingDTable_internal_body()`, and it's very annoying, with swing in the 10% range following minor changes which should be performance neutral.
    This one is more problematic, as I don't know any good way to be shielded from such impact.

PR #1024, opened 2018-02-24
Split the window state into substructure

terrelln:
This commit splits the window state into a substructure, and factors out operations on the window. There should be no functional change.

The next PR, which gives LDM its own window management logic, is ready. I split it up into two PRs for ease of review, so I'll submit it once this is merged. For reference, the branch is [lrm-window](https://github.com/terrelln/zstd/tree/lrm-window).


lib/compress/zstd_compress.c
```diff
@@ -2145,15 +2091,8 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
     ZSTD_compressionParameters const* cParams = &params->cParams;
 
-    /* input becomes current prefix */
-    ms->lowLimit = ms->dictLimit;
-    ms->dictLimit = (U32)(ms->nextSrc - ms->base);
-    ms->dictBase = ms->base;
-    ms->base = ip - ms->dictLimit;
-    ms->nextToUpdate = ms->dictLimit;
-    ms->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ms->base);
+    ZSTD_window_update(&ms->window, src, srcSize);
```
terrelln: I'm pretty sure this is correct, but could you double check?

lib/compress/zstd_compress_internal.h
```diff
@@ -441,6 +445,132 @@ MEM_STATIC size_t ZSTD_hashPtr(const void* p, U32 hBits, U32 mls)
...
+
+/**
+ * ZSTD_window_clear():
+ * Clears the window containing the history by simply setting it to empty.
+ */
+MEM_STATIC void ZSTD_window_clear(ZSTD_window_t* window)
+{
+    size_t const endT = (size_t)(window->nextSrc - window->base);
+    U32 const end = (U32)endT;
+    assert(endT < (3U<<30));
```
Cyan4973: Is the constant `(3U<<30)` used elsewhere ?
If yes, could it be named ?

lib/compress/zstd_compress_internal.h
```diff
@@ -441,6 +445,132 @@ MEM_STATIC size_t ZSTD_hashPtr(const void* p, U32 hBits, U32 mls)
...
+}
+
+/**
+ * ZSTD_window_needOverflowCorrection():
+ * Returns non-zero if the indices are getting too large and need overflow
+ * protection.
+ */
+MEM_STATIC U32 ZSTD_window_needOverflowCorrection(ZSTD_window_t const window)
+{
+    return window.lowLimit > (3U << 29);
```
Cyan4973: Same kind of comment : 
could this constant be named ?
Relation with previous constant  `(3U << 30)` ? Should they be explicitly related (aka, one derive from the other) ?

lib/compress/zstd_compress_internal.h
```diff
@@ -441,6 +445,132 @@ MEM_STATIC size_t ZSTD_hashPtr(const void* p, U32 hBits, U32 mls)
...
+MEM_STATIC U32 ZSTD_window_correctOverflow(ZSTD_window_t* window, U32 cycleLog,
+                                           U32 maxDist, void const* src)
+{
+    U32 const cycleMask = (1U << cycleLog) - 1;
+    U32 const current = (U32)((BYTE const*)src - window->base);
+    U32 const newCurrent = (current & cycleMask) + maxDist;
+    U32 const correction = current - newCurrent;
+    assert((maxDist & cycleMask) == 0);
+    assert(current > newCurrent);
+    assert(correction > 1<<28); /* Loose bound, should be about 1<<29 */
```
Cyan4973: Why is this `assert()` required ?
Could the function work properly if `correction < (1<<28)` ? Or are there side-effects which are too important to neglect ?

lib/compress/zstd_compress_internal.h
```diff
@@ -441,6 +445,132 @@ MEM_STATIC size_t ZSTD_hashPtr(const void* p, U32 hBits, U32 mls)
...
+             window->lowLimit);
+    return correction;
+}
+
+/**
+ * ZSTD_window_enforceMaxDist():
+ * Sets lowLimit such that indices earlier than (src - base) - lowLimit are
+ * invalid. This allows a simple check index >= lowLimit to see if it is valid.
+ */
+MEM_STATIC void ZSTD_window_enforceMaxDist(ZSTD_window_t* window,
```
Cyan4973: Could you indicate in code comments that `src` must be set at the _end_ of the target memory segment (`ip + blockSize`) ?

lib/compress/zstd_compress.c
Cyan4973: It looks fine to me.

Cyan4973:
The patch itself looks fine to me.
It's a bit complex to follow at times, because it modifies many little things in many places.
I mostly asked for a few code comments to help the reader.

lib/compress/zstd_compress_internal.h
terrelln: Its not strictly required. There isn't a problem if it is less than `1<<28`, but there would potentially be a problem if it is less than the block size, and a severe performance penalty if it is not large enough to happen infrequently. I'll say basically that in a comment.

lib/compress/zstd_compress_internal.h
terrelln: Yes, I'll call the parameter `srcEnd`.

terrelln:
I had to reduce `ZSTD_CHAINLOG_MAX_32` to 29, while leaving `ZSTD_CHAINLOG_MAX_64` alone. After correction `newCurrent < 1<<windowLog + 1<<chainLog <= 1<<32-1`, which leaves no room for the next block, without overflowing the `ptrdiff_t`.

Reducing the maximum chain log to 29 gives us a maximum safe chunk size of `1<<29` before we need to call the overflow protection again.

terrelln:
I have one more patch for the LDM once this lands. Then I'm changing zstdmt to use a single buffer. Then I'll come back to the LDM to expose an API to use an external `ZSTD_ldmState_t`, and use it in zstdmt.

PR #1026, opened 2018-02-27
LDM manages its own window round buffer

terrelln:
* Give the LDM its own `windowLog` parameter, which can't yet be set to anything other than the `ZSTD_CCtx`'s window log.
* Give the LDM its own round buffer management, separate from the `ZSTD_CCtx`'s.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -983,6 +983,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
 
     if (params.ldmParams.enableLdm) {
         /* Adjust long distance matching parameters */
+        params.ldmParams.windowLog = params.cParams.windowLog;
         ZSTD_ldm_adjustParameters(&params.ldmParams, &params.cParams);
         assert(params.ldmParams.hashLog >= params.ldmParams.bucketSizeLog);
         assert(params.ldmParams.hashEveryLog < 32);
@@ -1067,6 +1068,8 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
             ptr = zc->ldmState.hashTable + ldmHSize;
             zc->ldmSequences = (rawSeq*)ptr;
             ptr = zc->ldmSequences + maxNbLdmSeq;
+
+            memset(&zc->ldmState.window, 0, sizeof(zc->ldmState.window));
         }
         assert(((size_t)ptr & 3) == 0); /* ensure ptr is properly aligned */
 
@@ -1277,19 +1280,6 @@ static void ZSTD_reduceTable_btlazy2(U32* const table, U32 const size, U32 const
     ZSTD_reduceTable_internal(table, size, reducerValue, 1);
 }
 
-
-/*! ZSTD_ldm_reduceTable() :
- *  reduce table indexes by `reducerValue` */
-static void ZSTD_ldm_reduceTable(ldmEntry_t* const table, U32 const size,
-                                 U32 const reducerValue)
-{
-    U32 u;
-    for (u = 0; u < size; u++) {
-        if (table[u].offset < reducerValue) table[u].offset = 0;
-        else table[u].offset -= reducerValue;
-    }
-}
-
 /*! ZSTD_reduceIndex() :
 *   rescale all indexes to avoid future overflow (indexes are U32) */
 static void ZSTD_reduceIndex (ZSTD_CCtx* zc, const U32 reducerValue)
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -983,6 +983,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
 
     if (params.ldmParams.enableLdm) {
         /* Adjust long distance matching parameters */
+        params.ldmParams.windowLog = params.cParams.windowLog;
```
Cyan4973: Do you foresee, one day, the possibility for these 2 values to be different ?
Should they always be identical ?

lib/compress/zstd_compress_internal.h
```diff
@@ -148,6 +149,7 @@ typedef struct {
     U32 bucketSizeLog;      /* Log bucket size for collision resolution, at most 8 */
     U32 minMatchLength;     /* Minimum match length */
     U32 hashEveryLog;       /* Log number of entries to skip */
+    U32 windowLog;          /* Window log for the LDM */
 } ldmParams_t;
```
Cyan4973: I'm slightly embarrassed here by the discrepancy between the name of the structure, and its actual function.

While `minMatchLength` and such do look like `params`,
`hashTable`, `bucketOffsets` and now `window` feel much more like a state context, that must be transported and evolves during compression.

It sounds like the 2 functions are mixed.

What I would expect in such a case is for such a structure to be actually named a `state` (or a `context`), _inside which_ one can find a nested `param` structure, which content is limited to just params (and might be used independently for other purpose, without all the `state` values attached).

What do you think ?

lib/compress/zstd_ldm.c
```diff
@@ -453,6 +453,56 @@ size_t ZSTD_ldm_generateSequences(
...
+
+size_t ZSTD_ldm_generateSequences(
+        ldmState_t* ldmState, rawSeq* sequences,
+        ldmParams_t const* params, void const* src, size_t srcSize,
+        int const extDict)
+{
+    U32 const maxDist = 1U << params->windowLog;
+    BYTE const* const istart = (BYTE const*)src;
+    size_t const kMaxChunkSize = 1 << 20;
+    size_t const nbChunks = (srcSize / kMaxChunkSize) + (srcSize != kMaxChunkSize);
```
Cyan4973: what's the intended value of `nbChunks` when `srcSize == (2 <<20)` ?

lib/compress/zstd_ldm.c
```diff
@@ -453,6 +453,56 @@ size_t ZSTD_ldm_generateSequences(
...
+        size_t const chunkSize = chunkEnd - chunkStart;
+
+        assert(chunkStart < srcSize);
+        if (ZSTD_window_needOverflowCorrection(ldmState->window)) {
+            U32 const ldmHSize = 1U << params->hashLog;
+            U32 const correction = ZSTD_window_correctOverflow(
+                &ldmState->window, /* cycleLog */ 0, maxDist, src);
+            ZSTD_ldm_reduceTable(ldmState->hashTable, ldmHSize, correction);
+        }
+        /* kMaxChunkSize should be small enough that we don't loose too much of
```
Cyan4973: 👍

lib/compress/zstd_ldm.c
```diff
@@ -453,6 +453,56 @@ size_t ZSTD_ldm_generateSequences(
...
+{
+    U32 const maxDist = 1U << params->windowLog;
+    BYTE const* const istart = (BYTE const*)src;
+    size_t const kMaxChunkSize = 1 << 20;
+    size_t const nbChunks = (srcSize / kMaxChunkSize) + (srcSize != kMaxChunkSize);
+    size_t nbSeq = 0;
+    size_t chunk;
+
+    assert(ZSTD_CHUNKSIZE_MAX >= kMaxChunkSize);
+    assert(ldmState->window.nextSrc >= (BYTE const*)src + srcSize);
```
Cyan4973: 1) In which case is it a valid situation to meet `nextSrc > (src+srcSize)` ?
2) Why is it asserted here ?

Cyan4973:
I erased a bogus comment related to `state` vs `param` structure,
as the `diff` view presented me both field lists as if they were combined into a single structure,
but after unfolding, I discovered they were actually already separated into 2 separated structures.

lib/compress/zstd_ldm.c
terrelln: Yeah, I'll test it when I have zstdmt exercising this path. I have another method to test as well. Only calling `ZSTD_window_enforceMaxDist()` after the generating the sequences, and checking offsets directly against `maxDist`. We still need to call `ZSTD_window_enforceMaxDist()` in this case, since the `lowLimit` must be updated to trigger overflow detection.

lib/compress/zstd_compress.c
terrelln: Thats a good point. I don't see a use for them to be different.

I was initially thinking each individual chunk in zstdmt would have a smaller window log, but it will be effectively limited using the chunk size and overlap log.

lib/compress/zstd_ldm.c
terrelln: The assert is checking that `ZSTD_window_update(&ldmState->window)` has been called for the input before calling the function.

PR #1029, opened 2018-02-28
FIO_addFInfo: Fully initialize output 'total' struct

cemeyer:
Silence a Coverity warning about 'windowSize' being uninitialized.
(Yes, nothing that calls this routine actually uses the windowSize
value.  Still, appeasing Coverity is pretty harmless in this case.)

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1963,6 +1963,7 @@ static void displayInfo(const char* inFileName, const fileInfo_t* info, int disp
 static fileInfo_t FIO_addFInfo(fileInfo_t fi1, fileInfo_t fi2)
 {
     fileInfo_t total;
+    memset(&total, 0, sizeof(total));
     total.numActualFrames = fi1.numActualFrames + fi2.numActualFrames;
     total.numSkippableFrames = fi1.numSkippableFrames + fi2.numSkippableFrames;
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks you for your patch @cemeyer .

Unfortunately, I believe this way of initializing a `struct` is not liked by many compilers.
So it will produce warnings on these compilers.

For a more "universal" patch, I would suggest : 
1) initialize the `struct` using `memset()`
2) add a line to initialize the missing member to `0`.

Considering there is only one missing member, I would recommend option 2.

cemeyer:
Hi Yann,

I believe most compilers prefer this style of initializing structs, actually.  Can you demonstrate the warning or error you have in mind?

The nice thing about this approach, or (1), is that it does not need to be changed if a new member is added.

Best,
Conrad

Cyan4973:
> Can you demonstrate the warning or error you have in mind?

Just have a look at the test list, as they all fail one after another.

I'm aware of the convenience of this notation.
By the way, it's accepted when initializing tables.
But not structures.
The C standard allows it, but compilers add a more restrictive set of rules in the form of warnings, because they believe it's a risky construction.

The closest equivalent and valid formulation is `memset()`.

cemeyer:
This compiler warning is totally spurious, unfortunately.

Switched patch to use memset().

Issue #1030, opened 2018-03-01
Does anynone know how to skip this compile error?


jackswan:
make
cc -O3   -I. -I./common -DXXH_NAMESPACE=ZSTD_ -I./legacy -DZSTD_LEGACY_SUPPORT=4  -c -o compress/huf_compress.o compress/huf_compress.c
In file included from compress/huf_compress.c:454:
compress/huf_compress_impl.h:35: error: attribute(target("bmi2")) is unknown
compress/huf_compress_impl.h:81: error: attribute(target("bmi2")) is unknown
make: *** [compress/huf_compress.o] Error 1

Cyan4973:
Which compiler are you using ?
(`cc -v`)

jackswan:
gcc version 4.4.7 20120313 (Red Hat 4.4.7-4) (GCC) has the problem.
It is ok with gcc version 4.8.3

PR #1031, opened 2018-03-01
force_inline HUF_decodeSymbol*()

Cyan4973:
`gcc v4.8` failed inlining these small functions,
resulting in major performance impact, especially for `bmi2`.

ex :
```
zstd -b1 silesia.tar
before : dec 680 MB/s
after  : dec 710 MB/s  (without bmi2)
after  : dec 770 MB/s  (with DYNAMIC_BMI2)
```

Also : 
limit `DYNAMIC_BMI2` to `gcc` >= 4.8

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -71,10 +71,11 @@
 #endif
 
 /* Enable runtime BMI2 dispatch based on the CPU.
- * Enabled for clang/gcc on x86 when BMI2 isn't enabled by default.
+ * Enabled for clang & gcc >=4.8 on x86 when BMI2 isn't enabled by default.
  */
 #ifndef DYNAMIC_BMI2
-  #if defined(__GNUC__) && (defined(__x86_64__) || defined(_M_X86)) && !defined(__BMI2__)
+  #if defined(__GNUC__) && (__GNUC__ >= 5 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8)) \
+      && (defined(__x86_64__) || defined(_M_X86)) && !defined(__BMI2__)
   #  define DYNAMIC_BMI2 1
   #else
   #  define DYNAMIC_BMI2 0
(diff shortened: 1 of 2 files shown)


terrelln:
Fixes #1030.

astratsx:
I recently encountered a compile time problem with DYNAMIC_BMI2 when compiling zstd on our RHEL 6.7 box using gcc 7.4 that I built for this box where it generated BMI2 instructions automatically which were not recognized by the older GNU assembler version 2.20.51.0.2-5.43.el6 20100205 that comes with RHEL 6.7. Yes, I could update the entire tools chain with a newer version of binutils to obtain a version of "as" that supports AVX2 an later, but I believe that it is an incorrect assumption to base DYNAMIC_BMI2 solely on the gcc version. Just because a particular version of gcc has the capability to generate BMI2 instructions doesn't mean that it should. The entire tool chain, including "as", must also have the ability to support BMI2. Also, I was intentionally building zstd targeting the SandyBridge architecture (yeah, we still have a lot of older HW)  which does NOT support BMI2. I was not expecting the compiler to generate BMI2 instructions since . The current DYNAMIC_BMI2 logic simply ignores the command line arguments setting for target machine architecture.  For example: -march=sandybridge
I was able to work around this by setting DYNAMIC_BMI2=0 when building, but this was obvious that this was the correct way to work around this issue.

terrelln:
I'm happy to take a PR that improves the detection of BMI2 support in the compilation toolchain.

We intentionally ignore the `-march` flag because we build in 2 versions of the function, one with BMI2 instructions, and one without. Then we detect BMI2 support at runtime (see [lib/common/cpu.h](https://github.com/facebook/zstd/blob/dev/lib/common/cpu.h)) and dispatch to the right function. Therefore, we can compile the binary for older hardware while getting the speed boost of BMI2 when we're running on newer hardware.

PR #1033, opened 2018-03-05
fix benchmark issue when measuring decoding speed only

Cyan4973:
zstd bench module can focus on decompression speed _only_.

This is useful when trying to measure decompression speed
on some large input data compressed using a high level,
as compression time becomes problematic (too long).

This mode is triggered by command : `zstd -b -d`

Problem was : in such a mode,
measured decoding speed was > 10% slower
than in nominal mode (compression + decompression).

The reason for such difference wasn't clear, and still isn't.
Nonetheless, this patch fixes the issue.
Moving the `memcpy()` operation sooner in the pipeline fixed it.

I can still measure small differences, but they are in the < 2% range,
so it's much more tolerable.

also : it doesn't matter anymore in which order are selected
commands `-b` and `-d`.
This combination always triggers bench_decodeOnly mode.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -260,7 +260,11 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
     }   }   }
 
     /* warmimg up memory */
-    RDG_genBuffer(compressedBuffer, maxCompressedSize, 0.10, 0.50, 1);
+    if (g_decodeOnly) {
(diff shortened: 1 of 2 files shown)


Cyan4973:
I hand-waived that maybe, for some reason, the `memcpy()` might be unfinished when starting the clock for decompression benchmark.
Though I didn't check assembly output for this hypothesis.

PR #1034, opened 2018-03-06
Dynamic selection of long offset mode

Cyan4973:
Long offset mode is a more complex decoding mode where sequences are decoded in advance in order to take advantage of `prefetch` capabilities. This is especially useful when some offsets are long, hence beyond memory cache capacity. Such memory access will stall the cpu, waiting idle for requested data to be loaded.

Long offset mode reduces the impact of stalling on long distance matches. But at the same time, it makes decoding of _every_ match more difficult, costing a baseline of 20% compared to regular decoder.
As a consequence, this mode is only enabled when the window size is large enough (> 16 MB).

So far, the triggering of this mode is solely based on window size, at the beginning of the frame.
This leads to an "all on (64-bit) / all off (32-bit)" design, which is not best in all circumstances.

For example, some files may have only a handful of long-distance matches, which is not enough to recoup the complexity cost on all other matches. This will notably happen when enabling "long range mode" associated with a fast or average compression level.
On the other hand, some files like `enwik` can feature _a lot_ of matches at long distance when compressed at `--ultra` level.

This patch improves the situation by making dynamic selection of `long offset mode`, per block.
To do so, it first decodes sequences tables, then analyze the offset distribution, and then select long-offset mode based on the proportion of long range matches.

This works reasonably well, improving performance in a number of scenario :
```
Mac OS-X, gcc-7, 64-bit :
Before :
3#silesia.tar (LRM) : 211984896 ->  65980439 (3.213), 101.7 MB/s , 534.6 MB/s
silesia-L20.tar.zst : 211984896 ->  53045812 (3.996),  0.00 MB/s , 570.1 MB/s
enwik8-L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 416.5 MB/s

After :
3#silesia.tar (LRM) : 211984896 ->  65980439 (3.213), 102.0 MB/s , 623.3 MB/s
silesia-L20.tar.zst : 211984896 ->  53045812 (3.996),  0.00 MB/s , 591.6 MB/s
enwik8-L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 420.2 MB/s
```

```
entropy, gcc-7, 64-bit:
Before :
3#silesia.tar (LRM) : 211948032 ->  65980564 (3.212), 170.4 MB/s ,1080.3 MB/s
silesia.tar.L20.zst : 211948032 ->  53003080 (3.999),  0.00 MB/s , 983.0 MB/s
enwik8.L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 726.5 MB/s

After:
3#silesia.tar (LRM) : 211948032 ->  65980564 (3.212), 176.8 MB/s ,1282.6 MB/s
silesia.tar.L20.zst : 211948032 ->  53003080 (3.999),  0.00 MB/s ,1094.8 MB/s
enwik8.L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 742.3 MB/s

```

```
Max OS-X, gcc, 32-bit
before:
3#silesia.tar (LRM) : 211984896 ->  65980439 (3.213),  59.6 MB/s , 419.6 MB/s
silesia-L20.tar.zst : 211984896 ->  53045812 (3.996),  0.00 MB/s , 361.0 MB/s
enwik8-L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 209.3 MB/s

after:
3#silesia.tar (LRM) : 211984896 ->  65980439 (3.213),  61.7 MB/s , 419.7 MB/s
silesia-L20.tar.zst : 211984896 ->  53045812 (3.996),  0.00 MB/s , 385.6 MB/s
enwik8-L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 303.4 MB/s
```

```
entropy, gcc, 32-bit:
before:
silesia.tar.L20.zst : 211948032 ->  53003080 (3.999),  0.00 MB/s , 635.8 MB/s
enwik8.L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 354.4 MB/s

after:
silesia.tar.L20.zst : 211948032 ->  53003080 (3.999),  0.00 MB/s , 626.1 MB/s
enwik8.L22.zst      : 100000000 ->  25380948 (3.940),  0.00 MB/s , 333.3 MB/s

```

One difficulty is that cpus react differently to these conditions. Cpus with larger caches suffer much less from long distance matches. 
This patch tries to settle for a "good" middle ground through testings. It's not necessarily the best for all platforms.
For example, `entropy` doesn't benefit from "long offset mode" within 32-bit x86 binary.
The current solution just makes it less probably to use "long offset mode" due to a high triggering threshold. It still happens sometimes, and on `entropy`, performance decreases a bit.
Same files, same parameters, but different cpu, on my laptop, gains are huge (up to 50% on `enwik8`).

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1248,47 +1248,49 @@ typedef enum { ZSTD_lo_isRegularOffset, ZSTD_lo_isLongOffset=1 } ZSTD_longOffset
 #endif
 
 typedef size_t (*ZSTD_decompressSequences_t)(
-    ZSTD_DCtx *dctx, void *dst, size_t maxDstSize, const void *seqStart,
-    size_t seqSize, const ZSTD_longOffset_e isLongOffset);
+    ZSTD_DCtx *dctx, void *dst, size_t maxDstSize,
+    const void *seqStart, size_t seqSize, int nbSeq,
+    const ZSTD_longOffset_e isLongOffset);
 
 static size_t ZSTD_decompressSequences(ZSTD_DCtx* dctx, void* dst, size_t maxDstSize,
-                                const void* seqStart, size_t seqSize,
+                                const void* seqStart, size_t seqSize, int nbSeq,
                                 const ZSTD_longOffset_e isLongOffset)
 {
 #if DYNAMIC_BMI2
     if (dctx->bmi2) {
-        return ZSTD_decompressSequences_bmi2(dctx, dst, maxDstSize, seqStart, seqSize, isLongOffset);
+        return ZSTD_decompressSequences_bmi2(dctx, dst, maxDstSize, seqStart, seqSize, nbSeq, isLongOffset);
     }
 #endif
-  return ZSTD_decompressSequences_default(dctx, dst, maxDstSize, seqStart, seqSize, isLongOffset);
+  return ZSTD_decompressSequences_default(dctx, dst, maxDstSize, seqStart, seqSize, nbSeq, isLongOffset);
 }
 
-static size_t ZSTD_decompressSequencesLong(ZSTD_DCtx* dctx, void* dst, size_t maxDstSize,
-                                    const void* seqStart, size_t seqSize,
-                                    const ZSTD_longOffset_e isLongOffset)
+static size_t ZSTD_decompressSequencesLong(ZSTD_DCtx* dctx,
+                                void* dst, size_t maxDstSize,
+                                const void* seqStart, size_t seqSize, int nbSeq,
+                                const ZSTD_longOffset_e isLongOffset)
 {
 #if DYNAMIC_BMI2
     if (dctx->bmi2) {
-        return ZSTD_decompressSequencesLong_bmi2(dctx, dst, maxDstSize, seqStart, seqSize, isLongOffset);
+        return ZSTD_decompressSequencesLong_bmi2(dctx, dst, maxDstSize, seqStart, seqSize, nbSeq, isLongOffset);
     }
 #endif
-  return ZSTD_decompressSequencesLong_default(dctx, dst, maxDstSize, seqStart, seqSize, isLongOffset);
+  return ZSTD_decompressSequencesLong_default(dctx, dst, maxDstSize, seqStart, seqSize, nbSeq, isLongOffset);
 }
 
 static unsigned
-ZSTD_getLongOffsetsShare(const FSE_DTable* offTable)
+ZSTD_getLongOffsetsShare(const ZSTD_seqSymbol* offTable)
 {
     const void* ptr = offTable;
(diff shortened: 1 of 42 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -1245,33 +1248,59 @@ typedef enum { ZSTD_lo_isRegularOffset, ZSTD_lo_isLongOffset=1 } ZSTD_longOffset
...
-        return ZSTD_decompressSequencesLong_bmi2(dctx, dst, maxDstSize, seqStart, seqSize, isLongOffset);
+        return ZSTD_decompressSequencesLong_bmi2(dctx, dst, maxDstSize, seqStart, seqSize, nbSeq, isLongOffset);
     }
 #endif
-  return ZSTD_decompressSequencesLong_default(dctx, dst, maxDstSize, seqStart, seqSize, isLongOffset);
+  return ZSTD_decompressSequencesLong_default(dctx, dst, maxDstSize, seqStart, seqSize, nbSeq, isLongOffset);
 }
 
+static unsigned
+ZSTD_getLongOffsetsShare(const ZSTD_seqSymbol* offTable)
```
terrelln: Can you comment that the result is the number of shares out of `1 << OffFSELog` that are long.

lib/decompress/zstd_decompress.c
```diff
@@ -1295,12 +1324,23 @@ static size_t ZSTD_decompressBlock_internal(ZSTD_DCtx* dctx,
...
-      && MEM_64bits() /* x86 benefits less from long mode than x64 */ )
-        return ZSTD_decompressSequencesLong(dctx, dst, dstCapacity, ip, srcSize, isLongOffset);
+    /* Build Decoding Tables */
+    {   int nbSeq;
+        size_t const seqHSize = ZSTD_decodeSeqHeaders(dctx, &nbSeq, ip, srcSize);
+        if (ZSTD_isError(seqHSize)) return seqHSize;
+        ip += seqHSize;
+        srcSize -= seqHSize;
+
+        if ( (dctx->fParams.windowSize > (1<<24) || !frame)
```
terrelln: We should check `frame` first, since otherwise the `fParams` may be uninitialized.

lib/decompress/zstd_decompress.c
```diff
@@ -1295,12 +1324,23 @@ static size_t ZSTD_decompressBlock_internal(ZSTD_DCtx* dctx,
...
+          && (nbSeq>0) ) {  /* could probably use a larger nbSeq limit */
+            U32 const shareLongOffsets = ZSTD_getLongOffsetsShare(dctx->OFTptr);
+            U32 const minShare = MEM_64bits() ? 7 : 20;
```
terrelln: Can you comment what fraction of long sequences this is?

PR #1035, opened 2018-03-07
Fix overflow protection with wlog=31

terrelln:
The overflow protection is broken when the window log is `> (3U << 29)`, so 31.
It doesn't work when `current` isn't around `1U << windowLog` ahead of `lowLimit`,
and the the assertion `current > newCurrent` fails. This happens when the same
context is used many times over, but with a large window log, like in zstdmt.

Fix it by triggering correction based on `nextSrc - base` instead of `lowLimit`.

The added test fails before the patch, and passes after.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1902,7 +1902,7 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
             return ERROR(dstSize_tooSmall);   /* not enough space to store compressed block */
         if (remaining < blockSize) blockSize = remaining;
 
-        if (ZSTD_window_needOverflowCorrection(ms->window)) {
+        if (ZSTD_window_needOverflowCorrection(ms->window, ip + blockSize)) {
             U32 const cycleLog = ZSTD_cycleLog(cctx->appliedParams.cParams.chainLog, cctx->appliedParams.cParams.strategy);
             U32 const correction = ZSTD_window_correctOverflow(&ms->window, cycleLog, maxDist, ip);
             ZSTD_STATIC_ASSERT(ZSTD_CHAINLOG_MAX <= 30);
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -453,13 +453,12 @@ MEM_STATIC size_t ZSTD_hashPtr(const void* p, U32 hBits, U32 mls)
 /*-*************************************
 *  Round buffer management
 ***************************************/
-
-#define ZSTD_LOWLIMIT_MAX (3U << 29) /* Max lowLimit allowed */
+/* Max current allowed */
+#define ZSTD_CURRENT_MAX ((3U << 29) + (1U << ZSTD_WINDOWLOG_MAX))
 /* Maximum chunk size before overflow correction needs to be called again */
-#define ZSTD_CHUNKSIZE_MAX                                                   \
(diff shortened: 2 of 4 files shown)


Cyan4973:
The solution _feels_ complex for the problem at hand.

The problem is triggered by the fact that `current` is too small.
But if it's small, no "overflow protection" should be triggered to begin with.

Overflow protection is triggered by `ZSTD_window_needOverflowCorrection()`.
Maybe that's the place to change ?

Currently, decision is triggered based on `window.lowLimit` value.
But `lowLimit` is just a proxy for `current` in this context.
And apparently, the relation between` lowLimit` and `current` is broken in the context of `zstdmt`.
So, maybe it's better to directly evaluate `current` instead, and test if next operation will make it overflow ?

terrelln:
Yeah, I'll look into that. I have a hard time understanding all the consequences of the overflow correction code.

terrelln:
The `aarch64` test is failing. AFAIK that test is flakey, but it is happening repeatedly, so I'm looking into it.

terrelln:
I wasn't able to reproduce the failure on my Macbook with sanitizers, or in a Ubuntu VM with `make aarch64fuzz`. I added logging for when the old overflow protection would apply, and when the new overflow protection does apply. Overflow protection is never triggered in the failing test https://travis-ci.org/facebook/zstd/jobs/353003570#L4999.

terrelln:
Lets see if rebasing on top of the latest `dev` branch helps.

Cyan4973:
The test failure is likely spurious.
We have an issue with Qemu ARM tests, especially when a lot of threads are created / freed, which is in essence what happens with `frametest --newapi`.
The fix will require an ability to resize `POOL` without the need to free and recreate a new `POOL`. But this is not a high priority item.

The patch looks good to me.
I was expecting you would merely check if `current` will overflow as a consequence of adding a new block, since [`ZSTD_window_needOverflowCorrection()` seems to be invoked before each new block](https://github.com/terrelln/zstd/blob/337f5e529081747d881d77a4c161cb0dd27b9b14/lib/compress/zstd_compress.c#L1881) .
But apparently you selected a more restrictive condition, for reasons I don't totally follow.
I'm convinced it will work fine though.

Note that github complains about some kind of "conflict" to be solved before merge.

terrelln:
Actually, using `nextSrc` isn't quite right, since the chunk size could be too large to work.

Cyan4973:
I'm a bit lost on this PR, as it changed significantly between comments (though there is no history to consult).

Is it considered ready at this stage ? or still in development ?

terrelln:
I don't have any changes planned.

I just changed to passing `srcEnd` into `needOverflowCorrection()` instead of using `nextSrc`.
The extra changes in `zstd_ldm.c` are not functional, just making things a bit nicer. I probably should have left them in another diff.

terrelln:
I think we should come up with a consistent policy for creating and merging PRs.

* I really don't like having one PR being many commits, since many of those commits are likely incomplete or won't compile, which hurts the ability to bisect for bugs in the future.
* Having only one single commit hurts the ability to see changes between versions of the pull request (although GitHub doesn't make this easy in any case).

Maybe we can develop PRs with multiple commits, but then squash them down to a single commit with the "Squash and merge pull request" option?

Cyan4973:
That's a good idea.
I never noticed that "squash and merge" option until now...

It would solve both issue : keep each patch consistent, and ability to notice changes during development.

PR #1037, opened 2018-03-07
Expose reference external sequence API

terrelln:
* Expose the reference external sequences API for zstdmt.
  Allows external sequences of any length, which get split when necessary.
* Reset the LDM window when the context is reset.
* Store the maximum number of LDM sequences.
* Sequence generation now returns the number of last literals.
* Fix sequence generation to not throw out the last literals when blocks of
  more than 1 MB are encountered.

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -213,12 +213,6 @@ MEM_STATIC void ZSTD_wildcopy_e(void* dst, const void* src, void* dstEnd)   /* s
 /*-*******************************************
 *  Private declarations
 *********************************************/
-typedef struct rawSeq_s {
-    U32 offset;
-    U32 litLength;
-    U32 matchLength;
-} rawSeq;
-
 typedef struct seqDef_s {
     U32 offset;
     U16 litLength;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -920,6 +920,8 @@ static size_t ZSTD_continueCCtx(ZSTD_CCtx* cctx, ZSTD_CCtx_params params, U64 pl
         (U32)pledgedSrcSize, cctx->appliedParams.fParams.contentSizeFlag);
     cctx->stage = ZSTDcs_init;
     cctx->dictID = 0;
+    if (params.ldmParams.enableLdm)
+        ZSTD_window_clear(&cctx->ldmState.window);
     ZSTD_invalidateMatchState(&cctx->blockState.matchState);
     ZSTD_reset_compressedBlockState(cctx->blockState.prevCBlock);
     XXH64_reset(&cctx->xxhState, 0);
@@ -1079,6 +1081,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
             ptr = zc->ldmState.hashTable + ldmHSize;
             zc->ldmSequences = (rawSeq*)ptr;
             ptr = zc->ldmSequences + maxNbLdmSeq;
+            zc->maxNbLdmSequences = maxNbLdmSeq;
 
             memset(&zc->ldmState.window, 0, sizeof(zc->ldmState.window));
         }
@@ -1103,6 +1106,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
             memset(ptr, 0, ldmBucketSize);
             zc->ldmState.bucketOffsets = (BYTE*)ptr;
             ptr = zc->ldmState.bucketOffsets + ldmBucketSize;
+            ZSTD_window_clear(&zc->ldmState.window);
         }
 
         /* buffers */
@@ -1834,17 +1838,33 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
             for (i = 0; i < ZSTD_REP_NUM; ++i)
                 zc->blockState.nextCBlock->rep[i] = zc->blockState.prevCBlock->rep[i];
         }
-        if (zc->appliedParams.ldmParams.enableLdm) {
-            size_t const nbSeq =
-                ZSTD_ldm_generateSequences(&zc->ldmState, zc->ldmSequences,
-                                           &zc->appliedParams.ldmParams,
-                                           src, srcSize, extDict);
+        if (zc->externSeqStore.pos < zc->externSeqStore.size) {
+            assert(!zc->appliedParams.ldmParams.enableLdm);
+            /* Updates ldmSeqStore.pos */
             lastLLSize =
-                ZSTD_ldm_blockCompress(zc->ldmSequences, nbSeq,
+                ZSTD_ldm_blockCompress(&zc->externSeqStore,
                                        ms, &zc->seqStore,
                                        zc->blockState.nextCBlock->rep,
                                        &zc->appliedParams.cParams,
                                        src, srcSize, extDict);
+            assert(zc->externSeqStore.pos <= zc->externSeqStore.size);
+        } else if (zc->appliedParams.ldmParams.enableLdm) {
+            rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0};
+
+            ldmSeqStore.seq = zc->ldmSequences;
+            ldmSeqStore.capacity = zc->maxNbLdmSequences;
+            /* Updates ldmSeqStore.size */
+            CHECK_F(ZSTD_ldm_generateSequences(&zc->ldmState, &ldmSeqStore,
+                                               &zc->appliedParams.ldmParams,
+                                               src, srcSize));
+            /* Updates ldmSeqStore.pos */
+            lastLLSize =
+                ZSTD_ldm_blockCompress(&ldmSeqStore,
+                                       ms, &zc->seqStore,
+                                       zc->blockState.nextCBlock->rep,
+                                       &zc->appliedParams.cParams,
+                                       src, srcSize, extDict);
+            assert(ldmSeqStore.pos == ldmSeqStore.size);
         } else {   /* not long range mode */
             ZSTD_blockCompressor const blockCompressor = ZSTD_selectBlockCompressor(zc->appliedParams.cParams.strategy, extDict);
             lastLLSize = blockCompressor(ms, &zc->seqStore, zc->blockState.nextCBlock->rep, &zc->appliedParams.cParams, src, srcSize);
@@ -2003,6 +2023,19 @@ size_t ZSTD_writeLastEmptyBlock(void* dst, size_t dstCapacity)
         MEM_writeLE24(dst, cBlockHeader24);
         return ZSTD_blockHeaderSize;
     }
+}
+
+size_t ZSTD_referenceExternalSequences(ZSTD_CCtx* cctx, rawSeq* seq, size_t nbSeq)
+{
+    if (cctx->stage != ZSTDcs_init)
+        return ERROR(stage_wrong);
+    if (cctx->appliedParams.ldmParams.enableLdm)
+        return ERROR(parameter_unsupported);
+    cctx->externSeqStore.seq = seq;
+    cctx->externSeqStore.size = nbSeq;
+    cctx->externSeqStore.capacity = nbSeq;
+    cctx->externSeqStore.pos = 0;
+    return 0;
 }
 
 
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -152,6 +152,19 @@ typedef struct {
     U32 windowLog;          /* Window log for the LDM */
 } ldmParams_t;
 
+typedef struct {
+    U32 offset;
+    U32 litLength;
+    U32 matchLength;
+} rawSeq;
+
+typedef struct {
+  rawSeq* seq;     /* The start of the sequences */
+  size_t pos;      /* The position where reading stopped. <= size. */
+  size_t size;     /* The number of sequences. <= capacity. */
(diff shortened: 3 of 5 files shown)


lib/compress/zstd_ldm.h
```diff
@@ -48,15 +47,17 @@ size_t ZSTD_ldm_generateSequences(
  * block compressor. The literals section of every sequence is passed to the
  * secondary block compressor, and those sequences are interspersed with the
  * predefined sequences. Returns the length of the last literals.
- * `nbSeq` is the number of sequences available in `sequences`.
```
Cyan4973: Just trying to understand the API at this stage.

Before calling `ZSTD_ldm_blockCompress()`: 
`*nbSeq` == nb of sequences in `sequences`.
So, it's basically the size of a table of `rawSeq`.

After calling `ZSTD_ldm_blockCompress()`: 
`*nbSeq` == nb of sequences consumed within `sequences`.
Hence, the index of the beginning of the remaining.

It works.
But I'm slightly annoyed by the change of nature of `*nbSeq`,
which starts by representing an "end of table", and then becomes a "beginning of table".
It's usable, it just pushes extra care into user territory.

The situation feels somewhat similar to streaming interface, which adopted `ZSTD_{in,out}Buffer*` format for its API. In this format, beginning and end of table are preserved after each function call, only `pos`, the starting position to read or write into the table, is updated.
This sounds like a little detail, but was pretty effective to make the API easier to read and use.

Does that sound applicable here ? Would it make the interface easier to use ?

lib/compress/zstd_ldm.h
```diff
@@ -28,18 +28,17 @@ extern "C" {
  * ZSTD_ldm_generateSequences():
  *
  * Generates the sequences using the long distance match finder.
- * The sequences completely parse a prefix of the source, but leave off the last
- * literals. Returns the number of sequences generated into `sequences`. The
- * user must have called ZSTD_window_update() for all of the input they have,
- * even if they pass it to ZSTD_ldm_generateSequences() in chunks.
+ * Generates at most `*nbSeqPtr` sequences into `sequences`, which parse a
+ * prefix of the source, and returns the number of unparsed literals at the end,
```
Cyan4973: small comment on return value : 
`returns the number of unparsed literals at the end`

This return value felt correct at a time when result of the "ldm" generator was passed to block compressor : it was more or less the amount of data the block compressor would have to work with.

Here I read it as : this is the place when "ldm" parsing will resume.

If that's correct, it feels more logical to return the amount of data processed from `src` instead.

What do you think ?

lib/compress/zstd_ldm.h
terrelln: Yeah thats make sense, I'll update it.

lib/compress/zstd_ldm.h
terrelln: I think I will just remove the return value for now, and fix it up later, in a patch where I only add the dynamic sequence resizing. The intention was to allow dynamically resizing the sequences storage, so it can resume parsing. However, this return value isn't enough to do that.

There are two "markers", the "nextToUpdate" position, where the LDM stopped inserting literals, and the number of trailing literals. We need to resume parsing at the nextToUpdate, even if there are more trailing literals, and also know how many extra literals to prepend to the first sequence.

Cyan4973: minor follow up : 
if return value becomes available, you might seize the opportunity to transform `size_t* nbSeqPtr ` from read/write argument,
into a more common `size_t nbSeq` read-only input argument, 
and provide resulting `nbSeq` as function return value.

sidenote : if this function cannot fail, it's probably good to mention it in documentation too.

terrelln:
I pushed the updates (also for the other PR), integrated it with my `mt-ldm` branch, and tested on some large files.

lib/common/zstd_internal.h
```diff
@@ -219,6 +219,12 @@ typedef struct rawSeq_s {
     U32 matchLength;
 } rawSeq;
 
+typedef struct {
+  rawSeq* seq;
+  size_t nbSeq;
+  size_t pos;
+} rawSeqStore_t;
```
Cyan4973: It seems `rawSeqStore_t` is only needed within `/lib/compress` (which feels logical).
In which case, it would be more accurate to define it within `/lib/compress/zstd_compress_internal.h`.

lib/compress/zstd_ldm.c
```diff
@@ -466,26 +468,33 @@ static void ZSTD_ldm_reduceTable(ldmEntry_t* const table, U32 const size,
 }
 
 size_t ZSTD_ldm_generateSequences(
-        ldmState_t* ldmState, rawSeq* sequences,
-        ldmParams_t const* params, void const* src, size_t srcSize,
-        int const extDict)
+        ldmState_t* ldmState, rawSeq* sequences, size_t nbSeq,
```
Cyan4973: `nbSeq` seems to be actually the size (or capacity) of `sequences`.
I believe the name should reflect that.

lib/compress/zstd_ldm.h
```diff
@@ -28,18 +28,18 @@ extern "C" {
  * ZSTD_ldm_generateSequences():
  *
  * Generates the sequences using the long distance match finder.
- * The sequences completely parse a prefix of the source, but leave off the last
- * literals. Returns the number of sequences generated into `sequences`. The
- * user must have called ZSTD_window_update() for all of the input they have,
- * even if they pass it to ZSTD_ldm_generateSequences() in chunks.
+ * Generates long range matching sequences in `sequences`, which parse a prefix
+ * of the source. `nbSeq` must be large enough to store every sequence.
```
Cyan4973: note 1 : `nbSeq` is actually the size / capacity of `sequences`.

note 2 : 
> `nbSeq` must be large enough to store every sequence.

is there a way to ensure this condition beforehand ?

terrelln: Yeah, there is `ZSTD_ldm_getMaxNbSeq()`.

lib/common/zstd_internal.h
Cyan4973: I regularly had hesitations about the meaning of `nbSeq`.
Sometimes, it's the capacity of `seq`, sometimes it's the nb of sequences actually present within `seq`, depending on read / write mode.

This might be because the internal API is regularly switching between `rawSeq*` and `rawSeqStore_t*` types, with `seqStore` becoming a relatively short-lived container.

I believe clarity would be helped by separating concepts `seqCapacity` and `nbSeq`.

terrelln: I'm changing `nbSeq` to `seqCapacity` where appropriate, and `rawSeqStore_t::nbSeq` to `rawSeqStore_t::end`.

The only function that doesn't take the `rawSeqStore_t` is `ZSTD_ldm_generateSequences()`, since it can use the return value, and doesn't support streaming. Do you think it would help clarity to take `rawSeqStore_t*` and store the number of generated sequences in `rawSeqStore_t::pos` instead?

Cyan4973: Yes, I believe so.

In isolation, the current design of `ZSTD_ldm_generateSequences()` would be fine,
but within a family of functions using `rawSeqStore_t*` as argument,
it seems better to keep the same type within interface.

terrelln:
The failed test looks like a transient infrastructure failure, but I can't rerun it on the Facebook account.

Cyan4973:
Indeed, test restarted

lib/compress/zstd_ldm.h
```diff
@@ -48,15 +49,17 @@ size_t ZSTD_ldm_generateSequences(
...
  * NOTE: The source must be at most the maximum block size, but the predefined
  * sequences can be any size, and may be longer than the block. In the case that
  * they are longer than the block, the last sequences may need to be split into
- * two. We handle that case correctly, and update `sequences` and `nbSeq`
- * appropriately.
+ * two. We handle that case correctly, and update `rawSeqStore` appropriately.
+ * NOTE: This function does not return any errors.
  */
-size_t ZSTD_ldm_blockCompress(rawSeq const* sequences, size_t nbSeq,
+size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
```
Cyan4973: Indicate what the return value is (`lastLLSize`?)

lib/compress/zstd_ldm.c
```diff
@@ -466,71 +468,92 @@ static void ZSTD_ldm_reduceTable(ldmEntry_t* const table, U32 const size,
...
+        if (prevPos < sequences->pos) {
+            sequences->seq[prevPos].litLength += (U32)leftoverSize;
+            leftoverSize = newLeftoverSize;
+        } else {
+            assert(newLeftoverSize == chunkSize);
+            leftoverSize += chunkSize;
+        }
     }
-    return nbSeq;
+    /* Return the number of generated sequences */
```
Cyan4973: The comment seems incorrect since function always returns `0` at this stage.

lib/compress/zstd_compress.c
```diff
@@ -1814,17 +1818,35 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
...
+
+            ldmSeqStore.seq = zc->ldmSequences;
+            ldmSeqStore.end = zc->maxNbLdmSequences;
+            ldmSeqStore.pos = 0;
+
+            CHECK_F(ZSTD_ldm_generateSequences(&zc->ldmState, &ldmSeqStore,
+                                               &zc->appliedParams.ldmParams,
+                                               src, srcSize));
+            /* We generated ldmSeqStore.pos sequences */
+            ldmSeqStore.end = ldmSeqStore.pos;
```
Cyan4973: The meaning of `end` and `pos` is changed
depending on being passed to `ZSTD_ldm_generateSequences()` or `ZSTD_ldm_blockCompress()`.

While I'm sure you properly control what's going on today,
I'm worried that some future contributor (which can be any one of us in the not-so-distance future), might be confused by such change of meaning half-way in the process.

I would really prefer `capacity`, `end`, and `pos` to be completely separated.
`ZSTD_ldm_generateSequences()` would update `end` by creating new sequences, and ensure it never goes beyond `capacity`,
`ZSTD_ldm_blockCompress()` would consume sequences by updating `pos`, ensuring it never goes beyond `end`.

lib/compress/zstd_ldm.c
```diff
@@ -466,71 +468,92 @@ static void ZSTD_ldm_reduceTable(ldmEntry_t* const table, U32 const size,
...
         /* Split the literals that we have out of the sequence.
          * They will become the last literals of this block.
          * The next block starts off with the remaining literals.
          */
-        sequences[seq].litLength -= remaining;
-        *nbSeq = seq;
+        rawSeqStore->seq[pos].litLength -= remaining;
         sequence.offset = 0;
     } else if (remaining < sequence.litLength + sequence.matchLength) {
         /* Split the match up into two sequences. One in this block, and one
```
Cyan4973: What about splitting at the boundary between literals and match ?

This would create variable size blocks. 
But I suspect the rest of the code is not yet ready for that ?
note : could be left as a refinement for later.

terrelln: Yeah, that would take some refactoring of the block compression code. `ZSTD_compressContinue_internal()`, `ZSTD_compress_frameChunk()`, and `ZSTD_compressBlock()` all expect `ZSTD_compressBlock_internal()` to compress all the input. It would also require a way to request not to do this, since it could be the last block in the frame, or a raw block, so all the input must be consumed.

lib/compress/zstd_compress_internal.h
```diff
@@ -152,6 +152,19 @@ typedef struct {
...
+    U32 litLength;
+    U32 matchLength;
+} rawSeq;
+
+typedef struct {
+  rawSeq* seq;     /* The start of the sequences */
+  size_t pos;      /* The position where reading stopped. <= size. */
+  size_t size;     /* The number of sequences. <= capacity. */
+  size_t capacity; /* The capacity of the `seq` pointer */
+} rawSeqStore_t;
```
Cyan4973: 👍

lib/compress/zstd_compress_internal.h
```diff
@@ -191,9 +204,11 @@ struct ZSTD_CCtx_s {
     ZSTD_customMem customMem;
     size_t staticSize;
 
-    seqStore_t seqStore;    /* sequences storage ptrs */
-    ldmState_t ldmState;    /* long distance matching state */
-    rawSeq* ldmSequences;   /* Storage for the ldm output sequences */
+    seqStore_t seqStore;      /* sequences storage ptrs */
+    ldmState_t ldmState;      /* long distance matching state */
+    rawSeq* ldmSequences;     /* Storage for the ldm output sequences */
```
Cyan4973: minor detail : why a `rawSeq*` here, instead of a `rawSeqStore_t` ?

Cyan4973:
The patch looks good to me.
I just have a feeling that the conversion from `rawSeq*` to `rawSeqStore_t` might be unnecessary (though maybe I'm missing something), in which case everything could be done by just passing along the same `rawSeqStore_t`. But that's just a minor detail.

lib/compress/zstd_compress_internal.h
terrelln: I wanted to make it clear that no sequences are maintained between calls to the block compressor. Using a `rawSeqStore_t` leaves the possibility that `pos != 0` or `size != 0` at the beginning of the call. Then it might look like we are throwing out sequences.

Cyan4973:
Looks good to me!
Just one last conflict to solve

PR #1041, opened 2018-03-12
Negative compression levels

Cyan4973:
This patch introduces faster compression levels, using negative values.
On the command line, they can be triggered using `--fast[=#]` command.

Negative compression levels use a new baseline level, stored at row `0`,
adding a `skip distance` between samples (simple rule : compression level `-3`, skip distance `3`), 
stored in field `ZSTD_p_targetLength`

Negative compression levels also disable Huffman compression of literals, translating into faster decoding speed.

It's possible to create one's own custom fast compression level
by using strategy `ZSTD_fast`, increasing `ZSTD_p_targetLength` to desired value,
and turning on or off literals compression, using `ZSTD_p_compressLiterals`.

Performance is generally on par or better than typical high speed algorithms such as `snappy` or `quicklz`. `lz4` still offers a better speed / compression combo.

name | ratio | compression | decompression
-- | -- | -- | --
zstd 1.3.4 --fast=5 | 1.996 | 770 MB/s | 2060 MB/s
lz4 1.8.1 | 2.101 | 750 MB/s | 3700 MB/s
zstd 1.3.4 --fast=4 | 2.068 | 720 MB/s | 2000 MB/s
zstd 1.3.4 --fast=3 | 2.153 | 675 MB/s | 1930 MB/s
lzo1x 2.09 -1 | 2.108 | 640 MB/s | 810 MB/s
zstd 1.3.4 --fast=2 | 2.265 | 610 MB/s | 1830 MB/s
quicklz 1.5.0 -1 | 2.238 | 540 MB/s | 720 MB/s
snappy 1.1.4 | 2.091 | 530 MB/s | 1820 MB/s
zstd 1.3.4 --fast=1 | 2.431 | 530 MB/s | 1770 MB/s
zstd 1.3.4 -1 | 2.877 | 470 MB/s | 1380 MB/s
brotli 1.0.2 -0 | 2.701 | 410 MB/s | 430 MB/s
lzf 3.6 -1 | 2.077 | 400 MB/s | 860 MB/s
zlib 1.2.11 -1 | 2.743 | 110 MB/s | 400 MB/s

It makes it possible for applications which consider Zstandard but are CPU-bounded to shift the load from CPU to bandwidth, even on a temporary basis.

Now for things which feel less neat : 

- Compression level `0` means `default`, which means `3`. In the continuum between `-1` and `1`, this makes it strange. To limit the impact, `bench` skip compression level `0` when requested to bench from a negative compression level to a positive one.
    One potential solution could be to make `1` the default level for API (as it used to be).
    In which case, compression level `0` would become the same as `1`, which would feel less strange than having a dip in compression speed at `0`. However, the "default" compression level of API would then differ from CLI side, which can be confusing too (actually, that's the reason why "default" was changed from `1` to `3` in the first place, to match CLI experience).

- The skip size is stored in `ZSTD_p_targetLength`. This is a misnomer : the intention is just to avoid creating a new field, re-using one which was so far idle with `ZSTD_fast` . However, the naming might now be confusing. Also, effect of this value is reversed compared to `ZSTD_opt` strategy : for `ZSTD_fast`, larger value increases speed and decreases ratio, the exact opposite of `ZSTD_opt`. This is all documented, but might nonetheless be confusing.

- Passing negative compression level through the advanced API `ZSTD_CCtx_setParameter()` : the argument is an `unsigned` type. The negative value is therefore casted to `unsigned` before being casted again to `int`. It works, but does not feel great.

- Huffman compression of literals is tracked through a new field `CCtxParams->disableLiteralsCompression` (named this way so that `0` is its default value). It works, but requires some extra care : now setting a negative compression level requires modifying this field too. This is done within `ZSTD_CCtxParams_setParameter()` (which means customizing `ZSTD_p_compressLiterals` must be done _after_ setting compression level). However, there are many other entry points in the API, so it's not yet totally guaranteed that all possible code paths handle this case properly.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -146,7 +146,7 @@ const seqStore_t* ZSTD_getSeqStore(const ZSTD_CCtx* ctx) { return &(ctx->seqStor
 static ZSTD_compressionParameters ZSTD_getCParamsFromCCtxParams(
         ZSTD_CCtx_params CCtxParams, U64 srcSizeHint, size_t dictSize)
 {
-    DEBUGLOG(4, "ZSTD_getCParamsFromCCtxParams: srcSize = %u, dictSize = %u",
+    DEBUGLOG(2, "ZSTD_getCParamsFromCCtxParams: srcSize = %u, dictSize = %u",
                 (U32)srcSizeHint, (U32)dictSize);
     return (CCtxParams.compressionLevel == ZSTD_CLEVEL_CUSTOM) ?
                 CCtxParams.cParams :
@@ -356,11 +356,17 @@ size_t ZSTD_CCtxParam_setParameter(
         CCtxParams->format = (ZSTD_format_e)value;
         return (size_t)CCtxParams->format;
 
-    case ZSTD_p_compressionLevel :
-        if ((int)value > ZSTD_maxCLevel()) value = ZSTD_maxCLevel();
-        if (value)  /* 0 : does not change current level */
-            CCtxParams->compressionLevel = value;
-        return CCtxParams->compressionLevel;
+    case ZSTD_p_compressionLevel : {
+        int cLevel = (int)value;  /* cast expected to restore negative sign */
+        if (cLevel > ZSTD_maxCLevel()) cLevel = ZSTD_maxCLevel();
+        DEBUGLOG(2, "ZSTD_CCtxParam_setParameter: set cLevel=%i", cLevel);
+        if (cLevel) {  /* 0 : does not change current level */
+            CCtxParams->disableLiteralCompression = (cLevel<0);  /* negative levels disable huffman */
+            CCtxParams->compressionLevel = cLevel;
+        }
+        if (CCtxParams->compressionLevel >= 0) return CCtxParams->compressionLevel;
+        return 0;  /* return type (size_t) cannot represent negative values */
+    }
 
     case ZSTD_p_windowLog :
         DEBUGLOG(4, "ZSTD_CCtxParam_setParameter: set windowLog=%u", value);
@@ -1393,6 +1399,9 @@ static size_t ZSTD_compressLiterals (ZSTD_entropyCTables_t const* prevEntropy,
     U32 singleStream = srcSize < 256;
     symbolEncodingType_e hType = set_compressed;
     size_t cLitSize;
+
+    DEBUGLOG(2,"ZSTD_compressLiterals (disableLiteralCompression=%i)",
+                disableLiteralCompression);
 
     /* Prepare nextEntropy assuming reusing the existing table */
     nextEntropy->hufCTable_repeatMode = prevEntropy->hufCTable_repeatMode;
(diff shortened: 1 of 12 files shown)


terrelln:
CC @felixhandte `-1` no longer means default.

lib/compress/zstdmt_compress.c
```diff
@@ -762,11 +763,12 @@ static size_t ZSTDMT_compress_advanced_internal(
     assert(jobParams.nbWorkers == 0);
     assert(mtctx->cctxPool->totalCCtx == params.nbWorkers);
 
-    DEBUGLOG(4, "ZSTDMT_compress_advanced_internal: nbJobs=%2u (rawSize=%u bytes; fixedSize=%u) ",
+    DEBUGLOG(2, "ZSTDMT_compress_advanced_internal: nbJobs=%2u (rawSize=%u bytes; fixedSize=%u) ",
```
terrelln: Change these debuglog's back from `2`.

lib/zstd.h
```diff
@@ -390,7 +390,7 @@ ZSTDLIB_API size_t ZSTD_DStreamOutSize(void);   /*!< recommended size for output
 #define ZSTD_SEARCHLOG_MIN       1
 #define ZSTD_SEARCHLENGTH_MAX    7   /* only for ZSTD_fast, other strategies are limited to 6 */
 #define ZSTD_SEARCHLENGTH_MIN    3   /* only for ZSTD_btopt, other strategies are limited to 4 */
-#define ZSTD_TARGETLENGTH_MIN    4   /* only useful for btopt */
+#define ZSTD_TARGETLENGTH_MIN    1   /* only useful for btopt */
```
terrelln: The comment is out of date.

lib/zstd.h
```diff
@@ -1001,17 +1007,25 @@ typedef enum {
...
     ZSTD_p_overlapSizeLog,   /* Size of previous input reloaded at the beginning of each job.
                               * 0 => no overlap, 6(default) => use 1/8th of windowSize, >=9 => use full windowSize */
 
     /* advanced parameters - may not remain available after API update */
+
+    ZSTD_p_compressLiterals=1000, /* control huffman compression of literals (enabled) by default.
+                              * disabling it improves speed and decreases compression ratio by a large amount.
+                              * note : this setting is updated when changing compression level.
+                              *        positive compression levels set literalCompression to 1.
+                              *        negative compression levels set literalCompression to 0. */
```
terrelln: Since this is in the unstable API we can always change it later. But if we do expose this somewhere more permanent, we might want to leave "space" for a middle ground, so we can say "only compress w/ huffman if there are large enough gains".

lib/compress/zstd_compress.c
```diff
@@ -146,7 +146,7 @@ const seqStore_t* ZSTD_getSeqStore(const ZSTD_CCtx* ctx) { return &(ctx->seqStor
 static ZSTD_compressionParameters ZSTD_getCParamsFromCCtxParams(
         ZSTD_CCtx_params CCtxParams, U64 srcSizeHint, size_t dictSize)
 {
-    DEBUGLOG(4, "ZSTD_getCParamsFromCCtxParams: srcSize = %u, dictSize = %u",
+    DEBUGLOG(2, "ZSTD_getCParamsFromCCtxParams: srcSize = %u, dictSize = %u",
```
terrelln: Change debuglog back to 4.

PR #1042, opened 2018-03-13
Made -H's printout specify the semantics of -T0

JesseTG:


Cyan4973:
Thanks for your patch @JesseTG !

Yes, it's a useful information to display.

PR #1044, opened 2018-03-13
Remove non-ASCII characters in header file comments

baldurk:
Currently there's a couple of UTF-8 characters in two header files, an en-dash & a non-breaking space. I guess they come out of copy-paste from elsewhere since I didn't see a reason for them otherwise. Without a UTF-8 BOM if the system codepage doesn't include the character this can lead to compile warnings in visual studio at least. I think this change is preferable to adding a BOM and potential headaches from that.

If this change isn't desired then feel free to reject the PR, though I'd like to fix the issue another way if possible.


Cyan4973:
Thanks for proposed patch @baldurk !
Yes, it's a no brainer fix.
It will be merged as soon as CLA get confirmed.

PR #1045, opened 2018-03-13
Use a single buffer in zstdmt

terrelln:
Allocate a single input buffer large enough to house each job, as well as
enough space for the IO thread to write 2 extra buffers. One goes in the
`POOL` queue, and one to fill, and then block on a full `POOL` queue.
Since we can't overlap with the prefix, we allocate space for 3 extra
input buffers.


terrelln:
Fix calling `XXH64_update()` with a `NULL` buffer on empty input, and specify some tests with `--long` as `--single-thread`.

lib/compress/zstdmt_compress.c
```diff
@@ -317,10 +324,8 @@ typedef struct {
     ZSTDMT_CCtxPool* cctxPool;           /* Thread-safe - used by mtctx and (all) workers */
     ZSTDMT_bufferPool* bufPool;          /* Thread-safe - used by mtctx and (all) workers */
     buffer_t dstBuff;                    /* set by worker (or mtctx), then read by worker & mtctx, then modified by mtctx => no barrier */
-    buffer_t srcBuff;                    /* set by mtctx, then released by worker => no barrier */
-    const void* prefixStart;             /* set by mtctx, then read and set0 by worker => no barrier */
-    size_t   prefixSize;                 /* set by mtctx, then read by worker => no barrier */
-    size_t   srcSize;                    /* set by mtctx, then read by worker & mtctx => no barrier */
+    range_t prefix;                      /* set by mtctx, then read by worker & mtctx => no barrier */
```
Cyan4973: 👍

lib/compress/zstdmt_compress.c
```diff
@@ -967,11 +971,30 @@ size_t ZSTDMT_initCStream_internal(
...
-    DEBUGLOG(4, "inBuff Size : %u KB", (U32)(mtctx->inBuff.targetCapacity>>10));
-    ZSTDMT_setBufferSize(mtctx->bufPool, MAX(mtctx->inBuff.targetCapacity, ZSTD_compressBound(mtctx->targetSectionSize)) );
+    DEBUGLOG(4, "inBuff Size : %u KB", (U32)(mtctx->targetSectionSize>>10));
+    ZSTDMT_setBufferSize(mtctx->bufPool, ZSTD_compressBound(mtctx->targetSectionSize));
+    {
+        /* Two buffers of slack, plus extra space for the overlap */
+        size_t const nbWorkers = MAX(mtctx->params.nbWorkers, 1);
+        size_t const nbSlackBuffers = MIN(nbWorkers, 2) + (mtctx->targetPrefixSize > 0);
+        size_t const nbSections = nbWorkers + nbSlackBuffers;
+        size_t const capacity = mtctx->targetSectionSize * nbSections;
```
Cyan4973: Should `capacity` be limited to `<= pledgedSrcSize` ?

lib/compress/zstdmt_compress.c
```diff
@@ -1066,14 +1088,13 @@ static size_t ZSTDMT_createCompressionJob(ZSTDMT_CCtx* mtctx, size_t srcSize, ZS
...
-        assert(mtctx->jobs[jobID].srcBuff.start == NULL);   /* no buffer left : supposed already released */
-        mtctx->jobs[jobID].srcBuff = mtctx->inBuff.buffer;
-        mtctx->jobs[jobID].prefixStart = mtctx->inBuff.buffer.start;
-        mtctx->jobs[jobID].prefixSize = mtctx->inBuff.prefixSize;
-        mtctx->jobs[jobID].srcSize = srcSize;
-        assert(mtctx->inBuff.filled >= srcSize + mtctx->inBuff.prefixSize);
+                    mtctx->nextJobID, (U32)srcSize, (U32)mtctx->inBuff.prefix.size);
+        mtctx->jobs[jobID].src.start = src;
+        mtctx->jobs[jobID].src.size = srcSize;
+        assert(mtctx->inBuff.filled >= srcSize);
```
Cyan4973: It seems this condition could be tighter ?
We want to ensure input buffer has been filled enough so that all data within `jobs[jobID].src` is actually initialized.

`mtctx->inBuff.filled >= srcSize` is fine when `src == inBuff.start`,
otherwise, it should be something like : 
`mtctx->inBuff.filled >= (src - inBuff.start) + srcSize` ?

lib/compress/zstdmt_compress.c
```diff
@@ -1089,29 +1110,20 @@ static size_t ZSTDMT_createCompressionJob(ZSTDMT_CCtx* mtctx, size_t srcSize, ZS
...
-                ZSTDMT_waitForAllJobsCompleted(mtctx);
-                ZSTDMT_releaseAllJobResources(mtctx);
-                return ERROR(memory_allocation);
-            }
-            mtctx->inBuff.filled -= (mtctx->inBuff.prefixSize + srcSize) - newPrefixSize;
-            memmove(mtctx->inBuff.buffer.start,   /* copy end of current job into next job, as "prefix" */
-                (const char*)mtctx->jobs[jobID].prefixStart + mtctx->inBuff.prefixSize + srcSize - newPrefixSize,
-                mtctx->inBuff.filled);
-            mtctx->inBuff.prefixSize = newPrefixSize;
+            size_t const newPrefixSize = MIN(srcSize, mtctx->targetPrefixSize);
```
Cyan4973: If an intermediate job is small (due to a `flush()` for example),
then the prefix of next job is going to be limited.

It seems there is no more any good reason for that limitation.
Since all input lies within a single big buffer, a prefix could encompass multiple previous jobs.
The comparison should rather be with the amount of data available within input buffer, hence : 

`size_t const newPrefixSize = MIN(inBuff.filled, mtctx->targetPrefixSize);`

note: I'm not sure at this stage if `inBuff` is really about the large single buffer.

lib/compress/zstdmt_compress.c
```diff
@@ -452,6 +456,8 @@ struct ZSTDMT_CCtx_s {
     ZSTD_CCtx_params params;
     size_t targetSectionSize;
     size_t targetPrefixSize;
+    buffer_t inBuffSpace;
+    size_t inBuffPos;
```
Cyan4973: `inBuffPos` seems to indicate how much data has been filled so far within `inBuffSpace`.
But since we also have some `inBuff` slices present the code, this feels confusing, as the naming implies the 2 are related.
Recommendation : change the name to `inBuffSpacePos`, or better make this value a member of `inBuffSpace` (which will have to change type).

Note : this is similar to what you did with `range_t`, and helps clarity.

lib/compress/zstdmt_compress.c
```diff
@@ -1232,6 +1243,112 @@ static size_t ZSTDMT_flushProduced(ZSTDMT_CCtx* mtctx, ZSTD_outBuffer* output, u
...
+        if (consumed < mtctx->jobs[wJobID].src.size) {
+            range_t range = mtctx->jobs[wJobID].prefix;
+            if (range.size == 0) {
+                /* Empty prefix */
+                range = mtctx->jobs[wJobID].src;
+            } else if (mtctx->jobs[wJobID].src.start < range.start) {
+                /* Two segments */
+                assert(mtctx->jobs[wJobID].src.start == mtctx->inBuffSpace.start);
+                range.start = mtctx->inBuffSpace.start;
+                range.size = mtctx->inBuffSpace.capacity;
```
Cyan4973: In this case, the "range of data used" covers the entire `inBuffSpace`, which means everything is in use, which means no space is available to load next job.

I'm wondering if that's a necessary restriction.

It seems enough to get the beginning of earliest job not yet completed,
which is basically `mtctx->jobs[wJobID].prefix.start`.

The assumption here is that this is a round buffer.
In which case `nextAvailableStart` immediately follows a busy segment.
What matters then is to check if there is enough "free" space in front of `nextAvailableStart`.

If `nextAvailableStart <= earliestJobStart`, the assumption is that the whole space between these 2 positions is available for loading new data, and we just have to check that's large enough to load a job.

If `nextAvailableStart > earliestJobStart`, the assumption is that everything from `nextAvailableStart` to the end of `inBuffSpace` is available to load new data, though we still have to check if that's large enough to load a job.
If not, then `nextAvailableStart = inBuffSpace.start`, and it becomes equivalent to previous case.

So, in case "earliest job not completed" spans the end and beginning of `inBuffSpace`, it's still possible to reserve some space to load some new job.

Feel free to comment.

lib/compress/zstdmt_compress.c
```diff
@@ -1303,10 +1419,11 @@ size_t ZSTDMT_compressStream_generic(ZSTDMT_CCtx* mtctx,
     }
 
     if ( (mtctx->jobReady)
-      || (mtctx->inBuff.filled >= newJobThreshold)  /* filled enough : let's compress */
+      || (mtctx->inBuff.filled >= mtctx->targetSectionSize)  /* filled enough : let's compress */
       || ((endOp != ZSTD_e_continue) && (mtctx->inBuff.filled > 0))  /* something to flush : let's go */
       || ((endOp == ZSTD_e_end) && (!mtctx->frameEnded)) ) {   /* must finish the frame with a zero-size block */
-        size_t const jobSize = MIN(mtctx->inBuff.filled - mtctx->inBuff.prefixSize, mtctx->targetSectionSize);
+        size_t const jobSize = MIN(mtctx->inBuff.filled, mtctx->targetSectionSize);
```
Cyan4973: maybe the `MIN()` is no longer necessary 
since the `assert()` implies that, by construction, 
we always guarantee that `mtctx->inBuff.filled` is the right data amount for next job.

lib/compress/zstdmt_compress.c
Cyan4973: Document what `inBuffPos` and `inBuff` are in this new construction.

Cyan4973:
There is a strange `memory allocation` error on Windows 32-bit test, as it happens during a relatively "small" job, with very reasonable input and window sizes : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/build/1.0.2412/job/5himp67l4mx9tg4c#L1721

Maybe some overflow ?

terrelln:
The problem is that it is asking for 1.5 GB. There is one worker, which has 2 extra slack buffers, each asking for 512 MB. This is fixed after the zstdmt ldm integration is done, so I'm just going to make the failing tests use `--single-thread` for now.

terrelln:
I just went ahead and made all the `--long` tests that weren't explicitly multithreaded use `--single-thread`.

lib/compress/zstdmt_compress.c
terrelln: Since we don't support 2-segment jobs yet, I'm just going to assert that it isn't the case.

lib/compress/zstdmt_compress.c
terrelln: I'm going to leave this until later, I created a task with the list of minor improvement follow-ups.

terrelln:
See D7284007 to look at the differences between updates.
* Create a separate roundBuffer_t to track the round buffer.
* Only allocate pledgedSrcSize if known.
* Remove MIN().

terrelln:
* Remove pledgedSrcSize adaption. 
  When we flush, we can run out of space. We need to work with
  ZSTD_tryGetInputRange() to only ask for however many bytes are remaining.
  I'd like to this in a separte diff, and not block the stack.
* Cast malloc result to BYTE*.

PR #1048, opened 2018-03-15
Master

HerveAncher:
Add support for VS2017 (MSbuild of Community edition)

[build.VS2017.zip](https://github.com/facebook/zstd/files/1815603/build.VS2017.zip)


Cyan4973:
Hi @HerveAncher 

Thanks for the proposed patch. Yes, this is a good idea.

It seems you have little experience of github. Please note :

- No PR is accepted into `master`. Any patch must stage at least through `dev` before reaching `master`. This is explained in the [contributing](https://github.com/facebook/zstd#contributing) section.
- Sending a `zip` file as link is not considered proper way to send a patch. You'll need to fork the project, create a new branch, modify the source code, push the change, and then issue a PR through github. This way, all changes can be properly tracked, attributed, commented, etc.
- zstd's `dev` branch already contains [3 scripts for VS2017](https://github.com/facebook/zstd/tree/dev/build/VS_scripts). They will become part of next release (and therefore be merged into master before that).

PR #1050, opened 2018-03-16
Restore setting loadedDictEnd

terrelln:
Setting `loadedDictEnd` was accidently removed from `ZSTD_loadDictionaryContent()`,
which means that dictionary compression will only be able to reference the parts of
the dictionary within the window. The spec allows us to reference the entire
dictionary so long as even one byte is in the window.

`ZSTD_enforceMaxDist()` incorrectly always allowed offsets up to `loadedDictEnd`
beyond the window, even once the dictionary was out of range.

When overflow protection kicked in, the check `current > loadedDictEnd + maxDist`
is incorrect if `loadedDictEnd` isn't reset back to zero. `current` could be reset
below the value, which would incorrectly allow references beyond the window. This
bug is present in `master`, but is very hard to trigger, since it requires both
dictionaries and data which triggers overflow correction.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1931,8 +1931,9 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
             ZSTD_reduceIndex(cctx, correction);
             if (ms->nextToUpdate < correction) ms->nextToUpdate = 0;
             else ms->nextToUpdate -= correction;
+            ms->loadedDictEnd = 0;
         }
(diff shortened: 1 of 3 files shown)


terrelln:
The `loadedDictEnd` code is slightly buggy. It is also buggy in `master`, just in a slightly different way, which is much harder to trigger.

terrelln:
The first bug is easy to see, the second is subtle. It only happens when overflow correction kicks in with a dictionary, and only if `current` is reset below `maxDist + loadedDictEnd`. This erroneously allows offsets up to `maxDist + loadedDictEnd` even when the dictionary is out of range. The solution is to reset `loadedDictEnd` when correcting the overflow. I also set it to zero once it is out of range, for extra safety.

Issue #880, opened 2017-10-03
Docker image for zstd


vaibhavsood:
Hi,

Want to check if there is a docker image/Dockerfile for zstd which is officially supported? I could not find one under dockerhub official images https://hub.docker.com/explore/

If not, want to check if there is any plan to add an official zstd image to dockerhub? https://docs.docker.com/docker-hub/official_repos/#how-do-i-create-a-new-official-repository

Cyan4973:
We haven't planned any docker image.
It seems like a reasonable idea, though the support of someone experienced in this process would be gladly welcomed.

gyscos:
To be clear, you're looking for a docker image that contains a pre-built zstd binary?
If so, here is a dockerfile that generates such an image:

```Dockerfile
# Dockerfile
FROM alpine

RUN apk --no-cache add make gcc libc-dev
COPY . /src
RUN cd /src && make && make install

FROM alpine

RUN mkdir -p /usr/local/share/licenses
COPY --from=0 /usr/local/bin/*zstd* /usr/local/bin/
COPY --from=0 /usr/local/share/man/man1/*zstd* /usr/local/share/man/man1/
COPY --from=0 /src/LICENSE /usr/local/share/licenses/zstd

CMD ["/usr/local/bin/zstd"]
```

Cyan4973:
Would it be useful to host such a script is `/contrib/docker` ?

gyscos:
I suppose. I think the main idea is to use this script to build an image, and then push this image to the official docker hub - ideally people would use this official image rather than create one themselves.

Cyan4973:
maybe calling it `zstd_image` or anything else
would avoid the confusion with the executable name ?

gyscos:
That's true; though the current standard is for images to have directly the name of the binary/service they provide: [mysql](https://hub.docker.com/_/mysql/), [openjdk](https://hub.docker.com/_/openjdk/), ...

As a note, there's an [unofficial lz4 image](https://hub.docker.com/r/sguthrie/lz4/). This one just sets a `CMD` to `bash`, so it needs to be called this way:
```
docker run sguthrie/lz4 lz4 -9
```

Cyan4973:
I've been trying to use the proposed `Dockerfile` by @gyscos to create a docker image.

As a warning, my docker skills are pretty limited, so I'm merely copy/pasting suggested script in a local file called `Dockerfile`. I'm using Ubuntu LTS 16.04 as baseline for this test.

[...]

PR #1052, opened 2018-03-18
Add Dockerfile

Varunram:
Dockerfile proposed by @gyscos at #880

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,0 +1,20 @@
+# Dockerfile
+# First image to build the binary
+FROM alpine as builder
+
+RUN apk --no-cache add make gcc libc-dev
+COPY . /src
+RUN mkdir /pkg && cd /src && make && make DESTDIR=/pkg install
+
+# Second minimal image to only keep the built binary
+FROM alpine
+
+# Copy the built files
+COPY --from=builder /pkg /
+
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @Varunram .

I've been considering doing this inclusion for some time now,
however having stopped on 2 points : 
- I could not reproduce the script result, due to a version requirement which was not fulfilled on my test machine
- [@gyscos' comment](https://github.com/facebook/zstd/issues/880#issuecomment-334294891) suggested users would be more interested in an image on official docker hub that in the script itself .

That being said, it is likely that having a script is nonetheless better than nothing.
Though I guess it'd better be documented (both how to run it, and the minimum version requirement).

Varunram:
@Cyan4973 Thanks for the review. I think the versioning problem should be solved by updating docker itself (I'm using docker-edge for reference). Regarding the comment, yes, you may use the script to generate an official docker build and maybe have the script as reference since most repos publish the Dockerfile and official build. This also ensure that if and when Docker decides to upgrade versions, we can keep track of the changes to the script pretty easily. However, just having the build is fine too, as long as it exists :)

gyscos:
Regarding the version, the officiel docker install docs use a ppa with a modern version available:
https://docs.docker.com/install/linux/docker-ce/ubuntu/

PR #1057, opened 2018-03-19
LRM parameters

Cyan4973:
This patch moves the LRM parameters from "experimental" section to the "supported" section of `ZSTD_cParameter`.

It also makes it possible to provide LRM parameters in any order. So `windowLog` can be specified before `enableLongDistanceMatching` for example, and still remain valid.

This capability is also extended to all compression parameters. It's now possible to specify `windowLog` or `hashLog` before `compressionLevel` (which used to reset all `cParams` members).

In order to introduce this feature, it was necessary to change the meaning of value `0` for most impacted parameters : previously, `0` meant "don't change the parameter". Now, it means "this parameter uses default settings". This is required because now parameters survive sessions and are not reset when setting compression level of LRM. So this is the only way to reset a parameter to default (well, technically, there is also `ZSTD_CCtx_reset()`, which resets _everything_).

The only parameter which still needs proper ordering is the experimental `ZSTD_p_compressLiterals`, as it's updated when setting `compressionLevel`. So for any user wanting a non-default setting, `ZSTD_p_compressLiterals` must be set _after_ `compressionLevel`.

It was relatively hard to assess that the patch does not introduce any "invisible" discrepancy, such as an ignored parameter (which would still produce a valid frame).
One key requirement is that `cParams` should always be extracted from `cctxParams` using `ZSTD_getCParamsFromCCtxParams()`, as this is where the proper construction rule is enforced. 
As far as I could review, it seems to be the case.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -359,43 +359,43 @@ size_t ZSTD_CCtxParam_setParameter(
     }
 
     case ZSTD_p_windowLog :
-        if (value)   /* 0 => use default */
+        if (value>0)   /* 0 => use default */
             CLAMPCHECK(value, ZSTD_WINDOWLOG_MIN, ZSTD_WINDOWLOG_MAX);
         CCtxParams->cParams.windowLog = value;
         return CCtxParams->cParams.windowLog;
 
     case ZSTD_p_hashLog :
-        if (value)   /* 0 => use default */
+        if (value>0)   /* 0 => use default */
             CLAMPCHECK(value, ZSTD_HASHLOG_MIN, ZSTD_HASHLOG_MAX);
         CCtxParams->cParams.hashLog = value;
         return CCtxParams->cParams.hashLog;
 
     case ZSTD_p_chainLog :
-        if (value)   /* 0 => use default */
+        if (value>0)   /* 0 => use default */
             CLAMPCHECK(value, ZSTD_CHAINLOG_MIN, ZSTD_CHAINLOG_MAX);
         CCtxParams->cParams.chainLog = value;
         return CCtxParams->cParams.chainLog;
 
     case ZSTD_p_searchLog :
-        if (value)   /* 0 => use default */
+        if (value>0)   /* 0 => use default */
             CLAMPCHECK(value, ZSTD_SEARCHLOG_MIN, ZSTD_SEARCHLOG_MAX);
         CCtxParams->cParams.searchLog = value;
         return value;
 
     case ZSTD_p_minMatch :
-        if (value)   /* 0 => use default */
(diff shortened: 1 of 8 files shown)


lib/compress/zstd_ldm.h
```diff
@@ -61,15 +60,12 @@ size_t ZSTD_ldm_generateSequences(
  * NOTE: This function does not return any errors.
  */
 size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
-    ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
-    ZSTD_compressionParameters const* cParams, void const* src, size_t srcSize,
-    int const extDict);
+        ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
```
terrelln: This is indented 4 less than `ZSTD_ldm_generateSequences()`.

lib/compress/zstd_compress.c
```diff
@@ -471,39 +447,32 @@ size_t ZSTD_CCtxParam_setParameter(
...
-            ZSTD_cLevelToCCtxParams(CCtxParams);
-            CCtxParams->cParams.windowLog = ZSTD_LDM_DEFAULT_WINDOW_LOG;
-        }
-        return ZSTD_ldm_initializeParameters(&CCtxParams->ldmParams, value);
+        CCtxParams->ldmParams.enableLdm = (value>0);
+        return CCtxParams->ldmParams.enableLdm;
 
     case ZSTD_p_ldmHashLog :
-        if (value) { /* 0 : does not change current ldmHashLog */
+        if (value>0)   /* 0 ==> auto */
```
terrelln: nit: pick one of `if (value)` (from above) and `if (value>0)`.

lib/compress/zstd_compress.c
```diff
@@ -2097,7 +2064,7 @@ size_t ZSTD_compressContinue (ZSTD_CCtx* cctx,
 size_t ZSTD_getBlockSize(const ZSTD_CCtx* cctx)
 {
     ZSTD_compressionParameters const cParams =
-            ZSTD_getCParamsFromCCtxParams(cctx->appliedParams, 0, 0);
+            ZSTD_getCParamsFromCCtxParams(&(cctx->appliedParams), 0, 0);
```
terrelln: Why can't we just use `cctx->appliedParams.cParams` here? Shouldn't `appliedParams` already be correctly adjusted?

Cyan4973: I wasn't completely sure, but indeed, `cParams` are supposed already built at this stage.
I'll replace this extraction by an `assert()`.

PR #1059, opened 2018-03-19
Integrate ldm with zstdmt

terrelln:
Integrate ldm into zstdmt by running it in serial and in order in the first
step of each job, in the same place as the hash gets updated. The input
buffer is sized to fit the whole LDM window and 2 full buffers of slack.
Input buffers cannot be reused until the LDM step is done with them.
After the LDM step is finished, the jobs don't actually have access to the
full window, only the overlap.

Tested on a few different multi-GB files with and without sanitizers,
and with different numbers of threads.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -922,6 +922,7 @@ static size_t ZSTD_continueCCtx(ZSTD_CCtx* cctx, ZSTD_CCtx_params params, U64 pl
     cctx->dictID = 0;
     if (params.ldmParams.enableLdm)
         ZSTD_window_clear(&cctx->ldmState.window);
+    ZSTD_referenceExternalSequences(cctx, NULL, 0);
     ZSTD_invalidateMatchState(&cctx->blockState.matchState);
     ZSTD_reset_compressedBlockState(cctx->blockState.prevCBlock);
     XXH64_reset(&cctx->xxhState, 0);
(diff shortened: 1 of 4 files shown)


Cyan4973:
Any potential benchmark, to compare how it works ?
Typically, `--single-thread` mode vs `1 worker` mode ?
Goal would be to evaluate speed, compression ratio and (if possible, as it is more difficult) memory usage.

terrelln:
I'll add benchmarks before it lands. I've looked at the speed and memory usage, and it looks the same, but I haven't written anything down yet.

terrelln:
To start the more in depth testing I've kicked off `zstreamtest` (`--mt`, `--opaqueapi`, `--newapi`) and `fuzzer`, each with no-sanitizers, UASAN, MSAN, and TSAN, to run for a day.

I'm also going to start the libFuzzer fuzzers, and make sure they get run with LDM and MT.

terrelln:
I benchmarked `--single-thread`, single worker, and multiple workers with and without long range mode. Without long range mode, the memory usage and speed looks the same before and after the patch. With long range mode and `--single-thread` the speed and memory usage look the same. With long range mode and 1+ worker, the memory usage is significantly lower, and the speed is faster for compression levels that aren't bottlenecked by the long range mode.

I'm preparing full benchmarks of the final result compared to `lrzip`.

PR #1061, opened 2018-03-20
Add reference to Ruby FFI bindings

msievers:
This PR adds a link to the [FFI based ruby bindings for zstd](https://github.com/msievers/zstandard-ruby). The differences to the already mentioned ruby bindings provided by SpringMT are

* the FFI bindings do not ship with a zstd implementation, but use the version which is installed on the system
* tested to work with zstd 0.6.0 - 1.3.3
* work with both major Ruby implementations MRI (CRuby) and JRuby (Java)


Cyan4973:
Great work @msievers !

PR #1062, opened 2018-03-21
Advanced decompression API

Cyan4973:
Implemented missing dictionary functions to be used with advanced API `ZSTD_decompress_generic()` : 

`ZSTD_DCtx_loadDictionary()`, `ZSTD_DCtx_loadDictionary_byReference()`, `ZSTD_DCtx_loadDictionary_advanced()`, `ZSTD_DCtx_refDDict()`, `ZSTD_DCtx_refPrefix()`, `ZSTD_DCtx_refPrefix_advanced()`


lib/zstd.h
```diff
@@ -433,10 +433,10 @@ typedef struct {
 typedef struct ZSTD_CCtx_params_s ZSTD_CCtx_params;
 
 typedef enum {
-    ZSTD_dm_auto=0,      /* dictionary is "full" if it starts with ZSTD_MAGIC_DICTIONARY, otherwise it is "rawContent" */
-    ZSTD_dm_rawContent,  /* ensures dictionary is always loaded as rawContent, even if it starts with ZSTD_MAGIC_DICTIONARY */
-    ZSTD_dm_fullDict     /* refuses to load a dictionary if it does not respect Zstandard's specification */
-} ZSTD_dictMode_e;
+    ZSTD_dct_auto=0,      /* dictionary is "full" when starting with ZSTD_MAGIC_DICTIONARY, otherwise it is "rawContent" */
```
terrelln: Why not `ZSTD_dict_auto` and such, it seems like adding one extra character is worthwhile.

lib/decompress/zstd_decompress.c
```diff
@@ -2720,14 +2775,17 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
...
                         /* legacy support is incompatible with static dctx */
                         if (zds->staticSize) return ERROR(memory_allocation);
                         CHECK_F(ZSTD_initLegacyStream(&zds->legacyContext,
                                     zds->previousLegacyVersion, legacyVersion,
                                     dict, dictSize));
                         zds->legacyVersion = zds->previousLegacyVersion = legacyVersion;
-                        return ZSTD_decompressLegacyStream(zds->legacyContext, legacyVersion, output, input);
-                    }
+                        {   size_t const hint = ZSTD_decompressLegacyStream(zds->legacyContext, legacyVersion, output, input);
+                            if (hint==0) zds->streamStage = zdss_init;   /* or stay in stage zdss_loadHeader */
```
terrelln: Were we missing the transparent reset in the legacy stream before?

Cyan4973: That was my concern when updating this part.
That being said, legacy stream was successfully passing tests so far, so either tests were incomplete, or there is a side effect that made it work nonetheless.

lib/zstd.h
Cyan4973: `dct` stands for Dictionary Content Type,
it's merely an accident if it ends up being close to `dict`.
That being said, why not if it helps clarity.

Issue #1043, opened 2018-03-13
[PATCH] CMakelist's fixes and addons.


Ne01eX:
This patch provides variables set for correct build zstd package on most unix platforms.

[zstd-1.3.3.cmake_fix-rtklinux-1.patch.gz](https://github.com/facebook/zstd/files/1808459/zstd-1.3.3.cmake_fix-rtklinux-1.patch.gz)

Best regards, Aleksandr "Ne01eX" Sayfulin

Cyan4973:
Thanks for this proposed patch @Ne01eX .

Any change this can become a PR ?

Ne01eX:
Yes, this may be used for PR ZSTD. :-)

Varunram:
@Ne01eX most of the changes were related to newline formatting, so I left them as is, I'll be putting in a PR shortly with the changes that aren't related to formatting and you can review it if desired. Thanks

PR #1063, opened 2018-03-21
Update man page variables in CMakeLists

Varunram:
Based on @Ne01eX's patch over at #1043. Omitted all formatting changes found over at the patch, so it condenses to a change in man page variables.


Cyan4973:
Looks good to me !
Thanks @Varunram !

PR #1065, opened 2018-03-21
Generic Visual Studio 2017 build script

HaydnTrigg:
This new build script automatically selects the installed Visual Studio 2017 implementation to use in order of Enterprise, Professional, and Community. This should be easier for people to reference without needing to specify which specific version of Visual Studio to use.

If a specific version is required, I have left the `build.VS2017Community` `build.VS2017Enterprise` and `build.VS2017Professional` build scripts so they can still be referenced.

There are now four `msbuild_version` options for 2017 `VS2017Community` `VS2017Professional` `VS2017Enterprise` and `VS2017` which automatically chooses the version to use.

### DIFF
--- a/build/VS_scripts/build.VS2017.cmd
+++ b/build/VS_scripts/build.VS2017.cmd
@@ -1,0 +1,7 @@
+@echo off
+
+rem build 32-bit
+call "%~p0%build.generic.cmd" VS2017 Win32 Release v141
+
+rem build 64-bit
+call "%~p0%build.generic.cmd" VS2017 x64 Release v141
--- a/build/VS_scripts/build.generic.cmd
+++ b/build/VS_scripts/build.generic.cmd
@@ -38,9 +38,9 @@ IF %msbuild_version% == VS2017Community SET msbuild="%msbuild_vs2017community%
 IF %msbuild_version% == VS2017Professional SET msbuild=%msbuild_vs2017professional%
 IF %msbuild_version% == VS2017Enterprise SET msbuild=%msbuild_vs2017enterprise%
 IF %msbuild_version% == VS2017 (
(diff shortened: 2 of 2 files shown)


Cyan4973:
Thanks for the patch @HaydnTrigg !

I fully agree with the objective.
Regarding the logic of the script, I have a few comments : 
- VS2017 `msbuild` path are written twice. It _seems_ better to re-use the same variable. For example : `IF %msbuild_version% == VS2017Community SET msbuild=%msbuild_vs2017community%`
- `msbuild` is going to be reset several times if `%msbuild_version% == VS2017`, ending with `VS2017Community` version if all 3 are installed. I would suggest : 
    + reverse the logic : `IF %msbuild_version% == VS2017`, then `if exist` nested
        * sidenote : some `IF` are uppercase, others `if` are lowercase, is there a reason ?
    + It's easier to follow the script logic when `msbuild` is set only once (rather than multiple times, with the last one wins). It probably needs using `else` or `exit` semantic (if it exists).
    + Document the priority order between VS2017 versions. This makes it easier to understand if the script behavior is correct.

Cyan4973:
Thanks for the refined script @HaydnTrigg !

Are you clear that, in the case of multiple VS2017 versions present on the same computer (if that's possible...), your script will select `VS2017Community` as higher priority ?

If yes, I believe the priority order deserves a mention somewhere,
typically within [local readme](https://github.com/facebook/zstd/blob/dev/build/VS_scripts/README.md).

HaydnTrigg:
@Cyan4973 You're totally right, I've swapped it around because I think the order makes more sense to go Enterprise > Professional > Community for the VS2017 option

Cyan4973:
I'll add a note about it in the readme then

PR #1068, opened 2018-03-21
Fix broken assertion

terrelln:
The `avgJobSize` must not be lower than 256 KB for single-pass mode.
In `zstd.h` we say the minimum value for `ZSTD_p_jobSize` is 1 MB,
so ensure that we always pick a size >= 1 MB.

Found by libFuzzer fuzzer tests with large input limits.

This bug has been around for a while, but only shows up when the
`windowLog` is very small, like in fuzzer tests, so I don't expect it to be
found in practice.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1029,7 +1029,7 @@ static size_t ZSTDMT_computeTargetJobLog(ZSTD_CCtx_params const params)
 {
     if (params.ldmParams.enableLdm)
         return MAX(21, params.cParams.chainLog + 4);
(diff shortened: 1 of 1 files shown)


Cyan4973:
Indeed.
For the general case, when `windowLog` is small (< 18), it happens because `srcSize` is small, and the whole thing is then deferred to the single-thread version.
However, it's possible to have unknown-but-large `srcSize` with a manually specified small `windowLog`, which I guess triggers the issue here.

terrelln:
Yeah, thats exactly the case, `windowLog=10`, `srcSize ~= 10 MB`.

PR #1072, opened 2018-03-24
Src package

Cyan4973:
This package automatically creates a `.tar.zst` source package of `zstd`/master.
It is only run when updating master, which should be done before any release.
It also calculates the sha256 of the archive.
Both files are then available on Appveyor Artifact menu,
on top of existing Windows pre-built binary archives.

minor issues : 
- that `.tar` archive comes from `git archive`. It contains files directly at root, in contrast with github's archive which contains a directory (like `zstd-1.3.3`) into which files are stored.
- the signature file contains whole output of Windows `CertUtil`, which includes a bit more verbosity than the standard `sha256sum` command on unix.


eli-schwartz:
> minor issues :
> 
> - that .tar archive comes from git archive. It contains files directly at root, in contrast with github's archive which contains a directory (like zstd-1.3.3) into which files are stored.

That's why the `git-archive(1)` command supports the `--prefix` option. In fact, GitHub's auto-generated source archives are *also* created using `git-archive(1)`. With the `--prefix` option.

>  - the signature file contains whole output of Windows CertUtil, which includes a bit more verbosity than the standard sha256sum command on unix.

It's also therefore useless for automated checking, since it cannot be used as file input for the `sha256sum --check` command and I bet certutil cannot read it either.

Wouldn't it make a lot more sense to implement this in travis or circleci instead of appveyor?

Added to the issues:
- Creating a .zst archive in order to distribute releases of the thing which lets you decompress .zst files is a pretty large bootstrap issue.
- The .sig file is erroneously named, since the standard naming for files containing sha256sum hashes is .sha256, whereas the .sig extension represents a claim that the file is a PGP signature, which it is not.

Cyan4973:
These are good points @eli-schwartz ,
I'll see what can be done to improve on these issues for next release.

Regarding the bootstrap issue :
note that there are many ways to get Zstandard source code, and the `.tar.zst` downloadable package is merely one of them. Alternative methods deliver different packaging/compression, so it seems reasonable to expect at least one of them to work. 
On the flip side, it is increasingly common for package managers to offer a compatible version of zstandard > 1.0 , so a `.tar.zst` package should already be readable on many systems, with more to follow in the future. I understand that there will always be outliers, and that's why alternative packaging methods are useful and will remain available.

eli-schwartz:
Yeah, the bootstrap issue is sort of iffy I agree, on the other hand I've recently been peripherally involved with people who had to bootstrap a distribution from ground zero.

For the average use case, now that libarchive finally supports zstd in a released version, we can make use of .zst archives in the general case, without writing special handling just for this.

PR #1074, opened 2018-03-25
fixed legacy/zstd_v* with older gcc version, by guarding builtin_*

rxrbln:
Fixed build error while compiling with slightly older GCCs, such as gcc-4.2.x on my am386dx40 by guarding the builtin_* as in other files.


Cyan4973:
Thanks for the patch @rxrbln !

Issue #1088, opened 2018-03-31
Failing regression test


bket:
OpenBSD fails on a specific test (v1.3.4, make check) with:

```
- Compress without dictID
tmp                  : 21.55%   ( 42430 =>   9142 bytes, tmp.zst)
tmp.zst : Decoding error (36) : Restored data doesn't match checksum
```

The test itself does something strange; it uses different dictionaries for compressing, and uncompressing.
 
https://github.com/facebook/zstd/blob/7188862d325ffb1c83b5495fa2da5ab1ad35b778/tests/playTests.sh#L373-L376

Concerning --no-dictID the manual states:

```
--nodictID
       do not store dictionary ID within frame header (dictionary
       compression). The decoder will have to rely on implicit
       knowledge about which dictionary to use, it won´t be able to
       check if it´s correct.
```

Taking the above into account the observed behaviour of failing this test seems to be expected. Btw, I think the manpage contains a typo: --nodictID should be --no-dictID.

If I change the test to use the same dictionary (like below) the test runs successfully.

```
$ZSTD -f tmp -D tmpDict1 --no-dictID 
$ZSTD -d tmp.zst -D tmpDict1 -fo result 
```

Now my question: Am I misinterpreting the test?

terrelln:
The two dictionaries should be the same, except for bytes [4, 8), which have different dictionary IDs. I suspect that they are different, can you verify with something like `sha1sum <(tail -c +9 tmpDict) <(tail -c +9 tmpDict1)`? I suspect that the order of the `*.c ../programs/*.c` isn't deterministic on OpenBSD. If that isn't the case, I suspect there is some non-determinism somewhere else.

bket:
> I suspect that they are different

Your suspicion seems correct, the checksums are different.

terrelln:
The fix should be as simple as (approximately)

```sh
dictFiles="*.c ../programs/*.c"
$ZSTD --train $dictFiles -o tmpDict
$ZSTD --train $dictFiles --dictID=1 -o tmpDict
```

terrelln:
@bket are you planning on submitting a PR to fix this? If not I will, but I can't test if it worked or not.

bket:
Running with `dictFiles=$(echo *.c ../programs/*.c) ` gives the same result: Failing 'Compress without dictID'. Success / failure rate is comparable to the previous series of runs.

terrelln:
[...]

PR #1091, opened 2018-03-31
Make sure that make check runs on OpenBSD.

bket:
OpenBSD uses md5 instead of md5sum, and has no device called full.

With this patch, make check runs until issue #1088. With the assumption made
in the issue, make check runs succesfully.

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -69,6 +69,7 @@ UNAME=$(uname)
 case "$UNAME" in
   Darwin) MD5SUM="md5 -r" ;;
   FreeBSD) MD5SUM="gmd5sum" ;;
+  OpenBSD) MD5SUM="md5" ;;
   *) MD5SUM="md5sum" ;;
 esac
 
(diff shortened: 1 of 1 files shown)


tests/playTests.sh
```diff
@@ -258,7 +259,7 @@ rm ./*.tmp ./*.zstd
 $ECHO "frame concatenation tests completed"
 
 
-if [ "$isWindows" = false ] && [ "$UNAME" != 'SunOS' ] ; then
+if [ "$isWindows" = false ] && [ "$UNAME" != 'SunOS' ] && [ "$(uname -s)" != "OpenBSD" ] ; then
```
terrelln: Does plain `$(uname)` not default to `-s` on OpenBSD? If it does, please use `$UNAME`.

bket:
> Does plain $(uname) not default to -s on OpenBSD?

Yes, it does. I changed the diff accordingly, and pushed it again.

terrelln:
Looks good! Once the tests finish we can merge it.

PR #1092, opened 2018-03-31
Fix #1083 and clean up support detection in compressFilename()

### LINKED ISSUE #1083: static suffix list displayed regardless of actual support
```Bash
$ unzstd FILE.bogus
zstd: FILE.bogus: unknown suffix (.gz/.xz/.zst/.lzma/.lz4 expected) -- ignored
$ unzstd FILE.gz 
zstd: FILE.gz: gzip file cannot be uncompressed (zstd compiled without HAVE_ZLIB) -- ignored
```

haneefmubarak:


haneefmubarak:
Welp. So due to `-Werror=switch-enum`, it doesn't like things being handled in `default`. I can redo that to be handled with more `#ifdef`s I suppose.

Meanwhile, `EXM_THROW` does not seem to like the part where the string sent in isn't a literal. Is there a way around that (enable typechecking some other way or override it for this particular case) and if not, what do you think would be a reasonable thing to do?

Cyan4973:
> due to `-Werror=switch-enum`, it doesn't like things being handled in default.

There are 2 ways to remain compliant with `-Werror=switch-enum` : 
- Do not use `switch() case:`, use a suite of `if () then else` instead
- Use `#define` _inside_ the `case`, so that the `case` is always present. You'll have to define a behavior should this `case` be hit while the corresponding `#define` is not enabled.

Cyan4973:
The second issue, impacting `EXM_THROW()` happens because we have additional rules on parsed strings, using [`-Wformat=2`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html).
This warning makes sure that the main string containing arguments can be parsed and controlled at compile time. This is to avoid "injection attacks", where an attacker could mess with the arguments of a string.

In your case, it happens because `error` pointer itself is not `const`, so it could be changed between the moment it is set, and the moment it is used.
To avoid this issue, just `const` the pointer, the compiler will be able to parse the string, removing the warning.

See this example : https://godbolt.org/g/L4jh2t

haneefmubarak:
So I `const`ed the inputs to `EXM_THROW()`: `error`, `suffix` :arrow_right: `csuffix`, and `macro` :arrow_right: `cmacro`; but it's still getting mad about `-Wformat=2`.

Does it require only compile-time literals for that or am I `const`ing wrong somewhere?

Cyan4973:
Only `error` needs to be `const`.
`suffix` and `macro` are parameters, they won't be parsed for additional fields, so they don't need to be `const` (plus they are obviously not determined at compile time, which is the whole point).

Well, I would have expected `const char* const error =` to work, as per the example.
But it appears this behavior depends in fact on `gcc` compiler version, and only works for gcc 5+.
For older versions, it doesn't.

Well, if a `const` pointer doesn't work, then a `#define` macro will.
See updated example : 
https://godbolt.org/g/fT3VaQ

programs/fileio.c
```diff
@@ -1754,8 +1822,26 @@ int FIO_decompressMultipleFilenames(const char** srcNamesTable, unsigned nbFiles
                     && strcmp(suffixPtr, ZSTD_EXTENSION)
                     && strcmp(suffixPtr, LZMA_EXTENSION)
                     && strcmp(suffixPtr, LZ4_EXTENSION)) ) {
-                DISPLAYLEVEL(1, "zstd: %s: unknown suffix (%s/%s/%s/%s/%s expected) -- ignored \n",
-                             srcFileName, GZ_EXTENSION, XZ_EXTENSION, ZSTD_EXTENSION, LZMA_EXTENSION, LZ4_EXTENSION);
+                DISPLAYLEVEL(1, "zstd: %s: unknown suffix ("
+                                ZSTD_EXTENSION
+#ifdef ZSTD_GZCOMPRESS
```
terrelln: Looks like Visual doesn't like macro definitions inside macro invocations.
We can refactor to something like:

```c
const char[] kSupportedExtensions = ZSTD_EXTENSION
#ifdef ZSTD_GZCOMPRESS
  /* ... */
#endif
;
DISPLAYLEVEL(1, "zstd: %s: unknown suffix (%s expected) -- ignored \n", srcFileName, kkSupportedExtensions);
```

haneefmubarak: So doing that (using a const built string) will make GCC unhappy (see 326224f)... I'm thinking that it might be time to break out some [string concatenation](https://gcc.gnu.org/onlinedocs/cpp/Concatenation.html) or to alternatively `ifdef` a special case for Visual.

Also, if compiler compatibility is so paramount, might it be worthwhile to consider adding jobs on Travis / Circle to check Clang too?

Cyan4973:
Seems this action has stalled, due to complexity of being compatible with many compilers.
It's a hard objective, agreed, and given `zstd` deployment ambitions, it's a top of requirement.

For information, a (similar) issue was solved some time ago in [`zstdcli.c`](https://github.com/facebook/zstd/blob/dev/programs/zstdcli.c#L333).

This function makes it possible to request `zstd -vV` and receive a list of supported codecs in return.
Given that this function passes all compilation tests, it might provide some hindsight on how to make this patch pass the same tests.

Cyan4973:
dropping, due to no feedback

PR #1110, opened 2018-04-21
Fix the DYNAMIC_BMI2 macro enables by mistake on ARM architecture with Clang

taigacon:
Just an operator precedence problem.


terrelln:
Thanks @taigacon!

PR #1114, opened 2018-04-30
Split library install target into pc, static, shared and include only target

pseiderer:
Enables fine grain compile/install (e.g. static/dyanamic library only, header-/pc-files only), used
for example by buildroot, see [1].

[1] https://git.buildroot.net/buildroot/tree/package/zstd/zstd.mk

Signed-off-by: Peter Seiderer <ps.report@gmx.net>


Cyan4973:
That's a nice suggestion, thanks @pseiderer

PR #1117, opened 2018-05-03
ZSTD_fast: Support Searching the Dictionary Context In-Place

felixhandte:
TBD.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -963,6 +963,7 @@ static void ZSTD_invalidateMatchState(ZSTD_matchState_t* ms)
     ms->nextToUpdate = ms->window.dictLimit + 1;
     ms->loadedDictEnd = 0;
     ms->opt.litLengthSum = 0;  /* force reset of btopt stats */
+    ms->dictMatchState = NULL;
 }
 
 /*! ZSTD_continueCCtx() :
@@ -1203,42 +1204,61 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
                             U64 pledgedSrcSize,
                             ZSTD_buffered_policy_e zbuff)
 {
+    /* We have a choice between copying the dictionary context into the working
+     * context, or referencing the dictionary context from the working context
+     * in-place. We decide here which strategy to use. */
+    /* TODO: pick reasonable cut-off size, handle ZSTD_CONTENTSIZE_UNKNOWN */
+    int attachDict = pledgedSrcSize < 64 KB
+                  && cdict->cParams.strategy == ZSTD_fast
+                  && ZSTD_equivalentCParams(cctx->appliedParams.cParams,
+                                            cdict->cParams);
+
     {   unsigned const windowLog = params.cParams.windowLog;
         assert(windowLog != 0);
         /* Copy only compression parameters related to tables. */
         params.cParams = cdict->cParams;
         params.cParams.windowLog = windowLog;
-        ZSTD_resetCCtx_internal(cctx, params, pledgedSrcSize, ZSTDcrp_noMemset, zbuff);
+        ZSTD_resetCCtx_internal(cctx, params, pledgedSrcSize,
+                                attachDict ? ZSTDcrp_continue : ZSTDcrp_noMemset,
+                                zbuff);
         assert(cctx->appliedParams.cParams.strategy == cdict->cParams.strategy);
         assert(cctx->appliedParams.cParams.hashLog == cdict->cParams.hashLog);
         assert(cctx->appliedParams.cParams.chainLog == cdict->cParams.chainLog);
     }
 
-    /* copy tables */
-    {   size_t const chainSize = (cdict->cParams.strategy == ZSTD_fast) ? 0 : ((size_t)1 << cdict->cParams.chainLog);
-        size_t const hSize =  (size_t)1 << cdict->cParams.hashLog;
-        size_t const tableSpace = (chainSize + hSize) * sizeof(U32);
-        assert((U32*)cctx->blockState.matchState.chainTable == (U32*)cctx->blockState.matchState.hashTable + hSize);  /* chainTable must follow hashTable */
-        assert((U32*)cctx->blockState.matchState.hashTable3 == (U32*)cctx->blockState.matchState.chainTable + chainSize);
-        assert((U32*)cdict->matchState.chainTable == (U32*)cdict->matchState.hashTable + hSize);  /* chainTable must follow hashTable */
-        assert((U32*)cdict->matchState.hashTable3 == (U32*)cdict->matchState.chainTable + chainSize);
-        memcpy(cctx->blockState.matchState.hashTable, cdict->matchState.hashTable, tableSpace);   /* presumes all tables follow each other */
-    }
-    /* Zero the hashTable3, since the cdict never fills it */
-    {   size_t const h3Size = (size_t)1 << cctx->blockState.matchState.hashLog3;
-        assert(cdict->matchState.hashLog3 == 0);
-        memset(cctx->blockState.matchState.hashTable3, 0, h3Size * sizeof(U32));
+    if (attachDict) {
+        DEBUGLOG(4, "attaching dictionary into context");
+        cctx->blockState.matchState.dictMatchState = &cdict->matchState;
+    } else {
+        DEBUGLOG(4, "copying dictionary into context");
+        /* copy tables */
+        {   size_t const chainSize = (cdict->cParams.strategy == ZSTD_fast) ? 0 : ((size_t)1 << cdict->cParams.chainLog);
+            size_t const hSize =  (size_t)1 << cdict->cParams.hashLog;
+            size_t const tableSpace = (chainSize + hSize) * sizeof(U32);
+            assert((U32*)cctx->blockState.matchState.chainTable == (U32*)cctx->blockState.matchState.hashTable + hSize);  /* chainTable must follow hashTable */
+            assert((U32*)cctx->blockState.matchState.hashTable3 == (U32*)cctx->blockState.matchState.chainTable + chainSize);
+            assert((U32*)cdict->matchState.chainTable == (U32*)cdict->matchState.hashTable + hSize);  /* chainTable must follow hashTable */
+            assert((U32*)cdict->matchState.hashTable3 == (U32*)cdict->matchState.chainTable + chainSize);
+            memcpy(cctx->blockState.matchState.hashTable, cdict->matchState.hashTable, tableSpace);   /* presumes all tables follow each other */
+        }
+
+        /* Zero the hashTable3, since the cdict never fills it */
+        {   size_t const h3Size = (size_t)1 << cctx->blockState.matchState.hashLog3;
+            assert(cdict->matchState.hashLog3 == 0);
+            memset(cctx->blockState.matchState.hashTable3, 0, h3Size * sizeof(U32));
+        }
+
+        /* copy dictionary offsets */
+        {
+            ZSTD_matchState_t const* srcMatchState = &cdict->matchState;
+            ZSTD_matchState_t* dstMatchState = &cctx->blockState.matchState;
+            dstMatchState->window       = srcMatchState->window;
+            dstMatchState->nextToUpdate = srcMatchState->nextToUpdate;
+            dstMatchState->nextToUpdate3= srcMatchState->nextToUpdate3;
+            dstMatchState->loadedDictEnd= srcMatchState->loadedDictEnd;
+        }
     }
 
-    /* copy dictionary offsets */
-    {
-        ZSTD_matchState_t const* srcMatchState = &cdict->matchState;
-        ZSTD_matchState_t* dstMatchState = &cctx->blockState.matchState;
-        dstMatchState->window       = srcMatchState->window;
-        dstMatchState->nextToUpdate = srcMatchState->nextToUpdate;
-        dstMatchState->nextToUpdate3= srcMatchState->nextToUpdate3;
-        dstMatchState->loadedDictEnd= srcMatchState->loadedDictEnd;
-    }
     cctx->dictID = cdict->dictID;
 
     /* copy block state */
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1207,7 +1207,6 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
     /* We have a choice between copying the dictionary context into the working
      * context, or referencing the dictionary context from the working context
      * in-place. We decide here which strategy to use. */
-    /* TODO: pick reasonable cut-off size, handle ZSTD_CONTENTSIZE_UNKNOWN */
     int attachDict = ( pledgedSrcSize <= 8 KB
                     || pledgedSrcSize == ZSTD_CONTENTSIZE_UNKNOWN )
                   && cdict->cParams.strategy == ZSTD_fast
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1207,11 +1207,12 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
     /* We have a choice between copying the dictionary context into the working
      * context, or referencing the dictionary context from the working context
      * in-place. We decide here which strategy to use. */
-    int attachDict = ( pledgedSrcSize <= 8 KB
-                    || pledgedSrcSize == ZSTD_CONTENTSIZE_UNKNOWN )
-                  && cdict->cParams.strategy == ZSTD_fast
-                  && ZSTD_equivalentCParams(cctx->appliedParams.cParams,
-                                            cdict->cParams);
+    const int attachDict = ( pledgedSrcSize <= 8 KB
+                          || pledgedSrcSize == ZSTD_CONTENTSIZE_UNKNOWN )
+                        && cdict->cParams.strategy == ZSTD_fast
+                        && ZSTD_equivalentCParams(cctx->appliedParams.cParams,
+                                                  cdict->cParams);
+
 
     {   unsigned const windowLog = params.cParams.windowLog;
         assert(windowLog != 0);
@@ -1227,7 +1228,9 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
     }
 
     if (attachDict) {
-        if (cdict->matchState.window.nextSrc - cdict->matchState.window.base == 0) {
+        const U32 cdictLen = (U32)( cdict->matchState.window.nextSrc
+                                  - cdict->matchState.window.base);
+        if (cdictLen == 0) {
             /* don't even attach dictionaries with no contents */
             DEBUGLOG(4, "skipping attaching empty dictionary");
         } else {
@@ -1236,15 +1239,12 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
 
             /* prep working match state so dict matches never have negative indices
              * when they are translated to the working context's index space. */
-            if (cctx->blockState.matchState.window.dictLimit <
-                (U32)(cdict->matchState.window.nextSrc - cdict->matchState.window.base)) {
+            if (cctx->blockState.matchState.window.dictLimit < cdictLen) {
                 cctx->blockState.matchState.window.nextSrc =
-                    cctx->blockState.matchState.window.base +
-                    ( cdict->matchState.window.nextSrc
-                    - cdict->matchState.window.base);
+                    cctx->blockState.matchState.window.base + cdictLen;
                 ZSTD_window_clear(&cctx->blockState.matchState.window);
             }
-            cctx->blockState.matchState.loadedDictEnd = cctx->blockState.matchState.window.dictLimit;
+            cctx->blockState.matchState.loadedDictEnd = params.forceWindow ? 0 : cdictLen;
         }
     } else {
         DEBUGLOG(4, "copying dictionary into context");
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1229,6 +1229,17 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
     if (attachDict) {
         DEBUGLOG(4, "attaching dictionary into context");
         cctx->blockState.matchState.dictMatchState = &cdict->matchState;
+
+        /* prep working match state so dict matches never have negative indices
+         * when they are translated to the working context's index space. */
+        if (cctx->blockState.matchState.window.dictLimit <
+            (U32)(cdict->matchState.window.nextSrc - cdict->matchState.window.base)) {
+            cctx->blockState.matchState.window.nextSrc =
+                cctx->blockState.matchState.window.base +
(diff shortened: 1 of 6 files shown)


Cyan4973:
Nice start Felix !

2 things : 
- as usual for such evolution, we'll need some comparison metrics before merge
- keep the `extDict` integration for another patch.
  It's a good goal, but execution is more complex than it seems, and you don't want to stack up too many difficulties in a single PR.

felixhandte:
Absolutely. Here's a representative benchmark result (showing levels -5, -1, 1, and 2):

montage-nci-gcc-00-b095bff-dbf373a-all-cls

`clang-4.0`, which is what I have on my desktop, seems to perform a little worse with some inputs. I'm going to kick off some benchmarking on my dev server where I have more up-to-date compilers.

lib/compress/zstd_compress.c
```diff
@@ -1199,42 +1200,61 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
                             U64 pledgedSrcSize,
                             ZSTD_buffered_policy_e zbuff)
 {
+    /* We have a choice between copying the dictionary context into the working
+     * context, or referencing the dictionary context from the working context
+     * in-place. We decide here which strategy to use. */
+    /* TODO: pick reasonable cut-off size, handle ZSTD_CONTENTSIZE_UNKNOWN */
+    int attachDict = pledgedSrcSize <= 8 KB
+                  && cdict->cParams.strategy == ZSTD_fast
+                  && ZSTD_equivalentCParams(cctx->appliedParams.cParams,
```
Cyan4973: What if `ZSTD_equivalentCParams()==0` ?
Does that mean the first time the dictionary is used, the strategy is to copy cdict tables ?

lib/compress/zstd_compress.c
```diff
@@ -1947,23 +1967,26 @@ MEM_STATIC size_t ZSTD_compressSequences(seqStore_t* seqStorePtr,
...
           ZSTD_compressBlock_lazy, ZSTD_compressBlock_lazy2, ZSTD_compressBlock_btlazy2,
           ZSTD_compressBlock_btopt, ZSTD_compressBlock_btultra },
         { ZSTD_compressBlock_fast_extDict  /* default for 0 */,
           ZSTD_compressBlock_fast_extDict, ZSTD_compressBlock_doubleFast_extDict, ZSTD_compressBlock_greedy_extDict,
           ZSTD_compressBlock_lazy_extDict,ZSTD_compressBlock_lazy2_extDict, ZSTD_compressBlock_btlazy2_extDict,
-          ZSTD_compressBlock_btopt_extDict, ZSTD_compressBlock_btultra_extDict }
+          ZSTD_compressBlock_btopt_extDict, ZSTD_compressBlock_btultra_extDict },
+        { ZSTD_compressBlock_fast_dictMatchState  /* default for 0 */,
+          ZSTD_compressBlock_fast_dictMatchState,
+          NULL, NULL, NULL, NULL, NULL, NULL, NULL /* unimplemented as of yet */ }
```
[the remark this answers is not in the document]
felixhandte: Yeah, shouldn't be possible, since `attachDict` is only selected for `ZSTD_fast` in this PR. But I can add an assert, sure.

lib/compress/zstd_compress.c
felixhandte: Correct. If we didn't, we would have to do a full reset of the working context anyways, which would be just as expensive. And accessing the dict from the working context is faster.

lib/compress/zstd_compress_internal.h
```diff
@@ -497,6 +502,16 @@ MEM_STATIC U32 ZSTD_window_hasExtDict(ZSTD_window_t const window)
     return window.lowLimit < window.dictLimit;
 }
 
+/**
+ * ZSTD_matchState_dictMode():
+ * Does what the label says.
```
Cyan4973: can this be clearer ?

lib/compress/zstd_compress_internal.h
```diff
@@ -497,6 +502,16 @@ MEM_STATIC U32 ZSTD_window_hasExtDict(ZSTD_window_t const window)
...
+ */
+MEM_STATIC ZSTD_dictMode_e ZSTD_matchState_dictMode(const ZSTD_matchState_t *ms)
+{
+    return ms->dictMatchState != NULL ? ZSTD_dictMatchState :
```
Cyan4973: use parenthesis and newline + indentation to help reader understand this line.

lib/compress/zstd_fast.c
```diff
@@ -59,9 +60,41 @@ size_t ZSTD_compressBlock_fast_generic(
     U32 offset_1=rep[0], offset_2=rep[1];
     U32 offsetSaved = 0;
 
+    /* This is all complicated by the fact that we need to handle positions
+     * specified in 3 different ways: by direct pointers, by indices relative
```
Cyan4973: just for my memory : in which case are positions tracked by direct pointers only ?
I thought everything was using indices, although relative to up to 2 different bases.

felixhandte: Right, no data structure tracks things by pointers. But pointers are used in this function for more than just accesses, they're compared, etc. That's all I'm saying. I guess this is really a to-do directed at me to see if I can clean all this up some.

lib/compress/zstd_compress.c
```diff
@@ -1199,42 +1200,61 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
                             U64 pledgedSrcSize,
                             ZSTD_buffered_policy_e zbuff)
 {
+    /* We have a choice between copying the dictionary context into the working
+     * context, or referencing the dictionary context from the working context
+     * in-place. We decide here which strategy to use. */
+    /* TODO: pick reasonable cut-off size, handle ZSTD_CONTENTSIZE_UNKNOWN */
```
terrelln: When we don't know the content size I think we should assume it is small since it is dictionary compression.

lib/compress/zstd_fast.c
```diff
@@ -73,26 +96,63 @@ size_t ZSTD_compressBlock_fast_generic(
         U32 const current = (U32)(ip-base);
         U32 const matchIndex = hashTable[h];
         const BYTE* match = base + matchIndex;
+        const ptrdiff_t repIndex = (ptrdiff_t)current + 1 - offset_1;
+        const BYTE* repBase = (hasDict == ZSTD_dictMatchState
```
terrelln: Heads up, I tried a similar strategy for merging single-segment and ext-dict compressors, and it worked for `ZSTD_fast`, but I couldn't get it to work on `ZSTD_doubleFast`. You might have better luck, but it is something to be aware of.

lib/compress/zstd_fast.c
```diff
@@ -45,23 +45,46 @@ FORCE_INLINE_TEMPLATE
...
+        ZSTD_dictMode_e const dictMode)
 {
     U32* const hashTable = ms->hashTable;
     const BYTE* const base = ms->window.base;
     const BYTE* const istart = (const BYTE*)src;
     const BYTE* ip = istart;
     const BYTE* anchor = istart;
-    const U32   lowestIndex = ms->window.dictLimit;
-    const BYTE* const lowest = base + lowestIndex;
+    const U32   localLowestIndex = ms->window.dictLimit;
```
Cyan4973: I believe the following naming convention is a better fit : `lowPrefixIndex`, followed by `lowPrefixPtr`.
Later in the code, it gives a clearer attribution of what this limit is referring to.

lib/compress/zstd_fast.c
```diff
@@ -45,23 +45,46 @@ FORCE_INLINE_TEMPLATE
...
+    const BYTE* const localLowest = base + localLowestIndex;
     const BYTE* const iend = istart + srcSize;
     const BYTE* const ilimit = iend - HASH_READ_SIZE;
     U32 offset_1=rep[0], offset_2=rep[1];
     U32 offsetSaved = 0;
 
+    const ZSTD_matchState_t* const dms = ms->dictMatchState;
+    const U32* const dictHashTable = dictMode == ZSTD_dictMatchState ?
+                                     dms->hashTable : NULL;
+    const U32 lowestDictIndex      = dictMode == ZSTD_dictMatchState ?
```
Cyan4973: naming consistency : 
all dictionary related variables seem to start with prefix `dict`,
except this one.
`dictLowestIndex` seems to fit better.

lib/compress/zstd_fast.c
```diff
@@ -45,23 +45,46 @@ FORCE_INLINE_TEMPLATE
...
+                                     dms->window.dictLimit : 0;
+    const BYTE* const dictBase     = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.base : NULL;
+    const BYTE* const dictLowest   = dictMode == ZSTD_dictMatchState ?
```
Cyan4973: Continuing on naming consistency : 
if you use `dictLowestIndex` above, `dictLowestPtr` here would help distinguishing the two.

lib/compress/zstd_fast.c
```diff
@@ -45,23 +45,46 @@ FORCE_INLINE_TEMPLATE
...
+                                     localLowestIndex;
+
+    assert(dictMode == ZSTD_noDict || dictMode == ZSTD_dictMatchState);
+
     /* init */
-    ip += (ip==lowest);
-    {   U32 const maxRep = (U32)(ip-lowest);
+    ip += (dictMode == ZSTD_noDict && ip == localLowest);
+    {   U32 const maxRep = dictMode == ZSTD_dictMatchState ?
+                           (U32)(ip - dictLowest) :
```
Cyan4973: `ip` is relative to `base`,
while `dictLowest` is relative to `dictBase`.
They are not comparable.

Prefer working directly within index referential to avoid such issues.

lib/compress/zstd_fast.c
```diff
@@ -45,23 +45,46 @@ FORCE_INLINE_TEMPLATE
...
+    const U32 lowestDictIndex      = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.dictLimit : 0;
+    const BYTE* const dictBase     = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.base : NULL;
+    const BYTE* const dictLowest   = dictMode == ZSTD_dictMatchState ?
+                                     dictBase + lowestDictIndex : NULL;
+    const BYTE* const dictEnd      = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.nextSrc : NULL;
+    const U32 dictIndexDelta       = dictMode == ZSTD_dictMatchState ?
+                                     localLowestIndex - (U32)(dictEnd - dictBase) :
```
Cyan4973: Is there a guarantee that `dictEnd - dictBase` (effectively `dictSize`) is necessarily < `localLowestIndex` ?
If not, does this still work ?

lib/compress/zstd_fast.c
felixhandte: Good catch!

lib/compress/zstd_fast.c
```diff
@@ -73,26 +96,63 @@ size_t ZSTD_compressBlock_fast_generic(
...
+        const BYTE* repBase = (dictMode == ZSTD_dictMatchState
+                            && repIndex < (ptrdiff_t)localLowestIndex) ?
+                               dictBase - dictIndexDelta : base;
+        const BYTE* repMatch = repBase + repIndex;
         hashTable[h] = current;   /* update hash table */
 
-        if ((offset_1 > 0) & (MEM_read32(ip+1-offset_1) == MEM_read32(ip+1))) {
+        if (dictMode == ZSTD_dictMatchState
+            && (((U32)((localLowestIndex-1) - repIndex) >= 3 /* intentional underflow */)
+                & (repIndex > dictLowestLocalIndex))
```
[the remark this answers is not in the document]
felixhandte: Aha, very good. Yes, you're absolutely right. Which means we can get rid of dictLowestLocalIndex!

lib/compress/zstd_fast.c
```diff
@@ -73,26 +96,63 @@ size_t ZSTD_compressBlock_fast_generic(
         U32 const current = (U32)(ip-base);
         U32 const matchIndex = hashTable[h];
         const BYTE* match = base + matchIndex;
+        const ptrdiff_t repIndex = (ptrdiff_t)current + 1 - offset_1;
+        const BYTE* repBase = (dictMode == ZSTD_dictMatchState
+                            && repIndex < (ptrdiff_t)localLowestIndex) ?
+                               dictBase - dictIndexDelta : base;
+        const BYTE* repMatch = repBase + repIndex;
```
Cyan4973: better define `repMatch` directly.
This will avoid any risk of potential undefined behavior related to pointer arithmetic (not sure what `dictBase - dictIndexDelta` can end up being).

lib/compress/zstd_fast.c
```diff
@@ -73,26 +96,63 @@ size_t ZSTD_compressBlock_fast_generic(
...
+        if (dictMode == ZSTD_dictMatchState
+            && (((U32)((localLowestIndex-1) - repIndex) >= 3 /* intentional underflow */)
+                & (repIndex > dictLowestLocalIndex))
+            && (MEM_read32(repMatch) == MEM_read32(ip+1)) ) {
+            const BYTE* repMatchEnd = repIndex < (ptrdiff_t)localLowestIndex ? dictEnd : iend;
+            mLength = ZSTD_count_2segments(ip+1+4, repMatch+4, iend, repMatchEnd, istart) + 4;
+            ip++;
+            ZSTD_storeSeq(seqStore, ip-anchor, anchor, 0, mLength-MINMATCH);
+        } else if ( dictMode == ZSTD_noDict
+                 && (offset_1 > 0) & (MEM_read32(ip+1-offset_1) == MEM_read32(ip+1))) {
```
Cyan4973: Use parenthesis to ensure `&&` and `&` are not mixed at the same level.

lib/compress/zstd_fast.c
```diff
@@ -102,19 +162,46 @@ size_t ZSTD_compressBlock_fast_generic(
...
+                        anchor = ip;
+                        continue;
+                    }
+                    break;
+                }
+            }
+
+            if (dictMode == ZSTD_noDict) {
+                while ( (ip <= ilimit)
+                     && (ip - offset_2 >= istart)
```
Cyan4973: This should be already guaranteed, by elimination of too large rep-offsets at the beginning.

lib/compress/zstd_compress.c
felixhandte: My concern with that approach is that failing to copy the dict in for a large compression is potentially a very expensive mistake.

So I think in order to do that, we would need a mechanism to perform the copy if it turns out that the input is large.

Cyan4973: > potentially a very expensive mistake

How much expensive ?

felixhandte: I thought about it and I take it back. Since the dictionary only applies to the first 128K block, the cost is bounded. A quick A/B test shows that an attached dict is like 5% slower at 128KB. So that might be fine.

I'll look into this more.

Cyan4973: Note that a dictionary is presumed valid as long as input `<= windowSize`,
so this can be a lot more than 128K.

That being said, it could also be that dictionary gains become really tiny, of not meaningless, beyond a certain input size. In which case, it could be an acceptable trade off to no longer look into the dictionary beyond a certain input size, irrespective of the `windowSize`.

lib/compress/zstd_compress.c
```diff
@@ -1203,42 +1204,74 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
...
+    int attachDict = ( pledgedSrcSize <= 8 KB
```
Cyan4973: So, 8 KB is the final selected cut-off threshold ?
Do you have graphs / data showing what happens when no cut-off is done, or done later ?

felixhandte: For the time being, yes. As I work with the other levels, we'll see if it needs to be level-dependent or whether 8 KB is best for all levels. But that seemed to be the crossover point in perf at the time when I was setting that up. I got rid of those graphs, but I'll kick off a new pass now with no cut-off so we can see.

lib/compress/zstd_fast.c
```diff
@@ -73,26 +98,62 @@ size_t ZSTD_compressBlock_fast_generic(
         U32 const current = (U32)(ip-base);
         U32 const matchIndex = hashTable[h];
         const BYTE* match = base + matchIndex;
+        const U32 repIndex = current + 1 - offset_1;
+        const BYTE* repBase = (dictMode == ZSTD_dictMatchState
+                            && repIndex < prefixLowestIndex) ?
+                               dictBase - dictIndexDelta : base;
```
Cyan4973: minor: 
`dictBase - dictIndexDelta` is supposed to be UB (it's an address outside of the object).

Since `repBase` will never be dereferenced, it's unlikely to degenerate into any kind of problem, therefore if there was some kind of benefit to this (speed typically), this could be debatable.
However, in this case, `repBase` is only used once, so I don't see any benefit invoking UB behavior.

On the other hand, `repMatch = repBase + (repIndex - dictIndexDelta)` is always correct (provided `repIndex >= dictIndexDelta`, which is supposed to be the case, and can be asserted).

lib/compress/zstd_fast.c
```diff
@@ -73,26 +98,62 @@ size_t ZSTD_compressBlock_fast_generic(
...
             ZSTD_storeSeq(seqStore, ip-anchor, anchor, 0, mLength-MINMATCH);
-        } else {
-            if ( (matchIndex <= lowestIndex)
-              || (MEM_read32(match) != MEM_read32(ip)) ) {
+        } else if ( (matchIndex <= prefixLowestIndex)
+                 || (MEM_read32(match) != MEM_read32(ip)) ) {
+            if (dictMode == ZSTD_dictMatchState) {
+                U32 const dictMatchIndex = dictHashTable[h];
+                const BYTE* dictMatch = dictBase + dictMatchIndex;
+                if (dictMatchIndex <= dictLowestIndex ||
```
terrelln: I think `dictLowestIndex` doesn't work with `forceWindow = 1` as it is written now.

lib/compress/zstd_fast.c
```diff
@@ -102,19 +163,44 @@ size_t ZSTD_compressBlock_fast_generic(
...
-                ZSTD_storeSeq(seqStore, 0, anchor, 0, rLength-MINMATCH);
-                ip += rLength;
-                anchor = ip;
-                continue;   /* faster when present ... (?) */
-    }   }   }
+            if (dictMode == ZSTD_dictMatchState) {
+                while (ip <= ilimit) {
+                    U32 const current2 = (U32)(ip-base);
+                    U32 const repIndex2 = current2 - offset_2;
+                    const BYTE* repMatch2 = dictMode == ZSTD_dictMatchState
```
terrelln: We don't need this ternary operator.

lib/compress/zstd_fast.c
```diff
@@ -73,26 +98,62 @@ size_t ZSTD_compressBlock_fast_generic(
...
+        const U32 repIndex = current + 1 - offset_1;
+        const BYTE* repMatch = (dictMode == ZSTD_dictMatchState
+                            && repIndex < prefixLowestIndex) ?
+                               dictBase + (repIndex - dictIndexDelta) :
+                               base + repIndex;
         hashTable[h] = current;   /* update hash table */
 
-        if ((offset_1 > 0) & (MEM_read32(ip+1-offset_1) == MEM_read32(ip+1))) {
+        if (dictMode == ZSTD_dictMatchState
+            && ((U32)((prefixLowestIndex-1) - repIndex) >= 3 /* intentional underflow */)
```
terrelln: I *think* it is impossible to have invalid repcodes when we are using a cdict with `forceWindow == 0`, since we check them during cdict construction. In which case we could turn this into an assert, and add some extra asserts to the beginning, and remove this check, which might get us some speed. If it doesn't we should leave it though.

lib/compress/zstd_fast.c
```diff
@@ -45,23 +45,48 @@ FORCE_INLINE_TEMPLATE
...
+    assert(dictMode != ZSTD_dictMatchState
+        || prefixLowestIndex >= (U32)(dictEnd - dictBase));
+
     /* init */
-    ip += (ip==lowest);
-    {   U32 const maxRep = (U32)(ip-lowest);
+    ip += (ip - prefixLowest + dictEnd - dictLowest == 0);
+    {   U32 const maxRep = dictMode == ZSTD_dictMatchState ?
+                           (U32)(ip - prefixLowest + dictEnd - dictLowest) :
+                           (U32)(ip - prefixLowest);
```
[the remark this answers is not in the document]
Cyan4973: If repCodes are necessarily valid by construction when `dictMode == ZSTD_dictMatchState`,
nonetheless `assert()` this condition.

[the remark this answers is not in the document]
felixhandte: Ah, I see what you mean. Because, in dictMatchState, the actual repCode checks don't check whether offset > 0. Ok. I'll assert.

lib/compress/zstd_fast.c
felixhandte: This check isn't to see if it's out of bounds, it's to see whether the 4 byte read is split across the end of the dictionary and the beginning of the prefix.

lib/compress/zstd_fast.c
felixhandte: So, the way `forceWindow` works with attached dictionaries is that if any part of the dictionary will be unreachable in the current block, the whole dictionary is made inaccessible. This overreaction is necessary because the dictionary is located right under `lowLimit`, so if `lowLimit` moves away from the actual starting offset of the blob, dictionary offsets will be incorrect.

Cyan4973: I think we just need to make sure both modes are not triggered together.
`forceWindow` is mostly there for `ZSTDMT`, which uses the "raw prefix" mode to simulate continuity, and where "dictionary"content is loaded directly into working context, and has no built-in statistics.
So this is a very different use case, which does not need to be compatible with `CDict` attached dictionaries.

lib/compress/zstd_fast.c
```diff
@@ -45,26 +45,57 @@ FORCE_INLINE_TEMPLATE
...
+    const U32 dictLowestIndex      = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.dictLimit : 0;
+    const BYTE* const dictBase     = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.base : NULL;
+    const BYTE* const dictLowest   = dictMode == ZSTD_dictMatchState ?
+                                     dictBase + dictLowestIndex : NULL;
+    const BYTE* const dictEnd      = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.nextSrc : NULL;
+    const U32 dictIndexDelta       = dictMode == ZSTD_dictMatchState ?
+                                     ms->loadedDictEnd - (U32)(dictEnd - dictBase) :
```
terrelln: Why `loadedDictEnd` instead of `prefixLowestIndex`, shouldn't they be the same?

felixhandte: `prefixLowestIndex`, i.e., `ms->window.dictLimit` gets slid forward in a lot of places, and in general, doesn't necessarily refer to the end of the dictionary/start of the stream. Granted, anywhere where we slide `lowLimit` or `dictLimit` forward is also a place we should stop using an attached dict. This seems more stable to me though.

PR #1123, opened 2018-05-08
lib/Makefile: create include directory before headers installation

baruchsiach:
Make sure that $(INCLUDEDIR) exists before copying the headers there.
Otherwise, the contest of header files is copied over
$(DESTDIR)$(INCLUDEDIR), making it a regular file.

While at it, remove $(DESTDIR)$(INCLUDEDIR) from the list of directories
to create in the install-pc target. The install-pc target does not need
this directory.


terrelln:
Looks good, but we should probably do the same for `$(LIBDIR)` in `install-shared` and `install-static`.

PR #1138, opened 2018-05-16
ZSTD_dfast: Support Searching the Dictionary Context In-Place

felixhandte:


lib/compress/zstd_compress.c
```diff
@@ -1203,42 +1204,78 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
                             U64 pledgedSrcSize,
                             ZSTD_buffered_policy_e zbuff)
 {
+    /* We have a choice between copying the dictionary context into the working
+     * context, or referencing the dictionary context from the working context
+     * in-place. We decide here which strategy to use. */
+    int attachDict = ( pledgedSrcSize <= 8 KB
```
Cyan4973: OK, so the in-place/copy threshold is applied here, irrespective of any compression level and strategy.

Given the relatively large differences in memory budget between strategies, I would expect the threshold to reflect that.

This could be classified as a later optimization.

Cyan4973:
The PR has a lot of common code updates with  the `fast` one.
I haven't noticed any obvious issue, so it looks good to me.

2 things you could do to improve the PR : 
- Publish some results. We want to know how much gain this PR provides compared to previous version.
We also want to control that the normal "no dictionary" path is safe of any regression.

- This PR is built on top of the `fast` one (it has to), but an unfortunate side effect is that it presents code changes from both patches, so it's difficult to review only the changes due to this PR.
Besides, this is no longer true, since `dev` has been updated with the `fast` path. 
Try to update your PR so that only the changes relevant to `dFast` are displayed.
Note: so far, I've only reviewed `zstd_double_fast.c` in an effort to avoid such confusion.

lib/compress/zstd_double_fast.c
```diff
@@ -86,34 +113,75 @@ size_t ZSTD_compressBlock_doubleFast_generic(
...
                     offset = (U32)(ip-matchL3);
-                    while (((ip>anchor) & (matchL3>lowest)) && (ip[-1] == matchL3[-1])) { ip--; matchL3--; mLength++; } /* catch up */
+                    while (((ip>anchor) & (matchL3>prefixLowest)) && (ip[-1] == matchL3[-1])) { ip--; matchL3--; mLength++; } /* catch up */
                 } else {
                     mLength = ZSTD_count(ip+4, match+4, iend) + 4;
                     offset = (U32)(ip-match);
-                    while (((ip>anchor) & (match>lowest)) && (ip[-1] == match[-1])) { ip--; match--; mLength++; } /* catch up */
+                    while (((ip>anchor) & (match>prefixLowest)) && (ip[-1] == match[-1])) { ip--; match--; mLength++; } /* catch up */
+                }
+            } else if (dictMode == ZSTD_dictMatchState) {
```
terrelln: Where else have you tried to put this?

Have you tried putting the long dict check after the long check, and the short dict check after the short check?

felixhandte: Yes, I have an experimental branch, `zstd-dfast-in-place-dict-test-1`, that uses `goto`s for flow control rather than ifs and this lets me match the  ([comparison](https://github.com/felixhandte/zstd/compare/zstd-dfast-in-place-dict...felixhandte:zstd-dfast-in-place-dict-test-1?expand=1)).

felixhandte:
Here are a bunch of benchmark images. I didn't want to try to pick one text to represent them all, so here's all of them (for gcc-7.2). Red is `dev`, blue is this branch, `zstd-dfast-in-place-dict`, and green is the `goto` fork, `zstd-dfast-in-place-dict-goto`.

montage-dickens-gcc-7 2-all-brs-all-cls
montage-mozilla-gcc-7 2-all-brs-all-cls
montage-mr-gcc-7 2-all-brs-all-cls
montage-nci-gcc-7 2-all-brs-all-cls
montage-ooffice-gcc-7 2-all-brs-all-cls
montage-osdb-gcc-7 2-all-brs-all-cls
montage-reymont-gcc-7 2-all-brs-all-cls
montage-samba-gcc-7 2-all-brs-all-cls
montage-sao-gcc-7 2-all-brs-all-cls
montage-webster-gcc-7 2-all-brs-all-cls
montage-xml-gcc-7 2-all-brs-all-cls
montage-x-ray-gcc-7 2-all-brs-all-cls

Cyan4973:
I see 2 lines of same color on each graph.
Do you have levels 3 and 4 blended together into each graph ?

felixhandte:
Ah yes, forgot to mention that. Yes.

Cyan4973:
The wins from the new strategy when `usingCDict` are fairly huge,
and the large drop of performance at 8K is likely the result of coming back to previous copy strategy when crossing that threshold.

It's a strong argument in favor of selecting a different threshold.

felixhandte:
Yes, I have a diff that adds per-strategy cutoffs. Gotta benchmark it.

felixhandte:
Closing in favor of #1151.

PR #1139, opened 2018-05-18
PREFETCH (or __builtin_prefetch) did not exist before gcc 4.

fbrosson:


fbrosson:
Thanks @Cyan4973  for the pointer to `lib/common/compiler.h`.
(I guess I had not searched hard enough... ;)

fbrosson:
BTW, I forgot to mention that I am **_not_** 100% about which version of gcc started providing `__builtin_prefetch`, but I remember having read a comment somewhere stating that it was probably with gcc 4. It might be 4.x with x>0, but at least this commit will help with gcc < 4 :)

Cyan4973:
Yeah, finding the exact version at which a feature is present and debugged is quite a pain.
`gcc` doesn't make any effort to clearly document it (and this is not limited to this builtin).
For portability, such documentation oversight is a serious problem.

Not that, looking around, I found documentation on this version : 
https://gcc.gnu.org/onlinedocs/gcc-3.1/gcc/Other-Builtins.html

which seems to imply that `__builtin_prefetch()` was at least available as far back as 3.1.
Though, without direct tests, it's unclear if this intrinsic works appropriately at this version for all supported targets, nor if the linked documentation is simply correct.

Cyan4973:
Nope. 3.1 is the lowest version we can rely on, according to available documentation.
It could even be that 3.1 is buggy and we should rather limit to 3.2 or later,
but this is not documented. So without direct tests, we can't do better.

lib/common/compiler.h
```diff
@@ -92,7 +92,7 @@
 #if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_I86))  /* _mm_prefetch() is not defined outside of x86/x64 */
 #  include <mmintrin.h>   /* https://msdn.microsoft.com/fr-fr/library/84szxsww(v=vs.90).aspx */
 #  define PREFETCH(ptr)   _mm_prefetch((const char*)ptr, _MM_HINT_T0)
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) && ((__GNUC__ >= 4) || (__GNUC__ == 3) && (__GNUC_MINOR__ >= 1))
```
Cyan4973: add parenthesis for `|| ( (__GNUC__ == 3) && (__GNUC_MINOR__ >= 1) )`

fbrosson: Would it be better to test against 3.0.3 with this change?
```diff
-#elif defined(__GNUC__) && ((__GNUC__ >= 4) || (__GNUC__ == 3) && (__GNUC_MINOR__ >= 1))
+#elif defined(__GNUC__) && (__GNUC__ >= 4 || __GNUC__ == 3 && (__GNUC_MINOR__ >= 1 || __GNUC_MINOR__ == 0 && defined(__GNUC_PATCHLEVEL__) && __GNUC_PATCHLEVEL__ >= 3))
```
BTW, I dropped some unneeded parenthesis but I need to check this.

Hmm, after all, shall I just change the commit message to `__builtin_prefetch did probably not exist before gcc 3.0.3.` and keep a simple test for `3.1` instead of `3.0.3`?

Cyan4973: The PR title is fine.
Stick to 3.1, and do not rely on operator priority to interpret the version test.
Parenthesis are fine. They help ensure the test performed is the wanted one.

fbrosson: I'm so sorry, I followed the link for gcc 3.1 link thinking I had followed the one for gcc 3.0.3!!!
(So when I found `builtin_prefetch` I wrongly though it was available in 3.0.3. :(

So everything is already OK. The commit title is OK as well as test.
Hmm, should I drop some parenthesis in the test? I guess this should be OK:
```
+#elif defined(__GNUC__) && (__GNUC__ >= 4 || __GNUC__ == 3 && __GNUC_MINOR__ >= 1)
```

Cyan4973: no you should ___add___ parenthesis in the test, anywhere there could be a doubt on operator priority.
At the very least : 
```
#elif defined(__GNUC__) && (__GNUC__ >= 4 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))
```

fbrosson: Done.
BTW, shall I squash my 2 commits?

Cyan4973: As you wish.
It's a bit better if they are squashed,
but not an absolute necessity.

fbrosson: Yep, but some purists like to live dangerously, putting parenthesis only where they are strictly required.
I tend to choose the safe side, but maybe we should replace
```
#elif defined(__GNUC__) && ((__GNUC__ >= 4) || ( (__GNUC__ == 3) && (__GNUC_MINOR__ >= 1) ) )
```
by
```
#elif defined(__GNUC__) && (__GNUC__ >= 4 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))
```
What would you recommend?

Cyan4973: either version is good to me

Cyan4973: Here also, I'm fine with whatever works best for HaikuPorts.
Squashing is slightly better, but I don't mind if it's not ,
so if there is a reason to prefer the non-squashed version, I'm fine with it and can merge it.

fbrosson: I also prefer the squashed version. It's there in a minute.

fbrosson: I've added a space after one of the opening parenthesis, otherwise it would break the symetry with the space before it"s corresponding closing parenthesis ;)

lib/common/compiler.h
```diff
@@ -92,7 +92,7 @@
 #if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_I86))  /* _mm_prefetch() is not defined outside of x86/x64 */
 #  include <mmintrin.h>   /* https://msdn.microsoft.com/fr-fr/library/84szxsww(v=vs.90).aspx */
 #  define PREFETCH(ptr)   _mm_prefetch((const char*)ptr, _MM_HINT_T0)
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) && ( (__GNUC__ >= 4) || ( (__GNUC__ == 3) && (__GNUC_MINOR__ >= 1) ) )
```
terrelln: Does this still work for clang?

fbrosson: This change only impacts compilers which define `__GNUC__`, and since I guess clang does not define it, I think we are safe here.
That said, I might be wrong about clang not defining `__GNUC__`. Maybe it does, although this would be counter-intuitive.

fbrosson: > This change only impacts compilers which define __GNUC__

To be more precise, I should say that this change is neutral to all compilers which do not define `__GNUC__`.

terrelln: It should work for clang https://godbolt.org/g/yaJ8Q5, it sets `__GNUC__ == 4` since clang-3.0 at least, which is good enough.

terrelln: @fbrosson clang pretends to be gcc, and it supports all or nearly all gcc extensions.

PR #1140, opened 2018-05-18
Drop colon in asm snippet to make old versions of gcc happy.

fbrosson:


lib/common/cpu.h
```diff
@@ -73,7 +73,7 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
           "popl %%ebx\n\t"
           : "=a"(f1a), "=c"(f1c), "=d"(f1d)
           : "a"(1)
-          :);
+          );
```
fbrosson: One of Haiku's platform, called `x86_gcc2`, uses a custom gcc `2.95.3` (aka gcc2 in Haiku speak) to achieve binary compatibility with closed-source software built for BeOS.
And our gcc2 (and probably non custom but old versions of gcc too) error on `:);`.
So this tiny patch would be handy for HaikuPorts :)
(That's OK if this PR does not get merged.)

BTW, should I fix this with a different style? I mean:
```diff
-          : "a"(1)
-          :);
+          : "a"(1));
```
Thanks!

Cyan4973:
I think that's fine, it passes all tests, and can be merged.
Yes, the second version proposed in comment seems nicer.

PR #1141, opened 2018-05-18
Random stuff on High Compression mode

Cyan4973:
This is a set of patches around high compression mode, 
developed in the same branch and provided here as a single PR,
though they could be separated into 3 different batches.
PR created mostly for reviewing, as I'm not sure of the value of merging them at this stage.

#### First patch is about removing the "static price" branch in the optimal parser.
   
This branch was only used with dictionary compression, on the first block.
    It consists in taking the costs directly from the dictionary, 
    and use them to evaluate choices during parsing stage.

This is replaced by "init_from_dict + dynamic price".
    The main idea was that it would result in mostly the same result as "static price",
    since, for small data, new statistics have little to no time to influence statistics initialized from dictionary.
    This is mostly true, though surprisingly, it resulted in _improved_ compression ratio (slightly) compared to "static price" on my sample set.

So it's all good ? Well, yes, in some ways, but it's also a bit hard to accept side-effects which are not well controlled.
The result is hard to believe because for small data, the final statistics will be selected from dictionary, so using them for price estimation _should_ lead to more optimal choices.

It seems the issue is that the parser does not work with "exact statistics". Rather it slightly favors uncommon symbols in order for them to have a better chance to be selected, which experimentally leads to better parsing choices (since the optimal parser is a single-scan strategy, there's no time to retrofit earlier choices with new prices).

Anyway, strange but positive outcome.
I think this patch is okay, if only because it simplifies the code (which was the main motivation).

#### Second patch introduces fractional bit cost evaluation

The optimal parser has always been working using bit approximation of prices.
A symbol's cost is evaluated to 2 bits, or 3 bits, but never 2.5 bits,
even if the entropy stage (FSE) is able to compress using fractional bits.

There is a good reason for that : as mentioned previously, all statistics are "approximate", and effectively a bit biased to avoid rejecting too strongly less common candidates. Therefore, since statistics are not "correct", what good is it to make them "accurate" ?

It turns out bit fractional cost evaluation does help compression a little (~0.2%), though at a measurable speed loss (~5%) due to extra complexity in the cost function.
The trade off is clearly not great. But when willing to reach highest compression ratio, it stills looks like an acceptable price to pay.

The suggestion done in this patch is to enable bit fractional evaluation for highest strategy `btultra` but keep current bit approximation for the faster variant `btopt`.

#### Third patch introduces first block statistics evaluation

A major drawback of dynamic price evaluation is that it starts from nothing, with some default "empty" statistics, and then use real input to fix statistics. This approach needs time to converge towards more-or-less correct price estimations. 
The convergence is reasonably fast. I believe it starts to provide approximately correct evaluations around 4KB, so that's not bad.
Still, this bad beginning will continue to influence parsing choices for some time before being overtaken by real statistics. It ends up negatively influencing the compression ratio.

Another approach, proposed in this patch, is to use a first pass to "seed" initial statistics, in order to make better choices during the second pass, improving compression ratio.

This "works", although benefits are disappointingly small. 
I used "small" inputs, less than 1 block (< 128 KB) to measure the efficiency of the scheme.
I measure a relatively "flat" compression gain of ~0.5%, at all input sizes, from 4KB to 64 KB.
This is much less than the ~4% gains observed with dictionary compression when using a similar strategy to seed initial statistics with statistics from dictionary.

The price to pay though is pretty hefty : small inputs (< 128 KB) now cost 2x more time to compress.
Even at high compression settings, it's pretty high tag for the small benefit.
Larger inputs will pay a proportionally less important cost, since this strategy is only used on the first block. But the gain will also be more or less limited to this first block. 0.5% of 128 KB, that's basically a 0.5 KB gain, so almost nothing.

So I'm a bit puzzled about this one.
Plus, it's not as if it was a sure gain : depending on files, sometimes it ends up being a more substantial 1% gain, and sometimes nothing, or even a small loss. 0.5% is just an average.

I suspect it can still have a role in the "let's get the highest possible compression ratio" game, but that's all. In which case, it could be useful to create a new strategy just for it.
The idea would be to create a new strategy `btmax`, which would be reserved to the highest compression level, and would be the only one to use this trick.

For discussion.


terrelln:
* The dynamic prices part looks good to me. I'm a bit puzzled as to why it works better though, but the simplification is welcome. Hopefully this is something that we can understand better with the new parser.
* I think the fractional bit-cost 5% speed for 0.2% ratio is a welcome trade-off for `btultra`.
* I'm not so sure about 2-pass parser. Unless we have a use case, other than benchmarking, that benefits from the 2x slower for 0.5% ratio, I think we're better off leaving `btultra` as a single-pass algorithm, and use the new parser for benchmarking. The code is very simple, but it seems fragile, since nothing else expects to support 2-pass parsing, and it has to reset external state.

Cyan4973:
I can disable the 2-passes strategy, either by ensuring that the test is always false, or by disabling the relevant code through macro guard.
This way, it can be easily re-enabled in the future, or just for temporary tests.

I agree it's not great that it has to reset external state.

Cyan4973:
I used this opportunity to make a more thorough measurement
of the fractional bit evaluation bit.

Here is a side by side evaluation 
of size compressed by `dev` and this PR at level -22 : 

filename | `dev` | frac bit | improvement
-- | -- | -- | --
enwik2 | 91 | 91 | 0.00%
enwik3 | 342 | 341 | -0.29%
enwik4 | 3,672 | 3,661 | -0.30%
obj1 | 10,027 | 9,960 | -0.67%
progl | 15,239 | 15,205 | -0.22%
enwik5 | 33,993 | 33,906 | -0.26%
geo | 66,551 | 66,415 | -0.20%
obj2 | 70,782 | 70,571 | -0.30%
pic | 44,172 | 43,723 | -1.02%
enwik6 | 300,929 | 300,358 | -0.19%
calgary.tar | 899,002 | 897,430 | -0.17%
xml | 453,960 | 453,051 | -0.20%
ooffice | 2,615,801 | 2,608,016 | -0.30%
sao | 5,015,054 | 5,024,841 | ___0.20%___
x-ray | 5,162,564 | 5,151,702 | -0.21%
mr | 3,116,877 | 3,123,810 | ___0.22%___
enwik7 | 2,798,643 | 2,794,339 | -0.15%
osdb | 3,067,338 | 3,074,601 | ___0.24%___
samba | 3,911,899 | 3,908,797 | -0.08%
enwik8 | 25,379,959 | 25,341,859 | -0.15%

As one can see, the improvement is generally in the 0.2% range.
There are only a few exceptions which do better.
More importantly, in a few cases (underlined in ___bold___), fractional bit evaluation leads to worse compression ratio. So it's not a guaranteed win.

I'm unsure if there is a way to improve this situation.

Cyan4973:
I slightly modified the cost formula to gently nudge the selection towards less sequences (presumed better for decompression speed). It also helps reducing impact of unfavorable cases.

New results are as follows : 

filename | `dev` | frac bit | size reduction
-- | -- | -- | --
enwik2 | 91 | 91 | 0.00%
enwik3 | 342 | 341 | -0.29%
enwik4 | 3,672 | 3,664 | -0.22%
obj1 | 10,027 | 9,979 | -0.48%
progl | 15,239 | 15,199 | -0.26%
enwik5 | 33,993 | 33,899 | -0.28%
geo | 66,551 | 66,771 | ___+0.33%___
obj2 | 70,782 | 70,595 | -0.26%
pic | 44,172 | 43,623 | -1.24%
enwik6 | 300,929 | 300,299 | -0.21%
calgary.tar | 899,002 | 897,184 | -0.20%
xml | 453,960 | 453,006 | -0.21%
ooffice | 2,615,801 | 2,608,563 | -0.28%
sao | 5,015,054 | 5,024,188 | ___+0.18%___
x-ray | 5,162,564 | 5,153,661 | -0.17%
mr | 3,116,877 | 3,103,751 | -0.42%
enwik7 | 2,798,643 | 2,794,015 | -0.17%
osdb | 3,067,338 | 3,069,567 | _+0.07%_
samba | 3,911,899 | 3,910,387 | -0.04%
enwik8 | 25,379,959 | 25,340,680 | -0.15%
silesia.tar | 52,801,490 | 52,750,281 | -0.10%

This feels like a better trade-off.

PR #1143, opened 2018-05-19
Update table of compression levels

Cyan4973:
The main difference is level 19 using `btultra` strategy
(which is now badly named, since it's no longer reserved for `--ultra` levels anymore)
so that we can make level 19 a "strongest compression level at window size <= 8 MB".

Several other levels have been updated, to reflect improvements in intermediate strategies,
and smooth the speed/compression curve on a slightly larger scale, due to level 1 being faster, and level 19 being stronger.


lib/compress/zstd_compress.c
```diff
@@ -3434,101 +3434,101 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
...
     { 14, 15, 14,  6,  3, 16, ZSTD_btopt   },  /* level 12.*/
     { 14, 15, 14,  6,  3, 24, ZSTD_btopt   },  /* level 13.*/
     { 14, 15, 15,  6,  3, 48, ZSTD_btopt   },  /* level 14.*/
     { 14, 15, 15,  6,  3, 64, ZSTD_btopt   },  /* level 15.*/
     { 14, 15, 15,  6,  3, 96, ZSTD_btopt   },  /* level 16.*/
     { 14, 15, 15,  6,  3,128, ZSTD_btopt   },  /* level 17.*/
-    { 14, 15, 15,  6,  3,256, ZSTD_btopt   },  /* level 18.*/
-    { 14, 15, 15,  7,  3,256, ZSTD_btopt   },  /* level 19.*/
+    { 14, 15, 15,  7,  3,256, ZSTD_btopt   },  /* level 18.*/
+    { 14, 15, 15,  6,  3,256, ZSTD_btultra },  /* level 19.*/
```
terrelln: It seems a bit strange to see search log get smaller for `btultra`.

Cyan4973: `btultra` is significantly slower than `btopt`, so even with a smaller `slog`, it's measurably slower, which is quite a thing for 16K blocks, as the amount of work to do is quickly maxed out.

On my latest devserver:
```
17#silesia.tar       : 211957760 ->  73223542 (2.895),  7.33 MB/s , 430.4 MB/s
18#silesia.tar       : 211957760 ->  73221967 (2.895),  7.17 MB/s , 426.9 MB/s
19#silesia.tar       : 211957760 ->  73167187 (2.897),  5.44 MB/s , 425.2 MB/s
```

Apparently, I should increase level 18's `slog` even more so that it takes a more intermediate position between 17 and 19.

PR #1146, opened 2018-05-23
[zstd] Fix decompression edge case

terrelln:
This edge case is only possible with the new optimal encoding selector,
since before zstd would always choose `set_basic` for small numbers of
sequences.

Fix `FSE_readNCount()` to support buffers < 4 bytes.

Credit to OSS-Fuzz


terrelln:
Looks like there is another bug in `FSE_readNCount()`, I'm looking into it.

PR #1144, opened 2018-05-22
Approximate FSE encoding costs for selection

terrelln:
Estimate the cost for using FSE modes `set_basic`, `set_compressed`, and
`set_repeat`, and select the one with the lowest cost.

* The cost of `set_basic` is computed using the cross-entropy cost
  function `ZSTD_crossEntropyCost()`, using the normalized default count
  and the count.
* The cost of `set_repeat` is computed using `FSE_bitCost()`. We check the
  previous table to see if it is able to represent the distribution.
* The cost of `set_compressed` is computed with the entropy cost function
  `ZSTD_entropyCost()`, together with the cost of writing the normalized
  count `ZSTD_NCountCost()`.

# Benchmarks

## Small data at high compression levels

Silesia800: `zstd silesia.tar -B800 -b19e19i15`
GitHub: `zstd -r github/ -b19e19i15`

Silesia800 | Ratio | Compression MB/s | Decompression MB/s
-- | -- | -- | --
Before | 1.897 | 10.6 | 228.2
After | 1.930 | 10.5 | 208.6

GitHub | Ratio | Compression MB/s | Decompression MB/s
-- | -- | -- | --
Before | 3.002 | 10.1 | 265.5
After | 3.060 | 9.70 | 251.2

## Dictionary with well adapted tables

"dict6" and "dict19" are created with the commands:
```
zstd --train-cover=d=8,steps=32 -T2 -r github/  -6 -o dict6
zstd --train-cover=d=8,steps=32 -T2 -r github/  -19 -o dict19
zstd --train-cover=d=8,steps=32 -T2 -r hg-changelog/  -6 -o hg-changelog-dict
zstd --train-cover=d=8,steps=32 -T2 -r hg-commands/  -19 -o hg-commands-dict
zstd --train-cover=d=8,steps=32 -T2 -r hg-manifest/  -19 -o hg-manifest-dict
```
GitHub6: `zstd -r github/ -b6e6i15 -D dict6`
GitHub19: `zstd -r github/ -b19e19i15 -D dict19`
hg-changelog: `zstd -r hg-changelog/ -b19e19i15 -D hg-changelog-dict`
hg-commands: `zstd -r hg-commands/ -b19e19i15 -D hg-commands-dict`
hg-manifest: `zstd -r hg-manifest/ -b19e19i15 -D hg-manifest-dict`

GitHub6 | Ratio | Compression MB/s | Decompression MB/s
-- | -- | -- | --
Before | 10.71 | 2.00 | 4.81.2
[...]

PR #1147, opened 2018-05-23
Work around bug in zstd decoder

terrelln:
Pull request #1144 exercised a new path in the zstd decoder that proved to
be buggy. Avoid the extremely rare bug by emitting an uncompressed block.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1983,8 +1983,7 @@ MEM_STATIC size_t ZSTD_compressSequences_internal(seqStore_t* seqStorePtr,
     BYTE* op = ostart;
     size_t const nbSeq = seqStorePtr->sequences - seqStorePtr->sequencesStart;
     BYTE* seqHead;
-    size_t const kMinLastCountSize = 4;
-    size_t lastCountSize = kMinLastCountSize;
+    BYTE* lastNCount = NULL;
 
     ZSTD_STATIC_ASSERT(HUF_WORKSPACE_SIZE >= (1<<MAX(MLFSELog,LLFSELog)));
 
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -2050,6 +2053,10 @@ MEM_STATIC size_t ZSTD_compressSequences_internal(seqStore_t* seqStorePtr,
                     workspace, HUF_WORKSPACE_SIZE);
             if (ZSTD_isError(countSize)) return countSize;
             op += countSize;
+            if (Offtype == set_compressed)
+                lastCountSize = countSize;
+            else if (countSize > 0)
```
Cyan4973: I would have expected : 
`else lastCountSize += countSize;`

lib/compress/zstd_compress.c
```diff
@@ -2077,6 +2088,27 @@ MEM_STATIC size_t ZSTD_compressSequences_internal(seqStore_t* seqStorePtr,
...
+        assert(lastCountSize == 0 || bitstreamSize > 0);
+        /* zstd versions <= 1.3.4 mistakenly report corruption when
+         * FSE_readNCount() recieves a buffer < 4 bytes.
+         * Fixed by https://github.com/facebook/zstd/pull/1146.
+         * This can happen when the last set_compressed table present is 2
+         * bytes and the bitstream is only one byte.
+         * In this exceedingly rare case, we will simply emit an uncompressed
+         * block, since it isn't worth optimizing.
+         */
+        if (bitstreamSize == 1 && lastCountSize == 2) {
```
Cyan4973: I would have expected : 
`if ((lastCountSize) && (bitstreamSize + lastCountSize <= 3))`

lib/compress/zstd_compress.c
terrelln: That would work too, and it makes the check a bit simpler.

PR #1149, opened 2018-05-24
Small fixes to fuzz.py

terrelln:
* Fix a typo.
* Make `zip` work for larger corpora, tested it by updating the regression test corpora.
* Define `BIT_DEBUG` in addition to `ZSTD_DEBUG`.
* Increase the maximum allowed file size to 128 MB.

### DIFF
--- a/tests/fuzz/fuzz.py
+++ b/tests/fuzz/fuzz.py
@@ -265,7 +265,7 @@ def build_parser(args):
         '--disable-fuzzing-mode',
         dest='fuzzing_mode',
         action='store_false',
-        help='Do not define FUZZING_BUILD_MORE_UNSAFE_FOR_PRODUCTION')
+        help='Do not define FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION')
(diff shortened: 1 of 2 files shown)


terrelln:
CC @Cyan4973 looks like there is a bug in `FSE_bitCost()` (or I'm using it wrong [here]()). Maybe this is why fractional costs are worse?

This assert is failing now that `BIT_DEBUG` is enabled https://github.com/facebook/zstd/blob/dev/lib/common/fse.h#L600. From this caller https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1666.

Cyan4973:
I'll look into it.
I'm suspecting a problem when `freq[s]==2`.
To be investigated ...

It's a good idea to enable `BIT_DEBUG` for these tests

Cyan4973:
It's actually happening when `freq[s]==0`.
There's a need to specify a behavior for this case.

Cyan4973:
`assert` fix merged into `dev`

PR #1151, opened 2018-05-25
ZSTD_dfast: Support Searching the Dictionary Context In-Place (Alternate `goto` Implementation)

felixhandte:
This PR has some additional refactoring on top of #1138 that rewrites `ZSTD_compressBlock_doubleFast_generic` to control flow with `goto`s rather than deeply nested if/else blocks. In doing so, it is able to more closely match the existing compression strategy.

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -69,6 +69,26 @@ size_t ZSTD_compressBlock_doubleFast_generic(
     U32 offset_1=rep[0], offset_2=rep[1];
     U32 offsetSaved = 0;
 
+    const ZSTD_matchState_t* const dms = ms->dictMatchState;
+    const U32* const dictHashLong  = dictMode == ZSTD_dictMatchState ?
+                                     dms->hashTable : NULL;
+    const U32* const dictHashSmall = dictMode == ZSTD_dictMatchState ?
+                                     dms->hashTable : NULL;
+    const U32 lowestDictIndex      = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.dictLimit : 0;
+    const BYTE* const dictBase     = dictMode == ZSTD_dictMatchState ?
+                                     dms->window.base : NULL;
+    const BYTE* const dictLowest   = dictMode == ZSTD_dictMatchState ?
+                                     dictBase + lowestDictIndex : NULL;
+    const BYTE* const dictEnd      = dictMode == ZSTD_dictMatchState ?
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_double_fast.c
```diff
@@ -62,70 +62,178 @@ size_t ZSTD_compressBlock_doubleFast_generic(
...
-                if ( (matchIndexL3 > lowestIndex) && (MEM_read64(matchL3) == MEM_read64(ip+1)) ) {
-                    mLength = ZSTD_count(ip+9, matchL3+8, iend) + 8;
+            goto _match_stored;
+        }
+
+        /* check prefix long match */
+        if ( (matchIndexL > prefixLowestIndex) && (MEM_read64(matchLong) == MEM_read64(ip)) ) {
+            mLength = ZSTD_count(ip+8, matchLong+8, iend) + 8;
+            offset = (U32)(ip-matchLong);
+            while (((ip>anchor) & (matchLong>prefixLowest)) && (ip[-1] == matchLong[-1])) { ip--; matchLong--; mLength++; } /* catch up */
```
terrelln: Do you think it makes sense to require the `match` variable be set at `_match_found` and move the backwards search into `_match_found`, since it is done on every `goto`?

Cyan4973:
I agree, this version feels clearer to follow.
It's an important quality, which makes it preferable to the user version.
The small speed benefit comes on top of this.

lib/compress/zstd_double_fast.c
felixhandte: Yeah, that makes sense. I'll give it a try.

felixhandte: So, I've updated the PR with the collected backwards match extension code. Benchmark spot-checks are mixed-to-negative. When the full benchmarking run has signal I'll let you know what the impact is.

felixhandte: Benchmarks actually look good!

Cyan4973:
I've been measuring performance of this branch compared to `dev`.
Unfortunately, it seems substantially slower (almost -10%)
and compress a bit less than `dev` when using no dictionary.
Could you look into it ?

Cyan4973:
With performance fixed in latest patch,
now seems the right time to have another at the "threshold" value,
which divides the world between copy/no-copy strategy.

It's currently a single value, more targeted for `_fast` strategy.
It probably needs to be adaptive, at least on strategy, maybe depending on some other value (hashlog / chainlog maybe ?).

PR #1154, opened 2018-05-29
minor optimal parser optimization

Cyan4973:
This is mostly a simplification of the optimal parser.

It gets rid of the "cache" structure for literals.
The cache was required for performance because the previous version of optimal parser would recalculate the full cost of all literals + litLength at each position.
In this new version, this cost is simply saved into the path table.

This removes a bunch of functions,
and slightly improves speed (mostly visible for small files).

Used this opportunity to change a few variable names, for clarity.

There are some minor ramifications beyond `zstd_opt.c` as I had to improve the debugging log functionality to properly validate this patch.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -189,10 +189,13 @@ static U32 ZSTD_rawLiteralsCost(const BYTE* const literals, U32 const litLength,
         return (litLength*6) * BITCOST_MULTIPLIER;  /* 6 bit per literal - no statistic used */
 
     /* dynamic statistics */
-    {   U32 price = litLength * optPtr->litSumBasePrice;
-        U32 u;
-        for (u=0; u < litLength; u++)
-            price -= WEIGHT(optPtr->litFreq[literals[u]], optLevel);
+    {   U32 price, u;
+        for (u=0, price=0; u < litLength; u++) {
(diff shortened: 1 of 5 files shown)


lib/compress/zstd_opt.c
```diff
@@ -807,27 +747,29 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
...
+                                maxML, sufficient_len);
                     cur = 0;
-                    last_pos = 1;
+                    last_pos = ZSTD_totalLen(lastSequence);
                     goto _shortestPath;
             }   }
 
             /* set prices for first matches starting position == 0 */
-            {   U32 const literalsPrice = ZSTD_fullLiteralsCost_cached(&cachedLitPrice, anchor, litlen, optStatePtr, optLevel);
+            {   U32 const literalsPrice = opt[0].price + ZSTD_litLengthPrice(0, optStatePtr, optLevel);
```
terrelln: To make sure I'm understanding this correctly, this `literalsPrice` now includes the literals before the match, and the 0 lengthed literals after the match, right?

Cyan4973: At this place in the code, `literalsPrice` includes all literals before the match,
and the complete price of `litLength`.

Previously, `opt[0].price` would include all literals before the match, and the _contribution_ of `litLength`, which means it was receiving a rebate, corresponding to a fixed amount, the price of a `litLength == 0`. We simply add it.

PR #1159, opened 2018-05-31
Unknown Suffix Error

GeorgeLu97:
Changed so only compiled formats are printed in list of supported extensions

Test Plan: Compiled w/o support for lzma/xz/lz4, did in fact print out the right thing.

zstd -d blah.abc
zstd: blah.abc: unknown suffix (.zst/.gz expected) -- ignored

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1754,16 +1754,17 @@ int FIO_decompressMultipleFilenames(const char** srcNamesTable, unsigned nbFiles
                     && strcmp(suffixPtr, ZSTD_EXTENSION)
                     && strcmp(suffixPtr, LZMA_EXTENSION)
                     && strcmp(suffixPtr, LZ4_EXTENSION)) ) {
-                char suffixlist[50] = ZSTD_EXTENSION;
-                #ifdef ZSTD_GZCOMPRESS
-                    strcat(suffixlist, "/" GZ_EXTENSION);
+                const char* suffixlist = ZSTD_EXTENSION
+                #ifdef ZSTD_GZCOMPRESS 
+                    "/" GZ_EXTENSION
                 #endif
                 #ifdef ZSTD_LZMACOMPRESS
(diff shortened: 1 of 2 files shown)


programs/fileio.c
```diff
@@ -1754,8 +1754,18 @@ int FIO_decompressMultipleFilenames(const char** srcNamesTable, unsigned nbFiles
                     && strcmp(suffixPtr, ZSTD_EXTENSION)
                     && strcmp(suffixPtr, LZMA_EXTENSION)
                     && strcmp(suffixPtr, LZ4_EXTENSION)) ) {
-                DISPLAYLEVEL(1, "zstd: %s: unknown suffix (%s/%s/%s/%s/%s expected) -- ignored \n",
-                             srcFileName, GZ_EXTENSION, XZ_EXTENSION, ZSTD_EXTENSION, LZMA_EXTENSION, LZ4_EXTENSION);
+                char suffixlist[50] = ZSTD_EXTENSION;
+                #ifdef ZSTD_GZCOMPRESS
```
terrelln: C supports string literals concatenation like this:

```c
"a""b""c"
#if A
"d";
#endif
```

Instead of using `strcat()` we can use that feature.

GeorgeLu97: ok, i've removed strcats

Cyan4973:
Looks good to me.

Here also, try to add a test into `playTests.sh` that will fail when your patch is not present.

tests/playTests.sh
```diff
@@ -620,6 +620,23 @@ else
     $ECHO "lz4 mode not supported"
 fi
 
+$ECHO "\n===> suffix list test"
+
+$ZSTD -d tmp.abc 2> tmplg || echo
```
Cyan4973: I suspect you use `|| echo` in order to cancel the return value of `$ZSTD` which just failed.

There is simpler and more useful : 
`! $ZSTD -d tmp.abc`
The will reverse the success / failure signal.
And `$ZSTD -d tmp.abc` must indeed fail.

PR #1160, opened 2018-05-31
Added --format=zstd

GeorgeLu97:
for scripts

Test Plan:
Ran with new flag over an old flag
gclu-mbp:programs gclu$ ./zstd --rm --format=gzip --format=zstd blah.txt
blah.txt             :183.33%   (    12 =>     22 bytes, blah.txt.zst)         
gclu-mbp:programs gclu$ ./zstd -d --rm --format=gzip --format=zstd blah.txt.zst
blah.txt.zst        : 12 bytes
Overrides previous format flags (and doesn't give error)

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -145,6 +145,7 @@ static int usage_advanced(const char* programName)
 #ifdef UTIL_HAS_CREATEFILELIST
     DISPLAY( " -r     : operate recursively on directories \n");
 #endif
+    DISPLAY( "--format=zstd : compress files to the .zstd format \n");
 #ifdef ZSTD_GZCOMPRESS
     DISPLAY( "--format=gzip : compress files to the .gz format \n");
 #endif
@@ -500,6 +501,7 @@ int main(int argCount, const char* argv[])
                     if (!strcmp(argument, "--rm")) { FIO_setRemoveSrcFile(1); continue; }
                     if (!strcmp(argument, "--priority=rt")) { setRealTimePrio = 1; continue; }
                     if (!strcmp(argument, "--single-thread")) { nbWorkers = 0; singleThread = 1; continue; }
+                    if (!strcmp(argument, "--format=zstd")) { suffix = ZSTD_EXTENSION; FIO_setCompressionType(FIO_zstdCompression); continue; }
 #ifdef ZSTD_GZCOMPRESS
                     if (!strcmp(argument, "--format=gzip")) { suffix = GZ_EXTENSION; FIO_setCompressionType(FIO_gzipCompression); continue; }
 #endif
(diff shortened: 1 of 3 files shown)


programs/zstdcli.c
```diff
@@ -145,6 +145,7 @@ static int usage_advanced(const char* programName)
 #ifdef UTIL_HAS_CREATEFILELIST
     DISPLAY( " -r     : operate recursively on directories \n");
 #endif
+    DISPLAY( "--format=zstd : compress files to the .zstd format \n");
```
Cyan4973: nit : add to description (default)

Cyan4973:
Great, you get it.

As suggested by @terrelln, the next step is to add a test to the test suite,
which will prove that your patch work, and ensure it remains the case in the future.

You already described a suitable strategy for this test.
All you need to do is to add it to [`tests/playTests.sh`](https://github.com/facebook/zstd/blob/dev/tests/playTests.sh),
typically as part of the section checking format compatibility.

Pay attention that you need a clear signal in case of failure : 
your test shall fail when your patch is disabled or ineffective.
Typically, use command `--list` on the resulted compressed file
and parse the result to check it's using `zstd` format.

programs/zstdcli.c
Cyan4973: Also : please update the relevant section in man page (via `zstd.1.md`)

tests/playTests.sh
```diff
@@ -485,6 +485,11 @@ $ZSTD -bi0 --fast tmp1
 $ECHO "with recursive and quiet modes"
 $ZSTD -rqi1b1e2 tmp1
 
+$ECHO "\n===>  zstd compatibility tests "
+
+./datagen > tmp
+$ZSTD --format=zstd tmp 2> tmplog
+grep "zst" tmplog > $INTOVOID || die "--format=zstd not supported"
```
Cyan4973: good !

You need one last test though : 
you need to ensure that, in a case of a command like this `$ZSTD --format=gzip --format=zstd`, the final format produced is actually `zstd`.

tests/playTests.sh
```diff
@@ -485,6 +485,11 @@ $ZSTD -bi0 --fast tmp1
 $ECHO "with recursive and quiet modes"
 $ZSTD -rqi1b1e2 tmp1
 
+$ECHO "\n===>  zstd compatibility tests "
+
+./datagen > tmp
+$ZSTD --format=zstd tmp 2> tmplog
```
Cyan4973: nit: you want to add `-f` command, in case `tmp.zst` already exist in current directory, which would trigger `$ZSTD` to request an overwrite confirmation.

tests/playTests.sh
```diff
@@ -485,6 +485,11 @@ $ZSTD -bi0 --fast tmp1
...
+grep "zst" tmplog > $INTOVOID || die "--format=zstd not supported"
```
Cyan4973: This test works.
Just, if all you want to do is to check that a file named `tmp.zst` was created,
the way we do it in other equivalent tests in this same script is : 
`test -f tmp.zst`

If you want to be sure that the file `tmp.zst` was actually created by your test, and not just already present due to some earlier artefact, `rm -f tmp.zst` before compression.

PR #1161, opened 2018-06-01
Added zstdcli-style format for compression parameters in paramgrill

GeorgeLu97:
Previous: paramgrill -Sw23c23h22s06l3t048S6
Added: paramgrill --zstd=windowLog=23,chainLog=23,hashLog=22,searchLog=6,searchLength=3,targetLength=48,strategy=6
Should behave equivalently to old command (which still exists)

Test Plan:
Ran above commands, produced same result.

### DIFF
--- a/tests/paramgrill.c
+++ b/tests/paramgrill.c
@@ -996,12 +996,13 @@ int main(int argc, const char** argv)
                 if (longCommandWArg(&argument, "searchLength=") || longCommandWArg(&argument, "slen=")) { g_params.searchLength = readU32FromChar(&argument); if (argument[0]==',') { argument++; continue; } else break; }
                 if (longCommandWArg(&argument, "targetLength=") || longCommandWArg(&argument, "tlen=")) { g_params.targetLength = readU32FromChar(&argument); if (argument[0]==',') { argument++; continue; } else break; }
                 if (longCommandWArg(&argument, "strategy=") || longCommandWArg(&argument, "strat=")) { g_params.strategy = (ZSTD_strategy)(readU32FromChar(&argument)); if (argument[0]==',') { argument++; continue; } else break; }
+                if (longCommandWArg(&argument, "level=") || longCommandWArg(&argument, "lvl=")) { g_params = ZSTD_getCParams(readU32FromChar(&argument), g_blockSize, 0); if (argument[0]==',') { argument++; continue; } else break; }
                 DISPLAY("invalid compression parameter \n");
                 return 1;
             }
 
             if (argument[0] != 0) {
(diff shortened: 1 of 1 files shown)


tests/paramgrill.c
```diff
@@ -972,7 +985,25 @@ int main(int argc, const char** argv)
         if(!strcmp(argument,"--no-seed")) { g_noSeed = 1; continue; }
 
         /* Decode command (note : aggregated commands are allowed) */
-        if (argument[0]=='-') {
+        if (longCommandWArg(&argument, "--zstd=")) {
+            g_singleRun = 1;
+            g_params = ZSTD_getCParams(2, g_blockSize, 0);
```
Cyan4973: This will initialize `g_params` using compression parameters from level 2.
Thus, if the user only provide a few arguments (for example only `wlog`), all other parameters will be initialized with values from level 2.

The problem is, that's too limited.

With the older scheme, it was possible to start from a different compression level, 
for example `SL9` meaning "use parameters from level 9", 
and then add a few modified parameters,
for example `SL9w18` meaning "start from level 9, and just modify `windowLog=18`".

We don't have this ability here, which is very limiting.

Suggestion : add a new sub-command `level=` to achieve same capability.

tests/paramgrill.c
```diff
@@ -972,7 +985,29 @@ int main(int argc, const char** argv)
         if(!strcmp(argument,"--no-seed")) { g_noSeed = 1; continue; }
 
         /* Decode command (note : aggregated commands are allowed) */
-        if (argument[0]=='-') {
+        if (longCommandWArg(&argument, "--zstd=")) {
+            g_singleRun = 1;
```
Cyan4973: This means that `--zstd=#` acts as a command, with same impact as `-S#`.

This is not entirely obvious.
In `zstd`, `--zstd=#` only selects parameters for compression, but it doesn't presume what is done with them afterwards. It could be for compression, it could be for benchmark.

So at the very least, this behavior should be documented.
Use `usage_advanced()` for that.

Cyan4973:
This patch only accepts advanced parameters using zstdcli's long command semantic.

Another goal is for the output of the optimizer (`-O#`) to produce results using the same long command semantic.
Do you want to make this second goal a separate PR ?

tests/paramgrill.c
```diff
@@ -1078,7 +1115,12 @@ int main(int argc, const char** argv)
     }
 
     if (filenamesStart==0) {
-        result = benchSample();
+        if (optimizer) {
+            DISPLAY("Optimizer Expects File")
```
Cyan4973: nit : missing `\n` and `;`

PR #1162, opened 2018-06-02
Fixed a corruption bug

Cyan4973:
recently introduced within the new dictionary mode (not part of any release).

The bug can be reproduced with this command :
`./zstreamtest -v --opaqueapi --no-big-tests -s4092 -t639`

Error was when invoking `ZSTD_count_2segments()` :
the beginning of the 2nd segment corresponds to `prefixStart`
as opposed to the beginning of the current block (`istart == src`).
Both pointers are identical during the first block,
so the issue only appears if `src` is large enough to require at least 2 blocks.
Then it results in comparing the wrong byte(s).

The bug was detected in `compressBlock_fast`
but is also present in `compressBlock_doubleFast`.
This patch fix both.

Changed a few variable names in the process,
in the hope to improve clarity.

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -74,18 +74,18 @@ size_t ZSTD_compressBlock_doubleFast_generic(
                                      dms->hashTable : NULL;
     const U32* const dictHashSmall = dictMode == ZSTD_dictMatchState ?
                                      dms->chainTable : NULL;
-    const U32 dictLowestIndex      = dictMode == ZSTD_dictMatchState ?
+    const U32 dictStartIndex       = dictMode == ZSTD_dictMatchState ?
                                      dms->window.dictLimit : 0;
     const BYTE* const dictBase     = dictMode == ZSTD_dictMatchState ?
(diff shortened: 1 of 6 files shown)


lib/compress/zstd_double_fast.c
```diff
@@ -126,7 +126,7 @@ size_t ZSTD_compressBlock_doubleFast_generic(
             && ((U32)((prefixLowestIndex-1) - repIndex) >= 3 /* intentional underflow */)
             && (MEM_read32(repMatch) == MEM_read32(ip+1)) ) {
             const BYTE* repMatchEnd = repIndex < prefixLowestIndex ? dictEnd : iend;
-            mLength = ZSTD_count_2segments(ip+1+4, repMatch+4, iend, repMatchEnd, istart) + 4;
+            mLength = ZSTD_count_2segments(ip+1+4, repMatch+4, iend, repMatchEnd, prefixLowest) + 4;
```
terrelln: nit: we should rename this to `prefixStart` as well.

felixhandte:
Aw man. Yeah, the fix is correct. *sigh*

PR #1164, opened 2018-06-04
Partial Compilation Macros

GeorgeLu97:
Makes it easier to compile subsets of features

Test Plan: ran
ZSTD_LIB_XXXX=0 make
with all of the macros, check only relevant files were compiled.


lib/Makefile
```diff
@@ -28,10 +28,41 @@ DEBUGFLAGS = -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
...
+ZSTDDECOMP_FILES := $(sort $(wildcard decompress/*.c))
+ZDICT_FILES := $(sort $(wildcard dictBuilder/*.c))
+ZDEPR_FILES := $(sort $(wildcard deprecated/*.c))
+ZSTD_FILES := $(ZSTDCOMMON_FILES)
 
 ZSTD_LEGACY_SUPPORT ?= 4
+ZSTD_LIB_COMPRESSION ?= 1
+ZSTD_LIB_DECOMPRESSION ?= 1
+ZSTD_LIB_DICTBUILDER ?= 1
+ZSTD_LIB_DEPRECATED ?= 1
```
Cyan4973: Note : `ZSTD_LIB_DEPRECATED` requires both `ZSTD_LIB_COMPRESSION` and `ZSTD_LIB_DECOMPRESSION` to be set to 1.

lib/Makefile
```diff
@@ -28,10 +28,43 @@ DEBUGFLAGS = -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
...
+
+ifneq ($(ZSTD_LIB_COMPRESSION), 0)
+	ZSTD_FILES += $(ZSTDCOMP_FILES) 
+endif
+
+ifneq ($(ZSTD_LIB_DECOMPRESSION), 0)
+	ZSTD_FILES += $(ZSTDDECOMP_FILES) 
+endif
+
+ifneq ($(ZSTD_DEPRECATED), 0)
```
Cyan4973: it should be `ZSTD_LIB_DEPRECATED`
instead of `ZSTD_DEPRECATED`.

Same mistake in 2 instances above.

lib/Makefile
```diff
@@ -28,10 +28,43 @@ DEBUGFLAGS = -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
...
+	ZSTD_FILE += $(ZDEPR_FILES)
+endif
+
+ifneq ($(ZSTD_LIB_DICTBUILDER), 0)
+	ZSTD_FILE += $(ZDICT_FILES)
```
Cyan4973: This should be `ZSTD_FILES`
instead of `ZSTD_FILE`.

Same typo above.

tests/partialTests.sh
```diff
@@ -0,0 +1,24 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+INTOVOID="/dev/null"
+case "$OS" in
+  Windows*)
+    INTOVOID="NUL"
+    ;;
+esac
+
```
Cyan4973: Looks good to me.
I guess you intend to add tests for the other macro directives ?

.travis.yml
```diff
@@ -31,6 +31,7 @@ matrix:
 
     - env: Cmd='make lz4install && make -C tests test-lz4'
 
+    - env: Cmd='bash tests/partialTests.sh'
```
Cyan4973: nit-picking here,
but `partialTests` is not a great name.
This could be easily confused.

Let's be a bit more specific.
Example : `conditional_libzstd_builds.sh`
Feel free to propose a better name if you like

Cyan4973:
Congratulation @GeorgeLu97, this patch will get you named in the release summary

PR #1165, opened 2018-06-05
Dynamic context downsize

Cyan4973:
In order to skip some allocation and initialization work,
it's recommended to preserve and re-use a compression context
across multiple compression jobs.
This can translate into substantial savings when dealing with a lot of small data.

The baseline scenario was that a process would fire up,
proceed with a batch of compression job, re-using a context in the process,
and then release the context at the end.

The context resizes up when it needs more than it has already allocated.
But it never resizes down.

In some cases though, some context might be preserved for a very long time (multiple days),
and have to process some various workload,
containing some rare random outliers (large files, high compression ratio).
This would result in a context being sized for this rare outlier,
even if the vast majority of the work consists of small data, requiring much less memory.

This patch dynamically sizes down the context in such a case.
To reduce fragmentation, the patch only sizes the context down when it's considered "wasteful",
defined as "too big" (i.e. > 3x larger necessary) for an extended period of time (> 128x).
Both constants can be tweaked later on.

cc @felixhandte , as @terrelln mentioned this patch can actually be useful for Managed Compression.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1054,10 +1054,10 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
     return ptr;
 }
 
-#define ZSTD_WORKSPACETOOLARGE_FACTOR 3 /* define "workspace is too large" as ZSTD_WORKSPACETOOLARGE_FACTOR times larger than needed */
-#define ZSTD_WORKSPACETOOLARGE_MAX 128  /* when workspace is continuously too large
-                                         * at least that number of times,
-                                         * context's memory usage is actually wasteful,
+#define ZSTD_WORKSPACETOOLARGE_FACTOR 3 /* define "workspace is too large" as this number of times larger than needed */
+#define ZSTD_WORKSPACETOOLARGE_MAXDURATION 128  /* when workspace is continuously too large
+                                         * during at least this number of times,
+                                         * context's memory usage is considered wasteful,
                                          * because it's sized to handle a worst case scenario which rarely happens.
                                          * In which case, resize it down to free some memory */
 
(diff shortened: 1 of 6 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1094,14 +1114,20 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
...
-
-            if (zc->workSpaceSize < neededSpace) {  /* too small : resize */
-                DEBUGLOG(4, "Need to update workSpaceSize from %uK to %uK",
-                            (unsigned)(zc->workSpaceSize>>10),
-                            (unsigned)(neededSpace>>10));
+
+            int const workSpaceTooSmall = zc->workSpaceSize < neededSpace;
+            int const workSpaceTooLarge = zc->workSpaceSize > ZSTD_WORKSPACETOOLARGE_FACTOR * neededSpace;
+            int const workSpaceWasteful = workSpaceTooLarge && (zc->workSpaceTooLarge > ZSTD_WORKSPACETOOLARGE_MAX);
+            zc->workSpaceTooLarge = workSpaceTooLarge ? zc->workSpaceTooLarge+1 : 0;
```
terrelln: nit: These variables have the same name but mean different things. Maybe call it `zc->timesWorkSpaceTooLarge`.

lib/compress/zstd_compress.c
```diff
@@ -1057,8 +1077,9 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
         if (ZSTD_equivalentParams(zc->appliedParams, params,
                                 zc->inBuffSize, zc->blockSize,
                                 zbuff, pledgedSrcSize)) {
-            DEBUGLOG(4, "ZSTD_equivalentParams()==1 -> continue mode (wLog1=%u, blockSize1=%u)",
-                        zc->appliedParams.cParams.windowLog, (U32)zc->blockSize);
+            DEBUGLOG(4, "ZSTD_equivalentParams()==1 -> continue mode (wLog1=%u, blockSize1=%zu)",
+                        zc->appliedParams.cParams.windowLog, zc->blockSize);
+            zc->workSpaceTooLarge += (zc->workSpaceTooLarge > 0);   /* if it was too large, it still is */
```
terrelln: Can we resize even when we are instructed to continue?

```c
if (zc->workSpaceTooLarge < ZSTD_WORKSPACETOOLARGE_MAX) {
  return ZSTD_continueCCtx(zc, params, pledgeSrcSize);
}
```

PR #1168, opened 2018-06-06
Have paramgrill share bench.c benchmarking function

GeorgeLu97:
Delete paramgrill's old benchmarking function to use bench.c's instead. Make bench.c's benchmarking return results instead of just printing.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -71,17 +71,16 @@ static U32 g_compressibilityDefault = 50;
 *  console display
 ***************************************/
 #define DISPLAY(...)         fprintf(stderr, __VA_ARGS__)
-#define DISPLAYLEVEL(l, ...) if (g_displayLevel>=l) { DISPLAY(__VA_ARGS__); }
-#define DISPLAYLEVELP(l, ...) if (g_displayLevel>=l && printable) { DISPLAY(__VA_ARGS__); }
-static int g_displayLevel = 2;   /* 0 : no display;   1: errors;   2 : + result + interaction + warnings;   3 : + progression;   4 : + information */
+#define DISPLAYLEVEL(l, ...) if (displayLevel>=l) { DISPLAY(__VA_ARGS__); }
+/* 0 : no display;   1: errors;   2 : + result + interaction + warnings;   3 : + progression;   4 : + information */
 
 static const U64 g_refreshRate = SEC_TO_MICRO / 6;
 static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 
-#define DISPLAYUPDATE(l, ...) { if (g_displayLevel>=l) { \
-            if ((UTIL_clockSpanMicro(g_displayClock) > g_refreshRate) || (g_displayLevel>=4)) \
+#define DISPLAYUPDATE(l, ...) { if (displayLevel>=l) { \
+            if ((UTIL_clockSpanMicro(g_displayClock) > g_refreshRate) || (displayLevel>=4)) \
             { g_displayClock = UTIL_getTime(); DISPLAY(__VA_ARGS__); \
-            if (g_displayLevel>=4) fflush(stderr); } } }
+            if (displayLevel>=4) fflush(stderr); } } }
 
 
 /* *************************************
@@ -93,9 +92,9 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 #define DEBUGOUTPUT(...) { if (DEBUG) DISPLAY(__VA_ARGS__); }
 #define EXM_THROW(error, ...)  {                      \
     DEBUGOUTPUT("%s: %i: \n", __FILE__, __LINE__);    \
-    DISPLAYLEVELP(1, "Error %i : ", error);            \
-    DISPLAYLEVELP(1, __VA_ARGS__);                     \
-    DISPLAYLEVELP(1, " \n");                           \
+    DISPLAYLEVEL(1, "Error %i : ", error);            \
+    DISPLAYLEVEL(1, __VA_ARGS__);                     \
+    DISPLAYLEVEL(1, " \n");                           \
     exit(error);                                      \
 }
 
@@ -106,30 +105,34 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 static int g_additionalParam = 0;
 static U32 g_decodeOnly = 0;
 
-void BMK_setNotificationLevel(unsigned level) { g_displayLevel=level; }
-
 void BMK_setAdditionalParam(int additionalParam) { g_additionalParam=additionalParam; }
 
+
+//TODO : Deal with DISPLAYLEVEL for all these set functions
+
 static U32 g_nbSeconds = BMK_TIMETEST_DEFAULT_S;
+
 void BMK_setNbSeconds(unsigned nbSeconds)
 {
     g_nbSeconds = nbSeconds;
-    DISPLAYLEVEL(3, "- test >= %u seconds per compression / decompression - \n", g_nbSeconds);
+    DISPLAY("- test >= %u seconds per compression / decompression - \n", g_nbSeconds);
 }
 
 static size_t g_blockSize = 0;
+
 void BMK_setBlockSize(size_t blockSize)
 {
     g_blockSize = blockSize;
-    if (g_blockSize) DISPLAYLEVEL(2, "using blocks of size %u KB \n", (U32)(blockSize>>10));
+    if (g_blockSize) DISPLAY("using blocks of size %u KB \n", (U32)(blockSize>>10));
 }
 
 void BMK_setDecodeOnlyMode(unsigned decodeFlag) { g_decodeOnly = (decodeFlag>0); }
 
 static U32 g_nbWorkers = 0;
+
 void BMK_setNbWorkers(unsigned nbWorkers) {
 #ifndef ZSTD_MULTITHREAD
-    if (nbWorkers > 0) DISPLAYLEVEL(2, "Note : multi-threading is disabled \n");
+    if (nbWorkers > 0) DISPLAY("Note : multi-threading is disabled \n");
 #endif
     g_nbWorkers = nbWorkers;
 }
@@ -188,12 +191,13 @@ typedef struct {
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 #define MAX(a,b)    ((a) > (b) ? (a) : (b))
 
-BMK_result_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
-                        const char* displayName, int cLevel,
-                        const size_t* fileSizes, U32 nbFiles,
+BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
+                        const size_t* fileSizes, unsigned nbFiles,
+                        const int cLevel, const ZSTD_compressionParameters* comprParams,
                         const void* dictBuffer, size_t dictBufferSize,
-                        const ZSTD_compressionParameters* const comprParams, 
-                        ZSTD_CCtx* ctx, ZSTD_DCtx* dctx, int printable)
+                        ZSTD_CCtx* ctx, ZSTD_DCtx* dctx,
+                        int displayLevel, const char* displayName)
+
 {
     size_t const blockSize = ((g_blockSize>=32 && !g_decodeOnly) ? g_blockSize : srcSize) + (!srcSize) /* avoid div by 0 */ ;
     U32 const maxNbBlocks = (U32) ((srcSize + (blockSize-1)) / blockSize) + nbFiles;
@@ -201,19 +205,19 @@ BMK_result_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
     size_t const maxCompressedSize = ZSTD_compressBound(srcSize) + (maxNbBlocks * 1024);   /* add some room for safety */
     void* const compressedBuffer = malloc(maxCompressedSize);
     void* resultBuffer = malloc(srcSize);
-    BMK_result_t results;
+    BMK_return_t results;
 
     size_t const loadedCompressedSize = srcSize;
     size_t cSize = 0;
     double ratio = 0.;
     U32 nbBlocks;
 
-    assert(ctx != NULL);
-    assert(dctx != NULL);
-
     /* checks */
-    if (!compressedBuffer || !resultBuffer || !blockTable || !ctx || !dctx)
+    if (!compressedBuffer || !resultBuffer || !blockTable)
         EXM_THROW(31, "allocation error : not enough memory");
+
+    if(!ctx || !dctx) 
+        EXM_THROW(31, "error: passed in null context");
 
     /* init */
     if (strlen(displayName)>17) displayName += strlen(displayName)-17;   /* display last 17 characters */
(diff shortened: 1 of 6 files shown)


Cyan4973:
Consider providing a short description of your work in the PR.

While I do know what you are doing and why,
an external contributor / reviewer will not.
Consider writing this short description for such an audience.

programs/bench.c
```diff
@@ -535,11 +551,11 @@ static void BMK_benchCLevel(const void* srcBuffer, size_t benchedSize,
         DISPLAY("bench %s %s: input %u bytes, %u seconds, %u KB blocks\n", ZSTD_VERSION_STRING, ZSTD_GIT_COMMIT_STRING, (U32)benchedSize, g_nbSeconds, (U32)(g_blockSize>>10));
 
     for (l=cLevel; l <= cLevelLast; l++) {
-        if (l==0) continue;  /* skip level 0 */
+        //if (l==0) continue;  /* skip level 0 */
```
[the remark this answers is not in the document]
GeorgeLu97: I thought I might need it, but I don't think it ended up mattering. I'll uncomment it

tests/paramgrill.c
```diff
@@ -63,11 +62,13 @@ static const int g_maxNbVariations = 64;
 #undef MAX
 #define MIN(a,b)   ( (a) < (b) ? (a) : (b) )
 #define MAX(a,b)   ( (a) > (b) ? (a) : (b) )
-
+#define CUSTOM_LEVEL 99
```
Cyan4973: 👍

programs/bench.c
```diff
@@ -189,25 +193,30 @@ typedef struct {
...
     void* resultBuffer = malloc(srcSize);
-    ZSTD_CCtx* const ctx = ZSTD_createCCtx();
-    ZSTD_DCtx* const dctx = ZSTD_createDCtx();
+    
     size_t const loadedCompressedSize = srcSize;
     size_t cSize = 0;
     double ratio = 0.;
     U32 nbBlocks;
 
+    int ctxFree = 0, dctxFree = 0;
```
Cyan4973: Consider 2 nested functions.

- The lower one will accept `cctx` and `dctx` as arguments, 
   and used them as is , i.e. not modify them (but still verify they are not `NULL`).
- The upper one will have the responsibility to create / free `cctx` and `dctx`,
   and simply invoke the lower one.

Since only the lower function is called by `paramgrill`, it's the only one to be published in `bench.h`.

programs/bench.c
```diff
@@ -369,10 +378,12 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
                 cSize = 0;
                 { U32 blockNb; for (blockNb=0; blockNb<nbBlocks; blockNb++) cSize += blockTable[blockNb].cSize; }
                 ratio = (double)srcSize / (double)cSize;
+                if(results != NULL) { results->cSize = cSize; }
```
Cyan4973: Consider making `results` your return type.

programs/bench.c
```diff
@@ -625,10 +642,12 @@ static void BMK_benchFileTable(const char* const * const fileNamesTable, unsigne
         U32 fileNb;
         for (fileNb=0; fileNb<nbFiles; fileNb++) {
             size_t const fileSize = fileSizes[fileNb];
+            BMK_result_t* resulti = (results == NULL) ? NULL : &(results[fileNb]);
```
Cyan4973: `BMK_result_t* const resulti`

We are basically obsessed by the `const` property in this project.
Try to `const` everything you can.

It's an important hint for the reviewer that the created value will not change.
When applied consistently, its absence means that this value _will_ change,
and therefore, the reviewer must concentrate his attention on it.

programs/bench.h
```diff
@@ -13,12 +13,32 @@
 #define BENCH_H_121279284357
 
 #include <stddef.h>   /* size_t */
+#include "mem.h"
```
Cyan4973: The least amount of `#include` is always the better.
It's even more important inside a `.h` header file, because the compilation overhead is multiplied by the nb of users of the `.h`, which can quickly ramp up after a few nested levels.

In this case, I guess you need `mem.h` for the `U32` type within `BMK_benchMem()`.

Consider replacing this type by `unsigned` instead.

tests/paramgrill.c
```diff
@@ -162,176 +169,39 @@ typedef struct
...
-    /* warmimg up memory */
-    RDG_genBuffer(compressedBuffer, maxCompressedSize, 0.10, 0.10, 1);
-
-    /* Bench */
-    {   U32 loopNb;
-        size_t cSize = 0;
-        double fastestC = 100000000., fastestD = 100000000.;
-        double ratio = 0.;
-        UTIL_time_t const benchStart = UTIL_getTime();
+    //fileSizes[0] == srcSize?
```
Cyan4973: ??

programs/bench.h
```diff
@@ -33,5 +53,6 @@ void BMK_setLdmMinMatch(unsigned ldmMinMatch);
 void BMK_setLdmHashLog(unsigned ldmHashLog);
 void BMK_setLdmBucketSizeLog(unsigned ldmBucketSizeLog);
 void BMK_setLdmHashEveryLog(unsigned ldmHashEveryLog);
+void BMK_setPrint(int printable);
```
Cyan4973: Could you make `printable` an argument of `BMK_benchMem()` instead ?

Cyan4973:
Try to quash your commits when you have a lot of "minor fix + minor fix + minor fix" succession.
Ideally you want each commit to have a title and do something meaningful, that you can describe.
Sometimes, minor patches are just necessary.
But let's just not multiply them.
This has a cost when trying to follow contribution history, or during bisection.

Cyan4973:
For some reason, your PR seems to aggregate all recent changes.

While it does not impact the final merge result,
it makes it a bit more difficult to isolate the changes from this patch only.

I'm surprised, merging `dev` into your working branch was supposed to clean these artifacts.

programs/bench.c
```diff
@@ -514,20 +570,28 @@ static size_t BMK_findMaxMem(U64 requiredMem)
     return (size_t)(requiredMem);
 }
 
-static void BMK_benchCLevel(const void* srcBuffer, size_t benchedSize,
+/* returns average stats over all range [cLevel, cLevelLast] */
```
Cyan4973: nope

programs/bench.c
```diff
@@ -71,6 +72,7 @@ static U32 g_compressibilityDefault = 50;
 ***************************************/
 #define DISPLAY(...)         fprintf(stderr, __VA_ARGS__)
 #define DISPLAYLEVEL(l, ...) if (g_displayLevel>=l) { DISPLAY(__VA_ARGS__); }
+#define DISPLAYLEVELP(l, ...) if (g_displayLevel>=l && printable) { DISPLAY(__VA_ARGS__); }
```
Cyan4973: That feels strange.
This macro now depends on both `g_displayLevel` and some local variable `printable` to exist within scope.

I've no doubt you can make it to work in this patch, but this is a recipe for problems in future patches.

Consider having `displayLevel` a direct argument of the function.
This way, setting to `0` is enough to disable any display.

This may enter in competition with currently defined `g_displayLevel`.
One cheap way to get it to work is to start your function with `g_displayLevel = displayLevel`.
Everything else stay the same.

A more involved but proper way is to remove `g_displayLevel` altogether.
Now that `bench.c` is no longer a stand-alone module, it becomes more desirable.
You would then have to take care of proper propagation of `displayLevel` argument.

programs/bench.c
```diff
@@ -536,33 +600,35 @@ static void BMK_benchCLevel(const void* srcBuffer, size_t benchedSize,
...
         if (l==0) continue;  /* skip level 0 */
-        BMK_benchMem(srcBuffer, benchedSize,
-                     displayName, l,
-                     fileSizes, nbFiles,
-                     dictBuffer, dictBufferSize, compressionParams);
+        resultarray[l - cLevel] = BMK_benchMemCtxless(srcBuffer, benchedSize,
+                    displayName, l, fileSizes, nbFiles,
+                    dictBuffer, dictBufferSize, compressionParams, printable);
     }
+    returnval = BMK_average(resultarray, cLevelLast - cLevel + 1, NULL);
```
Cyan4973: Averaging the performance of multiple levels to generate a `BMK_result_t` return type is not good.
Such measure makes no sense.
If the only reason to do that is to retrieve a `BMK_result_t` for `paramgrill`, then consider skipping this stage.

There is still a need for a function that loops through multiple compression levels, but that's not for `paramgrill`.

programs/bench.c
```diff
@@ -617,43 +683,50 @@ static void BMK_benchFileTable(const char* const * const fileNamesTable, unsigne
...
-            BMK_benchCLevel(srcPtr, fileSize,
+            resultarray[fileNb] = BMK_benchCLevel(srcPtr, fileSize,
                             fileNamesTable[fileNb], cLevel, cLevelLast,
-                            fileSizes+fileNb, 1,
-                            dictBuffer, dictBufferSize, compressionParams);
+                            fileSizes+fileNb, 1, dictBuffer, dictBufferSize, 
+                            compressionParams,printable);
             srcPtr += fileSize;
         }
+        returnval =  BMK_average(resultarray, nbFiles, fileSizes);
```
Cyan4973: It's a bit more useful to generate an average result over multiple files (rather than multiple compression levels)
but there are 2 problems : 
- The formula in `BMK_average()` is wrong. Speeds don't add up, time / duration would be more sensible.
- This functionality is already integrated in `benchMem()`.
  It's actually the default behavior.
  This function is actually a hack around that, to prevent `benchMem()` from generating an aggregate result,
   and have one clean result per file instead, in order to precisely observe the differences.

So, here also, I believe it's not a correct entry point for `paramgrill`.
Skip that stage, create an intermediate one, that fits `paramgrill` needs, and also fits nicely with the other benchmark functions, which are created for other needs.

Cyan4973:
Minor thing : 
now that `bench.c` becomes a module that can be invoked in multiple programs,
consider decorating the declaration in `bench.h` with some `extern "C"` statement.
This is useful to invoke the functions from C++.
Have a look into any `.h` function within `/lib` for details.

programs/bench.h
```diff
@@ -16,9 +16,28 @@
...
+    double cSpeed;   /* bytes / sec */
+    double dSpeed;
+} BMK_result_t;
+
+/* called in cli */
 int BMK_benchFiles(const char** fileNamesTable, unsigned nbFiles, const char* dictFileName,
                    int cLevel, int cLevelLast, const ZSTD_compressionParameters* compressionParams);
 
+/* more options */
+BMK_result_t BMK_benchFilesFull(const char** fileNamesTable, unsigned nbFiles, const char* dictFileName,
```
Cyan4973: So apparently you already skip intermediate functions when invoking from `paramgrill`,
by calling directly `BMK_benchMem()`.

So I guess the issue is limited to `BMK_benchFilesFull()`.
It simply makes no sense to provide a single "unified" `BMK_result_t ` result to measure multiple compression levels.
I guess it's more appropriate for this function to have a `void` or `int` type.

Also, if there is no user of this function, it might be better to just remove it entirely.
It's a complex function, with a complex behavior, so if it does not serve an immediate identified need, it's just a maintenance tax.

programs/bench.c
```diff
@@ -182,32 +183,34 @@ typedef struct {
...
 
-
-
 #undef MIN
 #undef MAX
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 #define MAX(a,b)    ((a) > (b) ? (a) : (b))
 
-static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
+BMK_result_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
```
Cyan4973: minor : 
I believe it's worthwhile to re-order arguments in a way which makes a bit more sense.
It's becoming a complex function, with an uncomfortably large nb of arguments, 
and more importantly, it's becoming a public (exposed) function, 
so it matters.
Here is my suggestion : 
````
const void* srcBuffer, size_t srcSize,        // like now
const size_t* fileSizes, unsigned nbFiles,    // describes the content of `srcBuffer`, so related. Also : prefer `unsigned` over `U32`, since the prototype is declared in `bench.h`
int cLevel, const ZSTD_compressionParameters* const comprParams,// they are related, keep them together
const void* dictBuffer, size_t dictBufferSize,// complementary compression argument
ZSTD_CCtx* ctx, ZSTD_DCtx* dctx,              // workspace contexts
int displayLevel, const char* displayName     // both are about display, they are related
````

File: programs/bench.c [hunk 15]
```diff
@@ -490,12 +501,27 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
...
+}
+
+static void BMK_benchMemCtxless(const void* srcBuffer, size_t srcSize,
+                        const size_t* fileSizes, unsigned nbFiles, 
+                        int cLevel, const ZSTD_compressionParameters* const comprParams,
+                        const void* dictBuffer, size_t dictBufferSize,
+                         int displayLevel, const char* displayName) 
+{
+    ZSTD_CCtx* ctx = ZSTD_createCCtx();
+    ZSTD_DCtx* dctx = ZSTD_createDCtx();
```
Cyan4973: Important : check that `ctx` and `dctx` are both non NULL !
Memory allocation can always fail !

Note : in this case, it's an active check, with an associated "error" behavior defined.
Not to be confused with an `assert()`, which describes a situation which should never happen.

programs/bench.h
```diff
@@ -16,8 +19,25 @@
...
+} BMK_result_t;
+
+/* called in cli */
 int BMK_benchFiles(const char** fileNamesTable, unsigned nbFiles, const char* dictFileName,
-                   int cLevel, int cLevelLast, const ZSTD_compressionParameters* compressionParams);
+                   int cLevel, int cLevelLast, const ZSTD_compressionParameters* compressionParams, 
+                   int displayLevel);
+
+/* basic benchmarking function, called in paramgrill
+ * results, cctx, dctx, dictbuffer can be null or passed in */
```
Cyan4973: The comment seems incorrect : `cctx` and `dctx` must be valid

programs/bench.h
```diff
@@ -16,8 +19,25 @@
 #define ZSTD_STATIC_LINKING_ONLY   /* ZSTD_compressionParameters */
 #include "zstd.h"     /* ZSTD_compressionParameters */
 
+typedef struct {
+    size_t cSize;
+    double cSpeed;   /* bytes / sec */
+    double dSpeed;
+} BMK_result_t;
```
Cyan4973: What happens if there is an error ?
Or : is there a guarantee that `BMK_benchMem()` will never fail ?

programs/bench.c
```diff
@@ -182,32 +186,35 @@ typedef struct {
...
     size_t cSize = 0;
     double ratio = 0.;
     U32 nbBlocks;
 
+    assert(ctx != NULL);
+    assert(dctx != NULL);
+
     /* checks */
     if (!compressedBuffer || !resultBuffer || !blockTable || !ctx || !dctx)
         EXM_THROW(31, "allocation error : not enough memory");
```
Cyan4973: `EXM_THROW()` was fine for a self-contained program,
but for a library function which is called by multiple other modules,
it's less desirable.

In theory, that would be the right time to consider implementing a cleaner error reporting system.
But in the interest of getting this PR completed with a reasonable amount of complexity,
we'll keep it this way for now.

programs/bench.c
```diff
@@ -182,32 +186,35 @@ typedef struct {
...
     size_t const loadedCompressedSize = srcSize;
     size_t cSize = 0;
     double ratio = 0.;
     U32 nbBlocks;
 
+    assert(ctx != NULL);
+    assert(dctx != NULL);
+
     /* checks */
     if (!compressedBuffer || !resultBuffer || !blockTable || !ctx || !dctx)
```
Cyan4973: `ctx` and `dctx` were already checked by `assert()` above.

PR #1174, opened 2018-06-08
Expose ZSTD_CLEVEL_DEFAULT and update documentation

### LINKED ISSUE #1148: Default compression level undocumented/unexposed
Is it possible to either (1) document in the public API that compression level zero will always be sanitized to ZSTD_CLEVEL_DEFAULT, or (2) expose ZSTD_CLEVEL_DEFAULT in the public API? Please correct me if I'm wrong, but I couldn't find either of these, in which case it doesn't seem there's a way to use ZSTD default compression level that's guaranteed not to change in the future.

duc0:
Fixes #1148 

Reviewer: @Cyan4973

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -8,15 +8,6 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
-
-/*-*************************************
-*  Tuning parameters
-***************************************/
-#ifndef ZSTD_CLEVEL_DEFAULT
-#  define ZSTD_CLEVEL_DEFAULT 3
-#endif
-
-
 /*-*************************************
 *  Dependencies
 ***************************************/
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks @duc0 .
That's exactly what needs to be done.

Now onto small presentation details : 
- Position within `zstd.h` : I guess it's better to have this constant defined just after the `--   Version   --` constants.
- Title : ` Tuning parameters` is no longer a correct title. `Default constant` is probably more adapted.

duc0:
@Cyan4973 Thanks for feedback, I made the suggested changes

Cyan4973:
Thanks @duc0 !

duc0:
I made multiple stacked commits here. I guess it didn't get squashed into one commit when merging to the dev branch. Is that fine or is there a way to fix that?

Cyan4973:
Yes, you are right, squashing would have been better.
Since the nb of commits is not large here, it's not a big deal, just preferable for history review and bisection. Think about it next time, it'll be all fine.

PR #1177, opened 2018-06-10
Remove duplicate CXXFLAGS

ryandesign:


Cyan4973:
Good point, thanks @ryandesign !

PR #1178, opened 2018-06-11
Add CXXFLAGS to ALL_LDFLAGS

ryandesign:
pzstd requires C++11, which older C++ standard libraries like libstdc++
as used on OS X 10.8 and earlier don't support. The user might address
this by setting "CXXFLAGS=-stdlib=libc++". This flag must be used both
at compile time and at link time. Asking the user to also put the flag
in LDFLAGS is undesirable because then the flag would also be used when
linking C code, which would be inappropriate.


Cyan4973:
cc @terrelln

PR #1179, opened 2018-06-11
Improves UX for --list command's lack of support for pipes

supertopher:
`--list` does not support piped input
This commit checks if the STDIN is not from a terminal and exits 1 
with a well formatted error message 

the output of this command may still be piped to another program

**examples of the code change below**
```
zstd [dev] :> ./zstd --list # exits 0

zstd [dev] :> echo "awdaw" | ./zstd --list
zstd --list does not support reading from standard input # exits 1

zstd [dev] :> ./zstd --list | echo # exits 0
```
 I don't believe this type of code is usually tested.  Please inform me if i am incorrect and I will add tests.

closes task `T29991838` @Cyan4973

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -2031,9 +2031,10 @@ static int FIO_listFile(fileInfo_t* total, const char* inFileName, int displayLe
 }
 
 int FIO_listMultipleFiles(unsigned numFiles, const char** filenameTable, int displayLevel){
-    if (!isatty(0))
-        DISPLAYOUT("zstd --list does not support reading from standard input\n");
+    if (!isatty(0)) {
+        DISPLAYOUT("zstd: --list does not support reading from standard input\n");
         return 1;
+    }
 
     if (numFiles == 0) {
         DISPLAYOUT("No files given\n");
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -61,6 +61,7 @@
 #if defined(MSDOS) || defined(OS2) || defined(WIN32) || defined(_WIN32) ||     \
     defined(__CYGWIN__)
 #include <io.h> /* _isatty */
+#include <sys/stat.h>
 #define IS_CONSOLE(stdStream) _isatty(_fileno(stdStream))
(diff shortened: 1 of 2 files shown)


programs/fileio.c
```diff
@@ -62,6 +57,20 @@
 #  include <lz4.h>
 #endif
 
+// Multi-OS support for determining if terminal is tty
```
Cyan4973: The code itself is fine.
Just, it competes with an existing and included code, which is here : 
https://github.com/facebook/zstd/blob/dev/programs/platform.h#L97

To be fair, I'm surprised the compiler doesn't complain about duplicate macro symbols (?).

Anyway, if your version is better, even in a limited way,
I believe a better outcome would be to improve the implementation in `platform.h`, so that the improvement can be shared with any other module using `IS_CONSOLE()` (such as `zstdcli.c`).

programs/fileio.c
```diff
@@ -62,6 +57,10 @@
 #  include <lz4.h>
 #endif
 
+#if defined (_MSC_VER)
```
Cyan4973: Any reason you prefer to move Windows - specific `#include` down the list ?

Cyan4973:
Good !
The change looks fine to me !

The next stage is to add a test, verifying your change works as intended.
More importantly, this test will be part of CI, and therefore ensure that it will work in the future, thus catching any potential mistake done by any future contributor.

The test should be added to `playTests.sh`, in the `--list` category : 
https://github.com/facebook/zstd/blob/dev/tests/playTests.sh#L711

A simple test will be enough.
Just ensure that it fails (return code > 0) when provided with `stdin` input.

programs/fileio.c
```diff
@@ -2030,6 +2029,12 @@ static int FIO_listFile(fileInfo_t* total, const char* inFileName, int displayLe
 }
 
 int FIO_listMultipleFiles(unsigned numFiles, const char** filenameTable, int displayLevel){
+
+    if (!IS_CONSOLE(stdin)) {
+        DISPLAYOUT("zstd: --list does not support reading from standard input\n");
+        return 1;
```
Cyan4973: 👍

Cyan4973:
Thanks @supertopher ! Looks good to me !

PR #1169, opened 2018-06-07
ZSTD_lazy: Support Searching the Dictionary Context In-Place

felixhandte:
Extending the work of #1151 and #1117, this change set allows the `ZSTD_greedy`, `ZSTD_lazy`, and `ZSTD_lazy2` strategies to reference a separate dictionary context in-place. Performance figures are being prepared.

PR #1185, opened 2018-06-13
ZSTD_btlazy2: Support Searching the Dictionary Context In-Place

felixhandte:
Continuing the work of #1169 (and #1151 and #1117), this PR allows the `ZSTD_btlazy2` to perform searches against an external, immutable dictionary context.

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -140,6 +140,84 @@ static void ZSTD_insertDUBT1(
 }
 
 
+static size_t ZSTD_DUBT_findBetterDictMatch (
+        ZSTD_matchState_t* ms, ZSTD_compressionParameters const* cParams,
+        const BYTE* const ip, const BYTE* const iend,
+        size_t* offsetPtr,
+        size_t bestLength,
+        U32 nbCompares,
+        U32 const mls,
+        const ZSTD_dictMode_e dictMode) {
+    const ZSTD_matchState_t * const dms = ms->dictMatchState;
+    const U32 * const dictHashTable = dms->hashTable;
+    U32         const hashLog = cParams->hashLog;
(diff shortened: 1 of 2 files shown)


Cyan4973:
This patch looks good to me.
It's clean, adding only a well localized complexity.

It also reminds me that, at some point, we'll probably need to make a more thorough evaluation of what are the "cut off" points for copy/no-copy strategy selection.

lib/compress/zstd_lazy.c
```diff
@@ -261,6 +339,10 @@ static size_t ZSTD_DUBT_findBestMatch (
 
         *smallerPtr = *largerPtr = 0;
 
+        if (dictMode == ZSTD_dictMatchState && nbCompares) {
```
terrelln: Should we give the dictionary a fixed lower limit on the number of compares, like at least 1 compare, or at least 8 compares? Or have benchmarks to show that it doesn't matter?

felixhandte: The short answer is, I haven't experimented with those kinds of changes. The scope of these changesets has been to produce identical behavior faster.

So yeah, there are probably opportunities to experiment with different strategies given the additional information available. I'd love to chat about how to go about doing that.

Cyan4973: I agree this seems like an additional topic.

This new strategy is likely to alter the speed/compression trade-off, spending more cpu for more compression opportunity, in a way which is probably content specific.
It would have to be tested against a corpus to observe the impact.

The nice thing though is that it seems relatively easy to test.
Just add a `nbCompares = MAX(nbCompares, CONSTANT);` before the dictionary section, and that's it from a code perspective. A more important question is the selection of a benchmark corpus, though Managed Compression might help on this front.

PR #1188, opened 2018-06-14
Bench module

GeorgeLu97:
Makes bench.c more useable by other files.

Test plan:
Ran fullbench, zstd -b#, and paramgrill before/after changes.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -110,13 +110,25 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
     return r;                                         \
 }
 
+/* error without displaying */
+#define EXM_THROW_ND(errorNum, retType, ...)  {          \
+    retType r;                                        \
+    memset(&r, 0, sizeof(retType));                   \
+    DEBUGOUTPUT("%s: %i: \n", __FILE__, __LINE__);    \
+    DEBUGOUTPUT("Error %i : ", errorNum);         \
+    DEBUGOUTPUT(__VA_ARGS__);                     \
+    DEBUGOUTPUT(" \n");                           \
+    r.error = errorNum;                               \
+    return r;                                         \
+}
+
 /* *************************************
 *  Benchmark Parameters
 ***************************************/
 
 #define BMK_LDM_PARAM_NOTSET 9999
 
-BMK_advancedParams_t BMK_defaultAdvancedParams(void) { 
+BMK_advancedParams_t BMK_initAdvancedParams(void) { 
     BMK_advancedParams_t res = { 
         BMK_both, /* mode */
         BMK_timeMode, /* loopMode */
@@ -124,7 +136,6 @@ BMK_advancedParams_t BMK_defaultAdvancedParams(void) {
         0, /* blockSize */
         0, /* nbWorkers */
         0, /* realTime */
-        1, /* separateFiles */
         0, /* additionalParam */
         0, /* ldmFlag */ 
         0, /* ldmMinMatch */
@@ -239,7 +250,7 @@ static size_t local_defaultCompress(
     return out.pos;
 }
 
-/* addiional argument is just the context */
+/* additional argument is just the context */
 static size_t local_defaultDecompress(
     const void* srcBuffer, size_t srcSize, 
     void* dstBuffer, size_t dstSize, 
@@ -269,14 +280,13 @@ static size_t local_defaultDecompress(
 /* initFn will be measured once, bench fn will be measured x times */
 /* benchFn should return error value or out Size */
 /* takes # of blocks and list of size & stuff for each. */
-BMK_customReturn_t BMK_benchCustom(
-    const char* functionName, size_t blockCount,
-    const void* const * const srcBuffers, const size_t* srcSizes,
-    void* const * const dstBuffers, const size_t* dstSizes,
-    size_t (*initFn)(void*), size_t (*benchFn)(const void*, size_t, void*, size_t, void*), 
-    void* initPayload, void* benchPayload,
-    unsigned mode, unsigned iter,
-    int displayLevel) {
+BMK_customReturn_t BMK_benchFunction(
+    size_t blockCount,
+    const void* const * const srcBlockBuffers, const size_t* srcBlockSizes,
+    void* const * const dstBlockBuffers, const size_t* dstBlockCapacities,
+    size_t (*initFn)(void*), void* initPayload,
+    size_t (*benchFn)(const void*, size_t, void*, size_t, void*), void* benchPayload,
+    unsigned mode, unsigned iter) {
     size_t srcSize = 0, dstSize = 0, ind = 0;
     unsigned toAdd = 1;
 
@@ -287,24 +297,22 @@ BMK_customReturn_t BMK_benchCustom(
     {
         unsigned i;
         for(i = 0; i < blockCount; i++) {
-            memset(dstBuffers[i], 0xE5, dstSizes[i]);  /* warm up and erase result buffer */
+            memset(dstBlockBuffers[i], 0xE5, dstBlockCapacities[i]);  /* warm up and erase result buffer */
         }
 
         UTIL_sleepMilli(5);  /* give processor time to other processes */
         UTIL_waitForNextTick();
     }
 
-    /* display last 17 char's of functionName*/
-    if (strlen(functionName)>17) functionName += strlen(functionName)-17;
     if(!iter) {
         if(mode == BMK_iterMode) {
-            EXM_THROW(1, BMK_customReturn_t, "nbLoops must be nonzero \n");
+            EXM_THROW_ND(1, BMK_customReturn_t, "nbLoops must be nonzero \n");
         }
 
     }
 
     for(ind = 0; ind < blockCount; ind++) {
-        srcSize += srcSizes[ind];
+        srcSize += srcBlockSizes[ind];
     }
 
     switch(mode) {
@@ -318,13 +326,13 @@ BMK_customReturn_t BMK_benchCustom(
                 unsigned i, j;
                 /* Overheat protection */
                 if (UTIL_clockSpanMicro(coolTime) > ACTIVEPERIOD_MICROSEC) {
-                    DISPLAYLEVEL(2, "\rcooling down ...    \r");
+                    DEBUGOUTPUT("\rcooling down ...    \r");
                     UTIL_sleep(COOLPERIOD_SEC);
                     coolTime = UTIL_getTime();
                 }
                 
                 for(i = 0; i < blockCount; i++) {
-                    memset(dstBuffers[i], 0xD6, dstSizes[i]);  /* warm up and erase result buffer */
+                    memset(dstBlockBuffers[i], 0xD6, dstBlockCapacities[i]);  /* warm up and erase result buffer */
                 }
 
                 clockStart = UTIL_getTime();
@@ -332,10 +340,10 @@ BMK_customReturn_t BMK_benchCustom(
 
                 for(i = 0; i < nbLoops; i++) {
                     for(j = 0; j < blockCount; j++) {
-                        size_t res = (*benchFn)(srcBuffers[j], srcSizes[j], dstBuffers[j], dstSizes[j], benchPayload);
+                        size_t res = (*benchFn)(srcBlockBuffers[j], srcBlockSizes[j], dstBlockBuffers[j], dstBlockCapacities[j], benchPayload);
                         if(ZSTD_isError(res)) {
-                            EXM_THROW(2, BMK_customReturn_t, "%s() failed on block %u of size %u : %s  \n",
-                                functionName, j, (U32)dstSizes[j], ZSTD_getErrorName(res));
+                            EXM_THROW_ND(2, BMK_customReturn_t, "Function benchmarking failed on block %u of size %u : %s  \n",
+                                j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
                         } else if (toAdd) {
                             dstSize += res;
                         }
@@ -362,10 +370,10 @@ BMK_customReturn_t BMK_benchCustom(
             clockStart = UTIL_getTime();
             for(i = 0; i < iter; i++) {
                 for(j = 0; j < blockCount; j++) {
-                    size_t res = (*benchFn)(srcBuffers[j], srcSizes[j], dstBuffers[j], dstSizes[j], benchPayload);
+                    size_t res = (*benchFn)(srcBlockBuffers[j], srcBlockSizes[j], dstBlockBuffers[j], dstBlockCapacities[j], benchPayload);
                     if(ZSTD_isError(res)) {
-                        EXM_THROW(2, BMK_customReturn_t, "%s() failed on block %u of size %u : %s  \n",
-                            functionName, j, (U32)dstSizes[j], ZSTD_getErrorName(res));
+                        EXM_THROW_ND(2, BMK_customReturn_t, "Function benchmarking failed on block %u of size %u : %s  \n",
+                            j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
                     } else if(toAdd) {
                         dstSize += res;
                     }
@@ -374,18 +382,18 @@ BMK_customReturn_t BMK_benchCustom(
             }
             totalTime = UTIL_clockSpanNano(clockStart);
             if(!totalTime) {
-                EXM_THROW(3, BMK_customReturn_t, "Cycle count (%u) too short to measure \n", iter);
+                EXM_THROW_ND(3, BMK_customReturn_t, "Cycle count (%u) too short to measure \n", iter);
             } else {
                 fastest = totalTime / iter;
             }
             break;
         }
         default:
-            EXM_THROW(4, BMK_customReturn_t, "Unknown Mode \n");
+            EXM_THROW_ND(4, BMK_customReturn_t, "Unknown Mode \n");
     }
     retval.error = 0;
-    retval.result.time = fastest;
-    retval.result.size = dstSize;
+    retval.result.nanoSecPerRun = fastest;
+    retval.result.sumOfReturn = dstSize;
     return retval;
 }
 
@@ -503,23 +511,23 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
             cctxprep.adv = adv;
             /* Compression */
             DISPLAYLEVEL(2, "%2s-%-17.17s :%10u ->\r", marks[markNb], displayName, (U32)srcSize);
-            compressionResults = BMK_benchCustom("ZSTD_compress_generic", nbBlocks, 
+            compressionResults = BMK_benchFunction(nbBlocks, 
                 srcPtrs, srcSizes, cPtrs, cSizes, 
-                &local_initCCtx, &local_defaultCompress, 
-                (void*)&cctxprep, (void*)(ctx),
-                adv->loopMode, adv->nbSeconds, displayLevel);
+                &local_initCCtx, (void*)&cctxprep,
+                &local_defaultCompress, (void*)(ctx),
+                adv->loopMode, adv->nbSeconds);
 
             if(compressionResults.error) {
                 results.error = compressionResults.error;
                 return results;
             }
 
-            results.result.cSize = compressionResults.result.size;
+            results.result.cSize = compressionResults.result.sumOfReturn;
             ratio = (double)srcSize / (double)results.result.cSize;
             markNb = (markNb+1) % NB_MARKS;
             {   
                 int const ratioAccuracy = (ratio < 10.) ? 3 : 2;
-                double const compressionSpeed = ((double)srcSize / compressionResults.result.time) * 1000;
+                double const compressionSpeed = ((double)srcSize / compressionResults.result.nanoSecPerRun) * 1000;
                 int const cSpeedAccuracy = (compressionSpeed < 10.) ? 2 : 1;
                 results.result.cSpeed = compressionSpeed * 1000000;
                 DISPLAYLEVEL(2, "%2s-%-17.17s :%10u ->%10u (%5.*f),%6.*f MB/s\r",
@@ -536,11 +544,11 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
             dctxprep.dctx = dctx;
             dctxprep.dictBuffer = dictBuffer;
             dctxprep.dictBufferSize = dictBufferSize;
-            decompressionResults = BMK_benchCustom("ZSTD_decompress_generic", nbBlocks, 
+            decompressionResults = BMK_benchFunction(nbBlocks, 
                 (const void * const *)cPtrs, cSizes, resPtrs, resSizes, 
-                &local_initDCtx, &local_defaultDecompress, 
-                (void*)&dctxprep, (void*)(dctx),
-                adv->loopMode, adv->nbSeconds, displayLevel);
+                &local_initDCtx, (void*)&dctxprep, 
+                &local_defaultDecompress, (void*)(dctx),
+                adv->loopMode, adv->nbSeconds);
 
             if(decompressionResults.error) {
                 results.error = decompressionResults.error;
@@ -551,7 +559,7 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
             {   int const ratioAccuracy = (ratio < 10.) ? 3 : 2;
                 double const compressionSpeed = results.result.cSpeed / 1000000;
                 int const cSpeedAccuracy = (compressionSpeed < 10.) ? 2 : 1;
-                double const decompressionSpeed = ((double)srcSize / decompressionResults.result.time) * 1000;
+                double const decompressionSpeed = ((double)srcSize / decompressionResults.result.nanoSecPerRun) * 1000;
                 results.result.dSpeed = decompressionSpeed * 1000000;
                 DISPLAYLEVEL(2, "%2s-%-17.17s :%10u ->%10u (%5.*f),%6.*f MB/s ,%6.1f MB/s \r",
                         marks[markNb], displayName, (U32)srcSize, (U32)results.result.cSize,
@@ -634,7 +642,7 @@ BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
                         ZSTD_CCtx* ctx, ZSTD_DCtx* dctx,
                         int displayLevel, const char* displayName) {
 
-    const BMK_advancedParams_t adv = BMK_defaultAdvancedParams();
+    const BMK_advancedParams_t adv = BMK_initAdvancedParams();
     return BMK_benchMemAdvanced(srcBuffer, srcSize,
                                 fileSizes, nbFiles,
                                 cLevel, comprParams,
@@ -685,31 +693,20 @@ static size_t BMK_findMaxMem(U64 requiredMem)
     return (size_t)(requiredMem);
 }
 
-ERROR_STRUCT(BMK_result_t*, BMK_returnPtr_t);
-
-/* returns average stats over all range [cLevel, cLevelLast] */
-static BMK_returnPtr_t BMK_benchCLevel(const void* srcBuffer, size_t benchedSize,
+static BMK_return_t BMK_benchCLevel(const void* srcBuffer, size_t benchedSize,
                             const size_t* fileSizes, unsigned nbFiles,
-                            const int cLevel, const int cLevelLast, const ZSTD_compressionParameters* comprParams,
+                            const int cLevel, const ZSTD_compressionParameters* comprParams,
                             const void* dictBuffer, size_t dictBufferSize,
                             int displayLevel, const char* displayName,
                             BMK_advancedParams_t const * const adv)
 {
-    int l;
-    BMK_result_t* res = (BMK_result_t*)malloc(sizeof(BMK_result_t) * (cLevelLast - cLevel + 1));
-    BMK_returnPtr_t ret;
+    BMK_return_t res;
 
     const char* pch = strrchr(displayName, '\\'); /* Windows */
-
-    ret.error = 0;
-    ret.result = res;
 
     if (!pch) pch = strrchr(displayName, '/'); /* Linux */
     if (pch) displayName = pch+1;
 
-    if(res == NULL) {
-        EXM_THROW(12, BMK_returnPtr_t, "not enough memory\n");
-    }
     if (adv->realTime) {
         DISPLAYLEVEL(2, "Note : switching to real-time priority \n");
         SET_REALTIME_PRIORITY;
@@ -718,23 +715,14 @@ static BMK_returnPtr_t BMK_benchCLevel(const void* srcBuffer, size_t benchedSize
     if (displayLevel == 1 && !adv->additionalParam)
         DISPLAY("bench %s %s: input %u bytes, %u seconds, %u KB blocks\n", ZSTD_VERSION_STRING, ZSTD_GIT_COMMIT_STRING, (U32)benchedSize, adv->nbSeconds, (U32)(adv->blockSize>>10));
 
-    for (l=cLevel; l <= cLevelLast; l++) {
-        BMK_return_t rettmp;
-        if (l==0) continue;  /* skip level 0 */
-        rettmp = BMK_benchMemCtxless(srcBuffer, benchedSize,
-                fileSizes, nbFiles, 
-                l, comprParams, 
-                dictBuffer, dictBufferSize, 
-                displayLevel, displayName, 
-                adv);
-        if(rettmp.error) {
-            ret.error = rettmp.error;
-            return ret;
-        }
-        res[l-cLevel] = rettmp.result;
-    }
+    res = BMK_benchMemCtxless(srcBuffer, benchedSize,
+            fileSizes, nbFiles, 
+            cLevel, comprParams, 
+            dictBuffer, dictBufferSize, 
+            displayLevel, displayName, 
+            adv);
 
-    return ret;
+    return res;
 }
 
 
@@ -776,8 +764,8 @@ static int BMK_loadFiles(void* buffer, size_t bufferSize,
     return 0;
 }
 
-static BMK_returnSet_t BMK_benchFileTable(const char* const * const fileNamesTable, unsigned const nbFiles,
-                               const char* const dictFileName, int const cLevel, int const cLevelLast,
+static BMK_return_t BMK_benchFileTable(const char* const * const fileNamesTable, unsigned const nbFiles,
+                               const char* const dictFileName, int const cLevel, 
                                const ZSTD_compressionParameters* const compressionParams, int displayLevel,
                                const BMK_advancedParams_t * const adv)
 {
@@ -786,23 +774,20 @@ static BMK_returnSet_t BMK_benchFileTable(const char* const * const fileNamesTab
     void* dictBuffer = NULL;
     size_t dictBufferSize = 0;
     size_t* const fileSizes = (size_t*)calloc(nbFiles, sizeof(size_t));
-    BMK_returnSet_t res;
+    BMK_return_t res;
     U64 const totalSizeToLoad = UTIL_getTotalFileSize(fileNamesTable, nbFiles);
 
-    res.result.cLevel = cLevel;
-    res.result.cLevelLast = cLevelLast;
-    res.result.results = NULL;
-    if (!fileSizes) EXM_THROW(12, BMK_returnSet_t, "not enough memory for fileSizes");
+    if (!fileSizes) EXM_THROW(12, BMK_return_t, "not enough memory for fileSizes");
 
     /* Load dictionary */
     if (dictFileName != NULL) {
         U64 const dictFileSize = UTIL_getFileSize(dictFileName);
         if (dictFileSize > 64 MB)
-            EXM_THROW(10, BMK_returnSet_t, "dictionary file %s too large", dictFileName);
+            EXM_THROW(10, BMK_return_t, "dictionary file %s too large", dictFileName);
         dictBufferSize = (size_t)dictFileSize;
         dictBuffer = malloc(dictBufferSize);
         if (dictBuffer==NULL)
-            EXM_THROW(11, BMK_returnSet_t, "not enough memory for dictionary (%u bytes)",
+            EXM_THROW(11, BMK_return_t, "not enough memory for dictionary (%u bytes)",
                             (U32)dictBufferSize);
         {
             int errorCode = BMK_loadFiles(dictBuffer, dictBufferSize, fileSizes, &dictFileName, 1, displayLevel);
@@ -819,7 +804,7 @@ static BMK_returnSet_t BMK_benchFileTable(const char* const * const fileNamesTab
     if (benchedSize < totalSizeToLoad)
         DISPLAY("Not enough memory; testing %u MB only...\n", (U32)(benchedSize >> 20));
     srcBuffer = malloc(benchedSize);
-    if (!srcBuffer) EXM_THROW(12, BMK_returnSet_t, "not enough memory");
+    if (!srcBuffer) EXM_THROW(12, BMK_return_t, "not enough memory");
 
     /* Load input buffer */
     {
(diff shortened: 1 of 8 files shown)


programs/bench.h
```diff
@@ -19,46 +19,135 @@ extern "C" {
 #define ZSTD_STATIC_LINKING_ONLY   /* ZSTD_compressionParameters */
 #include "zstd.h"     /* ZSTD_compressionParameters */
 
+#define BMK_COMPRESS_ONLY 2
+#define BMK_DECODE_ONLY 1
```
Cyan4973: If you need to switch between 2 modes depending on a parameter, prefer an `enum` type.

programs/bench.h
```diff
@@ -19,46 +19,140 @@ extern "C" {
...
+    BMK_iterMode = 1
+} BMK_loopMode_t;
+
+typedef enum {
+    BMK_both = 0,
+    BMK_decodeOnly = 1,
+    BMK_compressOnly = 2
+} BMK_mode_t;
+
+#define ERROR_STRUCT(baseType, typeName) typedef struct { \
```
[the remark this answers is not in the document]
GeorgeLu97: What should I do instead of this? Since exiting the program is undesirable now, passing out an error result seems necessary. Should error just be manually written as a field of `BMK_result_t` and `BMK_customResult_t`?

programs/bench.h
```diff
@@ -19,46 +19,140 @@ extern "C" {
...
+    int error;       \
+    baseType result; \
+} typeName
+
 typedef struct {
     size_t cSize;
     double cSpeed;   /* bytes / sec */
     double dSpeed;
 } BMK_result_t;
```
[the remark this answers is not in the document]
GeorgeLu97: This is sort of what `BMK_customResult_t` is, and `benchMem()` does use it as you described. I think the differences are .time (now .nanoSecPerRun) is just the inverse of nbRunsPerSec, and that customBench adds all returns of the function run on each `srcBuffers[i]` (so it assumes returns are the size written to in the destination buffer or an error code using `ZSTD_isError()`), which wouldn't make sense with other kinds of return values. 

I'm not quite sure how to best generalize this though, since having the an array field `size_t* returnValue` in the struct would need freeing, which was the same problem had by BMK_resultSet_t (making the caller free the result). The other option I've considered is for `benchCustom` to only accept a single source (so it could return just the return value of the passed in function), which also has some complications, as the only ways I've thought of for this to happen while maintaining the same functionality for `benchMem` and `fullbench` are calling `benchCustom` multiple times (once per block) or leaving the block division up to `*benchFn`. 

In the former solution, we would still need to measure initFn only once, so the first call would have the normal initFn and for later sources we would just have initFn do nothing. It would also complicate timeMode for multiple sources, since the logic for that is internal to customBench. Even with the changes, it seems like this implementation would have different cache behavior than the old benchmarking functions.

The latter seems unnatural since a lot of the benchmarking functions in fullbench don't seem to have a parameter for this (as far as I'm aware of), so it would need to be passed in and implemented. This would also only be possible in the case where the data is stored contiguously, or we'd have to use the former again.

`BMK_mode_t` was intended to be used with paramgrill (decode_only was already an existing flag, and compress_only could be useful for saving an unnecessary operation at times).

programs/bench.h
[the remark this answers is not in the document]
Cyan4973: This construction is correct.

Just, writing it as a macro like this, implies that you are going to use this scheme in multiple places. Which is effectively the case in following code.

I initially thought you would need only to have one case like this, 
and maybe I'm wrong on this point.

Bottom line : if you need it only once, it's more straightforward to write the same thing but directly.

The difference is for the reader of the API : it's easier to understand the relation between the 2 types for a newcomer when they are written directly.
Here, doing the same thing but through a magic macro, it takes time to understand what is going on.
At the very least, I believe it would deserve some more detailed instructions in code comment.

Cyan4973:
General comment : 
`bench.h` presents your contract to users.
The contract should be clear just from reading the `*.h`, 
without a need to look at actual implementation in `*.c`.

As a general structure of the `*.h`,
try to start from the simplest functionality, that newcomer will easily understand,
and then gradually introduce some more complexity as user reads deeper into the `*.h`.

programs/bench.h
```diff
@@ -19,46 +19,130 @@ extern "C" {
...
 
-/* 0 = no Error */
 typedef struct {
-	int errorCode;
-	BMK_result_t result;
-} BMK_return_t;
+    size_t size;    /* sum of return values */
```
Cyan4973: Is the "sum of return values" always meaningful ?
In particular, what happens if one of the return values is an error code ?

Also : if this value is a sum of return values, better name it this way.
Here it's called `size`, but it might not be a size at all.

programs/bench.h
```diff
@@ -19,46 +19,130 @@ extern "C" {
...
+    U64 nanoSecPerRun; /* time per iteration */
+} BMK_customResult_t;
```
Cyan4973: From a naming perspective, `BMK_result_t` and `BMK_customResult_t` feel close.
But they are not.
They represent 2 very different levels of the API.
So it ends up being confusing.

Unfortunately, I'm short of any proposal at this stage.
So let's think about it a bit more.
Naming can be hard at times.

programs/bench.h
```diff
@@ -19,46 +19,130 @@ extern "C" {
...
+ * srcBuffers - an array of buffers to be operated on by benchFn
+ * srcSizes - an array of the sizes of above buffers
+ * dstBuffers - an array of buffers to be written into by benchFn
+ * dstCapacities - an array of the capacities of above buffers.
+ * initFn - (*initFn)(initPayload) is run once per benchmark
+ * benchFn - (*benchFn)(srcBuffers[i], srcSizes[i], dstBuffers[i], dstCapacities[i], benchPayload)
+ *      is run a variable number of times, specified by mode and iter args
+ * mode - if 0, iter will be interpreted as the minimum number of seconds to run
+ * iter - see mode
+ * displayLevel - what gets printed
```
Cyan4973: to update : `displayLevel` is no longer part of the interface of `BMK_benchCustom()`

programs/bench.h
```diff
@@ -19,46 +19,130 @@ extern "C" {
...
+ *      0 : no display;   
+ *      1 : errors;   
+ *      2 : + result + interaction + warnings;   
+ *      3 : + progression;   
+ *      4 : + information
+ * return 
+ *      .error will give a nonzero value if any error has occured
+ *      .result will contain the speed (B/s) and time per loop (ns)
```
Cyan4973: to update, since it doesn't match current content of `BMK_customResult_t`.
Also, what happens to `.result` when `.error != 0` ?

programs/bench.h
```diff
@@ -19,46 +19,130 @@ extern "C" {
...
-int BMK_benchFiles(const char** fileNamesTable, unsigned nbFiles, const char* dictFileName,
-                   int cLevel, int cLevelLast, const ZSTD_compressionParameters* compressionParams, 
-                   int displayLevel);
 
-/* basic benchmarking function, called in paramgrill
- * ctx, dctx must be valid */
+ERROR_STRUCT(BMK_result_t, BMK_return_t);
+ERROR_STRUCT(BMK_customResult_t, BMK_customReturn_t);
+
+/* functionName - name of function
```
Cyan4973: Is `functionName` argument still necessary ? since `BMK_benchCustom()` does no longer display anymore ?

programs/bench.h
```diff
@@ -19,46 +19,130 @@ extern "C" {
...
+/* basic benchmarking function, called in paramgrill ctx, dctx must be provided */
+/* srcBuffer - data source, expected to be valid compressed data if in Decode Only Mode
+ * srcSize - size of data in srcBuffer
+ * cLevel - compression level  
+ * comprParams - basic compression parameters
+ * dictBuffer - a dictionary if used, null otherwise
+ * dictBufferSize - size of dictBuffer, 0 otherwise
+ * ctx - Compression Context
+ * dctx - Decompression Context
+ * diplayLevel - see BMK_benchCustom
```
Cyan4973: to update, since `BMK_benchCustom()` does no longer have this argument.

programs/bench.h
```diff
@@ -19,46 +19,150 @@ extern "C" {
 #define ZSTD_STATIC_LINKING_ONLY   /* ZSTD_compressionParameters */
 #include "zstd.h"     /* ZSTD_compressionParameters */
 
+typedef enum {
```
Cyan4973: It's probably better to introduce these types closer to where they are actually needed.

programs/bench.h
```diff
@@ -19,46 +19,150 @@ extern "C" {
...
+ *          the time per run of benchFn (.nanoSecPerRun). For the former, this
+ *          is generally intended to be used on functions which return the # of bytes written 
+ *          into dstBuffer, hence this value will be the total amount of bytes written to 
+ *          dstBuffer.
+ */
+BMK_customReturn_t BMK_benchFunction(size_t blockCount,
+                        const void* const * const srcBuffers, const size_t* srcSizes,
+                        void* const * const dstBuffers, const size_t* dstCapacities,
+                        size_t (*initFn)(void*), void* initPayload,
+                        size_t (*benchFn)(const void*, size_t, void*, size_t, void*), void* benchPayload,
```
Cyan4973: minor : 
I would select to show `benchFn` at the top of argument list,
because it's the most important element of the interface : 
that's what we want to actually bench.

programs/bench.h
```diff
@@ -19,46 +19,151 @@ extern "C" {
...
+ *      .result - if .error = 0, then .result will contain the sum of all return values of 
+ *          benchFn on the first iteration through all of the blocks (.sumOfReturn) and also 
+ *          the time per run of benchFn (.nanoSecPerRun). For the former, this
+ *          is generally intended to be used on functions which return the # of bytes written 
+ *          into dstBuffer, hence this value will be the total amount of bytes written to 
+ *          dstBuffer.
+ */
+BMK_customReturn_t BMK_benchFunction(                        
+                        size_t (*benchFn)(const void*, size_t, void*, size_t, void*), void* benchPayload,
+                        size_t (*initFn)(void*), void* initPayload,
```
Cyan4973: Is `initFn` optional ?
I think it should be.

Within `fullbench`, we want to benchmark _only_ the target function.
Sometimes, the preparation work can be pretty significant.
For example, it might be necessary to compress data to measure decompression speed,
but we don't want the dominant compression time to pollute decompression speed measurements.

This situation is different from `benchFiles`, 
which tries to attribute a time contribution to `loadDictionary` operation,
by counting it only once (which is a debatable strategy).

Bottom line : 
I'm not sure if `initFn` is actually necessary,
though I understand it's the only way to preserve current strategy used in `benchFiles`.
At the very least, it should be possible to not use it.

programs/bench.c
```diff
@@ -64,9 +64,9 @@
 
 static const size_t maxMemory = (sizeof(size_t)==4)  ?  (2 GB - 64 MB) : (size_t)(1ULL << ((sizeof(size_t)*8)-31));
 
+/* remove this in the future? */
```
Cyan4973: or make it programmable ?

This parameter is used when generating a synthetic sample, to benchmark without any source (`zstd -b`).

We could use the same convention as `fullbench` here, by using `-P#` to modify the compressibility.
The option could remain hidden, and therefore not need any external documentation (just code comments).

programs/bench.c
```diff
@@ -90,121 +90,339 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
-#define MIN(a,b)    ((a) < (b) ? (a) : (b))
-#define MAX(a,b)    ((a) > (b) ? (a) : (b))
+    }
 
-BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
+    for(ind = 0; ind < blockCount; ind++) {
+        srcSize += srcBlockSizes[ind];
+    }
+
+    switch(mode) {
```
Cyan4973: Since these 2 modes are completely different,
could it be better to separate them into 2 functions ?

programs/bench.c
```diff
@@ -213,317 +431,251 @@ BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
     U32 nbBlocks;
 
     /* checks */
-    if (!compressedBuffer || !resultBuffer || !blockTable)
-        EXM_THROW(31, "allocation error : not enough memory");
+    if (!compressedBuffer || !resultBuffer ||  
+        !srcPtrs || !srcSizes || !cPtrs || !cSizes || !resPtrs || !resSizes)
+        EXM_THROW(31, BMK_return_t, "allocation error : not enough memory");
```
Cyan4973: OK, we have a new problem here.

Previously, `EXM_THROW` was exiting the program.
As a consequence, any allocated memory would be freed there just while mopping up the process.

But now, `EXM_THROW` just returns, 
which is a significantly different behavior (likely worth a renaming).
As consequence, if 6 buffers are allocated but the 7th fail, we have a memory leak on exiting the function here.

Make sure all buffers allocated in this function are freed when exiting the function.

This can be difficult when there are multiple exit points, as you'll have to free all resources at each and every exit point. That's a significant burden.

A strategy for this situation is to group all exit responsibilities in one place.
One I like is to "nest" functions into 2 levels : 

```
int function1(...)
{
   ... Allocate stuff ...
   if (allocation_successful)
      result = function2(...)
   ... Free stuff ...
   return result;
}

int function2(...)
{
   /* do not allocate anything here */
   ... do stuff...
   ... exit 1 ...
   ... do more stuff ...
   ... exit 2 ...
   ... do more stuff ...
   ... final exit ...
}
```

In some cases, the amount of variables to pass from function1 to function2 can be too large to feel comfortable.
A second possibility then is to use a `goto` to an exit section.

programs/bench.c
```diff
@@ -213,317 +431,251 @@ BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
...
+            BMK_customReturn_t compressionResults;
+            cctxprep.ctx = ctx;
+            cctxprep.dictBuffer = dictBuffer;
+            cctxprep.dictBufferSize = dictBufferSize;
+            cctxprep.cLevel = cLevel;
+            cctxprep.comprParams = comprParams;
+            cctxprep.adv = adv;
+            /* Compression */
+            DISPLAYLEVEL(2, "%2s-%-17.17s :%10u ->\r", marks[markNb], displayName, (U32)srcSize);
+            compressionResults = BMK_benchFunction(
```
Cyan4973: A side effect of this design is that there is no more any "intermediate refresh".
Now, it's an "all or nothing" strategy.

It still works, though it's less "interactive".
For example, before, it was possible to call something like `zstd -b -i99` and just look at the converging numbers, eventually stopping processing with a Ctrl^C anytime. 
Now it's necessary to wait 99 seconds (with no feedback) to get the compression result, and then another 99 seconds for decompression one. 

I get it that it's more complex to make result display more progressive. We can discuss together a few ideas to see if one is convincing enough to be attempted.

Cyan4973:
The implementation is generally good.
Coding style is fine, expression is clear, so that's a good beginning. This module is getting somewhere.
Now, there are still a number of details to fix, so keep focused on these ones.

programs/bench.c
```diff
@@ -90,440 +87,662 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+
+                    {   U64 loopDuration = compressionResults.result.nanoSecPerRun * nbLoops;
+                        totalLoops += nbLoops;
+                        totalTime += loopDuration;
+                        if (loopDuration > 0) { // nanoSec / run
+                            fastest = MIN(fastest, compressionResults.result.nanoSecPerRun);
+                            nbLoops = (U32)(TIMELOOP_NANOSEC / fastest) + 1;
+                        } else {
+                            assert(nbLoops < 40000000);  /* avoid overflow */
+                            nbLoops *= 2;
```
Cyan4973: if `loopDuration == 0`, you'll want to increase `nbLoops` by more than a factor 2,
otherwise next measurement will likely be insignificant too.

programs/bench.c
```diff
@@ -90,440 +87,662 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+                    }
+
+                    {   U64 loopDuration = compressionResults.result.nanoSecPerRun * nbLoops;
+                        totalLoops += nbLoops;
+                        totalTime += loopDuration;
+                        if (loopDuration > 0) { // nanoSec / run
+                            fastest = MIN(fastest, compressionResults.result.nanoSecPerRun);
+                            nbLoops = (U32)(TIMELOOP_NANOSEC / fastest) + 1;
+                        } else {
+                            assert(nbLoops < 40000000);  /* avoid overflow */
```
Cyan4973: note that this overflow `assert()` was designed to be paired with a `nbLoops` multiplicator of 10.
It should be adapted if you plan using a different multiplicator.

programs/bench.c
```diff
@@ -90,440 +87,662 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+
+                    compressionResults = BMK_benchFunction(&local_defaultCompress, (void*)ctx, &local_initCCtx, (void*)&cctxprep,
+                    nbBlocks, srcPtrs, srcSizes, cPtrs, cSizes, nbLoops);
+                    if(compressionResults.error) {
+                        results.error = compressionResults.error;
+                        return results;
+                    }
+
+                    {   U64 loopDuration = compressionResults.result.nanoSecPerRun * nbLoops;
+                        totalLoops += nbLoops;
```
Cyan4973: I was kind of wondering what was this `totalLoops` for,
but I now understand that you effectively calculate the average speed over all runs.

This is incorrect.
The purpose of having several runs is to select the better one.
The logic behind this choice is that the cpu can be slowed down by a multitude of incontrollable reasons, 
while you can't expect the cpu to perform mysteriously better.

There is an exception to this second part, which is rounding error.
More on this later.

programs/bench.c
```diff
@@ -90,440 +87,662 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+                    }
+
+                    compressionResults = BMK_benchFunction(&local_defaultCompress, (void*)ctx, &local_initCCtx, (void*)&cctxprep,
+                    nbBlocks, srcPtrs, srcSizes, cPtrs, cSizes, nbLoops);
+                    if(compressionResults.error) {
+                        results.error = compressionResults.error;
+                        return results;
+                    }
+
+                    {   U64 loopDuration = compressionResults.result.nanoSecPerRun * nbLoops;
```
Cyan4973: `const` this variable.
It's an important signal for the reviewer that this variable is not going to be modified in some later part of the code.
Otherwise, I have to scrub the code to ensure nothing fancy happen.

Also : 
try to reduce the lifetime of a variable to a minimum possible, whenever practical.

programs/bench.c
```diff
@@ -90,440 +87,662 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+                        totalLoops += nbLoops;
+                        totalTime += loopDuration;
+                        if (loopDuration > 0) { // nanoSec / run
```
Cyan4973: There is another use case which is interesting : 
whenever `loopDuration` is "too small" (to be defined, typically ~0.5sec if the intended run is one second),
there is a risk of a rounding error, which will impact speed results.
This is especially the case when clock resolution is poor,
which can happen : 
some systems can have clock resolution of ~10ms, so it only progresses in huge steps.
In such a situation, if "real" time is somewhere between 10-19ms, the clock will nonetheless measure only 10ms,
so that's an error of up to a factor 2.

Even at 100ms, such discrepancy rounding error remains significant (up to 10%!).

You want to reduce scope for such error,
hence impose a minimum measurement time for a result to be considered valid.

programs/bench.c
```diff
@@ -90,440 +87,662 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
-static U32 g_ldmBucketSizeLog = BMK_LDM_PARAM_NOTSET;
-void BMK_setLdmBucketSizeLog(unsigned ldmBucketSizeLog) {
-    g_ldmBucketSizeLog = ldmBucketSizeLog;
-}
+/* initFn will be measured once, bench fn will be measured x times */
+/* benchFn should return error value or out Size */
+/* takes # of blocks and list of size & stuff for each. */
+/* only does iterations*/
+/* note time/iter could be zero if interval too short */
+BMK_customReturn_t BMK_benchFunction(
```
Cyan4973: Thinking again about it,
I'm wondering if it would be better to have 2 levels of benchFunction.

The lower one is the one defined here, using a nb of iterations as input argument.

But a higher level could use a "time objective" instead,
and proceed with some automatic iter adjustment to reach intended objective.

It feels like a useful level, because otherwise, such logic must be duplicated in almost every place where `BMK_benchFunction()` is called.

programs/bench.h
```diff
@@ -19,46 +19,174 @@ extern "C" {
...
+                        const int cLevel, const ZSTD_compressionParameters* comprParams,
+                        const void* dictBuffer, size_t dictBufferSize,
+                        ZSTD_CCtx* ctx, ZSTD_DCtx* dctx,
+                        int displayLevel, const char* displayName,
+                        const BMK_advancedParams_t* adv);
+
+/* This function times the execution of 2 argument functions, benchFn and initFn  */
+
+/* benchFn - (*benchFn)(srcBuffers[i], srcSizes[i], dstBuffers[i], dstCapacities[i], benchPayload)
+ *      is run a variable number of times, specified by mode and iter args
```
Cyan4973: note that mode is no longer part of the interface

programs/bench.c
GeorgeLu97: I was thinking about this too when I was making the previous changes. But all the current uses of the timing (fullbench, BMK_benchMem) use the timing to print various info as well, so it seems the timed higher level function would need to be able to support this functionality if it wants to serve as a replacement. But you mentioned portability problems with the approach of giving a callback function argument for printing messages, so I wasn't sure how else to do this. Though I suppose benchFunction already unavoidably takes in function pointer arguments to benchmark, so maybe any problems adding the callback would entail were a lost cause anyway.

programs/bench.c
GeorgeLu97: My thought here was that if loopDuration was 0, the execution time of the function is rather minimal, so the execution time of these loops is mostly dominated by the first time loopDuration isn't zero. For example, assuming a 10ms resolution, a 9 ms time could be potentially displayed as 0. With the previous 100x multiplier, this would mean the first nonzero loopDuration would be almost a second long, which means the total execution time could be close to 2 seconds even though we pass in nbSeconds = 1. The main downside is that the getTime functions would need to be called more times, but the overhead for that at most ~25 calls of that, which doesn't seem too large.

Cyan4973:
minor : 
try, in general, to have a commit title a bit more descriptive than "changes" or "updates".

The commit title is what will appear when looking at the history of commits,
which can happen, for example when an investigation requires looking back several commits in the past while hunting a bug or a regression.
So a clean title helps.

The commit title is a short one-liner, so it's pretty limited. In many cases, it cannot describe the full scope of a commit. But it's often enough to transport the most important "idea" behind a commit.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+    DISPLAYLEVEL(1, "Error %i : ", errorNum);         \
     DISPLAYLEVEL(1, __VA_ARGS__);                     \
     DISPLAYLEVEL(1, " \n");                           \
-    exit(error);                                      \
+    r.error = errorNum;                               \
+    return r;                                         \
 }
 
+/* error without displaying */
+#define EXM_THROW_ND(errorNum, retType, ...)  {          \
```
terrelln: nit: align these `\`.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+    if (adv->nbWorkers==1) {
+        ZSTD_CCtx_setParameter(ctx, ZSTD_p_nbWorkers, 0);
+    } else {
+        ZSTD_CCtx_setParameter(ctx, ZSTD_p_nbWorkers, adv->nbWorkers);
+    }
+    ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionLevel, cLevel);
+    ZSTD_CCtx_setParameter(ctx, ZSTD_p_enableLongDistanceMatching, adv->ldmFlag);
+    ZSTD_CCtx_setParameter(ctx, ZSTD_p_ldmMinMatch, adv->ldmMinMatch);
+    ZSTD_CCtx_setParameter(ctx, ZSTD_p_ldmHashLog, adv->ldmHashLog);
+    if (adv->ldmBucketSizeLog != BMK_LDM_PARAM_NOTSET) {
```
terrelln: Why do you need this? Setting these 2 params to `0` should be the same as not setting them.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+    void* addArgs) {
+    size_t moreToFlush = 1;
+    ZSTD_CCtx* ctx = (ZSTD_CCtx*)addArgs;
+    ZSTD_inBuffer in;
+    ZSTD_outBuffer out;
+    in.src = srcBuffer;
+    in.size = srcSize;
+    in.pos = 0;
+    out.dst = dstBuffer;
+    out.size = dstSize;
```
terrelln: We check if `dstSize >= ZSTD_compressBound(srcSize)` and error if it isn't true. Since we don't make more size, we might infinite loop if it isn't true. Or we could error out of the loop if `out` is empty before we call `ZSTD_compress_generic()`.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+    out.pos = 0;
+    while (moreToFlush) {
+        moreToFlush = ZSTD_decompress_generic(dctx,
```
terrelln: We should error if `out` is empty before this call.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+    }
+
+    for(ind = 0; ind < blockCount; ind++) {
+        srcSize += srcBlockSizes[ind];
+    }
+
+    {        
+        unsigned i, j;
+        clockStart = UTIL_getTime();
+        if(initFn != NULL) { (*initFn)(initPayload); }
```
terrelln: Nit: Function pointers dereference into... function pointers, so you can just call `initFn(initPayload)`. Or if you're feeling creative `(********initFn)(initPayload)`.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+            for(j = 0; j < blockCount; j++) {
+                size_t res = (*benchFn)(srcBlockBuffers[j], srcBlockSizes[j], dstBlockBuffers[j], dstBlockCapacities[j], benchPayload);
+                if(ZSTD_isError(res)) {
+                    EXM_THROW_ND(2, BMK_customReturn_t, "Function benchmarking failed on block %u of size %u : %s  \n",
+                        j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
+                } else if(toAdd) {
+                    dstSize += res;
+                }
+            }
+            toAdd = 0;
```
terrelln: Nit: a better name is probably `firstIter`.

programs/bench.c
terrelln: Why would a callback have portability issues? It gets a bit tricky if you're calling from a FFI, but thats the only thing that I can see.

programs/bench.h
```diff
@@ -19,46 +19,206 @@ extern "C" {
...
+ *          into dstBuffer, hence this value will be the total amount of bytes written to 
+ *          dstBuffer.
+ */
+BMK_customReturn_t BMK_benchFunction(                        
+                        size_t (*benchFn)(const void*, size_t, void*, size_t, void*), void* benchPayload,
+                        size_t (*initFn)(void*), void* initPayload,
+                        size_t blockCount,
+                        const void* const * const srcBuffers, const size_t* srcSizes,
+                        void* const * const dstBuffers, const size_t* dstCapacities,
+                        unsigned sec);
```
Cyan4973: I don't get it.
Is that parameter `iter` or `sec` ?

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
-    size_t srcSize;
-    void*  cPtr;
-    size_t cRoom;
-    size_t cSize;
-    void*  resPtr;
-    size_t resSize;
-} blockParam_t;
+#define MINUSABLETIME 500000000ULL
+
+//how to use minusabletime?
```
terrelln: Have you finished this comment?

I recommend marking temporary comments with `TODO`, or `XXX`, or some easily `grep`able string, so you can easily find and remove them.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
-    size_t cSize;
-    void*  resPtr;
-    size_t resSize;
-} blockParam_t;
+#define MINUSABLETIME 500000000ULL
+
+//how to use minusabletime? 
+//only report times which are > minUsable
+void BMK_benchFunctionTimed(
+    size_t (*benchFn)(const void*, size_t, void*, size_t, void*), void* benchPayload,
```
terrelln: This is super verbose, you should define:

```
typedef size_t (*BMK_benchFn_t)(const void*, size_t, void*, size_t, void*);
```

and the same for `initFn`. Then you can just write `BMK_benchFn_t benchFn`.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+    size_t (*initFn)(void*), void* initPayload,
+    size_t blockCount,
+    const void* const * const srcBlockBuffers, const size_t* srcBlockSizes,
```
terrelln: Nit: `const void* const* const`.

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+            if (loopDuration > 0) { 
+                fastest = MIN(fastest, cont->intermediateResult.result.nanoSecPerRun);
+                cont->intermediateResult.result.nanoSecPerRun = fastest;
+                cont->state.nbLoops = (U32)(TIMELOOP_NANOSEC / fastest) + 1;
+            } else {
+                const unsigned multiplier = 2;
+                assert(cont->state.nbLoops < ((unsigned)-1) / multiplier);  /* avoid overflow */
+                cont->state.nbLoops *= multiplier;
+            }
+            if(loopDuration < MINUSABLETIME) { /* don't report results which have time too low */
```
terrelln: Should this be before the previous if/else? Otherwise we will report the `fastest` even when < `MINUSABLETIME`.

programs/bench.h
```diff
@@ -19,46 +19,206 @@ extern "C" {
...
+/* 
+ * initializes the last argument of benchFunctionTimed, with iter being the number of seconds to bench (see below)
+ */
+BMK_customResultContinuation_t BMK_init_customResultContinuation(unsigned iter);
+
+/* 
+ * Benchmarks custom functions like BMK_benchFunction(), but runs for iter seconds rather than a fixed number of iterations
+ * arguments mostly the same other than BMK_benchFunction()
+ * Usage - benchFunctionTimed will return in approximately one second, where the intermediate results can be found in 
+ * the *cont passed in and be displayed/used as wanted. Keep calling BMK_benchFunctionTimed() until cont->completed = 1
```
Cyan4973: What's the benefit of `cont->completed` ?
Can't the caller just iterate around `BMK_benchFunctionTimed()` as it wishes, and stop when it wishes ?

programs/bench.c
```diff
@@ -90,440 +87,677 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+            srcBuffer, srcSize, fileSizes, nbFiles, cLevel, comprParams,
+            dictBuffer, dictBufferSize, ctx, dctx, displayLevel, displayName, adv);
+    }
     /* clean up */
-    free(blockTable);
     free(compressedBuffer);
     free(resultBuffer);
-    results.errorCode = 0;
+
+    free((void*)srcPtrs);
```
terrelln: Make sure you are testing that everything gets freed. Run with ASAN, which should catch leaks. To double check remove one of these frees and make sure it crashes.

programs/bench.h
```diff
@@ -19,46 +19,206 @@ extern "C" {
...
+    unsigned nbLoops;
+    U64 timeRemaining;
+    UTIL_time_t coolTime;
+} BMK_timeState_t;
+
+typedef struct {
+    int completed;
+    BMK_customReturn_t intermediateResult; /* since the wrapper can't err, don't need ERROR_STRUCT(cRC, just check here) */
+    BMK_timeState_t state; 
+} BMK_customResultContinuation_t;
```
Cyan4973: Try to distinguish a _state_ from a _result_.

A _state_ is something we have to carry around, as a way to maintain some relation between successive invocations, or for performance. We would pretty much prefer __not__ have a _state_, if we could get the same benefits without it. It's all technical debt.
A _state_ is passed as a parameter, and is preferably opaque (we should not mess with its internals). 
The best solution is a pointer to incomplete type, which is fully opaque.
But it then requires a constructor and destructor (it's effectively the equivalent of an object).
It's sometimes beneficial to define a state directly as a struct, so that it can be easily allocated on stack, or as part of a larger struct. In which case, since __C__ doesn't have a concept of "private member", mention in documentation that the structure must be considered opaque, and make sure that it's effectively the case (users never need to invoke its members).

A _result_ is something we actually want.
Except when the result requires some big buffer, 
it's generally preferable to have the result provided directly within return type.

programs/bench.c
```diff
@@ -601,44 +834,65 @@ static void BMK_loadFiles(void* buffer, size_t bufferSize,
...
+        if (dictBuffer==NULL) {
+            free(fileSizes);
+            EXM_THROW(11, BMK_return_t, "not enough memory for dictionary (%u bytes)",
                             (U32)dictBufferSize);
-        BMK_loadFiles(dictBuffer, dictBufferSize, fileSizes, &dictFileName, 1, displayLevel);
+        }
+        {
+            int errorCode = BMK_loadFiles(dictBuffer, dictBufferSize, fileSizes, &dictFileName, 1, displayLevel);
+            if(errorCode) {
+                res.error = errorCode;
```
terrelln: Looks like you're leaking `fileSizes`, and `dictBuffer` here. You might want to use a `goto out`, or a `_internal()` function to help with the cleanup.

programs/bench.c
```diff
@@ -647,96 +901,80 @@ static void BMK_benchFileTable(const char* const * const fileNamesTable, unsigne
...
-            BMK_benchCLevel(srcPtr, fileSize,
-                            fileSizes+fileNb, 1, 
-                            cLevel, cLevelLast, compressionParams,
-                            dictBuffer, dictBufferSize, 
-                            displayLevel, fileNamesTable[fileNb]);
-            srcPtr += fileSize;
+    {
+        int errorCode = BMK_loadFiles(srcBuffer, benchedSize, fileSizes, fileNamesTable, nbFiles, displayLevel);
+        if(errorCode) {
+            res.error = errorCode;
```
terrelln: Same here.

programs/bench.c
```diff
@@ -647,96 +901,80 @@ static void BMK_benchFileTable(const char* const * const fileNamesTable, unsigne
...
+                    cLevel, compressionParams, 
                     NULL, 0, 
-                    displayLevel, name);
+                    displayLevel, name, adv);
 
     /* clean up */
-    free(srcBuffer);
-}
-
+    free((void*)srcBuffer);
```
terrelln: No need to cast to `void*`.

programs/bench.h
```diff
@@ -19,46 +19,206 @@ extern "C" {
...
 
+/* Creates a struct of type typeName with an int type .error field
+ * and a .result field of some baseType. Functions with return
+ * typeName pass a successful result with .error = 0 and .result
+ * with the intended result, while returning an error will result
+ * in .error != 0. 
+ */
+#define ERROR_STRUCT(baseType, typeName) typedef struct { \
+    int error;       \
+    baseType result; \
```
terrelln: nit: Put `result` first. It is likely to have higher alignment requirements than an `int`.

programs/bench.h
```diff
@@ -19,46 +19,206 @@ extern "C" {
...
+} BMK_loopMode_t;
+
+typedef enum {
+    BMK_both = 0,
+    BMK_decodeOnly = 1,
+    BMK_compressOnly = 2
+} BMK_mode_t;
+
+typedef struct {
+    BMK_mode_t mode; /* 0: all, 1: compress only 2: decode only */
```
terrelln: nit: align these comments.

programs/zstdcli.c
```diff
@@ -700,11 +702,24 @@ int main(int argCount, const char* argv[])
...
+                            adv.additionalParam = (int)readU32FromChar(&argument);
                         } else
 #endif
                             main_pause=1;
                         break;
+
+                        /* Select compressibility of synthetic sample */
+                    case 'P': 
+                    {   U32 proba32 = 0;
+                        while ((argument[1]>= '0') && (argument[1]<= '9')) {
```
terrelln: Can we use `readU32FromChar()` here?

Also, should we document this parameter in the help/man pages?

programs/zstdcli.c
```diff
@@ -801,21 +816,46 @@ int main(int argCount, const char* argv[])
...
+        adv.blockSize = blockSize;
+        adv.nbWorkers = nbWorkers;
+        adv.realTime = setRealTimePrio;
+        adv.nbSeconds = bench_nbSeconds;
+        adv.ldmFlag = ldmFlag;
+        adv.ldmMinMatch = g_ldmMinMatch;
+        adv.ldmHashLog = g_ldmHashLog;
         if (g_ldmBucketSizeLog != LDM_PARAM_DEFAULT) {
-            BMK_setLdmBucketSizeLog(g_ldmBucketSizeLog);
+            adv.ldmBucketSizeLog = g_ldmBucketSizeLog;
```
terrelln: Note: I think we can remove the `LDM_PARAM_DEFAULT` logic. We can leave it for now, and I can clean it up in a later patch.

tests/fullbench.c
```diff
@@ -288,8 +290,9 @@ static size_t benchMem(const void* src, size_t srcSize, U32 benchNb)
     size_t const dstBuffSize = ZSTD_compressBound(srcSize);
     void*  buff2;
     const char* benchName;
-    size_t (*benchFunction)(void* dst, size_t dstSize, void* verifBuff, const void* src, size_t srcSize);
-    double bestTime = 100000000.;
+    size_t (*benchFunction)(const void* src, size_t srcSize, void* dst, size_t dstSize, void* verifBuff);
```
terrelln: Lets use the same typedef from earlier.

programs/bench.c
GeorgeLu97: VS gives a 'different const qualifiers' warning on the free() call without the cast.

programs/bench.c
Cyan4973: `valgrind` can be a useful tool to detect leaks too.
Though it requires a linux VM.

programs/bench.c
Cyan4973: I would expect that in previous function `free((void*)srcPtrs);`, due to a bug in Visual,

but here, Visual should not complain. `srcBuffer` is clearly defined as `void*`, not `const void*`.

programs/bench.h
```diff
@@ -19,46 +19,205 @@ extern "C" {
...
+                        BMK_benchFn_t benchFn, void* benchPayload,
+                        BMK_initFn_t initFn, void* initPayload,
+                        size_t blockCount,
+                        const void* const * const srcBuffers, const size_t* srcSizes,
+                        void* const * const dstBuffers, const size_t* dstCapacities,
+                        unsigned nbLoops);
+
+
+/* state information needed to advance computation for benchFunctionTimed */
+typedef struct BMK_timeState_t* BMK_timedFnState_t;
```
Cyan4973: Don't hide the `*` ptr nature in a typedef.
This makes it impossible to used `const` property later on.

programs/bench.h
```diff
@@ -19,46 +19,205 @@ extern "C" {
...
+/* initializes timeState object with desired number of seconds */
+BMK_timedFnState_t BMK_initTimeState(unsigned nbSeconds);
```
Cyan4973: This is not an initialization. 
It is an object creation, implying an allocation, implying a requirement to free the object on exit.
Prefer a name like `*_create*()`.

programs/bench.h
```diff
@@ -19,46 +19,205 @@ extern "C" {
...
+typedef struct {
+    BMK_customReturn_t result;
+    int completed;
+} BMK_customTimedReturn_t;
+
+/* 
+ * Benchmarks custom functions like BMK_benchFunction(), but runs for nbSeconds seconds rather than a fixed number of loops
+ * arguments mostly the same other than BMK_benchFunction()
+ * Usage - benchFunctionTimed will return in approximately one second, where the intermediate results can be found in 
+ * the *cont passed in and be displayed/used as wanted. Keep calling BMK_benchFunctionTimed() until cont->completed = 1
```
Cyan4973: Comment out of date.
Information is in return type.

programs/bench.c
GeorgeLu97: Oops, yeah. I accidentally deleted the wrong cast when looking at this comment last time, it should be fixed now.

Cyan4973:
I've been tested d6121ad on my laptop,
using `zstd -b FILE`.

Things noticed : 
- The display updates, but does compression only first, and only proceed to decompression once compression is completed (instead of interleaving them).
- Decompression speed tests are skipped, result is `0.0 MB/s`.

The second point is recent, since it does not happen in ab26f24c9c8ec8e6435c09d8c69b1a8faa19b852.

Make sure you test your changes.
Automated tests are always better when possible, since they will free your time and your mind having to think and control everything.

Cyan4973:
OK, I believe this PR is fine.

The last regression I noticed is that
when requesting to benchmark a file across multiple compression levels,
the file gets reloaded between each level.

Depending on the file size, this is an invisible additional load, or a noticeable one.
For example, when benchmarking `enwik9`, an almost 1 GB file, 
loading delay becomes quite perceptible.
Loading a lot of small file can also result in perceptible delays.

I think this is an optimization, which could be done in a separate PR.
This PR is fine as is, basically good to merge.
It will be merged into `dev` after the creation of v1.3.5, later this week.

Cyan4973:
merged into `dev`

PR #1189, opened 2018-06-14
histogram module

Cyan4973:
It always bugged me that `HUF_compress()` needs to invoke `FSE_count()` to build its histogram.
Histogram functions can be useful for more than just FSE. As a recent example, the optimal parser uses this capability to populate its initial statistics.

This patch regroups all histogram functions into `hist.c`
and rename them with `HIST_*` prefix.
This focus also helped make the interface slightly cleaner (within `hist.h`).

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -55,6 +55,7 @@ SET(Sources
 
 SET(Headers
         ${LIBRARY_DIR}/zstd.h
+        ${LIBRARY_DIR}/common/debug.h
         ${LIBRARY_DIR}/common/pool.h
         ${LIBRARY_DIR}/common/threading.h
         ${LIBRARY_DIR}/common/bitstream.h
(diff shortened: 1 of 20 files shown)


build/cmake/lib/CMakeLists.txt
```diff
@@ -63,6 +64,7 @@ SET(Headers
         ${LIBRARY_DIR}/common/huf.h
         ${LIBRARY_DIR}/common/mem.h
         ${LIBRARY_DIR}/common/zstd_internal.h
+        ${LIBRARY_DIR}/compress/hist.h
```
terrelln: Do we need `debug.h` here as well?

lib/compress/hist.c
```diff
@@ -0,0 +1,200 @@
...
+
+    return largestCount;
+}
+
+
+/* HIST_count_parallel_wksp() :
+ * Same as HIST_count_parallel(), but using an externally provided scratch buffer.
+ * `workSpace` size must be a table of size >= HIST_WKSP_SIZE_U32.
+ * @return : largest histogram frequency,
+ *           or an error code (notably when histogram would be larger than *maxSymbolValuePtr). */
```
terrelln: Can we just delete the doc-blocks from the `.c` file, so we don't have to keep them in sync with the `.h` documentation.

Cyan4973: Note that `HIST_count_parallel_wksp()` is not part of the API,
it's only an internal function.
So I guess we need to keep the doc block of this function in side `hist.c`.

Duplicated ones can be removed.

PR #1192, opened 2018-06-15
Document parallel zstd options in zstd.1

bmwiedemann:
pulled from
https://build.opensuse.org/request/show/617021


Cyan4973:
The current plan is to deprecate `pzstd`, since `zstd` already supports multi-threading directly.

Cyan4973:
Yes : 
https://github.com/facebook/zstd/blob/dev/programs/zstd.1#L118

PR #1193, opened 2018-06-18
removed constants from `zstd.h`

Cyan4973:
Some constants were defined as `static const` within `zstd.h`, experimental section (`ZSTD_STATIC_LINKING_ONLY`).

Problem is, some compilers complain when those constants are not used by the user program,
typically when combined with `-Wunused-const-variable` on `gcc` :
https://trac.torproject.org/projects/tor/ticket/26272

Reverted to `#define`, which do not trigger such warning.

Note : as a consequence, some constants are now `#define` under 2 different names,
in an attempt to preserve availability of both names,
in order to not break potential source code depending on the `static const` name.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -408,10 +408,12 @@ ZSTDLIB_API size_t ZSTD_DStreamOutSize(void);   /*!< recommended size for output
 #define ZSTD_FRAMEHEADERSIZE_PREFIX 5   /* minimum input size to know frame header size */
 #define ZSTD_FRAMEHEADERSIZE_MIN    6
 #define ZSTD_FRAMEHEADERSIZE_MAX   18   /* for static allocation */
-static const size_t ZSTD_frameHeaderSize_prefix = ZSTD_FRAMEHEADERSIZE_PREFIX;
-static const size_t ZSTD_frameHeaderSize_min = ZSTD_FRAMEHEADERSIZE_MIN;
-static const size_t ZSTD_frameHeaderSize_max = ZSTD_FRAMEHEADERSIZE_MAX;
-static const size_t ZSTD_skippableHeaderSize = 8;  /* magic number + skippable frame length */
+
+#define ZSTD_frameHeaderSize_prefix ZSTD_FRAMEHEADERSIZE_PREFIX
+#define ZSTD_frameHeaderSize_min    ZSTD_FRAMEHEADERSIZE_MIN
+#define ZSTD_frameHeaderSize_max    ZSTD_FRAMEHEADERSIZE_MAX
+#define ZSTD_skippableHeaderSize    8
+
 
 
 /*--- Advanced types ---*/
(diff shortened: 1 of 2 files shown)


Cyan4973:
mmmh, not sure if this is a right thing to do after all.

While it's indeed possible to change `static const` into a `#define` for simple values like integers,
it's a less straightforward thing for pre-defined structures.
Visual will complain this is outside of C standard (see `C4576` : "a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax" : https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c4400-through-c4599)

I have no great solution for the second case.
Instead of a constant, an initializer function could be used to achieve the same objective, though it would introduce 2 drawbacks : 
- Implementation is separated, typically into `zstd_common.c` 
- Any call to `ZSTD_defaultCMem` becomes a function call, hence introduces one indirection. This happens in many places throughout the code, though mostly within constructors, so I'm unsure if it really matters.

Last idea : keep it this way, since it's considered "better" to have `static const` instead of `#define`, users should simply avoid triggering `-Wunused-const-variable` (which is not enabled by default)

lib/zstd.h
```diff
@@ -408,10 +408,12 @@ ZSTDLIB_API size_t ZSTD_DStreamOutSize(void);   /*!< recommended size for output
 #define ZSTD_FRAMEHEADERSIZE_PREFIX 5   /* minimum input size to know frame header size */
 #define ZSTD_FRAMEHEADERSIZE_MIN    6
 #define ZSTD_FRAMEHEADERSIZE_MAX   18   /* for static allocation */
-static const size_t ZSTD_frameHeaderSize_prefix = ZSTD_FRAMEHEADERSIZE_PREFIX;
-static const size_t ZSTD_frameHeaderSize_min = ZSTD_FRAMEHEADERSIZE_MIN;
-static const size_t ZSTD_frameHeaderSize_max = ZSTD_FRAMEHEADERSIZE_MAX;
-static const size_t ZSTD_skippableHeaderSize = 8;  /* magic number + skippable frame length */
+
+#define ZSTD_frameHeaderSize_prefix ZSTD_FRAMEHEADERSIZE_PREFIX
```
terrelln: Do we want to keep these around?

Cyan4973: That's the intention : any code using `ZSTD_frameHeaderSize_prefix` will continue to work properly with this `#define`.

Now, since it's in the `ZSTD_STATIC_LINKING_ONLY` section, we have no obligation.
It's just that I felt it would be "nicer" for any potential user of these constants.
I have no strong feeling about it.

lib/zstd.h
```diff
@@ -600,7 +602,7 @@ ZSTDLIB_API const ZSTD_DDict* ZSTD_initStaticDDict(
 typedef void* (*ZSTD_allocFunction) (void* opaque, size_t size);
 typedef void  (*ZSTD_freeFunction) (void* opaque, void* address);
 typedef struct { ZSTD_allocFunction customAlloc; ZSTD_freeFunction customFree; void* opaque; } ZSTD_customMem;
-static ZSTD_customMem const ZSTD_defaultCMem = { NULL, NULL, NULL };  /**< this constant defers to stdlib's functions */
+#define ZSTD_defaultCMem ((ZSTD_customMem){ NULL, NULL, NULL})   /**< this defers memory management to stdlib's functions */
```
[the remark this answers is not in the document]
Cyan4973: In most cases, `ZSTD_defaultCMem` is used directly as function argument.
It would be quite a change to force the creation of a local variable just to store `{NULL, NULL, NULL}` into it.

Cyan4973:
I'll close this.
I'm no longer convinced it's worth it.
Gain is tiny, and side effects, though small too, seems larger, so it does not feel like a great deal.

PR #1196, opened 2018-06-19
ZSTD_btopt: Support Searching the Dictionary Context In-Place

felixhandte:
This is the final PR in the stack of #1185, #1169, #1151, and #1117 that modify Zstd's compression strategies to allow searching an independent, immutable dictionary context. This PR extends support to the `ZSTD_btopt` and `ZSTD_btultra` levels.

Two further work items are planned (in separate PRs):

- Improve the heuristic that controls when to attach a dictionary vs when to copy it into the working context.
- Provide a user-accessible override of that heuristic to allow forcing / disallowing attaching the dictionary.


lib/compress/zstd_opt.c
```diff
@@ -589,17 +606,47 @@ U32 ZSTD_insertBtAndGetAllMatches (
...
+                    matchIndex3 = dmsMatchIndex3 + dmsIndexDelta;
+                    DEBUGLOG(8, "found small dms match with hlog3, of length %u",
+                                (U32)mlen);
+                    bestLength = mlen;
+                    assert(current > matchIndex3);
+                    assert(mnum==0);  /* no prior solution */
+                    matches[0].off = (current - matchIndex3) + ZSTD_REP_MOVE;
+                    matches[0].len = (U32)mlen;
+                    mnum = 1;
+                    if ( (mlen > sufficient_len) |
```
terrelln: We could put this outside the `if`, with `if (mnum > 0 && (matches[0].len > ...)`, and maybe add an `assert(mnum == 1);`

lib/compress/zstd_opt.c
```diff
@@ -620,9 +667,10 @@ U32 ZSTD_insertBtAndGetAllMatches (
             matches[mnum].off = (current - matchIndex) + ZSTD_REP_MOVE;
             matches[mnum].len = (U32)matchLength;
             mnum++;
-            if (matchLength > ZSTD_OPT_NUM) break;
-            if (ip+matchLength == iLimit) {  /* equal : no way to know if inf or sup */
-                break;   /* drop, to preserve bt consistency (miss a little bit of compression) */
+            if ( (matchLength > ZSTD_OPT_NUM)
```
terrelln: @Cyan4973 should this be `sufficient_len` like above, to save a little bit of searches?

lib/compress/zstd_opt.c
```diff
@@ -643,6 +691,45 @@ U32 ZSTD_insertBtAndGetAllMatches (
...
+
+            if (matchLength > bestLength) {
+                matchIndex = dictMatchIndex + dmsIndexDelta;
+                if (matchLength > matchEndIdx - matchIndex)
+                    matchEndIdx = matchIndex + (U32)matchLength;
+                bestLength = matchLength;
+                matches[mnum].off = (current - matchIndex) + ZSTD_REP_MOVE;
+                matches[mnum].len = (U32)matchLength;
+                mnum++;
+                if ( (matchLength > ZSTD_OPT_NUM)
```
terrelln: Same comment with `sufficient_len` here.

lib/compress/zstd_opt.c
```diff
@@ -399,7 +400,7 @@ static U32 ZSTD_insertBt1(
     assert(ip <= iend-8);   /* required for h calculation */
     hashTable[h] = current;   /* Update Hash Table */
 
-    while (nbCompares-- && (matchIndex > windowLow)) {
+    while (nbCompares-- && (matchIndex >= matchLow)) {
```
terrelln: Why change this?

felixhandte:
Comments as to the current state of this PR:

- In non-dict scenarios, when reusing contexts, the `nextToUpdate`/`matchLow` changes seem to produce very small compression ratio improvements (~.04% @ 1KB inputs).
- Rarely, when compressing with a dictionary, there can be very very small compression ratio regressions (~.004%, i.e. 1 byte, in the example I found). This seems to be related to running out of comparisons earlier due to having to start at the root of the dict binary tree, rather than having the benefit of already having narrowed the search range with the comparisons spent on matches in the input. It's not clear that there's anything we can do about this. I'm surprised it isn't a larger effect.
- Compression still has perf problems (up to 0-5% slower).

lib/compress/zstd_opt.c
```diff
@@ -589,17 +606,47 @@ U32 ZSTD_insertBtAndGetAllMatches (
...
                     return 1;
-    }   }   }   }
+                }
+            }
+        } else if (dictMode == ZSTD_dictMatchState) {
+            /* should we perform this search even if the working ctx search
+             * found a match? */
+            U32 const hashLog3 = ms->hashLog3;
+            size_t const hash3 = ZSTD_hash3Ptr(ip, hashLog3);
+            U32 const dmsMatchIndex3 = dms->hashTable3[hash3];
```
terrelln: This hash table isn't actually populated in the dictionaries currently, which is causing the segfault you're seeing.

terrelln: If you add it, make sure to update https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1300. Not sure if it is worth adding it though.

lib/compress/zstd_opt.c
felixhandte: This allows us to match against the first position in the window, which was previously not possible. A separate limit variable needs to be introduced so we can avoid trying to match against index 0, which has a lot of spurious matches pointing at it.

lib/compress/zstd_opt.c
Cyan4973: Initially, the search used to stop on reaching `sufficient_len`.
But, if my memory serves me right, 
it was found that using `sufficient_len` as a search limit directly within the binary tree
was hurting compression performance too much.

This conclusion obviously depends on test sample,
but I kind of remember that some corner cases were significantly impacted,
with `btlazy2` achieving better performance just because of this.

The trade-off was to use `sufficient_len` mainly within the parser,
as opposed to the match finder.

Note though that `sufficient_len` is still used in hc3 and repcode finder,
probably because the conversion was half-done,
and the benchmark results were likely fixed just with this half.

This could be re-visited if need be.
This is part of the compression / speed trade-off territory.

However, it may take time to properly benchmark this scenario in order to argue that the new solution is better.

Cyan4973:
I couldn't find the same performance issue.
Here are benchmark results from a few configurations :
- macOS, llvm9 : -2% slower than `dev`
- macOS, gcc7 : identical performance
- ubuntu, clang3.8 : -3% slower than `dev`
- ubuntu, gcc5.4 : +3% faster than `dev`

So results do not point towards a single direction.
As importantly, these results do not match your experience, which is basically reversed (clang better, gcc worse).

This suggests performance impact is almost random.
So it becomes hard to optimize that.

I've also looked at the code, and couldn't find any thing obvious to optimize.

felixhandte:
@Cyan4973, yeah. I'm sort of all out of ideas here, other than the one commit I just pushed which has maybe marginal positive impact. I see Travis has identified a segfault, so I'm investigating that, but otherwise how do we feel about the perf? Are we comfortable interpreting the perf differences as random churn and accepting this as good enough to merge?

Cyan4973:
The performance difference isn't a clear loss,
so yes, it qualifies as "random churn".

At this stage, the most important step is to fix the segfault.
Who knows, maybe the segfault fix will also impact performance in subtle ways ...

Cyan4973:
Results after this last change, on macOS : 
macOS, llvm9 : identical performance
macOS, gcc7 : -3% slower than dev

This is in line with @felixhandte experience, 
but also different from previous measurement, using 5d81f71.
It confirms the random nature of performance impact related to these changes.

PR #1197, opened 2018-06-20
Thread Pool resize

Cyan4973:
This patch makes `POOL_ctx` resizable.
Up to now, when number of threads was changed, it was only possible to `free` and then `create` a new `POOL_ctx`.

This capability, combined with a few more resizable structures inside `ZSTDMT`, dramatically reduces the number of threads created in situation where the number of requested threads frequently changes,
which is the case with `zstreamtest --newapi`.

The expected benefits are :
- thread sanitizer : this test max out pretty quickly when the nb of threads created becomes large. It stops on reaching a limit. The limit should now be a lesser problem.
- less flaky qemu ARM test : we regularly have random crashes when testing ARM with qemu, but specifically during test of `zstreamtest --newapi`. This behavior is suspected related to the high number of created and destroyed threads during this test. This patch reducing the number of thread creation / destruction, it's expected to have a positive impact on qemu ARM test, making them more stable.

As a reference point, running ~3000 tests with `zstreamtest --newapi` used to create ~2000 threads, it now only creates ~50 threads, a 40x reduction.
The test continues creating threads even with this patch because, on top of regularly changing the nb of threads, it also free / create CCtx randomly. In order to reduce the nb of created threads even more, it would be enough to reduce the probability to free / create CCtx.

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -33,8 +33,9 @@ typedef struct POOL_job_s {
 struct POOL_ctx_s {
     ZSTD_customMem customMem;
     /* Keep track of the threads */
-    ZSTD_pthread_t *threads;
-    size_t numThreads;
+    ZSTD_pthread_t* threads;
+    size_t threadCapacity;
+    size_t threadLimit;
 
     /* The queue is a circular buffer */
     POOL_job *queue;
@@ -58,10 +59,10 @@ struct POOL_ctx_s {
 };
 
 /* POOL_thread() :
-   Work thread for the thread pool.
-   Waits for jobs and executes them.
-   @returns : NULL on failure else non-null.
-*/
+ * Work thread for the thread pool.
+ * Waits for jobs and executes them.
+ * @returns : NULL on failure else non-null.
+ */
 static void* POOL_thread(void* opaque) {
     POOL_ctx* const ctx = (POOL_ctx*)opaque;
     if (!ctx) { return NULL; }
@@ -103,16 +104,17 @@ POOL_ctx* POOL_create(size_t numThreads, size_t queueSize) {
     return POOL_create_advanced(numThreads, queueSize, ZSTD_defaultCMem);
 }
 
-POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize, ZSTD_customMem customMem) {
+POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize,
+                               ZSTD_customMem customMem) {
     POOL_ctx* ctx;
-    /* Check the parameters */
+    /* Check parameters */
     if (!numThreads) { return NULL; }
     /* Allocate the context and zero initialize */
     ctx = (POOL_ctx*)ZSTD_calloc(sizeof(POOL_ctx), customMem);
     if (!ctx) { return NULL; }
     /* Initialize the job queue.
-     * It needs one extra space since one space is wasted to differentiate empty
-     * and full queues.
+     * It needs one extra space since one space is wasted to differentiate
+     * empty and full queues.
      */
     ctx->queueSize = queueSize + 1;
     ctx->queue = (POOL_job*)ZSTD_malloc(ctx->queueSize * sizeof(POOL_job), customMem);
@@ -126,7 +128,7 @@ POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize, ZSTD_customM
     ctx->shutdown = 0;
     /* Allocate space for the thread handles */
     ctx->threads = (ZSTD_pthread_t*)ZSTD_malloc(numThreads * sizeof(ZSTD_pthread_t), customMem);
-    ctx->numThreads = 0;
+    ctx->threadCapacity = 0;
     ctx->customMem = customMem;
     /* Check for errors */
     if (!ctx->threads || !ctx->queue) { POOL_free(ctx); return NULL; }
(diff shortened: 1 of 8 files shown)


lib/common/pool.h
```diff
@@ -30,40 +30,49 @@ typedef struct POOL_ctx_s POOL_ctx;
...
+ */
 void POOL_free(POOL_ctx* ctx);
 
+/*! POOL_resize() :
+ *  Expands or shrinks pool's number of threads.
+ *  This is more efficient than releasing and creating a new context.
+ * @return : a new pool context on success, NULL on failure
+ *    note : new pool context might have same address as original one, but it's not guaranteed.
+ *           consider starting context as consumed, only rely on returned one.
+ *    note 2 : only numThreads can be resized, queueSize is unchanged.
```
terrelln: We should add a note that the number of threads can only be effectively reduced when `queueSize == 1`, or change the behavior.

Cyan4973: Not sure to understand this one.
I don't see a relation between the capability to change the nb of threads, and `queueSize` value ?

lib/common/pool.c
```diff
@@ -172,24 +175,69 @@ void POOL_free(POOL_ctx *ctx) {
...
- * then a queue is empty if there is a thread free and no job is waiting.
+ * When queueSize is 1 (pool was created with an intended queueSize of 0),
+ * then a queue is empty if there is a thread free _and_ no job is waiting.
  */
 static int isQueueFull(POOL_ctx const* ctx) {
     if (ctx->queueSize > 1) {
         return ctx->queueHead == ((ctx->queueTail + 1) % ctx->queueSize);
     } else {
-        return ctx->numThreadsBusy == ctx->numThreads ||
+        return ctx->numThreadsBusy == ctx->threadLimit ||
```
terrelln: `threadLimit` is only used when `queueSize == 1`, so you can't reduce the number of threads when `queueSize > 1`.

lib/common/pool.c
```diff
@@ -58,25 +60,25 @@ struct POOL_ctx_s {
...
     POOL_ctx* const ctx = (POOL_ctx*)opaque;
     if (!ctx) { return NULL; }
     for (;;) {
         /* Lock the mutex and wait for a non-empty queue or until shutdown */
         ZSTD_pthread_mutex_lock(&ctx->queueMutex);
 
-        while (ctx->queueEmpty && !ctx->shutdown) {
+        while ( ctx->queueEmpty
+            || (ctx->numThreadsBusy >= ctx->threadLimit) ) {
+             if (ctx->shutdown) {
```
terrelln: What if `ctx->shutdown && !ctx->queueEmpty && ctx->numThreadsBusy >= ctx->threadLimit`? Might it drop work in this case?

lib/common/pool.c
```diff
@@ -172,24 +177,69 @@ void POOL_free(POOL_ctx *ctx) {
...
 size_t POOL_sizeof(POOL_ctx *ctx) {
     if (ctx==NULL) return 0;  /* supports sizeof NULL */
     return sizeof(*ctx)
         + ctx->queueSize * sizeof(POOL_job)
-        + ctx->numThreads * sizeof(ZSTD_pthread_t);
+        + ctx->threadCapacity * sizeof(ZSTD_pthread_t);
+}
+
+
+/* note : only works if no job is running ! */
```
terrelln: Lets assert/check this condition.

lib/common/pool.c
```diff
@@ -89,30 +91,32 @@ static void* POOL_thread(void* opaque) {
             job.function(job.opaque);
 
             /* If the intended queue size was 0, signal after finishing job */
+            ZSTD_pthread_mutex_lock(&ctx->queueMutex);
+            ctx->numThreadsBusy--;
             if (ctx->queueSize == 1) {
-                ZSTD_pthread_mutex_lock(&ctx->queueMutex);
-                ctx->numThreadsBusy--;
-                ZSTD_pthread_mutex_unlock(&ctx->queueMutex);
                 ZSTD_pthread_cond_signal(&ctx->queuePushCond);
```
terrelln: We always need to signal now, or at least signal if `numThreadsBusy == theadLimit` before decrementing.

terrelln: That comment isn't quite right. However, I think we should move the increment/decrement of `numThreadsBusy`.

Shouldn't we decrement right before we actually sleep (and start with it at one).

lib/common/pool.c
Cyan4973: Good point !
Yes, it might.

I'm surprised it passes the new test, because this case is supposed to happen.

terrelln: I think that condition is only possible when the pools is downsized.

lib/common/pool.c
Cyan4973: Actually, the comment is incorrect.

lib/common/pool.c
Cyan4973: Another strategy could be to have a local variable `isActive`
which knows if the thread was just active before or was awaken.
It would then be able to "fix" the real nb of active threads, or the decision to do `numThreads++`.

But this alternative design feels fragile, easier to break.

lib/common/pool.c
```diff
@@ -225,13 +225,16 @@ static POOL_ctx* POOL_resize_internal(POOL_ctx* ctx, size_t numThreads)
  *    note : starting context is considered consumed. */
 POOL_ctx* POOL_resize(POOL_ctx* ctx, size_t numThreads)
 {
-    POOL_ctx* newCtx;
     if (ctx==NULL) return NULL;
     ZSTD_pthread_mutex_lock(&ctx->queueMutex);
-    newCtx = POOL_resize_internal(ctx, numThreads);
+    {   POOL_ctx* const newCtx = POOL_resize_internal(ctx, numThreads);
+        if (newCtx!=ctx) {
+            POOL_free(ctx);
```
terrelln: Destroying a locked mutex is undefined behavior. http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_destroy.html

Cyan4973: I'll change the behavior then, and just use an error code return type.
`free` will be an external operation.

PR #1200, opened 2018-06-21
Add CCtx Param Controlling Dict Attachment Behavior

felixhandte:


Cyan4973:
Is there a scenario for which `ZSTD_p_forceAttachDict` is useful or necessary ?

felixhandte:
Necessary? No. This is after all an optimization on an optimization.

Useful? As discussed, the real negative component of the cost-benefit trade-off offered by attaching a dict is a cost per-search. The number of searches that will end up being performed is a product of the size of the input and its compressibility. I know that some (all?) of the strategies have conditions where they try to avoid spending a lot of time trying to compress totally incompressible data, but it's still the case that very highly compressible data will want to attach a dict at larger sizes than relatively incompressible data.

I'm also just thinking that these cut-offs were selected for server-class x86_64 hardware, and the trade-off is probably different on other platforms. But maybe that simply indicates a need for a better heuristic. For example, if the dictionary itself included an expected compression ratio or something, Zstd could make better decisions.

Cyan4973:
I believe I understand the additional flexibility this option offers.

I was mostly wondering if this is going to be useful, 
in the following sense : is there at least one identified use case where it's going to be used ? 

For example, is it going to be used by Managed Compression ?

felixhandte:
Yes, I plan to use this in Managed Compression for JSON (or equivalently high compression categories).

PR #1203, opened 2018-06-22
Install pkg-config file on non-Unix platforms

Hello71:
This is standard practice; pkg-config is fully usable on Windows too. MXE generates its own pkg-config file, which is incompatible with zstd's.

I believe this change is not copyrightable or patentable in any jurisdiction, but to the extent that it is, all such rights are released under the CC0 license.


Hello71:
Tor for example uses only pkg-config to locate zstd.

Cyan4973:
Thanks for proposed patch @Hello71 .

Minor detail : could you please direct your patch at the `dev` branch instead ?
Our process disallow direct commit to `master` branch.

Hello71:
I mean, it's the same patch, but OK, whatever.

Cyan4973:
Sure ! It's a trivial patch !
Note that the [CLA]( https://code.facebook.com/cla) need to be signed before a merge.

Alternatively, I can manually reproduce your patch if you prefer.

Hello71:
I mainly wish to avoid an agreement that applies to any future contributions. I don't think "manually reproducing" satisfies copyright or patent law, but I don't think any of this is copyrightable or patentable anyways.

If it helps, I agree to the CLA with respect to this present contribution (i.e. relating to installing pkg-config files on non-Unix platforms) but not with respect to future contributions. All license grants for this contribution are still perpetual and irrevocable though, as already released by my CC0 declaration.

build/cmake/lib/CMakeLists.txt
```diff
@@ -157,20 +157,18 @@ IF (ZSTD_BUILD_STATIC)
             OUTPUT_NAME ${STATIC_LIBRARY_BASE_NAME})
 ENDIF (ZSTD_BUILD_STATIC)
 
-IF (UNIX)
```
tonytheodore: A standard way is `IF (UNIX or MINGW)` to avoid installing for MSVC.

PR #1204, opened 2018-06-23
Error code on no forward progress

Cyan4973:
Streaming decoders, such as `ZSTD_decompressStream()` or `ZSTD_decompress_generic()`,
may end up making no forward progress,
(aka no byte read from input __and__ no byte written to output),
due to unusual parameters conditions,
such as providing an output buffer already full.

In such case, the caller may be caught in an infinite loop,
calling the streaming decompression function again and again,
without making any progress.

This version detects such situation, and generates an error instead :
`ZSTD_error_dstSize_tooSmall` when output buffer is full,
`ZSTD_error_srcSize_wrong` when input buffer is empty.

The detection tolerates a number of attempts before triggering an error,
controlled by `ZSTD_NO_FORWARD_PROGRESS_MAX` macro constant,
which is set to 16 by default, and can be re-defined at compilation time.
This behavior tolerates potentially existing implementations
where such cases happen sporadically, like once or twice,
which is not dangerous (only infinite loops are),
without generating an error, hence without breaking these implementations.

added: corresponding test case.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -41,6 +41,17 @@
 #endif
 
 
+/*!
+ *  NO_FORWARD_PROGRESS_MAX :
+ *  maximum allowed nb of calls to ZSTD_decompressStream() and ZSTD_decompress_generic()
+ *  without any forward progress
+ *  (defined as: no byte read from input, and no byte flushed to output)
+ *  before triggering an error.
+ */
+#ifndef ZSTD_NO_FORWARD_PROGRESS_MAX
+#  define ZSTD_NO_FORWARD_PROGRESS_MAX 16
+#endif
+
 /*-*******************************************************
 *  Dependencies
 *********************************************************/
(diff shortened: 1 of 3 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -2620,6 +2633,7 @@ size_t ZSTD_initDStream_usingDict(ZSTD_DStream* zds, const void* dict, size_t di
 {
     DEBUGLOG(4, "ZSTD_initDStream_usingDict");
     zds->streamStage = zdss_init;
+    zds->noForwardProgress = 0;
```
terrelln: Does `ZSTD_initDCtx_internal()` not get called during initialization?

Cyan4973: I don't think so. `initDCtx` is only invoked when creating a new context.

Re-using a context triggers a much lighter context init.
For `initDStream`, it's entirely contained in this rather short function.

PR #1205, opened 2018-06-25
[zstdmt] Fix jobsize bugs

terrelln:
* `ZSTDMT_serialState_reset()` should use `targetSectionSize`, not `jobSize` when sizing the seqstore.
  Add an assert that checks that we sized the seqstore using the right job size.
* `ZSTDMT_compressionJob()` should check if `rawSeqStore.seq == NULL`.
* `ZSTDMT_initCStream_internal()` should not modify `mtctx->params.jobSize`.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1007,6 +1007,21 @@ size_t ZSTDMT_setMTCtxParameter(ZSTDMT_CCtx* mtctx, ZSTDMT_parameter parameter, 
     }
 }
 
+size_t ZSTDMT_getMTCtxParameter(ZSTDMT_CCtx* mtctx, ZSTDMT_parameter parameter, unsigned* value)
+{
+    switch (parameter) {
+    case ZSTDMT_p_jobSize:
+        *value = mtctx->params.jobSize;
+        break;
+    case ZSTDMT_p_overlapSectionLog:
+        *value = mtctx->params.overlapSizeLog;
+        break;
+    default:
+        return ERROR(parameter_unsupported);
+    }
+    return 0;
+}
+
 /* Sets parameters relevant to the compression job,
  * initializing others to default values. */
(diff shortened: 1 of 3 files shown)


Cyan4973:
> `ZSTDMT_initCStream_internal()` should not modify `mtctx->params.jobSize`

Not sure to follow this one.
If I read it correctly, this patch still modifies `params.jobSize` ?
https://github.com/facebook/zstd/pull/1205/files#diff-92cfacb89db0f858d089c8021c329941R1292

terrelln:
We clamp it to `ZSTDMT_JOBSIZE_MIN` and `ZSTDMT_JOBSIZE_MAX`. When we set it using the advanced API, we already do this clamping.

lib/compress/zstdmt_compress.c
```diff
@@ -1321,7 +1328,9 @@ size_t ZSTDMT_initCStream_internal(
     mtctx->targetPrefixSize = (size_t)1 << ZSTDMT_computeOverlapLog(params);
     DEBUGLOG(4, "overlapLog=%u => %u KB", params.overlapSizeLog, (U32)(mtctx->targetPrefixSize>>10));
     mtctx->targetSectionSize = params.jobSize;
-    if (mtctx->targetSectionSize < ZSTDMT_JOBSIZE_MIN) mtctx->targetSectionSize = ZSTDMT_JOBSIZE_MIN;
+    if (mtctx->targetSectionSize == 0) {
```
terrelln: This is where we do the adjustment that is parameter-specific.

Cyan4973:
OK, what you mean is  : 
when `params.jobSize` is set to `0` to mean "auto", 
it keeps this value
(whereas previously it would be updated, and have a non-zero value afterwards).

`params.jobSize` may still be modified, but only if it's `!= 0`, in order to be clamped.

Cyan4973:
Does it sound possible to create a new test 
which would fail without this patch ?
Notably regarding the `params.jobSize` update issue ?

PR #1207, opened 2018-06-26
Correct multithread logic, fixing 'unsupported parameter' error

oleid:
The original conditions only worked, when both, static and shared variants where built, resulting in an inconsistency between programs and library. The program was built with MT support enabled, the library not. That lead to _error 11 'unsupported parameter'_ when compressing anything with the command line tool.

When changing the AND condition to `ZSTD_MULTITHREAD_SUPPORT AND (ZSTD_BUILD_SHARED OR ZSTD_BUILD_SHARED)`, cmake stopps complaining one of the targets wasn't built. This commit works for any case.


Cyan4973:
Looks good, thanks @oleid !

PR #1213, opened 2018-06-27
re-enable arm64 tests

Cyan4973:
arm64 tests have been failing recently
due to some strange bug in `ld` after travis CI updated `ld` version for arm64.

Just attempting to re-enable arm64 tests, see if they work correctly again


Cyan4973:
arm64 tests do work fine again.
So let's re-enable them.

PR #1214, opened 2018-06-27
Make --fast=0 fail

jennifermliu:


tests/playTests.sh
```diff
@@ -108,6 +108,7 @@ $ECHO "test : --fast aka negative compression levels"
 $ZSTD --fast -f tmp  # == -1
 $ZSTD --fast=3 -f tmp  # == -3
 $ZSTD --fast=200000 -f tmp  # == no compression
+! $ZSTD -c --fast=0  tmp # should fail
```
Cyan4973: minor detail : 
`zstd -c` will push the compressed result into `stdout`,
which, if it's not explicitly redirected, will be the console by default.

It's not a problem here, because the test is expected to fail before any compression takes place,
but should it _not_ fail, then the console will be filled with garbage binary values, some of them possibly corresponding to interpreted commands, and quite quickly, the console becomes an unusable mess.

recommandation : push the result into null (`INTOVOID`)

PR #1124, opened 2018-05-08
Write to /dev/random for test

terrelln:
Fixes #1116.

PR #1222, opened 2018-06-30
2 separate issues found when running zstd's tests on OpenBSD

bket:
1.) OpenBSD is unable to write to /dev/zero

#1124 fixes an issue with GNU/Hurd being unable to write to /dev/zero. Implemented fix is writing to /dev/random instead.

On OpenBSD a regular user is unable to write to /dev/random because of permissions set on this device. Result is failing a regression test.

Proposed solution should work for all platforms.

2.) xz/lzma warning causes test to fail
    
OpenBSD's port building infrastructure is able to build in a privilege separated mode. It uses a privilege drop model. Regression tests fail in this mode as xz/lzma is unable to set file group and errors out with:
    
```
xz: tmp.xz: Cannot set the file group: Operation not permitted
gmake[1]: *** [Makefile:307: zstd-playTests] Error 2
```

Actually it is not a xz/lzma error but a warning causing zstd's regression test to fail. Proposed fix is to have xz/lzma not set the exit status to 2 even if a condition worth a warning was detected (-Q flag).

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -548,16 +548,16 @@ $ZSTD --format=xz -V || LZMAMODE=0
 if [ $LZMAMODE -eq 1 ]; then
     $ECHO "xz support detected"
     XZEXE=1
-    xz -V && lzma -V || XZEXE=0
+    xz -Q -V && lzma -Q -V || XZEXE=0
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good ! Thanks @bket !

PR #1225, opened 2018-07-03
Split samples when building dictionary for COVER

jennifermliu:


lib/dictBuilder/cover.c
```diff
@@ -531,25 +538,50 @@ static void COVER_ctx_destroy(COVER_ctx_t *ctx) {
  */
 static int COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
                           const size_t *samplesSizes, unsigned nbSamples,
-                          unsigned d) {
+                          unsigned d, double splitPoint) {
   const BYTE *const samples = (const BYTE *)samplesBuffer;
-  const size_t totalSamplesSize = COVER_sum(samplesSizes, nbSamples);
+  const unsigned first = 0;
```
terrelln: nit: this is a compile time constant, call it `kFirst`.

lib/dictBuilder/cover.c
```diff
@@ -531,25 +538,50 @@ static void COVER_ctx_destroy(COVER_ctx_t *ctx) {
...
+  const size_t totalSamplesSize = COVER_sum(samplesSizes, first, nbSamples);
+  /* Split samples into testing and training sets */
+  double tmp = (double)nbSamples * splitPoint;
+  const unsigned nbTrainSamples = (unsigned)tmp;
```
terrelln:
```c
const unsigned nbTrainSamples = (unsigned)((double)nbSamples * splitPoint);
```

lib/dictBuilder/cover.c
```diff
@@ -531,25 +538,50 @@ static void COVER_ctx_destroy(COVER_ctx_t *ctx) {
...
     DISPLAYLEVEL(1, "Total samples size is too large (%u MB), maximum size is %u MB\n",
                  (U32)(totalSamplesSize>>20), (COVER_MAX_SAMPLES_SIZE >> 20));
     return 0;
   }
+  /* Check if there are at least 5 training samples */
+  if (nbTrainSamples < 5) {
+    DISPLAYLEVEL(1, "Total number of training samples is %u and is invalid.", nbTrainSamples);
+    return 0;
+  }
+  /* Check if there's testing sample when splitPoint is nonzero */
```
terrelln: This comment isn't quite right.

lib/dictBuilder/zdict.h
```diff
@@ -86,6 +86,7 @@ typedef struct {
     unsigned d;                  /* dmer size : constraint: 0 < d <= k : Reasonable range [6, 16] */
     unsigned steps;              /* Number of steps : Only used for optimization : 0 means default (32) : Higher means more parameters checked */
     unsigned nbThreads;          /* Number of threads : constraint: 0 < nbThreads : 1 means single-threaded : Only used for optimization : Ignored if ZSTD_MULTITHREAD is not defined */
+    double splitPoint;           /* Percentage of samples used for training: the first nbSamples * splitPoint samples will be used to training */
```
terrelln: What is the default value?

programs/zstdcli.c
```diff
@@ -277,15 +277,21 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
...
+        if (longCommandWArg(&stringPtr, "split=")) {
+          splitPercentage = readU32FromChar(&stringPtr);
+          params->splitPoint = (double)splitPercentage / 100.0;
+          if (stringPtr[0]==',') { stringPtr++; continue; } else break;
+        }
         return 0;
     }
     if (stringPtr[0] != 0) return 0;
-    DISPLAYLEVEL(4, "cover: k=%u\nd=%u\nsteps=%u\n", params->k, params->d, params->steps);
+    DISPLAYLEVEL(4, "cover: k=%u\nd=%u\nsteps=%u\nsplitPoint=%d\n", params->k, params->d, params->steps, splitPercentage);
```
terrelln: `split` not `splitPoint`.

lib/dictBuilder/zdict.h
jennifermliu: Right now the splitPoint is default to 0.8. I store it in a constant named DEFAULT_SPLITPOINT. I use this constant when I initialize the parameters because I have not yet decided on the "best" splitPoint and just save the default value as a constant so it's easier to change it later. Would you recommend having the actual value of 0.8 instead of the stored constant? It seems like the actual values are used when d and steps are initialized in ZDICT_cover_params_t defaultCoverParams(void)in zstdcli.c.

lib/dictBuilder/cover.c
```diff
@@ -531,25 +538,50 @@ static void COVER_ctx_destroy(COVER_ctx_t *ctx) {
...
+                          unsigned d, double splitPoint) {
   const BYTE *const samples = (const BYTE *)samplesBuffer;
-  const size_t totalSamplesSize = COVER_sum(samplesSizes, nbSamples);
+  const unsigned kFirst = 0;
+  const size_t totalSamplesSize = COVER_sum(samplesSizes, kFirst, nbSamples);
+  /* Split samples into testing and training sets */
+  const unsigned nbTrainSamples = splitPoint < 1.0 ? (unsigned)((double)nbSamples * splitPoint) : nbSamples;
+  const unsigned nbTestSamples = splitPoint < 1.0 ? nbSamples - nbTrainSamples : nbSamples;
+  const size_t trainingSamplesSize = splitPoint < 1.0 ? COVER_sum(samplesSizes, kFirst, nbTrainSamples) : totalSamplesSize;
+  const size_t testSamplesSize = splitPoint < 1.0 ? COVER_sum(samplesSizes, nbTrainSamples, nbSamples) : totalSamplesSize;
```
terrelln: nit: Don't add an extra parameter to `COVER_sum()`. Instead, do:

```
COVER_sum(sampleSizes + nbTrainSamples, nbTestSamples);
```

lib/dictBuilder/cover.c
```diff
@@ -531,25 +538,50 @@ static void COVER_ctx_destroy(COVER_ctx_t *ctx) {
...
+    DISPLAYLEVEL(1, "Total number of training samples is %u and is invalid.", nbTrainSamples);
+    return 0;
+  }
+  /* Check if there's testing sample */
+  if (nbTestSamples < 1) {
+    DISPLAYLEVEL(1, "Total number of testing samples is %u and is invalid.", nbTestSamples);
+    return 0;
+  }
+  /* Check if nbTrainSamples plus nbTestSamples add up to nbSamples when splitPoint is less than 1*/
+  if (nbTrainSamples + nbTestSamples != nbSamples && splitPoint < 1.0) {
```
terrelln: This can be an assert right? We use asserts when the test should never fail.

lib/dictBuilder/cover.c
```diff
@@ -958,6 +994,10 @@ ZDICTLIB_API size_t ZDICT_optimizeTrainFromBuffer_cover(
   POOL_ctx *pool = NULL;
 
   /* Checks */
+  if (splitPoint <= 0 || splitPoint > 1) {
```
terrelln: Lets only make `0.0` the default. Otherwise this check will never fail, and people might get unexpected results if they pass 99 for example.

lib/dictBuilder/zdict.h
terrelln: You should mention the default value "0.0 means default (0.8)" or whatever we decide the default is.

lib/dictBuilder/cover.c
jennifermliu: When checking if splitPoint is 0.0, is it ok that I use <=0.0 instead since the number from cli can't be negative? Or should I do the floating point comparison with epsilon to check if this is 0.0?

terrelln: You should use `== 0.0`, the API doesn't have the same restriction. We can use equal here, since we know we will be passing exactly `0.0`.

terrelln:
This is looking good. The last thing I want to see is if the split point helps for small amounts of training data. Then we can decide on the default value and merge.

terrelln:
nit: Explain how splitPoint=1.0/100 is special.

PR #1230, opened 2018-07-10
zstdcli: Allow -o before --train

### LINKED ISSUE #1227: -o has no effect when training with --train (v. 1.3.5)
zstd -o gt.zstd --train gt.sample10k/*
writes to file dictionary and not to gt.zstd

terrelln:
Only set the default value if `outFileName` is unset.

Fixes #1227.


Cyan4973:
Looks good to me !

Just some minor conflict to take care of.

Issue #1231, opened 2018-07-10
GENie/premake build


codecat:
I wrote a [GENie](https://github.com/bkaradzic/GENie)/premake build script for zstd to incorporate within our builds. Would it be worth adding to the `contrib` folder? Additionally, I can make a pull request.

https://gist.github.com/codecat/5ff66a9ab4fce068b6229bcd12f6d985

terrelln:
I'm not sure how popular that build system is, but if you think others might find it useful please submit a pull request to a new directory under `contrib`.

codecat:
GENie (a fork of Premake) is quite popular in game development, while premake itself is quite popular across the board. I will test this script tomorrow in actual premake as well and then do a pull request. Thanks!

PR #1234, opened 2018-07-11
Added premake4/GENie script to contrib folder

codecat:
This adds support for premake4 and [GENie](https://github.com/bkaradzic/GENie) as earlier mentioned in #1231.

* `zstd.lua`: contains the function for adding the project to a solution.
* `premake4.lua`: example build script which will build the zstd library.

Tested on Windows (GENie) & MacOS (GENie and premake4).

### DIFF
--- a/contrib/premake/zstd.lua
+++ b/contrib/premake/zstd.lua
@@ -1,0 +1,79 @@
+-- This GENie/premake file copies the behavior of the Makefile in the lib folder.
+-- Basic usage: project_zstd(ZSTD_DIR)
+
+function project_zstd(dir, compression, decompression, deprecated, dictbuilder, legacy)
+	if compression == nil then compression = true end
+	if decompression == nil then decompression = true end
+	if deprecated == nil then deprecated = false end
+	if dictbuilder == nil then dictbuilder = false end
+
+	if legacy == nil then legacy = 0 end
+
+	if compression then
+		dictbuilder = false
+		deprecated = false
+	end
+
+	if decompression then
+		legacy = 0
+		deprecated = false
+	end
+
+	project 'zstd'
+		kind 'StaticLib'
+		language 'C'
+
+		files {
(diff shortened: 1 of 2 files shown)


contrib/premake/zstd.lua
```diff
@@ -0,0 +1,79 @@
...
+function project_zstd(dir, compression, decompression, deprecated, dictbuilder, legacy)
+	if compression == nil then compression = true end
+	if decompression == nil then decompression = true end
+	if deprecated == nil then deprecated = false end
+	if dictbuilder == nil then dictbuilder = false end
+
+	if legacy == nil then legacy = 0 end
+
+	if compression then
+		dictbuilder = false
```
terrelln: Why turn `dictBuilder` and `deprecated` off?

contrib/premake/zstd.lua
```diff
@@ -0,0 +1,79 @@
...
+		deprecated = false
+	end
+
+	if decompression then
+		legacy = 0
```
terrelln: Same here

contrib/premake/zstd.lua
```diff
@@ -0,0 +1,79 @@
...
+		if deprecated then
+			files {
+				dir .. 'deprecated/**.c',
+				dir .. 'deprecated/**.h'
+			}
+		end
+
+		if legacy < 8 then
+			files {
+				dir .. 'legacy/zstd_v0' .. (legacy - 7) .. '.*'
```
terrelln: When legacy == 4, we want to support >= 4, for example. Except 0 means no legacy support.

terrelln:
Thanks for the PR @codecat, just have some small comments then it will be good to merge.

contrib/premake/zstd.lua
codecat: Ugh, nevermind, I think I was tired writing this. I got these wrong too, you're right.

PR #1236, opened 2018-07-13
ParamgrillConstraints

GeorgeLu97:
Adds additional Functionality to paramgrill
-Accepts decompression speed and compression memory as constraints
-Accepts fixed parameters as constraints (ex: windowLog=8)
-Accepts dictionaries and multiple files
-Improves convergence speed

Testing:
Ran with various constraint configurations and files.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -294,7 +294,7 @@ BMK_customReturn_t BMK_benchFunction(
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
     const void* const * const srcBlockBuffers, const size_t* srcBlockSizes,
-    void* const * const dstBlockBuffers, const size_t* dstBlockCapacities, size_t* cSizes, 
+    void* const * const dstBlockBuffers, const size_t* dstBlockCapacities, size_t* blockResult, 
     unsigned nbLoops) {
     size_t dstSize = 0;
     U64 totalTime;
@@ -332,8 +332,8 @@ BMK_customReturn_t BMK_benchFunction(
                         j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
                 } else if(i == nbLoops - 1) {
                     dstSize += res;
-                    if(cSizes != NULL) {
-                        cSizes[j] = res;
+                    if(blockResult != NULL) {
+                        blockResult[j] = res;
                     }
                 } 
             }
@@ -358,6 +358,9 @@ void BMK_resetTimeState(BMK_timedFnState_t* r, unsigned nbSeconds) {
 
 BMK_timedFnState_t* BMK_createTimeState(unsigned nbSeconds) {
     BMK_timedFnState_t* r = (BMK_timedFnState_t*)malloc(sizeof(struct BMK_timeState_t));
+    if(r == NULL) {
+        return r;
+    }
     BMK_resetTimeState(r, nbSeconds);
     return r;
 }
@@ -373,7 +376,7 @@ BMK_customTimedReturn_t BMK_benchFunctionTimed(
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
     const void* const* const srcBlockBuffers, const size_t* srcBlockSizes,
-    void * const * const dstBlockBuffers, const size_t * dstBlockCapacities, size_t* dstSizes) 
+    void * const * const dstBlockBuffers, const size_t * dstBlockCapacities, size_t* blockResults) 
 {
     U64 fastest = cont->fastestTime;
     int completed = 0;
@@ -390,7 +393,7 @@ BMK_customTimedReturn_t BMK_benchFunctionTimed(
         }
         /* reinitialize capacity */
         r.result = BMK_benchFunction(benchFn, benchPayload, initFn, initPayload,
-        blockCount, srcBlockBuffers, srcBlockSizes, dstBlockBuffers, dstBlockCapacities, dstSizes, cont->nbLoops);
+        blockCount, srcBlockBuffers, srcBlockSizes, dstBlockBuffers, dstBlockCapacities, blockResults, cont->nbLoops);
         if(r.result.error) { /* completed w/ error */
             r.completed = 1;
             return r;
@@ -718,38 +721,36 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
     U32 const maxNbBlocks = (U32) ((srcSize + (blockSize-1)) / blockSize) + nbFiles;
 
     /* these are the blockTable parameters, just split up */
-    const void ** const srcPtrs = (const void** const)malloc(maxNbBlocks * sizeof(void*));
-    size_t* const srcSizes = (size_t* const)malloc(maxNbBlocks * sizeof(size_t));
+    const void ** const srcPtrs = (const void**)malloc(maxNbBlocks * sizeof(void*));
+    size_t* const srcSizes = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
 
 
-    void ** const cPtrs = (void** const)malloc(maxNbBlocks * sizeof(void*));
-    size_t* const cSizes = (size_t* const)malloc(maxNbBlocks * sizeof(size_t));
-    size_t* const cCapacities = (size_t* const)malloc(maxNbBlocks * sizeof(size_t));
+    void ** const cPtrs = (void**)malloc(maxNbBlocks * sizeof(void*));
+    size_t* const cSizes = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
+    size_t* const cCapacities = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
 
-    void ** const resPtrs = (void** const)malloc(maxNbBlocks * sizeof(void*));
-    size_t* const resSizes = (size_t* const)malloc(maxNbBlocks * sizeof(size_t));
+    void ** const resPtrs = (void**)malloc(maxNbBlocks * sizeof(void*));
+    size_t* const resSizes = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
 
     BMK_timedFnState_t* timeStateCompress = BMK_createTimeState(adv->nbSeconds);
     BMK_timedFnState_t* timeStateDecompress = BMK_createTimeState(adv->nbSeconds);
 
-    void* compressedBuffer;
     const size_t maxCompressedSize = dstCapacity ? dstCapacity : ZSTD_compressBound(srcSize) + (maxNbBlocks * 1024);
+
+    void* const internalDstBuffer = dstBuffer ? NULL : malloc(maxCompressedSize);
+    void* const compressedBuffer = dstBuffer ? dstBuffer : internalDstBuffer;
+
     void* resultBuffer = malloc(srcSize);
 
-
     BMK_return_t results = { { 0, 0, 0, 0 }, 0 };
-    int allocationincomplete;
+    int allocationincomplete = !srcPtrs || !srcSizes || !cPtrs || 
+        !cSizes || !cCapacities || !resPtrs || !resSizes || 
+        !timeStateCompress || !timeStateDecompress || !compressedBuffer || !resultBuffer;
 
-    if(!dstCapacity) {
-        compressedBuffer = malloc(maxCompressedSize);
-    } else {
-        compressedBuffer = dstBuffer;
-    }
+    int parametersConflict = !dstBuffer ^ !dstCapacity;
 
-    allocationincomplete = !compressedBuffer || !resultBuffer ||  
-        !srcPtrs || !srcSizes || !cPtrs || !cSizes || !resPtrs || !resSizes;
 
-    if (!allocationincomplete) {
+    if (!allocationincomplete && !parametersConflict) {
         results = BMK_benchMemAdvancedNoAlloc(srcPtrs, srcSizes, cPtrs, cCapacities, cSizes,
             resPtrs, resSizes, &resultBuffer, compressedBuffer, maxCompressedSize, timeStateCompress, timeStateDecompress,
             srcBuffer, srcSize, fileSizes, nbFiles, cLevel, comprParams,
@@ -759,9 +760,8 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
     /* clean up */
     BMK_freeTimeState(timeStateCompress);
     BMK_freeTimeState(timeStateDecompress);
-    if(!dstCapacity) { /* only free if not given */
-        free(compressedBuffer);
-    }
+    
+    free(internalDstBuffer);
     free(resultBuffer);
 
     free((void*)srcPtrs); 
@@ -774,6 +774,10 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
 
     if(allocationincomplete) {
         EXM_THROW(31, BMK_return_t, "allocation error : not enough memory");
+    }
+    
+    if(parametersConflict) {
+        EXM_THROW(32, BMK_return_t, "Conflicting input results");   
     }
     return results;
 }
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -188,7 +188,7 @@ static void BMK_initCCtx(ZSTD_CCtx* ctx,
     ZSTD_CCtx_setParameter(ctx, ZSTD_p_searchLog, comprParams->searchLog);
     ZSTD_CCtx_setParameter(ctx, ZSTD_p_minMatch, comprParams->searchLength);
     ZSTD_CCtx_setParameter(ctx, ZSTD_p_targetLength, comprParams->targetLength);
-    ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionStrategy , comprParams->strategy);
+    ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionStrategy, comprParams->strategy);
     ZSTD_CCtx_loadDictionary(ctx, dictBuffer, dictBufferSize);
 }
 
@@ -281,8 +281,6 @@ static size_t local_defaultDecompress(
 
 }
 
-volatile char g_touched;
-
 /* initFn will be measured once, bench fn will be measured x times */
 /* benchFn should return error value or out Size */
 /* takes # of blocks and list of size & stuff for each. */
@@ -293,7 +291,7 @@ BMK_customReturn_t BMK_benchFunction(
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
     const void* const * const srcBlockBuffers, const size_t* srcBlockSizes,
-    void** const dstBlockBuffers, size_t* dstBlockCapacities,
+    void* const * const dstBlockBuffers, size_t* dstBlockCapacitiesToSizes,
     unsigned nbLoops) {
     size_t srcSize = 0, dstSize = 0, ind = 0;
     U64 totalTime;
@@ -310,20 +308,13 @@ BMK_customReturn_t BMK_benchFunction(
     }
 
     {
-
-        unsigned i, j;
+        size_t i;
         for(i = 0; i < blockCount; i++) {
-            for(j = 0; j < srcBlockSizes[i]; j++) {
-                g_touched = ((const char*)srcBlockBuffers[i])[j]; /* touch */
-            }
-        }
-        for(i = 0; i < blockCount; i++) {
-            memset(dstBlockBuffers[i], 0xE5, dstBlockCapacities[i]);  /* warm up and erase result buffer */
+            memset(dstBlockBuffers[i], 0xE5, dstBlockCapacitiesToSizes[i]);  /* warm up and erase result buffer */
         }
 
         //UTIL_sleepMilli(5);  /* give processor time to other processes */
         //UTIL_waitForNextTick();
-
     }
 
     {      
@@ -332,17 +323,13 @@ BMK_customReturn_t BMK_benchFunction(
         if(initFn != NULL) { initFn(initPayload); }
         for(i = 0; i < nbLoops; i++) {
             for(j = 0; j < blockCount; j++) {
-                size_t res = benchFn(srcBlockBuffers[j], srcBlockSizes[j], dstBlockBuffers[j], dstBlockCapacities[j], benchPayload);
+                size_t res = benchFn(srcBlockBuffers[j], srcBlockSizes[j], dstBlockBuffers[j], dstBlockCapacitiesToSizes[j], benchPayload);
                 if(ZSTD_isError(res)) {
                     EXM_THROW_ND(2, BMK_customReturn_t, "Function benchmarking failed on block %u of size %u : %s  \n",
-                        j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
+                        j, (U32)dstBlockCapacitiesToSizes[j], ZSTD_getErrorName(res));
                 }  else if(firstIter) {
                     dstSize += res;
-                    //Make compressed blocks continuous
-                    if(j != blockCount - 1) {
-                        dstBlockBuffers[j+1] = (void*)((char*)dstBlockBuffers[j] + res);
-                        dstBlockCapacities[j] = res;
-                    }
+                    dstBlockCapacitiesToSizes[j] = res;
                 } 
             }
             firstIter = 0;
@@ -382,7 +369,7 @@ BMK_customTimedReturn_t BMK_benchFunctionTimed(
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
     const void* const* const srcBlockBuffers, const size_t* srcBlockSizes,
-    void** const dstBlockBuffers, size_t* dstBlockCapacities) 
+    void * const * const dstBlockBuffers, size_t * dstBlockCapacitiesToSizes) 
 {
     U64 fastest = cont->fastestTime;
     int completed = 0;
@@ -399,7 +386,7 @@ BMK_customTimedReturn_t BMK_benchFunctionTimed(
         }
 
         r.result = BMK_benchFunction(benchFn, benchPayload, initFn, initPayload,
-        blockCount, srcBlockBuffers, srcBlockSizes, dstBlockBuffers, dstBlockCapacities, cont->nbLoops);
+        blockCount, srcBlockBuffers, srcBlockSizes, dstBlockBuffers, dstBlockCapacitiesToSizes, cont->nbLoops);
         if(r.result.error) { /* completed w/ error */
             r.completed = 1;
             return r;
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -85,7 +85,7 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 *  Exceptions
 ***************************************/
 #ifndef DEBUG
-#  define DEBUG 1
+#  define DEBUG 0
 #endif
 #define DEBUGOUTPUT(...) { if (DEBUG) DISPLAY(__VA_ARGS__); }
 
@@ -188,7 +188,7 @@ static void BMK_initCCtx(ZSTD_CCtx* ctx,
     ZSTD_CCtx_setParameter(ctx, ZSTD_p_searchLog, comprParams->searchLog);
     ZSTD_CCtx_setParameter(ctx, ZSTD_p_minMatch, comprParams->searchLength);
     ZSTD_CCtx_setParameter(ctx, ZSTD_p_targetLength, comprParams->targetLength);
-    ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionStrategy, comprParams->strategy);
+    ZSTD_CCtx_setParameter(ctx, ZSTD_p_compressionStrategy , comprParams->strategy);
     ZSTD_CCtx_loadDictionary(ctx, dictBuffer, dictBufferSize);
 }
 
@@ -293,7 +293,7 @@ BMK_customReturn_t BMK_benchFunction(
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
     const void* const * const srcBlockBuffers, const size_t* srcBlockSizes,
-    void* const * const dstBlockBuffers, const size_t* dstBlockCapacities,
+    void** const dstBlockBuffers, size_t* dstBlockCapacities,
     unsigned nbLoops) {
     size_t srcSize = 0, dstSize = 0, ind = 0;
     U64 totalTime;
@@ -338,6 +338,11 @@ BMK_customReturn_t BMK_benchFunction(
                         j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
                 }  else if(firstIter) {
                     dstSize += res;
+                    //Make compressed blocks continuous
+                    if(j != blockCount - 1) {
+                        dstBlockBuffers[j+1] = (void*)((char*)dstBlockBuffers[j] + res);
+                        dstBlockCapacities[j] = res;
+                    }
                 } 
             }
             firstIter = 0;
@@ -370,13 +375,14 @@ void BMK_freeTimeState(BMK_timedFnState_t* state) {
     free(state);
 }
 
+/* make option for dstBlocks to be */
 BMK_customTimedReturn_t BMK_benchFunctionTimed(
     BMK_timedFnState_t* cont,
     BMK_benchFn_t benchFn, void* benchPayload,
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
     const void* const* const srcBlockBuffers, const size_t* srcBlockSizes,
-    void* const* const dstBlockBuffers, const size_t* dstBlockCapacities) 
+    void** const dstBlockBuffers, size_t* dstBlockCapacities) 
 {
     U64 fastest = cont->fastestTime;
     int completed = 0;
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -441,7 +441,7 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
     int displayLevel, const char* displayName, const BMK_advancedParams_t* adv) 
 {
     size_t const blockSize = ((adv->blockSize>=32 && (adv->mode != BMK_decodeOnly)) ? adv->blockSize : srcSize) + (!srcSize); /* avoid div by 0 */
-    BMK_return_t results = { { 0, 0., 0., 0 }, 0 } ;
+    BMK_return_t results = { { 0, 0, 0, 0 }, 0 } ;
     size_t const loadedCompressedSize = srcSize;
     size_t cSize = 0;
     double ratio = 0.;
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -692,6 +692,7 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
     }
     DISPLAYLEVEL(2, "%2i#\n", cLevel);
     }   /* Bench */
+    results.result.cMem = ZSTD_sizeof_CCtx(ctx);
     return results;
 }
 
@@ -731,6 +732,7 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
             srcBuffer, srcSize, fileSizes, nbFiles, cLevel, comprParams,
             dictBuffer, dictBufferSize, ctx, dctx, displayLevel, displayName, adv);
     }
+    
     /* clean up */
     BMK_freeTimeState(timeStateCompress);
     BMK_freeTimeState(timeStateDecompress);
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -85,7 +85,7 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 *  Exceptions
 ***************************************/
 #ifndef DEBUG
-#  define DEBUG 0
+#  define DEBUG 1
 #endif
 #define DEBUGOUTPUT(...) { if (DEBUG) DISPLAY(__VA_ARGS__); }
 
@@ -281,12 +281,14 @@ static size_t local_defaultDecompress(
 
 }
 
+volatile char g_touched;
+
 /* initFn will be measured once, bench fn will be measured x times */
 /* benchFn should return error value or out Size */
 /* takes # of blocks and list of size & stuff for each. */
 /* only does looping */
 /* note time per loop could be zero if interval too short */
-BMK_customReturn_t BMK_benchFunction(
+BMK_customReturn_t __attribute__((optimize("O0"))) BMK_benchFunction(
     BMK_benchFn_t benchFn, void* benchPayload,
     BMK_initFn_t initFn, void* initPayload,
     size_t blockCount,
@@ -299,22 +301,28 @@ BMK_customReturn_t BMK_benchFunction(
     BMK_customReturn_t retval;
     UTIL_time_t clockStart;
 
-    {
-        unsigned i;
-        for(i = 0; i < blockCount; i++) {
-            memset(dstBlockBuffers[i], 0xE5, dstBlockCapacities[i]);  /* warm up and erase result buffer */
-        }
-
-        UTIL_sleepMilli(5);  /* give processor time to other processes */
-        UTIL_waitForNextTick();
-    }
-
     if(!nbLoops) {
         EXM_THROW_ND(1, BMK_customReturn_t, "nbLoops must be nonzero \n");
     }
 
     for(ind = 0; ind < blockCount; ind++) {
         srcSize += srcBlockSizes[ind];
+    }
+
+    {
+        unsigned i, j;
+        for(i = 0; i < blockCount; i++) {
+            for(j = 0; j < srcBlockSizes[i]; j++) {
+                g_touched = ((const char*)srcBlockBuffers[i])[j]; /* touch */
+            }
+        }
+        for(i = 0; i < blockCount; i++) {
+            memset(dstBlockBuffers[i], 0xE5, dstBlockCapacities[i]);  /* warm up and erase result buffer */
+            //this is written at end proc? where did compressed data get overwritten ny this?
+        }
+
+        UTIL_sleepMilli(5);  /* give processor time to other processes */
+        UTIL_waitForNextTick();
     }
 
     {      
@@ -327,9 +335,9 @@ BMK_customReturn_t BMK_benchFunction(
                 if(ZSTD_isError(res)) {
                     EXM_THROW_ND(2, BMK_customReturn_t, "Function benchmarking failed on block %u of size %u : %s  \n",
                         j, (U32)dstBlockCapacities[j], ZSTD_getErrorName(res));
-                } else if(firstIter) {
+                }  else if(firstIter) {
                     dstSize += res;
-                }
+                } 
             }
             firstIter = 0;
         }
@@ -393,7 +401,7 @@ BMK_customTimedReturn_t BMK_benchFunctionTimed(
         {   U64 const loopDuration = r.result.result.nanoSecPerRun * cont->nbLoops;
             r.completed = (cont->timeRemaining <= loopDuration);
             cont->timeRemaining -= loopDuration;
-            if (loopDuration > 0) { 
+            if (loopDuration > (TIMELOOP_NANOSEC / 100)) { 
                 fastest = MIN(fastest, r.result.result.nanoSecPerRun);
                 if(loopDuration >= MINUSABLETIME) {
                     r.result.result.nanoSecPerRun = fastest;
@@ -420,7 +428,7 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
     const void ** const srcPtrs, size_t* const srcSizes,
     void** const cPtrs, size_t* const cSizes,
     void** const resPtrs, size_t* const resSizes,
-    void* resultBuffer, void* compressedBuffer,
+    void** resultBufferPtr, void* compressedBuffer,
     const size_t maxCompressedSize,
     BMK_timedFnState_t* timeStateCompress, BMK_timedFnState_t* timeStateDecompress,
 
@@ -432,7 +440,7 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
     int displayLevel, const char* displayName, const BMK_advancedParams_t* adv) 
 {
     size_t const blockSize = ((adv->blockSize>=32 && (adv->mode != BMK_decodeOnly)) ? adv->blockSize : srcSize) + (!srcSize); /* avoid div by 0 */
-    BMK_return_t results;
+    BMK_return_t results = { { 0, 0., 0., 0 }, 0 } ;
     size_t const loadedCompressedSize = srcSize;
     size_t cSize = 0;
     double ratio = 0.;
@@ -454,13 +462,14 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
             srcPtr += fileSizes[fileNb];
         }
         {   size_t const decodedSize = (size_t)totalDSize64;
-            free(resultBuffer);
-            resultBuffer = malloc(decodedSize);
-            if (!resultBuffer) { 
+            free(*resultBufferPtr);
+            //TODO: decodedSize corrupted? 
+            *resultBufferPtr = malloc(decodedSize);
+            if (!(*resultBufferPtr)) {
                 EXM_THROW(33, BMK_return_t, "not enough memory"); 
             }
             if (totalDSize64 > decodedSize) {
-                free(resultBuffer); 
+                free(*resultBufferPtr); 
                 EXM_THROW(32, BMK_return_t, "original size is too large");   /* size_t overflow */
             }
             cSize = srcSize;
@@ -472,7 +481,7 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
     /* Init data blocks  */
     {   const char* srcPtr = (const char*)srcBuffer;
         char* cPtr = (char*)compressedBuffer;
-        char* resPtr = (char*)resultBuffer;
+        char* resPtr = (char*)(*resultBufferPtr);
         U32 fileNb;
         for (nbBlocks=0, fileNb=0; fileNb<nbFiles; fileNb++) {
             size_t remaining = fileSizes[fileNb];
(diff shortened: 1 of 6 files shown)


programs/bench.h
```diff
@@ -121,6 +122,8 @@ BMK_return_t BMK_syntheticTest(int cLevel, double compressibility,
  * (cLevel, comprParams + adv in advanced Mode) */
 /* srcBuffer - data source, expected to be valid compressed data if in Decode Only Mode
  * srcSize - size of data in srcBuffer
+ * dstBuffer - destination buffer to write compressed output in, optional (NULL)
```
terrelln: What do you mean by (NULL). Also, it seems like this documentation is added to the wrong doc block.

programs/bench.h
```diff
@@ -143,6 +146,7 @@ BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
 
 /* See benchMem for normal parameter uses and return, see advancedParams_t for adv */
 BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
+                        void* dstBuffer, size_t dstCapacity,
```
terrelln: Why are we adding `dstBuffer`?

tests/paramgrill.c
```diff
@@ -340,76 +541,273 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    size_t arrayLen = 1;
+    int i;
+    for(i = 0; i < varyLen; i++) {
+        arrayLen *= rangetable[varyParams[i]];
+    }
+    return arrayLen;
 }
 
+//sort of ~lg2 (replace 1024 w/ 999, and add 0 at lower end of range) for memoTableInd Tlen
+static unsigned lg2(unsigned x) {
```
terrelln: Check out `ZSTD_highBit32()` function.

tests/paramgrill.c
```diff
@@ -340,76 +541,273 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+            case SLOG_IND: ptr->searchLog    = ind % SLOG_RANGE + ZSTD_SEARCHLOG_MIN;                             ind /= SLOG_RANGE; break;
+            case SLEN_IND: ptr->searchLength = ind % SLEN_RANGE + ZSTD_SEARCHLENGTH_MIN;                          ind /= SLEN_RANGE; break;
+            case TLEN_IND: ptr->targetLength = (ind % TLEN_RANGE) ? MIN(1 << ((ind % TLEN_RANGE) - 1), 999) : 0;  ind /= TLEN_RANGE; break;
+        }
+    }
+}
+
+
+/* Initialize memotable, immediately mark redundant / obviously infeasible params as */
+static void memoTableInit(U8* memoTable, ZSTD_compressionParameters paramConstraints, const constraint_t target, const U32* varyParams, const int varyLen, const size_t srcSize) {
```
terrelln: I want to know what % of hits the memo table has. We should only keep it if the hit % is high enough.

GeorgeLu97: It's about 15%

programs/bench.c
```diff
@@ -317,19 +307,30 @@ BMK_customReturn_t BMK_benchFunction(
         srcSize += srcBlockSizes[ind];
     }
 
+    {
+        size_t i;
+        for(i = 0; i < blockCount; i++) {
+            memset(dstBlockBuffers[i], 0xE5, dstBlockCapacitiesToSizes[i]);  /* warm up and erase result buffer */
+        }
+
+        //UTIL_sleepMilli(5);  /* give processor time to other processes */
```
Cyan4973: explain why you believe it's better to disable these lines,
prefer using preprocessor directives, such as `#if 0`

tests/paramgrill.c
```diff
@@ -27,7 +27,8 @@
 #include "xxhash.h"
 #include "util.h"
 #include "bench.h"
-
+#include "zstd_errors.h"
+#include "zstd_internal.h"
 
 /*-************************************
```
Cyan4973: `paramgrill.c` rights seem to have unduly changed from `644` to `755` (executable).
Should be fixed, to remain `644`.

tests/paramgrill.c
```diff
@@ -56,17 +54,57 @@ static const int g_maxNbVariations = 64;
 *  Macros
 **************************************/
 #define DISPLAY(...)  fprintf(stderr, __VA_ARGS__)
+#define TIMED 0
+#ifndef DEBUG
+#  define DEBUG 1
```
Cyan4973: is it intended to have `DEBUGOUTPUT` enabled by default ?

tests/paramgrill.c
```diff
@@ -56,17 +54,57 @@ static const int g_maxNbVariations = 64;
...
+#define DEBUGOUTPUT(...) { if (DEBUG) DISPLAY(__VA_ARGS__); }
 
 #undef MIN
 #undef MAX
 #define MIN(a,b)   ( (a) < (b) ? (a) : (b) )
 #define MAX(a,b)   ( (a) > (b) ? (a) : (b) )
 #define CUSTOM_LEVEL 99
 
+/* indices for each of the variables */
+#define WLOG_IND 0
```
Cyan4973: Another possibility for such usage is to use `enum` instead.
This makes it possible to create a dedicated type, which tends to help later on, as the compiler can distinguish type errors.

tests/paramgrill.c
```diff
@@ -56,17 +54,57 @@ static const int g_maxNbVariations = 64;
...
+#define CLOG_IND 1
+#define HLOG_IND 2
+#define SLOG_IND 3
+#define SLEN_IND 4
+#define TLEN_IND 5
+//#define STRT_IND 6
+//#define NUM_PARAMS 7
+#define NUM_PARAMS 6
+//just don't use strategy as a param.
```
Cyan4973: 👍

Cyan4973:
> -Accepts decompression speed and compression ratio as constraints

In general, `paramgrill` is expected to find the best possible compression ratio
given a set of constraints, such as speed and memory.

How does that work then if compression ratio itself is a constraint ?
What else is being optimized for a given compression ratio ?
Is there a scenario where that makes sense ?

tests/paramgrill.c
```diff
@@ -140,6 +179,126 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
     return result;
 }
 
+//assume that clock can at least measure .01 second intervals?
```
Cyan4973: I believe it's reasonable to assume that `clock_t` offers a resolution of at least 10ms, if not better. And `clock_t` is the "backup solution", `util.h` use better libraries when available.

tests/paramgrill.c
```diff
@@ -140,6 +179,126 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
...
+}
+
+static void BMK_translateAdvancedParams(const ZSTD_compressionParameters params)
+{
+    DISPLAY("--zstd=windowLog=%u,chainLog=%u,hashLog=%u,searchLog=%u,searchLength=%u,targetLength=%u,strategy=%u \n",
+             params.windowLog, params.chainLog, params.hashLog, params.searchLog, params.searchLength, params.targetLength, (U32)(params.strategy));
+}
+
+/* checks results are feasible */
+static int feasible(const BMK_result_t results, const constraint_t target) {
```
Cyan4973: `feasible` is not specific enough, hence confusing.
I believe what you are checking here is if `result` respects `constraints`.
Use a name that better reflects the intent.

tests/paramgrill.c
```diff
@@ -140,6 +179,126 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
...
+/* checks exact equivalence to 0, to stop compiler complaining fpeq */
+static int eqZero(const double c1) {
+    const double z1 = 0.0;
+    const double z2 = -0.0;
+    return !(memcmp(&c1, &z1, sizeof(double))) || !(memcmp(&c1, &z2, sizeof(double)));
+}
+
+/* returns 1 if result2 is strictly 'better' than result1 */
+/* strict comparison / cutoff based */
+static int objective_lt(const BMK_result_t result1, const BMK_result_t result2) {
```
Cyan4973: Here also, I'm surprised by the name of the function.
The associated comment is clear enough though.
Try to find a name that reflects the intent.

tests/paramgrill.c
```diff
@@ -173,26 +344,406 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+                        void** dstPtrs, size_t* dstCapacityToSizes, U32 const nbBlocks, 
+                        const int cLevel, const ZSTD_compressionParameters* comprParams,
+                        const void* dictBuffer, const size_t dictBufferSize,
+                        ZSTD_CCtx* ctx, ZSTD_DCtx* dctx,
+                        const BMK_mode_t mode, const BMK_loopMode_t loopMode, const unsigned nbSeconds) {
+    U32 i;
+    BMK_return_t results = { { 0, 0., 0., 0 }, 0 } ;
+    size_t srcSize = 0;
+    void** const resPtrs = malloc(sizeof(void*) * nbBlocks); /* only really needed in both mode. */
+    size_t* const resSizes = malloc(sizeof(size_t) * nbBlocks);
```
Cyan4973: Important : check if `malloc()` was successful before using either `resSizes` or `resPtrs`.

tests/paramgrill.c
```diff
@@ -173,26 +344,406 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+    int freeDST = 0;
+
+    BMK_advancedParams_t adv = BMK_initAdvancedParams();
+    adv.mode = mode;
+    adv.loopMode = loopMode;
+    adv.nbSeconds = nbSeconds;
+
+    /* resSizes == srcSizes, but modifiable */
+    memcpy(resSizes, srcSizes, sizeof(size_t) * nbBlocks);
```
Cyan4973: `resSizes` is filled before checking if it's `NULL` (verified later).

tests/paramgrill.c
```diff
@@ -173,26 +344,406 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+    if(dstPtrs == NULL) {
+        freeDST = 1;
+        dstPtrs = malloc(nbBlocks * sizeof(void*));
+        dstCapacityToSizes = malloc(nbBlocks * sizeof(size_t));
+        if(dstPtrs == NULL) {
+            results.error = 33;
+            DISPLAY("error: allocation failed\n");
+            free(resPtrs);
+            free(resSizes);
+            return results;
```
Cyan4973: `dstCapacityToSizes` was allocated just before,
and is not freed there, 
so it becomes a memory leak.

tests/paramgrill.c
```diff
@@ -173,26 +344,406 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+/*-*******************************************************
+*  From Paramgrill End
+*********************************************************/
+
+/* Replicate function of benchMemAdvanced, but with pre-split src / dst buffers, with relevant info to invert it (compressedSizes) passed out. */
+/*BMK_benchMemAdvanced(srcBuffer,srcSize, dstBuffer, dstSize, fileSizes, nbFiles, 0, &cParams, dictBuffer, dictSize, ctx, dctx, 0, "File", &adv); */
+/* nbSeconds used in same way as in BMK_advancedParams_t, as nbIters when in iterMode */
+
+/* if in decodeOnly, then srcPtr's will be compressed blocks, and uncompressedBlocks will be written to dstPtrs? */
+/* dictionary nullable, nothing else though. */
```
Cyan4973: General comment : 
this function seems to have a lot of exit points.
This becomes a problem to properly manage memory, as it becomes too complex to track what has been allocated and what must be freed at each exit point.

In such case, divide the function in 2 nested parts.
The first one is only about allocating and freeing, and never early exits.
The second one does the real job, and can have as many exits as necessary.

tests/paramgrill.c
```diff
@@ -173,26 +344,406 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+        }
+    }
+
+    /* warmimg up memory */
+    for(i = 0; i < nbBlocks; i++) {
+        RDG_genBuffer(dstPtrs[i], dstCapacityToSizes[i], 0.10, 0.50, 1);
+    }
+
+    /* Bench */
+    {
```
Cyan4973: why double-nesting ?

tests/paramgrill.c
```diff
@@ -173,26 +344,406 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+
+        if(TIMED) { fprintf(f, " - %1lu:%2lu:%05.2f", (unsigned long) minutes / 60,(unsigned long) minutes % 60, (double)(time - minutes * TIMELOOP_NANOSEC * 60ULL)/TIMELOOP_NANOSEC); }
+        fprintf(f, "\n");
+        if(objective_lt(g_winner.result, result) && feasible(result, g_targetConstraints)) {
+            BMK_translateAdvancedParams(params);
+            g_winner.result = result;
+            g_winner.params = params;
+        }
+    }  
+    //else {
```
Cyan4973: prefer disabling code behind `#if 0` if you want to keep it around.

tests/paramgrill.c
```diff
@@ -340,76 +891,264 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
 /* nullified useless params, to ensure count stats */
-static ZSTD_compressionParameters* sanitizeParams(ZSTD_compressionParameters params)
+/* no point in windowLog < chainLog (no point 2x chainLog for bt) */
+/* now with built in bounds-checking */
+/* no longer does anything with sanitizeVarArray + clampcheck */
+static ZSTD_compressionParameters sanitizeParams(ZSTD_compressionParameters params)
 {
-    g_params = params;
     if (params.strategy == ZSTD_fast)
         g_params.chainLog = 0, g_params.searchLog = 0;
```
Cyan4973: since you return `params` at the end of the function,
you should modify `params` instead of `g_params`.

Also : if `g_params` is no longer useful, you can remove it entirely.

tests/paramgrill.c
```diff
@@ -340,76 +891,264 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    if (params.strategy != ZSTD_btopt && params.strategy != ZSTD_btultra && params.strategy != ZSTD_fast)
         g_params.targetLength = 0;
-    return &g_params;
+
+    return params;
 }
 
+/* new length */
+/* keep old array, will need if iter over strategy. */
+static int sanitizeVarArray(const int varLength, const U32* varArray, U32* varNew, const ZSTD_strategy strat) {
```
Cyan4973: minor : 
interface convention : 
prefer presenting the modified state (here `varNew`) at the beginning of parameters list.

tests/paramgrill.c
```diff
@@ -340,76 +891,264 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    }
+    if(!paramConstraints.targetLength) {
+        res[j] = TLEN_IND;
+        j++;
+    }
+    return j;
+}
+
+/* amt will probably always be \pm 1? */
+/* slight change from old paramVariation, targetLength can only take on powers of 2 now (999 ~= 1024?) */
```
Cyan4973: issue is, experience showed we need intermediate `targetLength` values (non power of 2) to properly scale compression levels.
You might consider a kind of "fixed point" representation, with just 1 or 2 bits representing fractional, so that it becomes possible to test values such as `48` or `24`.

tests/paramgrill.c
```diff
@@ -340,76 +891,264 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    U8** mtAll = malloc(sizeof(U8*) * (ZSTD_btultra + 1));
+    int i;
+    if(mtAll == NULL) {
+        return NULL;
+    }
+    for(i = 1; i <= (int)ZSTD_btultra; i++) {
+        varLenNew = sanitizeVarArray(varyLen, varyParams, varNew, i);
+        mtAll[i] = malloc(sizeof(U8) * memoTableLen(varNew, varLenNew));
+        if(mtAll[i] == NULL) {
+            return NULL;
```
Cyan4973: and what about already allocated data ?

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+    DEBUGOUTPUT("Feas:\n");
+    benchres = BMK_benchMemInvertible(srcPtrs, srcSizes, dstPtrs, dstCapacityToSizes, nbBlocks, 0, &cParams, dictBuffer, dictBufferSize, ctx, dctx,
+                        BMK_both, BMK_iterMode, 1);
+    if(benchres.error) {
+        DISPLAY("ERROR %d !!\n", benchres.error);
+    }
+    for(i = 0; i < nbBlocks; i++) {
+        srcSize += srcSizes[i];
+    }
+    BMK_printWinner(stdout, CUSTOM_LEVEL, benchres.result, cParams, srcSize);
```
Cyan4973: what happens here if `benchres.error` ?

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+
+    if(!benchres.error) { 
+        *resultPtr = benchres.result;
+        /* if speed is 0 (only happens when time = 0) */
+        if(eqZero(benchres.result.cSpeed)) {
+            loopDurationC = 0;
+            uncertaintyConstantC = 2;
+        } else {
+            loopDurationC = (U64)((double)(srcSize * TIMELOOP_NANOSEC) / benchres.result.cSpeed); 
+            //problem - tested in fullbench, saw speed vary 3x between iters, maybe raise uncertaintyConstraint up?
```
Cyan4973: Such large measurement variation is a big issue.
We need to investigate.

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+        free(srcSizes);
+        free(dstPtrs);
+        free(dstCapacities);
+        return res;
+    }
+}
+
+/* specifically feasibleBenchMemo and infeasibleBenchMemo */
+//maybe not necessary 
+typedef int (*BMK_benchMemo_t)(BMK_result_t*, const void*, size_t, void*, size_t, ZSTD_CCtx*, ZSTD_DCtx*,
```
Cyan4973: while parameter names are not strictly necessary,
they help understanding what each parameter is about.

tests/paramgrill.c
```diff
@@ -831,13 +2572,20 @@ int main(int argc, const char** argv)
         filenamesStart=0,
         result;
     const char* exename=argv[0];
-    const char* input_filename=0;
+    const char* input_filename = 0;
```
Cyan4973: nit: prefer `NULL` instead of `0` for pointers.
It's _generally_ equivalent, though sometimes not : 
http://c-faq.com/null/machexamp.html

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+        if(ec) {
+            DISPLAY("Error Loading Files");
+            ret = ec;
+            goto _cleanUp;
+        }
+        benchedSize = 0;
+        for(i = 0; i < nbFiles; i++) {
+            benchedSize += fileSizes[i];
+        }
+        origBuff = realloc(origBuff, benchedSize);
```
Cyan4973: the usage of `realloc()` can be pretty dangerous.
If it fails to allocate a new buffer of size `benchedSize`, it will actually return `NULL`,
but the initial buffer will still be allocated.
By overwriting `origBuff`, you lose track of the original buffer, making it impossible to `free()`.

You should also test the result of `realloc()`, since success is not guaranteed.

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
         winnerInfo_t winner;
-        BMK_result_t candidate;
+        U32 varNew[NUM_PARAMS];
         const size_t blockSize = g_blockSize ? g_blockSize : benchedSize;
+        U32 const maxNbBlocks = (U32) ((benchedSize + (blockSize-1)) / blockSize) + 1;
+        const size_t maxCompressedSize = ZSTD_compressBound(benchedSize) + (maxNbBlocks * 1024);
+        void* compressedBuffer = malloc(maxCompressedSize);
 
         /* init */
         if (ctx==NULL) { DISPLAY("\n ZSTD_createCCtx error \n"); free(origBuff); return 14;}
```
Cyan4973: what about `dctx` ?
if it's allocated, that's a memory leak.

Consider nesting, or using `goto`, for proper cleaning before leaving the function.

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+                                    winner.result = candidate;
+                                    winner.params = CParams;
+                                }
+                            }
+                        }
+                    }
+                } //best, -1, +1, ..., 
+
+            } else if (paramTarget.strategy == 0) { //constrained
+                int feas = 0, i, j;
```
Cyan4973: what is `i` ?
Could the name be more descriptive ?

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+                    }
                 }
-                if ( (candidate.cSize < winner.result.cSize)
-                   | ((candidate.cSize == winner.result.cSize) & (candidate.cSpeed > winner.result.cSpeed)) )
-                {
-                    winner.params = CParams;
-                    winner.result = candidate;
-                    BMK_printWinner(stdout, i, winner.result, winner.params, benchedSize);
-            }   }
+            }
```
Cyan4973: when there are so many levels nested,
it becomes hard to understand what is being closed.

Consider indicating in comment to which begin each end refers to,
at least for the outer most ones.

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+               const constraint_t target,
+               BMK_result_t* winnerResult, U8* memoTable,
+               const U32* varyParams, const int varyLen) {
+    size_t memind = memoTableInd(&cParams, varyParams, varyLen);
+    if(memoTable[memind] >= INFEASIBLE_THRESHOLD) {
+        return INFEASIBLE_RESULT; //see feasibleBenchMemo for concerns
+    } else {
+        const size_t blockSize = g_blockSize ? g_blockSize : srcSize;
+        U32 const maxNbBlocks = (U32) ((srcSize + (blockSize-1)) / blockSize) + (U32)nbFiles;
+        const void ** const srcPtrs = (const void** const)malloc(maxNbBlocks * sizeof(void*));
```
terrelln: Lets split blocks at the same place we are loading files, so we don't have to worry about splitting blocks here.

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+            int i, d;
+            better = 0;
+            DEBUGOUTPUT("Start\n");
+            cparam = winnerInfo.params;
+            BMK_printWinner(stdout, CUSTOM_LEVEL, winnerInfo.result, winnerInfo.params, srcSize);
+            candidateInfo.params = cparam;
+            //all dist-1 targets
+            //if we early end this, we should also randomize the order these are picked. 
+            for(i = 0; i < varLen; i++) {
+                paramVaryOnce(varArray[i], 1, &candidateInfo.params); /* +1 */
```
terrelln:
```c
candidateInfo.params = cparam;
```

tests/paramgrill.c
```diff
@@ -632,131 +1391,1112 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
-            for (i=1; i<=maxSeeds; i++) {
-                ZSTD_compressionParameters const CParams = ZSTD_getCParams(i, blockSize, 0);
-                BMK_benchParam(&candidate, origBuff, benchedSize, ctx, dctx, CParams);
-                if (candidate.cSpeed < (double)targetSpeed) {
-                    break;
+        {
+            /* strategy selection */
+            const int maxSeeds = g_noSeed ? 1 : ZSTD_maxCLevel();
+            DEBUGOUTPUT("Strategy Selection\n");
+            if(varLen == NUM_PARAMS && paramTarget.strategy == 0) { /* no variable based constraints */
```
terrelln: Combine these two branches. Instead of doing random initialization in the `else`, start with `ZSTD_getCParams()` and then apply the fixed parameters.

tests/README.md
```diff
@@ -88,3 +88,35 @@ as well as the 10,000 original files for more detailed comparison of decompressi
 will choose a random seed, and for 1 minute,
 generate random test frames and ensure that the
 zstd library correctly decompresses them in both simple and streaming modes.
+
+#### `paramgrill` - tool for generating compression table parameters and optimizing parameters on file given constraints
```
Cyan4973: 👍

tests/README.md
```diff
@@ -88,3 +88,35 @@ as well as the 10,000 original files for more detailed comparison of decompressi
...
+    S# - strategy
+    L# - level
+ --zstd=      : Single run, parameter selection syntax same as zstdcli
+ --optimize=  : find parameters to maximize compression ratio given parameters
+    Can use all --zstd= commands to constrain the type of solution found in addition to the following constraints
+    cSpeed= - Minimum compression speed
+    dSpeed= - Minimum decompression speed
+    cMem= - compression memory
+    lvl= - Automatically sets compression speed constraint to the speed of that level
+ --optimize=  : same as -O with more verbose syntax
```
Cyan4973: there are two `--optimize=` but no `-O`

programs/bench.c
```diff
@@ -709,44 +720,60 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
     const void ** const srcPtrs = (const void** const)malloc(maxNbBlocks * sizeof(void*));
```
Cyan4973: nit : 
the final `const` in variable declaration is useful : it states the variable cannot change after its creation, which is great for reviewer, since it's not longer necessary to track these variables in case they would change later in the function.

But in the casting, `(type const)`, such final `const` does not achieve anything useful.

programs/bench.c
```diff
@@ -709,44 +720,60 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
...
+    BMK_return_t results = { { 0, 0, 0, 0 }, 0 };
+    int allocationincomplete;
+
+    if(!dstCapacity) {
+        compressedBuffer = malloc(maxCompressedSize);
+    } else {
+        compressedBuffer = dstBuffer;
+    }
+
+    allocationincomplete = !compressedBuffer || !resultBuffer ||
```
Cyan4973: `cCapacities` seems not checked for `malloc()` success.

Since allocation success is tested much later, and there are so many buffers,
try some simple rule,
such as keeping variable order identical,
which would help detecting such missing test.

programs/bench.c
```diff
@@ -691,10 +699,13 @@ static BMK_return_t BMK_benchMemAdvancedNoAlloc(
     }
     DISPLAYLEVEL(2, "%2i#\n", cLevel);
     }   /* Bench */
+    results.result.cMem = (1ULL << (comprParams->windowLog)) + ZSTD_sizeof_CCtx(ctx);
+    results.error = 0;
     return results;
 }
 
 BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
+                        void* dstBuffer, size_t dstCapacity,
```
Cyan4973: according to function description,
if `dstBuffer==NULL`, then `dstCapacity==0`.
And I guess the reverse condition is also expected ?

`assert()` such condition.
Otherwise, you would have to deal with such case in the code flow,
and currently, a case like `dstBuffer==NULL` but `dstCapacity>0` doesn't end well.

programs/bench.c
```diff
@@ -709,44 +720,60 @@ BMK_return_t BMK_benchMemAdvanced(const void* srcBuffer, size_t srcSize,
...
             srcBuffer, srcSize, fileSizes, nbFiles, cLevel, comprParams,
             dictBuffer, dictBufferSize, ctx, dctx, displayLevel, displayName, adv);
     }
+    
     /* clean up */
     BMK_freeTimeState(timeStateCompress);
     BMK_freeTimeState(timeStateDecompress);
-    free(compressedBuffer);
+    if(!dstCapacity) { /* only free if not given */
+        free(compressedBuffer);
```
Cyan4973: alternative way of doing the same thing, 
given a `dstBuffer` which might optionally be `NULL` : 
```
void* const internalDstBuffer = dstBuffer ? NULL : malloc(maxCompressedSize);
void* const compressedBuffer = dstBuffer ? dstBuffer : internalDstBuffer;

(... test malloc() success on compressedBuffer ...)
(... use compressedBuffer ...)

free(internalDstBuffer);   // free() unconditionnally
```

Advantage : 
the rule to create and free `internalDstBuffer` is not distributed : 
it's located in a single place, where `internalDstBuffer` is created.

This avoids potential discrepancies when, in the future, this condition is updated for whatever reason.

programs/bench.h
```diff
@@ -170,8 +175,9 @@ typedef size_t (*BMK_initFn_t)(void*);
  * srcBuffers - an array of buffers to be operated on by benchFn
  * srcSizes - an array of the sizes of above buffers
  * dstBuffers - an array of buffers to be written into by benchFn
- * dstCapacities - an array of the capacities of above buffers.
+ * dstCapacitiesToSizes - an array of the capacities of above buffers. Output modified to compressed sizes of those blocks.
```
Cyan4973: this comment is no longer correct

tests/paramgrill.c
```diff
@@ -340,76 +823,269 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    if(mtAll == NULL) { return; }
+    for(i = 1; i <= (int)ZSTD_btultra; i++) {
+        free(mtAll[i]);
+    }
+    free(mtAll);
+}
+
+/* inits memotables for all (including mallocs), all strategies */
+/* takes unsanitized varyParams */
+static U8** memoTableInitAll(ZSTD_compressionParameters paramConstraints, constraint_t target, const varInds_t* varyParams, const int varyLen, const size_t srcSize) {
```
Cyan4973: It's more than an `init`,
this function is actually allocating the space returned,
which means, it will have to be `free()` afterwards.

In our coding convention, we use the word `create` to make such distinction.
Typically `createMemoTable()`,
which implies that a corresponding `freeMemoTable()` necessarily exist.

`initMemoTable()` would be reserved for situations where `U8** mtAll` already exists, 
and needs to be set to some specific values.
This is useful if there is more than one way to reserve `mtAll` area.

tests/paramgrill.c
```diff
@@ -491,7 +1187,7 @@ static void BMK_benchFullTable(ZSTD_CCtx* cctx, ZSTD_DCtx* dctx, const void* src
     winnerInfo_t winners[NB_LEVELS_TRACKED+1];
     const char* const rfName = "grillResults.txt";
     FILE* const f = fopen(rfName, "w");
-    const size_t blockSize = g_blockSize ? g_blockSize : srcSize;   /* cut by block or not ? */
+    const size_t blockSize = g_blockSize ? g_blockSize : ZSTD_BLOCKSIZE_MAX;   /* cut by block or not ? */
```
Cyan4973: Any reason you prefer to change the default value to `ZSTD_BLOCKSIZE_MAX` ?
This will directly impact the selection of `l1params` below, and I'm not sure to understand why it's preferable.

tests/paramgrill.c
```diff
@@ -631,132 +1327,684 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+    buff->srcSizes = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
+
+    buff->dstPtrs = (void**)calloc(maxNbBlocks, sizeof(void*));
+    buff->dstCapacities = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
+    buff->dstSizes = (size_t*)malloc(maxNbBlocks * sizeof(size_t));
+
+    buff->resPtrs = (void**)calloc(maxNbBlocks, sizeof(void*));
+    buff->resSizes = (size_t*)malloc(maxNbBlocks * sizeof(size_t)); 
+
+    if(!buff->srcPtrs || !buff->srcSizes || !buff->dstPtrs || !buff->dstCapacities || !buff->dstCapacities || !buff->resPtrs || !buff->resSizes) {
```
Cyan4973: `dstCapacities` is tested twice, while `dstSizes` is not tested

programs/bench.h
```diff
@@ -182,12 +188,11 @@ typedef size_t (*BMK_initFn_t)(void*);
...
  *          dstBuffer.
  */
-BMK_customReturn_t BMK_benchFunction(                        
-                        BMK_benchFn_t benchFn, void* benchPayload,
+BMK_customReturn_t BMK_benchFunction(BMK_benchFn_t benchFn, void* benchPayload,
                         BMK_initFn_t initFn, void* initPayload,
                         size_t blockCount,
                         const void* const * const srcBuffers, const size_t* srcSizes,
-                        void* const * const dstBuffers, const size_t* dstCapacities,
+                        void * const * const dstBuffers, const size_t* dstCapacities, size_t* cSizes,
```
Cyan4973: see previous comment on `cSizes` being too specific for `benchFunction`

Cyan4973:
Next stage : 
if this patch is going to be merged, it's going to be maintained.
Give future developers enough information so that they can maintain your code.

This includes explaining the general strategy of your algorithm, and how it's implemented, in code comment.

tests/paramgrill.c
```diff
@@ -631,132 +1328,684 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+    ZSTD_freeCCtx(ctx.cctx);
+    ZSTD_freeDCtx(ctx.dctx);
+}
+
+static int createContexts(contexts_t* ctx, const char* dictFileName) {
+    FILE* f;
+    size_t readSize;
+    ctx->cctx = ZSTD_createCCtx();
+    ctx->dctx = ZSTD_createDCtx();
+    if(dictFileName == NULL) {
```
Cyan4973: what happens if `cctx==NULL` or `dctx==NULL` ?

tests/paramgrill.c
```diff
@@ -631,132 +1328,684 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+    free(ctx.dictBuffer);
+    ZSTD_freeCCtx(ctx.cctx);
+    ZSTD_freeDCtx(ctx.dctx);
+}
+
+static int createContexts(contexts_t* ctx, const char* dictFileName) {
+    FILE* f;
+    size_t readSize;
+    ctx->cctx = ZSTD_createCCtx();
+    ctx->dctx = ZSTD_createDCtx();
```
Cyan4973: what happens if `cctx==NULL` or `dctx==NULL` ?

tests/paramgrill.c
```diff
@@ -631,132 +1328,684 @@ int benchFiles(const char** fileNamesTable, int nbFiles)
...
+                DISPLAY("Climb Part 2\n");
+            }
+        }
+        winnerInfo = bestFeasible1;
+    }
+
+    return winnerInfo;
+}
+
+//optimizeForSize but with fixed strategy
```
Cyan4973: nit: 
prefer `/* ... */` comment style

tests/paramgrill.c
```diff
@@ -140,30 +175,137 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
...
+static double resultScore(const BMK_result_t res, const size_t srcSize, const constraint_t target) {
+    double cs = 0., ds = 0., rt, cm = 0.;
+    const double r1 = 1, r2 = 0.1, rtr = 0.5;
+    double ret;
+    if(target.cSpeed) { cs = res.cSpeed / (double)target.cSpeed; }
+    if(target.dSpeed) { ds = res.dSpeed / (double)target.dSpeed; }
+    if(target.cMem != (U32)-1) { cm = (double)target.cMem / res.cMem; }
+    rt = ((double)srcSize / res.cSize);
+
+    ret = (MIN(1, cs) + MIN(1, ds)  + MIN(1, cm))*r1 + rt * rtr +
```
Cyan4973: Could you explain, in comment, the scoring function ?

tests/paramgrill.c
```diff
@@ -140,30 +175,125 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
...
 
 const char* g_stratName[ZSTD_btultra+1] = {
                 "(none)       ", "ZSTD_fast    ", "ZSTD_dfast   ",
                 "ZSTD_greedy  ", "ZSTD_lazy    ", "ZSTD_lazy2   ",
                 "ZSTD_btlazy2 ", "ZSTD_btopt   ", "ZSTD_btultra "};
 
-/* TODO: support additional parameters (more files, fileSizes) */
-static size_t
-BMK_benchParam(BMK_result_t* resultPtr,
+/* benchParam but only takes in one file. */
```
Cyan4973: apparently, it only accepts one _input buffer_

tests/paramgrill.c
```diff
@@ -173,26 +303,364 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+    w1.result.dSpeed = 0.;
+    w1.result.cMem = (size_t)-1;
+    w1.result.cSize = (size_t)-1;
+    w1.params = p;
+    return w1;
+}
+
+typedef struct {
+    size_t srcSize;
+    void** srcPtrs;
```
Cyan4973: could `srcPtrs` be a `const void**` ?

tests/paramgrill.c
```diff
@@ -173,26 +303,364 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+    ZSTD_CCtx* ctx;
+    const void* dictBuffer;
+    size_t dictBufferSize;
+    int cLevel;
+    const ZSTD_compressionParameters* comprParams;
+    const BMK_advancedParams_t* adv;
+} BMK_initCCtxArgs;
+
+static size_t local_initCCtx(void* payload) {
+    BMK_initCCtxArgs* ag = (BMK_initCCtxArgs*)payload;
```
Cyan4973: nit: `const` since it's read only,
it's nice hint to the reviewer that the content of this object is not going to be modified.

tests/paramgrill.c
```diff
@@ -173,26 +303,364 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+            }
+        }
+    }
+   /* Bench */
+    results.result.cMem = (1 << (comprParams->windowLog)) + ZSTD_sizeof_CCtx(cctx);
+    return results;
+}
+
+/* global winner used for display. */ 
+static winnerInfo_t g_winner = { { 0, 0, (size_t)-1, (size_t)-1 } , { 0, 0, 0, 0, 0, 0, ZSTD_fast } };
```
Cyan4973: If it's used only within `BMK_printWinner()`, 
it's probably better to declare this variable within `BMK_printWinner()`, 
so that its scope remains limited to this function.

tests/paramgrill.c
```diff
@@ -173,26 +303,364 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+        }
+    }
+   /* Bench */
+    results.result.cMem = (1 << (comprParams->windowLog)) + ZSTD_sizeof_CCtx(cctx);
+    return results;
+}
+
+/* global winner used for display. */ 
+static winnerInfo_t g_winner = { { 0, 0, (size_t)-1, (size_t)-1 } , { 0, 0, 0, 0, 0, 0, ZSTD_fast } }; 
+static constraint_t g_targetConstraints;
```
Cyan4973: Could it be a parameter of `BMK_printWinner()` instead ?

tests/paramgrill.c
```diff
@@ -340,76 +808,270 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+            case clog_ind: ptr->chainLog     = ind % CLOG_RANGE + ZSTD_CHAINLOG_MIN;                              ind /= CLOG_RANGE; break;
+            case hlog_ind: ptr->hashLog      = ind % HLOG_RANGE + ZSTD_HASHLOG_MIN;                               ind /= HLOG_RANGE; break;
+            case slog_ind: ptr->searchLog    = ind % SLOG_RANGE + ZSTD_SEARCHLOG_MIN;                             ind /= SLOG_RANGE; break;
+            case slen_ind: ptr->searchLength = ind % SLEN_RANGE + ZSTD_SEARCHLENGTH_MIN;                          ind /= SLEN_RANGE; break;
+            case tlen_ind: ptr->targetLength = tlen_table[(ind % TLEN_RANGE)];                                    ind /= TLEN_RANGE; break;
+        }
+    }
+}
+
+/* Initialize memotable, immediately mark redundant / obviously infeasible params as such */
```
Cyan4973: Explain what `memotable` is, its role within this algorithm.

tests/paramgrill.c
```diff
@@ -340,76 +808,270 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+static void createMemoTable(U8* memoTable, ZSTD_compressionParameters paramConstraints, const constraint_t target, const varInds_t* varyParams, const int varyLen, const size_t srcSize) {
```
Cyan4973: My understanding is that `memoTable` is supposed already allocated,
and passed as parameter of this function.

In which case, it's not a `create` action, rather it is an `init` one.

tests/paramgrill.c
```diff
@@ -340,76 +808,270 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    size_t i;
+    size_t arrayLen = memoTableLen(varyParams, varyLen);
+    int cwFixed = !paramConstraints.chainLog || !paramConstraints.windowLog;
+    int scFixed = !paramConstraints.searchLog || !paramConstraints.chainLog;
+    int whFixed = !paramConstraints.windowLog || !paramConstraints.hashLog;
+    int wFixed = !paramConstraints.windowLog;
+    int j = 0;
+    memset(memoTable, 0, arrayLen);
```
Cyan4973: nit: `assert(memoTable != NULL);`

tests/paramgrill.c
```diff
@@ -340,76 +808,270 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+    }
+    return ind;
+}
+
+/* inverse of above function (from index to parameters) */
+static void memoTableIndInv(ZSTD_compressionParameters* ptr, const varInds_t* varyParams, const int varyLen, size_t ind) {
+    int i;
+    for(i = varyLen - 1; i >= 0; i--) {
+        switch(varyParams[i]) {
+            case wlog_ind: ptr->windowLog    = ind % WLOG_RANGE + ZSTD_WINDOWLOG_MIN;                             ind /= WLOG_RANGE; break;
```
Cyan4973: These lines are a bit long.
It makes reading difficult.
Use 2 lines instead,

tests/paramgrill.c
```diff
@@ -340,76 +808,270 @@ static int BMK_seed(winnerInfo_t* winners, const ZSTD_compressionParameters para
...
+        }
+    }
+    DEBUGOUTPUT("%d / %d Invalid\n", j, (int)i);
+    if((int)i == j) {
+        DEBUGOUTPUT("!!!Strategy %d totally infeasible\n", (int)paramConstraints.strategy)
+    }
+}
+
+/* frees all allocated memotables */
+static void memoTableFreeAll(U8** mtAll) {
```
Cyan4973: If this function is the inverse of `createMemoTableArray()`,
it should be called `freeTableArray()`.

tests/paramgrill.c
```diff
@@ -831,24 +2098,57 @@ int main(int argc, const char** argv)
...
+                if (longCommandWArg(&argument, "strategy=") || longCommandWArg(&argument, "strat=")) { paramTarget.strategy = (ZSTD_strategy)(readU32FromChar(&argument)); if (argument[0]==',') { argument++; continue; } else break; }
+                if (longCommandWArg(&argument, "compressionSpeed=") || longCommandWArg(&argument, "cSpeed=")) { target.cSpeed = readU32FromChar(&argument); if (argument[0]==',') { argument++; continue; } else break; }
+                if (longCommandWArg(&argument, "decompressionSpeed=") || longCommandWArg(&argument, "dSpeed=")) { target.dSpeed = readU32FromChar(&argument); if (argument[0]==',') { argument++; continue; } else break; }
+                if (longCommandWArg(&argument, "compressionMemory=") || longCommandWArg(&argument, "cMem=")) { target.cMem = readU32FromChar(&argument); if (argument[0]==',') { argument++; continue; } else break; }
+                if (longCommandWArg(&argument, "level=") || longCommandWArg(&argument, "lvl=")) { optimizerCLevel = readU32FromChar(&argument); if (argument[0]==',') { argument++; continue; } else break; }
```
Cyan4973: minor : it seems there is a lot of repetition in this code section.
You might be able to factor in the `if (argument[0]==',') { argument++; continue; } else break;` for example,
or avoid repeating the logic through function calls or macros.

Cyan4973:
Try compiling with `gcc-8` if possible, as it provides useful additional warning statements.

This is what I current get with it : 
```
paramgrill.c: In function ‘optimizeForSize’:
paramgrill.c:1685:9: warning: ‘buf’ may be used uninitialized in this function [-Wmaybe-uninitialized]
         freeBuffers(*buff);
         ^~~~~~~~~~~~~~~~~~
paramgrill.c:1883:15: note: ‘buf’ was declared here
     buffers_t buf;
               ^~~
In function ‘createContexts’,
    inlined from ‘optimizeForSize’ at paramgrill.c:1896:8:
paramgrill.c:1792:23: warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
     ctx->dictBuffer = malloc(ctx->dictSize);
                       ^~~~~~~~~~~~~~~~~~~~~
In file included from ../programs/util.h:24,
                 from paramgrill.c:15:
paramgrill.c: In function ‘optimizeForSize’:
/usr/include/stdlib.h:539:14: note: in a call to allocation function ‘malloc’ declared here
 extern void *malloc (size_t __size) __THROW __attribute_malloc__ __wur;
              ^~~~~~
In function ‘createContexts’,
    inlined from ‘optimizeForSize’ at paramgrill.c:1896:8:
paramgrill.c:1792:23: warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
     ctx->dictBuffer = malloc(ctx->dictSize);
                       ^~~~~~~~~~~~~~~~~~~~~
In file included from ../programs/util.h:24,
                 from paramgrill.c:15:
paramgrill.c: In function ‘optimizeForSize’:
/usr/include/stdlib.h:539:14: note: in a call to allocation function ‘malloc’ declared here
 extern void *malloc (size_t __size) __THROW __attribute_malloc__ __wur;
              ^~~~~~
```

tests/paramgrill.c
Cyan4973: Try giving more meaningful variable names.
The scoring function is relatively important in a hill climbing algorithm.
The ability to tune it is important.
This function must be clear enough for any maintainer.

programs/bench.h
```diff
@@ -141,8 +142,12 @@ BMK_return_t BMK_benchMem(const void* srcBuffer, size_t srcSize,
                         ZSTD_CCtx* ctx, ZSTD_DCtx* dctx,
```
Cyan4973: What is the intended benefit
of providing `ctx` and `dctx` in parameters ?

Cyan4973:
Please merge `facebook/zstd/dev` into your feature branch.
It includes latest fixes and tests.

Specifically, it enables `gcc-8` on travis, which will flag a few places for fixes.

tests/paramgrill.c
```diff
@@ -173,26 +308,370 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+/* dictionary nullable, nothing else though. */
+static BMK_return_t BMK_benchMemInvertible(buffers_t buf, contexts_t ctx, 
+                        const int cLevel, const ZSTD_compressionParameters* comprParams,
+                        const BMK_mode_t mode, const BMK_loopMode_t loopMode, const unsigned nbSeconds) {
+
+    U32 i;
+    BMK_return_t results = { { 0, 0., 0., 0 }, 0 } ;
+    const void *const *const srcPtrs = (const void *const *const)buf.srcPtrs;
+    size_t const *const srcSizes = buf.srcSizes;
+    void** dstPtrs = buf.dstPtrs;
```
Cyan4973: Try to use `const` whenever possible.

Most of the pointer variables I see here are not going to change after their initialization (while their content might be modified, which is a different topic).
So for example, `dstPtrs` should be declared : `void** const dstPtrs = (...)`.

When applied consistently, the _absence of `const`_ becomes a pretty useful signal for the reviewer 
underlining that a variable is going to change, hence needs to be tracked carefully.

tests/paramgrill.c
```diff
@@ -173,26 +308,370 @@ BMK_benchParam(BMK_result_t* resultPtr,
...
+    }
+    return out.pos;
+
+}
+
+/*-*******************************************************
+*  From Paramgrill End
+*********************************************************/
+
+/* Replicate function of benchMemAdvanced, but with pre-split src / dst buffers, with relevant info to invert it (compressedSizes) passed out. */
```
Cyan4973: This function is still not clear to me.
I get it that it is a variant of `benchMemAdvanced()`,
but I don't understand what is meant by `invert it`.

Cyan4973:
There seems to be some issue with `--fast` mode and `ZSTD_fast` strategy within `paramgrill`.

Here are some results from single benchmark runs : 
```
./paramgrill -SL1 silesia.tar
*** ZSTD parameters tester 1.3.5 64-bits, by Yann Collet ***
1: -SL1
2: silesia.tar
using silesia.tar :
Compression Ratio: 2.421  Compress Speed: 360.5 MB/s Decompress Speed: 1003.2 MB/s
```

In contrast, using `./zstd -b` internal memory benchmark (using the same `bench.c` module) : 
```
$ ../zstd -b -1 silesia.tar
 1#silesia.tar       : 211984896 ->  73651053 (2.878), 301.6 MB/s , 702.8 MB/s
$ ../zstd -b --fast=1 silesia.tar
-1#silesia.tar       : 211984896 ->  87204222 (2.431), 329.9 MB/s , 889.0 MB/s
```

Looking at the compression ratio, it seems that `paramgrill` triggers the `--fast` mode when requested to benchmark level 1.

Cyan4973:
minor thing to change : 
when comparing `zstd -b#` with `paramgrill -SL#`, 
speed results differ, by a non-negligible amount, with `paramgrill` being always faster.

I suspect a difference in meaning,
such as one (`zstd`) using 1024 bytes per KB, and the other (`paramgrill`) using 1000 bytes per KB.

It would be better to use a uniform metric across both implementations.

Since most programmers expect 1 KB to mean 1024 bytes, and consequently 1 MB to mean 1024 KB, I believe it's better to adhere to that scale.

Cyan4973:
Thanks @GeorgeLu97 !
It's a good patch for `paramgrill` !

I'll merge it now.
You can still add new features or fixes by creating new PR.

PR #1238, opened 2018-07-13
Add random dictionary builder

jennifermliu:


contrib/randomDictBuilder/random.c
```diff
@@ -0,0 +1,455 @@
...
+ * Selects a random segment from totalSamplesSize - k + 1 possible segments
+ */
+static RANDOM_segment_t RANDOM_selectSegment(const RANDOM_ctx_t *ctx,
+                                            ZDICT_random_params_t parameters) {
+    const U32 k = parameters.k;
+    RANDOM_segment_t segment;
+    unsigned index;
+
+    /* Seed random number generator */
+    srand((unsigned)time(NULL));
```
terrelln: Don't seed `srand()`, we want a deterministic dictionary builder.

contrib/randomDictBuilder/random.c
```diff
@@ -0,0 +1,455 @@
...
+  }
+}
+
+
+/**
+ * Prepare a context for dictionary building.
+ * Returns 1 on success or zero on error.
+ * The context must be destroyed with `RANDOM_ctx_destroy()`.
+ */
+static int RANDOM_ctx_init(RANDOM_ctx_t *ctx, const void *samplesBuffer,
```
terrelln: I don't think we need to create a context. It looks like we only use `ctx->samples` and `ctx->totalSamplesSize`. We can just pass those in as parameters to avoid the extra complexity.

contrib/randomDictBuilder/random.c
```diff
@@ -0,0 +1,455 @@
...
+            DISPLAYLEVEL(2, "Constructed dictionary of size %u\n",
+                          (U32)dictSize);
+        }
+        RANDOM_ctx_destroy(&ctx);
+        return dictSize;
+      }
+}
+
+
+int RANDOM_trainFromFiles(const char* dictFileName, unsigned maxDictSize,
```
terrelln: Lets refactor this function so it just loads the files and returns the `srcBuffer`, `sampleSizes`, and `nbSamples`, and move it into `main.c`. This way we can reuse it for the next dictionary builder, and only need to duplicate the dictionary builder function call.

contrib/randomDictBuilder/random.c
```diff
@@ -0,0 +1,455 @@
...
+        const char* const tmp = fileNamesTable[j];
+        fileNamesTable[j] = fileNamesTable[i];
+        fileNamesTable[i] = tmp;
+    }
+}
+
+
+
+/*-********************************************************
+*  Dictionary training functions
```
terrelln: Lets also move all of these into `main.c`.

contrib/randomDictBuilder/random.c
```diff
@@ -0,0 +1,455 @@
...
+/** loadFiles() :
+ *  load samples from files listed in fileNamesTable into buffer.
+ *  works even if buffer is too small to load all samples.
+ *  Also provides the size of each sample into sampleSizes table
+ *  which must be sized correctly, using DiB_fileStats().
+ * @return : nb of samples effectively loaded into `buffer`
+ * *bufferSizePtr is modified, it provides the amount data loaded within buffer.
+ *  sampleSizes is filled with the size of each sample.
+ */
+static unsigned loadFiles(void* buffer, size_t* bufferSizePtr,
```
terrelln: Lets also move all of these into `main.c`.

contrib/randomDictBuilder/main.c
```diff
@@ -0,0 +1,399 @@
...
+  const char** filenameTable = (const char**)malloc(argCount * sizeof(const char*));
+  unsigned filenameIdx = 0;
+
+  for (int i = 1; i < argCount; i++) {
+    const char* argument = argv[i];
+    if (longCommandWArg(&argument, "k=")) { k = readU32FromChar(&argument); continue; }
+    if (longCommandWArg(&argument, "dictID=")) { dictID = readU32FromChar(&argument); continue; }
+    if (longCommandWArg(&argument, "maxdict=")) { maxDictSize = readU32FromChar(&argument); continue; }
+    if (longCommandWArg(&argument, "in=")) {
+      inputFile = malloc(strlen(argument) + 1);
```
terrelln: This currently leaks `inputFile`. Since `argv` outlives `main()` and `inputFile` is unsued, we can simply do:

```
fileNameTable[filenameIdx] = argument;
filenameIdx++;
```

Same for the `out=` argument.

contrib/randomDictBuilder/main.c
```diff
@@ -0,0 +1,198 @@
...
+  zParams.compressionLevel = DEFAULT_CLEVEL;
+  zParams.notificationLevel = displayLevel;
+  zParams.dictID = dictID;
+  params.zParams = zParams;
+  params.k = k;
+
+  sampleInfo* info= getSampleInfo(filenameTable,
+                    filenameIdx, blockSize, maxDictSize, zParams.notificationLevel);
+  operationResult = RANDOM_trainFromFiles(outputFile, info, maxDictSize, &params);
+
```
terrelln: We should call `UTIL_freeFileList()` here so we don't leak memory.

contrib/randomDictBuilder/io.c
```diff
@@ -0,0 +1,243 @@
...
+        fs.nbSamples += nbSamples;
+    }
+    DISPLAYLEVEL(4, "Preparing to load : %u KB \n", (U32)(fs.totalSizeToLoad >> 10));
+    return fs;
+}
+
+
+
+
+sampleInfo* getSampleInfo(const char** fileNamesTable, unsigned nbFiles, size_t chunkSize,
```
terrelln: Since we `malloc()` in this function, we'll need a `freeSampleInfo` function which `free()`s the memory that we call before exiting `main()`.

contrib/randomDictBuilder/io.h
```diff
@@ -0,0 +1,33 @@
...
+} fileStats;
+
+typedef struct {
+  const void* srcBuffer;
+  const size_t *samplesSizes;
+  size_t nbSamples;
+}sampleInfo;
+
+
+sampleInfo* getSampleInfo(const char** fileNamesTable, unsigned nbFiles, size_t chunkSize,
```
terrelln: Can you add some short documentation to these functions. I know they are only used by you, but it is good practice.

contrib/randomDictBuilder/main.c
```diff
@@ -0,0 +1,200 @@
...
+    /* Checks */
+    if (!dictBuffer)
+        EXM_THROW(12, "not enough memory for trainFromFiles");   /* should not happen */
+
+    {   size_t dictSize;
+        dictSize = ZDICT_trainFromBuffer_random(dictBuffer, maxDictSize, info->srcBuffer,
+                                             info->samplesSizes, info->nbSamples, *params);
+        DISPLAYLEVEL(2, "k=%u\n", params->k);
+        if (ZDICT_isError(dictSize)) {
```
terrelln: If this happens we will free `dictBuffer`, then use `dictBuffer` after free in `saveDict`, you're probably missing a `return 1`, or a `goto _cleanUp`.

contrib/randomDictBuilder/random.h
```diff
@@ -0,0 +1,36 @@
...
+#endif
+#include "zdict.h"
+
+/**
+ * A segment is an inclusive range in the source.
+ */
+typedef struct {
+  U32 begin;
+  U32 end;
+} RANDOM_segment_t;
```
terrelln: nit: The segment isn't used externally, so it can be declared in the `random.c` file.

contrib/randomDictBuilder/Makefile
```diff
@@ -0,0 +1,51 @@
...
+.PHONY: run
+run:
+	echo "Building a random dictionary with given arguments"
+	./main $(ARG)
+
+main: main.o io.o random.o libzstd.a
+	gcc main.o io.o random.o libzstd.a -o main
+
+main.o: main.c
+	gcc -c main.c -I io.h -I random.h -I ../../programs -I ../../lib/common -I ../../lib -I ../../lib/dictBuilder
```
terrelln: You don't need to pass `-I` for header files in the same directory, only in different directories, and you should put it in a variable.

The Makefile doesn't need to be 100% standard, but you should use the `$(CC)` variable and `$(CFLAGS)` variables like here. This also adds compilation with `-O3` which should speed it up considerably.

```
CC ?= gcc
CFLAGS ?= -O3
INCLUDES := -I ../../programs -I ../../lib/common -I ../../lib -I ../../lib/dictBuilder

main.o: main.c
	$(CC) $(CFLAGS) $(INCLUDES) -c main.c
```

contrib/randomDictBuilder/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+random.o: random.c
+	gcc -c random.c -I random.h -I ../../lib/common -I ../../lib/dictBuilder
```
terrelln:
```
	$(CC) $(CFLAGS) $(INCLUDES) -c random.c 
```

contrib/randomDictBuilder/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+io.o: io.c $(PROGRAM_FILES)
+	gcc -c io.c $(PROGRAM_FILES) -I io.h -I ../../programs -I ../../lib/common -I ../../lib -I ../../lib/dictBuilder
```
terrelln:
```
	$(CC) $(CFLAGS) $(INCLUDES) -c io.c $(PROGRAM_FILES)
```

contrib/randomDictBuilder/Makefile
```diff
@@ -0,0 +1,51 @@
+PROGRAM_FILES := ../../programs/fileio.c
```
terrelln: Do you actually need `fileio.c`?

contrib/randomDictBuilder/Makefile
```diff
@@ -0,0 +1,51 @@
...
+	zstd -be3 -D $(TEST_OUTPUT) -r $(TEST_INPUT) -q
+	rm -f $(TEST_OUTPUT)
+
+.PHONY: testshell
+testshell: test.sh
+	sh test.sh
+	echo "Finish running test.sh"
+
+.PHONY: clean
+clean:
```
terrelln:
```
clean:
	rm -f *.o main
	$(MAKE) -C ../../lib clean
	echo "Cleaning is completed"
```

jennifermliu: Should I also remove libzstd.a?

terrelln: Good point, yeah.

PR #1240, opened 2018-07-14
Add FreeBSD to list of zstd consumers

allanjude:
FreeBSD now supports ZSTD compression for user and kernel crash dumps


Cyan4973:
Sure !

PR #1244, opened 2018-07-17
fileio: Error in compression on read errors

terrelln:
We can write a corrupted file if the input file errors during a read.
We should return a non-zero error code in this case.


Cyan4973:
Looks good to me !

PR #1246, opened 2018-07-20
Benchmark dictionary builders

jennifermliu:


contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+                                                info->samplesSizes, info->nbSamples, coverParams);
+        } else {
+          size_t totalSize= 0;
+          for (int i = 0; i < info->nbSamples; i++) {
+            totalSize += info->samplesSizes[i];
+          }
+          size_t const maxMem = findMaxMem(totalSize * MEMMULT) / MEMMULT;
+          size_t loadedSize = (size_t) MIN ((unsigned long long)maxMem, totalSize);
+          fillNoise((char*)(info->srcBuffer) + loadedSize, NOISELENGTH);
+          dictSize = ZDICT_trainFromBuffer_unsafe_legacy(dictBuffer, maxDictSize, info->srcBuffer,
```
terrelln: You can avoid the extra complexity by just using `ZDICT_trainFromBuffer_legacy()`.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+                                               info->samplesSizes, info->nbSamples, *legacyParams);
+        }
+        if (ZDICT_isError(dictSize)) {
+            DISPLAYLEVEL(1, "dictionary training failed : %s \n", ZDICT_getErrorName(dictSize));   /* should not happen */
+            free(dictBuffer);
+            freeSampleInfo(info);
```
terrelln: We shouldn't free `info` here, the caller expects to own it.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+_cleanup:
+  freeDictInfo(dInfo);
+  return result;
+}
+
+/** benchmarkCover() :
+ *  Measure how long random dictionary builder takes and compression ratio with the cover dictionary
+ *  @return 0 if benchmark successfully, 1 otherwise
+ */
+int benchmarkCover(sampleInfo *srcInfo, unsigned maxDictSize,
```
terrelln: Lets merge these three functions, like you did with `createDictFromFiles()`.

You can derive the name to print from which parameter is non-null, or just pass it as an argument.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+ * @return 0 and doesn't modify *stringPtr otherwise.
+ */
+static unsigned longCommandWArg(const char** stringPtr, const char* longCommand){
+    size_t const comSize = strlen(longCommand);
+    int const result = !strncmp(*stringPtr, longCommand, comSize);
+    if (result) *stringPtr += comSize;
+    return result;
+}
+
+static void fillNoise(void* buffer, size_t length)
```
terrelln: We won't need this if we switch to the regular legacy dict builder.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+typedef struct {
+  const void* dictBuffer;
+  size_t dictSize;
+} dictInfo;
+
+
+/*-*************************************
+*  Commandline related functions
+***************************************/
+static unsigned readU32FromChar(const char** stringPtr){
```
terrelln: Lets put this function and `longCommandWArg()` in `io.h` so we can share them between this and random's `main.c`.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+                    filenameIdx, blockSize, maxDictSize, displayLevel);
+
+  /* set up zParams */
+  ZDICT_params_t zParams;
+  zParams.compressionLevel = cLevel;
+  zParams.notificationLevel = displayLevel;
+  zParams.dictID = dictID;
+
+  /* for random */
+  ZDICT_random_params_t randomParam;
```
terrelln: nit: prefer to put these in their own scope. This way I can clearly see that these variables won't be used later on in the function, and you can't accidentally use the wrong variable, like when checking error codes.

```c
{
  ZDICT_random_params_t randomParams;
  // ...
  if (randomResult) { /* ... */ }
}
```

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+  randomParam.zParams = zParams;
+  randomParam.k = k;
+  int randomResult = benchmarkRandom(srcInfo, maxDictSize, &randomParam);
+  if(randomResult) {
+    result = 1;
+    goto _cleanup;
+  }
+
+  /* for cover */
```
terrelln: This is good for now. Later on when we have the fast cover ready, we'll want to benchmark both the "strong" cover optimize, and regular cover with the same parameters we pass to fast cover, so we can see how much faster it is on an even field.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+  ZDICT_params_t zParams;
+  zParams.compressionLevel = cLevel;
+  zParams.notificationLevel = displayLevel;
+  zParams.dictID = dictID;
+
+  /* for random */
+  ZDICT_random_params_t randomParam;
+  randomParam.zParams = zParams;
+  randomParam.k = k;
+  int randomResult = benchmarkRandom(srcInfo, maxDictSize, &randomParam);
```
terrelln: nit: make this `const`, so it is clear that we won't change it. Any variable that you don't modify should be const, I wish it were the default in C.

contrib/benchmarkDictBuilder/benchmark.c
```diff
@@ -0,0 +1,458 @@
...
+
+
+
+int main(int argCount, const char* argv[])
+{
+  int displayLevel = 2;
+  const char* programName = argv[0];
+  int result = 0;
+  /* Initialize arguments to default values */
+  unsigned k = 200;
```
terrelln: nit: make these `const`, we can always mark them non-const if we allow them to be modified later on.

contrib/benchmarkDictBuilder/README.md
```diff
@@ -0,0 +1,43 @@
...
+| Algorithm     | Speed(sec)    | Compression Ratio  |
+| ------------- |:-------------:| ------------------:|
+| random        | 1.095083      |  2.309485          |
+| cover         | 517.999132    |  2.575331          |
+| legacy        | 10.789509     |  2.506775          |
+
+hg-changelog
+| Algorithm     | Speed(sec)    | Compression Ratio  |
+| ------------- |:-------------:| ------------------:|
+| random        | 0.639630      |  2.096785          |
```
terrelln: It looks like in general `random` doesn't perform as well as `legacy` or `cover`, but it is super interesting that `random` can get close to the `legacy` dictionary builder in some cases.

PR #1250, opened 2018-07-25
Add Fast Cover Dictionary Builder

jennifermliu:


contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -0,0 +1,749 @@
...
+  activeSegment.end = begin;
+  activeSegment.score = 0;
+  /* Slide the activeSegment through the whole epoch.
+   * Save the best segment in bestSegment.
+   */
+  while (activeSegment.end < end) {
+    /* Get hash value of current dmer  */
+    const size_t index = FASTCOVER_hashPtrToIndex(ctx->samples + activeSegment.end, parameters.f, ctx->d);
+    /* Add frequency of this index to score */
+    activeSegment.score += freqs[index];
```
terrelln: Try adding another map of `U16` the same size of `freqs`. We want to keep track of the number of the number of times that `index` shows up in the segment, and only increment the score if it is the first time it shows up (and only decrement if it is the only remaining instance).

I think this will be an important change, but we will have to benchmark it.

contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -0,0 +1,749 @@
...
+    ctx->offsets = NULL;
+  }
+}
+
+/**
+ * Calculate for frequency of hash value of each dmer in ctx->samples
+ */
+static void FASTCOVER_getFrequency(U32 *freqs, unsigned f, FASTCOVER_ctx_t *ctx){
+  /* inCurrSample keeps track of this hash value has already be seen in previous dmers in the same sample*/
+  size_t* inCurrSample = (size_t *)malloc((1<<f)*sizeof(size_t));
```
terrelln: This can be a `BYTE*`, since it is either 0 or 1. We could switch to a bitmap, but that can be a next step.

contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -0,0 +1,749 @@
...
+  size_t start; /* start of current dmer */
+  for (unsigned i = 0; i < ctx->nbTrainSamples; i++) {
+    memset(inCurrSample, 0, (1 << f)); /* Reset inCurrSample for each sample */
```
terrelln:
```
memset(inCurrSample, 0, (1 << f) * sizeof(*inCurrSample));
```

contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -0,0 +1,762 @@
...
+static void FASTCOVER_computeFrequency(U32 *freqs, unsigned f, FASTCOVER_ctx_t *ctx){
+  /* inCurrSample keeps track of this hash value has already be seen in previous dmers in the same sample*/
+  BYTE* inCurrSample = (BYTE *)malloc((1 << f) * sizeof(BYTE));
+  size_t start; /* start of current dmer */
+  for (unsigned i = 0; i < ctx->nbTrainSamples; i++) {
+    memset(inCurrSample, 0, (1 << f) * sizeof(*inCurrSample)); /* Reset inCurrSample for each sample */
+    size_t currSampleStart = ctx->offsets[i];
+    size_t currSampleEnd = ctx->offsets[i+1];
+    start = currSampleStart;
+    while (start + f < currSampleEnd) {
```
terrelln: Should this be `start + ctx->d <= currSampleEnd`?

contrib/experimental_dict_builders/benchmarkDictBuilder/benchmark.c
```diff
@@ -344,10 +352,9 @@ int main(int argCount, const char* argv[])
     memset(&coverParam, 0, sizeof(coverParam));
     coverParam.zParams = zParams;
     coverParam.splitPoint = 1.0;
-    coverParam.d = d;
     coverParam.steps = 40;
     coverParam.nbThreads = 1;
-    const int coverOptResult = benchmarkDictBuilder(srcInfo, maxDictSize, NULL, &coverParam, NULL);
+    const int coverOptResult = benchmarkDictBuilder(srcInfo, maxDictSize, NULL, &coverParam, NULL, NULL);
```
terrelln: `ZDICT_optimizeTrainFromBuffer_cover()` will fill `coverParam` with the selected params. We can use the parameters to run it a second time with a fixed `k` and `d`, to get the speed when it doesn't optimize. We can also use that `k` value to run `FASTCOVER` without having to optimize.

contrib/experimental_dict_builders/benchmarkDictBuilder/README.md
```diff
@@ -14,46 +14,107 @@ make ARG="in=../../../lib/dictBuilder in=../../../lib/compress"
...
+FAST17       0.268376       9.605798
+FAST18       9.988885       10.626382
+FAST18       0.311769       9.130565
+FAST19       10.737259       10.411729
+FAST19       0.331885       9.271814
+FAST20       10.479782       10.388895
+FAST20       0.498416       9.194115
+FAST21       21.189883       10.376394
+FAST21       1.098532       9.244456
+FAST22       39.849935       10.432555
```
terrelln: For each value of `f`, does approximately the same value for `k` get chosen?

contrib/experimental_dict_builders/benchmarkDictBuilder/benchmark.c
```diff
@@ -360,16 +361,50 @@ int main(int argCount, const char* argv[])
...
+    if(coverOptResult) {
+      result = 1;
+      goto _cleanup;
+    }
+
+    k = coverParam.k;
+    d = coverParam.d;
+
+    /* for COVER with k and d provided */
+    ZDICT_cover_params_t covernParam;
```
terrelln: nit: You should just be able to use the `coverParam` as is.

contrib/experimental_dict_builders/benchmarkDictBuilder/benchmark.c
```diff
@@ -379,17 +414,21 @@ int main(int argCount, const char* argv[])
...
     {
       ZDICT_fastCover_params_t fastParam;
       memset(&fastParam, 0, sizeof(fastParam));
       fastParam.zParams = zParams;
       fastParam.splitPoint = 1.0;
-      fastParam.d = 8;
+      fastParam.d = d;
       fastParam.f = f;
-      fastParam.k = 200;
+      fastParam.k = k;
```
terrelln: Does FASTCOVER select approximately the same `k` value as COVER? Does `d=6` or `d=8` win more often, or is it about evenly split?

contrib/experimental_dict_builders/benchmarkDictBuilder/README.md
jennifermliu: For github and hg-changelog, the chosen k value is very similar. For hg-commands and hg-manifest, the k value varies a lot across different f values.

contrib/experimental_dict_builders/benchmarkDictBuilder/benchmark.c
jennifermliu: For hg-changelog, COVER chooses k =98, which is the k value for most f values of FASTCOVER. For github, FASTCOVER with all f values chooses k=1778, which is very different from l=1298 chosen by COVER. For hg-commands and hg-manifest,  the k value varies greatly across different f values, so k value for some f values is close to COVER while others vary greatly. COVER chooses d=8 for all 4 sample sets. d=8 always wins for github and hg-changelog. For hg-changelog, d=8 wins more than d=6. For hg-manifest, there's an even split between d=8 and d=6.

PR #1254, opened 2018-07-27
 Save segmentFreqs in ctx instead of malloc and memset in SelectSegment

jennifermliu:


contrib/experimental_dict_builders/benchmarkDictBuilder/README.md
```diff
@@ -124,3 +126,114 @@ FAST23       377.196211       2.601416        8          194
 FAST23       13.497286       2.601416        8          194
 FAST24       503.208577       2.602830        6          194
 FAST24       29.538585       2.602830        6          194
+
+---------------------------------------------------------------
+Version 2 (save segmentFreqs in ctx instead of malloc and memset in every call to SelectSegment):
+
+github:
```
terrelln: Its probably easier to just overwrite what you had before. That way I get a nice diff side-by-side.

Since we have the commit history, we can see the progression over time by looking at the file after each optimization commit.

PR #1255, opened 2018-07-28
[FSE] Fix division by zero

terrelln:
When the primary normalization method fails, and
`(1 << tableLog) == (maxSymbolValue + 1)`, and every symbol gets assigned
normalized weight 1 or -1 in the first loop, then the next division can
raise `SIGFPE`.

I'm looking into getting a test case.

### DIFF
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -394,6 +394,9 @@ static size_t FSE_normalizeM2(short* norm, U32 tableLog, const unsigned* count, 
     }
     ToDistribute = (1 << tableLog) - distributed;
 
+    if (ToDistribute == 0)
+        return 0;
+
     if ((total / ToDistribute) > lowOne) {
         /* risk of rounding to zero */
         lowOne = (U32)((total * 3) / (ToDistribute * 2));
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -331,7 +331,7 @@ void FSE_freeCTable (FSE_CTable* ct) { free(ct); }
 /* provides the minimum logSize to safely represent a distribution */
 static unsigned FSE_minTableLog(size_t srcSize, unsigned maxSymbolValue)
 {
-    U32 minBitsSrc = BIT_highbit32((U32)(srcSize - 1)) + 1;
+    U32 minBitsSrc = BIT_highbit32((U32)(srcSize)) + 1;
     U32 minBitsSymbols = BIT_highbit32(maxSymbolValue) + 2;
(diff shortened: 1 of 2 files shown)


Cyan4973:
Interesting.
I thought that the case `(1 << tableLog) == (maxSymbolValue + 1)` was impossible,
because `tableLog` is supposed to be sized appropriately before entering this function
in a way which ensures that `(1 << tableLog) > (maxSymbolValue + 1)`.

Within `zstd`, this property is ensured there : 
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1903

I haven't checked lately, and there might be an off-by-one error somewhere, maybe enough to make the equality possible. Though [this line](https://github.com/facebook/zstd/blob/dev/lib/compress/fse_compress.c#L335) makes me think that `tableLog` should always be large enough, relative to `maxSymbolValue`.

Anyway, if equality is possible, your fix will solve this case.
And if it is impossible, an `assert()` should be present.

How to try generating a test case ?
In term of distribution, I would consider something like a 64 bytes input where the first 32 values (`[0-31])` are each present twice. This should ensure they all get a share of 1.
If that doesn't work, try 65 bytes, with the last byte being any of the first 32 bytes, on the ground that the last occurrence might be removed from statistics.

That will probably still not work. The issue here is that the normal `FSE` functions are supposed to be naturally protected from such cases, with an early detector which will tell that this distribution is not compressible. So it might be necessary to circumvent such protection, by calling directly inner functions.

It probably sounds like a stretch, but thing is, that's exactly what `zstd` does : to compress sequences, `zstd` re-assemble inner functions the way it wishes, instead of calling directly `FSE_compress()`.

But still, since it [invokes FSE_optimalTableLog() properly](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1903), I'm still surprised it ends up generating an equality, which is a necessary condition to trigger this division by zero.

terrelln:
The reproduction condition is calling:

```c
FSE_normalizeCount(tableLog=5, count={1, 1, 1, ...}, total=32, maxSymbolValue=31)
```

Cyan4973:
It's a great test case to add !

`FSE_normalizeCount()` is supposed to start with a [`tableLog` check](https://github.com/facebook/zstd/blob/dev/lib/compress/fse_compress.c#L456) : 

`if (tableLog < FSE_minTableLog(total, maxSymbolValue)) return ERROR(GENERIC);`

This test should fail. But apparently, it doesn't ?
Maybe something to fix there.

terrelln:
It is allowed to be equal to the number of sequences, but not less: https://github.com/facebook/zstd/blob/dev/lib/compress/fse_compress.c#L334

PR #1256, opened 2018-07-30
Add splitLimit to COVER

jennifermliu:
- Size of all training samples cannot exceed splitLimit.
- If splitPoint is 1.0, the same training samples are used for testing
- Otherwise, use remaining samples for testing, which may be more than (1-splitPoint) * 100% of all samples if number of training samples is updated so that total training size is below splitLimit.


Cyan4973:
@terrelln , what should be done about this PR ?

PR #1259, opened 2018-07-31
Some Makes Don't Allow Setting Variables Before Command

felixhandte:
Fixes b5207aadfaa0769a63bda0837a662957cbcd1f55.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ zlibwrapper:
 
 .PHONY: test
 test:
-	MOREFLAGS+="-g -DDEBUGLEVEL=1 -Werror" $(MAKE) -C $(PRGDIR) allVariants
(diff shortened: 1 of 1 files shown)


Cyan4973:
Setting a variable before `make` is actually a very important capability, 
it's the only way to start from a non-empty variable, and have it completed later on by the `Makefile` script.

I believe the issue is `+=`. This one works fine on OS-X, but is not guaranteed to be supported.

I changed it in this patch : 
https://github.com/facebook/zstd/commit/0840d02ecf74eae656e0df0d900000d9b0154cde

```
test: MOREFLAGS+="-g -DDEBUGLEVEL=1 -Werror"
test: 
    MOREFLAGS=$(MOREFLAGS) $(MAKE) -C $(PRGDIR) allVariants
```

which solves it.

This patch is part of a branch dedicated to improving test coverage on travis CI
which is not yet merged, 
as I discovered issues in the code recently merged into `bench.c`
which must be solved too for travis CI tests to pass.

PR #1260, opened 2018-08-01
Add non-optimize FASTCOVER

jennifermliu:
Changes Made:
- Call ZDICT_trainFromBuffer_fastCover instead of ZDICT_optimizeTrainFromBuffer_fastCover when both k and d provided. The new function eliminates the time spent on trying out different k and d and compressing with built dictionary to find the best parameters. 

Performance: 
- Compression ratio remains the same.
- For FASTCOVER with provided k and d, speed gets significantly faster.


contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -629,6 +629,55 @@ static void FASTCOVER_tryParameters(void *opaque) {
   }
 }
 
+ZDICTLIB_API size_t ZDICT_trainFromBuffer_fastCover(
+    void *dictBuffer, size_t dictBufferCapacity, const void *samplesBuffer,
+    const size_t *samplesSizes, unsigned nbSamples, ZDICT_fastCover_params_t *parameters) {
```
terrelln: We don't use `parameters` as an in-out value, so it can just be a value instead of a pointer.

Also, do we check that neither `k` nor `d` are zero?

jennifermliu: I think in FASTCOVER_checkParameter if either k or d is 0 the function would return 0.

PR #1261, opened 2018-08-02
Experiment with hashing different numbers of dmers together

jennifermliu:
Changes Made:
-  In computeFrequency, instead of hashing 1 dmer then increment frequency, hash "stride" number of dmers first and then increment frequency for those dmer hash values. 
- Added stridetest.c, which gets sampleInfo from input files, initializes a FASTCOVER context, then passes this context to the computeFrequency function. The amount of time spent on computeFrequency is recorded.
- Tried stride=1 (original condition), stride=4, and stride=8

Performance:
- For the stride test, I ran each stride value in 7 trials, with 200 iterations each. In each iteration, computeFrequency is run 100 times, and the overall time of each iteration is the total time used to run computeFrequency 100 times. The minimum of each trial is calculated. Overall, it seems like stride=8 has slightly less time than stride=4, which has slightly less time than stride=1. But since the difference is roughly 0.01 for 100 runs of computeFrequency, it's not clear whether changing the stride value affects computeFrequency.
- Based on benchmarking result with stride=8, compression ratio didn't change, which is expected since the exact same dmers are hashed. There's no obvious trend in how the speed changes. For github, the new speed is slightly better than old speed. For the other three sample sets, the new speed is actually slight slower than old speed for both optimize and non-optimize FASTCOVER.


terrelln:
It looks like there is basically no difference. What happens when we don't break the source up into samples?

I also recommend running something like this to benchmark:

```c
const size_t kRunsPerTrial = 20;
const size_t kTrials = 20;
for (size_t i = 0; i < kTrials; ++i) {
  // start timer
  for (size_t j = 0; j < kRunsPerTrial; ++j) {
    computeFrequencies()
  }
  // stop timer
  // Take min of each trial
}
// Only report the min value
```

We run multiple runs of `computeFrequencies()`, since each one is so fast the timer overhead can start to matter. We run it multiple times and take the min to avoid noise.

Tune the parameters until the results are fairly stable across runs.

jennifermliu:
I am going to benchmark not breaking source into samples in another pull request.

terrelln:
Can you update the summary with the new benchmarking method? Are the other 2 sample sets still consistently slower with `stride=8`?

terrelln:
If there is little difference, lets opt for the simpler code.

PR #1262, opened 2018-08-02
Experiment with not zeroing frequency for the selected segment

jennifermliu:
Changes Made:
- In selectSegment, the frequency of the dmer index in the selected segment is zeroed to avoid adding the same dmers multiple times. In this pull request, this step is deleted to see effect of zeroing frequency.

Performance:
- Compression ratio drops significantly, and new speed isn't better than old speed, so it seems like zeroing frequency is actually an important step.


terrelln:
Awesome, that is expected, lets drop this change.

PR #1264, opened 2018-08-04
Experiment with not breaking source into samples in computeFrequency

jennifermliu:
Changes Made:
- Originally, the dmers used to compute frequency are from the same sample. For a sample of size n, number of dmers in this sample is n-d+1. In this pull request, the sample boundary is removed so that a dmer can be formed from 2 samples. Overall, the number of dmers from a set of samples increases.

Performance:
- The new compression ratio is sometimes slightly better than old one and sometimes slightly worse than the old one. 
- Non-optimize FASTCOVER is benchmarked 5 times and the result in README.md is the minimum of the 5 trials. This is added to minimize noise. However, compression speed can still sometimes be worse than old compression speed. 
- In a separate test for computeFrequency, I ran old computeFrequency function and new computeFrequency function in 5 trials, with 200 iterations each. In each iteration, computeFrequency is run 100 times, and the overall time of each iteration is the total time used to run computeFrequency 100 times. The minimum of each trial is calculated. It seems like this change in computeFrequency has no significant change in time.


terrelln:
Cool, it looks like this change is within the noise for both compression ratio and speed, so we can skip it.

PR #1265, opened 2018-08-06
Experiment with passing in half of the training samples in finalizeDictionary

jennifermliu:
Changes Made:
- Originally, all training samples are passed to finalizeDictionary. In this pull request, only the first half of the training samples are passed to finalizeDictionary. This change is supposed to increase speed since finalizeDictionary takes a significant portion of the overall runtime. 

Performance:
- Compression ratio drops a little bit for some cases and slightly increases for other cases. 
- Overall, new compression speed is better than old speed. 
- Based on analysis from Instruments, time spent on finalizeDictionary went from 1.97s to 967ms for optimize FASTCOVER and from 31ms to 16 ms for non-optimize FASTCOVER. The new time is approximately 1/2 of old time, which is expected given that half of the training samples are used in finalizeDictionary.


jennifermliu:
Added in #1274

PR #1266, opened 2018-08-06
Experiment with skipping every other dmer in computeFrequency

jennifermliu:
Changes Made:
- Originally, every dmer is used to computeFrequency. In this pull request, every other dmer is skipped, so computeFrequency should only use half of the dmers.

Performance:
- Compression ratio drops a little bit for some cases and slightly increases for other cases.
- Overall, new compression speed is better than old speed.
- Based on analysis from Instruments, time spent on computeFrequency went from 51ms to 38ms. The new time is approximately 3/4 of old time.


jennifermliu:
Added in #1274

PR #1267, opened 2018-08-07
Experiment with using hash table from COVER 

jennifermliu:
Changes Made:
- Originally, an array called segmentFreqs of size 2^f is used to keep track of number of occurrence of a dmer in the current segment. In this pull request, the hash map from COVER is used to keep track of this information.

Performance:
- Time increases for all f values for all sample sets except for non-optimize FASTCOVER with f=24 for github.
- Compression ratio remains the same.
- Since the compression ratio doesn't change from this change but the speed got slower, it seems like this change should not be merged.


contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -80,12 +80,291 @@ typedef struct {
...
+/*-*************************************
+*  Cover functions
+***************************************/
+
+/**
+ * Called on each group of positions with the same dmer.
+ * Counts the frequency of each dmer and saves it in the suffix array.
+ * Fills `ctx->dmerAt`.
+ */
+static void COVER_group(FASTCOVER_ctx_t *ctx, const void *group,
```
terrelln: You don't need this.

contrib/experimental_dict_builders/fastCover/fastCover.c
```diff
@@ -150,19 +431,24 @@ static FASTCOVER_segment_t FASTCOVER_selectSegment(const FASTCOVER_ctx_t *ctx,
       /* Get hash value of current dmer */
       const size_t index = FASTCOVER_hashPtrToIndex(ctx->samples + activeSegment.end, parameters.f, ctx->d);
       /* Add frequency of this index to score if this is the first occurence of index in active segment */
-      if (ctx->segmentFreqs[index] == 0) {
+      U32 newDmer = ctx->dmerAt[activeSegment.end];
+      U32 *newDmerOcc = COVER_map_at(activeDmers, newDmer);
```
terrelln: The key is `index`

```c
U32 *newDmerOcc = COVER_map_at(activeDmers, (U32)index);
```

terrelln:
Cool, looks like this is slightly slower still, even with `f=23`, lets stick with the `segmentFreqs` table.

PR #1275, opened 2018-08-14
MultiLvl

GeorgeLu97:
Add Display - 
	Prints table of pareto frontier when in lvl mode.
Performance Improvements - 
	Added back in strategy switching
	Stop Strategy selection if too slow
Interface options - 
	Added separate files option
	Added option to adjust strictness (how much slack in lvl constraints)
	Added option to adjust speed/ratio tradeoff preference
	Added option to change number of tries before restarting with a different strategy

### DIFF
--- a/tests/paramgrill.c
+++ b/tests/paramgrill.c
@@ -139,8 +139,9 @@ struct ll_node {
 static ll_node* g_winners; /* linked list sorted ascending by cSize & cSpeed */
 static BMK_result_t g_lvltarget;
 
-/* range 0 - 99 */
-static U32 g_strictness = 99;
+/* range 0 - 99, measure of how strict  */
+#define DEFAULT_STRICTNESS 99999
+static U32 g_strictness = DEFAULT_STRICTNESS;
 
 void BMK_SetNbIterations(int nbLoops)
 {
@@ -318,8 +319,8 @@ static int compareResultLT(const BMK_result_t result1, const BMK_result_t result
 
 static constraint_t relaxTarget(constraint_t target) {
     target.cMem = (U32)-1;
-    target.cSpeed *= ((double)(g_strictness + 1) / 100); 
-    target.dSpeed *= ((double)(g_strictness + 1) / 100);
+    target.cSpeed *= ((double)(g_strictness) / 100); 
+    target.dSpeed *= ((double)(g_strictness) / 100);
     return target;
 }
 
@@ -2080,8 +2081,7 @@ static ZSTD_compressionParameters maskParams(ZSTD_compressionParameters base, ZS
 }
 
 /* experiment with playing with this and decay value */
-#define MAX_TRIES 8
-#define TRY_DECAY 3
+
 /* main fn called when using --optimize */
 /* Does strategy selection by benchmarking default compression levels
  * then optimizes by strategy, starting with the best one and moving 
@@ -2094,6 +2094,9 @@ static ZSTD_compressionParameters maskParams(ZSTD_compressionParameters base, ZS
  * paramTarget - parameter constraints (i.e. restriction search space to where strategy = ZSTD_fast)
  * cLevel - compression level to exceed (all solutions must be > lvl in cSpeed + ratio)
  */
+
+#define MAX_TRIES 3
+#define TRY_DECAY 1
 
 static int optimizeForSize(const char* const * const fileNamesTable, const size_t nbFiles, const char* dictFileName, constraint_t target, ZSTD_compressionParameters paramTarget, int cLevel)
 {
@@ -2164,6 +2167,21 @@ static int optimizeForSize(const char* const * const fileNamesTable, const size_
         goto _cleanUp;
     }
  
+    /* default strictness = Maximum for */
+    if(g_strictness == DEFAULT_STRICTNESS) {
+        if(cLevel) {
+            g_strictness = 99;
+        } else {
+            g_strictness = 90;
+        }
+    } else {
+        if(0 >= g_strictness || g_strictness > 100) {
+            DISPLAY("Strictness Outside of Bounds\n");
+            ret = 4;
+            goto _cleanUp;
+        }
+    }
+
     /* use level'ing mode instead of normal target mode */
     if(cLevel) {
         winner.params = ZSTD_getCParams(cLevel, maxBlockSize, ctx.dictSize);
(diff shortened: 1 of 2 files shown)


tests/paramgrill.c
```diff
@@ -108,17 +110,57 @@ static double g_compressibility = COMPRESSIBILITY_DEFAULT;
 static U32 g_blockSize = 0;
 static U32 g_rand = 1;
 static U32 g_singleRun = 0;
+static U32 g_optimizer = 0;
 static U32 g_target = 0;
 static U32 g_noSeed = 0;
-static ZSTD_compressionParameters g_params = { 0, 0, 0, 0, 0, 0, ZSTD_greedy };
+static ZSTD_compressionParameters g_params;
```
Cyan4973: explain it's going to be initialized somewhere else.
explain why it's guaranteed to not be used initialized.

note : the memory sanitizer and valgrind can typically be used to check that. Some static analyzer might also be able to prove it, thought success rate can vary.

tests/paramgrill.c
```diff
@@ -482,6 +506,175 @@ static size_t local_defaultDecompress(
...
+}
+
+/* allocates buffer's arguments. returns success / failuere */
+static int createBuffers(buffers_t* buff, const char* const * const fileNamesTable, 
+                          const size_t nbFiles)
+{
+    size_t pos = 0;
+    size_t n;
+    U64 const totalSizeToLoad = UTIL_getTotalFileSize(fileNamesTable, (U32)nbFiles);
+    size_t benchedSize = MIN(BMK_findMaxMem(totalSizeToLoad * 3) / 3, totalSizeToLoad);
```
Cyan4973: `const` benchedSize

tests/paramgrill.c
```diff
@@ -482,6 +506,175 @@ static size_t local_defaultDecompress(
...
+static void freeContexts(const contexts_t ctx) {
+    free(ctx.dictBuffer);
+    ZSTD_freeCCtx(ctx.cctx);
+    ZSTD_freeDCtx(ctx.dctx);
+}
+
+static int createContexts(contexts_t* ctx, const char* dictFileName) {
+    FILE* f;
+    size_t readSize;
+    ctx->cctx = ZSTD_createCCtx();
```
Cyan4973: do you still need `cctx` and `dctx` anywhere ?

tests/paramgrill.c
```diff
@@ -1087,6 +1468,17 @@ static U8** createMemoTableArray(ZSTD_compressionParameters paramConstraints, co
     return mtAll;
 }
 
+static ZSTD_compressionParameters maskParams(ZSTD_compressionParameters base, ZSTD_compressionParameters mask) {
```
Cyan4973: minor naming detail : 

a `mask` is generally something different : 
it's typically a bitmap, which selectively enable / disable each target field, using as low as one bit per signal.

Here, this is more about a vector `ZSTD_compressionParameters` which may contain optional parameters.
In this case, they overwrite existing parameters.

Another name like `new` or equivalent would better translate the intention.

tests/paramgrill.c
GeorgeLu97: It's used in benchMemInvertible. I suppose I could remove it as an argument from there and explicitly allocate in benchMemInvertible, but I don't really see a benefit there since then the contexts would pretty much be the only thing benchMemInvertible manages, where right now contexts_t / buffers_t allocation is in one place.

tests/paramgrill.c
```diff
@@ -1175,174 +1567,199 @@ static void randomConstrainedParams(ZSTD_compressionParameters* pc, varInds_t* v
...
+    size_t splitSize = 0;
 
-    void* origBuff = malloc(benchedSize);
-    if (!origBuff) { perror("not enough memory"); return 12; }
+    buffers_t buf;
+    contexts_t ctx;
 
-    /* Fill buffer */
-    RDG_genBuffer(origBuff, benchedSize, g_compressibility, 0.0, 0);
+    buf.srcPtrs = (const void**)calloc(maxNbBlocks, sizeof(void*));
```
Cyan4973: is there a reason to not use `createBuffers()` here ?

tests/paramgrill.c
```diff
@@ -482,6 +506,175 @@ static size_t local_defaultDecompress(
...
+    }
+    
+    if(ctx->dictSize > 64 MB || !(ctx->dictBuffer)) {
+        DISPLAY("dictionary too large\n");
+        fclose(f);
+        freeContexts(*ctx);
+        return 1;
+    }
+    readSize = fread(ctx->dictBuffer, 1, ctx->dictSize, f);
+    if(readSize != ctx->dictSize) {
```
Cyan4973: If `(readSize == ctx->dictSize)`, which is the intended behavior,
then `f` will not be `fclose(f)`, 
which is a form of resource leak.

Prefer `fclose(f)` unconditionally.

In term of testings, I suspect a static analyzer such as `scanbuild` could have been able to detect such leak.

tests/paramgrill.c
GeorgeLu97: createBuffers took in file names. This is fixed in forceAttachDict https://github.com/facebook/zstd/pull/1276/commits/3072fbe786f2c17de3d5746a9e756530f46f7f36

tests/paramgrill.c
```diff
@@ -629,14 +824,143 @@ static BMK_return_t BMK_benchMemInvertible(const buffers_t buf, const contexts_t
...
+        }
+        first_node->next = NULL;
+        first_node->res = w;
+        g_winners = first_node;
+        return 0;
+    }
+
+    while(cur_node->next != NULL) {
+        switch(speedSizeCompare(r, cur_node->res.result)) {
+            case BETTER_RESULT:
```
Cyan4973: in term of naming : 
it feels a bit strange here.

`insertWinner()` objective seems to insert a new candidate if it's better than existing ones ? 
But if the result is `BETTER_RESULT`, it's not inserted, 
while if it's `WORSE_RESULT`, it is inserted.

So it feels strange.
Maybe a few words to explain all this ?

GeorgeLu97: It's just the comparison direction. If an existing winner is better than the result, it is inserted (so our candidate is worse than an existing solution). Is it more intuitive if I flip the comparison?

Cyan4973: I guess so.
There is also `SPEED_RESULT` and `SIZE_RESULT` which seem less obvious to invert.

GeorgeLu97: They are opposite of each other, so they would just be swapped (speed_result means better speed but worse ratio, and size_result is better ratio, worse speed)

PR #1276, opened 2018-08-14
forceAttachDict

GeorgeLu97:
Depends on MultiLvl Branch (Only last 4 commits new)

allow for force attach dictionary
-change ZSTD_compressionParameters struct to paramValues_t struct
time limit works for optimizer
fix turning off asserts

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -118,8 +118,8 @@ Full list of arguments
                 (Lower value will begin with stronger strategies) (Default 90%)
     preferSpeed= / preferRatio=
               : Only affects lvl = invocations. Defines value placed on compression speed or ratio
-                when determining overall winner (default 1 for both, higher = more valued).
-    tries=    : Maximum number of random restarts on a single strategy before switching (Default 5)
+                when determining overall winner (default speed = 1, ratio = 5 for both, higher = more valued).
+    tries=    : Maximum number of random restarts on a single strategy before switching (Default 3)
                 Higher values will make optimizer run longer, more chances to find better solution.
  -P#          : generated sample compressibility 
  -t#          : Caps runtime of operation in seconds (default : 99999 seconds (about 27 hours )) 
--- a/tests/README.md
+++ b/tests/README.md
@@ -105,7 +105,8 @@ Full list of arguments
     t# - targetLength
     S# - strategy
     L# - level
- --zstd=      : Single run, parameter selection syntax same as zstdcli
+ --zstd=      : Single run, parameter selection syntax same as zstdcli. 
+                When invoked with --optimize, this represents the sample to exceed. 
  --optimize=  : find parameters to maximize compression ratio given parameters
                 Can use all --zstd= commands to constrain the type of solution found in addition to the following constraints
     cSpeed=   : Minimum compression speed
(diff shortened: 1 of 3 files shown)


tests/paramgrill.c
```diff
@@ -94,17 +81,153 @@ typedef enum {
...
+    return 0; /* should never happen, stop compiler warnings */
+}
+
+typedef struct {
+    U32 vals[NUM_PARAMS];
+} paramValues_t;
+
+//TODO: unset -> 0?
+static ZSTD_compressionParameters pvalsToCParams(paramValues_t p) {
+    ZSTD_compressionParameters c;
```
Cyan4973: yes, it would be better to `memset()` this structure,
just in case its definition would change in the future,
leading to several members being not initialized.

tests/paramgrill.c
```diff
@@ -240,45 +368,63 @@ static void findClockGranularity(void) {
...
-    }
-    if(paramTarget.strategy > ZSTD_btultra) {
-        DISPLAY("INVALID PARAMETER CONSTRAINTS\n");
-        return 0;
+static int paramValid(paramValues_t paramTarget) {
+    U32 i;
+    for(i = 0; i < NUM_PARAMS; i++) {
+        CLAMPCHECK(paramTarget.vals[i], mintable[i], maxtable[i]);
     }
+    //TODO: Strategy could be valid at 0 before, is that right?
```
Cyan4973: answering the comment : 
yes, it used to be true, `ZSTD_fast` used to be `==0`,
but it changed some time ago, and now `0` is no longer a valid value for this `enum`.

tests/paramgrill.c
```diff
@@ -1245,80 +1381,31 @@ static int sanitizeVarArray(varInds_t* varNew, const int varLength, const varInd
...
-        case slog_ind: ptr->searchLog    += amt; break;
-        case slen_ind: ptr->searchLength += amt; break;
-        case tlen_ind: 
-            ptr->targetLength = tlen_table[MAX(0, MIN(TLEN_RANGE - 1, tlen_inv(ptr->targetLength) + amt))];
-            break;
-        case strt_ind: ptr->strategy     += amt; break;
-        default: break;
-    }
+static void paramVaryOnce(const varInds_t paramIndex, const int amt, paramValues_t* ptr) {
+    ptr->vals[paramIndex] = rangeMap(paramIndex, invRangeMap(paramIndex, ptr->vals[paramIndex]) + amt); //TODO: bounds check.
```
Cyan4973: which bound check ?

GeorgeLu97: It was added in invRangeMap + rangeMap, I just forgot to remove this comment

tests/README.md
```diff
@@ -117,10 +118,9 @@ Full list of arguments
                 (Lower value will begin with stronger strategies) (Default 90%)
     preferSpeed= / preferRatio=
               : Only affects lvl = invocations. Defines value placed on compression speed or ratio
-                when determining overall winner (default 1 for both, higher = more valued).
-    tries=    : Maximum number of random restarts on a single strategy before switching (Default 5)
+                when determining overall winner (default speed = 1, ratio = 5 for both, higher = more valued).
```
Cyan4973: 👍 
These are better default values

PR #1277, opened 2018-08-14
memoTableOptions

GeorgeLu97:
Based on forceAttachDict (Last 3 commits new)

Simplify adding new params
Add option for memoization table to be hash table or to be turned off altogether
Change format of speed/ratio tradeoff on command line

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -113,14 +113,15 @@ Full list of arguments
     dSpeed=   : Minimum decompression speed
     cMem=     : Maximum compression memory
     lvl=      : Searches for solutions which are strictly better than that compression lvl in ratio and cSpeed, 
-    stc=      : When invoked with lvl=, represents percentage slack in ratio/cSpeed allowed for a solution to be considered (Default 99%)
+    stc=      : When invoked with lvl=, represents percentage slack in ratio/cSpeed allowed for a solution to be considered (Default 100%)
               : In normal operation, represents percentage slack in choosing viable starting strategy selection in choosing the default parameters
                 (Lower value will begin with stronger strategies) (Default 90%)
-    preferSpeed= / preferRatio=
-              : Only affects lvl = invocations. Defines value placed on compression speed or ratio
-                when determining overall winner (default speed = 1, ratio = 5 for both, higher = more valued).
+    speedRatio=   (accepts decimals)
+              : determines value of gains in speed vs gains in ratio
+                when determining overall winner (default 5 (1% ratio = 5% speed)).
     tries=    : Maximum number of random restarts on a single strategy before switching (Default 3)
                 Higher values will make optimizer run longer, more chances to find better solution.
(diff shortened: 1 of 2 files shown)


Cyan4973:
Same idea here,
better resync from `dev`,
so that only changes relevant to this PR get displayed.

tests/README.md
```diff
@@ -113,14 +113,15 @@ Full list of arguments
...
     cMem=     : Maximum compression memory
     lvl=      : Searches for solutions which are strictly better than that compression lvl in ratio and cSpeed, 
-    stc=      : When invoked with lvl=, represents percentage slack in ratio/cSpeed allowed for a solution to be considered (Default 99%)
+    stc=      : When invoked with lvl=, represents percentage slack in ratio/cSpeed allowed for a solution to be considered (Default 100%)
               : In normal operation, represents percentage slack in choosing viable starting strategy selection in choosing the default parameters
                 (Lower value will begin with stronger strategies) (Default 90%)
-    preferSpeed= / preferRatio=
-              : Only affects lvl = invocations. Defines value placed on compression speed or ratio
-                when determining overall winner (default speed = 1, ratio = 5 for both, higher = more valued).
+    ratioSpeedTradeoff=#:#
```
Cyan4973: I would simplify this setting to a single value `speedRatio=#`,
which would effectively be the same as `preferRatio=#` and pin `preferSpeed` to `1`.

Should you worry about a potential lack of available values at lower ranges (1,2,3, etc.),
you may allow fractional values.

Also : pay attention to value `0` : what does it mean ? how is it handled ?

GeorgeLu97: Zero just means setting that multiplier to zero, so speed / ratio isn't factored at all when comparing two solutions. I was thinking of making the ratio a single number, but I thought it'd be annoying to add a function to parse doubles just for this parameter.

tests/paramgrill.c
```diff
@@ -1399,41 +1500,41 @@ static int variableParams(const paramValues_t paramConstraints, varInds_t* res,
 /* slight change from old paramVariation, targetLength can only take on powers of 2 now (999 ~= 1024?) */
 /* take max/min bounds into account as well? */
 static void paramVaryOnce(const varInds_t paramIndex, const int amt, paramValues_t* ptr) {
-    ptr->vals[paramIndex] = rangeMap(paramIndex, invRangeMap(paramIndex, ptr->vals[paramIndex]) + amt); //TODO: bounds check. 
+    ptr->vals[paramIndex] = rangeMap(paramIndex, invRangeMap(paramIndex, ptr->vals[paramIndex]) + amt);
 }
 
 /* varies ptr by nbChanges respecting varyParams*/
-static void paramVariation(paramValues_t* ptr, const varInds_t* varyParams, const int varyLen, const U32 nbChanges)
+//TODO: get this to work with playAround()
```
Cyan4973: is this comment still valid ?

GeorgeLu97: no, I'll delete it

Issue #1081, opened 2018-03-29
mention DCMAKE_BUILD_TYPE in documentation


pdknsk:
I think [`DCMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) should be mentioned in the documentation.
```Bash
$ cmake ../build/cmake/ -DCMAKE_BUILD_TYPE=Release
```
`zstd` is significantly slower otherwise.

PR #1278, opened 2018-08-14
Cmake now builds with CMAKE_BUILD_TYPE=Release by default

edenzik:
Both invocation from from the main Makefile (via `cmakebuild`) or directly from
the `build/cmake` directory. Suggested by @pdknsk (#1081).

Changes to `.gitignore` to exclude additional autogenerated files produced by `cmake`.


Cyan4973:
Thanks @edenzik !

pitrou:
Apologies for being dense, but how do I get a debug build after this change? Passing `-DCMAKE_BUILD_TYPE=Debug` doesn't seem to have any effect.

veprbl:
I agree with @pitrou. Hardcoding `CMAKE_BUILD_TYPE` is a terrible antipattern in CMake.

Cyan4973:
I guess `CMAKE_BUILD_TYPE` should only be set if it's not already set.

Patch welcomed.

Note that `zstd` projects officially maintains the `Makefile` only.
In contrast, `build/cmake` is community maintained.

pitrou:
I'm not a CMake expert, but I would say that if debug builds are the default with CMake, there is no reason for zstd to override that convention.

If we want to keep release by default (why? :-)), it seems we may be able to use a "cache" variable:
https://cmake.org/cmake/help/v3.6/command/set.html#set-cache-entry
https://cgold.readthedocs.io/en/latest/tutorials/variables/cache.html#d

There's more discussion here:
https://stackoverflow.com/questions/46820155/should-i-cache-cmake-build-type/

LLVM does this:
https://reviews.llvm.org/D17398

veprbl:
@pitrou 
Last time I checked the default in cmake is neither Debug nor Release, cmake will just *not* pass `-g` or `-O2` to the compiler.

veprbl:
<offtopic>
Speaking of overrides, from what I understand, `arrow-cpp` also messes with CMAKE_BUILD_TYPE and allows only `Debug` or `Release`, but not empty or `RelWithDebInfo`, `MinSizeRel`. So it techically has the same problem :)
</offtopic>

Cyan4973:
It seems something like that may work : 
```cmake
IF( NOT CMAKE_BUILD_TYPE )
   SET( CMAKE_BUILD_TYPE "Release" )
ENDIF()
```

pitrou:
@veprbl Feel free to open an Arrow issue :-)

pitrou:
> It seems something like that may work :

Unfortunately it doesn't seem to set optimization flags. I think we need to force the user to pass the right setting.

PR #1280, opened 2018-08-15
Better Paramgrill Display

GeorgeLu97:
option to not print certain parameters (--display)
quiet and verbose modes
normal - prints table as before
-q Prints parameters + results whenever a new best is found
-q -q Only prints parameters whenever a new best is found, prints final parameters + results
-q -q -q Only print final parameters + results 
-q -q -q -q Only prints final parameter set in the form --zstd
-v Prints all candidate parameters and results


Cyan4973:
Same issue as usual,
can you sync this PR with `dev` so that only new changes are displayed ?

tests/README.md
```diff
@@ -122,11 +122,14 @@ Full list of arguments
     tries=    : Maximum number of random restarts on a single strategy before switching (Default 3)
                 Higher values will make optimizer run longer, more chances to find better solution.
     memLog    : Limits the log of the size of each memotable (1 per strategy). Setting memLog = 0 turns off memoization 
+ --display=   : which params to display, uses all --zstd parameter names and 'cParams' to display only compression parameters.
```
Cyan4973: It's not too clear for me what this command does.
Could you elaborate a bit more ?

PR #1296, opened 2018-08-23
Store a Copy of the `ZSTD_compressionParameters` in the `ZSTD_matchState_t`

felixhandte:
The `ZSTD_matchState_t` cannot be interpreted or manipulated without knowledge of the compression parameters it has been configured with. As a consequence, pretty much anywhere the match state is passed, a pointer to the cparams had to be passed as well. This changeset includes a copy of the cparams in the match state struct, which allows us to streamline the signatures of a great many functions.

The immediate motivation for this change is to allow the cparams for the working context's match state to differ from the dictionary context's match state. Previously, this would mean adding another cparams argument to the block compressor function signature and its many implementations. With this change, the dict match state's cparams will be accessible via `ms->dms->cParams`, which is much simpler.

Unfortunately, it's not feasible to remove the cparams from the cctx params struct. This means that the applied params are stored twice in the cctx, and must be kept in sync with each other. This changeset includes assertions at appropriate places to check that that is the case.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1060,6 +1060,8 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
     ms->hashTable3 = ms->chainTable + chainSize;
     ptr = ms->hashTable3 + h3Size;
 
+    ms->cParams = *cParams;
+
     assert(((size_t)ptr & 3) == 0);
     return ptr;
 }
@@ -1329,6 +1331,7 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
             dstMatchState->nextToUpdate = srcMatchState->nextToUpdate;
             dstMatchState->nextToUpdate3= srcMatchState->nextToUpdate3;
             dstMatchState->loadedDictEnd= srcMatchState->loadedDictEnd;
+            dstMatchState->cParams      = srcMatchState->cParams;
         }
     }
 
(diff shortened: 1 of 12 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1257,29 +1274,30 @@ static size_t ZSTD_resetCCtx_usingCDict(ZSTD_CCtx* cctx,
         32 KB, /* ZSTD_btopt */
         8 KB /* ZSTD_btultra */
```
Cyan4973: unrelated to this patch : 
I'm surprised that the cut-off value for `ZSTD_btultra` goes down to 8 KB, significantly lower than previous strategies, and especially lower than `ZSTD_btopt`, which I would expect to be pretty close.
Is that intentional ?

felixhandte: I'd have to dig up the data to find the graphs that supported that choice, but yeah. My rationalization was that btultra does *so many searches* that double searching both contexts gets prohibitively expensive very quickly. I dunno.

Cyan4973:
Is there a need to merge this patch yet ?
Or could it be part of the larger change which 
> allow the cparams for the working context's match state to differ from the dictionary context's match state.
?

felixhandte:
@Cyan4973, there's no motivation for this change other than the separate params project. So in that sense this could certainly be folded in. However, I wanted to put this up and get it reviewed first, to verify that this approach is acceptable before I invest a huge amount of time building on top of it.

If the contents of this patch set are acceptable though, I'm happy to close the PR and eventually just include it in the separate parameters PR when that's ready.

Cyan4973:
I feel the patch quality is acceptable.

It's a pity that data in params struct needs to be duplicated,
but the `assert()` mitigation proposed feels acceptable,
and we could use more time later to eventually find a better solution, if that's possible.

PR #1298, opened 2018-08-25
Refactored bench.c

Cyan4973:
and fixed associated user programs (`fullbench`, `paramgrill`).

This is a pre-requisite for the multi-dictionaries test tool.

This refactoring is larger than anticipated.
Hopefully, the resulting API, defined in `bench.h`, should prove usable for additional projects.

I'm still not satisfied with the large nb of parameters that some of these functions require.
Improving this aspect will need another refactoring round.

### DIFF
--- a/tests/paramgrill.c
+++ b/tests/paramgrill.c
@@ -241,7 +241,7 @@ static U32 g_noSeed = 0;
 static paramValues_t g_params; /* Initialized at the beginning of main w/ emptyParams() function */
 static double g_ratioMultiplier = 5.;
 static U32 g_strictness = PARAM_UNSET; /* range 1 - 100, measure of how strict  */
-static BMK_result_t g_lvltarget;
+static BMK_benchResult_t g_lvltarget;
 
 typedef enum {
     directMap,
@@ -258,7 +258,7 @@ typedef struct {
 } memoTable_t;
 
 typedef struct {
-    BMK_result_t result;
+    BMK_benchResult_t result;
     paramValues_t params;
 } winnerInfo_t;
 
@@ -497,8 +497,11 @@ static void findClockGranularity(void) {
 **************************************/
 
 /* checks results are feasible */
-static int feasible(const BMK_result_t results, const constraint_t target) {
-    return (results.cSpeed >= target.cSpeed) && (results.dSpeed >= target.dSpeed) && (results.cMem <= target.cMem) && (!g_optmode || results.cSize <= g_lvltarget.cSize);
+static int feasible(const BMK_benchResult_t results, const constraint_t target) {
+    return (results.cSpeed >= target.cSpeed)
+        && (results.dSpeed >= target.dSpeed)
+        && (results.cMem <= target.cMem)
+        && (!g_optmode || results.cSize <= g_lvltarget.cSize);
 }
 
 /* hill climbing value for part 1 */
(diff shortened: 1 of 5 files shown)


tests/paramgrill.c
```diff
@@ -1425,130 +1440,112 @@ static BMK_return_t BMK_benchMemInvertible(const buffers_t buf, const contexts_t
...
-    BMK_result_t resultMax, benchres;
+                BMK_benchResult_t* winnerResult, int feas)
+{
+    BMK_benchResult_t resultMax, benchres;
     U64 loopDurationC = 0, loopDurationD = 0;
     double uncertaintyConstantC = 3., uncertaintyConstantD = 3.;
     double winnerRS;
     /* initial benchmarking, gives exact ratio and memory, warms up future runs */
-    CBENCHMARK(1, benchres, tmp, BMK_both, BMK_iterMode, 1);
+    CBENCHMARK(1, benchres, tmp, BMK_both, 1);
```
terrelln: Did you find that this slows down paramgrill? I asked George if the complexity of iter mode was worthwhile, and he said it got a 3x? speedup.

tests/paramgrill.c
```diff
@@ -1589,33 +1586,33 @@ static int allBench(BMK_result_t* resultPtr,
         return WORSE_RESULT;
     }
 
-    CBENCHMARK(loopDurationC < TIMELOOP_NANOSEC, benchres, tmp, BMK_compressOnly, BMK_timeMode, 1);
-    CBENCHMARK(loopDurationD < TIMELOOP_NANOSEC, benchres, tmp, BMK_decodeOnly,   BMK_timeMode, 1);
+    CBENCHMARK(loopDurationC < TIMELOOP_NANOSEC, benchres, tmp, BMK_compressOnly, 1);
```
terrelln: Do we need both of these runs? Line 1572 and line 1589.

Cyan4973: Looking at these lines, I suspect these tests are no longer active,
because `loopDurationC` and `loopDurationD` should no longer be `< TIMELOOP_NANOSEC / 10`,
since all tests use `BMK_timeMode` now.

I could add an assert() to ensure it, and then remove these lines.

tests/paramgrill.c
Cyan4973: I did not checked `paramgrill` speed,
just verified that `paramgrill` was still running properly.

I suspect `BMK_iterMode` can be faster on small data sets,
since in this case it's possible to complete a compression / decompression test in a fraction of a second.
(I've been using `silesia.tar`, for which such conclusion is less true).

But a consequence of faster tests
would be an increased jitter in speed measurements, hence less reliable hill climbing.
That's why the initial code enforced a "1 second minimum per measurement" rule. If I do remember properly, it would even request a "2 second measurement period" (by default), meaning that, on average, 2 measurements will be taken, and the better one will be preserved. The goal of this policy is to reduce jitter to acceptable levels.

In George's code, jitter is taken into consideration by calculating `uncertaintyConstantX`, though it's limited to measured clock granularity. This is nonetheless much better than nothing.

Maybe we could use that capability to justify reducing measurement times ?

If that matters, I would suggest to keep using the `BMK_timeMode`, but make measurement time programmable (instead of the hard-coded "1 second per measurement"). We could then enforce a different (faster) policy within `paramgrill`.

As mentioned before, this is only going to be useful for short data sets,
and cost a bit of accuracy, hence potentially impact hill climbing.
The hill-climbing side effect is non-trivial to measure, and would require some dedicated tools.

Cyan4973:
`assert()` proved that additional tests are never triggered,
since `timeMode` guarantees that each measurement receive a minimum amount of time.

Still to be investigate : convergence speed, and comparison of convergence speed.
I've been looking again at the impact `uncertaintyConstantX`, and I actually don't understand how it works. So I'm not sure that this will actually help counter balance measurement jitter.

PR #1299, opened 2018-08-28
Fixed memory leak detected by cppcheck

modbw:
cppcheck (which is run regularly in our CI environment)  detected a possible memory leak.


Cyan4973:
Yes, this is a legit warning for a static analyzer.
Thanks for the patch @modbw !

PR #1302, opened 2018-08-30
 Always use splitPoint=1.0 for non-optimize cover and fastcover and some minor changes on comments

jennifermliu:
1) I changed the split point to use the default split point if not provided even for non-optimize cover and fastcover in the last pull request, but it seems like it doesn't make sense to have a split point since non-optimize doesn't use any test samples. The default split point is 0.75 for fastcover, so we are ignoring those 25% samples completely for non-optimize fastcover. I think we should probably make it clear that split point is only used for optimize cover and fastcover and set it to 1.0 for non-optimize. 

2) The default steps for both cover and fastcover should be 40 instead of 32. I changed that in zdict.h.

3) Updated some comments on default dictionary builder.


terrelln:
Nice catch!

Issue #1305, opened 2018-09-03
Zstd Spec: potential misleading statement regarding direct Huffman Tree representation


ulikunitz:
The Zstd Compression Format specification contains following two sentences in the description of the direct Huffman Tree Representation:

> Note that maximum `Number_of_Symbols` is 255-127 = 128. If any literal has a value > 128, raw header mode is not possible. In such case, it's necessary to use FSE compression.

I believe the value `Number_of_Symbols` is wrongly named and should be `Number_of_Weights`. The  number of symbols is actually one higher than the number of weights, because the weight for the last symbol will be reconstructed.  At least the educational decoder does this. So according to my understanding the second sentence should be something like this:

> If any literal has a value > 129, raw header mode is not possible. (The number of encoded symbols  is one higher than number of weights encoded in the Huffmann Tree representation, because the weight of the last symbol will be reconstructed.)

Cyan4973:
Thanks for looking into it @ulikunitz .

I believe you are correct on both counts : 
- `Number_of_Symbols` can be confusing, correctly interpreting it depends on taking in consideration local context. `Number_of_Weights` would be better, indeed, since it's specific enough to only apply to this paragraph.
- On alphabet size, you are also totally right. Since the last weight is not present in the compact representation, the upper limit can be increased by 1. Hence, the direct representation is compatible with alphabet sizes of up to 129 symbols.

I'll update the specification to reflect your insightful comments.

Cyan4973:
Proposed a spec update at : 
https://github.com/facebook/zstd/pull/1307
following your suggestions.
You're welcomed to comment.

PR #1307, opened 2018-09-05
spec clarification

Cyan4973:
following #1305 comments from @ulikunitz


ulikunitz:
I agree with the changes. They are much more clear right now.

PR #1310, opened 2018-09-07
 reduce DDict size by 2KB

Cyan4973:
by removing `workspace`
which is only needed while building huffman table.

`workspace` is now either present within `DCtx`,
or replaced by temporary use of available FSE table space.


lib/decompress/zstd_decompress.c
```diff
@@ -114,7 +114,6 @@ typedef struct {
     ZSTD_seqSymbol OFTable[SEQSYMBOL_TABLE_SIZE(OffFSELog)];
```
terrelln: Can we add a comment here saying that all data up to the `hufTable` may be overwritten when constructing the huffman table.

Cyan4973: That's only true when building `DDict`,
because the content of these tables is not yet initialized, 
and is ensured, by construction, to be initialized after huffman.

However, this assumption is no longer correct in a "normal" `DCtx` context,
as FSE tables might contain valid distributions, to be re-used in next blocks.

PR #1311, opened 2018-09-07
Print a stack trace on unexpected term signal (e.g. SIGABRT)

cmcginty:
For OSX and Linux, add a signal handler to SIGABRT, SGIFPE, SIGILL,
SIGSEGV, and SIGBUS. When the program terminates unexpectedly the
handler will print the current stack to the terminal to help determine
the location of the failure.

On OSX the output will look like:

```
Stack trace:
4   zstd                                0x000000010927ed96 main + 16886
5   libdyld.dylib                       0x00007fff767d1015 start + 1
6   ???                                 0x0000000000000001 0x0 + 1
```

On Linux the output will look like:

```
Stack trace:
./zstd() [0x4b8e1b]
./zstd() [0x4b928a]
./zstd() [0x403dc2]
/lib64/libc.so.6(__libc_start_main+0xf5) [0x7f5e0fbb0445]
./zstd() [0x405754]
```

As is, the code does not function on WIN32.

See also: https://oroboro.com/stack-trace-on-crash/


Cyan4973:
Thanks for attempting this difficult patch @cmcginty !

I've spent some time testing it on both OS-X and Linux,
and it works as advertised. But I've 2 issues worth mentioning : 

- This functionality is expected to be provided as part of the CLI program, but not as part of the library. All associated code should therefore be present in `/programs` only.
- Traces on Mac are fine : they list functions in correct nesting order. Here is an example, with a deliberately introduced `free()` error in `ZSTD_compressEnd()` : 

```
zstd(42696,0x7fffb42d3380) malloc: *** error for object 0x7fc9f9502860: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Stack trace:
4   libsystem_malloc.dylib              0x00007fff7b970822 free + 521
5   zstd                                0x000000010fd478c7 ZSTD_compressEnd + 263
6   zstd                                0x000000010fd49738 ZSTD_compressStream_generic + 360
7   zstd                                0x000000010fd4a24e ZSTD_compress_generic + 1198
8   zstd                                0x000000010fe2e290 FIO_compressZstdFrame + 752
9   zstd                                0x000000010fe2de48 FIO_compressFilename_internal + 184
10  zstd                                0x000000010fe2ac9b FIO_compressFilename_srcFile + 219
11  zstd                                0x000000010fe29fda FIO_compressFilename_dstFile + 282
12  zstd                                0x000000010fe2a7e4 FIO_compressMultipleFilenames + 1668
13  zstd                                0x000000010fe25c4e main + 9214
14  libdyld.dylib                       0x00007fff7b7c6015 start + 1
15  ???                                 0x0000000000000002 0x0 + 2
Abort trap: 6
```

This is great.
But Linux output, in contrast, is fairly unexpressive : 

```
*** Error in `./zstd': double free or corruption (!prev): 0x0000000000897060 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f60f65c97e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7f60f65d237a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7f60f65d653c]
./zstd[0x4190e4]
./zstd[0x41ae79]
./zstd[0x41b986]
./zstd[0x4cbe71]
./zstd[0x4cc57d]
./zstd[0x4cc81b]
./zstd[0x4cca46]
./zstd[0x4cd23f]
./zstd[0x4c73aa]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f60f6572830]
./zstd[0x402069]
```

We are missing function names, which is critical for debugging.
(and Linux is unfortunately the platform where we'll need this feature most, typically associated with `Qemu` tests).

The [reference website](https://oroboro.com/stack-trace-on-crash/) mentioned something about "name demangling", which I would expect to be a C++ concept, but might (?) actually prove useful for C code too. The example stack trace is : 

```
Caught signal 6 (SIGABRT)
stack trace:
/lib/libc.so.6 ( abort + 0x180 ) [0x7f1b6987ff70]
/lib/libc.so.6 ( __assert_fail + 0xf1 ) [0x7f1b698762b1]
./bin/linux-x86_64/debug/kxf_test ( ) [0x435a5d]
./bin/linux-x86_64/debug/kxf_test ( testPath() + 0x29b4) [0x415748]
./bin/linux-x86_64/debug/kxf_test ( doTest(char const*, void (*)()) + 0xc8 ) [0x424ce9]
./bin/linux-x86_64/debug/kxf_test ( main + 0xe4 ) [0x424e42]
/lib/libc.so.6 ( __libc_start_main + 0xfd ) [0x7f1b69869c4d]
./bin/linux-x86_64/debug/kxf_test ( ) [0x412cd9]
```

As one can see, function names are present, inside parenthesis.

Would it be possible to reach the same level of functionality ?

cmcginty:
@Cyan4973 Do you have time to review this?

cmcginty:
I think the issue is with this comment in https://www.gnu.org/software/libc/manual/html_node/Backtraces.html

`Currently, the function name and offset can only be obtained on systems that use the ELF binary format for programs and libraries. ... Also, you may need to pass additional flags to the linker to make the function names available to the program. (For example, on systems using GNU ld, you must pass -rdynamic.)`

Cyan4973:
Indeed, as suggested, when adding `-rdynamic` to linker flags, the resulting backtrace is better.
On the same example :  
```
*** Error in `./zstd': double free or corruption (!prev): 0x0000000001a9b060 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f0b9da427e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7f0b9da4b37a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7f0b9da4f53c]
./zstd(ZSTD_compressEnd+0xf7)[0x421b84]
./zstd(ZSTD_compressStream_generic+0x189)[0x423919]
./zstd(ZSTD_compress_generic+0x4e0)[0x424426]
./zstd[0x4d4911]
./zstd[0x4d501d]
./zstd[0x4d52bb]
./zstd[0x4d54e6]
./zstd(FIO_compressMultipleFilenames+0x591)[0x4d5cdf]
./zstd(main+0x2251)[0x4cfe4a]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f0b9d9eb830]
./zstd(_start+0x29)[0x40ab09]
```

It's less complete than the OS-X backtrace, but that's nonetheless an improvement, several function names are present, and most importantly, the leaf one.

The `-rdynamic` flag got GNU `ld` likely deserves a code comment, at least.

So the only remaining comment is about the location of the PR. `/lib` is a lot more conservative in term of adding features, so let's stage first through `/programs`.

lib/common/zstd_common.c
```diff
@@ -15,10 +15,17 @@
 ***************************************/
 #include <stdlib.h>      /* malloc, calloc, free */
 #include <string.h>      /* memset */
+#include <stdio.h>       /* fprintf(), stderr */
```
Cyan4973: `<stdio.h>` : could you remove this dependency ?

lib/common/zstd_common.c
```diff
@@ -15,10 +15,17 @@
...
 #include "error_private.h"
 #include "zstd_internal.h"
 
 
+/*-************************************
+*  Display Macros
+**************************************/
+#define DISPLAY(...)         fprintf(stderr, __VA_ARGS__)
```
Cyan4973: Could you remove this macro ?

cmcginty:
Added the requested changes and did some more digging. The missing symbols on the Linux side seem to be limited to the static functions. Interestingly using `nm zstd` on the binary indicates the symbols are present, yet the `bactrace_symbols` method can not resolve them.

Researching this I stumbled onto a debugging library that can be preloaded and will inject a debug signal handler (if none exists). It can be run as so:

```
env SEGFAULT_SIGNALS="abrt segv" LD_PRELOAD=/lib64/libSegFault.so ./zstd
```
This provides more output including the CPU registers and memory map. However the backtrace it provides is also missing the static symbol names. So there is no gain vs the current code.

Another option I found was a [libbacktrace](https://github.com/gcc-mirror/gcc/tree/master/libbacktrace) which is include with gcc. I didn't dig too far into this but it basically includes a standalone elf decoder and symbol matcher.  There might be some code to leverage from this package or directly integrate the lib into zstd if it is detected in the environment.

Cyan4973:
Awesome investigation @cmcginty !

programs/Makefile
```diff
@@ -48,6 +48,7 @@ DEBUGFLAGS+=-Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
             -Wvla -Wformat=2 -Winit-self -Wfloat-equal -Wwrite-strings \
             -Wredundant-decls
 CFLAGS  += $(DEBUGFLAGS) $(MOREFLAGS)
+LDFLAGS += -rdynamic
```
Cyan4973: I'm unsure of the potential side effect of enabling this flag _always by default_.
I would suggest to keep it for debugging purpose,
hence document it (typically in `programs/README.md`).

programs/fileio.c
```diff
@@ -157,7 +159,60 @@ static void clearHandler(void)
...
+      case SIGABRT: name = "SIGABRT"; break;
+      case SIGFPE: name = "SIGFPE"; break;
+      case SIGILL: name = "SIGILL"; break;
+      case SIGINT: name = "SIGINT"; break;
+      case SIGSEGV: name = "SIGSEGV"; break;
+      default: name = "UNKNOWN";
+   }
+
+   DISPLAY("Caught %s signal, printing stack:\n", name);
+   // Retrieve current stack addresses.
```
Cyan4973: For better portability, keep comment style to `/* ... */`.

programs/fileio.c
```diff
@@ -157,7 +159,60 @@ static void clearHandler(void)
...
+   addrlen = backtrace(addrlist, sizeof(addrlist) / sizeof(void*));
```
Cyan4973: minor : I would have selected `sizeof(addrlist) / sizeof(*addrlist)` for automatic type adaptation

programs/Makefile
cmcginty: The only issue I am aware of would be more data in the symbol table. I will modify to use the `DEBUGFLAGS` option so it does not impact the `zstd-release` target. How would you want this documented?

Cyan4973: Even `DEBUGFLAGS` is too general, I'm afraid. 
`DEBUGFLAGS` is enabled for anything _but_ the `zstd-release` target.
However, `-rdynamic` is not compatible with MinGW for example. There could be more examples.
So automatically enabling this flag would require some advanced detection mechanisms, which would make this PR more complex. 

I believe the simpler solution would be to let this flag under user control. But it needs to be documented, for the user to be aware of it.

The way I see it, `programs/README.md` is the right place to document this flag, since it impacts the build process.
However, a paragraph must be added to document the new backtrace feature.

Cyan4973:
Thanks @cmcginty ! Great PR !

Cyan4973:
Note that excluding Windows might not be enough.

Roaming Internet, I found comments such as : 
> Including execinfo.h fails compilation on Solaris 

So, instead of excluding certain platforms, it might be better to allow only those for which we know this include will work.

Cyan4973:
https://github.com/facebook/zstd/pull/1331

PR #1314, opened 2018-09-12
[libzstd] Fix cpu for MSFT ARM

### LINKED ISSUE #1313: Cannot build for Windows ARM
The code makes use of `__cpuid` and `__cpuidex` when MSVC is detected, but these are [not available for the ARM architecture](https://docs.microsoft.com/en-us/cpp/intrinsics/cpuid-cpuidex?view=vs-2017#requirements)
https://github.com/facebook/zstd/blob/3b53bfe4f3f0a9bcc4f9be014b83896a370ba5ff/lib/common/cpu.h#L39-L54

terrelln:
The `__cpuid()` and `__cpuidex()` intrinsics are only available on x86 and x86_64.

Fixes #1313.

### DIFF
--- a/lib/common/cpu.h
+++ b/lib/common/cpu.h
@@ -36,7 +36,7 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
     U32 f1d = 0;
     U32 f7b = 0;
     U32 f7c = 0;
-#if defined(_MSC_VER) && defined(_M_X64)
+#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))
     int reg[4];
     __cpuid((int*)reg, 0);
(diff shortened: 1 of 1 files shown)


lib/common/cpu.h
```diff
@@ -36,7 +36,7 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
     U32 f1d = 0;
     U32 f7b = 0;
     U32 f7c = 0;
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && defined(_M_X64)
```
Cyan4973: The comment says that `__cpuid()` function is available for both x86 and x64,
but the test only check for x64.
Is that intentional ?

PR #1315, opened 2018-09-12
Add build define to manually enable -rdynamic LD flag

cmcginty:


Cyan4973:
Excellent, that's exactly what's needed ! Thanks !

PR #1317, opened 2018-09-13
Independent Dictionary and Working Context Table Logs

felixhandte:
This PR includes two broad changes.

## ZSTD_compressionParameters in ZSTD_matchState_t

(I.e., this PR includes a resubmission of #1296.)

The `ZSTD_matchState_t` cannot be interpreted or manipulated without knowledge of the compression parameters it has been configured with. As a consequence, pretty much anywhere the match state is passed, a pointer to the cparams had to be passed as well. This changeset adds a copy of the cparams in the match state struct, which allows us to streamline the signatures of a great many functions.

This change is made here because, without it, the block compressors do not have access to the cdict's compression parameters. The only alternative to this change would be to add another cparams argument to the block compressor function signature and its many implementations. With this change, the dict match state's cparams will be accessible via `ms->dms->cParams`, which is much simpler.

Unfortunately, it's not feasible to remove the cparams from the cctx params struct. This means that the applied params are stored twice in the cctx, and must be kept in sync with each other. This changeset includes assertions at appropriate places to check that that is the case.

## Separate Table Logs

This PR adds support to the block compressors to use different hash and chain logs for the dictionary and working contexts. This allows us, when we are compressing small inputs, to shrink the working context tables, to be more friendly to the CPU cache and memory bandwidth.

For large inputs, the existing copying behavior is retained: the dictionary's tables are copied into the working context.

For small inputs, where we want to attach the dictionary, the working context is reset with parameters derived from the dictionary through `ZSTD_adjustCParams(cdict_cparams, pledgedSrcSize, 0)`, which produces tables appropriately sized for just the input without the dictionary.

Further optimizations (changing the default table parameters for cdicts? refining when we attach vs copy the dict? making resizing the working context cheaper?) should be possible.

## Performance

Compression speed is largely (though not entirely) unaffected in single-context benchmarks, where the working context is permanently the hottest thing in the CPU cache. When memory pressure is added (by rotating through a number of working contexts), compression speed degrades significantly less quickly than the baseline dev branch.

I believe there is only one negative speed consequence of these changes. When compressing with an attached dictionary, the position is now hashed twice, where it used to be hashed once. Otherwise, performance for non-attached scenarios should be identical (and testing so far supports that hypothesis).

Compression ratio is sometimes slightly negatively affected (as expected, since we're shrinking the working tables).

As I generate firm benchmarking data, I will update the PR.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -936,6 +936,18 @@ static U32 ZSTD_equivalentCParams(ZSTD_compressionParameters cParams1,
          & ((cParams1.searchLength==3) == (cParams2.searchLength==3));  /* hashlog3 space */
 }
 
+static U32 ZSTD_equalCParams(ZSTD_compressionParameters cParams1,
+                             ZSTD_compressionParameters cParams2)
+{
+    return (cParams1.windowLog    == cParams2.windowLog)
+         & (cParams1.chainLog     == cParams2.chainLog)
+         & (cParams1.hashLog      == cParams2.hashLog)
+         & (cParams1.searchLog    == cParams2.searchLog)
+         & (cParams1.searchLength == cParams2.searchLength)
+         & (cParams1.targetLength == cParams2.targetLength)
+         & (cParams1.strategy     == cParams2.strategy);
+}
+
 /** The parameters are equivalent if ldm is not enabled in both sets or
  *  all the parameters are equivalent. */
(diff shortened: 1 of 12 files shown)


Cyan4973:
The resulting simplification of compression functions is welcomed.

The added complexity of managing tables with different parameters seems worth it, for better memory frugality and locality. Though a few figures explaining the performance benefits would be great to justify this complexity.

PR #1318, opened 2018-09-14
Don't Search Dictionary Context When Working Context Search Resulted in Mismatch

felixhandte:
This PR causes `ZSTD_fast` and `ZSTD_dfast` in `ZSTD_dictMatchState` mode to actually perfectly mirror the behavior of `ZSTD_extDict` (aka copying the dict ctx into the working ctx). Previously, in DMS mode, these strategies would search the dictionary tables if the working tables didn't have an entry (`matchIndex < prefixLowestIndex`) **or** if the working tables' entry was a mismatch (`MEM_read32(match) != MEM_read32(ip)`). This latter situation caused behavior to differ, since in the copied-dict scenario, the existence of the mismatched entry would have overwritten the dictionary match.

So this PR changes their behaviors to only search the dictionary tables in the first situation, but not the second.

Perf results are forthcoming. I expect this to actually be perf-positive for these strategies, since we will do fewer lookups into the dictionary tables. We'll see.

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -141,16 +141,16 @@ size_t ZSTD_compressBlock_doubleFast_generic(
             goto _match_stored;
         }
 
-        /* check prefix long match */
-        if ( (matchIndexL > prefixLowestIndex) && (MEM_read64(matchLong) == MEM_read64(ip)) ) {
-            mLength = ZSTD_count(ip+8, matchLong+8, iend) + 8;
-            offset = (U32)(ip-matchLong);
-            while (((ip>anchor) & (matchLong>prefixLowest)) && (ip[-1] == matchLong[-1])) { ip--; matchLong--; mLength++; } /* catch up */
-            goto _match_found;
-        }
-
-        /* check dictMatchState long match */
-        if (dictMode == ZSTD_dictMatchState) {
+        if (matchIndexL > prefixLowestIndex) {
+            /* check prefix long match */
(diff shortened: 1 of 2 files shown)


Cyan4973:
The proposed changes look good to me.
Just, I would appreciate a few performance comparison figures when available,
and I guess there is a follow-up to consider with `zstd_lazy.c`

felixhandte:
## Speeds and Ratio Diffs for Compilers and Compression Level

All silesia texts. Benchmarked at 1, 4, 16, 64, and 256 contexts. On Intel Xeon Broadwell. Red is `dev` i.e. b76c8884. Green is `shadow-dict-matches` i.e. b287b8f7 Blue is the ratio PR:dev.

### Clang 7.0

Level 1: all-clang-7 0-00-b287b8f7-b76c8884-01-all

Level 2: all-clang-7 0-00-b287b8f7-b76c8884-02-all

Level 3: all-clang-7 0-00-b287b8f7-b76c8884-03-all

Level 4: all-clang-7 0-00-b287b8f7-b76c8884-04-all

### GCC 7.2

Level 1: all-gcc-7 2-00-b287b8f7-b76c8884-01-all

Level 2: all-gcc-7 2-00-b287b8f7-b76c8884-02-all

Level 3: all-gcc-7 2-00-b287b8f7-b76c8884-03-all

Level 4: all-gcc-7 2-00-b287b8f7-b76c8884-04-all

The benchmark only ran for ~24 hours so it's still fairly noisy. Especially since I was doing other things on the server at the same time. So I don't think the outliers are real.

PR #1319, opened 2018-09-16
Fix clang build

ko-zu:
Cleanup some typo to make clangbuild target working.

Clang-6.0.0 on ubuntu bionic emits errors on the broken doxygen comment.


Cyan4973:
Nice catch ! Thanks @ko-zu !

PR #1321, opened 2018-09-16
'head -c BYTES' is non-portable.

bket:
`tests/playTests.sh` uses `head -c` in a couple of tests to truncate the last byte of a file. The '-c' option is non-portable (not in POSIX). Instead use a wrapper around `dd` (truncateLastByte).

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -48,6 +48,10 @@ fileRoundTripTest() {
     $DIFF -q tmp.md5.1 tmp.md5.2
 }
 
+truncateLastByte() {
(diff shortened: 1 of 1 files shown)


terrelln:
It looks like `stat -f %z $1` doesn't work correctly on Centos. The equivalent command would be `stat -c %s`.

Can we use `wc -c < $1` instead?

bket:
I checked a couple of manpages and I think that `wc -c "$1" ` is the best option. It is supported by Linux, *BSD, and OSX.

Tested with OpenBSD.

Commit has been changed to reflect the above.

Issue #75, opened 2015-11-19
Idea: adaptive compression


annulen:
I need to send large core dumps from embedded device to server, and do it in minimal time. I've found zstd to be optimal solution because it is very fast and has good compression ratio. I'm using it like this:

```
zstd  -c | curl -T - $url
```

where kernel fills stdin of zstd

However, if user has narrow bandwidth for upload, it could be benifical to switch from fast compression to high compression which in my case is 30% more effective. For example, if in first 10 seconds zstd detects that compression throughput is N times larger than write speed (in this case to stdout), it automatically switches high compression and uses it up to the end of input stream.

Would such feature make sense?

Cyan4973:
Yes. I have this idea in mind for quite some time.

I believe it is first necessary to implement multi-threading, in order to achieve IO and compression in parallel. Then it seems possible to check and compare both speeds.

That's not a easy feature, so it will take some time to make it right.

stati64:
We do this with:
1. a monotone list of Pareto Convex Frontier compressor+parameter.
e.g. raw, lz4+256, lz4+32, lz4+2, zstd, lzlib+0, lzlib+3
2. an idle time goal e.g. 5%.
3. 16 step Pulse Width Modulation between adjacent members of  1, e.g.
typedef union{
	u32 level;
	struct{
		u08 smooth2:4;					//smoothing
		u08 blend2:4;						//blending adjacent compressors
		u08 basec2:8;						//base compressor
		u16 fill2:16;						//unused
	}bit2;
}blender;

renevdzee:
I've been thinking about the same thing for about a year and came up with the following algorithm to optimize for the slowest resource in the chain:
Create n+2 threads; n compression threads, 1 input thread and 1 output thread. 
* the input-thread reads data blocks from the input and puts these in an input buffer
* the output-thread reads compressed blocks from an output buffer and writes them to the output
* the compression threads allocate and compress blocks from the inputbuffer, where the compression level selection is based on how full the input and output buffers are. 

When the inputbuffer is below a certain level or the output buffer is near-full a higher compression level is selected. With a full inputbuffer or an empty output buffer a faster compression level, etcetera.

When used over a network connection, whenever there is congestion in the network or on the receiving side (cpu or i/o) the output buffer will fill up and the compression threads will switch to a higher compression ratio. Thus optimizing for highest throughput end-to-end.

regards, René

Cyan4973:
[...]

PR #1327, opened 2018-09-20
Adaptive compression 

Cyan4973:
This patch brings adaptive compression level driven by I/O conditions,
provided by command `--adapt`.
This is a long awaited feature request (#75).

This a pure CLI feature, and implemented into `fileio.c`,
though it also forced the creation or update of some reporting functions inside the library
to properly track MT progresses.

Some notes on current behavior : 

- `--adapt` is currently incompatible with `--single-thread` mode : since it's based on comparing I/O speed with compression speed, both operations must be performed in parallel, hence at least 1 worker (default mode)
- In single worker mode, adaptation works well, in both direction.
   + but adaptation is damn slow when compression level is very high
- In multi-workers mode, accelerating works fine, as well as lightly correcting speed when close to cruise regime
   + however, when starting from a way too slow speed, it's not able to "correct" that and remains stuck at low speed

So it's far from perfect.

Issue is, I suspect a major engine update would be necessary to improve this behavior.

Major issues facing the current mechanism are latency, and I/O bursts.

To smooth I/O bursts, adaptation only happens when at least one job is completed.
But one job is 32 MB by default, so at level 19, it's very slow.

Even then, it takes time before the order to accelerate goes through the pipeline and takes effect, due to existing jobs, stacked buffers and decisions. And a problem is : during that time, statistics will be collected and believe they are measuring speed for the new parameters, resulting in over correction.
So, to tame that, correction is limited to one level per job.

Both issues combined cannot be compensated well at very slow speeds, hence the issue with acceleration when starting from a too low speed.

A better engine would need to : 
- ensure its new parameters are being applied immediately, or almost immediately
   + if necessary, by cancelling or shortening prior jobs stacked in the queue
- ensure it knows which parameters it is measuring
- compensate for the "queue of running jobs" in multi-threading mode, typically by measuring only the speed of the last job, and extrapolating from there.
- even then, it will suffer from I/O bursts, and require some smoothing capabilities

These modifications look complex, and I'm not sure they can be completed on time for next release.

So these leaves us with 2 possibilities : 
- merge the feature as it is today, warn about its limitations, improve later.
- do not merge now, improve the feature first

I suspect solution 1 is preferable.
The current algorithm nonetheless works correctly as long as target speed is "fast enough", which should be the case for most datacenter connections.
It will bring some use cases, provide some insights, give us opportunities to learn to improve the feature more accurately.

Possible follow-up : 
- set a min and max level, to ensure the adaptation mechanism does not drift too far in either direction.

Notes on tests : 
- while the new command `--adapt` can be added to the test suite, verifying that it indeed adapts properly is in another complexity league. Such tests are currently done manually, using typically `pv` and `tests/rateLimiter.py`.


programs/fileio.c
```diff
@@ -812,41 +836,134 @@ FIO_compressZstdFrame(const cRess_t* ressPtr,
...
+
+                    /* check output speed */
+                    if (zfp.currentJobID > 1) {  /* only possible if nbWorkers >= 1 */
+
+                        unsigned long long newlyProduced = zfp.produced - previous_zfp_update.produced;
+                        unsigned long long newlyFlushed = zfp.flushed - previous_zfp_update.flushed;
+                        assert(zfp.produced >= previous_zfp_update.produced);
+                        assert(g_nbWorkers >= 1);
+
+                        if ( (zfp.ingested == previous_zfp_update.ingested)   /* no data read : input buffer full */
```
terrelln: Doesn't this condition suggest that compression is too slow? When would this happen if compression is too fast?

programs/fileio.c
```diff
@@ -812,41 +836,134 @@ FIO_compressZstdFrame(const cRess_t* ressPtr,
...
+                          && (zfp.consumed == previous_zfp_update.consumed)   /* no data compressed : no more buffer to compress OR compression is really slow */
+                          && (zfp.nbActiveWorkers == 0)          /* confirmed : no compression : either no more buffer to compress OR not enough data to start first worker */
```
terrelln: Do we need the previous 2 checks, or is this one sufficient?

programs/fileio.c
Cyan4973: In the general case, yes : if ingestion is blocked, it means input is faster than compression can consume.

But this case is a bit different : we are actually testing a situation where both compression and input are blocked because output speed is too slow.

In which case, it doesn't matter that input is faster : compression should nonetheless slow down in an attempt to match output speed.

I shall add a comment to explain the nature of this test.
Also, thinking about it again, it seems I don't really need to check ingestion rate. The test should work too with just consumption rate and `nbActiveWorkers` (which is a recently added indicator).
I will make more tests to check that hypothesis.

Cyan4973: I kind of remember now that controlling ingestion rate was meant to cover the initial part,
when the compressor is waiting for at least one job to be filled,
and cannot even start as long as the first buffer is not completely filled.

In such a case, `nbActiveWorkers = 0` and `consumptionRate == 0`. But it doesn't mean that output speed is too slow. It's actually the reverse situation : input speed is too slow.

Note that the outcome is still the same : compression must slow down.

But input speed is too slow _compared to what_ ?
While waiting for the first job, it's only compared to the notification update rate. 
We don't know yet if input speed is too slow compared to compression speed.
If initial compression level is set to 19, it might very well be fast enough.
It's just we did not had a chance yet to compare both.

So, in order to avoid triggering a "slow down" order while waiting for the first job to fill up, I added a test looking at ingestion rate.
Since then, test order has been changed, and now this test is behind an `if (zfp.currentJobID > 1)` condition, which means the first job has long been started.

Now, a situation where all buffers are empty, resulting in compression waiting for more input, may still happen in the future : we could imagine a sudden prolonged drop in input speed. In which case, measuring `consumptionRate` and `nbActiveWorkers` will trigger this condition.
But this time, this is really because input is too slow. So we can slow down compression.

Hence this test can be simplified by removing the `ingestionRate` part.

programs/fileio.c
Cyan4973: I believe it's useful to check `consumptionRate`.

`zfp.nbActiveWorkers` is the _current_ nb of active workers, at the moment update is done.
It doesn't tell that "no compression happened" since last update.
Maybe some compression happened and was completed, and at the exact moment of the update, there are no active workers.

Now, one could argue that just the fact that `nbActiveWorkers == 0` even for a short period is ground for a speed reduction, as it proves that compression is currently "waiting for something" (either input or output).

That's correct. In this context, ensuring that, on top of that, no compression actually happened since last update, acts as a kind of confirmation signal, reducing over-reactions.

Both strategies can be defended. I just selected one which is a bit more cautious to take this decision.

X-Ryl669:
I think you should add a `generation` counter that clocks everything. When you trigger an adaptation,  you've to record the actual generation in the request. If the adaptation generation is greater than the global generation, then there is no adaptation possible and you skip adaptation.  When the compression job take a decision it has to ensure it's *actual*, that is, it's the same generation as the global generation.
If it is, it increment the global generation. That way, whatever the timing of the different threads, you'll have, **at worst** on step late, so you'll adapt in all cases. 
This can be done completely atomically, fortunately without stalling anything and should solve the issues you've mentioned.

terrelln:
I agree with landing it as-is. It would be nice if we adjusted down from high levels faster, but I think this will be a useful feature without it. If adjusting down is critical, then releasing will help us get that feedback.

I ran some tests to make sure that we are adapting to about the right compression level:

```
zstd -cq -X some-lage.file | zstd -dq | zstd --adapt -vo /dev/null
zstd --adapt -v -c | zstd -dq | zstd -cq -11 -o /dev/null
```

I'm very happy with the results when the target level isn't too high (say < 10), but once the target level is > 10, the adaptive compression ends up stabilizing a bit lower than the target level (7-9 for level 12). It could be an artifact of the test, like the input is bursty, but that is expected in the real world. This just shows that there is room for improvement, which we already know.

PR #1330, opened 2018-09-21
Enable building zstd on Haiku.

korli:


Cyan4973:
Thanks @korli .

I'm very favorable to enable `zstd` on `Haiku`. I note that this specific patch makes those simultaneous assumptions : 
- It defines haiku as compatible with `PLATFORM_POSIX_VERSION 200112L` , with no need to check this fact within `unistd.h`.
- It creates an exception for haiku regarding support of `setpriority()`, which is supposed to be part of [posix standard libraries](https://en.wikipedia.org/wiki/C_POSIX_library).

So, which one is it ? Does Haiku need to be declared posix compliant ?

korli:
setpriority() is still absent in Haiku, a patch is pending https://review.haiku-os.org/#/c/haiku/+/78/
It probably should have been merged notwithstanding the problems to avoid having to discriminate like this.
I propose to leave this part out of the patch, so that it's future proof, if it suits you.

Cyan4973:
I agree.

Quick questions, to improve my understanding : 
- Does haiku delivers a `unistd.h` header ?
- If yes, does it define `_POSIX_VERSION` ?
- If yes, what's its value ?

korli:
Haiku includes a `unistd.h` header which defines `_POSIX_VERSION` to `199009L`. It likely needs to be updated.

Cyan4973:
Thanks for feedback @korli .

I'm no expert at posix libraries, as these things tend to be complex,
but it may be that `_POSIX_VERSION = 199009L` is actually correct.
Haiku might be complete with regards to `199009L` posix library specification,
and may be missing just a few symbols with regards to more recent versions, such as `200112L`.

In which case, `_POSIX_VERSION` would be correct, and sufficient to determine appropriate level of library support.

What is missing in the picture is which `_POSIX_VERSION` requires support for `setpriority()`.

Cyan4973:
So, apparently, the posix version which requires presence of `setpriority()` is `200112L`.

Note that the current source code already test presence of macro constant `PRIO_PROCESS`. So in order to fail that test, one would need an implementation which defines the constant, but does not define the function using the constant, which sounds unlikely.

A more probably issue is a posix platform which does not define `sys/ressource.h`, which will break the `#include` statement. That's where the `_POSIX_VERSION` guard is supposed to help.

I can rewrite the test to ensure this condition is tested before the include, which should help Haiku, since it defines `_POSIX_VERSION = 199009L`.

korli:
Patch is now updated. 

`sys/resource.h` provides `PRIO_PROCESS` https://github.com/haiku/haiku/blob/master/headers/posix/sys/resource.h#L55

Cyan4973:
OK, I'll finish off the last details on my side

Cyan4973:
I've been creating a feature branch, [`haikubuild`](https://github.com/facebook/zstd/tree/haikubuild), which is supposed to fix the last compilation issues on Haiku.

Since I'm using a different laptop today, with more storage available, I've downloaded Haiku OS and installed it under a VM. It works fine. The guest OS is surprisingly frugal, so it doesn't cost much resource to run.

I could check that compilation of `haikubuild` works flawlessly (`make`). Performance is also as good as host system.

However, installation process fails (`make install`).
I'm not sure why, and it's likely a consequence of my ignorance of Haiku specifics.
The usual `sudo make install` doesn't work, because Haiku doesn't know `sudo`.
`make install` doesn't work, because "install cannot create regular file".
However, `make install DESTDIR=sometestDirectory` works great,
so it's likely an issue with right managements.

korli:
Right, in the ports build, `make install` is run in a chroot with appropriate paths. That's because the hierarchy `/system` is read-only, being an amalgamation of the presently installed packages at the system level.
Thanks for your help and support!

PR #1340, opened 2018-09-26
Publish artifacts with CircleCI

terrelln:
* Updates CircleCI to use workflows.
  We can now specify any number of test jobs to run in parallel.
* Switch the image to `buildpack-deps:trusty` which is only 500 MB
  instead of 7 GB, so that saves 7 minutes to download it if it isn't
  already cached on the host.
* Publish the source tarball and sha256sum as artifacts for tag builds only.
  Whenever a new tag is created, a build is triggered that runs the tests,
  and then publishes the artifacts to the job.


terrelln:
Nice catch!

I fixed the silent errors (by removing `&&`) and fixed the resulting issues.

PR #1342, opened 2018-09-26
fix : huge (>4GB) chain of blocks

Cyan4973:
Experimental function `ZSTD_compressBlock()` is designed for very small data in mind,
for situations where saving the ~12 bytes of frame header can actually make a difference.

Some systems though may have to deal with data blobs of any size, small and large entangled.
When data is larger than a block (> 128KB), `ZSTD_compressBlock()` cannot compress them in a single round.
That's why it's also possible to compress in multiple rounds.
This generates a chain of compressed blocks.

Some users push this extended capability to the limit, encoding gigantic chains of blocks.
On crossing the 4GB limit, some internal overflow occurs.

This fix moves the existing overflow correction mechanism higher in the call chain,
so that it's applied also to gigantic chains of blocks.

Added a test case in fuzzer.c, which crashes before the fix, and pass now.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2333,6 +2333,7 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
     ZSTD_matchState_t* const ms = &zc->blockState.matchState;
     DEBUGLOG(5, "ZSTD_compressBlock_internal (dstCapacity=%zu, dictLimit=%u, nextToUpdate=%u)",
                 dstCapacity, ms->window.dictLimit, ms->nextToUpdate);
+    assert(srcSize <= ZSTD_BLOCKSIZE_MAX);
 
     if (srcSize < MIN_CBLOCK_SIZE+ZSTD_blockHeaderSize+1) {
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_compress.c
```diff
@@ -2603,6 +2590,20 @@ static size_t ZSTD_compressContinue_internal (ZSTD_CCtx* cctx,
     if (cctx->appliedParams.ldmParams.enableLdm)
         ZSTD_window_update(&cctx->ldmState.window, src, srcSize);
 
+    if (ZSTD_window_needOverflowCorrection(ms->window, (const char*)src + srcSize)) {
```
terrelln: I think we will need to duplicate this logic in both here or `ZSTD_compressBlock()` and in the loop in `ZSTD_compress_frameChunk()`.

I'd rather put it in `ZSTD_compressBlock()`, since I don't think the logic will work as expected when `srcSize` is very large.

Cyan4973:
updated overflow correction,
now duplicated in both `ZSTD_compressBlock()` and `ZSTD_compress_frameChunk()`.

PR #1343, opened 2018-09-27
[zstreamtest] Reduce memory of newapi tests

terrelln:
We could allocate up to 2^28 bytes of memory for the zstdmt buffers when
using 2 threads with window log = 24. Now, we limit it to 2^26 bytes of memor
when not running big tests. It will also reduce the hashlog / chainlog memory.

I chose max window log = 22 since that is the maximum source size when
big tests are disabled. Hopefully this will be enough to reduce or
eliminate the test failures.

### DIFF
--- a/tests/zstreamtest.c
+++ b/tests/zstreamtest.c
@@ -1808,7 +1808,7 @@ static int fuzzerTests_newAPI(U32 seed, U32 nbTests, unsigned startTest,
             }
             {   U64 const pledgedSrcSize = (FUZ_rand(&lseed) & 3) ? ZSTD_CONTENTSIZE_UNKNOWN : maxTestSize;
                 ZSTD_compressionParameters cParams = ZSTD_getCParams(cLevel, pledgedSrcSize, dictSize);
-                static const U32 windowLogMax = 24;
+                const U32 windowLogMax = bigTests ? 24 : 22;
                 if (dictSize)
                     DISPLAYLEVEL(5, "t%u: with dictionary of size : %zu \n", testNb, dictSize);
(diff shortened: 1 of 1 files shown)


Cyan4973:
Apparently, reducing memory budget wasn't enough ?

terrelln:
It looks like reducing the number of times we change the number of workers is enough. All the tests have passed several times. I'm going to run them a few more times to be sure.

terrelln:
I've ran all 4 QEMU tests 5+ times, and haven't seen any failures, so this seems to have at least improved things.

tests/zstreamtest.c
```diff
@@ -1808,7 +1808,8 @@ static int fuzzerTests_newAPI(U32 seed, U32 nbTests, unsigned startTest,
             }
             {   U64 const pledgedSrcSize = (FUZ_rand(&lseed) & 3) ? ZSTD_CONTENTSIZE_UNKNOWN : maxTestSize;
                 ZSTD_compressionParameters cParams = ZSTD_getCParams(cLevel, pledgedSrcSize, dictSize);
-                static const U32 windowLogMax = 24;
+                const U32 windowLogMax = bigTests ? 24 : 20;
```
Cyan4973: note :
while this will limit the amount of memory when custom compression parameters are passed,
it will not affect tests which select a compression level instead, since it's a different path.

I suspect that memory usage blows up in this other path.

terrelln: It is limited to level 4 in the other path, which is also window log = 20, so it should be okay.

PR #1344, opened 2018-09-27
Fixes for GRUB

terrelln:
This is the first batch of fixes for using upstream zstd in GRUB. There may be more, but it is building now, so I wanted to get this batch in now.

* Enable `-Wmissing-prototypes`. This is a generally useful flag, it found a bunch of functions that can be static. The ones that have "hidden" declarations, we now have to explicitly declare it right before the definition.
* Include `stddef.h` in xxhash for `size_t`.
* Rename a variable named `byte` because of a conflicting typedef.

### DIFF
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -321,19 +321,6 @@ size_t FSE_writeNCount (void* buffer, size_t bufferSize,
 /*-**************************************************************
 *  FSE Compression Code
 ****************************************************************/
-/*! FSE_sizeof_CTable() :
-    FSE_CTable is a variable size structure which contains :
-    `U16 tableLog;`
-    `U16 maxSymbolValue;`
-    `U16 nextStateNumber[1 << tableLog];`                         // This size is variable
-    `FSE_symbolCompressionTransform symbolTT[maxSymbolValue+1];`  // This size is variable
-Allocation is manual (C standard does not support variable-size structures).
-*/
-size_t FSE_sizeof_CTable (unsigned maxSymbolValue, unsigned tableLog)
-{
-    if (tableLog > FSE_MAX_TABLELOG) return ERROR(tableLog_tooLarge);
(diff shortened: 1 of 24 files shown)


lib/compress/fse_compress.c
```diff
@@ -321,19 +321,6 @@ size_t FSE_writeNCount (void* buffer, size_t bufferSize,
 /*-**************************************************************
 *  FSE Compression Code
 ****************************************************************/
```
Cyan4973: `FSE_sizeof_CTable()` seems purely removed, without replacement solution ?

lib/compress/zstd_compress.c
```diff
@@ -1531,15 +1531,6 @@ static void ZSTD_reduceIndex (ZSTD_CCtx* zc, const U32 reducerValue)
 
 /* See doc/zstd_compression_format.md for detailed format description */
```
Cyan4973: I'm surprised, I would expect `ZSTD_noCompressBlock()` to be useful ?
Is it not used for non-compressible blocks ?

lib/compress/fse_compress.c
terrelln: Yes, it isn't exposed in `fse.h`, and there are no callers that use a "hidden declaration". There is still the macro.

lib/compress/zstd_compress.c
terrelln: Not currently, it is inlined.

The two differences are we need `lastBlock` support, and it uses `MEM_writeLE32()` since it knows the block isn't empty.

We could extract it out into this function.

terrelln: Oh wow, this function is incorrect for the current format, it puts `bt_raw` in the lowest bit.

Cyan4973: So it's a (very old) pre-v1.0 leftover. 
Good thing this patch can detect that one !

PR #1348, opened 2018-10-02
Fix #1082

### LINKED ISSUE #1082: (un)zstd deletes files it shouldn't
```Bash
$ touch FILE
$ unzstd -f FILE.zst
zstd: FILE.zst is not a regular file -- ignored 
$ ls FILE
ls: cannot access FILE: No such file or directory
```
This happens with any suffix `zstd` knows.

Cyan4973:
Fix #1082 : 
`zstd -f` and `zstd -d -f` do no longer overwrite destination file
when source file does not exist.

This is achieved by reversing operation order within `fileio.c`.
opening source before destination,
thus skipping destination opening where there is a pb with source.

Added associated tests cases.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1914,9 +1914,9 @@ int FIO_decompressMultipleFilenames(const char** srcNamesTable, unsigned nbFiles
             EXM_THROW(72, "Write error : cannot properly close output file");
     } else {
         size_t suffixSize;
-        size_t dfnSize = FNSPACE;
+        size_t dfnbCapacity = FNSPACE;
         unsigned u;
-        char* dstFileName = (char*)malloc(FNSPACE);
+        char* dstFileName = (char*)malloc(dfnbCapacity);
         if (dstFileName==NULL)
             EXM_THROW(73, "not enough memory for dstFileName");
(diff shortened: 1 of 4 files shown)


programs/fileio.c
```diff
@@ -1792,11 +1836,72 @@ static int FIO_decompressFrames(dRess_t ress, FILE* srcFile,
...
+          && strcmp(dstFileName, stdoutmark)  /* special case : don't remove() stdout */
+          ) {
+            FIO_remove(dstFileName);  /* remove decompression artefact; note: don't do anything special if remove() fails */
+        } else {  /* operation success */
+            if ( strcmp(dstFileName, stdoutmark) /* special case : don't chmod stdout */
+              && strcmp(dstFileName, nulmark)    /* special case : don't chmod /dev/null */
+              && transfer_permissions )          /* file permissions correctly extracted from src */
+                UTIL_setFileStat(dstFileName, &statbuf);  /* transfer file permissions from src into dst */
+        }
+        signal(SIGINT, SIG_DFL);
```
terrelln: Delete this call, we already call `clearHandler()` above.

PR #1350, opened 2018-10-03
fixed msan warning

Cyan4973:
on btlazy2 strategy with dictAttach


Cyan4973:
I need to merge that patch to resume fuzzer/resilience tests.

PR #1351, opened 2018-10-04
 portability macro flags updates, for Haiku

Cyan4973:
some non-trivial changes to `platform.h` and `util.h`,
initially related to compilation for Haiku,
but I used this opportunity to make them cleaner
and add some documentation.

Tested on a "Haiku" VM, compilation works fine (while it fails with current `dev` branch).

I noticed several platform tests could be improved
(too harsh conditions, useless exception, etc.)
but I did not dare modifying too many tests just before release.


Cyan4973:
must be merged for relase

PR #1353, opened 2018-10-04
fixed paramgrill wrong assert() conditions

Cyan4973:
and slightly refactored surrounding function.

I believe the formula calculating variance should get a second reviewing round,
as it's not clear if it's correct.


Cyan4973:
must be merged for release

PR #1356, opened 2018-10-06
don't assume __linux__ means __GLIBC__ on BACKTRACES_ENABLE

maxice8:
check for __GLIBC__ instead of __linux__ as musl libc doesn't provide
execinfo.h

fixes compilation on Alpine Linux and Void Linux musl arches.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -21,7 +21,7 @@
 #endif
 
 #if !defined(BACKTRACES_ENABLE) && \
-   (defined(__linux__) || (defined(__APPLE__) && defined(__MACH__)) )
+   (defined(__GLIBC__) || (defined(__APPLE__) && defined(__MACH__)) )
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @maxice8 .

Would it be preferable to test for both ?
Is there a minimum version required for glibc to work with this capability ?

maxice8:
> Would it be preferable to test for both ?

The most common (and more correct than just hardcoding __linux__ or __GLIBC__) way is to check for the existence of the header (autoconf users do AC_CHECK_HEADERS([execinfo.h])), but i have no clue how to do that with makefiles.

>  Is there a minimum version required for glibc to work with this capability ?

__GLIBC__ is a very old macro it has been available since (problably older) than 1997 https://github.com/lattera/glibc/blob/5107cf1d7d27f17c6de68ec15a8e8d9dd5b471c1/include/features.h

PR #1363, opened 2018-10-09
backtrace support compiled with more conditions

Cyan4973:
following #1356,
only enable backtrace compilation on linux+glibc.

Also, disable backtrace by default from "release" compilation,
so that less platforms get impacted by the new requirements.
Can be manually enabled/disabled using `BACKTRACE=1/0`.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -136,7 +136,6 @@ endif
 
 # enable backtrace symbol names for Linux & Darwin
 BACKTRACE ?= 0
-DEBUGFLAGS = -DBACKTRACE_ENABLE=$(BACKTRACE)
 ifeq (,$(filter Windows%, $(OS)))
 ifeq ($(BACKTRACE), 1)
 DEBUGFLAGS += -DBACKTRACE_ENABLE=1
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -134,11 +134,13 @@ else
 LZ4_MSG := $(NO_LZ4_MSG)
 endif
 
-# enable backtrace symbol names for Linux/Darwin
-ALL_SYMBOLS := 0
+# enable backtrace symbol names for Linux & Darwin
+BACKTRACE ?= 0
+DEBUGFLAGS = -DBACKTRACE_ENABLE=$(BACKTRACE)
 ifeq (,$(filter Windows%, $(OS)))
-ifeq ($(ALL_SYMBOLS), 1)
-DEBUGFLAGS_LD+=-rdynamic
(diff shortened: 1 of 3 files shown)


programs/Makefile
```diff
@@ -134,11 +134,13 @@ else
...
-# enable backtrace symbol names for Linux/Darwin
-ALL_SYMBOLS := 0
+# enable backtrace symbol names for Linux & Darwin
+BACKTRACE ?= 0
+DEBUGFLAGS = -DBACKTRACE_ENABLE=$(BACKTRACE)
 ifeq (,$(filter Windows%, $(OS)))
-ifeq ($(ALL_SYMBOLS), 1)
-DEBUGFLAGS_LD+=-rdynamic
+ifeq ($(BACKTRACE), 1)
+DEBUGFLAGS += -DBACKTRACE_ENABLE=1
```
terrelln: This isn't necessary, since it is already covered by line 139, right?

Cyan4973: Actually, line 139 is a left-over and should not be there...

programs/fileio.c
```diff
@@ -167,7 +158,24 @@ static void clearHandler(void)
...
+#  endif /* __has_feature(address_sanitizer) */
+#elif defined(__SANITIZE_ADDRESS__) && !defined(BACKTRACE_ENABLE) /* GCC compiler */
+#  define BACKTRACE_ENABLE 0
+#endif
+
+#if !defined(BACKTRACE_ENABLE)          \
+   && ((defined(__linux__) && defined(__GLIBC__)) \
+       || (defined(__APPLE__) && defined(__MACH__)) )
+#  define BACKTRACE_ENABLE 1
+#endif
```
terrelln: else define `BACKTRACE_ENABLE = 0`.

programs/fileio.c
```diff
@@ -167,7 +158,24 @@ static void clearHandler(void)
...
+
+
+#if defined(BACKTRACE_ENABLE) && (BACKTRACE_ENABLE>=1)
```
terrelln: `#if BACKTRACE_ENABLE`

programs/fileio.c
```diff
@@ -208,7 +216,7 @@ static void ABRThandler(int sig) {
 
 void FIO_addAbortHandler()
 {
-#ifdef BACKTRACES_ENABLE
+#if defined(BACKTRACE_ENABLE) && (BACKTRACE_ENABLE>=1)
```
terrelln: `#if BACKTRACE_ENABLE`

PR #1367, opened 2018-10-11
Enable use of bswap intrinsics in clang

atdt:
Necessary because clang disguises itself as an older (__GNUC_MINOR__ = 2) GCC.

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -39,6 +39,10 @@ extern "C" {
 #  define MEM_STATIC static  /* this version may generate warnings for unused static functions; disable the relevant warning */
 #endif
 
+#ifndef __has_builtin
+#  define __has_builtin(x) 0  /* compat. with non-clang compilers */
+#endif
+
 /* code only tested on 32 and 64 bits systems */
 #define MEM_STATIC_ASSERT(c)   { enum { MEM_static_assert = 1/(int)(!!(c)) }; }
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for the patch @atdt !

Great find, once again. Do you have a tool to detect these lines missed by `clang` ?

Also : could you please rebase your development branch on latest version for `facebook/zstd/dev` ?
I would like to check the (potential) impact on speed.
I expect it to be neutral (which would be fine), but right now, it's negative, simply because it doesn't contain latest changes from `facebook/zstd/dev` branch.

atdt:
Thanks for taking a look!

> Do you have a tool to detect these lines missed by `clang` ?

No; I just grepped the repo for `__GNUC_MINOR__`. There is another opportunity in: https://github.com/facebook/zstd/blob/d28b80e363fc7c76879476fc4c8cfeeeba749f81/lib/common/xxhash.c#L216

..but I wasn't sure whether that should be changed here or in https://github.com/Cyan4973/xxHash.

> Also : could you please rebase your development branch on latest version for `facebook/zstd/dev` ?

Done.

Cyan4973:
Yes, changes to xxhash are better done in https://github.com/Cyan4973/xxHash .

I tested the updated version, and all looks good.
It's performance neutral, as expected.
I believe the `bswap` functions are mostly useful when running on big-endian systems, in order to read or write the bitstream in a compatible manner.

PR #1368, opened 2018-10-11
Moving Code To util.c

rkjain89:
Testing it for the first time.


programs/util.c
```diff
@@ -0,0 +1,654 @@
...
+    unsigned n;
+    for (n=0; n<nbFiles; n++) {
+        U64 const size = UTIL_getFileSize(fileNamesTable[n]);
+        error |= (size == UTIL_FILESIZE_UNKNOWN);
+        total += size;
+    }
+    return error ? UTIL_FILESIZE_UNKNOWN : total;
+}
+
+  #ifdef _WIN32
```
felixhandte: Fix indent.

programs/util.h
```diff
@@ -39,7 +38,7 @@ extern "C" {
 #include "mem.h"          /* U32, U64 */
 
 
-/* ************************************************************
+/*-************************************************************
```
felixhandte: What is the significance of this change?

programs/util.c
```diff
@@ -0,0 +1,654 @@
...
+
+#else   /* relies on standard C (note : clock_t measurements can be wrong when using multi-threading) */
+    typedef clock_t UTIL_time_t;
+    #define UTIL_TIME_INITIALIZER 0
+    UTIL_time_t UTIL_getTime(void) { return clock(); }
+    U64 UTIL_getSpanTimeMicro(UTIL_time_t clockStart, UTIL_time_t clockEnd) { return 1000000ULL * (clockEnd - clockStart) / CLOCKS_PER_SEC; }
+    U64 UTIL_getSpanTimeNano(UTIL_time_t clockStart, UTIL_time_t clockEnd) { return 1000000000ULL * (clockEnd - clockStart) / CLOCKS_PER_SEC; }
+#endif
+
+#define SEC_TO_MICRO 1000000
```
felixhandte: This is still in the header also.

programs/util.c
```diff
@@ -0,0 +1,654 @@
...
+/*-****************************************
+*  Console log
+******************************************/
+int g_utilDisplayLevel;
+
+/*-****************************************
+*  Time functions
+******************************************/
+#if defined(_WIN32)   /* Windows */
+    UTIL_time_t UTIL_getTime(void) { UTIL_time_t x; QueryPerformanceCounter(&x); return x; }
```
felixhandte: I think these should all be un-indented?

programs/util.h
rkjain89: Every other comment was starting like that. Just to make it look the same

Cyan4973:
Clearly, `cmake` script and Visual studio projects will need to be updated too.

For `cmake`, it should be possible to test it locally.

For Visual though, options are more limited. One is to create a Windows box/vm and download a free version of Visual Studio for testings.
One could also manually edit the project file directly. This is more in the "blind fix" category, but it can work, once one understand how they are structured.

build/VS2008/zstd/zstd.vcproj
```diff
@@ -518,6 +518,10 @@
 				RelativePath="..\..\..\lib\common\threading.h"
 				>
 			</File>
+			<File
+				RelativePath="..\..\..\programs\util.c"
```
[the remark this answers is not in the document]
rkjain89: VS2010 was failing without the util.c being present in zstd project file. I added them in both to make them compile.

tests/fuzz/Makefile
```diff
@@ -52,7 +52,7 @@ FUZZ_SRC       := \
 	$(ZSTDCOMMON_SRC) \
 	$(ZSTDCOMP_SRC)
 
-FUZZ_OBJ := $(patsubst %.c,%.o, $(wildcard $(FUZZ_SRC)))
+FUZZ_OBJ := $(patsubst %.c,%.o, $(wildcard $(FUZZ_SRC))) $(PRGDIR)/util.o
```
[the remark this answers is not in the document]
rkjain89: Probably it will work, I haven't tried it.

build/VS2008/zstd/zstd.vcproj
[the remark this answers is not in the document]
Cyan4973: Well, this script if for Visual Studio 2008 specifically, because it uses a different semantic than Visual 2010+.

And indeed, I second @felixhandte, it seems pretty strange to have `util.c` in the "header" section, rather than the "source files" section just above.

Note that the Visual compilation test on Appveyor passes because it only tests VS2015 compilation (and I do not know why, since this test was supposed to run multiple versions of Visual studio in the past, something to look at).

Anyway, bottom line : passing Appveyor test does not prove that VS2008 compilation works.

Cyan4973:
Almost there, one minor change requested,
and then I'll merge the PR right after v1.3.7 release.

PR #1384, opened 2018-10-22
Clarify special case of offset history update

ulikunitz:
If the current sequence has literal length of zero then an offset value
of three is handled in a special manner. While I implemented a golang
decoder I had to consult the educational decoder for clarification on
the update of the offset history in that case. This commit provides the
clarification that the offset value Repeated_Offset1-1 is handled as a
new offset is added to the offset history accordingly.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -919,6 +919,10 @@ This means that when `Repeated_Offset1` (most recent) is used, history is unmodi
 When `Repeated_Offset2` is used, it's swapped with `Repeated_Offset1`.
 If any other offset is used, it becomes `Repeated_Offset1` and the rest are shift back by one.
 
+In the case of an `offset_value` of 3 and the literal length of the current
+sequence is zero the value `Repeasted_Offset1 - 1_byte` is a new offset,
+becoming the lead of the offset history and the first two repeated offsets will
+be shifted back.
 
 Skippable Frames
 ----------------
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -920,7 +920,7 @@ When `Repeated_Offset2` is used, it's swapped with `Repeated_Offset1`.
 If any other offset is used, it becomes `Repeated_Offset1` and the rest are shift back by one.
 
 In the case of an `offset_value` of 3 and the literal length of the current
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for feedback @ulikunitz .

I naively thought that [L902](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#repeat-offsets) was providing this information.
Though even if it does, it's apparently not clear enough.

I believe the documentation is clear regarding what `offset_value_3` means in the context of `literals_length == 0`.
However, it seems that the consequence of it, with regards to "offset update rules", wasn't obvious, and can be misinterpreted ?

doc/zstd_compression_format.md
```diff
@@ -919,6 +919,10 @@ This means that when `Repeated_Offset1` (most recent) is used, history is unmodi
 When `Repeated_Offset2` is used, it's swapped with `Repeated_Offset1`.
 If any other offset is used, it becomes `Repeated_Offset1` and the rest are shift back by one.
 
+In the case of an `offset_value` of 3 and the literal length of the current
+sequence is zero the value `Repeasted_Offset1 - 1_byte` is a new offset,
```
terrelln: `Repeated_Offset1` (you have an extra s).

ulikunitz:
I'm not confused of the value of the offset in the special case. Line 902 makes it clear. My confusion is caused by a sentence in the Offset Update Rules section:

> This means that when `Repeated_Offset1` (most recent) is used, history is unmodified.

In the special case `Repeated_Offset1` is clearly used and therefore I saw a possibility that the history remains unmodified instead of handling the offset as a new value. So I had to check the source code.

As I stated there is no error in the documentation, but a tiny lack of clarity.

PS: I find that the observation that the last offset  will not repeat after an empty (or no) literal in the sequence very clever. xz doesn't handle this case.

Cyan4973:
Clear enough, thanks for feedback @ulikunitz .
We could certainly rephrase this paragraph to reduce opportunities for confusion,
starting with your PR.

(note the minor comment from @terrelln before merge)

doc/zstd_compression_format.md
ulikunitz: You are right, I have to fix it.

PR #1392, opened 2018-10-25
provide consistent spacing to enum field

coetry:
Consistency is important. If we ignore little things like this, we aren't perfecting our craft. Right?

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -473,7 +473,7 @@ typedef struct {
 typedef struct ZSTD_CCtx_params_s ZSTD_CCtx_params;
 
 typedef enum {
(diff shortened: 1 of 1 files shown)


lib/zstd.h
```diff
@@ -473,7 +473,7 @@ typedef struct {
 typedef struct ZSTD_CCtx_params_s ZSTD_CCtx_params;
 
 typedef enum {
-    ZSTD_dct_auto=0,      /* dictionary is "full" when starting with ZSTD_MAGIC_DICTIONARY, otherwise it is "rawContent" */
+    ZSTD_dct_auto = 0,      /* dictionary is "full" when starting with ZSTD_MAGIC_DICTIONARY, otherwise it is "rawContent" */
```
Cyan4973: Could you ensure that the code comments remain vertically aligned ? for the sake of consistency ?

coetry: sure thing!

PR #1395, opened 2018-10-26
created zstd_decompress_block module 

Cyan4973:
isolate all logic associated with block decompression into its own module.

`zstd_decompress.c` is still the main entry module, in charge of 
context creation/destruction, frames, headers, streaming, special blocks, etc.

Only compressed blocks are handled within `zstd_decompress_block.c`.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -510,7 +510,7 @@ static size_t ZSTD_copyRawBlock(void* dst, size_t dstCapacity,
     DEBUGLOG(5, "ZSTD_copyRawBlock");
     if (dst == NULL) {
         if (srcSize == 0) return 0;
-        return ERROR(dstSize_tooSmall);
+        return ERROR(dstBuffer_null);
     }
     if (srcSize > dstCapacity) return ERROR(dstSize_tooSmall);
     memcpy(dst, src, srcSize);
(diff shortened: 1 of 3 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -1761,6 +477,53 @@ size_t ZSTD_findFrameCompressedSize(const void *src, size_t srcSize)
...
+}
+
+
+static size_t ZSTD_copyRawBlock(void* dst, size_t dstCapacity,
+                          const void* src, size_t srcSize)
+{
+    DEBUGLOG(5, "ZSTD_copyRawBlock");
+    if (dst == NULL) dstCapacity = 0;  /* better safe than sorry */
+    if (srcSize > dstCapacity) return ERROR(dstSize_tooSmall);
+    memcpy(dst, src, srcSize);
```
terrelln: Looks like the static analyzer complains that `src` may be `NULL`, so we should add a check.

Cyan4973:
The main point is to improve code clarity.
So if you believe some more changes would fit in the picture and should be present in this PR,
sure, your suggestions are welcomed.

terrelln:
Sorry, I'm just asking if there are any functional changes in this diff that I might miss, since the diff isn't very useful.

Cyan4973:
No there is no functional evolution.
The patch just tries to isolate block compression logic into its own module.

It would have been great if it was also able to isolate the relevant data structures too,
but with the exception of a few tables of constants, it's not happening
This would require more significant changes in the way `ZSTD_DCtx` is defined.
Maybe a later patch ...

PR #1398, opened 2018-10-29
Fix building zstd-pgo

bmwiedemann:
Fix building zstd-pgo,
especially with make -j2 or more


Cyan4973:
Good catch ! Thanks @bmwiedemann !

PR #1408, opened 2018-11-08
Prevent deadlock on malloc() failure.

szreder:
If one of the helper threads executing `COVER_best_finish()` fails to allocate memory, it exits without releasing the mutex. Moreover, if every thread fails on `malloc()` then the main thread sleeps indifinetely on condition variable - that's why `cond_signal()` is also needed.

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -868,6 +868,8 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
         if (!best->dict) {
           best->compressedSize = ERROR(GENERIC);
           best->dictSize = 0;
+          ZSTD_pthread_cond_signal(&best->cond);
+          ZSTD_pthread_mutex_unlock(&best->mutex);
           return;
         }
       }


Cyan4973:
Thanks for the patch @szreder, looks legit to me (but let's wait for @terrelln feedback).

Quick question : how did you find this issue ? Visual code review, or some kind of automated tool ?

szreder:
> Thanks for the patch @szreder, looks legit to me (but let's wait for @terrelln feedback).
> 
> Quick question : how did you find this issue ? Visual code review, or some kind of automated tool ?

TL;DR - I noticed it myself, though I was driven to this place by ThreadSanitizer. Full story below.


I used ZSTD-1.3.5 and had ThreadSanitizer warnings about data race on `best->cond` from two places: `COVER_best_finish()`, specifically:

```
    ZSTD_pthread_mutex_unlock(&best->mutex);
    if (liveJobs == 0) {
      ZSTD_pthread_cond_broadcast(&best->cond); // <---- here
    }
```

and `COVER_best_destroy()`, called from the main thread executing `ZDICT_optimizeTrainFromBuffer_cover()`.

I believe that TSAN was warning me against spurious wakeup which might happen in the following scenario:
- main thread sleeps on condition variable `best->cond` associated with `best->mutex`;
- worker thread unlocks `best->mutex`;
- main thread wakes up spuriously, acquires the `best->mutex`, notices that `best->liveJobs == 0` and proceeds with `COVER_best_destroy()`, thus destroying `best->cond` on which the worker thread wants to send the broadcast.

As I wanted to report/fix this issue by broadcasting first and unlocking later, I noticed that the 1.3.7 version had already reversed the order of operations in `9d6ed9def34fa042291d48637dabc695bfa5d23f`. As an aside: when I read 1.3.5 code, I noticed a few more places in which the mutexes are unlocked first and the CondVar gets signalled later. I didn't analyze whether such places are vulnerable to spurious wakeups. Also, I haven't checked 1.3.7 version for this pattern.

Just as I was calling it a day, I noticed that `COVER_best_finish()` bails out on `malloc()` failure without releasing the mutex. After some poking around and deliberate `malloc()` sabotage, I added CondVar signalling because otherwise the main thread sleeps indefinitely.

Cyan4973:
Nice story @szreder !
We use thread sanitizer for compression / decompression code,
but I don't see we do the same for dictionary generation.
That could be something to add to our automated test system ...

terrelln:
Thanks for the fix @szreder! I double checked the other signals/broadcasts, and the only other one I saw was in `pool.c`, but that looks safe to me. I'll go ahead and change it so there isn't anything to think about. I'll also be adding dictionary generation to the TSAN tests, though it looks like this bug would've still been missed.

PR #1412, opened 2018-11-12
Remove redundant check

vnair81:
Reported by static analysis tool: value being unsigned, both (value) and (value > 0) are equivalent.

zstd-dead-code

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -413,9 +413,7 @@ size_t ZSTD_CCtxParam_setParameter(
         return CCtxParams->forceWindow;
 
     case ZSTD_p_forceAttachDict :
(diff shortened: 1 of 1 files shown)


terrelln:
Thanks for the report @vnair81! It looks like the check is buggy, since it intends for `value` to be signed.

felixhandte:
@vnair81, thanks for reporting!

This is indeed a bug. This was written with the expectation that `value` is an `int`. It intends to distinguish positive, negative, and zero values. Your proposed fix matches the actual behavior, which is wrong. I'll put up a fix in a minute.

Cyan4973:
Thanks for detecting the issue @vnair81 !
We'll be fixing the root cause using @felixhandte PR at #1413

PR #1413, opened 2018-11-12
Fix #1412: Perform Signed Comparison When Setting Attach Dict Param

felixhandte:


lib/compress/zstd_compress.c
```diff
@@ -412,11 +412,12 @@ size_t ZSTD_CCtxParam_setParameter(
         CCtxParams->forceWindow = (value > 0);
         return CCtxParams->forceWindow;
 
-    case ZSTD_p_forceAttachDict :
-        CCtxParams->attachDictPref = value ?
-                                    (value > 0 ? ZSTD_dictForceAttach : ZSTD_dictForceCopy) :
-                                     ZSTD_dictDefaultAttach;
+    case ZSTD_p_forceAttachDict : {
+        ZSTD_dictAttachPref_e pref = (ZSTD_dictAttachPref_e)value;
```
Cyan4973: nit : `const`

PR #1415, opened 2018-11-13
Separating benchfn and benchzstd

Cyan4973:
During experiment, I frequently need to benchmark something.
Up to now, I had to grab `bench.c` and manually modify it so that it fit into the experiment environment.

This patch separates `benchfn.c` from `benchzstd.c`.

`benchfn.c` does no longer drag zstd dependency (it only needs `mem.h`), making it a lot easier to re-use "as is" into any environment.
`benchzstd.c` is a user of `benchfn`, dedicated to zstd compression/decompression speed.

I also used this opportunity to modify `benchfn` API, using structures to replace the long list of parameters required to start benchmarking function. I believe it helps readability.


Cyan4973:
Good point,
I tested `fullbench`, but not `paramgrill`.
I can do that now.

Cyan4973:
Tests ongoing on `core9`, everything seems to work fine

Issue #1249, opened 2018-07-23
The new advanced API should have a simple one-pass function.


terrelln:
Right now, to get the equivalent of `ZSTD_compressCCtx()` you need to write:

```c
size_t ZSTD_compress_generic_onePass(ZSTD_CCtx* cctx, void* dst, size_t capacity, void const* src, size_t size) {
  ZSTD_inBuffer in = {src, size, 0};
  ZSTD_outBuffer out = {dst, capacity, 0};
  ZSTD_CCtx_reset(cctx);
  size_t const ret = ZSTD_compress_generic(cctx, &out, &in, ZSTD_e_end);
  if (ZSTD_isError(ret))
    return ret;
  if (ret != 0) {
    assert(out.pos == out.capacity);
    return ZSTD_error_dstSizeTooSmall;
  }
  return 0;
}
```

I'm not sure I'm happy with the function name though.

Cyan4973:
`ZSTD_compress2()` is the one-pass variant, now part of the advanced API.

PR #1417, opened 2018-11-15
Advanced API

Cyan4973:
This patch creates a "staging area",
with experimental API designed to reach "stable" status by `v1.4.0`.
It mostly concerns a core part of the "advanced API", and necessary dependencies.

There are some questions remaining, which I believe should be addressed before merging this PR : 

- scope : The staged API has been limited to well-known advanced functionalities (sticky dictionaries, compression parameters, frame parameters, multi-threading, long distance mode) for which we have clear usage scenarios.  More advanced ones have been left into "experimental" status. But maybe it's too restrictive ?    
    + For example, there is a very new `--rsyncable` capability, it is currently labelled "experimental" : should it be part of "staging area" instead ?
- Some `enum` values in `ZSTD_cParameter` are branded "experimental", but they are still present in staging area, because an `enum` definition cannot be broken into 2 separate parts.
    + should these `enum` be moved away into "experimental" section ? For example, using `#define` instead ? Note this could cause some side-effects regarding `enum`-related checks, such as `-Wswitch-enum` and `-Wc++-compat`.
- There is no `ZSTD_DCtx_setParameter()` function which would mirror `ZSTD_CCtx_setParameter()`. A key reason is that there is only one parameter (`MaxWindowSize`) on the decompression side, hence a `ZSTD_DCtx_setParameter()` would have a single `enum` value. Moreover, even `ZSTD_CCtx_setParameter()` cannot set all parameters : several parameters require dedicated setter functions, such as dictionaries, and `pledgedSrcSize`. Still, should the list of decoder parameters increase in the future, this will either re-open the topic, or lead to 2 different logics, with decompression side using dedicated setters, rather than a single `ZSTD_DCtx_setParameter()` entry point.
- `ZSTD_?Ctx_reset()` works differently on compression and decompression sides : compression makes it possible to reset parameter and/or session, decompression resets everything only. Should their behavior converge ?
- What about `_simpleArgs()` variant ? are they relevant candidates for "stable" status ?

Some more related points, to be addressed in a separate PR : 

- #1249 : The new advanced API should have a simple one-pass function : and I guess it extends to both compression and decompression sides.
- Function naming : is `ZSTD_compress_generic` good enough ?
    + should this function be merged with `ZSTD_compressStream` instead ?

_Edit_ : 
- `ZSTD_DCtx_reset()` signature and behavior changed to match `ZSTD_CCtx_reset()`
- experimental `enum` are now pushed into experimental section of the API
- `_simpleArgs()` variants are now relegated to experimental section
- `rsyncable` is still part of experimental. We can revisit this choice after v1.3.8 (it can still be part of stable within v1.4.0 if it feels suitable).
- `ZSTD_compress_generic()` is renamed `ZSTD_compressStream2()`
- created `ZSTD_compress2()` as one-pass variant

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -226,6 +226,66 @@ static ZSTD_CCtx_params ZSTD_assignParamsToCCtxParams(
     return ret;
 }
 
+ZSTD_bounds ZSTD_cParam_getBounds(ZSTD_cParameter param)
+{
+    ZSTD_bounds bounds = { 0, 0, 0 };
+
+    switch(param)
+    {
+    case ZSTD_p_compressionLevel:
+        bounds.lowerBound = ZSTD_minCLevel();
+        bounds.upperBound = ZSTD_maxCLevel();
+        return bounds;
+
+    case ZSTD_p_windowLog:
+        bounds.lowerBound = ZSTD_WINDOWLOG_MIN;
+        bounds.upperBound = ZSTD_WINDOWLOG_MAX;
+        return bounds;
+
+    case ZSTD_p_hashLog:
+        bounds.lowerBound = ZSTD_HASHLOG_MIN;
+        bounds.upperBound = ZSTD_HASHLOG_MAX;
+        return bounds;
+
+    case ZSTD_p_chainLog:
+        bounds.lowerBound = ZSTD_CHAINLOG_MIN;
+        bounds.upperBound = ZSTD_CHAINLOG_MAX;
+        return bounds;
+
+    case ZSTD_p_searchLog:
+        bounds.lowerBound = ZSTD_SEARCHLOG_MIN;
+        bounds.upperBound = ZSTD_SEARCHLOG_MAX;
+        return bounds;
+
+    case ZSTD_p_minMatch:
+        bounds.lowerBound = ZSTD_MINMATCH_MIN;
+        bounds.upperBound = ZSTD_MINMATCH_MAX;
+        return bounds;
+
+    case ZSTD_p_targetLength:
+    case ZSTD_p_compressionStrategy:
+    case ZSTD_p_format:
+    case ZSTD_p_contentSizeFlag:
+    case ZSTD_p_checksumFlag:
+    case ZSTD_p_dictIDFlag:
+    case ZSTD_p_forceMaxWindow :
+    case ZSTD_p_nbWorkers:
+    case ZSTD_p_jobSize:
+    case ZSTD_p_overlapSizeLog:
+    case ZSTD_p_rsyncable:
+    case ZSTD_p_enableLongDistanceMatching:
+    case ZSTD_p_ldmHashLog:
+    case ZSTD_p_ldmMinMatch:
+    case ZSTD_p_ldmBucketSizeLog:
+    case ZSTD_p_ldmHashEveryLog:
+    case ZSTD_p_forceAttachDict:
+    default:
+        {   ZSTD_bounds const boundError = { ERROR(parameter_unsupported), 0, 0 };
+            return boundError;
+        }
+    }
+}
+
 #define CLAMPCHECK(val,min,max) {            \
     if (((val)<(min)) | ((val)>(max))) {     \
         return ERROR(parameter_outOfBound);  \
@@ -379,9 +439,9 @@ size_t ZSTD_CCtxParam_setParameter(ZSTD_CCtx_params* CCtxParams,
 
     case ZSTD_p_minMatch :
         if (value>0)   /* 0 => use default */
-            CLAMPCHECK(value, ZSTD_SEARCHLENGTH_MIN, ZSTD_SEARCHLENGTH_MAX);
-        CCtxParams->cParams.searchLength = value;
-        return CCtxParams->cParams.searchLength;
+            CLAMPCHECK(value, ZSTD_MINMATCH_MIN, ZSTD_MINMATCH_MAX);
+        CCtxParams->cParams.minMatch = value;
+        return CCtxParams->cParams.minMatch;
 
     case ZSTD_p_targetLength :
         /* all values are valid. 0 => use default */
@@ -511,7 +571,7 @@ size_t ZSTD_CCtxParam_getParameter(
         *value = CCtxParams->cParams.searchLog;
         break;
     case ZSTD_p_minMatch :
-        *value = CCtxParams->cParams.searchLength;
+        *value = CCtxParams->cParams.minMatch;
         break;
     case ZSTD_p_targetLength :
         *value = CCtxParams->cParams.targetLength;
@@ -698,7 +758,7 @@ size_t ZSTD_checkCParams(ZSTD_compressionParameters cParams)
     CLAMPCHECK(cParams.chainLog, ZSTD_CHAINLOG_MIN, ZSTD_CHAINLOG_MAX);
     CLAMPCHECK(cParams.hashLog, ZSTD_HASHLOG_MIN, ZSTD_HASHLOG_MAX);
     CLAMPCHECK(cParams.searchLog, ZSTD_SEARCHLOG_MIN, ZSTD_SEARCHLOG_MAX);
-    CLAMPCHECK(cParams.searchLength, ZSTD_SEARCHLENGTH_MIN, ZSTD_SEARCHLENGTH_MAX);
+    CLAMPCHECK(cParams.minMatch, ZSTD_MINMATCH_MIN, ZSTD_MINMATCH_MAX);
     ZSTD_STATIC_ASSERT(ZSTD_TARGETLENGTH_MIN == 0);
     if (cParams.targetLength > ZSTD_TARGETLENGTH_MAX)
         return ERROR(parameter_outOfBound);
@@ -721,7 +781,7 @@ ZSTD_clampCParams(ZSTD_compressionParameters cParams)
     CLAMP(cParams.chainLog, ZSTD_CHAINLOG_MIN, ZSTD_CHAINLOG_MAX);
     CLAMP(cParams.hashLog, ZSTD_HASHLOG_MIN, ZSTD_HASHLOG_MAX);
     CLAMP(cParams.searchLog, ZSTD_SEARCHLOG_MIN, ZSTD_SEARCHLOG_MAX);
-    CLAMP(cParams.searchLength, ZSTD_SEARCHLENGTH_MIN, ZSTD_SEARCHLENGTH_MAX);
+    CLAMP(cParams.minMatch, ZSTD_MINMATCH_MIN, ZSTD_MINMATCH_MAX);
     ZSTD_STATIC_ASSERT(ZSTD_TARGETLENGTH_MIN == 0);
     if (cParams.targetLength > ZSTD_TARGETLENGTH_MAX)
         cParams.targetLength = ZSTD_TARGETLENGTH_MAX;
@@ -795,7 +855,7 @@ ZSTD_compressionParameters ZSTD_getCParamsFromCCtxParams(
     if (CCtxParams->cParams.hashLog) cParams.hashLog = CCtxParams->cParams.hashLog;
     if (CCtxParams->cParams.chainLog) cParams.chainLog = CCtxParams->cParams.chainLog;
     if (CCtxParams->cParams.searchLog) cParams.searchLog = CCtxParams->cParams.searchLog;
-    if (CCtxParams->cParams.searchLength) cParams.searchLength = CCtxParams->cParams.searchLength;
+    if (CCtxParams->cParams.minMatch) cParams.minMatch = CCtxParams->cParams.minMatch;
     if (CCtxParams->cParams.targetLength) cParams.targetLength = CCtxParams->cParams.targetLength;
     if (CCtxParams->cParams.strategy) cParams.strategy = CCtxParams->cParams.strategy;
     assert(!ZSTD_checkCParams(cParams));
@@ -808,7 +868,7 @@ ZSTD_sizeof_matchState(const ZSTD_compressionParameters* const cParams,
 {
     size_t const chainSize = (cParams->strategy == ZSTD_fast) ? 0 : ((size_t)1 << cParams->chainLog);
     size_t const hSize = ((size_t)1) << cParams->hashLog;
-    U32    const hashLog3 = (forCCtx && cParams->searchLength==3) ? MIN(ZSTD_HASHLOG3_MAX, cParams->windowLog) : 0;
+    U32    const hashLog3 = (forCCtx && cParams->minMatch==3) ? MIN(ZSTD_HASHLOG3_MAX, cParams->windowLog) : 0;
     size_t const h3Size = ((size_t)1) << hashLog3;
     size_t const tableSpace = (chainSize + hSize + h3Size) * sizeof(U32);
     size_t const optPotentialSpace = ((MaxML+1) + (MaxLL+1) + (MaxOff+1) + (1<<Litbits)) * sizeof(U32)
@@ -829,7 +889,7 @@ size_t ZSTD_estimateCCtxSize_usingCCtxParams(const ZSTD_CCtx_params* params)
     {   ZSTD_compressionParameters const cParams =
                 ZSTD_getCParamsFromCCtxParams(params, 0, 0);
         size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << cParams.windowLog);
-        U32    const divider = (cParams.searchLength==3) ? 3 : 4;
+        U32    const divider = (cParams.minMatch==3) ? 3 : 4;
         size_t const maxNbSeq = blockSize / divider;
         size_t const tokenSpace = WILDCOPY_OVERLENGTH + blockSize + 11*maxNbSeq;
         size_t const entropySpace = HUF_WORKSPACE_SIZE;
@@ -954,7 +1014,7 @@ static U32 ZSTD_equivalentCParams(ZSTD_compressionParameters cParams1,
     return (cParams1.hashLog  == cParams2.hashLog)
          & (cParams1.chainLog == cParams2.chainLog)
          & (cParams1.strategy == cParams2.strategy)   /* opt parser space */
-         & ((cParams1.searchLength==3) == (cParams2.searchLength==3));  /* hashlog3 space */
+         & ((cParams1.minMatch==3) == (cParams2.minMatch==3));  /* hashlog3 space */
 }
 
 static void ZSTD_assertEqualCParams(ZSTD_compressionParameters cParams1,
@@ -966,7 +1026,7 @@ static void ZSTD_assertEqualCParams(ZSTD_compressionParameters cParams1,
     assert(cParams1.chainLog     == cParams2.chainLog);
     assert(cParams1.hashLog      == cParams2.hashLog);
     assert(cParams1.searchLog    == cParams2.searchLog);
-    assert(cParams1.searchLength == cParams2.searchLength);
+    assert(cParams1.minMatch     == cParams2.minMatch);
     assert(cParams1.targetLength == cParams2.targetLength);
     assert(cParams1.strategy     == cParams2.strategy);
 }
@@ -997,7 +1057,7 @@ static U32 ZSTD_sufficientBuff(size_t bufferSize1, size_t maxNbSeq1,
 {
     size_t const windowSize2 = MAX(1, (size_t)MIN(((U64)1 << cParams2.windowLog), pledgedSrcSize));
     size_t const blockSize2 = MIN(ZSTD_BLOCKSIZE_MAX, windowSize2);
-    size_t const maxNbSeq2 = blockSize2 / ((cParams2.searchLength == 3) ? 3 : 4);
+    size_t const maxNbSeq2 = blockSize2 / ((cParams2.minMatch == 3) ? 3 : 4);
     size_t const maxNbLit2 = blockSize2;
     size_t const neededBufferSize2 = (buffPol2==ZSTDb_buffered) ? windowSize2 + blockSize2 : 0;
     DEBUGLOG(4, "ZSTD_sufficientBuff: is neededBufferSize2=%u <= bufferSize1=%u",
@@ -1101,7 +1161,7 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
 {
     size_t const chainSize = (cParams->strategy == ZSTD_fast) ? 0 : ((size_t)1 << cParams->chainLog);
     size_t const hSize = ((size_t)1) << cParams->hashLog;
-    U32    const hashLog3 = (forCCtx && cParams->searchLength==3) ? MIN(ZSTD_HASHLOG3_MAX, cParams->windowLog) : 0;
+    U32    const hashLog3 = (forCCtx && cParams->minMatch==3) ? MIN(ZSTD_HASHLOG3_MAX, cParams->windowLog) : 0;
     size_t const h3Size = ((size_t)1) << hashLog3;
     size_t const tableSpace = (chainSize + hSize + h3Size) * sizeof(U32);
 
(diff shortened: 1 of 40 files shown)


terrelln:
The decompression side also as the `format` parameter that can be used in the set parameter function. In the future, we expect we might have a `nbWorkers` parameter. I can also imagine us potentially wanting to expose prefetching as a parameter.

I think we should stick with the set parameter function.

lib/zstd.h
```diff
@@ -533,452 +465,501 @@ ZSTDLIB_API size_t ZSTD_frameHeaderSize(const void* src, size_t srcSize);
...
+    ZSTD_p_forceAttachDict,  /* Controls whether the contents of a CDict are
+                              * used in place, or whether they are copied into
+                              * the working context.
+                              *
+                              * Accepts values from the ZSTD_dictAttachPref_e
+                              * enum. See the comments on that enum for an
+                              * explanation of the feature.
+                              */
+    /* Question : should rsyncable be categorized as experimental, or be candidate for stable ? */
+    ZSTD_p_rsyncable,        /* Enables rsyncable mode, which makes compressed
```
terrelln: I think this could be a candidate for stable, but its also fine to leave it in experimental for a release or two.

lib/zstd.h
```diff
@@ -1028,496 +1009,591 @@ typedef enum {
...
+*  Frame size functions
+***************************************/
+
+/*! ZSTD_findFrameCompressedSize() :
+ *  `src` should point to the start of a ZSTD encoded frame or skippable frame
+ *  `srcSize` must be >= first frame size
+ *  @return : the compressed size of the first frame starting at `src`,
+ *            suitable to pass to `ZSTD_decompress` or similar,
+ *            or an error code if input is invalid */
+ZSTDLIB_API size_t ZSTD_findFrameCompressedSize(const void* src, size_t srcSize);
```
terrelln: I'm wondering if this should be part of the stable API. It is necessary if you want to use the simple decompression function, but don't know exactly where the compressed data ends. I need this, for example, in GRUB zstd decompression.

lib/zstd.h
```diff
@@ -1028,496 +1009,591 @@ typedef enum {
...
+ */
+typedef void* (*ZSTD_allocFunction) (void* opaque, size_t size);
+typedef void  (*ZSTD_freeFunction) (void* opaque, void* address);
+typedef struct { ZSTD_allocFunction customAlloc; ZSTD_freeFunction customFree; void* opaque; } ZSTD_customMem;
+static ZSTD_customMem const ZSTD_defaultCMem = { NULL, NULL, NULL };  /**< this constant defers to stdlib's functions */
+
+ZSTDLIB_API ZSTD_CCtx*    ZSTD_createCCtx_advanced(ZSTD_customMem customMem);
+ZSTDLIB_API ZSTD_CStream* ZSTD_createCStream_advanced(ZSTD_customMem customMem);
+ZSTDLIB_API ZSTD_DCtx*    ZSTD_createDCtx_advanced(ZSTD_customMem customMem);
+ZSTDLIB_API ZSTD_DStream* ZSTD_createDStream_advanced(ZSTD_customMem customMem);
```
terrelln: I'm also wondering if these deserve to be in the stable API.

We don't expect `customMem` to change, and that is the only thing they do.

lib/zstd.h
```diff
@@ -1028,496 +1009,591 @@ typedef enum {
...
+ZSTDLIB_API size_t ZSTD_checkCParams(ZSTD_compressionParameters params);
+
+/*! ZSTD_adjustCParams() :
+ *  optimize params for a given `srcSize` and `dictSize`.
+ *  both values are optional, select `0` if unknown. */
+ZSTDLIB_API ZSTD_compressionParameters ZSTD_adjustCParams(ZSTD_compressionParameters cPar, unsigned long long srcSize, size_t dictSize);
+
+/*! ZSTD_compress_advanced() :
+*   Same as ZSTD_compress_usingDict(), with fine-tune control over each compression parameter */
+ZSTDLIB_API size_t ZSTD_compress_advanced (ZSTD_CCtx* cctx,
```
terrelln: Nit: there is an extra space before the `(`

lib/zstd.h
```diff
@@ -1028,496 +1009,591 @@ typedef enum {
...
- *           Use ZSTD_CCtx_loadDictionary_byReference() to reference dictionary content instead.
- *           In such a case, dictionary buffer must outlive its users.
- *  Note 4 : Use ZSTD_CCtx_loadDictionary_advanced()
- *           to precisely select how dictionary content must be interpreted. */
-ZSTDLIB_API size_t ZSTD_CCtx_loadDictionary(ZSTD_CCtx* cctx, const void* dict, size_t dictSize);
-ZSTDLIB_API size_t ZSTD_CCtx_loadDictionary_byReference(ZSTD_CCtx* cctx, const void* dict, size_t dictSize);
-ZSTDLIB_API size_t ZSTD_CCtx_loadDictionary_advanced(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, ZSTD_dictLoadMethod_e dictLoadMethod, ZSTD_dictContentType_e dictContentType);
 
+/********************************************************************
+*  Advanced streaming functions
```
terrelln: Can we add that these are soon to be deprecated in favor of the advanced API?

lib/zstd.h
Cyan4973: Associated question :
does moving an `enum` from `experimental` to `stable` increases the minor release number (`v1.4.0` -> `v1.5.0`) everything else being equal ?

lib/zstd.h
Cyan4973: This seems a good reason to upgrade to "stable".

lib/zstd.h
Cyan4973: Yeah, I've been hesitating on that point too.
Ultimately, it did not felt needed: use cases for `customMem` are pretty rare and distant so far, mostly limited to kernel stuff (which can accommodate the static linking condition).

Cyan4973:
There is another detail I'm now wondering about : 
within `ZSTD_CCtx_setParameter(ZSTD_CCtx* cctx, ZSTD_cParameter param, unsigned value)`
note that `value` is of type `unsigned`.

Why `unsigned`, and not `int` ?

I suspect that's because the underlying (private) structures use mostly `unsigned` types for its fields too, presumably as a way to ensure that all values are always `>=0` by design.

But from an interface perspective, is it needed ?

There is actually one case where this choice is detrimental : for compression levels, expressing negative values require a prior cast to `unsigned`.
It's not pretty, but it works.

Enter query functions `ZSTD_CCtx_getParameter()`, `ZSTD_cParam_upperBound()` and `ZSTD_cParam_lowerBound()`, all returning `unsigned` since it's the base type of `value`. How should negative compression levels be expressed now ? It's actually a mess.

So let's look in the other direction : what would go wrong if `value` parameter was of type `int` instead ?
Well, very little it seems.
Apart from occasional casting, which might not be even needed given proper clamping (maybe just some `assert()` for good measure), all values are actually pretty short, so it never happens that we need anything even close to `INT_MAX`.

The only exception is `ZSTD_DCtx_setMaxWindowSize()` : if it's transformed into `ZSTD_DCtx_setParameter()`, then we should probably keep a signature consistent with `ZSTD_CCtx_setParameter()`, in which case, `value ` would have type `int`, and the maximum window size that can be expressed is now `INT_MAX` (`2 GB - 1`) (and probably less if we want some good test margins).

There are several solutions possible, such as introducing a different `enum`, like `windowLogMax`, which can now easily express gigantic values, though at reduced resolution (and is more consistent with the rest of the API, which is mostly concerned by `windowLog`, not `windowSize`).

So apparently using `int` instead of `unsigned` as the type of `value` does not seem to produce bad side effects, but it improves compatibility with negative `compressionLevel` (and potentially some future negative values for new parameters).

My last problem is about signaling an error when using `ZSTD_cParam_upperBound()` and `ZSTD_cParam_lowerBound()` : this can happen if the `enum` provided in argument is bogus. But `zstd` error system is based on `size_t`, so it's not compatible with an `int` return type.
It happens we have the same problem with `ZSTD_getFrameContentSize()`, and the solution was to define a set of macro constants to define errors. Maybe that's the way to go ...

terrelln:
I think switching to an int is good, and setting the maximum window log on the decompression side will work.

We already can't use `size_t ZSTD_cParam_upperBound(ZSTD_cParam)`, since negative values are allowed for some parameters, we can't know for sure if it is a negative value or an error code.

I think the signature should be `size_t ZSTD_cParam_lowerBound(ZSTD_cParam, int* bound)`, or `size_t ZSTD_cParam_bounds(ZSTD_cParam, int* lowerBound, int* upperBound)`, or `size_t ZSTD_cParam_bounds(ZSTD_cParam, ZSTD_bounds* bounds)`

Cyan4973:
Whenever possible, I tend to prefer function signatures which feature this behavior :
`result_type function(read_arguments);`,
they tend to be more usable.

So far, my temporary design to query bounds is : 
`int ZSTD_cParam_lowerBound(ZSTD_cParam);`.
It has the nice advantage that a function invocation can be trivially inserted into expressions (typically for comparison purpose).

This feels more expressive than a close variant : 
`ZSTD_bounds ZSTD_cParam_getBounds(cParam);`
which requires storing result in an intermediate structure and then extract `lowerBound` or `upperBound`
(though it's also usable directly in expression, but more verbose).

The possibility of errors however significantly curtails this argument.
Sure, there is no error whenever `cParam` argument can be guaranteed to be correct, 
but since it is a user-provided argument, there is no way to guarantee "a priori" this condition.
So the function can fail when the `cParam` argument is bogus.
And the function signature must reflect this possibility.

As always, as soon as errors are introduced, the design becomes more complex.

These days, I like to deploy Phantom Types to represent something that may be either a valid answer or an error. Similar to Rust's `Result<T,E>`, this pattern forces error evaluation in order to extract the useful result, eventually defaulting to a simple `panic` (`abort()` in standard C) if the user doesn't want to take care of this case by himself. For shorter codes, both operations can be combined in a single prototype (if `abort()` is deemed an acceptable outcome, which is certainly not the case for a library). It feels clean.

However, there is no phantom type anywhere in the `zstd.h` API, so it would be too weird to introduce that concept just for one set of functions.

This makes `ZSTD_bounds ZSTD_cParam_getBounds(cParam);` more relevant, because it's easy to sneak in a `size_t error;` field into the structure return type. This field can use the same error code convention as the rest of the API (`ZSTD_isError()` notably) . 
(In contrast, `int ZSTD_cParam_lowerBound(ZSTD_cParam);` must create its own specific set of `#define` constants to declare new errors).

This prototype gets in competition with `size_t ZSTD_cParam_bounds(ZSTD_cParam, ZSTD_bounds* bounds);` as you suggest, which is itself closer to an existing prototype : `size_t ZSTD_getFrameHeader(ZSTD_frameHeader* zfhPtr, const void* src, size_t srcSize);`.

aka, the return type is first and foremost for the error code, any non-`size_t` output uses write pointers instead.

2 years ago, I would have made this choice.

Nowadays, I believe that the pattern `result_type function(read_arguments);` makes an API more expressive, and therefore should be preferred whenever applicable. Hence, I feel more in favor of :
`ZSTD_bounds ZSTD_cParam_getBounds(cParam);` .

lib/compress/zstd_compress.c
```diff
@@ -278,6 +304,13 @@ ZSTD_bounds ZSTD_cParam_getBounds(ZSTD_cParameter param)
     case ZSTD_p_ldmMinMatch:
     case ZSTD_p_ldmBucketSizeLog:
     case ZSTD_p_ldmHashEveryLog:
+
```
terrelln: This sets the bounds for all of these parameters to `0` and `1`.

Cyan4973: sorry, the function wasn't complete yet.
It is now.

Cyan4973:
Modified the name of the main advanced compression function,
from `ZSTD_compress_generic()` to `ZSTD_compressStream2()`.

Generally speaking, the new name "feels better" in user source code, blending well into the existing stable API.

I also created the one-pass variant `ZSTD_compress2()`,
which is great simplification, saving many lines of code on the user side.

What's not done : 
both advanced and "stable" API use different configuration planes : 
advanced parameters do not influence `ZSTD_compressStream()`,
and using `ZSTD_initCStream()` does not influence `ZSTD_compressStream2()`.

The new naming seems imply that both `Stream()` and `Stream2()` are close relative though.

Cyan4973:
On the decompression side, situation feels a bit more _weird_ : 
signature of `ZSTD_decompress_generic()`, which can be renamed `ZSTD_decompressStream2()` for consistency, is _exactly_ identical to `ZSTD_decompressStream()`.

In fact, both functions are completely equivalent, and the only reason for `ZSTD_decompressStream2()` to exist is to accept advanced parameters, which `ZSTD_decompressStream()` doesn't (different configuration planes).

But if joining both configuration planes is desirable, then `ZSTD_decompressStream2()` becomes completely redundant.

The situation is the same for a potential "one-pass" variant `ZSTD_decompress2()`, which would be exactly equivalent to `ZSTD_decompressDCtx()` if both accept advanced parameters.

As a consequence, the mere existence of these prototypes is an opened question mark...

Cyan4973:
Some more consistency shenanigans.

Compression parameters can be selected using an enum `ZSTD_cParameter`.
These `enum` parameters follow the naming convention `ZSTD_p_*`.
`p` stands for "parameter".

It's unspecific to either compression or decompression.

This was fine as long as only compression side was supporting a `setParameter()` method.
But since [this comment](https://github.com/facebook/zstd/pull/1417#issuecomment-439228623), it seems preferable to ensure interface consistency, so now the decompression side will need a `ZSTD_DCtx_setParameter()` method.

This method will need an `enum` to select its own parameters, and for consistency, it will be called `ZSTD_dParameter`.

Now what's the convention to name these enum themselves ?

- solution 1  : re-use `ZSTD_p_*`. All parameters, wether compression or decompression, would use the same convention. Sounds reasonable, especially since compilation will catch confusion errors (using a compression parameter with `ZSTD_DCtx_setParameter()` or vice-versa). However, it introduces risk of naming collision, when the same parameter exist on both compression and decompression side. It's not theoretical : there is a first case already with the experimental parameter `ZSTD_p_format`.

- solution 2 : rename all parameters. Compression side will use `ZSTD_c_*`, decompression side will use `ZSTD_d_*`. Feels more consistent. But on immediate term, it will generate a substantial amount of rewrite of existing user code. Not sure how large is current user code base though.

- solution 3 : keep `ZSDT_p_*` for compression, introduce `ZSTD_d_*` for decompression. Reduce pain for user-side code changes. But consistency is less good (why `_p_` for compression ?)

- solution 4 : keep solution 2, plus add some `#define` in `STATIC_ONLY` section to rename all parameters `ZSDT_p_*` into `ZSTD_c_*`. Ease the transition on short term. Also delays the problem. There is also a (small) risk than more users will be invoking `ZSTD_p_*` instead of `ZSTD_c_*`, increasing the size of obsolete code base.

Cyan4973:
By the way,
we have a `ZSTD_CCtx_params`, 
but we don't have a mirror `ZSTD_DCtx_params`.

It's not nearly as much necessary for decompression as it is for compression.
After all, the main justification for `ZSTD_CCtx_params` to exist is `ZSTD_estimateCCtxSize_usingCCtxParams()`. There is no equivalent need on decompression side.

So the main reason would be consistency, which seems too little.

Another potential scenario : 
Let's imagine a situation where decompression contexts `DCtx` are shared through a pool.
As a consequence, each `DCtx` context comes (potentially) with its own set of parameters, depending on previous usage.
Therefore, the only way to safely use such context is to reset all parameters.
In such a setup, it would be easier to have a `ZSTD_DCtx_params` to apply, rather than resetting each parameter one by one.

Note that the current advanced API only has 1 stable parameter and 1 experimental one.
This will likely grow in the future, but the immediate need is quite moot.

Maybe it's better to reconsider this scenario later, when it actually starts to matter....

Cyan4973:
Another consistency issue : 

While trying to make `ZSTD_compressStream()` compatible with `ZSTD_compressStream2()`,
I realized it would also require to deal with `ZSTD_flushStream()` and `ZSTD_endStream()`.

Each of them is basically a different variant of the end directive `ZSTD_e_continue`, `ZSTD_e_flush` and `ZSTD_e_end`.

But there is more.
The return value of these functions has a different meaning.

For `ZSTD_compressStream()`, the return value is an input size hint. It's only a hint, any input size is allowed and work well. Respecting the hint is supposed to help performance (though I've not measured by how much, and it's quite likely to be small). 

For `ZSTD_flushStream()`, it returns the amount of data still present within internal buffer, and therefore still needed to be flushed.

For `ZSTD_endStream()`, it's the same, plus the potential additional budget required to finalize the frame.

So far, for `ZSTD_compressStream2()` return value has same meaning as `ZSTD_flushStream()`, whatever the end directive.

Given that the first 3 functions are already in "stable", it's not possible to change their behavior. So their definitions must hold.

For the newer `ZSTD_compressStream2()`, it opens different perspectives. Namely : 
- Change the meaning of the return value of `ZSTD_compressStream2()`, depending on the end directive, in order to match the 3 existing definitions.
- Keep the definition as it is : `ZSTD_compressStream2()` returns the same value as `ZSTD_flushStream()`.

Solution 1 makes the definition of `ZSTD_compressStream2()` complex. Having a return value which means different things depending on context feels weird, and is likely a source of confusion.

Solution 2 makes it impossible to get an "input size hint" like `ZSTD_compressStream()`, but given that the efficiency of this hint is poor, it's not a big loss. 
Also, wrappers from existing streaming API towards the new advanced one become a bit more complex, because it's no longer possible to just "delegate" the return value to `ZSTD_compressStream2()` : it requires additional calculation. But at least, it's an internal thing, so it doesn't bother users.

Short of a better proposal, it seems that solution 2 is preferable.

terrelln:
I'm in favor of solution 2.

Cyan4973:
At this stage, I've got the regular streaming compression and the new advanced functions compatibles with the new advanced parameters.
On the decompression side, both streaming and one-shot decompression functions are compatible with advanced parameters.

The single-shot compression functions, namely `ZSTD_compressCCtx()`, `ZSTD_compressUsingDict()` and `ZSTD_compressUsingCDict()`, continue using their own independent configuration plane.
I'm still hesitant on this front.

Another opened question is about the `enum` names of advanced parameters, as underlined [in this post](https://github.com/facebook/zstd/pull/1417#issuecomment-443335277). I'm considering version 3 at this stage, even though it's a lot of refactoring work, but solution 2 (currently implemented) seems to lack consistency.

lib/zstd.h
```diff
@@ -392,28 +392,50 @@ ZSTDLIB_API size_t ZSTD_DStreamOutSize(void);   /*!< recommended size for output
...
+
+/* all magic numbers are supposed read/written to/from files/memory using little-endian convention */
+#define ZSTD_MAGICNUMBER            0xFD2FB528    /* valid since v0.8.0 */
+#define ZSTD_MAGIC_DICTIONARY       0xEC30A437    /* valid since v0.7.0 */
+#define ZSTD_MAGIC_SKIPPABLE_START  0x184D2A50    /* all 16 values, from 0x184D2A50 to 0x184D2A5F, signal the beginning of a skippable frame */
+#define ZSTD_MAGIC_SKIPPABLE_MASK   0xFFFFFFF0
+
+/* note : should this limit be smaller ? like 23 (8 MB) as suggested in the spec ?
+ * can it be different from zstd cli default limit, which is designed to match --ultra and --long default (27) ? */
+#define ZSTD_WINDOWLOG_LIMIT_DEFAULT 27   /* by default, the streaming decoder will refuse any frame
```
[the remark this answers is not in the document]
Cyan4973: `ZSTD_dParam_getBounds()` will only provide the limits of `ZSTD_d_windowLogMax`, not its default value.

Therefore, providing such a value dynamically from the library seems to require a dedicated prototype.
It somehow _feel_ overkill but I don't see an escape here. 
The value seems necessary, if only for documentation, so that users can understand how reacts the library on receiving frames with gigantic window sizes.

An alternative could be to create a "generic" get function,
such as `ZSTD_dParam_getDefault()`, 
valid for any `ZSTD_dParameter` (well, there are only 2 currently, and only `ZSTD_d_windowLogMax` matters).

However, note that the equivalent variant on the compression side, `ZSTD_cParam_getDefault()`, does not exist.
I suspect creating one variant on one side would warrant creating the other one, for consistency ?
Obviously, the compression variant is more complex, as it contains a lot more parameters.

And now, replacing this simple constant generates a surprisingly large amount of refactoring ...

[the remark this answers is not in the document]
Cyan4973: Whatever the solution (presence or not of a function to query this value),
the macro constant itself must go into _static_linking_only_ area.

lib/compress/zstd_compress.c
```diff
@@ -3874,32 +4030,50 @@ size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
...
+    size_t const cErr = ZSTD_compressStream2(cctx, &output, &input, endOp);
     *dstPos = output.pos;
     *srcPos = input.pos;
     return cErr;
 }
 
+size_t ZSTD_compress2(ZSTD_CCtx* cctx,
+                      void* dst, size_t dstCapacity,
+                      const void* src, size_t srcSize)
+{
```
terrelln: Should `ZSTD_compress2()` first call `ZSTD_CCtx_reset(cctx, ZSTD_reset_session_only)`?

Cyan4973: Good point !
Yes, I think it is better with a `reset`.
This way, it won't accidentally "continue" a non-finished stream.

lib/compress/zstd_compress.c
```diff
@@ -3874,32 +4030,50 @@ size_t ZSTD_compress_generic (ZSTD_CCtx* cctx,
...
+    size_t oPos = 0;
+    size_t iPos = 0;
+    size_t const result = ZSTD_compressStream2_simpleArgs(cctx,
+                                    dst, dstCapacity, &oPos,
+                                    src, srcSize, &iPos,
+                                    ZSTD_e_end);
+    assert(iPos == srcSize);
```
terrelln: Nit: This `assert()` should be after the `ZSTD_isError()` check.

lib/zstd.h
```diff
@@ -392,593 +402,604 @@ ZSTDLIB_API size_t ZSTD_DStreamOutSize(void);   /*!< recommended size for output
...
+     * ZSTD_c_format
+     * ZSTD_c_forceMaxWindow
+     * ZSTD_c_forceAttachDict
+     * Because they are not stable, it's necessary to define ZSTD_STATIC_LINKING_ONLY to access them.
+     * note : never ever use experimentalParam? names directly
+     */
+     ZSTD_c_experimentalParam1=500,
+     ZSTD_c_experimentalParam2=10,
+     ZSTD_c_experimentalParam3=1000,
+     ZSTD_c_experimentalParam4
```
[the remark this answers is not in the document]
Cyan4973: ok, merely for consistency then, as none of these values is supposed to be "final"

Cyan4973:
Thanks, requested modifications pushed

lib/compress/zstd_compress.c
```diff
@@ -4060,19 +4060,21 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
-        return ERROR(dstSize_tooSmall);
+    ZSTD_CCtx_reset(cctx, ZSTD_reset_session_only);
+    {   size_t oPos = 0;
+        size_t iPos = 0;
+        size_t const result = ZSTD_compressStream2_simpleArgs(cctx,
+                                        dst, dstCapacity, &oPos,
+                                        src, srcSize, &iPos,
+                                        ZSTD_e_end);
+        if (ZSTD_isError(result)) return result;
+        assert(iPos == srcSize);
```
terrelln: Oops, does this also need to go after `result != 0`?

Cyan4973: Indeed !
if result is non-zero, we expect failure due to insufficient `dstCapacity`,
so there is no guarantee that input was entirely consumed !

PR #1418, opened 2018-11-16
Fix i386 build failure "Junk character 13"

ryandesign:
This fixes a build failure on i386 with gcc-4.2 on Mac OS X 10.6 and earlier:

```
.tmp/cc).co70.s:899:Junk character 13 (
.tmp/cc8Vco70.s:899:Rest of line ignored. 1st junk character valued 112 (p).
.tmp/cc).co70.s:2502:Junk character 13 (
.tmp/cc8Vco70.s:2502:Rest of line ignored. 1st junk character valued 112 (p).
```

Thanks to @rmottola.

### DIFF
--- a/lib/common/cpu.h
+++ b/lib/common/cpu.h
@@ -78,7 +78,7 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
       __asm__(
           "pushl %%ebx\n\t"
           "cpuid\n\t"
-          "movl %%ebx, %%eax\n\r"
+          "movl %%ebx, %%eax\n\t"
           "popl %%ebx"
           : "=a"(f7b), "=c"(f7c)
           : "a"(7), "c"(0)


rmottola:
Thank you @ryandesign my preparing the patch to upstream. It was too late this morning to make an official patch :)

Just wanted to add that it is not just gcc-4.2 but also gcc-6.5

Cyan4973:
Thanks for the patch !
I was wondering if we could add a test that would be able to trigger this issue ?
So far, I've been unable to reproduce it (though I'm using more recent versions of OS-X and `gcc`).
`gcc-6` on Linux does not seem affected either.

terrelln:
I ran this command on OS X

```
> gcc-6 --version
gcc-6 (Homebrew GCC 6.5.0) 6.5.0
> CC=gcc-6 make clean libzstd MOREFLAGS="-m32 -O0"
```

I'm certain it is triggering the right code path, but I wasn't able to reproduce the issue either. I am on OS X 10.13 though.

terrelln:
Yeah, I added `#error` to that code path to double check.

terrelln:
Did this only happen on 10.6, or can it be reproduced on a newer OS X?

ryandesign:
I observed the problem on 10.6 i386. Another user observed the problem on 10.5 i386. Xcode on 10.7 and later do not include gcc anymore (they use clang instead) so the problem is not observed there.

PR #1419, opened 2018-11-16
Fix feature detection with multiple -arch flags

ryandesign:
When multiple -arch flags are used, the compiler invokes itself once for
each architecture. Apparently, input on stdin is consumed by the
compilation of the first arch and is no longer available to the
compilation of the second arch, which results in a build failure and the
potentially incorrect determination that a feature is not available. So
write the feature detection source to a file instead of using stdin.

------

To explain the error I saw which this PR fixes:

If I remove the redirection of stderr to /dev/null from the feature detection part of programs/Makefile, I can see that when for example I have `-arch x86_64 -arch i386` in CFLAGS/CXXFLAGS/LDFLAGS, the error is:

```
Undefined symbols for architecture i386:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture i386
```

If I rearrange the flags to `-arch i386 -arch x86_64` then the error is:

```
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64
```

It's always the second listed arch that has the problem. If I use only one arch or the other, there is no error.

I saw the same behavior with Apple clang 902.0.39.2 on macOS High Sierra and Apple gcc 4.2.1 on Mac OS X Snow Leopard.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -91,7 +91,7 @@ VOID = /dev/null
 
 # thread detection
 NO_THREAD_MSG := ==> no threads, building without multithreading support
-HAVE_PTHREAD := $(shell printf '\#include <pthread.h>\nint main(void) { return 0; }' | $(CC) $(FLAGS) -o have_pthread$(EXT) -x c - -pthread 2> $(VOID) && rm have_pthread$(EXT) && echo 1 || echo 0)
+HAVE_PTHREAD := $(shell printf '\#include <pthread.h>\nint main(void) { return 0; }' > have_pthread.c && $(CC) $(FLAGS) -o have_pthread$(EXT) have_pthread.c -pthread 2> $(VOID) && rm have_pthread$(EXT) && echo 1 || echo 0; rm have_pthread.c)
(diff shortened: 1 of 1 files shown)


ryandesign:
Here is the MacPorts project's ticket about this: https://trac.macports.org/ticket/56646

This was the output I saw, despite those features and libraries actually being available:

```
==> no threads, building without multithreading support
==> no zlib, building zstd without .gz support
==> no liblzma, building zstd without .xz/.lzma support
```

terrelln:
IIUC, this will enable support for the intersection of features supported on each architecture, is that right?

ryandesign:
I don't know how to interpret your word "intersection". I'm just talking about fixing a problem that occurs when trying to build for multiple architectures—a [universal binary](https://en.wikipedia.org/wiki/Universal_binary).

ryandesign:
Yes, that's correct.

Cyan4973:
Does it seem possible to create a test which would trigger (and detect!) the problem,
and be fixed by this patch ?
This, to ensure the requested property will remain correctly supported after future updates.

ryandesign:
It's probably possible to create a test, but I'm not familiar with your test suite.

PR #1427, opened 2018-11-27
Update meson build and add Travis test for it

lzutao:
#### Update meson build

WARNING: Need to test on Windows.

* Move contrib/meson to build/meson
* Update relevant files after moving meson to build/
* Update legacy support to 5
* Fix manpage symlinks
* Install zstdmt.1 manpage
* Install missed zstdgrep and zstdless
* No install zstd_manual.html
* The run_command is run from an unspecified directory. Therefore we cannot assume
  which directory it is running our command.
* Bump to 1.3.8 and fix run_command function
* Fix meson tests build
* Fix lib soversion and no install cover.h header
* Replace CopyFile.py and CreateSymlink.py with InstallSymlink.py script
* Apply new InstallSymlink script
* Add enable_lz4 build option and fix lzma dependency
* Initial support for Windows build
* Add almost all test cases in tests/Makefile
* meson: Change build options' name

#### Init CI test for meson build
* Build both shared and static lib on CI
* ci: Add valgrind and lz4 to meson build
* ci: Use new meson auto_features option


lzutao:
@Cyan4973 I think an Travis CI job for meson build is needed. And it might allow failures.
This job might only if meson build is broken or not.

Cyan4973:
I believe you are right.
Having an automated test which checks that meson builds will still work in the future is very important.
I guess one test can be added in `.travis.tml` for that purpose.
Alternatively, it could be combined with `cmake` build test, since it's the same category.

lzutao:
Added Travis test for meson build.

lzutao:
This pull request is not ready yet. Do not merge it!

lzutao:
@Cyan4973 It is really weird. Gcc-8 build on my machine is fine. But it failed in Travis job [5005.5](https://travis-ci.org/facebook/zstd/jobs/460745897).

lzutao:
Ready to merge now! :sparkles:

lzutao:
@Cyan4973 Can we just move meson to `build` directory?

Cyan4973:
ok

Cyan4973:
At which point do you believe your patch will be ready @lzutao ?

lzutao:
If I don't push anything in one day, it would be ready to merge.
Do you want me to squash this patch @Cyan4973 ?

Cyan4973:
> Do you want me to squash this patch

Not necessarily.
Just, if there was a clean "go" signal, it would make the following "review and merge" step less hazardous.

lzutao:
Ok, I will inform you when it is ready.

lzutao:
The `ZSTD_LEGACY_SUPPORT` is different in these Makefile. Is it intended or not?
I am getting the warning `'ZSTD_LEGACY_SUPPORT' macro redefined` because of 
defining `ZSTD_LEGACY_SUPPORT` as project macro.

https://github.com/facebook/zstd/blob/386c9ab58a3cd3d01682ecd4b16e2884f542ef3d/tests/Makefile#L221-L223

https://github.com/facebook/zstd/blob/386c9ab58a3cd3d01682ecd4b16e2884f542ef3d/lib/Makefile#L43-L47

lzutao:
I think I will control `ZSTD_LEGACY_SUPPORT` through `-Dlegacy_level` build option.

Cyan4973:
I don't remember exactly why there is `legacy` target in `tests/Makefile`.
Maybe it's a way to guarantee that it's still possible to compile `zstd` while using a value of `ZSTD_LEGACY_SUPPORT` different from default.

Indeed, this test hard-defines a specific value for `ZSTD_LEGACY_SUPPORT`,
which leads to problem if one wants to use a different value at the same time.

Possible work around : 
- do not use `legacy` target in your scripts
- change the `legacy` target so that, instead of using a hard-coded value, it uses a variable with a default value, which can be overridden on command line.

lzutao:
@Cyan4973 I think it is OK to merge now. I  wouldn't touch this PR and would create another PR if needed.

lzutao:
@Cyan4973 After using `git blame` I guess that increasing `ZSTD_LEGACY_SUPPORT` to 5 in `tests/Makefile` is just forgot. What do you think about increasing the macro to 5?

Cyan4973:
> What do you think about increasing the macro to 5?

It seems fine.

Note that I'm also wondering what's the point of this `legacy` target in `tests/Makefile`. 
What does it test ?
Maybe the target itself is redundant and could be removed ...

lzutao:
@Cyan4973 A `tests/legacy.c` comment state that
> This program uses hard-coded data compressed with Zstd legacy versions
> and tests that the API decompresses them correctly

BTW, could you review this patch? I think it is OK now.

build/meson/README.md
```diff
@@ -0,0 +1,38 @@
+Meson build system for zstandard
+================================
+
+Meson is a build system designed to optimize programmer productivity.
+It aims to do this by providing simple, out-of-the-box support for
+modern software development tools and practices, such as unit tests,
+coverage reports, Valgrind, CCache and the like.
+
+This Meson build system is provided with no guarantee and maintained
+by Dima Krasner <dima@dimakrasner.com>.
```
Cyan4973: You might want to add your name here @lzutao, considering the very large update you provided to the meson build,

PR #1429, opened 2018-11-28
fixed a typo

denis2342:
Since zstd is generating files with the ending .zst I corrected the help output.


Cyan4973:
Nice catch ! Thanks @denis2342 !

PR #1430, opened 2018-11-28
Fix  #1428 - zstdgrep now returns 1 on unmatch

### LINKED ISSUE #1428: zstdgrep returns 0 on non-match
The return code of grep is often used—particularly with "grep -q", but not always—in a script to detect the presence or absence of a match.  And zgrep preserves this behavior.  But zstdgrep does not:
```
$ echo a | gzip | zgrep a; echo $?
a
0
$ echo a | gzip | zgrep b; echo $?
1
$ echo a | zstd | zstdgrep a; echo $?
a
0
$ echo a | zstd | zstdgrep b; echo $?
0
```
Even with "zstdgrep -q", a non-match still returns 0.
I'd recommend bringing zstdgrep in line.

lzutao:
Fix some shellcheck warnings.

### DIFF
--- a/programs/zstdgrep
+++ b/programs/zstdgrep
@@ -103,9 +103,11 @@ EXIT_CODE=0
 # call grep ...
 if [ "$#" -lt 1 ]; then
     # ... on stdin
+    set -f # Disable file name generation (globbing).
     # shellcheck disable=SC2086
     "${zcat}" -fq - | "${grep}" ${grep_args} -- "${pattern}" -
     EXIT_CODE=$?
+    set +f
 else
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for this timely fix @lzutao !

Would it be possible to add a test of `zstdgrep` ?
It would typically be added to, or triggered from, `tests/Makefile`.

Cyan4973:
Thanks @lzutao !

Cyan4973:
It seems this introduced a new problem : 
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=918390

PR #1431, opened 2018-11-29
Refactor examples to avoid code duplication

hungptit:
Below are changes that I have made:
1. Move common functions into a shared header file to avoid code duplication
2. Replace hard coded exit values with enums to improve readability of example code.

### DIFF
--- a/examples/dictionary_compression.c
+++ b/examples/dictionary_compression.c
@@ -7,71 +7,13 @@
  * in the COPYING file in the root directory of this source tree).
  * You may select, at your option, one of the above-listed licenses.
  */
-
-
 #include <stdlib.h>    // malloc, exit
 #include <stdio.h>     // printf
 #include <string.h>    // strerror
 #include <errno.h>     // errno
 #include <sys/stat.h>  // stat
 #include <zstd.h>      // presumes zstd library is installed
-
-
-static off_t fsize_orDie(const char *filename)
-{
-    struct stat st;
-    if (stat(filename, &st) == 0) return st.st_size;
-    /* error */
-    perror(filename);
-    exit(1);
-}
-
-static FILE* fopen_orDie(const char *filename, const char *instruction)
-{
-    FILE* const inFile = fopen(filename, instruction);
-    if (inFile) return inFile;
-    /* error */
-    perror(filename);
-    exit(2);
-}
-
-static void* malloc_orDie(size_t size)
-{
-    void* const buff = malloc(size);
-    if (buff) return buff;
-    /* error */
-    perror("malloc");
-    exit(3);
-}
-
-static void* loadFile_orDie(const char* fileName, size_t* size)
-{
-    off_t const buffSize = fsize_orDie(fileName);
-    FILE* const inFile = fopen_orDie(fileName, "rb");
-    void* const buffer = malloc_orDie(buffSize);
-    size_t const readSize = fread(buffer, 1, buffSize, inFile);
-    if (readSize != (size_t)buffSize) {
-        fprintf(stderr, "fread: %s : %s \n", fileName, strerror(errno));
-        exit(4);
-    }
-    fclose(inFile);
-    *size = buffSize;
-    return buffer;
-}
-
-static void saveFile_orDie(const char* fileName, const void* buff, size_t buffSize)
-{
-    FILE* const oFile = fopen_orDie(fileName, "wb");
-    size_t const wSize = fwrite(buff, 1, buffSize, oFile);
-    if (wSize != (size_t)buffSize) {
-        fprintf(stderr, "fwrite: %s : %s \n", fileName, strerror(errno));
-        exit(5);
-    }
-    if (fclose(oFile)) {
-        perror(fileName);
-        exit(6);
-    }
-}
+#include "utils.h"
 
 /* createDict() :
    `dictFileName` is supposed to have been created using `zstd --train` */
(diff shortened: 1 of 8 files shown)


Cyan4973:
These are rather good changes @hungptit .

One small issue is that I'm not used to implement functions directly into `*.h`.
It works here, because they are just simple examples,
but it supposes that `utils.h` is only included once per program,
otherwise it will generate symbol duplicates.

While it's not a big deal, since these are just examples,
one issue is that some people like to copy/paste the examples directly.
Hence some code comment would be welcomed to explain the rationale and the limitations.

Alternatively, you could `static` all functions defined in `utils.h`.
Then, some compilers will complain that some `static` functions are not used,
requiring compiler-specific directives, such as `__attribute__((unused))` for `gcc`.
Refer to https://github.com/facebook/zstd/blob/v1.3.5/programs/util.h#L92 for guidance.

hungptit:
I use **#pragma once** at the beginning of utils.h so it is only included one time. I have verified my changes by maximizing the warning level in Makefile and including utils.h several times in one source file. 

Below are what I have done to test my changes:

**Modify the Makefile and include utils several times in dictionary_compression.c**

```
2 files changed, 3 insertions(+)
examples/Makefile                 | 1 +
examples/dictionary_compression.c | 2 ++

modified   examples/Makefile
@@ -10,6 +10,7 @@
 # This Makefile presumes libzstd is installed, using `sudo make install`

 CPPFLAGS += -I../lib
+CFLAGS += -Wall -Wextra -march=native -O3
 LIB = ../lib/libzstd.a

 .PHONY: default all clean test
modified   examples/dictionary_compression.c
@@ -14,6 +14,8 @@
 #include <sys/stat.h>  // stat
 #include <zstd.h>      // presumes zstd library is installed
 #include "utils.h"
+#include "utils.h"
+#include "utils.h"

 /* createDict() :
    `dictFileName` is supposed to have been created using `zstd --train` */

```

**Compile all examples with a default compiler and gcc-8 in my MacBook Pro**
```
make clean; make -j5
Cleaning completed
cc -I../lib -Wall -Wextra -march=native -O3 simple_compression.c ../lib/libzstd.a  -o simple_compression
cc -I../lib -Wall -Wextra -march=native -O3 simple_decompression.c ../lib/libzstd.a  -o simple_decompression
cc -I../lib -Wall -Wextra -march=native -O3 dictionary_compression.c ../lib/libzstd.a  -o dictionary_compression
cc -I../lib -Wall -Wextra -march=native -O3 dictionary_decompression.c ../lib/libzstd.a  -o dictionary_decompression
cc -I../lib -Wall -Wextra -march=native -O3 streaming_compression.c ../lib/libzstd.a  -o streaming_compression
cc -I../lib -Wall -Wextra -march=native -O3 streaming_decompression.c ../lib/libzstd.a  -o streaming_decompression
cc -I../lib -Wall -Wextra -march=native -O3 multiple_streaming_compression.c ../lib/libzstd.a  -o multiple_streaming_compression
cc -I../lib -Wall -Wextra -march=native -O3 streaming_memory_usage.c ../lib/libzstd.a  -o streaming_memory_usage
```

lzutao:
@hungptit  Please read this [stackoverflow answer](https://stackoverflow.com/questions/558122/what-is-a-static-function) to know about static function.

Cyan4973:
Great job on code comments @hungptit !

> I use #pragma once at the beginning of utils.h

The issue is not multiple inclusions (directly or indirectly) of `utils.h` within the same `*.c` file.
It is indeed well protected by `#pragma once` (though I prefer the classical `#ifndef X ...` guard construction for maximum portability, but that's minor).

The issue appears when a project contains at least 2 `*.c` files, and at least 2 of them include `utils.h`.
In this case, symbol duplication will appear on public symbols (those not `static`).

This situation is not possible in `/examples`, because all projects are short, single `*.c` file.

The issue will appear if someone copy/paste `utils.h` into its project and starts to include it into multiple files.
Or when, in a hypothetical future, a 2-files example is created.

So it's true this doesn't immediately impact `/examples` compilation.
Consider it just as good practice for functions in `*.h` files.

Cyan4973:
There are still 2 symbols which are public in `utils.h`. 
Unfortunately, a single public symbol is enough to trigger symbol duplication issues.

I suspect they are public because they would generate an "unused function" warning in some examples, since they are not "always" necessary. The warning itself is also an issue when employing a "zero warning" policy, which is recommended for high quality software. The solution is to [selectively silence the warning](https://github.com/facebook/zstd/pull/1431#issuecomment-443044177) (and not globally, since the warning remains useful to detect unwanted situations).

That being said, I don't want to bother you too much with this issue, since it's not a requirement for existing `/examples`.
The rest of the patch is pretty good quality, code comments are great, and I believe your patch is a valuable addition.

I'll complete it on my side.

Maybe in some future, if you find yourself in a similar situation, with the need to implement functions in a `*.h` included from multiple `*.c` files (in the same program), this discussion will make a bit more sense, and hopefully prove useful.

hungptit:
I missed the last two functions by mistake. I will fix it in the next
round. It is a big challenge to write C code after so many years with C++.

Regards,
Hung

Cyan4973:
Damn, I'm unable to get your patch applied to `dev` branch.
Patching directly `facebook:dev` from `hungptit:master` generates conflicts.

I could copy/paste your files into my repository, but it would remove attribution.

I guess best solution is, that you copy/paste the files in your repo `hundptit:dev` and submit a new PR from `hungptit:dev` into `facebook:dev`,
this way attribution will be preserved.

hungptit:
OK. I can pull a request to the dev branch. If you see any thing that can
be better or improved then feel free to add the inline code comments.

Thanks,
Hung

PR #1433, opened 2018-11-30
Small changes for buck build

obelisk:
I apologize if this diff is incorrect incase I'm missing something but when I tried to buck build the zstd program it failed because it couldn't find the debug headers/code.

This diff does four things:

- This adds a new buck target for the debug code and adds it common
- Fixes a typo in the word deprecated
- Added the `hist.c` file to the compress target
- Adds `zstd.h` as an exported header for the entire library

### DIFF
--- a/lib/BUCK
+++ b/lib/BUCK
@@ -1,6 +1,7 @@
 cxx_library(
     name='zstd',
     header_namespace='',
+    exported_headers=['zstd.h'],
     visibility=['PUBLIC'],
     deps=[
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @obelisk !
This modifications looks good, though I will defer final verdict to @terrelln .

It seems that the buck build script is not part of our CI test suite.
It would probably be better to have some test that validates it still works as intended.
What do you think @terrelln ?

PR #1444, opened 2018-12-11
btultra2

Cyan4973:
This patch adds a new strategy, `btultra2`.
This strategy passes the first block twice through the `btultra` parser,
the first pass being only useful in order to populate statistics,
which makes the second pass more efficient.

Gains are typically small but measurable (a bit less than 1%, on first block).

Due to changes to compression level tables,
enabling this strategy at highest compression levels,
and rebalancing a few other levels for a smoother or more optimal speed / compression trade off,
a lot of levels now produce a different result (generally more compact and a bit slower).


terrelln:
Once you land, we can update `results.csv` once the cron job runs. Or you can run it manually: `make zstd && cd tests/regression && make test && ./test -o results.csv -c data-cache -z ../../zstd`

Cyan4973:
Note : there is another patch ongoing, which will also impact compression ratio, 
when associated with high compression levels and multithreading (or at least 1 worker, which is the default mode of the cli).

Cyan4973:
`results.csv` has been updated

PR #1446, opened 2018-12-12
[libzstd] Fix infinite loop in decompression

terrelln:
When we switched `ZSTD_SKIPPABLEHEADERSIZE` to a macro, the places where we do:

    MEM_readLE32(ptr) + ZSTD_SKIPPABLEHEADERSIZE

can now overflow `(unsigned)-8` to `0` and we infinite loop. The fix is to make
the variable an `unsigned long long`, so it is always 64-bits. I've also made
the other related variables `unsigned long long` to mitigate similar potential
issues.

Note that this bug never made it into a release, and was only in the dev branch
for a few days.

Credit to OSS-Fuzz

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -343,6 +343,21 @@ unsigned long long ZSTD_getFrameContentSize(const void *src, size_t srcSize)
     }   }
 }
 
+static size_t readSkippableFrameSize(void const* src, size_t srcSize)
+{
+    size_t const skippableHeaderSize = ZSTD_SKIPPABLEHEADERSIZE;
+    U32 sizeU32;
+
+    if (srcSize < ZSTD_SKIPPABLEHEADERSIZE)
+        return ERROR(srcSize_wrong);
+
+    sizeU32 = MEM_readLE32((BYTE const*)src + ZSTD_FRAMEIDSIZE);
+    if ((U32)(sizeU32 + ZSTD_SKIPPABLEHEADERSIZE) < sizeU32)
+        return ERROR(frameParameter_unsupported);
+
+    return skippableHeaderSize + sizeU32;
+}
+
 /** ZSTD_findDecompressedSize() :
  *  compatible with legacy mode
  *  `srcSize` must be the exact length of some number of ZSTD compressed and/or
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for catching that issue.

So the scenario proves that it's not possible to blindly add `ZSTD_SKIPPABLEHEADERSIZE == 8` to skippable frame size, since skippable frame size is read as a `U32`, and could be set to `(4 GB - 8)`, resulting in a wrap to zero, since the addition happens within `U32` space.

It's worth noting that we have another problem, if the skippable frame size is `> (4 GB - 8)`. In which case, the addition also wraps and the result, while not zero, is incorrect (the result is defined from a `C` standard perspective, since wrapping is the defined behavior of unsigned integers, but it's not what we want).

Making all constant `unsigned long long` is likely to break 32-bit builds, and could have other potential side effects related to unwanted casting.
I believe we are better off looking at the places where this situation occurs, and fix that in place (occurrences are pretty rare).

In all instances I could find, the result of the addition is saved into a `size_t`. Hence, for the addition to be completed in the correct space, it's enough to cast the result of `MEM_readLE32()` into `(size_t)`, it will make the operation 64-bit, and the result will be correct.

However, the result of the addition will still be incorrect in 32-bit mode.
And I don't see an easy way to fix that : it would require to change the return or storage type, from `size_t` to `unsigned long long` (or `U64`), so that it can express values up to `4 GB + 7`.

Alternatively, it could also make sense to just reject such sizes, and return an error code instead.
It's a step back from the specification, but it could be simply labelled a decoder limitation, and realistically, I can't think of a valid scenario needing a skippable frame size of `4 GB - 8` or more. 
Since this operation was previously not supposed to fail, it's first necessary to ensure that returning an error code is fine. And from what I can tell, it is : there are only 3 places where this addition happens, and they all play well with error code.
Then it would be necessary to select an error code.
We don't have one which would be specifically accurate, but there is one which could be good enough : [`ZSTD_error_frameParameter_unsupported`](https://github.com/facebook/zstd/blob/dev/lib/common/zstd_errors.h#L57).

For discussion

terrelln:
Looks like this fix doesn't work on 32-bit platforms.

terrelln:
I've switched to checking for overflow.

PR #1447, opened 2018-12-13
meson: More accurate Windows build support

lzutao:


Cyan4973:
Is the patch ready for review ? Do you need to work on it a bit more  ?

lzutao:
@Cyan4973 I think it is needed to test Meson build on Appveyor, but I am not familiar with that build service.

Btw, this PR is ready.

PR #1448, opened 2018-12-13
travis: Use script instead of Cmd env

lzutao:


Cyan4973:
Thanks, I agree it's better style

PR #1451, opened 2018-12-14
pzstd: Fix wrong condition

lzutao:
`stemSize` is always greater than or equal to zero.

### DIFF
--- a/contrib/pzstd/Options.cpp
+++ b/contrib/pzstd/Options.cpp
@@ -415,14 +415,17 @@ std::string Options::getOutputFile(const std::string &inputFile) const {
   }
   // Attempt to add/remove zstd extension from the input file
   if (decompress) {
-    int stemSize = inputFile.size() - kZstdExtension.size();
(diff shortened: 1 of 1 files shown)


terrelln:
@lzutao I don't see anything wrong with the current code. Do you have a test case that fails?

It is broken when the input file name is over 2 GB, but that isn't a scenario I feel we need to handle.

lzutao:
You're right. But when the `inputFile` is `".zst"`, which is the `kZstdExtension`, 
I think that is a valid filename, but the old code seems ignore that.

Another problem is that we should return an empty `std::string()` instance instead of
an empty C string `""`.

terrelln:
This code treats the empty return result as an error, so both before and after this patch we won't handle the file `".zst"`. I don't want to create a file with an empty filename, so that is okay.

Returning `""` is equivalent to returning `std::string()`. `std::string()` has an implicit conversion from `char const*`. See the 5th constructor here https://en.cppreference.com/w/cpp/string/basic_string/basic_string.

Since there is no functional change before and after, I prefer to leave the code alone.

PR #1454, opened 2018-12-14
README: Explain a little bit about Meson build

lzutao:
- Increase one heading level. Now headings is bigger and prettier.
- Explain a bit more about Meson build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,11 @@ and a command line utility producing and decoding `.zst`, `.gz`, `.xz` and `.lz4
 Should your project require another programming language,
 a list of known ports and bindings is provided on [Zstandard homepage](http://www.zstd.net/#other-languages).
 
-Development branch status : [![Build Status][travisDevBadge]][travisLink]   [![Build status][AppveyorDevBadge]][AppveyorLink]   [![Build status][CircleDevBadge]][CircleLink]
+**Development branch status:**
+
+[![Build Status][travisDevBadge]][travisLink]
+[![Build status][AppveyorDevBadge]][AppveyorLink]
+[![Build status][CircleDevBadge]][CircleLink]
 
 [travisDevBadge]: https://travis-ci.org/facebook/zstd.svg?branch=dev "Continuous Integration test suite"
 [travisLink]: https://travis-ci.org/facebook/zstd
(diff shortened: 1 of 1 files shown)


lzutao:
cc @Cyan4973

Cyan4973:
There are 3 parts in this patch.
I'm fine for the last 2 ones,
but regarding the first (LGTM badge), 
I think we would need first to select LGTM as a static analyzer that we want to support.

Static analyzers tend to generate false positives, by design (they require more guarantees).
It's hard to support many of them. 
Handling false positives generally requires rewriting a segment of code differently, even when previous one was correct. And when unlucky, 2 static analyzers will contradict each other, each requiring a different way to write the same thing.

We already support 2 static analyzers, `scan-build` and `cppcheck`.

Looking at [LGTM results](https://lgtm.com/projects/g/facebook/zstd/alerts/?mode=list), the report complains about 44 warnings. So they would need to be solved first, before adding a badge.
Many of these warnings are about "Missing header guards" for example, even though header guards are actually right there, just a few lines below. That's not a great sign. I'm not yet convinced going through this list of warnings will be worth the investment.

For the time being, it would be fine if this part could be separated, so that the 2 other elements in this PR (meson doc, readme levels) can get merged.

README.md
```diff
@@ -9,16 +9,26 @@ and a command line utility producing and decoding `.zst`, `.gz`, `.xz` and `.lz4
 Should your project require another programming language,
 a list of known ports and bindings is provided on [Zstandard homepage](http://www.zstd.net/#other-languages).
 
-Development branch status : [![Build Status][travisDevBadge]][travisLink]   [![Build status][AppveyorDevBadge]][AppveyorLink]   [![Build status][CircleDevBadge]][CircleLink]
+**Development branch status:**
+
+[![Build Status][travisDevBadge]][travisLink]
+[![Build status][AppveyorDevBadge]][AppveyorLink]
+[![Build status][CircleDevBadge]][CircleLink]
+[![Total alerts][LgtmAlertBadge]][LgtmAlertLink]
```
Cyan4973: Please remove `LGTM` badges.
The rest of the patch is good to go.

PR #1456, opened 2018-12-17
Add multiple_simple_compression.c (demonstrates the use of ZSTD_compressCCtx(...)) to examples directory.

yijinfb:
1. Add multiple_simple_compression.c to examples directory.

2. In utils.h:
   (a) refactor fsize_orDie()- it now returns size_t instead of off_t after checking for off_t -> size_t cast overflow.
   (b) generalize loadFile_orDie() to use (optional) pre-allocated memory buffer.

3. Amend Makefile to account for dependency on utils.h.


examples/Makefile
```diff
@@ -17,40 +17,45 @@ LIB = ../lib/libzstd.a
...
+	$(CC) $(CPPFLAGS) $(CFLAGS) $< $(LIB) $(LDFLAGS) -o $@
 
-dictionary_compression : dictionary_compression.c $(LIB)
-	$(CC) $(CPPFLAGS) $(CFLAGS) $^ $(LDFLAGS) -o $@
+multiple_simple_compression : multiple_simple_compression.c utils.h $(LIB)
+	$(CC) $(CPPFLAGS) $(CFLAGS) $< $(LIB) $(LDFLAGS) -o $@
 
-dictionary_decompression : dictionary_decompression.c $(LIB)
-	$(CC) $(CPPFLAGS) $(CFLAGS) $^ $(LDFLAGS) -o $@
+dictionary_compression : dictionary_compression.c utils.h $(LIB)
```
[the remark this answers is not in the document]
Cyan4973: I believe that's fine.
`utils.h` in this case is just a dependency, which is correct.
It's not compiled, since it's not part of `$<`.

The advantage is that now, when `utils.h` gets modified, any depending program (basically all examples) gets recompiled too, which is also correct.
Previously, any change to `utils.h` would have remained invisible to `Makefile`.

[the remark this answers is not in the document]
lzutao: agreed!

examples/utils.h
```diff
@@ -128,39 +141,41 @@ static void* malloc_orDie(size_t size)
...
+ *
  * Note: This function will send an error to stderr and exit if it
  * cannot read data from the given file path.
- * 
+ *
  * @return If successful this function will return a pointer to read
  * data otherwise it will printout an error to stderr and exit.
  */
-static void* loadFile_orDie(const char* fileName, size_t* size)
+static void* loadFile_orDie(const char* fileName, size_t* size, void* buffer, int bufferSize)
```
Cyan4973: This works. All paths look properly checked.

However, in term of _interface design_,
I suspect it's not great to have a function which sometimes allocate a buffer and sometimes doesn't (depending on its arguments), even if it's clearly documented.

I feel rather in favor of 2 versions,
one which just load data into an existing buffer,
and another one which creates a buffer and uses the first version to load it.
The second version should have a name which makes it very clear that an allocation is happening, since it means a `free` operation will have to be triggered later on.

cc @terrelln for additional point of view.

examples/multiple_simple_compression.c
```diff
@@ -0,0 +1,115 @@
...
+    size_t fBufferSize;
+    size_t cBufferSize;
+    ZSTD_CCtx* cctx;
+} resources;
+
+/*
+ * allocate memory for buffers big enough to compress all files
+ * as well as memory for output file name (outFilename)
+ */
+static resources createResources_orDie(int argc, const char** argv, char **outFilename)
```
Cyan4973: Could `outFileName` be part of `resources`, since it's basically just a buffer to manipulate strings ?

examples/multiple_simple_compression.c
```diff
@@ -0,0 +1,115 @@
...
+    /* memory allocation for outFilename and resources */
+    char* outFilename;
+    resources ress = createResources_orDie(argc, argv, &outFilename); 
+
+    /* compress files with shared context, input and output buffers */
+    int argNb;
+    for (argNb = 1; argNb < argc; argNb++) {
+        const char* const inFilename = argv[argNb];
+        memset(outFilename, 0, 1);
+        strcat(outFilename, inFilename);
```
Cyan4973: We received some push back in the past against using `strcat`, due to potential buffer overflow.

I know that, in this case, there should be no problem, because you already sized `outFileName` to be large enough for any `inFilename`. Still, such condition should be spelled explicitly, using `assert()`.

FYI, note that unfortunately, some static analyzer will still complain "blindly" against any use of `strcat` even when everything is done correctly, 
Replacing it with `strncat`, which is often suggested because it offers control over size, is source of other problems, such as portability.
So we ended up using `memcpy()` instead.

Replacement is optional. I believe `assert()` is enough to guarantee good behavior.

Cyan4973:
Don't worry about the `travis` error on `aarch64` : it's an `ld` bug, not directly related to any fault in the source code.

examples/utils.h
yijinfb: Totally agree! I feel a bit hesitant to do that too...

examples/multiple_simple_compression.c
Cyan4973: Totally reasonable !

For the record, another design pattern could be to create an encapsulating `struct` as return type of `createResources_orDie()`, containing both the current `resources` and `outFilename` buffer, so that they can be easily separated later for `compressFile_orDie()`.

However, that would define 2 different types of `resources`, requiring different names to avoid confusing. 
It's probably overkill for such a simple example.

examples/multiple_simple_compression.c
yijinfb: Good point. Resolved with both `assert()` and `memcpy()`.

examples/utils.h
yijinfb: code refactored.

yijinfb:
All concerns should be addressed by the latest commit.

Cyan4973:
Excellent ! Thanks @yijinfb !

PR #1457, opened 2018-12-18
btultra2 and very small input

Cyan4973:
`btultra2` presumed that providing more accurate statistics to the parser would result in better compression ratio.

While that's generally true, this does not apply when input is very small, since default distribution tables will be used at entropy stage, whatever the real distribution of values.

The optimal parser uses a cheap heuristic to determine this mode : it uses default distribution tables and keep the cost static when `srcSize <= 1024`.
The same heuristic is now used in `btultra2` to disable initial statistics estimation.

This solves the problem of `btultra2` generating worse compression than `btultra` on some small files.

Note that, a proper way would be to selectively determine which symbol will use dynamic tables and which one will use predefined tables, then selectively initialize relevant symbols. Unfortunately, the parser is currently unable to make a "selective" decision regarding cost evolution : it's all static and predefined, or all dynamic. Some underlying changes will be needed to unlock this possibility.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -1116,8 +1116,8 @@ ZSTD_initStats_ultra(ZSTD_matchState_t* ms,
     U32 tmpRep[ZSTD_REP_NUM];  /* updated rep codes will sink here */
     memcpy(tmpRep, rep, sizeof(tmpRep));
 
-    DEBUGLOG(2, "ZSTD_initStats_ultra (srcSize=%zu)", srcSize);
-    DEBUGLOG(2, "repCodes: %u, %u, %u", tmpRep[0], tmpRep[1], tmpRep[2]);
+    DEBUGLOG(5, "ZSTD_initStats_ultra (srcSize=%zu)", srcSize);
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_opt.c
```diff
@@ -1067,28 +1097,32 @@ static U32 ZSTD_upscaleStat(U32* table, U32 lastEltIndex, int bonus)
...
+ZSTD_initStats_ultra(ZSTD_matchState_t* ms,
+                     seqStore_t* seqStore,
+                     U32 rep[ZSTD_REP_NUM],
+               const void* src, size_t srcSize)
 {
     U32 tmpRep[ZSTD_REP_NUM];  /* updated rep codes will sink here */
+    memcpy(tmpRep, rep, sizeof(tmpRep));
 
-    DEBUGLOG(5, "ZSTD_initStats_ultra (srcSize=%zu)", srcSize);
+    DEBUGLOG(2, "ZSTD_initStats_ultra (srcSize=%zu)", srcSize);
```
terrelln: Switch these back to 5.

PR #1461, opened 2018-12-20
[regression] Add more configs

terrelln:
* Add configs that test multithreading, LDM, and setting explicit
  parameters.
* Update the `compress cctx` method to accept `ZSTD_parameters`.
* Compile against the multithreaded `libzstd.a`.
* Update `results.csv` for the new configs.

Unless you think there are more configs/methods I should test, I think
we have a fairly wide set of configs/methods, so I'll pause adding
more for now.

### DIFF
--- a/tests/regression/Makefile
+++ b/tests/regression/Makefile
@@ -10,7 +10,7 @@
 CFLAGS ?= -O3
 
 CURL_CFLAGS := $(shell curl-config --cflags)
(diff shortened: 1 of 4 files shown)


Cyan4973:
> Unless you think there are more configs/methods I should test, I think we have a fairly wide set of configs/methods

I agree

PR #1463, opened 2018-12-20
Add support for environment variable ZSTD_CLEVEL in CLI

yijinfb:
1. Add support for environment variable ZSTD_CLEVEL in CLI, both positive and negative values are allowed.
2. refactor readU32FromChar(...)
3. Add test cases and update documentation.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -234,12 +234,12 @@ static void errorOut(const char* msg)
     DISPLAY("%s \n", msg); exit(1);
 }
 
-/*! _readU32FromChar() :
+/*! readU32FromCharChecked() :
  * @return 0 if success, and store the result in *value.
  *  allows and interprets K, KB, KiB, M, MB and MiB suffix.
  *  Will also modify `*stringPtr`, advancing it to position where it stopped reading.
  * @return 1 if an overflow error occurs */
-static int _readU32FromChar(const char** stringPtr, unsigned* value)
+static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
 {
     static unsigned const max = (((unsigned)(-1)) / 10) - 1;
     unsigned result = 0;
@@ -271,7 +271,7 @@ static int _readU32FromChar(const char** stringPtr, unsigned* value)
 static unsigned readU32FromChar(const char** stringPtr) {
     static const char errorMsg[] = "error: numeric value too large";
     unsigned result;
-    if (_readU32FromChar(stringPtr, &result)) { errorOut(errorMsg); }
+    if (readU32FromCharChecked(stringPtr, &result)) { errorOut(errorMsg); }
     return result;
 }
 
@@ -483,15 +483,15 @@ static int init_cLevel(void) {
 
         if ((*ptr>='0') && (*ptr<='9')) {
             unsigned absLevel;
-            if (_readU32FromChar(&ptr, &absLevel)) { 
-                DISPLAYLEVEL(2, "Ignore environment variable %s=%s: numeric value too large\n", ENV_CLEVEL, env);
+            if (readU32FromCharChecked(&ptr, &absLevel)) { 
+                DISPLAYLEVEL(2, "Ignore environment variable setting %s=%s: numeric value too large\n", ENV_CLEVEL, env);
                 return ZSTDCLI_CLEVEL_DEFAULT;
             } else if (*ptr == 0) {
                 return sign * absLevel;
             }
         }
 
-        DISPLAYLEVEL(2, "Ignore environment variable %s=%s: not a valid integer value\n", ENV_CLEVEL, env);
+        DISPLAYLEVEL(2, "Ignore environment variable setting %s=%s: not a valid integer value\n", ENV_CLEVEL, env);
     }
 
     return ZSTDCLI_CLEVEL_DEFAULT;
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -469,7 +469,7 @@ static void printVersion(void)
 #define ENV_CLEVEL "ZSTD_CLEVEL"
 
 /* functions that pick up environment variables */
-int init_cLevel() {
+int init_cLevel(void) {
     const char* const env = getenv(ENV_CLEVEL);
     if (env) {
         const char *ptr = env;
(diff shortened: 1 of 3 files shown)


programs/zstdcli.c
```diff
@@ -452,6 +465,38 @@ static void printVersion(void)
 #endif
 }
 
+/* Environment variables for parameter setting */
+#define ENV_CLEVEL "ZSTD_CLEVEL"
+
+/* functions that pick up environment variables */
+int init_cLevel() {
```
Cyan4973: minor : prefer declaration `static int init_cLevel(void)`

programs/zstdcli.c
```diff
@@ -467,6 +512,8 @@ typedef enum { zom_compress, zom_decompress, zom_test, zom_bench, zom_train, zom
 
 int main(int argCount, const char* argv[])
 {
+    g_displayOut = stderr;
```
Cyan4973: We use declaration-after-statement, for broader compatibility with a larger class of compilers.
It forbids assigning a value to `g_displayOut` before variable declarations.

programs/zstdcli.c
yijinfb: Sorry, forgot that one.

programs/zstdcli.c
Cyan4973: A usual way to make an assignment before variable declaration is to create a new indentation level `{ }`.
In this case though, it would mean that the whole code inside `main()` would have to indent by 4 spaces, just to initialize `g_displayOut`, which feels overkill.

Another way could be to combine it to an assignment using the comma operator.
Typically : 
`int cLevel = ( g_displayOut = stderr, init_cLevel() );`

Only `init_cLevel()` return value will be used, but `g_displayOut` will be initialized before (the order is guaranteed).
The downside is that readability is less good, simply because less coders understand the comma operator.

tests/playTests.sh
```diff
@@ -122,6 +122,18 @@ $ZSTD --fast=5000000000 -f tmp && die "too large numeric value : must fail"
...
+ZSTD_CLEVEL=-12 $ZSTD -f tmp # negative compression level
+ZSTD_CLEVEL=+12 $ZSTD -f tmp # valid: verbose '+' sign 
+ZSTD_CLEVEL=    $ZSTD -f tmp # empty env var, warn and revert to default setting
+ZSTD_CLEVEL=-   $ZSTD -f tmp # malformed env var, warn and revert to default setting
+ZSTD_CLEVEL=a   $ZSTD -f tmp # malformed env var, warn and revert to default setting
+ZSTD_CLEVEL=+a  $ZSTD -f tmp # malformed env var, warn and revert to default setting
+ZSTD_CLEVEL=3a7 $ZSTD -f tmp # malformed env var, warn and revert to default setting
+ZSTD_CLEVEL=50000000000  $ZSTD -f tmp # numeric value too large, warn and revert to default setting
+$ECHO "test : override ZSTD_CLEVEL with command line option"
+ZSTD_CLEVEL=12  $ZSTD --fast=3 -f tmp # overridden by command line option
```
Cyan4973: great test cases ! 👍

programs/zstdcli.c
```diff
@@ -493,7 +540,8 @@ int main(int argCount, const char* argv[])
     size_t blockSize = 0;
     zstd_operation_mode operation = zom_compress;
     ZSTD_compressionParameters compressionParams;
-    int cLevel = ZSTDCLI_CLEVEL_DEFAULT;
+    int cLevel = init_cLevel();
+    printf("init cLevel = %d\n", cLevel);
```
Cyan4973: I guess these `printf()` are debug left-over ?
To be removed.

programs/zstdcli.c
yijinfb: Thanks for the tip! I will just rearrange the code to make it more readable.

programs/zstdcli.c
```diff
@@ -233,32 +234,44 @@ static void errorOut(const char* msg)
...
-/*! readU32FromChar() :
- * @return : unsigned integer value read from input in `char` format.
+/*! _readU32FromChar() :
+ * @return 0 if success, and store the result in *value.
  *  allows and interprets K, KB, KiB, M, MB and MiB suffix.
  *  Will also modify `*stringPtr`, advancing it to position where it stopped reading.
- *  Note : function will exit() program if digit sequence overflows */
-static unsigned readU32FromChar(const char** stringPtr)
+ * @return 1 if an overflow error occurs */
+static int _readU32FromChar(const char** stringPtr, unsigned* value)
```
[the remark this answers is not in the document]
yijinfb: That's a good point.

programs/zstdcli.c
```diff
@@ -452,6 +465,38 @@ static void printVersion(void)
...
+        const char *ptr = env;
+        int sign = 1;
+        if (*ptr == '-') {
+            sign = -1;
+            ptr++;
+        } else if (*ptr == '+') {
+            ptr++;
+        }
+
+        if ((*ptr>='0') && (*ptr<='9')) {
```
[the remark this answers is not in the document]
yijinfb: That's a good idea! It requires a bit more refactoring work as `readU32FromChar()` is used in many places under slightly different contexts.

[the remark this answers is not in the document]
Cyan4973: We are too close to a release to modify any behavior, so no need to refactor.

Also, I suspect that the `if ((*ptr>='0') && (*ptr<='9'))` test is extraneous,
because it's already included in `readU32FromCharChecked()`, 
so the function will fail and return `1` if the first character does not respect this condition anyway.

Cyan4973:
The test issues were unrelated to this patch.
I'm restarting them so that we can get a clear green signal.

programs/zstdcli.c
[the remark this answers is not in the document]
yijinfb: readU32FromCharChecked() in its current form is pretty bullet-proof, and it will return 0 as the value read if the first char is not a numeral. This feature has implications on allowed command line option format, e.g., currently,
`$ZSTD -bi0 --fast tmp1`
is equivalent to
`$ZSTD -bi --fast tmp1`.

If we remove `if ((*ptr>='0') && (*ptr<='9'))` in `int_cLevel(...)`, setting `ZSTD_CLEVEL` to an empty string will be considered valid and interpreted as ZSTD_CLEVEL=0. 

Also the same check appears in current release: (zstdcli.c, line 754)
```
#ifndef ZSTD_NOCOMPRESS
                    /* compression Level */
                    if ((*argument>='0') && (*argument<='9')) {
                        dictCLevel = cLevel = readU32FromChar(&argument);
                        continue;
                    }
#endif
```
which is also necessary.

In the future, we can spend some time rationalizing the acceptable option format for the command line. Also it would be helpful to add an option to display all the parameter settings actually used by our compression/decompression algorithms.

PR #1465, opened 2018-12-20
 fixed : detection of non-existing file 

Cyan4973:
provides a better error message, consistent with `gzip` one

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -388,6 +388,12 @@ static FILE* FIO_openSrcFile(const char* srcFileName)
         return stdin;
     }
 
+    if (!UTIL_fileExist(srcFileName)) {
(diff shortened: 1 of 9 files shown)


programs/fileio.c
```diff
@@ -388,6 +388,12 @@ static FILE* FIO_openSrcFile(const char* srcFileName)
         return stdin;
     }
 
+    if (!UTIL_fileExist(srcFileName)) {
+        DISPLAYLEVEL(1, "zstd: %s : No such file or directory (can't stat) -- ignored \n",
```
terrelln: I think we should print

```c
DISPLAYLEVEL(1, "zstd: %s: %s -- ignored \n", srcFileName, strerror(errno));
```

If the file doesn't exist the strerror is `No such file or directory`.

Cyan4973: I wasn't sure if it would work properly after `stat()`, but indeed, it does !
Fixed

PR #1472, opened 2018-12-23
Modernize cmake syntax

hjmjohnson:


Cyan4973:
Thanks @hjmjohnson ,
my understanding is that this patch does not change behavior, just coding style,
with the most visible being all keywords converted from uppercase to lowercase.

Quick question : do you have any link where "preferred style" is defined ?

hjmjohnson:
@Cyan4973 You are correct that this only changes the coding style.

Regarding the style, all cmake commands in the documentation (back to version 2.6 https://cmake.org/cmake/help/cmake2.6docs.html) have been in lower case.  The original motivation for the ITK, VTK, Slicer, BRAINSTools, projects was to make the style consistent, and in consultation with the cmake developers at Kitware we moved to lower case formatting.

Brad King (one of the primary architects of cmake and one of the original developers) help to write the scripts applied to this code base.

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/ITKv5Preparation/cmakeRemoveBlockEndCruft.sh

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/ITKv5Preparation/cmakeToLowerCase.sh

I've successfully applied these scripts to dozens of repositories (some of them very large repos like the ITK project above.).

Thank you for considering this patch!

Cyan4973:
Thanks @hjmjohnson,
I believe the patch itself is fine.

I was more thinking along the line : "what will happen when a future contributor will eventually introduce his own changes, messing with the style rules introduced by this patch ?"

In order to guide future developers in preserving the recommended code style,
it would be great to have a short doc file in `build/cmake/README.md` 
which merely states that contributors are invited to respect a coding style,
and linking to a place where said coding style is described.

build/cmake/CMakeLists.txt
```diff
@@ -7,9 +7,12 @@
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
 
-cmake_minimum_required(VERSION 2.8.9)
+cmake_minimum_required(VERSION 3.0.0...3.13.2 FATAL_ERROR)
```
Cyan4973: Why increasing the version requirement from v2.8.9 ?
That change can be problematic, because not all systems are up to date.

build/cmake/CMakeLists.txt
```diff
@@ -7,9 +7,12 @@
...
+
+project(zstd
+VERSION 1.3.7
```
Cyan4973: Is there a way to determine version programmatically from `lib/zstd.h` ?
It would avoid future synchronization errors.

build/cmake/README.md
```diff
@@ -0,0 +1,65 @@
+# Cmake contributions
```
Cyan4973: :+1:

hjmjohnson:
@Cyan4973 Sorry about the cmake version change.  That was an incomplete test that was a mistake in hast to get the readme submitted.

PR #1474, opened 2018-12-24
Cmake set default build type

hjmjohnson:


build/cmake/CMakeLists.txt
```diff
@@ -7,14 +7,43 @@
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
 
-cmake_minimum_required(VERSION 2.8.9)
+cmake_minimum_required(VERSION 2.8.9...3.13.2 FATAL_ERROR)
+
+#-- Extract version information from zstd.h file -------------------------
```
lzutao: I think we could use cmake module from `build/cmake/CMakeModules`.

build/cmake/CMakeLists.txt
```diff
@@ -7,14 +7,43 @@
...
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/../../lib/zstd.h" zstd_main_header)
+string(REGEX MATCH "define *ZSTD_VERSION_MAJOR *([0-9]*)"   _ ${zstd_main_header})
+set(PROJECT_VERSION_MAJOR ${CMAKE_MATCH_1})
+string(REGEX MATCH "define *ZSTD_VERSION_MINOR *([0-9]*)"   _ ${zstd_main_header})
+set(PROJECT_VERSION_MINOR ${CMAKE_MATCH_1})
+string(REGEX MATCH "define *ZSTD_VERSION_RELEASE *([0-9]*)" _ ${zstd_main_header})
+set(PROJECT_VERSION_PATCH ${CMAKE_MATCH_1})
```
lzutao: I think we could use cmake module from `build/cmake/CMakeModules`.

hjmjohnson: @lzutao I do not understand your comment.  Could you please elaborate on the intended change you would like to see?

lzutao: The file `build/cmake/CMakeModules/GetZstdLibraryVersion.cmake` do extract version from `libzstd.h` too. You could use that module instead.

hjmjohnson:
Decoupled from other improvements so that this change can be considered in isolation.

build/cmake/CMakeLists.txt
hjmjohnson: @lzutao Thank you for pointing this out. I had not seen that file.  The requested change was performed in a separate #1475.

PR #1475, opened 2018-12-24
Cmake version setting

hjmjohnson:


build/cmake/CMakeLists.txt
```diff
@@ -7,17 +7,54 @@
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
 
-cmake_minimum_required(VERSION 2.8.9)
+cmake_minimum_required(VERSION 2.8.9...3.13.2 FATAL_ERROR)
```
Cyan4973: I'm trying to understand this line, as it's new to me.

It states that the _minimum_ `cmake` version required is _anything between `2.8.9` and `3.13.2`_ ?
In which way is it different from stating that the minimum needed is `2.8.9` ?

Also : I read that `FATAL_ERROR` is the resulting action if the `cmake` version is `< 2.8.9`. Is that correct ?

What happens if `cmake` version is `> 3.13.2` ?
I guess this is dealt with in following paragraph, but just want to be sure I understand this correctly.

hjmjohnson: This is a new syntax that is only properly supported in 3.12+.   While not supported in previous versions, it is silently ignored in cmake less than 3.11.

I have removed the range of values. It makes more sense once the minimum is at least 3.0 (or at least closer to 3.12).    

You are correct that if cmake version is less than 2.8.9 that configuration will stop immediately with a meaning ful message.   

Thank you for the feedback.

Hans

PR #1477, opened 2018-12-26
ENH: Update to -DZSTD_LEGACY_SUPPORT=5

hjmjohnson:
Update ZSTD_LEGACY_SUPPORT to 5 based on a request by @lzutao in #1474.


Cyan4973:
Yes, it's more in line with current default setting in the `Makefile`  
(see https://github.com/facebook/zstd/tree/dev/lib#modular-build) .

PR #1491, opened 2019-01-05
Fix #1425 - Use physical core count API on FreeBSD

### LINKED ISSUE #1425: Zstd -T0 infers physical CPU cores on Linux and MacOS, but uses all logical CPUs on FreeBSD
[Enhancement]

Admittedly, this is our (FreeBSD) fault.  We do not yet expose this information in an accessible way.  I do not consider this enhancement request actionable for now, I just want to track the issue.  I'm working on the FreeBSD side to provide the needed information.

([Source code link](https://github.com/facebook/zstd/blob/master/programs/util.h#L772).)

Thanks!

Edit: Committed upstream as [r342771](https://svnweb.freebsd.org/base?view=revision&revision=r342771); was: [FreeBSD review here](https://reviews.freebsd.org/D18322).

cemeyer:
Similar to Apple, use the native physical core count sysctl, when available.

This is a little repetitive (it's basically the `__APPLE__` method plus the
otherBSD method concatenated together) but seemed clearer than any way that
would totally eliminate repetition.

The `__FreeBSD_version` check only tests the version of the FreeBSD kernel
that zstd is compiled on; importantly, it may be run on a different version.
So the compile-time check is a little naive and needs to be able to fallback
to work on older versions of FreeBSD.  For a similar reason, it may make
sense to simply eliminate the `__FreeBSD_version` check entirely.  The
tradeoff is that a spurious `sysctlbyname` would be issued when `-T0` is used on
older kernels.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -640,10 +640,42 @@ failed:
     }
 }
 
-#elif defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__DragonFly__)
+#elif defined(__FreeBSD__)
 
-/* Use apple-provided syscall
- * see: man 3 sysctl */
+#include <sys/param.h>
+#include <sys/sysctl.h>
+
+/* Use physical core sysctl when available
+ * see: man 4 smp, man 3 sysctl */
+int UTIL_countPhysicalCores(void)
+{
+    static int numPhysicalCores = 0; /* freebsd sysctl is native int sized */
+    if (numPhysicalCores != 0) return numPhysicalCores;
+
+#if __FreeBSD_version >= 1300008
+    {   size_t size = sizeof(numPhysicalCores);
+        int ret = sysctlbyname("kern.smp.cores", &numPhysicalCores, &size, NULL, 0);
+        if (ret == 0) return numPhysicalCores;
+        if (errno != ENOENT) {
+            perror("zstd: can't get number of physical cpus");
+            exit(1);
+        }
+        /* sysctl not present, fall through to older sysconf method */
+    }
+#endif
+
+    numPhysicalCores = (int)sysconf(_SC_NPROCESSORS_ONLN);
+    if (numPhysicalCores == -1) {
+        /* value not queryable, fall back on 1 */
+        numPhysicalCores = 1;
+    }
+    return numPhysicalCores;
+}
+
+#elif defined(__NetBSD__) || defined(__OpenBSD__) || defined(__DragonFly__)
+
+/* Use POSIX sysconf
+ * see: man 3 sysconf */
 int UTIL_countPhysicalCores(void)
 {
     static int numPhysicalCores = 0;


cemeyer:
Testing notes:

```
::: Machine topology:
$ dmesg | grep FreeBSD/SMP
FreeBSD/SMP: Multiprocessor System Detected: 4 CPUs
FreeBSD/SMP: 1 package(s) x 2 core(s) x 2 hardware threads
```
```
::: Old kernel:

$ sysctl kern.smp.{cores,threads_per_core,cpus}
sysctl: unknown oid 'kern.smp.cores'
sysctl: unknown oid 'kern.smp.threads_per_core'
kern.smp.cpus: 4

:: New Zstd (built on newer kernel with the sysctl available):

$ build/cmake/programs/zstd -vvv -T0 -c < README.md > /dev/null
...
Note: 4 physical core(s) detected
set nb workers = 4
...

:: Old Zstd

$ zstd -vvv -T0 -c < README.md >/dev/null
...
Note: 4 physical core(s) detected
set nb workers = 4
```
(I.e., fallback to sysconf(3) is not a regression on older systems.)

```
::: New kernel:

$ sysctl kern.smp.{cores,threads_per_core,cpus}
kern.smp.cores: 2
kern.smp.threads_per_core: 2
kern.smp.cpus: 4

:: New Zstd

$ build/cmake/programs/zstd -vvv -T0 -c < README.md > /dev/null
...
Note: 2 physical core(s) detected
set nb workers = 2
...

:: Old Zstd

$ zstd -vvv -T0 -c < README.md >/dev/null
...
Note: 4 physical core(s) detected
set nb workers = 4
...
```
(With new kernel sysctls available, and this change, zstd correctly detects physical CPUs on FreeBSD.)

Cyan4973:
Everything looks good to me !

By the way, opened question,
do you know a way to run tests on *BSD using public CI systems ?

Our current test chain includes multiple flavors of Linux, OS-X and Windows,
but we have not found an accessible way to test on FreeBSD in CI so far.
Such a capability would be an improvement over manual testings in a local VM.

cemeyer:
> Everything looks good to me !

Thanks!
 
> By the way, opened question,
> do you know a way to run tests on *BSD using public CI systems ?

I know Cirrus CI supports FreeBSD VMs on GCE: https://cirrus-ci.org/guide/FreeBSD/ .  I think small VMs are free for open source projects.  But only released images published by FreeBSD Foundation, which won't include things like this (not yet in any release).

[Here is a half-baked example use](https://github.com/cemeyer/hadoofus/blob/master/.cirrus.yml).

> Our current test chain includes multiple flavors of Linux, OS-X and Windows,
> but we have not found an accessible way to test on FreeBSD in CI so far.
> Such a capability would be an improvement over manual testings in a local VM.

Yes, makes a lot of sense to me!

programs/util.c
```diff
@@ -640,10 +640,42 @@ int UTIL_countPhysicalCores(void)
...
+int UTIL_countPhysicalCores(void)
+{
+    static int numPhysicalCores = 0; /* freebsd sysctl is native int sized */
+    if (numPhysicalCores != 0) return numPhysicalCores;
+
+#if __FreeBSD_version >= 1300008
+    {   size_t size = sizeof(numPhysicalCores);
+        int ret = sysctlbyname("kern.smp.cores", &numPhysicalCores, &size, NULL, 0);
+        if (ret == 0) return numPhysicalCores;
+        if (errno != ENOENT) {
```
lzutao: Does it need to set `errno` to 0 before `sysctlbyname` call?

cemeyer: I don't believe so — the earlier `ret == 0` early return is sufficient to avoid the cases where `errno` is not set.

From the manual page on FreeBSD:
```
RETURN VALUES
     Upon successful completion, the value 0 is returned; otherwise the value -1
     is returned and the global variable errno is set to indicate the error.
```

lzutao: :+1:

programs/util.c
```diff
@@ -640,10 +640,42 @@ int UTIL_countPhysicalCores(void)
...
+            perror("zstd: can't get number of physical cpus");
+            exit(1);
```
lzutao: Could it be recoverable instead of exit?

cemeyer: It is possible to add, but I don't think a recovery path makes sense here.

I followed the existing convention from the `__APPLE__` [section above](https://github.com/facebook/zstd/pull/1491/files#diff-0d55920a6e9a92f8408dbf7a76c994bfR567) for unexpected sysctl failures.

If the `sysctl` succeeds, we're done; if the sysctl fails, and the `errno` was `ENOENT` (i.e., we're running on an older kernel predating the interface), we have a recovery path via the older `sysconf` method.

The only case where this termination occurs is when `sysctl` returns error status and `errno` is not `ENOENT`.  The only plausible scenarios for that occurring are programmer error (userspace side), which does not seem to be the case; or a radical change in kernel sysctl topology (`kern.smp.cores` is modified into an intermediary node with leaves below it, e.g., `kern.smp.cores.foo`).  I think those scenarios can be set aside, for much the same reason(s) they are set aside on MacOS.

cemeyer:
Thanks @Cyan4973 and @lzutao !

PR #1493, opened 2019-01-06
Fix #1428 - zstdgrep returned 1 on both matched and unmatched file

### LINKED ISSUE #1428: zstdgrep returns 0 on non-match

lzutao:


tests/Makefile
```diff
@@ -355,6 +355,10 @@ test-gzstd: gzstd
 test-zstdgrep: gzstd
 	@echo a | $(PRGDIR)/zstd | $(PRGDIR)/zstdgrep a
 	@echo a | $(PRGDIR)/zstd | $(PRGDIR)/zstdgrep b && return 1 || return 0
+	-[ -f zstdcat ] || ln -s $(PRGDIR)/zstd zstdcat
```
Cyan4973: The test seems to leave a `zstdcat` link in the working directory.
Is that intentional and beneficial ?
The links seems only useful for the test itself.

lzutao: The `test-lz4` test does leave `lz4` in cwd too. But I'd move `zstdcat` link to `/tmp`.

lzutao:
@Cyan4973 Off-topic, I think we could run `scan-build` with meson and ninja. 

**Pros**:
- It is faster than `make staticAnalyze`
- It is directly supported by meson and ninja

**Cons**:
- Some test files (`.c`) are not included in Meson build

tests/Makefile
Cyan4973: Yep, that's a good choice

Cyan4973:
I agree on the `meson` `scan-build` proposal.

The current `make staticAnalyze` is way too long, 
because it analyzes the same files multiple times (the ones which belong to multiple projects). 
It would need to be refactored.

lzutao:
Oops, there's an error in `test-zstdgrep` when creating link in `/tmp`.
Let's me fix this.

Issue #1489, opened 2019-01-03
VERSION_LESS_EQUAL in CMake < 3.7 - Unknown arguments specified


erikwebb:
I am working on building [mcrouter](https://github.com/facebook/mcrouter) and am unable to build the zstd dependency due to the following error -

```
+ git clone https://github.com/facebook/zstd
Cloning into 'zstd'...
+ cd /usr/local/mcrouter/pkgs/zstd
+ cmake -DCMAKE_INSTALL_PREFIX=/usr/local/mcrouter/install build/cmake/
CMake Error at CMakeLists.txt:18 (if):
  if given arguments:

    "3.5.1" "VERSION_LESS_EQUAL" "3.13.2"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
```

According to release notes, `VERSION_LESS_EQUAL` was not added to CMake [until 3.7](https://cmake.org/cmake/help/v3.7/release/3.7.html?highlight=version_less_equal). This change was recently made by @hjmjohnson in 97d1de3d22c5ad53a465d76a5708a4ba90d0ed8b.

Before I move forward with a fix, is the right path forward to change the [CMakeLists.txt conditional](https://github.com/facebook/zstd/blob/dev/build/cmake/CMakeLists.txt#L15-L23) to check for `${CMAKE_MINOR_VERSION}` as well? I'm not very familiar with CMake.

Cyan4973:
I'm not familiar with `cmake` either,
but your suggestion to use `${CMAKE_MINOR_VERSION}` looks good to me.

Cyan4973:
Also : minor detail : 
I would recommend using the `master` branch for dependency build, 
instead of default `dev` branch. 

`master` is very stable, basically a mirror of latest release, 
while `dev` changes all the time, so it can contain transcient errors/problems.

erikwebb:
@Cyan4973 Interesting. I am just using the default mcrouter build. Maybe I should update their build as well.

vitaut:
Looks like the fix in https://github.com/facebook/zstd/commit/9448a3790dec147ac05a3ba02bcbb52a5b9b090a#diff-2f38feedf9af0f5288a86bd6f203d7ea introduces an error for CMake versions [3.0, 3.6) because `ZSTD_CMAKE_POLICY_VERSION` is no longer set:

```
CMake Error at CMakeLists.txt:24 (cmake_policy):
  cmake_policy VERSION not given an argument
```

[...]

PR #1496, opened 2019-01-08
ENH: Simplify conditional logic

hjmjohnson:
Based on excellent comment in #1489, the logic
to set cmake policies was simplified.  This will
provide an easier mechanism for maintaining the
conditional logic across many versions of cmake.


build/cmake/CMakeLists.txt
```diff
@@ -11,15 +11,14 @@ cmake_minimum_required(VERSION 2.8.9 FATAL_ERROR)
...
 # Set and use the newest cmake policies that are validated to work 
-set(ZSTD_MAX_VALIDATED_CMAKE_VERSION "3.13.2") 
-if("${CMAKE_MAJOR_VERSION}" LESS 3) # Cmake version <3.7 does not understand the VERSION_LESS_EQUAL operator
+set(ZSTD_MAX_VALIDATED_CMAKE_MAJOR_VERSION "3") 
+set(ZSTD_MAX_VALIDATED_CMAKE_MINOR_VERSION "13") #Policies never changed at PATCH level
+if("${CMAKE_MAJOR_VERSION}" LESS 3)
   set(ZSTD_CMAKE_POLICY_VERSION "${CMAKE_VERSION}") 
-elseif(("${CMAKE_MAJOR_VERSION}" EQUAL 3 AND "${CMAKE_MINOR_VERSION}" GREATER 6) OR ("${CMAKE_MAJOR_VERSION}" GREATER 3))
-  if("${CMAKE_VERSION}" VERSION_LESS_EQUAL "${ZSTD_MAX_VALIDATED_CMAKE_VERSION}") 
+elseif( "${ZSTD_MAX_VALIDATED_CMAKE_MINOR_VERSION}" GREATER "${CMAKE_MINOR_VERSION}")
```
Cyan4973: What if, sometimes in the future, some `4.0` version gets released ?
Would it use this branch ?

hjmjohnson: No, not until zstd is verified to work with 4.0+, and then we would need to update lines 14 & 15 to the latest validated version.

In fact if new policies are added in cmake 3.14, those would not be enforced until they are validated and lines 14 & 15 are updated.

Cyan4973: I'm concerned by the following scenario : 
if a `4.0` version exist one day, I expect : 
- `CMAKE_MAJOR_VERSION == 4`
- `CMAKE_MINOR_VERSION == 0`

Consequently, first test `("${CMAKE_MAJOR_VERSION}" LESS 3)` is false,
but second test `( "${ZSTD_MAX_VALIDATED_CMAKE_MINOR_VERSION}" GREATER "${CMAKE_MINOR_VERSION}")` is true
because `ZSTD_MAX_VALIDATED_CMAKE_MINOR_VERSION =13` which is `>` to `CMAKE_MINOR_VERSION == 0`.

Hence the second branch is taken,
which results in `set(ZSTD_CMAKE_POLICY_VERSION "${CMAKE_VERSION}") ` with `CMAKE_VERSION == 4.0`,
and I suspect that's not an intended outcome, since you want policy pinned to `3.13`.

hjmjohnson:
@Cyan4973 You are 100% correct.  Sorry for the lazy review of your previous comment.  I will fix soon.

PR #1505, opened 2019-01-22
Remove global parameters, pass into public functions instead

kostmo:
This is a work in progress.

### DIFF
--- a/programs/fileio.h
+++ b/programs/fileio.h
@@ -42,44 +42,52 @@ extern "C" {
 ***************************************/
 typedef enum { FIO_zstdCompression, FIO_gzipCompression, FIO_xzCompression, FIO_lzmaCompression, FIO_lz4Compression } FIO_compressionType_t;
 
+typedef struct FIO_prefs_s FIO_prefs_t;
+
+FIO_prefs_t* FIO_createPreferences(void);
+void FIO_freePreferences(FIO_prefs_t* const prefs);
+
+typedef struct FIO_display_prefs_s FIO_display_prefs_t;
 
 /*-*************************************
 *  Parameters
 ***************************************/
-void FIO_setCompressionType(FIO_compressionType_t compressionType);
-void FIO_overwriteMode(void);
-void FIO_setAdaptiveMode(unsigned adapt);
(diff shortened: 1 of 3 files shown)


Cyan4973:
Hi @kostmo . You will need a `struct`, as the number of parameters to transmit is too large to become part of argument list.

Also, @tzakian made a similar patch for `lz4`, at : https://github.com/lz4/lz4/pull/640
It might be inspiring, as both programs share a similar code structure.

kostmo:
All CI checks passed.  I kept `displayLevel` the `noProgress` as `static` in the module due to their use by the `DISPLAY*` macros.

programs/fileio.h
```diff
@@ -42,44 +42,52 @@ extern "C" {
 ***************************************/
 typedef enum { FIO_zstdCompression, FIO_gzipCompression, FIO_xzCompression, FIO_lzmaCompression, FIO_lz4Compression } FIO_compressionType_t;
 
+typedef struct ZSTDIO_prefs_s ZSTDIO_prefs_t;
+
+ZSTDIO_prefs_t* ZSTDIO_defaultPreferences(void);
```
Cyan4973: some minor details, on naming convention : 

1. All symbols in this unit are prefixed `FIO_`.
Is there a reason to prefer prefixing these new symbols with `ZSTDIO_` instead ?

2. `ZSTDIO_defaultPreferences()` is likely inspired by `LZ4IO_defaultPreferences()`, so that's understandable.
I would nonetheless prefer `*_createPreferences()` instead if possible.

Reasoning : I would expect `*_defaultPreferences()` to return the structure itself, by value.
But what's happening here is slightly different : we get a pointer to some heap memory area.
As a consequence, the caller must also remember to deallocate this space afterwards.

`create` helps to realize there is this responsibility at stake.

Cyan4973:
Thanks @kostmo , it's great work !
Only some minor comments on naming convention, and it's basically good to go !

programs/fileio.h
kostmo: Fixed both naming issues.

PR #1508, opened 2019-01-25
fixed fileio.c compilation with LZ4 enabled

Cyan4973:
was broken by #1505.

I'm surprised it passed CI tests.
LZ4 tests are part of the "Extended" tests on Travis CI,
which are run on "master" and in "cron" jobs.

Since latest cron job did not failed,
especially this one : https://travis-ci.org/facebook/zstd/jobs/484365040
it suggests cron jobs are no longer using `dev` branch.

To be investigated

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -195,7 +195,7 @@ static void ABRThandler(int sig) {
     const char* name;
     void* addrlist[MAX_STACK_FRAMES];
     char** symbollist;
-    U32 addrlen, i;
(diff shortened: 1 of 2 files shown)


Cyan4973:
The end signal is red, due to a spurious error on Appveyor, that cannot be reproduced. For some reason, I can't remove the red flag (it's not even listed on Appveyor's Dashboard !), so it sticks around, but the test result should actually be green.

Issue #1435, opened 2018-12-03
Add Debug Log Statements for Decompression Error Conditions?


felixhandte:
When diagnosing corrupt compressed blobs, I frequently get a generic "Corrupted block detected" error message, which can be caused by all sorts of different things.

What do you think about adding a `DEBUGLOG` statement everywhere we return an error, so that diagnosing these blobs becomes as simple as rebuilding zstd with `-DDEBUGLEVEL=999`. In many cases the comment that is already on the condition being checked would be the perfect message.

I'm happy to work up a patch that does this.

What do you think about introducing a macro? E.g.,

```
#define RETURN_ERROR(code, ...) do { DEBUGLOG(2, __VA_ARGS__); return code; } while (0)
#define CHECK_M(code, ...) do { if (ERR_isError(code)) { RETURN_ERROR(code, __VA_ARGS__); } } while(0)
```

And then converting existing `return ERROR(corruption_detected);` sites to use the new format with some error message?

Cyan4973:
Agreed.
Level 2 is for temporary debug, for conditions that do not remain in the code.
Level 3+ are adapted for debug messages that remain present in source code.
It's also possible to use a constant with a default value that can be overridden on command line.

Cyan4973:
Is there a PR associated to this issue ?

PR #1509, opened 2019-01-28
Add Debug Information to Error Checks

felixhandte:
This diff wraps the error handling code in the decompressor so that debug information tracking the source of errors can be logged. This is intended to help diagnose issues with corrupt compressed blobs.

Addresses #1435.

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -53,7 +53,7 @@ extern "C" {
 #undef MAX
 #define MIN(a,b) ((a)<(b) ? (a) : (b))
 #define MAX(a,b) ((a)>(b) ? (a) : (b))
-#define CHECK_F(f) { size_t const errcod = f; if (ERR_isError(errcod)) return errcod; }  /* check and Forward error code */
+
 #define CHECK_E(f, e) { size_t const errcod = f; if (ERR_isError(errcod)) return ERROR(e); }  /* check and send Error code */
 
 /**
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -63,7 +63,7 @@ extern "C" {
  */
 #define RETURN_ERROR_IF(cond, err, ...) \
   if (cond) { \
-    RAWLOG(3, "%s:%d: check %s failed, returning %s", __FILE__, __LINE__, ZSTD_QUOTE(cond), ZSTD_QUOTE(ERROR(err))); \
+    RAWLOG(3, "%s:%d: ERROR!: check %s failed, returning %s", __FILE__, __LINE__, ZSTD_QUOTE(cond), ZSTD_QUOTE(ERROR(err))); \
     RAWLOG(3, ": " __VA_ARGS__); \
     RAWLOG(3, "\n"); \
     return ERROR(err); \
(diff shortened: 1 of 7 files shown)


lib/common/zstd_internal.h
```diff
@@ -56,6 +56,48 @@ extern "C" {
 #define CHECK_F(f) { size_t const errcod = f; if (ERR_isError(errcod)) return errcod; }  /* check and Forward error code */
 #define CHECK_E(f, e) { size_t const errcod = f; if (ERR_isError(errcod)) return ERROR(e); }  /* check and send Error code */
 
+/**
+ * Return the specified error if the condition evaluates to true.
+ *
+ * In debug modes, prints additional information.
+ */
+#define RETURN_ERROR_IF(cond, err, ...) \
```
Cyan4973: So, the main reason why `RETURN_ERROR_IF(...)` is not a simple `if (cond) RETURN_ERROR(...)`
is to be able to log `cond` itself ?

That would a good enough reason, since I don't see any other way around.

Maybe add a comment about that in the macro description.

lib/common/zstd_internal.h
```diff
@@ -56,6 +56,48 @@ extern "C" {
...
+    RAWLOG(3, "\n"); \
+    return ERROR(err); \
+  } while(0);
+
+/**
+ * If the provided expression evaluates to an error code, returns that error code.
+ *
+ * In debug modes, prints additional information.
+ */
+#define FORWARD_ERROR(err, ...) \
```
Cyan4973: `FORWARD_ERROR()` is equivalent to `CHECK_F()`,
but add support for traces.

Is there a reason to have 2 separated variants ?
Would it be reasonable to fold logging support into `CHECK_F()` ?

lib/common/zstd_internal.h
felixhandte: Yep, that's right. Happy to add a comment.

lib/common/zstd_internal.h
felixhandte: Yeah, I think it makes sense to eventually replace the `CHECK_*` macros. The barriers to that are:

1. I've only converted the zstd decompression code. I'd need to also convert all the compression code. I'm happy to do that work, but I wanted to put this PR up first to make sure everyone's happy with the approach taken.
2. The huffman codebase also defines its own `CHECK_{E,F}` macros, I guess because it needs to be freestanding from the zstd codebase. Should I convert those too? Or leave them be?

Cyan4973: 1. If the modifications are done directly within `CHECK_F`, it's unnecessary to change the rest of the code to take advantage of it. 
  - There is a point though that maybe `FORWARD_ERROR()` is more expressive than `CHECK_F()`. I'm fine with that. That would require to change all `CHECK_F()` invocations though.
  - In which case, I was wondering about the name `FORWARD_IF_ERROR()`, because `FORWARD_ERROR()` feels like a definitive statement. But it's not what's happening.

2. Yes, keep Huffman code separate for the time being. We can try to unify that later on.

Cyan4973:
From my perspective, this patch is good to go.
If you don't have anything left to add, we could merge it.

felixhandte:
Yep, I'm done.

PR #1499, opened 2019-01-15
fix --list on truncated files

kostmo:
fseek() doesn't indicate when it moves past the end of a file.
Consequently, if a file is truncated within its last block, the error would't be detected.

This PR adds a test scenario that induces this situation using a small compressed file of only one block in size.
This test is added to tests/playTests.sh

Check is implemented by ensuring that the filehandle position is equal to the filesize upon exit.

PR #1518, opened 2019-02-11
'head -c BYTES' is non portable

bket:
#1499 added a new test, which uses `head -c`. The `-c` option is non-portable (not in POSIX). Instead use `dd`. Similar issue has been resolved in the past (#1321).


terrelln:
Thanks for the patch! We have a new CI, Cirrus-CI, which runs on BSD. Right now it only compiles the project, but I'll look into having it run `playTests.sh`, which should catch this issue.

terrelln:
Well, `head` is available on FreeBSD, so the test doesn't catch it. However, it looks like FreeBSD doesn't support `lstat()` unless you define a particular macro #1520. Do you know if OpenBSD is the same?

This should emit a warning to use `--force` to compress symbolic links, and not compress `world`.

```
echo hello > hello
ln -s hello world
zstd world
```

bket:
> Do you know if OpenBSD is the same?

Guess it is. See #1521

Issue #1520, opened 2019-02-11
Detect symbolic links on FreeBSD


terrelln:
When compressing files, zstd will typically ignore symbolic links and warn that you need to pass `--force`. On FreeBSD, zstd doesn't detect that [lstat()](https://github.com/facebook/zstd/blob/dev/programs/util.c#L90) is supported, so it will compress the target of symbolic links.

While this isn't the end of the world, it is slightly, silently inconsistent behavior. We would likely need to define some macros in [platform.h](https://github.com/facebook/zstd/blob/dev/programs/platform.h) to enable `lstat()` on FreeBSD.

This issue might also affect OpenBSD.

PR #1521, opened 2019-02-11
Detection of symbolic links on OpenBSD

bket:
In #1520 it is described that FreeBSD doesn't detect symbolic links. The same is true for OpenBSD. This diff fixes this issue for OpenBSD. I'm guessing that something similar works for FreeBSD as well. However, I'm unable to test this.

While working on the above I noticed some inconsistent behaviour:

```
$ echo hello > hello
$ ln -s hello world
$ zstd hello world
Warning : world is a symbolic link, ignoring
hello                :316.67%   (     6 =>     19 bytes, hello.zst
$ ls *.zst
hello.zst
$ zstd world
world                :316.67%   (     6 =>     19 bytes, world.zst)
$ ls *.zst
hello.zst world.zst

```
I'm guessing that the latter is not OpenBSD specific?

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -951,6 +951,8 @@ int main(int argCount, const char* argv[])
                 filenameTable[fileNamesNb++] = filenameTable[u];
             }
         }
+        if (fileNamesNb == 0 && filenameIdx > 0)
+            CLEAN_RETURN(1);
         filenameIdx = fileNamesNb;
     }
     if (recursive) {  /* at this stage, filenameTable is a list of paths, which can contain both files and directories */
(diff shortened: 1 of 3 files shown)


programs/zstdcli.c
```diff
@@ -947,6 +947,8 @@ int main(int argCount, const char* argv[])
         for (u=0, fileNamesNb=0; u<filenameIdx; u++) {
             if (UTIL_isLink(filenameTable[u])) {
                 DISPLAYLEVEL(2, "Warning : %s is a symbolic link, ignoring\n", filenameTable[u]);
+                if (filenameIdx == 1)
```
terrelln: We'll still be open to the same bug if we have 2 files which are both symlinks.

bket: Yes, your feedback makes sense. Commit received an update.
I have updated the test case for symbolic links, and have enabled it for OpenBSD.

Tested OK on OpenBSD.

PR #1522, opened 2019-02-13
Add ZSTD_c_literalCompressionMode flag

### LINKED ISSUE #1517: Compression with negative levels worse on latest version compared to 1.3.4
We are trying to update from 1.3.4 to 1.3.8. Most of our compression is done streaming at level -3, which seems to work well for us. With version 1.3.8 compression at -3 is faster (5% - 10%) but the compression rate is worse. In one case the rate went from 0.27 to 0.55. I had a look at individual parameters, but I can only improve the rate by switching strategy from fast to dfast, at level -3. Sadly, now the compression speed is significantly lower than with 1.3.4.

Whatever I try, I don't seem to be able to get the same speed and compression rate with 1.3.8 as we get with 1.3.4 at level -3. What could be the cause of this? We compress large amounts of small images, no more than a few kb each, using ZSTD_compressStream2().

terrelln:
It controls the literals compression. It is either `auto`, `huffman`, or `uncompressed`. It defaults to `auto`, which is the current behavior.

Add a regression test that tests a positive compression level with uncompressed literals, and a negative compression level with compressed literals. I checked the `results.csv` table and made sure that the compressed sizes make sense.

Fixes #1517.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -360,6 +360,12 @@ ZSTD_bounds ZSTD_cParam_getBounds(ZSTD_cParameter param)
         bounds.upperBound = ZSTD_dictForceCopy;       /* note : how to ensure at compile time that this is the highest value enum ? */
         return bounds;
 
+    case ZSTD_c_literalCompressionMode:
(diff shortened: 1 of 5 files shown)


Cyan4973:
That's excellent.
Minimum amount of change, and great continuous tests !

PR #1525, opened 2019-02-15
Fix optimal parser prices with uncompressed literals

terrelln:
* Disallow updating `ZSTD_c_literalCompressionMode` during compression, since the optimal parsers statistics would be messed up.
* Tell the optimal parser that literals are 1 byte each when using `ZSTD_lcm_uncompressed`, and avoid updating the literal frequencies.
* Add a flag on the CLI to control literals compression `--[no-]compress-literals`.
* Update the regression tests to test level 19 with uncompressed literals. 
3d7377b shows the difference after fixing the pricing (-2% on silesia, -5% on github).
* Fix a bug in the `compress cctx` method that was causing decompression failures.
* Respect `--[no-]compress-literals` in benchmark mode
* Add CLI tests to `playTests.sh` so they get executed with different configurations.


Cyan4973:
Looks great !

2 minor points  : 

> * Disallow updating `ZSTD_c_literalCompressionMode` during compression, since the optimal parsers statistics would be messed up.

- What prevents the optimal parser from switching mode between blocks ?

> * Add a flag on the CLI to control literals compression `--[no-]compress-literals`.

- Could this flag be combined with benchmark mode ?

terrelln:
> * What prevents the optimal parser from switching mode between blocks ?

We skip updating the literals statistics when we are using fixed 1-byte literals. The optimal parser wouldn't have up to date literal costs when it switched back. Also, a more minor issue, is that all the stats will be a little bit off if you switch between blocks, since we were previously using a 1-byte literal cost model.

This is not a hard limitation, but for ease of implementation, since I don't expect many people will want to switch on and off literal compression between blocks. If that isn't the case, we can lift the restriction.

Issue #1527, opened 2019-02-19
Windows release issues / NuGet


edenhill:
Found the following issues with the win32 & win64 releases when creating a NuGet package of zstd for use with [librdkafka](https://github.com/edenhill/librdkafka):
 * libzstd.dll, .lib are missing "advanced/experimental" APIs, such as ZSTD_getErrorCode(). Since DLLs are typically bundled with the application on Windows (i.e., no system install), it would be better for these libs to be built with full support as versioning is not really an issue in that case.
 * zstd_error.h is in example/, but should really be in include/ and part of the non-experimental build. (see Debian and Brew packages)
 * zstd_static.lib seems to be built with mingw and is not compatible with MSVC/VS builds (___chkstk_ms symbol missing)
 * win32 libzstd.lib is built without [SAFESEH](https://docs.microsoft.com/en-us/cpp/build/reference/safeseh-image-has-safe-exception-handlers?view=vs-2017), requiring the dependee (e.g., librdkafka) to pass `/SAFESEH:NO` to the MSVC linker.

I've created a NuGet package builder that uses the win32 & win64 release zips, it is usable with some minor caller hacks (such as mimicking the missing ZSTD_getErrorCode(), disable experimental options, ..).
NuGet package available here https://www.nuget.org/packages/confluent.libzstd.redist, but it is for experimental use only. (hint: NuGet packages are zip files with a .nupkg extension)

I intend to submit a PR with the NuGet builder, it would be great if the zstd project could maintain an official NuGet package.

terrelln:
I can pretty easily fix 1 and 2. I'll look into fixing points 3 and 4, but it might take me a while, since I'm not a Windows native. In the meantime, we build our Windows binary with Appveyor, so any PRs to our [appveyor.yml](https://github.com/facebook/zstd/blob/dev/appveyor.yml) are extremely welcome!

terrelln:
It looks like `libzstd.dll` exposes all the experimental symbols, but `libzstd.lib` doesn't (using `nm`). If you found that `libzstd.dll` doesn't export some symbol please let me know.

terrelln:
* Some googling around `SAFESEH` shows that it is MSVC specific, so MinGW doesn't have an equivalent flag. It would be a larger change to compile artifacts with MSVC, so I'm going to leave it for now (and keep this issue open).
* Why are MinGW static libraries not compatible with MSVC? Is there a way MinGW can generate static libraries compatible with MSVC?

[...]

PR #1528, opened 2019-02-19
NuGet package builder

edenhill:
Also see the issues reported in #1527.

Example usage from a C project in  https://github.com/edenhill/librdkafka/commit/f5107430c1d32f5be850f122c4832d69f42d7cf5

### DIFF
--- a/contrib/nuget/.gitignore
+++ b/contrib/nuget/.gitignore
@@ -1,0 +1,2 @@
+stage-*
+*.nupkg
--- a/contrib/nuget/README.md
+++ b/contrib/nuget/README.md
@@ -1,0 +1,12 @@
+Builds a NuGet package using the Windows release artifacts
+from https://github.com/facebook/zstd/releases
+
+Usage:
+
+    # 1. Build the package (requires docker)
+    $ ./build.sh v1.3.8
+
+    # 2. Check that package looks okay
+    $ unzip -l libzstd.redist.1.3.8.nupkg
+
+    # 3. Upload to https://NuGet.org (via web UI or with 'nuget push')
--- a/contrib/nuget/build.sh
+++ b/contrib/nuget/build.sh
@@ -1,0 +1,116 @@
+#!/bin/bash
+#
+
+#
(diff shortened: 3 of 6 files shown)


terrelln:
Taking a look at the README, do you need some sort of permissions to upload a new package to NuGET (out of curiosity)?

edenhill:
Sign up on NuGet.org (preferably with a Microsoft account), and then upload packages manually through the web UI.
You can also create an access key and use the `nuget push ..` utility to upload from the command line.
I also think AppVeyor can upload artifacts directly.

edenhill:
Will update this PR when new win releases are available with the new structure, as changed in #1531

terrelln:
zstd-1.4.0 has been released, which has the new structure.

bruno-garcia:
Any updates on this? I'm stuck trying to use `zlib` from .NET on a macOS (.NET Core).
Could send a PR after to build the NuGet packages to support not only Windows but also macOS.

Cyan4973:
Joining in.

If I understand correctly, this PR offers a script to generate a Nuget package.

Its dependencies are : names provided to Windows binary packages in the "release" section, and their content (and directory tree).
It's worth mentioning because, also we don't plan to change these things often, it's all too easy to have a change once in a while, and it would be enough to break the script.

And by the way, it seems @terrelln mentions a change in the Windows binary package for `v1.4.0`, so it seems the script in this PR should be updated ?

Also : 
Running the script to create a package locally is one thing, actually pushing it on nuget.org is another one, with its own administrative burden. While I might be able to do it, I wonder if it's the best course of action, since I'm not myself a nuget user, hence I probably don't understand the use cases, and can easily miss its maintenance.

Also, I note that there are already 3 nuget packages proposed with some `zstd` flavor, listed at : 
https://www.nuget.org/packages?q=Tags%3A%22zstd%22

and I'm not yet clear what this one would offer on top of existing ones. 

I must read more the history of exchanges...

Cyan4973:
This PR hasn't changed much in the last few months.
Is it still active ? or should it be closed ?

Issue #1536, opened 2019-02-21
Provide ZSTD_boundDecompressedSize()


terrelln:
When the decompressed size isn't written into the frame, and you want to use the one pass function, you are forced to guess the decompressed size. We could estimate the decompressed size using the number of blocks. It will be no larger than 128 KB larger than the decompressed size if flush isn't used. If flush is used, or some other block level API, then we could estimate way too large.

Example usage (without error checking):

```c++
std::string decompress(void* src, size_t size, size_t maxAlloc) {
    size_t const bound = std::min(maxAlloc, ZSTD_boundDecompressedSize(src, size));
    std::string out;
    out.resize(bound);
    out.resize(ZSTD_decompress(&out[0], out.size(), src, size));
    return out;
}
```

Edit: Implemented by @shakeelrao in PR #1543. This issue is left open to add support for legacy mode before the release.

shakeelrao:
Hi @terrelln,

I'm fairly new to the project (have been using the CLI for benchmarking), but would be interested in implementing this feature. Do you have any additional starting points?

terrelln:
1. Read the block header section of the [format spec](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#blocks). Each block has at most `Block_Maximum_Decompressed_Size` bytes.
2. We will loop over each frame in the input and either use the content size if available, or count the number of blocks and multiple by `Block_Maximum_Decompressed_Size`.
3. We can reuse the code in [ZSTD_findFrameCompressedSize](https://github.com/facebook/zstd/blob/dev/lib/decompress/zstd_decompress.c#L442). We can write a helper function which finds both the compressed size, and bounds the decompress size. Then `ZSTD_findFrameCompressedSize()` can return only the compressed size.
4. We should support multiple frames like [ZSTD_findDecompressedSize()](https://github.com/facebook/zstd/blob/dev/lib/decompress/zstd_decompress.c#L364).

shakeelrao:
Thanks @terrelln. Is the following pseudocode the right approach?

```
bound = 0
for each frame in src:
  if Frame_Content_Size of frame exists:
    bound += Frame_Content_Size of frame
  else
    for each block in frame:
       bound += Block_Maximum_Decompressed_Size
```

Also, where is `Block_Maximum_Decompressed_Size` defined in the code?

shakeelrao:
[...]

PR #1543, opened 2019-02-28
Provide a function to estimate decompressed size.

shakeelrao:
Provides a function to estimate the upper-bound of a compressed message.

Implements https://github.com/facebook/zstd/issues/1536.


shakeelrao:
Sorry for all the failed CI runs. For some reason, the tests don't run to completion on my machine.

For example: `make shortest` outputs:

```
test : detect truncated compressed file 

truncatable-input.txt : 29.53%   ( 50000 =>  14766 bytes, truncatable-input.txt.zst) 
dd: unknown operand status
make[1]: *** [test-zstd] Error 1
make: *** [shortest] Error 2
```

Any suggestions?

lib/decompress/zstd_decompress.c
```diff
@@ -483,10 +480,79 @@ size_t ZSTD_findFrameCompressedSize(const void *src, size_t srcSize)
...
+ *  `src` must point to the start of a ZSTD frame or a skippeable frame
+ *  `srcSize` must be at least as large as the frame contained
+ *  @return : maximum decompressed size of the compressed source
+ */
+unsigned long long ZSTD_decompressBound(const void* src, size_t srcSize)
+{
+    unsigned long long totalDstSize = 0;
+#if defined(ZSTD_LEGACY_SUPPORT) && (ZSTD_LEGACY_SUPPORT >= 1)
+    if (ZSTD_isLegacy(src, srcSize))
+        return ERROR(version_unsupported);
```
terrelln: The error codes are `size_t`, but we are returning `unsigned long long`, so we can't return them.

When this function errors, return `ZSTD_CONTENTSIZE_ERROR`.

terrelln:
To add a test with the frame content size isn't available, you can compress like this:

```c
ZSTD_CCtx* cctx = ZSTD_createCCtx();
CHECK_Z( ZSTD_CCtx_setParameter(cctx, ZSTD_c_contentSizeFlag, 0) );
CHECKPLUS(r, ZSTD_compress2(cctx,
                            compressedBuffer, compressedBufferSize,
                            CNBuffer, CNBuffSize),
                  cSize=r );
ZSTD_freeCCtx(cctx);
```

Look at the documentation for `ZSTD_c_contentSizeFlag` and `ZSTD_compress2()` to see whats going on here.

Just make sure that you do it after anything that relies on `compressedBuffer`, so either the last test, or if the next test also compresses into `compressedBuffer`.

terrelln:
What are you running `make shortest` on? It works for me on Mac and Linux.

You can rely on the CI to run your tests. Facebook's CI will automatically run when you update the PR, but sometimes the queues get long, so you can set up CircleCI / Travis to run on your own repo if you like.

shakeelrao:
Thanks for the detailed feedback, @terrelln! Really appreciate your patience on this :) Will start tackling some of the requested changes tonight after work. 

I am running `makeshortest` on Mac. For some reason the test fails when dealing with the `
truncatable-input.txt` test case.

lib/decompress/zstd_decompress.c
```diff
@@ -434,12 +434,7 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
...
-/** ZSTD_findFrameCompressedSize() :
- *  compatible with legacy mode
- *  `src` must point to the start of a ZSTD frame, ZSTD legacy frame, or skippable frame
- *  `srcSize` must be at least as large as the frame contained
- *  @return : the compressed size of the frame starting at `src` */
-size_t ZSTD_findFrameCompressedSize(const void *src, size_t srcSize)
+static size_t ZSTD_findFrameCompressedSize_internal(const void *src, size_t srcSize, size_t* bound)
 {
 #if defined(ZSTD_LEGACY_SUPPORT) && (ZSTD_LEGACY_SUPPORT >= 1)
     if (ZSTD_isLegacy(src, srcSize))
```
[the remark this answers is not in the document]
shakeelrao: Good idea! Do you think it's reasonable to return `ERROR(version_unsupported)` for now and add legacy support in another PR?

lib/decompress/zstd_decompress.c
shakeelrao: Should I change the function definition to `size_t ZSTD_decompressBound(const void* src, size_t srcSize)` instead? I'm a bit confused between the use of `size_t` and `unsigned long long` in this project.

lib/decompress/zstd_decompress.c
```diff
@@ -433,60 +433,134 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
...
+                frameSizeInfo.compressedSize = ERROR(srcSize_wrong);
+                frameSizeInfo.decompressedBound = ERROR(srcSize_wrong);
+                return frameSizeInfo;
+            }
             ip += 4;
         }
 
-        return ip - ipstart;
+        frameSizeInfo.compressedSize = ip - ipstart;
+        frameSizeInfo.decompressedBound = (zfh.frameContentSize != ZSTD_CONTENTSIZE_UNKNOWN)
```
shakeelrao: Looks like here we assign `size_t` to `unsigned long long`, but changed `decompressedBound` to `unsigned long long` means we won't be able to use the standard error codes. Do you have any suggestions?

shakeelrao:
I think I implemented most of your suggest changes, including the documentation. Let me know if there are any more API level changes. I want to make sure the overall implementation is good before adding the final tests :)

lib/decompress/zstd_decompress.c
```diff
@@ -433,60 +433,135 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
...
+/**
+ * Contains the compressed frame size and an upper-bound for the decompressed frame size.
+ * Note: before using `compressedSize` you must check for errors using ZSTD_isError().
+ *       similarly, before using `decompressedBound`, you must check for errors using:
+ *          `decompressedBound` != ZSTD_CONTENTSIZE_UNKNOWN
+ */
+typedef struct {
+    size_t compressedSize;
+    unsigned long long decompressedBound;
+} ZSTD_frameSizeInfo;
```
terrelln: You'll save some code duplication, and make it harder miss setting a field if you use this helper function.

```c
static ZSTD_frameSizeInfo ZSTD_errorFrameSizeInfo(size_t err)
{
    ZSTD_frameSizeInfo frameSizeInfo;
    frameSizeInfo.compressedSize = err;
    frameSizeInfo.decompressedBound = ZSTD_CONTENTSIZE_UNKNOWN;
    return frameSizeInfo;
}
```

lib/decompress/zstd_decompress.c
```diff
@@ -433,60 +433,135 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
...
+        size_t nbBlocks = 0;
         ZSTD_frameHeader zfh;
 
         /* Extract Frame Header */
-        {   size_t const ret = ZSTD_getFrameHeader(&zfh, src, srcSize);
-            if (ZSTD_isError(ret)) return ret;
-            RETURN_ERROR_IF(ret > 0, srcSize_wrong);
+        {
+            size_t const ret = ZSTD_getFrameHeader(&zfh, src, srcSize);
+            if (ZSTD_isError(ret)) {
```
terrelln:
```c
if (ZSTD_isError(ret))
    return ZSTD_errorFrameSizeInfo(ret);
```

lib/decompress/zstd_decompress.c
```diff
@@ -433,60 +433,135 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
...
+ *  `src` must point to the start of a ZSTD frame or a skippeable frame
+ *  `srcSize` must be at least as large as the frame contained
+ *  @return : the maximum decompressed size of the compressed source
+ */
+unsigned long long ZSTD_decompressBound(const void* src, size_t srcSize)
+{
+    unsigned long long bound = 0;
+    /* Iterate over each frame */
+    while (srcSize > 0) {
+        ZSTD_frameSizeInfo frameSizeInfo = ZSTD_findFrameSizeInfo(src, srcSize);
```
terrelln: nit: Make every variable you can const. Every non-const variable should be mutated. It is a hint to the reader that this variable gets updated.

```c
    ZSTD_frameSizeInfo const frameSizeInfo = ...;
    size_t const compressedSize = ...;
    unsigned long long const decompressedBound = ...;
```

lib/zstd.h
```diff
@@ -1103,6 +1103,21 @@ typedef enum {
...
+ *  currently incompatible with legacy mode
+ *  `src` must point to the start of a ZSTD frame or a skippeable frame
+ *  `srcSize` must be at least as large as the frame contained
+ *  @return : - the maximum decompressed size of the compressed source
+ *            - if an error occured: ZSTD_CONTENTSIZE_ERROR
+ *
+ *  note 1  : an error can occur if `src` points to a legacy frame or an invalid/incorrectly formatted frame.
+ *  note 2  : the bound is exact when Frame_Content_Size field is available in EVERY frame of `src`.
+ *  note 3  : when Frame_Content_Size isn't provided, the upper-bound for that frame is calculated by:
+ *              upper-bound = min(128 KB, Window_Size)
```
terrelln: `upper-bound = # blocks * min(128 KB, Window_Size)`

lib/zstd.h
```diff
@@ -1103,6 +1103,21 @@ typedef enum {
  *            however it does mean that all frame data must be present and valid. */
 ZSTDLIB_API unsigned long long ZSTD_findDecompressedSize(const void* src, size_t srcSize);
 
+/** ZSTD_decompressBound() :
+ *  currently incompatible with legacy mode
+ *  `src` must point to the start of a ZSTD frame or a skippeable frame
+ *  `srcSize` must be at least as large as the frame contained
```
terrelln: This has the same requirements as `ZSTD_findDecompressedSize()`.

```
 *  `src` should point the start of a series of ZSTD encoded and/or skippable frames
 *  `srcSize` must be the _exact_ size of this series	 *  `srcSize` must be the _exact_ size of this series
 *       (i.e. there should be a frame boundary exactly at `srcSize` bytes after `src`)	 *       (i.e. there should be a frame boundary exactly at `srcSize` bytes after `src`)
```

lib/decompress/zstd_decompress.c
shakeelrao: Great idea! I'll add this.

lib/decompress/zstd_decompress.c
shakeelrao: Should I apply this style consistently throughout the code (i.e. if there's only one statement after the if, omit the brackets)?

lib/decompress/zstd_decompress.c
shakeelrao: What do you think about this?

```
static ZSTD_frameSizeInfo ZSTD_errorFrameSizeInfo(size_t err)
{
    return ZSTD_createFrameSizeInfo(err, ZSTD_CONTENTSIZE_ERROR);
}

static ZSTD_frameSizeInfo ZSTD_createFrameSizeInfo(size_t compressedSize, unsigned long long decompressedBound) {
    ZSTD_frameSizeInfo frameSizeInfo;
    frameSizeInfo.compressedSize = compressedSize;
    frameSizeInfo.decompressedBound = decompressedBound;
    return frameSizeInfo;
}
```

lib/zstd.h
shakeelrao: Nice catch!

tests/fuzzer.c
```diff
@@ -901,6 +908,11 @@ static int basicUnitTests(U32 seed, double compressibility)
         if (r != CNBuffSize / 2) goto _output_error; }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : get tight decompressed bound of multiple frames : ", testNb++);
+    {   unsigned long long const r = ZSTD_decompressBound(compressedBuffer, cSize);
+        if (r != CNBuffSize / 2) goto _output_error; }
+    DISPLAYLEVEL(3, "OK \n");
```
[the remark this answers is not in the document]
shakeelrao: If I'm not mistaken, I think `compressedBuffer` contains a skippable [frame](https://github.com/facebook/zstd/pull/1543/files#diff-800cd24573dee9ac3476b9b256b07bf3L888).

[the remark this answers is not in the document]
shakeelrao: Thanks! I believe this test case has a skippable frame [already](https://github.com/facebook/zstd/pull/1543/files#diff-800cd24573dee9ac3476b9b256b07bf3R916).

lib/decompress/zstd_decompress.c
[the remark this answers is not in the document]
shakeelrao: Do you think it's reasonable to return an error for now and implement legacy support in a different PR?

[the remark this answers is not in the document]
terrelln: Yeah, it should be done before the release, but you or I can do it in a separate PR.

shakeelrao:
Thanks for all the help, @terrelln :) The final solution is much cleaner than my original commit. 

As a side note: I was under the impression that my commits would be squashed before merging into dev, so I was pretty liberal in my use of commits. Is this the standard for this project?

PR #1547, opened 2019-03-13
Fix incorrect error code in ZSTD_errorFrameSizeInfo

shakeelrao:
While reading through the code to implement legacy compatibility, I noticed that `decompressBound` was assigned to the wrong error code in `ZSTD_errorFrameSizeInfo`.

In `ZSTD_decompressBound`, we check if `decompressedBound` equals `ZSTD_CONTENTSIZE_ERROR`, not `ZSTD_CONTENTSIZE_UNKNOWN` [1].

[1] https://github.com/facebook/zstd/blob/dev/lib/decompress/zstd_decompress.c#L545

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -437,7 +437,7 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
  * Contains the compressed frame size and an upper-bound for the decompressed frame size.
  * Note: before using `compressedSize` you must check for errors using ZSTD_isError().
  *       similarly, before using `decompressedBound`, you must check for errors using:
- *          `decompressedBound` != ZSTD_CONTENTSIZE_UNKNOWN
+ *          `decompressedBound` != ZSTD_CONTENTSIZE_ERROR
  */
 typedef struct {
     size_t compressedSize;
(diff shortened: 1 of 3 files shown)


terrelln:
Nice catch @shakeelrao, and thanks for adding the test!

shakeelrao:
Yeah, that's right. In all error cases, compressedSize will always be an error, unless we were dealing with a legacy record.

terrelln:
@shakeelrao yeah we should test both conditions, and I will merge this patch once we have a test case that can expose the bug. When we find a bug, we want to test that we fixed the problem.

You can add a  test for your function to `tests/legacy.c`, which contains multiple legacy compressed frames concatenated together. We expect the result to be `ZSTD_CONTENTSIZE_ERROR` until we add legacy support. Then verify that the test fails before your patch, and passes after.

When you add legacy support, you will then just update that test. You should also add a test for legacy frames and `ZSTD_findCompressedSize()`, since you'll be modifying its implementation for legacy frames.

shakeelrao:
@terrelln Sounds good! Thanks for the feedback. 

>You should also add a test for legacy frames and ZSTD_findCompressedSize()

Just to clarify, which function are you referring to (`ZSTD_findFrameCompressedSize` or `ZSTD_findDecompressedSize`)?

terrelln:
`ZSTD_findFrameCompressedSize()`

Issue #1544, opened 2019-03-02
Compression dicts and advance API


luben:
Hi,

I am trying to port existing code (zstd-jni) to the new Advanced API but it looks like there is something that I don't understand. I have already ported the decompression side allowing to attach dictionaries (fast of plain) using the new API but there is something missing compression side. Here is minimal change that I was expecting to work:

```patch
     ZSTD_CCtx_setParameter(stream, ZSTD_c_checksumFlag, checksum);
-    result = ZSTD_initCStream_usingDict(stream, dict_buff, dict_size, level);
+    result = ZSTD_CCtx_loadDictionary(stream, dict_buff, dict_size);
+    if (result != 0) goto E2;
+    result = ZSTD_initCStream(stream, level);
```
 With this change we can round-trip compression/decompression but the dictionary `Id` is lost from the compressed frame header - it is always 0.

Is it an omission on the library side or I am doing something wrong?

Cyan4973:
Hi @luben ,
a few things I can think of : 

- `ZSTD_initCStream()` is not supposed to be necessary. In the new advanced API, one just set the compression level with `ZSTD_CCtx_setParameter()`, and that's fine, sending data afterwards triggers an automatic `init` if they are the first bytes of the frames. 

- `ZSTD_initCStream()` is still useful for support of legacy code, obviously.
It can also be used to start a new frame without finishing properly the previous one, though this functionality is in competition with `ZSTD_CCtx_reset()`, which is now preferable.

- In theory, using `ZSTD_initCStream()` is expected to be neither necessary nor detrimental. It should merely reset the context for a new frame and set the compression level. But I believe it's a case of bad side effects when mixing legacy and new API entry points. `ZSTD_initCStream()` [enforces its own frame parameters](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3848), and one of them is removing the dictionary ID, which wasn't necessary before. I think we should update this entry point, so that it play nicer with the new API.

Short term : 
remove `ZSTD_initCStream()` in your code, just set the compression level with `ZSTD_CCtx_setParameter()`.

Medium term (for us) : 
update `ZSTD_initCStream()` internal so that it plays nicer with the new advanced API.

luben:
Thanks, that solved the problem. The lazy initialization will also allow to simplify the bindings. Does it apply also for the decompression context?

Cyan4973:
Yes, it does.

bkramer-moz:
[...]

PR #1551, opened 2019-03-14
Refactor old stream API to completely wrap the advanced API

terrelln:
Refactor the `ZSTD_initCStream*()` functions and `ZSTD_resetCStream()` to call:

```c
ZSTD_CCtx_reset(cctx, ZSTD_reset_session_only);
// Set the pledged src size if known
// Set parameters
// Set or clear the dictionary
```

This is a small difference in behavior when you mix the two APIs, since the parameters are now sticky.

The biggest advantages of this change are that it reduces the maintenance of the old streaming API, since it is now just a small wrapper, and we can point users to the implementation to help them switch to the new API. We can document these functions in terms of what they do in the new API.

The disadvantages are a small change in behavior, and users will end up mixing the two APIs.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3769,13 +3769,17 @@ static size_t ZSTD_resetCStream_internal(ZSTD_CStream* cctx,
 
 /* ZSTD_resetCStream():
  * pledgedSrcSize == 0 means "unknown" */
-size_t ZSTD_resetCStream(ZSTD_CStream* zcs, unsigned long long pledgedSrcSize)
+size_t ZSTD_resetCStream(ZSTD_CStream* zcs, unsigned long long pss)
 {
-    ZSTD_CCtx_params params = zcs->requestedParams;
+    /* temporary : 0 interpreted as "unknown" during transition period.
+     * Users willing to specify "unknown" **must** use ZSTD_CONTENTSIZE_UNKNOWN.
+     * 0 will be interpreted as "empty" in the future.
+     */
+    U64 const pledgedSrcSize = (pss==0) ? ZSTD_CONTENTSIZE_UNKNOWN : pss;
     DEBUGLOG(4, "ZSTD_resetCStream: pledgedSrcSize = %u", (unsigned)pledgedSrcSize);
-    if (pledgedSrcSize==0) pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN;
-    params.fParams.contentSizeFlag = 1;
-    return ZSTD_resetCStream_internal(zcs, NULL, 0, ZSTD_dct_auto, zcs->cdict, params, pledgedSrcSize);
+    FORWARD_IF_ERROR( ZSTD_CCtx_reset(zcs, ZSTD_reset_session_only) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_setPledgedSrcSize(zcs, pledgedSrcSize) );
+    return 0;
 }
 
 /*! ZSTD_initCStream_internal() :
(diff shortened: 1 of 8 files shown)


Cyan4973:
My understanding is that this patch does not modify the behavior of existing API as long as it's not mixed with the new API. That's great.

In case of API mixing (which ideally should not happen), it makes the existing API compatible with the new behavior of the advanced API, which seems like the right choice.

The expectation is that it will reduce confusion, like [this recent example](https://github.com/facebook/zstd/issues/1544).

lib/compress/zstd_compress.c
```diff
@@ -3769,13 +3769,18 @@ static size_t ZSTD_resetCStream_internal(ZSTD_CStream* cctx,
...
+     * 0 will be interpreted as "empty" in the future.
+     */
+    U64 const pledgedSrcSize = (pss==0) ? ZSTD_CONTENTSIZE_UNKNOWN : pss;
     DEBUGLOG(4, "ZSTD_resetCStream: pledgedSrcSize = %u", (unsigned)pledgedSrcSize);
-    if (pledgedSrcSize==0) pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN;
-    params.fParams.contentSizeFlag = 1;
-    return ZSTD_resetCStream_internal(zcs, NULL, 0, ZSTD_dct_auto, zcs->cdict, params, pledgedSrcSize);
+    FORWARD_IF_ERROR( ZSTD_CCtx_reset(zcs, ZSTD_reset_session_only) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_setPledgedSrcSize(zcs, pledgedSrcSize) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_setParameter(zcs, ZSTD_c_contentSizeFlag, 1) );
```
terrelln: Here and later we set `ZSTD_c_contentSizeFlag = 1` to preserve current behavior. But, since it defaults to 1, and nothing implicitly sets it to 0, should we just remove this line?

lib/compress/zstd_compress.c
```diff
@@ -3848,33 +3837,54 @@ size_t ZSTD_initCStream_usingCDict(ZSTD_CStream* zcs, const ZSTD_CDict* cdict)
...
+    /* temporary : 0 interpreted as "unknown" during transition period.
+     * Users willing to specify "unknown" **must** use ZSTD_CONTENTSIZE_UNKNOWN.
+     * 0 will be interpreted as "empty" in the future.
+     */
+    U64 const pledgedSrcSize = (pss==0) ? ZSTD_CONTENTSIZE_UNKNOWN : pss;
+    DEBUGLOG(4, "ZSTD_initCStream_srcSize");
+    FORWARD_IF_ERROR( ZSTD_CCtx_reset(zcs, ZSTD_reset_session_only) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_refCDict(zcs, NULL) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_setParameter(zcs, ZSTD_c_compressionLevel, compressionLevel) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_setParameter(zcs, ZSTD_c_contentSizeFlag, 1) );
```
terrelln: Here as well

lib/compress/zstd_compress.c
```diff
@@ -3848,33 +3837,54 @@ size_t ZSTD_initCStream_usingCDict(ZSTD_CStream* zcs, const ZSTD_CDict* cdict)
...
+    FORWARD_IF_ERROR( ZSTD_CCtx_setPledgedSrcSize(zcs, pledgedSrcSize) );
+    return 0;
 }
 
 size_t ZSTD_initCStream(ZSTD_CStream* zcs, int compressionLevel)
 {
     DEBUGLOG(4, "ZSTD_initCStream");
-    return ZSTD_initCStream_srcSize(zcs, compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN);
+    FORWARD_IF_ERROR( ZSTD_CCtx_reset(zcs, ZSTD_reset_session_only) );
+    FORWARD_IF_ERROR( ZSTD_CCtx_refCDict(zcs, NULL) );
```
terrelln: I'm clearing the dictionary here and above. I added this to preserve current behavior. I think it is the right call, because people are probably using `ZSTD_initCStream_usingCDict()` in conjunction with `ZSTD_initCStream()`, using the latter to compress without a dictionary.

lib/compress/zstd_compress.c
Cyan4973: Yes, I believe relying on default is fine.
It won't change behavior for users sticking to the existing API.

lib/compress/zstd_compress.c
Cyan4973: Same, default looks enough.

lib/compress/zstd_compress.c
Cyan4973: I agree.

terrelln:
I've updated the documentation, but I still need to add tests.

I will make a separate PR that moves the advanced API into stable, and moves the old stable streaming API to the bottom of stable, and marks the old experimental streaming API as deprecated.

terrelln:
I added regression tests that make sure that all of the `ZSTD_initCStream*()` variants compress the same as before.

lib/zstd.h
```diff
@@ -347,9 +347,24 @@ ZSTDLIB_API ZSTD_CStream* ZSTD_createCStream(void);
...
+ *
+ *     ZSTD_CCtx_reset(zcs, ZSTD_reset_session_only);
+ *     ZSTD_CCtx_refCDict(zcs, NULL); // clear the dictionary (if any)
+ *     ZSTD_CCtx_setParameter(zcs, ZSTD_c_compressionLevel, compressionLevel);
+ */
 ZSTDLIB_API size_t ZSTD_initCStream(ZSTD_CStream* zcs, int compressionLevel);
+/**
+ * Alternative for ZSTD_compressStream2(zcs, output, input, ZSTD_e_continue).
+ * NOTE: The return value is different. ZSTD_compressStream() returns a hint for
+ * the next read size (if non-zero and not an error). ZTD_compressStream2()
```
Cyan4973: nit: `ZSTD_compressStream2()`

Cyan4973:
Looks good to me !

PR #1553, opened 2019-03-15
Add legacy support to decompressBound

shakeelrao:
Allows `decompressBound` to handle legacy formatted records.

Implements https://github.com/facebook/zstd/issues/1536.

### DIFF
--- a/lib/legacy/zstd_legacy.h
+++ b/lib/legacy/zstd_legacy.h
@@ -186,9 +186,9 @@ MEM_STATIC void ZSTD_errorFrameSizeInfoLegacy(size_t* cSize, unsigned long long*
 
 MEM_STATIC ZSTD_frameSizeInfo ZSTD_findFrameSizeInfoLegacy(const void *src, size_t srcSize)
 {
+    U32 const version = ZSTD_isLegacy(src, srcSize);
     ZSTD_frameSizeInfo frameSizeInfo;
     memset(&frameSizeInfo, 0, sizeof(ZSTD_frameSizeInfo));
-    U32 const version = ZSTD_isLegacy(src, srcSize);
     switch(version)
     {
 #if (ZSTD_LEGACY_SUPPORT <= 1)
--- a/lib/legacy/zstd_legacy.h
+++ b/lib/legacy/zstd_legacy.h
@@ -188,43 +188,50 @@ MEM_STATIC ZSTD_frameSizeInfo ZSTD_findFrameSizeInfoLegacy(const void *src, size
 #if (ZSTD_LEGACY_SUPPORT <= 1)
         case 1 :
             ZSTDv01_findFrameSizeInfoLegacy(src, srcSize,
-                &frameSizeInfo.compressedSize, &frameSizeInfo.decompressedBound);
+                &frameSizeInfo.compressedSize,
+                &frameSizeInfo.decompressedBound);
             break;
 #endif
 #if (ZSTD_LEGACY_SUPPORT <= 2)
         case 2 :
             ZSTDv02_findFrameSizeInfoLegacy(src, srcSize,
-                &frameSizeInfo.compressedSize, &frameSizeInfo.decompressedBound);
+                &frameSizeInfo.compressedSize,
+                &frameSizeInfo.decompressedBound);
             break;
 #endif
 #if (ZSTD_LEGACY_SUPPORT <= 3)
         case 3 :
             ZSTDv03_findFrameSizeInfoLegacy(src, srcSize,
-                &frameSizeInfo.compressedSize, &frameSizeInfo.decompressedBound);
+                &frameSizeInfo.compressedSize,
+                &frameSizeInfo.decompressedBound);
             break;
 #endif
 #if (ZSTD_LEGACY_SUPPORT <= 4)
         case 4 :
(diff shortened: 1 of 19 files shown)


shakeelrao:
This is my initial draft for this issue. It's not cleanest solution, but I think it gets the job done. I had some difficulties with being able to include various headers, so certain macros/functions are often redefined.

As a side note: documentation for the `ZSTD_findFrameSizeInfoLegacy` methods aren't updated yet. I will update them once the initial design is approved.

lib/legacy/zstd_legacy.h
```diff
@@ -178,43 +178,73 @@ MEM_STATIC size_t ZSTD_decompressLegacy(
...
 #if (ZSTD_LEGACY_SUPPORT <= 7)
         case 7 :
-            return ZSTDv07_findFrameCompressedSize(src, compressedSize);
+            ZSTDv07_findFrameSizeInfoLegacy(src, srcSize,
+                &frameSizeInfo.compressedSize, &frameSizeInfo.decompressedBound);
+            break;
 #endif
         default :
-            return ERROR(prefix_unknown);
+            ZSTD_errorFrameSizeInfoLegacy(&frameSizeInfo.compressedSize,
```
terrelln: This function is only used here, so lets just inline it:

```c
frameSizeInfo.compressedSize = ERROR(prefix_unknown);
frameSizeInfo.decompressedBound = ZSTD_CONTENTSIZE_ERROR;
break;
```

tests/legacy.c
```diff
@@ -133,8 +133,8 @@ static int testStreamingAPI(void)
 
 static int testFrameDecoding(void)
 {
-    if (ZSTD_decompressBound(COMPRESSED, COMPRESSED_SIZE) != ZSTD_CONTENTSIZE_ERROR) {
-        DISPLAY("ERROR: ZSTD_decompressBound: Expected to receive ZSTD_CONTENTSIZE_ERROR\n");
+    if (strlen(EXPECTED) > ZSTD_decompressBound(COMPRESSED, COMPRESSED_SIZE)) {
```
terrelln: Can you also add a test for `ZSTD_findFrameCompressedSize()`, since we are modifying that code? This data contains multiple frames, since we should make sure if we sum them all up we get the right size.

lib/zstd.h
```diff
@@ -1085,6 +1085,17 @@ typedef enum {
...
+/**
+ * Contains the compressed frame size and an upper-bound for the decompressed frame size.
+ * Note: before using `compressedSize` you must check for errors using ZSTD_isError().
+ *       similarly, before using `decompressedBound`, you must check for errors using:
+ *          `decompressedBound` != ZSTD_CONTENTSIZE_ERROR
+ */
+typedef struct {
+    size_t compressedSize;
+    unsigned long long decompressedBound;
+} ZSTD_frameSizeInfo;
```
terrelln: This doesn't belong in `zstd.h`, since it is a public header, and this is not a public function.  You can put it in `zstd_internal.h` and include that in `zstd_legacy.h`.

shakeelrao: I was wondering how to get around this. Thanks!

shakeelrao:
Thanks, @terrelln! Ideally, each of the legacy implementations would return a `ZSTD_frameSizeInfo` struct and reuse `ZSTD_errorFrameSizeInfo` to set error values, rather than defining `ZSTD_errorFrameSizeInfoLegacy` and `ZSTD_CONTENTSIZE_ERROR` in each legacy file, but I wasn't sure how to achieve this.

tests/legacy.c
shakeelrao: Definitely. I added a test for `ZSTD_findFrameCompressedSize` in the latest commit. Let me know if that's what you were thinking.

lib/legacy/zstd_v02.c
```diff
@@ -3312,37 +3314,57 @@ static size_t ZSTD_decompress(void* dst, size_t maxDstSize, const void* src, siz
     return ZSTD_decompressDCtx(&ctx, dst, maxDstSize, src, srcSize);
 }
 
-static size_t ZSTD_findFrameCompressedSize(const void *src, size_t srcSize)
+MEM_STATIC void ZSTD_errorFrameSizeInfoLegacy(size_t* cSize, unsigned long long* dBound, size_t ret)
```
terrelln: nit: This function only needs to be `static`, not `MEM_STATIC`. That macro just tells the compiler it is static and it may be unused, but this function isn't unused. Same for the rest of the implementations of `ZSTD_errorFrameSizeInfoLegacy()`.

shakeelrao: Should I also add error checking? For instance:

```
if (cSize == NULL || dBound == NULL) return;
```

terrelln: I would just `assert(cSize != NULL && dBound != NULL)` at the top of each `ZSTDv0X_findFrameInfoLegacy()`. Since these functions are internal only, and not exposed (no-one includes `zstd_v0X.h` directly except `zstd_legacy.h`) we can assume that they get valid parameters, but adding an assert helps verify that during testing.

terrelln: Actually, these files don't have any asserts. I would just write in the docs that neither may be `NULL`. Since it is a `NULL` check debug builds would crash anyways.

Issue #1548, opened 2019-03-13
Update examples to use the new advanced API


terrelln:
We have example usages of zstd under [examples/](https://github.com/facebook/zstd/tree/dev/examples). We need to update all of the streaming APIs to use the new [advanced API](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L472). Each file should be updated in a separate PR.

- [x] [Streaming compression](https://github.com/facebook/zstd/blob/dev/examples/streaming_compression.c)
- [x] [Multiple streaming compression](https://github.com/facebook/zstd/blob/dev/examples/multiple_streaming_compression.c)
- [x] [Streaming decompression](https://github.com/facebook/zstd/blob/dev/examples/streaming_decompression.c)
- [x] [Streaming memory usage](https://github.com/facebook/zstd/blob/dev/examples/streaming_memory_usage.c)

We should also update the dictionary examples to use the new API.

- [x] [Dictionary compression](https://github.com/facebook/zstd/blob/dev/examples/dictionary_compression.c)
- [x] [Dictionary decompression](https://github.com/facebook/zstd/blob/dev/examples/dictionary_decompression.c)

Blocks Issue #1533.

PR #1557, opened 2019-03-22
[examples] Update streaming_compression to the new API

terrelln:
Update streaming_compression to the new API and add instructive comments.

Issue #1548.

### DIFF
--- a/examples/utils.h
+++ b/examples/utils.h
@@ -20,6 +20,7 @@
 #include <errno.h>     // errno
 #include <assert.h>    // assert
 #include <sys/stat.h>  // stat
+#include <zstd.h>
 
 /*
  * Define the returned error code from utility functions.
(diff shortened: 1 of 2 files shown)


examples/utils.h
```diff
@@ -204,4 +205,30 @@ static void saveFile_orDie(const char* fileName, const void* buff, size_t buffSi
     }
 }
 
+/*! CHECK
+ * Check that the condition holds. If it doesn't print a message and die.
+ */
+#define CHECK(cond, ...)                        \
```
Cyan4973: `CHECK` only writes a report when a condition is not met.
It doesn't `exist()` nor `abort()` execution, which therefore proceed to next instruction.
Is that the wanted behavior ?

terrelln: Yeah thats not the intention, nice catch!

PR #1558, opened 2019-03-23
[libzstd] Specify soversion and version corectly for CMake build

### LINKED ISSUE #1512: soname inconsistency b/w cmake and non-cmake build
```
$ find ./ | grep libzstd.so

# CMake build produces:
./build/lib/libzstd.so
./build/lib/libzstd.so.1.3.8

# Non-CMake build produces (running make in the top-level directory):
./lib/libzstd.so
./lib/libzstd.so.1.3.8
./lib/libzstd.so.1
```
The symlink `./lib/libzstd.so.1` -> `libzstd.so.1.3.x` is missing from the CMake builds. Is that a bug or has that been omitted deliberately?

nehaljwani:
Fixes #1512

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -149,7 +149,8 @@ if (ZSTD_BUILD_SHARED)
             libzstd_shared
             PROPERTIES
             OUTPUT_NAME zstd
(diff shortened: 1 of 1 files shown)


terrelln:
Thanks for the patch! I've tested it by running `make cmake build && ls -l build/cmake/build` and looking at the symlinks.

PR #1561, opened 2019-03-29
Update comments in zstd.h and fileio.c

shakeelrao:
Fix typo in `zstd.h` and add `UTIL_isSameFile` to comment in `fileio.c`.

Simple text change for completeness.


terrelln:
Thanks for the patch! It seems small, but fixing small mistakes like this really adds to the polish of the library.

PR #1563, opened 2019-03-29
[libzstd] Split out zstd_fast dict match state function

terrelln:
I'm working on optimizing the fast search algorithm, but it is hard to work with the two functions combined. Each function wants a different optimization.

This patch has no behavior changes, it just splits the two functions out exactly as-is. I will have a second PR that optimizes these functions.

I benchmarked with clang and gcc on my Intel i9 and clang on my Mac on silesia and the GitHub dataset with a dictionary. This change has a neutral to small positive (up to 5 MB/s) performance difference.


Cyan4973:
Looks good,
it is more readable.
it might even help the compiler a bit, as I measured a ~1% speed improvement in simple benchmarks.

PR #1568, opened 2019-04-02
Update streaming_memory_usage.c and fix ZSTD_estimateCStreamSize_usingCCtxParams()

terrelln:
* Fix `ZSTD_estimateCStreamSize_usingCCtxParams()` and add a test.
* Make setting `ZSTD_d_maxWindowLog` to 0 reset it to the default value.
* Check that the bounds hold in the example.
* Set the maximum allowed window size in the decompressor to allow larger window logs.
* Update to the new API making progress on #1548.

### DIFF
--- a/examples/streaming_memory_usage.c
+++ b/examples/streaming_memory_usage.c
@@ -19,6 +19,7 @@
 #include <stdio.h>   /* printf */
 #define ZSTD_STATIC_LINKING_ONLY
 #include "zstd.h"
+#include "utils.h"
 
 
 /*===   functions   ===*/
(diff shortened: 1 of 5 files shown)


examples/streaming_memory_usage.c
```diff
@@ -19,6 +19,7 @@
 #include <stdio.h>   /* printf */
 #define ZSTD_STATIC_LINKING_ONLY
 #include "zstd.h"
+#include "utils.h"
```
Cyan4973: Is this dependency really necessary ?

Generally, the examples should better be as self-contained as possible, depending only on standard libs and `libzstd`.

lib/compress/zstd_compress.c
```diff
@@ -1106,9 +1106,11 @@ size_t ZSTD_estimateCCtxSize(int compressionLevel)
 size_t ZSTD_estimateCStreamSize_usingCCtxParams(const ZSTD_CCtx_params* params)
 {
     RETURN_ERROR_IF(params->nbWorkers > 0, GENERIC, "Estimate CCtx size is supported for single-threaded compression only.");
-    {   size_t const CCtxSize = ZSTD_estimateCCtxSize_usingCCtxParams(params);
-        size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << params->cParams.windowLog);
-        size_t const inBuffSize = ((size_t)1 << params->cParams.windowLog) + blockSize;
+    {   ZSTD_compressionParameters const cParams =
```
Cyan4973: Is that just to avoid the `params->` indirection ? 
or is there another benefit I missed ?

tests/zstreamtest.c
```diff
@@ -344,6 +344,20 @@ static int basicUnitTests(U32 seed, double compressibility)
         DISPLAYLEVEL(3, "OK (%u bytes) \n", (unsigned)(cstreamSize + cdictSize));
     }
 
+    /* context size functions */
```
Cyan4973: 👍

examples/streaming_memory_usage.c
```diff
@@ -61,90 +62,74 @@ int main(int argc, char const *argv[]) {
...
+        /* Set the compression level. */
+        CHECK_ZSTD( ZSTD_CCtxParams_setParameter(cctxParams, ZSTD_c_compressionLevel, compressionLevel) );
+        /* Set the window log.
+         * The value 0 means use the default window log, which is equivalent to
+         * not setting it.
+         */
+        CHECK_ZSTD( ZSTD_CCtxParams_setParameter(cctxParams, ZSTD_c_windowLog, wLog) );
+
+        /* Force the compressor to allocate the maximum memory size for a given
+         * level by not provding the pledged source size, or calling
```
Cyan4973: :typo `providing`

examples/streaming_memory_usage.c
```diff
@@ -61,90 +62,74 @@ int main(int argc, char const *argv[]) {
...
-                printf("ZSTD_initCStream error : %s \n", ZSTD_getErrorName(error));
-                return 1;
-            }
-        }
-
+        /* the ZSTD_CCtx_params structure is a way to save parameters and use
+         * them across multiple contexts. We use them here so we can call the
+         * function ZSTD_estimateCStreamSize_usingCCtxParams().
+         */
+        ZSTD_CCtx_params* const cctxParams = ZSTD_createCCtxParams();
```
Cyan4973: `NULL` check seems missing

examples/streaming_memory_usage.c
```diff
@@ -61,90 +62,74 @@ int main(int argc, char const *argv[]) {
...
-            return 1;
-        }
-        {   size_t const error = ZSTD_initDStream(dstream);
-            if (ZSTD_isError(error)) {
-                printf("ZSTD_initDStream error : %s \n", ZSTD_getErrorName(error));
-                return 1;
-            }
-        }
-        /* forces decompressor to use maximum memory size, as decompressed size is not known */
+        ZSTD_DCtx* dctx = ZSTD_createDCtx();
```
Cyan4973: minor: `const`ify all the world : `ZSTD_DCtx* const`

lib/compress/zstd_compress.c
terrelln: This is the bug fix. This function will set `cParams.windowLog`, where previously just the compression level was set.

examples/streaming_memory_usage.c
terrelln: I'll put up a separate PR to rename `utils.h` after these have landed.

terrelln:
I made setting `ZSTD_d_maxWindowLog` to `0` mean reset the value to the default value, so the behavior is consistent with `ZSTD_c_windowLog`. I slightly prefer this behavior, but don't have a strong preference, so I could back out that commit if you want.

Cyan4973:
> I made setting ZSTD_d_maxWindowLog to 0 mean reset the value to the default value

I think it's a good choice for consistency.

Cyan4973:
oops, sorry, looking for "review approval" but "merge"d instead...

PR #1569, opened 2019-04-02
Stabilize the advanced API

terrelln:
* Bump the library version to 1.4.0
* Move the candidate advanced API to the stable section.
It makes some minor whitespace changes, but it doesn't change any
of the wording of the documentation.

The biggest thing to review in this PR is the order of the functions.
Also, if there are any documentation changes you want to see, let me know.
I'm planning on updating the streaming documentation block to mention
`ZSTD_compressStream2()`.

I'll put up a separate PR that tweaks some of the documentation
once this lands, so that it is easier to review.

NOTE: Even though these functions are now in stable, they aren't
stable until the next release (in under 1 month). It is possible
that they change until then.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -90,6 +90,21 @@ ZSTDLIB_API const char* ZSTD_versionString(void);   /* requires v1.3.0+ */
 #endif
 
 /***************************************
+*  Constants
+***************************************/
+
+/* All magic numbers are supposed read/written to/from files/memory using little-endian convention */
+#define ZSTD_MAGICNUMBER            0xFD2FB528    /* valid since v0.8.0 */
+#define ZSTD_MAGIC_DICTIONARY       0xEC30A437    /* valid since v0.7.0 */
+#define ZSTD_MAGIC_SKIPPABLE_START  0x184D2A50    /* all 16 values, from 0x184D2A50 to 0x184D2A5F, signal the beginning of a skippable frame */
+#define ZSTD_MAGIC_SKIPPABLE_MASK   0xFFFFFFF0
+
+#define ZSTD_BLOCKSIZELOG_MAX  17
+#define ZSTD_BLOCKSIZE_MAX     (1<<ZSTD_BLOCKSIZELOG_MAX)
+
+
+
+/***************************************
 *  Simple API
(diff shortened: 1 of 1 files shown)


lib/zstd.h
```diff
@@ -941,6 +710,207 @@ ZSTDLIB_API size_t ZSTD_DCtx_refPrefix(ZSTD_DCtx* dctx,
 ZSTDLIB_API size_t ZSTD_DCtx_reset(ZSTD_DCtx* dctx, ZSTD_ResetDirective reset);
 
 
+/****************************
+*  Streaming
```
Cyan4973: Comment, for discussion : 

"Streaming" seems pretty far in the priority order of API presentation.

There is a logic here : all one-shot compression functions are presented first, and one-shot looks like an "easier" use case than streaming. Also, presenting the advanced parameters makes them re-usable in the "streaming" section.

I'm just puzzled by the feeling that "streaming" scenarios are probably more common that "dictionary" ones. Yet, a reader will have to go pretty far down the list to discover streaming capabilities.

Should we swap this order, and have streaming before dictionary ?

It's probably not a big deal, just curious what's your take on this.

lib/zstd.h
```diff
@@ -941,6 +710,207 @@ ZSTDLIB_API size_t ZSTD_DCtx_refPrefix(ZSTD_DCtx* dctx,
...
+ *            only ZSTD_e_end or ZSTD_e_flush operations are allowed.
+ *            Before starting a new compression job, or changing compression parameters,
+ *            it is required to fully flush internal buffers.
+ */
+ZSTDLIB_API size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
+                                         ZSTD_outBuffer* output,
+                                         ZSTD_inBuffer* input,
+                                         ZSTD_EndDirective endOp);
+
+/**
```
Cyan4973: Maybe add a note, or separation,
to indicate that it's a legacy, yet fully supported, streaming API.
This would explain/ remind why the redundancy.

terrelln:
The last commit in the stack rewords the streaming API doc block. The rest of the commits are just reordering, or adding a some headers.

PR #1576, opened 2019-04-10
Add new fuzzers and fix exposed bugs

terrelln:
Changes to the zstd library:

* Make `ZSTD_decompressDCtx()` use the loaded dictionary and add a test.
* Fix `ZSTD_compress2()` for multithreaded compression and add a test.
* Make `ZSTD_compressStream2()` with `ZSTD_e_flush` and `ZSTD_end` make maximal forward progress, and update the docs.
* Don't check the dictionary ID when `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined.
* Make the corpus size warning in the dictionary builder respect the current displayLevel.

Changes to the fuzzers:
* Add `dictionary_round_trip` and `dictionary_decompress` fuzzers that generate a dictionary to use using fast-cover based on the input. So this tests using a valid dictionary. The dictionary is used in both prefix mode and zstd mode, and using both the old and new APIs.
* Fuzz with less than `ZSTD_compressBound()` output size half the time.
* Use the new advanced API in streaming fuzzers, and fuzz some new options like `ZSTD_c_nbWorkers` and `ZSTD_c_rsyncable`.
* Small improvements to the fuzzers.

Sorry this change got big. The commits are broken up into small pieces, and the changes to zstd are broken out into their own commits.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -788,7 +788,7 @@ size_t ZSTD_decompress_usingDict(ZSTD_DCtx* dctx,
 
 size_t ZSTD_decompressDCtx(ZSTD_DCtx* dctx, void* dst, size_t dstCapacity, const void* src, size_t srcSize)
 {
-    return ZSTD_decompress_usingDict(dctx, dst, dstCapacity, src, srcSize, NULL, 0);
+    return ZSTD_decompress_usingDDict(dctx, dst, dstCapacity, src, srcSize, dctx->ddict);
 }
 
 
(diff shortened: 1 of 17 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -783,7 +788,7 @@ size_t ZSTD_decompress_usingDict(ZSTD_DCtx* dctx,
 
 size_t ZSTD_decompressDCtx(ZSTD_DCtx* dctx, void* dst, size_t dstCapacity, const void* src, size_t srcSize)
 {
-    return ZSTD_decompress_usingDict(dctx, dst, dstCapacity, src, srcSize, NULL, 0);
+    return ZSTD_decompress_usingDDict(dctx, dst, dstCapacity, src, srcSize, dctx->ddict);
```
Cyan4973: 👍

Cyan4973:
Very good,
these changes are welcomed as they improve API consistency.

PR #1580, opened 2019-04-10
[libzstd] Fix decompression dictionary bugs and clean up initialization

terrelln:
Bugs:

* `ZSTD_DCtx_refPrefix()` didn't clear the dictionary after the first
  use. Fix and add a test case.
* `ZSTD_DCtx_reset()` always cleared the dictionary. Fix and add a test
  case.
* After calling `ZSTD_resetDStream()` you could no longer load a
  dictionary, since the stage was set to `zdss_loadHeader`. Fix and add
  a test case.

Cleanup:

* Make `ZSTD_initDStream*()` and `ZSTD_resetDStream()` wrap the new
 advanced API, and add test cases.
* Document the equivalent of these functions in the advanced API and
  document the unstable functions as deprecated.

### DIFF
--- a/lib/decompress/zstd_decompress_internal.h
+++ b/lib/decompress/zstd_decompress_internal.h
@@ -89,6 +89,12 @@ typedef enum { ZSTDds_getFrameHeaderSize, ZSTDds_decodeFrameHeader,
 typedef enum { zdss_init=0, zdss_loadHeader,
                zdss_read, zdss_load, zdss_flush } ZSTD_dStreamStage;
 
+typedef enum {
+    ZSTD_use_indefinitely = -1,  /* Use the dictionary indefinitely */
+    ZSTD_dont_use = 0,           /* Do not use the dictionary (if one exists free it) */
+    ZSTD_use_once = 1            /* Use the dictionary once and set to ZSTD_dont_use */
+} ZSTD_dictUses_e;
+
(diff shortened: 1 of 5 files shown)


lib/decompress/zstd_decompress_internal.h
```diff
@@ -123,6 +123,9 @@ struct ZSTD_DCtx_s
     const ZSTD_DDict* ddict;     /* set by ZSTD_initDStream_usingDDict(), or ZSTD_DCtx_refDDict() */
     U32 dictID;
     int ddictIsCold;             /* if == 1 : dictionary is "new" for working context, and presumed "cold" (not in cpu cache) */
+    int dictUsesRemaining;       /* if == 1 : dictionary should be used once.
+                                  * if == 0 : dictionary should be forgotten now.
+                                  * if < 0 : dictionary should be used indefinitely. */
```
Cyan4973: minor : while it works, it nonetheless feels strange to declare a member `dictUsesRemaining` to actually mean "use indefinitely when value is negative".

Would it be clearer to have an `enum`, with typically 3 values, for example `use_indefinitely`, `use_once`, `do_no_longer_use` ?

Cyan4973:
These are very good points, 
it's definitely better for API convergence.

I just have a minor comment regarding clarity of one internal mechanism.
Feel free to proceed or comment.

PR #1589, opened 2019-04-16
 [libzstd] Require ZSTD_MULTITHREAD to create a ZSTDMT_CCtx

terrelln:
ZSTDMT was broken when compiled without ZSTD_MULTITHREAD defined, because `ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, nbWorkerss)` failed. It was detected by the MSVC test which runs the fuzzer with multithreading disabled.

This is a very niche use case of a deprecated API, because the API is inefficient and synchronous, since `threading.h` will be synchronous. Users almost certainly don't want this, and anyone who tested their code should realize that it is broken. Therefore, I think it is safe to require `ZSTD_MULTITHREAD` to be defined to use ZSTDMT.

Also, fix Visual `fuzzer` builds to compile with `ZSTD_MULTITHREAD=1`. This branch `appveyorTests` runs the tests that `master` runs, so passing Appveyor means that the tests are fixed.


Cyan4973:
Nice catch !
There are still a few minor warnings to clean up for Windows, but nothing that change the diagnosis nor the solution.

PR #1593, opened 2019-04-18
[fuzzer] Run fuzzers in legacy mode and fix legacy code

terrelln:
* Run fuzzers with legacy support.
* Size the decompression output buffer randomly and run 10x with different output sizes to try and catch buffer overruns better.
* Fix a bug in `ZSTDv6_findFrameSizeInfoLegacy()`.
* Fix a bug in `ZSTD_decodeSeqHeaders()`.


terrelln:
The `stream_decompress` fuzzer running with legacy mode enabled just found the reported bug in `ZSTD_decodeSeqHeaders()`.

PR #1602, opened 2019-04-26
meson: Update default project version

lzutao:
* Update to use ninja v1.9.0 on CI

### DIFF
--- a/build/meson/meson.build
+++ b/build/meson/meson.build
@@ -16,9 +16,7 @@ project('zstd',
     'cpp_std=c++11',
     'buildtype=release'
   ],
-  # Fall-back version in case of extracting version number from
-  # `zstd.h` failed.
-  version: '1.4.0',
+  version: 'DUMMY',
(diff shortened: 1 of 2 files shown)


build/meson/meson.build
```diff
@@ -16,7 +16,7 @@ project('zstd',
     'cpp_std=c++11',
     'buildtype=release'
   ],
-  version: '1.3.8',
+  version: '1.4.0',
```
Cyan4973: Same comment (as the `lz4` PR) regarding the version number

lzutao: This is fallback version number in case of extracting version number from
`zstd.h` failed.

terrelln: Is it possible to always use the version in `zstd.h` and fail the build if the extraction fails? That would be preferable, because then we don't have to remember to update the meson build file.

lzutao: Done and done.

PR #1604, opened 2019-04-27
Make zstdcat to follow symbolic links

vejnar:
Since [#1520](https://github.com/facebook/zstd/issues/1520) has been merged, commands like zstdcat properly detect symbolic links on Linux. But the consequence is that they output nothing when executed on a symbolic link. According to the manual, zstdcat includes -f, which should follow symbolic links. This is also how most *cat commands like zcat behave.

Please comment and I would be glad if you could consider this request for merging into master. Thanks.
Charles


terrelln:
Oh, nearly forgot! Would you mind adding a test to [tests/playTests.sh](https://github.com/facebook/zstd/blob/dev/tests/playTests.sh#L303) that makes sure it works `zstdcat` works with symlinks? There is already a section that tests `zstdcat`, so you can just append a test there.

If you don't have the time, let me know and I can add the test myself. If I don't hear back in a few days I'll go ahead and merge and add the test.

PR #1616, opened 2019-05-22
Modifying many ERROR(generic)'s

Tyler-Tran:
To make error messages more descriptive I changed a large portion of the ERROR(generics)'s besides the ones that returned from the init function, which I will look into after.


lib/dictBuilder/cover.c
```diff
@@ -729,11 +729,11 @@ ZDICTLIB_API size_t ZDICT_trainFromBuffer_cover(
   /* Checks */
   if (!COVER_checkParameters(parameters, dictBufferCapacity)) {
     DISPLAYLEVEL(1, "Cover parameters incorrect\n");
-    return ERROR(GENERIC);
+    return ERROR(parameter_unsupported);
```
terrelln: I think `parameter_outOfBound` is a better error code, here and elsewhere.

Generally `parameter_unsupported` is meant to mean zstd doesn't understand the parameter, but here its the parameter value that is bad.

lib/dictBuilder/cover.c
```diff
@@ -1075,7 +1075,7 @@ ZDICTLIB_API size_t ZDICT_optimizeTrainFromBuffer_cover(
         COVER_best_destroy(&best);
         COVER_ctx_destroy(&ctx);
         POOL_free(pool);
-        return ERROR(GENERIC);
+        return ERROR(parameter_unsupported);
```
terrelln: This looks like a memory allocation failure, since `malloc` failed.

lib/dictBuilder/fastcover.c
```diff
@@ -687,7 +687,7 @@ ZDICT_optimizeTrainFromBuffer_fastCover(
           COVER_best_destroy(&best);
           FASTCOVER_ctx_destroy(&ctx);
           POOL_free(pool);
-          return ERROR(GENERIC);
+          return ERROR(parameter_unsupported);
```
terrelln: memory allocation failure

lib/dictBuilder/zdict.c
```diff
@@ -834,7 +834,7 @@ static size_t ZDICT_analyzeEntropy(void*  dstBuffer, size_t maxDstSize,
     }
 
     if (maxDstSize<12) {
-        eSize = ERROR(GENERIC);
+        eSize = ERROR(workSpace_tooSmall);
```
terrelln: I think `dstSize_tooSmall` is a better error code here.

lib/dictBuilder/cover.c
```diff
@@ -529,7 +529,7 @@ static void COVER_ctx_destroy(COVER_ctx_t *ctx) {
  * Returns 1 on success or zero on error.
  * The context must be destroyed with `COVER_ctx_destroy()`.
  */
-static int COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
+size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
```
terrelln: The CI error is a bit cryptic, but it is basically just telling you that you need to mark this function `static`.

`static` means that this function is only used in this translation unit, so don't export it. All functions not marked static are implicitly exported.

As a general rule, unless a function is present in a header file, it should be marked static.

lib/dictBuilder/cover.c
```diff
@@ -544,17 +544,17 @@ static int COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
       totalSamplesSize >= (size_t)COVER_MAX_SAMPLES_SIZE) {
     DISPLAYLEVEL(1, "Total samples size is too large (%u MB), maximum size is %u MB\n",
                  (unsigned)(totalSamplesSize>>20), (COVER_MAX_SAMPLES_SIZE >> 20));
-    return 0;
+    return 72;
```
terrelln: You want to return an error here, like this

```
return ERROR(srcSize_wrong);
```

`ERROR(srcSize_wrong)` gets translated to `-(size_t)ZSTD_error_srcSize_wrong == -72`.

You can check for an error using

```
    size_t const ret = COVER_ctx_init(...);
    if (ZSTD_isError(ret)) {
        return ret;
    }
```

This "multiplexes" a return value and an error code into a single value. If the value is less than `ERROR(maxCode)`, which is a very large value (`(size_t)-120`), then the function succeeded and the returned value is a size. Otherwise the function failed and we can find the message with `ZSTD_getErrorName(ret)`.

You can take a look at `zstd_errors.h`, and `error_private.[hc]` for details on how it works.

Since we don't need a return value we can just return `0` on success.

lib/dictBuilder/cover.c
```diff
@@ -729,27 +729,31 @@ ZDICTLIB_API size_t ZDICT_trainFromBuffer_cover(
...
                  ZDICT_DICTSIZE_MIN);
     return ERROR(dstSize_tooSmall);
   }
   /* Initialize context and activeDmers */
-  if (!COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
-                      parameters.d, parameters.splitPoint)) {
-    return ERROR(GENERIC);
+  size_t init_val = COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
+                      parameters.d, parameters.splitPoint);
+  if (init_val == 72) {
```
terrelln:
```
if (ZSTD_isError(initCode)) {
    return initCode;
}
```

lib/dictBuilder/cover.c
```diff
@@ -729,27 +729,31 @@ ZDICTLIB_API size_t ZDICT_trainFromBuffer_cover(
...
   if (dictBufferCapacity < ZDICT_DICTSIZE_MIN) {
     DISPLAYLEVEL(1, "dictBufferCapacity must be at least %u\n",
                  ZDICT_DICTSIZE_MIN);
     return ERROR(dstSize_tooSmall);
   }
   /* Initialize context and activeDmers */
-  if (!COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
-                      parameters.d, parameters.splitPoint)) {
-    return ERROR(GENERIC);
+  size_t init_val = COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
```
terrelln: nit: Mark this `const`. We want to mark any variable which doesn't get modified as `const`. This helps the reader identify when a variable will be modified easily.

Also, our convention is to name variables camelCase.

terrelln:
For an example of how it works, look at this function https://github.com/facebook/zstd/blob/dev/lib/decompress/zstd_ddict.c#L85. You don't need to bother what it does, but you can see it returns `ERROR(...)` on error, and `0` on success. You would call it like this:

```
size_t const ret = ZSTD_loadEntropy_intoDDict(...);
if (ZSTD_isError(ret)) {
    return ret;
}
```

This particular function just returns 0 on success. Another example is `ZSTD_decompress()`, which returns an error, or the size of the decompressed output.

```
size_t const dstSize = ZSTD_decompress(dst, dstCapacity, src, srcSize);
if (ZSTD_isError(dstSize)) {
    fprintf(stderr, "ZSTD returned an error: %s\n", ZSTD_getErrorName(dstSize);
} else {
    fprintf(stderr, "Zstd decompression succeeded and the output is %zu bytes!\n", dstSize);
}
```

lib/dictBuilder/cover.c
```diff
@@ -724,32 +724,37 @@ ZDICTLIB_API size_t ZDICT_trainFromBuffer_cover(
...
   if (dictBufferCapacity < ZDICT_DICTSIZE_MIN) {
     DISPLAYLEVEL(1, "dictBufferCapacity must be at least %u\n",
                  ZDICT_DICTSIZE_MIN);
     return ERROR(dstSize_tooSmall);
   }
   /* Initialize context and activeDmers */
-  if (!COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
-                      parameters.d, parameters.splitPoint)) {
-    return ERROR(GENERIC);
+  init_val = COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
```
terrelln: So this is an annoying thing about C. It doesn't allow mixed declarations and code. That means to create a new variable, you have to start a new scope, like:

```
{
    size_t const initCode = COVER_ctx_init();
    /* use initCode */
}
```

Still, it's better to create a new scope than move the variable declaration to the top of the code.

lib/dictBuilder/cover.c
```diff
@@ -729,27 +729,30 @@ ZDICTLIB_API size_t ZDICT_trainFromBuffer_cover(
...
     return ERROR(dstSize_tooSmall);
   }
   /* Initialize context and activeDmers */
-  if (!COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
-                      parameters.d, parameters.splitPoint)) {
-    return ERROR(GENERIC);
+  {
+    size_t const initVal = COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
+                      parameters.d, parameters.splitPoint);
+    if(ZSTD_isError(initVal)){
```
terrelln: nit: space after `if` and between `){`

lib/dictBuilder/cover.c
```diff
@@ -1022,15 +1025,15 @@ ZDICTLIB_API size_t ZDICT_optimizeTrainFromBuffer_cover(
   /* Checks */
   if (splitPoint <= 0 || splitPoint > 1) {
     LOCALDISPLAYLEVEL(displayLevel, 1, "Incorrect parameters\n");
-    return ERROR(GENERIC);
+    return ERROR(parameter_unsupported);
```
terrelln: `parameter_outOfBounds`

lib/dictBuilder/cover.c
```diff
@@ -1022,15 +1025,15 @@ ZDICTLIB_API size_t ZDICT_optimizeTrainFromBuffer_cover(
...
   }
   if (kMinK < kMaxD || kMaxK < kMinK) {
     LOCALDISPLAYLEVEL(displayLevel, 1, "Incorrect parameters\n");
-    return ERROR(GENERIC);
+    return ERROR(parameter_unsupported);
```
terrelln: `parameter_outOfBounds`. I think we should change all `parameter_unsupported` to `outOfBounds`.

lib/dictBuilder/cover.c
```diff
@@ -1054,11 +1057,11 @@ ZDICTLIB_API size_t ZDICT_optimizeTrainFromBuffer_cover(
     /* Initialize the context for this value of d */
     COVER_ctx_t ctx;
     LOCALDISPLAYLEVEL(displayLevel, 3, "d=%u\n", d);
-    if (!COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples, d, splitPoint)) {
+    if (COVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples, d, splitPoint) != 1) {
```
terrelln: This looks wrong, you'll have to check `ZSTD_isError()`

lib/dictBuilder/fastcover.c
```diff
@@ -550,11 +550,11 @@ ZDICT_trainFromBuffer_fastCover(void* dictBuffer, size_t dictBufferCapacity,
     if (!FASTCOVER_checkParameters(coverParams, dictBufferCapacity, parameters.f,
                                    parameters.accel)) {
       DISPLAYLEVEL(1, "FASTCOVER parameters incorrect\n");
-      return ERROR(GENERIC);
+      return ERROR(parameter_unsupported);
```
terrelln: `outOfBounds`

lib/dictBuilder/fastcover.c
```diff
@@ -564,11 +564,14 @@ ZDICT_trainFromBuffer_fastCover(void* dictBuffer, size_t dictBufferCapacity,
...
     /* Initialize context */
-    if (!FASTCOVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
+    {
+      size_t const initVal = FASTCOVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples,
                             coverParams.d, parameters.splitPoint, parameters.f,
-                            accelParams)) {
-      DISPLAYLEVEL(1, "Failed to initialize context\n");
-      return ERROR(GENERIC);
+                            accelParams);
+      if(ZSTD_isError(initVal)){
```
terrelln: nit: spaces like above

lib/dictBuilder/fastcover.c
```diff
@@ -666,11 +669,14 @@ ZDICT_optimizeTrainFromBuffer_fastCover(
...
       FASTCOVER_ctx_t ctx;
       LOCALDISPLAYLEVEL(displayLevel, 3, "d=%u\n", d);
-      if (!FASTCOVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples, d, splitPoint, f, accelParams)) {
-        LOCALDISPLAYLEVEL(displayLevel, 1, "Failed to initialize context\n");
-        COVER_best_destroy(&best);
-        POOL_free(pool);
-        return ERROR(GENERIC);
+      {
+        size_t const initVal = FASTCOVER_ctx_init(&ctx, samplesBuffer, samplesSizes, nbSamples, d, splitPoint, f, accelParams);
+        if(ZSTD_isError(initVal)){
```
terrelln: nit: spaces

tests/playTests.sh
```diff
@@ -508,6 +508,8 @@ println "- Create dictionary using all samples for both training and testing"
 $ZSTD --train-fastcover=split=100 -r "$TESTDIR"/*.c "$PRGDIR"/*.c
 println "- Create dictionary using f=16"
 $ZSTD --train-fastcover=f=16 -r "$TESTDIR"/*.c "$PRGDIR"/*.c
+println "- Create dictionary using accel=0 (should error)"
+$ZSTD --train-fastcover=accel=0 -r "$TESTDIR"/*.c "$PRGDIR"/*.c
```
terrelln: The comment says it should error, so do you need `&& die` here as well?

terrelln: It looks like it passes with `accel=0`, so either pick a different accel value for the test, or update the comment.

PR #1620, opened 2019-05-26
Skip --adapt and --rsyncable tests when built without thread support

michaelforney:
These options require ZSTD_MULTITHREAD.


felixhandte:
Thanks for your contribution! This looks good. A possible enhancement would be to also add a straightforward test if zstd was built without multi-threading, that checks that these options fail, but do not, e.g., segfault. Do you want to add that? Otherwise I can merge this as is.

michaelforney:
Sorry, I don't really have time to look at that. Please merge as-is.

Issue #1503, opened 2019-01-19
CMake does not install zstdgrep and zstdless shell scripts


Mno-hime:
I ported zstd for the OpenIndiana project userland. I opt to build it via the CMake infrastructure in `build/cmake/`, however `zstdgrep` and `zstdless` shell scripts were not installed to `usr/bin/` (their respective man pages were installed correctly), so, I had to install them manually" in our build infrastructure.

terrelln:
The CMake build is community maintained. We'd gladly accept a PR to improve the situation!

LeeYoung624:
This PR should help
[Cmake bug fix: CMake does not install zstdgrep and zstdless](https://github.com/facebook/zstd/pull/1625)

Cyan4973:
@LeeYoung624 patch merged

PR #1625, opened 2019-06-03
Cmake bug fix: CMake does not install zstdgrep and zstdless

LeeYoung624:
As it's described here
[CMake does not install zstdgrep and zstdless shell scripts](https://github.com/facebook/zstd/issues/1503)

Copy zstdgrep and zstdless in CMake make install, and grant them execute permission

### DIFF
--- a/build/cmake/programs/CMakeLists.txt
+++ b/build/cmake/programs/CMakeLists.txt
@@ -38,6 +38,8 @@ if (UNIX)
     add_custom_target(unzstd ALL ${CMAKE_COMMAND} -E create_symlink zstd unzstd DEPENDS zstd COMMENT "Creating unzstd symlink")
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/zstdcat DESTINATION "bin")
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/unzstd DESTINATION "bin")
+    install(PROGRAMS ${CMAKE_SOURCE_DIR}/../../programs/zstdgrep DESTINATION "bin")
+    install(PROGRAMS ${CMAKE_SOURCE_DIR}/../../programs/zstdless DESTINATION "bin")
 
     add_custom_target(zstd.1 ALL
         ${CMAKE_COMMAND} -E copy ${PROGRAMS_DIR}/zstd.1 .
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1290,7 +1290,6 @@ static void ZSTD_invalidateMatchState(ZSTD_matchState_t* ms)
     ZSTD_window_clear(&ms->window);
 
     ms->nextToUpdate = ms->window.dictLimit;
-    ms->nextToUpdate3 = ms->window.dictLimit;
     ms->loadedDictEnd = 0;
     ms->opt.litLengthSum = 0;  /* force reset of btopt stats */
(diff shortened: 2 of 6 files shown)


build/cmake/programs/CMakeLists.txt
```diff
@@ -38,6 +38,8 @@ if (UNIX)
     add_custom_target(unzstd ALL ${CMAKE_COMMAND} -E create_symlink zstd unzstd DEPENDS zstd COMMENT "Creating unzstd symlink")
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/zstdcat DESTINATION "bin")
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/unzstd DESTINATION "bin")
+    install(FILES ${CMAKE_SOURCE_DIR}/../../programs/zstdgrep DESTINATION "bin" PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+    install(FILES ${CMAKE_SOURCE_DIR}/../../programs/zstdless DESTINATION "bin" PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
```
Mno-hime: Perhaps there's too much toggling or permissions?
```suggestion
    install(FILES ${CMAKE_SOURCE_DIR}/../../programs/zstdless DESTINATION "bin")
```

LeeYoung624: well, the default permissions is OWNER_WRITE, OWNER_READ, GROUP_READ, and WORLD_READ
[cmake install](https://cmake.org/cmake/help/v3.13/command/install.html)
which means the users can not directly use zstdless and zstdgrep after make install.
Users need to chmod for zstdless and zstdgrep.
I tested on Ubuntu, Redhat, Centos and suse, and all of them show the same result if I didn't add the execute permissions
`-bash: /usr/local/bin/zstdless: Permission denied`

I think this should be done in cmake scripts.

And of course, the commits should be squashed. Thank you so much

felixhandte: Can you explain why these permissions need to be explicitly set here, while they don't on the `zstd` install command? I guess it's because `zstd` is a `TARGETS` object, while these are `FILES` objects. Would it make more sense for these to be `TARGETS` themselves?

Or maybe these should be `PROGRAMS`?

> The PROGRAMS form is identical to the FILES form except that the default permissions for the installed file also include OWNER_EXECUTE, GROUP_EXECUTE, and WORLD_EXECUTE. This form is intended to install programs that are not targets, such as shell scripts. Use the TARGETS form to install targets built within the project.

LeeYoung624: @felixhandte I checked the camke documentation and ask some guys familiar with CMake. I think you are right. These should be `PROGRAMS`. Thank you so much.

felixhandte:
@LeeYoung624, can you fix your stack so that it doesn't include rebases of commits already merged?

LeeYoung624:
Yes @felixhandte I will rebase this

LeeYoung624:
@felixhandte seems I mess up the stack of this PR.. I will close thie PR and open a new one for this cmake fix. Sorry about this

PR #1631, opened 2019-06-04
zstdcli : expose cpu load indicator for each file on -vv mode

ephiepark:
Moved the logic to measure elapsed time and cpu load from FIO_compressFilename to FIO_compressFilename_internal.

1. FIO_compressFilename is only used for single file case. If we want to show elapsed time and cpu load for each file in multiple file scenario, the measurement needs to be done in FIO_compressFilename_internal, which is called by both single and multiple file scenarios. 
2. I could have added the logic to FIO_compressFilename_internal, FIO_compressFilename_dstFile, or FIO_compressFilename_srcFile. I thought FIO_compressFilename_internal is a good fit since it already has logic to print stats about compression (compression ratio). I see elapsed time and cpu load as the extension to the stats.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1236,6 +1236,9 @@ static int FIO_compressFilename_dstFile(FIO_prefs_t* const prefs,
                                         const char* srcFileName,
                                         int compressionLevel)
 {
+    UTIL_time_t timeStart;
+    clock_t cpuStart;
+
     int closeDstFile = 0;
     int result;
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good to me !

I believe you selected the right place to provide this capability.

Also, the patch uses a single location, doesn't bleed its logic outside of its block, and is easy to follow.
These are good properties for its maintenance.

PR #1632, opened 2019-06-05
CircleCI : use custom Docker Image with pre-installed dependencies

ephiepark:
Instead of installing dependencies before test, use custom Docker Image with pre-installed dependencies.


Cyan4973:
Thanks @ephiepark !
This is great, this change make "CircleCI" tests start faster,
they are now below the 5mn mark, which was the goal.

Issue #1634, opened 2019-06-05
Higher strategies have dramatic slowdowns on repetitive data


bruceg:
When compressing strings of a single character (ie streams of zeroes), strategies 6-9 (btlazy2, btopt, btultra, and btultra2) show dramatic slowdowns. In particular, btlazy2 is about 6.3x slower than lazy2, btopt and btultra are about 3.7x slower than btlazy2 (23.3x lazy2), and btultra2 is about 3.9x slower than btultra (92.5x lazy2). Combined with the dictionary size increase, this makes zstd -19 about 400x slower than zstd -9 on this kind of data.

Obviously, I would expect higher level strategies to use more time, but the increase is out of proportion to the time increase with less repetitive data (more like 10x slower for -19 vs -9).

My test:
```
> for n in $(seq 9); do echo strategy=$n; dd status=none if=/dev/zero bs=1024k count=1000 | command time -f %U zstd --zstd=strategy=$n >/dev/null; done
strategy=1
0.34
strategy=2
0.36
strategy=3
0.41
strategy=4
0.93
strategy=5
1.33
strategy=6
8.41
strategy=7
31.03
strategy=8
31.13
strategy=9
123.05
```

The context: I was compressing a raw disk image (produced by ntfs_clone) with zstd, and decided to try increasing the compression with zstd -19. The image produced by ntfs_clone contains zero blocks wherever the space is unallocated. Two days later it finished compressing. This seemed somewhat ridiculous, so I dug in to see at what point it became very slow. At some points I wondered if zstd was hung somehow, given that it didn't seem to be making progress.

terrelln:
What version of zstd are you using?

terrelln:
Aha: This is a problem with multithreading interactions. When I add `--single-thread` the problem goes away. I was a bit confused because I know we handle repetitive data specially, but we must be missing that optimization with multithreading somehow. I'll look into it.

```
for n in $(seq 9); do echo strategy=$n; dd status=none if=/dev/zero bs=1024k count=1000 | command time -f %U zstd --single-thread --zstd=strategy=$n,tlen=128 >/dev/null; done
strategy=1
0.27
strategy=2
0.27
strategy=3
0.30
strategy=4
0.54
strategy=5
0.63
strategy=6
0.61
strategy=7
0.36
strategy=8
0.40
strategy=9
0.43
```

terrelln:
[...]

PR #1635, opened 2019-06-06
[libzstd] Optimize ZSTD_insertBt1() for repetitive data

terrelln:
We would only skip at most 192 bytes at a time before this diff.
This was added to optimize long matches and skip the middle of the
match. However, it doesn't handle the case of repetitive data.

This patch keeps the optimization, but also handles repetitive data
by taking the max of the two return values.

**Before:**
```
> for n in $(seq 9); do echo strategy=$n; dd status=none if=/dev/zero bs=1024k count=1000 | command time -f %U zstd --zstd=strategy=$n >/dev/null; done
strategy=1
0.34
strategy=2
0.36
strategy=3
0.41
strategy=4
0.93
strategy=5
1.33
strategy=6
8.41
strategy=7
31.03
strategy=8
31.13
strategy=9
123.05
```

**After:**
```
> for n in $(seq 9); do echo strategy=$n; dd status=none if=/dev/zero bs=1024k count=1000 | command time -f %U ./zstd --zstd=strategy=$n >/dev/null; done
strategy=1
0.27
strategy=2
0.23
strategy=3
0.27
strategy=4
0.43
strategy=5
0.56
strategy=6
0.43
strategy=7
0.34
strategy=8
0.34
strategy=9
0.35
```

At level 19 with multithreading the compressed size of `silesia.tar` regresses 300 bytes, and `enwik8` regresses 100 bytes. In single threaded mode `enwik8` is also within 100 bytes, and I didn't test `silesia.tar`. The compression speed is not significantly effected.

The regression tests show changes in the order of single digit number of bytes.

Fixes Issue #1634.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -506,9 +506,11 @@ static U32 ZSTD_insertBt1(
     }   }
 
     *smallerPtr = *largerPtr = 0;
-    if (bestLength > 384) return MIN(192, (U32)(bestLength - 384));   /* speed optimization */
-    assert(matchEndIdx > current + 8);
-    return matchEndIdx - (current + 8);
+    {   U32 positions = 0;
+        if (bestLength > 384) positions = MIN(192, (U32)(bestLength - 384));   /* speed optimization */
+        assert(matchEndIdx > current + 8);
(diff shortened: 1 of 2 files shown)


Cyan4973:
What about speed benefits ?

PR #1640, opened 2019-06-07
Protect lstat() With Better Macro Guard

### LINKED ISSUE #1607: zstd is compressing symbolic links with `std=c99`
@Cyan4973 Hmm, I might find a test error that zstd have been 
compressed the symbolic link.

_Originally posted by @lzutao in https://github.com/facebook/zstd/issues/1605#issuecomment-487978531_

felixhandte:
Resolves #1607.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -107,19 +107,11 @@ int UTIL_isSameFile(const char* file1, const char* file2)
 U32 UTIL_isLink(const char* infilename)
 {
 /* macro guards, as defined in : https://linux.die.net/man/2/lstat */
-#ifndef __STRICT_ANSI__
-#if defined(_BSD_SOURCE) \
-    || (defined(_XOPEN_SOURCE) && (_XOPEN_SOURCE >= 500)) \
-    || (defined(_XOPEN_SOURCE) && defined(_XOPEN_SOURCE_EXTENDED)) \
-    || (defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 200112L)) \
-    || (defined(__APPLE__) && defined(__MACH__)) \
-    || defined(__OpenBSD__) \
-    || defined(__FreeBSD__)
+#if PLATFORM_POSIX_VERSION >= 200112L
     int r;
     stat_t statbuf;
     r = lstat(infilename, &statbuf);
     if (!r && S_ISLNK(statbuf.st_mode)) return 1;
-#endif
 #endif
     (void)infilename;
     return 0;
--- a/.travis.yml
+++ b/.travis.yml
@@ -153,6 +153,12 @@ matrix:
         - make clean
         - make travis-install    # just ensures `make install` works
 
+    - name: c99 compatibility
+      if: branch = master
+      script:
+        - make c99build
+        - make -C tests test-zstd
+
     - name: PPC64
       if: branch = master
       script:
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -247,7 +247,7 @@ clean:
 	$(MAKE) -C $(ZSTDDIR) clean
 	$(MAKE) -C $(PRGDIR) clean
 	@$(RM) -fR $(TESTARTEFACT)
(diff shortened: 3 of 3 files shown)


terrelln:
Can you explain why this works, and how we know it works? What platform/configurations have you tested?

felixhandte:
The goal is to maximally make use of `lstat()` when the platform is known to provide it, while ensuring we never try to use it when the platform doesn't. This macro, defined [here](https://github.com/facebook/zstd/blob/dev/programs/platform.h#L68), detects the level with which the environment conforms to POSIX. I believe that that test should be sufficient to only enable this on platforms that provide `lstat()`. I'm less confident about it succeeding on all platforms that provide `lstat()`. It does however work with `-std=c99`, where the previous did not.

programs/util.c
```diff
@@ -107,19 +107,11 @@ int UTIL_isSameFile(const char* file1, const char* file2)
...
 /* macro guards, as defined in : https://linux.die.net/man/2/lstat */
-#ifndef __STRICT_ANSI__
-#if defined(_BSD_SOURCE) \
-    || (defined(_XOPEN_SOURCE) && (_XOPEN_SOURCE >= 500)) \
-    || (defined(_XOPEN_SOURCE) && defined(_XOPEN_SOURCE_EXTENDED)) \
-    || (defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 200112L)) \
-    || (defined(__APPLE__) && defined(__MACH__)) \
-    || defined(__OpenBSD__) \
-    || defined(__FreeBSD__)
+#if PLATFORM_POSIX_VERSION >= 200112L
```
Cyan4973: One way to improve the likelihood that this code path selector is correct (i.e. will compile properly)
is to also check if the macro `S_ISLNK` (used below) is defined.

Cyan4973:
We probably need a test somewhere.
Something that would break without this patch, and pass with it.
Maybe compressing a link with a cli compiled with `std=c99` for example ?

felixhandte:
@Cyan4973, the obvious way to do that is to add a `c99build` to Travis and invoke `playTests.sh` after building. It does take a while to run, though. Is that acceptable? I'm not sure what's important in optimizing contbuild efficiency. It looks like the other language variants only run on master.

I'll put up a proposal commit, let me know what you think.

Cyan4973:
Running the full test suite just to check this minor patch seems oversized. We need to pay attention to test times.
Besides, I don't even think it tests the issue at stake (#1607 : compressing symbolic link).

Piggy backing on the `c99` build test is fine. 
I suspect we need a _new_ small test which verifies that #1607 is fixed.
It should not spend more than 0.1s on this.

felixhandte:
The following does fail before this patch and pass afterwards:

```
make clean
make c99build
make -C tests test-zstd
```

(It fails [here](https://github.com/facebook/zstd/blob/dev/tests/playTests.sh#L349)).

But ok, if you'd rather run a narrow targeted test on this issue, I can gin that up.

Cyan4973:
Wow, we have a whole test section on compression of symbolic links. I totally forgot about that !
So yes, that's indeed the kind of test we need.

I see this test will only play during daily `cron` job, and not at every commit, reducing its impact on the development loop. It seems fine considering its scope.

Although it tests more than we need for this patch, it can be argued that it is preferable to have all CLI basic tests run against the `c99` build, and that it should be part for CI anyway.

So it's good for me.

felixhandte:
Aha, I just put up a specific test. I can remove that and go back to the full test suite, or I can retain this new test, or I can do both. Let me know what you think is best.

Cyan4973:
If it was only for this PR, we would only need the symlinks tests.

But considering the broader picture, 
`c99build` has been a compilation test only up to now.
It feels that having coverage over all basic cli features when compiling with `std=c99` would be a good thing for the project.
So why not fixing that too.

Selecting the "fast" `test-zstd` target, as you did, seems the right fit for this goal.

PR #1644, opened 2019-06-09
[programs] set chmod 600 after opening destination file

### LINKED ISSUE #1630: zstd adds read permissions to files while being compressed or uncompressed
While the final file mode is reflective of the input file, when compressing or uncompressing, the file can temporarily gain greater permissions than the input and potentially leading to security issues (especially if large files are being handled).

Example: `file` has mode 600 -> `zstd file` -> `file.zst` has mode 644 while compression is happening. (Same happens with `unzstd file.zst` and `file`)

chungy:
This resolves a race condition where zstd or unzstd may expose read
permissions beyond the original file allowed.  Mode 600 is used
temporarily during the compression and decompression write stage
and the new file inherits the original file’s mode at the end.

Fixes #1630

chmod() should silently fail in situations where it isn't possible to set mode 600, such as on FAT file systems. I'm also not really able to tell if this builds okay on Windows. mingw-w64 seems to build it fine (as long as I changed timefn.h to have `<windows.h>` instead of `<Windows.h>`), but i don't know if chmod() has any effect on that OS -- it's well possible that someone should make a Windows-specific equivalent to this.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -564,9 +564,11 @@ static FILE* FIO_openDstFile(FIO_prefs_t* const prefs, const char* srcFileName, 
     }   }
 
     {   FILE* const f = fopen( dstFileName, "wb" );
-        if (f == NULL)
+        if (f == NULL) {
             DISPLAYLEVEL(1, "zstd: %s: %s\n", dstFileName, strerror(errno));
-        chmod(dstFileName, 00600);
+        } else {
+            chmod(dstFileName, 00600);
+        }
         return f;
     }
 }
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -566,6 +566,7 @@ static FILE* FIO_openDstFile(FIO_prefs_t* const prefs, const char* srcFileName, 
     {   FILE* const f = fopen( dstFileName, "wb" );
         if (f == NULL)
             DISPLAYLEVEL(1, "zstd: %s: %s\n", dstFileName, strerror(errno));
+        chmod(dstFileName, 00600);
         return f;
     }
 }


Cyan4973:
Thanks @chungy .
Thankfully, we have automated Windows tests, using both `mingw` and Visual Studio.
It seems your patch passed them, which suggests compilation (at least) works fine.

`chmod` is a Unix command, but it seems to have a corresponding Windows hook.
We already use it in multiple other places, so if it was a bad thing, we would have already triggered some bug reports.

cc @felixhandte for the review.

programs/fileio.c
```diff
@@ -566,6 +566,7 @@ static FILE* FIO_openDstFile(FIO_prefs_t* const prefs, const char* srcFileName,
     {   FILE* const f = fopen( dstFileName, "wb" );
         if (f == NULL)
             DISPLAYLEVEL(1, "zstd: %s: %s\n", dstFileName, strerror(errno));
+        chmod(dstFileName, 00600);
```
Cyan4973: I understand that `chmod()` can fail silently,
yet it would feel cleaner if `chmod()` wasn't attempted when `f == NULL`.
an `if { } else { } ` should do the trick.

chungy: Totally right, I did miss that. :)

felixhandte:
I wonder about the race condition that is potentially introduced here, where `zstd` first opens a file with liberal permissions, and then closes them down. It seems like a malicious process could open the file during that gap and then read the contents as they are written.

An alternative approach would be to temporarily set the process umask to `umask | 00077`, do the open, and then restore the original umask. It seems though that windows does not have an equivalent of umask, so that approach would be less portable.

What do you think?

Cyan4973:
I believe this patch solves the more pressing issue of ensuring that the temporary file cannot be read during compression, which is a long time compared to the very tiny window of opportunity between file creation and rights setting.
Also, this simple patch does not introduce additional dependencies nor portability complexity.

svenha:
This PR shows an unexpected change in 1.4.1 (compared to previous versions): if compressing from stdin to a file, the file always has final rights 0600 instead of umask. This breaks a lot of code I fear.
Example:

    echo "123456789" | zstd -14 -f -q -o test.zst

chungy:
Possible workaround: use shell redirection (`zstd > test.zst`) instead of `-o test.zst`

Though yes, it is a new bug and you should file a regular issue on it so it doesn't become forgotten in this merged PR :)

PR #1652, opened 2019-06-14
[programs] set restrictive umask before file creation

chungy:
This resolves a condition where zstd or unzstd may expose read
permissions beyond what the original file allowed.  umask 077
blocks out all permissions for group and other while the file
is being written to by zstd, and gets reset to the source file’s
mode afterward.

As suggested by @felixhandte in PR #1644, I think this solution is cleaner and removes the chance of reading the file before a chmod() call could be done. That PR should be closed if this one is accepted instead (one or the other).

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -533,6 +533,11 @@ static FILE* FIO_openDstFile(FIO_prefs_t* const prefs, const char* srcFileName, 
         prefs->sparseFileSupport = ZSTD_SPARSE_DEFAULT;
     }
 
+    /* set the umask so group and other permissions are not set,
+     * preventing possible security leaks when compressing or
+     * uncompressing */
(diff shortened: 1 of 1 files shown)


chungy:
the chmod PR was pulled and one of the CI tests on this failed for some reason (the log doesn't tell me anything...). Closing. :)

Cyan4973:
The error was : 
```
c:\projects\zstd-p0yf0\programs\fileio.c(539): warning C4996: 
'umask': The POSIX name for this item is deprecated. 
Instead, use the ISO C and C++ conformant name: _umask.
```

at https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/builds/25271621/job/4j0mfu1kvn7317nt#L40.

PR #1656, opened 2019-06-18
Adding shrinking flag for cover and fastcover

Tyler-Tran:
Created a flag that allows the dictionary builder to select the smallest dictionary that is within a 1.01 regression of the largest dictionary.


lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
   }
 }
 
+static void COVER_selectDict(void* dict, size_t* dictBufferCapacity, const void* customDictContent, size_t dictContentSize,
+        const void* samplesBuffer, const size_t* samplesSizes, size_t nbTrainSamples, size_t nbSamples, ZDICT_cover_params_t params,
+        size_t* offsets, size_t* totalCompressedSize){
+
+  size_t largestDict = 0;
+  size_t largestCompressed = 0;
+  double defaultRegression = 1.01;
```
terrelln: Let's add the `autoShrinkDict` and `autoShrinkDictMaxRegression` parameters to COVER and FASTCOVER here https://github.com/facebook/zstd/blob/dev/lib/dictBuilder/zdict.h#L96, and use those in this function.

The regression should be an `unsigned` percent, instead of a `double`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
   }
 }
 
+static void COVER_selectDict(void* dict, size_t* dictBufferCapacity, const void* customDictContent, size_t dictContentSize,
```
terrelln: Instead of using out-parameters, lets return a struct. We also need to handle returning errors.

```c
struct COVER_dictSelection {
  const void* dictContent;
  size_t dictSize;
  size_t totalCompressedSize;
};

COVER_dictSelection COVER_dictSelectionError(size_t error) {
    const COVER_dictSelection selection = { NULL, 0, error };
    return selection;
}

bool COVER_dictSelectionIsError(COVER_dictSelection selection) {
  return ZSTD_isError(selection.totalCompressedSize);
}
```

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+  double defaultRegression = 1.01;
+
+  /* Initial dictionary size and compressed size */
+  {
+    *dictBufferCapacity = ZDICT_finalizeDictionary(
+      dict, *dictBufferCapacity, customDictContent, dictContentSize,
+      samplesBuffer, samplesSizes, (unsigned)nbTrainSamples, params.zParams);
+
+    if (ZDICT_isError(*dictBufferCapacity)) {
+      return;
```
terrelln: `return ZSTD_dictSelectionError(*dictBufferCapacity);`

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+    }
+  }
+
+  *totalCompressedSize = COVER_checkTotalCompressedSize(params, samplesSizes,
```
terrelln: We should check for an error. If this call errors we should error.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+      *dictBufferCapacity = ZDICT_finalizeDictionary(
+        dict, *dictBufferCapacity, customDictContent, dictContentSize,
+        samplesBuffer, samplesSizes, (unsigned)nbTrainSamples, params.zParams);
+
+      if (ZDICT_isError(*dictBufferCapacity)) {
+        return;
+      }
+    }
+
+    *totalCompressedSize = COVER_checkTotalCompressedSize(params, samplesSizes,
```
terrelln: Should check for an error here and skip this dictionary if it errors.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+                                                         nbTrainSamples, nbSamples,
+                                                         (BYTE *const)dict, *dictBufferCapacity);
+
+    if (*totalCompressedSize <= largestCompressed * defaultRegression){
+      break;
+    }
+    dictContentSize *= 2;
+    *dictBufferCapacity *= 2;
+  }
+  return;
```
terrelln: If we don't find any dictionaries better than the default one, this is buggy. Since we will have the wrong `dictBufferCapacity`, and `totalCompressedSize`.

lib/dictBuilder/zdict.h
```diff
@@ -82,6 +82,7 @@ typedef struct {
     int      compressionLevel;   /* optimize for a specific zstd compression level; 0 means default */
     unsigned notificationLevel;  /* Write log to stderr; 0 = none (default); 1 = errors; 2 = progression; 3 = details; 4 = debug; */
     unsigned dictID;             /* force dictID value; 0 means auto mode (32-bits random value) */
+    unsigned shrinkDict;         /* shrink dictionaries to select smallest within 1% regression */
```
terrelln: This belongs in the `COVER` and `FASTCOVER` parameters, since the `ZDICT` dictionary builder doesn't support this.

Let's make the regression an unsigned percent, instead of a double.

programs/zstdcli.c
```diff
@@ -77,6 +77,7 @@ static const unsigned g_defaultMaxDictSize = 110 KB;
 static const int      g_defaultDictCLevel = 3;
 static const unsigned g_defaultSelectivityLevel = 9;
 static const unsigned g_defaultMaxWindowLog = 27;
+static const double   g_defaultDictRegression = 1.01;
```
terrelln: Add support for this.

We can't accept doubles on the command line. Instead, can we take a percent 0, 1, etc?

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+    return;
+  }
+
+  largestDict = *dictBufferCapacity;
+  largestCompressed = *totalCompressedSize;
+  *dictBufferCapacity = ZDICT_DICTSIZE_MIN;
+  dictContentSize = ZDICT_CONTENTSIZE_MIN;
+
+  /* Largest dict is initially at least ZDICT_DICTSIZE_MIN */
+  while (*dictBufferCapacity <= largestDict){
```
terrelln: No point checking the largest dict twice, this should be `<`

terrelln:
I haven't reviewed yet, but can you please add a test case in `tests/playTests.sh`, where the other dictionary builder tests are?

terrelln:
Also, lets add a test case in `tests/fuzzer.c` where the other dict builder tests are

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,108 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+  size_t largestDict = 0;
+  size_t largestCompressed = 0;
+  double regressionTolerance = 1 + ((double)params.shrinkDict / 100);
+  char* customDictContentEnd = ((char*)customDictContent + dictContentSize);
+
+  BYTE *const dict = (BYTE * const)malloc(dictContentSize);
+  BYTE *const dictBuffer = (BYTE * const)malloc(dictContentSize);
+
+  if (!dict || !dictBuffer) {
+    if (dict) free(dict);
```
terrelln: nit: just `free(dict)`, free handles `NULL` correctly, here and elsewhere.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,108 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+      const COVER_dictSelection_t selection = { dictBuffer, dictContentSize, totalCompressedSize };
+      if (dict) free(dict);
+      return selection;
+    }
+    dictContentSize *= 2;
+  }
+  dictContentSize = largestDict;
+  totalCompressedSize = largestCompressed;
+  {
+    const COVER_dictSelection_t selection = { dictBuffer, dictContentSize, totalCompressedSize };
```
terrelln: I love the const!

lib/dictBuilder/cover.c
```diff
@@ -969,21 +1071,22 @@ static void COVER_tryParameters(void *opaque) {
...
-  /* Check total compressed size */
-  totalCompressedSize = COVER_checkTotalCompressedSize(parameters, ctx->samplesSizes,
-                                                       ctx->samples, ctx->offsets,
-                                                       ctx->nbTrainSamples, ctx->nbSamples,
-                                                       dict, dictBufferCapacity);
+    dictBufferCapacity = selection.dictSize;
+    totalCompressedSize = selection.totalCompressedSize;
+    memcpy(dict, selection.dictContent, dictBufferCapacity);
 
+    if (selection.dictContent) {
```
terrelln: nit: Lets make a function `ZSTD_dictSelectionFree(selection)` that will free `dictContent`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,65 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+  }
+
+  largestDict = *dictBufferCapacity;
+  largestCompressed = *totalCompressedSize;
+  *dictBufferCapacity = ZDICT_DICTSIZE_MIN;
+  dictContentSize = ZDICT_CONTENTSIZE_MIN;
+
+  /* Largest dict is initially at least ZDICT_DICTSIZE_MIN */
+  while (*dictBufferCapacity <= largestDict){
+    {
```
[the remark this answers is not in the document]
terrelln: This still needs to be fixed. It can just be

```
memcpy(...);
```

instead of 

```
{
  memcpy();
}
``

File: lib/dictBuilder/cover.c [hunk 14]
```diff
@@ -929,6 +929,108 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+COVER_dictSelection_t COVER_selectDict(void* customDictContent,
+        size_t dictContentSize, const void* samplesBuffer, const size_t* samplesSizes, size_t nbFinalizeSamples,
+        size_t nbCheckSamples, size_t nbSamples, ZDICT_cover_params_t params, size_t* offsets, size_t totalCompressedSize) {
+
+  size_t largestDict = 0;
+  size_t largestCompressed = 0;
+  double regressionTolerance = 1 + ((double)params.shrinkDict / 100);
+  char* customDictContentEnd = ((char*)customDictContent + dictContentSize);
+
+  BYTE *const dict = (BYTE * const)malloc(dictContentSize);
```
terrelln: nit: Lets have better names than `dict` and `dictBuffer`.

Maybe `largestDictbuffer` and `smallerDictBuffer` or `candidateDictBuffer`?

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+}
+
+COVER_dictSelection_t COVER_selectDict(void* customDictContent,
+        size_t dictContentSize, const void* samplesBuffer, const size_t* samplesSizes, size_t nbFinalizeSamples,
+        size_t nbCheckSamples, size_t nbSamples, ZDICT_cover_params_t params, size_t* offsets, size_t totalCompressedSize) {
+
+  size_t largestDict = 0;
+  size_t largestCompressed = 0;
+  double regressionTolerance = 1 + ((double)params.shrinkDict / 100);
+  char* customDictContentEnd = ((char*)customDictContent + dictContentSize);
```
terrelln: nit: `BYTE` not `char`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+  if (!largestDictbuffer || !candidateDictBuffer) {
+    free(largestDictbuffer);
+    free(candidateDictBuffer);
+    return COVER_dictSelectionError(dictContentSize);
+  }
+
+  /* Initial dictionary size and compressed size */
+  memcpy(largestDictbuffer, customDictContent, dictContentSize);
+  dictContentSize = ZDICT_finalizeDictionary(
+    largestDictbuffer, dictContentSize, (const void*)customDictContent, dictContentSize,
```
terrelln: You don't need the `customDictContent` cast.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+    samplesBuffer, samplesSizes, (unsigned)nbFinalizeSamples, params.zParams);
```
terrelln: Instead of casting, have the function take an `unsigned`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+
+  if (ZDICT_isError(dictContentSize)) {
+    free(largestDictbuffer);
+    free(candidateDictBuffer);
+    return COVER_dictSelectionError(dictContentSize);
+  }
+
+  totalCompressedSize = COVER_checkTotalCompressedSize(params, samplesSizes,
+                                                       (const BYTE*)samplesBuffer, offsets,
```
terrelln: Instead of casting `samplesBuffer` have the function take a `BYTE const*`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+                                                       nbCheckSamples, nbSamples,
+                                                       (BYTE *const)largestDictbuffer, dictContentSize);
```
terrelln: You don't need this cast.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+    return selection;
+  }
+
+  largestDict = dictContentSize;
+  largestCompressed = totalCompressedSize;
+  dictContentSize = ZDICT_DICTSIZE_MIN;
+
+  /* Largest dict is initially at least ZDICT_DICTSIZE_MIN */
+  while (dictContentSize < largestDict) {
+    {
```
terrelln: Sorry, this comment got hidden by GitHub. You don't need this block (see the resolved conversation).

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+      memcpy(candidateDictBuffer, largestDictbuffer, largestDict);
+      dictContentSize = ZDICT_finalizeDictionary(
+        candidateDictBuffer, dictContentSize, (const void*)(customDictContentEnd - dictContentSize), dictContentSize,
```
terrelln: You don't need this cast.

Any pointer type is implicitly convertible to a `void*`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+        samplesBuffer, samplesSizes, (unsigned)nbFinalizeSamples, params.zParams);
```
terrelln: You don't need this cast if you take in an `unsigned`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+
+      if (ZDICT_isError(dictContentSize)) {
+        free(largestDictbuffer);
+        free(candidateDictBuffer);
+        return COVER_dictSelectionError(dictContentSize);
+      }
+    }
+
+    totalCompressedSize = COVER_checkTotalCompressedSize(params, samplesSizes,
+                                                         (const BYTE*)samplesBuffer, offsets,
```
terrelln: Same with these casts.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,112 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+      const COVER_dictSelection_t selection = { candidateDictBuffer, dictContentSize, totalCompressedSize };
+      free(largestDictbuffer);
+      return selection;
+    }
+    dictContentSize *= 2;
+  }
+  dictContentSize = largestDict;
+  totalCompressedSize = largestCompressed;
+  {
+    const COVER_dictSelection_t selection = { candidateDictBuffer, dictContentSize, totalCompressedSize };
```
terrelln: Is this a bug?
1. We're returning the `candidateDictBuffer`, but giving it the size/compressed size of `largestDict`.
2. Shouldn't we return `largestDict` if we didn't find any smaller ones that were better?

lib/dictBuilder/zdict.h
```diff
@@ -94,6 +94,7 @@ typedef struct {
     unsigned steps;              /* Number of steps : Only used for optimization : 0 means default (40) : Higher means more parameters checked */
     unsigned nbThreads;          /* Number of threads : constraint: 0 < nbThreads : 1 means single-threaded : Only used for optimization : Ignored if ZSTD_MULTITHREAD is not defined */
     double splitPoint;           /* Percentage of samples used for training: Only used for optimization : the first nbSamples * splitPoint samples will be used to training, the last nbSamples * (1 - splitPoint) samples will be used for testing, 0 means default (1.0), 1.0 when all samples are used for both training and testing */
+    unsigned shrinkDict;         /* shrink dictionaries to select smallest within specified regression */
```
terrelln: nit: Better comment here. What does the value mean? What does 0 mean?

programs/zstdcli.c
```diff
@@ -185,6 +186,7 @@ static int usage_advanced(const char* programName)
     DISPLAY( " -o file : `file` is dictionary name (default: %s) \n", g_defaultDictName);
     DISPLAY( "--maxdict=# : limit dictionary to specified size (default: %u) \n", g_defaultMaxDictSize);
     DISPLAY( "--dictID=# : force dictionary ID to specified value (default: random)\n");
+    DISPLAY( "--shrink-dict=#: select the smallest dictionary within given regression of the largest dictionary (default: %u)\n", g_defaultDictRegression);
```
terrelln: nit: better description and what does `0` mean?

programs/zstdcli.c
```diff
@@ -644,6 +649,7 @@ int main(int argCount, const char* argv[])
                     if (!strcmp(argument, "--train")) { operation=zom_train; if (outFileName==NULL) outFileName=g_defaultDictName; continue; }
                     if (!strcmp(argument, "--maxdict")) { nextArgumentIsMaxDict=1; lastCommand=1; continue; }  /* kept available for compatibility with old syntax ; will be removed one day */
                     if (!strcmp(argument, "--dictID")) { nextArgumentIsDictID=1; lastCommand=1; continue; }  /* kept available for compatibility with old syntax ; will be removed one day */
+                    if (!strcmp(argument, "--shrink-dict")) { nextArgumentIsShrinkDict=1; lastCommand=1; continue; }  /* kept available for compatibility with old syntax ; will be removed one day */
```
terrelln: Add this to `parseFastCoverParameters` and `parseCoverParameters`. Maybe just call it `shrink`?

Check how `split` is implemented.

programs/zstdcli.c
```diff
@@ -557,6 +561,7 @@ int main(int argCount, const char* argv[])
     unsigned dictID = 0;
     int dictCLevel = g_defaultDictCLevel;
     unsigned dictSelect = g_defaultSelectivityLevel;
+    unsigned shrinkDict = g_defaultDictRegression;
```
terrelln: You don't need this when you move to `parseCoverParameters()`.

tests/fuzzer.c
```diff
@@ -1104,8 +1104,38 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+            fastCoverParams.shrinkDict = 1;
+            dictSize = ZDICT_optimizeTrainFromBuffer_fastCover(
+                dictBuffer, dictBufferCapacity,
+                CNBuffer, samplesSizes, nbSamples,
+                &fastCoverParams);
+            if (ZDICT_isError(dictSize)) goto _output_error;
+        }
+        DISPLAYLEVEL(3, "OK, created dictionary of size %u \n", (unsigned)dictSize);
+
+        DISPLAYLEVEL(3, "test%3i : FASTCOVER dictBuilder with shrinkDict: ", testNb++);
```
terrelln: nit: I think you switched the names.

tests/playTests.sh
```diff
@@ -499,6 +499,9 @@ $ZSTD --train-fastcover=k=56,d=8 && die "Create dictionary without input file"
 println "- Create dictionary with short dictID"
 $ZSTD --train-fastcover=k=46,d=8,f=15,split=80 "$TESTDIR"/*.c "$PRGDIR"/*.c --dictID=1 -o tmpDict1
 cmp tmpDict tmpDict1 && die "dictionaries should have different ID !"
+println "- Create dictionaries with shrink-dict flag enabled"
+$ZSTD --train-fastcover=steps=256 --shrink-dict=1 "$TESTDIR"/*.c "$PRGDIR"/*.c --dictID=1 -o shrinkDict
```
terrelln: e.g. will be `$ZSTD --train-fastcover=steps=256,shrink=1 <files> --dictID=1 -o shrinkDict` with the parsing parameters.

Also, don't need `--dictID=1`, thats an unnecessary parameter.

tests/playTests.sh
```diff
@@ -499,6 +499,9 @@ $ZSTD --train-fastcover=k=56,d=8 && die "Create dictionary without input file"
...
+$ZSTD --train-fastcover=steps=256 --shrink-dict=5 "$TESTDIR"/*.c "$PRGDIR"/*.c --dictID=1 -o shrinkDict1
```
terrelln: Call it `tmpShrinkDict` and `tmpShrinkDict1`. The script removes files starting with `tmp*`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +929,111 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
...
+        size_t dictContentSize, const BYTE* samplesBuffer, const size_t* samplesSizes, unsigned nbFinalizeSamples,
+        size_t nbCheckSamples, size_t nbSamples, ZDICT_cover_params_t params, size_t* offsets, size_t totalCompressedSize) {
+
+  size_t largestDict = 0;
+  size_t largestCompressed = 0;
+  BYTE* customDictContentEnd = customDictContent + dictContentSize;
+
+  BYTE *const largestDictbuffer = (BYTE * const)malloc(dictContentSize);
+  BYTE *const candidateDictBuffer = (BYTE * const)malloc(dictContentSize);
+  double regressionTolerance = params.shrinkDict + 1.00;
```
terrelln: I liked taking the parameter as an `unsigned`, but we should distinguish:

1. Auto shrinking enabled vs disabled. If you need fast dictionary training, you don't want auto shrinking.
   It can be disabled by default.
2. The regression we allow. A regression of 0 could make sense if you want to take a smaller dictionary only if it is better.

I think we could take two flags `shrinkDict` and `shrinkDictMaxRegression`, both `unsigned`. It was a mistake to take `split` as a double, so I don't want to repeat it here.

On the library side, we can default to disabled, and the regression defaults to 0.

On the CLI, we can default to disabled, but we could say if someone sets `--train-fastcover=shrink` we enable with default to a regression of 1, but `--train-fastcover=shrink=0` or `--train-fastcover=shrink=2` should both still work.

lib/dictBuilder/cover.c
```diff
@@ -969,21 +1074,22 @@ static void COVER_tryParameters(void *opaque) {
...
     }
-  }
-  /* Check total compressed size */
-  totalCompressedSize = COVER_checkTotalCompressedSize(parameters, ctx->samplesSizes,
-                                                       ctx->samples, ctx->offsets,
-                                                       ctx->nbTrainSamples, ctx->nbSamples,
-                                                       dict, dictBufferCapacity);
+    dictBufferCapacity = selection.dictSize;
+    totalCompressedSize = selection.totalCompressedSize;
+    memcpy(dict, selection.dictContent, dictBufferCapacity);
```
terrelln: We no longer need `dict`. If we move the declaration of `selection` to the top scope, we can refactor `COVER_best_finish()` to take a `COVER_dictSelection` instead of the three parameters (dict, totalCompressedSize, dictBufferCapacity).


```c
COVER_dictSelection_t selection = COVER_dictSelectionError(ERROR(GENERIC));
/* ... */
{
  const size_t tail = COVER_buildDictionary(/* ... */);
  selection = COVER_selectDict(/* ... */);
}
_cleanup:
  COVER_best_finish(data->best, parameters, selection); /* Copies the dict out of selection */
  COVER_dictSelectionFree(selection);

``

File: programs/zstdcli.c [hunk 34]
```diff
@@ -311,10 +311,15 @@ static unsigned parseCoverParameters(const char* stringPtr, ZDICT_cover_params_t
           params->splitPoint = (double)splitPercentage / 100.0;
           if (stringPtr[0]==',') { stringPtr++; continue; } else break;
         }
+        if (longCommandWArg(&stringPtr, "shrink=")) {
```
terrelln: Lets allow `shrink` to set a default regression amount of `1`.

```c
if (longCommandWArg(&stringPtr, "shrink")) {
  unsigned regression = kDefaultRegression;
  if (stringPtr[0] == '=') {
    stringPtr++;
    regression = readU32FromChar(&stringPtr);
  }
  /* ... */
```

lib/dictBuilder/zdict.h
```diff
@@ -94,7 +94,8 @@ typedef struct {
...
     unsigned nbThreads;          /* Number of threads : constraint: 0 < nbThreads : 1 means single-threaded : Only used for optimization : Ignored if ZSTD_MULTITHREAD is not defined */
     double splitPoint;           /* Percentage of samples used for training: Only used for optimization : the first nbSamples * splitPoint samples will be used to training, the last nbSamples * (1 - splitPoint) samples will be used for testing, 0 means default (1.0), 1.0 when all samples are used for both training and testing */
-    double shrinkDict;           /* Shrink dictionaries to select the smallest dictionary within the % specified by the auto-shrink-dict flag: 0 disables shrinking  */
+    unsigned shrinkDict;         /* Train dictionaries to shrink in size starting from the minimum size and selects the smallest dictionary within the max regression. 0 means no shrinking and 1 means shrinking  */
+    unsigned shrinkDictMaxRegression; /* Sets the value that determines how much worse a dictionary can be than the dictionary of max dictionary size */
```
terrelln: Can you describe what this value means? E.g. A smaller dict is allowed to be `shrinkDictMaxRegression` worse than the full sized dictionary.

programs/zstdcli.c
```diff
@@ -179,8 +179,8 @@ static int usage_advanced(const char* programName)
     DISPLAY( "\n");
     DISPLAY( "Dictionary builder : \n");
     DISPLAY( "--train ## : create a dictionary from a training set of files \n");
-    DISPLAY( "--train-cover[=k=#,d=#,steps=#,split=#,shrink=#] : use the cover algorithm with optional args\n");
-    DISPLAY( "--train-fastcover[=k=#,d=#,f=#,steps=#,split=#,accel=#,shrink=#] : use the fast cover algorithm with optional args\n");
+    DISPLAY( "--train-cover[=k=#,d=#,steps=#,split=#,shrink] : use the cover algorithm with optional args\n");
```
terrelln: `shrink[=#]` brackets means optional arguments

programs/zstdcli.c
```diff
@@ -377,7 +394,8 @@ static ZDICT_cover_params_t defaultCoverParams(void)
     params.d = 8;
     params.steps = 4;
     params.splitPoint = 1.0;
-    params.shrinkDict = 0.0;
+    params.shrinkDict = 0;
+    params.shrinkDictMaxRegression = 1;
```
terrelln: `kDefaultRegression`

programs/zstdcli.c
```diff
@@ -390,7 +408,8 @@ static ZDICT_fastCover_params_t defaultFastCoverParams(void)
     params.steps = 4;
     params.splitPoint = 0.75; /* different from default splitPoint of cover */
     params.accel = DEFAULT_ACCEL;
-    params.shrinkDict = 0.0;
+    params.shrinkDict = 0;
+    params.shrinkDictMaxRegression = 1;
```
terrelln: `kDefaultRegression`

lib/dictBuilder/cover.c
```diff
@@ -969,29 +1077,20 @@ static void COVER_tryParameters(void *opaque) {
...
-                                                       dict, dictBufferCapacity);
-
 _cleanup:
-  COVER_best_finish(data->best, totalCompressedSize, parameters, dict,
-                    dictBufferCapacity);
+  COVER_best_finish(data->best, parameters, selection);
   free(data);
   COVER_map_destroy(&activeDmers);
-  if (dict) {
-    free(dict);
```
terrelln: It looks like you're leaking `dict`.

lib/dictBuilder/fastcover.c
```diff
@@ -473,27 +473,23 @@ static void FASTCOVER_tryParameters(void *opaque)
...
-                                                       ctx->samples, ctx->offsets,
-                                                       ctx->nbTrainSamples, ctx->nbSamples,
-                                                       dict, dictBufferCapacity);
 _cleanup:
-  COVER_best_finish(data->best, totalCompressedSize, parameters, dict,
-                    dictBufferCapacity);
+  COVER_best_finish(data->best, parameters, selection);
   free(data);
   free(segmentFreqs);
-  free(dict);
```
terrelln: Looks like you're leaking `dict`.

lib/dictBuilder/cover.c
```diff
@@ -929,6 +934,111 @@ void COVER_best_finish(COVER_best_t *best, size_t compressedSize,
   }
 }
 
+COVER_dictSelection_t COVER_dictSelectionError(size_t error) {
+    COVER_dictSelection_t selection = { NULL, error, error };
```
terrelln: You need to set `dictSize == 0` here.

In `COVER_best_finish()` when `totalCompressedSize` is an error, but still `< -1` we will allocate space for the dictionary. We will try to allocate ~2^64 bytes, which will probably fail, but still we don't want to do that.

The dictionary size should always be the right number here, which is `0`.

Issue #1653, opened 2019-06-14
Decoding error on spruce genome


KirillKryukov:
Using command line zstd, experimenting with maximum compression options. All is fine until getting to a 13 GB file. Command lines:

zstd --ultra -22 --zstd=wlog=31,clog=30,hlog=30,slog=26 -c <pa.fna >pa.zstd
zstd --zstd=wlog=31,clog=30,hlog=30,slog=26 -dc pa.zstd >pa.fna2

Decompression prints this error message:

"pa.zstd : Decoding error (36) : Restored data doesn't match checksum"

The decompressed file has correct size, but corrupted contents.

zstd: Version 1.4.0, built from source using bundled Makefile
OS: Ubuntu 18.04.1 LTS (64-bit)
gcc: 7.4.0
CPU: dual Xeon E5-2643v3 (hyperthreading is off)
RAM: 128 GB DDR4-2133 ECC Registered

Test data size: 13,409,043,938 bytes
Test data: ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/900/067/695/GCA_900067695.1_Pabies01/GCA_900067695.1_Pabies01_genomic.fna.gz
(Un-gzipped and renamed into "pa.fna")

This experiment was for the Sequence Compression Benchmark ( http://kirr.dyndns.org/sequence-compression-benchmark/ ). zstd is working fine with default levels. With the above options it worked fine for smaller files, producing a bit stronger compression than just "--ultra -22".

I'd like to include a stronger compression mode in the benchmark, but only if it works on inputs of all sizes.

Let me know if you need any additional information or help with testing the fix.

Cyan4973:
We recently received a similar report associated with usage og `wlog=31`.
This is a pretty rare setting, which is likely badly tested.

We mostly need a reproducible test case to investigate.
The links proposed are unfortunately down currently (both http and ftp).

If we can get access to the source file, this will be very useful to reproduce the scenario, understand what's going wrong, and ship a fix.

KirillKryukov:
Link works for me. But OK, I'll mirror the file on my server. Please wait a bit.

KirillKryukov:
Please try this one: http://kirill.med.u-tokai.ac.jp/data/temp/Picea%20abies%20[genbank%20GCA_5F900067695.1%202016-11-09].fna.gz

(I also checked the original FTP link again - still working fine for me).

Cyan4973:
Thanks @KirillKryukov ,
[...]

PR #1659, opened 2019-06-22
Fix data corruption in niche use case

terrelln:
* Extract the overflow correction into a helper function.
* Load the dictionary `ZSTD_CHUNKSIZE_MAX = 512 MB` bytes at a time
  and overflow correct between each chunk.
* Add a test case that made overflow asserts fail, and now passes.
   I couldn't figure out how to make a fast test that causes data corruption.

Data corruption could happen when all these conditions are true:

* You are using multithreading mode
* Your overlap size is >= 512 MB (implies window size >= 512 MB)
* You are using a strategy >= ZSTD_btlazy
* You are compressing more than 4 GB

The problem is that when loading a large dictionary we don't do
overflow correction. We can only load 512 MB at a time, and may
need to do overflow correction before each chunk.

Fixes https://github.com/facebook/zstd/issues/1653.

### DIFF
--- a/tests/bigdict.c
+++ b/tests/bigdict.c
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2017-present, Yann Collet, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
+#include <assert.h>
+#include <stdio.h>
+#include <stddef.h>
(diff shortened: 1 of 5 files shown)


tests/bigdict.c
```diff
@@ -0,0 +1,128 @@
...
+}
+
+int main(int argc, const char** argv)
+{
+    ZSTD_CCtx* cctx;
+    ZSTD_DCtx* dctx;
+    const size_t dataSize = (size_t)1 << 30;
+    const size_t outSize = ZSTD_compressBound(dataSize);
+    const size_t bufferSize = (size_t)1 << 31;
+    char* buffer = (char*)malloc(bufferSize);
```
Cyan4973: nit: I guess all these `malloc()` should be `assert()` too ?

Issue #1636, opened 2019-06-06
[stack usage] Reduce stack usage of HUF_sort()


terrelln:
Maninder Singh found that `HUF_sort()` is using 256 bytes of stack space, and reduced it to 128 bytes in https://lkml.org/lkml/2019/5/10/89.

`HUF_buildCTable_wksp()` takes a `workspace` parameter that we can move the `rank` table into. We'll just need to increase the size to account for it.

felixhandte:
Closed via #1984.

PR #1660, opened 2019-06-22
Reduce stack usage of HUF_sort()

Kriyszig:
Issue in focus: #1636 
Changes in Linux Kernel: <https://lkml.org/lkml/2019/5/10/89>

Description:
* Using `U16` instead of `U32` cuts stack usage in half for `rankPos`

### DIFF
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -311,8 +311,8 @@ static U32 HUF_setMaxHeight(nodeElt* huffNode, U32 lastNonNull, U32 maxNbBits)
 
 
 typedef struct {
-    U32 base;
-    U32 current;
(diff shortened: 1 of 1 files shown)


Cyan4973:
The memory saving is clear, but what about performance ?
Cpu nowadays tend to prefer reading 32-bit data, meaning 16-bit ones tend to incur a small performance penalty.
It would be better if this impact could be measured.
(note: I believe @terrelln made such a measurement recently ?)

terrelln:
I'd prefer to have `HUF_buildCTable_wksp()` take a workspace and use that for the `rank` table. This will give us the full 256 bytes of stack savings, and is likely to perform better. But, we'll still need to measure the performance.

PR #1664, opened 2019-06-27
decodecorpus 

ephiepark:
Decoder was enforcing a constraint that is more strict than the spec. Sequence section for a compressed block can be less than the constraint that was enforced by decoder. When a block with rle mode for all (literal length, match length, and offset) is followed by another block with repeat mode for all (literal length, match length, and offset), the second block's sequence section can be 2 bytes (1 byte for number of sequences and 1 byte for compression mode for each field). 

This pull request includes 2 changes. 
1. fixing the decoder code to remove the constraint. 
2. Updating decodecorpus code to generate repeat after rle mode. 

Run `./decodecorpus -ptestfiles -otestfiles -n1 -s1236947365 -v` to generate an input that has an issue before the fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,10 +24,9 @@ jobs:
     # TODO: Create a small custom docker image with all the dependencies we need
     #       preinstalled to reduce installation time.
     docker:
-      - image: circleci/buildpack-deps:bionic
+      - image: fbopensource/zstd-circleci-primary:0.0.1
     steps:
       - checkout
-      - *install-dependencies
       - run:
           name: Test
           command: |
@@ -41,10 +40,9 @@ jobs:
   # the second half of the jobs are in this test
   short-tests-1:
     docker:
-      - image: circleci/buildpack-deps:bionic
+      - image: fbopensource/zstd-circleci-primary:0.0.1
     steps:
       - checkout
-      - *install-dependencies
       - run:
           name: Test
           command: |
@@ -61,12 +59,11 @@ jobs:
   # tagged release.
   publish-github-release:
     docker:
-      - image: circleci/buildpack-deps:bionic
+      - image: fbopensource/zstd-circleci-primary:0.0.1
     environment:
       CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
     steps:
       - checkout
-      - *install-dependencies
       - run:
           name: Publish
           command: |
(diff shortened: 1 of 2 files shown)


ephiepark:
@terrelln
It takes a few minutes. I would say less than 10 minutes. They are cases where sequences section is 4 bytes, which does cause issues. But I don't think I have ever seen sequences section being <= 3 so far. I added a unittest to cover that though.

Cyan4973:
A targeted test case for an extreme outlier,
and a generic test case generator which reaches the target issue in a reasonable time frame,
It's the right mix I believe.

tests/zstreamtest.c
```diff
@@ -1184,6 +1184,68 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+        inBuff.pos = 0;
+        outBuff.dst = decodedBuffer;
+        outBuff.size = CNBufferSize;
+        outBuff.pos = 0;
+
+        while (inBuff.pos < inBuff.size) {
+            CHECK_Z( ZSTD_decompressStream(zds, &outBuff, &inBuff) );
+        }
+
+        {   XXH64_state_t xxhStateIn, xxhStateOut;
```
Cyan4973: nit: 
if all you want to do is compare `decompressed` and `decodedBuffer ` content and ensure they are equal, you could simply `memcmp()` them.

tests/zstreamtest.c
```diff
@@ -1184,6 +1184,68 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+        CHECK_Z( ZSTD_initDStream(zds) );
+        inBuff.src = compressed;
+        inBuff.size = compressedSize;
+        inBuff.pos = 0;
+        outBuff.dst = decodedBuffer;
+        outBuff.size = CNBufferSize;
+        outBuff.pos = 0;
+
+        while (inBuff.pos < inBuff.size) {
+            CHECK_Z( ZSTD_decompressStream(zds, &outBuff, &inBuff) );
```
Cyan4973: nit: it's slightly stronger than `CHECK_Z()` :
presuming the frame is complete (is it ?), you want a return value `0`, to indicate that the decoder has reached the end of frame. You also want to ensure it has consumed the entire input (`inBuff.pos == inBuff.compressedSize`).

Cyan4973:
Looks good !
Just some minor comments on the "extreme" test case.

tests/zstreamtest.c
ephiepark: Yes, it's a complete frame.

PR #1671, opened 2019-07-03
Adding targetCBlockSize param

ephiepark:
This pull request only adds targetCBlockSize as an advanced parameter for ZSTD compression. At the moment changing its value won't have any impact on compression. The implementation of super block compression will follow as a separate PR.


Cyan4973:
Looks fine,

there is a just minor implicit cast warning to fix, nothing alarming.

PR #1674, opened 2019-07-05
Fix #1673 Unnecessary records in `build/.gitignore`

### LINKED ISSUE #1673: Unnecessary records in `build/.gitignore`
Hello.

At the end of the file `build/.gitignore` recently added two new entries:

```
build
lib
```

Which are too general and conflict, for example, with an existing `build/cmake/lib` directory. In addition, the `build/.gitignore` file already contains the entry` cmake/build/`.

I put the zstd source codes in the Node.js package. Npm parses the .gitignore files and skips the `build/cmake/lib` directory.

I propose to delete these records. Or, if these directories are created somewhere, limit the effect of these entries to only one level, using a slash:

```
/build/
/lib/
```

dr15x86:


Cyan4973:
The issue _seems_ fixed in latest `dev` branch update,
although it's bundled with a number of other `.gitignore` rule changes.

I recommend having a look at latest `dev` branch update, to ensure it properly covers your use case.

PR #1675, opened 2019-07-06
Factor out the logic to build sequences

ephiepark:
Factor out the logic to build sequence store. ZSTD_buildSeqStore() will be used by ZSTD_compress_superBlock in the following PR.


Cyan4973:
Looks good to me !

PR #1677, opened 2019-07-09
fix gitignore errors

LeeYoung624:
Some files are ignored by some .gitignore rules. These files should be exclude from gitignore

Ignored files:
`build/VS2008/zstd/`
`build/VS2010/zstd/`
`build/cmake/lib/`
`build/meson/lib/`
`contrib/VS2005/zstd/`
`examples/dictionary_compression.c`
`examples/dictionary_decompression.c`
`mygittestproject.zip`
`programs/windres/generate_res.bat`
`tests/fuzz/dictionary_decompress.c`
`tests/fuzz/dictionary_round_trip.c`
`tests/gzip/hufts-segv.gz`
`zlibWrapper/examples/example.c`
`zlibWrapper/examples/fitblk.c`
`zlibWrapper/examples/zwrapbench.c`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@
 
 # Executables
 zstd
+!build/VS2008/zstd
+!build/VS2010/zstd
+!contrib/VS2005/zstd
 zstdmt
 *.exe
 *.out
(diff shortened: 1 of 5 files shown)


Cyan4973:
Agreed, some of these "ignore" rules are over-reaching, and should be changed.
As a quickfix, I'll accept this list of exception, which achieve the same outcome, provided the list of files remains constant.

PR #1678, opened 2019-07-09
[ldm] Fix bug in overflow correction with large job size

### LINKED ISSUE #1672: Incompatibility between --long=31 and -22 --ultra?
I tried to compress a file with `zstd -22 --ultra --long=31` and then decompress it with `zstd -d --long=31` and got the following error:

```
tmp.tar.zst          : 6031 MB...     tmp.tar.zst : Decoding error (36) : Corrupted block detected
```

Is this a known issue? Just checking before I try to create a repro and/or debug this since it takes about 2 hours to compress the file (it's ~7GB), so it may take a while to reproduce the issue. Unfortunately, the file I'm compressing isn't something I can pass around, so I can't just send it to you.

I've tried this with the same file on two different platforms (different hardware and different Linux kernel versions) and have run into the same things both times. In both cases, the code I had was built from c0eb6c9c81e2dd31571267b37c12cdaa92b7ffd7, by running with `make` with no extra options specified.

terrelln:
`src` is the wrong pointer here. When the job size is 2GB the second job does overflow correction, but `src - base == 2 GB` for every chunk. When we do overflow correction we get `current & cycleLog + maxDist == 2 GB & 2^26 + 2 GB == 2 GB`, meaning we effectively never do overflow correction that job. Then the next job we overflow, and things go terribly wrong.

Instead, we need to pass the beginning of the current chunk, which is only 1 MB large and is well under the maximum chunk size of 512 MB.

There are asserts that failed before this change, and don't fail after, but I accidentally defined the wrong macro during the build for my initial tests.

The test I added to `playTests.sh` has assertion failures and produces corrupted data before this change. The test I added to `zstreamtest.c` doesn't fail before the change, but exercises the max job size and window log which is useful.

Fixes #1672.

### DIFF
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -447,7 +447,7 @@ size_t ZSTD_ldm_generateSequences(
         if (ZSTD_window_needOverflowCorrection(ldmState->window, chunkEnd)) {
             U32 const ldmHSize = 1U << params->hashLog;
             U32 const correction = ZSTD_window_correctOverflow(
-                &ldmState->window, /* cycleLog */ 0, maxDist, src);
+                &ldmState->window, /* cycleLog */ 0, maxDist, chunkStart);
             ZSTD_ldm_reduceTable(ldmState->hashTable, ldmHSize, correction);
         }
         /* 2. We enforce the maximum offset allowed.
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1153,12 +1153,16 @@ size_t ZSTDMT_toFlushNow(ZSTDMT_CCtx* mtctx)
 
 static unsigned ZSTDMT_computeTargetJobLog(ZSTD_CCtx_params const params)
 {
-    if (params.ldmParams.enableLdm)
+    unsigned jobLog;
+    if (params.ldmParams.enableLdm) {
         /* In Long Range Mode, the windowLog is typically oversized.
          * In which case, it's preferable to determine the jobSize
          * based on chainLog instead. */
-        return MAX(21, params.cParams.chainLog + 4);
-    return MAX(20, params.cParams.windowLog + 2);
+        jobLog = MAX(21, params.cParams.chainLog + 4);
+    } else {
+        jobLog = MAX(20, params.cParams.windowLog + 2);
+    }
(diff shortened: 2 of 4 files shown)


tests/zstreamtest.c
```diff
@@ -1028,6 +1028,55 @@ static int basicUnitTests(U32 seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    /* Large source and large job size multithreading compression test */
```
Cyan4973: how long does this test last ?

terrelln: Less than 10 seconds

Cyan4973: It's very reasonable time duration to test such condition,

but for _this section of the test system_ (`basicUnitTests()` in `zstreamtest`), it might be a bit too long :
we ideally want all tests in `basicUnitTests()` to be a fraction of a second each,
so that the sum of all these tests remain in the "acceptable" range to be run as a preamble test.

Any chance this could become an equivalent test in the `--test-large-data` of `playTests.sh` ?
That's generally where we play all long tests with large files, making it possible to disable them selectively.

One downside is that `playTests.sh` uses the CLI, rather than the API.
On first look, it seems it could nonetheless be compatible with below test ?

terrelln: We can't control the job size on the CLI except implicitly. I also have a test there that exposes the bug.

I think we should leave the test there for now, because we don't currently have a better place.

But, I think we should invest some time in switching to gtest for our unit tests. It is the standard for C/C++ tests. There are a couple of benefits:

* All the tests can start with a clean environment and share setup/teardown code. We can reset or recreate contexts between tests, populate input buffers, etc.
* gtest has great helpers for printing why a test failed.
* We can do fancy things like run the same test on multiple types of data (if we want).
* When one test fails it still runs all the other tests
* It is the standard, so on-boarding engineers and open source contributors will be easier.

The downside is that we have a C++/cmake dependency for testing (gtest can be a submodule). I think that is reasonable since it is only needed for tests.

Cyan4973: Agreed !

terrelln: I ended up removing the test because it is failing in 32-bit mode with ASAN.

When we introduce gtest we'll have to skip the "expensive" tests in 32-bit ASAN mode like we do with playTests.sh.

PR #1679, opened 2019-07-11
Restructure the source files

ephiepark:
This pull request reorganizes codes into different files.
This changes have the following side effect:

- functions like ZSTD_noCompressLiterals, ZSTD_compressRleLiteralsBlock, ZSTD_selectEncodingType, ZSTD_buildCTable, ZSTD_encodeSequences are no longer static functions.


terrelln:
This is great! `zstd_compress.c` is gigantic

Cyan4973:
The change itself looks good to me.

Now, the main issue is around build scripts.
Some of them will have to be updated, because they can't just "grab the directory content", as `Makefile` does (or are not designed this way, and maybe they should be changed).

This includes :
 - cmake
- Visual Studio
- meson

For the PR to pass, all build systems must be updated.
This is a major source of slow down.

At this stage, a quickfix is good enough.
though if you can find a way to make some of these build systems a bit more "generic", 
this would be a great gain too.

ephiepark:
@Cyan4973  Thanks for the comment! I was confused about what was causing the issue. Let me look into them.

ephiepark:
This pull request had one failure on Travis CI on (Trusty (gcc-8 + ASan + UBSan + Fuzz Test)). This cause of the failure is independent from this pull request. It's caused because gcc8 is not properly suppressing signed-integer-overflow (pointer-overflow) for ubsan. I have created an issue for it (https://github.com/facebook/zstd/issues/1686).

build/cmake/lib/CMakeLists.txt
```diff
@@ -20,62 +20,32 @@ endif()
 # Define library directory, where sources and header files are located
 include_directories(${LIBRARY_DIR} ${LIBRARY_DIR}/common)
 
+file(GLOB CommonSources ${LIBRARY_DIR}/common/*.c)
+file(GLOB CompressSources ${LIBRARY_DIR}/compress/*.c)
+file(GLOB DecompressSources ${LIBRARY_DIR}/decompress/*.c)
+file(GLOB DictBuilderSources ${LIBRARY_DIR}/dictBuilder/*.c)
+file(GLOB DeprecatedSources ${LIBRARY_DIR}/deprecated/*.c)
```
Cyan4973: Thanks @ephiepark , this new `GLOB` is much better for file layout flexibility

Cyan4973:
Thanks @ephiepark ,
this patch is good to go.

As we are close to a new release, we'll wait a bit for the "freeze" snapshot, and merge this patch right after that.

Cyan4973:
Release completed,
we can merge.

PR #1685, opened 2019-07-15
Add Check if Block Size Exceeds Maximum

vivekmig:
Adding check if block size exceeds maximum.

Tested by creating a frame with a fake frame with raw block of size approximately 500 kb, and verifying the output:
Screen Shot 2019-07-15 at 12 16 24 PM

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -909,7 +909,7 @@ size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t dstCapacity, c
         {   blockProperties_t bp;
             size_t const cBlockSize = ZSTD_getcBlockSize(src, ZSTD_blockHeaderSize, &bp);
             if (ZSTD_isError(cBlockSize)) return cBlockSize;
-            RETURN_ERROR_IF(cBlockSize > ZSTD_BLOCKSIZE_MAX, corruption_detected, "Block Size Exceeds Maximum");
+            RETURN_ERROR_IF(cBlockSize > dctx->fParams.blockSizeMax, corruption_detected, "Block Size Exceeds Maximum");
             dctx->expected = cBlockSize;
             dctx->bType = bp.blockType;
             dctx->rleSize = bp.origSize;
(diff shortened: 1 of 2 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -909,6 +909,7 @@ size_t ZSTD_decompressContinue(ZSTD_DCtx* dctx, void* dst, size_t dstCapacity, c
         {   blockProperties_t bp;
             size_t const cBlockSize = ZSTD_getcBlockSize(src, ZSTD_blockHeaderSize, &bp);
             if (ZSTD_isError(cBlockSize)) return cBlockSize;
+            RETURN_ERROR_IF(cBlockSize > ZSTD_BLOCKSIZE_MAX, corruption_detected, "Block Size Exceeds Maximum");
```
Cyan4973: Yes, this is a first good place for this check.
It controls the compressed size, and no block shall have its compressed size larger than maximum block size.

However : 
- No block shall have its _decompressed_ size larger than maximum block size. This is more complex, and must be controlled elsewhere.
- `ZSTD_BLOCKSIZE_MAX` is the "maximum maximum block size". But the [specifications](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#blocks) makes it clear that the maximum block size can be smaller than that, depending on frame header instructions. It actually depends on the window size. This value should be computed once, at header decoding stage, stored into `dctx`, and then used for each check.

vivekmig: Thanks so much, this was really helpful! I think the maximum block size (min of window size and block size) seemed to already be stored in dctx->fParams.blockSizeMax, so I switched to using that. I've also added a check for the decompressed block size by checking rSize returned after decompressing a block.

Cyan4973:
Thanks @vivekmig !
The change looks good to me !

Now, it seems it generates errors in the `decodecorpus` test.

It _might_ be because `decodecorpus` is generating incorrect frames.
This seems plausible, as this block size property was never enforced so far.
But this is pure speculation. One would need an investigation to understand the problem.

vivekmig:
Updated decodecorpus test to not create an RLE block if frame content size is 0, this eliminates the possibility that an RLE block exceeds the max block size.

Cyan4973:
Great work @vivekmig !

We are in the middle of a new release, so we'll wait a little bit for the actual merge.
But this patch is validated, and it will be merged soon after.

Cyan4973:
Release completed, we can merge

PR #1688, opened 2019-07-16
Implement superBlock compression

ephiepark:
Superblock compression compresses a super block into multiple smaller sub blocks (the size of sub blocks should be close to targetCBlockSize param). First sub block will have entropy information like huffman description and fse tables and subsequent sub blocks will leverage that data by compressing with repeat mode. If it fails to compress with Superblock compression, the super block will not be compressed and will be written as a raw block to output buffer.


lib/compress/zstd_compress_superblock.c
```diff
@@ -0,0 +1,588 @@
...
+                                        LL_defaultNorm, LL_defaultNormLog,
+                                        ZSTD_defaultAllowed, strategy);
+        assert(set_basic < set_compressed && set_rle < set_compressed);
+        assert(!(LLtype < set_compressed && nextEntropy->litlength_repeatMode != FSE_repeat_none)); /* We don't copy tables */
+        {   size_t const countSize = ZSTD_buildCTable(op, oend - op, CTable_LitLength, LLFSELog, (symbolEncodingType_e)LLtype,
+                                                    countWksp, max, llCodeTable, nbSeq, LL_defaultNorm, LL_defaultNormLog, MaxLL,
+                                                    prevEntropy->litlengthCTable, sizeof(prevEntropy->litlengthCTable),
+                                                    cTableWksp, cTableWkspSize);
+            FORWARD_IF_ERROR(countSize);
+            // if (LLtype == set_compressed)
```
ephiepark: I need to handle this for the bug in past version of decoder https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L2661 . I am thinking of storing lastCountSize instead of storing pointer to fseMetadata. I just have a weird feeling when I think of passing pointer around across distant functions for pointer arithmetic. Does this approach make sense?

lib/compress/zstd_compress_superblock.c
```diff
@@ -0,0 +1,588 @@
...
+    /* Build Huffman Tree */
+    memset(nextHuf->CTable, 0, sizeof(nextHuf->CTable));
+    huffLog = HUF_optimalTableLog(huffLog, srcSize, maxSymbolValue);
+    {   size_t const maxBits = HUF_buildCTable_wksp((HUF_CElt*)nextHuf->CTable, countWksp,
+                                                    maxSymbolValue, huffLog,
+                                                    nodeWksp, nodeWkspSize);
+        FORWARD_IF_ERROR(maxBits);
+        huffLog = (U32)maxBits;
+        hufMetadata->hType = set_compressed;
+        return HUF_writeCTable(hufMetadata->hufDesBuffer, sizeof(hufMetadata->hufDesBuffer),
```
ephiepark: Copying over @Cyan4973 's comment from previous PR. 
===================================
At this point, you both write the literals' section header, and decide to go for huffman compression of literals.
All "obvious" corner cases have already been filtered out, but it's still possible that the sum of header size + compressed literals is > uncompressed literals, especially when nb of literals is small.

In the "normal" implementation, this is checked after compression, on compressed data itself.
But here, compression isn't happening, so the check has simply been suppressed.

Suggestion : ensure that compression of literals will at least produce some gains.
It's possible to quickly evaluate the cost of compressed literals just from the histogram and the CTable : just sum the bits. I suspect we already have a function for this evaluation. If not, it should be easy to create.

Stretch goal : consider the cost of headers. I believe uncompressed literals have a small advantage regarding header size. Compressed blocks must over-match that. Don't try to be precise, a loose or slightly generous estimation will do fine enough.

lib/compress/zstd_compress.c
```diff
@@ -2274,6 +2285,39 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
     return cSize;
 }
 
+static size_t ZSTD_compressSuperBlock(ZSTD_CCtx* zc,
```
ephiepark: @terrelln @Cyan4973  naming wise, I am thinking of changing this function to be ZSTD_compress_targetCBlockSize() and changing ZSTD_compressSuperBlock_internal() to be ZSTD_compressSuperBlock(). Does that sound okay?

lib/compress/zstd_compress_superblock.h
```diff
@@ -0,0 +1,31 @@
...
+/*-*************************************
+*  Target Compressed Block Size
+***************************************/
+
+/* ZSTD_compressSuperBlock_internal() :
+ * Used to compress a super block when targetCBlockSize is being used.
+ * The given block will be compressed into multiple sub blocks that are around targetCBlockSize. */
+size_t ZSTD_compressSuperBlock_internal(ZSTD_CCtx* zc,
+                               void* dst, size_t dstCapacity,
+                               U32 lastBlock);
```
Cyan4973: nit: very minor : 
`U32` requires `mem.h`.
Rather than adding `mem.h` as a header dependency just for that, 
it might be better to use `unsigned` type instead, which is standard.

lib/compress/zstd_compress.c
Cyan4973: Yes, I agree, it feels better.

In term of API design, it's important to evaluate "how does the API look like in the `*.h`" ?
The naming should make sense, the scope should make sense, etc.

And I believe the new proposed naming fits better in `zstd_compress_superblock.h`.

lib/compress/zstd_compress_superblock.c
Cyan4973: While I do remember a decoder limitation around `set_repeat` following a `set_rle`, 
I'm not sure to follow what's the issue addressed here with `set_compressed`...

Cyan4973: OK, reading the decoder side, I understand this is a completely different issue (in `FSE_readNCount ()` than the one I had in mind.

lib/compress/zstd_compress.c
```diff
@@ -2274,6 +2285,39 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
...
+    }
+    /* We check that dictionaries have offset codes available for the first
+     * block. After the first block, the offcode table might not have large
+     * enough codes to represent the offsets in the data.
+     */
+    if (zc->blockState.prevCBlock->entropy.fse.offcode_repeatMode == FSE_repeat_valid)
+        zc->blockState.prevCBlock->entropy.fse.offcode_repeatMode = FSE_repeat_check;
+
+    if (cSize == 0) {
+      cSize = ZSTD_noCompressBlock(dst, dstCapacity, src, srcSize, lastBlock);
```
terrelln: Do we want to emit many small uncompressed blocks instead of one large block?

lib/compress/zstd_compress_superblock.c
```diff
@@ -0,0 +1,611 @@
...
+    litSize = 0;
+    seqCount = 0;
+    while (sp + seqCount < send) {
+        // TODO this is crude estimate for now...
+        // Ask Yann, Nick for feedback.
+        const seqDef* const sequence = sp + seqCount;
+        const U32 lastSequence = sequence+1 == send;
+        litSize = (sequence == send) ? (size_t)(lend-lp) : litSize + sequence->litLength;
+        seqCount++;
+        cBlockSizeEstimate = sizeof(seqDef) * seqCount + litSize;
```
terrelln: I haven't kept up too much with the discussion, but why do we want to break based on compressed size not uncompressed size? The latter we can easily compute.

Cyan4973: This is more in line with a transmission scenario : break data flow into small packets of roughly predictable size. Now, every ~`N` bytes, we have one block to decode, containing a variable number of bytes.

In contrast, splitting on uncompressed size, we end up with some very tiny blocks (big match) and others which are much larger (poorly compressible content) . It may also cut long matches in the middle.

lib/compress/zstd_compress.c
Cyan4973: Yes, it seems to match the scenario better :
we want the decoder to regularly decode some output, every packet or so.

When the block is not compressed, it's in theory possible to "stream intra-block",
thus emitting some output without waiting for the full block.
In practice, this ability is not present yet in the reference decoder, 
which reads the block header, determine the block size, and then wait for the full block.

Emitting one big uncompressed block kind of rely on such capability to be developed one day. 
But it doesn't address older versions.

lib/compress/zstd_compress_superblock.c
ephiepark: What would be the better way to estimate the compressed data size?

ephiepark:
Now, I am running into cases where libfuzzer test fails due to compressed data being larger than the raw data. 

I think the scenario goes like the following. Roundtrip test allocates an output buffer assuming some worst case scenario. However, because of the small targetCBlock size and header overhead, the compressed data is larger than assumed worst case scenario and compression crashes with "Destination buffer is too small". Should I update Roundtrip tests to allocate more generously?

ephiepark:
Another option I can do is that if `ZSTD_compressBlock_targetCBlockSize` fails for whatever reason, I can call `ZSTD_compressBlock_internal` to compress normally instead of forwarding error from `ZSTD_compressBlock_targetCBlockSize`. I think this is more reasonable than changing `ZSTD_compressBound` function to increase the buffer size.

Cyan4973:
We can document `ZSTD_compressBound()`, to make it clearer its evaluation is only compatible with the "normal" full-block strategy.

Even today, it's possible to blow up `ZSTD_compressBound()` budget in streaming mode, with an abundant use `ZSTD_flushStream()`: it's possible to create a lot tiny blocks, inflating the proportion of block headers.

The situation of your superBlock patch is basically equivalent. So documentation should underline this limitation for both scenarios.

> Now, I am running into cases where libfuzzer test fails due to compressed data being larger than the raw data.

I'm not sure which test fails, 
but note that we _never guarantee_ that compressed data will _always be smaller_ than original data. Trivially, when "compressing" an encrypted channel, we end up with the raw encrypted data, and all the headers around, so it's necessarily larger.

ephiepark:
@Cyan4973  Make sense, I can add comments to `ZSTD_compressBound()`. 

The round trip tests are failing not because there is something wrong in encoding / decoding logic but because it allocates buffer memory using `ZSTD_compressBound()` and the compressed data turns out to be larger than the allocated memory. 

I think this can be mitigated by ... 
1. Make simple round trip and dictionary round trip test to allocate larger amount of buffer. (streaming round trip is allocating `ZSTD_compressBound()` X 2 . So, I think I can do the same for simple and dictionary round trip.) 
2. Fall back to compress normally when superblock compression fails. 
3. Do both 1 and 2. 

I feel like I need to do both 1 and 2, but I want to hear your opinion.

Cyan4973:
I believe using the same method as streaming round trip (1) is appropriate,
since it's about compensating the same limitation of `ZSTD_compressBound()`.

I'm not sure if (2) is useful. Then it would become a "normal compression test", but that's a scenario which is already pretty well tested, so it probably doesn't add value.

ephiepark:
@Cyan4973 
For streaming mode, it seems like it's almost required that a block gets compressed within `ZSTD_compressBound()` limit. The reason why I think this way is the following:
1. In streaming mode, out buffer can be smaller than compressed data size of in buffer. 
2. This is possible because when out buffer is smaller, it will write the output to `outBuff` in compress context instead of provided out buffer. 
3. The size of `outBuff` is determined using `ZSTD_compressBound` on block size from param. This is safe since a single block compression shouldn't go over `ZSTD_compressBound`. 
4. However, single block compress can go over `ZSTD_compressBound` when it is broken down into multiple subBlocks. 

So, it seems like I need to guarantee that a compressed superblock doesn't exceed `ZSTD_compressBound` for the superblock to make it work properly with streaming mode. Does this understanding make sense? 

I think possible mitigations are:
1. When a compressed superblock is larger than the `ZSTD_compressBound`, fall back to normal compression. I think this is okay since targetCBlockSize doesn't give guarantees on block size. It's best effort. 
2. Specify that the caller always has to make sure that out buffer is big enough if s/he wants to use streaming mode with targetCBlockSize option. (In this case, we will need to modify streaming_round_trip test to make sure to always pass big enough out buffer when testing with targetCBlockSize option.)

Cyan4973:
Indeed.

`ZSTD_compressBound()` is more about the maximum compressed size of a _frame_.
Using it for a _block_ is quite generous, since it accounts for header costs which will not be present.
But that's a safe way to size a buffer designed to receive any compressed block.

However, this was designed with a _single_ block in mind.
Therefore, trying to fit a lot of small blocks into this budget may be able to overcome this budget.

I guess the architecture doesn't allow to output just a few blocks, up to the buffer limit, flush, and then resume generating next blocks on next invocation.
So this is more about a work-around to compensate the shortcomings of current design.

I agree that (1) is allowed (no guarantee on compressed block size yet), and is likely the simplest solution.

hsiangkao:
Hi all,
I am interested this feature as well, I am not quite sure, maybe it can be used for fixed-sized output compression such as EROFS?

one more about
https://github.com/facebook/zstd/commit/90077016701522259a0969f622d7d14bf1289c1f

ZSTD_TARGETCBLOCKSIZE_MAX  ZSTD_BLOCKSIZE_MAX  --> is it possible to change this value by users? because I want to implement in-place decompression for super block with small targetCBlockSize (such as 4k) as well...

Thanks,
Gao Xiang

Cyan4973:
> I am interested this feature as well, I am not quite sure, maybe it can be used for fixed-sized output compression such as EROFS?

We can look at this scenario in a second stage.
Right now, `targetCBlockSize` is only a "target" which is roughly approximated.
I suspect you will look for some "maximum guaranteed" CBlockSize instead, so that it always fit into a sector. This is probably an accessible objective, though not one that was part of this PR.

> ZSTD_TARGETCBLOCKSIZE_MAX ZSTD_BLOCKSIZE_MAX --> is it possible to change this value by users? 

I believe there is no need to.
These constants are absolute limits.
The _actual_ limit is a runtime value, that can be selected using `ZSTD_c_targetCBlockSize` parameter. As long as the selected value is between the bounds, all is fine. And 4k is definitely between bounds.

Cyan4973:
@ephiepark : it seems we are getting there.
The PR looks good to me, although `CircleCI` seems to complain about the last test failure.
Unsure if there is an actual problem, or if the test itself is failing.

hsiangkao:
> > I am interested this feature as well, I am not quite sure, maybe it can be used for fixed-sized output compression such as EROFS?
> 
> We can look at this scenario in a second stage.
> Right now, `targetCBlockSize` is only a "target" which is roughly approximated.
> I suspect you will look for some "maximum guaranteed" CBlockSize instead, so that it always fit into a sector. This is probably an accessible objective, though not one that was part of this PR.

I think I can try to recompress this again around the previous result if actual compressed size is larger than targetCBlockSize. It seems better than directly bisect blindly. I will play with it after this gets merged :)

> 
> > ZSTD_TARGETCBLOCKSIZE_MAX ZSTD_BLOCKSIZE_MAX --> is it possible to change this value by users?
> 
> I believe there is no need to.
> These constants are absolute limits.
> The _actual_ limit is a runtime value, that can be selected using `ZSTD_c_targetCBlockSize` parameter. As long as the selected value is between the bounds, all is fine. And 4k is definitely between bounds.

OK, I will look into how to do in-place decompression by some safe margin. I thought maybe it is a per-subblock basic margin since I have no idea it can do in-place decompression like lz4 within a single block...

Thanks,
Gao Xiang

lib/compress/zstd_compress_superblock.c
Cyan4973: CBlock Size estimation :
I see mostly 3 parts : 
1) Literals : starting from the huffman distribution presumed already calculated, the cost of each literal is known, in nb of bits. Just add these bits together. This requires reading the actual byte values from the literals buffer.
2) Sequences : same recipe, merely a bit more complex : the proportion of each symbol is presumed known, and from this information, it's possible to derive the cost of each symbol. There are several functions dedicated to this objective, at the beginning of`zstd_compress_sequences.c` .  The cost is supposed to be fractional (aka more precise that nb of bits). I think it uses a fixed point format.
3) Headers : Each block will cost a few additional bytes due to the format. Block size itself, but also size of literals section, nb of sequences, symbol type, etc. While this cost is not strictly speaking a "fixed cost", it's relatively stable, and can probably be cheaply estimated with a constant.

ephiepark: Thanks for this! @Cyan4973 I have a question on #2. I am not sure if my understanding is correct but the functions to estimate cost of symbol only accounts for Code and does not account for additional bits for bitstream. Does those functions accounts for both Code and additional bits?

Cyan4973: You are totally right @ephiepark , the functions only estimate cost of the symbol. Additional bits must be added on top of that.

Cyan4973:
Thanks for completing this great feature @ephiepark !
Let's merge it into its feature branch.
Then, we'll wait for `v1.4.3` release (which is a quick bugfix release) before merging the feature into `dev`.

PR #1690, opened 2019-07-18
fix compiling errors with clang-8

piguin:
Compiling with clang-8 fails with the following errors:

largeNbDicts.c:562:37: error: implicit conversion turns floating-point
number into integer: 'const double' to 'U64' (aka 'unsigned long')
[-Werror,-Wfloat-conversion]
        U64 const dTime_ns = result.nanoSecPerRun;
                  ~~~~~~~~   ~~~~~~~^~~~~~~~~~~~~

zstdcli.c:300:5: error: '@return' command used in a comment that is
not attached to a function or method declaration
[-Werror,-Wdocumentation]
 * @return 1 means that cover parameters were correct
   ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

zstdcli.c:301:5: error: '@return' command used in a comment that is
not attached to a function or method declaration
[-Werror,-Wdocumentation]
 * @return 0 in case of malformed parameters
   ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -559,7 +559,7 @@ static int benchMem(slice_collection_t dstBlocks,
         CONTROL(BMK_isSuccessful_runOutcome(outcome));
 
         BMK_runTime_t const result = BMK_extract_runTime(outcome);
-        U64 const dTime_ns = result.nanoSecPerRun;
+        double const dTime_ns = result.nanoSecPerRun;
         double const dTime_sec = (double)dTime_ns / 1000000000;
         size_t const srcSize = result.sumOfReturn;
         double const dSpeed_MBps = (double)srcSize / dTime_sec / (1 MB);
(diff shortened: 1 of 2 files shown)


contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -559,7 +559,7 @@ static int benchMem(slice_collection_t dstBlocks,
         CONTROL(BMK_isSuccessful_runOutcome(outcome));
 
         BMK_runTime_t const result = BMK_extract_runTime(outcome);
-        U64 const dTime_ns = result.nanoSecPerRun;
+        U64 const dTime_ns = (U64)result.nanoSecPerRun;
```
Cyan4973: I believe it would actually be better to make `dTime_ns` a `double`,
this would avoid a useless double-conversion.

I think this issue happened because `nanoSecPerRun` changed from a `U64` to a `double` type 
well after creation of `largeNbDicts.c` test program.

piguin: thanks, let me update it

piguin:
@Cyan4973 anything else I could on the PR, could you take another look? :)

Cyan4973:
I think that's fine, thanks @piguin !

PR #1698, opened 2019-07-24
Bump Library Version Number to 1.4.2

felixhandte:
In preparation for a new release, rev the version number and update the CHANGELOG.


Cyan4973:
hint : after a change of version in `zstd.h`, run `make all` locally,
it will update a bunch of files that depend on the version number,
and these changes will be part of the PR.

PR #1706, opened 2019-07-29
add NULL pointer check in util.c

LeeYoung624:
I believe the original code was meant to deal with case that `/proc/cpuinfo` has come across a error: the formatting of `/proc/cpuinfo` was broken. In that way, the `sep` could also be `NULL` because it's the return of `strchr`.
I think adding a `NULL` check is better here.


Cyan4973:
Good point @LeeYoung624 !

PR #1708, opened 2019-07-31
zstd: Don't use utime on Linux

neheb:
utime is deprecated by POSIX 2008 and optionally not available with
uClibc-ng.

Got rid of a few useless headers in timefn.h.


neheb:
Realized that utimes creates a year 2038 problem for 32-bit platforms.

utimensat uses a different structure that specifies time_t instead of long. Switched.

felixhandte:
The effect of this is to use `utimensat()` for all non-win32 platforms (not just linux). I don't have confidence that all of those platforms define that function. Can you wrap the changes you've made in a feature macro test? E.g.,

```c
#if PLATFORM_POSIX_VERSION >= 200809L
#  include <fcntl.h>      /* AT_FDCWD */
#  include <sys/stat.h>   /* utimensat */
#endif
```

etc. If the platform doesn't support posix 2008, can you retain the old `utime()` call?

neheb:
@felixhandte that example is wrong since the commit sets the version to 2008 if it is not set. It has to be set to 2008, otherwise utimensat would not be defined (an issue I and the buildbots were running into).

Are non-POSIX 2008 platforms still being used?

felixhandte:
This commit sets `_POSIX_C_SOURCE` to `200809L`. That will cause the platform's headers (`unistd.h`, via `features.h` for my glibc install) to set `_POSIX_VERSION` to the highest supported version less than or equal to `200809L`. Zstd than `#define`s `PLATFORM_POSIX_VERSION` to `_POSIX_VERSION` (mostly).

You would not believe some of the platforms people try to compile zstd for.

neheb:
Modified. Please review.

I also noticed that for utimensat, the time() call can be removed by setting UTIME_NOW for the field.

Good idea?

programs/util.h
```diff
@@ -25,12 +25,17 @@ extern "C" {
 #include <stdio.h>        /* fprintf */
 #include <sys/types.h>    /* stat, utime */
 #include <sys/stat.h>     /* stat, chmod */
-#if defined(_MSC_VER)
+#if defined(_WIN32)
```
felixhandte: Can you comment on the distinction between `_MSC_VER` and `_WIN32`? Why the change?

I haven't actually used a windows machine in years, so I'm unfamiliar with the dev ecosystem. @Cyan4973?

programs/util.c
```diff
@@ -54,14 +54,24 @@ int UTIL_getFileStat(const char* infilename, stat_t *statbuf)
 int UTIL_setFileStat(const char *filename, stat_t *statbuf)
 {
     int res = 0;
+#if defined(_WIN32) || (_POSIX_C_SOURCE < 200809L)
```
felixhandte: So you should be checking the value of `PLATFORM_POSIX_VERSION`, not `_POSIX_C_SOURCE`, in all of these.

I also have a weak preference for making these positive checks, putting the preferred behavior first (e.g., `#if PLATFORM_POSIX_VERSION >= 200809L` etc., not `#if PLATFORM_POSIX_VERSION < 200809L`).

felixhandte:
Great, thanks! Yeah, just passing in `UTIME_NOW` makes sense to me.

programs/util.h
neheb: _MSC_VER is only for Visual Studio. MinGW (POSIX wrapper around actual WIndows APIs) does not support utimensat because Windows does not support utimensat.

programs/util.c
neheb: PLATFORM_POSIX_VERSION >= 200809L is the wrong define here. Should be less than since it's for utime, not utimensat. Could make it > 200112L.

neheb: Actually that's wrong too...

I'll keep it as is.

felixhandte:
Were you going to switch to `UTIME_NOW`?

programs/util.c
```diff
@@ -54,14 +54,24 @@ int UTIL_getFileStat(const char* infilename, stat_t *statbuf)
...
 
     if (!UTIL_isRegularFile(filename))
         return -1;
 
+#if defined(_WIN32) || (PLATFORM_POSIX_VERSION < 200809L)
     timebuf.actime = time(NULL);
     timebuf.modtime = statbuf->st_mtime;
     res += utime(filename, &timebuf);  /* set access and modification times */
+#else
+    timebuf[0].tv_sec = time(NULL);
```
felixhandte: Also, can you initialize the `tv_nsec` fields to 0?

neheb: Initialized.

neheb:
I think that should be it.

programs/timefn.h
```diff
@@ -19,12 +19,6 @@ extern "C" {
 /*-****************************************
 *  Dependencies
 ******************************************/
-#include <sys/types.h>    /* utime */
-#if defined(_MSC_VER)
-#  include <sys/utime.h>  /* utime */
-#else
-#  include <utime.h>      /* utime */
-#endif
```
Cyan4973: 👍

Cyan4973:
As we are moving towards `v1.4.3`, which is a bugfix version with limited scope,
let's keep that PR in stock, and consider it again for a merge after `v1.4.3`.

I'm not against it, it looks good, and if `utime()` is now deprecated it seems a valid reason to consider it,
however this kind of change is typical of a modification that can break builds for several systems which are not part of our CI environment, hence are not directly tested, and I prefer to avoid such risk for `v1.4.3`.

neheb:
1.4.3 got merged.

PR #1709, opened 2019-08-01
Fix compression ratio inefficiency 

Cyan4973:
Fix compression ratio inefficiency 
introduced in #1624
under following scenario :
- dictionary compression
- at levels 1 to 3
- with medium size payloads (larger than `attachDict` threshold)

The patch restore compression ratio to previous level.

A new compression efficiency test is added into `fuzzer/unitTests` in order to control gross negative variations in compression ratio, playing all levels with normal compression and flat dictionary compression.

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -619,12 +619,14 @@ FORCE_INLINE_TEMPLATE size_t ZSTD_HcFindBestMatch_extDict_selectMLS (
 /* *******************************
 *  Common parser - lazy strategy
 *********************************/
-FORCE_INLINE_TEMPLATE
-size_t ZSTD_compressBlock_lazy_generic(
+typedef enum { search_hashChain, search_binaryTree } searchMethod_e;
+
+FORCE_INLINE_TEMPLATE size_t
+ZSTD_compressBlock_lazy_generic(
                         ZSTD_matchState_t* ms, seqStore_t* seqStore,
                         U32 rep[ZSTD_REP_NUM],
                         const void* src, size_t srcSize,
-                        const U32 searchMethod, const U32 depth,
+                        const searchMethod_e searchMethod, const U32 depth,
                         ZSTD_dictMode_e const dictMode)
 {
     const BYTE* const istart = (const BYTE*)src;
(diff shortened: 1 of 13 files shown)


terrelln:
Do you want to fix the rest of the strategies in this diff as well? The repro that exposes the bad behavior for every level is:

```
> head -c 40000 /dev/urandom > rand
> cp rand dict
> head -c 40000 /dev/urandom >> dict
> ./zstd -$LEVEL rand -D dict -o /dev/null
rand                 :100.04%   ( 40000 =>  40014 bytes, /dev/null)
```

Cyan4973:
Yes, I will

Cyan4973:
All strategies fixed, able to access the entire dictionary as long as at least one byte is within window size.

Still need to understand why the new size-control test fails on gcc-6 + 32-bit.

Also worthwhile to note :
the `usan` 32-bit build complains about integer overflow issues when reaching overflow protection tests for large `windowLog`.
It's a completely separate issue, but might be worth looking into.

Cyan4973:
The failing test due to lack of reproducibility between 32 and 64-bit builds is now fixed.
And now, another completely unrelated test fails later in the serie.
Great, such a pleasure ....

terrelln:
> As one can already guess, this test wasn't modified by a single line.

I'd love to switch these tests to gtest, given the time.

lib/compress/zstd_lazy.c
```diff
@@ -243,8 +243,10 @@ ZSTD_DUBT_findBestMatch(ZSTD_matchState_t* ms,
     const BYTE* const base = ms->window.base;
     U32    const current = (U32)(ip-base);
     U32    const maxDistance = 1U << cParams->windowLog;
-    U32    const windowValid = ms->window.lowLimit;
-    U32    const windowLow = (current - windowValid > maxDistance) ? current - maxDistance : windowValid;
+    U32    const lowestValid = ms->window.lowLimit;
+    U32    const withinWindow = (current - lowestValid > maxDistance) ? current - maxDistance : lowestValid;
+    U32    const isDictionary = (ms->loadedDictEnd != 0);
+    U32    const windowLow = isDictionary ? lowestValid : withinWindow;
```
terrelln: The only variable you need for the rest of the function is `windowLow` right?

Can you extract this out into a function since it is duplicated in every match finder?

Cyan4973: Yep, that's a good point

Cyan4973:
Last item completed, with `windowLow` logic factored into `zstd_compress_internal.h` .

PR #1712, opened 2019-08-08
Allocate Internal Buffers via Workspace Abstraction

felixhandte:
This pull request introduces a new abstraction, the `ZSTD_cwksp` and associated functions, which wraps and manage the single buffer in which all of the `ZSTD_CCtx` internal datastructures are allocated.

The goal of this pull request is to centralize and simplify the logic around allocating these datastructures. This enables future work to allow tables to use the same space whenever possible, even when we change cparams, which will allow us to avoid re-memsetting that table space, which is one avenue towards improving compression efficiency, especially for small inputs. Even independent of those future possibilities, this PR seems worth merging based solely on the cleanliness and separation of concern that it introduces.

This PR itself should change nothing about the behavior or performance of Zstd. (Other than re-ordering the allocations in the workspace, and minor changes that were enabled by this refactor, e.g., moving the `ZSTD_CDict` itself into the workspace rather than making a separate allocation.)

The correctness and performance has been verified to a basic degree. The built-in benchmarker shows no difference in performance. And `make test` all passes.


felixhandte:
Note: this current version has had no clean-up effort put into it. It's just the minimal version that mostly passes `make test` on my machine. I just want to get it up early to start showing you guys the approach I'm taking.

lib/compress/zstd_compress.c
```diff
@@ -38,15 +38,209 @@ size_t ZSTD_compressBound(size_t srcSize) {
...
+                                         * during at least this number of times,
+                                         * context's memory usage is considered wasteful,
+                                         * because it's sized to handle a worst case scenario which rarely happens.
+                                         * In which case, resize it down to free some memory */
+
+/**
+ * Align must be a power of 2.
+ */
+static size_t ZSTD_workspace_align(size_t size, size_t align) {
+    return size + align - 1 - ((size - 1) & (align - 1));
```
terrelln: This is a simpler version of alignment.

```c
size_t const mask = align - 1;
assert((align & mask) == 0);
return (size + mask) & ~mask;
```

lib/compress/zstd_compress_internal.h
```diff
@@ -220,6 +220,95 @@ struct ZSTD_CCtx_params_s {
...
+    void* objectEnd;
+
+    void* tableEnd;
+
+    void* allocStart;
+
+    int allocFailed;
+    int workspaceOversizedDuration;
+    ZSTD_workspace_alloc_phase_e phase;
+} ZSTD_CCtx_workspace;
```
Cyan4973: Naming : 
`ZSTD_CCtx_workspace` seems to mean that this object is the actual `workspace` of `CCtx`.

However, reading its declaration, it's a different thing : 
it contains pointers inside the actual `workspace`,
making it possible to track its different regions.

So it's related, but different.
I suggest changing the type name to something more specific, as the current name can be misleading when the declaration is not known (which is quite possible, since there is a large distance between declaration and usage of the type).

Possible candidates : 
- `ZSTD_CCtx_workspace_descriptor` 
- `ZSTD_CCtx_workspace_layout` 
- `ZSTD_CCtx_wksp_desc`
- etc.

Cyan4973: Btw, is `CCtx` meaningful ?
It seems `CDict` uses the same object too ?

terrelln:
This got me thinking if we actually need a memset at all. I prototyped an implementation for `ZSTD_fast` and `ZSTD_dfast` and I only saw a 0.5% regression on silesia for both strategies. The prototype commit is https://github.com/terrelln/zstd/commit/ba8bdf5dd73daceee1b2ec77512c2481a416910d (totally broken for ext-dict and dms modes).

I benchmarked the case of streaming compression of 1KB at level 3 without context reuse and it is 4x faster, and 20% faster at level 1. Compressing using `ZSTD_compressCCtx()` with a fresh CCtx is 2.5% faster at level 3, and the same speed at level 1.

I think that it could be an alternative to this PR, because with this PR we still have to memset the tables the first time. It could be done one strategy at a time to ease the transition. It would remove the need to use a context pool in HHVM zstd compression.

The downside is that we lose < 1% of compression speed for large compressions and context reuse with the same parameters. I think it is an acceptable trade off, but we should discuss tomorrow or in our meeting.

Cyan4973:
I investigated the same idea some time ago, though did not dare to actually measure it as you did (and to be fair these results are better than expected, they might depend on the actual content of the "garbage" values, because it would change the proportions, hence the efficiency of the branch predictor) .

The blocking point was the Binary Tree implementation.

For single-probe and hash table, it's fine, because every attempt checks both the validity of index and the full match length.

For the binary tree though, ensuring index validity can still be done, but the (current) algorithm implies that, after a few candidates, a certain nb of bytes are guaranteed to be correct at the beginning of the match. So it only checks the match after that guaranteed length.

Hence binary tree is exposed to a scenario where the index is garbage, but it happens to be within correct range, and it happens to correspond to a match after position `n`, but prior bytes before `n` are actually different. The outcome is data corruption.
Triggering this chain of events requires some quite extraordinary conditions, but it's not a good news : it means it's very hard to test.

That's where I stopped.

But that's no necessarily the end, there are possible solutions.
One possibility could be to differentiate tables with values guaranteed within `[0 - current]`, like this patch does, and tables with garbage values, and then disable the skipping part of BT match check when tables contain garbage values.

There is also the question of the memory sanitizer tests...

terrelln:
1. We could always memset for BT to start out with, and then go further with this patch by selectively enabling the match check, as you say.
2. We can detect when memory sanitizer is enabled and memset the tables.
3. Yeah, we would definitely need further testing with context reuse with different parameters.

felixhandte:
Ok, I'm pretty happy with the state of this PR.

Cyan4973:
The changes look good, and offer improved readability on what the code is doing.

I guess you also benchmarked them, and noticed no performance difference ?

@felixhandte , you mentioned a need to tidy up a bit the source code before merge ?

felixhandte:
Further performance testing has confirmed that performance is totally unchanged for large inputs. Small inputs can suffer somewhat. As this PR stands, I've observed compression slowing down up to 3% for small inputs (<=1KB), with dictionaries. I have an experiment which inlines all of the workspace functions and that improves things by about 1%.

I don't really have good theories about how to address that gap.

lib/compress/zstd_cwksp.c
```diff
@@ -0,0 +1,21 @@
...
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
+#include "zstd_cwksp.h"
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+/* all functions have been inlined... */
```
Cyan4973: very minor: 
some compiler(s) may complain about an empty unit.
I remember having to declare a useless symbol once in a similar case to avoid that.

Cyan4973:
So far, I've not reproduced the described performance issue with small inputs and dictionaries.
Any performance difference I could measure was insignificant.

Tested on a fairly stable desktop core i7-9700k, at level 1, on the `Github_Users` corpus (all files < 1 KB), with and without dictionary. Compiled with `gcc` v8.3.0.

Maybe the test settings need to be altered to observe this effect.

PR #1713, opened 2019-08-08
Fix the build on GCC 4.x after 812e8f2a1

cemeyer:
Noticed and fixed by Warner Losh <imp@FreeBSD.org>.

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -128,7 +128,7 @@
 }
 
 /* vectorization */
-#if !defined(__clang__) && defined(__GNUC__)
+#if !defined(__clang__) && defined(__GNUC__) && __GNUC__ >= 5
 #  define DONT_VECTORIZE __attribute__((optimize("no-tree-vectorize")))
 #else
 #  define DONT_VECTORIZE


lib/common/compiler.h
```diff
@@ -128,7 +128,7 @@
 }
 
 /* vectorization */
-#if !defined(__clang__) && defined(__GNUC__)
+#if !defined(__clang__) && defined(__GNUC__) && __GNUC__ > 5
```
Cyan4973: 1) What about `clang` ?
2) Can you point at some online doc where the version requirement is explained ?

cemeyer: 1. The `DONT_VECTORIZE` macro to disable the option is not needed on Clang.  This is described in the original commit https://github.com/facebook/zstd/commit/812e8f2a1 and covered correctly by the preexisting `#if !defined(__clang__)` portion of the condition.

2. https://gcc.gnu.org/gcc-4.4/changes.html:

> ### New Languages and Language specific improvements:
> **C family:**
> A new `optimize` attribute was added …

Here's a link to FreeBSD's [CI run (warning: large text)](https://ci.freebsd.org/job/FreeBSD-head-sparc64-build/14315/console) (we build with `-Werror`):

> /usr/src/sys/contrib/zstd/lib/common/zstd_internal.h:213: warning: 'optimize' attribute directive ignored

It seems Numpy [had a similar issue](https://github.com/numpy/numpy/issues/5450) on OpenBSD in the past (same 4.2.1 GCC).

Cyan4973: Thanks for sharing the details !

> https://gcc.gnu.org/gcc-4.4/changes.html:

So it's available since `gcc v4.4` then ?
Hence, should it rather be `__GNUC__ >= 5` ?

cemeyer: Yes, I think since 4.4.  `>=` is fine with me.

terrelln:
@Cyan4973 @felixhandte do we want to merge this before the next release?

Cyan4973:
It seems that current code base breaks compilation on platforms using gcc < v4.4 .
(Unfortunately, I could not test it directly, because none of my VM provides a gcc < v4.4.)
Nonetheless, I'm rather favorable to this fix, which improves situation for these older systems, and seems mostly harmless for all others.

cwoffenden:
I realised I overwrote this change with [my recent commit here](https://github.com/facebook/zstd/blob/e9c0fc12d28a2d0c826752e0fd73c8eb0074983f/lib/common/compiler.h#L133). I'd discovered the same thing (with < 4.4 as the cutoff) and added a compatible `_Pragma("GCC optimize(\"no-tree-vectorize\")")` in it's place. This was successfully tested with GCC 4.2 (on an old Mac PPC machine we keep around for big endian testing).

PR #1732, opened 2019-08-19
Add --size-hint=# option

nmagerko:
Certain streaming situations can result in significantly different compression ratios between a file and stdin (streaming the same file). This is mainly an issue for smaller source sizes, where the default compression parameters are not appropriate.

This diff adds a --size-hint option, which allows the user to specify the approximate size of a source stream. The hint does not need to be exactly the same size as the incoming stream. 

The following are outputs from a few different play tests with hints that are not exact:

```
/*stdin*\         : 21.79%   ( 11000 =>   2397 bytes, tmp.zst)   # 50 bytes too big
/*stdin*\         : 21.79%   ( 11000 =>   2397 bytes, tmp.zst)   # 50 bytes too small
/*stdin*\         : 21.85%   ( 11000 =>   2403 bytes, tmp.zst)   # half actual stream size
/*stdin*\         : 22.15%   ( 11000 =>   2436 bytes, tmp.zst)   # twice actual stream size
```

Compressing the streamed content from a file yields a ratio of `21.80%` while streaming the same content without a hint yields a ratio of `21.95%`. Thus, correct or close-to-correct hints improve the compression ratio. However, it would seem that serious overestimates hurt the compression ratio quite a bit while underestimating (at least at this source size) is not as severe.

Tests have been added to ensure that this parameter keeps the compression ratio close to the expected ratio (compared to file-based compression).

Resolves #1720.


nmagerko:
Closing, need to rebase from `dev`.

PR #1744, opened 2019-08-22
Generate RLE blocks in the encoder

bimbashrestha:
Using RLE blocks when the size upper bound criteria is met and the block is a RLE.


lib/compress/zstd_compress.c
```diff
@@ -2372,9 +2381,21 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
...
+                    We check to see if the regularly compressed block size is
+                    less than 10 (the upper bound for RLE blocks) and we check
+                    to see if the block is an RLE
+                */
+                if (cSize < 10 && ZSTD_isRLE(ip, blockSize)) {
+                    op[ZSTD_blockHeaderSize] = ip[0];
+                    cBlockHeader = lastBlock + (((U32)bt_rle)<<1) + (U32)(blockSize << 3);
+                    cSize = ZSTD_blockHeaderSize + 1;
+                } else {
+                    cBlockHeader = lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
```
terrelln: Add `assert(!ZSTD_isRLE(ip, blockSize))` here so we can know if the upper bound is wrong.

lib/compress/zstd_compress.c
```diff
@@ -2372,9 +2381,21 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
                 cSize = ZSTD_noCompressBlock(op, dstCapacity, ip, blockSize, lastBlock);
                 FORWARD_IF_ERROR(cSize);
             } else {
-                U32 const cBlockHeader24 = lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
-                MEM_writeLE24(op, cBlockHeader24);
-                cSize += ZSTD_blockHeaderSize;
+                U32 cBlockHeader;
```
terrelln: This needs to be done inside of `ZSTD_compressBlock_internal()`, because when we use an RLE block we need to not swap `prevCBlock` and `nextCBlock` https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L2296.

Then we'll ned some way to signal if the block is RLE or not. Returning `cSize == 1` for RLE would work.

lib/compress/zstd_compress.c
```diff
@@ -2287,12 +2299,19 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
             zc->entropyWorkspace, HUF_WORKSPACE_SIZE /* statically allocated in resetCCtx */,
             zc->bmi2);
 
+    if (cSize < 10 && ZSTD_isRLE(ip, srcSize)) {
```
Cyan4973: 1) I guess it means that the threshold to determine that a block _might_ be RLE is 10 bytes ?
In which case, give a name to this constant, it's useful for code documentation.

2) A previous bootcamper tried to generate statistics on this topic.
His results suggested that the limit should be higher, because many RLE-possible blocks ended up being larger than example. His suggestion was rather in the ~25 bytes.
The question of the threshold is probably worth exploring.

terrelln: The `dictionary_round_trip` regression test suggests ~25 bytes as well.

However, it also has an example where an RLE block is 423 bytes. In that case literal compression is disabled. So unless we don't use a heuristic, we'll never get it exactly right.

A middle ground could be to use a larger than necessary heuristic, but spend a little bit of time optimizing `ZSTD_isRLE()` so we can afford to run it more often.

Cyan4973: At this stage, I believe we are fine missing a few outliers.
Catching the remaining ones could be seen as a later refinement, if it ever becomes a topic.

PR #1754, opened 2019-08-28
Tests for single-file decompressor

Cyan4973:
This patch simply automates the amalgation + compilation test
of the single-file decompressor provided by @cwoffenden 
in `contrib/single-file-decoder/`.

A few script names were updated to help users better understand their purpose.


terrelln:
Looks like `simple.c` needs to be fixed.

cwoffenden:
I created #1755 to fix the C99-related error (which fixes the CI error above).

Cyan4973:
Thanks @cwoffenden ! The fix works great !

PR #1760, opened 2019-08-29
Adding api for extracting sequences from seqstore

bimbashrestha:


bimbashrestha:
Not ready for review yet:)

lib/compress/zstd_compress.c
```diff
@@ -13,6 +13,7 @@
 ***************************************/
 #include <limits.h>         /* INT_MAX */
 #include <string.h>         /* memset */
+#include <stdlib.h>
```
Cyan4973: why do you need `<stdlib.h>` ?

lib/compress/zstd_compress.c
```diff
@@ -2394,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
     }
 }
 
+static void ZSTD_copyBlockSequences(const seqStore_t* seqStore, seqDef* seqs,
```
Cyan4973: minor: prototype style : parameter order, 
generally prefer the mutable target state first, then the arguments (like immutable input).
like `memcpy(out, in, size)`, or `ZSTD_compress(outPtr, outSize, inPtr, inSize, lvl)` .

lib/compress/zstd_compress.c
Cyan4973: OK, here is a more difficult design principle for `libzstd` : 
we are always trying to reduce `#include` dependencies to a minimum.
So adding one is a big deal,
especially an external dependency such as `<stdlib.h>`.

We have made our code mostly `<stdlib.h>` free, so this is a step back.

I think it's possible to preserve this property without changing the application logic
by using `ZSTD_malloc()` and `ZSTD_free()` instead.
They need a `ZSTD_customMem` argument, but since you have none specified, you can simply use `ZSTD_defaultCMem` instead.

Without exploiting the ability to introduce custom allocators, the only advantage is that it allows removing `#include <stdlib.h>`. I know it looks small, but it matters.

Cyan4973: For the record, 
another to avoid this issue is to not need a `dst` buffer altogether, hence no need to allocate it.

This is possible, since we never use the compressed result, only the intermediate `seqStore`.

However, this would require to use a different call than `ZSTD_compressCCtx()`, invoking some more internal function instead, to intercept the result right after `seqStore` generation. This may require to refactor some of these internal functions.

As one can guess, it's a more difficult change.

lib/compress/zstd_compress.c
```diff
@@ -2394,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+static void ZSTD_getBlockSequences(ZSTD_CCtx* cctx, const seqStore_t* seqStore)
+{
+    size_t seqsSize = seqStore->sequences - seqStore->sequencesStart;
+
+    assert(cctx->seqCollector.maxSequences >
+        (cctx->seqCollector.seqCurrent - cctx->seqCollector.seqStart) + seqsSize);
+
+    ZSTD_copyBlockSequences(seqStore, seqStore->sequencesStart,
+        cctx->seqCollector.seqCurrent, seqsSize);
+    cctx->seqCollector.seqCurrent += seqsSize;
```
Cyan4973: I will have to read more to understand this,

but at this stage, I'm surprised that `cctx` contains a `seqCollector`,
and I'm not sure to understand why it would be necessary.

A consequence of this design choice is that `cctx` is a mutable state for `ZSTD_getBlockSequences()`.
If the `seqCollector` was external, `cctx` could be just read, hence immutable, which "feels" more logical and safer.

But as stated, I may miss something important, so I must read more of this PR.

lib/compress/zstd_compress.c
```diff
@@ -2394,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+}
+
+size_t ZSTD_getSequences(ZSTD_CCtx* zc, const void* src,
```
Cyan4973: minor : prototype design : 
as mentioned before,
it seems that, in this prototype, the target is `outSeqs`, so it should come before `src`.

Regarding `ZSTD_CCtx* zc`, I'm wondering if it's really necessary.
The usage seems self-contained : no dependency on what happened before on `zc`, no need to do something with `zc` after this function.
So, maybe this state could be created internally ?
This would remove one mutable parameter from this prototype, making it easier and safer to invoke.

lib/compress/zstd_compress.c
```diff
@@ -2394,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+    size_t srcSize, ZSTD_Sequence* outSeqs, size_t outSeqsSize,
+    int level)
+{
+    size_t dstCapacity = ZSTD_compressBound(srcSize * sizeof(void*));
+    void* dst = malloc(dstCapacity);
```
Cyan4973: one must check if `malloc()` was successful (i.e. `dst != NULL`).

lib/compress/zstd_compress.c
```diff
@@ -2394,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+    size_t seqsSize;
+
+    SeqCollector seqCollector;
+    seqCollector.collectSequences = 1;
+    seqCollector.seqStart = outSeqs;
+    seqCollector.seqCurrent = outSeqs;
+    seqCollector.maxSequences = outSeqsSize;
+    zc->seqCollector = seqCollector;
+
+    ZSTD_compressCCtx(zc, dst, dstCapacity, src, srcSize, level);
```
Cyan4973: one must check if `ZSTD_compressCCtx()` was successful (i.e. `! ZSTD_isError(return_value)`).

lib/compress/zstd_compress.c
```diff
@@ -2394,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+    size_t dstCapacity = ZSTD_compressBound(srcSize * sizeof(void*));
+    void* dst = malloc(dstCapacity);
+    size_t seqsSize;
+
+    SeqCollector seqCollector;
+    seqCollector.collectSequences = 1;
+    seqCollector.seqStart = outSeqs;
+    seqCollector.seqCurrent = outSeqs;
+    seqCollector.maxSequences = outSeqsSize;
+    zc->seqCollector = seqCollector;
```
Cyan4973: OK, so that's why there is a `seqCollector` into `cctx` : 
it's actually a pointer to this stack structure.

OK, I'm still interrogative if this is the better design choice,
though I understand the simplicity of it when paired with `ZSTD_compressCCtx()`.

lib/compress/zstd_compress.c
```diff
@@ -2438,6 +2443,9 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
                                 op+ZSTD_blockHeaderSize, dstCapacity-ZSTD_blockHeaderSize,
                                 ip, blockSize);
             FORWARD_IF_ERROR(cSize);
+            if (cctx->seqCollector.collectSequences) {
```
Cyan4973: OK, so that's where multiple blocks of sequences get concatenated.

lib/zstd.h
```diff
@@ -1210,6 +1218,9 @@ ZSTDLIB_API unsigned long long ZSTD_decompressBound(const void* src, size_t srcS
  *           or an error code (if srcSize is too small) */
 ZSTDLIB_API size_t ZSTD_frameHeaderSize(const void* src, size_t srcSize);
 
+ZSTDLIB_API size_t ZSTD_getSequences(ZSTD_CCtx* zc, const void* src,
```
Cyan4973: minor: document this function, using code comments.

bimbashrestha:
Thanks for the comments @Cyan4973. I think the core decision to made now is this. 

1) We can either compromise mutability of cctx in ZSTD_getSequences() (as I have done above) and have no new redundancy in compression logic. That is, we just reuse the flow from ZSTD_compressCCtx() and take sequences from the store after every block compression. 

2) Or we can externalize the SeqCollector (thus maintaining mutability) and introduce some redundancy in compression logic by effectively recreating the flow of ZSTD_compressCCtx() in a new API. Correct me if I'm wrong but it looks like doing this would be a substantially bigger endeavor than 1).

Not totally sure what the best course of action here. We want to extract sequences because doing so will allow us to do compression analysis without needing to know the internals of zstd. If we expect to want more internal properties of the algorithm extracted for analysis in the future, creating a new external framework 'ZSTD_DataCollector' (this would be more in line with 2)) might be a good idea. Thoughts?

lib/compress/zstd_compress_internal.h
```diff
@@ -192,6 +192,13 @@ typedef struct {
   size_t capacity; /* The capacity starting from `seq` pointer */
 } rawSeqStore_t;
 
+typedef struct {
+    int collectSequences;
+    ZSTD_Sequence* seqStart;
+    ZSTD_Sequence* seqCurrent;
```
Cyan4973: minor: API style:

we have something loosely equivalent in the streaming interface : [`ZSTD_outBuffer`](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L555).

The general idea is that it's a bit better to represent a limited and partially filled destination buffer as `ptr, usize, usize` rather than `ptr, ptr, usize`.
One reason is that it's easier to reason about bounds, hence to detect and reject incorrect inputs.

Applied here, this would result in : 
```C
   ZSTD_Sequence* seqStart;
   size_t nbSeqs;
   size_t scCapacity;
```

lib/compress/zstd_compress.c
Cyan4973: Also: is there any need for `seqDef* seqs` to be mutable ?
It seems to be read only,
in which case prefer `const seqDef* seqs`.

Cyan4973:
Yes, I believe option 2 is better.

However, recreating a parallel compression logic is bad. We should rely on exactly the same compression logic as normal compression, to avoid discrepancies in the future.

This however requires some code refactoring, untying hard-wired operations to create new interfaces. The complexity of which I have not yet assessed.

lib/compress/zstd_compress.c
terrelln: Why do we need to pass in `seqs`? Can we just get them from the `seqStore`?

lib/compress/zstd_compress.c
```diff
@@ -2323,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
 }
 
+static void ZSTD_copyBlockSequences(const seqStore_t* seqStore, seqDef* seqs,
+    ZSTD_Sequence* outSeqs, size_t seqsSize)
+{
+    size_t i; size_t position; int repIdx;
+    for (i = 0, position = 0; i < seqsSize; ++i) {
+        outSeqs[i].offset = seqs[i].offset;
+        outSeqs[i].litLength = seqs[i].litLength;
+        outSeqs[i].matchLength = seqs[i].matchLength + 3 /* min match */;
```
terrelln: Instead of `3` use `MINMATCH`.

lib/compress/zstd_compress.c
```diff
@@ -2323,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+
+        if (i == seqStore->longLengthPos) {
+            if (seqStore->longLengthID == 1) {
+                outSeqs[i].litLength += 0x10000;
+            } else if (seqStore->longLengthID == 2) {
+                outSeqs[i].matchLength += 0x10000;
+            }
+        }
+
+        if (outSeqs[i].offset <= 3 /* num reps */) {
```
terrelln: This block doesn't work for `litLength == 0`. See https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#repeat-offsets.

lib/compress/zstd_compress.c
```diff
@@ -2323,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+            outSeqs[i].rep = 1;
```
terrelln: Lets set this to 1-4. 1-3 mean that offset. 4 means repcode 3 with litlength == 0.

So normally it will be 1-3. When litLength == 0 it will be 2-4.

lib/compress/zstd_compress.c
```diff
@@ -2323,6 +2324,81 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms, ZSTD_CCtx_params
...
+
+            if (repIdx == -1) {
+                outSeqs[i].offset = 1;
+            } else if (repIdx == -2) {
+                outSeqs[i].offset = 4;
+            } else if (repIdx == -3) {
+                outSeqs[i].offset = 8;
+            }
+        } else {
+            outSeqs[i].offset -= 3 /* num reps */;
```
terrelln: Instead of 3 use `ZSTD_REP_NUM`

lib/compress/zstd_compress.c
terrelln: I want to pass a `ZSTD_CCtx` into this function. That way we can configure it however we want, like adding a dictionary.

That means that we shouldn't pass a `level`, and should use the new advanced API to compress.

lib/compress/zstd_compress.c
terrelln: We should use `ZSTD_compress2()` here.

lib/compress/zstd_compress.c
```diff
@@ -2291,11 +2291,22 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
             outSeqs[i].rep = 1;
             repIdx = i - outSeqs[i].offset;
 
+            /* Not first block */
             if (repIdx >= 0) {
-                outSeqs[i].offset = outSeqs[repIdx].offset;
-            }
-
-            if (repIdx == -1) {
+                /* Special case where litLength == 0 */
```
bimbashrestha: This seems consistent with the docs to me. Please look at this carefully and let me know if I misunderstood.

lib/compress/zstd_compress.c
```diff
@@ -2291,11 +2291,22 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
...
+                    /* When the offset is either 1 or 2 */
+                    } else {
+                        outSeqs[i].offset = outSeqs[repIdx - 1].offset;
+                    }
+                } else {
+                    outSeqs[i].offset = outSeqs[repIdx].offset;
+                }
+            /* First block */
+            } else if (repIdx == -1) {
                 outSeqs[i].offset = 1;
```
bimbashrestha: I assume the macros for these default first block offset constants should be in zstd_internal.h but I can't find them there.

terrelln: It is `repStartValue`

lib/compress/zstd_compress.c
terrelln: This isn't quite right, it will access out of bounds when `repIdx == 0`.

It might be simpler to do something like:

```c
if (outSeqs[i].litLength == 0) {
    if (outSeqs[i].offset < 3) {
        --repIdx;
    } else {
        repIdx = i - 1;
    }
}
assert(repIdx >= -3);
outSeqs[i].offset = repIdx >= 0 ? outSeqs[repIdx].offset : repStartValue[-repIdx - 1];
if (outSeqs[i].offset == 3) {
    --outSeqs[i].offset;
}
```

lib/compress/zstd_compress.c
```diff
@@ -2264,6 +2264,88 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+            outSeqs[i].rep = 1;
+            repIdx = i - outSeqs[i].offset;
+
+            /* Not first block */
+            if (repIdx >= 0) {
+                /* Special case where litLength == 0 */
+                if (outSeqs[i].litLength == 0) {
+                    /* When the offset is 3 */
+                    if (outSeqs[i].offset > 2) {
+                        outSeqs[i].offset = outSeqs[repIdx - 1].offset - 1;
```
terrelln: I believe that this should be `i - 1`, not `repIdx - 1`.

lib/compress/zstd_compress.c
```diff
@@ -2264,6 +2264,78 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+            if (outSeqs[i].litLength == 0) {
+                if (outSeqs[i].offset < 3) {
+                    --repIdx;
+                } else {
+                    repIdx = i - 1;
+                }
+            }
+            assert(repIdx >= -3);
+            outSeqs[i].offset = repIdx >= 0 ? outSeqs[repIdx].offset : repStartValue[-repIdx - 1];
+            if (outSeqs[i].offset == 3) {
```
terrelln: `if (outSeqs[i].rep == 4)`.

lib/compress/zstd_compress.c
```diff
@@ -2264,6 +2264,78 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+        if (i == seqStore->longLengthPos) {
+            if (seqStore->longLengthID == 1) {
+                outSeqs[i].litLength += 0x10000;
+            } else if (seqStore->longLengthID == 2) {
+                outSeqs[i].matchLength += 0x10000;
+            }
+        }
+
+        if (outSeqs[i].offset <= ZSTD_REP_NUM) {
+            outSeqs[i].rep = 1;
```
terrelln: Lets save the actual rep number here.

```
outSeqs[i].rep = outSeqs[i].offset;
```

Then when `litLength == 0` we'll have to increment it by 1.

lib/compress/zstd_compress.c
```diff
@@ -2287,6 +2360,10 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
             zc->entropyWorkspace, HUF_WORKSPACE_SIZE /* statically allocated in resetCCtx */,
             zc->bmi2);
 
+    if (zc->seqCollector.collectSequences) {
```
Cyan4973: As mentioned previously,
I believe it's reasonable to consider that,
while collecting sequences, 
users are not expecting compressed data to be actually produced. 
This is, btw, exactly the outcome of `ZSTD_getSequences()`, which throws away the compressed result.

Therefore, don't `ZSTD_compressSequences()`.
Just invoke `ZSTD_copyBlockSequences()` before that point, and bail out with a `0` result.
This will free `ZSTD_getSequences()` from allocating a large memory block.

tests/fuzzer.c
```diff
@@ -1960,6 +1960,11 @@ static int basicUnitTests(U32 const seed, double compressibility)
         DISPLAYLEVEL(3, "OK \n");
     }
 
+    DISPLAYLEVEL(3, "test%3i : ZSTD_getSequences zeros : ", testNb++);
+    memset(CNBuffer, 0, 1000000);
+    assert(ZSTD_getSequences(ZSTD_createCCtx(), compressedBuffer, 1000000,
```
Cyan4973: this `assert()` produces a side effect and a memory leak.

lib/compress/zstd_compress.c
```diff
@@ -2264,6 +2264,79 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+        position += outSeqs[i].matchLength;
+    }
+    zc->seqCollector.seqIndex += seqsSize;
+}
+
+/* We call compress2() and collect sequences after each block
+ * compression. The function stores the ZSTD_Sequences in outSeqs
+ * and returns the number of collected sequences from all blocks.
+ */
+size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
```
Cyan4973: I'm unconvinced that `ZSTD_CCtx* zc` is required in this interface.

If it doesn't transport any information to/from user code,
prefer allocating and freeing it internally,
it will make this interface easier to use & integrate.

terrelln: We need to keep `ZSTD_CCtx* zc` in this interface. We need it to set compression parameters and use a `ZSTD_CDict*`.

Cyan4973: Yes that's indeed a good reason. 
The change from `compressCCtx()` to `compress2()` makes this possible now.

lib/zstd.h
```diff
@@ -1072,6 +1072,14 @@ ZSTDLIB_API size_t ZSTD_sizeof_DDict(const ZSTD_DDict* ddict);
 
 typedef struct ZSTD_CCtx_params_s ZSTD_CCtx_params;
 
+typedef struct {
```
Cyan4973: We probably need to document that structure, so that any user can quickly understand what it's about,
and the fields' meaning is clearly established.
For example, I'm not sure just from the naming if `matchPos` is the position of the match at its destination point (I presume), or at its source point. Also it's relative to something (likely `src`).
There was also this discussion about the meaning of `rep` values.
Just clarify, for any 3rd party.

lib/zstd.h
```diff
@@ -1210,6 +1218,9 @@ ZSTDLIB_API unsigned long long ZSTD_decompressBound(const void* src, size_t srcS
  *           or an error code (if srcSize is too small) */
 ZSTDLIB_API size_t ZSTD_frameHeaderSize(const void* src, size_t srcSize);
 
+ZSTDLIB_API size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
```
Cyan4973: Document the function too.
For example, tell that `zc` can be used to insert custom compression parameters.

tests/fuzzer.c
bimbashrestha: Perhaps a silly question:) What is the issue here?

Cyan4973: `ZSTD_createCCtx()` allocates a context, which is never freed afterwards, so it's a leak.
This is assuming that allocation is successful, otherwise it returns a `NULL` pointer, and using it triggers undefined behavior.

It's also a general rule that no `assert()` should ever trigger any side effect. Heap allocation is a side effect.

Cyan4973:
There is still a minor issue to fix on this PR : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/builds/27451730/job/v1xf493k9ya4d6a7#L32

tests/fuzzer.c
```diff
@@ -1960,6 +1960,16 @@ static int basicUnitTests(U32 const seed, double compressibility)
         DISPLAYLEVEL(3, "OK \n");
     }
 
+    {
+        ZSTD_CCtx* cctx = ZSTD_createCCtx();
```
Cyan4973: minor : 
just FYI,
as a programming pattern, we absolutely love the `const` property, and try to use it anywhere possible.

For example, in this case, the declaration could be : 
`ZSTD_CCtx* const cctx = ZSTD_createCCtx();` .

This means that, after it's initialized, `cctx` address will never change. This is different from the content, which is still mutable.

When used extensively, the `const` property is very useful to protect the code from accidental modifications, and for debugging : now the _absence of_ `const` means that a variable _is __mutable___, and this is a useful signal, as most bugs are concentrated in the mutation of states.
Obviously, this signal is only meaningful if variables are consistently labelled `const` anytime they are.

It's absolutely not a blocker, and I'll validate and merge this patch "as is". This comment is made just to influence your coding style in future contributions.

lib/compress/zstd_compress.c
```diff
@@ -2265,6 +2265,75 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+                if (outSeqs[i].offset < 3) {
+                    --repIdx;
+                } else {
+                    repIdx = (unsigned int)i - 1;
+                }
+                ++outSeqs[i].rep;
+            }
+            assert(repIdx >= -3);
+            outSeqs[i].offset = repIdx >= 0 ? outSeqs[repIdx].offset : repStartValue[-repIdx - 1];
+            if (outSeqs[i].offset == 4) {
```
terrelln: `if (outSeqs[i].rep == 4)`

lib/compress/zstd_compress.c
```diff
@@ -2265,6 +2265,75 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+        position += outSeqs[i].matchLength;
+    }
+    zc->seqCollector.seqIndex += seqsSize;
+}
+
+size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
+    size_t outSeqsSize, const void* src, size_t srcSize)
+{
+    const size_t dstCapacity = ZSTD_compressBound(srcSize * sizeof(void*));
+    void* dst = ZSTD_malloc(dstCapacity, ZSTD_defaultCMem);
```
terrelln: We need to check this malloc for `NULL`.

terrelln: Since we aren't compressing the sequences, do we actually need the `dst` buffer? Will it work with an empty output buffer? If it doesn't, can we easily make it work?

lib/zstd.h
```diff
@@ -1077,6 +1077,14 @@ ZSTDLIB_API size_t ZSTD_sizeof_DDict(const ZSTD_DDict* ddict);
 
 typedef struct ZSTD_CCtx_params_s ZSTD_CCtx_params;
 
+typedef struct {
+    unsigned int matchPos; /* match pos in dst */
+    unsigned int offset; /* offset taking into account rep (different from seqdef) */
+    unsigned int litLength; /* literal length */
+    unsigned int matchLength; /* match length */
+    unsigned int rep; /* 0 when seq not rep and seqDef.offset otherwise */
```
terrelln: The docs aren't quite accurate in the `litLength == 0` case. Don't be afraid to add a multiline comment here.

lib/compress/zstd_compress.c
bimbashrestha: Have a look at the new commit. I just passed in src to get rid of the malloc. We never write to dst anyway when collecting sequences so this should be harmless as far as I can tell.

lib/compress/zstd_compress.c
```diff
@@ -2265,6 +2265,72 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+{
+    SeqCollector seqCollector;
+    seqCollector.collectSequences = 1;
+    seqCollector.seqStart = outSeqs;
+    seqCollector.seqIndex = 0;
+    seqCollector.maxSequences = outSeqsSize;
+    zc->seqCollector = seqCollector;
+
+    /* We never write to dst when collecing sequences so setting dst = src is harmless */
+    ZSTD_compress2(zc, (void*)src, srcSize, src, srcSize);
```
Cyan4973: nope, don't use a `const` buffer as a parameter requiring a mutable buffer,
even if you know or suspect it won't be written into.

If all you need is a fake buffer to satisfy the interface, you may either : 
- provide `NULL`, and a size of `0`
- create a minimal buffer in stack, like a `size_t buffer[1]`, and refer to this one.

The first case doesn't need any resource but may fail, depending on code to follow.
The second case is sure to work, but uses a (very minimal) amount of resource.

lib/compress/zstd_compress.c
Cyan4973: It's unclear (to me) if destination buffer is never written into. Maybe, but there is a lot more code being traversed than just sequence collection. Frame header, for example, might be written from a different part of the code base. I don't know for sure, so it deserves to be tested.

Even if just a few bytes need to be written, this can be solved, by creating a small temporary buffer on stack, and providing that as output of the function. This is still a significant win compared to allocating a larger buffer on heap.

lib/compress/zstd_compress.c
bimbashrestha: Passing NULL and 0 results in only one block being acted upon. There must be a check somewhere that triggers this but it's not clear to me where. 

A buffer of size 1 has the same effect as above. I have to cross some  threshold for buffer size before I start seeing all the blocks acted upon. But I can't tell if this threshold is variable with srcSize or not. Will have to dig some more.

bimbashrestha: Just did a test. It looks like the threshold does increase with srcSize. I changed it back to what I had before. Allocating a ZSTD_compressBound(srcSize) dst buffer on the heap.

lib/compress/zstd_compress.c
terrelln: We need to check this for `NULL`.

lib/compress/zstd_compress.c
```diff
@@ -2321,8 +2321,10 @@ size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
 {
     const size_t dstCapacity = ZSTD_compressBound(srcSize);
     void* dst = ZSTD_malloc(dstCapacity, ZSTD_defaultCMem);
-
     SeqCollector seqCollector;
+
+    assert(dst != NULL);
```
Cyan4973: An `assert()` means that this situation should _never_ happen.
It's only activated during debug sessions, to ensure that the condition is indeed valid as promised, and then is removed at production time.

In this case, we can't _guarantee_ that `malloc()` will necessarily succeed.
It may fail and return `NULL`.

Therefore, what's needed here is to actively check this condition, and _return an __error code___ when it's triggered.

`memory_allocation` seems the correct error code.
`RETURN_ERROR_IF()` seems the right macro to deal with this situation.

lib/compress/zstd_compress.c
Cyan4973: Regarding the need for a `dst` buffer :
well, it's a pity we can't avoid that, 
and it would be interesting to understand why 
there is a need to provide a (large) `dst` buffer for sequence-extraction to work, 
this would have allowed us to get rid of this `malloc`.

But this can be done as a later step.

tests/fuzzer.c
```diff
@@ -1960,6 +1960,34 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        size_t expectedMatchLengths[] = {ZSTD_BLOCKSIZE_MAX - 2, ZSTD_BLOCKSIZE_MAX - 2};
+        size_t expectedReps[] = {1, 1};
+        size_t expectedMatchPos[] = {2, 1};
+        size_t expectedSequencesSize = 2;
+        size_t sequencesSize;
+        size_t i = 0;
+        ZSTD_Sequence* sequences = (ZSTD_Sequence*)compressedBuffer;
+        DISPLAYLEVEL(3, "test%3i : ZSTD_getSequences zeros : ", testNb++);
+        assert(cctx != NULL);
+        memset(CNBuffer, 0, zerosLength);
```
terrelln: Instead of compressing only zeros, can we have a more realistic scenario?

It should be easy to write a function

```c
size_t decompressSequences(BYTE* dst, size_t dstCapacity, BYTE const* src, size_t srcSize, ZSTD_Sequence const* seqs, size_t nbSeq);
```

since the sequences tell you exactly what to do. Then you can call `ZSTD_getSequences()` on `CNBuffer` and validate that `decompressSequences()` works.

tests/fuzzer.c
```diff
@@ -1960,6 +1980,34 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        ZSTD_CCtx* cctx = ZSTD_createCCtx();
+        ZSTD_Sequence* seqs = (ZSTD_Sequence*)malloc(srcSize * sizeof(ZSTD_Sequence));
+        size_t seqsSize; size_t i;
+        U32 randSeed = seed;
+
+        if (seqs == NULL) goto _output_error;
+        assert(cctx != NULL);
+
+        /* Populate src with random data */
+        for (i = 0; i < srcSize / sizeof(U32); ++i) {((U32*)CNBuffer)[i] = FUZ_rand(&randSeed);}
```
terrelln: This is likely to produce completely incompressible data, meaning 1 large litlength. We have a function to do this. You can set `srcSize == CNBuffSize`.

```
RDG_genBuffer(CNBuffer, CNBuffSize, compressibility, 0., seed);
```

bimbashrestha: The memcmp assert fails when I'm using data generated using RDG_genBuffer(...). FUZ_decodeSequences looks fine to me so I think there is something wrong with the rep code parsing in the getSequences api. Need to double check this

lib/compress/zstd_compress.c
```diff
@@ -2272,37 +2272,46 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
...
-                    case 1: outSeqs[i].offset = repeatOffset2; break;
-                    case 2: outSeqs[i].offset = repeatOffset3; break;
-                    /* corner case where offsetValue == 3 */
-                    case 3: outSeqs[i].offset = repeatOffset1 - 1; break;
+        outSeqs[i].matchLength = seqs[i].matchLength + MINMATCH;
+
+        if (i == seqStore->longLengthPos) {
+            if (seqStore->longLengthID == 1) {
+                outSeqs[i].litLength += 0x10000;
+            } else if (seqStore->longLengthID == 2) {
```
bimbashrestha: Ignore changes to this file. This just reverts changes from the last commit (which was accidental). Only the fuzzer.c changes are new

bimbashrestha:
This one looks ready to merge to me:)

PR #1761, opened 2019-08-30
[regression] small improvement from RLE blocks

terrelln:


Cyan4973:
Sure, it's small, but I wasn't expecting any impact at all on our regular samples, so it's good news that it's able to show up.

I see it impacts "small window log", which probably means smaller blocks, which makes it possible for some full block of `\0` to happen.

PR #1769, opened 2019-09-01
Fix streaming compression/decompression examples

cybojanek:
* Handle compression of empty file
* Error in decompression in case of trailing data

### DIFF
--- a/examples/streaming_compression.c
+++ b/examples/streaming_compression.c
@@ -44,8 +44,8 @@ static void compressFile_orDie(const char* fname, const char* outName, int cLeve
      * and writes all output produced to the output file.
      */
     size_t const toRead = buffInSize;
-    size_t read;
-    while ((read = fread_orDie(buffIn, toRead, fin))) {
+    for (;;) {
+        size_t read = fread_orDie(buffIn, toRead, fin);
         /* Select the flush mode.
          * If the read may not be finished (read == toRead) we use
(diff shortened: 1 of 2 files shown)


examples/streaming_compression.c
```diff
@@ -78,6 +80,20 @@ static void compressFile_orDie(const char* fname, const char* outName, int cLeve
               "Impossible: zstd only returns 0 when the input is completely consumed!");
     }
 
+    if (isEmpty) {
```
Cyan4973: Thanks for looking into that case @cybojanek !

I was wondering if the same objective could be achieved without a dedicated second code block.

Looking into at the existing code, it seems it could almost work, if : 
- `while()` becomes `for(;;)` , hence is always true
- Loop always start with `read = fread_orDie()`
- `lastChunk = (read < toRead)` remains valid, and thus encompass `read==0`
- `read==0` also suggests a potential read error, requiring a check with ferror()`
- `if (lastChunk) break;` becomes the exit condition

It seems it would cover the "empty file" case, while being a shorter solution, keeping most of existing code in place.

What do you think ?

cybojanek: Simplified according to your suggestion. I think the check for an error in `read==0` is handled by `fread_orDie` which checks for `feof(file)` before it returns.

Cyan4973: Yep, you are right @cybojanek !

Issue #1637, opened 2019-06-06
[stack usage] Take ZSTD_parameters as a const pointer


terrelln:
The parameters are 40 bytes, so for each call site that takes them we can shave off 32-40 bytes by taking them as a const-pointer instead. See https://lkml.org/lkml/2019/5/10/317.

PR #1771, opened 2019-09-05
Passing ZSTD_CCtx_params by const pointer

vnair81:
Static analyzer reported issue of performance inefficiency due to big parameter passed by value (`ZSTD_CCtx_params`).
Size of the parameter is reported as 144 bytes on 64-bit, thus passing by const pointer saves 136 bytes on the stack on each function call and potentially provides a small performance boost due to avoided copy.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -239,9 +239,9 @@ size_t ZSTD_CCtxParams_init_advanced(ZSTD_CCtx_params* cctxParams, ZSTD_paramete
 /* ZSTD_assignParamsToCCtxParams() :
  * params is presumed valid at this stage */
 static ZSTD_CCtx_params ZSTD_assignParamsToCCtxParams(
-        ZSTD_CCtx_params cctxParams, ZSTD_parameters params)
+        const ZSTD_CCtx_params* cctxParams, ZSTD_parameters params)
 {
-    ZSTD_CCtx_params ret = cctxParams;
+    ZSTD_CCtx_params ret = *cctxParams;
     ret.cParams = params.cParams;
     ret.fParams = params.fParams;
     ret.compressionLevel = ZSTD_CLEVEL_DEFAULT;   /* should not matter, as all cParams are presumed properly defined */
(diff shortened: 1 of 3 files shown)


Cyan4973:
cc @felixhandte, as we were discussing this exact issue yesterday, 
and I vaguely remember there was a problem in an equivalent kernel community patch, but your PR might have successfully dodged it, so let's make a review.

felixhandte:
The issue with the linux kernel patch was that they had extended the change to the public API, changing how `ZSTD_compressCCtx()` takes its cparams. I see that this PR does not do that.

Cyan4973:
I don't expect these changes to impact performance,
but could it be checked, just to be sure ?

vnair81:
@Cyan4973  the linux kernel patch (#1637?) was for `ZSTD_parameters`, while this one targets `ZSTD_CCtx_params`.
Regarding performance, I am solely relying on the static analyzer's recommendation, which has served me well in the past in these kinds of issues. I suspect the performance gain, if any, would be tiny, but the stack saving could be real win here.

Cyan4973:
I'm registering a very small performance decrease with `gcc` v8.3.0 on my test system.
It's very small, within noise margin, so I had to run the test multiple times to extract some reliable average.
At level 1, compression speed of `dev` is ~373 MB/s
while `vnair:dev` is ~371 MB/s .

With `clang` v8.0.0, the situation is reversed,
with `dev` at ~380 MB/s,
while `vnair:dev` reaches ~381.5 MB/s .

So that's not a very clear signal.

Ideally, we would like no platform to suffer a performance loss.
But it may be hard to spot where the issue could lie for `gcc`...

Cyan4973:
Looking again at the PR code, I don't see any good reason for these performance changes.

I don't think that the measurements are incorrect : I tested multiple corpuses, they were consistent in the proportion and direction of measured impact. Rather, I suspect that differences come from other factors, such as a slightly different instruction layout, with knock-off impact on uop cache, essentially a random ingredient not controllable from `C` side.

So we are mostly left with the stack savings, which should be consistently positive (at worst neutral when inlining kicks in).

Note that it's not _always_ better to pass by reference, as suggested by this recent article : 
https://lemire.me/blog/2019/09/05/passing-integers-by-reference-can-be-expensive/

But you targeted a large enough structure (`ZSTD_CCtx_params`), for which passing by reference is a reasonable choice.

Issue #1334, opened 2018-09-22
Extending the option -o to directories?


svenha:
I have some use cases that require many zstd-decompress invocations: the directory compressed/ contains 20m files. In regular intervals, I have to decompress around 2m files to the directory uncompressed/.
I do this with 2 million invocations of `zstd-decompress -d -o uncompressed/f f.zst`
How about something like:
`xargs -a my-2m-files | zstd-decompress -d -o uncompressed`
Thanks for your time.

Cyan4973:
It's an interesting feature request.
Clearly, we don't have it yet.

It's not trivial to implement either, especially as this proposal conflicts with current definition of `-o`, which defines a destination __file__ name.
Now, to re-use the same command, we would need the behavior to differ, depending on the existence of `-o` target and whether it is a directory or not. This could introduce some confusion. Typically : 

- `zstd -d someFiles*.zst -o someNonexistingName` => decompress everything and append into single destination file `someNonexistingName`
- `zstd -d someFiles*.zst -o someExistingFile` => prompt for overwrite, if yes, decompress everything and append into single destination file `someExistingFile`
- `zstd -d someFiles*.zst -o someExistingDirectory` => Now : error out. Current proposal : decompress each file separately into directory `someExistingDirectory` using their original name

We can already guess some confusion scenario : 
- If a user wants to decompress all files into some target directory, but does provide a non-existing destination name instead, the result will be a single file, featuring all decompressed contents appended.
- Conversely, if a user wants to decompress a single file at some explicit destination, and it happens that this destination exists and is a directory, the order will be silently converted into "decompress into target directory using existing filename", instead of an error as it behaves today.

An opened question is therefore if creating such confusion scenarios is an acceptable cost for the feature.

An alternative would be to use a different command.
Unfortunately, `-d` and `-D`, which could have been handy to mean "directory", are already in use.
But a relative to `-o`, `-O`, is still available, and could be suitable.

svenha:
The danger of confusion is real as you pointed out. Hence, a new option -O is a good idea.

Cyan4973:
This feature request is in our backlog for next release, either `v1.3.9` or `v1.4.0` depending on user feedback.

svenha:
Thanks for including my feature request. I am really curious how it will speed up my use cases sketched above.

senhuang42:
Hey @svenha ! I'm currently working on this - I was wondering if you had any thoughts for how this should behave?

Basically, there are two options: 
[...]

PR #1772, opened 2019-09-06
[RFC] Adding --output-dir-flat and --output-dir-mirrored to output multiple files into folders

senhuang42:
addresses #1334 
description tbd, not currently looking for serious review since code is def not code-review-level quality just yet - just putting this here to keep track rn

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1388,14 +1388,15 @@ int FIO_compressFilename(FIO_prefs_t* const prefs, const char* dstFileName,
 static const char*
 FIO_determineCompressedName(const char* srcFileName, const char* suffix)
 {
-    if (srcFileName == NULL) {
-        DISPLAYLEVEL(1, "ERROR: Invalid path for --output-dir-mirrored: needs to be downstream from current directory\n");
-        return NULL;
-    }
+    
     static size_t dfnbCapacity = 0;
     static char* dstFileNameBuffer = NULL;   /* using static allocation : this function cannot be multi-threaded */
     size_t const sfnSize = strlen(srcFileName);
     size_t const suffixSize = strlen(suffix);
+    if (srcFileName == NULL) {
+        DISPLAYLEVEL(1, "ERROR: Invalid path for --output-dir-mirrored: needs to be downstream from current directory\n");
+        return NULL;
+    }
     if (dfnbCapacity <= sfnSize+suffixSize+1) {
         /* resize buffer for dstName */
         free(dstFileNameBuffer);
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1422,7 +1422,6 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs, const char** inFileN
                                   const char* suffix, const char* dictFileName,
                                   int compressionLevel, ZSTD_compressionParameters comprParams)
 {
-    printf("compressing multiple...\n");
     int error = 0;
     cRess_t ress = FIO_createCResources(prefs, dictFileName, compressionLevel, comprParams);
 
@@ -1457,7 +1456,7 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs, const char** inFileN
     }   }
 
     FIO_freeCResources(ress);
-    UTIL_freeDestinationFilenameTable(dstFileNamesTable, nbFiles);
+    /*UTIL_freeDestinationFilenameTable(dstFileNamesTable, nbFiles);*/
     return error;
 }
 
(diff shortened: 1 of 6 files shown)


senhuang42:
TODO:
-change to long commands for invoking
-add support for directory structure preservation
-add name collision detection
-make it actually work on windows

programs/util.c
```diff
@@ -87,6 +94,239 @@ U32 UTIL_isDirectory(const char* infilename)
...
+/* function requires posix >= 2001 */
+int UTIL_getRealPath(const char* relativePath, char* absolutePath) {
+#if defined(_MSC_VER) || defined (__MINGW32__) || defined (__MSVCRT__)
+    if (_fullpath(absolutePath, relativePath, LIST_SIZE_INCREASE) == NULL)
+        return 1;
+#elif (PLATFORM_POSIX_VERSION >= 200112L)
+#define _BSD_SOURCE
+    char* r;
+    r = realpath(relativePath, absolutePath);
+    if (r == NULL && errno != ENOENT /* we will be using this to generate "FAKE" paths, so ignore this err */) {
```
felixhandte: Does this work? Do you get the results that you expect if you try to realpath something that doesn't exist? My read of the man page suggests that you won't.

programs/util.c
```diff
@@ -87,6 +94,239 @@ U32 UTIL_isDirectory(const char* infilename)
...
+        errno = 0;
+        return 1;
+    }
+#else
+    UTIL_DISPLAYLEVEL(1, "Unable to process realpath due to platform limitations\n", relativePath, absolutePath);
```
felixhandte: You need substitutions for the additional parameters, don't you ("`%s`")?

programs/zstdcli.c
```diff
@@ -136,6 +136,8 @@ static int usage_advanced(const char* programName)
     DISPLAY( " -q     : suppress warnings; specify twice to suppress errors too\n");
     DISPLAY( " -c     : force write to standard output, even if it is the console\n");
     DISPLAY( " -l     : print information about zstd compressed files \n");
+    DISPLAY( " --output-dir-flat dir    : result(s) stored into toplevel of target existing directory `dir`. If multiple files have the same name, -f will cause the final file to be the output, otherwise the first will be output. Requires POSIX >= 2001.12 \n");
```
felixhandte: Rather than describe the platform requirements here, simply choose whether to display this line at all depending on whether zstd was build with support for this functionality.

programs/fileio.c
```diff
@@ -1392,12 +1388,15 @@ int FIO_compressFilename(FIO_prefs_t* const prefs,
...
 {
+    
     static size_t dfnbCapacity = 0;
     static char* dstFileNameBuffer = NULL;   /* using static allocation : this function cannot be multi-threaded */
-
     size_t const sfnSize = strlen(srcFileName);
     size_t const suffixSize = strlen(suffix);
-
+    if (srcFileName == NULL) {
+        DISPLAYLEVEL(1, "ERROR: Invalid path for --output-dir-mirrored: needs to be downstream from current directory\n");
```
felixhandte: Why? I thought we were going to require that all **inputs** be below the cwd, but not the output.

programs/util.c
```diff
@@ -20,6 +20,13 @@ extern "C" {
 #include <errno.h>
 #include <assert.h>
 
+#if defined(_MSC_VER) || defined(__MINGW32__) || defined (__MSVCRT__)
+#include <direct.h>     /* needed for _mkdir in windows */
+#define currDir _getcwd
```
felixhandte: Why this vs. a `UTIL_getcwd()` wrapper?

programs/util.c
```diff
@@ -87,6 +94,239 @@ U32 UTIL_isDirectory(const char* infilename)
...
+    c = '\\';
+    #endif
+    result = 0;
+
+    /* approach is to start from the first folder in path, and iteratively try to construct a dir at each '/' */
+    for (ptr = strchr(path+1, c); ptr; ptr = strchr(ptr+1, c)) {
+        *ptr = '\0';
+        result = UTIL_createDir((const char*) path);
+        if (result) {
+            UTIL_DISPLAYLEVEL(9, "Unsuccessful directory creation: either already exists or system error\n");
```
felixhandte: This is inaccurate, I think? `UTIL_createDir()` succeeds when the directory already exists.

programs/fileio.c
senhuang42: Yeah that also makes sense - I'll switch it over.

programs/util.c
senhuang42: Should be fixed in the latest version.

programs/util.c
senhuang42: Totally - was thinking of mkdir.

Cyan4973:
With `--output-dir-flat` merged, I guess we keep this PR out of the loop.
It may serve as inspiration source for a future `--output-dir-mirror` command.

PR #1773, opened 2019-09-06
Removing redundant condition in decompression, making first block rle…

bimbashrestha:
Removing redundant condition in decompression, making first block rles valid to deocmpress


Cyan4973:
There are 2 parts combined in this patch : 
- make the CLI decompression able to decode frames with a single RLE block.
- allow the libzstd compression to generate such frame.

I believe the second part is not wanted.

Even if, after this patch, the CLI becomes able to decode single-rle-block frames, _older_ versions are still unable to, and they are deployed everywhere. These versions are the one we want to protect by forbidding the compressor to generate such frames.

Instead, update the comment, in order to indicate that the issue is for `zstd <= v1.4.3` only. This will make it possible, in some long-distance future, to remove this restriction from the encoder, once it feels reasonable that "everyone should have a more recent version".
That will take time ....

lib/compress/zstd_compress.c
```diff
@@ -2305,6 +2307,11 @@ static size_t ZSTD_compressBlock_internal(ZSTD_CCtx* zc,
             zc->bmi2);
 
     if (frame &&
+        /* We don't want to emit our first block as a RLE even if it qualifies because
+         * doing so will cause the decoder to throw a "should consume all input error."
```
Cyan4973: minor: mentions that it's only for the __CLI__ decoder.
This is a very big difference compared to the `libzstd` decoder.
This makes it more probable that we may remove this restriction in some (distant) future.

Cyan4973:
Oh, I forgot,
could we add one test for this decoding of 1st RLE block ?
It should fail with previous version of `zstd` cli, and work with this one.
More importantly, it should be part of our CI test suite, so that we can ensure the property remains valid in the future.

bimbashrestha:
@Cyan4973 I was using https://github.com/facebook/zstd/blob/e9c0fc12d28a2d0c826752e0fd73c8eb0074983f/tests/playTests.sh#L371 to reproduce the decompression bug so the test case should already be covered I believe.

Cyan4973:
The existing test makes a full round-trip, 
and since the compressor cannot generate the 1st block RLE condition,
it cannot test that the decoder supports it.

bimbashrestha:
@Cyan4973  Ah you are right:) I will add it

Cyan4973:
One remaining issue on FreeBSD, as the version of `head` there accepts exact numbers only.

Cyan4973:
Thanks @bimbashrestha ! Looks good to me !

Issue #1723, opened 2019-08-15
[fuzz] Improve the data generation


terrelln:
The fuzzers currently use the first 4 bytes of the input as a seed for the random number generator. See [FUZZ_seed()](https://github.com/facebook/zstd/blob/dev/tests/fuzz/fuzz_helpers.h#L61), [FUZZ_rand*()](https://github.com/facebook/zstd/blob/dev/tests/fuzz/fuzz_helpers.h#L71), and an example of how it is used in the [simple_decompress.c](https://github.com/facebook/zstd/blob/dev/tests/fuzz/simple_decompress.c) fuzzer.

This isn't great for the fuzzer because when it mutates the seed, it changes all the decisions at once. It can't change a single decision at a time.

Instead, we should consume bytes from the end of the input for our decisions, and keep each decision separate.

PR #1777, opened 2019-09-09
[Fuzz] Improve data generation #1723

darxsys:
Addressing the fuzz data generation by following the lz4 approach of reading data from the end to generate parameters, instead of only hashing the first few bytes and relying on the hash to be the seed.

Only changing one test in this PR to make sure it's the right way to do it. I've tested locally and the test passes.

### DIFF
--- a/tests/fuzz/fuzz_data_producer.h
+++ b/tests/fuzz/fuzz_data_producer.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ */
+
+/**
+ * Helper APIs for generating random data from input data stream.
+ */
+
+#ifndef FUZZ_DATA_PRODUCER_H
+#define FUZZ_DATA_PRODUCER_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "fuzz_helpers.h"
+
+/* Struct used for maintaining the state of the data */
+typedef struct FUZZ_dataProducer_s FUZZ_dataProducer_t;
+
+/* Returns a data producer state struct. Use for producer initialization. */
+FUZZ_dataProducer_t *FUZZ_dataProducer_create(const uint8_t *data, size_t size);
+
+/* Frees the data producer */
+void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer);
+
+/* Returns value between [min, max] */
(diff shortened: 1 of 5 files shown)


Cyan4973:
Thanks @darxsys .

While your solution works, it also has the subtle issue that it mixes content to decompress and parameters : in order to change a parameter, the fuzzer must modify some end bytes, but in the process, it also modify the content to decompress.

Double usage of bits is something what we want to avoid in this PR : the goal is to give more accurate control to the fuzzer.

Requiring a certain amount of input to provide parameters means the input stream might be too small to populate all parameters _and_ still have some input left to decompress. Provide and document a clean behavior to this situations.

Finally, I would also suggest to only do only 1 decompression per invocation, as opposed to 10 in current code. I think the intention was to improve performance. doing more work per invocation, since each invocation has some startup cost, but it goes against the objective of better control by the fuzzer, which has higher priority today.

terrelln:
Yeah, we should switch to 1 decompression instead of 10.

We did 10 to allow the fuzzer a better chance to "get lucky" with the decompressed size. Now that it has explicit control, we only want to do 1.

tests/fuzz/simple_decompress.c
```diff
@@ -23,23 +23,26 @@ static ZSTD_DCtx *dctx = NULL;
...
-
-    FUZZ_dataProducer_free(producer);
+  int i;
+  if (!dctx) {
+      dctx = ZSTD_createDCtx();
+      FUZZ_ASSERT(dctx);
+  }
+
+  size_t const bufSize = FUZZ_dataProducer_uint32Range(producer, 0, 2 * size);
+  void* rBuf = malloc(bufSize);
```
darxsys: Nit: this should likely be: `void *rBuf = malloc(bufSize);`

tests/fuzz/fuzz_data_producer.h
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ */
+
+/**
+ * Helper APIs for generating random data from input data stream.
```
terrelln: Nit: Can you add a longer description of what this does?
* How does it generate data at a high level?
* What does it do when it is out of bytes?

tests/fuzz/simple_decompress.c
```diff
@@ -17,26 +17,32 @@
...
-        ZSTD_decompressDCtx(dctx, rBuf, bufSize, src, size);
-        free(rBuf);
-    }
+  int i;
+  if (!dctx) {
+      dctx = ZSTD_createDCtx();
+      FUZZ_ASSERT(dctx);
+  }
+
+  size_t const bufSize = FUZZ_dataProducer_uint32Range(producer, 0, 2 * size);
```
terrelln: Lets make this `10 * size` to give it more opportunity to decompress very compressible data.

PR #1780, opened 2019-09-10
Avoid Clearing Tables Even When Changing CParams

felixhandte:
This changeset builds on the work in #1712, and uses the new workspace abstraction introduced in that PR to perform more efficient context resets, even when the compression parameters change.

Overall this PR should produce compression speed wins for repeated compressions on contexts where the parameters vary (whether because the compression level changes, or the size of the input changes), and be otherwise perf neutral. The gains are very situational, and range from 0% to 4%. There are also outsize wins that this PR brings from fixing existing pathological cases, like constantly switching compression levels on streaming compression of small responses without a pledged src size, which range up 40x improvements.

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -55,7 +55,23 @@ MEM_STATIC void MEM_check(void) { MEM_STATIC_ASSERT((sizeof(size_t)==4) || (size
 #endif
 
 #if defined (MEMORY_SANITIZER)
-#  include <sanitizer/msan_interface.h>
+/* Not all platforms that support msan provide sanitizers/msan_interface.h.
+ * We therefore declare the functions we need ourselves, rather than trying to
+ * include the header file... */
+
+#include <stdint.h> /* intptr_t */
+
+/* Make memory region fully initialized (without changing its contents). */
+void __msan_unpoison(const volatile void *a, size_t size);
+
+/* Make memory region fully uninitialized (without changing its contents).
+   This is a legacy interface that does not update origin information. Use
+   __msan_allocated_memory() instead. */
+void __msan_poison(const volatile void *a, size_t size);
+
+/* Returns the offset of the first (at least partially) poisoned byte in the
+   memory range, or -1 if the whole range is good. */
+intptr_t __msan_test_shadow(const volatile void *x, size_t size);
 #endif
 
 #if defined (MEMORY_SANITIZER)
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -47,6 +47,23 @@ extern "C" {
 #define MEM_STATIC_ASSERT(c)   { enum { MEM_static_assert = 1/(int)(!!(c)) }; }
 MEM_STATIC void MEM_check(void) { MEM_STATIC_ASSERT((sizeof(size_t)==4) || (sizeof(size_t)==8)); }
 
+/* detects whether we are being compiled under msan */
+#if defined (__has_feature)
+#  if __has_feature(memory_sanitizer)
+#    define MEMORY_SANITIZER 1
+#  endif
+#endif
+
+#if defined (MEMORY_SANITIZER)
+#  include <sanitizer/msan_interface.h>
+#endif
+
+#if defined (MEMORY_SANITIZER)
+#  define MEM_SKIP_MSAN __attribute__((no_sanitize("memory")))
+#else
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1715,12 +1638,14 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
...
         params.cParams.windowLog = windowLog;
         FORWARD_IF_ERROR(ZSTD_resetCCtx_internal(cctx, params, pledgedSrcSize,
-                                                 ZSTDcrp_noMemset, zbuff));
+                                                 ZSTDcrp_leaveDirty, zbuff));
         assert(cctx->appliedParams.cParams.strategy == cdict_cParams->strategy);
         assert(cctx->appliedParams.cParams.hashLog == cdict_cParams->hashLog);
         assert(cctx->appliedParams.cParams.chainLog == cdict_cParams->chainLog);
     }
 
+    ZSTD_cwksp_mark_tables_dirty(&cctx->workspace);
```
terrelln: Shouldn't this already be done as part of `ZSTD_reset_matchState()`. Should we move this to an assert that they are marked dirty?

lib/compress/zstd_compress.c
```diff
@@ -1798,24 +1726,29 @@ static size_t ZSTD_copyCCtx_internal(ZSTD_CCtx* dstCCtx,
...
-                                ZSTDcrp_noMemset, zbuff);
+                                ZSTDcrp_leaveDirty, zbuff);
         assert(dstCCtx->appliedParams.cParams.windowLog == srcCCtx->appliedParams.cParams.windowLog);
         assert(dstCCtx->appliedParams.cParams.strategy == srcCCtx->appliedParams.cParams.strategy);
         assert(dstCCtx->appliedParams.cParams.hashLog == srcCCtx->appliedParams.cParams.hashLog);
         assert(dstCCtx->appliedParams.cParams.chainLog == srcCCtx->appliedParams.cParams.chainLog);
         assert(dstCCtx->blockState.matchState.hashLog3 == srcCCtx->blockState.matchState.hashLog3);
     }
 
+    ZSTD_cwksp_mark_tables_dirty(&dstCCtx->workspace);
```
terrelln: Same here

lib/compress/zstd_cwksp.h
```diff
@@ -256,9 +261,39 @@ MEM_STATIC void* ZSTD_cwksp_reserve_object(ZSTD_cwksp* ws, size_t bytes) {
     }
     ws->objectEnd = end;
     ws->tableEnd = end;
+    ws->tableValidEnd = end;
     return start;
 }
 
+MEM_STATIC void ZSTD_cwksp_mark_tables_dirty(ZSTD_cwksp* ws) {
```
terrelln: Can we have a flag like `ZSTD_c_randomizeDirtyTables` and when it is set to true we will randomize the dirty tables?

Then we can sometimes enable it in `zstreamtest`, `fuzzer`, and in the oss-fuzz `FUZZ_setRandomParameters()`.

felixhandte: As discussed offline, I'm going to hook this mechanism into msan.

lib/common/mem.h
```diff
@@ -47,6 +47,23 @@ extern "C" {
...
 
+/* detects whether we are being compiled under msan */
+#if defined (__has_feature)
+#  if __has_feature(memory_sanitizer)
+#    define MEMORY_SANITIZER 1
+#  endif
+#endif
+
+#if defined (MEMORY_SANITIZER)
+#  include <sanitizer/msan_interface.h>
```
terrelln: Sadly, when compiling with `clang.par` and `clang++.par` on my devserver, it wasn't able to find this header. But, it did have the symbols, so declaring the functions we need here would work.

Repro:
```
sudo yum install fb-tp2-compilers
CC=clang.par CXX=clang.par ./fuzz.py build all --enable-fuzzer --enable-msan
```

felixhandte: Oof. That sucks. Building on my desktop (clang-6.0 on Ubuntu 18.04.2) fails if I don't include the header (the symbols are not defined, and we disallow implicit definition). Not sure what an appropriate test is.

felixhandte: Fixed by declaring the functions we need.

PR #1781, opened 2019-09-10
Improve data generation

darxsys:
In this PR, I've moved all tests to use the new producer to generate random parameters. I am using Nick's idea where for each test, we first generate a random number to decide how many bytes to give to the producer, and leave the rest to (de)compression.


tests/fuzz/block_round_trip.c
```diff
@@ -52,10 +52,15 @@ static size_t roundTripTest(void *result, size_t resultCapacity,
 
 int LLVMFuzzerTestOneInput(const uint8_t *src, size_t size)
 {
-    size_t neededBufSize;
+    /* Give a random portion of src data to the producer, to use for
+    parameter generation. The rest will be used for (de)compression */
+    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(src, size);
+    size_t producerSliceSize = FUZZ_dataProducer_uint32Range(producer, 0, size);
+    size = FUZZ_dataProducer_contract(producer, producerSliceSize);
```
terrelln: When we can make all the decisions up front, like in this case and some of the other simpler fuzzers, we don't need to split the data in two. We can just use the `remainingBytes()` function.

That will make the fuzzers job slightly easier, but if it is verbose/ugly to go that route it is fine to keep this method.

tests/fuzz/block_round_trip.c
```diff
@@ -52,10 +52,15 @@ static size_t roundTripTest(void *result, size_t resultCapacity,
 
 int LLVMFuzzerTestOneInput(const uint8_t *src, size_t size)
 {
-    size_t neededBufSize;
+    /* Give a random portion of src data to the producer, to use for
+    parameter generation. The rest will be used for (de)compression */
+    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(src, size);
+    size_t producerSliceSize = FUZZ_dataProducer_uint32Range(producer, 0, size);
```
terrelln: Lets combine these two calls into one function like:

```
size = FUZZ_dataProducer_reserveDataPrefix(producer);
```

tests/fuzz/block_round_trip.c
darxsys: I did it this way for 2 reasons: consistency and safety. Consistency makes it easier to see what our preferred/general approach is no matter where you look. Safety is really about someone some day potentially making changes to existing tests and introducing new places where we use the producer to generate data, after the data has already been used for (de)compression. If the author or reviewer then don't notice this new instance of mixing data for producing and compression, we have a slight problem. Therefore I feel like it's nice to "force" all these tests to just use the same approach. If you feel differently, I am okay with reverting this to previous approach, too.

tests/fuzz/block_round_trip.c
darxsys: Nice, thanks!

tests/fuzz/block_round_trip.c
terrelln: Yeah, I get that. I'm torn because that makes sense to me, but it does make the fuzzers job a little bit harder.

Could you run a fuzzer like `simple_compress_fuzzer` with and without this, and see if it gains coverage about as quickly either way? You will need several runs because it is non-deterministic. It doesn't need to be very scientific, just if it seems about the same we can leave it.

Issue #754, opened 2017-07-11
Files with unknown file extensions shouldn't be ignored


dchest:
I tried extracting an archive with `.zstd` extension and got this error:

```
$ zstd -d Mail.zstd
zstd: Mail.zstd: unknown suffix (.gz/.xz/.zst/.lzma/.lz4 expected) -- ignored
```

This is unexpected and unlike most other command-line tools — in my opinion, it's not the program's business to tell me how to name my files, especially since it can tell what kind of file it is by reading the magic string.

PS Thank you for the great file compressor.

mcmilk:
This is the exact the exact `gzip` behavior... so this should be okay.

```bash
[mcmilk@latitude ~/tmp]$ echo 12345 > testing
[mcmilk@latitude ~/tmp]$ cat testing 
12345
[mcmilk@latitude ~/tmp]$ cat testing | gzip > testing.x
[mcmilk@latitude ~/tmp]$ gzip -d testing.x > testing2
gzip: testing.x: unknown suffix -- ignored
[mcmilk@latitude ~/tmp]$ 
```
Nearly the same for `bzip2` ... it's just using some filename it guesses:
```bash
[mcmilk@latitude ~/tmp]$ cat testing|bzip2 > testing.x
[mcmilk@latitude ~/tmp]$ bzip2 -d testing.x > testing2
bzip2: Can't guess original name for testing.x -- using testing.x.out
[mcmilk@latitude ~/tmp]$ cat testing.x.out
12345
```

So this is okay for zstd also. If you want to use such filenames, then just use `zstdcat` - this will work as you want.

dchest:
Ha, interesting, thanks for pointing that out. So for the sake of consistency, I'm closing this.

terrelln:
You can also specify an output file `zstd -d Mail.zstd -o Mail` should work. When the extension isn't `.zst` we don't know how to derive the output file name.

stokito:
Meanwhile tar just don't care about an extension because it don't need to determine output file name since it already knows it. What is interesting is that gzip also save an original file name so it should also work.
Anyway, could you improve the error message and add a note that output file name should be specified?

chungy:
> What is interesting is that gzip also save an original file name so it should also work.

This kind of metadata has largely been considered a misfeature (reduced determinism in output) and isn't present in other compressors like bzip2 or xz.

stokito:
[...]

PR #1790, opened 2019-09-14
Add short tar's extensions .tgz (.tar.gz), .txz (.tar.xz), .tzst (.tar.zst)

stokito:
The `tar` supports short file extensions like `.tgz` instead of `.tar.gz`. It also supports `.tzst` extension but when I tried to decompress an archive with the `tzst` extension I got an error `unknown suffix` like in #754.
So here is an improvement for the error message itself and adding the missing `tzst` extension.


felixhandte:
This seems like a sensible direction. Right now, though, this seems incomplete. If I understand correctly, this will decompress a file like "`foo.tzst`" into a file named "`foo`", not "`foo.tar`". Would you consider adding that additional functionality?

stokito:
Yes, I'll add the tar extension.

stokito:
Ok, I added a simplest possible solution: if the suffix starts with `t` that means that suffix is one of the `tgz`, `txz`, `tlz4`, `tzst` so we'll add `.tar` to the end of destFileName.
But it has one problem: if in the feature will be created a new compressor with a name starts from `t` then here it may be false recognized as a tar's short extension.
But I guess this unlikely to happen or even if it happen then it will be easy to fix.

programs/fileio.c
```diff
@@ -2220,8 +2210,18 @@ FIO_determineDstName(const char* srcFileName)
 
     /* return dst name == src name truncated from suffix */
     assert(dstFileNameBuffer != NULL);
-    memcpy(dstFileNameBuffer, srcFileName, sfnSize - suffixSize);
-    dstFileNameBuffer[sfnSize-suffixSize] = '\0';
+    size_t dstFileNameEndPos = sfnSize - suffixSize;
```
felixhandte: You're failing the contbuild because you declare this variable after other statements, which is not allowed in C90.

programs/fileio.c
```diff
@@ -2169,43 +2169,33 @@ FIO_determineDstName(const char* srcFileName)
...
-        #ifdef ZSTD_LZMADECOMPRESS
-            "/" XZ_EXTENSION "/" LZMA_EXTENSION
-        #endif
-        #ifdef ZSTD_LZ4DECOMPRESS
-            "/" LZ4_EXTENSION
-        #endif
-        ;
-        DISPLAYLEVEL(1, "zstd: %s: unknown suffix (%s expected) -- ignored \n",
-                     srcFileName, suffixlist);
+        || (strstr(SUFFIX_LIST, suffixPtr) == NULL)) {
```
felixhandte: Won't this also accept a suffix like `.z` (i.e., a prefix of a valid extension)?

felixhandte:
@stokito, I'd like to merge this changeset! Do you plan to add the requested fixes in the next couple of days? Otherwise, I can fix it up myself.

stokito:
I hope that I'll do this on Wednesday, sorry for delay

felixhandte:
Merging into temporary feature branch to fix up.

PR #1793, opened 2019-09-16
Removing assert added when handling related oss fuzz issue

bimbashrestha:
We added this last week when oss-fuzz found a bug in flushBitsFast(). Line 265 handles the overflow for this function unlike flushBitsFast so I think we don't need the assert?


Cyan4973:
The `assert()` seems correct. Try to find what could break it first.

Cyan4973:
I may have an explanation for that one. Let's discuss it when you have time.

PR #1795, opened 2019-09-16
Add Poisoned Redzones to the Workspace When Compiling with ASAN

felixhandte:
Adds 8-byte spaces on either side of allocations into the workspace (which requires touching all of the places we calculate the size of the workspace), and poisons them when we run under ASAN. This allows us to verify that accesses into the individual objects in the workspace don't bleed over into each other.

This means removing any assumptions made in the rest of the codebase concerning contiguity of objects, which I kind of view as a side benefit...

Note that we do not put redzones around the table allocations, because that memory region is complicated... We depend on the contiguity of the tables to know that the compressor is maintaining the invariants around the values in the [objectEnd, tableEnd) range, while the workspace is responsible for the [tableEnd, tableValidEnd) range.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1107,7 +1107,9 @@ size_t ZSTD_estimateCCtxSize_usingCCtxParams(const ZSTD_CCtx_params* params)
         size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << cParams.windowLog);
         U32    const divider = (cParams.minMatch==3) ? 3 : 4;
         size_t const maxNbSeq = blockSize / divider;
-        size_t const tokenSpace = ZSTD_cwksp_alloc_size(WILDCOPY_OVERLENGTH + blockSize + 11*maxNbSeq);
+        size_t const tokenSpace = ZSTD_cwksp_alloc_size(WILDCOPY_OVERLENGTH + blockSize)
+                                + ZSTD_cwksp_alloc_size(maxNbSeq * sizeof(seqDef))
+                                + 3 * ZSTD_cwksp_alloc_size(maxNbSeq * sizeof(BYTE));
         size_t const entropySpace = ZSTD_cwksp_alloc_size(HUF_WORKSPACE_SIZE);
         size_t const blockStateSpace = 2 * ZSTD_cwksp_alloc_size(sizeof(ZSTD_compressedBlockState_t));
         size_t const matchStateSize = ZSTD_sizeof_matchState(&cParams, /* forCCtx */ 1);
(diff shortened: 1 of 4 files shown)


felixhandte:
Performance figures look unchanged, as expected.

lib/compress/zstd_compress.c
```diff
@@ -130,24 +130,23 @@ static void ZSTD_freeCCtxContent(ZSTD_CCtx* cctx)
...
     ZSTD_clearAllDicts(cctx);
 #ifdef ZSTD_MULTITHREAD
     ZSTDMT_freeCCtx(cctx->mtctx); cctx->mtctx = NULL;
 #endif
+    ZSTD_cwksp_free(&cctx->workspace, cctx->customMem);
 }
 
 size_t ZSTD_freeCCtx(ZSTD_CCtx* cctx)
 {
+    int cctxInWorkspace = ZSTD_cwksp_owns_buffer(&cctx->workspace, cctx);
```
terrelln: This must be moved after the `NULL` check.

terrelln:
This looks reasonable to me. The allocation size code is still ugly... But it is no worse than before.

Cyan4973:
The patch seems to be in a territory of debatable benefit / complexity trade-off.

__Benefit__ : poisoned redzones are supposed to help detect out-of-bound accesses.
This seems useful in this case, since we have a single memory object, which is internally divided into multiple objects. So an out-of-bound access to an internal table may end up into a different table, and therefore still remain valid, hence undetected from a `malloc`+`asan` perspective.

However, what can redzones detect ?
Adding a small poisoned space between tables makes it possible to detect an off-by-one error.
So, if a table has a size `S`, accessing `table[S]` will get right into the red zone, and trigger `asan`.

However, a wild out-of-bound access, like `table[2*S]`, simply  jumps over the red zone, and may end up in another (valid) table area. This is all the more probable as tables are gigantic while red zones are tiny.

Therefore, red zones are somewhat useful, but don't catch all out-of-bound accesses. How useful is that ? I have no idea. In my view, it does half the expected job, which is still something.

__Complexity__ : well, if the complexity was all contained in a single place, I think this would be okay.
But that's not completely the case. We have 4 files modified. 
2 of them are `*.h`, and they seem like the right place for this change (`asan` function declaration, and new allocation mechanism).
The 2 `*.c` are paying the consequence of this new allocation strategy, and need to adapt their code. `resetCCtx` is the most impacted, but hardly the only one. The changes are numerous, though they feel relatively well contained : 2 files is not that large.

Also, I'm just not completely sure if that's really the end of it, and how we can ensure that all consequences are properly accounted for.
What about `ZSTD_sizeof_CCtx()` for example ?

Bottom line : there is a detection benefit, but it's not that large. There is an associated complexity cost, which is clearly not negligible, but also not intractable, hence difficult to balance.

Cyan4973:
Question :
what would be the complexity of allocating one memory area per object (instead of a single workspace which is then split among objects) when compiling with `asan` mode ?

The benefit is that such an approach should catch most out-of-bound accesses when combined with `asan`, beyond off-by-one errors.

A disadvantage is that some type of problems, typically related to context re-use while also changing object sizes, can no longer be re-created in this new environment. Therefore, there is also a need to be able to run `asan` with the "normal" code (single workspace).

PR #1803, opened 2019-09-19
[zstd] Fix and speed up wildcopy

terrelln:
Fix a memory over read introduced by PR #1756 by increasing `WILDCOPY_OVERLENGTH` to 16. This means that we can avoid some of the branches added, and simplify wildcopy.

Preliminary results show 3+% faster decompression on gcc and clang, and maybe faster compression as well. More measurements to come.

Will test more to ensure that `WILDCOPY_OVERLENGTH` is respected everywhere, and the constant `8` isn't used.

Credit to OSS-Fuzz.

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -207,7 +207,13 @@ typedef enum {
 } ZSTD_overlap_e;
 
 /*! ZSTD_wildcopy() :
- *  custom version of memcpy(), can overwrite up to WILDCOPY_OVERLENGTH bytes (if length==0) */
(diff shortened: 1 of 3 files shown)


felixhandte:
Love the branch name.

terrelln:
This check needs to be fixed (and probably some others).

https://github.com/facebook/zstd/blob/b83059958246dfcb5b91af9c187fad8c706869a0/lib/decompress/zstd_decompress_block.c#L691-L700

terrelln:
This has now passed ~1 hour of fuzz tests, so its ready for review. Any further fuzz failures should quickly be found by oss-fuzz.

terrelln:
I'm going to close this PR and roll the rest of my optimizations into the same PR. That way it is easier to measure.

PR #1804, opened 2019-09-20
Optimize (de)compression and fix wildcopy overread

terrelln:
* Bump `WILDCOPY_OVERLENGTH` to 32 to fix the wildcopy overread introduced in #1756.
* Optimize `ZSTD_wildcopy()` by removing unnecessary branches, unrolling the loop, and bumping to a 32 byte copy in stripes of 16.
* Extract `ZSTD_overlapCopy8()` into its own function.
* Add `ZSTD_safecopy()` for `ZSTD_execSequenceEnd()`. It is
  optimized for single long sequences, since that is the important
  case that can end up in `ZSTD_execSequenceEnd()`. Without this
  optimization, decompressing a block with 1 long match goes
  from 5.7 GB/s to 800 MB/s.
* Refactor `ZSTD_execSequenceEnd()`.
* Increase the literal copy shortcut to 16.
* Add a shortcut for offset >= 16.
* Simplify `ZSTD_execSequence()` by pushing more cases into
  `ZSTD_execSequenceEnd()`.
* Delete `ZSTD_execSequenceLong()` since it is exactly the
  same as `ZSTD_execSequence()`.
* Refactor and fix `ZSTD_storeSeq()` to remove the implicit requirement that the literals can be wildcopied.

| Scenario | gcc-9 | clang-8 |
|---|---|---|
| Decompression | +12.4% | +22.8% |
| Level 1 | +3.8% | +5.0% |
| Level 2 | -0.9% | +4.2% |
| Level 3 | +2.4% | +2.7% |
| Level 4 | +1.8% | +3.1% |
| Level 5 | +3.3% | +2.0% |

The numbers are gathered on an Intel i9-9900k with turbo disabled taskset to CPU 0. It is the average percent change over silesia.tar, enwik8, and calgary.tar. The decompression speed is the average of the percent change over levels 1 to 5.

I got very similar results with the fbcode gcc and clang on my devserver for compression and decompression. On OS X gcc sees a 9% decompression speed boost, and clang a 13% speed boost, and compression is neutral.

Edge case single long match:
The file `head -c 128K /dev/zero` triggers a corner case in the new code. Level 1 compression is now 12.7 GB/s (previously 8.8 GB/s), and decompression is neutral at 5.8 GB/s.

Crdit to OSS-Fuzz for finding the wildcopy overread.

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -61,6 +61,13 @@
 #  define HINT_INLINE static INLINE_KEYWORD FORCE_INLINE_ATTR
 #endif
 
+/* UNUSED_ATTR tells the compiler it is okay if the function is unused. */
+#if defined(__GNUC__)
+#  define UNUSED_ATTR __attribute__((unused))
+#else
+#  define UNUSED_ATTR
+#endif
+
 /* force no inlining */
 #ifdef _MSC_VER
 #  define FORCE_NOINLINE static __declspec(noinline)
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -197,7 +197,7 @@ static void ZSTD_copy8(void* dst, const void* src) { memcpy(dst, src, 8); }
 static void ZSTD_copy16(void* dst, const void* src) { memcpy(dst, src, 16); }
 #define COPY16(d,s) { ZSTD_copy16(d,s); d+=16; s+=16; }
 
-#define WILDCOPY_OVERLENGTH 16
+#define WILDCOPY_OVERLENGTH 32
 #define WILDCOPY_VECLEN 16
 
 typedef enum {
@@ -237,10 +237,10 @@ void ZSTD_wildcopy(void* dst, const void* src, ptrdiff_t length, ZSTD_overlap_e 
          * On clang-8 unrolling once is +1.4%, twice is +3.3%, thrice is +3%.
          */
         COPY16(op, ip);
-        if (op >= oend) return;
         COPY16(op, ip);
         if (op >= oend) return;
         do {
+            COPY16(op, ip);
             COPY16(op, ip);
         }
         while (op < oend);
@@ -257,7 +257,7 @@ MEM_STATIC void ZSTD_wildcopy8(void* dst, const void* src, ptrdiff_t length)
     BYTE* op = (BYTE*)dst;
     BYTE* const oend = (BYTE*)op + length;
     do {
-        COPY8(op, ip)
+        COPY8(op, ip);
     } while (op < oend);
 }
 
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -197,8 +197,8 @@ static void ZSTD_copy8(void* dst, const void* src) { memcpy(dst, src, 8); }
 static void ZSTD_copy16(void* dst, const void* src) { memcpy(dst, src, 16); }
 #define COPY16(d,s) { ZSTD_copy16(d,s); d+=16; s+=16; }
 
-#define WILDCOPY_OVERLENGTH 8
-#define VECLEN 16
+#define WILDCOPY_OVERLENGTH 16
+#define WILDCOPY_VECLEN 16
 
 typedef enum {
     ZSTD_no_overlap,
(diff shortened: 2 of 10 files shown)


Cyan4973:
That's a hell of a patch @terrelln !

Quick results on my bench system (Core i7-9700k, gcc 8.3.0) :

silesia : 1312 -> 1400
calgary : 1189 -> 1288
enwik8 : 1059 -> 1152

Slightly less than 10%, still excellent !
I also tested the specific case of `--ultra -22`, since it triggers long distance matches, with resulting caches misses (compensated with prefetching) : 

silesia -22 : 1118 -> 1243
enwik9 -22 : 865 -> 975

This time, this is more than 10% !

With clang 8.0, as announced, the gains are much larger : 

silesia : 1082 -> 1268
calgary : 963 -> 1144
enwik8 : 854 -> 1016

Congratulations !!

davidbolvansky:
And if you remove "dont vectorize" macro and enable -ftree-vectorize again in Makefile, you should see another small again. Seen with gcc9.

davidbolvansky:
Consider 
COPY32(op, ip);

instead of 
COPY16(op, ip);
COPY16(op, ip);

Brings another perf gain.

davidbolvansky:
zstd's master + GCC 9
./zstd -b -d bench/nci.zst -f
 3#nci.zst           :  33553445 ->   2857582 (11.74),  0.00 MB/s ,1295.1 MB/s

This patch + things I mentioned + unrolled building loop in ZSTD_buildFSETable + GCC 9

/zstd -b -d bench/nci.zst -f
 3#nci.zst           :  33553445 ->   2857582 (11.74),  0.00 MB/s ,1645.9 MB/s 

22% percent improvement

terrelln:
Looks like a new assert in `ZSTD_storeSeq()` uncovered an existing bug where there wasn't at least 8 bytes after the literals length for wildcopy.

Cyan4973:
Indeed, the end of buffer condition was previously implicit.

nit-picking here, but the exact condition is that end of buffer must be at least 8 bytes beyond _start_ of literals, and at least **7** bytes beyond end of literals. I don't think it makes a difference but just in case it does, I mention it here.

The implicit end of buffer condition relied on the fact that the last match starts effectively at least 8 bytes before end of block, due to hashing constraints. `repcode` can evade the hashing issue, but I think the same condition is applied to them anyway.

The new code in `decodeCorpus` doesn't respect this condition, since end of buffer is declared right at `literals + litlen`, so without sufficient margin : 
https://github.com/facebook/zstd/pull/1804/files#diff-771ac86c47832895df8cc02374f8da6cR761

However, the errors detected in `travisCI` do not depend on `decodeCorpus`, so there is something else going on.

terrelln:
LDM also doesn't respect this condition, which is where the errors are showing up.

I'm working on rewriting `ZSTD_storeSeq()` to not require any over length, since that cleans up implicit assumptions.

Cyan4973:
Yes, this version is even better.
Testing at level 1, which is likely the most impacted by `ZSTD_storeSeq()`, I now measure a _gain_ in compression speed : 

silesia : 498 -> 523
calgary : 417 -> 434
enwik8 : 365 -> 378

I would even say, the benefits are larger than I expected.

Decompression speed also improves a little bit more compared to latest measurement.

terrelln:
@davidbolvansky I haven't been able to reproduce speed gains of removing `DONT_VECTORIZE` and using `COPY32()` with gcc-9 on an Intel i9-9900K. I will leave them out of this patch, but if you put up a PR with those changes I'll benchmark them.

What do you mean by `ZSTD_buildFSETable()` unrolling, are you planning on putting a PR up for that?

davidbolvansky:
I will measure (my) things again after this patch is merged.

terrelln:
I've fixed the compression perf regression on OS X clang by using `HINT_INLINE` for `ZSTD_storeSeq()`. Clang on Mac wasn't inlining it for level 1...

I've verified that it doesn't change the performance of anything else. I've been running fuzzers for a few hours, and no failures so far. I see good performance on my home server, devserver, and Macbook (detailed in summary).

I'm happy this this code now.

Cyan4973:
Some minor detail left : 
https://travis-ci.org/facebook/zstd/jobs/587703350#L3968

`HINT_INLINE` needs an additional attribute to avoid triggering the `unused-function` warning.

davidbolvansky:
> I will measure (my) things again after this patch is merged.

With this patch:
 ./zstd -b -d bench/nci.zst -f 
 3#nci.zst           :  33553445 ->   2857582 (11.74),  0.00 MB/s ,1611.7 MB/s 

Some my changes:
 ./zstd -b -d bench/nci.zst -f 
 3#nci.zst           :  33553445 ->   2857582 (11.74),  0.00 MB/s ,1632.9 MB/s 

https://pastebin.com/3uumCPrf

PR #1805, opened 2019-09-20
Improve ZSTD_highbit32's codegen

davidbolvansky:
GCC is unable to simplify '31 - __builtin_clz(val)' and produces ugly assembly. Clang can simplify it.

Use alternative expr "__builtin_clz(val) ^ 31" which is optimized well by Clang and GCC too.

Proof: https://godbolt.org/z/y32jQq

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -339,7 +339,7 @@ MEM_STATIC U32 ZSTD_highbit32(U32 val)   /* compress, dictBuilder, decodeCorpus 
         _BitScanReverse(&r, val);
         return (unsigned)r;
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* GCC Intrinsic */
-        return 31 - __builtin_clz(val);
+        return __builtin_clz(val) ^ 31;
 #   elif defined(__ICCARM__)    /* IAR Intrinsic */
         return 31 - __CLZ(val);
 #   else   /* Software version */


Cyan4973:
Great find @davidbolvansky !

terrelln:
Awesome! I'll benchmark this later today, so we know how much we're getting from this.

Cyan4973:
I haven't found any measurable speed difference. Noise is larger than this impact.

The more convincing argument is about the `godbolt` assembly listing, which is pretty straightforward.

terrelln:
@Cyan4973 I may have measured a ~0.5% speed difference on top of my PR. But, I need to measure it more carefully.

terrelln:
We should also update `BIT_highBit32()`.

terrelln:
clang-8 generates exactly the same binary before/after. I've measured a ~0.3% decompression speed boost with gcc-9 (1227 MB/s -> 1232 MB/s on silesia.tar), and I'm fairly confident it isn't just noise.

davidbolvansky:
> We should also update `BIT_highBit32()`.

I included this patch in my other patch: https://github.com/facebook/zstd/pull/1807

So I will close this PR.

>> clang-8 generates exactly the same binary before/after. I've measured a ~0.3% decompression speed boost with gcc-9 (1227 MB/s -> 1232 MB/s on silesia.tar), and I'm fairly confident it isn't just noise.

Thanks, thr results make sense.

Cyan4973:
> I included this patch in my other patch: #1807 . So I will close this PR.

I'm not sure if it's a good idea.
#1807 entangles several optimizations in one package, some of which are bets, and we don't know yet if they are all good or need to be cherry-picked.

In contrast, this little patch targets a single precise optimization, which is presumed to always be positive (gcc) or neutral (clang). It's an easier sell.

davidbolvansky:
Ok, I had to create a new branch anyway, so please follow this patch here: https://github.com/facebook/zstd/pull/1808

PR #1808, opened 2019-09-23
Improve ZSTD_highbit32's codegen

davidbolvansky:
GCC is unable to simplify '31 - __builtin_clz(val)' and produces ugly assembly. Clang can simplify it.

Use alternative expr "__builtin_clz(val) ^ 31" which is optimized well by Clang and GCC too.

Proof: https://godbolt.org/z/y32jQq

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -164,7 +164,7 @@ MEM_STATIC unsigned BIT_highbit32 (U32 val)
         _BitScanReverse ( &r, val );
         return (unsigned) r;
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* Use GCC Intrinsic */
-        return 31 - __builtin_clz (val);
+        return __builtin_clz (val) ^ 31;
 #   elif defined(__ICCARM__)    /* IAR Intrinsic */
         return 31 - __CLZ(val);
 #   else   /* Software version */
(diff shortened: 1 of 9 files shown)


Cyan4973:
This patch takes care of `BIT_highbit32()`,
but doesn't include the earlier update of `ZSTD_highbit32()`.
Is that intentional ?

Cyan4973:
Btw, maybe we should make `ZSTD_highbit32()` just a follow up of `BIT_highbit32()`.
These 2 functions just do the same thing, so better having a single place to optimize and maintain.

davidbolvansky:
Oh, no :[

The pattern in sed missed that place.

davidbolvansky:
> Btw, maybe we should make `ZSTD_highbit32()` just a follow up of `BIT_highbit32()`.
> These 2 functions just do the same thing,
> and `zstd_internal.h` already depends on `bitstream.h` (transitively, through `huff.h` and `fse.h`),
> so better have a single place to optimize and maintain.

Any idea where to move it?

Cyan4973:
> Any idea where to move it?

Well, I was just thinking either
`#define ZSTD_highbit32 BIT_highbit32`
or its equivalent and cleaner function call equivalent.

Don't worry if you are not comfortable doing it, 
I can take care of it after this patch is merged if you prefer.

davidbolvansky:
Ok, agree (I will leave this patch as is).

PR #1812, opened 2019-10-02
Support for --output-dir-flat: output into a directory

senhuang42:
Compress multiple files into one directory. New pull request to separate this from separate recursive directory construction --output-dir-mirrored.


programs/zstdcli.c
```diff
@@ -1163,10 +1175,23 @@ int main(int argCount, const char* argv[])
         if (adaptMin > cLevel) cLevel = adaptMin;
         if (adaptMax < cLevel) cLevel = adaptMax;
 
+        if (outDirName) {
+            if (UTIL_isDirectory(outDirName)) {
+                DISPLAY("Output of files will be in directory: %s\n", outDirName);
+                dstFilenameTable = (char**)malloc(filenameIdx * sizeof(char*));
```
Cyan4973: Always check is a `malloc()` was successful.

programs/fileio.c
```diff
@@ -1453,6 +1453,9 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
     }   }
 
     FIO_freeCResources(ress);
+    if (dstFileNamesTable)
+        UTIL_freeDestinationFilenameTable(dstFileNamesTable, nbFiles);
```
Cyan4973: Programming patterns :
for better code maintenance, it's recommended
to free an object in the same code block as the one where it's created.

```C
{
    TYPE* const object = createObject();
    (...)
    freeObject(object);
}
```

It seems this is possible here too.

programs/zstdcli.c
Cyan4973: Also, what's the purpose of `dstFilenameTable` ?

senhuang42: We use `dstFilenameTable` to hold the output-dir + filename string, such that the file from `filenameTable[i]` will end up being (de)compressed into `dstFilenameTable[i]`.

In `FIO_compressMultipleFilenames()', the usual behavior is to just use `filenameTable` and append or remove '.zst' to get our final path to the file, so in order to output files to a new directory, we need to create a new mapping from original path to new path based on the specified output dir.

Cyan4973: My understanding is as follows :
given a list of files : `{ file1 file2 file3 }`
the algorithm starts by creating a new list (table) of files, prefixing the destination directory in front of each one of them : `{ outdir/file1 outdir/file2 outdir/file3 }`.
This new list is passed to `FIO_compressMultipleFilenames()` which, as usual, generates the final name on the fly : `outdir/file1.zst` .

A question is :
why not generating the directly final name completely on the fly ?
`(outdir, file1) => outdir/file1.zst` .

A possible answer : to reduce modifications inside `FIO_compressMultipleFilenames()`.

But `FIO_compressMultipleFilenames ()` is not exactly free of modifications anyway, and an important downside is : now it's necessary to generate a (potentially large) table with names stored on heap, bringing with it the corresponding responsibility to manage the memory, aka to free it properly. It also makes interfaces more complex, with lifetime issues to document.

This seems more complex and, given a long-term maintenance perspective, more dangerous than the other solution, which concentrates the topic in a single place, thus largely freeing the rest of the system from having to deal with it.

senhuang42: Yep that was definitely my initial approach - minimize my code footprint overall. However, I also see the merits in modifying FIO_compressMultipleFilenames() some more in favor of introducing a new `dstFilenameTable`. 

I think the approach then is to modify my function that takes a table and generates a table into a function that takes a `srcFileName` and `outDirName`, returns the final path, and call that in FIO_compressMultipleFilenames() at some point to modify our final file destination.

Cyan4973: Do you even need a `table` at all ?

senhuang42: Nope! Not using new table at all in the new implementation, we only make a table of pointers when we need to check whether or not it there are name collisions (our approach is sorting the table and then checking for consecutive matches).

programs/fileio.c
```diff
@@ -628,6 +628,104 @@ static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName)
...
+    return 0;
+}
+
+/* FIO_determineDstFilenameOutdir() :
+ * Takes a source file name and specified output directory, and
+ * allocates memory for and returns a pointer to final path.
+ * This function never returns an error (it may abort() in case of pb)
+ */
+static char*
+FIO_determineDstFilenameOutdir(const char* srcFilename, const char* outDirName, const size_t suffixLen)
```
Cyan4973: minor : naming consistency : 

the result of this function is an allocated buffer.

Because its ownership is now transferred, the new owner must ensure that it will `free()` it when no longer useful.
It's an important responsibility.

To make it clear that such a responsibility is now transferred onto the caller,
the convention in `zstd` code base is to use the word `create` in the name of the function.

In this case, it could be something like : 
`FIO_createFilename_from*()`. In this example, there is still a need to select a meaningful suffix.

senhuang42: Good to know moving forward, I've changed the name to fit convention.

Cyan4973:
Some relatively minor remaining issue, such as : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/builds/27863595/job/63v86s5n5g16gv8g#L269

Just ensure that parameters get same qualifiers in both declaration and definition. Visual is a little picky on these ones.

senhuang42:
Interesting, the C4090 warning appears to be a bug with the VS compiler. We are modifying a non-const table of `const char*`, not the `const char*` itself. So on lines 661 and 670 when we try to `qsort()` and `free()` the table, so it should be a valid operation.

I did some digging and came across this: 
1) https://stackoverflow.com/questions/10403713/why-does-visual-c-warn-on-implicit-cast-from-const-void-to-void-in-c-but
2) https://stackoverflow.com/questions/34347455/const-char-and-free (funny coincidence! looks like you came across this same problem a while ago!)

Seems like we can either:
1) Use some `#pragma` to suppress warning locally for just that particular code block if we're on windows.
2) Use `char** filenameTableSorted` instead of `const char** filenameTableSorted` , which means we must also now allocate space in the heap for strings in `filenameTableSorted` which are copies of strings in `const char** filenameTable`. This is because we can't directly point at `const char*` with `char*`.

I personally prefer option 1) since we avoid allocating on the heap (part of reason why we don't create intermediate destination filename table).

senhuang42:
Probably desirable to only check if filenames are the same if we're using the new flag.

PR #1814, opened 2019-10-04
Adding ZSTDMT_pThreadPool (Thread pool that respects ordering)

bimbashrestha:
Just pushed the header for now so that I can get comments on the API. This abstraction will allow us to design more complex multi-threaded algorithms than is currently supported by the regular thread pool in pzstd. We need this for the coming ZSTDMT_decompress() feature. 

The current thread pool inside pzstd uses a regular FIFO queue to store jobs. When a thread becomes available, the queue is popped and the job that comes out is run. The only notable modification here is to use a priority queue in place of the queue. The time complexity for adding a job will go from O(1) to O(log(#jobs))


lib/common/zstdmt_priority_threadpool.h
```diff
@@ -0,0 +1,59 @@
...
+ * @return : ptr to ZSTDMT_pThreadPoolCtx_t for thread pool */
+static ZSTDMT_pThreadPoolCtx_t* ZSTDMT_pThreadPoolCtx_create(size_t queueSize,
+  size_t nbThreads);
+
+/* Destroies the context */
+static size_t ZSTDMT_pThradPoolCtx_free(ZSTDMT_pThreadPoolCtx_t* ctx);
+
+/* Enqueue a job into the priority queue of the thread pool with given 'priority'
+ * Note: time complexity will be log(#jobs)
+ * Note: lower priorities are going to be excuted first (ie. priority 0 jobs will come first)
```
Cyan4973: Is there a maximum priority value ?

lib/common/zstdmt_priority_threadpool.h
```diff
@@ -0,0 +1,59 @@
...
+ * @return : ID of the job */
```
Cyan4973: What's the purpose of returning the ID of the job ?

lib/common/zstdmt_priority_threadpool.h
Cyan4973: Also : what happens when 2 jobs have the same priority value ?
Is there a specific order guaranteed ?

bimbashrestha: Yes, whatever queueSize is supplied in create(). If every job is a new priority, that will be the max. Or do you mean absolute maximum priority value?

lib/common/zstdmt_priority_threadpool.h
```diff
@@ -0,0 +1,59 @@
...
+static size_t ZSTDMT_pThreadPool_addJob(ZSTDMT_pThreadPoolCtx_t* ctx, size_t priority,
+  size_t addedTimeMs, ZSTDMT_pThreadPool_fn fn, void* data);
+
+/* Dequeue the job from the priority queue of the thread pool if it exists in the queue
+ * @return : 1 for success or some error code */
+static size_t ZSTDMT_pThreadPool_removeJob(ZSTDMT_pThreadPoolCtx_t* ctx, size_t jobId);
```
Cyan4973: We will have to go to the specifics of the algorithm that makes you need a `removeJob` action.
I see several actions below related to some form of "manual" maintenance of the queue,
it's important to study if that's the best way to do it.

lib/common/zstdmt_priority_threadpool.h
bimbashrestha: I wasn't planning on maintaining order when priority values clash. That should be just fine for the way I'm planning on using this in the multi-threaded decoder. Jobs that have the same priority value will not require a particular execution order.

lib/common/zstdmt_priority_threadpool.h
bimbashrestha: I don't see the multi-threaded decoder algorithm using this. Just thought it would be useful to support for future use cases. Perhaps this would be better left to when such a use case comes about?

Cyan4973: Yes, probably.
Without a clear use case, we can't be sure that this functionality is useful, nor if it's the right abstraction.
Try to reduce the API to what you will need.

PR #1819, opened 2019-10-08
Thread pool with dependencies

bimbashrestha:
EDIT: The thread pool api accepts jobs with dependencies. 

-------------

The api is pretty simple here. The thread pool accepts jobs with either no dependencies or one dependency. Still need to do error handling on the pthread methods but pushing this now so that I can get some early feedback. Also need to add better tests than the super simple one I have right now in fuzzer.c.

I'm not handling multiple dependencies because the multi-threaded decoder only needs to handle jobs with one dependency (i.e the previous block)


lib/common/zstdmt_depthreadpool.h
```diff
@@ -0,0 +1,45 @@
...
+ZSTDMT_depThreadPoolCtx_t* ZSTDMT_depThreadPool_createCtx(size_t maxNbJobs, size_t maxNbThreads);
+
+/* Wait until all the jobs queued finish and then frees resources
+ * Destroys all mutexes and cond variables used */
+void ZSTDMT_depThreadPool_destroyCtx(ZSTDMT_depThreadPoolCtx_t* ctx);
+
+/* Adds a job with potential dependencies to the thread pool
+ * @fn: This is the callback function
+ * @data: Any args that you want to pass to your callback function
+ * @independent: 1 if the job has no dependencies, 0 otherwise
```
Cyan4973: Could this field `independent` be folded into `depJobId` ?

lib/common/zstdmt_depthreadpool.h
```diff
@@ -0,0 +1,45 @@
...
+#define ZSTDMT_DEPTHREADPOOL_H_MODULE
+
+#include <stddef.h>
+
+/***************************************
+* Definitions
+***************************************/
+
+typedef void (*ZSTDMT_depThreadPoolFn)(void* data);                          /* Job callback function */
+typedef struct ZSTDMT_depThreadPoolJob_s ZSTDMT_depThreadPoolJob_t;          /* Job management struct */
```
Cyan4973: minor : naming style : 
types should have the first letter in Capital.
Also, try to avoid suffix `*_t` when possible.
So here it would be : ZSTDMT_DepThreadPoolJob.

We have not been very consistent in enforcing these rules in the past, so you can find plenty of counter-examples in the code base. Let's try to make some progresses now.

Cyan4973:
> I'm not handling multiple dependencies because the multi-threaded decoder only needs to handle jobs with one dependency (i.e the previous block)

I believe you need at least 2, even in the "simple" MT variant.

The LZ section of a block depends : 
- on the entropy section of the same block
- on the LZ section of the previous block

bimbashrestha:
@Cyan4973 Agh. you are right:) I'll make the change

lib/common/zstdmt_depthreadpool.h
```diff
@@ -0,0 +1,45 @@
...
+/* Wait until all the jobs queued finish and then frees resources
+ * Destroys all mutexes and cond variables used */
+void ZSTDMT_depThreadPool_destroyCtx(ZSTDMT_DepThreadPoolCtx* ctx);
+
+/* Adds a job with potential dependencies to the thread pool
+ * @fn: This is the callback function
+ * @data: Any args that you want to pass to your callback function
+ * @nbDeps: Number of dependencies
+ * @depJobIds: The dependencies
+ * Note: This only supports having jobs with one dependency.
```
Cyan4973: I guess this comment is no longer up to date ?

lib/common/zstdmt_depthreadpool.c
```diff
@@ -0,0 +1,159 @@
...
+	free(ctx->jobs);
+	pthread_mutex_destroy(&ctx->mutex);
+	pthread_cond_destroy(&ctx->cond);
+	free(ctx);
+}
+
+size_t ZSTDMT_depThreadPool_addJob(ZSTDMT_DepThreadPoolCtx* ctx, ZSTDMT_depThreadPoolFn fn,
+	void* data, size_t nbDeps, size_t* depJobIds)
+{
+	ZSTDMT_DepThreadPoolJob* job = malloc(sizeof(ZSTDMT_DepThreadPoolJob));
```
Cyan4973: check `malloc()` for success.

lib/common/zstdmt_depthreadpool.c
```diff
@@ -0,0 +1,159 @@
...
+	job->fn = fn;
+	job->data = data;
+	job->started = 0;
+	job->finished = 0;
+	job->nbDeps = nbDeps;
+	job->depJobIds = depJobIds;
```
Cyan4973: This is likely to cause problems :
`depJobIds` is a pointer to a table.
I recommend to modify the name of the variable to make that fact clear, as it's easy to lose track of that.

The consequence is that what is copied is the reference to the table, not its content.
But what will happen to these data afterwards ?
If the user `free` the table, or re-purpose the table for other usages, this will trigger an undefined behavior.

Bottom line : ensure that the `jobID` values in the table are well preserved throughout job posting and execution.

tests/fuzzer.c
```diff
@@ -366,6 +370,18 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
     DISPLAYLEVEL(3, "OK \n");
 
+    /* Need to write some better tests for this */
+    DISPLAYLEVEL(3, "test%3u : ZSTDMT_depThreadPool dep test : ", testNb++);
+    {
+      int data = 0;
+      ZSTDMT_DepThreadPoolCtx* ctx = ZSTDMT_depThreadPool_createCtx(2, 2);
+      size_t jobId = ZSTDMT_depThreadPool_addJob(ctx, FUZ_depThreadPool_callbackFn1, &data, 0, NULL);
+      size_t deps[1] = {jobId};
+      jobId = ZSTDMT_depThreadPool_addJob(ctx, FUZ_depThreadPool_callbackFn2, &data, 1, deps);
```
Cyan4973: Both jobs write to the same `data`.
The only protection is that the second job must wait for the first one to be completed before starting,
thus solving risks of race conditions.
It basically tests that the dependency is implemented correctly.

I recommend to explain the logic of this test in code comment.

lib/common/zstdmt_depthreadpool.c
```diff
@@ -0,0 +1,159 @@
...
+	ZSTDMT_DepThreadPoolCtx* ctx = malloc(sizeof(ZSTDMT_DepThreadPoolCtx));
+	size_t i;
+
+	pthread_mutex_init(&ctx->mutex, NULL);
+	pthread_cond_init(&ctx->cond, NULL);
+	ctx->nbJobs = 0;
+	ctx->nbJobsRemaining = maxNbJobs;
+	ctx->maxNbThreads = maxNbThreads;
+	ctx->threads = malloc(sizeof(ZSTDMT_DepThreadPoolThread*) * ctx->maxNbThreads);
+	ctx->jobs = malloc(sizeof(ZSTDMT_DepThreadPoolJob*) * ctx->nbJobsRemaining);
```
Cyan4973: check both `malloc()`

Cyan4973:
The new API looks fine to me.
There are a few implementation details left, but the API itself seems in good shape.

tests/fuzzer.c
bimbashrestha: Yes. Better tests to come. This is just a place holder:)

lib/common/zstdmt_depthreadpool.c
```diff
@@ -0,0 +1,178 @@
...
+{
+	ZSTDMT_DepThreadPoolJob* job = (ZSTDMT_DepThreadPoolJob*)malloc(sizeof(ZSTDMT_DepThreadPoolJob));
+	if (job == NULL) err_abort(-1, "Malloc job");
+	job->fn = fn;
+	job->data = data;
+	job->started = 0;
+	job->finished = 0;
+	job->nbDeps = nbDeps;
+	job->depJobIds = (size_t*)malloc(sizeof(size_t) * nbDeps);
+	if (job->depJobIds == NULL) err_abort(-1, "Malloc depJobIds");
```
Cyan4973: note : if using this branch `(job->depJobIds == NULL)`, there is a memory leak with regards to `ZSTDMT_DepThreadPoolJob* job` which has been already allocated.

bimbashrestha: Hmm, should I free job here after the branch? Seems clumsy. What do you think is the best way to handle something like this?

Cyan4973: Sorry, I missed that `err_abort()` was actually aborting the process,
so forget my earlier comment, it's fine.

A higher level discussion could be : should the program recover from such an allocation error ? or give up ?
And I believe that "give up" is the right call here.

bimbashrestha:
I'm having some trouble getting rid of this c99build failure. I feel like it should just be as simple as linking pthread somewhere in the Makefile but I'm not sure.

Cyan4973:
From what I guess from the traces : 
- This fails while compiling `paramgrill`, technically at link stage
- `paramgrill` compiles without `ZSTD_MULTITHREAD`
- hence it doesn't link with `pthread`
- then `zstdmt_depthreadpool.c` fails at link stage, due to missing `pthread_*` symbols
- this seems completely unrelated to `c99`

Suggestion : 
- make sure `zstdmt_depthreadpool.c` can be compiled without `ZSTD_MULTITHREAD`
- to that end, do _not_ use `pthread_*` object directly ! This is not portable anyway. Use `ZSTD_pthread_*` objects like the rest of the code base.

bimbashrestha:
Closing. I will include this again in the future zstdmt_decompress PR

Issue #1703, opened 2019-07-26
Why dictionary compression can make the ratio so much worse than without one


dislimit:
Compressing an XML file of >10MB with a dictionary trained from dozens of XMLs (samples include the big XML), results in about 26% ratio; while compressing it without a dictionary gets <10% ratio, both using default compression level.

I read that dictionary compression works best with many small files, but trying to understand why dictionary compression can make the ratio so much worse than without one? A dictionary is basically a table prepended to the corpus for offseting when compressing, right?

Would be great to have some insights on how Zstd uses dictionaries. I'd hope that dictionaries do not hurt the performance even in cases when they do not help.

Cyan4973:
Indeed, dictionaries are mostly useful for small files. 
The reason is, they typically save a few KB, which is negligible for large multi-MB file, but is _a lot_ for small files of a few KB.

_In theory_, dictionary compression should not be detrimental for large files either. After all, it's some "additional data", that can simply be discarded if deemed not useful.

In practice though, the algorithm use some decision shortcuts, and kind of "trust" that the dictionary is used for a "suitable" use case : small data, similar to training stage. When that's not the case, it makes bad choices.

We could improve the situation, especially as we have new capabilities now, allowing the algorithm to make more educated choices.

dislimit:
Thanks @Cyan4973, could you give a concrete example when "bad choice" can hurt the ratio so much? Is there big difference when we offset into dictionary or not?

And sounds like the improvement will not be done any time soon?

Cyan4973:
There are multiple sources. An important one is the size of structures being used for search operations during compression.

Whenever possible, we try to "adapt" the amount of resource needed depending on compression level and input size : it doesn't make sense to reserve and initialize many MB of memory to compress a few KB. But on the other hand, using a few KB of resource to compress MB of input will seriously alter search capabilities, translating into worse compression ratio. So it's a balancing act.

In "normal" compression, the heuristic err towards "large size", because it helps compression ratio.
In "dictionary compression", the heuristic err towards "small size", because it helps speed and memory, and it's supposed to be the target scenario.

Now, we can adapt and correct initial decision during later stages, but sometimes we don't have the information (streaming mode with no srcSize hint), sometimes we don't have the cpu budget (fast modes), or sometimes flexibility is pinned down by choices in the early stages of the decision tree (copy-dictionary-tables mode for example).

[...]

PR #1824, opened 2019-10-15
Avoid using CDict params when input is large.

senhuang42:
Instead of using the CDict params, we use params based on the source size for compression. This should allow for better compression ratios, particularly when compressing with large files and small dictsizes, where small dictsizes will use hashtable sizes too small for these large inputs.

In addition to changes to param selection logic:

- We introduce a new macro: `#define ZSTD_USE_CDICT_PARAMS_CUTOFF (1 MB)` to set the cutoff at which we do not use CDict params. I propose that we conservatively set this to `1 MB` based on my initial benchmarking, but this may change as I continue to do some more benchmarking.

- We introduce a new `BYTE compressionLevel` field to `struct CDict`. This field is overloaded with specification of whether or not the user used the advanced API to set custom `cParams` - if `compressionLevel == 0` then this signifies that the user used the advanced API to construct the CDict, so we won't touch their parameters. 

Much more detailed benchmarks/graphs are forthcoming.


senhuang42:
issues with merge 😢hold on...

senhuang42:
Some initial benchmarking results using small dicts of size 2000, compression levels of {3, 11}, and the dickens file from Silesia. The large filesizes is where we want to focus, since that where the impact will be.

Ratio tends to be better than old ZSTD using CDict, only a tiny bit better than no dict at all. Compression speed is more complicated - it behaves more like compression at {x} level without a dictionary, so sometimes it is better than using CDict, sometimes worse.

image

image

image

image

image

Cyan4973:
The major idea is to no longer suffer a compression ratio penalty when compressing large files with a dictionary, compared to no dictionary. 

And I believe your patch achieves just that.

Other metrics are secondary at this stage, they are good to know, but do not influence the outcome, especially as they are small, such as the small speed loss measured on certain lengths, and the tiny compressed size improvement.

lib/compress/zstd_compress.c
```diff
@@ -50,6 +50,7 @@ struct ZSTD_CDict_s {
     ZSTD_compressedBlockState_t cBlockState;
     ZSTD_customMem customMem;
     U32 dictID;
+    BYTE compressionLevel; /* 0 indicates that advanced API was used to select CDict params */
```
Cyan4973: I guess it should be an `int` instead, to respect the API contract.
`int` can also represent negative values, which are useful in some high speed scenarios.

Cyan4973:
Looks mostly good to me !

The only requested change is about the type of `compressionLevel`, which should be `int`.

senhuang42:
TODO: Switching the hard cutoff of 1MB to a multiplier based on dictsize may make more sense actually - this way if we have a huge dict and huge input size we're not choosing the new path incorrectly, plus benchmarks with large dictionaries do look a bit different than the ~2kb ones I posted above.

Cyan4973:
> TODO: Switching the hard cutoff of 1MB to a multiplier based on dictsize 

Yes !

senhuang42:
From eyeballing some more data it seems like a multiplier of `10*dictSize` is actually more than enough to start getting better results on most files, compressionlevels, and dict sizes.

However, I still think an additional min filesize requirement would be nice. For example: for dictSize 2000 and filesize 16K-32K, MB/s and ratio are very comparable to current behavior, but for something like dictSize 32K and filesize 256K and up, behavior is often better by a lot across files like nci, dickens, mozilla in both MB/s and ratio. With the extra min filesize requirement in addition to a multiplier, the behavior of using the new path will be more conservative, and revert to normal behavior for edge cases like using a 100 byte dict to compress 1000 bytes (though in this case neither the typical CDict behavior nor the new path behavior is ideal).

In particular, when filesize + dictsize > 256KB, Zstd will use the next set of cParams with the new patch, which seems to give some nice MB/s gains (in addition to inherently better ratio).

I propose we require both a minimum filesize threshold and a multiplier, set at:
`#define ZSTD_USE_INPUT_PARAMS_SRCSIZE_THRESHOLD (256 KB)`
and
`#define ZSTD_USE_INPUT_PARAMS_DICTSIZE_MULTIPLIER_THRESHOLD (10)`

Though perhaps this is over-complicating the issue or being too conservative.

Cyan4973:
Acknowledging that a linear scale doesn't work as well for very small sizes than for larger ones is a very good point. The idea of introducing a threshold to select different behaviors between the 2 worlds looks good too.

But I haven't really understood what will happen in your proposal depending on being under or above the threshold.

Suggestion : design the primary multiplier with larger dictionary sizes in mind (like the default 110KB one). 10x is likely too much. 2x is most likely too little. So a "good" multiplier is probably somewhere between 4x and 8x.

Then apply the same formula with small dictionaries, including limit cases using very small dictionaries (I believe the trainer applies a minimum size of 256 bytes), and explain what is lost by applying the same formula. Then describe the mitigation, using the threshold if it's still the wanted technique.

A table, with src size on one side, and dict size on the other one, is likely going to give a good snapshot of what happens.

senhuang42:
As per Felix's suggestion, I've added (and tested) a new enum option to `ZSTD_dictAttachPref_e`, which allows the user to force using the source size as the way to determine the `cParams`, which would be done through `CCtx_setParameter`.

I re-ran a batch of benchmarks too, focused on smaller dictsizes and srcsizes. This time, the benchmarks were ran on a Xeon E5-2680 v4 @ 2.40GHz, with 10 repetitions of each `{filesize, compression level, dictsize, filename}` combination. The speed results are a bit noisy. Each column represents a filesize, and each row represents a dictionary size, with each entry representing the percentage difference between the new CDict path vs. the old. Mbps results are represented as the P90 of speed of each 10 runs.

Based on the results, I think that it's best for us to use a filesize cutoff threshold of `128KB`, and a multiplier of `6*dictSize` to use the new CDict path. Generally we see that speed of compression is comparable or better than the old path across the board for large filesizes with relatively large dictionaries, and that we generally don't see the same sort of general gains for smaller filesizes on smaller inputs in general. If we were to cut off each table and discard anything left of the column `128000`, then the results look pretty good in general.

image

image

Cyan4973:
Thanks for detailed feedback @senhuang42 !

It's interesting to see that the "better threshold" seems to also depend on the strategy, with level 11 apparently receiving benefits sooner than level 3. Anyway, that could be a refinement for later, I'm fine with a generic threshold for all strategies for the time being.

> I think that it's best for us to use a filesize cutoff threshold of 128KB, and a multiplier of 6*dictSize to use the new CDict path.

Just to be sure, about the effect of the cutoff threshold, do you mean : 
- All inputs > 128 KB will reload the dictionary, and below the threshold, they will reload if `inputSize > 6*dictSize` ? or, 
- Dictionary is reloaded if both conditions are satisfied : `(inputSize > 128 KB) && (inputSize > 6*dictSize)`  ?

senhuang42:
Dictionary is reloaded only if both conditions are satisfied.

The way it'll be written the code will be in the opposite form though, i.e. dictionary is not reloaded either condition is not satisfied.

Cyan4973:
👍 
Right, that's a good start !

Just a little file conflict to fix, and this thing should be ready for a merge.

senhuang42:
And yeah, the point about compression levels is very relevant - what we are doing is creating heuristics (cutoff/multiplier) based on heuristics (compression levels mapped to params), which can make the results surprising at times. In general, I do notice that the new path tends to perform better and better relative to the old behavior as compression level increases.

Cyan4973:
fix #1703

Issue #1817, opened 2019-10-05
Windows releases higher than v 1.4.0 runs with error when compressing with a dictionary


artyompetrov:
When I run windows binary of zstd v.1.4.0 or higher:
.\zstd.exe .\file.txt -D .\dictName
I get an error:
zstd: cannot use .\file.txt as an input file and dictionary

But when I use zstd 1.3.8 everything goes great:
.\zstd.exe .\file.txt -D .\dictName
.\file.txt        :  3.75%   (671180 =>  25144 bytes, .\file.txt.zst)

Cyan4973:
could it be an issue of `/` vs `\` ?

artyompetrov:
no, without slashes I get the same result:
C:\Users\user\Desktop>zstd.exe file.txt -D dictName -19
zstd: cannot use file.txt as an input file and dictionary

but zstd v1.3.8 works great:
C:\Users\user\Desktop>zstd138.exe file.txt -D dictName -19
file.txt          :  2.45%   (671180 =>  16418 bytes, file.txt.zst)

Cyan4973:
So, it seems that a recent protection, preventing to use the same file as both the input and the dictionary, is kicking in on the Windows binary.

Indeed, that's weird, and something to fix.

Cyan4973:
Issue reproduced, using `mingw64` on Windows 10 + msys2

Cyan4973:
I was surprised that this issue, which can be reproduced locally, did not show up during CI tests.
It appears that CI tests on Windows are limited to compilation test, they don't have a runtime session afterwards.

This should be updated, it would have make it possible to detect this issue earlier.

Cyan4973:
Strangely enough, runtime tests on Windows are supposed to be enabled, albeit with Visual binaries :
https://github.com/facebook/zstd/blob/dev/appveyor.yml#L166

And I'm positive that this used to work.
But it's unclear why they do no longer run recently...

Cyan4973:
Issue fixed in latest `dev` branch update

PR #1834, opened 2019-10-18
Windows fixes

Cyan4973:
This started as a small patch for #1817 : 
the function protecting users from the scenario where both source and dictionary are the same file would always be triggered within the Windows binary.
This is because file identification by inode does not seem to work on Windows, so it's better to just compare file names, which is an existing code path for Visual, just extend it to `mingw`.

But then the questions was : why has this problem not already been detected by tests ? Well, because there is no runtime test on Appveyor. So let's just enable them.

And then, it became a rabbit hole. Many little things (corner cases mostly, new features too) would start failing, making it impossible to complete the runtime on Windows. Fixing them one by one resulted into this more generous patch.

- Runtime tests would not start on MinGW environment
- Tests would fail console detection inside `mintty` environment
- `nul` and `NUL` trigger different behaviors on Windows
- Test mode `-t` was not working on Windows
   + optimized test mode, so that it doesn't open output file nor `fwrite()` any data
- Tests for `--output-dir-flat` were failing on Windows
- `harness` tests for the educational decoder would not compile
- `zstd file -D dict` was always failing on Windows
- Improved speed of `make check`
- Scores of cast warnings on Windows + MinGW (failing `-Werror`)
- Enable `make check` on Appveyor


Cyan4973:
There is still a pb to fix : 
in `util.c`, a function pointer is casted into a different function pointer.
This is fine for Visual, but not for `gcc9`, which complains that these functions have different prototypes.
Using `void*` as an intermediate is fine for `gcc9` (and `clang`). But it's no longer okay for Visual...

programs/fileio.c
```diff
@@ -671,56 +686,50 @@ int FIO_checkFilenameCollisions(const char** filenameTable, unsigned nbFiles) {
...
-        }
-        strcpy(filename, srcFilename);
-    } else {
-        filename = (char*) malloc((strlen(filenameBegin+1)+1) * sizeof(char));
-        if (!filename) {
-            EXM_THROW(30, "zstd: %s", strerror(errno));
-        }
-        strcpy(filename, filenameBegin+1);
-    }
+    filenameStart = extractFilename(path, separator);
```
terrelln: Unrelated to this patch, since it doesn't change behavior, but what is the desired behavior if the source directory is `path/to/my/file/`?

I would expect the output paths to be relative to `path/to/my/file/`, but I believe the current behavior is to output `outdir/to/my/file`.

Cyan4973: The functionality developed by @senhuang42 is called `--output-dir-flat`.

It places every compressed file, whatever its source directory, into a single, flat, target directory. It has the advantage of being simpler, both to describe and implement. 

It has the disadvantage of only allowing one file with a given name, so multiple files with same name from different directories will overwrite each other (there will be a prompt, similar to what happens today when overwriting a file).

Another command was considered, `--output-dir-mirror`, which would reproduce the source directory tree into a target directory. This patch was dropped for the time being. It could be resumed after `v1.4.4` release.

PR #1836, opened 2019-10-21
Added --output-dir-flat=

Cyan4973:
which is a preferred construction for long options when applicable.

Also, added `--output-dir-flat` entry into man page.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -136,11 +136,10 @@ static int usage_advanced(const char* programName)
     DISPLAY( " -q     : suppress warnings; specify twice to suppress errors too\n");
     DISPLAY( " -c     : force write to standard output, even if it is the console\n");
     DISPLAY( " -l     : print information about zstd compressed files \n");
-    DISPLAY( " --output-dir-flat directory: results stored into `directory`. Filename collisions mean first file will be compressed. With -f, the last file will be compressed.\n");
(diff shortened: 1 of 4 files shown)


programs/zstdcli.c
```diff
@@ -756,6 +756,7 @@ int main(int argCount, const char* argv[])
                     if (longCommandWArg(&argument, "--stream-size=")) { streamSrcSize = readU32FromChar(&argument); continue; }
                     if (longCommandWArg(&argument, "--target-compressed-block-size=")) { targetCBlockSize = readU32FromChar(&argument); continue; }
                     if (longCommandWArg(&argument, "--size-hint=")) { srcSizeHint = readU32FromChar(&argument); continue; }
+                    if (longCommandWArg(&argument, "--output-dir-flat=")) { outDirName = argument; continue; }
```
felixhandte: I'm confused. How is this different from the existing test above (line 692)? Should the other still be present with this added here?

felixhandte: Ohhhh, I see, this is if it's provided as a single arg in `--option=val` style.

Cyan4973: They are both enabled.
`--output-dir-flat DIR` is just equivalent to `--output-dir-flat=DIR`.

I tend to prefer the second formulation, which is only possible with long commands.

PR #1838, opened 2019-10-22
fix initCStream_advanced() for fast strategies

### LINKED ISSUE #1787: ZSTD_initCStream_advanced results in different compression settings than ZSTD_initCStream or ZSTD_CCtx_setParameter
Consider the following fragments:

1.
```
ZSTD_CCtx* compressor = ZSTD_createCCtx();
ZSTD_initCStream(compressor, compression_level);
```
2.
```
ZSTD_CCtx* compressor = ZSTD_createCCtx();
ZSTD_CCtx_setParameter(compressor, ZSTD_c_compressionLevel, compression_level);
```
3.
```
ZSTD_CCtx* compressor = ZSTD_createCCtx();
ZSTD_parameters params =
    ZSTD_getParams(compression_level, ZSTD_CONTENTSIZE_UNKNOWN, 0);
ZSTD_initCStream_advanced(compressor, nullptr, 0, params, ZSTD_CONTENTSIZE_UNKNOWN);
```

They look equivalent, and the first two in fact are. But the third yields very different compression settings in compression levels 1 and 2, and slightly different compression settings in higher levels.

There are two independent problems here.

The first problem causes a significantly weaker compression for levels 1 and 2 if `ZSTD_initCStream_advanced` is used to set the compression level. Here is my analysis what goes on (in zstd 1.4.0, but I believe nothing changed later):
* `ZSTD_initCStream_advanced` calls `ZSTD_assignParamsToCCtxParams`.
* `ZSTD_assignParamsToCCtxParams` sets `compressionLevel` to `ZSTD_CLEVEL_DEFAULT` (i.e. 3) because it “should not matter, as all cParams are presumed properly defined”.
* `ZSTD_compressStream2` calls `ZSTD_getCParamsFromCCtxParams`.
* `ZSTD_getCParamsFromCCtxParams` initializes `cParams` from `compressionLevel`, i.e. `ZSTD_CLEVEL_DEFAULT`. This includes setting `cParams.targetLength` to 1.
* `ZSTD_getCParamsFromCCtxParams` copies nonzero values from `CCtxParams->cParams` to `cParams`.
* `CCtxParams->cParams.targetLength` happens to be 0 for compression levels 1 and 2 (see `ZSTD_getCParams`), so it is not copied, and it is left at 1.
* This causes `ZSTD_disableLiteralsCompression` to return 1 instead of 0.

The first problem can be worked around by adding an apparently redundant `ZSTD_CCtx_setParameter` call after `ZSTD_initCStream_advanced`, thus specifying the compression level twice. I am not sure what should be the proper fix in zstd.

Here is the second problem. If `ZSTD_getParams` is used (either with `ZSTD_initCStream_advanced` or with `ZSTD_CCtxParams_init_advanced` + `ZSTD_CCtx_setParametersUsingCCtxParams`), it takes an `estimatedSrcSize`, which can be `ZSTD_CONTENTSIZE_UNKNOWN`. But if `ZSTD_getParams` is not used at all, the effect is different than specifying the `estimatedSrcSize` of `ZSTD_CONTENTSIZE_UNKNOWN`.

This can be observed if the first `ZSTD_compressStream2` call is with `ZSTD_e_end` (i.e. zstd knows that the input is complete on the first streaming call). In this case compression settings are automatically tuned for the source size. This automatic tuning does not happen when `ZSTD_getParams` was used with `ZSTD_CONTENTSIZE_UNKNOWN`, in which case settings remain tuned for an infinite estimated size. I did not investigate further why this happens.

I can prepare a program which demonstrates these issues, but I do not have a standalone version ready (I only have one in the Google internal build system), so maybe this will not be needed if these descriptions are sufficient.

Cyan4973:
Compression ratio of ` initCStream_advanced()` when using fast strategies (levels 1 & 2)
was seriously reduced, due to accidental disabling of Literals compression.

Credit to @QrczakMK, which perfectly described the issue, and implementation details,
making the fix straightforward.

Example : initCStream with level 1 on synthetic sample P50 :
```
Before : 5,273,976 bytes
After  : 3,154,678 bytes
ZSTD_compress (for comparison) : 3,154,550
```
Fix #1787.

To follow : refactor the test which was supposed to catch this issue (and failed)

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -238,10 +238,10 @@ size_t ZSTD_CCtxParams_init_advanced(ZSTD_CCtx_params* cctxParams, ZSTD_paramete
     RETURN_ERROR_IF(!cctxParams, GENERIC);
     FORWARD_IF_ERROR( ZSTD_checkCParams(params.cParams) );
     memset(cctxParams, 0, sizeof(*cctxParams));
+    assert(!ZSTD_checkCParams(params.cParams));
     cctxParams->cParams = params.cParams;
     cctxParams->fParams = params.fParams;
     cctxParams->compressionLevel = ZSTD_CLEVEL_DEFAULT;   /* should not matter, as all cParams are presumed properly defined */
-    assert(!ZSTD_checkCParams(params.cParams));
     return 0;
 }
 
@@ -251,10 +251,10 @@ static ZSTD_CCtx_params ZSTD_assignParamsToCCtxParams(
         const ZSTD_CCtx_params* cctxParams, ZSTD_parameters params)
 {
     ZSTD_CCtx_params ret = *cctxParams;
+    assert(!ZSTD_checkCParams(params.cParams));
     ret.cParams = params.cParams;
     ret.fParams = params.fParams;
     ret.compressionLevel = ZSTD_CLEVEL_DEFAULT;   /* should not matter, as all cParams are presumed properly defined */
-    assert(!ZSTD_checkCParams(params.cParams));
     return ret;
 }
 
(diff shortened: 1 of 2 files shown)


terrelln:
Once I've fixed the tests, this patch shows the fix in the regression testing tool.

```diff
521c521
< silesia,                            level 1,                            old streaming advanced,             6250645
---
> silesia,                            level 1,                            old streaming advanced,             5314109
547c547
< silesia.tar,                        level 1,                            old streaming advanced,             6255920
---
> silesia.tar,                        level 1,                            old streaming advanced,             5336879
577,578c577,578
< github,                             level 1,                            old streaming advanced,             184796
< github,                             level 1 with dict,                  old streaming advanced,             43560
---
> github,                             level 1,                            old streaming advanced,             143682
> github,                             level 1 with dict,                  old streaming advanced,             42430

```

Issue #1752, opened 2019-08-28
Zstd refuses to compress named pipes


terrelln:
```
> zstd -19 <(head -c 1M nci) -o /dev/null
Warning : /proc/self/fd/12 is a symbolic link, ignoring
> zstd -19 <(head -c 1M nci) -c > /dev/null
Warning : /proc/self/fd/12 is a symbolic link, ignoring
```

This differs from gzip
```
gzip -9 <(head -c 1M nci) -c > /dev/null
```

svenha:
This should be fixed because other pipe uses already work:
    ` > head -c 1M nci | zstd -19 -o /dev/null`

terrelln:
Yeah my bad, I meant named pipes, `stdin` works fine.

Cyan4973:
cc @bimbashrestha

PR #1839, opened 2019-10-22
Allowing named pipes to go through in zstdcli

bimbashrestha:
https://github.com/facebook/zstd/issues/1752 for context

Just added a check to see if the file is a named pipe. 

`./zstd -19 <(head -c 1M ~/silesia/nci) -o /dev/null`
`/dev/fd/63           :  6.73%   (1048576 =>  70589 bytes, /dev/null)`


Cyan4973:
Could you please add a test in `tests/playTests.sh` along the lines described in the commit ?

This will ensure that the functionality works now, but also remains well supported in the future.

Cyan4973:
Any reason why you preferred to open a new `playTests_bash.sh` file, as opposed to adding a new test into existing `playTests.sh` ?

Cyan4973:
Looks great, thanks !

Cyan4973:
Unfortunately, tests on Windows are failing now, likely due to (lack of) support of named pipe on Windows.
See for example : https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/builds/28364947/job/jag0kq499x0swbe2#L3663

1) Find a way to gate the feature and the tests when running on Windows

2) It's surprising we did not notice that issue before the merge. It seems that, at PR time, only compilation tests are run, not runtime one. That could be something to fix too.

PR #1841, opened 2019-10-23
Adding arm64 make check to travis.ci

bimbashrestha:
https://travis-ci.com/bimbashrestha/zstd/jobs/248513778
Make check passes on arm64


.travis.yml
```diff
@@ -222,5 +222,12 @@ matrix:
...
       after_failure:
         - cat "$TRAVIS_BUILD_DIR"/builddir/meson-logs/testlog.txt
+
+    # arm 64
+    - name: arm64
+      os: linux
+      arch: arm64
+      script:
+        - make check
   allow_failures:
```
Cyan4973: I believe `allow_failures` is part of the `meson` test.

By introducing the `arm64` test in the middle, it presumably transfers the `allow_failures` property from the `meson` test to the `arm64` one.

bimbashrestha: Ah I see. I'll try moving that outside and see what happens.

Cyan4973:
Great !
Over time, we should transfer our `qemu-arm64` test towards this new platform.

Note that `qemu` can also do other platforms (`ppc` notably), including the legacy `arm` 32-bit instruction set, which is different from `arm64`, so we need to keep that one.

PR #1843, opened 2019-10-23
Take ZSTD_parameters as a const pointer

moozzyk:
Fixes: #1637

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -248,12 +248,12 @@ size_t ZSTD_CCtxParams_init_advanced(ZSTD_CCtx_params* cctxParams, ZSTD_paramete
 /* ZSTD_assignParamsToCCtxParams() :
  * params is presumed valid at this stage */
 static ZSTD_CCtx_params ZSTD_assignParamsToCCtxParams(
-        const ZSTD_CCtx_params* cctxParams, ZSTD_parameters params)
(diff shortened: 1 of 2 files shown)


moozzyk:
I assumed that the public API should remain unchanged. Let me know if this assumption is incorrect.

Cyan4973:
cc @terrelln, as author of #1637

terrelln:
Yeah, we should leave the public API as-is, since it is deprecated. If we were going to change it we would just delete it.

terrelln:
Can we also change these functions:

* `ZSTD_compressBegin_advanced_internal()`
* `ZSTD_compress_internal()`

moozzyk:
@terrelln - I am not sure I understand the ask. `ZSTD_compressBegin_advanced_internal()` does not take `ZSTD_parameters`:

https://github.com/facebook/zstd/blob/fb77afc62679d24d79df34ab2be0fb01f870c331/lib/compress/zstd_compress_internal.h#L965-L971

and `ZSTD_compress_internal` was updated in my PR:

https://github.com/facebook/zstd/pull/1843/files#diff-62f99f3a22fe685ec153abb360cca5d5R3067

terrelln:
Thanks @moozzyk! Sorry for the noise, this looks great to me!

Cyan4973:
We are in feature freeze for the next release.

As a consequence, since this patch modifies the recently validated core library, even in a minor way, it seems better to wait a bit to trigger the merge.
We will be able to merge this PR right after the release.

PR #1844, opened 2019-10-23
Adding --file=FILE feature

AhmedAbdellah19:
Adding --file=FILE as feature in any command by adding file names in FILE in filenameTable.


programs/zstdcli.c
```diff
@@ -824,7 +823,7 @@ int main(int argCount, const char* argv[])
                         DISPLAYLEVEL(4, "[TRACE] call read function is finished\n");
                         DISPLAYLEVEL(4, "[TRACE] extendedFileNamesTable:\n");
 
-                        unsigned i = 0;
+                        unsigned i;
```
Cyan4973: Open a new block here, and close it right after the `for` loop.

Do the same for each `for` loop using an `i` counter.
There is no reason for a counter `i` to be shared between loop, because the outcome from a previous loop is not applicable to the next loop. They are effectively independent counters.

programs/util.c
```diff
@@ -174,6 +174,228 @@ U64 UTIL_getTotalFileSize(const char* const * const fileNamesTable, unsigned nbF
...
+  table->fileNames = filenames;
+  table->buf = buf;
+  table->tableSize = tableSize;
+  return table;
+}
+
+void UTIL_freeFileNamesTable(FileNamesTable* table) {
+  if(table) {
+    if(table->fileNames) {
+      free(table->fileNames);
```
Cyan4973: It seems to be the last test error : 
`free(table->fileNames);` emits some warning on Visual, though it seems to be a Visual bug.

`table->fileNames` is a `const char**`, aka a table of `const char*`.
It's not possible to free a `const char*`, because operations on `const char*` shall not change any related state.
But it should be possible to `free()` a mutable table of pointers, even if those pointers point to some read-only memory. In this respect `const char* *table` should be no different from `object *table`.

However, Visual seems to be confused by the first `const`, and emits a warning. I think this is wrong. I would classify it as a bug, though it has been so long lasting, that I'm not even sure if it's a bug of a feature.

Anyway, the usual fix for this case is to cast the pointer to `(void*)`.

Cyan4973:
It seems the code is in good shape. The last warning should be easy to fix.

Now, there is one last thing missing from this PR, it's a test.
You may already have developed a few tests on your side while developing this feature, we merely want the relevant ones into the repository. 
The tests shall prove that the capability works.
More importantly, it shall ensure that the capability _continue to work_ in the future, if some future contributor accidentally breaks the code by introducing some subtle change.

To test CLI capabilities, we have a shell test script, in `tests/playTests.sh`.
You can add a section into it to test the `--file=FILE` command.
Just add a few basic scenarios you can think about.
We will likely add more complex scenarios in the future, but for a first release, a few basic ones are enough.

The tests will be automatically played in CI environment.

Cyan4973:
Moving the patch into feature branch `ahmed_file`, for further investigation

PR #1847, opened 2019-10-24
fix zlibWrapper for Visual Studio

### LINKED ISSUE #1800: zlibWrapper/*.c files cannot be compiled with VS
Thank you for your great work!

Visual studio (any version) cannot compile *.c files in zlibWrapper folder.

Following line(s)(25 times) requires -Wstrict-aliasing=1 option.
`state = (gz_statep)file;`

But, VS doesn't support strict aliasing.
So, I think it's better to change those lines to:
`*(void**)(&state) = (void*)file;`
or
`state.file = file;`

And, VS doesn't define ssize_t type, but it is used in gzread.c
So, something like this is needed in gzread.c
`#if defined(_MSC_VER) && !defined(ssize_t)`
`#include <BaseTsd.h>`
`typedef SSIZE_T ssize_t;`
`#endif`

What do you think about these issue?

Cyan4973:
Fix #1800

As per https://github.com/facebook/zstd/issues/1800#issuecomment-545945050,
Visual does not support `ssize_t` type,
which is an issue for `gzread.c`.

Added a work around, suggested by @bluenlive

Note : I have not been able to confirm the problem,
so this is a blind fix.
This seems safe outside of Visual, since it is gated by `_MSC_VER` macro.

### DIFF
--- a/zlibWrapper/gzread.c
+++ b/zlibWrapper/gzread.c
@@ -8,6 +8,14 @@
 
 #include "gzguts.h"
 
+/* fix for Visual Studio, which doesn't support ssize_t type.
+ * see https://github.com/facebook/zstd/issues/1800#issuecomment-545945050 */
+#if defined(_MSC_VER) && !defined(ssize_t)
+#  include <BaseTsd.h>
+   typedef SSIZE_T ssize_t;
+#endif
+
+
 /* Local functions */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Merging, as the failed tests are unrelated

PR #1850, opened 2019-10-25
Gating named pipe support on windows

bimbashrestha:
Test plan: waiting on this test to finish running.

https://ci.appveyor.com/project/bimbashrestha/zstd/builds/28390954


programs/fileio.c
```diff
@@ -513,12 +513,24 @@ static FILE* FIO_openSrcFile(const char* srcFileName)
         return NULL;
     }
 
+#ifdef _MSC_VER
+
+    if (!UTIL_isRegularFile(srcFileName)) {
```
felixhandte: I don't love duplicating the same test with the same code.

I'd much rather have a single `UTIL_isRegularFile()` check on both platforms, and then an addition `UTIL_isFIFO()` check on non-win platforms.

programs/zstdcli.c
```diff
@@ -992,11 +992,25 @@ int main(int argCount, const char* argv[])
...
+
+            if (UTIL_isLink(filenameTable[u])) {
+                DISPLAYLEVEL(2, "Warning : %s is a symbolic link, ignoring\n", filenameTable[u]);
+            } else {
+                filenameTable[fileNamesNb++] = filenameTable[u];
+            }
+
+#else
+
             if (UTIL_isLink(filenameTable[u]) && !UTIL_isFIFO(filenameTable[u])) {
```
felixhandte: So are we controlling whether we allow pipes based on the follow links option? Has that flag's documentation been updated? I note that this error message no longer really applies.

programs/zstdcli.c
```diff
@@ -992,11 +992,25 @@ int main(int argCount, const char* argv[])
     if (!followLinks) {
         unsigned u;
         for (u=0, fileNamesNb=0; u<filenameIdx; u++) {
+
+#ifdef _MSC_VER
+
+            if (UTIL_isLink(filenameTable[u])) {
```
felixhandte: Same preference.

programs/zstdcli.c
bimbashrestha: Yes, we are allowing pipes based on this and I just added a comment

programs/util.c
```diff
@@ -115,6 +115,8 @@ int UTIL_isSameFile(const char* file1, const char* file2)
 #endif
 }
 
+#ifndef _MSC_VER
+/* Using this to distinguish named pipes */
 U32 UTIL_isFIFO(const char* infilename)
```
Cyan4973: minor : if `UTIL_isFIFO()` doesn't exist on Windows, its declaration within `util.h` should also be gated.

bimbashrestha:
I had to remove tests for both visual 2008 and visual 2010 to get the appveyor "pr" ci tests to pass. We had previously suspected that the culprit might have just been 2008 but it seems  having 2010 results in the same failure.

Test plan: merged to master on my branch and here is the appveyor test passing (https://ci.appveyor.com/project/bimbashrestha/zstd/builds/28448884)

PR #1852, opened 2019-10-25
Added "--exclude-compressed" flag feature

shashank0791:
Through the Zstd CLI, this flag skips files that are precompressed (with extension .lz4, .xz, .zst, .gz). Added relevant test case in playTests.sh. All tests are passing as well.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1496,17 +1496,17 @@ FIO_determineCompressedName(const char* srcFileName, const char* outDirName, con
     static char* dstFileNameBuffer = NULL;   /* using static allocation : this function cannot be multi-threaded */
     char* outDirFilename = NULL;
     size_t sfnSize = strlen(srcFileName);
-    size_t const suffixSize = strlen(suffix);
+    size_t const srcSuffixLen = strlen(suffix);
     if (outDirName) {
-        outDirFilename = FIO_createFilename_fromOutDir(srcFileName, outDirName, suffixSize);
+        outDirFilename = FIO_createFilename_fromOutDir(srcFileName, outDirName, srcSuffixLen);
         sfnSize = strlen(outDirFilename);
         assert(outDirFilename != NULL);
     }
 
-    if (dfnbCapacity <= sfnSize+suffixSize+1) {
+    if (dfnbCapacity <= sfnSize+srcSuffixLen+1) {
         /* resize buffer for dstName */
         free(dstFileNameBuffer);
-        dfnbCapacity = sfnSize + suffixSize + 30;
+        dfnbCapacity = sfnSize + srcSuffixLen + 30;
         dstFileNameBuffer = (char*)malloc(dfnbCapacity);
         if (!dstFileNameBuffer) {
             EXM_THROW(30, "zstd: %s", strerror(errno));
@@ -1520,7 +1520,7 @@ FIO_determineCompressedName(const char* srcFileName, const char* outDirName, con
     } else {
         memcpy(dstFileNameBuffer, srcFileName, sfnSize);
     }
-    memcpy(dstFileNameBuffer+sfnSize, suffix, suffixSize+1 /* Include terminating null */);
+    memcpy(dstFileNameBuffer+sfnSize, suffix, srcSuffixLen+1 /* Include terminating null */);
     return dstFileNameBuffer;
 }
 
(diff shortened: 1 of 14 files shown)


programs/util.c
```diff
@@ -326,6 +326,27 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
 
 #endif /* #ifdef _WIN32 */
 
+/* Check if the file is precompressed (.zst, .lz4, .gz, .xz).
+YES => Skip the file (return 0)
+NO => return 1
+*/
+int UTIL_isPrecompressedFile(const char *inputName)
```
Cyan4973: Considering the name of the function `isPrecompressedFile()`,
it feels more logical to return `1` (==true) when the file is indeed compressed.

Also, maybe `isCompressedFile()` is enough.
Not sure what `Pre` in `Precompressed` implies.

programs/util.h
```diff
@@ -127,6 +127,8 @@ extern int g_utilDisplayLevel;
     typedef struct stat stat_t;
 #endif
 
+int g_excludeCompressedFiles;
+static const char compressedFileExtensions[4][10] = {".zst",".gz",".xz",".lz4"};
```
Cyan4973: We may need to `typedef` a type here,
in order to make sure that all invocations of this type use the same constants,
and also to make it easier to evolve this list in the future.

programs/zstdcli.c
```diff
@@ -118,6 +118,7 @@ static int usage(const char* programName)
 #endif
     DISPLAY( " -D file: use `file` as Dictionary \n");
     DISPLAY( " -o file: result stored into `file` (only if 1 input file) \n");
+    DISPLAY( "--exclude-compressed:  only compress files that are not previously compressed \n");
```
Cyan4973: That's nice to have it in the `-h` screen,
though this rather belong to the "advanced" usage category.

tests/playTests.sh
```diff
@@ -215,6 +215,25 @@ $ZSTD tmp -c --compress-literals    -19      | $ZSTD -t
 $ZSTD -b --fast=1 -i0e1 tmp --compress-literals
 $ZSTD -b --fast=1 -i0e1 tmp --no-compress-literals
 
+println "test: --exclude-compressed flag"
+mkdir precompressedFilterTestDir
+./datagen $size > precompressedFilterTestDir/input.5
+./datagen $size > precompressedFilterTestDir/input.6
+$ZSTD --exclude-compressed --long --rm -r precompressedFilterTestDir
+sleep 5
```
Cyan4973: Why is there a need for `sleep 5` ?

programs/util.c
```diff
@@ -326,6 +326,27 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
...
+    return compareExtensions(inputName,compressedFileExtensions);
+}
+
+int compareExtensions(const char* infilename, const char extensionList[4][10])
+{
+   int i=0;
+   //char* ext = strchr(infilename, '.');
+   for(i=0;i<4;i++)
+   {
+     char* ext = strstr(infilename,extensionList[i]);
```
Cyan4973: `const char* ext`

Cyan4973:
Thanks for the patch @shashank0791 ,
the strategy proposed to handle this issue looks pretty good !

There are a number of tactical decisions which are challenged in comments, but that's relatively minor.

programs/util.h
Cyan4973: Also,
we already have a list of extensions, but in `fileio.h` :
https://github.com/facebook/zstd/blob/dev/programs/fileio.h#L33

I'm wondering if it would make sense to join these lists, so that there is a single point of definition.

felixhandte: Indeed. Also note that my PR #1848 creates an actual null-terminated list of extensions. It is better to use that pattern (or alternatively a `sizeof(array)/sizeof(array[0])` idiom) rather than rely on a hard-coded length at every use.

programs/fileio.c
```diff
@@ -1453,7 +1453,12 @@ FIO_compressFilename_srcFile(FIO_prefs_t* const prefs,
 
     ress.srcFile = FIO_openSrcFile(srcFileName);
     if (ress.srcFile == NULL) return 1;   /* srcFile could not be opened */
-
+    if (g_excludeCompressedFiles && !UTIL_isPrecompressedFile(srcFileName)) {  /* precompressed file (--exclude-compressed). DO NOT COMPRESS */
+        DISPLAYLEVEL(4, "Precompressed file: %s \n", srcFileName);
```
felixhandte: This message should not only identify the category of the file (that it is already compressed), it should also describe the action that zstd is taking as a result (that we are not going to compress it again).

programs/util.c
shashank0791: Resolved in latest commit "Fixing tests and moving flag to advanced usage 0e9a37d"

programs/zstdcli.c
shashank0791: Addressed in latest commit "Fixing tests and moving flag to advanced usage 0e9a37d"

shashank0791:
Closing this pull request because of git resyncing issues. Will be opening a new PR.

Issue #1507, opened 2019-01-23
zstd --train FILEs: passing files via file?


braibant:
Looking at zstd (the command line tool) in version 1.3.8, it looks like dictionary training is restricted to passing the training files as command line arguments. Would it be possible to obtain the names of the files to be used in training from a file, a la `grep ---file=FILE`, receiving files from FILE, one per line? 

(This would allow for having arbitrary number of files available for training, irrespective of the size of those files.)

Cyan4973:
Thanks for the suggestion @braibant .

`zstd` command line doesn't have this capability yet.
Closest thing available is the `-r` recursive mode, which can dynamically discover a list of files belonging to a directory (and its sub-directories). And of course shell expansion, but this one can explode when the list of files is really large.

Of course, it's not equivalent : `-r` takes everything within a directory, while `--file=FILE` is explicit, hence more selective.

We will have to add this feature in a future version.

bra-fsn:
Would be a very useful feature indeed. BTW, is training `zstd` feasible with a LOT of files/data?
Does the training process's memory needs scale with the amount of the input?

Cyan4973:
> is training zstd feasible with a LOT of files/data?

Yes. I generally use the recursive mode (`-r`) for such cases, as it's free of command line expansion limitations. But that requires __all__ files within a given directory (and its sub-directories) to be samples, there is no way to "filter" them, only keep some and discards others.

> Does the training process's memory needs scale with the amount of the input?

Yes it does, almost linearly.

Cyan4973:
See #1844

Cyan4973:
This functionality is available in `v1.4.4`

PR #1853, opened 2019-10-26
--filelist=FILE feature

Cyan4973:
This is a refactor of @AhmedAbdellah19 patch (#1844), 
providing feature request #1507 ,
so that a list of files can be provided by the content of a file.

The command name has been changed to `--filelist=`, as it is more specific than original proposal.

A few basic tests are added, including some simple attempt to trick the parser with garbage noise.


programs/util.c
```diff
@@ -174,28 +200,197 @@ U64 UTIL_getFileSize(const char* infilename)
...
+
+/* condition : @file must be valid, and not have reached its end.
+ * @return : length of line written into buf, without the final '\n',
+ *           or 0, if there is no new line */
+static size_t readLineFromFile(char* buf, size_t len, FILE* file)
+{
+    assert(!feof(file));
+    CONTROL( fgets(buf, (int) len, file) == buf );  /* requires success */
+    if (strlen(buf)==0) return 0;
+    return strlen(buf) - (buf[strlen(buf)-1] == '\n');   /* ignore final '\n' character */
```
terrelln: nit: Can we write the `\0` in this function, since it already has the newline logic?

```c
{
    size_t len = strlen(buf);
    if (buf[len-1] == '\n') --len;
    buf[len] = '\0';
    return len;
}
```

programs/util.c
```diff
@@ -174,28 +200,197 @@ U64 UTIL_getFileSize(const char* infilename)
...
+/*Note: buf is not freed in case function successfully created table because filesTable->fileNames[0] = buf*/
+FileNamesTable*
+UTIL_createFileNamesTable_fromFileName(const char* inputFileName)
+{
+    size_t nbFiles = 0;
+    char* buf;
+    size_t bufSize;
+    size_t pos = 0;
+
+    if (!UTIL_fileExist(inputFileName) || !UTIL_isRegularFile(inputFileName))
```
terrelln: Can we test the `inputFileName` as a symlink? I don't know if it passes this test.

programs/util.c
```diff
@@ -174,28 +200,197 @@ U64 UTIL_getFileSize(const char* infilename)
...
+
+    buf = (char*) calloc(newTotalTableSize, sizeof(*buf));
+    CONTROL ( buf != NULL );
+
+    newTable->buf = buf;
+    newTable->tableSize = table1->tableSize + table2->tableSize;
+    newTable->fileNames = (const char **) calloc(newTable->tableSize, sizeof(*(newTable->fileNames)));
+    CONTROL ( newTable->fileNames != NULL );
+
+    {   unsigned idx1;
```
terrelln: How do we want to handle duplicates? I don't think we necessarily have to handle them, just want to bring it up.

programs/util.c
```diff
@@ -174,28 +200,197 @@ U64 UTIL_getFileSize(const char* infilename)
...
+                pos += strlen(buf+pos)+1;  /* +1 for the finishing `\0` */
+        }   }
+        assert(pos <= bufSize);
+
+        return UTIL_createFileNamesTable(filenamesTable, nbFiles, buf);
+    }
+}
+
+FileNamesTable*
+UTIL_createFileNamesTable(const char** filenames, size_t tableSize, char* buf)
```
terrelln: Can you write a comment that this takes ownership of `filenames` and `buf`?

programs/zstd.1.md
```diff
@@ -191,6 +191,9 @@ the last one takes effect.
     This is the default behavior.
 * `-r`:
     operate recursively on directories
+* `--filelist=FILE`
```
terrelln: nit: `--file-list` is more consistent with the rest of the CLI args.

programs/zstdcli.c
```diff
@@ -796,6 +798,57 @@ int main(int argCount, const char* argv[])
...
+                         *    and only start to load & interpret the file after command line is parsed.
+                         *    For a single file, it would be easy to just store its name here, and parse later.
+                         *    However, this implementation makes it possible to read multiple files.
+                         *    An equivalent will have to be able to store multiple file names.
+                         */
+                        FileNamesTable* extendedTable;
+                        FileNamesTable* curTable;
+                        FileNamesTable* concatenatedTables;
+
+                        if (!UTIL_fileExist(argument) || !UTIL_isRegularFile(argument)){
```
terrelln: We already do this check in `UTIL_createFileNamesTable_fromFileName()`. Should we only do it in one place?

programs/zstdcli.c
```diff
@@ -796,6 +798,57 @@ int main(int argCount, const char* argv[])
...
+                        concatenatedTables = UTIL_concatenateTwoTables(curTable, extendedTable);
+                        if (!concatenatedTables) {
+                            UTIL_freeFileNamesTable(curTable);
+                            UTIL_freeFileNamesTable(extendedTable);
+                            CLEAN_RETURN(badusage(programName));
+                        }
+
+                        /* transfer ownership */
+                        filenameTable = concatenatedTables->fileNames;
+                        filenameTableSize = concatenatedTables->tableSize;
```
terrelln: I think at this point we've lost the guarantee that we have enough space for all the filenames. So we could segfaults on line 1034 of the updated diff. If that is correct, can we add a test case that exposes this bug?

https://github.com/facebook/zstd/blob/6d5e0f1c9f57dd74edc5ccb4a96c0bbb5de96f7d/programs/zstdcli.c#L981

terrelln: Actually, it looks like we won't segfault, since we allocate a table of the correct size, `filenameTableSize` is wrong though.

terrelln: Well, I think it would segfault if we had 2 input files, since we use `filenameTableSize` to determine the table size.

programs/util.c
Cyan4973: Symlink test added, and passed

programs/util.c
Cyan4973: Completed documentation, though it's present in `util.h`

programs/util.c
Cyan4973: Duplicate are not specially handled.

The same file will be compressed twice, generating the same artifact twice, with usual behavior in this circumstance : never ask and never overwrite but return non-zero (default when compressing multiple files, or `-q`), or always overwrite and success (`-f`).

The situation is not different from today, if someone requests to compress the same file twice on the command line. The generated list of files goes through the same processing.

programs/zstdcli.c
Cyan4973: There is a slight difference.

Here, in case of error, we return `badusage()`, hence it complements the error message with a help screen.

Within `UTIL_createFileNamesTable_fromFileName()`, it just errors out.

`UTIL_createFileNamesTable_fromFileName()` is a library entry point that could be used for other purposes, beyond this sole invocation from `zstdcli.c`. So tests within this library function can't be avoided.

We could remove the tests from `zstdcli.c`, but then we would miss the help screen.

programs/zstdcli.c
Cyan4973: Yep, segfault generated.
This patch is not ready for merge.

Cyan4973:
All "simple" comments are addressed, 
but the segfault scenario is not.

I think it does hint at a need to refactor the source code even deeper,
such as ensuring that all list of files are using the same `FileNamesTable*` object,
and preferring to only "store" the order to read from file at command line parsing,
and then "act" later, presumably at the same place that `-r` "recursive" mode is applied.

Considering how close we are from a release, and my limited availability in this timeframe, I suggest we skip this patch for next release, giving time to refine it for the following one.

programs/zstd.1.md
Cyan4973: not sure why, but when I write it as` --file-list=`, it does not feel right ...

Cyan4973:
This is a much larger refactoring now.

All lists of file names are now handled by `FileNamesTable*`. The use of a common abstraction is now enforced in all projects, by removing the previous multi-fields abstraction from `util.h`.

I under-estimated the nb of projects which depended on it, hence there are a large nb of files modified by this patch, mostly converting from the old to the new abstraction.

The intended effect is reached : there is no longer an out-of-bound access when combining a list of files provided on command line on top of reading the content of a file containing file names right in the middle. A test has been added, which could trigger this bug in previous commit, and passes with latest one.

Cyan4973:
Last issue is about the `clang` static analyzer "`scanbuild`", which doesn't like the transfer of ownership policy for pointer parameters of `UTIL_createFileNamesTable()`. As a consequence, it misses this transfer, and confuses it for a memory leak.

No clue yet on how to fix that.

Cyan4973:
I believe all points are answered, and the work-around for `scan-build` false positive seems to work as intended. So this PR should be ready for another review.

programs/util.c
```diff
@@ -176,28 +220,218 @@ U64 UTIL_getFileSize(const char* infilename)
...
+
+        return UTIL_assembleFileNamesTable(filenamesTable, nbFiles, buf);
+    }
+}
+
+FileNamesTable*
+UTIL_assembleFileNamesTable(const char** filenames, size_t tableSize, char* buf)
+{
+    FileNamesTable* const table = (FileNamesTable*) malloc(sizeof(*table));
+    if(!table) return NULL;
```
terrelln: Does this leak `filenames` and `buf`?

programs/util.c
```diff
@@ -176,28 +220,218 @@ U64 UTIL_getFileSize(const char* infilename)
...
+    FileNamesTable* fnt;
+    if (fnTable==NULL) return NULL;
+    fnt = UTIL_assembleFileNamesTable(fnTable, tableSize, NULL);
+    fnt->tableSize = 0;   /* the table is empty */
+    return fnt;
+}
+
+void UTIL_refFilename(FileNamesTable* fnt, const char* filename)
+{
+    if (fnt->tableCapacity <= fnt->tableSize) abort();
```
terrelln: Should this be a `CONTROL()`?

programs/util.c
Cyan4973: No, it takes ownership and transfers `filenames` and `buf` into the newly created `FileNamesTable *` object.

programs/util.c
Cyan4973: I believe it should rather be an `assert()`, because the interface contract specifies that `UTIL_refFilename()` must not be invoked if the internal table is not large enough.

Now, the issue is, `assert()` can be disabled with `-DNDEBUG` (note : by default, they are not). So I believe that's why I put that `abort()` explicitly, as a way to be sure that it would not disappear.

`CONTROL()` is close enough, but slightly different : it's designed to catch normal failures for which we don't want to maintain a dedicated error path. This is typically the case for some  `malloc()` returning `NULL`, which may happen rarely, and can be terminated with a cleaner `exit()`.  Note the difference with an `abort()`.

Cyan4973:
In latest policy, `UTIL_assembleFileNamesTable()` is now obliged to succeed. It `exit()` through `CONTROL()` otherwise.

This solves situations regarding potential leak of arguments in case of `NULL` return value.

PR #1855, opened 2019-10-28
Added "--exclude-compressed" flag feature

shashank0791:
Introducing a "--exclude-compressed" that skips compression of pre-compressed input files. Addressing Issue#1579.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1430,6 +1430,9 @@ static int FIO_compressFilename_dstFile(FIO_prefs_t* const prefs,
     return result;
 }
 
+/* List used to compare file extensions (used with --exclude-compressed flag)
+* Different from the suffixList and should only apply to ZSTD compress operationResult
+*/
 static const char *compressedFileExtensions[] = {
     ZSTD_EXTENSION,
     TZSTD_EXTENSION,
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1453,8 +1453,13 @@ FIO_compressFilename_srcFile(FIO_prefs_t* const prefs,
 
     ress.srcFile = FIO_openSrcFile(srcFileName);
     if (ress.srcFile == NULL) return 1;   /* srcFile could not be opened */
-    if (g_excludeCompressedFiles && !UTIL_isPrecompressedFile(srcFileName)) {  /* precompressed file (--exclude-compressed). DO NOT COMPRESS */
-        DISPLAYLEVEL(4, "Precompressed file: %s \n", srcFileName);
+
+    /* Check if "srcFile" is compressed. Only done if --exclude-compressed flag is used
+    * YES => ZSTD will not compress the file.
+    * NO => ZSTD will resume with compress operation.
+    */
+    if (g_excludeCompressedFiles && UTIL_isCompressedFile(srcFileName)) {  /* precompressed file (--exclude-compressed). DO NOT COMPRESS */
+        DISPLAYLEVEL(4, "File is already compressed : %s \n", srcFileName);
         fclose(ress.srcFile);
         ress.srcFile = NULL;
         return 0;
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -319,6 +319,8 @@ struct FIO_prefs_s {
     /* Computation resources preferences */
     unsigned memLimit;
     int nbWorkers;
+
+    int excludeCompressedFiles;
 };
 
 
@@ -359,6 +361,7 @@ FIO_prefs_t* FIO_createPreferences(void)
     ret->srcSizeHint = 0;
     ret->testMode = 0;
     ret->literalCompressionMode = ZSTD_lcm_auto;
+    ret->excludeCompressedFiles = 0;
     return ret;
 }
 
@@ -401,6 +404,8 @@ void FIO_setNbWorkers(FIO_prefs_t* const prefs, int nbWorkers) {
 #endif
     prefs->nbWorkers = nbWorkers;
 }
+
+void FIO_setExcludeCompressedFile(FIO_prefs_t* const prefs, int excludeCompressedFiles) { prefs->excludeCompressedFiles = excludeCompressedFiles; }
 
 void FIO_setBlockSize(FIO_prefs_t* const prefs, int blockSize) {
     if (blockSize && prefs->nbWorkers==0)
@@ -1425,6 +1430,18 @@ static int FIO_compressFilename_dstFile(FIO_prefs_t* const prefs,
     return result;
 }
 
+static const char *compressedFileExtensions[] = {
+    ZSTD_EXTENSION,
+    TZSTD_EXTENSION,
+    GZ_EXTENSION,
+    TGZ_EXTENSION,
+    LZMA_EXTENSION,
+    XZ_EXTENSION,
+    TXZ_EXTENSION,
+    LZ4_EXTENSION,
+    TLZ4_EXTENSION,
+    NULL
+};
 
 /*! FIO_compressFilename_srcFile() :
  *  @return : 0 : compression completed correctly,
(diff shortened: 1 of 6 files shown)


programs/fileio.c
```diff
@@ -1454,6 +1454,16 @@ FIO_compressFilename_srcFile(FIO_prefs_t* const prefs,
     ress.srcFile = FIO_openSrcFile(srcFileName);
     if (ress.srcFile == NULL) return 1;   /* srcFile could not be opened */
 
+    /* Check if "srcFile" is compressed. Only done if --exclude-compressed flag is used
+    * YES => ZSTD will not compress the file.
+    * NO => ZSTD will resume with compress operation.
+    */
+    if (g_excludeCompressedFiles && UTIL_isCompressedFile(srcFileName)) {  /* precompressed file (--exclude-compressed). DO NOT COMPRESS */
```
Cyan4973: minor : optimization :
could this rejection test be done _before_ `FIO_openSrcFile()` ?
This would avoid a needless `fopen()` / `fclose()` cycle.

programs/util.c
```diff
@@ -326,6 +326,29 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
...
+YES => Skip the file (return 0)
+NO => return 1
+*/
+
+int UTIL_isCompressedFile(const char *inputName)
+{
+    return compareExtensions(inputName,g_compressedFileExtensions);
+}
+
+int compareExtensions(const char* infilename, const char* extensionList[])
```
Cyan4973: minor: coding style :
if this function is exposed outside of its unit, its name should respect the same `PREFIX_` convention,
in this case `UTIL_`.

Alternatively, ask yourself if exposing this function is really needed.
If not, don't expose it, a declare it `static`.

programs/util.h
```diff
@@ -39,7 +39,7 @@ extern "C" {
 #endif
 #include <time.h>         /* clock_t, clock, CLOCKS_PER_SEC, nanosleep */
 #include "mem.h"          /* U32, U64 */
-
+#include "fileio.h"
```
Cyan4973: In term of dependency, I don't like the idea to have `util.h` depending on `fileio.h`,
keeping in mind that `fileio.c` depends on `util.h`.

Dependency should be a clean tree, with no circular relations.

programs/util.h
```diff
@@ -127,6 +127,19 @@ extern int g_utilDisplayLevel;
     typedef struct stat stat_t;
 #endif
 
+int g_excludeCompressedFiles;
```
Cyan4973: do not declare global variables inside `*.h`, since is a recipe to generate duplicate symbols.

programs/util.h
```diff
@@ -127,6 +127,19 @@ extern int g_utilDisplayLevel;
...
+static const char *g_compressedFileExtensions[] = {
```
Cyan4973: I believe a better design would be to create this list inside `fileio.c`,
and provide it as argument of the `UTIL_` function from `fileio.c`.

This would get rid of the circular dependency issue.

programs/zstdcli.c
```diff
@@ -708,7 +709,7 @@ int main(int argCount, const char* argv[])
                     if (!strcmp(argument, "--compress-literals")) { literalCompressionMode = ZSTD_lcm_huffman; continue; }
                     if (!strcmp(argument, "--no-compress-literals")) { literalCompressionMode = ZSTD_lcm_uncompressed; continue; }
                     if (!strcmp(argument, "--no-progress")) { FIO_setNoProgress(1); continue; }
-
+                    if (!strcmp(argument, "--exclude-compressed")) { g_excludeCompressedFiles = 1; continue; }
```
Cyan4973: We have a generic method to set parameters in `fileio` module : 
https://github.com/facebook/zstd/blob/dev/programs/fileio.h#L65 .

It's better to use the same convention.

This will require to create the variable inside `fileio.c`, and expose a helper setter function.

programs/fileio.c
shashank0791: Addressed in latest commit.

programs/util.c
shashank0791: Removed the function in the latest commit since it is no longer required.

programs/util.h
shashank0791: Addressed in latest commit.

programs/util.h
shashank0791: Addressed in latest commit.

programs/zstdcli.c
shashank0791: Addressed in latest commit.

programs/util.c
```diff
@@ -326,6 +326,23 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
 
 #endif /* #ifdef _WIN32 */
 
+/* Check if the file is precompressed (.zst, .lz4, .gz, .xz).
+YES => Skip the file (return 0)
```
felixhandte: This comment is out of date.

programs/util.c
```diff
@@ -326,6 +326,23 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
...
+NO => return 1
+*/
+
+int UTIL_isCompressedFile(const char *inputName, const char *extensionList[])
+{
+   while(*extensionList!=NULL)
+   {
+     const char* ext = strstr(inputName,*extensionList);
```
felixhandte: This will find the extension anywhere in the filename. E.g.: `foo.zst.baz.tar`. The extension check needs to check the end of the file.

programs/fileio.c
```diff
@@ -1425,6 +1430,18 @@ static int FIO_compressFilename_dstFile(FIO_prefs_t* const prefs,
     return result;
 }
 
+static const char *compressedFileExtensions[] = {
```
felixhandte: Can you distinguish this list from the `suffixList` added in #1848? I see that this always includes all compression algorithms, even when zstd isn't built with support for them, whereas the other only has extensions zstd itself can decompress. It seems reasonable that the behavior you're adding shouldn't depend on zstd's build settings in that way, so maybe this should be a separate list. If so though, can you add a comment describing the difference between the two and how they're used?

programs/util.c
shashank0791: Fixed in latest commit.

programs/util.c
shashank0791: I do agree that the function does not explicitly check for file extension although it is not required for solving the issue#1579.
However, I have added a utility function in the latest commit that extracts extension from filename and compares only the extension with the list.

programs/fileio.c
shashank0791: Agreed. Addressed in latest commit.

programs/util.c
```diff
@@ -326,6 +326,35 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
...
+*/
+
+int UTIL_isCompressedFile(const char *inputName, const char *extensionList[])
+{
+  const char* ext = UTIL_getFileExtension(inputName);
+  if(strcmp(ext,""))
+  {
+     while(*extensionList!=NULL)
+     {
+       const char* isCompressedExtension = strstr(ext,*extensionList);
```
felixhandte: This still isn't valid. This fails to reject files like `foo.zstbar`. You need a `strcmp()`.

programs/util.c
```diff
@@ -326,6 +326,35 @@ int UTIL_prepareFileList(const char *dirName, char** bufStart, size_t* pos, char
...
 
+/* Check if the file is Compressed by comparing it with compressFileExtension list.
+YES => Skip the file (return 1)
+NO => return 0
+*/
+
+int UTIL_isCompressedFile(const char *inputName, const char *extensionList[])
+{
+  const char* ext = UTIL_getFileExtension(inputName);
+  if(strcmp(ext,""))
```
felixhandte: This shouldn't be necessary?

felixhandte:
Consider adding adversarial test cases to ensure not only that your code correctly catches the cases it's meant to, but also that it does not incorrectly fire on things it shouldn't.

programs/util.c
shashank0791: Removed the comments. The if statement when true (i.e. ext != "") will run the while loop for comparing the extension. If false, the extension comparison will be skipped since file will not have any extension (ext == "").

programs/util.c
shashank0791: Addressed in latest commit.

programs/util.c
felixhandte: This comparison, if I understand it correctly, is not necessary for correctness. You've added it as an optimization. Given that, I would rather not include it. This is not performance-sensitive code, and instead we care much more about ease-of-understanding and ease-of-verifying-correctness, which I think this harms.

Alternatively, if you feel strongly that this is a valuable optimization, it should be rephrased as:

```
if (!strcmp(ext, "")) {
  return 0;
}
```

tests/playTests.sh
```diff
@@ -225,15 +225,25 @@ sleep 5
...
+#File Extension check.
+./datagen $size > precompressedFilterTestDir/input.zstbar
+$ZSTD --exclude-compressed --long --rm -r precompressedFilterTestDir
+#ZSTD should compress input.zstbar
+test -f precompressedFilterTestDir/input.zstbar.zst
+#Check without the --exclude-compressed flag
+$ZSTD --long --rm -r precompressedFilterTestDir
+#Files should get compressed again without the --exclude-compressed flag.
+test -f precompressedFilterTestDir/input.5.zst.zst
+test -f precompressedFilterTestDir/input.6.zst.zst
```
felixhandte: Great!

programs/util.c
shashank0791: Agreed. Thanks for pointing this out. Addressed in latest commit.

Issue #1715, opened 2019-08-12
Provide a method to find the dictionary header size


terrelln:
The dictionary has two parts, the header and the content. Provide a method that returns the size of the dictionaries header.

PR #1858, opened 2019-10-29
Method to get dictionary header size

senhuang42:
For #1715 

Tested locally and validated that the sizes come out as expected.

Main hurdle is that getting the size of the dict header can only come from the functions that "copy" the dictionary entropy tables into a CCtx or DCtx, and output the size as a side effect. So, I guess there are a couple of ways to do this.

1. It looks like "\x01\x00\x00\x00\x04\x00\x00\x00\x08\x00\x00\x00" is hardcoded into the end of the entropy tables as of now. We could just search for this, but it doesn't seem like this will always be the case.
2. Borrow some of the logic the logic currently existing in `ZSTD_loadZstdDictionary()` in `zstd_compress.c` and/or `ZSTD_loadDEntropy()` in `zstd_decompress.c` and re-write an dict header/entropy-table parsing function that focuses purely on getting the size (so theoretically we could reuse/overwrite buffers as we go, since we don't care about copy the entropy tables to anywhere meaningful).
3. Use the already existing `ZSTD_loadDEntropy()` function which is essentially an auxiliary function in `ZSTD_decompress_insertDictionary()` that returns exactly the header size of the dictionary, and feed it a dummy `ZSTD_entropyDTables_t` object to let it write into as we get the header.

I took approach 3, since it seems like the cleanest way to integrate it into the ZDICT api without polluting it with a bunch of new code that is basically replicated elsewhere in the codebase already. 

Only downside is that it's a little wasteful/a bit slower, but this method isn't really made for performance-sensitive code anyways.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2775,10 +2775,6 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
     const BYTE* dictPtr = (const BYTE*)dict + 8;    /* skip magic num and dict ID */
     const BYTE* const dictEnd = dictPtr + dictSize;
 
-    ZSTD_STATIC_ASSERT(HUF_WORKSPACE_SIZE >= (1<<MAX(MLFSELog,LLFSELog)));
-    assert(dictSize >= 8);
-    assert(MEM_readLE32(dictPtr) == ZSTD_MAGIC_DICTIONARY);
-
     {   unsigned maxSymbolValue = 255;
         size_t const hufHeaderSize = HUF_readCTable((HUF_CElt*)bs->entropy.huf.CTable, &maxSymbolValue, dictPtr, dictEnd-dictPtr);
         RETURN_ERROR_IF(HUF_isError(hufHeaderSize), dictionary_corrupted);
@@ -2862,6 +2858,10 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
     unsigned offcodeMaxValue = MaxOff;
     size_t dictID;
     size_t eSize;
+
+    ZSTD_STATIC_ASSERT(HUF_WORKSPACE_SIZE >= (1<<MAX(MLFSELog,LLFSELog)));
+    assert(dictSize >= 8);
+    assert(MEM_readLE32(dictPtr) == ZSTD_MAGIC_DICTIONARY);
 
     dictID = params->fParams.noDictIDFlag ? 0 :  MEM_readLE32(dictPtr + 4 /* skip magic number */ );
     eSize = ZSTD_loadCEntropy(bs, workspace, offcodeNCount, &offcodeMaxValue, dict, dictSize);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2772,7 +2772,7 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
                          short* offcodeNCount, unsigned* offcodeMaxValue,
                          const void* const dict, size_t dictSize) 
 {
-    const BYTE* dictPtr = (const BYTE*)dict + 8;
+    const BYTE* dictPtr = (const BYTE*)dict + 8;    /* skip magic num and dict ID */
     const BYTE* const dictEnd = dictPtr + dictSize;
 
     {   unsigned maxSymbolValue = 255;
@@ -2869,7 +2869,7 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
     dictID = params->fParams.noDictIDFlag ? 0 :  MEM_readLE32(dictPtr);
     dictPtr += 4;
 
-    dictPtr += eSize - 8;   /* size of header + magic number already accounted for */
+    dictPtr += eSize - 8;
 
     {   size_t const dictContentSize = (size_t)(dictEnd - dictPtr);
         U32 offcodeMax = MaxOff;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2772,8 +2772,9 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
                          short* offcodeNCount, unsigned* offcodeMaxValue,
                          const void* const dict, size_t dictSize) 
 {
-    const BYTE* dictPtr = (const BYTE*)dict + 8;    /* skip magic num and dict ID */
+    const BYTE* dictPtr = (const BYTE*)dict;    /* skip magic num and dict ID */
     const BYTE* const dictEnd = dictPtr + dictSize;
+    dictPtr += 8;
 
     {   unsigned maxSymbolValue = 255;
         size_t const hufHeaderSize = HUF_readCTable((HUF_CElt*)bs->entropy.huf.CTable, &maxSymbolValue, dictPtr, dictEnd-dictPtr);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2768,37 +2768,12 @@ static size_t ZSTD_checkDictNCount(short* normalizedCounter, unsigned dictMaxSym
     return 0;
 }
 
-
-/* Dictionary format :
- * See :
- * https://github.com/facebook/zstd/blob/master/doc/zstd_compression_format.md#dictionary-format
- */
-/*! ZSTD_loadZstdDictionary() :
- * @return : dictID, or an error code
- *  assumptions : magic number supposed already checked
- *                dictSize supposed >= 8
- */
-static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
-                                      ZSTD_matchState_t* ms,
-                                      ZSTD_cwksp* ws,
-                                      ZSTD_CCtx_params const* params,
-                                      const void* dict, size_t dictSize,
-                                      ZSTD_dictTableLoadMethod_e dtlm,
-                                      void* workspace)
+size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
+                         short* offcodeNCount, unsigned* offcodeMaxValue,
+                         const void* const dict, size_t dictSize) 
 {
     const BYTE* dictPtr = (const BYTE*)dict;
-    const BYTE* const dictEnd = dictPtr + dictSize;
-    short offcodeNCount[MaxOff+1];
-    unsigned offcodeMaxValue = MaxOff;
-    size_t dictID;
-
-    ZSTD_STATIC_ASSERT(HUF_WORKSPACE_SIZE >= (1<<MAX(MLFSELog,LLFSELog)));
-    assert(dictSize >= 8);
-    assert(MEM_readLE32(dictPtr) == ZSTD_MAGIC_DICTIONARY);
-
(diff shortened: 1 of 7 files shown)


Cyan4973:
Question, notably for @terrelln : 
Should it be a `ZDICT_*` function, therefore intended to be used at dictionary building time,
or a `ZSTD_*` function, for usage in a broader range of use cases ?

Cyan4973:
@senhuang42 , try to squash your most minor commits once you figure out the correct code to pass tests. While a PR can have multiple commit, each commit should preferably add some meaningful step to the edifice.

Also, just FYI, note that you can run most automated tests directly from your fork, by subscribing your github account to Travis CI, Circle CI, Appveyor CI and/or Cirrus CI (all free). This may sometimes help to reduce amount of surprises at PR time.

terrelln:
I think it should be a ZDICT API, that way we have no obligation for it to be performant. I don't see any reason to strip the header, except during dictionary construction time, or dictionary analysis.

We don't want to encourage people to use this API to find the dictionary's header size at runtime, so they can only pass the content, since that wouldn't be efficient.

lib/dictBuilder/zdict.c
```diff
@@ -99,6 +100,17 @@ unsigned ZDICT_getDictID(const void* dictBuffer, size_t dictSize)
     return MEM_readLE32((const char*)dictBuffer + 4);
 }
 
+size_t ZDICT_getDictHeaderSize(const void* dictBuffer, size_t dictSize)
+{
+    if (dictSize <= 8 || MEM_readLE32(dictBuffer) != ZSTD_MAGIC_DICTIONARY) return 0;
+
+    {   ZSTD_entropyDTables_t dummyEntropyTables;
```
Cyan4973: `ZSTD_entropyDTables_t` is a 28 KB structure.
If it was a `ZSTD_*` function, it would be considered too large to store it on stack.
Though as a `ZDICT_*` entry with a more narrow use case, I'm less sure. 
@terrelln, what do you think ?

Cyan4973:
To be merged after the release (`libzstd`, including `ZDICT_*`,  is frozen right now).

lib/dictBuilder/zdict.c
terrelln: I'd prefer to allocate it in the heap. It might be okay as-is, but I'd rather be safe.

lib/dictBuilder/zdict.c
```diff
@@ -104,10 +104,11 @@ size_t ZDICT_getDictHeaderSize(const void* dictBuffer, size_t dictSize)
 {
     if (dictSize <= 8 || MEM_readLE32(dictBuffer) != ZSTD_MAGIC_DICTIONARY) return 0;
 
-    {   ZSTD_entropyDTables_t dummyEntropyTables;
-        size_t headerSize;
-        dummyEntropyTables.hufTable[0] = (HUF_DTable)((HufLog)*0x1000001);
-        headerSize = ZSTD_loadDEntropy(&dummyEntropyTables, dictBuffer, dictSize);
+    {   size_t headerSize;
+        ZSTD_entropyDTables_t* dummyEntropyTables = (ZSTD_entropyDTables_t*)malloc(sizeof(ZSTD_entropyDTables_t));
```
Cyan4973: Could you check `malloc()` success ?
It might return `NULL`.

senhuang42: Will do!

lib/dictBuilder/zdict.c
```diff
@@ -48,6 +48,7 @@
 #  define ZDICT_STATIC_LINKING_ONLY
 #endif
 #include "zdict.h"
+#include "decompress/zstd_decompress_internal.h" /* ZSTD_entropyDTables_t */
```
Cyan4973: In https://github.com/facebook/zstd/tree/dev/lib#modular-build, we have specified the following dependency graph : 

> lib/dictBuilder : (...) This module depends on both `lib/common` and `lib/compress` .

This `#include` changes that, by making `zdict` now _also_ dependent on `lib/decompress`.

I believe this is significant. It would for example break the build of `dictBuilder` if `lib/decompress` is not present.
(btw, it also shows that we are probably missing one test about this possible build).

I would prefer the dependency graph to not change.

Potential work-around (for discussion) : 
- Move `ZSTD_entropyDTables_t` definition somewhere else, likely in `lib/common`, for example `zstd_internal.h`. Make sure to add code comments to explain why it's there.
- More complex, and debatable : move `ZDICT_getDictHeaderSize` into `lib/common`, though it would also change its name (`ZSTD_*`) and its exposition path (through `zstd.h`). It also still requires `ZSTD_entropyDTables_t` in `lib/common`, since `lib/common` _must not_ depend on `lib/decompress`.

senhuang42: Ah, I did wonder if there were some specification limitations with what `zdict.c` could `#include`. I think the first option is better - I'd feel better with this in the `ZDICT_*` API since it's use-case isn't as generalized as those in `ZSTD_*`, and it is a more expensive function than it would seem at first glance.

However, another issue is that `ZSTD_loadDEntropy()` also would need to be moved to `zstd_internal.h`. So in total, we will be moving the symbols `ZSTD_loadDEntropy()`, `ZSTD_seqSymbol`, `ZSTD_entropyDTables_t`, and `SEQSYMBOL_TABLE_SIZE` into `zstd_internal.h`. It does feel a little weird to me to have these in `zstd_internal.h`, but it does fit the note at the top of `zstd_internal.h`:

```
/* this module contains definitions which must be identical
 * across compression, decompression and dictBuilder.
 * It also contains a few functions useful to at least 2 of them
 * and which benefit from being inlined */
```

Cyan4973: Something I'm wondering :
a dictionary can be loaded from `lib/compress`, since it can be used for compression.
Hence `lib/compress` must be able to parse the dictionary header.

Maybe there is a way to extract the information you are looking for (header size) while using only functions already present in `lib/common` and `lib/compress` ?

senhuang42: Ahh, I was wondering this when first approaching this issue, since I assumed that decompress and compress would have very similar dict parsing functions, but this is not really the case: `lib/compress` uses `ZSTD_loadZstdDictionary()` to parse the dictionary into the context and returns a dictID, while `lib/decompress` uses `ZSTD_loadDEntropy()` to parse the dictionary and returns the header size. 

`ZSTD_loadZstdDictionary()` only ephemerally computes a header size during the parsing, and it's function signature is much more complicated than `ZSTD_loadDEntropy()`:
```
static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
                                      ZSTD_matchState_t* ms,
                                      ZSTD_cwksp* ws,
                                      ZSTD_CCtx_params const* params,
                                      ZSTD_dictTableLoadMethod_e dtlm,
                                      void* workspace)
```
So I think there isn't really much that's directly usable in `lib/compress`. While I think it would be theoretically possible to generate the header size from functions in `lib/common`, it would be akin to taking approach #2 I outlined in my first comment, which I don't think is the best, since `ZSTD_loadDEntropy()` already does close to the minimum amount of work required to get the dictionary header size.

senhuang42: Went ahead and updated this with the first option you suggested, moving some symbols into `zstd_internal.h`. Let me know how that looks to you!

lib/dictBuilder/zdict.h
```diff
@@ -64,6 +64,7 @@ ZDICTLIB_API size_t ZDICT_trainFromBuffer(void* dictBuffer, size_t dictBufferCap
 
 /*======   Helper functions   ======*/
 ZDICTLIB_API unsigned ZDICT_getDictID(const void* dictBuffer, size_t dictSize);  /**< extracts dictID; @return zero if error (not a valid dictionary) */
+ZDICTLIB_API size_t ZDICT_getDictHeaderSize(const void* dictBuffer, size_t dictSize);  /* returns dict header size; returns zero if error (not a valid dictionary or mem alloc failure) */
```
Cyan4973: Opened question :

For consistency with other functions returning a `size_t`, 
should we rather use 
an error code, compatible with `ZDICT_isError()`, 
instead of `0`
when an error happens ?

cc @terrelln

Cyan4973:
Here are my current thoughts : 

It's not enough to transfer the _declaration_ into `lib/common/zstd_internal.h`,
the symbol itself still needs to be generated, which means that link stage will fail in absence of `lib/decompress`, where it's defined.

I think we were missing a test to show this dependency relationship, so I just added one in `/programs`, by defining a new target `zstd-dictBuilder`. Hopefully, it should show this impact.

For the currently proposed design to work, it's now necessary to also transfer the definition of the function `ZSTD_loadDEntropy()` into `lib/common`, alongside all required dependencies.

Intuitively, this starts to be a bit too many changes for the target objective : we are modifying the structure of the core library, by pushing code from `/decompress` into `/common`, for the sole benefit of a function pushed aside into `ZDICT_*` because its usefulness is considered "rare". This does not feel balanced.

As a consequence, I would prefer a solution which preserves the current structure of `/lib`.
This seems to point at using functions available in `lib/compress`, since they are also accessible from `lib/dictBuilder`.

Now, I understand that `ZSTD_loadZstdDictionary()` is too complex to invoke.
But looking into the source code, I believe this is superficial.

The first part of the function seems to cover a scope exactly equivalent to `ZSTD_loadDEntropy()`, and therefore could certainly be extracted to create a new function `ZSTD_loadCEntropy()`. Such new function only needs as arguments `ZSTD_compressedBlockState_t* bs` and `void* workspace`, thus being essentially a mirror of `ZSTD_loadDEntropy()`.
`ZSTD_loadZstdDictionary()` should obviously be refactored to invoke `ZSTD_loadCEntropy()`. `ZSTD_loadCEntropy()` would be declared in `zstd_compress_internal.h`, which can be included from `zdict.c`.

I believe this could work. Thoughts ?

lib/dictBuilder/zdict.h
terrelln: We should definitely return an error code when an error happens.

lib/compress/zstd_compress.c
```diff
@@ -2768,37 +2768,12 @@ static size_t ZSTD_checkDictNCount(short* normalizedCounter, unsigned dictMaxSym
...
- *                dictSize supposed >= 8
- */
-static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
-                                      ZSTD_matchState_t* ms,
-                                      ZSTD_cwksp* ws,
-                                      ZSTD_CCtx_params const* params,
-                                      const void* dict, size_t dictSize,
-                                      ZSTD_dictTableLoadMethod_e dtlm,
-                                      void* workspace)
+size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
```
senhuang42: We need to include `offcodeNCount[]` and `offcodeMaxValue` in this function's args, since they're modified during loading of entropy tables and then their modified values are used later in `ZSTD_loadZstdDictionary()`

Cyan4973:
Does that seem possible to squash some of these commits ?
Not all of them look very useful : 

> new function and include  a2db071
whitespaces   9b9ccc9
fix everything   108a771
formatting conventions?   69102d2
eh it looks better in the same line   f7a3911

Consider that commit points do not "disappear" from commit history, they become "stages" for potential future debugging by bisection

Cyan4973:
The patch is globally good. Great work @senhuang42 !

There is only one minor comment regarding commit history, 
and it's basically ready for a merge.

senhuang42:
Force pushed the new history with useless commits squashed out, I think this rewrites the old history correctly.

Cyan4973:
Excellent ! Thanks @senhuang42 !

PR #1868, opened 2019-11-05
Superblocks rebased for merge

senhuang42:
Superblocks patch was pretty old so had to make some merge conflict fixes in order to be able to merge this into dev automatically.

Not sure if the convention is to try to merge to dev directly from this PR, or to the superBlocks feature branch. Still a couple things to do to validate correctness.


Cyan4973:
already ??

senhuang42:
I still need to read over the code in detail to make sure it's doing what I think it is. Just wanted to put this draft PR here so that automated tests run on it (for some reason I can only get CircleCI and Travis CI on my forked repo).

lib/compress/zstd_compress.c
```diff
@@ -20,6 +20,7 @@
 #include "fse.h"
 #define HUF_STATIC_LINKING_ONLY
 #include "huf.h"
+#include "zstd_compress_superblock.h"
```
Cyan4973: nit: very minor,

the general idea is that `#include "*.h"` should be listed from more generic to more specific,
in which case, I would expect `"zstd_compress_superblock.h"` to come last, or at least after  `"zstd_compress_internal.h"`.

Cyan4973:
Looks good to me !

terrelln:
@senhuang42 looks like this causes a ~400 byte regression for small window logs https://circleci.com/gh/facebook/zstd/4888?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Is this expected? If so that is fine, and we just need to update `results.csv`. If not, please look into it.

senhuang42:
Hmm, could be some of the minor changes introduced into the main compression loop with this patch. I'll try to see if reverting those changes would fix this.

lib/compress/zstd_compress.c
```diff
@@ -2500,21 +2572,28 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
...
+                cSize = ZSTD_compressBlock_internal(cctx,
+                                        op+ZSTD_blockHeaderSize, dstCapacity-ZSTD_blockHeaderSize,
+                                        ip, blockSize, 0 /* frame */);
+                FORWARD_IF_ERROR(cSize);
+
+                if (cSize == 0) {  /* block is not compressible */
+                    cSize = ZSTD_noCompressBlock(op, dstCapacity, ip, blockSize, lastBlock);
+                    FORWARD_IF_ERROR(cSize);
+                } else {
+                    U32 const cBlockHeader24 = lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
```
Cyan4973: At this line,
the previous version used to differentiate `cSize==1`, to mean "RLE block".
Now the new code only creates `bt_compressed` blocks.

I believe that's because the "RLE block generation" patch has been created after the Superblock one.

Now, it's a mistake, because an RLE block of `cSize==1` will be labelled a compressed block, which is incorrect, and I guess it will become a corruption bug at decompression time.
It seems to prove that we are missing a test able to generate RLE block and check they are correctly encoded and decoded.

senhuang42: Yep, I just switched back to the old logic and the tests look good.

senhuang42: Just pushed a new commit onto this branch and it doesn't look like Github picks it up - should I open a new PR for the modification?

lib/compress/zstd_compress.c
```diff
@@ -2500,21 +2572,28 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
...
                 FORWARD_IF_ERROR(cSize);
             } else {
-                const U32 cBlockHeader = cSize == 1 ?
-                    lastBlock + (((U32)bt_rle)<<1) + (U32)(blockSize << 3) :
-                    lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
-                MEM_writeLE24(op, cBlockHeader);
-                cSize += ZSTD_blockHeaderSize;
+                cSize = ZSTD_compressBlock_internal(cctx,
+                                        op+ZSTD_blockHeaderSize, dstCapacity-ZSTD_blockHeaderSize,
+                                        ip, blockSize, 0 /* frame */);
```
terrelln: @senhuang42 the bug is on this line. This disables RLE blocks.

It also looks like RLE block support is buggy, and needs to be fixed up, as Yann pointed out.

lib/compress/zstd_compress.c
terrelln: > It seems to prove that we are missing a test able to generate RLE block and check they are correctly encoded and decoded.

The bug is hidden because we've disabled RLE block generation.

Cyan4973: Yes, a new PR will be needed.

Cyan4973:
So we do have some follow up after all : 
- fix the bug that made the compressor no longer generate RLE block
- fix the bug that made the compressor generate buggy header for RLE block (if it generated some)
- fix the test that should have noticed that the compressor was no longer generating RLE blocks.

This last test already partially exists, but it seems it doesn't check that the produced frame actually produces RLE block. This could be approximated by controlling compressed size, since RLE blocks are cheaper than any other kind of block for this kind of data.

lib/compress/zstd_compress.c
senhuang42: I can go ahead and create a test in which the compressor compresses something into an RLE block and prove that it can roundtrip.

Cyan4973: I believe we already have this test. But it wasn't checking that compressed data was actually using RLE blocks.
So, in this case, since RLE blocks were no longer generated, round trip was fine.

PR #1875, opened 2019-11-07
Change c90 to c89

ryandesign:
[c89 and c90 are the same language](https://en.wikipedia.org/wiki/ANSI_C#C90). Old compilers like gcc 4.2.1 don't know the c90 alias.

See https://trac.macports.org/ticket/59620


ryandesign:
> We really only use this target to test compatibility. It shouldn't be needed to build the project.

For the target in the Makefile, I'm sure that's true. I just changed that for completeness. The build failure occurred due to the flag in zlibWrapper/Makefile. See the [log attached to the MacPorts ticket](https://trac.macports.org/attachment/ticket/59620/main.log):

```
cc1: error: unrecognized command line option "-std=c90"
```

Issue #1877, opened 2019-11-07
windows (mingw-w64): name of import lib


vtorri:
on Windows, with mingw-w64, usually the extension of the import lib is not .lib but .dll.a (see https://github.com/facebook/zstd/blob/dev/lib/Makefile#L164 )

I can provide a PR if you want

terrelln:
That would be great!

vtorri:
see #1889

PR #1883, opened 2019-11-12
Fix MinGW cross

Ericson2314:
The batch file was very unnecessary.


programs/Makefile
```diff
@@ -88,6 +88,12 @@ ifneq (,$(filter x86_64%,$(shell $(CC) -dumpmachine)))
 else
     RES_FILE = $(RES32_FILE)
 endif
+RC ?= windres
+# http://stackoverflow.com/questions/708238/how-do-i-add-an-icon-to-a-mingw-gcc-compiled-executable
+$(RES64_FILE): windres/zstd.rc
```
Cyan4973: A rule in first position becomes the default rule (triggered when invoking `make` with no argument).

Currently, it's `zstd`, but with this patch, it becomes `$(RES64_FILE)`.

Possible work-around : add a `default` rule which delegates to `zstd`
early enough in the `Makefile` to ensure it is the first rule.

Ericson2314: I'll move it down

Ericson2314:
Should be all good now

Ericson2314:
CC @vtorri  @jdxAtGitHub

programs/Makefile
```diff
@@ -235,8 +235,16 @@ zstdmt: zstd
...
-generate_res:
-	windres/generate_res.bat
+generate_res: $(RES64_FILE) $(RES32_FILE)
+
+ifneq (,$(filter Windows%,$(OS)))
+RC ?= windres
+# http://stackoverflow.com/questions/708238/how-do-i-add-an-icon-to-a-mingw-gcc-compiled-executable
+$(RES64_FILE): windres/zstd.rc
+	$(RC) -o $@ -I ../lib -I windres -i $< -O coff -F pe-x86-64
+$(RES64_FILE): windres/zstd.rc
```
vtorri: is it really RES64_FILE ? not RES32_FILE ?

vtorri:
it would be nice to fix import lib name : see issue #1877

Ericson2314:
@vtorri it would, but I would do that separately as I was able to link fine with this name.

vtorri:
the name of the import lib can have .lib extension, but the static lib can also have .lib extension. That's what visual studio does, but it is not a good idea. libtool, and the vast majority of open source project uses .dll.a as extension to enforce the fact that it's the import lib

ok to do that separately

programs/Makefile
Ericson2314: Good catch

vtorri: you can also use a variable that store pe-x86-64 or pe-i386 (where you set RES_FILE) to avoid a check here

Ericson2314: `make generate_res` builds both for some reason, so I'll keep it as is (with fix).

Cyan4973:
Looks good to me !

The only point I wonder is : how to test this capability ?

I believe it's an important point, otherwise it would be easy to break this capability in any future patch.

I guess this should be done on AppveyorCI, our current Windows based CI.

Ericson2314:
Do what https://github.com/facebook/rocksdb does, it is also a Facebook project so shouldn't be too dificult.

https://github.com/facebook/rocksdb/blob/master/.travis.yml#L115-L116

vtorri:
i'm using zstd in my package manager and i  test each new release with MSYS2 + mingw-w64, so i also don't think it will bitrot

Cyan4973:
In https://travis-ci.org/facebook/zstd/jobs/617904275, I've added a `mingw` cross-compilation test to the `travisTest` branch, in preparation for this patch.

As expected, if fails, since it doesn't include this patch : 
> timefn.h:44:47: fatal error: Windows.h: No such file or directory
https://travis-ci.org/facebook/zstd/jobs/617904275#L343

I guess it's the issue of casing that this patch fixes.

Also notable, library compilation doesn't seem to like the `-fPIC` directive for `mingw` cross-compilation. 
That's only a problem if one wants to neutralize any kind of warning during CI tests, which is generally a good policy. In which case, this issue should be fixed too.
Another way would be to restrict the scope of the test, and by extension of the patch : is it just meant to compile only the CLI in `/programs` ? Or is library compilation also part of the scope ?

vtorri:
always use lowercase header file name on windows for cross compilation

Cyan4973:
The `mingw` cross-compilation test passes after this patch.
Still, it doesn't solve the `-fPIC` warning for the library,
so I think I will restrict the scenario to compilation of the CLI only.
This will make it possible to enforce the strictest policy with regards to zero-warning tolerance, associated with multiple correctness compilation flags.

Cyan4973:
redirecting the PR towards `travisTest`, where the `mingw`cross-compilation test is present.

Cyan4973:
Well, I was expected CI tests to be triggered after the change of destination branch, but apparently, they are not.

Merging. Tests to be continued onto `travisTest` branch.

Ericson2314:
Can you make an issue for the testing the library? I was building this to build rocksdb in https://github.com/NixOS/nixpkgs/pull/73316, so I can assure you the library is being cross compiled in practice.

Cyan4973:
The CLI-limited `mingw` cross-compilation test has been added in #1910.

Compiling the library is likely to work fine too, it's just that it generates a ton of `-fPIC` warnings.
The issue would be to find a way to avoid these warnings, so that CI test results can remain clean.

Ericson2314:
Right, the issue would mention the PIC warnings as the blocker.

Cyan4973:
#1911

Ericson2314:
Thanks! Subscribed

Issue #1886, opened 2019-11-14
man page misstake


xdu-diagnoa:
In `zstd.1.md` ADVANCED COMPRESSION OPTIONS,  max `hashLog`
> The minimum hlog is 6 (64 B) and the maximum is 26 (128 MiB).

pow(2,26) is 64MiB

PR #1887, opened 2019-11-14
Fixing hashLog error 2^26=64Mib

bimbashrestha:
https://github.com/facebook/zstd/issues/1886


Cyan4973:
While the _correction_ is valid,
I believe the value itself is outdated.

Maximum values are documented in `zstd.h`, at :
https://github.com/facebook/zstd/blob/v1.4.4/lib/zstd.h#L1042

I presume multiple "maximum" expressed in the documentation are likely outdated.
Let's try to get them closer to actual implementation.

PR #1888, opened 2019-11-15
RLE test and re-enable RLE in main compression loop

senhuang42:
Modifies the existing RLE test to compress into two RLE blocks instead of one, and now we also check that there is no significant size regression (arbitrarily set to 10% of 39 bytes) as a proxy for testing that the blocks were indeed compressed with RLE (without RLE it would compress to 47 bytes).

Curiously, the existing test for RLE would compress to the same size regardless of whether RLE blocks were enabled.

Also validated that previous patch will fail this test.

### DIFF
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -1999,14 +1999,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
 
     /* long rle test */
     {   size_t sampleSize = 0;
+        size_t expectedCompressedSize = 39; /* two RLE blocks, zstd 1.4.4 */
         DISPLAYLEVEL(3, "test%3i : Long RLE test : ", testNb++);
-        RDG_genBuffer(CNBuffer, sampleSize, compressibility, 0., seed+1);
         memset((char*)CNBuffer+sampleSize, 'B', 256 KB - 1);
         sampleSize += 256 KB - 1;
-        RDG_genBuffer((char*)CNBuffer+sampleSize, 96 KB, compressibility, 0., seed+2);
+        memset((char*)CNBuffer+sampleSize, 'A', 96 KB);
         sampleSize += 96 KB;
         cSize = ZSTD_compress(compressedBuffer, ZSTD_compressBound(sampleSize), CNBuffer, sampleSize, 1);
-        if (ZSTD_isError(cSize)) goto _output_error;
+        if (ZSTD_isError(cSize) || cSize > expectedCompressedSize * 1.1) goto _output_error;
         { CHECK_NEWV(regenSize, ZSTD_decompress(decodedBuffer, sampleSize, compressedBuffer, cSize));
           if (regenSize!=sampleSize) goto _output_error; }
         DISPLAYLEVEL(3, "OK \n");
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -1999,14 +1999,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
 
     /* long rle test */
     {   size_t sampleSize = 0;
-        size_t expectedCompressedSize = 39; /* block 1: RLE, block 2: compressed, block 3: RLE, zstd 1.4.4 */
+        size_t expectedCompressedSize = 39; /* block 1, 2: compressed, block 3: RLE, zstd 1.4.4 */
         DISPLAYLEVEL(3, "test%3i : Long RLE test : ", testNb++);
         memset((char*)CNBuffer+sampleSize, 'B', 256 KB - 1);
         sampleSize += 256 KB - 1;
         memset((char*)CNBuffer+sampleSize, 'A', 96 KB);
         sampleSize += 96 KB;
         cSize = ZSTD_compress(compressedBuffer, ZSTD_compressBound(sampleSize), CNBuffer, sampleSize, 1);
-        if (ZSTD_isError(cSize) || cSize > expectedCompressedSize * 1.1) goto _output_error;
+        if (ZSTD_isError(cSize) || cSize > expectedCompressedSize) goto _output_error;
         { CHECK_NEWV(regenSize, ZSTD_decompress(decodedBuffer, sampleSize, compressedBuffer, cSize));
           if (regenSize!=sampleSize) goto _output_error; }
         DISPLAYLEVEL(3, "OK \n");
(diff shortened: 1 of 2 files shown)


tests/fuzzer.c
```diff
@@ -1999,14 +1999,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
 
     /* long rle test */
     {   size_t sampleSize = 0;
+        size_t expectedCompressedSize = 39; /* two RLE blocks, zstd 1.4.4 */
         DISPLAYLEVEL(3, "test%3i : Long RLE test : ", testNb++);
-        RDG_genBuffer(CNBuffer, sampleSize, compressibility, 0., seed+1);
```
senhuang42: Correct me if I'm wrong, but it seems like neither of these `RDG_genBuffer()` actually do what we want? 

The first time it's called with `size = 0`, which means nothing gets generated. And I checked the compressed block that's the result of the second call, and it actually was not an RLE block. It seems like what we want to do is just generate two blocks of repeated characters, so it seems like we just want two memsets.

Cyan4973: You are completely right. This test doesn't help at all producing RLE blocks. It's no surprise it was ineffective in detecting RLE issues.

tests/fuzzer.c
```diff
@@ -1999,14 +1999,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
 
     /* long rle test */
     {   size_t sampleSize = 0;
+        size_t expectedCompressedSize = 39; /* two RLE blocks, zstd 1.4.4 */
```
Cyan4973: minor, but needs to be correctly reflected in comments :

The below test will actually produce one compressed block, and then one RLE block.

The reason is : by default, and especially at level 1, the algorithm makes no educated choice regarding block boundaries, and therefore always cut at 128 KB boundaries.

As a consequence, if I understand the test correctly, it will generate a first block with 96 KB of 'A' and 32 KB of 'B', which is not RLE. Then the second block will be `128 KB - 1` of 'B', and this one should be RLE.

That's still good enough for tests, it's just that the code comment should accurately reflects the outcome.
Also, manually check that, when the algorithm _doesn't_ produce RLE blocks, then the test correctly fires up.

P.S : Selecting block boundaries for optimal compression ratio could be an interesting follow up for high compression modes. It's complex though, so drop me a word if you feel interested.

senhuang42: Ah, good to know! I mistakenly assumed some boundary selection was already happening.

In that case, since we consume 96KB of 'A' and 256KB of 'B' in 128KB block increments, then it seems like there are three blocks right? First is 96KB 'A' + 32KB 'B', second is 128KB 'B', and 3rd is 96KB 'B', right?

Cyan4973: Sorry, I misread the code.
I though that the second `memset(,'A',)` was overwriting the first 96 KB of `'B'.
But since `sampleSize` is updated, it actually follows, so it's a different outcome.

I would predict 2 RLE blocks, the 1st and the 3rd.
Without intelligent boundary selection, the 2nd block cannot become RLE, because of the presence of a lone 'A' at the end of the block.

Still, good enough and appropriate for an RLE test.

senhuang42: Yep you're right, I just printed out the blocks to be sure. I've updated the comment again
```
test126 : Long RLE test : 
RLE
notRLE
RLE
```

tests/fuzzer.c
```diff
@@ -1999,14 +1999,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
 
     /* long rle test */
     {   size_t sampleSize = 0;
+        size_t expectedCompressedSize = 39; /* block 1: RLE, block 2: compressed, block 3: RLE, zstd 1.4.4 */
```
terrelln: Block 1 shouldn't be RLE, are you sure that it is encoded as RLE? It should be block1 compressed, block2 RLE.

tests/fuzzer.c
```diff
@@ -1999,14 +1999,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
         DISPLAYLEVEL(3, "test%3i : Long RLE test : ", testNb++);
-        RDG_genBuffer(CNBuffer, sampleSize, compressibility, 0., seed+1);
         memset((char*)CNBuffer+sampleSize, 'B', 256 KB - 1);
         sampleSize += 256 KB - 1;
-        RDG_genBuffer((char*)CNBuffer+sampleSize, 96 KB, compressibility, 0., seed+2);
+        memset((char*)CNBuffer+sampleSize, 'A', 96 KB);
         sampleSize += 96 KB;
         cSize = ZSTD_compress(compressedBuffer, ZSTD_compressBound(sampleSize), CNBuffer, sampleSize, 1);
-        if (ZSTD_isError(cSize)) goto _output_error;
+        if (ZSTD_isError(cSize) || cSize > expectedCompressedSize * 1.1) goto _output_error;
```
terrelln: I'd say allow zero margin here.

tests/fuzzer.c
terrelln: It shouldn't be RLE because we should be disallowing the first block to be RLE, because of a decoder bug.

senhuang42: I added the following check in `ZSTD_compressBlock_internal()` to look for RLE:
```
if (ZSTD_isRLE(ip, srcSize)) fprintf(stderr, "RLE\n");
else fprintf(stderr, "notRLE\n");
```
But just noticed the fix at the bottom to prevent emitting first block as RLE, so yeah, the first block wouldn't be RLE. 

If the second block is indeed just 128KB of 'B', then I suppose the compressor should make it RLE, but according to the `ZSTD_isRLE()` function, which looks like it determines whether or not an input block will be compressed into an RLE block, it returns false on the second block during this test.

Cyan4973: The second block seems to consists of 128 KB __- 1 bytes__ of 'B' and then one lone 'A', preventing it from being encoded as RLE.

If that's not what you want, you could change the size of 'B' section to be exactly 128 KB instead (256 KB in total with first block).

senhuang42: Aha, my mistake was remembering the compression format wrong: I thought that the compressor read the input stream backwards for some reason (in addition to the decompressor), which would mean: 
`block 1: 96KB 'A' + 32KB 'B'`, `block 2: 128KB 'B'`, and `block 3: 96KB - 1byte 'B'`, and therefore block 2-3 are RLE. But read the correct way, then block 3 should be the only RLE block.

As long as one block is RLE, the test will still serve its purpose.

PR #1889, opened 2019-11-15
shared library: rename import library with .dll.a extension

vtorri:


terrelln:
Thanks for the patch @vtorri!

PR #1890, opened 2019-11-15
[fuzz] Iteratively double output buffer size of stream_round_trip

bimbashrestha:
Double the output buffer size up to ten times if we fail the the test because of dst too small error.


bimbashrestha:
After some more investigation, the problem was deeper than what this pr addresses. Closing.

PR #1891, opened 2019-11-16
[fuzz] Superblock fuzz issues

bimbashrestha:
Fixing 2 fuzzer issues
* makes sure enough room is present for checksum in superblock
* output a regular uncompressed block if compressSequences fails on dstSize_tooSmall error


Cyan4973:
Great fix @bimbashrestha !

Could you also add a unit test able to catch this bug please ?

lib/compress/zstd_compress.c
```diff
@@ -2493,12 +2496,18 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
                     srcSize,
                     zc->entropyWorkspace, HUF_WORKSPACE_SIZE /* statically allocated in resetCCtx */,
                     zc->bmi2);
-            if (!ZSTD_isError(cSize) && cSize != 0) {
+
```
terrelln: We need to check `cSize` for an error.

lib/compress/zstd_compress.c
```diff
@@ -2493,12 +2496,18 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
...
+                 * uncompressed block */
+                cSize = ZSTD_noCompressBlock(dst, dstCapacity, src, srcSize, lastBlock);
+                FORWARD_IF_ERROR(cSize);
+            } else {
                 U32 const cBlockHeader24 = lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
                 MEM_writeLE24(ostart, cBlockHeader24);
                 cSize += ZSTD_blockHeaderSize;
             }
-        }
+        } else return cSize;
```
terrelln: Please reverse the if condition. `if (!blah) return cSize;`

A common pattern is to have "guards" or "short-circuits" that return if a condition isn't met.  In this function we want to return as soon as we hit an error we can't recover from. This keeps the nesting level down, and keeps the control flow linear, which is easy to read.

lib/compress/zstd_compress.c
```diff
@@ -2480,11 +2480,13 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
...
          * In such case, fall back to normal compression. This is possible because
          * targetCBlockSize is best effort not a guarantee. */
-        if (cSize == ERROR(dstSize_tooSmall) || (dstCapacity - cSize) < 4) {
-            /* We check (dstCapacity - cSize) < 4 above because we have to make sure
+        if (cSize != ERROR(dstSize_tooSmall) && (dstCapacity - cSize) >= 4)
+            /* We check (dstCapacity - cSize) >= 4 above because we have to make sure
              * to leave enough room for the checksum that will eventually get added in
              * the epilogue. Otherwise, we're just going to throw the dstSize_tooSmall
              * error there instead of here */
+            return cSize;
```
terrelln: This is a bug. We cannot early return non-error cases, because then we miss the entropy confirmation.

Do we need to `return` in this case? Can we go back to the original `if` statement, but just delete the `else`?

lib/compress/zstd_compress.c
```diff
@@ -2507,7 +2510,7 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
                 MEM_writeLE24(ostart, cBlockHeader24);
                 cSize += ZSTD_blockHeaderSize;
             }
-        } else return cSize;
+        }
     }
 
     if (!ZSTD_isError(cSize) && cSize != 0) {
```
terrelln: We cannot skip this block when we emit a compressed block.

bimbashrestha:
I can't seem to figure out why the dictionary_round_trip is now failing. Here is the end of the debug trace. Any thoughts? 

`../../lib/decompress/zstd_decompress_block.c: ZSTD_decompressSequences_body 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decompressSequences_body: after decode loop, remaining nbSeq : 0 
../../lib/decompress/zstd_decompress.c: ZSTD_copyRawBlock 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decompressBlock_internal (size : 62) 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decodeLiteralsBlock 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decodeLiteralsBlock : 11 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decodeSeqHeaders 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decompressSequences 
../../lib/decompress/zstd_decompress_block.c: ZSTD_decompressSequences_body 
../../lib/decompress/zstd_decompress_block.c:675: ERROR!: check oMatchEnd > oend failed, returning ERROR(dstSize_tooSmall): last match must fit within dstBuffer
dictionary_round_trip.c: 97: Assertion: !ZSTD_isError(result) failed. Destination buffer is too small
Aborted (core dumped)`

terrelln:
I would check and see what new code is being triggered that causes the problem. Is `(dstCapacity - cSize) < 4`? Are we emitting an uncompressed block?

bimbashrestha:
Ah I think I found it. We can't rely on cSize == 0 to discriminate noCompress blocks when we confirm rep codes. Checking explicitly now.

tests/fuzzer.c
```diff
@@ -490,6 +489,19 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        /* This tests whether or not we leave enough room for the checksum at the end
+         * of the dst buffer. The bug that motivated this test was found by the
+         * stream_round_trip fuzzer but this crashes for the same reason and is
+         * far more compact than re-creating the stream_round_trip fuzzer's code path */
+        ZSTD_CCtx *cctx = ZSTD_createCCtx();
+        ZSTD_CCtx_setParameter(cctx, ZSTD_c_targetCBlockSize, 64);
+        assert(!ZSTD_isError(ZSTD_compress2(cctx, compressedBuffer, 1339, CNBuffer, 1278)));
+        ZSTD_freeCCtx(cctx);
+    }
+    DISPLAYLEVEL(3, "OK \n");
```
bimbashrestha: Checked that this fails before this patch.

lib/compress/zstd_compress.c
```diff
@@ -2491,15 +2496,22 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
...
+            FORWARD_IF_ERROR(cSize);
+            if (cSize == 0) {
+                /* If compressSequences didn't work, we just output a regular
+                 * uncompressed block */
+                cSize = ZSTD_noCompressBlock(dst, dstCapacity, src, srcSize, lastBlock);
+                FORWARD_IF_ERROR(cSize);
+            } else {
                 U32 const cBlockHeader24 = lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
                 MEM_writeLE24(ostart, cBlockHeader24);
                 cSize += ZSTD_blockHeaderSize;
```
terrelln: Do we need to set `noCompressSuperBlock = 0` in this branch?

lib/compress/zstd_compress.c
```diff
@@ -2455,6 +2455,7 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
                                const void* src, size_t srcSize,
                                U32 lastBlock) {
     size_t cSize = 0;
+    int usingNoCompressSuperBlock = 0;
```
terrelln: nit: Lets call this `compressedSuperBlock` to avoid negated booleans.

lib/compress/zstd_compress.c
terrelln: Assuming that is correct, can we add a test that catches this bug, since it indicates lacking test coverage.

I would be okay if you can show that `fuzzer.c`, `zstreamtest.c`, or any of the libfuzzer fuzzers can find the bug within 1 hour. I just want to know that if we landed this change as-is, we would be alerted of the bug before we cut a release. Ideally we would know before landing.

lib/compress/zstd_compress.c
```diff
@@ -2491,15 +2496,22 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
...
                 U32 const cBlockHeader24 = lastBlock + (((U32)bt_compressed)<<1) + (U32)(cSize << 3);
                 MEM_writeLE24(ostart, cBlockHeader24);
                 cSize += ZSTD_blockHeaderSize;
             }
         }
     }
 
-    if (!ZSTD_isError(cSize) && cSize != 0) {
+    if (!ZSTD_isError(cSize) && !usingNoCompressSuperBlock) {
         /* confirm repcodes and entropy tables when emitting a compressed block */
```
Cyan4973: OK, something I'm not completely clear here :
the comment states : 
> confirm repcodes and entropy tables when emitting a compressed block

So, obviously, when `ZSTD_isError(cSize)`, no compressed block is emitted, since nothing is generated.

However, when `(!usingNoCompressSuperBlock)`, there is still a change that a compress block was sent. It's just no longer divided into sub-blocks.

So I'm now wondering what would happen in this case, especially if a `repcode` is used in the next block.

lib/compress/zstd_compress.c
```diff
@@ -2513,7 +2515,7 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
         }
     }
 
-    if (!ZSTD_isError(cSize) && !usingNoCompressSuperBlock) {
+    if (!ZSTD_isError(cSize) && compressSuperBlock) {
```
bimbashrestha: > However, when (!usingNoCompressSuperBlock), there is still a possibility that a compress block was sent. It's just no longer divided into sub-blocks.

I think this was a mistake. Now we should only enter this branch if the block sent is compressed. And this should take care of the rep codes too I believe.

lib/compress/zstd_compress.c
```diff
@@ -2455,6 +2455,7 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
                                const void* src, size_t srcSize,
                                U32 lastBlock) {
     size_t cSize = 0;
+    int compressSuperBlock = 1;
```
Cyan4973: I believe the name of this variable is slightly misleading.
`compressSuperBlock` seems to underline a link with `SuperBlock`, but if I do understand the code properly, there is no such direct link, as this variable only tells if the generated block is compressed or not.
Thus it should rather be called `compressedBlock` or something equivalent.

bimbashrestha: Ah yes. good point

lib/compress/zstd_compress.c
```diff
@@ -2471,7 +2471,7 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
      */
     if (cSize == 0) {
         cSize = ZSTD_noCompressSuperBlock(dst, dstCapacity, src, srcSize, zc->appliedParams.targetCBlockSize, lastBlock);
-        compressSuperBlock = 0;
+        bss = ZSTDbss_noCompress;
```
bimbashrestha: @Cyan4973 I think maintaining a condition might be a reasonable option from the ones I considered for this after we talked. Creating a new bool wasn't necessary though. We can just reuse bss. The other options were:

1. Create a new function for the confirm repcodes and entropy block and invoke it within the branches as appropriate. I don't know if adding a new function call for a rather trivial operation is worth it. It doesn't necessarily make this much easier to read as the function call would have to be inside the branches (notably following the else at the bottom of the first branch). The upside of having a new function would mean we could call it from compressBlock_internal where the repocode and entropy checking code is currently duplicated. But this is an experimental branch so I don't know if changing something that part of the stable flow is a good idea for this feature's sake alone.

2. We could refactor both this function as well as frameChunk and compressBlock_internal to reduce code duplication. I think this would yield the most elegant solution but again, this is an experimental branch and changing parts of the stable flow might not be a good idea.

Cyan4973:
Let's discuss that again tomorrow.

lib/compress/zstd_compress.c
```diff
@@ -2457,54 +2457,46 @@ static void ZSTD_confirmRepcodesAndEntropyTables(ZSTD_CCtx* zc)
...
+                               const size_t bss, void* dst, size_t dstCapacity,
                                const void* src, size_t srcSize,
                                U32 lastBlock)
 {
-    /* Superblock compression was successful */
-    if (cSize != 0) {
-        ZSTD_confirmRepcodesAndEntropyTables(zc);
-        return cSize;
+    /* Attempt superblock compression and return early if successful */
+    {
```
Cyan4973: minor : one level of indentation is probably unnecessary here, since there is no variable declaration needed.

lib/compress/zstd_compress.c
```diff
@@ -2457,54 +2457,46 @@ static void ZSTD_confirmRepcodesAndEntropyTables(ZSTD_CCtx* zc)
...
+        if (bss == ZSTDbss_compress) {
+            size_t cSize = ZSTD_compressSuperBlock(zc, dst, dstCapacity, lastBlock);
```
Cyan4973: minor : `size_t const cSize =`.
Try to `const` any variable that's not modified after its creation.

Sure, it's a few more keystrokes, but it offers a useful signal to the reader : 
this variable is not going to change, so I can "trust" its value.

Bonus, and probably as important : when this rule is applied consistently, now when a variable is _not_ `const`, it's supposed to mean that this variable _is_ going to change during its lifetime, so the reader can focus its attention on it. This is an important signal because mutable variables are generally the ones where bugs happen.

As usual, the goal is to simplify the reader's understanding.

There are several other variables below that can also be made `const`.

lib/compress/zstd_compress.c
```diff
@@ -2517,13 +2509,17 @@ static size_t ZSTD_compressBlock_targetCBlockSize(ZSTD_CCtx* zc,
...
     size_t cSize = 0;
+    const size_t bss = ZSTD_buildSeqStore(zc, src, srcSize);
     DEBUGLOG(5, "ZSTD_compressBlock_targetCBlockSize (dstCapacity=%u, dictLimit=%u, nextToUpdate=%u, srcSize=%zu)",
                 (unsigned)dstCapacity, (unsigned)zc->blockState.matchState.window.dictLimit, (unsigned)zc->blockState.matchState.nextToUpdate, srcSize);
-    cSize = ZSTD_compressBlock_targetCBlockSize_init(zc, dst, dstCapacity, src, srcSize, lastBlock);
-    FORWARD_IF_ERROR(cSize);
-    cSize = ZSTD_compressBlock_targetCBlockSize_body(zc, cSize, dst, dstCapacity, src, srcSize, lastBlock);
+    FORWARD_IF_ERROR(bss);
+
+    cSize = ZSTD_compressBlock_targetCBlockSize_body(zc, bss, dst, dstCapacity, src, srcSize, lastBlock);
```
Cyan4973: minor : order or parameters.

In general, try to provide mutable states and destination buffers first, 
then source and read-only parameters.

`bss` qualifies as a read-only parameter. 
So it should be positioned in the last part, around `lastBlock`.

Cyan4973:
Good ! This is getting pretty close !

Only a few minor details remaining.

terrelln:
Looks like there are some new OSS-Fuzz issues with this patch.

terrelln:
It may help to run the fuzzers locally for ~1 hour once you have the fix, in case there are more bugs. You can follow the instructions in `tests/fuzz/README.md`.

terrelln:
It could be a case of this bug blocking oss-fuzz from making progress and now it can find a new one.

PR #1892, opened 2019-11-16
Decompression speed optimizations

terrelln:
Silesia level 1

| Compiler | Before | After | Gain |
|---|---|---|---|
| gcc-9.2 | 1192.7 MB/s | 1256.7 MB/s | +5.3% |
| clang-9.0.0 | 1083 MB/s | 1203.2 MB/s | +11.1% |

enwik8 level 1

| Compiler | Before | After | Gain |
|---|---|---|---|
| gcc-9.2 | 1035.6 MB/s | 1103.8 MB/s | +6.5% |
| clang-9.0.0 | 923.2 MB/s | 1044.8 MB/s | +13.1% |

enwik8 level 22 (`ZSTD_decompressSequencesLong()`)

| Compiler | Before | After | Gain |
|---|---|---|---|
| gcc-9.2 | 739.2 MB/s | 781.2 MB/s | +5.6% |
| clang-9.0.0 | 694.5 MB/s | 746.3 MB/s | +7.5% |

Streaming decompression of enwik8 and silesia at level 1: `zstd -t enwik8.zst silesia.tar.zst`

| Compiler | Before | After | Gain |
|---|---|---|---|
| gcc-9.2 | 927.3 MB/s | 970.3 MB/s | +4.6%|
| clang-9.0.0 | 847.1 MB/s | 935.5 MB/s | +10.4% |

Measured with `perf stat -r 10 ./build/programs/zstd -t ../enwik8.zst ../silesia.tar.zst` gcc-9 does 6% less instructions and 10% less branches, and clang-9.2 does 15% less instructions and 9% less branches. Each with slightly lower IPC.

I ran the `tests/fuzzer` decompression fuzzers for about an hour, and ran `fuzzer.c` and `zstreamtest.c` for 10K iterations. There aren't any obvious bugs found, but more intensive fuzzer tests may expose some.

### DIFF
--- a/lib/decompress/huf_decompress.c
+++ b/lib/decompress/huf_decompress.c
@@ -282,6 +282,7 @@ HUF_decompress4X1_usingDTable_internal_body(
     {   const BYTE* const istart = (const BYTE*) cSrc;
         BYTE* const ostart = (BYTE*) dst;
         BYTE* const oend = ostart + dstSize;
+        BYTE* const olimit = oend - 3;
         const void* const dtPtr = DTable + 1;
         const HUF_DEltX1* const dt = (const HUF_DEltX1*)dtPtr;
 
@@ -306,9 +307,9 @@ HUF_decompress4X1_usingDTable_internal_body(
         BYTE* op2 = opStart2;
         BYTE* op3 = opStart3;
         BYTE* op4 = opStart4;
-        U32 endSignal = BIT_DStream_unfinished;
         DTableDesc const dtd = HUF_getDTableDesc(DTable);
         U32 const dtLog = dtd.tableLog;
+        U32 endSignal = 1;
 
         if (length4 > cSrcSize) return ERROR(corruption_detected);   /* overflow */
         CHECK_F( BIT_initDStream(&bitD1, istart1, length1) );
@@ -317,8 +318,7 @@ HUF_decompress4X1_usingDTable_internal_body(
         CHECK_F( BIT_initDStream(&bitD4, istart4, length4) );
 
         /* up to 16 symbols per loop (4 symbols per stream) in 64-bit mode */
-        endSignal = BIT_reloadDStream(&bitD1) | BIT_reloadDStream(&bitD2) | BIT_reloadDStream(&bitD3) | BIT_reloadDStream(&bitD4);
-        while ( (endSignal==BIT_DStream_unfinished) && (op4<(oend-3)) ) {
+        for ( ; (endSignal) & (op4 < olimit) ; ) {
             HUF_DECODE_SYMBOLX1_2(op1, &bitD1);
             HUF_DECODE_SYMBOLX1_2(op2, &bitD2);
             HUF_DECODE_SYMBOLX1_2(op3, &bitD3);
@@ -335,10 +335,10 @@ HUF_decompress4X1_usingDTable_internal_body(
             HUF_DECODE_SYMBOLX1_0(op2, &bitD2);
             HUF_DECODE_SYMBOLX1_0(op3, &bitD3);
             HUF_DECODE_SYMBOLX1_0(op4, &bitD4);
-            BIT_reloadDStream(&bitD1);
-            BIT_reloadDStream(&bitD2);
-            BIT_reloadDStream(&bitD3);
-            BIT_reloadDStream(&bitD4);
+            endSignal &= BIT_reloadDStreamFast(&bitD1) == BIT_DStream_unfinished;
+            endSignal &= BIT_reloadDStreamFast(&bitD2) == BIT_DStream_unfinished;
+            endSignal &= BIT_reloadDStreamFast(&bitD3) == BIT_DStream_unfinished;
+            endSignal &= BIT_reloadDStreamFast(&bitD4) == BIT_DStream_unfinished;
         }
 
         /* check corruption */
@@ -757,7 +757,6 @@ HUF_decompress1X2_usingDTable_internal_body(
     return dstSize;
 }
 
-
 FORCE_INLINE_TEMPLATE size_t
 HUF_decompress4X2_usingDTable_internal_body(
           void* dst,  size_t dstSize,
@@ -769,6 +768,7 @@ HUF_decompress4X2_usingDTable_internal_body(
     {   const BYTE* const istart = (const BYTE*) cSrc;
         BYTE* const ostart = (BYTE*) dst;
         BYTE* const oend = ostart + dstSize;
+        BYTE* const olimit = oend - (sizeof(size_t)-1);
         const void* const dtPtr = DTable+1;
         const HUF_DEltX2* const dt = (const HUF_DEltX2*)dtPtr;
 
@@ -793,7 +793,7 @@ HUF_decompress4X2_usingDTable_internal_body(
         BYTE* op2 = opStart2;
         BYTE* op3 = opStart3;
         BYTE* op4 = opStart4;
-        U32 endSignal;
+        U32 endSignal = 1;
         DTableDesc const dtd = HUF_getDTableDesc(DTable);
         U32 const dtLog = dtd.tableLog;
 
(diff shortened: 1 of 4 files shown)


lib/decompress/huf_decompress.c
```diff
@@ -270,6 +278,16 @@ HUF_decompress1X1_usingDTable_internal_body(
     return dstSize;
 }
 
+MEM_STATIC int BIT_reloadDStreamFast(BIT_DStream_t* bitD)
```
terrelln: This helps gcc, but is surprisingly neutral for clang.

The idea is that we don't need to include all the end of stream code in the main loop body, since we have the completion calls at the end anyways. This removes branches and code from the large hot loop.

lib/decompress/huf_decompress.c
```diff
@@ -282,6 +300,7 @@ HUF_decompress4X1_usingDTable_internal_body(
     {   const BYTE* const istart = (const BYTE*) cSrc;
         BYTE* const ostart = (BYTE*) dst;
         BYTE* const oend = ostart + dstSize;
+	BYTE* const olimit = oend - 3;
```
terrelln: Putting this in a variable helps clang.

lib/decompress/huf_decompress.c
```diff
@@ -335,10 +353,10 @@ HUF_decompress4X1_usingDTable_internal_body(
             HUF_DECODE_SYMBOLX1_0(op2, &bitD2);
             HUF_DECODE_SYMBOLX1_0(op3, &bitD3);
             HUF_DECODE_SYMBOLX1_0(op4, &bitD4);
-            BIT_reloadDStream(&bitD1);
-            BIT_reloadDStream(&bitD2);
-            BIT_reloadDStream(&bitD3);
-            BIT_reloadDStream(&bitD4);
+            endSignal &= BIT_reloadDStreamFast(&bitD1);
```
terrelln: Fast is slightly better here, but no large difference (< 5 MB/s).

lib/decompress/huf_decompress.c
```diff
@@ -822,8 +861,11 @@ HUF_decompress4X2_usingDTable_internal_body(
             HUF_DECODE_SYMBOLX2_0(op2, &bitD2);
             HUF_DECODE_SYMBOLX2_0(op3, &bitD3);
             HUF_DECODE_SYMBOLX2_0(op4, &bitD4);
-
-            endSignal = BIT_reloadDStream(&bitD1) | BIT_reloadDStream(&bitD2) | BIT_reloadDStream(&bitD3) | BIT_reloadDStream(&bitD4);
+	    endSignal &= BIT_reloadDStreamFast(&bitD1);
```
terrelln: Using `BIT_reloadDStreamFast()` makes a huge difference for gcc, like +3%.

lib/decompress/huf_decompress.c
```diff
@@ -804,8 +822,29 @@ HUF_decompress4X2_usingDTable_internal_body(
         CHECK_F( BIT_initDStream(&bitD4, istart4, length4) );
 
         /* 16-32 symbols per loop (4-8 symbols per stream) */
-        endSignal = BIT_reloadDStream(&bitD1) | BIT_reloadDStream(&bitD2) | BIT_reloadDStream(&bitD3) | BIT_reloadDStream(&bitD4);
-        for ( ; (endSignal==BIT_DStream_unfinished) & (op4<(oend-(sizeof(bitD4.bitContainer)-1))) ; ) {
+        for ( ; (endSignal) & (op4 < olimit); ) {
+#ifdef __clang__
```
terrelln: Clang likes this ordering better, +2%.

lib/decompress/zstd_decompress_block.c
```diff
@@ -715,7 +715,7 @@ size_t ZSTD_execSequence(BYTE* op,
 
     /* Errors and uncommon cases handled here. */
     assert(oLitEnd < oMatchEnd);
-    if (iLitEnd > litLimit || oMatchEnd > oend_w)
+    if (UNLIKELY(iLitEnd > litLimit || oMatchEnd > oend_w))
```
terrelln: All the LIKELY() and UNLIKELY() I've added actually make a difference in perf. Normally neutral for gcc and helpful for clang.

lib/decompress/zstd_decompress_block.c
```diff
@@ -819,19 +827,20 @@ FORCE_INLINE_TEMPLATE seq_t
...
+    BYTE const mlBits = mlDInfo.nbAdditionalBits;
+    BYTE const ofBits = ofDInfo.nbAdditionalBits;
+    BYTE const totalBits = llBits+mlBits+ofBits;
 
     /* sequence */
     {   size_t offset;
-        if (!ofBits)
-            offset = 0;
-        else {
+        if (ofBits > 1) {
```
terrelln: I've rewritten this part to reduce the number of branches.

I got the idea from gcc, which was already doing a rewriting similar to this. So it is about neutral for gcc, but helps clang a few percent.

lib/decompress/zstd_decompress_block.c
```diff
@@ -845,49 +854,63 @@ ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets)
...
+    if (llBits > 0)
+        seq.litLength += BIT_readBitsFast(&seqState->DStream, llBits/*>0*/);
     if (MEM_32bits())
         BIT_reloadDStream(&seqState->DStream);
 
     DEBUGLOG(6, "seq: litL=%u, matchL=%u, offset=%u",
                 (U32)seq.litLength, (U32)seq.matchLength, (U32)seq.offset);
 
     /* ANS state update */
+#if defined(__GNUC__) && !defined(__clang__)
```
terrelln: The other version is really bad for gcc, but I don't really know why. It should be better...

lib/decompress/zstd_decompress_block.c
```diff
@@ -928,17 +952,19 @@ ZSTD_decompressSequences_body( ZSTD_DCtx* dctx,
...
 
-        for ( ; (BIT_reloadDStream(&(seqState.DStream)) <= BIT_DStream_completed) && nbSeq ; ) {
-            nbSeq--;
-            {   seq_t const sequence = ZSTD_decodeSequence(&seqState, isLongOffset);
-                size_t const oneSeqSize = ZSTD_execSequence(op, oend, sequence, &litPtr, litEnd, prefixStart, vBase, dictEnd);
-                DEBUGLOG(6, "regenerated sequence size : %u", (U32)oneSeqSize);
-                if (ZSTD_isError(oneSeqSize)) return oneSeqSize;
-                op += oneSeqSize;
-        }   }
+        for ( ; ; ) {
```
terrelln: Rewriting the loop in this way slightly helps gcc and clang. gcc really likes the `error = oneSeqSize` trick. Breaking early hurts speed, probably stops some optimization pass.

terrelln:
This is ready for review now

Cyan4973:
Great gains !

caoyzh:
The optimizations of modifications in “lib/decompress/huf_decompress.c” is not very well in aarch64，it would be good if you could give us some more info about your test environment. Thanks.

terrelln:
Hi @caoyzh, what platform are you testing on (e.g. raspberry pi or a particular phone)? I did my testing on an Intel i9-9900K. I also tested on a 2018 Macbook Pro and an Intel Skylake processor.

I'm currently looking into mobile performance, so I can look into this once I know what platform you are seeing the regression on.

lib/decompress/huf_decompress.c
terrelln: Never mind, you are compiling with gcc.

terrelln:
@caoyzh what is the raw speed of gcc before/after? Can you compare it to clang?

Yikun:
@terrelln 
> What platform are you testing on (e.g. raspberry pi or a particular phone)?

Thanks for reply, the platform we are testing is the server with [Kunpeng processer](http://www.hisilicon.com/en/Products/ProductList/Kunpeng) (which is a 64-bit Arm-compatible server CPU released by Huawei Hisilicon).

caoyzh:
I test it on KungPeng based on armv8.
the raw speed is as follow:
搜狗截图20200123114019
the average decompression optimizations of Level 1~19 in clang-9.0.0 is -1.19%

terrelln:
> Could you give some explanation about the optimization idea？The code is more complex than before, but I’m not very sure if the generated assembly will be simpler. Thanks.

@caoyzh can you please reply in the main thread. I don't seem to be able to view your comment on GitHub, maybe because of a GitHub bug. So I'm replying here.

The optimization is to replace `BIT_reloadDStream()` with `BIT_reloadDStreamFast()`, which has less code, so the loop is smaller. The edge cases are pulled out and handled in the `HUF_decodeStream*()` calls below. The cost is in the `HUF_decompress4X1` loop we now have to check the `endSignal`. The check isn't expensive, and should have been done anyways. The check is already present in the `HUF_decompress4X2` loop.

I've found the compilers to be very picky about this code, and it is tricky to get an incantation that works well for all compilers and architectures. The more architectures we add to the matrix, the harder it gets.

I'm sorry to say that I'm not going to be able to debug the performance on your server. Without access to the hardware it will be impossible. However, I will accept a patch that fixes the performance, as long as it doesn't hurt any of the architecture/compiler combinations that we care about most. Since it is only a 1-1.5% regression, I don't think it is worth having a conditional compilation that has special code for your compiler/architecture combination. If there were a 10% regression I would consider it.

I'm going to put up a patch later today or this week that slightly improves aarch64 performance with clang on mobile. It isn't an optimization really, just a slight rewrite to help guide the compiler. Maybe it will help your performance.

The architectures and compilers we care about most, in order, are:
* x86_64 clang and gcc
* aarch64 and armv7 clang decompression only (phones)
* i386 clang and gcc
* everything else

Of course, we care about correctness on every architecture deeply, and will always fix any correctness bugs.

PR #1893, opened 2019-11-19
Add Tool to Diagnose Whether Corrupt Blobs are Plausibly Bit-Flips

felixhandte:
I spend an increasing amount of my time looking at "Corrupted block detected" failures in decompression. Not infrequently, I suspect that it is the result of hardware failure, and that the blob has become bit-flipped or otherwise corrupted somewhere along the line.

For that reason I was motivated to write a little tool to inspect blobs that fail to decompress, to try modifying them, and then check whether they decompress successfully. This seems like potentially a generally useful tool, so I figured it might be worth putting in `contrib/`.

### DIFF
--- a/contrib/diagnose_corruption/.gitignore
+++ b/contrib/diagnose_corruption/.gitignore
@@ -1,0 +1,1 @@
+check_flipped_bits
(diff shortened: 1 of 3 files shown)


terrelln:
BTW, I have a similar tool in `//experimental/terrelln:bit_checker`. It checks for any single byte change (not just bits), but is the same idea.

terrelln:
BTW, I have a similar tool in `//experimental/terrelln:bit_checker`. It checks for any single byte change (not just bits), but is the same idea.

terrelln:
Ignore the Appveyor failure, I'm fixing it.

PR #1898, opened 2019-11-19
[fuzz] Only set HUF_repeat_valid if loaded table has all non-zero weights

bimbashrestha:
Fixes a fuzz issue where dictionary_round_trip failed because the compressor was generating corrupt files thanks to zero weights in the table.

EDIT: Also updating the fse tables to do the same check


terrelln:
Awesome!

Can you add a test that we still use the dictionaries literals table for small inputs when it has a zero-weighted symbol that is unused, and doesn't when it has a zero-weighted symbol that is used?

Do we need a similar fix for the FSE tables?

lib/compress/zstd_compress.c
```diff
@@ -2853,14 +2853,23 @@ static size_t ZSTD_checkDictNCount(short* normalizedCounter, unsigned dictMaxSym
...
-        size_t const hufHeaderSize = HUF_readCTable((HUF_CElt*)bs->entropy.huf.CTable, &maxSymbolValue, dictPtr, dictEnd-dictPtr);
+        size_t hasNoZeroWeights;
+        size_t const hufHeaderSize = HUF_readCTable((HUF_CElt*)bs->entropy.huf.CTable, &maxSymbolValue, dictPtr,
+            dictEnd-dictPtr, &hasNoZeroWeights);
+
+        /* We only set the loaded table as valid if it contains all non-zero
+         * weights. Otherwise, we set it to check */
+        if (hasNoZeroWeights)
+            bs->entropy.huf.repeatMode = HUF_repeat_valid;
+        else bs->entropy.huf.repeatMode = HUF_repeat_check;
```
bimbashrestha: Maybe this should be HUF_repeat_none instead of HUF_repeat_check? Since the dictionary is 'corrupt' at this point right?

Cyan4973: It depends on if we want to forbid such cases (zero-weighted symbols in the dictionary header) or we want to authorize them.

At this point, the specification doesn't forbid it, and I believe it's correct.

We can imagine a use case where a dictionary is employed to compress ASCII files, in which case many byte values are simply impossible. It then becomes advantageous to zero-weight the impossible byte values, for better compression.

Of course, if the file to compress contains a few non-ASCII characters, the assumption doesn't work anymore, and the compressed payload is penalized by the requirement to define its own table. But since it's _not_ supposed to happen, it's okay to shift this cost onto bogus exceptions.

terrelln: I agree we should allow it.

bimbashrestha:
> Can you add a test that we still use the dictionaries literals table for small inputs when it has a zero-weighted symbol that is unused, and doesn't when it has a zero-weighted symbol that is used?

Sure. I'm not totally sure what you mean...can you elaborate a bit plz?:)

> Do we need a similar fix for the FSE tables?

Yes, good point. I'll add the check now

terrelln:
We should have unit tests two scenarios:
1. The input is small and we are using a dictionary. The dictionary has a zero weighted symbol (so it is set to `HUF_repeat_check`). The compression strategy is < `ZSTD_lazy` so we prefer repeat with small inputs. And the zero weighted symbol is unused. In this case the dictionary's Huffman table should be repeated.
2. The same as above, but the zero weighted symbol is used in the data, so the dictionary's Huffman table should not be used.

lib/compress/huf_compress.c
```diff
@@ -444,6 +444,7 @@ static int HUF_validateCTable(const HUF_CElt* CTable, const unsigned* count, uns
   int s;
   for (s = 0; s <= (int)maxSymbolValue; ++s) {
     bad |= (count[s] != 0) & (CTable[s].nbBits == 0);
+    bad |= (count[s] == 0) & (CTable[s].nbBits != 0);
```
bimbashrestha: This makes sense right? We don't want the nbBits to be non-zero when the weight is zero. This gets rid of the other dictionary_loader fuzz failure but I'm not sure we want to actually do this. @terrelln, @Cyan4973?

terrelln: This would make nearly every CTable invalid. We construct dictionary CTables to have non-zero probability for every symbol, so that they always work.

With this change, any source that has any zero probability symbol will call the dictionary's CTable invalid.

lib/compress/huf_compress.c
```diff
@@ -196,7 +196,7 @@ size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void
     {   U32 n; for (n=0; n<nbSymbols; n++) {
             const U32 w = huffWeight[n];
             *hasNoZeroWeights *= w;
-            CTable[n].nbBits = (BYTE)(tableLog + 1 - w);
+            CTable[n].nbBits = (w != 0) * (BYTE)(tableLog + 1 - w);
```
bimbashrestha: Adding this conditional will fail HUF_validateCTable and so we will set repeatMode to HUF_repeat_none and thus, ignore the loaded dictionary during compression. This seems to get rid of both dictionary fuzz issues.

bimbashrestha:
That didn't seem to fix the msan issue. I'm going to move the FSE changes to another PR and leave this one clean with just the HUF changes.

lib/compress/huf_compress.c
```diff
@@ -192,9 +192,11 @@ size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void
     }   }
 
     /* fill nbBits */
+    *hasNoZeroWeights = 1;
```
terrelln: I would suggest switching this to something like:

```c
    *hasNoZeroWeights = 0;
    {   U32 n; for (n=0; n<nbSymbols; n++) {
            const U32 w = huffWeight[n];
            *hasNoZeroWeights |= w;
            CTable[n].nbBits = (BYTE)(tableLog + 1 - w) & -(w != 0);
```

It achieves the same results without any multiplications.

We should also benchmark this function before/after this change. I would suggest writing a micro-benchmark that loads some CTable repeatedly. It doesn't need to be a perfect benchmark, but we should show that the difference is small.

lib/compress/huf_compress.c
```diff
@@ -169,7 +169,7 @@ size_t HUF_writeCTable (void* dst, size_t maxDstSize,
 }
 
 
-size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void* src, size_t srcSize)
+size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void* src, size_t srcSize, size_t* hasNoZeroWeights)
```
terrelln: Nit: `hasNoZeroWeights` should be an `int` or an `unsigned`. It is unusual to use a `size_t` as a boolean type.

terrelln: I would also recommend making the parameter `hasZeroWeights`, to avoid a negative. Generally, it is best to keep boolean variables as positive (`X` instead of `notX`).

bimbashrestha:
I'm getting ~1% regression on my desktop with the last commit. I presume that is small enough. Mircobenchmarked loading a 253 byte dict using:

```C
for (samples = 0; samples < 10000; ++samples) {
    double start = get_time();
    for (runsPerSample = 0; runsPerSample < 10000; ++runsPerSample) {
      HUF_readCTable((HUF_CElt *)hufTable.CTable, &maxSymbolValue, dict,
                     dictSize, &hasZeroWeights);
    }
    double delta = get_time() - start;
    minDelta = delta < minDelta ? delta : minDelta;
  }
  return minDelta;
```

terrelln:
Out of curiosity, does the multiplication vs bit manipulation make a difference?

bimbashrestha:
> Out of curiosity, does the multiplication vs bit manipulation make a difference?

Yeah, I was curious about that too. I didn't seem to make much difference. Both were around ~1%. The multiplication approach was actually slightly faster in my benchmark but it was less than 0.01% so its probably just noise.

Cyan4973:
`HUF_readCTable()` is not too performance critical, since in most cases, it's supposed to be loaded offline. Hence, a 1% performance regression on this operation only is not an issue.

lib/common/huf.h
```diff
@@ -247,7 +247,7 @@ size_t HUF_readStats(BYTE* huffWeight, size_t hwSize,
 
 /** HUF_readCTable() :
  *  Loading a CTable saved with HUF_writeCTable() */
-size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void* src, size_t srcSize);
+size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void* src, size_t srcSize, unsigned *hasZeroWeights);
```
Cyan4973: minor point : I'm not too thrilled that the function signature had to be changed (and complexified) to support this new scenario, even though it's a valid solution.
I suppose an alternative could have been to scan the resulting `CTable` for potential presence of 0-freq slots.
It may be suspected that this alternative operation may cost more runtime, just it wasn't measured.

PR #1905, opened 2019-11-25
fix #1904

### LINKED ISSUE #1904: Zstd modifies the ownership of /dev/null
```
sudo zstd input -o /dev/null
```

`/dev/null` is now owned by root. I know we've fixed this before, but it looks like we've regressed.

Cyan4973:
fix #1904

`/dev/null` permissions were modified when targeted as destination file using `sudo` rights.
Bug fixed for both compression and decompression.

More importantly, this patch adds a test, [triggered in TravisCI](https://travis-ci.org/facebook/zstd/jobs/616825323#L676),
ensuring unaltered `/dev/null` permissions when using `sudo` rights.

minor : improves speed of `make check`, by removing some inefficiencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ matrix:
 
     - name: make test (complete)
       script:
-        - make test
+        # DEVNULLRIGHTS : will request sudo rights to test permissions on /dev/null
+        - DEVNULLRIGHTS=test make test
 
     - name: gcc-6 + gcc-7 compilation
       script:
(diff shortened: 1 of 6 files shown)


terrelln:
Could we add a `UTIL_chmod(char const* filename, mode_t permissions)` that refuses to change permissions on `/dev/null`, and use that everywhere? That way the edge case logic is centralized.

We would also add that same logic to `UTIL_setFileStat()`. So we probably want a helper function `UTIL_shouldModifyFileMetadata(char const* filename) { return strcmp(filename, devnul); }`

Cyan4973:
`UTIL_chmod()` created

PR #1906, opened 2019-11-26
[bench] Automated benchmarking script

bimbashrestha:
Currently have this running on my desktop. I have this set to give me email alerts if the there is speed regression in compression or decompression greater than 1% on silesia. Maybe that is too sensitive? We'll see. I'll adjust it if I get too many alerts.

Changes to test-zstd-speed.py
* More than 2 iterations for greater stability
* Simpler api (ie. compare_versions(old, new, filenames))


tests/automated-benchmarking.py
```diff
@@ -0,0 +1,189 @@
...
+LEVELS = [1, 2, 3, 4, 5]
+BUILDS = [
+    {"compiler": "clang", "mode": "64"},
+    {"compiler": "gcc", "mode": "64"}
+]
+
+# Not sure what the threshold for triggering alarms should be
+# 1% regression sounds like a little too sensitive but the desktop
+# that I'm running it on is pretty stable so I think this is fine
+CSPEED_REGRESSION_TOLERANCE = 0.01
```
Cyan4973: Try to make it possible to modify these values on command line.
You can keep constants in the source code as "default" ones, it's a good practice.

This makes the script a lot more flexible, eventually usable on different platforms than current target one.
Generally speaking, consider that any script in `/tests` can be used by any user or contributor of `zstd`.
This elusive "user" will eventually happen to be you in a few months or years, and you'll be glad that the script is flexible enough and do not make too many hard-coded assumptions.

Cyan4973:
Nice !

2 points :
- Can you describe the differences between this new script and [`test-zstd-speed.py`](https://github.com/facebook/zstd/blob/dev/tests/test-zstd-speed.py) ?
- Could you add a help file or paragraph explaining the purpose of and how to use this script, typically in the [`README.md`](https://github.com/facebook/zstd/blob/dev/tests/README.md) ?

Cyan4973:
Also : could we consider running it in CI ?
Not as a regression test, since speed conditions are uncontrollable, but just as a way to test that the script is still working as intended, in case something unexpected would break it in some future ?

tests/automated-benchmarking.py
bimbashrestha: Yep!

bimbashrestha:
> * Can you describe the differences between this new script and [`test-zstd-speed.py`](https://github.com/facebook/zstd/blob/dev/tests/test-zstd-speed.py) ?
* More stable: test-zstd-speed.py just benchmarked twice using its double_check(). Here, we can specify the number of iterations and this gave more stable results on my desktop.
* This api is easier to use: The functions in test-zstd-speed.py were a little convoluted (too many unnecessary arguments to pass in and not super modular). I think this will be easier to build a more robust automated benchmarking framework on. The core function here is super simple and easy to use.

def compare_versions(old_hash, new_hash, filenames_list)
 
> * Could you add a help file or paragraph explaining the purpose of and how to use this script, typically in the [`README.md`](https://github.com/facebook/zstd/blob/dev/tests/README.md) ?
Ah yes, good idea. Sounds good.

bimbashrestha:
> Also : could we consider running it in CI ?

Do you mean run automated_benchmarking.py in the cli (yes, this should be possible. we can just download the silesia corporus live and just run compare_versions(...) with the previous two git hashes)

or do you write tests to make sure the benchmaking code itself automated_benchmarking.py is correct?

Cyan4973:
> this should be possible. we can just download the silesia corporus live and just run compare_versions(...)

We should target something simpler. If possible, use existing resources from the repository, this would avoid dependencies and delays from a 3rd party source.

bimbashrestha:
> > this should be possible. we can just download the silesia corporus live and just run compare_versions(...)
> 
> We should target something simpler. If possible, use existing resources from the repository, this would avoid dependencies and delays from a 3rd party source.

Ah okay. I can start with just compression the dir.

tests/DEPRECATED-test-zstd-speed.py
```diff
@@ -1,4 +1,5 @@
 #! /usr/bin/env python3
+# THIS BENCHMARK IS BEING REPLACED BY automated-bencmarking.py
```
Cyan4973: nit: minor spelling error (missing `h`)

Cyan4973:
Still waiting for a presentation of the tool in `README.md`,
and if possible some quick test to check the script is running fine
(can be part of the `master`/`cron` tests on travisCI).

bimbashrestha:
> Still waiting for a presentation of the tool in `README.md`,
> and if possible some quick test to check the script is running fine
> (can be part of the `master`/`cron` tests on travisCI).

Yes. I'm need to update this benchmarking to compare pr and not just the latest hashes like it currently is. I'll update the README and include the ci test after I do that.

bimbashrestha:
Just included automated_benchmarking.py as part of make test. Not sure if we want to do this or not?

bimbashrestha:
> Just included automated_benchmarking.py as part of make test. Not sure if we want to do this or not?

Talked offline. We don't want to include it in make test because make test needs to be able to run everywhere and automated_benchmarking won't

tests/README.md
```diff
@@ -20,9 +20,39 @@ This script creates `versionsTest` directory to which zstd repository is cloned.
...
+```
+usage: automated_benchmarking.py [-h] directory levels mode emails
+
+positional arguments:
+  directory   directory with files to benchmark
+  levels      levels to test eg ('1,2,3')
+  mode        0: regular, 1: pull request ci, 2: continuous
+  iterations  number of benchmark iterations to run
+  emails      email addresses of people who will be alerted upon regression.
+              Only for mode 2
```
Cyan4973: Only for mode 2 ? or do you mean mode 3 ?

tests/README.md
```diff
@@ -20,9 +20,39 @@ This script creates `versionsTest` directory to which zstd repository is cloned.
...
+Example usage: python automated_benchmarking.py fuzz "1,2,3" "0" "1" ""
+```
+
+```
+usage: automated_benchmarking.py [-h] directory levels mode emails
+
+positional arguments:
+  directory   directory with files to benchmark
+  levels      levels to test eg ('1,2,3')
+  mode        0: regular, 1: pull request ci, 2: continuous
```
Cyan4973: `mode` : Is there a way to use _names_ instead of numbers here ?
That would make the associated command line much more readable, hence more maintainable.

tests/automated_benchmarking.py
```diff
@@ -0,0 +1,220 @@
...
+import glob
+import json
+import os
+import pickle as pk
+import subprocess
+import urllib.request
+
+
+GITHUB_API_PR_URL = "https://api.github.com/repos/facebook/zstd/pulls?state=open"
+GIRHUB_URL_TEMPLATE = "https://github.com/{}/zstd"
```
Cyan4973: `GIRHUB` or `GITHUB` ?

tests/README.md
```diff
@@ -20,9 +20,39 @@ This script creates `versionsTest` directory to which zstd repository is cloned.
...
+#### `automated-benchmarking.py` - script for benchmarking zstd prs to dev
+
+This script benchmarks facebook:dev and changes from pull requests made to zstd and compares
+them against facebook:dev to detect regressions. This script currently runs on a dedicated
+desktop machine for every pull request that is made to the zstd repo but can also
+be run on any machine via the command line interface.
+
+There are three modes of usage for this script: mode 0 will fetch all open pull requests
+in the repo and benchmark them against dev, mode 1 will fetch the pull request
+associated with the current hash and benchmark that against dev, and mode 3 will
```
Cyan4973: mode 0, mode 1, mode 3, 
but no mode 2 ?

tests/README.md
```diff
@@ -20,9 +20,39 @@ This script creates `versionsTest` directory to which zstd repository is cloned.
...
 
+#### `automated-benchmarking.py` - script for benchmarking zstd prs to dev
+
+This script benchmarks facebook:dev and changes from pull requests made to zstd and compares
+them against facebook:dev to detect regressions. This script currently runs on a dedicated
+desktop machine for every pull request that is made to the zstd repo but can also
+be run on any machine via the command line interface.
+
+There are three modes of usage for this script: mode 0 will fetch all open pull requests
+in the repo and benchmark them against dev, mode 1 will fetch the pull request
```
Cyan4973: > mode 1 will fetch the pull request associated with the current hash and benchmark that against dev

What the scenario for this mode ?

Could it measure speed of the current hash against `dev` if it's different, irrespective of being associated to a PR or not ?

tests/README.md
bimbashrestha: I mean mode 3. Will update

tests/README.md
bimbashrestha: Took me a second to remember what these modes were after not looking at this code for a week so yes, good call:) I'll give them labels.

tests/automated_benchmarking.py
bimbashrestha: Ah its supposed to be used in line 68 but it looks like I forgot. And yes, GITHUB:)

tests/README.md
bimbashrestha: This will be clearer with labels

tests/README.md
bimbashrestha: This scenario was for being run in the CI tests. It finds the pr associated with the hash and then runs the benchmarking. The actual metrics that it gets might not be super meaningful since we don't know can't control the stability of the CI host machines but it might still be nice to have the output printed. 

If the hash isn't associated with a PR, it will crash. It would be trivial to have it compare the hash to facebook:dev regardless of association with a PR or not so I'll go ahead and add that feature in the update.

tests/README.md
```diff
@@ -20,9 +20,38 @@ This script creates `versionsTest` directory to which zstd repository is cloned.
...
 In the following step interoperability between zstd versions is checked.
 
+#### `automated-benchmarking.py` - script for benchmarking zstd prs to dev
+
+This script benchmarks facebook:dev and changes from pull requests made to zstd and compares
+them against facebook:dev to detect regressions. This script currently runs on a dedicated
+desktop machine for every pull request that is made to the zstd repo but can also
+be run on any machine via the command line interface.
+
+There are three modes of usage for this script: fastmone will just run a minimal single
```
Cyan4973: nit : fastmone -> fastmode

tests/README.md
Cyan4973: Should the doc be updated ? It seems numbers are no longer useful for `mode`.

tests/README.md
Cyan4973: same comment on doc update

bimbashrestha:
Just tested this on my forked repo with this PR (https://github.com/bimbashrestha/zstd/pull/25) where I compress twice in benchzstd and it looks like emails are working. Got the following regression alert just now: 

```
[COMPRESSION REGRESSION] (level=1 filename=golden-compression/huffman-compressed-larger)
        facebook:dev (None) -> bimbashrestha:bench-test (7d896d8a1a8597d267efd854a35df7fc38cac6e0)
        25.63 -> 12.62 (50.76%)
```

tests/automated_benchmarking.py
```diff
@@ -10,6 +11,9 @@
 GITHUB_API_PR_URL = "https://api.github.com/repos/facebook/zstd/pulls?state=open"
 GITHUB_URL_TEMPLATE = "https://github.com/{}/zstd"
 MASTER_BUILD = {"user": "facebook", "branch": "dev", "hash": None}
+
+# check to see if there are any new PRs every minute
+MAX_API_CALL_FREQUENCY_SEC = 60
```
Cyan4973: 👍 

minor : ideally, this should be a parameter selectable on command line.
Starting with a constant is still fine, it can later be converted into a nice default.

Cyan4973:
Provided remaining tests pass, this looks good to me !

Issue #1895, opened 2019-11-19
The return value of 'fread()' should be checked


QiAnXinCodeSafe:
https://github.com/facebook/zstd/blob/6a7f65117e8f270ff30f43855ef0efe63fb4b31f/programs/fileio.c#L1777

terrelln:
We do check the output of `fread()` (it is in a while loop). But you're right, we should have a `ferror(finput)` after the loop.

Cyan4973:
fixed by @selavy

PR #1907, opened 2019-11-27
Check for fread failure

selavy:
On failure fread may return either a short read or 0. Need to use
ferror to detect error versus eof.

For https://github.com/facebook/zstd/issues/1895


Cyan4973:
Looks good to me ! Thanks @selavy !

PR #1908, opened 2019-11-27
[fuzz] msan uninitialized unsigned value

bimbashrestha:
Fixes new fuzz issue

Credit to OSS-Fuzz

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2861,6 +2861,7 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
 
     {   unsigned maxSymbolValue = 255;
         unsigned hasZeroWeights = 1;
+        bs->entropy.huf.repeatMode = HUF_repeat_check;
         size_t const hufHeaderSize = HUF_readCTable((HUF_CElt*)bs->entropy.huf.CTable, &maxSymbolValue, dictPtr,
             dictEnd-dictPtr, &hasZeroWeights);
 
@@ -2868,7 +2869,6 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
          * weights. Otherwise, we set it to check */
         if (!hasZeroWeights)
             bs->entropy.huf.repeatMode = HUF_repeat_valid;
-        else bs->entropy.huf.repeatMode = HUF_repeat_check;
 
         RETURN_ERROR_IF(HUF_isError(hufHeaderSize), dictionary_corrupted);
         RETURN_ERROR_IF(maxSymbolValue < 255, dictionary_corrupted);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2857,11 +2857,11 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
 {
     const BYTE* dictPtr = (const BYTE*)dict;    /* skip magic num and dict ID */
     const BYTE* const dictEnd = dictPtr + dictSize;
-    unsigned hasZeroWeights = 1;
     dictPtr += 8;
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -2860,7 +2860,7 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
     dictPtr += 8;
 
     {   unsigned maxSymbolValue = 255;
-        unsigned hasZeroWeights;
+        unsigned hasZeroWeights = 0;
```
Cyan4973: Yes, this should work.
But just for the record, there are a number of small design choices that could be selected here in order to produce a more defensive pattern.

- One could initialize `hasZeroWeights` to `1` rather than `0`, since it's more conservative.
- One could `RETURN_ERROR_IF()` before doing anything with `hasZeroWeights`
- One could initialize `bs->entropy.huf.repeatMode` to `HUF_repeat_check` first, unconditionally, before calling `HUF_readCTable()`, and only set it to `HUF_repeat_valid` after it gets confirmed.

These tiny differences are unlikely to make a visible impact _now_, but it could make a difference _tomorrow_, when _someone else_ change the code and break an assumption. In which case, the other mechanisms kick in, and fall back on defensive positions, rather than assume all is right an move strainght into UB territory.

bimbashrestha: I see. Noted for the future.

> One could initialize bs->entropy.huf.repeatMode to HUF_repeat_check first, unconditionally, before calling HUF_readCTable(), and only set it to HUF_repeat_valid after it gets confirmed.

This seems the most clean to me. I'll go ahead and make the changes in this PR now.

bimbashrestha: > One could RETURN_ERROR_IF() before doing anything with hasZeroWeights

Not sure I quite understand the reasoning behind this one. Can you elaborate a little?

lib/compress/zstd_compress.c
```diff
@@ -2860,15 +2860,15 @@ size_t ZSTD_loadCEntropy(ZSTD_compressedBlockState_t* bs, void* workspace,
     dictPtr += 8;
 
     {   unsigned maxSymbolValue = 255;
-        unsigned hasZeroWeights;
+        unsigned hasZeroWeights = 1;
+        bs->entropy.huf.repeatMode = HUF_repeat_check;
```
Cyan4973: This will cause issue with `C90` compatibility, due to declaration order.
Set the value before the block.

bimbashrestha: Hmm I would not have guessed that. Moved now

bimbashrestha: Moved the wrong line:( just fixed

lib/compress/zstd_compress.c
Cyan4973: Sure. 
`hasZeroWeights` used to have no value before being set inside `HUF_readCTable()`.
The `msan` warning was triggered, stating that there are cases where the value in _not_ set.
When does it happen ?
Essentially, when there is an error.
In which case, `RETURN_ERROR_IF()` should get triggerred.

Putting `RETURN_ERROR_IF()` first essentially means the function will return early, thus avoiding to read `hasZeroWeights`, which might not be set.

This seems no longer important if `hasZeroWeights` is set beforehand, true, but this is part of a defensive pattern. 
Maybe, in some future, someone else will make a mistake, and introduce a bug, which either initialize `hasZeroWeights` to some random garbage value, or which set `hasZeroWeights` to some incorrect (too optimistic) value before exiting from `HUF_readCTable()` with an error. We would prefer not to act on such signals.

bimbashrestha:
I think this is good to go now:)

terrelln:
Please add "Credit to OSS-Fuzz" to the summary. Also, please comment on the oss-fuzz issue (in their tracker) and say that this is a short-lived bug that never made it into a release.

PR #1909, opened 2019-11-27
Fix mingw cross

Cyan4973:
Added on top of the `mingw` cross-compilation test on travisCI


Cyan4973:
cancelling. this PR was only to observe travis CI tests. I will redirected #1883 to `travisTest` instead.

PR #1920, opened 2019-12-04
Use statbuf->st_mtim Again

felixhandte:
Let's try this again. Changes:

1. I believe checking `defined(_WIN32)` is redundant with checking the POSIX version. So I've removed that.
2. I flipped the order, we should do the preferred behavior first and the fallback next.
3. I added a check for whether `st_mtime` is defined. When `st_mtim` replaced `st_mtime`, `st_mtime` became a macro that expanded to `st_mtim.tv_sec` for backwards compatibility. Since we have found examples where the platform advertising POSIX 2008 still doesn't have `st_mtim`, this additional check should be defensive. The only concern is whether there are platforms that have `st_mtim` but not the `st_mtime` macro. I don't believe that there should be: the `st_mtime` macro is also part of POSIX 2008 (although we've seen where that gets us...). At any rate, failing this check will just cause us to fall back to the less preferable behavior.

Reference: [POSIX 2008 `sys/stat.h` Spec](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/basedefs/sys_stat.h.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -137,6 +137,13 @@ matrix:
         - make arminstall
         - make armfuzz
 
+    # Introduced to check compat with old toolchains, to prevent e.g. #1872
+    - name: ARM Build Test (on Trusty)
+      dist: trusty
+      script:
+        - make arminstall
+        - make armbuild
+
     - name: Qemu PPC + Fuzz Test    # ~13mn
       dist: trusty   # it seems ppc cross-compilation fails on "current"
       script:
--- a/programs/platform.h
+++ b/programs/platform.h
@@ -109,6 +109,15 @@ extern "C" {
 #endif   /* PLATFORM_POSIX_VERSION */
 
 
+#if PLATFORM_POSIX_VERSION > 1
+   /* glibc < 2.26 may not expose struct timespec def without this.
+    * See issue #1920. */
+#  ifndef _ATFILE_SOURCE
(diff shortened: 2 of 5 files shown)


Cyan4973:
Thanks @felixhandte , this is indeed much cleaner.

sergey-dryabzhinsky:
https://github.com/facebook/zstd/pull/1897#issuecomment-561900415
https://github.com/facebook/zstd/pull/1897#issuecomment-561914281

sergey-dryabzhinsky:
Debian Wheezy builds fine: gcc 4.7 + eglibc 2.13

sergey-dryabzhinsky:
https://stackoverflow.com/a/42606673 - this definition fixed issue with `timebuf`

felixhandte:
So, this looks like the result of a [bug](https://sourceware.org/bugzilla/show_bug.cgi?id=21371) that got [fixed in glibc 2.26](https://sourceware.org/git/gitweb.cgi?p=glibc.git;a=commitdiff;h=d76d3703551a362b472c866b5b6089f66f8daa8e), where the definition of `struct timespec` could be hidden even when the `struct timespec st_{a,m,c}tim` fields are present in `struct stat`.

IIUC, we can fix this if we additionally define `_ATFILE_SOURCE`. Does that seem reasonable?

PR #1921, opened 2019-12-04
Only Modify Mode on Regular Files

### LINKED ISSUE #1904: Zstd modifies the ownership of /dev/null

felixhandte:
Rather than do string comparisons to exclude `/dev/null` specifically, it makes more sense to me to avoid all special files (i.e.: devices, FIFOs, symlinks, sockets, and directories). This change does that, although it finds that in most places that chmod / chown are called from, there are already `UTIL_isRegularFile()` guards.

So I guess this is my proposal to amend #1905 to fix #1904 better.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -509,7 +509,7 @@ static int FIO_remove(const char* path)
 #if defined(_WIN32) || defined(WIN32)
     /* windows doesn't allow remove read-only files,
      * so try to make it writable first */
-    UTIL_chmod(path, _S_IWRITE);
+    UTIL_chmod(path, _S_IWRITE, 0 /* onlyRegularFiles: already checked above */);
 #endif
     return remove(path);
 }
@@ -618,7 +618,7 @@ FIO_openDstFile(FIO_prefs_t* const prefs,
                && strcmp (srcFileName, stdinmark)
                && strcmp(dstFileName, nulmark) ) {
             /* reduce rights on newly created dst file while compression is ongoing */
-            UTIL_chmod(dstFileName, 00600);
+            UTIL_chmod(dstFileName, 00600, 1 /* onlyRegularFiles */);
         }
         return f;
     }
--- a/programs/util.c
+++ b/programs/util.c
@@ -138,17 +138,25 @@ int UTIL_chmod(char const* filename, mode_t permissions, int onlyRegularFiles)
     return chmod(filename, permissions);
 }
 
-int UTIL_chown(char const* filename, uid_t owner, gid_t group, int onlyRegularFiles) {
+#if !defined(_WIN32)
+static int UTIL_chown(char const* filename,
+                      uid_t owner,
+                      gid_t group,
+                      int onlyRegularFiles) {
     if (onlyRegularFiles && !UTIL_isRegularFile(filename)) {
         return 0;
     }
-#if !defined(_WIN32)
     return chown(filename, owner, group);
-#else
-    /* windows doesn't have the chown concept. */
-    return 0;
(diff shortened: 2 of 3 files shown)


Cyan4973:
Yes, it makes sense.

The situation has been made a bit more complex since support for named pipes has been added,
but for the purpose of setting permissions, it does not matter : named pipes should not be modified by `chmod()` nor `chown()`, no more than `stdout`.

So there is room for simplification.

programs/util.h
```diff
@@ -118,7 +120,8 @@ U64 UTIL_getFileSize(const char* infilename);
 U64 UTIL_getTotalFileSize(const char* const * fileNamesTable, unsigned nbFiles);
 int UTIL_getFileStat(const char* infilename, stat_t* statbuf);
 int UTIL_setFileStat(const char* filename, stat_t* statbuf);
-int UTIL_chmod(char const* filename, mode_t permissions);   /*< like chmod, but avoid changing permission of /dev/null */
+int UTIL_chmod(char const* filename, mode_t permissions, int onlyRegularFiles);   /*< like chmod, but avoid changing permission of special files */
```
Cyan4973: Suggestion : if we agree that `chmod()` should only be applied to regular files, then it's not even necessary to provide a parameter `onlyRegularFiles`. It's implied as always `1`.

Same thing for `chown()` below.

programs/fileio.c
```diff
@@ -611,7 +611,7 @@ FIO_openDstFile(FIO_prefs_t* const prefs,
                && strcmp (srcFileName, stdinmark)
                && strcmp(dstFileName, nulmark) ) {
```
Cyan4973: If `UTIL_chmod()` guarantees that it will only applies permissions to regular files, then there is no more need to check if `dstFileName` == `nulmark`, simplifying this test.

note : conditions on `srcFileName` must still be checked !

programs/util.h
```diff
@@ -118,7 +120,8 @@ U64 UTIL_getFileSize(const char* infilename);
...
+int UTIL_chown(char const* filename, uid_t owner, gid_t group, int onlyRegularFiles);   /*< like chown, but avoid changing permission of special files */
```
Cyan4973: note : if `UTIL_chown()` is only used inside `util.c`, maybe there is no need to publish its interface ?

programs/util.c
```diff
@@ -130,13 +130,26 @@ int UTIL_getFileStat(const char* infilename, stat_t *statbuf)
...
 {
-    if (!strcmp(filename, "/dev/null")) return 0;   /* pretend success, but don't change anything */
+    if (onlyRegularFiles && !UTIL_isRegularFile(filename)) {
+        return 0;
+    }
     return chmod(filename, permissions);
 }
 
+int UTIL_chown(char const* filename, uid_t owner, gid_t group, int onlyRegularFiles) {
+    if (onlyRegularFiles && !UTIL_isRegularFile(filename)) {
```
Cyan4973: If the only known usage of `UTIL_chown()` sets `onlyRegularFiles == 0`, then this test could be replaced by an `assert()`.

Cyan4973:
still in need of fixes

terrelln:
Ping @felixhandte

Cyan4973:
It seems there were a number of opened issues on this PR.
Is it still valid, or should we close it ?

felixhandte:
Going to replace this with a better version soon.

PR #1923, opened 2019-12-05
Fix playTests.sh to Work under QEMU

felixhandte:
`playTests.sh` needs to distinguish the `qemu` invocation from the path to the `zstd` binary, so it can construct symlinks and invoke `zstd` through those symlinks.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -330,7 +330,7 @@ test-zstd-nolegacy: zstd-nolegacy
 
 test-zstd test-zstd32 test-zstd-nolegacy: datagen
 	file $(ZSTD)
-	ZSTD="$(QEMU_SYS) $(ZSTD)" ./playTests.sh $(ZSTDRTTEST)
+	EXE_PREFIX="$(QEMU_SYS)" ZSTD_BIN="$(ZSTD)" ./playTests.sh $(ZSTDRTTEST)
 
 
 test-gzstd: gzstd
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -102,9 +102,11 @@ case "$UNAME" in
   SunOS) DIFF="gdiff" ;;
 esac
 
-println "\nStarting playTests.sh isWindows=$isWindows ZSTD='$ZSTD'"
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -137,6 +137,12 @@ matrix:
         - make arminstall
         - make armfuzz
 
+    - name: ARM Build Test (on Trusty)
+      dist: trusty
+      script:
+        - make arminstall
+        - make armbuild
```
felixhandte: The specific issue that motivates this change is a build issue with an old toolchain, not a runtime problem. So in the interests of saving time, I've only built zstd, not run any tests. Does that seem reasonable? Should we do `armtest` instead?

Cyan4973: `armbuild` is fine, it answers the issue.

Maybe a comment reminding why this test is here would be useful.
Without context, it's easy to wonder why this test is not simply folded into the one just above.

Cyan4973:
The new test just checks the compilation outcome (`armbuild`), which is the intended objective.

According to report in #1897, build is supposed to fail after #1920 merge, and the test is basically there to guard against such risk. Is that the case ?

felixhandte:
So, no, it failed to fail: https://travis-ci.org/felixhandte/zstd/jobs/621714178.

:(

Cyan4973:
Maybe it needs some more specific setup ?
Such as being built directly on `arm` natively, rather than cross-compiled from `x64` ?
It might influence the exact library being linked, which may lead to a different outcome.

TravisCI recently added the ability to run on `arm64` platform, and we do have a test setup here : 
https://github.com/facebook/zstd/blob/dev/.travis.yml#L30

But it's unclear if this capability can be combined with `dist: trusty` (not attempted yet).

Cyan4973:
What do we do on this topic ?
It seems the test proposed doesn't catch the intended issue.
Should we drop this PR ?

felixhandte:
I still couldn't repro with Trusty on ARM64. I guess we should just drop the new travis build. The first commit is still useful though, fixing `playTests.sh`.

Issue #1922, opened 2019-12-05
Make zstd cmake use idomatically correct cmake install


ned14:
zstd currently uses idiomatically incorrect cmake install which breaks direct usage within other cmake which expects install to work as cmake says it ought to. I have a patchset fixing zstd's cmake at:

https://github.com/ned14/zstd/commit/ff6021046353cd91cbb1963ac943256029f0927c#commitcomment-36287624

... however I am unsure how compatible my changes are with cmake 2.8.

Are you open to accepting a PR fixing these issues, or would potential cmake 2.8 compatibility be a blocker?

ned14:
For reference, the link above is to https://github.com/cpp-pm/hunter, a 100% cmake-based package manager. It only accepts packages with correct cmake install.

bimbashrestha:
Thanks for submitting @ned14! I believe it would be good to retain cmake 2.8 compatibility. I'm not too familiar with the topic. While a fix for the issues you mentioned would certainly be very nice, we should first scope out what specifically the compatibility concerns on 2.8 would be with the patch you have in mind.

Cyan4973:
> Are you open to accepting a PR fixing these issues, or would potential cmake 2.8 compatibility be a blocker?

The primary reason for the existence of such a limit is to remain compatible with CI environments. Since this limit was established, CI environments have evolved, and now generally support newer versions.

However, we can't just go directly to latest version, because the script is meant to be broadly usable for anyone trying to compile or install `zstd`. And there are various systems out there, too high version requirement would limit the nb of systems able to use the `cmake` script. 

So there must be a limit. But which one ?
As a generic rule, we have to be conservative.

Hence here is what I would suggest : 

- If there is a need for newer capabilities, which can't be used with current limit, ___define___ a new limit. 

   + The new limit should be defined as tight as possible, aka, the minimum version absolutely required to be compatible with the new `cmake` script. Do not over-provision by setting a large encompassing limit, this will reduce the nb of compatible systems.

    + The new limit must remain compatible with CI tests. If I've not missed something, it has been increased to <= 3.9.2. And if I'm wrong, a test is likely to fail somewhere.

ned14:
@bimbashrestha My initial concern about cmake 2.8 compatibility was the use of `install(TARGETS ... EXPORT ...)` which I had thought a cmake 3-ism. But according to https://cmake.org/cmake/help/v2.8.12/cmake.html, it might just be supported, though how bug free I cannot tell.

[...]

PR #1924, opened 2019-12-06
Make idiomatic cmake install so third party cmake installers work right.

ned14:
Please see issue #1922

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -127,6 +127,7 @@ if (ZSTD_BUILD_STATIC)
     set_target_properties(
             libzstd_static
             PROPERTIES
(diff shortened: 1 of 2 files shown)


build/cmake/lib/CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ if (ZSTD_BUILD_STATIC)
     set_target_properties(
             libzstd_static
             PROPERTIES
+            POSITION_INDEPENDENT_CODE On
```
ned14: You may wish a separate target for this. It's a tricky one, it's not uncommon that static libraries get linked into dynamic libraries, however sometimes fPIC does slightly damage performance.

Cyan4973:
Thanks @ned14 ! Looks good to me !

PR #1928, opened 2019-12-13
[fuzz] Adding bool to check if there is enough room to emit noCompress Superblocks

bimbashrestha:
noCompress superblocks are larger than the original data because of the addition of a block header and epilogue per block. If we add enough of these during streaming compression, we're going to run out of room and throw a dstSize_tooSmall error. To prevent that I just branch to regular compression when we might run out of room. 

Credit to oss-fuzz

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2562,6 +2562,14 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
     BYTE* const ostart = (BYTE*)dst;
     BYTE* op = ostart;
     U32 const maxDist = (U32)1 << cctx->appliedParams.cParams.windowLog;
+
+    /* This bool is set if there is enough room to output all noCompress superblocks.
+     * Just checks if the number of compressed blocks we can fit in dstCapacity is
+     * greater than the optimistic number of blocks we still have remaining.
+     * This might be UNset when data is uncompressable and we're streaming.  */
+
+    int enoughDstCapacityForNoCompressSuperBlocks =
+        (dstCapacity / (blockSize + 7 /* header + checksum */)) > (srcSize / blockSize);
     assert(cctx->appliedParams.cParams.windowLog <= ZSTD_WINDOWLOG_MAX);
 
     DEBUGLOG(5, "ZSTD_compress_frameChunk (blockSize=%u)", (unsigned)blockSize);
@@ -2586,7 +2594,7 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
 
         {   size_t cSize;
             int useTargetCBlockSize = ZSTD_useTargetCBlockSize(&cctx->appliedParams);
-            if (useTargetCBlockSize) {
+            if (useTargetCBlockSize && enoughDstCapacityForNoCompressSuperBlocks) {
                 cSize = ZSTD_compressBlock_targetCBlockSize(cctx, op, dstCapacity, ip, blockSize, lastBlock);
                 FORWARD_IF_ERROR(cSize);
             } else {
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2568,7 +2568,7 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
      * greater than the optimistic number of blocks we still have remaining.
      * This might be UNset when data is uncompressable and we're streaming.  */
 
-    int enoughDstCapacityForNoCompressSuperBlocks =
+    const int enoughDstCapacityForNoCompressSuperBlocks =
         (dstCapacity / (blockSize + 7 /* header + checksum */)) > (srcSize / blockSize);
     assert(cctx->appliedParams.cParams.windowLog <= ZSTD_WINDOWLOG_MAX);
 
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -489,6 +489,53 @@ static int basicUnitTests(U32 const seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3d: superblock uncompressable data, too many nocompress superblocks : ", testNb++)
+    {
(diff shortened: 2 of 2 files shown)


bimbashrestha:
I managed to break my clang:( Going to run fuzzers on this patch for an hour locally after I fix it.

lib/compress/zstd_compress.c
```diff
@@ -2562,6 +2562,14 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
     BYTE* const ostart = (BYTE*)dst;
     BYTE* op = ostart;
     U32 const maxDist = (U32)1 << cctx->appliedParams.cParams.windowLog;
+
+    /* This bool is set if there is enough room to output all noCompress superblocks.
+     * Just checks if the number of compressed blocks we can fit in dstCapacity is
+     * greater than the optimistic number of blocks we still have remaining.
+     * This might be UNset when data is uncompressable and we're streaming.  */
+
+    int enoughDstCapacityForNoCompressSuperBlocks =
```
Cyan4973: I guess it can be `const` ?

tests/fuzzer.c
```diff
@@ -489,6 +489,53 @@ static int basicUnitTests(U32 const seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3d: superblock uncompressable data, too many nocompress superblocks : ", testNb++)
```
Cyan4973: 👍

tests/fuzzer.c
```diff
@@ -489,6 +489,53 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+    {
+        ZSTD_CCtx* cctx = ZSTD_createCCtx();
```
Cyan4973: nit : `const`ify : `ZSTD_CCtx* const cctx`
Also : check that `cctx` allocation is successful (`!= NULL`).

tests/fuzzer.c
```diff
@@ -489,6 +489,53 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        BYTE* src = (BYTE*)CNBuffer; BYTE* dst = (BYTE*)compressedBuffer;
```
Cyan4973: nit: `const`ify :
`const BYTE* const src`
`BYTE* const dst`
`size_t const srcSize`
`size_t const dstCapacity`

tests/fuzzer.c
```diff
@@ -489,6 +489,53 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        size_t srcSize = 321656; size_t dstCapacity = ZSTD_compressBound(srcSize);
+
+        /* This is the number of bytes to stream before ending. This value
+         * was obtained by trial and error :/. */
+
+        const size_t streamCompressThreshold = 161792;
+        const size_t streamCompressDelta = 1024;
+
+        /* The first 1/3 of the buffer is compressable and the last 2/3 is
+         * uncompressable. This is an approximation of the type of data
```
Cyan4973: nit: compressable ? or compressible ?

Cyan4973:
Looks good !
Just nagging you about some left-over `const` property (and one missing `malloc` check).
I believe it's time to apply this policy more consistently.

tests/fuzzer.c
bimbashrestha: Wasn't sure haha. I'll change it to compressible to be consistent with the rest of the file.

tests/fuzzer.c
bimbashrestha: dst, srcSize and dstCapacity change below but I'll const src.

bimbashrestha:
Just checked the test case for the new oss-fuzz failure that came in today. The problem there is also nocompress superblocks eating too much dstcapacity so this patch should take care of that too

tests/fuzzer.c
Cyan4973: That's a good example actually !
I missed that these variables are modified inside the inner loop.

Once the ` const` property is consistently applied, the _absence_ of `const` becomes a useful signal : it means "these variables are really mutable, pay attention to them", and helps focus the reviewers/readers/future contributor's attention.

Cyan4973:
All is good !
I presume you checked the test fails when the new boolean variable is not present.

bimbashrestha:
> All is good !
> I presume you also checked that the new test fails when the new boolean variable is not present (or always true).

Yep

bimbashrestha:
This is good to be merged but I plan on running fuzzers locally for a little while today to make sure this patch didn't quickly uncover any new code paths

bimbashrestha:
Ran ./fuzz.py libfuzzer stream_round_trip for an hour this morning. It didn't catch anything so I think this is good now.

lib/compress/zstd_compress.c
```diff
@@ -2562,6 +2562,14 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
...
     BYTE* op = ostart;
     U32 const maxDist = (U32)1 << cctx->appliedParams.cParams.windowLog;
+
+    /* This bool is set if there is enough room to output all noCompress superblocks.
+     * Just checks if the number of compressed blocks we can fit in dstCapacity is
+     * greater than the optimistic number of blocks we still have remaining.
+     * This might be UNset when data is uncompressable and we're streaming.  */
+
+    const int enoughDstCapacityForNoCompressSuperBlocks =
+        (dstCapacity / (blockSize + 7 /* header + checksum */)) > (srcSize / blockSize);
```
terrelln: minor: This is a bit pessimistic since we only have 1 checksum per frame, but that is okay since we'll still compress successfully.

bimbashrestha: Yeah, figured it would be better to be pessimistic here

bimbashrestha:
> Does this fix all the open oss-fuzz issues?

The stream round trip aborts yes. I think the pointer overflow alerts are going to go away right @Cyan4973? Because we rely overflowed logic routinely in zstd.

Cyan4973:
>  I think the pointer overflow alerts are going to go away right

right, this was announced by the oss-fuzz team.

PR #1930, opened 2019-12-16
pkgconfig: Don't hardcode libdir.

orbea:
This fixes the build for the current mesa master on my mulitlib Slackware system when `ld(1)` is given `/usr/lib/libzstd.so` instead of `/usr/lib64/libzstd.zo`.

For an example pkgconfig file see `libpkgconf.pc.in` from the pkgconf project.

https://github.com/pkgconf/pkgconf/blob/c862e030cf83447f679e4f49876f5298f0fc9691/libpkgconf.pc.in

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -244,6 +244,7 @@ libzstd.pc:
 libzstd.pc: libzstd.pc.in
 	@echo creating pkgconfig
 	@sed -e 's|@PREFIX@|$(PREFIX)|' \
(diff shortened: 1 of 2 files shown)


Cyan4973:
ping @felixhandte : do you want to follow up on this PR ?

felixhandte:
@orbea, can you check if #2116 solves your issue?

felixhandte:
I believe this need should be met by #2116. Closing for now. If you're still having issues, @orbea, let me know!

PR #1936, opened 2019-12-27
[fuzz] Dividing by targetCBlockSize instead of blockSize for nbBlocks fit

bimbashrestha:
I added this check last week because noCompress blocks in streaming mode were eating up all of dstCapacity when superblock was enabled. I used blockSize to determine the number of blocks that would fit inside srcSize and dstCapacity when I should have used targetCBlockSize. 

Also added some debug statements. Currently running libFuzzer on this locally

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2548,29 +2548,6 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms,
     }
 }
 
-/*! ZSTD_enoughSuperBlockDstCapacityRemaining()
- * Returns true if there is enough room in dstCapacity to fit the remaining
- * data if the remaining data were all compressed as noCompress superblocks.
- * Note: assumes that targetCBlockSize is not 0 (ie. we are using superblocks)
- * and dstCapacity > 4 */
-static int ZSTD_enoughSuperBlockDstCapacityRemaining(ZSTD_CCtx* cctx, size_t dstCapacity, size_t srcSize)
-{
-    const size_t targetCBlockSize = cctx->appliedParams.targetCBlockSize;
-    const size_t nbDstCapacityBlocks = (dstCapacity - ZSTD_FRAMECHECKSUMSIZE) / (targetCBlockSize + ZSTD_BLOCKHEADERSIZE);
-    const size_t nbSrcSizeBlocks = srcSize / targetCBlockSize;
-    int enoughCapacity = nbDstCapacityBlocks >= nbSrcSizeBlocks;
-
-    assert(dstCapacity > 4);
-
-    DEBUGLOG(6, "ZSTD_enoughDstCapacityRemaining(dstCapacity=%u srcSize=%u)",
-        (unsigned)dstCapacity, (unsigned)srcSize);
-    DEBUGLOG(9, "nbDstCapacityBlocks=%u nbSrcSizeBlocks=%u enoughCapacity=%u",
-        (unsigned)nbDstCapacityBlocks, (unsigned)nbSrcSizeBlocks, (unsigned)enoughCapacity);
-
-    return enoughCapacity;
-}
-
-
 /*! ZSTD_compress_frameChunk() :
 *   Compress a chunk of data into one or multiple blocks.
 *   All blocks will be terminated, all input will be consumed.
@@ -2613,8 +2590,7 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
         if (ms->nextToUpdate < ms->window.lowLimit) ms->nextToUpdate = ms->window.lowLimit;
 
         {   size_t cSize;
-            if (ZSTD_useTargetCBlockSize(&cctx->appliedParams)
-              && ZSTD_enoughSuperBlockDstCapacityRemaining(cctx, dstCapacity, srcSize)) {
+            if (ZSTD_useTargetCBlockSize(&cctx->appliedParams)) {
                 cSize = ZSTD_compressBlock_targetCBlockSize(cctx, op, dstCapacity, ip, blockSize, lastBlock);
                 FORWARD_IF_ERROR(cSize);
             } else {
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3796,6 +3796,7 @@ static size_t ZSTD_compressStream_generic(ZSTD_CStream* zcs,
 
         case zcss_load:
             if ( (flushMode == ZSTD_e_end)
+              && !ZSTD_useTargetCBlockSize(&zcs->appliedParams)
               && ((size_t)(oend-op) >= ZSTD_compressBound(iend-ip))  /* enough dstCapacity */
               && (zcs->inBuffPos == 0) ) {
                 /* shortcut to compression pass directly into output buffer */
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2462,6 +2462,7 @@ static size_t ZSTD_compressBlock_targetCBlockSize_body(ZSTD_CCtx* zc,
                                const void* src, size_t srcSize,
                                const size_t bss, U32 lastBlock)
 {
+    DEBUGLOG(6, "Attempting ZSTD_compressSuperBlock()");
     /* Attempt superblock compression and return early if successful */
     if (bss == ZSTDbss_compress) {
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_compress.c
```diff
@@ -2568,8 +2572,9 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
      * greater than the optimistic number of blocks we still have remaining.
      * This might be UNset when data is uncompressable and we're streaming.  */
 
-    const int enoughDstCapacityForNoCompressSuperBlocks =
-        (dstCapacity / (blockSize + 7 /* header + checksum */)) > (srcSize / blockSize);
+    const size_t targetCBlockSize = cctx->appliedParams.targetCBlockSize;
+    const int enoughDstCapacityForNoCompressSuperBlocks = !targetCBlockSize ? 0 :
```
Cyan4973: Is it possible for `targetCBlockSize` to be `== 0` at this point in  the code ?
If not, this should be an `assert()`, rather than a permanent test.

lib/compress/zstd_compress.c
```diff
@@ -2568,8 +2572,9 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
...
+        (dstCapacity / (targetCBlockSize + 7 /* header + checksum */)) > (srcSize / targetCBlockSize);
```
Cyan4973: minor comment :
`header` is present at each block,
while `checksum` can only exist once per frame.

More accurate formula should be `(dstCapacity - 4) / (targetCBlockSize + 3)`.

This, however, requires to ensure that `dstCapacity >= 4`.

Cyan4973:
Is it possible to add a test
that would fail with previous version of the code
and pass with the new version ?

lib/compress/zstd_compress.c
Cyan4973: Also :
try to avoid "magic" numbers, prefer referring to some constant defined for the project.

For example, the block size is [`ZSTD_BLOCKHEADERSIZE`](https://github.com/facebook/zstd/blob/dev/lib/common/zstd_internal.h#L127).
Sure it still means `3`, but this is more reliable for long-term maintenance (meaning is clear).

_note_ : it seems we are missing a kind of `ZSTD_FRAMECHECKSUMSIZE` constant for the second part.

lib/compress/zstd_compress.c
bimbashrestha: Yes, it is possible for `targetCBlockSize == 0`. I had to include this because fuzzer.c crashed so having an assert here would do that same. Maybe we should enforce `targetCBlockSize != 0` before it reaches this function and add an assert here instead?

Cyan4973: Well, I confused the position in the source code : reading `enoughDstCapacityForNoCompressSuperBlocks`, I assumed this code was in a superblock-dedicated function. But it's not, it's in `ZSTD_compress_frameChunk()`, which is generic, and therefore `targetCBlockSize` can be `0`, meaning "no superblock".

However, this asks another question : why is this test done in `ZSTD_compress_frameChunk()` ?

The way I understand the code, the codec will now disable superblock when `dstCapacity` is not large enough for the pessimistic case. This feels a bit too strong. Especially, I'm worried that "normal" scenarios, such as "average size" data (~100 KB) receiving `dstCapacity = ZSTD_compressBound(srcSize)`, may end up constantly disabling superblock. That would defeat the purpose of superblock.

bimbashrestha: Yeah, that's a good point. It might be a little too sensitive to put it here. I'll move it inside superblock

terrelln:
This disabling logic seems tricky. I don't remember exactly how all the superblock code works, so maybe this obviously doesn't work.

Can we fall back to emitting a raw block if the total sum of the super block sizes is larger than the original input size + block header size? If we do this we ensure that we never violate the `ZSTD_compressBound()`. This still leaves the case where we now have higher latency than before, because we have a 1 large block. But, since it should be an uncompressed block, we can fix that in the streaming API by "streaming" the uncompressed block without requiring the whole block to be present.

bimbashrestha:
I now perform the dstCapacity check only when we're using superblock mode. With this setup, we might alternate between having enough capacity and not having enough capacity for noCompress superblocks in streaming mode (which is fine because we produce a superblock whenever possible now).

lib/compress/zstd_compress.c
```diff
@@ -2544,6 +2548,29 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms,
...
+ * Note: assumes that targetCBlockSize is not 0 (ie. we are using superblocks)
+ * and dstCapacity > 4 */
+static int ZSTD_enoughSuperBlockDstCapacityRemaining(ZSTD_CCtx* cctx, size_t dstCapacity, size_t srcSize)
+{
+    const size_t targetCBlockSize = cctx->appliedParams.targetCBlockSize;
+    const size_t nbDstCapacityBlocks = (dstCapacity - ZSTD_FRAMECHECKSUMSIZE) / (targetCBlockSize + ZSTD_BLOCKHEADERSIZE);
+    const size_t nbSrcSizeBlocks = srcSize / targetCBlockSize;
+    int enoughCapacity = nbDstCapacityBlocks >= nbSrcSizeBlocks;
+
+    assert(dstCapacity > 4);
```
Cyan4973: are you sure that `dstCapacity` can't be `> 4` ?

lib/compress/zstd_compress.c
```diff
@@ -2593,8 +2613,8 @@ static size_t ZSTD_compress_frameChunk (ZSTD_CCtx* cctx,
         if (ms->nextToUpdate < ms->window.lowLimit) ms->nextToUpdate = ms->window.lowLimit;
 
         {   size_t cSize;
-            int useTargetCBlockSize = ZSTD_useTargetCBlockSize(&cctx->appliedParams);
-            if (useTargetCBlockSize && enoughDstCapacityForNoCompressSuperBlocks) {
+            if (ZSTD_useTargetCBlockSize(&cctx->appliedParams)
+              && ZSTD_enoughSuperBlockDstCapacityRemaining(cctx, dstCapacity, srcSize)) {
```
Cyan4973: The test is now played only when superblock is active.

But it doesn't change the issue that 
it effectively disables superblock whenever `dstCapacity` is not large enough to handle the worst case scenario (noCompress superblocks).

This pre-emptive disabling seems too strong because :
- The disabling happens without notification nor explanation. That's fine if if fixes an exceptional scenario, not if this happens regularly.
- most users will select either `dstCapacity = ZSTD_compressBound(srcSize)`, or `dstCapacity = srcSize` 
- `compressBound()` is the recommended method, but it's unclear to me if it provides enough to meet the new condition related to superblock. If not, that means that many (if not most) users will trigger this disabling, all transparently.

I do believe this has too big an impact.

Disabling superblock is still a valid strategy, but only _after_ the algorithm get a chance to compress the data with superblock. Only when that fails does it make sense to trigger backup strategies.

bimbashrestha:
> Can we fall back to emitting a raw block if the total sum of the super block sizes is larger than the original input size + block header size?

I missed this last week for some reason. Yes, this should work.

> Disabling superblock is still a valid strategy, but only after the algorithm get a chance to compress the data with superblock. Only when that fails does it make sense to trigger backup strategies.

Ah makes sense. In that case, I think @terrelln's proposed approach would be preferable to this. I'll update this PR with that

tests/fuzzer.c
```diff
@@ -501,14 +501,14 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
-        /* The first 1/3 of the buffer is compressible and the last 2/3 is
+        /* The first 1/5 of the buffer is compressible and the last 4/5 is
          * uncompressible. This is an approximation of the type of data
          * the fuzzer generated to catch this bug. Streams like this were making
          * zstd generate noCompress superblocks (which are larger than the src
          * they come from). Do this enough times, and we'll run out of room
          * and throw a dstSize_tooSmall error. */
 
-        const size_t compressiblePartSize = srcSize/3;
+        const size_t compressiblePartSize = srcSize/5;
```
Cyan4973: minor question :
what's the reason behind this change of proportion ?

bimbashrestha: The error caught was effectively the same as before (we just can't catch it with what we had because there wasn't enough uncompressible data and thus not enough noCompress superblocks). This just makes it so that there is more of the uncompressible part. As for the 3 -> 5, that was kind of arbitrary (probably would work just fine with 6 or 7 but didn't check)

PR #1938, opened 2020-01-03
[doc] Adding -M# decompression operation modifier documentation

bimbashrestha:


programs/zstd.1.md
```diff
@@ -122,6 +122,10 @@ the last one takes effect.
 
     Note: If `windowLog` is set to larger than 27, `--long=windowLog` or
     `--memory=windowSize` needs to be passed to the decompressor.
+* `-M#`:
```
Cyan4973: There is also an equivalent --long-command, named : `--memory=#`

programs/zstd.1.md
```diff
@@ -122,6 +122,10 @@ the last one takes effect.
...
+    Set a memory usage limit for decompression. By default, Zstandard uses 128 MB
+    as the maximum amount of memory the decompressor is allowed ot use, but you can
+    override this manually if need be.
```
Cyan4973: both upwards and downward.

programs/zstd.1.md
```diff
@@ -122,6 +122,10 @@ the last one takes effect.
 
     Note: If `windowLog` is set to larger than 27, `--long=windowLog` or
     `--memory=windowSize` needs to be passed to the decompressor.
+* `-M#`:
+    Set a memory usage limit for decompression. By default, Zstandard uses 128 MB
+    as the maximum amount of memory the decompressor is allowed ot use, but you can
```
Cyan4973: nit: `ot` => `to`

Cyan4973:
I'll do the conversion into `man` format in another commit,
since it requires a specific tool with dependency on `ruby`, that may not be setup on your environment.

PR #1939, opened 2020-01-04
fix UBSAN's invalid-null-argument error in zstd_decompress.c

igorsugak:
This was exposed when running another project's tests that use zstd with UBSAN instrumentation.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1482,7 +1482,9 @@ size_t ZSTD_estimateDStreamSize_fromFrame(const void* src, size_t srcSize)
 MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src, size_t srcSize)
 {
     size_t const length = MIN(dstCapacity, srcSize);
-    memcpy(dst, src, length);
+    if (length > 0) {
+        memcpy(dst, src, length);
+    }
     return length;
 }
 


terrelln:
Thanks for the patch! Before we merge, I would like to add a test that exposes it, since we're missing this coverage. It looks like it is triggered by a `NULL` input/output buffer in the streaming decompression API.

terrelln:
> Out of curiosity, what project exposed this?

Our tests in `folly/compression/test:compression_test` when compiled with a new build mode with more UBSAN options enabled.

terrelln:
The test in PR #1945 catches this bug, merging now.

Cyan4973:
2 points :
- the fix in `folly` is different : the branch is `if (src != NULL)`,
which might provide better latency, since `length` must be calculated first.
- It would have been nice to measure potential speed difference introduced by this patch. I don't anticipate any, just it would be nice to measure it.

terrelln:
* You would also have to check `dst != NULL`.
* This is only used to copy into/out-of in/out buffers. I measured the speed of a decompression on the CLI before/after:

```
Before:
> sudo cset shield --exec -- perf stat -r 10 -- ./bin/gcc-9.2.0/b77ad810c9bd7b33be22ff5e4d317d275496603c/zstd -d ~/datasets/silesia.tar.zst -o /dev/null

 Performance counter stats for './bin/gcc-9.2.0/b77ad810c9bd7b33be22ff5e4d317d275496603c/zstd -d /home/nick/datasets/silesia.tar.zst -o /dev/null' (10 runs):

            221.39 msec task-clock                #    0.999 CPUs utilized            ( +-  0.03% )
                 0      context-switches          #    0.001 K/sec                    ( +- 50.92% )
                 0      cpu-migrations            #    0.000 K/sec
               368      page-faults               #    0.002 M/sec
       797,002,268      cycles                    #    3.600 GHz                      ( +-  0.03% )
     2,010,782,507      instructions              #    2.52  insn per cycle           ( +-  0.00% )
       191,675,313      branches                  #  865.774 M/sec                    ( +-  0.00% )
         4,399,170      branch-misses             #    2.30% of all branches          ( +-  0.01% )

         0.2215405 +- 0.0000674 seconds time elapsed  ( +-  0.03% )

After:
> sudo cset shield --exec -- perf stat -r 10 -- ./bin/gcc-9.2.0/03ffda7b88da5a37c9c31478494b8768da50a008/zstd -d ~/datasets/silesia.tar.zst -o /dev/null

 Performance counter stats for './bin/gcc-9.2.0/03ffda7b88da5a37c9c31478494b8768da50a008/zstd -d /home/nick/datasets/silesia.tar.zst -o /dev/null' (10 runs):

            220.25 msec task-clock                #    0.999 CPUs utilized            ( +-  0.06% )
                 0      context-switches          #    0.001 K/sec                    ( +- 66.67% )
                 0      cpu-migrations            #    0.000 K/sec
               368      page-faults               #    0.002 M/sec
       792,897,382      cycles                    #    3.600 GHz                      ( +-  0.06% )
     2,010,789,554      instructions              #    2.54  insn per cycle           ( +-  0.00% )
       191,679,200      branches                  #  870.274 M/sec                    ( +-  0.00% )
         4,382,577      branch-misses             #    2.29% of all branches          ( +-  0.01% )

          0.220431 +- 0.000124 seconds time elapsed  ( +-  0.06% )

dst != NULL && src != NULL
> sudo cset shield --exec -- perf stat -r 10 -- ./zstd -d ~/datasets/silesia.tar.zst -o /dev/null

 Performance counter stats for './zstd -d /home/nick/datasets/silesia.tar.zst -o /dev/null' (10 runs):

            220.35 msec task-clock                #    0.999 CPUs utilized            ( +-  0.05% )
                 0      context-switches          #    0.001 K/sec                    ( +- 66.67% )
                 0      cpu-migrations            #    0.000 K/sec
               368      page-faults               #    0.002 M/sec
       793,241,466      cycles                    #    3.600 GHz                      ( +-  0.05% )
     2,010,797,362      instructions              #    2.53  insn per cycle           ( +-  0.00% )
       191,680,706      branches                  #  869.903 M/sec                    ( +-  0.00% )
         4,381,078      branch-misses             #    2.29% of all branches          ( +-  0.01% )

          0.220532 +- 0.000109 seconds time elapsed  ( +-  0.05% )

```

Decompression is reliably 0.5% faster after this patch with gcc. It's probably just jitter in how the code is laid out. Both the length check and the NULL checks are the same speed, so lets go with the simpler one.

Issue #1935, opened 2019-12-25
idea/request for help: zstd as a binary diff producer/patcher?


gima:
### Summary
I would like to produce a diff file (quickly) of the changes from some binary file A to B (where B is a changed version of A). *(i managed to do this with zstd, but, see below: "What I tried" section)*.

It could be said (from my naive viewpoint) that finding differences between files is somewhat the domain of dictionary-based compression programs. So why re-invent the wheel (and create yet another new software)?

### Reasoning
For starters, there currently seems to be a lack of stand-alone utilities that do this. All of them seem to be tied to something else. Be it zsync (unmaintained, as far as I see) being tied to urls and http, bsdiff taking nearly 30 seconds to generate the diff file (whereas zstd does this in under a second). Then bigger tools such as casync require all-or-nothing adoption of their way of doing things.

Secondly, most(?) Linux distros provide package updates as a totally new files-to-be-downloaded. There are major bandwidth (and monetary) savings that could be had here if an efficient (and easy / stand alone) binary diff could be had.

.. And again, since zstd needs to find repetitions and their positions in files, exposing functionality that supports using all of this to produce and use diff files (or at least stapling this functionality to public api) could be a good fit here.

### What I tried
I managed to use zstd to produce a very small diff file of changes from binary file A to B (with very fast creation time; less than 1sec) of around 1-2KB for both test cases ("simple" and "complex"). This small diff file was then given to zstd as the-file-to-be-decompressed and the original binary file (A) was given to zstd to use as dictionary. This procedure was able to reproduce the binary file B.

<details><summary>[Click to show transcript of the commands used]</summary>
preparing the file that's being used in this experiment:
```fish
$ cp /bin/qemu-system-x86_64 bin
```

splitting the binary file in two and showing that the when combined, the splits are equal to the original
```fish
$ split -n 2 -d bin bin.split-

$ cat bin.split-00 bin.split-01 | diff -s bin /dev/stdin
Files bin and /dev/stdin are identical
```

putting second half of the file in place
```fish
$ cat bin.split-01 bin.split-00 > binrev

$ cat bin.split-01 > bintailhalf
```

listing current state of directory
```fish
$ lf
16227960  ./bin
8113980   ./bin.split-00
8113980   ./bin.split-01
16227960  ./binrev
8113980   ./bintailhalf
```

test #1 ("simple"): compressing, decompressing and comparing (using the original binary file as dictionary)
```fish
# compressing
[...]

PR #1940, opened 2020-01-05
Adding new cli endpoint --patch-from=

bimbashrestha:
* Add new cli endpoint --diff-from 
* Overrides 32 MB dict limit with --memory=# when using --diff-from
* Ensures that windowSize > srcSize on compression using --diff-from
https://github.com/facebook/zstd/issues/1935

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -77,6 +77,7 @@
 
 #define FNSPACE 30
 
+#define DIFFFROM_WINDOWSIZE_EXTRA_BYTES 1 KB
 
 /*-*************************************
 *  Macros
@@ -321,6 +322,7 @@ struct FIO_prefs_s {
     int nbWorkers;
 
     int excludeCompressedFiles;
+    int diffFromMode;
 };
 
 
@@ -487,6 +489,10 @@ void FIO_setLdmHashRateLog(FIO_prefs_t* const prefs, int ldmHashRateLog) {
     prefs->ldmHashRateLog = ldmHashRateLog;
 }
 
+void FIO_setDiffFromMode(FIO_prefs_t* const prefs, int diffFromMode)
+{
+    prefs->diffFromMode = diffFromMode;
+}
 
 /*-*************************************
 *  Functions
@@ -624,7 +630,7 @@ FIO_openDstFile(FIO_prefs_t* const prefs,
  * @return : loaded size
  *  if fileName==NULL, returns 0 and a NULL pointer
  */
-static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName)
+static size_t FIO_createDictBuffer(FIO_prefs_t* const prefs, void** bufferPtr, const char* fileName)
 {
     FILE* fileHandle;
     U64 fileSize;
@@ -638,9 +644,12 @@ static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName)
     if (fileHandle==NULL) EXM_THROW(31, "%s: %s", fileName, strerror(errno));
 
     fileSize = UTIL_getFileSize(fileName);
-    if (fileSize > DICTSIZE_MAX) {
-        EXM_THROW(32, "Dictionary file %s is too large (> %u MB)",
-                        fileName, DICTSIZE_MAX >> 20);   /* avoid extreme cases */
+    {
+        size_t dictSizeMax = prefs->diffFromMode ? prefs->memLimit : DICTSIZE_MAX;
+        if (fileSize >  dictSizeMax) {
+            EXM_THROW(32, "Dictionary file %s is too large (> %u bytes)",
+                            fileName,  (unsigned)dictSizeMax);   /* avoid extreme cases */
+        }
     }
     *bufferPtr = malloc((size_t)fileSize);
     if (*bufferPtr==NULL) EXM_THROW(34, "%s", strerror(errno));
@@ -759,9 +768,16 @@ typedef struct {
     ZSTD_CStream* cctx;
 } cRess_t;
 
+static unsigned int FIO_log2(const size_t x)
+{
+    size_t tmp = x; int res = 0;
+    while (tmp >>= 1) res++;
+    return res;
+}
+
 static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
-                                    const char* dictFileName, int cLevel,
-                                    ZSTD_compressionParameters comprParams) {
+                                    const char* dictFileName, const size_t maxSrcFileSize,
+                                    int cLevel, ZSTD_compressionParameters comprParams) {
     cRess_t ress;
     memset(&ress, 0, sizeof(ress));
 
@@ -779,13 +795,17 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
 
     /* Advanced parameters, including dictionary */
     {   void* dictBuffer;
-        size_t const dictBuffSize = FIO_createDictBuffer(&dictBuffer, dictFileName);   /* works with dictFileName==NULL */
+        size_t const dictBuffSize = FIO_createDictBuffer(prefs, &dictBuffer, dictFileName);   /* works with dictFileName==NULL */
         if (dictFileName && (dictBuffer==NULL))
             EXM_THROW(32, "allocation error : can't create dictBuffer");
         ress.dictFileName = dictFileName;
 
         if (prefs->adaptiveMode && !prefs->ldmFlag && !comprParams.windowLog)
             comprParams.windowLog = ADAPT_WINDOWLOG_DEFAULT;
+
+        if (prefs->diffFromMode) {
+            comprParams.windowLog = FIO_log2(maxSrcFileSize + DIFFFROM_WINDOWSIZE_EXTRA_BYTES);
+        }
 
         CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_c_contentSizeFlag, 1) );  /* always enable content size when available (note: supposed to be default) */
         CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_c_dictIDFlag, prefs->dictIDFlag) );
@@ -1515,7 +1535,7 @@ int FIO_compressFilename(FIO_prefs_t* const prefs, const char* dstFileName,
                          const char* srcFileName, const char* dictFileName,
                          int compressionLevel,  ZSTD_compressionParameters comprParams)
 {
-    cRess_t const ress = FIO_createCResources(prefs, dictFileName, compressionLevel, comprParams);
+    cRess_t const ress = FIO_createCResources(prefs, dictFileName, (size_t)UTIL_getFileSize(srcFileName), compressionLevel, comprParams);
     int const result = FIO_compressFilename_srcFile(prefs, ress, dstFileName, srcFileName, compressionLevel);
 
 
@@ -1563,6 +1583,15 @@ FIO_determineCompressedName(const char* srcFileName, const char* outDirName, con
     return dstFileNameBuffer;
 }
 
+static size_t FIO_getLargestFileSize(const char** inFileNames, unsigned nbFiles)
+{
+    size_t i; size_t fileSize; size_t maxFileSize = 0;
+    for (i = 0; i < nbFiles; i++) {
+        fileSize = UTIL_getFileSize(inFileNames[i]);
+        maxFileSize = fileSize > maxFileSize ? fileSize : maxFileSize;
+    }
+    return maxFileSize;
+}
 
 /* FIO_compressMultipleFilenames() :
  * compress nbFiles files
@@ -1578,7 +1607,9 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
                                   ZSTD_compressionParameters comprParams)
 {
     int error = 0;
-    cRess_t ress = FIO_createCResources(prefs, dictFileName, compressionLevel, comprParams);
+    cRess_t ress = FIO_createCResources(prefs, dictFileName,
+        FIO_getLargestFileSize(inFileNamesTable, nbFiles),
+        compressionLevel, comprParams);
 
     /* init */
     assert(outFileName != NULL || suffix != NULL);
@@ -1648,7 +1679,7 @@ static dRess_t FIO_createDResources(FIO_prefs_t* const prefs, const char* dictFi
 
     /* dictionary */
     {   void* dictBuffer;
-        size_t const dictBufferSize = FIO_createDictBuffer(&dictBuffer, dictFileName);
+        size_t const dictBufferSize = FIO_createDictBuffer(prefs, &dictBuffer, dictFileName);
         CHECK( ZSTD_initDStream_usingDict(ress.dctx, dictBuffer, dictBufferSize) );
         free(dictBuffer);
     }
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -630,7 +630,7 @@ FIO_openDstFile(FIO_prefs_t* const prefs,
  * @return : loaded size
  *  if fileName==NULL, returns 0 and a NULL pointer
  */
-static size_t FIO_createDictBuffer(FIO_prefs_t* const prefs, void** bufferPtr, const char* fileName)
+static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName, FIO_prefs_t* const prefs)
 {
     FILE* fileHandle;
     U64 fileSize;
@@ -801,7 +801,7 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
 
     /* Advanced parameters, including dictionary */
     {   void* dictBuffer;
-        size_t const dictBuffSize = FIO_createDictBuffer(prefs, &dictBuffer, dictFileName);   /* works with dictFileName==NULL */
+        size_t const dictBuffSize = FIO_createDictBuffer(&dictBuffer, dictFileName, prefs);   /* works with dictFileName==NULL */
         if (dictFileName && (dictBuffer==NULL))
             EXM_THROW(32, "allocation error : can't create dictBuffer");
         ress.dictFileName = dictFileName;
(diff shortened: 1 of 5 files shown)


programs/fileio.c
```diff
@@ -487,6 +489,10 @@ void FIO_setLdmHashRateLog(FIO_prefs_t* const prefs, int ldmHashRateLog) {
     prefs->ldmHashRateLog = ldmHashRateLog;
 }
 
+void FIO_setDiffFromMode(FIO_prefs_t* const prefs, int diffFromMode)
+{
+    prefs->diffFromMode = diffFromMode;
```
Cyan4973: minor : 
`diffFromMode` is essentially a flag,
so it's better to consider it a `0` / `1` value, and not have to worry afterwards that the "real" value it carries might be different.

One way to ensure that outcome is : 
`prefs->diffFromMode = (diffFromMode != 0);`

programs/zstdcli.c
```diff
@@ -691,6 +693,7 @@ int main(int const argCount, const char* argv[])
                 if (!strcmp(argument, "--rm")) { FIO_setRemoveSrcFile(prefs, 1); continue; }
                 if (!strcmp(argument, "--priority=rt")) { setRealTimePrio = 1; continue; }
                 if (!strcmp(argument, "--output-dir-flat")) {nextArgumentIsOutDirName=1; lastCommand=1; continue; }
+                if (!strcmp(argument, "--diff-from")) {nextArgumentIsDiffFromDictFileName=1; lastCommand=1; continue; }
```
Cyan4973: consider using `--diff-from=#` instead.

Advantages are : 
- Reading the file name is done in the same operation. This avoids to "split" the logic in 2 different places and times, adding more code, and with potentially something going wrong in between.
- This policy is more consistent with other CLI arguments. I know we do have a few old arguments left that still use the 2-reads strategy, but this is now discouraged, and may eventually be displaced if possible (as an example, I really believe that the recent `--output-dir-flat` should be removed, and only `--output-dir-flat=#` should be advertised and remain (this is a different topic, unrelated to this PR)).

bimbashrestha: Ah sure. I was wondering why there were two instances of the same command. I have --diff-from=# in there. I'll just go ahead and remove this endpoint then.

programs/zstdcli.c
```diff
@@ -597,6 +598,7 @@ int main(int const argCount, const char* argv[])
     const char* outFileName = NULL;
     const char* outDirName = NULL;
     const char* dictFileName = NULL;
+    const char* diffFromDictFileName = NULL;
```
Cyan4973: Having 2 different variables `dictFileName` and `diffFromDictFileName` makes it now possible to define them both.
What happens then ?
The behavior should be defined, and documented.

Suggestion : 
- Only one of them should be possible. If both are defined, then, it's either an error, or the latest one takes over. I have a small preference for `error` in this case.
- Having a single variable for both makes it impossible to define both. But it requires another state variable to define what the value means. So there is no reduction in nb of variables. Hence, I have no preference : select whichever feels better to you.

bimbashrestha: Yes, just having one variable is very possible. Will update

bimbashrestha:
> My sense of the UNIX philosophy has always been that `--option arg` is more popular / canonical than `--option=arg`, and really that tools should accept both. Do you feel differently?

I've seen both used but I'm not sure what the standard for argument syntax on POSIX is. GNU seems to prefer `--option=arg` (https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) 

I don't really have a strong preference either way. Although because it is slightly annoying to have to support two different parsings, I think I slightly prefer `--option=arg` for options with arguments. Of course, I think `--option` is still just fine for argument-less options.

bimbashrestha:
> Given all that, I suppose I'm leaning more towards a CLI that looks like zstd --large -D v1/foo v2/foo -o foo-v1-to-v2.zst.

I agree that `--diff-from=FILE` might be misleading because the generated patch ins't bidirectional. This was chosen mostly because the issue that brought this up used the language of `diff`. I'm fine with switching `zstd --large -D`

programs/zstd.1.md
```diff
@@ -122,11 +122,20 @@ the last one takes effect.
 
     Note: If `windowLog` is set to larger than 27, `--long=windowLog` or
     `--memory=windowSize` needs to be passed to the decompressor.
+* `--diff-from=#`:
```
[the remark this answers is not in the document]
bimbashrestha: Ah good catch. I'll update this once we decide between keeping this UI or adopting the new one your proposed.

felixhandte:
@bimbashrestha, good find on that source!

One interesting point they make is that the `--option=value`  syntax has functionality that is superior to `--option value`, since it can handle optional values, which is very, very tricky with `--option value`. I had never considered that.

Nonetheless, they do say:

> An option and its argument may or may not appear as separate tokens.

My sense is generally that the upkeep burden of supporting both argument types is worth it. I'm curious if @Cyan4973 feels differently?

terrelln:
We haven't been super consistent in supporting both `--option=value` and `--option value`. I think it is best to support both.

Cyan4973:
There are a few topics now in this thread.
Let's handle the first one, regarding `--long-command=arg` vs `--long-command arg` syntax.

At this stage, I'm in favor of supporting `--long-command=arg` syntax _only_ .
There is a temporary exception for older `--long-command` which, initially, were not supporting the `=arg` syntax, because some scripts may already exist using the separate ` arg` syntax, but that's it, all new commands shouldn't allow it, and even the older ones should be expected to phase out gradually.

Now, an important reason for this preference is that our current code to support `-long-command arg` is brittle. It creates a multi-stages scenarios, separates the code logic into multiple places, requiring a dedicated tracking state per command, and introduces potential side-effects when the sequence of arguments is not the expected one.
For example, let's consider the sequence `--output-dir-flat -q dir`. How should the CLI behave ? Is that an error ? is `-q` now the name of target directory ? or is it skipped in favor of `dir` ? what about `--output-dir-flat -v --diff-from file dir input` then ? where is it defined ? is it consistent at every place where such situation may happen ? 

There is then a question of consistency : 
should `--long-command arg` syntax be only available when `arg` is a _name_, or should it also support _numbers_ ? We have a lot of commands with numbers already.
For example, `--long=#` is currently supported. Should the CLI also support `--long #` ?
In that case, what about default arguments ? Currently `--long` means "just long mode, use default argument". It doesn't expect any argument on command line. How to do that if `--long` can be followed by a numerical argument ?

The advantage of `--long-command=arg` is that it is clear and simple, both to code and to explain.

If _really_ it is important to support both syntaxes, I would then insist on improving the code in `zstdcli.c` which handles the `--long-command arg` case. It shall be modified to ensure that code complexity is well located, in a single place. This may be possible, using a macro, to hide all the dirty parts. One issue is that it will have to modify the loop counter inside the loop, which is instinctively a bad thing, but could be made to work nonetheless. That won't help for consistency issues though ...

felixhandte:
I agree: `--long-command arg` is more complicated to handle. But it is easier to invoke. And I suspect it's a lot more popular in usage. I basically never use the `--option=value` syntax, I suspect most folks are the same (you could survey StackOverflow answers shell commands, if you wanted to gather data).

I am all in favor of cleaning up `zstdcli.c` so that both versions get parsed in the same way. I might even be willing to volunteer to do the clean up.

terrelln:
When I use CLIs I always use `--option value`, and I expect them to support both.

For now, I think we should continue to add new options as `--option=value`, since the code is a mess otherwise. If we decide to support `--option value` uniformly, it should be done together with a code refactor that simplifies and cleans up `zstdcli.c`. Ideally in a way that can be easily reused.

> For example, let's consider the sequence --output-dir-flat -q dir. How should the CLI behave ? Is that an error ? is -q now the name of target directory ? or is it skipped in favor of dir ? what about --output-dir-flat -v --diff-from file dir input then ? where is it defined ? is it consistent at every place where such situation may happen ?

I would say that would be treated as an error if the argument was required, and the default no argument if it is optional. If the user wanted the argument to be `-q` they would have to use the `--option=-q` syntax.

terrelln:
@Cyan4973 pointed out the problem with `--long 30` as an optional argument. Is 30 an input file or a number? What about `--long asdf`, parsing error or input file?

I see two paths forwards:
1. If the argument is optional only allow `--option=value`.
2. Decide that optional arguments were a mistake and deprecate them. Replace the argument with a second flag, like `--long --wlog 30`.

felixhandte:
Yeah. Both of those seem reasonable to me.

Cyan4973:
Option 2 would break existing convention, hence could break deployed scripts.
I'm more in favor of Option 1.

bimbashrestha:
Option 1 sounds good to me as well. Changes to ensure that can go in a different PR. I can take on that task. 

Thoughts about the other changes on this PR and the UI suggestion?

terrelln:
Either is fine for me, but by deprecate I meant not introduce any new optional values, and provide alternative ways to write the values that are currently optional, but still support the legacy optional parameters with `--option=value`.

Cyan4973:
Regarding the definition of "diff" : the name comes from [original request](https://github.com/facebook/zstd/issues/1935), but if a diff _must_ be bi-directional, then the feature described here is clearly not, so maybe it could get another name. Which one then ? patch ? update ?

`--diff-from=` is initially a shortcut : it's supposed to combine dictionary compression with all necessary parameters (such as window size and memory usage) in a single command, to make this scenario accessible.

However, it turns out that, for this functionality to work properly, it actually needs to do a few things differently (so it's effectively not imitable by `-D dictionary`) : 

- Yes, to answer a question by @felixhandte , it shall force the reference content to be loaded as "raw" dictionary
- Presuming that some patches will be created from large reference files (think hundreds of MB), it must be compatible with the `--long` mode
- `--long` is not compatible with `cdict`, so it shall use `ZSTD_CCtx_loadDictionary()` instead, bypassing the `CDict` creation stage.

We still have a few issue though : 
- with our current design, the reference content must hold entirely in memory, putting a limit to its size.
This could be mitigated by mmaping the dictionary content if it's too large, though it will impact performance.
   + This feels achievable, though could be categorized as a "stretch goal".
- The `--long` mode still can't get beyond 2 GB. We would need to change the index map to use 64-bit values instead.
   + this is a much more involved change, potentially a candidate for a separate PR.

programs/fileio.c
```diff
@@ -1563,6 +1583,15 @@ FIO_determineCompressedName(const char* srcFileName, const char* outDirName, con
     return dstFileNameBuffer;
 }
 
+static size_t FIO_getLargestFileSize(const char** inFileNames, unsigned nbFiles)
+{
+    size_t i; size_t fileSize; size_t maxFileSize = 0;
```
Cyan4973: minor : it's common to define multiple variables of same type in the same declaration,
such as : 
`size_t i, fileSize, maxFileSize=0;`

programs/fileio.c
```diff
@@ -779,14 +795,18 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
+        size_t const dictBuffSize = FIO_createDictBuffer(prefs, &dictBuffer, dictFileName);   /* works with dictFileName==NULL */
         if (dictFileName && (dictBuffer==NULL))
             EXM_THROW(32, "allocation error : can't create dictBuffer");
         ress.dictFileName = dictFileName;
 
         if (prefs->adaptiveMode && !prefs->ldmFlag && !comprParams.windowLog)
             comprParams.windowLog = ADAPT_WINDOWLOG_DEFAULT;
 
+        if (prefs->diffFromMode) {
+            comprParams.windowLog = FIO_log2(maxSrcFileSize + DIFFFROM_WINDOWSIZE_EXTRA_BYTES);
```
Cyan4973: Big issue : the `cdict` will not preserve the `windowlog` information.

`windowlog` is derived from the actual `srcSize` at compression time. `cdict` will not be able to influence it.

Cyan4973: I confused that part initially : it seems that `comprParams` is used to setup parameters in `ress.cctx`, using `ZSTD_CCtx_setParameter()`. In which case, that's fine : `ZSTD_c_windowLog` is a sticky parameter, it should still be valid when starting next compression.

Cyan4973:
When code is ready, try the new command on a patching scenario : create a fake `fileB` from a `fileA` + a few inserted perturbations. Check how well the `fileA -> fileB` patch compresses.

You'll need to test several sizes, including some very large ones, in the ~100s MB range.

Without the `--long` mode, I would expect the patching process to lose efficiency after a certain size (to be determined).

That will provide a baseline, offering good reason and a good comparison scenario to introduce the `--long` mode in the patching process.

bimbashrestha:
Tried it with silesia/mozilla (truncated for the old version) and it seems to get worse with higher sizes. 
```
old_size=5000 && new_size=5100 && head -c $old_size silesia/mozilla > old_version && head -c $new_size silesia/mozilla > new_version && ./zstd -f --patch-from=old_version new_version -o patch && ./zstd -df --patch-from=old_version patch -o updated_version
new_version          : 10.47%   (  5100 =>    534 bytes, patch)
patch                : 5100 bytes
old_size=5000000 && new_size=5100000 && head -c $old_size silesia/mozilla > old_version && head -c $new_size silesia/mozilla > new_version && ./zstd -f --patch-from=old_version new_version -o patch && ./zstd -df --patch-from=old_version patch -o updated_version
new_version          :  7.77%   (5100000 => 396126 bytes, patch)
patch                : 5100000 bytes
old_size=50000000 && new_size=51000000 && head -c $old_size silesia/mozilla > old_version && head -c $new_size silesia/mozilla > new_version && ./zstd -f --patch-from=old_version new_version -o patch && ./zstd -df --patch-from=old_version patch -o updated_version
new_version          : 36.18%   (51000000 => 18450530 bytes, patch)
patch                : 51000000 bytes
```

programs/fileio.c
```diff
@@ -1585,7 +1585,7 @@ FIO_determineCompressedName(const char* srcFileName, const char* outDirName, con
 
 static size_t FIO_getLargestFileSize(const char** inFileNames, unsigned nbFiles)
 {
-    size_t i; size_t fileSize; size_t maxFileSize = 0;
+    size_t i, fileSize, maxFileSize = 0;
```
bimbashrestha: Hmm. freeBSD tests failed after this. I'm just going to change this back to what I had before @Cyan4973 '

Cyan4973: This [multi-variables declaration](https://www.tutorialspoint.com/cprogramming/c_variables.htm) is standard C, 
so all compilers are expected to support it.

The freeBSD tests [work fine](https://github.com/facebook/zstd/runs/380215789).

The failing test was a flacky one, named "Minimal Decompressor Macros", on travisCI.
It's the first time I can remember seeing this specific test flacky, 
but unfortunately all tests can be occasionally flacky, 
so it's good to check the actual reason of the failure.

In this case, compilation failed while parsing a strange symbol in an unrelated file (unmodified in the commit).
Restarting the test, it now passes without issue : 
https://travis-ci.org/facebook/zstd/jobs/634432963?utm_medium=notification&utm_source=github_status

bimbashrestha: Ah okay. Noted!

programs/fileio.c
```diff
@@ -638,9 +644,12 @@ static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName)
     if (fileHandle==NULL) EXM_THROW(31, "%s: %s", fileName, strerror(errno));
 
     fileSize = UTIL_getFileSize(fileName);
-    if (fileSize > DICTSIZE_MAX) {
-        EXM_THROW(32, "Dictionary file %s is too large (> %u MB)",
-                        fileName, DICTSIZE_MAX >> 20);   /* avoid extreme cases */
+    {
+        size_t dictSizeMax = prefs->patchFromMode ? prefs->memLimit : DICTSIZE_MAX;
```
Cyan4973: nit: `size_t const`

programs/fileio.c
```diff
@@ -759,9 +768,16 @@ typedef struct {
     ZSTD_CStream* cctx;
 } cRess_t;
 
+static unsigned int FIO_log2(const size_t x)
```
Cyan4973: note : `fileio.c` already defines `FIO_highbit64()`, which does the same,
except that input type is `unsigned long long` instead of `size_t`.

Recommendation : either use existing one,
or create another one due to input type difference, but ensure naming consistency.

bimbashrestha: Ah okay, didn't know about that function. I'll just use that one.

programs/fileio.c
```diff
@@ -624,7 +630,7 @@ FIO_openDstFile(FIO_prefs_t* const prefs,
  * @return : loaded size
  *  if fileName==NULL, returns 0 and a NULL pointer
  */
-static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName)
+static size_t FIO_createDictBuffer(FIO_prefs_t* const prefs, void** bufferPtr, const char* fileName)
```
Cyan4973: minor : API : argument order

Prefer grouping the _mutable states_ at the beginning.

In this case, `bufferPtr` is the actual of this function (it will be modified), while `prefs` is merely a parameter (it will only be read).

programs/fileio.c
Cyan4973: minor : it seems you reverted this declaration.

Cyan4973:
Almost there.
There are a few comments left, but only on coding style, so this is mostly minor and straightforward.

Also : tests show that it works, but only for files of "reasonable" size (a few MB).
So the next stage (`--long` mode) is necessary to properly advertise this new capability.

programs/fileio.c
```diff
@@ -638,9 +644,12 @@ static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName)
     if (fileHandle==NULL) EXM_THROW(31, "%s: %s", fileName, strerror(errno));
 
     fileSize = UTIL_getFileSize(fileName);
-    if (fileSize > DICTSIZE_MAX) {
-        EXM_THROW(32, "Dictionary file %s is too large (> %u MB)",
-                        fileName, DICTSIZE_MAX >> 20);   /* avoid extreme cases */
+    {
+        size_t const dictSizeMax = prefs->patchFromMode ? prefs->memLimit : DICTSIZE_MAX;
```
Cyan4973: I guess this is temporary, and will be changed once the `--long` mode gets integrated ?

bimbashrestha: Yes

vrubleg:
Very interesting use case.

@Cyan4973 before it is released, may be it is worth to consider adding a new frame to "patch" files with checksum of the original file which was used as a dictionary?

Cyan4973:
We could use a skippable frame to transport additional metadata.
This is a bit early though, the functionality still needs to be improved.

PR #1944, opened 2020-01-08
[fuzz] Fix regression_driver.c with directory input

terrelln:
The `numFiles` variable wasn't updated, so the fuzzer only tested one input in the directory. I did two things to fix this:

1. Remove the `numFiles` variable entirely.
2. Error if we can't open a file and print the number of files tested.

### DIFF
--- a/tests/fuzz/regression_driver.c
+++ b/tests/fuzz/regression_driver.c
@@ -20,23 +20,28 @@ int main(int argc, char const **argv) {
   int const kFollowLinks = 1;
   FileNamesTable* files;
   const char** const fnTable = argv + 1;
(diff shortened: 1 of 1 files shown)


Cyan4973:
> Remove the numFiles variable entirely.

nit-picking here, just I was surprised to see `numFiles` in the source, 
albeit with a shorter lifetime.

I guess what you mean is that you are now using `files->tableSize` directly as the bound of the loop, instead of `numFiles`, because it does not automatically follow updates in `files`.

terrelln:
> nit-picking here, just I was surprised to see numFiles in the source, albeit with a shorter lifetime.

I'm just using it as the input for the 2 file name table invocations, so as not to repeat `argc - 1`.

PR #1948, opened 2020-01-11
Fix the c99/cmake build under Cygwin/MSYS2

lazka:
When building zst under cygwin or msys2 with std=c99 the build would fail because
of an undefined fileno()/_fileno(), which is used by the IS_CONSOLE() macro.

When building with -std=c99 (gcc otherwise defaults to gnu, which implies POSIX),
which is the default of the cmake build, then including unistd.h wont define
_POSIX_VERSION and all other headers also wont expose POSIX API.

To fix this make sure to define _POSIX_C_SOURCE with the version we want before including
unistd.h and so that _POSIX_VERSION is set to the version provided by the system.

Since Cygwin/MSYS2 just follow POSIX we can also remove their special cases for
defining IS_CONSOLE().

And, for completeness, also explicitly include stdio.h which is what actually declares fileno().

Tested with the normal make file and cmake under MSYS2 and Cygwin.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -188,6 +188,9 @@
   environment:
     matrix:
     - COMPILER: "gcc"
+      HOST:     "cygwin"
+      PLATFORM: "x64"
+    - COMPILER: "gcc"
       HOST:     "mingw"
       PLATFORM: "x64"
       SCRIPT:   "CPPFLAGS=-DDEBUGLEVEL=2 CFLAGS=-Werror make -j allzstd DEBUGLEVEL=2"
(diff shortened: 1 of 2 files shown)


lazka:
IMO _POSIX_C_SOURCE should always be set before including unistd.h, but I wanted to keep this change minimal and only affect Cygwin.

terrelln:
Looks good to me, but I don't know Windows stuff at all.

Do you know if it is possible to run the cmake Cygwin build on Appveyor? If this is something you know how to do, would you be willing to add it to Appveyor? If we aren't building Cygwin in CI it is likely to break again.

lazka:
Sure, I'll have a look.

lazka:
I've added something, feedback welcome

terrelln:
Awesome! I verified that the new Appveyor test fails without your patch. It adds 1.5 minutes to the CI time, which is acceptable.

Thanks so much for the patch and especially adding the CI! The only thing we're missing is running the `make check` script on the cmake build. Ideally we should set it up in cmake's testing framework, so I'll open an issue for that.

lazka:
Thanks! Also for fixing the tests!

PR #1954, opened 2020-01-16
[refactor] Centralize match finding logic inside lazy 

bimbashrestha:
In preparation for new dictionary search structure. I don't think I nee to refactor opt since there is only one place where dictionary searching happens.

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -148,7 +148,6 @@ ZSTD_insertDUBT1(ZSTD_matchState_t* ms,
     *smallerPtr = *largerPtr = 0;
 }
 
-
 static size_t
 ZSTD_DUBT_findBetterDictMatch (
         ZSTD_matchState_t* ms,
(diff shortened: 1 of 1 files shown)


terrelln:
Can you include the perf impact in the PR summary please?

bimbashrestha:
> Can you include the perf impact in the PR summary please?

Yep. benchmarks running:)

Cyan4973:
Looks good to me.
Benchmarks welcomed,
also a few remaining minor conversion warnings to fix.

bimbashrestha:
> Benchmarks welcomed,

I ran benchmarks for the wrong commit:( Spinning up new ones now. I didn't see much change (+-1%) with the following but will confirm: 

```
tests/datagen -g10000 -P50 > dict; tests/datagen -g10000 -P10 > data
./zstd -b7 -e15 -D dict data
```

terrelln:
datagen isn't the right dataset to use for benchmarking. I recommend using silesia and enwik.

Cyan4973:
just a reminder: complete the topic of benchmark

bimbashrestha:
After some longer benchmarks, it looks like there is a ~1-2% dictionary compression regression. I tested silesia, using the first half of the data as a dictionary for the second half.

bimbashrestha:
I think that's too much of a regression for this change but I'm really sure how to improve that either. I think I'll close this PR for now and we can do further benchmarking (with this refactor and without) when I get a prototype of the dictionary search structure up.

Cyan4973:
I'll complete some benchmarks on my side too, using this PR.

Cyan4973:
I made a few performance tests, comparing this branch to `facebook/dev`, compiled with `gcc` `v9.2.1` on a Linux Ubuntu station.

Good news : when there is no dictionary involved, performance seems preserved. I could not find a reproducible difference higher than noise threshold.

Bad news : when dictionary is involved, compression speed is impacted, by a significant and reproducible margin.
Tested using the `github_users` corpus and a dictionary produced by the trainer : 

| level | `dev` | this branch | diff |
| --- | --- | --- | --- |
| 4 | 770 | 770 |  _dfast_ (unaffected) |
| 5 | 280 | 230 | -21% |
| 6 | 156 | 140 | -11% |
| 7 | 123 | 113 | -8% |
| 8 | 98 | 92 | -6% |
| 9 | 74 | 71 | -4% |
| 10 |  50 | 50 | (unaffected) |

The impact starts at _greedy_ strategy, and reduces with higher compression level, which is expected since other costs tend to become dominant.

The measured performance difference is indeed way too large, and worth an investigation before considering a merge.

PR #1956, opened 2020-01-17
Add vcpkg installation instructions

PhoebeHui:
zstd is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build zstd , ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for zstd and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,11 @@ example about how Meson is used to build this project.
 
 Note that default build type is **release**.
 
+### VCPKG
+If you are using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager,
+you can build and install zstd with CMake integration in a single command:
+- `vcpkg install zstd`
+
 ### Visual Studio (Windows)
 
 Going into `build` directory, you will find additional possibilities:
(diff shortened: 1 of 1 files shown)


felixhandte:
Actually, as I understand it, if someone uses `vcpkg` to install Zstd, they are pulling a version of the source code from some repository Microsoft maintains, that may have patches on top of vanilla zstd, or may lag behind, or whatever. I see that lz4/lz4#825, for example, calls that out. It would be nice to have similar language here clarifying that this installation method, unlike the others in this section, does not necessarily give you a build of `facebook/zstd:master`.

@PhoebeHui, does that make sense? Would you mind adding something along those lines?

PhoebeHui:
@felixhandte, thanks for your reviewing!

I have updated the comments.

PR #1958, opened 2020-01-17
Add --single-thread to zstd --help

terrelln:
It's already present in the man page, just missing from the help command.


Cyan4973:
Should it be mentioned that resulting compressed size can be slightly different from default `-T1` mode ?

PR #1959, opened 2020-01-18
Adding --long support for --patch-from

bimbashrestha:
**Patch From**
Zstandard is introducing a new command line option —patch-from= which leverages our existing compressors, dictionaries and the long range match finder to deliver a high speed engine for producing and applying patches to files.

Patch from increases the previous maximum limit for dictionaries from 32 MB to 4 GB. Additionally, it maintains fast speeds on lower compression levels without compromising patch size by using the long range match finder (now extended to find dictionary matches). By default, Zstandard uses a heuristic based on file size and internal compression parameters to determine when to activate long mode but it can also be manually specified as before. 

Patch from also works with multi-threading mode at a minimal compression ratio cost loss single threaded mode. 

**Example usage:**
```
# create the patch
zstd —patch-from=<oldfile> <newfile> <patchfile>

# apply the patch
zstd -d —patch-from=<oldfile> <patchfile> <newfile>
``` 

**Benchmarks:** 
We compared zstd to bsdiff, a popular industry grade diff engine. Our testing data were tarballs of different versions of source code from popular GitHub repositories. Specifically 

```
repos = {
    # ~31mb (small file)
    "zstd": {"url": "https://github.com/facebook/zstd", "dict-branch": "refs/tags/v1.4.2", "src-branch": "refs/tags/v1.4.3"},
    # ~273mb (medium file)
    "wordpress": {"url": "https://github.com/WordPress/WordPress", "dict-branch": "refs/tags/5.3.1", "src-branch": "refs/tags/5.3.2"},
    # ~1.66gb (large file)
    "llvm": {"url": "https://github.com/llvm/llvm-project", "dict-branch": "refs/tags/llvmorg-9.0.0", "src-branch": "refs/tags/llvmorg-9.0.1"}
}
```
alt text
Patch from on level 19 (with chainLog=30 and targetLength=4kb) remains competitive with bsdiff when comparing patch sizes.  
alt text
And patch from greatly outperforms bsdiff in speed even on its slowest setting of level 19 boasting an average speedup of ~7X. Patch from is >200X faster on level 1 and >100X faster on level 3 vs bsdiff while still delivering patch sizes less than 0.5% of the original file size.

And of course, there is no change to the fast zstd decompression speed.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -945,9 +945,11 @@ size_t ZSTD_CCtx_refPrefix_advanced(
 {
     RETURN_ERROR_IF(cctx->streamStage != zcss_init, stage_wrong);
     ZSTD_clearAllDicts(cctx);
-    cctx->prefixDict.dict = prefix;
-    cctx->prefixDict.dictSize = prefixSize;
-    cctx->prefixDict.dictContentType = dictContentType;
+    if (prefix != NULL && prefixSize > 0) {
+        cctx->prefixDict.dict = prefix;
+        cctx->prefixDict.dictSize = prefixSize;
+        cctx->prefixDict.dictContentType = dictContentType;
+    }
     return 0;
 }
 
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2782,6 +2782,7 @@ size_t ZSTD_compressBlock(ZSTD_CCtx* cctx, void* dst, size_t dstCapacity, const 
  *  @return : 0, or an error code
  */
 static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
+                                         ldmState_t* ls,
                                          ZSTD_cwksp* ws,
                                          ZSTD_CCtx_params const* params,
                                          const void* src, size_t srcSize,
@@ -2792,6 +2793,11 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
 
     ZSTD_window_update(&ms->window, src, srcSize);
     ms->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ms->window.base);
+
+    if (params->ldmParams.enableLdm && ls != NULL) {
+        ZSTD_window_update(&ls->window, src, srcSize);
+        ls->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ls->window.base);
+    }
 
     /* Assert that we the ms params match the params we're being given */
     ZSTD_assertEqualCParams(params->cParams, ms->cParams);
@@ -2804,6 +2810,11 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
         const BYTE* const ichunk = ip + chunk;
 
         ZSTD_overflowCorrectIfNeeded(ms, ws, params, ip, ichunk);
+
+        if (params->ldmParams.enableLdm && ls != NULL) {
+            ZSTD_ldm_fillHashTable(ls, (const BYTE*)src, (const BYTE*)src + srcSize, &params->ldmParams);
+            break;
+        }
 
         switch(params->cParams.strategy)
         {
@@ -2983,7 +2994,7 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
         bs->entropy.fse.matchlength_repeatMode = FSE_repeat_valid;
         bs->entropy.fse.litlength_repeatMode = FSE_repeat_valid;
         FORWARD_IF_ERROR(ZSTD_loadDictionaryContent(
-            ms, ws, params, dictPtr, dictContentSize, dtlm));
+            ms, NULL, ws, params, dictPtr, dictContentSize, dtlm));
         return dictID;
     }
 }
@@ -2993,6 +3004,7 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
 static size_t
 ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
                                ZSTD_matchState_t* ms,
+                               ldmState_t* ls,
                                ZSTD_cwksp* ws,
                          const ZSTD_CCtx_params* params,
                          const void* dict, size_t dictSize,
@@ -3010,13 +3022,13 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
 
     /* dict restricted modes */
     if (dictContentType == ZSTD_dct_rawContent)
-        return ZSTD_loadDictionaryContent(ms, ws, params, dict, dictSize, dtlm);
+        return ZSTD_loadDictionaryContent(ms, ls, ws, params, dict, dictSize, dtlm);
 
     if (MEM_readLE32(dict) != ZSTD_MAGIC_DICTIONARY) {
         if (dictContentType == ZSTD_dct_auto) {
             DEBUGLOG(4, "raw content dictionary detected");
             return ZSTD_loadDictionaryContent(
-                ms, ws, params, dict, dictSize, dtlm);
+                ms, ls, ws, params, dict, dictSize, dtlm);
         }
         RETURN_ERROR_IF(dictContentType == ZSTD_dct_fullDict, dictionary_wrong);
         assert(0);   /* impossible */
@@ -3059,12 +3071,12 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
     {   size_t const dictID = cdict ?
                 ZSTD_compress_insertDictionary(
                         cctx->blockState.prevCBlock, &cctx->blockState.matchState,
-                        &cctx->workspace, &cctx->appliedParams, cdict->dictContent,
+                        &cctx->ldmState, &cctx->workspace, &cctx->appliedParams, cdict->dictContent,
                         cdict->dictContentSize, dictContentType, dtlm,
                         cctx->entropyWorkspace)
               : ZSTD_compress_insertDictionary(
                         cctx->blockState.prevCBlock, &cctx->blockState.matchState,
-                        &cctx->workspace, &cctx->appliedParams, dict, dictSize,
+                        &cctx->ldmState, &cctx->workspace, &cctx->appliedParams, dict, dictSize,
                         dictContentType, dtlm, cctx->entropyWorkspace);
         FORWARD_IF_ERROR(dictID);
         assert(dictID <= UINT_MAX);
@@ -3342,7 +3354,7 @@ static size_t ZSTD_initCDict_internal(
         params.fParams.contentSizeFlag = 1;
         params.cParams = cParams;
         {   size_t const dictID = ZSTD_compress_insertDictionary(
-                    &cdict->cBlockState, &cdict->matchState, &cdict->workspace,
+                    &cdict->cBlockState, &cdict->matchState, NULL, &cdict->workspace,
                     &params, cdict->dictContent, cdict->dictContentSize,
                     dictContentType, ZSTD_dtlm_full, cdict->entropyWorkspace);
             FORWARD_IF_ERROR(dictID);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2813,10 +2813,8 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
 
         ZSTD_overflowCorrectIfNeeded(ms, ws, params, ip, ichunk);
 
-        if (params->ldmParams.enableLdm && ls != NULL && srcSize >= params->ldmParams.minMatchLength) {
+        if (params->ldmParams.enableLdm && ls != NULL && srcSize >= params->ldmParams.minMatchLength)
             ZSTD_ldm_fillHashTable(ls, (const BYTE*)src, (const BYTE*)src + srcSize, &params->ldmParams);
-            break;
-        }
 
         switch(params->cParams.strategy)
         {
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2813,7 +2813,7 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
 
         ZSTD_overflowCorrectIfNeeded(ms, ws, params, ip, ichunk);
 
-        if (params->ldmParams.enableLdm && ls != NULL) {
+        if (params->ldmParams.enableLdm && ls != NULL && srcSize >= params->ldmParams.minMatchLength) {
             ZSTD_ldm_fillHashTable(ls, (const BYTE*)src, (const BYTE*)src + srcSize, &params->ldmParams);
             break;
         }
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1011,7 +1011,7 @@ ZSTD_clampCParams(ZSTD_compressionParameters cParams)
 
 /** ZSTD_cycleLog() :
  *  condition for correct operation : hashLog > 1 */
-static U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
+U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
 {
     U32 const btScale = ((U32)strat >= (U32)ZSTD_btlazy2);
     return hashLog - btScale;
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -224,6 +224,17 @@ static U64 ZSTD_ldm_fillLdmHashTable(ldmState_t* state,
     return rollingHash;
 }
 
+void ZSTD_ldm_fillHashTable(
+            ldmState_t* state, const BYTE* ip,
+            const BYTE* iend, ldmParams_t const* params)
+{
+    U64 startingHash = ZSTD_rollingHash_compute(ip, params->minMatchLength);
+    ZSTD_ldm_fillLdmHashTable(
+        state, startingHash, ip, iend - params->minMatchLength, state->window.base,
+        params->hashLog - params->bucketSizeLog,
+        *params);
+}
+
 
 /** ZSTD_ldm_limitTableUpdate() :
  *
@@ -459,7 +470,7 @@ size_t ZSTD_ldm_generateSequences(
          *       * Try invalidation after the sequence generation and test the
          *         the offset against maxDist directly.
          */
-        ZSTD_window_enforceMaxDist(&ldmState->window, chunkEnd, maxDist, NULL, NULL);
+        ZSTD_window_enforceMaxDist(&ldmState->window, chunkEnd, maxDist, &ldmState->loadedDictEnd, NULL);
         /* 3. Generate the sequences for the chunk, and get newLeftoverSize. */
         newLeftoverSize = ZSTD_ldm_generateSequences_internal(
             ldmState, sequences, params, chunkStart, chunkSize);
@@ -567,7 +578,6 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
         if (sequence.offset == 0)
             break;
 
-        assert(sequence.offset <= (1U << cParams->windowLog));
         assert(ip + sequence.litLength + sequence.matchLength <= iend);
 
         /* Fill tables for block compressor */
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -461,34 +461,36 @@ typedef struct {
     ZSTD_window_t ldmWindow;  /* A thread-safe copy of ldmState.window */
 } serialState_t;
 
-static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* seqPool, ZSTD_CCtx_params params, size_t jobSize)
+static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* seqPool, 
+                        ZSTD_CCtx_params* paramsPtr, size_t jobSize, const void* dict, 
+                        const size_t dictSize)
 {
     /* Adjust parameters */
-    if (params.ldmParams.enableLdm) {
-        DEBUGLOG(4, "LDM window size = %u KB", (1U << params.cParams.windowLog) >> 10);
-        ZSTD_ldm_adjustParameters(&params.ldmParams, &params.cParams);
-        assert(params.ldmParams.hashLog >= params.ldmParams.bucketSizeLog);
-        assert(params.ldmParams.hashRateLog < 32);
+    if (paramsPtr->ldmParams.enableLdm) {
+        DEBUGLOG(4, "LDM window size = %u KB", (1U << paramsPtr->cParams.windowLog) >> 10);
+        ZSTD_ldm_adjustParameters(&paramsPtr->ldmParams, &paramsPtr->cParams);
+        assert(paramsPtr->ldmParams.hashLog >= paramsPtr->ldmParams.bucketSizeLog);
+        assert(paramsPtr->ldmParams.hashRateLog < 32);
         serialState->ldmState.hashPower =
-                ZSTD_rollingHash_primePower(params.ldmParams.minMatchLength);
+                ZSTD_rollingHash_primePower(paramsPtr->ldmParams.minMatchLength);
     } else {
-        memset(&params.ldmParams, 0, sizeof(params.ldmParams));
+        memset(&paramsPtr->ldmParams, 0, sizeof(paramsPtr->ldmParams));
     }
     serialState->nextJobID = 0;
-    if (params.fParams.checksumFlag)
+    if (paramsPtr->fParams.checksumFlag)
         XXH64_reset(&serialState->xxhState, 0);
-    if (params.ldmParams.enableLdm) {
-        ZSTD_customMem cMem = params.customMem;
-        unsigned const hashLog = params.ldmParams.hashLog;
+    if (paramsPtr->ldmParams.enableLdm) {
+        ZSTD_customMem cMem = paramsPtr->customMem;
+        unsigned const hashLog = paramsPtr->ldmParams.hashLog;
         size_t const hashSize = ((size_t)1 << hashLog) * sizeof(ldmEntry_t);
         unsigned const bucketLog =
-            params.ldmParams.hashLog - params.ldmParams.bucketSizeLog;
+            paramsPtr->ldmParams.hashLog - paramsPtr->ldmParams.bucketSizeLog;
         size_t const bucketSize = (size_t)1 << bucketLog;
         unsigned const prevBucketLog =
             serialState->params.ldmParams.hashLog -
             serialState->params.ldmParams.bucketSizeLog;
         /* Size the seq pool tables */
-        ZSTDMT_setNbSeq(seqPool, ZSTD_ldm_getMaxNbSeq(params.ldmParams, jobSize));
+        ZSTDMT_setNbSeq(seqPool, ZSTD_ldm_getMaxNbSeq(paramsPtr->ldmParams, jobSize));
         /* Reset the window */
         ZSTD_window_init(&serialState->ldmState.window);
         serialState->ldmWindow = serialState->ldmState.window;
@@ -507,8 +509,14 @@ static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* 
         memset(serialState->ldmState.hashTable, 0, hashSize);
         memset(serialState->ldmState.bucketOffsets, 0, bucketSize);
     }
-    serialState->params = params;
+    serialState->params = *paramsPtr;
     serialState->params.jobSize = (U32)jobSize;
+
+    /* Update window state and fill hash table with dict */
+    if (paramsPtr->ldmParams.enableLdm && dict) {
+        ZSTD_window_update(&serialState->ldmState.window, dict, dictSize);
+        ZSTD_ldm_fillHashTable(&serialState->ldmState, (const BYTE*)dict, (const BYTE*)dict + dictSize, &paramsPtr->ldmParams);
+    }
     return 0;
 }
 
@@ -1267,7 +1275,7 @@ static size_t ZSTDMT_compress_advanced_internal(
 
     assert(avgJobSize >= 256 KB);  /* condition for ZSTD_compressBound(A) + ZSTD_compressBound(B) <= ZSTD_compressBound(A+B), required to compress directly into Dst (no additional buffer) */
     ZSTDMT_setBufferSize(mtctx->bufPool, ZSTD_compressBound(avgJobSize) );
-    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, params, avgJobSize))
+    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, &params, avgJobSize, NULL, 0))
         return ERROR(memory_allocation);
 
     FORWARD_IF_ERROR( ZSTDMT_expandJobsTable(mtctx, nbJobs) );  /* only expands if necessary */
@@ -1500,7 +1508,7 @@ size_t ZSTDMT_initCStream_internal(
     mtctx->allJobsCompleted = 0;
     mtctx->consumed = 0;
     mtctx->produced = 0;
-    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, params, mtctx->targetSectionSize))
+    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, &params, mtctx->targetSectionSize, dict, dictSize))
         return ERROR(memory_allocation);
     return 0;
 }
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -461,36 +461,36 @@ typedef struct {
     ZSTD_window_t ldmWindow;  /* A thread-safe copy of ldmState.window */
 } serialState_t;
 
-static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* seqPool, 
-                        ZSTD_CCtx_params* paramsPtr, size_t jobSize, const void* dict, 
-                        const size_t dictSize)
+static int ZSTDMT_serialState_reset(serialState_t* serialState, 
+                        ZSTDMT_seqPool* seqPool, ZSTD_CCtx_params params, 
+                        size_t jobSize, const void* dict, size_t const dictSize)
 {
     /* Adjust parameters */
-    if (paramsPtr->ldmParams.enableLdm) {
-        DEBUGLOG(4, "LDM window size = %u KB", (1U << paramsPtr->cParams.windowLog) >> 10);
-        ZSTD_ldm_adjustParameters(&paramsPtr->ldmParams, &paramsPtr->cParams);
-        assert(paramsPtr->ldmParams.hashLog >= paramsPtr->ldmParams.bucketSizeLog);
-        assert(paramsPtr->ldmParams.hashRateLog < 32);
+    if (params.ldmParams.enableLdm) {
+        DEBUGLOG(4, "LDM window size = %u KB", (1U << params.cParams.windowLog) >> 10);
+        ZSTD_ldm_adjustParameters(&params.ldmParams, &params.cParams);
+        assert(params.ldmParams.hashLog >= params.ldmParams.bucketSizeLog);
+        assert(params.ldmParams.hashRateLog < 32);
         serialState->ldmState.hashPower =
-                ZSTD_rollingHash_primePower(paramsPtr->ldmParams.minMatchLength);
+                ZSTD_rollingHash_primePower(params.ldmParams.minMatchLength);
     } else {
-        memset(&paramsPtr->ldmParams, 0, sizeof(paramsPtr->ldmParams));
+        memset(&params.ldmParams, 0, sizeof(params.ldmParams));
     }
     serialState->nextJobID = 0;
-    if (paramsPtr->fParams.checksumFlag)
+    if (params.fParams.checksumFlag)
         XXH64_reset(&serialState->xxhState, 0);
-    if (paramsPtr->ldmParams.enableLdm) {
-        ZSTD_customMem cMem = paramsPtr->customMem;
-        unsigned const hashLog = paramsPtr->ldmParams.hashLog;
+    if (params.ldmParams.enableLdm) {
+        ZSTD_customMem cMem = params.customMem;
+        unsigned const hashLog = params.ldmParams.hashLog;
         size_t const hashSize = ((size_t)1 << hashLog) * sizeof(ldmEntry_t);
         unsigned const bucketLog =
-            paramsPtr->ldmParams.hashLog - paramsPtr->ldmParams.bucketSizeLog;
+            params.ldmParams.hashLog - params.ldmParams.bucketSizeLog;
         size_t const bucketSize = (size_t)1 << bucketLog;
         unsigned const prevBucketLog =
             serialState->params.ldmParams.hashLog -
             serialState->params.ldmParams.bucketSizeLog;
         /* Size the seq pool tables */
-        ZSTDMT_setNbSeq(seqPool, ZSTD_ldm_getMaxNbSeq(paramsPtr->ldmParams, jobSize));
+        ZSTDMT_setNbSeq(seqPool, ZSTD_ldm_getMaxNbSeq(params.ldmParams, jobSize));
         /* Reset the window */
         ZSTD_window_init(&serialState->ldmState.window);
         serialState->ldmWindow = serialState->ldmState.window;
@@ -509,14 +509,15 @@ static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* 
         memset(serialState->ldmState.hashTable, 0, hashSize);
         memset(serialState->ldmState.bucketOffsets, 0, bucketSize);
     }
-    serialState->params = *paramsPtr;
-    serialState->params.jobSize = (U32)jobSize;
 
     /* Update window state and fill hash table with dict */
-    if (paramsPtr->ldmParams.enableLdm && dict) {
+    if (params.ldmParams.enableLdm && dict) {
         ZSTD_window_update(&serialState->ldmState.window, dict, dictSize);
-        ZSTD_ldm_fillHashTable(&serialState->ldmState, (const BYTE*)dict, (const BYTE*)dict + dictSize, &paramsPtr->ldmParams);
+        ZSTD_ldm_fillHashTable(&serialState->ldmState, (const BYTE*)dict, (const BYTE*)dict + dictSize, &params.ldmParams);
     }
+
+    serialState->params = params;
+    serialState->params.jobSize = (U32)jobSize;
     return 0;
 }
 
@@ -1275,7 +1276,7 @@ static size_t ZSTDMT_compress_advanced_internal(
 
     assert(avgJobSize >= 256 KB);  /* condition for ZSTD_compressBound(A) + ZSTD_compressBound(B) <= ZSTD_compressBound(A+B), required to compress directly into Dst (no additional buffer) */
     ZSTDMT_setBufferSize(mtctx->bufPool, ZSTD_compressBound(avgJobSize) );
-    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, &params, avgJobSize, NULL, 0))
+    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, params, avgJobSize, NULL, 0))
         return ERROR(memory_allocation);
 
     FORWARD_IF_ERROR( ZSTDMT_expandJobsTable(mtctx, nbJobs) );  /* only expands if necessary */
@@ -1508,7 +1509,7 @@ size_t ZSTDMT_initCStream_internal(
     mtctx->allJobsCompleted = 0;
     mtctx->consumed = 0;
     mtctx->produced = 0;
-    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, &params, mtctx->targetSectionSize, dict, dictSize))
+    if (ZSTDMT_serialState_reset(&mtctx->serial, mtctx->seqPool, params, mtctx->targetSectionSize, dict, dictSize))
         return ERROR(memory_allocation);
     return 0;
 }
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -807,7 +807,7 @@ static void FIO_adjustParamsForPatchFromMode(FIO_prefs_t* const prefs,
                                     int cLevel)
 {
     unsigned const fileWindowLog = FIO_highbit64(maxSrcFileSize) + 1;
-    ZSTD_compressionParameters const cParams = ZSTD_getCParams(cLevel, maxSrcFileSize, dictSize);
+    ZSTD_compressionParameters const cParams = ZSTD_getCParams(cLevel, (size_t)maxSrcFileSize, (size_t)dictSize);
     FIO_adjustMemLimitForPatchFromMode(prefs, dictSize, maxSrcFileSize);
     if (fileWindowLog > ZSTD_WINDOWLOG_MAX)
         DISPLAYLEVEL(1, "Max window log exceeded by file (compression ratio will suffer)\n");
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -814,6 +814,7 @@ static void FIO_adjustParamsForPatchFromMode(FIO_prefs_t* const prefs,
     }
     if (cParams.strategy >= ZSTD_btopt) {
         DISPLAYLEVEL(1, "[Optimal parser notes] Consider the following to improve patch size at the cost of speed:\n");
+		DISPLAYLEVEL(1, "- Use --single-thread mode in the zstd cli");
         DISPLAYLEVEL(1, "- Set a larger targetLength (eg. --zstd=targetLength=4096)\n");
         DISPLAYLEVEL(1, "- Set a larger chainLog (eg. --zstd=chainLog=31)\n");
         DISPLAYLEVEL(1, "Also consdier playing around with searchLog and hashLog\n");
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -799,12 +799,13 @@ typedef struct {
 
 static void FIO_adjustParamsForPatchFromMode(FIO_prefs_t* const prefs, 
                                     ZSTD_compressionParameters* comprParams,
-                                    size_t const dictSize, size_t const maxSrcFileSize,
+                                    unsigned long long const dictSize, 
+                                    unsigned long long const maxSrcFileSize,
                                     int cLevel)
 {
-    unsigned const fileWindowLog = FIO_highbit64((unsigned long long)maxSrcFileSize) + 1;
+    unsigned const fileWindowLog = FIO_highbit64(maxSrcFileSize) + 1;
     ZSTD_compressionParameters const cParams = ZSTD_getCParams(cLevel, maxSrcFileSize, dictSize);
-    FIO_adjustMemLimitForPatchFromMode(prefs, (unsigned long long)dictSize, (unsigned long long)maxSrcFileSize);
+    FIO_adjustMemLimitForPatchFromMode(prefs, dictSize, maxSrcFileSize);
     if (fileWindowLog > ZSTD_WINDOWLOG_MAX)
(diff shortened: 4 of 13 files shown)


bimbashrestha:
Tested this using tar of 4 clang versions: 
Before
```
head -c 210000000 clangs > dict && head -c 200000000 clangs > data && ./zstd -f --memory=210000000 --single-thread --patch-from=dict data
data                 : 16.14%   (200000000 => 32275377 bytes, data.zst)
```
After 
```
head -c 210000000 clangs > dict && head -c 200000000 clangs > data && ./zstd --long -f --memory=210000000 --single-thread --patch-from=dict data
data                 :  7.77%   (200000000 => 15542116 bytes, data.zst)
```

Cyan4973:
I presume that, in the best case, the difference is only the last 10.000.000 bytes. 
And since `data` is actually` smaller than `dict`, it should consists entirely of references into `dict`.

We can't avoid spending a few bytes per block, at least to indicate one large match,
but such a budget should be < 30 bytes per block. That's still much less than the 15MB in the test.

Hence, there is ground for improvement.

One first question :
`--long`, by default, set the `windowLog` to `27`, 128 MB.
Since the size of `data` to compress is 200.000.000 bytes, it requires a larger `windowLog`, of `28`.
I presume the `--patch-from=FILE` command does this adaptation automatically ?

bimbashrestha:
> I presume the --patch-from=FILE command does this adaptation automatically ?

Yes, see: https://github.com/bimbashrestha/zstd/blob/a2917fe814a3c9f41b1723de8a26faaff28bfd7f/programs/fileio.c#L821

Cyan4973:
OK,
let's spend some time on this scenario next week.
I think it's a simple one, it's relatively simple to define an "optimal" behavior for it,
and we should aim at getting it right, even if it demands new adaptations to work properly.

bimbashrestha:
Tried this with a more likely scenario. The input file is just the dictionary with 10% of its bytes perturbed by +-5 and the gains from --long are basically non-existent. 

Generating file for compression
```Python
import random
clangs = open("clangs", "rb").read(); clangs_str = clangs.decode('latin-1')

# randomly add int in range (-5, 5) inclusive to the original clangs on 10% of the bytes
clangs_mod_str = "".join(chr(ord(a)+random.randrange(-5,5)) if random.random() > 0.9 and ord(a)+5 < 256 and ord(a)-5 >= 0 else a for a in clangs_str)

clangs_mod = clangs_mod_str.encode('latin-1'); open("clangs_mod", "wb").write(clangs_mod)
```

With --long
```Bash
./zstd --single-thread -f --memory=300000000 --long --patch-from=clangs clangs_mod && ./zstd -f -d --memory=300000000 --patch-from=clangs clangs_mod.zst -o tmp && diff tmp clangs_mod
clangs_mod           : 32.04%   (288859136 => 92544650 bytes, clangs_mod.zst)
clangs_mod.zst       : 288859136 bytes
```

Without --long
```Bash
./old --single-thread -f --memory=300000000 --patch-from=clangs clangs_mod && ./old -f -d --memory=300000000 --patch-from=clangs clangs_mod.zst -o tmp && diff tmp clangs_mod
clangs_mod           : 32.07%   (288859136 => 92651457 bytes, clangs_mod.zst)
clangs_mod.zst       : 288859136 bytes
```

bimbashrestha:
It's better with less perturbations but still not super stellar. Here, the input file is just the dictionary with 1% of the bytes perturbed by 1 or -1. 

Generation: 
```Python
import random
clangs = open("clangs", "rb").read(); clangs_str = clangs.decode('latin-1')

# randomly add  1 or -1 to the original clangs on 1% of the bytes
clangs_mod_str = "".join(chr(ord(a)+random.choice([1, -1])) if random.random() < 0.01 and ord(a)+5 < 256 and ord(a)-5 >= 0 else a for a in clangs_str)

# check what percentage is different 
print(sum(a != b for a, b in zip(clangs_str, clangs_mod_str))/len(clangs_str))

clangs_mod = clangs_mod_str.encode('latin-1'); open("clangs_mod", "wb").write(clangs_mod)
```
With --long 
```
./zstd --single-thread -f --memory=300000000 --long --patch-from=clangs clangs_mod && ./zstd -f -d --memory=300000000 --patch-from=clangs clangs_mod.zst -o tmp && diff tmp clangs_mod
clangs_mod           : 9.26%   (288859136 => 92544650 bytes, clangs_mod.zst)
clangs_mod.zst       : 288859136 bytes
```
Without --long
```
./old --single-thread -f --memory=300000000 --patch-from=clangs clangs_mod && ./old -f -d --memory=300000000 --patch-from=clangs clangs_mod.zst -o tmp && diff tmp clangs_mod
clangs_mod           : 17.95%  (288859136 => 92651457 bytes, clangs_mod.zst)
clangs_mod.zst       : 288859136 bytes
```

Cyan4973:
These scenarios are actually _less_ representative of target use case.

In no scenario does it make sense to randomly change one byte here and there, especially at such density (yes, even 1/100 is a lot).

The expected scenario is that entire sections of reference will be preserved _unmodified_ into target. target differs from reference by having : entire sections removed, and/or entire sections added, in no particular order nor position. Of course, a few random changes may also be present here and there, but sparsely (if they are dense, it's equivalent to a complete section change).

With this definition, the very first scenario was actually closer to the goal.
But there is a twist : this particular scenario was built from aggregation of 4 similar tarball, meaning internal repetitions are very high. We'll see together that this is an issue.

bimbashrestha:
Hmm I see. Maybe a better scenario would be tar(clang x, clang x+2, clang x+4, clang x+6) as dict and tar(clang x+1, clang x+3, clang x+5, clang x+7) as data?

Cyan4973:
nope, still the same issue. A better scenario would be `clang x` as reference, and `clang x+1` as target.

bimbashrestha:
As seen in the commit above, I made a stupid mistake:( I was populating the ldm hash table with the src, not the dictionary. Just an FYI. The results still don't change much though (probably because the dictionary and src have been nearly identical in the tests above). I ran another with subsequent versions of clang using the default parameters from the commit above.
```
tar cvf data ~/Downloads/clangs/llvm-9.0.1.src
tar cvf dict ~/Downloads/clangs/llvm-9.0.0.src

# with --long
./zstd --single-thread -f --memory=500000000 --long --patch-from=dict data
data                 : 11.72%   (425910272 => 49898506 bytes, data.zst)

# without --long
./old --single-thread -f --memory=500000000 --patch-from=dict data
data                 : 11.72%   (425910272 => 49905559 bytes, data.zst)

# without dictionary 
./zstd -f --long data
data                 : 11.31%   (425910272 => 48162576 bytes, data.zst)
```

I think the implementation is doing what we would like it to now. Can someone give it a look and check to see if I'm missing something obvious?

Cyan4973:
> Forces --long when using --patch-from on dictionaries larger than 128 mb

`--long` will likely be needed way before that point.
But it's more complex, as the threshold depends on the compression parameters.

For example, at level 1, the hash table contains only 16K positions, so it will saturate very quickly (likely in the ~100KB range, in any case, way before 128 MB). `--long` will be able to compensate beyond saturation point.

However, this is a bit more complex to setup, as it now requires to find a rule depending on other compression parameters. This could be investigated later, in a dedicated optimization pass.

terrelln:
This looks about correct, but it isn't working so there must be something wrong. I think it is really close to working, and if you fix a few bugs it'll be there.

bimbashrestha:
Looks like there were a few things missing. The main thing wrong was that I was computing the rolling hash starting at src when I should have been computing it starting at window.base. I also needed to increase the window log (to contain both dictionary and src) and update the window inside ldmstate when loading the dictionary.

Anyway, the results are much better now.

```
tar cf llvm-9.0.1 ~/Downloads/clangs/llvm-9.0.1.src
tar cf llvm-9.0.0 ~/Downloads/clangs/llvm-9.0.0.src

./zstd --single-thread --memory=500000000 --long --patch-from=llvm-9.0.0 llvm-9.0.1 \
  && ./zstd -d --memory=500000000 --patch-from=llvm-9.0.0 llvm-9.0.1.zst -o tmp \
  && diff -s tmp llvm-9.0.1

llvm-9.0.1           :  0.48%   (425910272 => 2062253 bytes, llvm-9.0.1.zst)
llvm-9.0.1.zst       : 425910272 bytes
Files tmp and llvm-9.0.1 are identical
```

Cyan4973:
> I also needed to increase the window log (to contain both dictionary and src) 

FYI, according to the zstd specification, it's enough for the current position in `src` to be `< windowSize` in order to access the _entire_ dictionary.
In a silly case, you could have `srcSize == 16 KB` , therefore `windowLog == 14`, and it would still give access to the entire dictionary, even if it's `128 MB` long, and therefore could generate `offset` way larger than `windowSize`.

This part of the specification is well respected by the "regular" match finder and the decoder.

However, the LDM has not been combined with dictionary so far, therefore it may not be aware of this capability, and as a consequence, may not be able to reference matches beyond `windowSize`. In which case, a work-around is indeed necessary.

terrelln:
> However, the LDM has not been combined with dictionary so far, therefore it may not be aware of this capability, and as a consequence, may not be able to reference matches beyond windowSize. In which case, a work-around is indeed necessary.

We should be able to fix it by respecting `loadedDictEnd` in the LDM match finder. We shouldn't increase the window log to workaround this issue, since that will increase the decompression memory usage.

bimbashrestha:
TODO: make dictionary ldm respect loaddictendl like @terrelln mentioned above.

Also linking the original thread here because I forgot to before: https://github.com/facebook/zstd/issues/1935

bimbashrestha:
Okay, I think this is ready for a review now. The functionality is still the same after the last few commits: 

```
./zstd --single-thread --memory=500000000 --long -f --patch-from=llvm-9.0.0 llvm-9.0.1 \
  && ./zstd -d --memory=500000000 -f --patch-from=llvm-9.0.0 llvm-9.0.1.zst -o tmp \
  && diff -s tmp llvm-9.0.1

llvm-9.0.1           :  0.48%   (425910272 => 2062279 bytes, llvm-9.0.1.zst)
llvm-9.0.1.zst      : 425910272 bytes
Files tmp and llvm-9.0.1 are identical
```

Cyan4973:
Is `--single-thread` still required ?

bimbashrestha:
> Is `--single-thread` still required ?

Without it, the current implementation doesn't compress exceptionally well. See below. It's still better than not using long mode but hardly. I'll have to look into how long mode works together with multi-threading to uncover why this is. I was thinking that could be part of a different pr? 

```
# without single-thread
./zstd --memory=500000000 -f --long --patch-from=llvm-9.0.0 llvm-9.0.1 \
  && ./zstd -d --memory=500000000 -f --patch-from=llvm-9.0.0 llvm-9.0.1.zst -o tmp \
  && diff -s tmp llvm-9.0.1

llvm-9.0.1           : 11.30%   (425910272 => 48130136 bytes, llvm-9.0.1.zst)
llvm-9.0.1.zst      : 425910272 bytes
Files tmp and llvm-9.0.1 are identical

# --single-thread without --long
./zstd --memory=500000000 -f --single-thread --patch-from=llvm-9.0.0 llvm-9.0.1 \
  && ./zstd -d --memory=500000000 -f --patch-from=llvm-9.0.0 llvm-9.0.1.zst -o tmp \
  && diff -s tmp llvm-9.0.1

llvm-9.0.1           : 11.72%   (425910272 => 49905685 bytes, llvm-9.0.1.zst)
llvm-9.0.1.zst      : 425910272 bytes
Files tmp and llvm-9.0.1 are identical
```

Cyan4973:
> I was thinking that could be part of a different pr?

Okay, that's fine.
Just keep in mind : at the end of this serie of patches, the resulting command must be "simple" to understand and invoke from an end-user perspective.
`--patch-from=` is.
`--memory=# --long --single-thread --patch-from=` isn't.

All the additional machinery must be simplified.

Special exception if we have good reasons to believe that some consequences shouldn't be transparently opt-in, such as huge memory usage (and then, it should be correctly documented).

bimbashrestha:
> Just keep in mind : at the end of this serie of patches, the resulting command must be "simple" to understand and invoke from an end-user perspective.

Yes, I believe --memory should be manually specified. Although, I think we should bump the default max to something larger than it is now. Maybe 1gb? Which would get rid of it for most use cases right?

Maybe we can get rid of having the end user specify --long by automatically using long by default on large files (>128mb)? The user can then override this by setting --long=0 or something (although I don't think many will want to do this)

Also I could just automatically turn off multithreading on patch from and notify the user instead of requiring that they explicitly input --single-thread until i get multithreaded working?

With those changes, the cli should just be --patch-from=FILE for most use cases.

terrelln:
> Yes, I believe --memory should be manually specified. Although, I think we should bump the default max to something larger than it is now. Maybe 1gb? Which would get rid of it for most use cases right?

I think I would be okay having `--patch-from=X` imply `--memory=FileSize` and `--long` when the file size is known. I do not want to increase the default `--memory`. That controls the maximum memory usage of the decoder in the standard case. 128 MB is already quite big. When users are using `--patch-from=X` they are getting what they pay for if we increase the memory usage, and it only affects them. Though we should note it in the help/man page.

> Also I could just automatically turn off multithreading on patch from and notify the user instead of requiring that they explicitly input --single-thread until i get multithreaded working?

Multithreading support should be easy to add. You just need to load the dictionary into its ldmState in the same way you are doing it for the single-threaded one now. Until that diff is landed we can fail if `--single-thread` isn't passed to the CLI, since we don't expect it to make it into a release.

terrelln:
Alternatively, for files larger than 128 MB, we could print a warning saying to use `--memory=X` to ensure the dictionary is always in-bounds, but still compress the file with reduced compression ratio.

Cyan4973:
> I think I would be okay having --patch-from=X imply --memory=FileSize

I agree.
Maybe a (non-blocking) warning message (displayLevel 2) would also be worthwhile when the memory required becomes very large, such as > 128 MB.

> Maybe we can get rid of having the end user specify --long by automatically using long by default on large files (>128mb)? 

Yes, although part of the game will be to determine what's the right threshold to trigger it. 
I expect it to be far smaller than 128 MB. Any files substantially larger than `cycleLog` is at risk of missing pointers from the early section of the reference files.
I guess it will be useful to make tests with intermediate sizes, and determine when it's possible to _not_ use `--long` mode.

> Multithreading support should be easy to add.

Maybe @bimbashrestha will need some directions on this topic. You seem to have a pretty clear idea of what to do.

bimbashrestha:
Okay then, I made --patch-from=file imply --memory=filesize (that removes --memory) and I put 32 mb as the threshold for automatically activating long mode in patch-from (that removes --long) We can experiment with the value for the threshold later.

> Multithreading support should be easy to add. You just need to load the dictionary into its ldmState in the same way you are doing it for the single-threaded one now. Until that diff is landed we can fail if --single-thread 

Didn't realize that was all we needed to do. I've just went ahead and included the multi-threaded changes in this pr. The ratio in multi-threaded mode is a little worse than in single-threaded mode though. Is this unusual? 

```
# --single-thread
./zstd --single-thread --patch-from=llvm-9.0.0 llvm-9.0.1
llvm-9.0.1           :  0.48%   (425910272 => 2062279 bytes, llvm-9.0.1.zst)

# vs no --single-thread
./zstd --patch-from=llvm-9.0.0 llvm-9.0.1 \
  && ./zstd -d  --patch-from=llvm-9.0.0 llvm-9.0.1.zst -o tmp \
  && diff -s tmp llvm-9.0.1
llvm-9.0.1           :  0.55%   (425910272 => 2337712 bytes, llvm-9.0.1.zst)
llvm-9.0.1.zst      : 425910272 bytes
Files tmp and llvm-9.0.1 are identical
```

terrelln:
> Didn't realize that was all we needed to do. I've just went ahead and included the multi-threaded changes in this pr. The ratio in multi-threaded mode is a little worse than in single-threaded mode though. Is this unusual?

This is probably expected, but I can help you investigate this tomorrow for learning and to double check. Basically we'll want to look at the sequences the single threaded encoder and the multithreaded encoder generate, and compare them.

programs/fileio.c
```diff
@@ -801,17 +800,16 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
         EXM_THROW(31, "allocation error : not enough memory");
 
     /* Advanced parameters, including dictionary */
-    {   void* dictBuffer;
-        size_t const dictBuffSize = FIO_createDictBuffer(&dictBuffer, dictFileName, prefs);   /* works with dictFileName==NULL */
-        if (dictFileName && (dictBuffer==NULL))
+    {   if (dictFileName && (dictBuffer==NULL))
```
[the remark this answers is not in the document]
bimbashrestha: I needed to prevent dictbuffer from being freed until compression was done for refPrefix().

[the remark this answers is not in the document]
terrelln: We can instead add the `dictBuffer` to the `cRess_t`, so it lasts for the entire lifetime of the compression call.

programs/fileio.c
```diff
@@ -854,8 +852,11 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
         CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_c_rsyncable, prefs->rsyncable) );
 #endif
         /* dictionary */
-        CHECK( ZSTD_CCtx_loadDictionary(ress.cctx, dictBuffer, dictBuffSize) );
-        free(dictBuffer);
+        if (prefs->patchFromMode) {
+            CHECK( ZSTD_CCtx_refPrefix(ress.cctx, dictBuffer, dictBuffSize) );
+        } else {
+            CHECK( ZSTD_CCtx_loadDictionary(ress.cctx, dictBuffer, dictBuffSize) );
```
terrelln: If we put the `dictBuffer` in the `ress` then we can reference the dictionary here as well, instead of copying it.

programs/zstdcli.c
```diff
@@ -1197,15 +1199,22 @@ int main(int const argCount, const char* argv[])
...
         if (compressionParams.windowLog == 0) {
             memLimit = (U32)1 << g_defaultMaxWindowLog;
         } else {
             memLimit = (U32)1 << (compressionParams.windowLog & 31);
     }   }
+    if (patchFromDictFileName != NULL) {
+        const unsigned long long dictSize = UTIL_getFileSize(patchFromDictFileName);
+        if (dictSize != UTIL_FILESIZE_UNKNOWN) {
+            memLimit = dictSize > memLimit ? (unsigned)dictSize : memLimit;
+#ifndef ZSTD_NOCOMPRESS
```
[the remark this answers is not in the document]
bimbashrestha: Do you just mean the part inside the ifndef? Because we want to set memLimit to dictSize for decompression too right?

bimbashrestha:
Comparisons to bsdiff (https://github.com/mendsley/bsdiff) on consecutive versions of some popular github repositories.
```
ls top-github
bootstrap-4.4.0.tar   d3-5.15.0.tar         linux-5.6-rc1.tar     react-16.12.0.tar     vscode-1.39.1.tar     vue-2.6.11.tar
bootstrap-4.4.1.tar   flutter-1.15.2.tar    linux-5.6-rc2.tar     tensorflow-1.15.2.tar vscode-1.39.2.tar     zstd-1.4.3.tar
d3-5.14.2.tar         flutter-1.15.3.tar    react-16.11.0.tar     tensorflow-2.0.1.tar  vue-2.6.10.tar        zstd-1.4.4.tar
```
I'm only reporting the average ratio loss and average time win. Link to google doc with per repo breakdown: https://docs.google.com/spreadsheets/d/1E9xrdTdLV8zKnkPug5StS2MRbUWFFCrthcZ-0FElcEc/edit?usp=sharing

| zstd level | avg ratio loss (bsdiff/zstd) | avg time win (bsdiff/zstd) |
|------------|------------------------------|----------------------------|
| zstd-4     | 8X worse                     | 127X better                |
| zstd-7     | 10X worse                    | 116X better                |
| zstd-10    | 7X worse                     | 106X better                |
| zstd-13    | 6X worse                     | 69X better                 |
| zstd-16    | 5X worse                     | 45X better                 |
| zstd-19    | 4X worse                     | 30X better                 |

Cyan4973:
Let's schedule a meeting to discuss the next steps.

Also, please display the file sizes in the comparison, they matter a lot in this context, and give us a more precise clue regarding comparison with `bsdiff`.

lib/compress/zstd_compress.c
```diff
@@ -2793,6 +2794,13 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
     ZSTD_window_update(&ms->window, src, srcSize);
     ms->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ms->window.base);
 
+    if (params->ldmParams.enableLdm && ls != NULL) {
+        /* Update window inside ldm state */
+        ZSTD_window_update(&ls->window, src, srcSize);
+        /* TODO: Still need to make sure ldm dictionaries work well in multi-threading mode */
```
terrelln: Is this todo still relevant?

lib/compress/zstd_ldm.c
```diff
@@ -223,6 +223,17 @@ static U64 ZSTD_ldm_fillLdmHashTable(ldmState_t* state,
     return rollingHash;
 }
 
+void ZSTD_ldm_fillHashTable(
+            ldmState_t* state, const BYTE* ip,
+            const BYTE* iend, ldmParams_t const* params)
+{
+    U64 startingHash = ZSTD_rollingHash_compute(ip, params->minMatchLength);
+    ZSTD_ldm_fillLdmHashTable(
+        state, startingHash, ip, iend, state->window.base,
```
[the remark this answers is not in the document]
terrelln: Did the fuzzer find this bug? It still needs to be fixed.

programs/fileio.c
```diff
@@ -797,21 +797,21 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
     ress.dstBuffer = malloc(ress.dstBufferSize);
+    ress.dictBufferSize = FIO_createDictBuffer(&ress.dictBuffer, dictFileName, prefs);   /* works with dictFileName==NULL */
     if (!ress.srcBuffer || !ress.dstBuffer)
         EXM_THROW(31, "allocation error : not enough memory");
 
     /* Advanced parameters, including dictionary */
-    {   void* dictBuffer;
-        size_t const dictBuffSize = FIO_createDictBuffer(&dictBuffer, dictFileName, prefs);   /* works with dictFileName==NULL */
-        if (dictFileName && (dictBuffer==NULL))
+    {   if (dictFileName && (ress.dictBuffer==NULL))
```
terrelln: nit: we no longer need this scope, since we aren't declaring any variables.

programs/fileio.c
```diff
@@ -797,21 +797,21 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
             EXM_THROW(32, "allocation error : can't create dictBuffer");
         ress.dictFileName = dictFileName;
 
         if (prefs->adaptiveMode && !prefs->ldmFlag && !comprParams.windowLog)
             comprParams.windowLog = ADAPT_WINDOWLOG_DEFAULT;
 
         if (prefs->patchFromMode) {
-            comprParams.windowLog = FIO_highbit64((unsigned long long)maxSrcFileSize + PATCHFROM_WINDOWSIZE_EXTRA_BYTES);
+            /* Make sure there is enough room for dict and src in long mode */
+            comprParams.windowLog = FIO_highbit64((unsigned long long)maxSrcFileSize) + 1;
```
terrelln: What about files larger than 2 GB? Should we silently cap the window log, print a warning, or error?

I believe we should print a warning that compression ratio will be degraded.

programs/fileio.c
```diff
@@ -865,6 +868,7 @@ static void FIO_freeCResources(cRess_t ress)
 {
     free(ress.srcBuffer);
     free(ress.dstBuffer);
+    if (ress.dictBuffer) free(ress.dictBuffer);
```
terrelln: nit: You can just say `free(ress.dictBuffer)`. `free()` is a no-op for `NULL`.

programs/zstdcli.c
```diff
@@ -544,6 +544,8 @@ static int init_cLevel(void) {
     return ZSTDCLI_CLEVEL_DEFAULT;
 }
 
+#define PATCHFROM_LONG_THRESH 32 MB
```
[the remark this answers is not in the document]
terrelln: What is the status of this? 32 MB is likely not the right threshold. The `zstd -1` compression ratio on `d3-5.14.2.tar->d3-5.15.0.tar` is `2.67`, which shows that long range mode should've been enabled in that case.

programs/zstdcli.c
```diff
@@ -1197,15 +1199,22 @@ int main(int const argCount, const char* argv[])
...
     if (memLimit == 0) {
         if (compressionParams.windowLog == 0) {
             memLimit = (U32)1 << g_defaultMaxWindowLog;
         } else {
             memLimit = (U32)1 << (compressionParams.windowLog & 31);
     }   }
+    if (patchFromDictFileName != NULL) {
+        const unsigned long long dictSize = UTIL_getFileSize(patchFromDictFileName);
+        if (dictSize != UTIL_FILESIZE_UNKNOWN) {
+            memLimit = dictSize > memLimit ? (unsigned)dictSize : memLimit;
```
terrelln: Doesn't `memLimit` have to be `fileSize`, not `dictSize`? The dictionary will stay in-bounds as long as a single byte of the dictionary is in-bounds, so the `memLimit` should be the file size.

programs/zstdcli.c
```diff
@@ -1197,15 +1199,22 @@ int main(int const argCount, const char* argv[])
...
+#ifndef ZSTD_NOCOMPRESS
+            ldmFlag = dictSize > PATCHFROM_LONG_THRESH;
```
terrelln: This should probably be `fileSize + dictSize`.

programs/zstdcli.c
```diff
@@ -1197,15 +1199,22 @@ int main(int const argCount, const char* argv[])
...
-    if (patchFromDictFileName != NULL) {
-        dictFileName = patchFromDictFileName;
-    }
     if (memLimit == 0) {
         if (compressionParams.windowLog == 0) {
             memLimit = (U32)1 << g_defaultMaxWindowLog;
         } else {
             memLimit = (U32)1 << (compressionParams.windowLog & 31);
     }   }
+    if (patchFromDictFileName != NULL) {
```
terrelln: We should check to ensure that we only have one file. Since we use `refPrefix()`, which is not sticky, only one file will get the dictionary.

Can you add a test to `playTests.sh` to make sure that `--patch-from` with more than one file errors?

tests/fuzzer.c
```diff
@@ -490,6 +490,24 @@ static int basicUnitTests(U32 const seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i :  : ldm fill dict out-of-bounds check", testNb++);
```
terrelln: If this is passing, then this test doesn't expose the bug, since it isn't fixed.

bimbashrestha:
TODO: find proper value for PATCHFROM_LONG_THRESH and the fuzzer stuff.

lib/compress/zstd_ldm.c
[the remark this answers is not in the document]
bimbashrestha: I'm having some trouble exposing this bug in fuzzer.c. I tried to get zstd to raise a seg fault by doing this but that didn't seem to work: 
1. Create a dict buffer of size 2*n
2. Create a src buffer of size n
3. Generate data of size n into src
4. memcpy n bytes from src to dict and dict + n
5. realloc(dict, n) 
Do you see a way to do it?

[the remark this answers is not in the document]
terrelln: The bug should be triggered by:
1. Pass in a dictionary of N bytes that has been malloc'd in a buffer with exactly N bytes.
2. Enable long range mode with the cdict (double check that this code is called).
3. Enable ASAN.

It would take quite unlucky (or engineered) circumstances to get the binary to crash without ASAN enabled, but ASAN should find it immediately.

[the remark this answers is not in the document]
bimbashrestha: Ah perfect. I wasn't enabling ASAN. That did it

tests/fuzz/dictionary_round_trip.c
```diff
@@ -42,21 +43,33 @@ static size_t roundTripTest(void *result, size_t resultCapacity,
                 dict.buff, dict.size,
                 cLevel);
     } else {
-        dictContentType = FUZZ_dataProducer_uint32Range(producer, 0, 2);
+        /* refPrefix seems to fail when not used with auto? */
+        if (!refPrefix)
+            dictContentType = FUZZ_dataProducer_uint32Range(producer, 0, 2);
```
bimbashrestha, line 48: Gives me dictionary mismatch error when I try to use refPrefix with full_dict. Not sure if this is a real bug or if I'm doing something wrong.

lib/compress/zstd_compress.c
```diff
@@ -2793,6 +2794,11 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
     ZSTD_window_update(&ms->window, src, srcSize);
     ms->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ms->window.base);
 
+    if (params->ldmParams.enableLdm && ls != NULL) {
+        ZSTD_window_update(&ls->window, src, srcSize);
+        ls->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ls->window.base);
```
Cyan4973, line 2799: I presume the combination `enableLdm && forceWindow` has never been tested ?

programs/fileio.c
```diff
@@ -803,75 +803,82 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
-        CHECK( ZSTD_CCtx_setParameter(ress.cctx, ZSTD_c_literalCompressionMode, (int)prefs->literalCompressionMode) );
-        /* multi-threading */
+    if (dictFileName && (ress.dictBuffer==NULL))
+        EXM_THROW(32, "allocation error : can't create dictBuffer");
+    ress.dictFileName = dictFileName;
+
+    if (prefs->adaptiveMode && !prefs->ldmFlag && !comprParams.windowLog)
+        comprParams.windowLog = ADAPT_WINDOWLOG_DEFAULT;
+
+    if (prefs->patchFromMode) {
```
terrelln, line 818: Can we separate all of this logic this into its own function `FIO_adjustParamsForPatchFromMode(prefs, &comprParams, maxSrcFileSize);`

When using the optimal parser, we can suggest parameters for slower but better patches here. E.g. suggest values for hashLog/chainLog/searchLog/targetLength.

programs/zstdcli.c
```diff
@@ -1240,25 +1242,37 @@ int main(int const argCount, const char* argv[])
...
 #ifndef ZSTD_NOCOMPRESS
+        if (patchFromDictFileName != NULL) {
+            const char* const srcFileName = filenames->fileNames[0];
+            const unsigned long long fileSize = UTIL_getFileSize(srcFileName);
+            const unsigned long long dictSize = UTIL_getFileSize(patchFromDictFileName);
+            if (fileSize != UTIL_FILESIZE_UNKNOWN && dictSize != UTIL_FILESIZE_UNKNOWN) {
+                memLimit = MAX(memLimit, MAX((unsigned)dictSize, (unsigned)fileSize));
+            }
+        }
+        FIO_setMemLimit(prefs, memLimit);
```
terrelln, line 1275: Do we need to do this adjustment during decompression as well?

tests/fuzz/dictionary_round_trip.c
```diff
@@ -42,21 +43,33 @@ static size_t roundTripTest(void *result, size_t resultCapacity,
                 dict.buff, dict.size,
                 cLevel);
     } else {
-        dictContentType = FUZZ_dataProducer_uint32Range(producer, 0, 2);
+        /* refPrefix seems to fail when not used with auto? */
```
terrelln, line 46: Is this comment out of date?

tests/fuzzer.c
```diff
@@ -566,6 +566,29 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        CHECK_Z(ZSTD_CCtx_setParameter(cctx, ZSTD_c_enableLongDistanceMatching, 1));
+        assert(!ZSTD_isError(ZSTD_compress_usingDict(cctx, dst, dstCapacity, src, size, dict, size, 3)));
+
+        ZSTD_freeCCtx(cctx);
+        free(dict);
+        free(src);
+        free(dst);
+    }
+    DISPLAYLEVEL(3, "OK \n");
+
```
terrelln, line 591: I would like to see a test that exercises refPrefix + LDM. Something that tests the compressed size.

E.g. use level 1 with and without LDM on a big buffer, and make sure it compresses to basically nothing with LDM, and is much larger without.

lib/compress/zstdmt_compress.c
```diff
@@ -461,34 +461,36 @@ typedef struct {
     ZSTD_window_t ldmWindow;  /* A thread-safe copy of ldmState.window */
 } serialState_t;
 
-static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* seqPool, ZSTD_CCtx_params params, size_t jobSize)
+static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool* seqPool, 
+                        ZSTD_CCtx_params* paramsPtr, size_t jobSize, const void* dict,
```
Cyan4973, line 465: I presume that the change from `ZSTD_CCtx_params params` to `ZSTD_CCtx_params* paramsPtr`
reflects the fact that the content of `paramsPtr` can now be updated by this function,
and this update must be propagated beyond the lifetime of this function.

I also presume it's related to dictionary loading into ldmState.

It's a very important change, fundamentally changing a core property of this interface (namely that `params` was safe) so I guess it's worth underlining what needs to be updated, and why it's not possible to do it differently.

programs/zstd.1.md
```diff
@@ -128,6 +128,12 @@ the last one takes effect.
     selection, namely that windowSize > srcSize.
 
     Note: cannot use both this and -D together
+    Note: consider using --patch-from with `--long` mode on large files
```
Cyan4973, line 131: Since `--long` mode is automatically activated when considered useful, I guess it's not that important to suggest to the user that it should do it manually.
Maybe it's still useful to suggest that the manual option is available.
In which case, I would recommend to swap the order of notes 2 & 3, or combine them.

programs/zstdcli.c
```diff
@@ -1240,25 +1242,37 @@ int main(int const argCount, const char* argv[])
...
         } else {
             memLimit = (U32)1 << (compressionParams.windowLog & 31);
     }   }
-    FIO_setMemLimit(prefs, memLimit);
+    if (patchFromDictFileName != NULL)
+        dictFileName = patchFromDictFileName;
     if (operation==zom_compress) {
 #ifndef ZSTD_NOCOMPRESS
+        if (patchFromDictFileName != NULL) {
+            const char* const srcFileName = filenames->fileNames[0];
```
Cyan4973, line 1268: Could this logic be encapsulated into `fileio.c` ?

We try to reduce the amount of logic within `zstdcli.c` to a minimum, 
mostly limited to just parsing command line.

Cyan4973:
Getting there.
There are still a few comments to address, but we are definitely closer to merge status.

lib/compress/zstdmt_compress.c
```diff
@@ -507,8 +509,14 @@ static int ZSTDMT_serialState_reset(serialState_t* serialState, ZSTDMT_seqPool*
...
     }
-    serialState->params = params;
+    serialState->params = *paramsPtr;
     serialState->params.jobSize = (U32)jobSize;
+
+    /* Update window state and fill hash table with dict */
+    if (paramsPtr->ldmParams.enableLdm && dict) {
+        ZSTD_window_update(&serialState->ldmState.window, dict, dictSize);
+        ZSTD_ldm_fillHashTable(&serialState->ldmState, (const BYTE*)dict, (const BYTE*)dict + dictSize, &paramsPtr->ldmParams);
+    }
```
bimbashrestha, line 519: >I believe it's worth underlining what needs to be updated, and why it's not possible to do it differently, or why it's the better choice.

If we want to prevent prefs from being changed by this function, we'd just have to move this if statement outside of it. I placed it in here originally because it seemed like a 'reset' type operation but looking back, moving this outside and keeping the rest as before might be a better solution.

programs/fileio.c
```diff
@@ -769,6 +769,15 @@ static unsigned FIO_highbit64(unsigned long long v)
...
 }
 
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+
+static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
+                                    unsigned long long const dictSize,  
+                                    unsigned long long const maxSrcFileSize)
+{
+    if (dictSize != UTIL_FILESIZE_UNKNOWN && maxSrcFileSize != UTIL_FILESIZE_UNKNOWN)
+        FIO_setMemLimit(prefs, MAX(prefs->memLimit, MAX((unsigned)dictSize, (unsigned)maxSrcFileSize)));
```
Cyan4973, line 779: Question :
what happens when `dictSize` or `maxSrcFileSize` is > 4 GB ?

It seems this invocation could suffer an overflow here,
unless a limit range condition is checked elsewhere
in which case it should be an `assert()` here.

lib/compress/zstdmt_compress.c
Cyan4973: Transforming an argument from "read only" into "read-write" is merely an early warning sign.
It's not an "automatic proof" that it's "bad", but the implications should be clearly understood.
In particular, where does it make more sense to do those changes, and what happens to those changes afterwards ?

In order to select between 2 implementation models, 
one has to understand why this function is invoked,
hence look at the places where it's invoked.

Right now, I see 2 invocations : in `ZSTDMT_initCStream_internal()`, and in ZSTDMT_compress_advanced_internal()` (which feels logical retrospectively).

Interestingly, in both cases, `params` is a read argument from the calling function,
which means the _changes enforced into `params` will not be propagated beyond the function_.

I presume it's fine, and exactly what's needed for `ZSTDMT_compress_advanced_internal()`.

However, for `ZSTDMT_initCStream_internal()`, it seems more strange : 
the function returns right after invoking `ZSTDMT_serialState_reset()`, 
which means that any change enforced into `params` is basically useless, since `params` becomes out of scope immediately after.

`ZSTDMT_initCStream_internal()` is the most important variant used by the `zstd` cli for most files (all non-tiny ones).

This suggests that _currently_, the changes made into `params.ldmParams` are basically cancelled immediately afterwards, hence not applied during the following compression session.
I would expect such an impact to be visible during tests ?

bimbashrestha: >This suggests that currently, the changes made into params.ldmParams are basically cancelled immediately afterwards, hence not applied during the following compression session.

```
serialState->params = *paramsPtr;
```

is set later in that same function btw so that is how the changes propagate. So params becomes useless afterwards yes but it changes the internal params after its used as a temp variable or sorts it looks like.

bimbashrestha: Going to make params immutable

programs/fileio.c
bimbashrestha: It should throw an error now

tests/fuzzer.c
```diff
@@ -566,6 +566,115 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        /* test round trip refPrefix + ldm*/
+        ZSTD_DCtx_refPrefix(dctx, dict, size);
+        reconSize = ZSTD_decompressDCtx(dctx, recon, size, dst, refPrefixLdmComrpessedSize);
+        assert(!ZSTD_isError(reconSize));
+        assert(reconSize == size);
+        assert(!memcmp(recon, src, size));
+
+        /* make sure that refPrefixCompressedSize is greater */
+        assert(refPrefixCompressedSize > refPrefixLdmComrpessedSize);
```
terrelln, line 665: nit: make sure that this is significantly greater. There should be at least a 10x difference if we're not able to capture the diff without LDM.

lib/compress/zstd_compress.c
```diff
@@ -2805,6 +2813,11 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
 
         ZSTD_overflowCorrectIfNeeded(ms, ws, params, ip, ichunk);
 
+        if (params->ldmParams.enableLdm && ls != NULL && srcSize >= params->ldmParams.minMatchLength) {
+            ZSTD_ldm_fillHashTable(ls, (const BYTE*)src, (const BYTE*)src + srcSize, &params->ldmParams);
+            break;
```
terrelln, line 2818: Why break here? This stops the dictionary from filling non-LDM tables. Could this be why compression ratio is hurt so much in the optimal parser when LDM is enabled?

bimbashrestha: Yeah good catch. Not sure why I had this originally honestly. Although, even with it removed, the compression ratio doesn't change. Tested with llvm and lz4. Still, I'm gong to remove it because it doesn't make sense to not fill the other tables

programs/fileio.c
```diff
@@ -784,12 +794,39 @@ typedef struct {
...
+                                    unsigned long long const maxSrcFileSize,
+                                    int cLevel)
+{
+    unsigned const fileWindowLog = FIO_highbit64(maxSrcFileSize) + 1;
+    ZSTD_compressionParameters const cParams = ZSTD_getCParams(cLevel, (size_t)maxSrcFileSize, (size_t)dictSize);
+    FIO_adjustMemLimitForPatchFromMode(prefs, dictSize, maxSrcFileSize);
+    if (fileWindowLog > ZSTD_WINDOWLOG_MAX)
+        DISPLAYLEVEL(1, "Max window log exceeded by file (compression ratio will suffer)\n");
+    comprParams->windowLog = MIN(ZSTD_WINDOWLOG_MAX, fileWindowLog);
+    if (fileWindowLog > comprParams->chainLog) {
```
terrelln, line 815: This should probably be the cycle log. See `ZSTD_cycleLog()`.

lib/compress/zstd_compress.c
bimbashrestha: Ah alright, found a very small difference in compression ratio when I compress with --single-thread on wordpress

bimbashrestha: Looks like there are small differences in file size that get exposed when compressing with --single-thread

Cyan4973: > when compressing with --single-thread

For my understanding, do you imply there is no difference _without_ `--single-thread` ?

Also, what's the status of `--single-thread` with regards to `--patch-from` btw ? Is it orthogonal ? It it implied ? etc.

programs/fileio.c
```diff
@@ -814,6 +814,7 @@ static void FIO_adjustParamsForPatchFromMode(FIO_prefs_t* const prefs,
     }
     if (cParams.strategy >= ZSTD_btopt) {
         DISPLAYLEVEL(1, "[Optimal parser notes] Consider the following to improve patch size at the cost of speed:\n");
+		DISPLAYLEVEL(1, "- Use --single-thread mode in the zstd cli");
```
Cyan4973, line 817: nit (very minor) : alignment seems messed up due to use of `tab` instead of `space`

Cyan4973:
Question : 
we mentioned a need for a paragraph describing the feature, to be used at release time. 
Where is it ?

programs/fileio.c
bimbashrestha: Ah yes good call. Switched to vim to make that quick change. I'll fix it

bimbashrestha:
I'll out up the blurb shortly:) Just need to also finish making a nice graph or two

bimbashrestha:
Alright finally merged!:)) Also updated the PR description with the blurb @Cyan4973  and @terrelln

RubenKelevra:
Hey guys, hey @bimbashrestha,

here's the requested look at the `--long` parameter. I also compared a full update with a compressed file vs a delta update with various real-world data.

While I'm actually more interested in patching backward, reasoning[0], this test is done forwards (which shouldn't matter much anyway).

My files are:

1) OpenStreetMap XML data for Luxembourg with a 2 day difference (Filesize ~390 MB)
2) OpenStreetMap ProtocolBufferFormat data for Luxembourg with a 2 day difference (Filesize ~25 MB)
3) OpenStreetMap ProtocolBufferFormat data for Australia & Oceania with a 1 month difference (Filesize ~766 MB)
4) OpenStreetMap ProtocolBufferFormat data for Australia & Oceania with a 1 day difference (Filesize ~766 MB)
5) ArchLinux x86_64 linux kernel package 5.6.4 to 5.6.5 uncompressed (Filesize ~80 MB)
6) ArchLinux x86_64 go-pie package 1.14.1 to 1.14.2 uncompressed (Filesize ~641 MB)
7) CentOS 7 x86_64 GenericCloud qcow2-image with version hop 1905 to 1907 (Filesize ~897 MB)
8) en wikipedia news articles (pages) as xml one month difference (Filesize ~195 MB)

I'm running zstd build from the git version 5b0a452c, compiled with `-march=native -O2`, running on an 'i3-2330M'. Data is stored and loaded on a SATA-SSD and the system has 12 GB memory.

1a) `delta -15`:
```
luxembourg-200417.osm :  0.88%   (410250506 => 3628882 bytes, luxembourg-200415-to-200417.osm.zst_patch)
real    1m53,091s
user    1m51,593s
sys     0m1,070s
```

1b) `delta -15 --long`:
```
luxembourg-200417.osm :  0.04%   (410250506 => 169502 bytes, luxembourg-200415-to-200417.osm.zst_patch)
real    1m57,944s
user    1m56,030s
sys     0m1,282s
```

1c) `full update -15 --long`:

41640286 bytes -> **delta saves 41470784 bytes (99.59%)**

2a) `delta -15`:

```
luxembourg-200417.osm.pbf : 98.79%   (25349898 => 25043001 bytes, luxembourg-200415-to-200417.osm.pbf.zst_patch)
real    0m5,329s
user    0m5,158s
sys     0m0,161s
```

2b) `delta -15 --long`:

```
luxembourg-200417.osm.pbf : 98.72%   (25349898 => 25025611 bytes, luxembourg-200415-to-200417.osm.pbf.zst_patch)
real    0m5,775s
user    0m5,636s
sys     0m0,133s
```

2c) `full update -15 --long`:

25350493 bytes -> **delta saves 324882 bytes (1.28%)**

3a) `delta -15`:

```
australia-oceania-200401.osm.pbf :100.00%   (803827262 => 803820669 bytes, australia-oceania-200301-to-200401.osm.pbf.zst_patch)
real    3m23,253s
user    3m9,292s
sys     0m3,464s
```

3b) `delta -15 --long`:

```
australia-oceania-200401.osm.pbf : 99.99%   (803827262 => 803737958 bytes, australia-oceania-200301-to-200401.osm.pbf.zst_patch)
real    3m44,705s
user    3m26,232s
sys     0m4,313s
```

3c) `full update -15 --long`:

803820673 bytes -> **delta saves 82715 bytes (0.01%)**

4a) `delta -15`:

```
australia-oceania-200417.osm.pbf :100.00%   (807853291 => 807844053 bytes, australia-oceania-200416-to-200417.osm.pbf.zst_patch)
real    3m21,324s
user    3m10,312s
sys     0m3,081s
```

4b) `delta -15 --long`:

```
australia-oceania-200417.osm.pbf : 99.59%   (807853291 => 804520667 bytes, australia-oceania-200416-to-200417.osm.pbf.zst_patch)
real    3m36,696s
user    3m27,023s
sys     0m3,033s
```

4c) `full update -15 --long`:

807843840 bytes -> **delta saves 3323173 bytes (0.41%)**

5a) `delta -15`:

```
linux-5.6.5.arch1-1-x86_64.pkg.tar : 86.53%   (83640320 => 72370676 bytes, linux-5.6.4-to-5.6.5.arch1-1-x86_64.pkg.tar.zst_patch)
real    0m20,083s
user    0m19,092s
sys     0m0,517s
```

5b) `delta -15 --long`:

```
linux-5.6.5.arch1-1-x86_64.pkg.tar : 86.33%   (83640320 => 72204691 bytes, linux-5.6.4-to-5.6.5.arch1-1-x86_64.pkg.tar.zst_patch)
real    0m20,919s
user    0m20,391s
sys     0m0,421s
```

5c) `full update -15 --long`:

72891396 bytes -> **delta saves 686705 bytes (0.94%)**

6a) `delta -15`:

```
go-pie-2_1.14.2-1-x86_64.pkg.tar : 12.51%   (672358400 => 84082084 bytes, go-pie-2_1.14.1-to-1.14.2-1-x86_64.pkg.tar.zst_patch)
real    5m12,541s
user    5m6,284s
sys     0m2,869s
```

6b) `delta -15 --long`:

```
go-pie-2_1.14.2-1-x86_64.pkg.tar :  8.34%   (672358400 => 56105936 bytes, go-pie-2_1.14.1-to-1.14.2-1-x86_64.pkg.tar.zst_patch)
real    4m55,772s
user    4m52,609s
sys     0m1,777s
```

6c) `full update -15 --long`:

156160084 bytes -> **delta saves 100054148 bytes (64.07%)**

7a) `delta -15`:

```
CentOS-7-x86_64-GenericCloud-1907.qcow2 : 27.55%   (942407680 => 259600155 bytes, CentOS-7-x86_64-GenericCloud-1905-to-1907.qcow2.zst_patch)
real    5m43,531s
user    5m34,114s
sys     0m3,793s
```

7b) `delta -15 --long`:

```
CentOS-7-x86_64-GenericCloud-1907.qcow2 : 15.18%   (942407680 => 143012290 bytes, CentOS-7-x86_64-GenericCloud-1905-to-1907.qcow2.zst_patch)
real    4m9,574s
user    4m4,923s
sys     0m3,366s
```

7c) `full update -15 --long`:

277690047 bytes -> **delta saves 134677757 bytes (48.50%)**

8a) `delta -15`:

```
enwikinews-20200401-pages-articles.xml :  4.53%   (204994714 => 9285772 bytes, enwikinews-20200301-to-20200401-pages-articles.xml.zst_patch)
real    2m25,614s
user    2m14,190s
sys     0m2,773s
```

8b) `delta -15 --long`:

```
enwikinews-20200401-pages-articles.xml :  0.80%   (204994714 => 1633204 bytes, enwikinews-20200301-to-20200401-pages-articles.xml.zst_patch)
real    1m44,225s
user    1m40,599s
sys     0m0,909s
```

8c) `full update -15 --long`:

39571286 bytes -> **delta saves 37938082 bytes (95.87%)**

Observations
---

OpenStreetMap data is sorted by ID, so new IDs appear at the end, changes don't move the data in its position, just change their size, while deletes just removes the data.

I suspect that the XML from Wikipedia is also sorted by key, but haven't confirmed that. But it contains a lot of large text chunks of static data.

go-pie either holds a lot of static files, which doesn't change much or the compilation creates very stable binaries.

The Linux kernel is hard to compress and extremely hard to diff - the second might be caused by an unstable compilation process?

qcow2 images are particularly good to diff when there are low amounts of changes (lts distributions)

The binary protocol buffer format (and probably also the export process) makes it nearly impossible to find unchanged data, therefore the diff barely makes an impact.

Conclusion
---

It looks like the worst case is the same size as a regular compressed full update.

[0] https://github.com/facebook/zstd/issues/2063#issuecomment-610694799

bimbashrestha:
Thanks for the update @RubenKelevra!

Just so we're on the same page here, I'm assuming `delta -15` is `zstd --patch-from` before this PR was merged, `delta -15 --long` is `zstd--patch-from` after this PR was merged and `full update` is just `zstd -15` without patch-from?

RubenKelevra:
> Thanks for the update @RubenKelevra!
> 
> Just so we're on the same page here, I'm assuming `delta -15` is `zstd --patch-from` before this PR was merged, `delta -15 --long` is `zstd--patch-from` after this PR was merged and `full update` is just `zstd -15` without patch-from? 

Nope, it's all done with the same version (the quoted commit). If you like to have a comparison between this version and a different one I can do this as well.

Which commit should I choose for a comparison? :)

Cyan4973:
That's unexpected. Shouldn't `--long` be automatically enabled when input size is large enough, which several of these examples definitely are ?

bimbashrestha:
>Nope, it's all done with the same version (the quoted commit)

Hmm yeah like @Cyan4973 mentioned, long mode should be automatically activated when the input is large enough. Is `delta` a script that just invokes the zstd cli with and without the `--long` option?

RubenKelevra:
> > Nope, it's all done with the same version (the quoted commit)
> 
> Hmm yeah like @Cyan4973 mentioned, long mode should be automatically activated when the input is large enough. Is `delta` a script that just invokes the zstd cli with and without the `--long` option?

No, I wasn't doing it that fancy, but you can think of it like this. ;)

I just didn't want to clutter my post with all those filenames :)

The `--long` parameter is definitely not set automatically at the moment (at least in my scenario).

bimbashrestha:
@RubenKelevra @Cyan4973  Ah I see the issue. It's this line here. 

https://github.com/facebook/zstd/blob/38a6d2a8f2b39d6b76155ef1dbac9e28ee1b7379/programs/fileio.c#L810

That should be `cParams.hashLog` and not `comprParams->hashLog`. We had a different heuristic that used `chainLog` before so I never tested to ensure that long mode was properly triggered when we use `ZSTD_cycleLog.` 

Glad we caught that. I'll put up a PR and a test fixing it shortly.

PR #1965, opened 2020-01-22
Fix timefn on android

terrelln:
Android doesn't provide `timespec_get()`.


Cyan4973:
Strange,
does that mean it does not provide `timespec_get()` while still claiming `C11` compliance __and__ providing `TIME_UTC` macro, which sole purpose is to be an argument of `timespec_get()` ?

terrelln:
Yes... https://github.com/tinycthread/tinycthread/issues/47

PR #1968, opened 2020-01-22
[automated_benchmarking] Make arguments optional and add --dict argument

bimbashrestha:
* Make the cli arguments optional
* Add early check for files count (bail when len(filenames == 0)
* Add --dict argument (when set, this will use the first half of the file as a dictionary for the second half)

Example: 
Benchmark using the first half of silesia files as the dict for the second half and compare the current hash to facebook:dev
```
python automated_benchmarking.py --mode current --dict True --dir ~/silesia
```

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -238,7 +238,7 @@ versionsTest: clean
 	$(PYTHON) test-zstd-versions.py
 
 automated_benchmarking: clean
-	$(PYTHON) automated_benchmarking.py golden-compression 1 current 1 "" 60
+	$(PYTHON) automated_benchmarking.py
 
 checkTag: checkTag.c $(ZSTDDIR)/zstd.h
 	$(CC) $(FLAGS) $< -o $@$(EXT)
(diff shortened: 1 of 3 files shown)


Cyan4973:
> when set, this will use the first half of the file as a dictionary for the second half

I'm intrigued by this selected scenario.
It's very specific, requires a specific command and code path for its own sake, and doesn't seem to represent any real scenario.

What is the motivation ?

bimbashrestha:
> It's very specific, requires a specific command and code path for its own sake, and doesn't seem to represent any real scenario. What is the motivation? 

I didn't have a very specific scenario in mind. I wanted to add dictionary compression benchmarking to the script and this seemed like a reasonable decision because: 

1. The content of a dictionary is usually very similar to the content that is being compressed (hence the data is split in half).

2. We don't have to explicitly specify dictionary files through the cli so its convenient to use. Just run `python automated_benchmarking.py --dict True --dir <whatever>`

Note: I used this to benchmark https://github.com/facebook/zstd/pull/1954

Cyan4973:
Unfortunately, I don't think this is representative of dictionary compression performance. It's actually misleading.

Measuring dictionary performance requires a real dictionary (and a lot of files to compress with the dictionary). It will result in a very different outcome.

I think it's better to keep `--dict DictionaryFile` for this more meaningful scenario.

bimbashrestha:
> I think it's better to keep `--dict DictionaryFile` for this more meaningful scenario.
Sounds good. I'll update this pr

PR #1970, opened 2020-01-23
[test-symbols] Make visibility hidden by default on cmake and add symbols test 

bimbashrestha:
I tested this by 
1. removing ZSTDLIB_API on ZSTD_compress
2. gcc symbols.c -o symbols -Wl,-rpath=../lib ../build/cmake/build/lib/libzstd.so.1.4.5 && ./symbols
Which fails

Next step will be to deprecate this test completely. Which should be as straightforward as ensuring that all of the symbols in symbols.c are used in unit tests somewhere. Then make cmakebuild will catch any missing ones.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -20,6 +20,7 @@
 
 ZSTDDIR = ../lib
 PRGDIR  = ../programs
+BUILDIR = ../build
 PYTHON ?= python3
 TESTARTEFACT := versionsTest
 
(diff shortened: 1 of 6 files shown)


Cyan4973:
Where are we with this patch ?

bimbashrestha:
> Where are we with this patch ?

Trying to fix some CI issues on my fork.

bimbashrestha:
Okay, I think this is ready now. I'm going to remove symbols.c in a subsequence PR after I make sure all of its symbols are covered in unit tests.

tests/Makefile
```diff
@@ -225,9 +225,9 @@ decodecorpus : $(filter-out zstdc_zstd_compress.o, $(ZSTD_OBJECTS)) $(ZDICT_FILE
 symbols  : symbols.c zstd-dll
 ifneq (,$(filter Windows%,$(OS)))
 	cp $(ZSTDDIR)/dll/libzstd.dll .
-	$(CC) $(FLAGS) $< -o $@$(EXT) -DZSTD_DLL_IMPORT=1 libzstd.dll
+	$(CC) $(FLAGS) $< -o $@$(EXT) -fvisibility=hidden -DZSTD_DLL_IMPORT=1 libzstd.dll
 else
-	$(CC) $(FLAGS) $< -o $@$(EXT) -Wl,-rpath=$(ZSTDDIR) $(ZSTDDIR)/libzstd.so   # broken on Mac
+	$(CC) $(FLAGS) $< -o $@$(EXT) -fvisibility=hidden -Wl,-rpath=$(ZSTDDIR) $(ZSTDDIR)/libzstd.so   # broken on Mac
```
Cyan4973: I think that `-fvisibility=hidden` is a directive that must be passed _at dynamic library compilation time_.
It sets the symbols that can be accessed at link time when a program links to the dynamic library.

Which means that, when compiling the caller program (`symbols` in this case), it's no longer useful to specify `-fvisibility=hidden` : the list of accessible symbols is already set within `libzstd.so` or `libzstd.dll`.

bimbashrestha: Just realized this is already being done over here: 

https://github.com/facebook/zstd/blob/6a426ef65e0864b658dc78bbb45eb23c29b1275d/lib/Makefile#L169

So I guess I don't need this pr at all and can just move forward with trying to deprecate symbols.c by ensuring all symbols are called by unit tests in a following pr?

Cyan4973: I agree

bimbashrestha: Okay, I'll go ahead and close this PR then.

Issue #1969, opened 2020-01-22
Use OpenBSD pledge() and unveil()


felixhandte:
OpenBSD has a nifty security syscall called [`pledge()`](https://man.openbsd.org/pledge), which lets processes limit their access to system functionality they don't expect to use. If the process is hijacked somehow, this limits the ability of an attacker to cause damage to the system. It would be nice for the CLI to take advantage of this proactive privilege dropping, since (a) we don't need access to very many system features at all, and (b) although we are careful, we are doing classically dangerous things (handling untrusted input, doing partially unchecked array accesses and copies, interacting with uninitialized memory, etc.).

I'd be interested in playing around with this myself, but I don't have an OpenBSD system and it doesn't look like any of our CI services support OpenBSD.

PR #1971, opened 2020-01-23
[experimental] Use `seccomp()` to Sandbox Zstd

felixhandte:
I realized Linux has analogous functionality to OpenBSD's `pledge()` (see #1969). This is an experiment exploring using the `seccomp()` syscall to limit what Zstd can do.

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -14,6 +14,7 @@
 #include "debug.h"     /* assert */
 #include "zstd_internal.h"  /* ZSTD_malloc, ZSTD_free */
 #include "pool.h"
+#include "zstd_seccomp.h" /* ZSTD_disable_syscalls_for_worker_thread */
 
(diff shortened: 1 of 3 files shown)


felixhandte:
As discussed, I think this is the kind of tool I'd like to see Zstd use in the future. However, at present, it seems like the downsides outweigh the benefits. Using `seccomp()` is brittle, and when it breaks, it's hard to debug.

As such, I'm going to close this for now and mark the issue `long-term`.

indolering:
While researching the Landlock ticket, I came across a Postgres discussion on `seccomp()` in which the author [suggested](https://lwn.net/Articles/800953/) using `seccomp()` to blacklist high-risk syscalls.  The ongoing maintenance cost of comprehensive sanndboxing via `seccomp()` is probably too high, but a blacklist might strike the right cost/benefit balance.

PR #1975, opened 2020-01-27
minor : update educational decoder

Cyan4973:
added an ability to remove error messages and dictionary support
using macro at compile time
for the benefit of smaller binary size.

### DIFF
--- a/doc/educational_decoder/Makefile
+++ b/doc/educational_decoder/Makefile
@@ -36,7 +36,7 @@ harness: $(HARNESS_FILES)
 	$(CC) $(FLAGS) $^ -o $@
 
 clean:
-	@$(RM) harness
+	@$(RM) harness *.o
 	@$(RM) -rf harness.dSYM  # MacOS specific
 
 test: harness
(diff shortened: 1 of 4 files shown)


Cyan4973:
That's a good point !
I'm not aware of any application using this version in production, which is understandable, as it was never branded as "production ready".

That being said, this decoder has been fuzzed, using its own fuzzer test suite (not oss-fuzz), mostly to ensure correctness (and not just resistance).

It's unclear to me if it would deserve a stronger continuous fuzzer test.

terrelln:
I think for a demo it is fine as-is. But, its own fuzzer is not enough for a production decoder. For that I would want to have it integrated into oss-fuzz for both round-trip and decompression fuzzing for at least a month.

Cyan4973:
Sounds reasonable.

One remaining opened question : would it be useful to modify the educational decoder for it to also become a "small decoder" version, suitable to be embedded into a small app ? Then it would be worth fuzzing.

However, I'm currently worried that such an additional goal could be at odd with the "educational" part, which is the primary objective of this port. So I'm not pushing too strongly in this direction.

PR #1976, opened 2020-01-27
Add Minification Variable to `lib/Makefile`

felixhandte:
This diff reorganizes the `lib/Makefile` to extract various settings that a
user would normally invoke together (supposing that they were aware of them)
if they were trying to build the smallest `libzstd` possible. It collects
these settings under a master setting `ZSTD_LIB_MIN_SIZE`.

Also document this new option.


Cyan4973:
Could we add a test for it (in CI) ?

Issue #1487, opened 2018-12-31
Document/show compression levels' --zstd options


hfuru:
Please provide a table with
  {compression level: --zstd options set by that level}
in man zstd(1), or provide a way to ask the zstd program.

Without this, it's hard to know what we are doing with
  zstd --zstd=<options modifying -level>.

Cyan4973:
That's a good point @hfuru .

From the time being, a possible work-around is to [consult these tables](https://github.com/facebook/zstd/blob/v1.3.8/lib/compress/zstd_compress.c#L4153) .

If you have access to the API, there is also [this function](https://github.com/facebook/zstd/blob/v1.3.8/lib/zstd.h#L1231) in the _experimental_ section which goes straight to the point.

A difficulty is that default parameters depend on both compression level and source size, making it more complex to request.

Cyan4973:
Possible implementation idea : if the point is _show_ (as opposed to _request_) advanced compression parameters being used for current compression, maybe this could be displayed as part of `-vv` (very verbose) command ?

hfuru:
I want to tune the compression level some scripts use.  That only
makes sense if my scripts knows more than zstd does.

So, the tables with a filesize key looks like the answer here.
Maybe until you modify them in a later zstd version and my tuning
becomes obsolete:-)

It also sounds nice nice to the see the params in -vv.. output, yes.
But if so, please add a "(params depend on file size)" note. Otherwise
that info becomes a trap instead: Users could start with what -vv..
shows, benchmark compression for a few files, tweak the params, and
miss out because zstd with default params would have been smarter with
real-world input files.

Back to tuning, I guess the flexible way is to be able to say
  zstd -level --ztd=params --zstd-defaults=params
This would add the zstd-defaults to --zstd, except when
"zstd -level" would use better values.

I'll write an zstd-wrapper doing something like that. With your
tables inverted a bit: {zstd-level: {filesize: params, ...}, ...}.

Question: When does zstd know the filesize - when the input is a
regular file?  That should be documented too.

Also, it looks like --sizehint=<filesize> or --zstd=sizehint=...
might be a useful option, for pipes/sockets where zstd does not
know what size to expect.  Or call it maxsize_hint / minsize_hint
to clarify whether it's better to guess to high or too low size.

terrelln:
The zstd library knows the file size when using a one-pass function where the whole input is available, or if it is explicitly provided when starting a streaming operation. The zstd CLI knows the size when the input is a regular file.

[...]

PR #1980, opened 2020-01-30
Adding --show-default-cparams to cli 

bimbashrestha:
Displays what the default cparams will be before compressing (for regular files only). Feature request: https://github.com/facebook/zstd/issues/1487
```
./zstd --show-default-cparams  appveyor.yml

appveyor.yml (13822 bytes)
 - windowLog    : 14
 - chainLog     : 14
 - searchLog    : 2
 - minMatch     : 4
 - targetLength : 0
 - strategy     : ZSTD_dfast
```


programs/zstdcli.c
```diff
@@ -1213,6 +1227,25 @@ int main(int const argCount, const char* argv[])
         if (adaptMin > cLevel) cLevel = adaptMin;
         if (adaptMax < cLevel) cLevel = adaptMax;
 
+        if (showDefaultCParams) {
+            size_t fileNb;
+            static const char* ZSTD_strategyMap[9 + 1] = { "", "ZSTD_fast",
+                "ZSTD_dfast", "ZSTD_greedy", "ZSTD_lazy", "ZSTD_lazy2", "ZSTD_btlazy2",
+                "ZSTD_btopt", "ZSTD_btultra", "ZSTD_btultra2"};
+            for (fileNb = 0; fileNb < (size_t)filenames->tableSize; fileNb++) {
+                const size_t fileSize = (size_t)UTIL_getFileSize(filenames->fileNames[fileNb]);
```
terrelln: We should check for failures here. It is possible that the first filename is a regular file, but the second isn't.

programs/zstdcli.c
```diff
@@ -1170,6 +1173,17 @@ int main(int const argCount, const char* argv[])
     }   }
 #endif
 
+    if (showDefaultCParams) {
+        if (operation == zom_decompress) {
+            DISPLAY("error : can't use --show-default-cparams in decomrpession mode \n");
+            CLEAN_RETURN(1);
+        }
+        if (!UTIL_isRegularFile(filenames->fileNames[0])) {
```
terrelln: What if there is more than one file?

programs/zstdcli.c
```diff
@@ -1213,6 +1227,25 @@ int main(int const argCount, const char* argv[])
...
 
+        if (showDefaultCParams) {
+            size_t fileNb;
+            static const char* ZSTD_strategyMap[9 + 1] = { "", "ZSTD_fast",
+                "ZSTD_dfast", "ZSTD_greedy", "ZSTD_lazy", "ZSTD_lazy2", "ZSTD_btlazy2",
+                "ZSTD_btopt", "ZSTD_btultra", "ZSTD_btultra2"};
+            for (fileNb = 0; fileNb < (size_t)filenames->tableSize; fileNb++) {
+                const size_t fileSize = (size_t)UTIL_getFileSize(filenames->fileNames[fileNb]);
+                const size_t cParamsIdx = (fileSize < 16 KB) ? 3 : ( (fileSize < 128 KB) ? 2 : ( (fileSize < 256 KB) ? 1 : 0) );
+                ZSTD_compressionParameters cParams = ZSTD_defaultCParameters[cParamsIdx][cLevel];
```
terrelln: There is a function for this, `ZSTD_getCParams()`. We don't need to move the `ZSTD_defaultCParameters` structure out of `zstd_compress.c`.

programs/zstdcli.c
```diff
@@ -1213,6 +1227,25 @@ int main(int const argCount, const char* argv[])
         if (adaptMin > cLevel) cLevel = adaptMin;
         if (adaptMax < cLevel) cLevel = adaptMax;
 
+        if (showDefaultCParams) {
+            size_t fileNb;
+            static const char* ZSTD_strategyMap[9 + 1] = { "", "ZSTD_fast",
```
terrelln: This should be a global constant.

programs/zstdcli.c
terrelln: When we don't know the file size, instead of just erroring, we can say `size: unknown` and give the "generic" parameters.

bimbashrestha: Ah okay. good point

programs/zstdcli.c
bimbashrestha: Yeah. good call. I'll check add the check for all files

programs/zstdcli.c
```diff
@@ -1213,6 +1227,25 @@ int main(int const argCount, const char* argv[])
...
+                "ZSTD_dfast", "ZSTD_greedy", "ZSTD_lazy", "ZSTD_lazy2", "ZSTD_btlazy2",
+                "ZSTD_btopt", "ZSTD_btultra", "ZSTD_btultra2"};
+            for (fileNb = 0; fileNb < (size_t)filenames->tableSize; fileNb++) {
+                const size_t fileSize = (size_t)UTIL_getFileSize(filenames->fileNames[fileNb]);
+                const size_t cParamsIdx = (fileSize < 16 KB) ? 3 : ( (fileSize < 128 KB) ? 2 : ( (fileSize < 256 KB) ? 1 : 0) );
```
Cyan4973: This way of doing embeds too much internal logic into `zstdcli.c` which is an external module (from a library perspective).
This is going to hurt maintenance.

Consider re-using a logic which is centrally defined.
In this case, there is `ZSTD_getCParams()` which is likely providing what you are looking for, without a need to expose and then access the internal tables directly.

programs/zstdcli.c
```diff
@@ -1213,6 +1227,23 @@ int main(int const argCount, const char* argv[])
         if (adaptMin > cLevel) cLevel = adaptMin;
         if (adaptMax < cLevel) cLevel = adaptMax;
 
+        if (showDefaultCParams) {
+            size_t fileNb;
+            for (fileNb = 0; fileNb < (size_t)filenames->tableSize; fileNb++) {
+                const size_t fileSize = UTIL_isRegularFile(filenames->fileNames[fileNb]) ? (size_t)UTIL_getFileSize(filenames->fileNames[fileNb]) : 0;
```
Cyan4973: The casting to `size_t` doesn't seem useful.
It seems to keep the information in its original format, `unsigned long long`.

programs/zstdcli.c
```diff
@@ -693,6 +699,7 @@ int main(int const argCount, const char* argv[])
                 if (!strcmp(argument, "--rm")) { FIO_setRemoveSrcFile(prefs, 1); continue; }
                 if (!strcmp(argument, "--priority=rt")) { setRealTimePrio = 1; continue; }
                 if (!strcmp(argument, "--output-dir-flat")) {nextArgumentIsOutDirName=1; lastCommand=1; continue; }
+                if (!strcmp(argument, "--show-default-cparams")) { showDefaultCParams = 1; continue; }
```
Cyan4973: Consider documenting this new command within `man zstd`, hence in `zstd.1.md`.

Very advanced command like this one don't need to be listed by `zstd -h`.

programs/zstdcli.c
```diff
@@ -1213,6 +1227,23 @@ int main(int const argCount, const char* argv[])
...
+                const size_t fileSize = UTIL_isRegularFile(filenames->fileNames[fileNb]) ? (size_t)UTIL_getFileSize(filenames->fileNames[fileNb]) : 0;
+                const size_t dictSize = dictFileName != NULL ? (size_t)UTIL_getFileSize(dictFileName) : 0;
+                const ZSTD_compressionParameters cParams = ZSTD_getCParams(cLevel, (unsigned long long)fileSize, dictSize);
+                if (fileSize) DISPLAY("%s (%u bytes)\n", filenames->fileNames[fileNb], (unsigned)fileSize);
+                else DISPLAY("%s (src size unknown)\n", filenames->fileNames[fileNb]);
+                DISPLAY(" - windowLog    : %u\n", (unsigned)cParams.windowLog);
+                DISPLAY(" - chainLog     : %u\n", (unsigned)cParams.chainLog);
+                DISPLAY(" - searchLog    : %u\n", (unsigned)cParams.searchLog);
+                DISPLAY(" - minMatch     : %u\n", (unsigned)cParams.minMatch);
+                DISPLAY(" - targetLength : %u\n", (unsigned)cParams.targetLength);
```
Cyan4973: minor: all these values are supposed to be already `unsigned`, so there is no value in casting them again.

programs/zstdcli.c
bimbashrestha: Whoops. sounds good!

Cyan4973:
Looks mostly good to me.
Only some minor comment on documentation and some casting simplification remaining.

programs/zstdcli.c
bimbashrestha: Okay, I'll add it to the man

programs/zstdcli.c
terrelln: `UTIL_getFileSize()` returns `UTIL_FILESIZE_UNKNOWN` when the file size is unknown. Instead of checking if it is a regular file, check if the file size is `UTIL_FILESIZE_UNKNOWN`.

programs/zstd.1.md
```diff
@@ -363,6 +363,11 @@ Compression of small files similar to the sample set will be greatly improved.
 
     `zstd --train-legacy=selectivity=8 FILEs`
 
+* `--show-default-cparams`:
```
Cyan4973: It's positioned in the "dictionary builder" category, which is probably not the right category for this command.

Also, could you explain to me the last sentence ?

programs/zstdcli.c
```diff
@@ -544,6 +544,11 @@ static int init_cLevel(void) {
     return ZSTDCLI_CLEVEL_DEFAULT;
 }
 
+#define ZSTD_NB_STRATEGIES 9
```
Cyan4973: Find a way to compare this value with its source of truth.
Strategy indexes have bounds, which can be consulted with `ZSTD_cParam_getBounds()`.
Check that it corresponds.

programs/zstdcli.c
```diff
@@ -1213,6 +1227,23 @@ int main(int const argCount, const char* argv[])
...
+                const size_t dictSize = dictFileName != NULL ? (size_t)UTIL_getFileSize(dictFileName) : 0;
+                const ZSTD_compressionParameters cParams = ZSTD_getCParams(cLevel, fileSize, dictSize);
+                if (fileSize != UTIL_FILESIZE_UNKNOWN) DISPLAY("%s (%u bytes)\n", filenames->fileNames[fileNb], (unsigned)fileSize);
+                else DISPLAY("%s (src size unknown)\n", filenames->fileNames[fileNb]);
+                DISPLAY(" - windowLog    : %u\n", cParams.windowLog);
+                DISPLAY(" - chainLog     : %u\n", cParams.chainLog);
+                DISPLAY(" - searchLog    : %u\n", cParams.searchLog);
+                DISPLAY(" - minMatch     : %u\n", cParams.minMatch);
+                DISPLAY(" - targetLength : %u\n", cParams.targetLength);
+                DISPLAY(" - strategy     : %s\n", ZSTD_strategyMap[(int)cParams.strategy]);
```
Cyan4973: Ensure that this array read cannot result in an out-of-bound access.
An `assert()` is good enough for this case, since it's not supposed to happen.

programs/zstd.1.md
bimbashrestha: Whoops, I meant to put it under options. I'm not sure which category it goes under in there. Integer suffixes and special values, Operation mode, Restricted usage of Environment Variables don't seem apt. Maybe I'll create a new section called misc. arguments?

programs/zstdcli.c
bimbashrestha: By check that it corresponds, you mean add an assert somewhere to make sure that it is within bounds?

programs/zstd.1.md
Cyan4973: It's a kind of request for some supplemental information display, 
so it could be in the same category as `-v` / `--verbose`.

programs/zstdcli.c
Cyan4973: Yes.

Though "within bound" is another topic (mentioned below).

Here, the objective is merely to ensure that the table of names is relevant. In case the nb of strategies change in the future, we need a clear signal that this part needs also be updated.

bimbashrestha: Actually I guess I could just use ZSTD_STRATEGY_MAX

Cyan4973: Yes, though : 
- It requires accessing the unstable section. We have successfully avoided that in `zstdcli.c` so far. In contrast, `ZSTD_cParam_getBounds()` is stable.
- It might still be worthy to use the constant instead of a function result _if_ the `assert()` was transformed into a _static assert_, in order to provide a compile-time signal. Unfortunately, no `static_assert()` macro is defined in `zstdcli.c`.

programs/zstdcli.c
```diff
@@ -544,6 +544,14 @@ static int init_cLevel(void) {
     return ZSTDCLI_CLEVEL_DEFAULT;
 }
 
+#define ZSTD_CLI_STATIC_ASSERT(c) (void)sizeof(char[(c) ? 1 : -1])
+
+#define ZSTD_NB_STRATEGIES ZSTD_STRATEGY_MAX
```
Cyan4973: I'm surprised this works, because there is no place where `zstdcli.c` depends explicitly on experimental functions within `zstd.h`.

It appears this capability is inherited transitively from `benchzstd.h`.

But this is not clean: `benchzstd.h` could change its dependencies anytime in the future and no longer depend on experimental declarations. Therefore, if `zstdcli.c` depends on experimental functions, it should be explicitly mentioned in its own list of imports.

Add the macro `ZSTD_STATIC_LINKING_ONLY` in front of `#include "zstd.h"`. Explain why it's needed (`ZSTD_STRATEGY_MAX`).

programs/zstdcli.c
```diff
@@ -544,6 +544,14 @@ static int init_cLevel(void) {
...
+
+static const char* ZSTD_strategyMap[ZSTD_NB_STRATEGIES + 1] = { "", "ZSTD_fast",
```
Cyan4973: OK, so this will warn you if `ZSTD_NB_STRATEGIES` is reduced in the future,
but not if it grows.

Unfortunately, growing is a more likely scenario (we may add a new strategy in the future).

Suggestion to detect any future change of `ZSTD_STRATEGY_MAX` : 
continue to separate `ZSTD_NB_STRATEGIES` from `ZSTD_STRATEGY_MAX`, and ensure their values are equal.
If you use a simple runtime `assert()`, you don't need `ZSTD_STRATEGY_MAX` : a runtime call to stable `ZSTD_cParam_getBounds()` will provide the same value.

If you use a `static_assert`, then `ZSTD_STRATEGY_MAX` is needed.

Cyan4973:
Almost there !

Just think about what could happen if `ZSTD_STRATEGY_MAX` increase in the future, and ensure that it results in some unmistakable signal, so the proper correction measures can be taken.

PR #1982, opened 2020-01-31
Using appliedParams instead of "supplied" params in compressBegin()

bimbashrestha:
EDIT: 

This PR doesn't change compression behavior, it is necessary for enabling ldm with dictionaries because resetCCtx changes ldm params

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3059,8 +3059,9 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
     {   size_t const dictID = cdict ?
                 ZSTD_compress_insertDictionary(
                         cctx->blockState.prevCBlock, &cctx->blockState.matchState,
-                        &cctx->workspace, &cctx->appliedParams, cdict->dictContent, cdict->dictContentSize,
-                        dictContentType, dtlm, cctx->entropyWorkspace)
+                        &cctx->workspace, &cctx->appliedParams, cdict->dictContent,
+                        cdict->dictContentSize, dictContentType, dtlm,
+                        cctx->entropyWorkspace)
               : ZSTD_compress_insertDictionary(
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -3059,11 +3059,11 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
     {   size_t const dictID = cdict ?
                 ZSTD_compress_insertDictionary(
                         cctx->blockState.prevCBlock, &cctx->blockState.matchState,
-                        &cctx->workspace, params, cdict->dictContent, cdict->dictContentSize,
+                        &cctx->workspace, &cctx->appliedParams, cdict->dictContent, cdict->dictContentSize,
```
felixhandte: Nit: please line-wrap.

Cyan4973:
> Using appliedParams instead of "supplied" params in compressBegin()

Is this what we want ?
Describe the difference, and why it's better.

felixhandte:
On further consideration, this is not a bug. If `ZSTD_resetCCtx_internal()` modifies cparams, then this is a bug that needs to be fixed. But it doesn't, and shouldn't.

@Cyan4973, there is no difference. But I guess the point is that if there were, it would be important for us to use the applied params, not the requested params.

bimbashrestha:
> On further consideration, this is not a bug. If `ZSTD_resetCCtx_internal()` modifies cparams, then this is a bug that needs to be fixed. But it doesn't, and shouldn't.

It calls ZSTD_ldm_adjustParameters when we're in ldm mode which sets the ldm params. And I need to be able to use that inside loadDictionaryContent when filling the ldm hash tables.

felixhandte:
Ok, sorry, yes. I was talking specifically about the cparams.

terrelln:
While not a bug, it is necessary to fill LDM tables while loading the dictionary.

PR #1984, opened 2020-02-02
Reduce stack usage of HUF_sort()

MeghnaM:
Addresses issue #1636.

### DIFF
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -299,7 +299,7 @@ static void HUF_sort(nodeElt* huffNode, const unsigned* count, U32 maxSymbolValu
 {
     U32 n;
 
-    memset(rank, 0, sizeof(rankPos)*32);
+    memset(rank, 0, sizeof(rankPos) * 32);
     for (n=0; n<=maxSymbolValue; n++) {
         U32 r = BIT_highbit32(count[n] + 1);
         rank[r].base ++;
@@ -334,11 +334,11 @@ size_t HUF_buildCTable_wksp (HUF_CElt* tree, const unsigned* count, U32 maxSymbo
     int lowS, lowN;
     U16 nodeNb = STARTNODE;
     U32 nodeRoot;
-    rankPos rank[32];
-
+    rankPos* rank = (rankPos *)(huffNode + 1);
+    DEBUGLOG(3, "workspace %p huffnode0 %p huffnode %p rank %p workspaceend %p", workSpace, huffNode0, huffNode, rank, ((BYTE*)workSpace) + wkspSize);
     /* safety checks */
     if (((size_t)workSpace & 3) != 0) return ERROR(GENERIC);  /* must be aligned on 4-bytes boundaries */
-    if (wkspSize < sizeof(huffNodeTable)) return ERROR(workSpace_tooSmall);
+    if (wkspSize < sizeof(huffNodeTable) + sizeof(rankPos) * 32) return ERROR(workSpace_tooSmall);
     if (maxNbBits == 0) maxNbBits = HUF_TABLELOG_DEFAULT;
     if (maxSymbolValue > HUF_SYMBOLVALUE_MAX) return ERROR(maxSymbolValue_tooLarge);
     memset(huffNode0, 0, sizeof(huffNodeTable));
@@ -403,8 +403,9 @@ size_t HUF_buildCTable_wksp (HUF_CElt* tree, const unsigned* count, U32 maxSymbo
  */
 size_t HUF_buildCTable (HUF_CElt* tree, const unsigned* count, unsigned maxSymbolValue, unsigned maxNbBits)
 {
-    huffNodeTable nodeTable;
-    return HUF_buildCTable_wksp(tree, count, maxSymbolValue, maxNbBits, nodeTable, sizeof(nodeTable));
+    U32 workspace[HUF_WORKSPACE_SIZE_U32];
+    DEBUGLOG(3, "!!!");
+    return HUF_buildCTable_wksp(tree, count, maxSymbolValue, maxNbBits, workspace, sizeof(workspace));
 }
 
 size_t HUF_estimateCompressedSize(const HUF_CElt* CTable, const unsigned* count, unsigned maxSymbolValue)
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -334,8 +334,8 @@ size_t HUF_buildCTable_wksp (HUF_CElt* tree, const unsigned* count, U32 maxSymbo
     int lowS, lowN;
     U16 nodeNb = STARTNODE;
     U32 nodeRoot;
-    rankPos* rank = (rankPos *)(huffNode + 1);
-    DEBUGLOG(3, "workspace %p huffnode0 %p huffnode %p rank %p workspaceend %p", workSpace, huffNode0, huffNode, rank, ((BYTE*)workSpace) + wkspSize);
+    rankPos* rank = (rankPos *)(huffNode0 + HUF_CTABLE_WORKSPACE_SIZE / sizeof(nodeElt*));
+
     /* safety checks */
     if (((size_t)workSpace & 3) != 0) return ERROR(GENERIC);  /* must be aligned on 4-bytes boundaries */
     if (wkspSize < sizeof(huffNodeTable) + sizeof(rankPos) * 32) return ERROR(workSpace_tooSmall);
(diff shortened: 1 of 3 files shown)


lib/compress/huf_compress.c
```diff
@@ -357,16 +356,17 @@ size_t HUF_buildCTable_wksp (HUF_CElt* tree, const unsigned* count, U32 maxSymbo
     int lowS, lowN;
     U16 nodeNb = STARTNODE;
     U32 nodeRoot;
+    rankPos* rank = (rankPos *)(huffNode0 + HUF_CTABLE_WORKSPACE_SIZE / sizeof(nodeElt*));
```
felixhandte: Looks like our contbuild is failing on this line, due to potential aliasing issues. You'll need to experiment to confirm, but I imagine you can fix this by casting through `char*` or `void*`.

Maybe:

```
rankPos* rank = (rankPos*)((BYTE*)workspace + HUF_CTABLE_WORKSPACE_SIZE);
```

You should be able to check with `make clean zstd MOREFLAGS="-Wstrict-aliasing -Werror"`.

felixhandte: It might also make more sense to give this variable a slightly more descriptive name now that its been hoisted out of `HUF_sort`.

Cyan4973: > Looks like our contbuild is failing on this line, due to potential aliasing issues

Indeed, strict typing will not allow a `rankPos*` to be derived from a `nodeElt*`.

Cyan4973:
I agree with @felixhandte , 
having magic constants throughout the code is not great for its maintenance.

Here is what I suggest to deal with it:

1) Declare a new structure, containing all the tables supposed to be active in the `workspace` at the same time. Give them proper type, and proper size.

2) On receiving the `void* workspace` in `HUF_buildCTable_wksp()`, alias it to a pointer of this new structure. This is allowed, because `workspace` is `void*`.

3) Now, when passing `rankPos* rank`, just pass a pointer to the relevant member of the structure. This will calculate the offsets automatically. No need for add-hoc mathematical formula nor constants to get this right.

4) The definition of this new structure doesn't need to be public, it can stay within `huf_compress.c`. This will keep the public API as lean as it is today. Just ensure that it remains in sync with `HUF_WORKSPACE_SIZE` with a `static_assert()`.

As can be guessed, while it's not a lot of lines of code, it's definitely subtle and tricky.

MeghnaM:
@felixhandte & @Cyan4973 Thank you both for your helpful comments. I declared a new struct containing all the tables active in workspace and was able to remove the magic constant `* 32`. It's still not clear to me how I can get rid of the `+ 256` constant, maybe because I don't understand why `HUF_WORKSPACE_SIZE` was originally `6 << 10`. I would appreciate some more guidance on how to fix this.

lib/compress/huf_compress.c
```diff
@@ -425,8 +433,8 @@ size_t HUF_buildCTable_wksp (HUF_CElt* tree, const unsigned* count, U32 maxSymbo
  */
 size_t HUF_buildCTable (HUF_CElt* tree, const unsigned* count, unsigned maxSymbolValue, unsigned maxNbBits)
 {
-    huffNodeTable nodeTable;
-    return HUF_buildCTable_wksp(tree, count, maxSymbolValue, maxNbBits, nodeTable, sizeof(nodeTable));
+    U32 workspace[HUF_WORKSPACE_SIZE_U32];
```
Cyan4973: Since you defined it just above, it would be slightly proper to use `HUF_buildCTable_wksp_tables` type here.
`HUF_WORKSPACE_SIZE_U32` is merely a proxy to get there.

lib/compress/huf_compress.c
```diff
@@ -693,7 +701,7 @@ HUF_compress_internal (void* dst, size_t dstSize,
     huffLog = HUF_optimalTableLog(huffLog, srcSize, maxSymbolValue);
     {   size_t const maxBits = HUF_buildCTable_wksp(table->CTable, table->count,
                                             maxSymbolValue, huffLog,
-                                            table->nodeTable, sizeof(table->nodeTable));
+                                            table->nodeTable, ((BYTE*)workSpace) + wkspSize - (BYTE*)table->nodeTable);
```
Cyan4973: This arithmetic feels weird. What's the goal ?

lib/compress/huf_compress.c
```diff
@@ -348,10 +353,11 @@ static void HUF_sort(nodeElt* huffNode, const unsigned* count, U32 maxSymbolValu
  *  `workSpace` must be aligned on 4-bytes boundaries, and be at least as large as a table of HUF_CTABLE_WORKSPACE_SIZE_U32 unsigned.
```
Cyan4973: Update this comment. The needed size of `workSpace` is now larger : `sizeof(HUF_buildCTable_wksp_tables)`.

lib/compress/huf_compress.c
Cyan4973: I believe you need to update the definition of `HUF_compress_tables_t`,
so that the space previously reserved as `huffNodeTable nodeTable;` now becomes larger, as `HUF_buildCTable_wksp_tables buildCTable_wksp;` instead.
And then you just pass this section as the `workspace` of `HUF_buildCTable_wksp()`.

lib/compress/huf_compress.c
```diff
@@ -311,29 +311,34 @@ static U32 HUF_setMaxHeight(nodeElt* huffNode, U32 lastNonNull, U32 maxNbBits)
...
     U32 base;
     U32 current;
 } rankPos;
 
-static void HUF_sort(nodeElt* huffNode, const unsigned* count, U32 maxSymbolValue)
+typedef nodeElt huffNodeTable[HUF_CTABLE_WORKSPACE_SIZE_U32];
+
+typedef struct {
+  rankPos rankPosition[32];
+  huffNodeTable huffNodeTbl;
```
Cyan4973: minor : it feels proper to reverse positions of `huffNodeTbl` and `rankPosition` in the `struct`.
That's because `huffNodeTbl`'s lifetime is the larger one (it starts before and ends after `rankPosition`).

Cyan4973:
Thanks for the update @MeghnaM .
The new code is much clearer.
Usage of proper structure definitions provides names and automates offset calculations, making this process a lot more safer and readable.

> It's still not clear to me how I can get rid of the + 256 constant

`256` seems to be the size of the `rankPos`, and that's probably why you added it.
I guess you presumed that, since the new strategy needs 256 bytes, it has to reserve them for this dedicated use.

Well, it may end up being correct, but this is not that simple. And I believe it comes down to understanding this : 
> maybe because I don't understand why HUF_WORKSPACE_SIZE was originally 6 << 10

You are right, this is a good question.
`6 << 10`, aka  6 KB, is the maximum workspace space size needed _at any one time_ by the Huffman compressor.

The `workspace` is actually a form of "scratch buffer" : every function uses it temporarily, and then forget about its content, making the buffer available again for next function. This is consistent  : we merely moved structures and tables that were originally onto the stack, and the stack would have acted more or less the same way.

It could probably be faithfully represented by defining a more complex structure which would contain all the different parts needed at any moment by the compression process, and employing `union` to show which parts of the memory are re-used by consecutive operations.
But that's a larger change, and would expand the scope of this PR. So let's keep that for a future change.

The same memory address can be re-used multiple times by consecutive function invocations. Of course, if some result or table needs to be preserved while calling another function, it must not be part of the `workspace` that is going to be modified. 
That's why you separate in your function `rankPosition` and `huffNodeTbl`, so that, when `rankPosition` is being written into, the source data into `huffNodeTbl` is not accidentally overwritten.

Now coming back to the `+256` question :
in your case, your are modifying `HUF_buildCTable_wksp()`, so how much `workspace` to you need ?
You need `sizeof(HUF_buildCTable_wksp_tables)`.
And how large is such space ? Well, it's much less than 6 KB.

But wait, maybe `HUF_buildCTable_wksp_tables` is needed _on top of_ something else ?
Yes, indeed! The actual total amount needed is visible in `HUF_compress_tables_t`, of which `HUF_buildCTable_wksp_tables` is (or rather should be) the 3rd elements.

So that's the size needed actually. `HUF_WORKSPACE_SIZE` should be `== sizeof(HUF_compress_tables_t)`/

By increasing the size of `HUF_buildCTable_wksp_tables`, it is increasing the size of `HUF_compress_tables_t`, and therefore, increasing `HUF_WORKSPACE_SIZE`.

The simpler way would be to define `HUF_WORKSPACE_SIZE` as `sizeof(HUF_compress_tables_t)`, but that's not possible, because `HUF_WORKSPACE_SIZE` is public, while `HUF_compress_tables_t` is private.
So the next best thing is to define `HUF_WORKSPACE_SIZE` as a constant, or a formula of constants, and then [use a `static_assert`](https://github.com/facebook/zstd/blob/v1.4.3/lib/compress/huf_compress.c#L62) to ensure that this condition holds true.
The static assert will ensure that the value is validated by the compiler.

N.B: Technically, increasing `HUF_WORKSPACE_SIZE` is an ABI break, which could pause problems with existing applications (mismatch between allocated and needed space). But since this is an internal interface, we are not supposed to have any external user of this interface, this change is under control, and we don't have to worry about ABI break at this level.

MeghnaM:
@Cyan4973 I updated the code in response to your comments. The explanation you gave was really useful, thank you! Some CI checks are still failing - it seems like a package version problem unrelated to the code change. It would be great if you could please help me fix the issue.

Issue #1983, opened 2020-02-01
Build issue with make 4.3


chenming1986:
When building zstd 1.4.4 with make 4.3, multi-threading/zlib/liblzma/liblz4 support is disabled at build time. It seems that make 4.3 interprets ```'\#'``` literally rather than interpreting it as escaped ```'#'``` in the shell function. For example for multi-threading support, the compilation test on line programs\Makefile:99 always fails with make 4.3 because ```\#include <pthread.h>``` is a wrong statement, so HAVE_PTHREAD is always set to 0. The issue doesn't exist with make 4.2.1

#### programs\Makefile:99
```
HAVE_PTHREAD := $(shell printf '\#include <pthread.h>\nint main(void) { return 0; }' > have_pthread.c && $(CC) $(FLAGS) -o have_pthread$(EXT) have_pthread.c -pthread 2> $(VOID) && rm have_pthread$(EXT) && echo 1 || echo 0; rm have_pthread.c)
```

#### Build log snippet with Make 4.3:
```
......
==> no threads, building without multithreading support
==> no zlib, building zstd without .gz support
==> no liblzma, building zstd without .xz/.lzma support
==> no liblz4, building zstd without .lz4 support
......
```

#### Build log snippet with Make 4.2.1:
```
......
==> building with threading support
==> building zstd with .gz compression support
==> building zstd with .xz/.lzma compression support
==> building zstd with .lz4 compression support
......
```

davidbolvansky:
https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

Can we use __has_include? gcc/clang supports this builtin.

bimbashrestha:
It looks like the workround that gnu provides for this is pretty straightforward (https://lwn.net/Articles/810071/) 

```Makefile
If you want to write makefiles portable to both versions
assign the number sign to a variable:

    H := \#
    foo := $(shell echo '$H')
```

I prefer just doing this to the current programs/Makefile over what @davidbolvansky proposes because it will require less change:). 

I'll put up a pr in a sec and we can move forward with this solution unless someone wants to make a strong case for the other approach.

PR #1987, opened 2020-02-03
[build-issue] More portable header prefix usage (#)

bimbashrestha:
https://github.com/facebook/zstd/issues/1983

```
../make-4.3/install/bin/make
...
==> building with threading support
==> building zstd with .gz compression support
==> building zstd with .xz/.lzma compression support
...
```

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -94,11 +94,12 @@ endif
 
 VOID = /dev/null
 
-HEADER_PREFIX := \#
+# Make 4.3 doesn't support '\#' anymore (https://lwn.net/Articles/810071/)
+NUM_SYMBOL := \#
 
 # thread detection
 NO_THREAD_MSG := ==> no threads, building without multithreading support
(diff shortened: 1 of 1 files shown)


chenming1986:
Just noticed there are 2 includes on programs\Makefile:136 and it seems the second one also needs the change.
```
HAVE_LZ4 := $(shell printf '$(NUM_SYMBOL)include <lz4frame.h>\n\#include <lz4.h>\nint main(void) { return 0; }' > have_lz4.c && $(CC) $(FLAGS) -o have_lz4$(EXT) have_lz4.c -llz4 2> $(VOID) && rm have_lz4$(EXT) && echo 1 || echo 0; rm have_lz4.c)
```

bimbashrestha:
> Just noticed there are 2 includes on programs\Makefile:136 and it seems the second one also needs the change.

Ah nice catch. Will update that today. Thanks @chenming1986!

PR #1988, opened 2020-02-04
Optimize decompression

caoyzh:
The performance optimization of the modifications is described as follows: 
image
1)	The value is the average of Level 1~19, and the testfile is silesia.tar
2)	Measured with lzbench, transplanting the code of Zstd’s develop branch.
image

### DIFF
--- a/lib/compress/zstd_fast.c
+++ b/lib/compress/zstd_fast.c
@@ -80,6 +80,13 @@ ZSTD_compressBlock_fast_generic(
     }
 
     /* Main Search Loop */
+#ifdef __INTEL_COMPILER
+    /* From intel 'The vector pragma indicates that the loop should be 
+     * vectorized if it is legal to do so'. Can be used together with 
+     * #pragma ivdep (but have opted to exclude that because intel 
+     * warns against using it).*/
+    #pragma vector always
+#endif
     while (ip1 < ilimit) {   /* < instead of <=, because check at ip0+2 */
         size_t mLength;
         BYTE const* ip2 = ip0 + 2;


terrelln:
Thanks for the patch @caoyzh! I’ll start measuring the performance tomorrow.

lib/decompress/zstd_decompress_block.c
```diff
@@ -974,11 +974,20 @@ ZSTD_decompressSequences_body( ZSTD_DCtx* dctx,
...
+        BIT_reloadUnfinishedDStream(&(seqState.DStream));
+#else
+        BIT_reloadDStream(&(seqState.DStream));
+#endif
         for ( ; ; ) {
             seq_t const sequence = ZSTD_decodeSequence(&seqState, isLongOffset, ZSTD_p_noPrefetch);
             size_t const oneSeqSize = ZSTD_execSequence(op, oend, sequence, &litPtr, litEnd, prefixStart, vBase, dictEnd);
             DEBUGLOG(6, "regenerated sequence size : %u", (U32)oneSeqSize);
+#ifdef __aarch64__
+            BIT_reloadUnfinishedDStream(&(seqState.DStream));
```
terrelln: I think this is incorrect. This could cause zstd to read out of bounds on corrupted inputs, since we’re removing the bounds check. I’m only mostly certain, but I remember looking into a similar change, and realizing it is wrong.

lib/decompress/zstd_decompress_block.c
```diff
@@ -789,7 +789,7 @@ ZSTD_initFseState(ZSTD_fseState* DStatePtr, BIT_DStream_t* bitD, const ZSTD_seqS
     DStatePtr->state = BIT_readBits(bitD, DTableH->tableLog);
     DEBUGLOG(6, "ZSTD_initFseState : val=%u using %u bits",
                 (U32)DStatePtr->state, DTableH->tableLog);
-    BIT_reloadDStream(bitD);
+    BIT_reloadUnfinishedDStream(bitD);
```
terrelln: I’d be surprised if this impacts performance since it is only called once per block

lib/decompress/zstd_decompress_block.c
```diff
@@ -974,11 +974,20 @@ ZSTD_decompressSequences_body( ZSTD_DCtx* dctx,
                 BIT_DStream_endOfBuffer < BIT_DStream_completed &&
                 BIT_DStream_completed < BIT_DStream_overflow);
 
+#ifdef __aarch64__
+        BIT_reloadUnfinishedDStream(&(seqState.DStream));
```
terrelln: Same here, this shouldn’t impact performance significantly

terrelln:
What gains come from member variable reordering?

We may want to split each individual optimization into its own PR so that we can selectively merge them as they’re ready. That will probably be the easiest way to work when both sides need to do measurements.

lib/decompress/zstd_decompress_block.c
caoyzh: @terrelln Thanks. Is the test result related to the test platform or test data? I executed “make test” on aarch64 and x86, and no error was reported. Are there other methods to check whether my modification is correct?

lib/decompress/zstd_decompress_block.c
caoyzh: > What gains come from member variable reordering?
> 
> We may want to split each individual optimization into its own PR so that we can selectively merge them as they’re ready. That will probably be the easiest way to work when both sides need to do measurements.

Ok, I will split them,  and measure their gains.

lib/decompress/zstd_decompress_block.c
terrelln: This test result https://travis-ci.org/facebook/zstd/jobs/645820930?utm_medium=notification&utm_source=github_status is an out of bounds access caused by this PR, probably this line.

You can run the fuzzers following this README https://github.com/facebook/zstd/tree/dev/tests/fuzz. The following is about what you need to do to run the `simple_decompress` fuzzer which is the most relevant.

```
cd tests/fuzz
make corpora # download corpus
CC=clang CXX=clang++ ./fuzz.py build --enable-fuzzer --enable-asan --enable-ubsan
./fuzz.py libfuzzer simple_decompress
```

You should also run zstreamtest and fuzzer. In this example they run for 3 minutes after running the "unit tests". These will be a faster turn around than the fuzzers.

```
cd tests
make -j zstreamtest fuzzer
./zstreamtest -T3m
./fuzzer -T3m
```

Those should give you a reasonable amount of confidence in your changes, along with the continuous integration passing when you put up your PR. After that we have oss-fuzz running continuously on our dev branch. So if something gets past all these tests, and code review, we should easily catch it before it merges into master.

terrelln: We should probably have better documentation for running tests for open source contributors. I'll file an issue.

lib/common/bitstream.h
```diff
@@ -109,11 +109,11 @@ MEM_STATIC size_t BIT_closeCStream(BIT_CStream_t* bitC);
 *  bitStream decoding API (read backward)
 **********************************************/
 typedef struct {
-    size_t   bitContainer;
-    unsigned bitsConsumed;
     const char* ptr;
     const char* start;
     const char* limitPtr;
+    size_t   bitContainer;
+    unsigned bitsConsumed;
```
terrelln: On x86 clang-9.0.1 I measured a 4% decompression speed boost from this change alone. It was neutral for gcc.

This one is very interesting, I will look into it further.

lib/decompress/zstd_decompress_block.c
```diff
@@ -563,13 +563,13 @@ typedef struct {
...
 typedef struct {
-    BIT_DStream_t DStream;
     ZSTD_fseState stateLL;
     ZSTD_fseState stateOffb;
     ZSTD_fseState stateML;
-    size_t prevOffset[ZSTD_REP_NUM];
     const BYTE* prefixStart;
     const BYTE* dictEnd;
+    BIT_DStream_t DStream;
+    size_t prevOffset[ZSTD_REP_NUM];
```
terrelln: I've measure that this helps gcc a bit and hurts clang a bit on x86 with clang-9.0.1 and gcc-9.2.0.

PR #1991, opened 2020-02-05
Displaying level in --show-default-cparams

bimbashrestha:
As per @Cyan4973's request. The end user isn't expected to know the mapping from compression level to compression strategy so I'm went ahead and included level in the display as well.


Cyan4973:
I think there is a misunderstanding. 
What I meant is that when users select a strategy, using command `--zstd=strategy=#` , they don't get to specify a name : instead, they have to specify a value.

The mapping is specified in the [man page](https://github.com/facebook/zstd/blob/dev/programs/zstd.1.md#--zstdoptions), however, it would be nicer if users could read an information directly relevant to the command.

bimbashrestha:
> display the value of the strategy itself

Ooooh I see. Okay. Will update

bimbashrestha:
Hmm FreeBSD has been failing randomly more than usual. It's been flakey the last few times and I think this one is too. Strange...

Cyan4973:
Unfortunately, some tests are flakey from time to time.
We have to look at the cause of error, and make a judgment call : is it an actual bug, or a flakey test ? Oftentimes, it's rather obvious. Tests considered flakey can then be restarted.

PR #1992, opened 2020-02-10
CIFuzz dogfood

Leo-Neat:
OSS-Fuzz was wondering if zstd would dogfood CIFuzz,
a new feature being rolled out in OSS-Fuzz. It allows for pull
requests to be fuzzed in CI using fuzz targets enabled in OSS-Fuzz.
It will be active in dry_run mode meaning that it will never fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,22 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        project-name: 'zstd'
+        dry-run: true
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        fuzz-time: 600
+        dry-run: true
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: bug_report
+        path: ./out/bug_report


felixhandte:
That sounds nifty! I think we would be interested in such a thing.

Leo-Neat:
@felixhandte It looks like signing the facebook CLA would take some effort and time. So if you are willing, I think it would be easier if you landed it yourselves. I don't plan to make any more patches to your repo other than the one suggested. The rest of the implementation is generic and lives in the OSS-Fuzz repo.

terrelln:
Yeah, we'd definitely be interested. I'll put up a PR later today. Is there documentation for this feature somewhere?

terrelln:
@Leo-Neat I put up a bad PR in #1994 and the CIFuzz found the issue in this run https://github.com/facebook/zstd/actions/runs/37230566. But where does the artifact get uploaded?

Leo-Neat:
> @Leo-Neat I put up a bad PR in #1994 and the CIFuzz found the issue in this run https://github.com/facebook/zstd/actions/runs/37230566. But where does the artifact get uploaded?

In order for the workflow to upload failure artifacts it needs to have dry-run set to false in both the build and run actions. This is because  dry_run mode prevents failures but a failure is what triggers the arifact upload. One way to test this is set dry-run: false temporarily and re-run the same test. Then the test will fail and an artifact with a test case and stack trace will be uploaded.

terrelln:
Thanks @Leo-Neat! Assuming things go well in the next week or so we'll turn off dry-run mode. I'm really excited for this!

Leo-Neat:
@terrelln No problem! Please reach out if you have any issues or UX comments and I would be happy to address them.

PR #2003, opened 2020-02-18
Fix integer parsing in cli

philj56:
`readU32FromCharChecked()` currently fails for numbers within the nearest multiple of 10 of the limit for 32-bit unsigned ints:
```
$ zstd -M4294967295
error: numeric value too large
```
This breaks e.g. [lesspipe](https://github.com/wofr06/lesspipe). As unsigned integer 'overflow' is well-defined behaviour in C, we can safely just let it happen and check afterwards.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -254,10 +254,12 @@ static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
 {
     unsigned result = 0;
     while ((**stringPtr >='0') && (**stringPtr <='9')) {
-        unsigned const max = (((unsigned)(-1)) / 10) - 1;
+        unsigned const max = ((unsigned)(-1)) / 10;
+        unsigned last = result;
         if (result > max) return 1; /* overflow error */
         result *= 10;
         result += (unsigned)(**stringPtr - '0');
+        if (result < last) return 1; /* overflow error */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good to me.
For my understanding, how does that break `lesspipe` ?
The benefit feels limited, since only the very last numbers of `UINT` range are impacted, but that's apparently enough to break `lesspipe` ?

philj56:
Turns out it doesn't actually break `lesspipe`, there was something else wrong as well. So you're right, this fix might not actually fix anything in practice.

The issue appeared to be that `lesspipe` was passing the `1<<32 - 1` as the memory limit to `zstd`, to allow decompressing archives created with `--long=28` or longer (see [the relevant commit](https://github.com/wofr06/lesspipe/commit/87ac1d025ba3417d7703b3ca57d52da055b5b21e)). However `zstd` doesn't accept numbers greater than `1<<31` anyway, so I assume something's changed in the meantime - I've submitted a [pull request](https://github.com/wofr06/lesspipe/pull/36) on the `lesspipe` repo as well.

Cyan4973:
Thanks for details @philj56 !
Yes, your analysis looks correct.

As for the patch, it may have a small impact, but it's nonetheless valid, so I'll merge it.

PR #2004, opened 2020-02-19
Update CIFuzz .yml file

Leo-Neat:
This change necessary to keep in sync with the CIFuzz action config. Sorry for the inconvenience, and thanks for dog fooding CIFuzz!


Leo-Neat:
We were wondering if you could integrate it @terrelln Thanks for the help!

PR #2007, opened 2020-02-20
Allow some CLI options to accept numeric values up to size_t

### LINKED ISSUE #2006: --stream-size doesn't allow values greater than 4GB
When I specify a stream size in the CLI that overflows a 32-bit unsigned int, I get an error like so:
```console
$ ./zstd --stream-size=4096MB
error: numeric value too large 
```

Internally, the CLI is using `size_t` as the type to store the stream size, so I think this is just a limitation of the option parser, which only supports U32 at the moment.

I've fixed this in my fork by adding another int parser that accepts `size_t` and applied it to the four CLI options whose internal variables are of type `size_t`.  The new int parser is just a copy-paste of the old one with the types changed to `size_t`, so this definitely isn't the most elegant way to solve the problem, just the quickest!  I've confirmed it fixes my original `--stream-size` problem.

Here's the diff: https://github.com/facebook/zstd/compare/dev...lgarrison:cli-allow-sizet

If you agree that this is the right place to fix the problem, I'm happy to open a PR and iterate on the solution.

lgarrison:
This is an attempt to fix #2006, in which the `--stream-size` CLI option is limited to 4 GB (32-bit unsigned int).  For CLI variables whose internal types are already `size_t`, this allows values up to the `size_t` limit to be specified.

This is something of a quick hack to get the CLI working for my application; happy to take suggestions on how to improve it!

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -284,9 +284,54 @@ static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
  *  Will also modify `*stringPtr`, advancing it to position where it stopped reading.
  *  Note : function will exit() program if digit sequence overflows */
 static unsigned readU32FromChar(const char** stringPtr) {
-    static const char errorMsg[] = "error: numeric value too large";
+    static const char errorMsg[] = "error: numeric value overflows 32-bit unsigned int";
     unsigned result;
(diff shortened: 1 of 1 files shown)


lgarrison:
Great, thanks!

PR #2009, opened 2020-02-21
Replaced targetLen with targetLength

eyherabh:
The actual name of the option is targetLength. Using targetLen as indicated in the man page resulted in errors.


Cyan4973:
Good catch @eyherabh !

Please modify `zstd.1.md` instead,
it is the editable source from which `zstd.1` is generated.

Cyan4973:
Nevermind, that's what you did in the other PR.
Closing this one then.

Also, note that the automated request for CLA is required before a merge.

PR #2018, opened 2020-02-27
[cirrus-ci] Removing pkg -y update and using 11.3-snap instead of 11.2 

bimbashrestha:
Testing to see if this will fix the FreeBSD failures

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,7 +1,10 @@
-freebsd_instance:
-  image_family: freebsd-12-1
-
 task:
+  name: FreeBSD (shortest)
+  matrix:
+      freebsd_instance:
+        image_family: freebsd-12-1
+      freebsd_instance:
+        image_family: freebsd-11-3
   install_script: pkg install -y gmake coreutils
   script: |
     MOREFLAGS="-Werror" gmake -j all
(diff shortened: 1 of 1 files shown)


bimbashrestha:
Well, it looks like this is passing (https://cirrus-ci.com/task/4853445320507392)

It looks like the changes are using 'image_family' instead of 'image', removing the pkg -y update, and removing amd. I think this is fine

Cyan4973:
That's great @bimbashrestha !

Minor detail : it seems we had 2 tests previously, for FreeBSD12 and FreeBSD11,
and now there is only the FreeBSD12 test left.

Is that possible to have FreeBSD11 running again ?

bimbashrestha:
> Is that possible to have FreeBSD11 running again ?

Hopefully that new commit will do it. Lets see:)

bimbashrestha:
Aha. They mention that 11.3 doesn't boot properly at the moment (https://cirrus-ci.org/guide/FreeBSD/). Strange but maybe we can just use freebsd-11-3-snap.

bimbashrestha:
Okay that finally seems to do it. We should probably switch back to using 11.3 stable once they fix the boot issue but until then the snapshot should be fine

Cyan4973:
Excellent ! Thanks @bimbashrestha !

PR #2022, opened 2020-03-01
Optimize decompression by reordering the member variables in structure.

caoyzh:
Merge partial modification in #1988.
https://github.com/facebook/zstd/pull/1988#discussion_r377995805

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -109,11 +109,11 @@ MEM_STATIC size_t BIT_closeCStream(BIT_CStream_t* bitC);
 *  bitStream decoding API (read backward)
 **********************************************/
 typedef struct {
-    size_t   bitContainer;
-    unsigned bitsConsumed;
     const char* ptr;
     const char* start;
     const char* limitPtr;
+    size_t   bitContainer;
+    unsigned bitsConsumed;
 } BIT_DStream_t;
 
 typedef enum { BIT_DStream_unfinished = 0,
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -563,13 +563,13 @@ typedef struct {
 } ZSTD_fseState;
 
 typedef struct {
+    BIT_DStream_t DStream;
     ZSTD_fseState stateLL;
     ZSTD_fseState stateOffb;
     ZSTD_fseState stateML;
+    size_t prevOffset[ZSTD_REP_NUM];
     const BYTE* prefixStart;
     const BYTE* dictEnd;
-    BIT_DStream_t DStream;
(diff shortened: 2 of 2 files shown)


terrelln:
Thanks @caoyzh! I'll measure this PR shortly!

caoyzh:
@terrelln 
On the latest dev-branch, this modification is negative optimization. I'm sorry for the trouble.

terrelln:
@caoyzh sorry for the delay. I went to measure these, but found performance stability problems with gcc on x86 on the current dev branch due to issues with the decompression loop falling out of the DSB. There are 10% performance swings from changing the code alignment only. That makes it very hard to measure performance differences, since any random change can swing the performance 10%.

I need to figure out how to stabilize gcc's decompression speed...

terrelln:
I measured the speed of this patch with gcc. It looks like the gains on gcc came from the code alignment noise.

caoyzh:
> any random change can swing the performance 10%.
I measured out the same effect with gcc on x86. Decompression had a negative optimization of about 10% ,  even if I only changed the order of assigment statements in the compression.  

> It looks like the gains on gcc came from the code alignment noise.
@terrelln  I agree with you, and it is difficult to consolidate this optimization. 
It is necessary to decide whether to give up this optimization for subsequent optimization.

terrelln:
By the way, I will be measure your patches tomorrow. Sorry it's been taking so long.

> I measured out the same effect with gcc on x86. Decompression had a negative optimization of about 10% , even if I only changed the order of assignment statements in the compression.

What CPU are you testing on?

Good to know. I think it might depend on the specific architecture. I've only been able to reproduce it on an Intel i9-9900K. I haven't been able to reproduce it on a Xeon or my Macbook (also an i9) so far. It is caused by the decompression loop falling out of the DSB. You can see if you have the same problem with this command:

```
perf stat -r20 -e cycles -e instructions -e frontend_retired.dsb_miss -e idq.all_dsb_cycles_any_uops -e idq.all_mite_cycles_any_uops -e idq.all_dsb_cycles_4_uops
```

When the performance is good you will see a high `idq.all_dsb_cycles_any_uops` and a low `idq.all_mite_cycles_any_uops`. When performance is bad `idq.all_mite_cycles_any_uops` will be higher than `idq.all_dsb_cycles_any_uops`.

I've found that this patch https://github.com/terrelln/zstd/commit/01fba2874edae05cbb75c75572a757baf46400e3 does a good job of stabilizing performance on my CPU at good performance on gcc-7 through 9, and clang-9 as well.

> @terrelln I agree with you, and it is difficult to consolidate this optimization. It is necessary to decide whether to give up this optimization for subsequent optimization.

We should pause on this one for now. I just wanted to measure to see if it would help the performance stability problem.

PR #2024, opened 2020-03-02
[zstdgrep] Log zstd decompression errors to stderr

### LINKED ISSUE #2020: zstdgrep wrapper not reporting errors such as access denied
Hello.

I've noticed a basic, and important, problem with zstdgrep.
Comparing the wrapper with similar ones, such as zgrep or xzgrep, it seems to lack some logic in how to deal with errors.
Which is made obvious in the problem I've come across. When zstdcat is called by the wrapper and it exits with an error, e.g. permission denied accessing its file arguments, nothing is reported back to the user either in stdout or stderr.
Therefore as far as the user is concerned, everything went well and there was just no line matching the pattern in the input files. Which is not at all the case.

Needless to say, this is a fundamental problem with the wrapper script.

terrelln:
Don't pass `-fq` to `zstdcat` because it is already implied. This allows errors to be logged to `stderr` and fixes #2020.

### DIFF
--- a/programs/zstdgrep
+++ b/programs/zstdgrep
@@ -109,7 +109,7 @@ if [ "$#" -lt 1 ]; then
     # ... on stdin
     set -f # Disable file name generation (globbing).
     # shellcheck disable=SC2086
-    "${zcat}" -f - | "${grep}" ${grep_args} -- "${pattern}" -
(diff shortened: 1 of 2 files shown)


Cyan4973:
Would it be useful to add a test ?

PR #2026, opened 2020-03-03
Add issue templates to zstd

bimbashrestha:
LZ4 just added this. I think it would be helpful here too.


Cyan4973:
Yes, it's a good start !
We can still tweak the exact wording later if that feels appropriate.

Issue #1951, opened 2020-01-14
blow up errors on Intel Compiler on O2: _BitScanReverse return code not checked


audioMirror:
**Platform:**
Windows 10
Visual Studio 2010
Intel Compiler 2013

**Result:**
Some streams blow-up on decompress

**Cause:**
Incorrect results leading to a blow-up are created by bitstream.h when doing an 02 optimized compile:

```
MEM_STATIC unsigned BIT_highbit32 (U32 val)
{
    assert(val != 0);
    {
#   if defined(_MSC_VER)   /* Visual */
        unsigned long r=0;
        _BitScanReverse ( &r, val );
        return (unsigned) r;
...
}
```

_BitScanReverse returns success (!=0)  or failure (0).  The Microsoft definition is:

```
__success(return!=0)
BOOLEAN
_BitScanReverse (
    __out DWORD *Index,
    __in DWORD Mask
    );

```
The Zstd code assumed that 'r' will be left at zero when the success if false.  However, I don't think the intrinsic is defined that way.  The 'r' value is likely undefined on failure.  Certainly, the Intel Compiler is making this assumption during its optimization at level O2 (O1 or below don't have the bug).

This could be an Intel Compiler bug, but it is more likely a bug in the zstd code assuming that 'r' has a valid value on failure.

This code stops giving the wrong value when changed to:

```
MEM_STATIC unsigned BIT_highbit32 (U32 val)
{
    assert(val != 0);
    {
#   if defined(_MSC_VER)   /* Visual */
        unsigned long r;
        return (_BitScanReverse ( &r, val )) ? (unsigned)r : 0;
...
}
```

The following Microsoft Intrinsics are used on zstd:

_BitScanForward
_BitScanReverse
_BitScanForward64
_BitScanReverse64

All of these calls will potentially be affected by this bug, and all cased should be fixed.

Cyan4973:
Does the proposed new code impacts performance ?

audioMirror:
I believe it is unlikely to affect performance.  These methods are called when building tables, not during compression or decompression.  

There is an extra if-statement, and one less assignment, though depending on how the intrinsic is defined, the result may be the same.

I doubt I will be able to measure a difference, but I will run my own tests on a Microsoft build (I can't compare Intel, because it blows up without the fix) and let you know.

audioMirror:
I have done tests on compression and decompression using 1MB, 100KB and 10KB file sizes, at level=1, both with and without context re-use.

[...]

PR #2027, opened 2020-03-03
[ci] Adding icc make check to travis

bimbashrestha:
Looks like we can run icc on travis. It might be helpful for issues like https://github.com/facebook/zstd/issues/1951

I think just make check should be fine here. Here is the run on my fork: 
https://travis-ci.com/bimbashrestha/zstd/jobs/293671236

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ matrix:
       env:
       - C_COMPILER=icc
       - CXX_COMPILER=icpc
-      - COMPILER_FLAGS='-wd13200 -wd13203'
       install:
       - source /opt/intel/inteloneapi/compiler/latest/env/vars.sh
       addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,23 @@ env:
 matrix:
   fast_finish: true
   include:
+    - name: icc
+      compiler: icc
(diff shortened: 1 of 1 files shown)


Cyan4973:
Excellent @bimbashrestha !

It seems it also brings some new warnings.
We'll have to look at them and see if they are worth fixing.

For the time being, just a test like `make check` is already a good first step.
We may need to add more, we can see that later.

.travis.yml
```diff
@@ -25,6 +25,23 @@ env:
 matrix:
   fast_finish: true
   include:
+    - name: icc
+      compiler: icc
+      env:
+      - C_COMPILER=icc
+      - CXX_COMPILER=icpc
+      - COMPILER_FLAGS='-wd13200 -wd13203'
```
Cyan4973: Just for my information,
what is this compiler flag about ?

bimbashrestha: Ah I'm pretty sure they are to disable warnings when using MMX instructions improperly or something like that. Saw that in some other travis.yml files where they were using icc and blindly copied:/

Actually looking into it, I'm pretty sure we don't need it in zstd since we don't use the relevant calls. I'll go head and remove it.

PR #2028, opened 2020-03-05
[opt] Don't recompute initial literals price

terrelln:
The pricing of `opt[0].price` doesn't matter at all for the parsing, since every path contains `opt[0]`. I've left the `litlen` price in so we don't get negative prices when we subtract it off.

I don't expect more than any real gains, except maybe on a specifically constructed dataset, but no need to do the extra work.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -249,27 +249,6 @@ static U32 ZSTD_litLengthPrice(U32 const litLength, const optState_t* const optP
     }
 }
 
-/* ZSTD_litLengthContribution() :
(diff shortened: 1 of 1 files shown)


Cyan4973:
OK, so the main idea seems to be that
the algorithm doesn't need the full price of all literals in front of the first match,
but only the cost of the literal length,
since the cost of literals themselves forms a kind of "static" sum which will not vary whatever the decision after that point ?

terrelln:
Yup, I've added a comment saying that. We don't "need" the cost of the literal length either, but I don't want the price to be negative when we subtract it off later.

PR #2029, opened 2020-03-05
[opt] Update repcodes less often

terrelln:
* Only update repcodes upon arrival
* Use the repcodes in `opt[cur]` for the next chunk instead of recomputing them

Measurements with gcc-9.2.1 and clang-9.0.1 on enwik7.

| Level | Compiler | Before | After |
|---|---|---|---|
| 16 | gcc | 4.70 MB/s | 4.71 MB/s |
| 16 | clang | 4.63 MB/s | 4.74 MB/s |
| 19 | gcc | 2.60 MB/s | 2.69 MB/s |
| 19 | clang |2.58 MB/s | 2.60 MB/s |

I expect this to be more important in two cases:
1. There are lots of match candidates. Like level 19 enwik7.
2. We're running at a faster speed.


terrelln:
Ping

Issue #1926, opened 2019-12-10
Does zstd handle compression of changing files?


anayrat:
Hello,

I tried to use zstd to compress log files which are continuously appended. 

Execute:
```bash
#!/bin/bash
while true
do
echo "line" >> /tmp/log.txt
done
```
When I tried to compress in another session I were faced to this issue:
```
zstd --compress --stdout /tmp/log.txt > /tmp/log.txt.zst
zstd: error 27 : Read error : Incomplete read : 66205544 / 66194080 B 
zstd -t /tmp/log.txt.zst
/tmp/log.txt.zst     : 0 MB...     /tmp/log.txt.zst : Decoding error (36) : Corrupted block detected 
```

Note gzip handle this:
```
gzip  --stdout /tmp/log.txt > /tmp/log.txt.gz 
gzip: /tmp/log.txt: file size changed while zipping
gzip -v -t /tmp/log.txt.gz
/tmp/log.txt.gz:	 OK
```

Same for lz4:
```
lz4 --compress --stdout /tmp/log.txt > /tmp/log.txt.lz4
lz4 -t /tmp/log.txt.lz4                                
/tmp/log.txt.lz4     : decoded 71455396 bytes                    
```

Regards,

Cyan4973:
Hi @anayrat , 
that's indeed unexpected.

Is it reproducible ?
And is it possible to share an example that produces this outcome ?

anayrat:
> Is it reproducible ?

Yes

> And is it possible to share an example that produces this outcome ?

You just have to append lines to a file and compress in another session.

FYI, I tried with zstd 1.4.4 (archlinux) and 1.3.8 (debian package)

Cyan4973:
OK, it took me a while to figure out what the problem was.

So what you mean is, if the file _is being modified (appended) at the same time that it is being compressed_, then `zstd` produces an error at the end. This is different from `gzip` which doesn't complain.

`zstd` produces an error because it detects that the size of the file is not the one that was claimed at the beginning of the session. This is classified as a file system error.

##### Why ?

When compressing a file, `zstd` queries the size of said file, and this becomes an information _embedded into the frame header_.
Once expressed in the header, the information is controlled. It must be correct, otherwise it's labelled an encoding error. The decoder will also complain if a frame with a mismatch between header and content size is decoded.

[...]

PR #2032, opened 2020-03-09
Adding --no-content-size to the cli

bimbashrestha:
https://github.com/facebook/zstd/issues/1926

Ran 
```
#!/bin/bash
while true
do
echo "line" >> /tmp/log.txt
done
```
and
```
./zstd --compress --no-content-size --stdout /tmp/log.txt > /tmp/log.txt.zst

```

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -634,7 +634,8 @@ int main(int const argCount, const char* argv[])
         setRealTimePrio = 0,
         singleThread = 0,
         showDefaultCParams = 0,
-        ultra=0;
+        ultra=0,
+        noContentSize=0;
     double compressibility = 0.5;
     unsigned bench_nbSeconds = 3;   /* would be better if this value was synchronized from bench */
     size_t blockSize = 0;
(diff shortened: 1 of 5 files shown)


programs/zstdcli.c
```diff
@@ -748,6 +749,7 @@ int main(int const argCount, const char* argv[])
                 if (!strcmp(argument, "--priority=rt")) { setRealTimePrio = 1; continue; }
                 if (!strcmp(argument, "--output-dir-flat")) {nextArgumentIsOutDirName=1; lastCommand=1; continue; }
                 if (!strcmp(argument, "--show-default-cparams")) { showDefaultCParams = 1; continue; }
+                if (!strcmp(argument, "--no-content-size")) { noContentSize = 1; continue; }
```
Cyan4973: Consider adding both `--content-size` and `--no-content-size`, 
with `--content-size` being the default.
If both are present, the latest wins.
This can happen when accumulating options on the command line through macros and scripts.

Cyan4973:
Consider adding a line about this new command in `programs/zstd.1.md`.

Cyan4973:
I think the only explicit use case is about compressing a file which is being appended to at the same time, thus changing its size during the compression process.

Setting the size to `UNKNOWN` disable any parameter adaptation, and make the behavior similar to what providing the input through a pipe would have done.

In contrast, just disabling the content size field, using `ZSTD_CCtx_setParameter(cctx, ZSTD_c_contentSizeFlag, 0);` would have given a chance to adjust the compression parameters to the first detected size of the file. This could make a small difference for smaller files.

My own preference would be for the second case, though I have no "strong" reason for that, just some "gut feeling", essentially because it feels closer to the expressed command : "no-content-size". I suspect the difference is pretty small and won't matter much.

terrelln:
I'm in favor of keeping parameter adjustment as well.

bimbashrestha:
Ah okay then! Haden't thought about changing parameters honestly but that makes sense to me. I'll change it to unset ZSTD_c_contentSizeFlag instead of setting filesize to unknown.

bimbashrestha:
I thought I noticed Travis failing but it's okay now. Guess it was flakey:/

terrelln:
> I thought I noticed Travis failing but it's okay now. Guess it was flakey:/

Yeah, I re-ran it and it passed. Seemed to be some infra flakiness, because it was a super weird issue.

PR #2041, opened 2020-03-16
Optimize compression by using neon function.

caoyzh:
The modification points interact on each other, and the effect of testing separately is not very well.

	Average gains(level 1~19)	gcc9.2.0	clang9.0.0
	Compression			1.67%		1.23%
	Decompression			0.02%		0.36%

**Test environment**

	1)	Measured with lzbench, transplanting the code of Zstd’s develop branch. 
	2)	The testfile is silesia.tar.
					aarch64
	Cpu name			Armv8-a
	CPU(s):				128
	Memory Device			DDR4 2666 MT/s 32 GB
	Number Of Memory Devices	16

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -21,6 +21,9 @@
 ***************************************/
 #include "compiler.h"
 #include "mem.h"
+#ifdef __aarch64__
+#include "arm_neon.h"
+#endif
 #include "debug.h"                 /* assert, DEBUGLOG, RAWLOG, g_debuglevel */
 #include "error_private.h"
 #define ZSTD_STATIC_LINKING_ONLY
@@ -193,10 +196,22 @@ static const U32 OF_defaultNormLog = OF_DEFAULTNORMLOG;
 /*-*******************************************
 *  Shared functions to include for inlining
 *********************************************/
-static void ZSTD_copy8(void* dst, const void* src) { memcpy(dst, src, 8); }
+static void ZSTD_copy8(void* dst, const void* src) {
+#ifdef __aarch64__
+    vst1_u8((uint8_t*)dst, vld1_u8((const uint8_t*)src));
+#else
+    memcpy(dst, src, 8);
+#endif
+}
 
 #define COPY8(d,s) { ZSTD_copy8(d,s); d+=8; s+=8; }
-static void ZSTD_copy16(void* dst, const void* src) { memcpy(dst, src, 16); }
+static void ZSTD_copy16(void* dst, const void* src) {
+#ifdef __aarch64__
+    vst1q_u8((uint8_t*)dst, vld1q_u8((const uint8_t*)src));
+#else
+    memcpy(dst, src, 16);
+#endif
+}
 #define COPY16(d,s) { ZSTD_copy16(d,s); d+=16; s+=16; }
 
 #define WILDCOPY_OVERLENGTH 32
@@ -238,8 +253,10 @@ void ZSTD_wildcopy(void* dst, const void* src, ptrdiff_t length, ZSTD_overlap_e 
 	 * one COPY16() in the first call. Then, do two calls per loop since
 	 * at that point it is more likely to have a high trip count.
          */
+#ifndef __aarch64__
         COPY16(op, ip);
         if (op >= oend) return;
+#endif
         do {
             COPY16(op, ip);
             COPY16(op, ip);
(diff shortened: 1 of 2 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1854,6 +1854,20 @@ ZSTD_reduceTable_internal (U32* const table, U32 const size, U32 const reducerVa
 
     for (rowNb=0 ; rowNb < nbRows ; rowNb++) {
         int column;
+#ifdef __aarch64__
```
terrelln: Do you see gains from this? This gets called once every few GB after compressing 3.5 GB. I suspect that this isn't worth the complexity, unless you have a use case that proves otherwise.

lib/common/zstd_internal.h
```diff
@@ -238,8 +253,10 @@ void ZSTD_wildcopy(void* dst, const void* src, ptrdiff_t length, ZSTD_overlap_e
 	 * one COPY16() in the first call. Then, do two calls per loop since
 	 * at that point it is more likely to have a high trip count.
          */
+#ifndef __aarch64__
```
terrelln: With this change on a pixel-2 compiled with android NDK r20 I see a 5-15% regression in decompression speed.

Keeping this shortcut is important for two reasons:
1. Most copies are <= 16 bytes, so doing a 2nd copy is wasteful until they become long.
2. This shortcut separates out the branch statistics for the first iteration `(op >= oend)` is nearly always true, from the loop, which is more likely to have a high trip count.

Without this change, I see neutral compression and decompression speed on the Pixel 2. If you remove this change do you still get positive compression performance?

caoyzh: > With this change on a pixel-2 compiled with android NDK r20 I see a 5-15% regression in decompression speed.
> 
> Keeping this shortcut is important for two reasons:
> 
> 1. Most copies are <= 16 bytes, so doing a 2nd copy is wasteful until they become long.
> 2. This shortcut separates out the branch statistics for the first iteration `(op >= oend)` is nearly always true, from the loop, which is more likely to have a high trip count.
> 
> Without this change, I see neutral compression and decompression speed on the Pixel 2. If you remove this change do you still get positive compression performance?

Without this change, there is still 1.5% positive compression on gcc9.2.0 , and almost on clang9.0.0 .

terrelln: We need to keep this code on aarch64 for the pixel2 benchmarks.

We could also try:

```c
#ifdef __aarch64__
do {
    COPY16(op, ip);
while (op < oend);
#else 
COPY16(op, ip);
if (op >= oend) return;
do {
    COPY16(op, ip);
    COPY16(op, ip);
while (op < oend);
#endif
```

Basically deleting the first if branch, but shortening the loop. I haven't measured it, so I don't know if it will work, but it is an idea.

lib/compress/zstd_compress.c
terrelln: Does this actually gain anything? I wouldn't expect it to, but your use case may be unique.

caoyzh: For the modification, I tested Level1~17 with lzbench (gcc9.2.0), and only several levels were impoved.

Yikun:
We saw the [travis result link](https://travis-ci.org/github/facebook/zstd/builds/670186866) is successful, but the check result is still `pending`, it looks like there is something wrong in status refresh?

lib/common/zstd_internal.h
caoyzh: | gcc-9.2.0 | clang-9.0.0
-- | -- | --
  | Compression | Decompression | Compression | Decompression
this modifcation | 1.11% | -3.58% | 0.05% | -1.05%
this   modifcationn +      neon function of ZSTD_copy8 and ZSTD_copy16 | 1.52% | -1.14% | -0.54% | -0.44%

caoyzh:
@terrelln hi, any other suggestion on this patch? 
Let me do some explanations about this PR and our current test results.

There are 3 modifications in this PR:
1. ZSTD_copy8/ZSTD_copy16 neon optimization, we see about 0.5+% gains in gcc and clang.
2. copy loop modfication, we do some test as your suggestion, we see about 1+% gains in gcc and clang.
3. ZSTD_reduceTable_internal optimization, looks like this modification doesn't bring stable gains, so we are going to remove this modification in next update.

So, we will keep the 1st modification, and update copy loop modfication in 2nd midification, and remove 3rd modification.

Yikun:
@Cyan4973 @terrelln All changes have been refreshed in our last updates, could you take a look? Thanks.

caoyzh:
@terrelln Could you take a look again? Please let us know if we still need do some other fix or somthing before merged.  Thanks

terrelln:
Sorry for the delay. This looks good to me. I measured neutral performance on a Pixel-2 compiled with the Android NDK clang compiled with -Oz, -Os, and -O3.

PR #2050, opened 2020-03-24
Align decompress sequences loop to 32+16 bytes

terrelln:
The alignment is added before the loop, so this shouldn't hurt
performance in any case. The only way it hurts is if there is already
performance instability, and we force it to be stable but in the bad
case.

This consistently gets us into the good case with gcc-{7,8,9} on an
Intel i9-9900K and clang-9. gcc-5 is 5% worse than its best case but has
stable performance. We get consistently good behavior on my Macbook Pro
compiled with both clang and gcc-8. It ends up in the 50% from DSB and
50% from MITE case, but the performance is the same as the 85% DSB case,
so thats fine.

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -974,6 +974,47 @@ ZSTD_decompressSequences_body( ZSTD_DCtx* dctx,
                 BIT_DStream_endOfBuffer < BIT_DStream_completed &&
                 BIT_DStream_completed < BIT_DStream_overflow);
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Great investigation @terrelln !

terrelln:
It is possible I'm wrong about Skylake not showing the issue. I haven't been able to get bad performance on Skylake. But, I don't have performance counters available on my Skylake machine.

All the other machines I've verified the DSB/MITE perf counters, so I'm confident about them.

PR #2053, opened 2020-03-26
[test] Fix playTests.sh with space in binary path

terrelln:
playTests.sh didn't work when `ZSTD_BIN` or `DATAGEN_BIN` had a space in
the path name. This happens for me because I split the cmake build
directory by compiler name, like "Clang 9.0.0".

The fix is to replace all instances of `$ZSTD` with the `zstd()`
function, and the replace `$DATAGEN` with `datagen()`. This will allow
us to change how we call zstd/datagen in the future without having to
change every callsite.

Note: I changed some tmp file suffixes from `.zstd` to `.zst` because they got messed up with my find and replace.

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -7,6 +7,26 @@ die() {
     exit 1
 }
 
+datagen() {
+    "$DATAGEN_BIN" $@
+}
+
+zstd() {
+    if [ -z "$EXEC_PREFIX" ]; then
+        "$ZSTD_BIN" $@
+    else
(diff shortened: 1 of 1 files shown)


terrelln:
The FreeBSD failure seems unrelated.

Cyan4973:
> The FreeBSD failure seems unrelated.

Indeed, some issue with `gmake` installation
https://github.com/facebook/zstd/pull/2053/checks?check_run_id=538347085

PR #2055, opened 2020-03-27
Fix using uninitialized value

wosk:
This commit is to fix critical warning of probably using uninitialized fields of structures.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -845,6 +845,7 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
     assert(optLevel <= 2);
     ZSTD_rescaleFreqs(optStatePtr, (const BYTE*)src, srcSize, optLevel);
     ip += (ip==prefixStart);
+    memset(&lastSequence, 0, sizeof(lastSequence));
 
     /* Match Loop */
     while (ip < ilimit) {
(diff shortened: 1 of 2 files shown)


terrelln:
I don't believe that these two places are using uninitialized values. I'd be happy to take the fix for `lastSequence`. But `seq` is in the hot decompression loop, so we'd have to measure performance before landing that one.

Our warnings complain that you haven't initialized all members. You'll either have to switch to a `memset` or add a `0` for every field in the struct.

wosk:
@terrelln rewrote the fix.
I mean these cases:
* zstd_opt.c:1048  opt[storeEnd] = lastSequence; <- Copy probably uninit fields `price` and `rep[]` when goes here through the `goto _shortestPath`
* zstd_decompress_block.c:936 return seq; <- return probably uninit `match` field

terrelln:
@wosk these variables are undefined only when they are un-accessed. So while the fix is good style, there isn't any undefined behavior in practice. We have memory sanitizer tests/fuzzing to prove it.

terrelln:
I'm going to close this PR.

We know that these fields are unused, so leaving them undefined is fine. There is no need to waste instructions to zero them.

In fact, since we expect them to be unused, any usage of them would be a bug. By leaving them undefined Memory Sanitizer will tell us if we use them on accident. If we zero them, we will no longer get this signal from memory sanitizer.

PR #2056, opened 2020-03-29
added NeoMutt (email reader)

flatcap:
NeoMutt uses zstd compression in its email cache.
(since release 2020-03-20)

- https://neomutt.org/feature/hccompress


flatcap:
> sign our Contributor License Agreement

Done

PR #2065, opened 2020-04-03
Single-file decoder script can now (optionally) create an encoder

cwoffenden:
To complement the single-file decoder new scripts were added to create an amalgamated single-file of _all_ the Zstd source, along with examples and (simple) tests.

Tested with Xcode/Clang, Visual Studio and Emscripten. Some notes and observations:

- This keeps `zstd.h` as-is and amalgamates the rest of the code (`common`, `compress` and `decompress`). It's possible to in-line this header file too but for a library it doesn't feel like the best approach.
- Dropping the resulting files into a small project Clang was taking slightly longer (in release builds, which I believe was the LTO, but didn't investigate much because on a larger project it didn't make a difference to multi-CPU compilation). Debug builds wasn't noticeable.
- The resulting binary is smaller than with the many-file version. Around 3% smaller on both Mac and Windows. 3% isn't a lot but it's 3% for free!
- The optimised Wasm build of `roundtrip.c` (compressing then decompressing data) is an amazing 237kB. This needs further investigation (DCE and LTO are probably doing their thing but it's worth a second look).

As few changes as possible were made to the Zstd source. It's probably worth though:

- A better solution for the `CHECK_F` macro, probably moving it to the shared `fse.h` (or `error_private.h`).
- Either moving the single library to its own `contrib` folder or renaming the decoder project. It's really easy to make single-file decoder, encoder and full codec now, so the instructions and demos could be expanded to show this.


cwoffenden:
Directory renamed and suggestions implemented.

cwoffenden:
I think I'd call this finished for now. At some point I will:

- Wrap the scripts with a Makefile (being able to `clean` would be a bonus)
- Document options for reducing the binary size (e.g. the roundtrip encode/decode example when compiled with `-Oz -g0 -flto` is 332kB on Mac and 232kB for Wasm, and that's with a 32kB embedded blob for testing)
- Test and document the difference (performance, size and compile time) from using a single-file implementation

Cyan4973:
Thanks @cwoffenden ! This is excellent.
I look forward to the follow up PR.

Issue #2051, opened 2020-03-24
Resource leak


QiAnXinCodeSafe:
Stream is opened
https://github.com/facebook/zstd/blob/07ad86675429218c8cc9871be13070b2cf9e9188/contrib/diagnose_corruption/check_flipped_bits.c#L94

Opened file never closed
https://github.com/facebook/zstd/blob/07ad86675429218c8cc9871be13070b2cf9e9188/contrib/diagnose_corruption/check_flipped_bits.c#L109

Cyan4973:
@felixhandte , this is part of newly added `contrib/diagnose_corruption/check_flipped_bits` tool.

It's a minor issue easily fixed.
More interesting, consider employing the nested-resources programming pattern, where all resources are created first (including `fopen()`), then checked, then action takes place, then all resources are released (including `fopen()`). This allows the action to feature multiple exit statements without a risk to forget releasing one resource.

PR #2067, opened 2020-04-03
[bug] adding fclose before return

bimbashrestha:
Addresses https://github.com/facebook/zstd/issues/2051.


felixhandte:
Cool! While you're here, I see there's a null check just above that doesn't actually return.

Issue #2066, opened 2020-04-03
Downsize decompression buffers when too large for too long


terrelln:
On the compression side we will downsize our allocation if it is too big for too long. But we don't do this on the decompression side. See the logic here.

https://github.com/facebook/zstd/blob/16654625738d26ea8344eb4fb615bcbffd30b7f0/lib/compress/zstd_cwksp.h#L517

PR #2069, opened 2020-04-03
[memory opt] Downsize decompression buffers when large for too long

bimbashrestha:
https://github.com/facebook/zstd/issues/2066

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1512,9 +1512,7 @@ static int ZSTD_isOversized(ZSTD_DStream* zds)
 {
     size_t const neededInBuffSize = MAX(zds->fParams.blockSizeMax, 4 /* frame checksum */);
     size_t const neededOutBuffSize = ZSTD_decodingBufferSize_min(zds->fParams.windowSize, zds->fParams.frameContentSize);
-    int const inOversized = zds->inBuffSize >= neededInBuffSize * ZSTD_OVERSIZED_FACTOR;
-    int const outOversized = zds->outBuffSize >= neededOutBuffSize * ZSTD_OVERSIZED_FACTOR;
-    return inOversized || outOversized;
+    return (zds->inBuffSize + zds->outBuffSize) >= (neededInBuffSize + neededOutBuffSize) * ZSTD_OVERSIZED_FACTOR;
 }
 
 static void ZSTD_updateOversizedDuration(ZSTD_DStream* zds)
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1510,7 +1510,7 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
 
 static int ZSTD_DCtx_isOverflow(ZSTD_DStream* zds, size_t const neededInBuffSize, size_t const neededOutBuffSize)
 {
-    return (zds->inBuffSize + zds->outBuffSize) >= (neededInBuffSize + neededOutBuffSize) * ZSTD_OVERSIZED_FACTOR;
+    return (zds->inBuffSize + zds->outBuffSize) >= (neededInBuffSize + neededOutBuffSize) * ZSTD_WORKSPACETOOLARGE_FACTOR;
 }
 
 static void ZSTD_DCtx_updateOversizedDuration(ZSTD_DStream* zds, size_t const neededInBuffSize, size_t const neededOutBuffSize)
@@ -1523,7 +1523,7 @@ static void ZSTD_DCtx_updateOversizedDuration(ZSTD_DStream* zds, size_t const ne
 
 static int ZSTD_DCtx_isOversizedTooLong(ZSTD_DStream* zds)
 {
-    return zds->oversizedDuration >= ZSTD_OVERSIZED_MAXDURATION;
+    return zds->oversizedDuration >= ZSTD_WORKSPACETOOLARGE_MAXDURATION;
 }
 
 size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inBuffer* input)
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1508,6 +1508,24 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
     return length;
 }
 
+static int ZSTD_isOversized(ZSTD_DStream* zds)
+{
+    size_t const neededInBuffSize = MAX(zds->fParams.blockSizeMax, 4 /* frame checksum */);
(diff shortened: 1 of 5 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -1507,6 +1508,24 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
     return length;
 }
 
+static int ZSTD_isOversized(ZSTD_DStream* zds)
+{
+    size_t const neededInBuffSize = MAX(zds->fParams.blockSizeMax, 4 /* frame checksum */);
+    size_t const neededOutBuffSize = ZSTD_decodingBufferSize_min(zds->fParams.windowSize, zds->fParams.frameContentSize);
+    int const inOversized = zds->inBuffSize >= neededInBuffSize * ZSTD_OVERSIZED_FACTOR;
+    int const outOversized = zds->outBuffSize >= neededOutBuffSize * ZSTD_OVERSIZED_FACTOR;
```
felixhandte, line 1516: The in and out buffers are actually adjacent and part of the same underlying allocation. It probably makes more sense to treat them that way, e.g.:

```
int const oversized = (inBuffSize + outBuffSize) >= (neededInBuffSize + neededOutBuffSize) * ZSTD_OVERSIZED_FACTOR;
```

lib/decompress/zstd_decompress.c
```diff
@@ -1507,6 +1508,24 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
     return length;
 }
 
+static int ZSTD_isOversized(ZSTD_DStream* zds)
```
felixhandte, line 1511: Although these are static functions (so they won't be published outside this translation unit), I'd prefer if these function names specified that they operate on `ZSTD_DCtx`es. Because just by looking at the name they could equally well apply to `ZSTD_CCtx`es or something else.

lib/decompress/zstd_decompress.c
```diff
@@ -1633,10 +1652,15 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
             RETURN_ERROR_IF(zds->fParams.windowSize > zds->maxWindowSize,
                             frameParameter_windowTooLarge);
 
+            ZSTD_updateOversizedDuration(zds);
             /* Adapt buffer sizes to frame header instructions */
             {   size_t const neededInBuffSize = MAX(zds->fParams.blockSizeMax, 4 /* frame checksum */);
                 size_t const neededOutBuffSize = ZSTD_decodingBufferSize_min(zds->fParams.windowSize, zds->fParams.frameContentSize);
-                if ((zds->inBuffSize < neededInBuffSize) || (zds->outBuffSize < neededOutBuffSize)) {
+
```
felixhandte, line 1659: Maybe easier if you just call `ZSTD_updateOversizedDuration()` duration here and pass `neededInBuffSize` and `neededOutBuffSize` in as arguments. That way we don't calculate them again in the function body in a way that could possibly become incorrect if someone updated this calculation here but not the one in the function.

terrelln:
Please add a test, you should be able to use `ZSTD_sizeof_DCtx()` to test it.

lib/decompress/zstd_decompress_internal.h
```diff
@@ -95,6 +95,9 @@ typedef enum {
     ZSTD_use_once = 1            /* Use the dictionary once and set to ZSTD_dont_use */
 } ZSTD_dictUses_e;
 
+#define ZSTD_OVERSIZED_MAXDURATION 128
```
felixhandte, line 98: Should we just use the same macros as on the compression side? Should we find a place to put them where they can be shared?

Otherwise we should disambiguate the names so this clearly refers to dctx oversized calculation and the existing ones should be renamed to scope them to cctx management.

lib/decompress/zstd_decompress.c
```diff
@@ -1507,6 +1508,24 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
...
+    size_t const neededInBuffSize = MAX(zds->fParams.blockSizeMax, 4 /* frame checksum */);
+    size_t const neededOutBuffSize = ZSTD_decodingBufferSize_min(zds->fParams.windowSize, zds->fParams.frameContentSize);
+    int const inOversized = zds->inBuffSize >= neededInBuffSize * ZSTD_OVERSIZED_FACTOR;
+    int const outOversized = zds->outBuffSize >= neededOutBuffSize * ZSTD_OVERSIZED_FACTOR;
+    return inOversized || outOversized;
+}
+
+static void ZSTD_updateOversizedDuration(ZSTD_DStream* zds)
+{
+    zds->oversizedDuration += ZSTD_isOversized(zds) != 0;
```
terrelln, line 1522: nit: No need to be clever here, just make it an if statement.

lib/decompress/zstd_decompress.c
```diff
@@ -1507,6 +1508,24 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
     return length;
 }
 
+static int ZSTD_isOversized(ZSTD_DStream* zds)
+{
+    size_t const neededInBuffSize = MAX(zds->fParams.blockSizeMax, 4 /* frame checksum */);
+    size_t const neededOutBuffSize = ZSTD_decodingBufferSize_min(zds->fParams.windowSize, zds->fParams.frameContentSize);
+    int const inOversized = zds->inBuffSize >= neededInBuffSize * ZSTD_OVERSIZED_FACTOR;
+    int const outOversized = zds->outBuffSize >= neededOutBuffSize * ZSTD_OVERSIZED_FACTOR;
+    return inOversized || outOversized;
```
terrelln, line 1517: We allocate these buffers together. Just do:

```c
size_t const neededSize = neededInBufSize + neededOutBufSize;
size_t const currentSize = zds->inBuffSize + zds->outBuffSize;
return currentSize >= neededSize * ZSTD_OVERSIZED_FACTOR;
```

lib/decompress/zstd_decompress_internal.h
```diff
@@ -95,6 +95,9 @@ typedef enum {
...
+#define ZSTD_OVERSIZED_FACTOR 3
```
terrelln, line 99: We should share these with compression instead of redefining them.

Cyan4973:
Looks good to me. It's a clean patch, great work @bimbashrestha !
I believe the only remaining comment is from @terrelln , about a need to add a test to ensure the downsizing happens as expected (there is already such a test for `CCtx`, it could be extended to also cover `DCtx`).

And just FYI, since I did not participate on earlier discussion on this topic, I was not opposed on having separate constants for compression and decompression states, since it would allow differentiated policy if need be. But indeed, the names of constants should reflect their respective scope.

That being said, there's no urgency. We can start by using the same constants and therefore same policy for both `DCtx` and `CCtx`. We can always add new specific constants later if a needs arise.

PR #2075, opened 2020-04-07
[bug] handling case where prefix is NULL or 0 sized in refPrefix_advanced 

bimbashrestha:
Adding refPrefix to our dictionary_round_trip fuzzer caused it to crash on 

https://github.com/facebook/zstd/blob/dde98d833db693020130996de4750e8052dfcc1c/lib/compress/zstd_compress.c#L3003-L3007

handling null case on refPrefix advanced like in loadDictionary.  I'll rebase the patch-from pr on top of this after it lands.


Cyan4973:
nice catch !

PR #2080, opened 2020-04-13
minor refactor of `fileio.c`

Cyan4973:
Made a few minor improvements to `fileio.c` :

- fix or clarify a few comments
- reorder some parameters, to enforce "mutable references first" rule
- simplified `fwriteSparse()`

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -100,7 +100,7 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 #define READY_FOR_UPDATE() (!g_display_prefs.noProgress && UTIL_clockSpanMicro(g_displayClock) > g_refreshRate)
 #define DELAY_NEXT_UPDATE() { g_displayClock = UTIL_getTime(); }
 #define DISPLAYUPDATE(l, ...) {                              \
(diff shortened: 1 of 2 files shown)


programs/fileio.c
```diff
@@ -1864,34 +1866,36 @@ FIO_zstdErrorHelp(const FIO_prefs_t* const prefs,
         unsigned long long const windowSize = header.windowSize;
         unsigned const windowLog = FIO_highbit64(windowSize) + ((windowSize & (windowSize - 1)) != 0);
         assert(prefs->memLimit > 0);
-        DISPLAYLEVEL(1, "%s : Window size larger than maximum : %llu > %u\n",
+        DISPLAYLEVEL(1, "%s : Window size larger than maximum : %llu > %u \n",
```
terrelln, line 1869: Why put a space here?

Cyan4973: It helps for the specific case where
the screen is already filled with characters on the same line,
whatever the reason (garbage, or other running process for example).
The final space character will help determine that this is the end of the line, 
and more importantly in this case the end of the numerical field.

It has helped me a few times in the past, so I tend to make it a habit nowadays.

PR #2082, opened 2020-04-20
Add nim_zstd

wltsmrz:


felixhandte:
Cool, thanks for your work to make zstd available to another community!

PR #2086, opened 2020-04-22
[bug] passing cParams.hashLog instead of comprParams->hashLog for ldm trigger

bimbashrestha:
Nit: looks like VS code also made some other whitespace changes so unfortunately the second commit isn't clean.


terrelln:
The code and tests look goo to me.

What was the previous behavior, and how did you catch this bug?

bimbashrestha:
Ah yeah should have linked: See https://github.com/facebook/zstd/pull/1959#issuecomment-617497982 for context @terrelln

PR #2087, opened 2020-04-22
[fuzz+bug] running mt compression every time when fuzzing (even on small files)

bimbashrestha:
Originally, I thought the first commit exposed a bug but it turns but it turns out there was another place where mt was being manually turned off when the input size <= ZSTDMT_JOBSIZE_MIN. So we needed another ifndef there.

Still, I don't think it would hurt to have this new path exposed to the fuzzer.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2813,7 +2813,7 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
 
         ZSTD_overflowCorrectIfNeeded(ms, ws, params, ip, ichunk);
 
-        if (params->ldmParams.enableLdm && ls != NULL && srcSize >= params->ldmParams.minMatchLength)
+        if (params->ldmParams.enableLdm && ls != NULL && srcSize > params->ldmParams.minMatchLength)
             ZSTD_ldm_fillHashTable(ls, (const BYTE*)src, (const BYTE*)src + srcSize, &params->ldmParams);
 
         switch(params->cParams.strategy)
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3950,9 +3950,14 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
 
 
 #ifdef ZSTD_MULTITHREAD
+
+/* Invoke fuzzing every time for mt even on small files */
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
         if ((cctx->pledgedSrcSizePlusOne-1) <= ZSTDMT_JOBSIZE_MIN) {
             params.nbWorkers = 0; /* do not invoke multi-threading when src size is too small */
         }
+#endif
+
         if (params.nbWorkers > 0) {
             /* mt context creation */
             if (cctx->mtctx == NULL) {
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1261,7 +1261,9 @@ static size_t ZSTDMT_compress_advanced_internal(
     unsigned const compressWithinDst = (dstCapacity >= ZSTD_compressBound(srcSize)) ? nbJobs : (unsigned)(dstCapacity / ZSTD_compressBound(avgJobSize));  /* presumes avgJobSize >= 256 KB, which should be the case */
     size_t frameStartPos = 0, dstBufferPos = 0;
     assert(jobParams.nbWorkers == 0);
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
     assert(mtctx->cctxPool->totalCCtx == params.nbWorkers);
+#endif
 
     params.jobSize = (U32)avgJobSize;
     DEBUGLOG(4, "ZSTDMT_compress_advanced_internal: nbJobs=%2u (rawSize=%u bytes; fixedSize=%u) ",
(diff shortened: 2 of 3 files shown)


bimbashrestha:
Oh it looks like the fuzzer found a heap-overflow. Is there a way to download the files from CIFuzz?

lib/compress/zstdmt_compress.c
```diff
@@ -1417,7 +1419,10 @@ size_t ZSTDMT_initCStream_internal(
     if (params.jobSize != 0 && params.jobSize < ZSTDMT_JOBSIZE_MIN) params.jobSize = ZSTDMT_JOBSIZE_MIN;
     if (params.jobSize > (size_t)ZSTDMT_JOBSIZE_MAX) params.jobSize = (size_t)ZSTDMT_JOBSIZE_MAX;
 
+/* always trigger mt when fuzzing */
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
terrelln, line 1423: If it was fixed when this ifdef was added, does that mean that there is a bug in `singleBlockingThread` mode then?

We should either find and fix that bug, or delete single blocking thread mode.

lib/compress/zstdmt_compress.c
```diff
@@ -1261,7 +1261,9 @@ static size_t ZSTDMT_compress_advanced_internal(
     unsigned const compressWithinDst = (dstCapacity >= ZSTD_compressBound(srcSize)) ? nbJobs : (unsigned)(dstCapacity / ZSTD_compressBound(avgJobSize));  /* presumes avgJobSize >= 256 KB, which should be the case */
     size_t frameStartPos = 0, dstBufferPos = 0;
     assert(jobParams.nbWorkers == 0);
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
terrelln, line 1264: Look 3 lines above (line 1261). I don't know what exactly `compressWithinDst` is used for off the top of my head, but the comment implies that it may need to be adjusted.

lib/compress/zstdmt_compress.c
bimbashrestha: Hmm maybe, what is `singleBlockingThread` mode for again? For long mode to run before the other compression jobs?

terrelln: No, `singleBlockingThread` is "single-threaded" mode inside of zstdmt. It is the equivalent of switching to single threaded for inputs `< ZSTDMT_JOBSIZE_MIN`.

Originally zstdmt was its own API, and handled automatic transition back to single threaded mode. Now it is wrapped by the zstd advanced API, and the zstd streaming API handles that transition.

That means `singleBlockingThread` is mostly dead code. The only use case is if people are using the `ZSTDMT_*` API directly. `python-zstandard` used to use this API, but they've switched away. They were the only user I was aware of, so maybe we can remove this functionality.

I'd say lets try to fix it, and see what @Cyan4973 thinks about deleting it.

lib/compress/zstd_compress.c
```diff
@@ -3950,9 +3950,14 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
 
 
 #ifdef ZSTD_MULTITHREAD
+
+/* Invoke fuzzing every time for mt even on small files */
```
Cyan4973, line 3954: I may have missed something here.

This code is defined to not use the ZSTDMT path when input size is too small.
What do we gain by forcing the fuzzer to test a path that is never used in production?

lib/compress/zstdmt_compress.c
Cyan4973: `compressWithinDst` means that the compress output will be written directly into `dst`, instead of using some additional intermediate buffer.
In which case, it is indeed extremely important that `avgJobSize >= 256 KB`.
This is required so that `sum(ZSTD_compressBound(jobSize)) <= ZSTD_compressBound(inputSize)`.
If this condition is not respected, some output could be written out of bound.

lib/compress/zstdmt_compress.c
bimbashrestha: I'll remove it in a different pr

lib/compress/zstd_compress.c
bimbashrestha: The original rationale was to get it to catch the refPrefix bug from https://github.com/facebook/zstd/pull/2088 but it didn't do that

>What do we gain by forcing the fuzzer to test a path that is never used in production?

Well, it exposed the heap-overflow above. I suppose that code path is one that should already be covered but this new ifndef caught it immediately where as without it, we hadn't caught it so far. The rationale here is that by forcing the fuzzer into more paths than are used in production, we might give it more opportunities to find bugs that are in production code paths.

Cyan4973: This is not about using a code path which is "rarely" used in production, which would be nonetheless worth fuzzing for extended coverage, just in case one scenario might actually trigger it.

This is about a code path which is _expressly forbidden_, and therefore _impossible_ to reach.

There are conditions for this function to work. They are safely checked. Why being surprised that it doesn't work anymore once the protections are removed ?

To give another example, one doesn't just test `memcpy()` with overlapping buffers and conclude that it's buggy: tests still have to respect the working conditions of the tested functions,  and `memcpy()` interface expressly forbids such a use case.
This is the same here.

terrelln: @Cyan4973 I think it is worth fixing the conditions to ensure we can run in multithreaded mode with smaller inputs. Larger inputs are slower to fuzz, and are not provided as often by oss-fuzz. Making it easier for the fuzzer to get into this code path should improve our coverage of zstdmt, and hopefully find real bugs.

bimbashrestha:
I'm going to close this. Maybe at some point this will be worth looking at again, but I think we've decided it isn't the way forward for now

PR #2089, opened 2020-04-22
[bench] Extending largeNbDicts to compression 

bimbashrestha:
Benchmark compression (default) with -z now 
```
 ./largeNbDicts [Options] filename(s)

Options :
-z          : benchmark compression (default)
-d          : benchmark decompression
-r          : recursively load all files in subdirectories (default: off)
-B#         : split input into blocks of size # (default: no split)
-#          : use compression level # (default: 3)
-D #        : use # as a dictionary (default: create one)
-i#         : nb benchmark rounds (default: 6)
--nbBlocks=#: use # blocks for bench (default: one per file)
--nbDicts=# : create # dictionaries for bench (default: one per block)
-h          : help (this text)
```

Sanity check: largeNbDicts with one dict benchmarks the same as hot compression with -b

```
zstd -b6 -i1 -D ~/data/github_dict -r ~/data/github
 6# 9114 files       :   7484607 ->    723641 (10.34),  96.3 MB/s ,1146.7 MB/s

./largeNbDicts -6 -z -D ~/data/github_dict -r ~/data/github --nbDicts=1 
loading 9114 files...
created src buffer of size 7.1 MB
split input into 9114 blocks
loading dictionary /home/bimbashrestha/data/github_dict
compressing at level 6 without dictionary : Ratio=3.03  (2471189 bytes)
compressed using a 112640 bytes dictionary : Ratio=10.34  (723641 bytes)
generating 1 dictionaries, using 0.5 MB of memory
Compression Speed : 96.1 MB/s
```

and then performance drops steadily off as expected: 

```
./largeNbDicts -6 -z -D ~/data/github_dict -r ~/data/github --nbDicts=1 
...
Compression Speed : 96.1 MB/s


./largeNbDicts -6 -z -D ~/data/github_dict -r ~/data/github --nbDicts=10
...
Compression Speed : 84.0 MB/s

./largeNbDicts -6 -z -D ~/data/github_dict -r ~/data/github --nbDicts=30
...
Compression Speed : 77.7 MB/s

./largeNbDicts -6 -z -D ~/data/github_dict -r ~/data/github --nbDicts=100
...
Compression Speed : 71.4 MB/s
```
and so on

### DIFF
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -415,7 +415,7 @@ static void freeCDictCollection(cdict_collection_t cdictc)
 /* returns .buffers=NULL if operation fails */
 static cdict_collection_t createCDictCollection(const void* dictBuffer, size_t dictSize, size_t nbCDict, int cLevel)
 {
-    ZSTD_CDict** const cdicts = malloc(nbCDict * sizeof(ZSTD_DDict*));
+    ZSTD_CDict** const cdicts = malloc(nbCDict * sizeof(ZSTD_CDict*));
     assert(cdicts != NULL);
     if (cdicts==NULL) return kNullCDictCollection;
     for (size_t dictNb=0; dictNb < nbCDict; dictNb++) {
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -419,7 +419,7 @@ static cdict_collection_t createCDictCollection(const void* dictBuffer, size_t d
     if (cdicts==NULL) return kNullCDictCollection;
     for (size_t dictNb=0; dictNb < nbCDict; dictNb++) {
         cdicts[dictNb] = ZSTD_createCDict(dictBuffer, dictSize, cLevel);
-        assert(cdicts[dictNb] != NULL);
+        CONTROL(cdicts[dictNb] != NULL);
     }
     cdict_collection_t cdictc;
     cdictc.cdicts = cdicts;
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -334,6 +334,18 @@ createBufferCollection_fromSliceCollectionSizes(slice_collection_t sc)
     return result;
 }
 
+static buffer_collection_t
+createBufferCollection_fromSliceCollection(slice_collection_t sc)
+{
+    buffer_collection_t result = createBufferCollection_fromSliceCollectionSizes(sc);
+    for (size_t i = 0; i < sc.nbSlices; i++) {
+        if (result.slices.slicePtrs[i] != NULL) {
+            memcpy(result.slices.slicePtrs[i], sc.slicePtrs[i], sc.capacities[i]);
+            result.slices.capacities[i] = sc.capacities[i];
+        }
+    }
+    return result;
+}
 
 /* @return : kBuffNull if any error */
 static buffer_collection_t
@@ -776,15 +788,9 @@ int bench(const char** fileNameTable, unsigned nbFiles,
 
         shuffleCDictionaries(cdictionaries);
 
-        buffer_collection_t resultCollection = createBufferCollection_fromSliceCollectionSizes(srcSlices);
+        buffer_collection_t resultCollection = createBufferCollection_fromSliceCollection(srcSlices);
         CONTROL(resultCollection.buffer.ptr != NULL);
 
-        for (size_t i = 0; i < srcSlices.nbSlices; i++) {
-            if (resultCollection.slices.slicePtrs[i] != NULL) {
-                memcpy(resultCollection.slices.slicePtrs[i], srcSlices.slicePtrs[i], srcSlices.capacities[i]);
-                resultCollection.slices.capacities[i] = srcSlices.capacities[i];
-            }
-        }
         result = benchMem(dstSlices, resultCollection.slices, ddictionaries, cdictionaries, nbRounds, benchCompression);
 
         freeBufferCollection(resultCollection);
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -582,6 +582,22 @@ void freeDecompressInstructions(decompressInstructions di)
 }
 
 /* benched function */
+size_t compress(const void* src, size_t srcSize, void* dst, size_t dstCapacity, void* payload)
+{
+    compressInstructions* const ci = (compressInstructions*) payload;
+
+    size_t const result = ZSTD_compress_usingCDict(ci->cctx,
+                                        dst, dstCapacity,
+                                        src, srcSize,
+                                        ci->dictionaries.cdicts[ci->dictNb]);
+
+    ci->dictNb = ci->dictNb + 1;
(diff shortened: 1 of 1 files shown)


contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -662,25 +761,53 @@ int bench(const char** fileNameTable, unsigned nbFiles,
...
+        size_t const dictMem = ZSTD_estimateCDictSize(dictBuffer.size, ZSTD_dlm_byCopy);
+        size_t const allDictMem = dictMem * nbDicts;
+        DISPLAYLEVEL(3, "generating %u dictionaries, using %.1f MB of memory \n",
+                        nbDicts, (double)allDictMem / (1 MB));
 
-    buffer_collection_t resultCollection = createBufferCollection_fromSliceCollectionSizes(srcSlices);
-    CONTROL(resultCollection.buffer.ptr != NULL);
+        shuffleCDictionaries(cdictionaries);
+
+        buffer_collection_t resultCollection = createBufferCollection_fromSliceCollectionSizes(srcSlices);
```
terrelln, line 780: This should be a function `createBufferCollection_fromSliceCollection()` that allocs + memcpys the data.

contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -485,6 +535,29 @@ static size_t compressBlocks(size_t* cSizes,   /* optional (can be NULL). If pre
...
+    size_t nbDicts;
+    size_t dictNb;
+    cdict_collection_t dictionaries;
+} compressInstructions;
+
+compressInstructions createCompressInstructions(cdict_collection_t dictionaries)
+{
+    compressInstructions ci;
+    ci.cctx = ZSTD_createCCtx();
+    assert(ci.cctx != NULL);
```
terrelln, line 549: `CONTROL()`

contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -397,6 +397,37 @@ typedef struct {
...
+        ZSTD_freeCDict(cdictc.cdicts[dictNb]);
+    }
+    free(cdictc.cdicts);
+}
+
+/* returns .buffers=NULL if operation fails */
+static cdict_collection_t createCDictCollection(const void* dictBuffer, size_t dictSize, size_t nbCDict, int cLevel)
+{
+    ZSTD_CDict** const cdicts = malloc(nbCDict * sizeof(ZSTD_CDict*));
+    assert(cdicts != NULL);
```
terrelln, line 419: `CONTROL()` here and below

Cyan4973: That's a subtle one :
`CONTROL()` would be necessary if the case `cdicts==NULL` wasn't handled afterwards.

But it is.
In which case, I guess even the `assert()` is superfluous.

contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -397,6 +397,37 @@ typedef struct {
...
+    if (cdicts==NULL) return kNullCDictCollection;
+    for (size_t dictNb=0; dictNb < nbCDict; dictNb++) {
+        cdicts[dictNb] = ZSTD_createCDict(dictBuffer, dictSize, cLevel);
+        assert(cdicts[dictNb] != NULL);
```
Cyan4973, line 423: In contrast, here, the condition `cdicts[dictNb] != NULL` is _not handled_ by the code afterwards.

Therefore, a `CONTROL()` (somehow equivalent to `assert()` but which is _never_ disabled) becomes necessary.

contrib/largeNbDicts/largeNbDicts.c
Cyan4973: And yes, here also, `CONTROL()` is preferable, since nothing handles the case `ci.cctx == NULL` afterwards.

contrib/largeNbDicts/largeNbDicts.c
terrelln: Yeah, but it should be a control, since we `CONTROL()` for dictionary creation failures. It seems strange to sometimes return `NULL` and sometimes crash depending on which malloc fails. We don't gain anything from returning `NULL` from the first one.

bimbashrestha:
@Cyan4973 or @terrelln Maybe a silly question: do you know how to fix the static analyzer complaint on travis? I'm not quite sure what to do about it?

contrib/largeNbDicts/largeNbDicts.c
Cyan4973: Yeah, right, so this is a matter of consistency now.

It depends on the end goal, do we want to lean towards a `CONTROL()` free code, which would require to chase down those `CONTROL()`, eventually in multiple rounds, and progressively replace them with proper return code, followed by appropriate error detection code at caller's side ?

Such a goal would be a necessity in a library code like `libzstd`.
But here, this is just a test tool.
So it doesn't seem really necessary.

Using `CONTROL()`simplifies the logic, because the caller can assume that the function never fails.

So, if I understand your point correctly, you believe that for consistency sake, rather than using `CONTROL()` as a "backup case", whenever an error case is not handled, the code should actively make use of it, and always use `CONTROL()` whenever applicable (there are still error cases that we want to detect and report, so it's not about pushing `CONTROL()` mindlessly everywhere).

terrelln: Yeah, I mean that this function both returns `NULL` and uses `CONTROL()`. That doesn't really make sense to me. Any particular function should either return an error code or use `CONTROL()`.

When a function uses both, we don't get the benefit of saying "we never crash", and we also have the burden of checking the return code, because not all conditions are `CONTROL()`ed.

bimbashrestha:
> You probably need to reproduce the static analyzer result in order to understand the diagnosis.
> My suspicion is that the analyzer doesn't understand that `createBufferCollection_fromSliceCollectionSizes()` is necessarily successful. Or it may not understand that `sc.nbSlices` is the same value as `result.slices.nbSlices`.
> A more detailed report will help understand the reason for this analyzer warning, and help deducting the appropriate mitigation.

Yeah, that is unfortunately the other problem I've been running into. I haven't been able to reproduce this on my local machine with scan-build installed (gcc and clang). Hence the attempts to fix it directly on CI...

contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -337,13 +337,37 @@ createBufferCollection_fromSliceCollectionSizes(slice_collection_t sc)
...
+        memcpy(slices[i], sc.slicePtrs[i], sc.capacities[i]);
+        capacities[i] = sc.capacities[i];
     }
+
+    buffer_collection_t result;
+    result.buffer = buffer;
+    result.slices.nbSlices = nbSlices;
+    result.slices.capacities = capacities;
+    result.slices.slicePtrs = slices;
+
```
bimbashrestha, line 370: Just copied this from the other function. This isn't ideal but it fixes the static_analyzer complaint (which I couldn't reproduce on my local machine) so I think it's fine.

Cyan4973:
>  I haven't been able to reproduce this on my local machine with scan-build installed (gcc and clang)

This is most likely a problem of version.

The problem with static analyzers is that each version tend to generate a different set of warnings. Therefore, in order to reproduce the same result, one would need the exact same version.

Issue #2064, opened 2020-04-03
zstdgrep parsing of grep options naively broken


trantor:
**Describe the bug**
zstdgrep as a wrapper for `zstdcat | grep` should handle grep options properly.
It doesn't in several cases. A few examples below.

**To Reproduce**
```
# The fact that the -e grep option can be specified multiple times
# for multiple pattern is not handled at all, nor is the use of equivalent
# long options with or without = between them and their arguments

$ echo start > /tmp/file
$ echo stop >> /tmp/file
$ grep -e start -e stop /tmp/file
start
stop
$ grep --regexp=start --regexp=stop /tmp/file
start
stop
$ grep --regexp start --regexp stop /tmp/file
start
stop
$ grep --regexp start --regexp stop -e start /tmp/file
start
stop
$ grep --regexp start --regexp=stop -e start /tmp/file
start
stop
$ zstdgrep --regexp=start --regexp=stop /tmp/file
/tmp/file:start
/tmp/file:stop
Incorrect parameters
$ zstdgrep --regexp start --regexp stop /tmp/file
zstd: can't stat --regexp : No such file or directory -- ignored 
grep: start: No such file or directory
zstd: can't stat stop : No such file or directory -- ignored 
grep: start: No such file or directory
grep: start: No such file or directory
$ zstdgrep -e start -e stop /tmp/file
zstd: can't stat -e : No such file or directory -- ignored 
zstd: can't stat stop : No such file or directory -- ignored 
/tmp/file:start

# --file= and -f with different behaviours for
# -, /dev/fd/0, /dev/stdin as pattern files 
# also regarding when to print the name of the examined file

$ echo start | grep -f - /tmp/file
start
$ echo start | zstdgrep -f - /tmp/file
$ echo start | grep --file=- /tmp/file
start
$ echo start | zstdgrep --file=- /tmp/file
/tmp/file:start

# --file= and -f with different behaviours
# Process substitutions with bash not working as with grep on its own

$ cp  /tmp/file /tmp/file2
$ grep --file=<( echo start ) /tmp/file /tmp/file2
/tmp/file:start
/tmp/file2:start
$ grep -f <( echo start ) /tmp/file /tmp/file2
/tmp/file:start
/tmp/file2:start
$ zstdgrep --file=<( echo start ) /tmp/file /tmp/file2
/tmp/file:start
/tmp/file2:start
$ zstdgrep -f <( echo start ) /tmp/file /tmp/file2
/tmp/file:start

```

[...]

PR #2095, opened 2020-04-28
[bugs] zstdgrep/grep inconsistencies

bimbashrestha:
Fixing zstdgrep inconsistencies listed on https://github.com/facebook/zstd/issues/2064

See each commit message for details

* Fix multiple -e
* Fix pipe in with -f
* Fix "incorrect parameters" output on regexp
* Don't show filename on single file (ie: no to /tmp/file:start and yes to start <- grep does this)
* Fixing --regexp giving lots of irelevant complaints

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -254,6 +254,38 @@ ZCAT=./zstdcat $ZSTDGREP 2>&1 "1234" tmp_grep_bad.zst && die "Should have failed
 ZCAT=./zstdcat $ZSTDGREP 2>&1 "1234" tmp_grep_bad.zst | grep "No such file or directory" || true
 rm -f tmp_grep*
 
+println "\n===> zstdgrep pipe in with -f "
+echo "start" > tmp_grep
+echo "stop" >> tmp_grep
+echo "start" | ZCAT=./zstdcat $ZSTDGREP -f - tmp_grep > tmp_grep_out1
+echo "start" | grep -f - tmp_grep > tmp_grep_out2
+$DIFF tmp_grep_out1 tmp_grep_out2
+rm -f tmp_grep*
+
+println "\n===> zstdgrep --regexp= multiple"
+echo "start" > tmp_grep
+echo "stop" >> tmp_grep
+ZCAT=./zstdcat $ZSTDGREP --regexp=start --regexp=stop tmp_grep > tmp_grep_out1
+grep -e start -e stop tmp_grep > tmp_grep_out2
+$DIFF tmp_grep_out1 tmp_grep_out2
+rm -f tmp_grep*
+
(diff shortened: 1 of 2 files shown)


Cyan4973:
Could we add some test case for this new capability ?

bimbashrestha:
FreeBSD failed saying it can't find zstdcat? I'm going to re-run it just to make sure it's not just being strange

bimbashrestha:
Nope, that's a real error:(

tests/playTests.sh
```diff
@@ -254,6 +254,14 @@ ZCAT=./zstdcat $ZSTDGREP 2>&1 "1234" tmp_grep_bad.zst && die "Should have failed
 ZCAT=./zstdcat $ZSTDGREP 2>&1 "1234" tmp_grep_bad.zst | grep "No such file or directory" || true
 rm -f tmp_grep*
 
+println "\n===> zstdgrep mulitple -e"
```
Cyan4973, line 257: nit : `mulitple`

Cyan4973:
One may need to setup a local FreeBSD VM to reproduce the issue and better diagnose it.

bimbashrestha:
> One may need to setup a local FreeBSD VM to reproduce the issue and better diagnose it.

I'll try that quick guess fix first. If that doesn't work, I'll setup of a local VM:/

bimbashrestha:
Okay that seemed to fix it. It was just a matter of adding ZCAT= in front. I'm going to include some other fixes from that issue in this PR too

bimbashrestha:
I tested zstdgrep on the list from the issue and these changes take care of everything but the bash stuff at the end. These: 

```
# Process substitutions with bash not working as with grep on its own

$ cp  /tmp/file /tmp/file2
$ grep --file=<( echo start ) /tmp/file /tmp/file2
/tmp/file:start
/tmp/file2:start
$ grep -f <( echo start ) /tmp/file /tmp/file2
/tmp/file:start
/tmp/file2:start
$ zstdgrep --file=<( echo start ) /tmp/file /tmp/file2
/tmp/file:start
/tmp/file2:start
$ zstdgrep -f <( echo start ) /tmp/file /tmp/file2
/tmp/file:start
```

bimbashrestha:
Okay, I think this is ready to merge. The only unsupported functionality are the the bash commands listed at the end of that issue but it looks like zgrep produces strange results for those too (so I don't think its worth trying to fix them for zstdgrep).

Cyan4973:
Could we add a few tests showing off the new capabilities ?

bimbashrestha:
Ah it looks like I deleted the changes to playTests when I force pushed. Yes. I"ll add those back

Cyan4973:
Looks good to me !

PR #2097, opened 2020-04-29
Fix underlinked libzstd

### LINKED ISSUE #2045: 'make -C lib lib-mt' produces underlinked `libzstd.so`
Original build failure was reported as a `gcc-10` build failure when `zstd` is installed in system. Shorter version of underlying cause is https://bugs.gentoo.org/713940.

The simplest way to reproduce it is to try to build multi-threaded `libzstd.so` with undefined symbols forbidden (there should be no undefined symbols in non-plugin libraries anyway).

I injected `-Wl,--no-undefined` manually as I don't see a nice way to append to existing LDFLAGS:

```diff
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -180,11 +180,13 @@ endif
 libzstd : $(LIBZSTD)

 libzstd-mt : CPPFLAGS += -DZSTD_MULTITHREAD
+libzstd-mt : LDFLAGS  += -Wl,--no-undefined
 libzstd-mt : libzstd

 lib: libzstd.a libzstd

 lib-mt: CPPFLAGS += -DZSTD_MULTITHREAD
+lib-mt: LDFLAGS  += -Wl,--no-undefined
 lib-mt: lib

 lib-release lib-release-mt: DEBUGFLAGS :=
```

Now build failure shows underlinked library:

```
$ LANG=C make -C lib lib-mt
compiling dynamic library 1.4.5
...
/usr/lib/gcc/x86_64-pc-linux-gnu/10.0.1/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/ccC81zhY.o: in function `POOL_free.part.0':
pool.c:(.text+0x16f): undefined reference to `pthread_join'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.0.1/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/ccC81zhY.o: in function `POOL_create_advanced':
pool.c:(.text+0x322): undefined reference to `pthread_create'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.0.1/../../../../x86_64-pc-linux-gnu/bin/ld: /tmp/ccC81zhY.o: in function `POOL_resize':
pool.c:(.text+0x4e1): undefined reference to `pthread_create'
make: *** [Makefile:172: libzstd.so.1.4.5] Error 1
```

 - OS: Linux
 - Compiler: gcc-10
 - Flags: defaults, multithreaded
 - Build system: Makefile

Reading https://github.com/facebook/zstd/blob/dev/lib/README.md#multithreading-support it looks like build system is supposed to pass `-pthread` when building/linking `libzstd.so`, but it does not:

```
Multithreading support

Multithreading is disabled by default when building with make. Enabling multithreading requires 2 conditions :

    set build macro ZSTD_MULTITHREAD (-DZSTD_MULTITHREAD for gcc)
    for POSIX systems : compile with pthread (-pthread compilation flag for gcc)

Both conditions are automatically applied when invoking make lib-mt target.

...
```

The following fix seems to be enough to fix underlinking:

```diff
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -187,6 +187,8 @@ lib: libzstd.a libzstd
 lib-mt: CPPFLAGS += -DZSTD_MULTITHREAD
 lib-mt: lib

+lib-mt: LDFLAGS  += -pthread
+
 lib-release lib-release-mt: DEBUGFLAGS :=
 lib-release: lib
 lib-release-mt: lib-mt
```

Cyan4973:
fix #2045
When compiling `libzstd` in multithreading mode,
the `libzstd-mt` recipe would not include `-pthread`,
resulting in an underlinked dynamic library.
As a consequence, user programs had to add `-pthread` during their own build.

Fixed it, and added a test on TravisCI to check that `libzstd` is fully linked.

This makes it possible, in some future release,
to provide a multi-threaded `libzstd` dynamic library by default
as it would no longer impact the build script of user programs.

Added another test, which show that compiling a user program (`zbufftest-dll`) linked to multi-threaded `libzstd` doesn't require `-pthread` on user side.

Thanks to @trofi for underlining the issue and solution.

note : not directly related to this PR, but these tests made me realize that almost all our executable targets can't be built by linking them to the `libzstd` dynamic library, due to usage of non-exposed symbols. They require building from source, or from the static library. 
We may want to consider if this is a scenario we would like to support in the future, and if yes for which executable(s).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,9 @@ matrix:
         - CC=gcc-7 CFLAGS=-Werror make -j all
         - make clean
         - LDFLAGS=-Wl,--no-undefined make -C lib libzstd-mt
+        - make -C tests zbufftest-dll
+        # LDFLAGS=-Wl,--no-undefined : will make the linker fail if dll is underlinked
+        # zbufftest-dll : test that a user program can link to multi-threaded libzstd without specifying -pthread
 
     - name: gcc-8 + ASan + UBSan + Test Zstd   # ~6.5mn
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,12 +56,14 @@ matrix:
         # DEVNULLRIGHTS : will request sudo rights to test permissions on /dev/null
         - DEVNULLRIGHTS=test make test
 
-    - name: gcc-6 + gcc-7 compilation   # ~ 3mn
+    - name: gcc-6 + gcc-7 + libzstdmt compilation   # ~ 6mn
       script:
         - make gcc6install gcc7install
         - CC=gcc-6 CFLAGS=-Werror make -j all
         - make clean
         - CC=gcc-7 CFLAGS=-Werror make -j all
+        - make clean
+        - LDFLAGS=-Wl,--no-undefined make -C lib libzstd-mt
 
     - name: gcc-8 + ASan + UBSan + Test Zstd   # ~6.5mn
       script:
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -154,9 +154,6 @@ libzstd.a: $(ZSTD_OBJ)
 	@echo compiling static library
 	@$(AR) $(ARFLAGS) $@ $^
 
-libzstd.a-mt: CPPFLAGS += -DZSTD_MULTITHREAD
-libzstd.a-mt: libzstd.a
-
 ifneq (,$(filter Windows%,$(OS)))
 
 LIBZSTD = dll\libzstd.dll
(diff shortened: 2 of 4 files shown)


Cyan4973:
> that sounds like a pretty big problem no (the not being able to link to the dynamic library)?

I think it depends on the exact executable being generated.

For example, regarding the `zstd` CLI, I believe it is indeed not great. We should probably lean towards making the CLI executable compatible with `libzstd`, hence invoke only public symbols. That will require some refactoring.

Regarding the tests programs, now it depends. Some tests may be created with the intention to check some internal functions, and therefore have no reason to be restricted to the list of public symbols only.
We may want to separate tests that are compatible with the restricted public list of symbols, and those that must access internals symbols.
So there is no trivial one-size-fits-all solution, and this property should be handled with care.

Issue #2035, opened 2020-03-12
Compiler macro warnings with -pedantic option


MarcusKlik:
Thanks for a fantastic library!

When compiling with the -pedantic option a lot of ISO C99 warnings are generated, an example of which is

```
d:/Compiler/gcc-4.9.3/mingw_32/bin/gcc  -I"D:/RCompile/recent/R/include" -DNDEBUG -I. -Ifstcore -Ifstcore_v1 -Ifstcore/LZ4 -Ifstcore/ZSTD -Ifstcore/ZSTD/common -Ifstcore/ZSTD/decompress -Ifstcore/ZSTD/compress -I'D:/RCompile/CRANpkg/lib/4.0/Rcpp/include'   -I"d:/Compiler/gcc-4.9.3/local330/include"     -pedantic -O2 -Wall  -std=gnu99 -mtune=core2 -c fstcore/ZSTD/decompress/zstd_decompress.c -o fstcore/ZSTD/decompress/zstd_decompress.o
fstcore/ZSTD/decompress/zstd_decompress.c: In function 'ZSTD_frameHeaderSize_internal':
fstcore/ZSTD/decompress/zstd_decompress.c:211:58: warning: ISO C99 requires rest arguments to be used
     RETURN_ERROR_IF(srcSize < minInputSize, srcSize_wrong);
```

These warnings are due to macro's defined with ellipses that are left empty when called, like the macro's defined [in this code](https://github.com/facebook/zstd/blob/dev/lib/common/zstd_internal.h#L63-L98).

Currently, R packages submitted to CRAN that include ZSTD as part of their source code fail the acceptance pre-tests because of these warnings. Would it be possible to include the -pedantic option in your gcc builds to help fix them?

thanks and all the best

bimbashrestha:
I think this is a duplicate issue: https://github.com/facebook/zstd/issues/1538

MarcusKlik:
Hi @bimbashrestha, thanks for your reply.

Yes, both issues are caused by calls to macro's defined with ellipsis arguments that are left unused. 
 Possible solutions are suggested [here](https://stackoverflow.com/questions/4100746/suppressing-iso-c99-requires-rest-arguments-to-be-used).

One of the suggestion in that post is to include the argument before the ellipsis in the ellipsis, that way the 'rest argument' is always used (by at least one parameter).

bimbashrestha:
[...]

PR #2099, opened 2020-04-30
Compile Under `-pedantic -Werror` and `-std=c90`

felixhandte:
This changeset allows Zstd to be built without warnings under `-pedantic`. Specifically, this was tested with:

```
make clean zstd CFLAGS="-std=c99 -Wall -Wextra -pedantic -Werror -DDEBUGLEVEL=3" CC=gcc
make clean zstd CFLAGS="-std=c99 -Wall -Wextra -pedantic -Werror -DDEBUGLEVEL=3" CC=clang
```

This addresses #2035. See also previous threads on this topic: #1989, #1540, and #1538.

It also lets Zstd build successfully (though not without warnings) under C90/C89/ANSI mode.

This PR accomplishes this in a different way than previous solutions. Rather than diverge into two kinds of macros with and without additional arguments, it simply complies with C99 by making sure every invocation of these macros does have at least one argument.

Known deficiencies:

- Every invocation of `RETURN_ERROR_IF`, `RETURN_ERROR`, and `FORWARD_IF_ERROR` must have an info string.

Why I selected this approach: adding `, ""` to the args is no more characters than adding `_MSG` to the name of the macro, and I'd rather bias contributors towards including explanatory text than away.

That being said, we could also introduce versions of these macros that don't take additional arguments. I'm open to suggestions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
           name: Test
           command: |
             ./tests/test-license.py
-            cc -v; CFLAGS="-O0 -Werror" make all && make clean
+            cc -v; CFLAGS="-O0 -Werror -pedantic" make all && make clean
             make c99build         ; make clean
             make c11build         ; make clean
             make aarch64build     ; make clean
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -17,7 +17,7 @@
 /* force inlining */
 
 #if !defined(ZSTD_NO_INLINE)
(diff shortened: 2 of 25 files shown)


terrelln:
Looks like the failing test is just noise, but this looks good to me once all the CI is passing!

terrelln:
We will need to compile with `-pedantic` in some CI. I suggest adding it to the CircleCI builds, since those are mostly compilation tests, and they complete the fastest by far.

Cyan4973:
The proposed solution looks good to me.
The only missing detail is a `-pedantic` test to ensure that the property remains respected in future changes.

PR #2100, opened 2020-04-30
Fix up superblock mode

### LINKED ISSUE #2096: ZSTD_c_targetCBlockSize == ZSTD_TARGETCBLOCKSIZE_MAX leads to improbably bad compression
Using `ZSTD_CCtxParams_setParameter(cctxParams, ZSTD_c_targetCBlockSize, ZSTD_TARGETCBLOCKSIZE_MAX)` leads to improbably bad compression for sources with large minimum sequences (>1kB).

Using a repeated concatenation of a minified CSS:
With setting targetCBlockSize == ZSTD_TARGETBLOCKSIZE_MAX:
`bootstrap.min.css    : 97.31%   (13862462 => 13488900 bytes, bootstrap.min.css.zst)`
Without setting targetCBlockSize:
`bootstrap.min.css    :  0.15%   (13862462 =>  20476 bytes, bootstrap.min.css.zst)`

Further note: I encountered this as a result of trying to reduce the block size on decompress regarding #2093. If instead of setting the TargetCompressedBlockSize, ZSTD_BLOCKSIZEMAX itself is reduced, in turn causing the maximum blocksize to be reduced as well, it has a much smaller impact on the compression.

-------

The source used was obtained via the following sequence:
`rm bootstrap.min.css; wget https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css && for i in {1..5}; do cat bootstrap.min.css >> bootstrap_2.min.css; cat bootstrap_2.min.css >> bootstrap.min.css; done && rm bootstrap_2.min.css`

terrelln:
Fixes:
* Enable RLE blocks for superblock mode
* Fix the limitation that the literals block must shrink. Instead, when we're within 200 bytes of the next header byte size, we will just use the next one up. That way we should (almost?) always have space for the table.
* Remove the limitation that the first sub-block MUST have compressed literals and be compressed. Now one sub-block MUST be compressed (otherwise we fall back to raw block which is okay, since that is streamable). If no block has compressed literals that is okay, we will fix up the next Huffman table.
* Handle the case where the last sub-block is uncompressed (maybe it is very small). Before it would skip superblock in this case, now we allow the last sub-block to be uncompressed. To do this we need to regenerate the correct repcodes.
* Respect disableLiteralsCompression in superblock mode
* Fix superblock mode to handle a block consisting of only compressed literals
* Fix a off by 1 error in superblock mode that disabled it whenever there were last literals
* Fix superblock mode with long literals/matches (> 0xFFFF)
* Allow superblock mode to repeat Huffman tables
* Respect `ZSTD_minGain()`.

Tests:
* Simple check for the condition in #2096.
* When the `simple_round_trip` fuzzer enables superblock mode, it checks that the compressed size isn't expanded too much.

Remaining limitations:
* O(targetCBlockSize^2) because we recompute statistics every sequence
* Unable to split literals of length > targetCBlockSize into multiple sequences
* Refuses to generate sub-blocks that don't shrink the compressed data, so we could end up with large sub-blocks. We should emit those sections as uncompressed blocks instead.
* ...

Fixes #2096

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -291,6 +291,31 @@ typedef struct {
     U32   longLengthPos;
 } seqStore_t;
 
+typedef struct {
+    U32 litLength;
+    U32 matchLength;
+} ZSTD_sequenceLength;
+
+/**
+ * Returns the ZSTD_sequenceLength for the given sequences. It handles the decoding of long sequences
+ * indicated by longLengthPos and longLengthID, and adds MINMATCH back to matchLength.
+ */
+MEM_STATIC ZSTD_sequenceLength ZSTD_getSequenceLength(seqStore_t const* seqStore, seqDef const* seq)
+{
+    ZSTD_sequenceLength seqLen;
+    seqLen.litLength = seq->litLength;
+    seqLen.matchLength = seq->matchLength + MINMATCH;
+    if (seqStore->longLengthPos == (U32)(seq - seqStore->sequencesStart)) {
+        if (seqStore->longLengthID == 1) {
+            seqLen.litLength += 0xFFFF;
+        }
+        if (seqStore->longLengthID == 2) {
+            seqLen.matchLength += 0xFFFF;
+        }
+    }
+    return seqLen;
+}
+
 /**
  * Contains the compressed frame size and an upper-bound for the decompressed frame size.
  * Note: before using `compressedSize`, check for errors using ZSTD_isError().
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1928,21 +1928,6 @@ void ZSTD_seqToCodes(const seqStore_t* seqStorePtr)
         mlCodeTable[seqStorePtr->longLengthPos] = MaxML;
 }
 
-static int ZSTD_disableLiteralsCompression(const ZSTD_CCtx_params* cctxParams)
-{
-    switch (cctxParams->literalCompressionMode) {
-    case ZSTD_lcm_huffman:
-        return 0;
-    case ZSTD_lcm_uncompressed:
-        return 1;
-    default:
-        assert(0 /* impossible: pre-validated */);
-        /* fall-through */
-    case ZSTD_lcm_auto:
-        return (cctxParams->cParams.strategy == ZSTD_fast) && (cctxParams->cParams.targetLength > 0);
-    }
-}
-
 /* ZSTD_useTargetCBlockSize():
  * Returns if target compressed block size param is being used.
  * If used, compression will do best effort to make a compressed block size to be around targetCBlockSize.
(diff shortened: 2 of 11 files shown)


lib/compress/zstd_compress.c
```diff
@@ -2463,6 +2460,16 @@ static size_t ZSTD_compressBlock_targetCBlockSize_body(ZSTD_CCtx* zc,
 {
     DEBUGLOG(6, "Attempting ZSTD_compressSuperBlock()");
     if (bss == ZSTDbss_compress) {
+        if (/* We don't want to emit our first block as a RLE even if it qualifies because
+            * doing so will cause the decoder (cli only) to throw a "should consume all input error."
```
Cyan4973, line 2464: 👍

lib/common/zstd_internal.h
```diff
@@ -291,6 +291,31 @@ typedef struct {
...
+typedef struct {
+    U32 litLength;
+    U32 matchLength;
+} ZSTD_sequenceLength;
+
+/**
+ * Returns the ZSTD_sequenceLength for the given sequences. It handles the decoding of long sequences
+ * indicated by longLengthPos and longLengthID, and adds MINMATCH back to matchLength.
+ */
+MEM_STATIC ZSTD_sequenceLength ZSTD_getSequenceLength(seqStore_t const* seqStore, seqDef const* seq)
```
Cyan4973, line 303: My understanding is that this function, `ZSTD_getSequenceLength()`, is only used within `zstd_compress_superblock.c`.

Why is it implemented in `common/zstd_internal.h` ?

lib/compress/zstd_compress_superblock.c
```diff
@@ -16,6 +16,7 @@
 #include "zstd_compress_sequences.h"
 #include "zstd_compress_literals.h"
 #include "zstd_compress_superblock.h"
+#include "zstd_internal.h"
```
Cyan4973, line 19: minor : document a reason for this dependency (`getSequenceLength()`)

lib/common/zstd_internal.h
terrelln: It is a generic helper function related to sequences. By putting it near the sequences, hopefully the next person who needs to read a litlen/matchlen will use this function, and see the reasoning behind it. A refactor of existing helper code, like the sequence collector, should switch it to using this function.

tests/fuzzer.c
```diff
@@ -12,6 +12,7 @@
 /*-************************************
 *  Compiler specific
 **************************************/
+#include "zstd_internal.h"
```
Cyan4973, line 15: minor : document a reason for this additional dependency.

Question : why does it have to be here, rather than in the `*  Includes` paragraph below ?

tests/fuzzer.c
```diff
@@ -708,16 +707,44 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
         { ZSTD_inBuffer in = {src, srcSize, 0};
           ZSTD_outBuffer out = {dst, dstCapacity, 0};
-          assert(!ZSTD_isError(ZSTD_compressStream2(cctx, &out, &in, ZSTD_e_end)));}
+          CHECK_Z(ZSTD_compressStream2(cctx, &out, &in, ZSTD_e_end));}
+        ZSTD_freeCCtx(cctx);
+    }
+    DISPLAYLEVEL(3, "OK \n");
+
+    DISPLAYLEVEL(3, "test%3d: superblock with no literals : ", testNb++);
+    // Generate the same data 20 times over
```
Cyan4973, line 725: minor nit : avoid `//` comment style

tests/fuzzer.c
terrelln: Ohhh, that is why I ended up with duplicate includes... I think vscode decided to start silently adding missing includes...

Cyan4973:
Thanks for this fairly large fix, complete with a new test set

Issue #2073, opened 2020-04-06
Meson build fails on Windows/MSVC


veloman-yunkan:
**Describe the bug**
The meson build fails on Windows+MSVC configuration with the following error:

```
meson.build:124:5: ERROR: Unknown variable "enable_static_runtime".
```

**Expected behavior**
The build must pass

**Desktop (please complete the following information):**
 - OS: Windows
 - Compiler: MSVC
 - Build system Meson

**Proposed fix**

```
diff --git a/build/meson/meson.build b/build/meson/meson.build
index 121811e..970cf3c 100644
--- a/build/meson/meson.build
+++ b/build/meson/meson.build
@@ -68,6 +68,7 @@ zstd_docdir = join_paths(zstd_datadir, 'doc', meson.project_name())
 # Built-in options
 use_debug = get_option('debug')
 buildtype = get_option('buildtype')
+default_library_type = get_option('default_library')
 
 # Custom options
 debug_level = get_option('debug_level')
@@ -121,7 +122,7 @@ elif cc_id == compiler_msvc
   if use_multi_thread
     msvc_compile_flags += '/MP'
   endif
-  if enable_static_runtime
+  if use_static_runtime
     msvc_compile_flags += '/MT'
   endif
   add_project_arguments(msvc_compile_flags, language: ['c', 'cpp'])
```

bimbashrestha:
Thanks for filing the issue. I've managed to reproduce this bug on my machine:

```
$ meson builddir
...
Run-time dependency zlib found: YES 1.2.11
Run-time dependency liblzma found: YES 5.2.4
Run-time dependency liblz4 found: NO (tried pkgconfig and cmake)

meson.build:124:5: ERROR: Unknown variable "enable_static_runtime".
```

And I can confirm that your proposed changes fix the issue on my machine: 
After patch
```
$ meson builddir
...
Message: Enable multi-threading support
Windows resource compiler: Microsoft (R) Windows (R) Resource Compiler Version 10.0.10011.16384
Program ../InstallSymlink.py found: YES (c:\python38\python.exe C:\Users\Bimba Shrestha\workspace\zstd1\build\meson\programs\../InstallSymlink.py)
Build targets in project: 5
```

I'll put up a PR soon.

PR #2101, opened 2020-05-01
[build] fix meson build issue

bimbashrestha:
https://github.com/facebook/zstd/issues/2073#issuecomment-622405927


Cyan4973:
Note : I believe this PR change is not tested in this CI round because the [meson test belongs to the extended test list](https://github.com/facebook/zstd/blob/dev/.travis.yml#L247). So it will be picked by the nightly CI round after merge.

Issue #1998, opened 2020-02-15
Headers include paths aren't relative..


Niadb:
I don't understand what is going on with zstds header include structure, the paths are wrong and seem to assume that various directories are globally available..

Example:
cover.h in dictBuilder has
#include "mem.h"
#include "pool.h"

Neither exists in the same folder as cover, so I have to fix these by changing to

#include "../common/mem.h" 
#include "../common/pool.h"

Then fix it again when I update zstd, after a few years I just have to ask..why?

lazka:
The build system adds "common" to the include path: https://github.com/facebook/zstd/blob/ff6350c098300ee3050dc6a6cdc0f48032755e84/lib/Makefile#L22

TrianglesPCT:
This is C++, nobody uses the same build system..(we don't use makefiles on windows..)

Even if I have my project for zstd add common, these paths are exposed to other projects, so they also now need to globally add the common folder, that is just broken imo.

Niadb:
This is C++, nobody uses the same build system..(we don't use makefiles on windows..)

Even if I have my project for zstd add common, these paths are exposed to other projects, so they also now need to globally add the common folder, that just seems broken..

lazka:
(disclaimer: not a zstd dev) I don't think cover.h is supposed to be a public header so it shouldn't be exposed to other projects.

stati64:
We edit them also.  Why crave ambiguity?

felixhandte:
@Niadb,

Our expectation has been, I think, that consumers of zstd will use it having been built separately (either by installing it as a system package or by building it on its own and then linking the resulting artifacts into their application). Sourced in either of those ways, zstd will have been built using any of the build configurations provided in the project (Make, CMake, Meson, Buck, Visual Studio), all of which are set up to support this include pattern.

That being said, this is not the first report we've had of this sort. I'm personally not opposed to re-writing includes to be relative. Feel free to put up a PR.

felixhandte:
Fixed by #2103.

PR #2103, opened 2020-05-01
Migrate Includes to Relative Paths

felixhandte:
This removes the need for adding our subdirectories to the compiler's include path (via `-Ilib`, `-Ilib/common`, etc.), both for our own builds, and for those who copy the zstd source in their own source tree and build using their own build system.

This PR addresses #1998.

Tested with make and cmake. Will let contbuild test the other build systems.


terrelln:
Looks like the single file decoder is still failing

Cyan4973:
Great work @felixhandte !

cwoffenden:
The changes to the single-file generator script inlines files multiple times. For example, `xxhash.h` is inlined once as `common/xxhash.h`, then as `compress/../common/xxhash.h`, etc.

The resulting binary file is the same, and the include guards do their job against these multiple copies, but the resulting `.c` file for the decoder is over twice the size.

PR #2109, opened 2020-05-05
faster cygwin test

Cyan4973:
The cygwin test on Appveyor is way too long (>30 mn).
The main issue is the very long `zstreamtest`, because no time out has been programmed.
Ensures that both `fuzzer` and `zstreamtest` receive a 30sec slot.
Should reduce cygwin tests duration by -30mn.


Cyan4973:
Yep, that works,
Appveyor tests should use a lot less time now.

PR #2113, opened 2020-05-07
[contrib] Support seek table-only API

mdittmer:
Memory constrained use cases that manage multiple archives benefit from
retaining multiple archive seek tables without retaining a `ZSTD_seekable`
instance for each.

* New opaque type for seek table:` ZSTD_seekTable`.
* `ZSTD_seekable_copySeekTable()` supports copying seek table out of a
  `ZSTD_seekable`.
* `ZSTD_seekTable_[eachSeekTableOp]()` defines seek table API that mirrors
  existing seek table operations.
* Existing `ZSTD_seekable_[eachSeekTableOp]()` retained; they delegate to
  `ZSTD_seekTable` the variant.

These changes allow the above-mentioned use cases to initialize a
`ZSTD_seekable`, extract its `ZSTD_seekTable`, then throw the `ZSTD_seekable`
away to save memory. Standard ZSTD operations can then be used to
decompress frames based on seek table offsets.

The copy and delegate patterns are intended to minimize impact on
existing code and clients. Using copy instead of move for the infrequent
operation extracting a seek table ensures that the extraction does not
render the `ZSTD_seekable` useless. Delegating to *new* seek
table-oriented APIs ensures that this is not a breaking change for
existing clients while supporting all meaningful operations that depend
only on seek table data.

### DIFF
--- a/contrib/seekable_format/zstd_seekable.h
+++ b/contrib/seekable_format/zstd_seekable.h
@@ -29,6 +29,7 @@ extern "C" {
 
 typedef struct ZSTD_seekable_CStream_s ZSTD_seekable_CStream;
 typedef struct ZSTD_seekable_s ZSTD_seekable;
+typedef struct ZSTD_seekTable_s ZSTD_seekTable;
 
 /*-****************************************************************************
 *  Seekable compression - HowTo
@@ -154,6 +155,10 @@ ZSTDLIB_API size_t ZSTD_seekable_writeSeekTable(ZSTD_frameLog* fl, ZSTD_outBuffe
 ZSTDLIB_API ZSTD_seekable* ZSTD_seekable_create(void);
 ZSTDLIB_API size_t ZSTD_seekable_free(ZSTD_seekable* zs);
 
+/*===== Independent seek table management =====*/
+ZSTDLIB_API size_t ZSTD_seekable_copySeekTable(ZSTD_seekable* zs, ZSTD_seekTable** out);
+ZSTDLIB_API size_t ZSTD_seekTable_free(ZSTD_seekTable* st);
+
 /*===== Seekable decompression functions =====*/
 ZSTDLIB_API size_t ZSTD_seekable_initBuff(ZSTD_seekable* zs, const void* src, size_t srcSize);
 ZSTDLIB_API size_t ZSTD_seekable_initFile(ZSTD_seekable* zs, FILE* src);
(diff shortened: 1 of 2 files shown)


mdittmer:
CC @Cyan4973

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -197,23 +197,63 @@ size_t ZSTD_seekable_free(ZSTD_seekable* zs)
...
+    }
+
+    st->checksumFlag = zs->seekTable.checksumFlag;
+    st->tableLen = zs->seekTable.tableLen;
+
+    /* Allocate an extra entry at the end to match logic of initial allocation */
+    size_t entriesSize = sizeof(seekEntry_t) * (zs->seekTable.tableLen + 1);
+    seekEntry_t* entries = (seekEntry_t*)malloc(entriesSize);
+    if (!entries) {
+        free(entries);
```
Cyan4973, line 215: did you meant `free(st);` ?

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -197,23 +197,63 @@ size_t ZSTD_seekable_free(ZSTD_seekable* zs)
     return 0;
 }
 
+size_t ZSTD_seekable_copySeekTable(ZSTD_seekable* zs, ZSTD_seekTable** out)
+{
+    ZSTD_seekTable* st = malloc(sizeof(ZSTD_seekTable));
+    if (!st) {
+        free(st);
```
Cyan4973, line 204: minor : 
if `st == NULL`, there is no need to `free()` it.

contrib/seekable_format/zstd_seekable.h
```diff
@@ -29,6 +29,7 @@ extern "C" {
 
 typedef struct ZSTD_seekable_CStream_s ZSTD_seekable_CStream;
 typedef struct ZSTD_seekable_s ZSTD_seekable;
+typedef struct ZSTD_seekTable_s ZSTD_seekTable;
```
Cyan4973, line 32: just a note : 
the name `ZSTD_seekTable` looks correct,
I just can't help but find it close to `ZSTD_seekable`, 
which could become a source of confusion.

I don't have a better idea in mind though.

contrib/seekable_format/zstd_seekable.h
```diff
@@ -154,21 +155,33 @@ ZSTDLIB_API size_t ZSTD_seekable_writeSeekTable(ZSTD_frameLog* fl, ZSTD_outBuffe
 ZSTDLIB_API ZSTD_seekable* ZSTD_seekable_create(void);
 ZSTDLIB_API size_t ZSTD_seekable_free(ZSTD_seekable* zs);
 
+/*===== Independent seek table management =====*/
+ZSTDLIB_API size_t ZSTD_seekable_copySeekTable(ZSTD_seekable* zs, ZSTD_seekTable** out);
```
Cyan4973, line 159: minor : naming.

I expect `copy` to represent an operation which would copy a content into an already allocated buffer.
But here this is different. 
This operation effectively allocates the required memory buffer(s), and then forwards ownership to the caller.
There is also a risk of failure, that the caller must check.

Naming is hard.
It looks to me that this is actually a _creation_ operation,
which _also_ requires a specific input object to proceed.
In order to respect the established naming convention of the `zstd` project, I would recommend the following prototype :

`ZSTD_seekTable* ZSTD_seekTable_create_fromSeekable(const ZSTD_seekable* zs);`

Note that the change of signature is non trivial :
- `create` is pretty clear in signifying that it allocates an object, that the caller must then manage.
- like any allocation operation, it may fail, and the failure will be signaled by a `NULL` return value, which is a common `C` idiom
- the argument is `const`, meaning that its content is guaranteed to remain intact during the operation. This is a _very strong_ guarantee, which scales greatly in multi-threaded environments, such an opportunity should never be missed.
- the naming scheme leaves room for other forms of creators, such as a potential `_fromInput()` for example

contrib/seekable_format/zstd_seekable.h
```diff
@@ -154,21 +155,33 @@ ZSTDLIB_API size_t ZSTD_seekable_writeSeekTable(ZSTD_frameLog* fl, ZSTD_outBuffe
...
+
 /*===== Seekable decompression functions =====*/
 ZSTDLIB_API size_t ZSTD_seekable_initBuff(ZSTD_seekable* zs, const void* src, size_t srcSize);
 ZSTDLIB_API size_t ZSTD_seekable_initFile(ZSTD_seekable* zs, FILE* src);
 ZSTDLIB_API size_t ZSTD_seekable_decompress(ZSTD_seekable* zs, void* dst, size_t dstSize, unsigned long long offset);
 ZSTDLIB_API size_t ZSTD_seekable_decompressFrame(ZSTD_seekable* zs, void* dst, size_t dstSize, unsigned frameIndex);
 
 #define ZSTD_SEEKABLE_FRAMEINDEX_TOOLARGE (0ULL-2)
-/*===== Seek Table access functions =====*/
+/*===== Seekable seek table access functions =====*/
```
Cyan4973, line 169: I only realized that now,
but it seems that the `const` statement is incorrectly positioned.

`ZSTD_seekable* const zs` just means that the pointer `zs` will not be modified, which says nothing about its content, and provides absolutely no value at prototype level.
In contrast, `const ZSTD_seekable* zs` guarantees that the content of `zs` will not be modified, which is very important.

All functions below seem to be read-only. So it would make sense that they all accept a `const ZSTD_seekable* zs` argument. I presume this was simply a mistake, that needs to be fixed.

It impacts your work. Your new API is essentially identical, just applied to `ZSTD_seekTable*` object. Therefore, it inherited the same issue in all the new prototypes.
I believe they should also be updated to accept a `const ZSTD_seekTable* st` argument.

Cyan4973:
The strategy followed by this patch is good.
Essentially, `seekTable_t` already existed, though only as an internal object.
You expose it as a `ZSTD_seekTable` public incomplete type, and make all the relevant functions public, as opposed to private. This minimizes implementation risks.

Nevertheless, your patch uncovered a few flaws of existing design, that should better be fixed in this PR (see code comments).

It is also preferable that the new API capability be delivered with a test set. This new test set can either be a modification of an existing one (maybe in `examples/seekable_decompress.c` ?), or be an entirely new test created for the new prototypes. The choice is entirely up to you.

Cyan4973:
@mdittmer , is it still a PR you would like to progress ?

Cyan4973:
I'll merge this PR into its own feature branch.

I believe there are a nb of points (listed above) that need to be taken care of before merging into mainline.
I'll likely take over the work from the feature branch.

PR #2118, opened 2020-05-07
programs/README.md: update help text

Artoria2e5:
I noticed that the help is outdated as there is no mention of --adapt. Y'all should probably put "update help text" into your release checklist.

(Huh, what are these silent changes with the trailing spaces?)


terrelln:
Thanks @Artoria2e5! I forgot this was in the programs README, we'll have to put this in our release notes!

PR #2120, opened 2020-05-08
updated zstd CLI documentation

Cyan4973:
make it clearer that `ZSTD_CLEVEL` is notably useful for `tar --zstd`

Also added documentation for build macros
`ZSTD_NOBENCH`, `ZSTD_NODICT`, `ZSTD_NOCOMPRESS`, `ZSTD_NODECOMPRESS`

### DIFF
--- a/programs/README.md
+++ b/programs/README.md
@@ -61,6 +61,24 @@ There are however other Makefile targets that create different variations of CLI
   In which case, linking stage will fail if `lz4` library cannot be found.
   This is useful to prevent silent feature disabling.
 
+- __ZSTD_NOBENCH__ : `zstd` cli will be compiled without its integrated benchmark module.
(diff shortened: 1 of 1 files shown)


programs/README.md
```diff
@@ -233,7 +236,7 @@ The below table illustrates this on the [Silesia compression corpus].
 | `zstd -10 --long`| `3.566` | `16.2 MB/s`       | `415.7 MB/s`   |
 
 
-#### zstdgrep
+### zstdgrep
```
bimbashrestha, line 268: Unrelated: I see below that we mention the limitations of zstdgrep. Guess we should also mention the grep options that we don't support currently? I can put up a pr soon

Cyan4973: Sure !

PR #2122, opened 2020-05-08
Compile Under `-Wunused-variable`

felixhandte:
CMake already passes this flag. Make should as well.

Also fixes up the one offender.


Cyan4973:
Interesting, I would expect `-Wunused-variable` to be already enabled as part of `-Wall` : 
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

felixhandte:
Ah I see yes. You're right. The confusion was because this was only failing when both `-Wunused-variable` (or `-Wall` or whatever) and `-DNDEBUG` (disabling the `assert()` that is using this variable) were both passed at the same time. We don't actually have any make targets that do both.

PR #2123, opened 2020-05-08
updated `zstd -h` 

Cyan4973:
to help readability.

Group arguments into separated titled categories,
try to give better visibility to more commonly used arguments.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -213,13 +213,13 @@ zstd-noxz : LZMALD  :=
 zstd-noxz : LZMA_MSG := - xz/lzma support is disabled
 zstd-noxz : zstd
 
-# note : the following target doesn't build
+## zstd-dll: zstd executable linked to dynamic library libzstd (must already exist)
+# note : the following target doesn't link
 #        because zstd uses non-public symbols from libzstd
 #        such as XXH64 (for benchmark),
-#        ZDICT_trainFromBuffer_unsafe_legacy for dictionary builder
-#        and ZSTD_cycleLog (likely for --patch-from)
-#        It's unclear at this stage if this is a scenario we want to support
-## zstd-dll: zstd executable linked to dynamic library libzstd (must already exist)
+#        ZDICT_trainFromBuffer_unsafe_legacy (for dictionary builder)
+#        and ZSTD_cycleLog (likely for --patch-from).
+#        It's unclear at this stage if this is a scenario that must be supported
 .PHONY: zstd-dll
 zstd-dll : LDFLAGS+= -L$(ZSTDDIR) -lzstd
 zstd-dll : ZSTDLIB_FILES =
@@ -227,6 +227,7 @@ zstd-dll : $(ZSTD_CLI_OBJ)
 	$(CC) $(FLAGS) $^ -o $@$(EXT) $(LDFLAGS)
 
 
+## zstd-pgo: zstd executable optimized with pgo. `gcc` only.
 zstd-pgo :
 	$(MAKE) clean
 	$(MAKE) zstd MOREFLAGS=-fprofile-generate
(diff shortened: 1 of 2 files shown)


programs/zstdcli.c
```diff
@@ -176,37 +184,44 @@ static void usage_advanced(const char* programName)
...
+
 #ifndef ZSTD_NODICT
     DISPLAYOUT( "\n");
     DISPLAYOUT( "Dictionary builder : \n");
     DISPLAYOUT( "--train ## : create a dictionary from a training set of files \n");
-    DISPLAYOUT( "--train-cover[=k=#,d=#,steps=#,split=#,shrink[=#]] : use the cover algorithm with optional args\n");
-    DISPLAYOUT( "--train-fastcover[=k=#,d=#,f=#,steps=#,split=#,accel=#,shrink[=#]] : use the fast cover algorithm with optional args\n");
-    DISPLAYOUT( "--train-legacy[=s=#] : use the legacy algorithm with selectivity (default: %u)\n", g_defaultSelectivityLevel);
-    DISPLAYOUT( " -o file : `file` is dictionary name (default: %s) \n", g_defaultDictName);
+    DISPLAYOUT( "--train-cover[=k=#,d=#,steps=#,split=#,shrink[=#]] : use the cover algorithm with optional args \n");
```
Artoria2e5, line 206: Maybe the change should go the other way, like remove all the trailing spaces before \n?

programs/Makefile
```diff
@@ -213,20 +213,21 @@ zstd-noxz : LZMALD  :=
...
+#        and ZSTD_cycleLog (likely for --patch-from).
+#        It's unclear at this stage if this is a scenario that must be supported
 .PHONY: zstd-dll
 zstd-dll : LDFLAGS+= -L$(ZSTDDIR) -lzstd
 zstd-dll : ZSTDLIB_FILES =
 zstd-dll : $(ZSTD_CLI_OBJ)
 	$(CC) $(FLAGS) $^ -o $@$(EXT) $(LDFLAGS)
 
 
+## zstd-pgo: zstd executable optimized with pgo. `gcc` only.
```
Artoria2e5, line 230: Bit of a shame -- is there a particular reason for that? The online documentation for clang does mention PGO at https://clang.llvm.org/docs/UsersManual.html#id38.

Cyan4973: maybe because the way `clang`'s pgo works is different enough to require a different script

Artoria2e5: Still doesn't make sense. Let me try it on my end.

Artoria2e5: Got it. Making a PR.

programs/zstdcli.c
```diff
@@ -137,34 +137,42 @@ static void usage_advanced(const char* programName)
...
 # endif
-    DISPLAYOUT( "--no-dictID : don't write dictID into header (dictionary compression)\n");
-    DISPLAYOUT( "--[no-]check : integrity check (default: enabled) \n");
+    DISPLAYOUT( "--exclude-compressed: only compress files that are not already compressed \n");
+    DISPLAYOUT( "--stream-size=# : specify size of streaming input from `stdin` \n");
+    DISPLAYOUT( "--size-hint=# optimize compression parameters for streaming input of approximately this size \n");
+    DISPLAYOUT( "--target-compressed-block-size=# : generate compressed block of approximately targeted size \n");
+    DISPLAYOUT( "--no-dictID : don't write dictID into header (dictionary compression only) \n");
+    DISPLAYOUT( "--[no-]check : add XXH64 integrity checksum to frame (default: enabled) \n");
     DISPLAYOUT( "--[no-]compress-literals : force (un)compressed literals \n");
```
terrelln, line 174: nit: Do we want to add `(default: enabled)` like `--[no-]check`?

Cyan4973: Problem is : default depends on the compression level. 
It's enabled for levels >=1 , but disabled for `--fast`.
While this could be explained in a man page paragraph, it starts to be a bit complex for a one-line summary.

PR #2125, opened 2020-05-10
programs/Makefile: make it actually work with clang

Artoria2e5:
Recent versions of clang seem to require a step of conversion from the profraw file to a "compiler-readable" profdata file to let PGO work.


terrelln:
Have you seen any gains from PGO? We have never found PGO to help zstd compression or zstd compression.

Artoria2e5:
No, but it's just a good idea to have it work I guess.

PR #2131, opened 2020-05-12
Fix rare scenario with lazy parser, dictionary, and repcodes

terrelln:
The bug is be triggered when:
* A dictionary is used.
* There are at least two blocks of data.
* First block ends and one of the repcodes in the state points into the dictionary at the end of the block.
* The dictionary is invalid for all/some of the second block.
* The second block uses one of the repcodes from the first block that point into the dictionary, past the point where the dictionary is invalidated.

The bug is fixed by always validating the repcodes. All other strategies are okay because the lower levels invalidate offsets early, and the optimal parser already does this validation.

This patch adds a macro `FUZZING_ASSERT_VALID_SEQUENCE` which tells the decompressor to check sequences with `assert()` because they are always expected to be valid. The round trip style fuzzers define this macro to better catch corruption. Now we don't have to use streaming compression to catch these types of bugs. The `dictionary_round_trip` fuzzer now catches this bug within 1 minute with no seed corpus.

This patch also adds an additional fuzzer that splits the input into two segments and uses 1 as the raw content dictionary and 1 as the source. This fuzzer catches the bug within ~1 minute with no seed corpus.

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -63,10 +63,8 @@ size_t ZSTD_compressBlock_doubleFast_generic(
     const BYTE* ip = istart;
     const BYTE* anchor = istart;
     const U32 endIndex = (U32)((size_t)(istart - base) + srcSize);
-    const U32 lowestValid = ms->window.dictLimit;
-    const U32 maxDistance = 1U << cParams->windowLog;
     /* presumes that, if there is a dictionary, it must be using Attach mode */
-    const U32 prefixLowestIndex = (endIndex - lowestValid > maxDistance) ? endIndex - maxDistance : lowestValid;
+    const U32 prefixLowestIndex = ZSTD_getLowestPrefixIndex(ms, endIndex, cParams->windowLog);
     const BYTE* const prefixLowest = base + prefixLowestIndex;
     const BYTE* const iend = istart + srcSize;
     const BYTE* const ilimit = iend - HASH_READ_SIZE;
@@ -104,13 +102,15 @@ size_t ZSTD_compressBlock_doubleFast_generic(
 
     /* if a dictionary is attached, it must be within window range */
     if (dictMode == ZSTD_dictMatchState) {
-        assert(lowestValid + maxDistance >= endIndex);
+        assert(ms->window.dictLimit + (1U << cParams->windowLog) >= endIndex);
     }
 
     /* init */
     ip += (dictAndPrefixLength == 0);
     if (dictMode == ZSTD_noDict) {
-        U32 const maxRep = (U32)(ip - prefixLowest);
+        U32 const current = (U32)(ip - base);
+        U32 const windowLow = ZSTD_getLowestPrefixIndex(ms, current, cParams->windowLog);
+        U32 const maxRep = current - windowLow;
         if (offset_2 > maxRep) offsetSaved = offset_2, offset_2 = 0;
         if (offset_1 > maxRep) offsetSaved = offset_1, offset_1 = 0;
     }
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -662,10 +662,14 @@ ZSTD_compressBlock_lazy_generic(
                                      0;
     const U32 dictAndPrefixLength = (U32)((ip - prefixLowest) + (dictEnd - dictLowest));
 
+    DEBUGLOG(5, "ZSTD_compressBlock_lazy_generic (dictMode=%u)", (U32)dictMode);
+
     /* init */
     ip += (dictAndPrefixLength == 0);
     if (dictMode == ZSTD_noDict) {
-        U32 const maxRep = (U32)(ip - prefixLowest);
+        U32 const current = (U32)(ip - base);
+        U32 const windowLow = ZSTD_getLowestPrefixIndex(ms, current, ms->cParams.windowLog);
+        U32 const maxRep = current - windowLow;
         if (offset_2 > maxRep) savedOffset = offset_2, offset_2 = 0;
         if (offset_1 > maxRep) savedOffset = offset_1, offset_1 = 0;
     }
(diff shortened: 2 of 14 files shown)


Cyan4973:
Great investigation !

terrelln:
The fuzzers locally and in the CI uncovered a similar bug in the optimal parser, fixed in the last commit.

terrelln:
Accidentally compared an offset to an index in the optimal parser fix. The bug caught by both the CI and running fuzzers locally.

terrelln:
Looks like fast has the same bug.

terrelln:
All of the fast, dfast, and lazy parsers didn't correctly validate repcodes in single-segment mode.

They assumed the repcodes were valid if they were within the prefix. However, if there was previously a dictionary, a repcode that points into the prefix could have previously pointed into the dictionary and beyond the window size. So when the dictionary goes out of bounds that offset becomes invalid.

Secnario:
1. Repcode points into dictionary beyond window size in first block.
2. Dictionary is out of bounds in the second block, but the repcode is preserved e.g. because it is not compressible or 1 big match.
3. Third block also not compressible or 1 big match.
3. The 4th block uses the repcode from the dictionary which is beyond the window size. This can still point into the prefix if block size == window size we have 2 full windows of data in our prefix.

terrelln:
Weird, the assert is only failing with MSAN enabled!

terrelln:
Was a bug in the assertion. It didn't handle the (rare) case where the dictionary happened to end up as the prefix of the destination buffer.

Cyan4973:
Is there any performance impact ?

lib/compress/zstd_opt.c
```diff
@@ -569,7 +569,10 @@ U32 ZSTD_insertBtAndGetAllMatches (
             U32 repLen = 0;
             assert(current >= dictLimit);
             if (repOffset-1 /* intentional overflow, discards 0 and -1 */ < current-dictLimit) {  /* equivalent to `current > repIndex >= dictLimit` */
-                if (ZSTD_readMINMATCH(ip, minMatch) == ZSTD_readMINMATCH(ip - repOffset, minMatch)) {
+                /* We must validate the repcode offset because when we're using a dictionary the
+                 * valid offset range shrinks when the dictionary goes out of bounds.
+                 */
+                if ((repIndex >= windowLow) & (ZSTD_readMINMATCH(ip, minMatch) == ZSTD_readMINMATCH(ip - repOffset, minMatch))) {
```
Cyan4973, line 575: Should it be a `&&` instead of `&`, as a way to guarantee that access at `ip - repOffset` is valid ?

lib/compress/zstd_lazy.c
```diff
@@ -953,10 +959,11 @@ size_t ZSTD_compressBlock_lazy_extDict_generic(
         U32 current = (U32)(ip-base);
 
         /* check repCode */
-        {   const U32 repIndex = (U32)(current+1 - offset_1);
+        {   const U32 windowLow = ZSTD_getLowestMatchIndex(ms, current+1, windowLog);
```
Cyan4973, line 962: minor : 
I wonder why it's necessary to invoke `ZSTD_getLowestMatchIndex()` at each position, 
as opposed to a constant calculated once ?
I presume this may cost performance (though it's arguably less important for the `_extDict()` variant).

lib/compress/zstd_opt.c
terrelln: Thats not necessary. We validate that it is within the prefix on line 571, so we know the memory is valid. This check just makes sure that the offset is allowed.

lib/compress/zstd_lazy.c
terrelln: We could "invalidate" bad repcodes at the beginning and set them to zero. Then we'd just need to compare against zero. But that is a slightly larger code change.

terrelln:
I've measured both ext-dict and single segment compression performance. I only see a speed regression for levels 5 and 6 for both single segment and ext-dict. The single segment regression is strange, since we don't change anything in that hot loop, will investigate.

I tested levels 1-9 & 16.

terrelln:
I've verified that the levels 5&6 speed in single-segment and ext-dict is an alignment thing. I aligned the loop to 32-bytes and the speed improved. I will put up a diff for that separately.

lib/compress/zstd_double_fast.c
```diff
@@ -104,13 +102,15 @@ size_t ZSTD_compressBlock_doubleFast_generic(
...
+        assert(ms->window.dictLimit + (1U << cParams->windowLog) >= endIndex);
     }
 
     /* init */
     ip += (dictAndPrefixLength == 0);
     if (dictMode == ZSTD_noDict) {
-        U32 const maxRep = (U32)(ip - prefixLowest);
+        U32 const current = (U32)(ip - base);
+        U32 const windowLow = ZSTD_getLowestPrefixIndex(ms, current, cParams->windowLog);
+        U32 const maxRep = current - windowLow;
```
terrelln, line 113: @Cyan4973 This is the change that causes the compressed size difference.

Before we were invalidating repcodes for all indices that are invalid by the end of the block.

Now we are only invalidating repcodes that are beyond our maximum offset. When we don't have a dictionary this is clearly okay. When we have a dictionary this is still okay, because we know the dictionary is valid for the entire block.

This shows up as a big difference for small window sizes because each block is exactly 1 window size, which means that the repcodes are never valid.

Do you agree?

Cyan4973: > Before we were invalidating repcodes for all indices that are invalid by the end of the block.

OK, it took me a while, but I'm finally reading the code the same way.

In `maxRep = (U32)(ip - prefixLowest)`, the trick is that `prefixLowest` was determined relative to the _end_ of the block, to ensure than any match found with an index `>= prefixLowest` is necessarily valid anywhere in the entire block.

As a consequence, `windowLow` can be as close as the beginning of the block, when window size == block size, effectively invalidating all repcodes.
The new code `maxRep = current - windowLow` checks more thoroughly vs the actual window size (and of course valid range within the prefix). So it effectively increases the range for which repcodes are valid, up to a maximum of `windowSize`, which is correct.

Hence, with more opportunities to find repcode matches at the beginning of the block, it can translate into a few more bytes saved here and there.

PR #2134, opened 2020-05-13
[greedy] Fix performance instability

terrelln:
PR #2131 caused a 5% loss in speed for levels 5 and 6. This is caused by code alignment. This PR gets the speed back up to par.

I've measured speed with clang and gcc levels 5-9 and the performance is now approximately neutral with performance before the PR (within 1 MB/s either way).

This is the same alignment command that we are using in `decompressSequences()` so we're not adding anything new right before the release.


Cyan4973:
Great find !

PR #2135, opened 2020-05-13
Revert to old zstdgrep because of potential license issue

bimbashrestha:
https://github.com/facebook/zstd/issues/2064


Cyan4973:
I presume test cases will have to be updated too

PR #2137, opened 2020-05-14
[actions] Migrating 1st batch of tests from Travis to Github Actions

bimbashrestha:
This is the first batch of tests I'm migrating from travis CI because of discontinued support in the coming months. This is just a subset of the dev pr tests. The master pr tests will be in a separate workflow file. 

Some tests weren't able to be migrated. Will have to do some additional work to make those work like setting up an arm self-hosted runner and trusty self-hosted runner with fb. 

* icc (need self-hosted)
* arm/qemu-arm (need self-hosted)
* versionTag 
* valgrindTest (keeps failing for some reason. need investigation)
* staticAnalyze (need trusty so need self-hosted)
* pcc-fuzz: (need trusty so need self-hosted)
* arm-build-test (need self-hosted)

### DIFF
--- a/.github/workflows/generic-dev.yml
+++ b/.github/workflows/generic-dev.yml
@@ -1,0 +1,172 @@
+name: generic-dev
+
+on:
+  pull_request:
+    branches: [ dev, master, actionsTest ]
+
+jobs:
+
+# Dev PR jobs that still have to be migrated from travis 
+# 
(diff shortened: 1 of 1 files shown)


bimbashrestha:
Well, actions finished slightly faster for me on this run. But it's an unfair comparison right now (because we're missing a few tests on actions :/)

PR #2138, opened 2020-05-14
[lib] Fix bug in loading LDM dictionary in MT mode

terrelln:
Exposed when loading a dictionary < LDM minMatch bytes in MT mode.

Test Plan:
Add a unit test that catches the bug, and run the zstreamtest reproducer.

```
CC=clang make -j zstreamtest fuzzer MOREFLAGS="-O0 -fsanitize=address"
./zstreamtest -vv -i100000000 -t1 --newapi -s7065 -t3925297
./fuzzer -v -i1
```


terrelln:
> We should indeed consider adding a relevant test as part of the PR.

Currently writing it.

terrelln:
Unit test added.

PR #2139, opened 2020-05-14
[fuzz] adding dictionary stream round trip fuzzer

bimbashrestha:
This exposes the bug mentioned in: 

https://github.com/facebook/zstd/pull/2138

```
$ ./dictionary_stream_round_trip corpora/dictionary_round_trip
...
../../lib/compress/zstdmt_compress.c: ZSTDMT_setBufferSize: bSize = 16236
../../lib/compress/zstd_compress_internal.h: ZSTD_window_update
../../lib/compress/zstd_compress_internal.h: Non contiguous blocks, new segment starts at 1
=================================================================
==25047==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61900000a480 at pc 0x00010a752f24 bp 0x7ffee569e640 sp 0x7ffee569e638
READ of size 1 at 0x61900000a480 thread T0
    #0 0x10a752f23 in ZSTD_rollingHash_append zstd_compress_internal.h:667
    #1 0x10a74c123 in ZSTD_rollingHash_compute zstd_compress_internal.h:677
    #2 0x10a74be53 in ZSTD_ldm_fillHashTable zstd_ldm.c:231
    #3 0x10a872295 in ZSTDMT_serialState_reset zstdmt_compress.c:516
    #4 0x10a86f3bf in ZSTDMT_initCStream_internal zstdmt_compress.c:1512
    #5 0x10a64ab78 in ZSTD_compressStream2 zstd_compress.c:4000
    #6 0x10a97e014 in compress dictionary_stream_round_trip.c:123
    #7 0x10a97cc32 in LLVMFuzzerTestOneInput dictionary_stream_round_trip.c:181
    #8 0x10a97f31a in main regression_driver.c:77
    #9 0x7fff68d3d3d4 in start (libdyld.dylib:x86_64+0x163d4)

0x61900000a480 is located 0 bytes to the right of 1024-byte region [0x61900000a080,0x61900000a480)
allocated by thread T0 here:
    #0 0x10aaee053 in wrap_malloc (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x5c053)
    #1 0x10a56631f in FUZZ_malloc fuzz_helpers.c:19
    #2 0x10a567a22 in FUZZ_train zstd_helpers.c:109
    #3 0x10a97ca34 in LLVMFuzzerTestOneInput dictionary_stream_round_trip.c:178
    #4 0x10a97f31a in main regression_driver.c:77
[...]

PR #2140, opened 2020-05-14
[fuzz] Adding dictionary_stream_round_trip fuzzer

bimbashrestha:
I can't find out why CIFuzz is failing on https://github.com/facebook/zstd/pull/2139

I'm creating a fresh pr to test if it fails here.

### DIFF
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -94,7 +94,8 @@ FUZZ_TARGETS :=       \
 	zstd_frame_info \
 	simple_compress \
 	dictionary_loader \
-	raw_dictionary_round_trip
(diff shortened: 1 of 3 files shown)


bimbashrestha:
Fixing the memory leak got CIFuzz to pass. Looks like CIFuzz didn't catch the ASAN failure and instead reported a passing build and a different failure. Going to file an issue with them

terrelln:
> Looks like CIFuzz didn't catch the ASAN failure and instead reported a passing build and a different failure.

I think it caught the error, which is why it failed. But then when trying to run the "reproduce" command it ran into a bug in oss-fuzz, which "swallowed" the error message.

PR #2143, opened 2020-05-16
[cmake] Generate libzstd.pc everywhere pkg-config is found

xantares:
Else it wont generate it when cross-compiling with mingw (target is WIN32, not UNIX)
Another solution would have been "if (NOT MSVC)" as MSVC is the only I can think of where pc files are useless, but it may be incomplete.
This is the cleanest idea I could think of.

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -134,7 +134,8 @@ if (ZSTD_BUILD_STATIC)
             OUTPUT_NAME ${STATIC_LIBRARY_BASE_NAME})
 endif ()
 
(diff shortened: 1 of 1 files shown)


xantares:
cc @felixhandte

felixhandte:
Hi @xantares,

This patch seems pretty straightforward. Happy to merge it.

You do need to agree to the contributor license agreement though before I can do that.

Thanks!

felixhandte:
Without a CLA from the author, we can't offer the contributed code under the project's license, since you hold the copyright on your work. Given that, I'm closing this PR for the time being. If you sign the CLA, please re-open.

PR #2146, opened 2020-05-18
Reduced single file lib size on macOS

cwoffenden:
Since #2103 the generated file sizes are much larger on macOS (and any other environment where `realpath` is missing). This adds a Python fallback to bring the file sizes back to what they were.


cwoffenden:
Travis CI failed with "too many requests" and the fuzzer failure is (I believe) unrelated (the codebase wasn't changed outside of the single file lib folder).

PR #2147, opened 2020-05-18
Update ClickHouse-related links to to their redirect targets

blinkov:


Cyan4973:
I'm happy to update the link.

Note though that CLA-compliance is not optional in order for us to merge external contributions.

blinkov:
@Cyan4973, unfortunately, it would have to be a corporate one, and bothering two legal departments for fixing a few links sound like overkill. Hopefully, someone else would do it.

Issue #2144, opened 2020-05-16
Test Suite broken due to using /bin/sh with [[ operator


davidfavor:
Several scripts in the test suite contain #!/bin/sh in first line.

Then in the script use "[[...]]" which is a bash-ism.

These scripts fail with all sorts of bogus errors related to this problem.

Here's an example...
```
# make check
make -C tests shortest
make[1]: Entering directory '/root/zstd.git/tests'
make -C ../programs zstd MOREFLAGS+="-g -DDEBUGLEVEL=1"
make[2]: Entering directory '/root/zstd.git/programs'
make[2]: 'zstd' is up to date.
make[2]: Leaving directory '/root/zstd.git/programs'
file ../programs/zstd
../programs/zstd: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=1c33f5eea4fcd0bd81be4a18a7c42952b5a848a4, for GNU/Linux 3.2.0, not stripped
EXE_PREFIX="" ZSTD_BIN="../programs/zstd" DATAGEN_BIN=./datagen ./playTests.sh 

Starting playTests.sh isWindows=false EXE_PREFIX='' ZSTD_BIN='../programs/zstd'

===>  simple tests 

test : basic compression 
tmp                  : 32.50%   ( 65537 =>  21298 bytes, tmp.zst)              
test : basic decompression
tmp.zst             : 65537 bytes                                              
test : too large compression level => auto-fix
Warning : compression level higher than max, reduced to 19 
tmp                  : 31.54%   ( 65537 =>  20669 bytes, tmp.zst)              
error: numeric value overflows 32-bit unsigned int 
test : --fast aka negative compression levels
tmp                  : 55.93%   ( 65537 =>  36656 bytes, tmp.zst)              
tmp                  : 56.31%   ( 65537 =>  36902 bytes, tmp.zst)              
tmp                  :100.02%   ( 65537 =>  65551 bytes, tmp.zst)              
error: numeric value overflows 32-bit unsigned int 
Incorrect parameters 
Usage : 
      zstd [args] [FILE(s)] [-o file] 

FILE    : a filename 
          with no FILE, or when FILE is - , read standard input
Arguments : 
 -#     : # compression level (1-19, default: 3) 
 -d     : decompression 
 -D DICT: use DICT as Dictionary for compression or decompression 
 -o file: result stored into `file` (only 1 output file) 
[...]

PR #2148, opened 2020-05-18
Don't Use `[[` in Shell Scripts

felixhandte:
Addresses #2144.

It's surprising that none of our tests caught this. We can address that shortfall in a follow-up issue.


felixhandte:
Follow-up in #2149.

PR #2150, opened 2020-05-18
[ldm] Reset loadedDictEnd when the context is reset

terrelln:
* Invalidate the `loadedDictEnd` when we reset the context
* Also invalidate `loadedDictEnd` when we do overflow correction

Added a test that produces an out of bounds offset before this fix, and works after.

Credit to OSS-Fuzz


terrelln:
Added a test.

PR #2157, opened 2020-05-21
Attempt to Fix Appveyor Master Tests

felixhandte:


felixhandte:
So the issue is now that the visual studio builds aren't producing a `datagen.exe`. I'm not sure how to get them to do that. Any ideas, @Cyan4973?

Cyan4973:
Have these tests ever worked ?
If yes, we could have a look at an older version, to determine what has changed.

Cyan4973:
Looking at the script currently present in `master` branch (==`v1.4.4`),
I don't see any difference regarding this part (`playTests.sh` with `visual` compiler).

`datagen` generation is probably controlled by an earlier step, 
either something has changed, resulting in `datagen` not being built,
or the script was already failing and we did not notice (?).

Cyan4973:
That's not what happens : `datagen.exe` is present in the local directory

PR #2159, opened 2020-05-22
Fix playTests.sh 

Cyan4973:
triggered following visual-release compilation test

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -169,8 +169,8 @@
   - SET "FUZZERTEST=-T30s"
   - if [%HOST%]==[visual] if [%CONFIGURATION%]==[Release] (
       CD tests &&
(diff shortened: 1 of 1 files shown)


terrelln:
Whats different between this patch and Felix's patch? Does the trailing space matter?

terrelln:
Woohoo, it looks like its fixed! That is such a dumb behavior for Appveyor...

Thanks @Cyan4973 for fixing all the test failures blocking the release!

Cyan4973:
> Does the trailing space matter?

Yes it does,
I presume it could be because in the new `playTests.sh`, @felixhandte made `zstd` and `datagen` an `sh` function, rather than a direct application of an environment variable.

terrelln:
Yeah, I presume it is because the variable is quoted. I made that change to support paths with spaces in the name.

PR #2163, opened 2020-05-22
Fix legacy build after #2103

orivej:
Fixes this error when building with `make ZSTD_LEGACY_SUPPORT=1`:

```
legacy/zstd_v04.c:77:10: fatal error: debug.h: No such file or directory                                                     
   77 | #include "debug.h"                                                                                                   
```

### DIFF
--- a/lib/legacy/zstd_v04.c
+++ b/lib/legacy/zstd_v04.c
@@ -74,7 +74,7 @@ extern "C" {
 /*-*************************************
 *  Debug
 ***************************************/
-#include "debug.h"
+#include "../common/debug.h"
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @orivej , that's a pretty straightforward fix !

Cyan4973:
cc @felixhandte 
I'm wondering why our test suite missed that one.
I was expecting one of our "last stage" test to cover compilation of older legacy units.
If that's not the case, we should have this test. Now, I'm unsure if similar issue are present in legacy 1-2-3.

PR #2167, opened 2020-05-23
[pzstd] Fix compilation error in MinGW

### LINKED ISSUE #2165: pzstd breaks on Windows build because ERROR is already defined on Windows
When building v1.4.5 for Windows with MinGW-w64 (from http://winlibs.com in my case) compilation fails because `ERROR` is already defined in Windows when it is redefined in `contrib/pzstd/Logging.h`.
I was able to build by adding `#define ERROR ZSTD_ERROR` at the top of `contrib/pzstd/Logging.h` for now, but the clean solution would be to replace `ERROR` with something else.

terrelln:
Fixes #2165

Tested with `make pzstd`

### DIFF
--- a/contrib/pzstd/Logging.h
+++ b/contrib/pzstd/Logging.h
@@ -13,10 +13,10 @@
 
 namespace pzstd {
 
-constexpr int ERROR = 1;
-constexpr int INFO = 2;
-constexpr int DEBUG = 3;
-constexpr int VERBOSE = 4;
+constexpr int kLogError = 1;
+constexpr int kLogInfo = 2;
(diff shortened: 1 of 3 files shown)


brechtsanders:
Thanks.
Note that this may make the `#undef ERROR` obsolete in the following files:

- contrib/seekable_format/zstdseek_compress.c
- contrib/seekable_format/zstdseek_decompress.c
- lib/common/error_private.h
- lib/common/threading.h 

In some places the comment for this line says: `already defined on Visual Studio`, so it was probably an earlier attempt to work around the same issue.

PR #2175, opened 2020-05-25
fix generalization of pattern rules in lib/Makefile

pixelb:
This was introduced in v1.4.4-383-gf77fd5ce
and results in the default rule in lib/ doing nothing.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -221,14 +221,14 @@ libzstd : $(LIBZSTD)
 lib : libzstd.a libzstd
 
 .PHONY: lib-mt
-%-mt : CPPFLAGS += -DZSTD_MULTITHREAD
-%-mt : LDFLAGS  += -pthread
-%-mt : %
+% : CPPFLAGS += -DZSTD_MULTITHREAD
+% : LDFLAGS  += -pthread
+% : lib
 	@echo multi-threading build completed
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
I'm surprised,
the `@echo` statement was supposed to ensure that the recipe was not empty, 
therefore avoiding the rule to be cancelled.

Also, I remember testing that rule manually, it used to work (but now it doesn't).
Finally, this issue should not have passed CI test, it was expected to be controlled there, so some test is missing.

Cyan4973:
Even more strange : 
`make libzstd-release` works
`make libzstd.a-release` works
but
`make lib-release` returns `make: Nothing to be done for 'lib-release'.`

That's the same `%-release` pattern rule ...

Cyan4973:
replacing `default : lib-release` by `default : all-release` 
make the default rule work again.

So this could be a solution.

However, it doesn't explain why `make lib-release` doesn't work ...

Cyan4973:
Also : 
if I remove `lib-release` from the list of `.PHONY` targets, it now works correctly.
But... why ???

Anyway, that's another simple solution.

Cyan4973:
Fix proposed in #2176

pixelb:
Yes I misunderstood % as pertaining only to the previous (phony) target

Issue #2173, opened 2020-05-25
--patch-from with files larger than 4GB


luzeagithub:
**Is your feature request related to a problem? Please describe.**
The --patch-from option only accept that are not larger than 4GB. My files are >5GB (the most) and very similar, they only differ in language, so the patch would be about 200MB.

**Describe the solution you'd like**
I would like zstd to support files larget than 4GB for the --patch-from option.

**Describe alternatives you've considered**
No alternatives, because I couldn't find any solutions regarding this error.

**Additional context**
```
PS Y:\zstd> .\zstd.exe --patch-from=.\en_windows_10_consumer_editions_version_2004_x64_dvd_8d28c5d7.iso .\de_windows_10_consumer_editions_version_2004_x64_dvd_7efdffc7.iso -o patch
zstd: error 42 : Can't handle files larger than 4 GB
```

bimbashrestha:
Thanks for the posting the issue. 

The restriction is actually 2GB instead of 4GB. That was a mistake and I just merged a pr to fix it. 

It would be nice to lift this restriction entirely at some point. But that will require quite a bit of work. I'm going to aim to have something like that for the next release if there is a lot of interest. 

A good (easier) first step might be to just bump the restriction to 4GB. I'd have to look into how involved that would be but it would definitely be a smaller patch than allowing for arbitrarily large files.

luzeagithub:
Why is it impossible to support files larger than 4GB?

felixhandte:
@luzeagithub, although the zstd codebase can compress arbitrarily large inputs, and although the Zstd format specification allows arbitrarily distant matches, this codebase internally uses 32 bit integers to represent match offsets. So in practice this implementation can only reference data within 4 GB of the current position in the stream. `--patch-from` mode operates as if you're compressing the new version using the old one as a dictionary, which places them in sequence like this: `[old file contents...][new file contents...]`. If the old file is larger than 4 GB, this implementation will therefore be unable to internally represent matches from a given part of the new file to the corresponding part of the old file, which is the whole point of `--patch-from`.

This could be fixed by using 64 bit integers to represent offsets internally, but that would be a large, scary refactor...

luzeagithub:
Ok, thanks for the explanation. Seems I have to use 3rd-party program (smartversion) to do it. It also has zstd support (which I am using) with files >4GB.

bimbashrestha:
I'm going to keep this issue open as a reminder that there is interest in this

[...]

PR #2178, opened 2020-05-26
Change --patch-from limit from 4gb to 2gb

bimbashrestha:
The effective limit was 2GB but the enforced limit was 4GB. Fixing the inconsistency. I see that there is another issue (https://github.com/facebook/zstd/issues/2173) open about increasing the limit all together. If we go down that route, that'll be a different pr. 

Test plan
```
$ make -C tests datagen
$ tests/datagen -g3000000000 > large-data
$ cp large-data large-dict
$ make -j
$ ./zstd --patch-from=large-dict large-data -o /dev/null
zstd: error 42 : Can't handle files larger than 2 GB
```

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -771,9 +771,10 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
                                     unsigned long long const maxSrcFileSize)
 {
     unsigned long long maxSize = MAX(prefs->memLimit, MAX(dictSize, maxSrcFileSize));
+    unsigned const maxWindowSize = (1U << ZSTD_WINDOWLOG_MAX);
     assert(maxSize != UTIL_FILESIZE_UNKNOWN);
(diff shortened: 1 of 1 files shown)


programs/fileio.c
```diff
@@ -772,8 +772,8 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
 {
     unsigned long long maxSize = MAX(prefs->memLimit, MAX(dictSize, maxSrcFileSize));
     assert(maxSize != UTIL_FILESIZE_UNKNOWN);
-    if (maxSize > UINT_MAX)
-        EXM_THROW(42, "Can't handle files larger than %u GB\n", UINT_MAX/(1 GB) + 1);
+    if (maxSize > INT_MAX)
```
Cyan4973, line 775: Prefer using `ZSTD_WINDOWLOG_MAX`

PR #2185, opened 2020-06-02
Install runtime components (i.e. DLLs) to BINDIR

tonytheodore:
The convention on windows-based systems (including mingw and cygwin) is to install DLLs in a `.../bin` directory. These [cmake target types](https://cmake.org/cmake/help/latest/command/install.html#installing-targets) handle the various shared, static, and import libraries across systems in a predictable way.


tonytheodore:
#2186 seems to be a duplicate of this, not sure why this is showing as having no conflicts with the base branch.

Feel free to close if there's any issues with merging.

Cyan4973:
You are right @tonytheodore , it's indeed a duplicate, tx !

Issue #2149, opened 2020-05-18
Test Shell Script Portability Better


felixhandte:
**Is your feature request related to a problem? Please describe.**
#2144 identified a use of `bash` features in a `sh` script, which escaped our notice in testing.

**Describe the solution you'd like**
We should run our shell scripts in a more diverse set of environments (in particular, finding environments where `sh` is not aliased to `bash`).

**Describe alternatives you've considered**
Alternatively, we could just rewrite our hashbangs to `/bin/bash`.

**Additional context**
N/A.

PR #2187, opened 2020-06-02
[ci] Adding bourne shell compatibility test

bimbashrestha:
Closes https://github.com/facebook/zstd/issues/2149

I'm using [shellcheck](https://github.com/koalaman/shellcheck#from-your-terminal) to check compatibility with bourne shell explicitly. I confirmed that this catches the incompatibility that was missed in the previous release (https://github.com/facebook/zstd/pull/2148): 

```
In tests/playTests.sh line 270:
if [[ $file2timestamp -ge $file1timestamp ]]; then
   ^-- SC2039: In POSIX sh, [[ ]] is undefined.
```

It also discovered 3 other warnings. Two of them were good suggestions so I fixed them. One of them isn't worth fixing imo so I suppressed it using shellcheck --exclude. This is the warning I excluded

```
In tests/playTests.sh line 401:
    ls -las $INTOVOID | grep "rw-rw-rw-"
    ^-- SC2010: Don't use ls | grep. Use a glob or a for loop with a condition to allow non-alphanumeric filenames.
```


Cyan4973:
Excellent ! Thanks @bimbashrestha !

PR #2190, opened 2020-06-03
Check environmental variables in playTests.sh and add default values

shaojing:
Improved zstd test script by making it more intuitive.
- Provided default values for both ZSTD_BIN and DATAGEN_BIN
- Added more actionable error messages in the script
- Added instructions for local testing in README

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -128,10 +128,25 @@ case "$UNAME" in
   SunOS) DIFF="gdiff" ;;
 esac
 
-println "\nStarting playTests.sh isWindows=$isWindows EXE_PREFIX='$EXE_PREFIX' ZSTD_BIN='$ZSTD_BIN' DATAGEN_BIN='$DATAGEN_BIN'"
 
-[ -n "$ZSTD_BIN" ] || die "\$ZSTD_BIN variable must be defined!"
-[ -n "$DATAGEN_BIN" ] || die "\$DATAGEN_BIN variable must be defined!"
+# check if ZSTD_BIN is defined. if not, use the default value
+if [[ -z "${ZSTD_BIN}" ]]; then
+  println "\nZSTD_BIN is not set. Using the default value..."
+  ZSTD_BIN="$PRGDIR/zstd"
+fi
+
+# check if DATAGEN_BIN is defined. if not, use the default value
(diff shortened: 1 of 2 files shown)


tests/playTests.sh
```diff
@@ -128,10 +128,25 @@ case "$UNAME" in
   SunOS) DIFF="gdiff" ;;
 esac
 
-println "\nStarting playTests.sh isWindows=$isWindows EXE_PREFIX='$EXE_PREFIX' ZSTD_BIN='$ZSTD_BIN' DATAGEN_BIN='$DATAGEN_BIN'"
 
-[ -n "$ZSTD_BIN" ] || die "\$ZSTD_BIN variable must be defined!"
-[ -n "$DATAGEN_BIN" ] || die "\$DATAGEN_BIN variable must be defined!"
+# check if ZSTD_BIN is defined. if not, use the default value
+if [[ -z "${ZSTD_BIN}" ]]; then
```
Cyan4973, line 133: Due to extreme portability requirements, we try to keep `playTests.sh` purely `sh` compatible.

This particular idiom is actually `bash` supported (that's why it [fails the bourne shell test](https://github.com/facebook/zstd/pull/2190/checks?check_run_id=735691305#step:3:53)).
This is a common mistake, as `bash` is popular and default on Linux. But unfortunately it is not guaranteed to be available on all unix platforms (FreeBSD is a good example).

_note (unrelated to this PR)_ : for some reason, the FreeBSD CI test did not "fail", even though this script is not `sh` compatible. It just complains, but without failing : 
`./playTests.sh: [[: not found`.
Fortunately, we have the bourne shell test to catch that, but I'm nonetheless curious as to why the FreeBSD test has not triggered a clear fail signal. It questions the validity / usefulness of this test. cc @bimbashrestha .

Cyan4973:
Thanks @shaojing , I believe this PR greatly answers task's requirements.

The only issue is reliance on a `bash` construct, which is disallowed in `playTests.sh`. Consider an `sh` compatible alternative.

tests/playTests.sh
shaojing: @Cyan4973 Yeah... saw the CI error message. Thank you for the detailed explanation! Fixed the double square bracket to single square bracket. Please help review again. Thanks.

PR #2194, opened 2020-06-05
[fuzz] Increase move forward amount for stream_decompress

bimbashrestha:
The new stream_decompress fuzz timeout happens when the output buffer size alternates between 1 and 0 and the decompressed output is quite large. The stream takes more than 60 seconds to decompress so it times out.

If the previous output size was 0, I'm bumping the output buffer size for the next stream by 10 instead of 1. This allows the decompression to finish within the time limit.

### DIFF
--- a/tests/fuzz/stream_decompress.c
+++ b/tests/fuzz/stream_decompress.c
@@ -70,6 +70,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *src, size_t size)
      * buffers in a row. */
     int prevInWasZero = 0;
     int prevOutWasZero = 0;
+
+    /* To avoid case where out.size alternates between 0 and non-0 and the
+     * decompressed output is large enough to cause a fuzzer timeout */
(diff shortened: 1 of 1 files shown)


bimbashrestha:
@Cyan4973 Talked to @terrelln about this offline. I think the preferred solution to this is to pass the whole src buffer in again when we run out (so it'd be a circular buffer I guess). And I guess we'd trigger that behavior here https://github.com/facebook/zstd/blob/89e90fc0a5a92c18ca22880ef88aef12fcd22cd4/tests/fuzz/stream_decompress.c#L57-L59 right @terrelln?

Cyan4973:
What's the status on this PR ?

bimbashrestha:
Ah yeah I simply just forgot about this PR. Thanks for the reminder @Cyan4973. If I remember right, the solution I proposed above didn't take care of the timeout. I'll find some time to look into this again soon.

terrelln:
I'll take this one over

PR #2196, opened 2020-06-08
Make --version respect -q for ouptut a shorter version string

tru:
This makes it easier to parse the version of zstd
we have instead of having to run a regex to parse
the WELCOME_MESSAGE

This was needed because we wanted to detect if
--patch-from was supported or not.

### DIFF
--- a/programs/zstd.1.md
+++ b/programs/zstd.1.md
@@ -238,6 +238,7 @@ the last one takes effect.
     display version number and exit.
     Advanced : `-vV` also displays supported formats.
     `-vvV` also displays POSIX support.
+    `-q` will only display the version number, suitable for machine reading.
 * `-v`, `--verbose`:
     verbose mode
 * `--show-default-cparams`:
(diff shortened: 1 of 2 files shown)


Cyan4973:
As there is no precedent for a `--dumpversion` command (at least across compression utilities),
I would rather avoid creating another keyword for this purpose, if that's possible.

It seems we can offer this capability by employing the verbosity level as it's designed.
The default verbosity level is `2`.
`-q` reduces the level to `1`, which basically means : reduce display output to essentials.

So we could use `-q -V` to display only the version number, as suggested in this PR.

tru:
I took dumpversion from GCC/clang. But I have no problem with your solution. I can see if I can rework this PR to do that.

tru:
I have redone the patch to instead just dump the version number if `g_displayLevel < DISPLAY_LEVEL_DEFAULT`

tru:
Anything more I should here before this can be approved?

Cyan4973:
Code is fine!

We _could_ also consider a mention of this new capability in the man page `programs/zstd.1.md`, but this is optional.

tru:
Documentation added - please let me know if that's fine and I'll merge this.

Cyan4973:
Thanks @tru !

PR #2197, opened 2020-06-09
meson: remove build requirement for distutils

neheb:
Tested on a default installation of Debian 10.


neheb:
Build failure seems unrelated.

PR #2201, opened 2020-06-12
Dedicated dict search rough draft 1

bimbashrestha:
PR just for review.

This is a first take at trying integrate the dict search optimizations. 

The idea behind this approach was to allow for a way to create a CDict with cParams specifically for the dedicated dict search algorithm (namely special hashLogs and chainLogs). So I introduced a new API method called `ZSTD_createCDict_advanced2(..)`. The difference between it and `ZSTD_createCDict_advanced` is that it takes a `ZSTD_CCtx_params` instead of `ZSTD_compressionParameters`. This way I can pass in the new parameter `ZSTD_c_enableDedicatdDictSearch` and select a set of dedicated defaults from the new `ZSTD_dedicatedDictSearch_defaultCompressionParameters`.

When the strategy picked by the new `ZSTD_dedicatedDictSearch_defaultCompressionParameters` falls outside `ZSTD_greedy`, `ZSTD_lazy` and `ZSTD_lazy2`, using the new param will have no effect. When the strategy picked is one of those and we have `ZSTD_c_enableDedicatedDictSearch` enabled, we force a dict to be attached instead of copied. 

The actual dedicated dict search algorithm needs to be tuned later but the one I have in right now has hash table buckets of size 4 and the remaining entries in the chainTable (used exactly as before so no reduction in size). It only prefetches the first hash entry into L1.

Also updated largeNbDicts.c. You can test the changes with --dedicated-dict-search. Here are some crude results for level 6 with http data (+ ~25-28% is where the wins plateau). I haven't property benchmarked or optimized yet so I'd expect this to be better.

plot

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3479,10 +3479,13 @@ ZSTDLIB_API ZSTD_CDict* ZSTD_createCDict_advanced2(const void* dict, size_t dict
 {
     int const enableDedicatedDictSearch = cctxParams->enableDedicatedDictSearch &&
         ZSTD_dedicatedDictSearch_isSupported(cctxParams->compressionLevel, dictSize);
-    if (!enableDedicatedDictSearch)
+    if (!enableDedicatedDictSearch) {
+        ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(
+            cctxParams->compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
         return ZSTD_createCDict_advanced(dict, dictSize,
-            dictLoadMethod, dictContentType, cctxParams->cParams,
+            dictLoadMethod, dictContentType, cParams,
             customMem);
+    }
     {
         ZSTD_compressionParameters const cParams = ZSTD_dedicatedDictSearch_getCParams(
             cctxParams->compressionLevel, dictSize);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2887,7 +2887,9 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
         case ZSTD_greedy:
         case ZSTD_lazy:
         case ZSTD_lazy2:
-            if (chunk >= HASH_READ_SIZE)
+            if (chunk >= HASH_READ_SIZE && params->enableDedicatedDictSearch)
+                ZSTD_lazy_loadDictionary(ms, ichunk-HASH_READ_SIZE);
+            else if (chunk >= HASH_READ_SIZE)
                 ZSTD_insertAndFindFirstIndex(ms, ichunk-HASH_READ_SIZE);
             break;
 
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3469,6 +3469,26 @@ ZSTD_CDict* ZSTD_createCDict_advanced(const void* dictBuffer, size_t dictSize,
     }
 }
 
+ZSTDLIB_API ZSTD_CDict* ZSTD_createCDict_advanced2(const void* dict, size_t dictSize,
+                                                  ZSTD_dictLoadMethod_e dictLoadMethod,
+                                                  ZSTD_dictContentType_e dictContentType,
+                                                  ZSTD_CCtx_params cctxParams,
+                                                  ZSTD_customMem customMem)
+{
+    int const enableDedicatedDictSearch = cctxParams.enableDedicatedDictSearch &&
+        ZSTD_dedicatedDictSearch_isSupported(cctxParams.compressionLevel, dictSize);
+    if (!enableDedicatedDictSearch)
+        return ZSTD_createCDict_advanced(dict, dictSize,
+            dictLoadMethod, dictContentType, cctxParams.cParams,
+            customMem);
+    {
+        ZSTD_compressionParameters const cParams = ZSTD_dedicatedDictSearch_getCParams(
+            cctxParams.compressionLevel, dictSize);
+        return ZSTD_createCDict_advanced(dict, dictSize,
+            dictLoadMethod, dictContentType, cParams, customMem);
+    }
+}
+
 ZSTD_CDict* ZSTD_createCDict(const void* dict, size_t dictSize, int compressionLevel)
 {
     ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3380,7 +3380,8 @@ static size_t ZSTD_initCDict_internal(
               const void* dictBuffer, size_t dictSize,
                     ZSTD_dictLoadMethod_e dictLoadMethod,
                     ZSTD_dictContentType_e dictContentType,
-                    ZSTD_compressionParameters cParams)
+                    ZSTD_compressionParameters cParams,
+                    ZSTD_CCtx_params params)
 {
     DEBUGLOG(3, "ZSTD_initCDict_internal (dictContentType:%u)", (unsigned)dictContentType);
     assert(!ZSTD_checkCParams(cParams));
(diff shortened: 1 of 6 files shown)


lib/zstd.h
```diff
@@ -356,6 +356,8 @@ typedef enum {
                               * Deviating far from default value will likely result in a compression ratio decrease.
                               * Special: value 0 means "automatically determine hashRateLog". */
 
+    ZSTD_c_enableDedicatedDictSearch=170,
```
Cyan4973, line 359: general rule for new advanced parameters :
- push it into experimental section (must be done in 2 parts, the `enum`, and then the `#define`)
- document what it does

When there is no strong expectation to become "stable"  on short term : 
- give it a high ID (> 1000)

lib/compress/zstd_lazy.c
```diff
@@ -475,6 +475,25 @@ U32 ZSTD_insertAndFindFirstIndex(ZSTD_matchState_t* ms, const BYTE* ip) {
     return ZSTD_insertAndFindFirstIndex_internal(ms, cParams, ip, ms->cParams.minMatch);
 }
 
+void ZSTD_dedicatedDictSearch_lazy_loadDictionary(ZSTD_matchState_t* ms, const BYTE* const ip)
+{
+    U32 const target = (U32)(ip - ms->window.base);
+    U32* const chainTable = ms->chainTable;
+    U32 const chainMask = (1 << ms->cParams.chainLog) - 1;
+    for (U32 idx = ms->nextToUpdate; idx < target; idx++) {
```
Cyan4973, line 483: This is `C99`.
`U32 idx; for(idx= ...` is `C90` portable.

lib/compress/zstd_lazy.c
```diff
@@ -531,7 +555,47 @@ size_t ZSTD_HcFindBestMatch_generic (
         matchIndex = NEXT_IN_CHAIN(matchIndex, chainMask);
     }
 
-    if (dictMode == ZSTD_dictMatchState) {
+    if (dictMode == ZSTD_dictMatchState && ms->dictMatchState->enableDedicatedDictSearch) {
```
Cyan4973, line 558: Does that sound possible to extract the dictionary search routine into its own function ?
This would centralize its implementation.

lib/compress/zstd_compress.c
```diff
@@ -4227,6 +4304,127 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
 },
 };
 
+static const ZSTD_compressionParameters
```
Cyan4973, line 4307: I'm wondering if having dedicated tables is really necessary,
or if it could be derived from the regular table with some local adjustments (such as `+ DD_BLOG`).

lib/compress/zstd_compress_internal.h
```diff
@@ -29,6 +29,12 @@ extern "C" {
 #endif
 
 
+/* Dedicated dict search bucket log:
+ * ---------------------------------
+ * This determines the additional space we need for the hash table.
+ * We will have 2^DD_BLOG slots in our bucket. */
+#define DD_BLOG 2
```
Cyan4973, line 36: The source code comment is fine,
but the name itself is not terribly self-descriptive.
I would suggest something longer, like ` DDS_BUCKETLOG`.

Cyan4973:
I believe the intention is that, once the new search method is sufficiently improved, tested and proven, it will become the new default one and just replace the old one.
Therefore, while it's certainly a good idea to have an enable/disable switch during the development period, such switch is not expected to survive beyond development, hence will never become part of the stable API.

lib/compress/zstd_compress.c
```diff
@@ -2866,7 +2887,9 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
         case ZSTD_greedy:
         case ZSTD_lazy:
         case ZSTD_lazy2:
-            if (chunk >= HASH_READ_SIZE)
+            if (chunk >= HASH_READ_SIZE && params->enableDedicatedDictSearch)
+                ZSTD_dedicatedDictSearch_lazy_loadDictionary(ms, ichunk-HASH_READ_SIZE);
```
terrelln, line 2891: This is called in a loop, is that okay?

lib/compress/zstd_compress.c
```diff
@@ -3357,7 +3380,8 @@ static size_t ZSTD_initCDict_internal(
               const void* dictBuffer, size_t dictSize,
                     ZSTD_dictLoadMethod_e dictLoadMethod,
                     ZSTD_dictContentType_e dictContentType,
-                    ZSTD_compressionParameters cParams)
+                    ZSTD_compressionParameters cParams,
+                    ZSTD_CCtx_params params)
```
terrelln, line 3384: Don't pass in both `ZSTD_CCtx_params` and `ZSTD_compressionParameters`.

If we're taking `ZSTD_CCtx_params` here now, the `cParams` should be transported inside of them.

terrelln: We'll need to think carefully about the parameters, and make sure everything is sane. In the past we were in a really bad state with parameter adjustment, so we need to be careful to not regress.

lib/compress/zstd_compress.c
```diff
@@ -3436,10 +3456,64 @@ ZSTD_CDict* ZSTD_createCDict_advanced(const void* dictBuffer, size_t dictSize,
...
+        ZSTD_dedicatedDictSearch_isSupported(cctxParams->compressionLevel, dictSize);
+    if (!enableDedicatedDictSearch) {
+        ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(
+            cctxParams->compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
+        return ZSTD_createCDict_advanced(dict, dictSize,
+            dictLoadMethod, dictContentType, cParams,
+            customMem);
+    }
+    {   ZSTD_compressionParameters const cParams = ZSTD_dedicatedDictSearch_getCParams(
+            cctxParams->compressionLevel, dictSize);
```
terrelln, line 3507: While creating these `cParams` you are taking the compression level. But it is possible that the user sets the `params.cParams` not the compression level, or sets some of the params. We need to handle that case.

Look at / use `ZSTD_getCParamsFromCCtxParams()`.

lib/compress/zstd_compress_internal.h
terrelln: It should also start with `ZSTD_`, so there is no chance of collision with other macros.

lib/compress/zstd_lazy.c
```diff
@@ -475,6 +475,25 @@ U32 ZSTD_insertAndFindFirstIndex(ZSTD_matchState_t* ms, const BYTE* ip) {
...
+    U32* const chainTable = ms->chainTable;
+    U32 const chainMask = (1 << ms->cParams.chainLog) - 1;
+    for (U32 idx = ms->nextToUpdate; idx < target; idx++) {
+        U32 const h = ZSTD_hashPtr(
+            ms->window.base + idx,
+            ms->cParams.hashLog - DD_BLOG,
+            ms->cParams.minMatch) << DD_BLOG;
+        chainTable[idx & chainMask] = ms->hashTable[h];
+        ms->hashTable[h] = idx;
+        /* Same logic as before. But now, just copy the chain into the bucket */
```
terrelln, line 490: This is non-optimal for large dictionaries `> chainSize`. It should be fine for now, but we should fix it up before landing.

lib/compress/zstd_lazy.c
```diff
@@ -506,6 +525,11 @@ size_t ZSTD_HcFindBestMatch_generic (
     /* HC4 match finder */
     U32 matchIndex = ZSTD_insertAndFindFirstIndex_internal(ms, cParams, ip, mls);
 
+    if (dictMode == ZSTD_dictMatchState && ms->dictMatchState->enableDedicatedDictSearch)
+        PREFETCH_L1(ms->dictMatchState->hashTable +
+            (ZSTD_hashPtr(ip, ms->dictMatchState->cParams.hashLog - DD_BLOG,
```
terrelln, line 530: We're calculating the hash here. If we cache it for use on line 570, does that improve speed?

lib/compress/zstd_lazy.c
```diff
@@ -531,7 +555,47 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+
+        U32 hash = ZSTD_hashPtr(ip, dms->cParams.hashLog - DD_BLOG,
+            dms->cParams.minMatch) << DD_BLOG;
+        U32 attemptNb = 0;
+        matchIndex = dms->hashTable[hash];
+
+        /* Empty chain */
+        if (!matchIndex)
+            return ml;
+
```
terrelln, line 578: Does prefetching the matches themselves help?

lib/compress/zstd_lazy.c
```diff
@@ -531,7 +555,47 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+        for ( ; (matchIndex>dmsLowestIndex) & (nbAttempts>0) ; nbAttempts--, attemptNb++) {
```
terrelln, line 579: It may be beneficial to split this into two loops. One for checking the bucket, and one for checking the chain.

bimbashrestha:
Closing this because it likely won't get worked on for a little while

PR #2202, opened 2020-06-13
[lib] Allow compression dictionaries with missing symbols

### LINKED ISSUE #2174: Custom dictionary: "error 11 : Allocation error : not enough memory"
When creating a custom dictionary I ran into something weird where I get the error above when trying to use it.

It is sort of clobbered together from a 'header' from another file and a custom payload, though I can't see why that would be an issue.

**To Reproduce**

* Ungz this dictionary: [dict-custom.gz](https://github.com/facebook/zstd/files/4677057/dict-custom.gz)
* Use command: `zstd -D dict-custom file.txt`
* zstd returns: `zstd: error 11 : Allocation error : not enough memory`. A 0 byte file is left behind.

**Expected behavior**

zstd should compress with the dictionary.

 - OS: Windows 10, 64 bit.
 - Version: `zstd command line interface 64-bits v1.4.5, by Yann Collet`
 - Compiler: Released binary

**Additional context**

As I said this is a hacked together dictionary, but AFAICT it should work. If I truncate the dictionary it works.

I tried creating a bigger dictionary with `zstd`, but that works.

terrelln:
Allow compression to use dictionaries with missing symbols in their
entropy tables. We set the FSE repeat mode to check when there are
missing symbols, and set the FSE repeat mode to valid when all symbols
are present.

Note that when not all symbols are present, the heuristics which favor
dictionary tables for lower compression levels won't activate.

Tested by manually creating a dictionary with missing symbols of every
type, and validating that the compressor rejects it before this change,
and accepts it after this change. Also, I ran the `dictionary_loader`
fuzzer for >1 hour of CPU time without running into cases where
compression succeeds, but decompression fails.

Fixes #2174.


terrelln:
@bimbashrestha I love the new Github Action tests! I got feedback in 1 minute that `make check` was failing, where before that would've taken > 20 minutes.

PR #2207, opened 2020-06-18
[actions] Migrating valgrind test

bimbashrestha:
As the title suggests. 

Note the comment about zstd without arguments not failing. Not sure what is happening here but I've flagged it for future examination. The rest of the valgrind test seems to work fine.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -262,7 +262,8 @@ valgrindTest: VALGRIND = valgrind --leak-check=full --show-leak-kinds=all --erro
 valgrindTest: zstd datagen fuzzer fullbench
 	@echo "\n ---- valgrind tests : memory analyzer ----"
 	$(VALGRIND) ./datagen -g50M > $(VOID)
-	$(VALGRIND) $(PRGDIR)/zstd ; if [ $$? -eq 0 ] ; then echo "zstd without argument should have failed"; false; fi
+	# doesn't seem to fail. need to re-examine
+	# $(VALGRIND) $(PRGDIR)/zstd ; if [ "$$?" -eq 0 ] ; then echo "zstd without argument should have failed"; false; fi
 	./datagen -g80 | $(VALGRIND) $(PRGDIR)/zstd - -c > $(VOID)
 	./datagen -g16KB | $(VALGRIND) $(PRGDIR)/zstd -vf - -c > $(VOID)
 	./datagen -g2930KB | $(VALGRIND) $(PRGDIR)/zstd -5 -vf - -o tmp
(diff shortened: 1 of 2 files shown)


tests/Makefile
```diff
@@ -262,7 +262,8 @@ valgrindTest: VALGRIND = valgrind --leak-check=full --show-leak-kinds=all --erro
 valgrindTest: zstd datagen fuzzer fullbench
 	@echo "\n ---- valgrind tests : memory analyzer ----"
 	$(VALGRIND) ./datagen -g50M > $(VOID)
-	$(VALGRIND) $(PRGDIR)/zstd ; if [ $$? -eq 0 ] ; then echo "zstd without argument should have failed"; false; fi
+	# doesn't seem to fail. need to re-examine
+	# $(VALGRIND) $(PRGDIR)/zstd ; if [ "$$?" -eq 0 ] ; then echo "zstd without argument should have failed"; false; fi
```
Cyan4973, line 266: I think the point is that `$(VALGRIND) $(PRGDIR)/zstd` _alone_, without the rest of the line, should fail.
The second part basically reverses the situation, and triggers an error if the first part hasn't.

Otherwise, the other tests below, which are the one which actually matter, won't trigger a clear failure signal if they do fail, making them barely useful in an automated CI environment (no one is gonna read the logs).

Cyan4973: OK, I may have misunderstood your code comment here : 
@bimbashrestha , do you mean that, in this instance, the `valgrind` command doesn't fail returning an error code,
while it's expected to, and as a consequence the _test_ itself fails,
and that's why you need to disable it, in order to make it run successfully on github action ?

bimbashrestha: @Cyan4973 yes

Cyan4973: OK, that's not the most important test, but I believe it was added for some reason.
I suspect that, at some point in the development process, just invoking `zstd` with no argument would produce some kind of buffer overflow or dangling pointer dereference. This is very easy to do : just access `argv[1]` without checking if `argc>=2` for example. This test was supposed to catch that.
So it would be better to keep the test active, or find a replacement (using a sanitizer for example).

bimbashrestha: Ahh I see. I just tried running the test with asan and couldn't get that line to fail on actions. See my fork: https://github.com/bimbashrestha/zstd/pull/55/checks?check_run_id=894893233

I'm also seeing some inconsistency between my ubuntu vm and the instance running on actions. Perhaps there are some issues with valgrind/compiler versions. 

I plan to look into this again to clean it up soon.

bimbashrestha: Looked into this again and I cannot get the behavior of github actions to reproduce on my ubuntu vm. Perhaps we should just leave this test in travis for now?

PR #2209, opened 2020-06-19
Explicitly use __cdecl for qsort, to avoid warning when default calling convention is not __cdecl

Niadb:
On MSVC there are other calling conventions such as __vectorcall, which may be enabled as the default. When this is done, zstd will not compile as it attempts to pass a __vectorcall sort function to qsort, which expects a __cdecl function.

 This explicitly marks those functions intended for usage with qsort as being __cdecl.


Cyan4973:
Thanks @Niadb ,
looks good to me,

could you add a comment at the place where the new macro is defined
explaining why it's useful ?
Otherwise, we will probably forget about it in some future.

Niadb:
hi @Cyan4973 

I added a comment explaining the reasoning behind the macro

Issue #2211, opened 2020-06-20
dictBuilder sources not in single file library


indygreg:
`contrib/single_file_libs/zstd-in.c` doesn't `#include` source files in `dictBuilder/`, meaning the various `ZDICT_` symbols aren't included in the single file library.

Note that if you add the `dictBuilder/` source files, you get various symbol collisions building the resulting file. Colliding symbols include `prime4bytes`, `DEFAULT_SPLITPOINT`, `g_displayLevel`, `g_time`, and others.

Should `dictBuilder/` source files be included in the single file full library by default?

Cyan4973:
My understanding is that this functionality, having the `dictBuilder` bundled inside the single-file library, is not supported yet. cc @cwoffenden .

cwoffenden:
I could add that to the full library. I'll take a look.

cwoffenden:
@Cyan4973 `cover.c` and `fastcover.c` contain a few sections of the same code. I could guard the macros and rename a few of the static vars. It would be more correct to share them in a common file though. I'll take a look when I can.

terrelln:
@cwoffenden feel free to move them to a common file. If they don't need to be inlined then `cover.c` would work with a declaration in `cover.h`. If they do need to be inlined for speed then putting them in `cover.h` would be appropiate.

cwoffenden:
This has now been merged as should work.

Cyan4973:
Thanks @cwoffenden !

@indygreg , you might want to test latest development version in `dev` branch,
see if the single-file library can now fit your needs.

yoshihitoh:
Thank you for this great feature!
I'm planning to use the amalgamated file on the Emscripten based project but failed to build it due to redefinitions of static vars.

Emscripten: 1.39.18
zstd branch:  dev (7afd5d85d359c4a778c0)

```bash
$ ./build_library_test.sh
Amalgamating files... this can take a while
Processing: zstd-in.c
...
Combine script: PASSED
Single file library creation script: PASSED
Compiling roundtrip.c: PASSED
Compression: PASSED (size: 3127, uncompressed: 32768)
Decompression: PASSED
Byte comparison: PASSED
Running roundtrip.c: PASSED
zstd.c:30822:21: error: redefinition of 'g_ctx' with a different type: 'COVER_ctx_t *' vs 'POOL_ctx' (aka 'struct POOL_ctx_s')
static COVER_ctx_t *g_ctx = NULL;
                    ^
[...]

PR #2212, opened 2020-06-22
Single-file libs now include dictBuilder

cwoffenden:
The `dictBuilder` sources were missing from the single-file libs, as per #2211. This PR:

- Adds the missing files to the generator script's input
- Fixes the duplicate 'console' macros and vars
- Fixes the duplicate hash functions

I'm not 100% happy with the approach for fixing the macros (adding `#undef`s) since it adds _noise_ around the definitions. That said... I'm not sure pulling out that code into a shared header would be any cleaner.

The hash functions were already present in `zstd_compress_internal.h` but instead of relying on those the choice was to prefix (with `COVER_` or `FASTCOVER_` accordingly). This was felt to be cleaner (than detecting `ZSTD_COMPRESS_H`, which felt like introducing future breakage).

### DIFF
--- a/lib/dictBuilder/fastcover.c
+++ b/lib/dictBuilder/fastcover.c
@@ -33,7 +33,7 @@
 #define FASTCOVER_MAX_SAMPLES_SIZE (sizeof(size_t) == 8 ? ((unsigned)-1) : ((unsigned)1 GB))
 #define FASTCOVER_MAX_F 31
 #define FASTCOVER_MAX_ACCEL 10
-#define DEFAULT_SPLITPOINT 0.75
+#define FASTCOVER_DEFAULT_SPLITPOINT 0.75
 #define DEFAULT_F 20
 #define DEFAULT_ACCEL 1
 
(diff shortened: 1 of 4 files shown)


lib/dictBuilder/fastcover.c
```diff
@@ -33,58 +33,67 @@
...
+static size_t FASTCOVER_hash8Ptr(const void* p, U32 h) { return FASTCOVER_hash8(MEM_readLE64(p), h); }
 
 
 /**
  * Hash the d-byte value pointed to by p and mod 2^f
  */
 static size_t FASTCOVER_hashPtrToIndex(const void* p, U32 h, unsigned d) {
   if (d == 6) {
-    return ZSTD_hash6Ptr(p, h) & ((1 << h) - 1);
+    return FASTCOVER_hash6Ptr(p, h) & ((1 << h) - 1);
```
terrelln, line 94: Wow, this `& ((1 << h) -1)` is completely redundant. I will put up a separate PR to remove it.

PR #2219, opened 2020-06-23
Add output-dir-mirror option

xxie24:


programs/fileio.c
```diff
@@ -1651,6 +1645,25 @@ static unsigned long long FIO_getLargestFileSize(const char** inFileNames, unsig
     return maxFileSize;
 }
 
+static const char*
+FIO_determineMirroredDestName(const char* srcFileName, const char* outDirName, const char* suffix)
+{
+    if (!UTIL_isFileNameValidForMirroredOutput(srcFileName))
+        return NULL;
+
+    char* srcFileNameCopy = strdup(srcFileName);
```
Cyan4973, line 1654: Be aware that `zstd` uses `C90` convention, for increased portability.
This requires all variables to be declared at the beginning of a block.

Sometimes, this requirement mandates creating new blocks, just for the purpose of declaring the variable,
which is actually useful whenever the variable also has a limited lifespan, 
as the block boundaries give it a clear end of life,
an often overlooked concept in managing software complexity.

Cyan4973: This project's coding style has a strong preference for `const` property,
which must be consistently applied whenever possible.

In this case, `srcFileNameCopy` should rather be a `char* const` pointer,
meaning that after its initialization, the pointer value never changes.

Cyan4973: `strdup` is actually creating an object in memory, and transferring its ownership into its return value.
You correctly identified that the memory object must be freed afterwards, which is great !

However, it also means that `strdup` may fail, and return a `NULL` pointer as return value.
This should be checked.

Given that it's not a situation we necessarily want to recover from (if the system can't even allocate enough memory for a simple filename, it's probably in a bad state already), you are allowed to end execution when it happens, in order to reduce the number of scenarios that the rest of the code must deal with.

programs/util.h
```diff
@@ -118,9 +124,16 @@ U64 UTIL_getFileSize(const char* infilename);
 U64 UTIL_getTotalFileSize(const char* const * fileNamesTable, unsigned nbFiles);
 int UTIL_getFileStat(const char* infilename, stat_t* statbuf);
 int UTIL_setFileStat(const char* filename, stat_t* statbuf);
+int UTIL_getDirectoryStat(const char* infilename, stat_t* statbuf);
 int UTIL_chmod(char const* filename, mode_t permissions);   /*< like chmod, but avoid changing permission of /dev/null */
 int UTIL_compareStr(const void *p1, const void *p2);
 const char* UTIL_getFileExtension(const char* infilename);
+int   UTIL_isFileNameValidForMirroredOutput(const char *filename);
+void  UTIL_convertPathnameToDirName(char *pathname);
```
Cyan4973, line 132: Even though that's implied by the function signature,
please document that this function is mutating its argument.

programs/util.c
```diff
@@ -633,6 +638,237 @@ const char* UTIL_getFileExtension(const char* infilename)
...
+{
+    if ((filename != NULL) && (filename[0] == PATH_SEP))
+        return filename + 1;
+    return filename;
+}
+
+static char*
+trimLeadingCurrentDir(char *filename)
+{
+    if ((filename != NULL) && (filename[0] == '.') && (filename[1] == PATH_SEP))
```
Cyan4973, line 710: Is it valid to invoke this function with a `NULL` pointer ?
If not, you can make it a pre-condition of the function, document it, and just `assert` it in the function body.

programs/util.h
```diff
@@ -118,9 +124,16 @@ U64 UTIL_getFileSize(const char* infilename);
...
+char* UTIL_trimPath(char *filename);
```
Cyan4973, line 133: Are there cases where the return pointer of this function must be able to mutate or `free()` its content ?
I did not notice one. If not, it would be better to return a `const char*`.
As a consequence, the argument could be `const char*` too, since it's not mutated either.
It also makes it a bit clearer that this function returns a reference, and not a new object.

programs/fileio.c
```diff
@@ -1655,6 +1649,28 @@ static unsigned long long FIO_getLargestFileSize(const char** inFileNames, unsig
...
+static const char*
+FIO_determineMirroredDestName(const char* srcFileName, const char* outDirName, const char* suffix)
+{
+    char *srcFileNameCopy = NULL, *srcDirName = NULL, *finalOutDirName = NULL;
+    const char *dest = NULL;
+
+    if (!UTIL_isFileNameValidForMirroredOutput(srcFileName))
+        return NULL;
+
+    srcFileNameCopy = strdup(srcFileName);
```
Cyan4973, line 1661: This fixes the `C90` declaration - statement order,
but `strdup()` may still fail and return `NULL` as a consequence.

programs/fileio.c
```diff
@@ -2536,10 +2563,31 @@ FIO_determineDstName(const char* srcFileName, const char* outDirName)
     /* note : dstFileNameBuffer memory is not going to be free */
 }
 
+static const char*
+FIO_determineMirroredDstName(const char* srcFileName, const char* outDirName)
```
Cyan4973, line 2567: So there are 2 functions, one called `FIO_determineMirroredDestName()`, and the other `FIO_determineMirroredDstName()`, doing essentially the same thing ?
If they do, they should be merged.
If they do not, the difference between them should be expressed in a more explicit way.

programs/util.c
```diff
@@ -633,6 +638,247 @@ const char* UTIL_getFileExtension(const char* infilename)
...
+        UTIL_DISPLAY("zstd: failed to create DIR %s: %s\n", dir, strerror(errno));
+    }
+    return ret;
+}
+
+void UTIL_convertPathnameToDirName(char *pathname)
+{
+    int len = 0;
+    char* pos = NULL;
+    // get dir name from pathname similar to 'dirname'
```
Cyan4973, line 679: `C90` comment style is limited to `/*  (...)  */`

programs/zstdcli.c
```diff
@@ -838,6 +847,7 @@ int main(int const argCount, const char* argv[])
                 if (longCommandWArg(&argument, "--target-compressed-block-size=")) { targetCBlockSize = readSizeTFromChar(&argument); continue; }
                 if (longCommandWArg(&argument, "--size-hint=")) { srcSizeHint = readSizeTFromChar(&argument); continue; }
                 if (longCommandWArg(&argument, "--output-dir-flat=")) { outDirName = argument; continue; }
+                if (longCommandWArg(&argument, "--output-dir-mirrored=")) { outMirroredDirName = argument; continue; }
```
Cyan4973, line 850: Is there a reason why this command is slightly different ?
`--output-dir-mirror` vs `--output-dir-mirrored` ?

programs/fileio.c
xxie24: yeah... I tried to merge it. The issue is each need call a static function which also has macro guard. 

I agree it is kinda ugly duplicated code here. Let me see if I can take another try to merge them.

Cyan4973:
Thanks @xxie24 ,
this is a rather good beginning.
You managed to integrate this capability in a way which makes sense within this project.
Also, the way you divide the topic into unitary components is appropriate.

We'll have a number of back and forth for the tactical details,
mostly around code conventions and specific precautions to C programming,
but I believe you've got the overall strategy right.

programs/fileio.c
xxie24: yeah, the C90 "declare before use" caught me be by surprise, I used to 100% following this rule. But after some years of C++, I forget it completely :D

programs/zstdcli.c
xxie24: good catch! Thanks!

programs/fileio.c
Cyan4973: I see, there is indeed one function invocation which is different.

Sometimes, I try to solve such issue by making the invoked function an argument, but that requires both functions to have the same prototype, and that's not the case here. This can be solved with a "shell" prototype, which will just be used to present the same interface, discard unused arguments and then invoke the "real" function.

Anyway, the issue is actually less about code duplication than it is about naming. The 2 function names are just too close, they do not describe what is the difference between them, hence why they both need to exist.

programs/util.c
xxie24: done.

programs/fileio.c
xxie24: done. will do a exit() upon strdup or malloc failure.

programs/util.h
xxie24: UTIL_trimPath will call trimLeadingCurrentDir(), trimLeadingCurrentDir() has user case of returning mutable object. 

It remind me the 'strchr()': (https://opensource.apple.com/source/BerkeleyDB/BerkeleyDB-18/db/clib/strchr.c.auto.html)
```
char *strchr(const char *p, int ch)
{
	char c;

	c = ch;
	for (;; ++p) {
		if (*p == c)
			return ((char *)p);
		if (*p == '\0')
			return (NULL);
	}
	/* NOTREACHED */
}
```
So seems like even C library is doing the const cast. There is an interesting analysis on this: https://stackoverflow.com/questions/14367727/how-does-strchr-implementation-work. So I can either write 1 function doing const-cast, or write a duplicated functions. 

trimLeadingCurrentDir() is 1 liner function, so I probably going to duplicate it into 2 funcitons that return const and non-const type.

Cyan4973: Indeed, `strchr()` is one of these cases where the standard got it seriously wrong, sacrificing the future for the short-term benefit of existing code. That being said, I'm a bit more sympathetic for errors made during this era, when many concepts where still being worked on. After all, `const` was introduced _after_ `strchr()`, and this lead to a situation where the function was already deployed, and the committee was under strong pressure to limit code breakage, so adaptation to `const` was a tricky situation.

Anyway, nowadays, we have no such excuse. `const` is an extremely important property, and it shall be applied and respected scrupulously. If there is a need for 2 versions, one safe which only works on references and provides references, and a more powerful but also more dangerous one which allows full ownership transfer, then it's better to write 2 functions.

programs/util.h
xxie24: done.

programs/util.c
xxie24: done.

programs/util.h
xxie24: I ended up doing a trick on this:
'''
static const char*
trimLeadingCurrentDirConst(const char *filename)
{
    assert(filename != NULL); // don't expect null string here
    if ((filename[0] == '.') && (filename[1] == PATH_SEP))
        return filename + 2;
    return filename;
}

static char*
trimLeadingCurrentDir(char *filename)
{
    /* 'union charunion' can do const-cast without compiler warning */
    union charunion {
        char *chr;
        const char* cchr;
    } ptr;
    ptr.cchr = trimLeadingCurrentDirConst(filename);
    return ptr.chr;
}
'''
seems to me a good compromise on this. Let me know what you think.

programs/fileio.c
xxie24: I put too much implementation details here in fileio.c, all these implementation should be only exist in the util.c. I export only 2 API nows in coming PR:
```
void  UTIL_mirrorSourceFilesDirectories(const char** fileNamesTable, unsigned int nbFiles, const char *outDirName);
char* UTIL_createMirroredDestDirName(const char* srcFileName, const char* outDirRootName);
```
 It seems reduce the need for me to write a dedicated function.

programs/util.h
Cyan4973: Interesting. That's a use of `union` I'm not used to. I'm wondering if it could be prone to strict aliasing issues. If not, then it's a clever use to circumvent `const` property.

programs/fileio.c
```diff
@@ -1689,12 +1684,28 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
...
             const char* const srcFileName = inFileNamesTable[u];
-            const char* const dstFileName = FIO_determineCompressedName(srcFileName, outDirName, suffix);  /* cannot fail */
+            const char* dstFileName = NULL;
+            if (outMirroredRootDirName) {
+                char* outMirroredDirName = UTIL_createMirroredDestDirName(srcFileName, outMirroredRootDirName);
+                if (outMirroredDirName) {
+                    dstFileName = FIO_determineCompressedName(srcFileName, outMirroredDirName, suffix);
+                    free(outMirroredDirName);
+                } else {
+                    continue;
```
Cyan4973, line 1700: Reaching that point is an error (basically `outMirroredDirName` allocation failed).
It should at least trigger some error signal (`error = 1;`).

programs/fileio.c
```diff
@@ -2558,12 +2569,23 @@ FIO_decompressMultipleFilenames(FIO_prefs_t* const prefs,
...
         for (u=0; u<nbFiles; u++) {   /* create dstFileName */
             const char* const srcFileName = srcNamesTable[u];
-            const char* const dstFileName = FIO_determineDstName(srcFileName, outDirName);
+            const char* dstFileName = NULL;
+            if (outMirroredRootDirName) {
+                char* outMirroredDirName = UTIL_createMirroredDestDirName(srcFileName, outMirroredRootDirName);
+                if (outMirroredDirName) {
+                    dstFileName = FIO_determineDstName(srcFileName, outMirroredDirName);
+                    free(outMirroredDirName);
+                }
```
Cyan4973, line 2584: what happens `if (!outMirroredDirName)` ?

I believe it's reasonable to stop execution if such a case happen, as we probably don't want to continue execution if the system can't even allocate a bit of memory for a file name.

programs/fileio.c
xxie24: right... 

I have a test case for this in the playTests.sh
```
    println "test: compress input dir will be ignored if it has '..'"
    zstd  -r tmpInputTestDir/we/must/../must --output-dir-mirror non-exist
    test ! -d non-exist
```
if I add that `error=1`, the zstd will exit with error signal and break the test suite. 

Is there a way to handle this in test script?

programs/fileio.c
xxie24: here the 'error=1' after outMirroredRootDirName, I think it will cause the program return non-0 error status.

For malloc issues, it the program will call exit.

The user case of 'outMirroredRootDir == NULL" is actually because the input dir has "..", according to the `tar`, it continue to work with other inputs. Should we follow the `tar` behavior? 

But I think need add an error print here.

programs/fileio.c
Cyan4973: Ah, that's different then.
Indeed, it matters to define a behavior.

First, is it certain that, on reaching that branch of the code, the reason necessarily is that a `..` is present in the filename ?
If yes, this should be documented.

I hear that `UTIL_createMirroredDestDirName()` can fail, but when it does, the failure reason cannot be an allocation error.
That's unusual for a function which returns a `NULL` pointer. But well, ok, unusual isn't impossible. Just, at least, it must be clearly documented.

Then is that a good idea to fail silently without sending an error signal ?
I don't think so.
For example, `zstd badfile goodfile` will fail if `goodfile` exists but `badfile` doesn't exist. 
It will nonetheless continue beyond `badfile`, and generate `goodfile.zst`.

I see no difference with the case described above, where compression of a file is skipped because of a problem, but the rest of the list proceed nonetheless. It should still return an error code.

As for the test itself, testing that an error is raised is not difficult.
We already have multiple tests which check that a given scenario is supposed to fail.
See for example : https://github.com/facebook/zstd/blob/v1.4.5/tests/playTests.sh#L235 .

programs/fileio.c
Cyan4973: Yes, an information message is a good idea.
Use level 2, as it's a normal warning, not a fatal error.

programs/fileio.c
xxie24: `That's unusual for a function which returns a NULL pointer. But well, ok, unusual isn't impossible. Just, at least, it must be clearly documented.`

Yeah, I overload the return meaning a bit. I changed the name from `outMirroredDirName` to `validMirroredDirName`, and add a comment to `UTIL_createMirroreDestDirName`, hopefully it reduce the confusion a bit. here. 

I decided not to throw error for this case (same as 'tar' case) but just do printf to stderr.

Cyan4973: That's better, but I still believe it's more in line with other functionalities of `zstd` to issue an error status in this case.
I understand that `tar` wouldn't. But for such a case, we do have a reference in `zstd` directly, which takes precedence.
(and not just `zstd` by they way, `gzip` and `xz` are also excellent references, and they all issue an error code when one element in the list of objects to compress or decompress is not electable for any sort of condition (file does not exist, or file is a directory, for example)).

xxie24: Agree, I will do so.

Cyan4973:
Thanks @xxie24 , this is a great feature, which is highly welcomed.

I note that one last test is failing (`min-decomp-macros`), 
though I somewhat remember that it might be one of the new tests that tend to be a bit flackey. cc @bimbashrestha

Issue #2215, opened 2020-06-22
Make no longer builds on SLES11 OS


dkcasset:
I believe I was able to build zstd 1.3.4 on SLES11 but 1.4.5 now fails for 2 reasons:

Issue 1: Use of undocumented -E option in sed (alias for -r)
Issue 2: Missing utime.h in programs/util.c

To reproduce:

1. make
sed: invalid option -- 'E'

2a. In lib/Makefile, replace all occurrences of -E option with -r
2b. make
util.c: In function 'UTIL_setFileStat':
util.c:161: error: storage size of 'timebuf' isn't known

3a. Add to programs/util.c
#include <utime.h>
3b. make
(Build successful)

Desktop
 - OS: SLES11 (x86-64_linux_3.0.101_ImageSLES11SP4-4)
 - Version: 1.4.5
 - Compiler: cc
 - Flags: Whatever is supplied in the Makefiles
 - Hardware specs: Intel(R) Xeon(R) CPU E5-2667
 - Build system: make

terrelln:
Thanks for the report @dkcasset!

If you would like to put up a PR for these changes we'd happily accept it. Otherwise, we'll be sure to make these fixes before our next release.

dkcasset:
I have since realized that the utime.h issue can be solved on the command line:
make CC='cc -DPLATFORM_POSIX_VERSION=1'
This is a reasonable workaround

felixhandte:
I believe this is fixed, so I'm going to close this.

PR #2220, opened 2020-06-26
Replace -E option with equivalent -r for older versions of sed

dkcasset:
For issue #2215

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -49,6 +49,7 @@ ifeq ($HAVE_COLORNEVER, 1)
 GREP_OPTIONS += --color=never
 endif
 GREP = grep $(GREP_OPTIONS)
+SED_ERE_OPT ?= -E
 
 ZSTDCOMMON_FILES := $(sort $(wildcard common/*.c))
 ZSTDCOMP_FILES := $(sort $(wildcard compress/*.c))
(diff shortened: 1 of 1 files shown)


dkcasset:
I see that the GNU documentation recommends -E for POSIX portability but warns that it may not port to older systems.  I would understand if you decide not to take any action to support an ancient version of sed.

terrelln:
@dkcasset I'm happy to merge something to fix older systems, but it can't break Mac compatibility.

We could test for `-E` and `-r` support and use `-E` if it works, and fall back to `-r`.

dkcasset:
@terrelln I'm thinking it would be simple and reasonable to have a Makefile variable that defaults to -E

terrelln:
That works for me. If we wanted to we could get more fancy like we do with [`HAVE_COLORNEVER`](https://github.com/facebook/zstd/blob/dev/lib/Makefile#L46) for example.

felixhandte:
Works for me. And automated detection ought to be pretty easy to add (either from the start, or as needed in the future).

lib/Makefile
```diff
@@ -49,6 +49,7 @@ ifeq ($HAVE_COLORNEVER, 1)
 GREP_OPTIONS += --color=never
 endif
 GREP = grep $(GREP_OPTIONS)
+SED_ERE_OPT ?= -E
```
Cyan4973, line 52: I'm wondering if it would be better to centralize these changes and options into a second variable `SED`, such as for example :
`SED ?= sed $(SED_ERE_OPT)`

dkcasset: Some of the sed commands use -E while others do not

terrelln: We could switch to `-E` for the other commands and simplify the logic, but let's do that in a separate PR.

I'd rather not add functional changes to a PR that is currently a functional no-op when the user doesn't override `SED_ERE_OPT`.

terrelln:
Thanks for the pull request @dkcasset!

PR #2221, opened 2020-06-29
Install MSVC runtime shared libraries (DLL) with cmake

emmenlau:
The cmake based build currently does not install MSVC shared libraries. This seems to be caused by a rather simple issue: The cmake install command is missing the `RUNTIME` component (see https://cmake.org/cmake/help/latest/command/install.html#installing-targets). The usual install target directory for the `RUNTIME` is `CMAKE_INSTALL_BINDIR`.

This PR adds the `RUNTIME` component to the install command. This should not have any negative consequences for any users. The only behavioral change should be that the shared library (dll) files will be installed into `CMAKE_INSTALL_BINDIR` on MSVC, which should be a purely positive effect.


terrelln:
Thanks for the PR @emmenlau!

emmenlau:
Thanks for the quick response!

PR #2226, opened 2020-07-03
Add ImpromptuNinjas.ZStd C# libzstd bindings link to GitHub Pages branch

Tyler-IN:
Should I add something like (full) or (block, streaming, dictionary, training)?
I think it's the only C# binding that provides OSX .dylib and Linux Musl compatible .so versions as well.
It ships stripped statically linked O3 LTO builds of libzstd for the gamut of platforms that .NET / Fw / Core runs on.
It builds libzstd-mt from zstd version tags and does generic tests for all the platforms using docker images, incl. armv7-hf and arm64v8 via qemu.
It exposes friendly wrappers as well as direct bindings, it has decent testing and coverage.
Performance overhead is minimal as it performs direct `calli` cdecl call site bindings.
It's also signed now, so corpos that need that sort of thing can link their signed assemblies against it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,21 @@
+version: 2.1
+
+jobs:
+  build:
+    docker:
+      - image: alpine:3.7
+    steps:
+      - run:
+          name: Ignored
+          command: echo Ignored
+
+workflows:
+ gh-pages-wf:
+   when:
+    false
+   jobs:
+    - build:
+        filters:
+          branches:
+            ignore:
+              - gh-pages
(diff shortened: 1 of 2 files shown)


Tyler-IN:
Signed CLA ... what's up with the CI?

Tyler-IN:
Looks like you guys need to add this to your**.circleci/config.yml**;
```yml
          branches:
            ignore:
              - gh-pages
```

Tyler-IN:
Turns out it was more annoying than that...

Cyan4973:
Thanks @Tyler-IN

PR #2231, opened 2020-07-08
Use threads in streaming_compression.c example.

### LINKED ISSUE #1584: How to use streaming compression with multi-thread?
I am trying to add multi-thread to [examples/streaming_compression.c](https://github.com/facebook/zstd/blob/470344d33e1d52a2ada75d278466da8d4ee2faf6/examples/streaming_compression.c). I added a few lines in  `compressFile_orDie()`(full code [here](https://gist.github.com/bennyyip/cc45ab593afa848c00fb312384fee525)):
```C
#ifdef ZSTD_MULTITHREAD
    assert(!ZSTD_isError(ZSTD_CCtx_setParameter(cstream, ZSTD_c_nbWorkers, 2)));
#endif

    // size_t const initResult = ZSTD_initCStream(cstream, cLevel);
    // if (ZSTD_isError(initResult)) {
    //     fprintf(stderr, "ZSTD_initCStream() error : %s \n",
    //                 ZSTD_getErrorName(initResult));
    //     exit(11);
    // }
```

It seems that https://github.com/facebook/zstd/blob/470344d33e1d52a2ada75d278466da8d4ee2faf6/lib/compress/zstd_compress.c#L4021 would create `mtctx` only if `cctx` is not initialized(i.e. `cctx->streamStage == zcss_init`). So I comment `ZSTD_initCStream` and set `ZSTD_c_nbWorkers` to 2.

But this modified version of `streaming_compression` dies after calling `ZSTD_endStream` and exit with message "not fully flushed". This can be reproduced on large enough input, for example, [silesia.tar](http://sun.aei.polsl.pl/~sdeor/index.php?page=silesia).

I think maybe I misunderstand the way to use ZSTD. So how to use streaming compression with multi-thread?

marxin:
Fixes #1584.

### DIFF
--- a/examples/streaming_compression.c
+++ b/examples/streaming_compression.c
@@ -39,6 +39,7 @@ static void compressFile_orDie(const char* fname, const char* outName, int cLeve
      */
     CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel, cLevel) );
     CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 1) );
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 4);
 
     /* This loop read from the input file, compresses that entire chunk,
      * and writes all output produced to the output file.


Cyan4973:
While the update correctly states how to use the API in order to use multithread,
the one issue I see is that 
`ZSTD_MULTITHREAD` is a build macro
which exists solely while the library is compiled, but then disappears.
By the time this example is compiled, this macro is no longer there.

Note that it doesn't mean that I've got anything better to propose.
It just underlines that testing this macro is unlikely to provide any signal
unless the local build system do something about it.

marxin:
You are fully right and I think it would be good to inform a potential user about the nice multi-threaded compression.
And wrapping into the very same macro as used for build of `zstd` seems to me logical.
Or should I mention the feature only in a comment?

Cyan4973:
So we have been thinking about it, the question being :
> when a user program links to a local `libzstd` library, which may, or may not, support multithreading,
how can the user program know if multi-threading is supported ?

There are 2 main methods considered, one is compiled based, the other is runtime based.
The compile-based method would typically check a macro.
But it depends on using, at compilation time, a header file which is guaranteed to be representative of the library executable that will be present at link time.

Ultimately, it feels much less powerful than a runtime method.
The runtime method is _always correct_ : it effectively tests the currently linked library.

So do we need a method to test the presence of multithreading support ?
Well, it appears that we already have it.
Invoking `ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, N)`, with N>=1, will return an error code if multithreading is not supported.

Note that, in such a case, execution is not aborted, and the resulting context `cctx` remains completely valid, it just will not trigger multithreading.

Someone willing to test if the locally linked library supports multithreading could just do `mtSupport = ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 1);` and then check `ZSTD_isError(mtSupport)`.

Alternatively, there might be no need to check anything.
`ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 4);` will merely _attempt_ to request 4 worker threads, but if multi-threading support is not enabled, this is essentially a no-op. All it takes is to ignore the return value.

Applying this to your updated example, it means that : 
- checking the build macro is not useful, the inner code will basically be always skipped
- `CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 4) );` will `abort()` execution if the library does not support multi-threading
- Just `ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 4);` alone will work with both single-threaded and multi-threaded libraries
- If you want to make the example more explicit, you could also check the return value and return an information message depending on the `ZSTD_isError()` status.

PR #2236, opened 2020-07-09
Update comment about PGO build.

marxin:
Clang is also supported, so no need to mention only the GCC compiler.


Cyan4973:
If one implies `linux` only, then yes, `clang` is also supported

PR #2239, opened 2020-07-09
cifuzz: don't upload artifacts when CIFuzz fails to build fuzz targets

evverx:
It should address https://github.com/google/oss-fuzz/issues/3982


evverx:
@terrelln I'm not sure whether you're interested in turning on UBSan and MSan on OSS-Fuzz but it's been supported for a while: https://google.github.io/oss-fuzz/getting-started/continuous-integration/#optional-configuration

PR #2244, opened 2020-07-14
Fix -Wunused-variable under FUZZING_BUILD_MODE...

hctim:
Fuzzing build modes (FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION) doesn't
necessarily imply that assert() is enabled, according to the manual.

When the current do-nothing is expanded under -Wunused-variable (-Wall),
it results in unused variables in some of the FUZZING_BUILD_MODE...
blocks.

This patch extends the do-nothing to avoid the unused variable.

### DIFF
--- a/lib/common/debug.h
+++ b/lib/common/debug.h
@@ -79,7 +79,7 @@ extern "C" {
 #  include <assert.h>
 #else
 #  ifndef assert   /* assert may be already defined, due to prior #include <assert.h> */
-#    define assert(condition) ((void)0)   /* disable assert (default) */
+#    define assert(condition) ((void) (condition))   /* disable assert (default) */
 #  endif
 #endif
 


Cyan4973:
Thanks @hctim ,
it seems your PR accidentally exposed a mistake in one `assert()` statement,
explaining the Appveyor test failure.

This can be dealt with in a separate PR.

Cyan4973:
Actually, it's unclear if this is the right thing to do.

In this PR, all `assert()` conditions are changed, now they are just prefixed by `(void)` in release mode.
The return value is not used, so that's fine for simple things such as `assert(value == 0);`.

But what about `assert(function(p) == f2(x));` ?
Chances are, function call will still be present in the code, in case they would generate some side effect. This would impact the release code (slower, heavier binary).
This outcome obviously depends on other factors, such as condition and function complexity, compiler and optimization levels.
But this is no longer "simple", as in "this `assert()` condition just disappears in release mode". Now it may, it may not. 
And it's silent, the worst kind.

Rather than impacting the whole code, if the goal is to fix `-Wunused-variable` under `FUZZING_BUILD_MODE`, it's likely possible to address the issue more directly, by modifying the related code. Effectively, it requires more changes than a single `#define`, but it's also less invasive for the rest of the code base.

PR #2250, opened 2020-07-17
CLI: simplified handling of separated-fields arguments

Cyan4973:
this patch reduces complexity associated with
commands requiring a separated arguments
such as :
`-o filename`
`-D dictionary`
`--output-dir-flat dir`
`--output-dir-mirror dir`
`--maxdict size`
`--dictID id`

It used to be a multi-stage process requiring maintenance of an explicit context,
it's now simplified as a single macro.

Thanks to this simplification,
separated arguments logic has also been extended to
`--patch-from XXX`
`--filelist XXX`
extending current syntax using `=XXX`
`--patch-from=XXX`
`--filelist=XXX`

Separated argument is useful for filenames and directories,
as it benefits from shell expansion
such as `~/dir/file`
where the `~` is automatically translated by the shell.
In contrast `--long-command=FILE` does not interpret `FILE`,
hence `~/` is transmitted as is to the `main()` function,
generally resulting in incorrect file name.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -615,18 +615,27 @@ static int init_cLevel(void) {
     return ZSTDCLI_CLEVEL_DEFAULT;
 }
 
-#define NEXT_FIELD(ptr) {      \
-    argNb++;                   \
-    if (argNb >= argCount) {   \
-        DISPLAY("error: missing command argument \n"); \
-        CLEAN_RETURN(1);       \
-    }                          \
-    ptr = argv[argNb];         \
-    assert(ptr != NULL);       \
-    if (ptr[0]=='-') {         \
-        DISPLAY("error: command cannot be separated from its argument by another command \n"); \
-        CLEAN_RETURN(1);       \
-    }                          \
+#define NEXT_FIELD(ptr) {         \
+    if (*argument == '=') {       \
+        ptr = ++argument;         \
+        argument += strlen(ptr);  \
+    } else {                      \
+        argNb++;                  \
+        if (argNb >= argCount) {  \
+            DISPLAY("error: missing command argument \n"); \
+            CLEAN_RETURN(1);      \
+        }                         \
+        ptr = argv[argNb];        \
+        assert(ptr != NULL);      \
+        if (ptr[0]=='-') {        \
+            DISPLAY("error: command cannot be separated from its argument by another command \n"); \
+            CLEAN_RETURN(1);      \
+}   }   }
+
+#define NEXT_UINT32(val32) {      \
+    const char* __nb;             \
+    NEXT_FIELD(__nb);             \
+    val32 = readU32FromChar(&__nb); \
 }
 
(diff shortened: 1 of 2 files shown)


programs/zstdcli.c
```diff
@@ -615,8 +615,21 @@ static int init_cLevel(void) {
     return ZSTDCLI_CLEVEL_DEFAULT;
 }
 
-#define ZSTD_NB_STRATEGIES 9
+#define NEXT_FIELD(ptr) {      \
```
terrelln, line 618: Would it make sense to handle `--flag ARG` and `--flag=ARG` with the same macro? That way we don't need to list each argument twice.

We could use `longCommandWArg()` to increment `argument` to the `{'=', '\0'}` character, then take care of that case within the macro.

```c
#define NEXT_FIELD(ptr) do { \
    if (argument[0] == '=') { ++argument; ptr = *argument; break; } \
    /* same as before */ \
} while (0);

...

if (longCommandWArg(&argument, "--flag")) { char char* arg; NEXT_FIELD(ptr); continue; }
```

Cyan4973: Yes, that's a good follow-up,
since it would simplify managing both `--long-command`.

Depending on how it's implemented, it might also introduce _new_ syntax, which is so far invalid
such as`-o=FILE`, which is so far not allowed because `-o` is a short command.

Not sure what to think about it.

programs/zstdcli.c
```diff
@@ -957,7 +962,10 @@ int main(int const argCount, const char* argv[])
                 case 't': operation=zom_test; argument++; break;
 
                     /* destination file name */
-                case 'o': nextArgumentIsOutFileName=1; lastCommand=1; argument++; break;
+                case 'o':
+                    NEXT_FIELD(outFileName);
+                    if (!strcmp(outFileName, "-")) outFileName = stdoutmark;
```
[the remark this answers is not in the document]
Cyan4973: ah yes, good point !
`-o -` will fail, since the second `-` will be confused as the beginning of a new command,
so the macro will complain about missing argument, display an error message and then `CLEAN_RETURN()`.

I'm not sure if we need to support this. Chances are, we don't, so it's just better to remove this line.

Cyan4973:
Updated macro to take care of both syntaxes.

programs/zstdcli.c
terrelln: That should be manageable. If `argument` isn't incremented with `-o` then `argument[0] == 'o'`. Only commands that increment `argument`, like with `longCommandWArg()`.

Cyan4973:
There are 2 CLI test errors, but they seem unrelated.

It seems to be some kind of race condition during build of `zwrapbench`, hence a completely different issue, which is potentially impacting other PR.
https://github.com/facebook/zstd/pull/2250/checks?check_run_id=892267138#step:3:387
https://github.com/facebook/zstd/pull/2250/checks?check_run_id=892266951#step:3:828

PR #2253, opened 2020-07-27
optimized histogram

Cyan4973:
new version easier to vectorize
leads to smaller code and faster execution
notably at the last recombination stage (fixed cost per block).

Assembly inspected with godbolt

On my `macosx` laptop, using default `clang` with `-mavx2` :
2K block : 1280 MB/s -> 1550 MB/s
8K block : 1750 MB/s -> 1860 MB/s

Also : fixed a minor error for a corner case (srcSize == 0)

### DIFF
--- a/lib/compress/hist.c
+++ b/lib/compress/hist.c
@@ -71,20 +71,21 @@ static size_t HIST_count_parallel_wksp(
 {
     const BYTE* ip = (const BYTE*)source;
     const BYTE* const iend = ip+sourceSize;
+    size_t const countSize = (*maxSymbolValuePtr + 1) * sizeof(*count);
     unsigned max=0;
     U32* const Counting1 = workSpace;
     U32* const Counting2 = Counting1 + 256;
     U32* const Counting3 = Counting2 + 256;
     U32* const Counting4 = Counting3 + 256;
 
-    memset(workSpace, 0, 4*256*sizeof(unsigned));
-
     /* safety checks */
+    assert(*maxSymbolValuePtr <= 255);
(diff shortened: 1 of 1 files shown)


Cyan4973:
Interestingly, I've been trying to reproduce the result on my desktop, using a slightly different CPU (9700k), a different compiler by default (`gcc-10`) and a different OS (Linux, Ubuntu 20.04).

And I don't reproduce the speed gains of `histogram` function in micro-benchmark.

Interestingly, once combined into a larger `fse_compress.c` workload, it does produce some speed benefit.
It could be either an indirect benefit from using less memory for instructions, or once again a random impact from instruction alignment.

lib/compress/hist.c
```diff
@@ -71,22 +71,21 @@ static size_t HIST_count_parallel_wksp(
...
     U32* const Counting2 = Counting1 + 256;
     U32* const Counting3 = Counting2 + 256;
     U32* const Counting4 = Counting3 + 256;
 
-    memset(workSpace, 0, 4*256*sizeof(unsigned));
-
     /* safety checks */
+    assert(*maxSymbolValuePtr <= 255);
     if (!sourceSize) {
-        memset(count, 0, maxSymbolValue + 1);
```
terrelln, line 85: Can this case be reached in zstd?

I would guess no, since we should emit uncompressed data.

Cyan4973: Within `zstd`, I don't think it can.

It's present to support a more a "general" case as part of the `fse` library.

PR #2257, opened 2020-07-28
MSVC: switch BitScanForward/Reverse to _lzcnt/_tzcnt_u32/64 when BMI2 is available. Added STATIC_BMI2 for detecting BMI2 based on build settings.

Niadb:
This adds a define for STATIC_BMI2 that is similar to DYNAMIC_BMI2 but for statically determining BMI2 support.

It is only targeting MSVC here.

When STATIC_BMI2 is on, many of the calls to BitScanForward/Reverse are replaced with calls to _lzcnt/_tzcnt. 
The codegen for the BitScanForward/Reverse calls was not great on MSVC.

Also if STATIC_BMI2 is enabled, which means the binary is targeting BMI2, DYNAMIC_BMI2 is forced on.   

**Compiler Explorer using _lzcnt_u32:** https://godbolt.org/z/sEqTEE

   cmp     ecx, 63                             ; 0000003fH
        jbe     SHORT $LN3@ZSTD_LLcod
        lzcnt   eax, ecx
        add     eax, 19
        ret     0

**Compiler Explorer using BitScanForward**:https://godbolt.org/z/W8TxT9

   cmp     ecx, 63                             ; 0000003fH
        jbe     SHORT $LN3@ZSTD_LLcod
        xor     eax, eax
        bsr     ecx, ecx
        cmovne  eax, ecx
        add     eax, 19
        ret     0

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -192,13 +192,15 @@
 #  endif
 #endif
 
+#ifndef STATIC_BMI2
+    #define STATIC_BMI2 0
+#endif
+
 /* If STATIC_BMI2 was enabled force DYNAMIC_BMI2 on */
 #if STATIC_BMI2 == 1
 #   ifndef DYNAMIC_BMI2
 #       define DYNAMIC_BMI2 1 //Build it statically targeting BMI2, so enable DYNAMIC_BMI2 if it wasn't already on
 #   endif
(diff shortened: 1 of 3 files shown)


Cyan4973:
What was the issue closing this PR ?

PR #2259, opened 2020-07-29
Fixed VS variable shadowing warning (single file lib)

cwoffenden:
Visual Studio 2019 was generating a variable shadowing warning after the dict builder was rolled-in. This prefixes the affected global and adds `-Wshadow` to the build scripts.


cwoffenden:
Single CI error looks unrelated (`No command 'icc' found`).

PR #2262, opened 2020-08-05
Share stat() Calls Where Convenient

felixhandte:
This stack introduces variants of various `UTIL_...()` functions that take an already prepared result of a `stat()` call, rather than making a `stat()` call themselves. In situations where several of these calls are made in sequence, it's easy to use these new functions to refactor the code to reuse the result of a single `stat()` call rather than make several.

This causes zstd to make 37% fewer `stat()` calls:

```
felixh@vali:~/prog/zstd (dev)$ strace ./zstd -r lib -o /dev/null |& grep -E '^stat\(' | wc -l
730
felixh@vali:~/prog/zstd (fewer-stat-syscalls)$ strace ./zstd -r lib -o /dev/null |& grep -E '^stat\(' | wc -l
460
```

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -155,7 +155,6 @@ int UTIL_chmod(char const* filename, const stat_t* statbuf, mode_t permissions);
  * compute the needed information.
  */
 
-int UTIL_fileExist(const char* filename);
 int UTIL_isRegularFile(const char* infilename);
 int UTIL_isDirectory(const char* infilename);
 int UTIL_isSameFile(const char* file1, const char* file2);
--- a/programs/util.h
+++ b/programs/util.h
@@ -120,18 +120,24 @@ extern int g_utilDisplayLevel;
  * Returns success (1) or failure (0).
  */
 int UTIL_stat(const char* filename, stat_t* statbuf);
-int UTIL_statFile(const char* infilename, stat_t* statbuf); /* also check it's a file */
-int UTIL_statDir(const char* infilename, stat_t* statbuf); /* also check it's a directory */
+/** Also checks that the target is a regular file. */
+int UTIL_statFile(const char* infilename, stat_t* statbuf);
+/** Also checks that the target is a directory. */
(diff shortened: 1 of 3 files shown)


programs/util.h
```diff
@@ -113,20 +115,31 @@ extern int g_utilDisplayLevel;
...
+/**
+ * Calls platform's equivalent of stat() on filename and writes info to statbuf.
+ * Returns success (1) or failure (0).
+ */
+int UTIL_stat(const char* filename, stat_t* statbuf);
+/** Also checks that the target is a regular file. */
+int UTIL_statFile(const char* infilename, stat_t* statbuf);
+/** Also checks that the target is a directory. */
+int UTIL_statDir(const char* infilename, stat_t* statbuf);
 int UTIL_fileExist(const char* filename);
```
Cyan4973, line 127: I was expecting that the new API, which reads `stat()` once and exploits the outcome across multiple subsequent invocations, was replacing the existing one. But I see that older API prototypes are still there, so it's not a complete replacement, both live side by side.

As a consequence, there are now multiple entry points in the `UTIL_` API which do approximately the same thing, with subtle differences.

I believe it would be helpful for the different API strategies to be grouped together.

Typically, we would have 2 families,
on one side, "integrated" functions, such as  `UTIL_isDirectory(const char* infilename);` (old API),
on the other side, `UTIL_stat()` + "read-only" functions, such as `UTIL_isDirectoryStat(const stat_t* statbuf); (new API).

This also opens the question of some new _public_ prototypes which are "between the two worlds", such as `UTIL_statFile(const char* infilename, stat_t* statbuf);`, which is a sort of combined `UTIL_stat()` + `UTIL_isRegularFileStat()`. It makes the API design a bit more messy, and the benefit does not seem obvious to me. I'm not convinced they are worth the complexity they introduce.

Cyan4973:
Excellent, thanks !

PR #2264, opened 2020-08-10
Prepare for freestanding library and add initial translator script

terrelln:
This pulls all stdlib dependencies (except `<stddef.h>` and `<limits.h>` in exported headers) into `zstd_deps.h`. It is separated into several sections:

1. Common definitions from: `<limits.h>`, `<stddef.h>`, `<string.h>`, and integer types.
2. Malloc / free / calloc
3. assert
4. stdio: Only requires a new macro `ZSTD_DEBUG_PRINT()` used for `DEBUGLOG()`

The common definitions are always provided. The other three sections are hidden behind macros. You have to define, for example, `ZSTD_DEPS_NEED_MALLOC`.

This also adds a prototype freestanding translator script. I still need to add some functionality, but the structure is there. A bare-bones test case compiled in freestanding mode `-ffreestanding` is also added. I will test all the functionality needed for the kernel needs, and run that test in CI. This will ensure that the translation continues to work over time, and updating the kernel version is easy.

The freestanding test will ensure that no one can accidentally introduce a call to `memcpy()` instead of `ZSTD_memcpy()`, for example. Since the compilation would emit a error with `-Werror`.


terrelln:
I think that I do want to rewrite all functions to be `ZSTD_` prefixed (excepting `<stddef.h>` and `<limits.h>`). That way a freestanding zstd can live in harmony with a library compiled with the stdlib. That will allow us, for example, to provide a malloc function that will fail, without having to compile in freestanding mode.

However, I will do that in a second PR that only does the renaming, so this doesn't become super noisy.

terrelln:
I went ahead and included it in this PR, but the two renamings are done in their own commits:

1. ZSTD_malloc -> ZSTD_customMalloc.
2. ZSTD_ prefix all stdlib functions.

I would recommend looking at the 4 commits individually.

contrib/freestanding_lib/freestanding.py
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env python3
+
```
Cyan4973, line 2: should this file contain a license header ?

terrelln: Yup

terrelln:
Good news, I just got a kernel with zstd-1.4.5 booting (with several more commits on top of this PR)! It is ~25% faster than the current zstd version.

lib/common/zstd_deps.h
```diff
@@ -0,0 +1,118 @@
...
+ * ptrdiff_t
+ * INT_MAX
+ * ...
+ */
+#ifndef ZSTD_DEPS_COMMON
+#define ZSTD_DEPS_COMMON
+
+#include <limits.h>
+#include <stddef.h>
+#include <string.h>
```
Cyan4973, line 30: This seems to include `<string.h>` everywhere, 
notably in numerous places where only `<stddef.h>` was previously required.

This feels slightly overkill.

`string.h` is notably useful for `memcpy()` and `memset()`, and occasionally for additional functions.
I suspect these places are not that numerous.
Typically, I expect this situation to happen in some `*.c` units, and (generally) not in `*.h` headers.

lib/compress/fse_compress.c
```diff
@@ -25,6 +23,8 @@
 #define FSE_STATIC_LINKING_ONLY
 #include "../common/fse.h"
 #include "../common/error_private.h"
+#define ZSTD_DEPS_NEED_MALLOC
+#include "../common/zstd_deps.h"  /* malloc, free, memcpy, memset */
```
Cyan4973: minor : with all these little changes (not just this one),
we have made it more difficult to synchronize `fse` between its own library and the `zstd` repository.

It's now more firmly integrated as a specific `zstd` unit,
making it a bit less generic, hence more difficult to re-use elsewhere

Maybe something to think about, though likely a topic for a different PR.

lib/common/zstd_deps.h
terrelln: `"mem.h"` currently includes `<string.h>`, so that means it is basically everywhere. `zstd_internal.h` needs memcpy.

Cyan4973: `mem.h` is indeed included in many `*.c` units,
but I believe we paid attention to not include it unduly into most `*.h` headers.

Cyan4973: You have already paid attention to not include "everything" when including `zstd_deps.h`
thanks to the `ZSTD_DEPS_NEED_XXX` pattern.

I guess it could be extended to `<string.h>`.

In which case, `mem.h` could include `zstd_deps.h` with `ZSTD_DEPS_NEED_STRING` (for example)
and that would make it equivalent to what we have today,
with `<string.h>` indirectly included when `mem.h` is included.

Note that I do not extend this request to `<limits.h>`.
`<limits.h>` only contains a few constants, so it's harmless.
`<mem.h>` in contrast includes many functions in an expanded "ready-to-inline" format, so it's an entirely different payload.

terrelln: All but a few headers in `common/` like `debug.h` get it. Most headers end up including one of `zstd_internal.h`, `zstd_compress_internal.h`, or `zstd_decompress_internal.h`. All of which indirectly include `<string.h>`.

I don't think the gains of not including `<string.h>` in a small number of `common/` headers pay for the cost of having to `#define ZSTD_DEPS_NEED_STRING` before nearly every include of `zstd_deps.h`.

The important thing is that we don't depend on `zstd_deps.h` in our public headers `zstd.h` and `zstd_errors.h`, which we don't.

terrelln: I'm not against a cleanup for headers that only need `size_t`. In those cases we don't need `zstd_deps.h`, instead we can include `<stddef.h>` like `zstd.h` does.

I'm also not against a cleanup of `zstd_*_internal.h` to split it up into smaller pieces that are only types/declarations that don't need `memcpy()`, and thus don't need `zstd_deps.h`.

But that isn't the current state of the repo, so I think that can happen in a separate PR.

Cyan4973: > I don't think the gains of not including `<string.h>` in a small number of `common/` headers pay for the cost of having to `#define ZSTD_DEPS_NEED_STRING` before nearly every include of `zstd_deps.h`.

I agree this wouldn't be a good trade-off.
But I was under the perhaps misleading assumption that this would only be necessary in `mem.h`, since most other files would include it through this channel.

terrelln: > But I was under the perhaps misleading assumption that this would only be necessary in mem.h, since most other files would include it through this channel.

That would work. Though I'm a fan of including what you use in the file directly, and not depending on transitive includes.

Cyan4973:
> There is still some more work to do though:

Which additional follow-ups do you believe should be part of this PR  ?

terrelln:
> Which additional follow-ups do you believe should be part of this PR ?

I have several more commits that I was planning on putting in a separate PR. These are their titles as a summary (with some comments):

* Remove <limits.h> dependency from zstd.h (I think this may actually be unnecessary, TODO)
* [API] Add the ZSTD_NO_MALLOC macro to hide allocating functions
* [lib] s/current/curr because it collides with Linux Kernel macro (rename all varaibles `current` to `curr`)
* [contrib] Add preprocessor hardwiring to freestanding.py (more functionality for `freestanding.py`)
* Add linux-kernel freestanding (update `contrib/linux-kernel` to use `freestanding.py`)

Cyan4973:
As I was going through these changes, I was considering a slightly different way of achieving the same goals.

The core idea is that, instead of having a single `zstd_deps.h`,
there would exist multiple libc equivalents,
such as `zstd_stdlib.h`, `zstd_assert.h`, `zstd_string.h`, etc.

Then these files would contain what's required in their respective context (standard clib, or freestanding environment).

The advantage is that the changes, meaning and scope feel fairly straightforward : 
where there used to be `#include <string.h>`, not there's `#include "zstd_string.h"`,
which feels intuitive, especially with regards to understanding the scope being included,
and avoids the constraint of manipulating build macros.

The disadvantage is that there are now several files to adjust for freestanding environment,
instead of a single place.

If that matters, this property could still be achieved in a manner which mostly preserves current code.
Typically, `zstd_stdlib.h` would merely contain something like :
``` 
#define ZSTD_DEPS_NEED_STDLIB
#include "zstd_deps.h"`
```
therefore preserving the current design with all adaptations happening in one place.

Mostly for discussion.
I know you have already invested a lot of time in this PR, 
so this is only appropriate to consider if it keeps the workload at an acceptable level.

terrelln:
> The disadvantage is that there are now several files to adjust for freestanding environment,
instead of a single place.

I'm okay with this direction. It wouldn't be much work to switch (this header is the easy part). I'm open to discussion.

The reason I chose one header is it is really easy to see what our dependencies are, and it is super easy to replace it. I think that it may be important, because someone compiling zstd in freestanding mode won't be as familiar with the codebase, so having one file to replace may help.

PR #2267, opened 2020-08-13
Fix clang -Wcomma warning (single file decoder)

cwoffenden:
Minor change but newer Clang/Xcode projects seem to have the `-Wcomma` warning enabled by default.

There are more of these in the codebase in general, and it's a style I use myself, so wasn't sure whether adding the various compiler equivalents of `-Wno-comma` would be better? For the single file _decoder_ there's only one so it became a one-liner.


Cyan4973:
There are cases where the comma style can be really useful,
so I wouldn't feel comfortable to blindly remove them all,

but in this case, there is no perceptible advantage to comma style,
so this change is totally fine.

PR #2268, opened 2020-08-13
[actions] removing broken icc test from travis and adding working version to actions

bimbashrestha:
The directory structure for icc's installation had changed so it couldn't find he setvars.sh it needed. I now just reference the compiler path for the latest icc directly which shouldn't change.


Cyan4973:
Thanks @bimbashrestha !

PR #2271, opened 2020-08-15
Small block optimizations

terrelln:
* Add a benchmark `smallbench` that benchmarks header decoding for small blocks
* Optimize `ZSTD_buildFSETable()` (see inline comments)
* Optimize `FSE_readNCount()` (see inline comments)

Benchmarks run on `silesia.tar` broken up into 4KB blocks. After is compiled with BMI2 support.

| Benchmark | Compiler | Before | After | % change |
|---|---|---|---|---|
| ZSTD_decodeLiteralsHeaders | gcc | 2218 MB/s | 4245 MB/s | +91% |
| ZSTD_decodeLiteralsHeaders | clang | 2161 MB/s | 4376 MB/s | +102% |
| ZSTD_decodeSeqHeaders | gcc | 2888 MB/s | 4535 MB/s | +57% |
| ZSTD_decodeSeqHeaders | clang | 2509 MB/s | 4530 MB/s | +80% |
| 4k block decompression (raw lits) | gcc | 828 MB/s | 956 MB/s | +15.4% |
| 4k block decompression (raw lits) | clang | 777 MB/s | 917 MB/s | +18% |
| 4k block decompression (huf lits) | gcc | 484 MB/s | 590  MB/s | +21.9% |
| 4k block decompression (huf lits) | clang | 465 MB/s | 581 MB/s | +24.9% |

Note: The clang after binary lucked into a bad alignment and has 5% less performance with 128KB blocks, which should be about neutral with these optimizations, so this actually going to be a slightly larger win.

TODO:
* Check out `HUF_readStats()`
* Emit -1 based on number of symbols, instead of `LOW_PROB` macro
* Clean up the code
* Add some tests
* Fuzz it


lib/decompress/zstd_decompress_block.h
```diff
@@ -53,7 +53,7 @@ size_t ZSTD_decompressBlock_internal(ZSTD_DCtx* dctx,
 void ZSTD_buildFSETable(ZSTD_seqSymbol* dt,
              const short* normalizedCounter, unsigned maxSymbolValue,
              const U32* baseValue, const U32* nbAdditionalBits,
-                   unsigned tableLog);
+                   unsigned tableLog, U32* wksp, size_t wkspSize);
```
Cyan4973, line 56: minor : doc : working conditions regarding `wkspSize` should be described in function description

Cyan4973:
minor : it seems that the reason for `smallbench` to exist
is to have a way to measure speed of internal functions, such as `ZSTD_decodeSeqHeaders()`,
but I note that `fullbench` was also created for the same purpose.
Am I missing something ? 
If not, could they be joined ?

terrelln:
> but I note that fullbench was also created for the same purpose.
> Am I missing something ?
> If not, could they be joined ?

They could potentially be joined. I split it out from fullbench because I wanted to run on all the blocks in the file, not just the first block, and it was faster to write a new benchmark. Before I merge, I could spend the time to upgrade fullbench to support splitting the input file into blocks with `-B` is set, and run the benchmark over all the blocks, like I do in smallbench.

terrelln:
Speed up literals header decoding by 90%+

terrelln:
I'm fixing up the tests now, then checking over the code.

Will remove the `[WIP]` label once I believe this is ready for merge/review.

tests/regression/results.csv
```diff
@@ -1,611 +1,611 @@
...
-silesia,                            multithreaded long distance mode,   compress cctx,                      4849491
-silesia,                            small window log,                   compress cctx,                      7078156
-silesia,                            small hash log,                     compress cctx,                      6554898
-silesia,                            small chain log,                    compress cctx,                      4931093
-silesia,                            explicit params,                    compress cctx,                      4794609
-silesia,                            uncompressed literals,              compress cctx,                      4849491
-silesia,                            uncompressed literals optimal,      compress cctx,                      4293262
-silesia,                            huffman literals,                   compress cctx,                      6178442
-silesia,                            multithreaded with advanced params, compress cctx,                      4849491
+silesia.tar,                        level -5,                           compress simple,                    6738593
```
Cyan4973, line 2: It seems there is a general tiny compression ratio regression with this PR.
I presume this is because blocks generating < 4K symbols actually exist quite naturally in `silesia.tar`,
even with 128K full-size blocks,
and the removal of `-1` weight in these cases is slightly detrimental to compactness.

terrelln: Yeah, that is correct

lib/common/entropy_common.c
```diff
@@ -118,31 +160,85 @@ size_t FSE_readNCount (short* normalizedCounter, unsigned* maxSVPtr, unsigned* t
...
+    /* Only possible when there are too many zeros. */
+    if (charnum > maxSV1) return ERROR(maxSymbolValue_tooSmall);
     if (bitCount > 32) return ERROR(corruption_detected);
     *maxSVPtr = charnum-1;
 
     ip += (bitCount+7)>>3;
     return ip-istart;
 }
 
+static size_t FSE_readNCount_body_default(
```
Cyan4973, line 206: I presume the purpose of this function is to avoid the `FORCE_INLINE` statement of `FSE_readNCount_body` ?
Maybe this could be documented.

lib/common/fse.h
```diff
@@ -137,10 +137,12 @@ FSE_PUBLIC_API unsigned FSE_optimalTableLog(unsigned maxTableLog, size_t srcSize
 /*! FSE_normalizeCount():
     normalize counts so that sum(count[]) == Power_of_2 (2^tableLog)
     'normalizedCounter' is a table of short, of minimum size (maxSymbolValue+1).
+    useLowProbCount is a bool param which is set to 1 to use count=-1 or set to 0 to
```
Cyan4973, line 140: While doc is technically correct, it will be hard for external `fse` users to understand what is going on here.
Also, it's a breaking API change, and there is no other way to request `normalizeCount()`, so a user _must_ understand `useLowProbCount` to invoke this function.

Suggestion : 
- Provide benefits pros/cons of each setting (partially done here for `useLowProbCount==0`). Insist more on compressed / speed benefit, rather than the "technicality" of `count-=1` which might feel obscure at this level of the interface.
- Provide guidance on "good" defaults

Cyan4973:
Looks good to me.
Only some minor documentation comments.

lib/common/entropy_common.c
terrelln: Yeah, this is the pattern that we've been using in all the other places we use BMI2.

Cyan4973:
It depends.
In many cases, `ZFS` and `BtrFS` define some pretty "large" blocks, in the range of ~128 KB.
This is larger than what this optimization is intended for (~4 KB blocks).

terrelln:
Yeah, this won't change much the speed of BtrFS decompression, because it uses a 128KB block size. The exception might be if you have a lot of small files, since I think each file is compressed individually (though I may be wrong). I would guess ZFS is the same, but I'm not 100% sure what ZFS's block size is.

Cyan4973:
ZFS block size is configurable, but 128 KB is a fairly standard setup.

terrelln:
> 16K is not unheard of for DB loads ;-)

Yeah, we would expect to see a 10% speed boost at this block size. https://github.com/facebook/zstd/releases/tag/v1.4.7 contains the approximate decompression speed gains by data size.

PR #2276, opened 2020-08-18
DDSS For Lazy: Implement a Dedicated Dictionary Hash Table

felixhandte:
Builds on top of @bimbashrestha's #2201 to implement a new search structure for dictionaries in the greedy, lazy, and lazy2 
strategies. This is a work-in-progress.

Expected performance increases are approximately as follows:

Level | Hot Dict | Cold Dict
----- | -------- | ---------
  5   |   ~+10%  |    ~+35%
  6   |   ~+10%  |    ~+30%
  7   |   ~+ 8%  |    ~+25%
  8   |   ~+ 6%  |    ~+17%
  9   |   ~+ 6%  |    ~+14%
 10   |   ~+ 5%  |    ~+12%

The improvement degrades as the input gets large. Compression ratio is slightly improved at level 5 (~0.05%). Benchmarked on an Intel Xeon E5-1650 v3 @ 3.50GHz (Haswell), with gcc 8.4.0. With recent fixes, I'm not aware of any scenario where perf is worse with the DDSS than without.

To-Do:

- [x] Document performance impacts, including corpora other than `http`.
- [x] Remove `ZSTD_dedicatedDictSearch_defaultCParameters` and replace with a lookup in the regular params, that is then adjusted?
- [x] Settle on a signature for `ZSTD_createCDict_advanced2()` and document it.

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -574,8 +574,7 @@ size_t ZSTD_HcFindBestMatch_generic (
         const U32 ddsIndexDelta        = dictLimit - ddsSize;
         const U32 ddsMinChain = ddsSize > ddsChainSize ? ddsSize - ddsChainSize : 0;
         const U32 bucketSize           = (1 << DD_BLOG);
-
-        U32 attemptNb = 1;
+        U32 attemptNb;
 
         matchIndex = dms->hashTable[ddsIdx];
 
@@ -583,7 +582,11 @@ size_t ZSTD_HcFindBestMatch_generic (
         if (!matchIndex)
             return ml;
 
-        for ( ; (matchIndex>ddsLowestIndex) & (nbAttempts>0) ; nbAttempts--, attemptNb++) {
+        for (attemptNb = 0; attemptNb < bucketSize; attemptNb++) {
+            PREFETCH_L1(ddsBase + dms->hashTable[ddsIdx + attemptNb]);
+        }
+
+        for (attemptNb = 1; (matchIndex>ddsLowestIndex) & (nbAttempts>0) ; nbAttempts--, attemptNb++) {
             size_t currentMl=0;
             const BYTE* const match = ddsBase + matchIndex;
             assert(match+4 <= ddsEnd);
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -475,7 +475,7 @@ U32 ZSTD_insertAndFindFirstIndex(ZSTD_matchState_t* ms, const BYTE* ip) {
     return ZSTD_insertAndFindFirstIndex_internal(ms, cParams, ip, ms->cParams.minMatch);
 }
 
-void ZSTD_lazy_loadDictionary(ZSTD_matchState_t* ms, const BYTE* const ip)
+void ZSTD_dedicatedDictSearch_lazy_loadDictionary(ZSTD_matchState_t* ms, const BYTE* const ip)
 {
     U32 const target = (U32)(ip - ms->window.base);
     U32* const chainTable = ms->chainTable;
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -258,6 +258,7 @@ ZSTD_DUBT_findBestMatch(ZSTD_matchState_t* ms,
 
     DEBUGLOG(7, "ZSTD_DUBT_findBestMatch (%u) ", curr);
     assert(ip <= iend-8);   /* required for h calculation */
+    assert(dictMode != ZSTD_dedicatedDictSearch);
 
     /* reach end of unsorted candidates list */
     while ( (matchIndex > unsortLimit)
@@ -525,7 +526,7 @@ size_t ZSTD_HcFindBestMatch_generic (
     /* HC4 match finder */
     U32 matchIndex = ZSTD_insertAndFindFirstIndex_internal(ms, cParams, ip, mls);
 
-    if (dictMode == ZSTD_dictMatchState && ms->dictMatchState->enableDedicatedDictSearch)
+    if (dictMode == ZSTD_dedicatedDictSearch)
         PREFETCH_L1(ms->dictMatchState->hashTable +
             (ZSTD_hashPtr(ip, ms->dictMatchState->cParams.hashLog - DD_BLOG,
             ms->dictMatchState->cParams.minMatch) << DD_BLOG));
@@ -555,7 +556,7 @@ size_t ZSTD_HcFindBestMatch_generic (
         matchIndex = NEXT_IN_CHAIN(matchIndex, chainMask);
     }
 
-    if (dictMode == ZSTD_dictMatchState && ms->dictMatchState->enableDedicatedDictSearch) {
+    if (dictMode == ZSTD_dedicatedDictSearch) {
         const ZSTD_matchState_t* const dms = ms->dictMatchState;
         const U32 dmsChainSize         = (1 << dms->cParams.chainLog);
         const U32 dmsChainMask         = dmsChainSize - 1;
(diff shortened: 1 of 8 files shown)


lib/compress/zstd_lazy.c
```diff
@@ -531,8 +564,55 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+                ml = currentMl;
+                *offsetPtr = current - (matchIndex + ddsIndexDelta) + ZSTD_REP_MOVE;
+                if (ip+currentMl == iLimit) break; /* best possible, avoids read overflow on next attempt */
+            }
+
+            if (matchIndex <= ddsMinChain) {
+                break;
+            }
+
+            if (attemptNb < bucketSize) {
```
terrelln, line 609: I wonder if it would be better to have two loops. One for the in-bucket attempts and one for the out of bucket attempts?

I guess it depends on how main chains fall out of the bucket.

lib/compress/zstd_lazy.c
```diff
@@ -531,8 +564,55 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+
+        /* Empty chain */
+        if (!matchIndex)
+            return ml;
+
+        for (attemptNb = 0; attemptNb < bucketSize; attemptNb++) {
+            PREFETCH_L1(ddsBase + dms->hashTable[ddsIdx + attemptNb]);
+        }
+
+        for (attemptNb = 1; (matchIndex>ddsLowestIndex) & (nbAttempts>0) ; nbAttempts--, attemptNb++) {
```
terrelln, line 589: small thing: It might be slightly beneficial to not keep track of `attemptNb` and `nbAttempts`. Might be a better way to write it.

lib/compress/zstd_lazy.c
```diff
@@ -531,8 +564,55 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+            }
+
+            /* save best solution */
+            if (currentMl > ml) {
+                ml = currentMl;
+                *offsetPtr = current - (matchIndex + ddsIndexDelta) + ZSTD_REP_MOVE;
+                if (ip+currentMl == iLimit) break; /* best possible, avoids read overflow on next attempt */
+            }
+
+            if (matchIndex <= ddsMinChain) {
```
terrelln, line 605: We should be able to get rid of this branch through table construction. Either set our chain big enough, or if we see a match index that doesn't fit in our chain set it to 0.

lib/compress/zstd_lazy.c
```diff
@@ -644,20 +741,21 @@ ZSTD_compressBlock_lazy_generic(
...
                                          : ZSTD_HcFindBestMatch_dictMatchState_selectMLS) :
+        (dictMode == ZSTD_dedicatedDictSearch ? ZSTD_HcFindBestMatch_dedicatedDictSearch_selectMLS :
         (searchMethod==search_binaryTree ? ZSTD_BtFindBestMatch_selectMLS
-                                         : ZSTD_HcFindBestMatch_selectMLS);
+                                         : ZSTD_HcFindBestMatch_selectMLS));
     U32 offset_1 = rep[0], offset_2 = rep[1], savedOffset=0;
 
     const ZSTD_matchState_t* const dms = ms->dictMatchState;
-    const U32 dictLowestIndex      = dictMode == ZSTD_dictMatchState ?
+    const U32 dictLowestIndex      = dictMode == ZSTD_dictMatchState || dictMode == ZSTD_dedicatedDictSearch ?
```
terrelln, line 750: nit: We should have a constant for this. If the compiler gets confused about propagating the constants it could be a macro instead.

Cyan4973: Parenthesis could help readability

lib/compress/zstd_lazy.c
felixhandte: Hmm, I'm not sure.

1. It's possible but non-trivial to enforce this in table construction: dictionaries get loaded in chunks and so the table-filling code doesn't 100% know where the end of the dictionary is.

2. That would mean potentially losing matches that are pointed at by a valid chainTable entry, but that themselves are not valid chainTable indices.

terrelln:
What are the plans for future work?

I ran a simple test where I disabled the chain table search (so we only search the hash table). This improves performance by 25%, with a loss to compression ratio. This shouldn't have a huge impact on the number of searches, since most chains are <= 4. But, halves the number of LLC misses from 37M to 18M.

I suspect we should be able to get at least 2/3 of that 25% win by improving the chain table structure. Is that something you are planning on doing, either in this PR or another?

lib/compress/zstd_lazy.c
```diff
@@ -531,8 +569,74 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+        const BYTE* const ddsBase = dms->window.base;
+        const BYTE* const ddsEnd  = dms->window.nextSrc;
+        const U32 ddsSize         = (U32)(ddsEnd - ddsBase);
+        const U32 ddsIndexDelta   = dictLimit - ddsSize;
+        const U32 ddsMinChain     = ddsSize > ddsChainSize ? ddsSize - ddsChainSize : 0;
+        const U32 bucketSize      = (1 << ZSTD_LAZY_DDSS_BUCKET_LOG);
+        const U32 bucketLimit     = nbAttempts < bucketSize ? nbAttempts : bucketSize;
+        U32 ddsAttempt;
+
+        for (ddsAttempt = 0; ddsAttempt < bucketSize; ddsAttempt++) {
```
terrelln, line 585: I see 20% of this functions cycles spent in this loop. That suggests to me that the earlier prefetch isn't early enough. Can we do better?

lib/compress/zstd_lazy.c
```diff
@@ -503,10 +528,23 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+    const U32 ddsHashLog = dictMode == ZSTD_dedicatedDictSearch
+                         ? dms->cParams.hashLog - ZSTD_LAZY_DDSS_BUCKET_LOG : 0;
+    const size_t ddsIdx = dictMode == ZSTD_dedicatedDictSearch
+                        ? ZSTD_hashPtr(ip, ddsHashLog, mls) << ZSTD_LAZY_DDSS_BUCKET_LOG : 0;
+
+    U32 matchIndex;
+
+    if (dictMode == ZSTD_dedicatedDictSearch) {
+        const U32* entry = &dms->hashTable[ddsIdx];
+        PREFETCH_L1(entry);
```
terrelln, line 541: I made this super crude change (which may access out of bounds memory, but whatever). I saw a speed boost of ~10%. That confirms that we aren't prefetching early enough.


```suggestion
        const size_t ddsIdx1 = dictMode == ZSTD_dedicatedDictSearch
                            ? ZSTD_hashPtr(ip+1, ddsHashLog, mls) << ZSTD_LAZY_DDSS_BUCKET_LOG : 0;
        const U32* entry = &dms->hashTable[ddsIdx];
        const U32* entry1 = &dms->hashTable[ddsIdx1];
        PREFETCH_L1(entry);
        PREFETCH_L1(entry1);
```

terrelln: That likely also means that we could also prefetch the match pointers earlier too.

lib/compress/zstd_lazy.c
terrelln: Even with the prefetch ip+1 change I suggested below we are spending 10% of cycles in this loop. So we could likely do even better.

Also prefetching ip+2 gives another 2% of speed in cold dictionaries. Fetching ip+2 we still spend 8% of cycles in this loop. I don't think that simple strategy is the right approach, but it shows that we can to do better in the final version of the code.

For example, we could move the prefetch out of the search function. Then:
* If we did not find a match prefetch the next (1, 2, 3, ...?) positions? Maybe # of positions depends on strategy. Would have to take care not to double-prefetch. This makes sense for lazy because we will search the next few positions, even if we find a match in the first search, since this is the lazy strategy. In greedy, it still makes sense, because we may predict we will find a few literals before a match. But maybe we should be less aggressive.
* If we find a match prefetch ip + matchLength + (0, 1, or 2, ...) positions before doing anything else. Similar to above. But we should also consider rep-codes, and we may be more likely to find a 2nd match directly after a match than searching a position where we just saw a literal.

terrelln:
At level 6, with the http data set, with 100 dictionaries in the `largeNbDicts` benchmark I see +22% performance with the current code. With the extra prefetch + removing the chain table search I see +75% speed and a 0.5% loss of compression ratio from dropping the chain table search.

felixhandte:
@terrelln, awesome experiments! Yeah, I've definitely not been focusing as much on the prefetching side of things. We should talk more about it in our meeting today.

I spent last evening and this morning putting together a prototype of the sorted chaintable which I just got working. Without match prefetching, I see an incremental ~+20% at the high end of the affected levels. So I do expect to follow this PR up with that shortly.

PR #2278, opened 2020-08-24
New advanced decompression param to ignore checksums

senhuang42:
This adds the advanced parameter `ZSTD_d_forceIgnoreChecksum` as a new experimental decompression parameter. If enabled, then decompression will ignore checksums, whether or not they were enabled during compression.

A few considerations/points of discussion:
- Open to other naming suggestions - not sure if there’s a general consensus against negative actions in param names (ignore), but I wanted to emphasize that we’re going out of our way to ignore the checksum behavior that the compressor specified, and that this isn't typical behavior, hence `forceIgnore`. But, done the other way, we avoid all the `!` in the logic implementing this.
- This strictly ignores the value of the checksum, i.e. if the checksum flag is enabled in the compressed stream, but the checksum size is not 4 bytes, enabling `ZSTD_d_forceIgnoreChecksum` will not prevent decompression from throwing an error. In my mind, the checksum size somehow being wrong seems like a very serious error, and the new functionality probably shouldn't ignore that - but, let me know what your thoughts are
- I noticed `--no-check` was a compression CLI flag. I’ve added support for `--no-check` to be used as as decompression parameter to enable this new advanced param. We can add a different one, but I think the overload makes sense since it's documented and seems a little less bloat-y.
- In addition to testing the new functionality, the added unit test in fuzzer.c also tests that if checksums are corrupt, they should fail to decompress under normal circumstances - I couldn't find an explicit test for this already

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1397,6 +1397,11 @@ size_t ZSTD_DCtx_setFormat(ZSTD_DCtx* dctx, ZSTD_format_e format)
     return ZSTD_DCtx_setParameter(dctx, ZSTD_d_format, format);
 }
 
+size_t ZSTD_DCtx_setForceSkipChecksum(ZSTD_DCtx* dctx, ZSTD_format_e format)
+{
+    return ZSTD_DCtx_setParameter(dctx, ZSTD_d_forceSkipChecksum, format);
+}
+
 ZSTD_bounds ZSTD_dParam_getBounds(ZSTD_dParameter dParam)
 {
     ZSTD_bounds bounds = { 0, 0, 0 };
@@ -1414,6 +1419,9 @@ ZSTD_bounds ZSTD_dParam_getBounds(ZSTD_dParameter dParam)
             bounds.lowerBound = (int)ZSTD_obm_buffered;
             bounds.upperBound = (int)ZSTD_obm_stable;
             return bounds;
+        case ZSTD_d_forceIgnoreChecksum:
+            bounds.lowerBound = (int)ZSTD_d_validateChecksum;
+            bounds.upperBound = (int)ZSTD_d_ignoreChecksum;
         default:;
     }
     bounds.error = ERROR(parameter_unsupported);
@@ -1453,6 +1461,9 @@ size_t ZSTD_DCtx_setParameter(ZSTD_DCtx* dctx, ZSTD_dParameter dParam, int value
             CHECK_DBOUNDS(ZSTD_d_stableOutBuffer, value);
             dctx->outBufferMode = (ZSTD_outBufferMode_e)value;
             return 0;
+        case ZSTD_d_forceIgnoreChecksum:
+            CHECK_DBOUNDS(ZSTD_d_forceIgnoreChecksum, value);
+            dctx->forceIgnoreChecksum = (ZSTD_ignoreChecksumMode_e)value;
         default:;
     }
     RETURN_ERROR(parameter_unsupported, "");
@@ -1524,7 +1535,7 @@ static void ZSTD_DCtx_updateOversizedDuration(ZSTD_DStream* zds, size_t const ne
 {
     if (ZSTD_DCtx_isOverflow(zds, neededInBuffSize, neededOutBuffSize))
         zds->oversizedDuration++;
-    else 
+    else
         zds->oversizedDuration = 0;
 }
 
@@ -1731,7 +1742,7 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
 
                 {   int const tooSmall = (zds->inBuffSize < neededInBuffSize) || (zds->outBuffSize < neededOutBuffSize);
                     int const tooLarge = ZSTD_DCtx_isOversizedTooLong(zds);
-                    
+
                     if (tooSmall || tooLarge) {
                         size_t const bufferSize = neededInBuffSize + neededOutBuffSize;
                         DEBUGLOG(4, "inBuff  : from %u to %u",
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1398,9 +1398,9 @@ size_t ZSTD_DCtx_setFormat(ZSTD_DCtx* dctx, ZSTD_format_e format)
     return ZSTD_DCtx_setParameter(dctx, ZSTD_d_format, format);
 }
 
-size_t ZSTD_DCtx_setForceIgnoreChecksum(ZSTD_DCtx* dctx, ZSTD_format_e format)
+size_t ZSTD_DCtx_setForceIgnoreChecksum(ZSTD_DCtx* dctx, ZSTD_forceIgnoreChecksum_e shouldIgnore)
 {
-    return ZSTD_DCtx_setParameter(dctx, ZSTD_d_forceIgnoreChecksum, format);
+    return ZSTD_DCtx_setParameter(dctx, ZSTD_d_forceIgnoreChecksum, shouldIgnore);
 }
 
 ZSTD_bounds ZSTD_dParam_getBounds(ZSTD_dParameter dParam)
@@ -1423,6 +1423,7 @@ ZSTD_bounds ZSTD_dParam_getBounds(ZSTD_dParameter dParam)
         case ZSTD_d_forceIgnoreChecksum:
             bounds.lowerBound = (int)ZSTD_d_validateChecksum;
             bounds.upperBound = (int)ZSTD_d_ignoreChecksum;
+            return bounds;
         default:;
     }
     bounds.error = ERROR(parameter_unsupported);
@@ -1465,6 +1466,7 @@ size_t ZSTD_DCtx_setParameter(ZSTD_DCtx* dctx, ZSTD_dParameter dParam, int value
         case ZSTD_d_forceIgnoreChecksum:
             CHECK_DBOUNDS(ZSTD_d_forceIgnoreChecksum, value);
             dctx->forceIgnoreChecksum = (ZSTD_forceIgnoreChecksum_e)value;
+            return 0;
         default:;
     }
     RETURN_ERROR(parameter_unsupported, "");
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -115,6 +115,7 @@ static void ZSTD_initDCtx_internal(ZSTD_DCtx* dctx)
     dctx->bmi2 = ZSTD_cpuid_bmi2(ZSTD_cpuid());
     dctx->outBufferMode = ZSTD_obm_buffered;
     dctx->forceIgnoreChecksum = ZSTD_d_validateChecksum;
+    dctx->validateChecksum = 1;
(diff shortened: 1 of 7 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -1397,6 +1403,11 @@ size_t ZSTD_DCtx_setFormat(ZSTD_DCtx* dctx, ZSTD_format_e format)
     return ZSTD_DCtx_setParameter(dctx, ZSTD_d_format, format);
 }
 
+size_t ZSTD_DCtx_setForceIgnoreChecksum(ZSTD_DCtx* dctx, ZSTD_forceIgnoreChecksum_e value)
```
felixhandte, line 1406: Do we need this?

I guess in my mind users would just use `ZSTD_DCtx_setParameter(..., ZSTD_d_forceIgnoreChecksum, ...)` directly.

lib/zstd.h
```diff
@@ -1160,6 +1162,11 @@ typedef enum {
                                  * Decoder cannot recognise automatically this format, requiring this instruction. */
 } ZSTD_format_e;
 
+typedef enum {
```
felixhandte, line 1165: Maybe just add a quick comment here that this is used with the `ZSTD_d_forceIgnoreChecksum` parameter?

programs/zstdcli.c
```diff
@@ -196,6 +196,7 @@ static void usage_advanced(const char* programName)
     DISPLAYOUT( " -l     : print information about zstd compressed files \n");
     DISPLAYOUT( "--test  : test compressed file integrity \n");
     DISPLAYOUT( " -M#    : Set a memory usage limit for decompression \n");
+    DISPLAYOUT( "--no-check : disable validation of checksums in compressed frame \n");
```
felixhandte, line 199: If both compression and decompression are compiled into the CLI, this will produce two help lines for `--no-check`, won't it?

We should consolidate them into one description line, which should vary based on whether both, neither, or one of compression and decompression are compiled in.

You might also, like the compression-side `--[no-]check`, want to describe and handle `--check`. (`zstd --no-check ... --check ...` should check, `zstd --check ... --no-check ...` shouldn't check. Is that already the case?)

tests/fuzzer.c
```diff
@@ -544,6 +544,41 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        ZSTD_CCtx* const cctx = ZSTD_createCCtx();
+        CHECK_Z( ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 1) );
+        CHECK_VAR(cSize, ZSTD_compress2(cctx,
+                            compressedBuffer, compressedBufferSize,
+                            CNBuffer, CNBuffSize) );
+        ZSTD_freeCCtx(cctx);
+    }
+    {   /* copy the compressed buffer and corrupt the checksum */
+        size_t r;
+        char* const corruptedChecksumCompressedBuffer = (char*)malloc(cSize);
```
felixhandte, line 558: Nit: this allocation and copy shouldn't be necessary, right? You can corrupt the buffer in-place, and then check decompression succeeds/fails as expected.

tests/fuzzer.c
```diff
@@ -544,6 +544,41 @@ static int basicUnitTests(U32 const seed, double compressibility)
       if (ZSTD_getErrorCode(r) != ZSTD_error_dstSize_tooSmall) goto _output_error; }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : decompress with corrupted checksum : ", testNb++);
+    {   /* create compressed buffer with checksumming enabled */
+        ZSTD_CCtx* const cctx = ZSTD_createCCtx();
```
felixhandte, line 549: Nit: `CHECK()` this allocation succeeded?

tests/playTests.sh
```diff
@@ -260,6 +260,11 @@ zstd tmp -c --compress-literals    --fast=1 | zstd -t
 zstd tmp -c --compress-literals    -19      | zstd -t
 zstd -b --fast=1 -i0e1 tmp --compress-literals
 zstd -b --fast=1 -i0e1 tmp --no-compress-literals
+println "test: --no-check for decompression"
```
felixhandte, line 263: Do you want to try corrupting the input here? Something like [this](https://stackoverflow.com/a/5586379/1217213):

```
printf '\xDE\xAD\xBE\xEF' | dd of=tmp.zst bs=1 seek=$(($(wc -c tmp.zst) - 4)) count=4 conv=notrunc
```

Not sure if we're ok taking deps on `printf` and `wc`. I know we already depend on `dd`. @Cyan4973?

Or maybe:

```
head -c -4 < tmp.zst > tmp.corrupt.zst
echo -ne "\xDE\xAD\xBE\xEF" >> tmp.corrupt.zst
```

lib/zstd.h
```diff
@@ -1690,13 +1697,28 @@ ZSTDLIB_API size_t ZSTD_DCtx_setMaxWindowSize(ZSTD_DCtx* dctx, size_t maxWindowS
  */
 #define ZSTD_d_stableOutBuffer ZSTD_d_experimentalParam2
 
+/* ZSTD_d_forceIgnoreChecksum
+ * Experimental parameter.
+ * Default is 0 == disabled. Set to 1 to enable
```
felixhandte, line 1702: Mention that the values are of type `ZSTD_forceIgnoreChecksum_e`.

tests/playTests.sh
Cyan4973: The difficulty is remain compatible with `sh` in `playTests.sh`, 
which is more limited than `bash`.
But I believe that all formulations proposed are compatible.

The `freeBSD` CI test is supposed to help checking `sh` compatibility.

lib/decompress/zstd_decompress.c
```diff
@@ -676,11 +677,13 @@ static size_t ZSTD_decompressFrame(ZSTD_DCtx* dctx,
...
-        RETURN_ERROR_IF(checkRead != checkCalc, checksum_wrong, "");
+        if (!dctx->forceIgnoreChecksum) {
+            U32 const checkCalc = (U32)XXH64_digest(&dctx->xxhState);
+            U32 checkRead;
+            RETURN_ERROR_IF(remainingSrcSize<4, checksum_wrong, "");
+            checkRead = MEM_readLE32(ip);
+            RETURN_ERROR_IF(checkRead != checkCalc, checksum_wrong, "");
+        }
         ip += 4;
         remainingSrcSize -= 4;
```
Cyan4973, line 688: Agreed with @felixhandte : the check `remainingSrcSize<4` remains valid and important to ensure that `remainingSrcSize` is not underflowed at this operation.

lib/decompress/zstd_decompress.c
Cyan4973: I presume this method may have been created due to the existence of `ZSTD_DCtx_setFormat()`.

I think the missing context here is that this capability existed before `ZSTD_DCtx_setParameter()`, so that's why it already had its own method. Furthermore, when `ZSTD_DCtx_setParameter()` started its existence, it only had _one_ parameter to look after, making it both more complex and less expressive than `ZSTD_DCtx_setFormat()`, so it was not completely obvious which one was the "redundant" one.

It's only now that `ZSTD_DCtx_setParameter()` can access multiple parameters that it starts to make more sense, and that dedicated access methods such as `ZSTD_DCtx_setFormat()` start to be seen as redundant.

senhuang42: Do we want to just get rid of both then? The params they control are still technically in the "experimental" section.

Cyan4973: Not in this PR, which is focused on `forceIgnoreChecksum`. 
Removing a symbol, even from the experimental section, is a "dangerous" operation, as we don't always know how many other code depend on it. We may nonetheless do it, if it seems that it's beneficial for API clarity, but likely in combination with a larger version change, like `v1.5.0`.

tests/fuzzer.c
senhuang42: ah yes... `compressedBuffer` is a `void* const`, not `const void*`  😧

lib/decompress/zstd_decompress.c
```diff
@@ -446,7 +447,7 @@ static size_t ZSTD_decodeFrameHeader(ZSTD_DCtx* dctx, const void* src, size_t he
     RETURN_ERROR_IF(dctx->fParams.dictID && (dctx->dictID != dctx->fParams.dictID),
                     dictionary_wrong, "");
 #endif
-    if (dctx->fParams.checksumFlag) XXH64_reset(&dctx->xxhState, 0);
+    if (dctx->fParams.checksumFlag && !dctx->forceIgnoreChecksum) XXH64_reset(&dctx->xxhState, 0);
```
Cyan4973, line 450: Food for though :
I wonder if it would be better to combine `(dctx->fParams.checksumFlag && !dctx->forceIgnoreChecksum)` into a 3rd parameter, named something like `dctx->validateChecksum`.

There is a small optimization here (one variable access vs 2), but I don't this it matters much.
More importantly, I'm wondering if it would be helpful both for code readability and (later) code modifications, as it seems relatively easy to miss one of those conditions.

The critical point here would be to ensure that `dctx->validateChecksum` is always necessarily initialized correctly, which I believe is probably doable.

senhuang42: Just for clarity - do you mean:
1) adding the field `validateChecksum` in the `dctx`, that we can initialize based on `fParams.checksumFlag` and `forceIgnoreChecksum`
or 2) changing the `dctx`  param `forceIgnoreChecksum` to `validateChecksum` and figuring out a way to initialize it properly each time, and typically only calling `validateChecksum` in our logic. But, I'm not totally sure how possible this is yet since we do need to know whether compression specified checksumming in some cases.

My only concern about option 1) is that having three flags to deal with checksumming might seem excessive? But I defer that judgement to you.

Cyan4973: Faced with these 2 choices, I'm more in favor of option 1, for the following reasons : 

- These variables have different nature and lifetime
   + `forceIgnoreChecksum` is a user-selectable parameter. It must be registered "as is", especially as parameters are "sticky" by default, meaning that it should still mean exactly the same for next frame. If it's tampered by some processing involving other local parameters, this property is lost.
   + `fParams.checksumFlag` is a property of the frame currently decoded. It only depends on input, and is not user-selectable. It's useful in itself for other purposes, and only valid for current frame. This flag will have to be determined again on next frame.
   + `validateChecksum` is combination of the previous 2. It's a processing request : it tells that a checksum exist in the frame and must be validated by the decoding process. In contrast, `forceIgnoreChecksum==0` (default) means that _if_ a checksum exist, it should be validated. This request is only valid for the current frame, and will be determined again on next frame.

There is a 3rd possibility, but I'm not sure if it's applicable : 
- If the combination `(dctx->fParams.checksumFlag && !dctx->forceIgnoreChecksum)` is only useful within one function (`ZSTD_decompressContinue()` typically), then `validateChecksum` could be processed as a local stack variable, instead of being an additional state parameter. `validateChecksum` would be recalculated several times per frame, instead of once, but I'm pretty sure it doesn't matter. More importantly, it would ensure that there is no inconsistency between `validateChecksum` and its constituents (`forceIgnoreChecksum` and `fParams.checksumFlag`).

Cyan4973:
Note : the `arm` test failures on `github actions` are unrelated

PR #2279, opened 2020-08-24
Add a warning whenever (de)compressing multiple files

senhuang42:
The following would trigger the warning:

`zstd tst/tmp1 tst/tmp2 tst/tmp3 -o out.zst`
`zstd tst/tmp1 tst/tmp2 tst/tmp3 -o out.zst --rm` (this one aborts fully)
`zstd tst/tmp1 tst/tmp2 tst/tmp3 -o out.zst --rm -f`
`zstd tst/tmp1 tst/tmp2 tst/tmp3 -o out.zst -f`
`zstd tst/tmp1 testfiles/tmp2 testfiles/tmp3 -c`
`zstd -r tst/ -o out.zst`
`zstd -r tst/ -c`
`zstd -r tst/ --rm -c` (this one aborts fully)
./zstd -r tst/ --rm -f`

For overwriting the file with `-f`, there's no warning triggered. Is that behavior that we want here as well? 

Some tests in `playTests.sh` had to be modified to work with the new warning.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1672,14 +1672,23 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
     assert(outFileName != NULL || suffix != NULL);
     if (outFileName != NULL) {   /* output into a single destination (stdout typically) */
         if (nbFiles > 1 && !prefs->overwrite) {
-            if (!strcmp (outFileName, stdoutmark)) {
-                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into stdout. ");
+            /* g_display_prefs.displayLevel <= 1 corresponds to -q flag */
+            DISPLAY("%d\n", g_display_prefs.displayLevel);
+            if (g_display_prefs.displayLevel <= 1) {
+                if (prefs->removeSrcFile) {
+                    DISPLAY("zstd: Aborting... not deleting files and processing into dst: %s", outFileName);
+                    return 1;
+                }
             } else {
-                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into a single output file: %s ", outFileName);
+                if (!strcmp (outFileName, stdoutmark)) {
+                    DISPLAY("zstd: WARNING: all input files will be processed and concatenated into stdout. ");
+                } else {
+                    DISPLAY("zstd: WARNING: all input files will be processed and concatenated into a single output file: %s ", outFileName);
+                }
+                if (prefs->removeSrcFile)
+                    error = g_display_prefs.displayLevel > 1 && UTIL_requireUserConfirmationToProceed("Proceed? (y/n): ", "Aborting...", "yY");
+                DISPLAY("\n");
             }
-            if (prefs->removeSrcFile)
-                error = UTIL_requireUserConfirmationToProceed("Proceed? (y/n): ", "Aborting...", "yY");
-            DISPLAY("\n");
         }
         ress.dstFile = FIO_openDstFile(prefs, NULL, outFileName);
         if (ress.dstFile == NULL) {  /* could not open outFileName */
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1722,8 +1722,10 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
     /* init */
     assert(outFileName != NULL || suffix != NULL);
     if (outFileName != NULL) {   /* output into a single destination (stdout typically) */
-        if (FIO_removeMultiFilesWarning(prefs, 1 /* displayLevelCutoff */, outFileName))
+        if (FIO_removeMultiFilesWarning(prefs, 1 /* displayLevelCutoff */, outFileName)) {
+            FIO_freeCResources(ress);
             return 1;
+        }
         ress.dstFile = FIO_openDstFile(prefs, NULL, outFileName);
         if (ress.dstFile == NULL) {  /* could not open outFileName */
             error = 1;
@@ -2618,8 +2620,10 @@ FIO_decompressMultipleFilenames(FIO_prefs_t* const prefs,
     dRess_t ress = FIO_createDResources(prefs, dictFileName);
 
     if (outFileName) {
-        if (FIO_removeMultiFilesWarning(prefs, 1 /* displayLevelCutoff */, outFileName))
+        if (FIO_removeMultiFilesWarning(prefs, 1 /* displayLevelCutoff */, outFileName)) {
+            FIO_freeDResources(ress);
             return 1;
+        }
         if (!prefs->testMode) {
             ress.dstFile = FIO_openDstFile(prefs, NULL, outFileName);
             if (ress.dstFile == 0) EXM_THROW(19, "cannot open %s", outFileName);
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1677,6 +1677,29 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
     /* init */
     assert(outFileName != NULL || suffix != NULL);
     if (outFileName != NULL) {   /* output into a single destination (stdout typically) */
+        if (nbFiles > 1) {
+            if (!strcmp (outFileName, stdoutmark)) {
+                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into stdout. ");
+            } else {
+                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into a single output file: %s ", outFileName);
+            }
+            if (prefs->removeSrcFile && !prefs->overwrite) {
+                DISPLAY("\nYou must specify -f as well in order to execute this command with --rm. Aborting...");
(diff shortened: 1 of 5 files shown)


programs/fileio.c
```diff
@@ -1677,6 +1677,29 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
...
+                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into stdout. ");
+            } else {
+                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into a single output file: %s ", outFileName);
+            }
+            if (prefs->removeSrcFile && !prefs->overwrite) {
+                DISPLAY("\nYou must specify -f as well in order to execute this command with --rm. Aborting...");
+                return 1;
+            }
+            
+            DISPLAY("Proceed? (y/n): ");
```
terrelln, line 1691: I think the behavior we want is:
* If `-f` is passed then proceed without waiting for confirmation. This should probably also silence the warnings.
* If `-f` is not passed and `--rm` is not passed then print a warning and proceed. This is a non-destructive operation, but maybe not what they want.
* If `-f` is not passed and `--rm` is passed then print a warning. At this point we could either abort, or ask for confirmation. Asking for confirmation is good for me.

programs/fileio.c
```diff
@@ -1677,6 +1677,29 @@ int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
...
+        if (nbFiles > 1) {
+            if (!strcmp (outFileName, stdoutmark)) {
+                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into stdout. ");
+            } else {
+                DISPLAY("zstd: WARNING: all input files will be processed and concatenated into a single output file: %s ", outFileName);
+            }
+            if (prefs->removeSrcFile && !prefs->overwrite) {
+                DISPLAY("\nYou must specify -f as well in order to execute this command with --rm. Aborting...");
+                return 1;
+            }
```
terrelln, line 1689: Why not have `\n` at the end of the displayed line?

tests/playTests.sh
```diff
@@ -349,19 +349,25 @@ rm tmp*
 println "\n===>  compress multiple files"
 println hello > tmp1
 println world > tmp2
-zstd tmp1 tmp2 -o "$INTOVOID" -f
-zstd tmp1 tmp2 -c | zstd -t
-zstd tmp1 tmp2 -o tmp.zst
+echo 'y' | zstd tmp1 tmp2 -o "$INTOVOID" -f  # echo 'y' confirms the warning prompt
```
terrelln, line 352: nit: Also test `Y` and `yes` to ensure they work.

programs/fileio.c
Cyan4973: minor : 
there is already a `(y/n)` prompt at line 602,
it might be ground to unify both implementations into a single function.

senhuang42: yep i was just working on this!

felixhandte: Prompting the user (i.e., blocking the process for a response) makes me a little nervous. At the very least, we should gate such a prompt behind `isatty(STDIN_FILENO)`.

Cyan4973: We already have this situation regarding confirmation to overwrite existing files.

Things we currently do to make such a scenario compatible with scripts (which can't answer prompt) : 
`-f` : is an automatic "yes"
`-q` : is an automatic "no"
`-f -q` : `-f` prevails

senhuang42: Oh interesting - wasn't aware of that flag. I'll add that in.

felixhandte: Right. But it would be very easy, without those flags, to get into this situation in a non-interactive environment, and then `zstd` would just hang.

Cyan4973: I believe that's what already happens now when doing `zstd FILE` and `FILE.zst` already exist.
`zstd FILE` asks for confirmation prompt.
But `zstd FILE < /dev/null`, which is similar to non-interactive, is an automatic "no".
I presume it's a natural consequence of `getchar()`

I tested the behavior using `gzip` and it's exactly the same.

felixhandte: Hmmmm, then I think this is actually an unlikely, but potentially bad, bug:

```
$ cat <<EOF >test
> y
> Hello, world!
> EOF
$ ./zstd test
test                 :181.25%   (    16 =>     29 bytes, test.zst)             
$ ./zstd -q -d test.zst -c
y
Hello, world!
$ ./zstd < test -o test.zst
/*stdin*\            :192.86%   (    14 =>     27 bytes, test.zst)             
$ ./zstd -q -d test.zst -c
Hello, world!
$ 
```

I.e., the prompt can end up swallowing part of the input when it's being presented via `stdin`.

Cyan4973: Indeed, there should not be any usage of `getchar()` when any input is `stdin`,
and therefore it should an automatic "no".

That's a bug to fix.

Unsure if it's "fair" to make it part of _this_ PR, as it increases its scope.

felixhandte: Yeah, makes sense. I'll put up a PR after this gets merged, since this is refactoring that code already.

Cyan4973: Good point !

I think the initial intention is that a level `2` enables live line status, 
such as `Read : 310 / 953 MB ==> 35%`, 
while level `1` quiets that, and only prints warnings.

It was deemed that line status was designed for a single file or stream, 
and as a consequence was too annoying for multiple files, 
due to a combination of confusion and space occupation.

On short term, the "easy" way out is that
processing multiple files is by default equivalent to `-q` , 
meaning it effectively disables prompt selection.

Note that prompt selection still remains a thing with an explicit `-v`.

Longer term, we may have to consider if we want a concise line status dedicated to multiple files,
so that level `2` can still exist in combination with multiple files.

senhuang42:
This should work properly now that #2283 is merged, and added function `FIO_removeMultiFilesWarning()` to further simplify the code

programs/fileio.c
```diff
@@ -796,6 +789,35 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
...
 }
 
+/* FIO_removeMultiFilesWarning() :
+ * Returns 1 if the console should abort, 0 if console should proceed.
+ * Based on displayLevelCutoff (typically == 1) and the global setting g_display_prefs.displayLevel, this
+ * function may print a warning message, a user prompt, both, or none.
+ */
+static int FIO_removeMultiFilesWarning(const FIO_prefs_t* const prefs, int displayLevelCutoff, const char* outFileName)
+{
+    int error;
```
Cyan4973, line 799: nit: this could be combined `int error = 0;`

Cyan4973:
The code itself looks good to me !

I believe this new protection feature would benefit from some small documentation updates :  
- `FIO_removeMultiFilesWarning()` : better describe the supported scenarios and expected behavior, directly in code documentation. For example, it took me a while to understand that `FIO_removeMultiFilesWarning()` also supports being called when `--rm` is _not_ triggered. It also correctly adapts the warning / user prompt logic in this case, but one needs to read the code to follow what the "expected behavior" is.
- Maybe document the new behavior in the `man` page, which is built from `programs/zstd.1.md`. While it doesn't need to be very detailed, a few words around `--rm` would be welcomed. Also indicate that `-f` can be use to remove the warning message, in case some user find it annoying.

programs/fileio.c
```diff
@@ -796,6 +789,35 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
...
+    error = 0;
+    if (prefs->nbFiles > 1 && !prefs->overwrite) {
+        if (g_display_prefs.displayLevel <= displayLevelCutoff) {
+            if (prefs->removeSrcFile) {
+                DISPLAYLEVEL(1, "zstd: Aborting... not deleting files and processing into dst: %s", outFileName);
+                error =  1;
+            }
+        } else {
+            if (!strcmp(outFileName, stdoutmark)) {
+                DISPLAYLEVEL(2, "zstd: WARNING: all input files will be processed and concatenated into stdout. ");
```
Cyan4973, line 809: The message is informational and correct : it tells what the action is doing.

But it doesn't tell what are the potential consequences, aka the concatenated output _cannot regenerate the original directory tree_.

This matters particularly in combination with `--rm`, because `zstd -r --rm * -o archive.zst` will simply destroy all information below current directory, with no way to regenerate that from `archive.zst`.
While there might be some valid use cases to do so,
there might also be some cases where some user just expects `archive.zst` to behave about the same as `archive.tar.zst`, the current warning message is not explicit enough to explain that this is a one-way transform with no way back possible.

Probably worth respelling.

terrelln: The like you added `if (prefs->removeSrcFile)` looks good to me. But I think we should always print that line, regardless of `--rm`. Since the user may think they have a full archive and manually delete the directory, for example

terrelln:
Just one minor comment and it looks good to me!

PR #2280, opened 2020-08-25
Add a Couple Use Cases to the Zstd Web-Page

felixhandte:
If there are others that I've forgotten, let me know and I'll add 'em in.


Cyan4973:
Thanks @felixhandte !

PR #2281, opened 2020-08-25
Fix building on AIX 5.1

likema:
When building `zstd` by **GCC 4.5.4** on **AIX 5.1** , GCC could not found `stdint.h`

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -45,7 +45,11 @@ extern "C" {
 *  Basic Types
 *****************************************************************/
 #if  !defined (__VMS) && (defined (__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */) )
-# include <stdint.h>
+#  if defined(_AIX)
+#    include <inttypes.h>
+#  else
+#    include <stdint.h> /* intptr_t */
+#  endif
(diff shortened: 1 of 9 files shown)


Cyan4973:
While I'm fine with the proposed change,
I remember that there is also an ongoing patch
which changes the way `zstd` includes `C` standard libraries.
Therefore, I would expect this PR to conflict.
cc @terrelln

Cyan4973:
With `mem.h` now updated,
could you please rebase your PR onto latest `dev` branch ?
Once conflict is cleared, I believe it's good to go.

likema:
I have rebased it onto `dev`. 

But I find that

- contrib/pzstd/test/RoundTrip.h
- contrib/pzstd/utils/FileSystem.h
- contrib/pzstd/Options.h
- contrib/pzstd/SkippableFrame.h
- contrib/pzstd/Pzstd.h
- contrib/single_file_libs/examples/roundtrip.c
- contrib/single_file_libs/examples/simple.c
- doc/educational_decoder/zstd_decompress.c

includes `cstdint`

Should I fix them?

Cyan4973:
Nope, these are different programs, just hosted in the same repository.
The current scope of this PR looks good to me.

Cyan4973 (approved):

PR #2282, opened 2020-08-25
[bug] Fix FSE_readNCount()

terrelln:
* Fix bug introduced in PR #2271
* Fix long-standing bug that is impossible to trigger inside of zstd
* Add a fuzzer that makes sure the normalized count always round trips
  correctly

I measured this change to be performance neutral, modulo small changes due to alignment.

Credit to OSS-Fuzz

### DIFF
--- a/lib/common/entropy_common.c
+++ b/lib/common/entropy_common.c
@@ -110,13 +110,14 @@ size_t FSE_readNCount_body(short* normalizedCounter, unsigned* maxSVPtr, unsigne
             int repeats = FSE_ctz(~bitStream | 0x80000000) >> 1;
             while (repeats >= 12) {
                 charnum += 3 * 12;
-                if (ip <= iend-7) {
+                if (LIKELY(ip <= iend-7)) {
                     ip += 3;
-                    bitStream = MEM_readLE32(ip) >> bitCount;
                 } else {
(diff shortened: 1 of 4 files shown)


lib/common/entropy_common.c
```diff
@@ -137,14 +138,15 @@ size_t FSE_readNCount_body(short* normalizedCounter, unsigned* maxSVPtr, unsigne
...
                 assert((bitCount >> 3) <= 3); /* For first condition to work */
                 ip += bitCount>>3;
                 bitCount &= 7;
-                bitStream = MEM_readLE32(ip) >> bitCount;
             } else {
-                bitStream >>= 2;
+                bitCount -= (int)(8 * (iend - 4 - ip));
+                ip = iend - 4;
             }
+            bitStream = MEM_readLE32(ip) >> bitCount;
```
Cyan4973, line 149: Some remaining corner case :
`../../lib/common/entropy_common.c:149:42: runtime error: shift exponent 40 is too large for 32-bit type 'U32' (aka 'unsigned int')`

terrelln: Fixed, and also removed the `& 31` from the common case.

Cyan4973:
Great work @terrelln !

PR #2283, opened 2020-08-25
Refreshing progress bar for processing multiple files

senhuang42:
video of how it works: https://vimeo.com/451651373

Summary:
Now, instead of line by line output, we have a single, live updating progress feed when compressing/decompressing multiple files.

```
zstd tmp1 tmp2 tmp3 tmp4.       # uses refreshing status line, displayLevel = 2
zstd tmp1 tmp2 tmp3 tmp4.  -v   # same as before, prints status updates on each line and more info, displayLevel = 3
zstd tmp1 tmp2 tmp3 tmp4   -q.  # same as before, emits nothing, displayLevel = 1
```

We add two new parameters in `FIO_prefs_t`: `int nbFiles` and `int currFileIdx`. Also, the function signature of `FIO_(de)compressMultipleFilenames()` no longer has a `nbFiles` field, since we already have one in `FIO_prefs_t`.
Now, we only force `displayLevel=1` in pipe mode which is actually desirable, as opposed to the previous situation where this was forced during multiple files mode too.

One small side effect of this correction of multi-file `displayLevel`: if we input `zstd file1 file2 -o conflictingOutfile.zst`, previously zstd would simply bail (due to `displayLevel=1`), but now it will trigger user confirmation prompt (due to `displayLevel=2`). We also trigger the prompt if we invoke `zstd file1 file2` in the presence of `file1.zst file2.zst`. This the exact same behavior as single file operations: `zstd file1 -o conflictingOutfile.zst`.

This PR also corrects a bug in the user prompt that caused it to skip over every other file if you answered "no" each time when prompting overwrite for multiple files.

Followup:
- Add a summary of total compression ratio of all files `totalBytes/compressedBytes` in a followup PR.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -624,9 +624,10 @@ FIO_openDstFile(FIO_prefs_t* const prefs,
                 {   int ch = getchar();
                     if ((ch!='Y') && (ch!='y')) {
                         DISPLAY("    not overwritten  \n");
+                        /* flush rest of input line */
+                        while ((ch!=EOF) && (ch!='\n')) ch = getchar();
                         return NULL;
                     }
-                    /* flush rest of input line */
                     while ((ch!=EOF) && (ch!='\n')) ch = getchar();
             }   }
             /* need to unlink */
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -319,6 +319,7 @@ struct FIO_prefs_s {
     int excludeCompressedFiles;
     int patchFromMode;
     int contentSize;
+    int currFileIdx;
     int nbFiles;
 };
 
@@ -362,6 +363,7 @@ FIO_prefs_t* FIO_createPreferences(void)
     ret->literalCompressionMode = ZSTD_lcm_auto;
     ret->excludeCompressedFiles = 0;
     ret->nbFiles = 1;
+    ret->currFileIdx = 0;
     return ret;
 }
 
@@ -500,6 +502,11 @@ void FIO_setContentSize(FIO_prefs_t* const prefs, int value)
 void FIO_setNbFiles(FIO_prefs_t* const prefs, int value)
 {
     prefs->nbFiles = value;
+}
+
+void FIO_setCurrFileIdx(FIO_prefs_t* const prefs, int value)
+{
+    prefs->currFileIdx = value;
 }
 
 /*-*************************************
@@ -1268,10 +1275,15 @@ FIO_compressZstdFrame(FIO_prefs_t* const prefs,
                                 (unsigned)(zfp.produced >> 20),
                                 cShare );
                 } else {   /* summarized notifications if == 2; */
-                    DISPLAYLEVEL(2, "\033[s Read : %u ", (unsigned)(zfp.consumed >> 20));
+                    if (prefs->nbFiles > 1) {
+                        DISPLAYLEVEL(2, "\rCompressing %u/%u files. Current source: %s ", prefs->currFileIdx+1, prefs->nbFiles, srcFileName);
+                    } else {
+                        DISPLAYLEVEL(2, "\r");
+                    }
+                    DISPLAYLEVEL(2, "Read : %u ", (unsigned)(zfp.consumed >> 20));
                     if (fileSize != UTIL_FILESIZE_UNKNOWN)
                         DISPLAYLEVEL(2, "/ %u ", (unsigned)(fileSize >> 20));
-                    DISPLAYLEVEL(2, "MB ==> %2.f%%\033[u", cShare);
+                    DISPLAYLEVEL(2, "MB ==> %2.f%%", cShare);
                     DELAY_NEXT_UPDATE();
                 }
 
@@ -1436,8 +1448,7 @@ FIO_compressFilename_internal(FIO_prefs_t* const prefs,
     /* Status */
     
     DISPLAYLEVEL(2, "\r%79s\r", "");
-    /* No status message in pipe mode (stdin - stdout) or multi-files mode */
-    if (g_display_prefs.displayLevel > 2 || (prefs->nbFiles == 1 && !((!strcmp(srcFileName, stdinmark) && dstFileName && !strcmp(dstFileName,stdoutmark))))) {
+    if (g_display_prefs.displayLevel > 2 || (g_display_prefs.displayLevel == 2 && prefs->nbFiles <= 1)) {
         if (readsize == 0) {
             DISPLAYLEVEL(2,"%-20s :  (%6llu => %6llu bytes, %s) \n",
                 srcFileName,
@@ -1673,7 +1684,7 @@ static unsigned long long FIO_getLargestFileSize(const char** inFileNames, unsig
  * or into a destination folder (specified with -O)
  */
 int FIO_compressMultipleFilenames(FIO_prefs_t* const prefs,
-                                  const char** inFileNamesTable, unsigned nbFiles,
+                                  const char** inFileNamesTable,
                                   const char* outMirroredRootDirName,
                                   const char* outDirName,
                                   const char* outFileName, const char* suffix,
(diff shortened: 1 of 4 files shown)


Cyan4973:
Regarding the design of the solution itself, I believe it's pretty good.

The one difference with one file is that there is no "summary" at the end.
It might be useful / interesting to produce one,
though it involved tracking all I/O amounts read and written,
and I'm unsure if the current code base make this easily possible.
If not, better leave that open for a future PR.

programs/zstdcli.c
```diff
@@ -1244,10 +1244,6 @@ int main(int const argCount, const char* argv[])
         CLEAN_RETURN(1);
     }
 
-    /* No status message in pipe mode (stdin - stdout) or multi-files mode */
-    if (!strcmp(filenames->fileNames[0], stdinmark) && outFileName && !strcmp(outFileName,stdoutmark) && (g_displayLevel==2)) g_displayLevel=1;
```
Cyan4973, line 1248: I really believe that reducing verbosity to quiet `-q` mode when `zstd` is used in a pipe configuration
remain the correct thing to do.

The goal of this patch is to add a status line when there are multiple input files.
So only the second line (below) should be removed.

programs/zstdcli.c
```diff
@@ -1306,10 +1302,12 @@ int main(int const argCount, const char* argv[])
...
         }
 
-        if ((filenames->tableSize==1) && outFileName)
-          operationResult = FIO_compressFilename(prefs, outFileName, filenames->fileNames[0], dictFileName, cLevel, compressionParams);
-        else
-          operationResult = FIO_compressMultipleFilenames(prefs, filenames->fileNames, (unsigned)filenames->tableSize, outMirroredDirName, outDirName, outFileName, suffix, dictFileName, cLevel, compressionParams);
+        if ((filenames->tableSize==1) && outFileName) {
+            operationResult = FIO_compressFilename(prefs, outFileName, filenames->fileNames[0], dictFileName, cLevel, compressionParams);
+        } else {
+            FIO_setNbFiles(prefs, (int)filenames->tableSize);
```
Cyan4973, line 1308: since `filenames->tableSize` is also provided as an argument of next function, `FIO_compressMultipleFilenames()`, would it be possible to skip the `FIO_setNbFiles()` stage altogether ?

programs/fileio.c
```diff
@@ -2316,7 +2332,9 @@ static int FIO_decompressFrames(dRess_t ress, FILE* srcFile,
 
     /* Final Status */
     DISPLAYLEVEL(2, "\r%79s\r", "");
-    DISPLAYLEVEL(2, "%-20s: %llu bytes \n", srcFileName, filesize);
+    /* No status message in pipe mode (stdin - stdout) or multi-files mode */
+    if (g_display_prefs.displayLevel > 2 || (prefs->nbFiles == 1 && !((!strcmp(srcFileName, stdinmark) && dstFileName && !strcmp(dstFileName,stdoutmark)))))
```
Cyan4973, line 2336: See comment regarding `zstd` in pipe mode below.
This should simplify this line.

programs/fileio.c
```diff
@@ -1427,18 +1434,22 @@ FIO_compressFilename_internal(FIO_prefs_t* const prefs,
...
-            (unsigned long long)readsize, (unsigned long long) compressedfilesize,
-            dstFileName);
-    } else {
-        DISPLAYLEVEL(2,"%-20s :%6.2f%%   (%6llu => %6llu bytes, %s) \n",
-            srcFileName,
-            (double)compressedfilesize / readsize * 100,
-            (unsigned long long)readsize, (unsigned long long) compressedfilesize,
-            dstFileName);
+    /* No status message in pipe mode (stdin - stdout) or multi-files mode */
+    if (g_display_prefs.displayLevel > 2 || (prefs->nbFiles == 1 && !((!strcmp(srcFileName, stdinmark) && dstFileName && !strcmp(dstFileName,stdoutmark))))) {
```
Cyan4973, line 1440: See comment (below) on `zstd` in pipe mode.
This should simplify this line.

programs/fileio.c
```diff
@@ -1261,10 +1268,10 @@ FIO_compressZstdFrame(FIO_prefs_t* const prefs,
                                 (unsigned)(zfp.produced >> 20),
                                 cShare );
                 } else {   /* summarized notifications if == 2; */
-                    DISPLAYLEVEL(2, "\rRead : %u ", (unsigned)(zfp.consumed >> 20));
+                    DISPLAYLEVEL(2, "\033[s Read : %u ", (unsigned)(zfp.consumed >> 20));
```
Cyan4973, line 1271: So far, we have avoided escape codes, due to portability concerns (unsure how this works on _all_ consoles).

I presume that if we want to avoid that risk,
existing design based on `\r` would require to rewrite the entire line, instead of just the count section.

Do you think this would be feasible ?

senhuang42: Within `FIO_compressZstdFrame()` we don't have any idea which # file we're currently compressing, which we'd need to regenerate the `compressing file currentFile/totalFiles` part of the output line. We do have access to `FIO_prefs` which maybe we could use to store more information, like a `currFileIdx`. I'm inclined to go with this, since we already have added a `nbFiles` field.

If we're willing to forgo the `currentFile/totalFiles`, we can definitely just have it output the typical: `[sourcefilename] : [some info about the progress` and have it continually overwrite that. 

If we could separate it into two lines, that might be another way, but I haven't come across anything that lets me replace already-printed console lines that looks portable.

programs/zstdcli.c
senhuang42: Yeah - however, in our discussion below it seems like we'll be moving any text output further down into the function:
```
FIO_compressFilename_srcFile(FIO_prefs_t* const prefs,
                             cRess_t ress,
                             const char* dstFileName,
                             const char* srcFileName,
                             int compressionLevel) 
```
which seems like if we want to pass a `nbFiles` and/or `currFileIdx` field, we'd have to modify either the function signature of `FIO_compressFilename_srcFile()` or add two fields to `FIO_prefs_t`. I think I prefer adding fields to `FIO_prefs_t` since I don't think `FIO_compressFilename_srcFile()` should need a `nbFiles` and/or `currFileIdx` argument, since it is an operation on a single file.

Cyan4973: As an alternative, and assuming you want to limit the amount of changes,
`FIO_setNbFiles()` could be invoked from _within_ `FIO_compressMultipleFilenames()`

senhuang42:
PR description has been updated

programs/fileio.c
```diff
@@ -319,6 +319,8 @@ struct FIO_prefs_s {
     int excludeCompressedFiles;
     int patchFromMode;
     int contentSize;
+    int currFileIdx;
+    int nbFiles;
 };
```
Cyan4973, line 324: out of scope for this PR, but nonetheless impacted by it : 

It seems like `struct FIO_prefs_s` is no longer a "preferences" structure,
and has become a kind of "state" at some point.

The difference is pretty substantial : 
"preferences" (which would probably be better named "parameters") 
implies a set of choices made externally by the calling process 
and which expression shall remain valid during the processing stage.
In essence, once set, it should remain `const` immutable.

By contrast, "state" contains internal (private) fields, often mutated during processing, 
in order to preserve synchronization between functions during a larger processing event,
such as compressing multiple files.

I presume it would be a good idea to separate both concepts more cleanly.
But that's obviously a larger change, for another PR.

senhuang42: Yeah, especially if we decide to move forward with the final "status" output that requires tracking compressed bytes throughout the operation - it functions more like a `ZSTD_CCtx`.

Perhaps a followup PR can include a new `FIO_Ctx_s` struct:
```
struct FIO_Ctx_s {
    int nbFiles;
    int currFileIdx;
    int bytesConsumed;
    int bytesEmitted;
};
```

senhuang42: I'd be happy to submit a PR for this a bit later.

Cyan4973: Indeed, good food for thoughts for a potential follow up.

Also, to keep in mind : distinction between public API, with parameters necessarily provided from outside,
and internal state information, which should remain private, hence unaccessible.
The less the external world has to know about, the better (both safer and easier to use).

We'll figure out what to do when it comes to this part.

PR #2288, opened 2020-08-27
[doc] Add ZSTD_versionString() to manual

animalize:
`ZSTD_versionString(void)` function is missing from the manual.
Below is the display effect in the Edge browser.

Before:
before

After:
after

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -73,15 +73,20 @@ extern "C" {
 #define ZSTD_VERSION_MAJOR    1
 #define ZSTD_VERSION_MINOR    4
 #define ZSTD_VERSION_RELEASE  5
-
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for the suggestion @animalize !
It's a good suggestion.

However, note that `zstd_manual.html` is a computer-generated file,
which means that, on next update, all manual changes will be erased.

A better way to make these changes permanent is to influence the source from which the manual is created.
This source is `lib/zstd.h`.
And it is transformed into an `html` with [`contrib/gen_html`](https://github.com/facebook/zstd/tree/v1.4.5/contrib/gen_html).
The update process can be invoked with `make manual`.

`gen_html` interprets the code declarations and code comments in a specific way,
explained in the `README.md` of the project.

The idea is to use these rules to modify `zstd.h`
so that its transformation using `gen_html` produces the intended `zstd_manual.html` content.

animalize:
Thank you very much for detailed explanation.

The PR and screenshot have been updated, the screenshot was generated by `gen_html`.

Cyan4973:
Excellent ! Thanks @animalize !

Issue #1097, opened 2018-04-01
How about sharing thread pools across contexts?


stati64:
Would this let multiple streams or blocks coexist with less dissonance?

Cyan4973:
I don't think it would bring any tangible benefit, neither speed nor memory wise,

but, on the other hand, it would markedly increase code complexity,
as it would introduce some "invisible shared object" between contexts.

terrelln:
Zstd isn't the right place to maintain a "global" thread pool. However, we could potentially use an external thread pool. Now that zstdmt allocates a single memory chunk for the input buffer, I think it is no longer dependent on the size of the thread pool for correctness (or close to independent).

Is this an issue you are running into in practice? If so can you provide some details?

marxin:
I see a similar need for `rpm`. Right now `rpm` creates packages in a parallel way (using openmp) where each of the `.rpm` files is then compressed. I added support for parallel compression, but it may lead to thread explosion as one can process N `.rpm` files each being compressed with N threads in `zstd`.
Maybe it would be handy to have a semaphore API where the compressing thread in `zstd` would need to acquire it?

PR #2290, opened 2020-08-27
[WIP] Add new ZSTD_customJobControl settings.

marxin:
As mentioned in #1097, I see it handy to have a mechanism that can limit threading among ZSTD contexts.
My motivation example comes from `rpm` where it does:
```
#omp parallel
foreach (package in packages)
   package->compress()
```

where each package can use threading compression of `zstd`. It is common that you can have a single file package (like Firefox) where you want to utilize threading compression. And you can also have very many sub-packages (like for gcc) where you can't spawn N threads for each of them.

I know the suggested code is not portable right now. It's a POC and I would like to see comments from community?

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -255,6 +255,7 @@ struct ZSTD_CCtx_s {
     unsigned long long producedCSize;
     XXH64_state_t xxhState;
     ZSTD_customMem customMem;
+    ZSTD_customJobControl customJobControl;
     size_t staticSize;
     SeqCollector seqCollector;
     int isFirstBlock;
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -343,6 +343,7 @@ typedef struct {
     int totalCCtx;
     int availCCtx;
     ZSTD_customMem cMem;
+    ZSTD_customJobControl cJobControl;
     ZSTD_CCtx* cctx[1];   /* variable size */
 } ZSTDMT_CCtxPool;
 
@@ -359,7 +360,8 @@ static void ZSTDMT_freeCCtxPool(ZSTDMT_CCtxPool* pool)
 /* ZSTDMT_createCCtxPool() :
  * implies nbWorkers >= 1 , checked by caller ZSTDMT_createCCtx() */
 static ZSTDMT_CCtxPool* ZSTDMT_createCCtxPool(int nbWorkers,
-                                              ZSTD_customMem cMem)
+                                              ZSTD_customMem cMem,
+                                              ZSTD_customJobControl cJobControl)
 {
     ZSTDMT_CCtxPool* const cctxPool = (ZSTDMT_CCtxPool*) ZSTD_customCalloc(
         sizeof(ZSTDMT_CCtxPool) + (nbWorkers-1)*sizeof(ZSTD_CCtx*), cMem);
@@ -370,9 +372,10 @@ static ZSTDMT_CCtxPool* ZSTDMT_createCCtxPool(int nbWorkers,
         return NULL;
     }
     cctxPool->cMem = cMem;
+    cctxPool->cJobControl = cJobControl;
     cctxPool->totalCCtx = nbWorkers;
     cctxPool->availCCtx = 1;   /* at least one cctx for single-thread mode */
-    cctxPool->cctx[0] = ZSTD_createCCtx_advanced(cMem);
+    cctxPool->cctx[0] = ZSTD_createCCtx_advanced(cMem, ZSTD_defaultJobControl);
     if (!cctxPool->cctx[0]) { ZSTDMT_freeCCtxPool(cctxPool); return NULL; }
     DEBUGLOG(3, "cctxPool created, with %u workers", nbWorkers);
     return cctxPool;
@@ -385,8 +388,9 @@ static ZSTDMT_CCtxPool* ZSTDMT_expandCCtxPool(ZSTDMT_CCtxPool* srcPool,
     if (nbWorkers <= srcPool->totalCCtx) return srcPool;   /* good enough */
     /* need a larger cctx pool */
     {   ZSTD_customMem const cMem = srcPool->cMem;
+        ZSTD_customJobControl const cJobControl = srcPool->cJobControl;
         ZSTDMT_freeCCtxPool(srcPool);
-        return ZSTDMT_createCCtxPool(nbWorkers, cMem);
+        return ZSTDMT_createCCtxPool(nbWorkers, cMem, cJobControl);
     }
 }
 
@@ -420,7 +424,7 @@ static ZSTD_CCtx* ZSTDMT_getCCtx(ZSTDMT_CCtxPool* cctxPool)
     }   }
     ZSTD_pthread_mutex_unlock(&cctxPool->poolMutex);
     DEBUGLOG(5, "create one more CCtx");
-    return ZSTD_createCCtx_advanced(cctxPool->cMem);   /* note : can be NULL, when creation fails ! */
+    return ZSTD_createCCtx_advanced(cctxPool->cMem, cctxPool->cJobControl);   /* note : can be NULL, when creation fails ! */
 }
 
 static void ZSTDMT_releaseCCtx(ZSTDMT_CCtxPool* pool, ZSTD_CCtx* cctx)
@@ -829,6 +833,7 @@ struct ZSTDMT_CCtx_s {
     unsigned long long consumed;
     unsigned long long produced;
     ZSTD_customMem cMem;
+    ZSTD_customJobControl cJobControl;
     ZSTD_CDict* cdictLocal;
     const ZSTD_CDict* cdict;
 };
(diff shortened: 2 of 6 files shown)


lib/compress/zstd_compress_internal.h
```diff
@@ -282,6 +282,7 @@ struct ZSTD_CCtx_s {
 #ifdef ZSTD_MULTITHREAD
     ZSTDMT_CCtx* mtctx;
 #endif
+    sem_t *semaphore;
```
Cyan4973, line 285: This should only be required when `ZSTD_MULTITHREAD` is defined.
Keep in mind that, in many systems, `libzstd` must be compiled within multithread support nor dependency.

lib/zstd.h
```diff
@@ -17,6 +17,7 @@ extern "C" {
 /* ======   Dependency   ======*/
 #include <limits.h>   /* INT_MAX */
 #include <stddef.h>   /* size_t */
+#include <semaphore.h> /* sem_t */
```
Cyan4973, line 20: as a PoC, it's probably okay,
but just to be clear,
we don't take it lightly to increase the dependency load of `libzstd`,
and especially `zstd.h`, since it's transitively included into all user projects.
As such, such a change is unlikely to be accepted into this repository.

lib/common/error_private.c
```diff
@@ -48,6 +48,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(frameIndex_tooLarge): return "Frame index is too large";
     case PREFIX(seekableIO): return "An I/O error occurred when reading/seeking";
     case PREFIX(dstBuffer_wrong): return "Destination buffer is wrong";
+    case PREFIX(semaphore): return "Semaphore is corrupted";
```
Cyan4973, line 51: prefer `id_status` naming convention for error codes,
such as `semaphore_corruption` for example.

Cyan4973:
The main issue of this proposal is the dependency on <semaphore.h>,
which must be dealt with very carefully,
enabled on compatible systems only,
and not break `libzstd` on systems which do not offer this capability.

A less risky solution would be an overlay library, which could typically start its life in `contrib/`,
it would depend on `libzstd`, and add <semaphore.h>` capabilities.
However, considering that many changes are being done directly within `zstdmt_compress.c`, this may not be possible.

Finally, a potentially better way to add this capability could be to make several contexts share the same thread pool.
`libzstd` thread pool implementation is hosted in `lib/common/pool`.
Right now, each `ZSTDMT_CCtx*` has its own copy.
So the idea would be to take a reference instead, and externalize lifetime management.
More easily said than done.

lib/zstd.h
davidbolvansky: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

terrelln:
> Finally, an alternative way to add this capability could be to make several contexts share the same thread pool.
> libzstd thread pool implementation is hosted in lib/common/pool.
> Right now, each ZSTDMT_CCtx* has its own copy.
> So the idea would be to take a reference instead, and externalize lifetime management.
> More easily said than done.
> But this could be done with existing tools and dependencies, hence no need to add another dependency to the library.

I'm in favor of this approach. This is more general, and allows us to integrate libraries that provide their own thread pools. We could define a "thread pool" concept in `zstd.h` that `pool.h` fulfills. Then allow users to pass in a custom thread-pool, which we would use instead of creating a local pool. I guess it would look something like this:

```
typedef void (*ZSTD_threadPoolJob_t)(void* opaque);
typedef struct {
    void (*add)(ZSTD_threadPoolJob_t job, void* opaque); /* Must succeed eventually. May block. */
    int (*tryAdd)(ZSTD_threadPoolJob_t job, void* opaque); /* Returns 1 on success and 0 on failure. Does not block. */
} ZSTD_threadPool_t;

size_t ZSTD_CCtx_setThreadPool(ZSTD_CCtx* cctx, ZSTD_threadPool_t threadPool);
```
 
I believe that will be enough to implement multithreaded zstd. But some care will need to be taken to ensure that all jobs are finished before freeing a `ZSTD_CCtx`. which is currently handled by `POOL_join()`. I think the function `ZSTDMT_waitForAllJobsCompleted()` should do the trick.

Cyan4973:
Yes, that solution looks good

terrelln:
Another consideration is that if the user wants to use all the threads in a system, the threads themselves are pretty cheap. The scheduler will handle it well. A problem could be memory utilization if compressing at a high level though, which would not be solved by this. To solve that you would need to limit the maximum number of parallel compression jobs, regardless of the number of threads that are actually running. Since the first the compression will do is allocate memory.

I guess thats another argument for the external thread-pool. It gives all the control to the user, so they can tune their parallelism as they see fit. You lose some ease of use, but this is already a pretty complex topic, so I favor more flexibility over ease of use.

marxin:
Thank you for the fruity discussion that has started. I have a few comments:

- `zstd` doesn't have to depend on `semaphore.h` as it can come up with 2 callbacks that will be called before and after job spawning. `zstd` consumer can implement the logic with semaphores
- as @terrelln mentioned, I actually do care about the memory allocation and not about the number of threads. So isn't `ZSTDMT_bufferPool` the resource that we want to share among multiple `ZSTD_CCtx`s?
- or can be `ZSTD_customMem` extended to control the total amount of allocated memory?

marxin:
May I please remind the discussion. Is there any approach I can try implementing?

terrelln:
@marxin 

You can limit memory usage by limiting the number of concurrent compressions. Most memory goes to the compression tables, and is unrelated to multi-threading. So you will need to limit memory with the number of concurrent compression jobs. You could do that with a semaphore outside of zstd.

If you would like to submit a PR providing an API similar to what I described, I'd be happy to review & merge it. Otherwise, we will ensure that it makes it into a future release.

marxin:
> If you would like to submit a PR providing an API similar to what I described, I'd be happy to review & merge it. Otherwise, we will ensure that it makes it into a future release.

All right. This introduces the concept, the patch is still missing documentation. But I would like to be given feedback ;)

marxin:
May I please ping the concept?

lib/compress/zstdmt_compress.c
```diff
@@ -1612,6 +1623,11 @@ static size_t ZSTDMT_createCompressionJob(ZSTDMT_CCtx* mtctx, size_t srcSize, ZS
     }
 
     if (!mtctx->jobReady) {
+        if (mtctx->cJobControl.canCreateJob && !mtctx->cJobControl.canCreateJob (mtctx->cJobControl.opaque)) {
+          DEBUGLOG(5, "ZSTDMT_createCompressionJob: canCreateJob returned false");
+          return 0;
```
Cyan4973, line 1628: Expected behavior should be well documented here.
Is `canCreateJob` a blocking function ?
What's the impact of `return 0` ?
Will it cancel the job ? Will it make the user process enter a busy-loop ?

Cyan4973:
Hi @marxin 
this concept design is much closer to what we can integrate within `libzstd`.
It doesn't add any new dependency, leaving the `ZSTD_customJobControl` implementation to the caller.
Note that the expected behavior of user-defined `ZSTD_customJobControl` must be documented, so that users can understand what they have to provide. Adding an example, and/or even a ready-to-use implementation in `contrib/`, can be part of this documentation.

It's not what I had in mind when mentioning the possibility of a shared thread pool, since it's more a "parallel" counter that selectively accepts to start a new thread or not depending on its own set of rules, but I understand the attractivity of the design in combination with `semaphore.h`.

Among the downsides, since thread pools are still created independently within each context, each pool still allocates up front all the stack space for maximum thread usage. For example, if 8 contexts are created with a thread pool for 8 jobs, that's 64 threads created and stack space reserved, even if they may all be associated with the same `ZSTD_customJobControl` limiting usage to 8 threads total. It's not a killer blow, just good to know.

More importantly, it's unclear to me if the suggested `return 0` within `zstdmt` is sufficient. This should be thoroughly tested, with scenarios actually triggering the job control limit, to ensure if behaves as expected.

lib/compress/zstdmt_compress.c
marxin: I tried to document it properly in `zstd.h`, can you please take a look?
It's not a blocking function and you are right, it enters a busy-loop. For a compression running for 40s, I see the `canCreateJob returned false` message 850 times in log. Is it acceptable?

marxin:
> Note that the expected behavior of user-defined `ZSTD_customJobControl` must be documented, so that users can understand what they have to provide. Adding an example, and/or even a ready-to-use implementation in `contrib/`, can be part of this documentation.

I provided first version of documentation in `zstd.h` and there's a example in `examples/streaming_compression.c` that shows an implementation using a semaphore.

lib/compress/zstdmt_compress.c
Cyan4973: This is indeed a concern. Busy loops must be avoided.

To avoid this problem, `zstd` currently ensures forward progress : it only returns if at least one byte is consumed or produced.
When nothing happened, typically as a consequence of having no thread available to start a new job, it blocks the calling thread, up to the moment that "something" happens, typically a job just produced some data to flush, and/or got freed, hence allowing to start a new job.

This requires a capability to block / sleep the thread,
and more importantly, it requires a capability to wake up,
right on time, when a thread finally becomes available.

This is not trivial, and is likely even more difficult when trying to achieve a similar goal outside of the thread pool.

marxin: I see.

Note that it's for me quite difficult to adopt the job control to the existing `zstd` codebase.
Maybe a closer look from a developer can introduce a more elegant way?

Cyan4973:
Here is the way I would consider implementing it : 

- Make the `threadPool` an object that can be managed from the public interface, with usual `create`, `free`, `setParameter` accessors.
- Give a way for any `cctx` to reference that `threadPool` instead of creating their own. This would likely require an additional accessor, similar to what already exists to reference `cdict` objects.
- When an external `threadPool` is defined, use that instead of letting the `cctx` create its own `threadPool`.

Such design would result in minimal code changes, and would preserve all properties of current code, including avoiding busy loops, by getting the caller into sleep mode while waiting for a `thread` to become available.

This design relies on good lifetime management : the `threadPool` must outlive all its users. Though that's the same rule for `cdict` references and the `semaphore` example.

Long term, we could imagine that a user could define its own threadPool, and wrap its essential accessors into a published structure so that the `cctx` would use that instead of the one `threadPool` already defined within `libzstd`. Though that's clearly a second stage.

PR #2291, opened 2020-08-31
Fix setParameter not falling back to default compression level

i-do-cpp:
`ZSTD_CCtx_setParameter` does not adhere to the documented behavior of setting `0` as `compressionLevel`:  
`Special: value 0 means default, which is controlled by ZSTD_CLEVEL_DEFAULT`.  
Instead it does not change the compression level at all.  

This pull request adjusts the behavior to conform to the documentation.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -544,7 +544,6 @@ size_t ZSTD_CCtxParams_setParameter(ZSTD_CCtx_params* CCtxParams,
             CCtxParams->compressionLevel = ZSTD_CLEVEL_DEFAULT; /* 0 == default */
         else
             CCtxParams->compressionLevel = value;
-        }
         if (CCtxParams->compressionLevel >= 0) return (size_t)CCtxParams->compressionLevel;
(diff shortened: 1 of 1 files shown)


terrelln:
Thanks for the fix @i-do-cpp!

I will merge this soon. I just want to quickly see if I can see anyone that relies on the `0` means no-op behavior.

terrelln:
I just checked and I don't see any callers who expect that level 0 means no-op. This should be good to go. But, since this is technically a breaking API change, I want to get input from @Cyan4973 before I merge.

Cyan4973:
Indeed,
during the development of the advanced API, we had a few places where `0` meant "no-op", while others used `0` to mean "default".
As the interface matured, it steadily leaned more and more towards "default".
But I can see how some "no-op" could have sneakily remained in the code even when it was no longer intended.

In this case, the documentation is pretty clear, and makes perfect sense : level `0` already has a meaning, which is controlled by `ZSTD_CLEVEL_DEFAULT`. And this PR puts the implementation in line with the intention and the documentation.

Thanks @i-do-cpp !

Cyan4973 (approved):

PR #2292, opened 2020-09-03
Make multifile (de)compression print out a summary

senhuang42:
When processing multiple files, we now print out a summary of the process that includes # of files and compression ratio or bytes output. Combined with #2283, the experience of dealing with multiple files is more in line with that of dealing with single files.

image

This PR removes the stateful fields (introduced in #2283) from `FIO_prefs_t`, restoring it to a true "preferences" struct, and adds a new struct `FIO_ctx_t` that should serve as the one-stop-shop for info regarding the state of file i/o during (de)compression. I plan on following this up with another PR that moves more relevant fields into this struct and out of function arguments.
So something like this:
```
int FIO_compressFilename (FIO_prefs_t* const prefs, FIO_ctx_t* const fCtx,
                          const char* outfilename, const char* infilename,
                          const char* dictFileName, int compressionLevel,
                          ZSTD_compressionParameters comprParams);
```
will become this:
```
int FIO_compressFilename (const FIO_prefs_t* const prefs, FIO_ctx_t* const fCtx,
                          int compressionLevel, ZSTD_compressionParameters comprParams);
```
and should make further adjustments in this vein a lot less annoying (as evidenced by how annoying it was to add just one argument to `FIO_compressFilename()`)


senhuang42:
Some silent merge incompatibilities popped up - hold on...

programs/fileio.h
```diff
@@ -96,22 +103,27 @@ void FIO_setNotificationLevel(int level);
 void FIO_setExcludeCompressedFile(FIO_prefs_t* const prefs, int excludeCompressedFiles);
 void FIO_setPatchFromMode(FIO_prefs_t* const prefs, int value);
 void FIO_setContentSize(FIO_prefs_t* const prefs, int value);
-void FIO_setNbFiles(FIO_prefs_t* const prefs, int value);
-void FIO_setCurrFileIdx(FIO_prefs_t* const prefs, int value);
+
+/* FIO_ctx_t functions */
+void FIO_setNbFilesTotal(FIO_ctx_t* const fCtx, int value);
```
senhuang42, line 108: So, I'm aware that I never actually use these functions - is it more idiomatic to have `set()`-like functions for structs or to just not have them in the first place?

programs/fileio.c
```diff
@@ -1478,18 +1530,24 @@ FIO_compressFilename_internal(FIO_prefs_t* const prefs,
...
+                DISPLAYLEVEL(2,"%-20s :%6.2f%%   (%6llu => %6llu bytes, %s) \n",
+                    srcFileName,
+                    (double)compressedfilesize / readsize * 100,
+                    (unsigned long long)readsize, (unsigned long long) compressedfilesize,
+                    dstFileName);
+            }
+        }
+        if (fCtx->nbFilesTotal > 1) {
+            fCtx->totalBytesInput += (size_t)readsize;
+            fCtx->totalBytesOutput += (size_t)compressedfilesize;
```
terrelln, line 1550: I think it would make sense to always keep track of `totalBytesInput` and `totalBytesOutput`. Even if `g_display_prefs.displayLevel < 2`. Future readers will probably expect that these fields are always accurate.

programs/fileio.c
```diff
@@ -2378,8 +2451,13 @@ static int FIO_decompressFrames(dRess_t ress, FILE* srcFile,
     /* Final Status */
     DISPLAYLEVEL(2, "\r%79s\r", "");
     /* No status message in pipe mode (stdin - stdout) or multi-files mode */
-    if (g_display_prefs.displayLevel > 2 || (g_display_prefs.displayLevel == 2 && prefs->nbFiles <= 1)) {
+    if (g_display_prefs.displayLevel >= 2) {
+        if (fCtx->nbFilesTotal <= 1 || g_display_prefs.displayLevel >= 3) {
             DISPLAYLEVEL(2, "%-20s: %llu bytes \n", srcFileName, filesize);
+        }
+        if (fCtx->nbFilesTotal > 1) {
+            fCtx->totalBytesOutput += (size_t)filesize;
```
terrelln, line 2459: Same comment here

programs/fileio.h
terrelln: We don't need them. They are present for `FIO_prefs_t` so that callers can set fields in `FIO_prefs_t` without having the struct definition.

We only need these functions if someone outside of `fileio.c` needs to set a field on the `FIO_ctx_t`.

terrelln (commented):
I love the refactor, just a few small comments!

programs/fileio.h
senhuang42: Ah that makes sense - I was also wondering why we always define the struct with a `_s` suffix and typedef it into `_t`, but that seems to explain that as well!

terrelln (approved):

terrelln:
Awesome!

programs/fileio.c
```diff
@@ -800,10 +830,10 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
  *         If neither flag is specified, zstd will prompt the user for confirmation to proceed.
  * If --rm is not specified, then zstd will print a warning to the user (which can be silenced with -q).
  */
-static int FIO_removeMultiFilesWarning(const FIO_prefs_t* const prefs, int displayLevelCutoff, const char* outFileName)
+static int FIO_removeMultiFilesWarning(const FIO_prefs_t* const prefs, FIO_ctx_t* const fCtx, int displayLevelCutoff, const char* outFileName)
```
Cyan4973, line 833: minor :
argument order (programming pattern)

Prefer having the object _state_ (when it exists) as the first argument in the list, in this case `fCtx`.
Then followed by mutable arguments when they exist (such as `void* dst`), then immutable ones.
`prefs` belong to the read-only parameter category, and should get further down the list of parameters.

It's merely a pattern we use everywhere.
Obviously, it doesn't change a thing to the way the function works.

programs/fileio.h
```diff
@@ -96,22 +103,23 @@ void FIO_setNotificationLevel(int level);
...
 
 /*-*************************************
 *  Single File functions
 ***************************************/
 /** FIO_compressFilename() :
  * @return : 0 == ok;  1 == pb with src file. */
-int FIO_compressFilename (FIO_prefs_t* const prefs,
+int FIO_compressFilename (FIO_prefs_t* const prefs, FIO_ctx_t* const fCtx,
```
Cyan4973, line 115: The comment regarding parameter order is a little bit more important here
because it's applied to public functions, exposed via `fileio.h`.

Also, a question : is `prefs` still mutable ?

Cyan4973:
Thanks @senhuang42 ,
it's a very nice improvement,
both on UI side, and on code refactoring.

My only comment is to try to apply a programming pattern we are used to regarding argument orders,
specifically, any function signature like `PREFIX_doAction(object, parameters)`
should try to present the object's state as the first argument of the function.
That's relatively minor.

programs/fileio.h
senhuang42: After this PR, none of my recent changes should mutate `prefs`, but I'm not sure if there's somewhere else in the code that might modify `prefs` during (de)compression.

In a followup refactor PR focused more on making full use of `fCtx` I will also try to modify the relevant function signatures to use `const FIO_prefs_t* const prefs` to ensure it's not mutable.

Cyan4973 (approved):

PR #2293, opened 2020-09-03
Resolve Coverity 1432392 Unintentional integer overflow

allanjude:
Unintentional integer overflow (OVERFLOW_BEFORE_WIDEN)
overflow_before_widen: Potentially overflowing expression:
cdict->dictContentSize * 6U
with type unsigned int (32 bits, unsigned) is evaluated using 32-bit arithmetic, and then used in a context that expects an expression of type U64 (64 bits, unsigned).


terrelln (approved):
LGTM! I will merge once CI passes.

PR #2295, opened 2020-09-04
DDSS for Lazy: Implement a Dedicated Dictionary Chain Table

felixhandte:
This PR builds on #2276 to also restructure the chain table for the strategies `ZSTD_greedy`, `ZSTD_lazy`, and `ZSTD_lazy2`.

With this PR, the cumulative perf improvement seen for the Dedicated Dictionary Search Structure stack, for the HTTP test corpus, is:

Level | Hot Dict | Cold Dict
----- | -------- | ---------
  5   |   ~+17%  |    ~+30%
  6   |   ~+12%  |    ~+45%
  7   |   ~+13%  |    ~+40%
  8   |   ~+16%  |    ~+50%
  9   |   ~+19%  |    ~+65%
 10   |   ~+24%  |    ~+70%

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -483,7 +483,6 @@ void ZSTD_dedicatedDictSearch_lazy_loadDictionary(ZSTD_matchState_t* ms, const B
     U32* const hashTable = ms->hashTable;
     U32* const chainTable = ms->chainTable;
     U32 const chainSize = 1 << ms->cParams.chainLog;
-    U32 const chainMask = chainSize - 1;
     U32 idx = ms->nextToUpdate;
     U32 const minChain = chainSize < target ? target - chainSize : idx;
     U32 const bucketSize = 1 << ZSTD_LAZY_DDSS_BUCKET_LOG;
@@ -493,24 +492,27 @@ void ZSTD_dedicatedDictSearch_lazy_loadDictionary(ZSTD_matchState_t* ms, const B
 
     /* We know the hashtable is oversized by a factor of `bucketSize`.
      * We are going to temporarily pretend `bucketSize == 1`, keeping only a
-     * single entry. We will use
-     * the rest of the space to construct a temporary chaintable.
+     * single entry. We will use the rest of the space to construct a temporary
+     * chaintable.
      */
     U32 const hashLog = ms->cParams.hashLog - ZSTD_LAZY_DDSS_BUCKET_LOG;
     U32* const tmpHashTable = hashTable;
     U32* const tmpChainTable = hashTable + (1 << hashLog);
+    U32 const tmpChainSize = ((1 << ZSTD_LAZY_DDSS_BUCKET_LOG) - 1) << hashLog;
+    U32 const tmpMinChain = tmpChainSize < target ? target - tmpChainSize : idx;
 
     U32 hashIdx;
 
     assert(ms->cParams.chainLog <= 24);
     assert(ms->cParams.hashLog >= ms->cParams.chainLog);
     assert(idx != 0);
+    assert(tmpMinChain <= minChain);
 
     /* fill conventional hash table and conventional chain table */
     for ( ; idx < target; idx++) {
         U32 const h = ZSTD_hashPtr(base + idx, hashLog, ms->cParams.minMatch);
-        if (idx >= minChain) {
-            tmpChainTable[idx & chainMask] = hashTable[h];
+        if (idx >= tmpMinChain) {
+            tmpChainTable[idx - tmpMinChain] = hashTable[h];
         }
         tmpHashTable[h] = idx;
     }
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_lazy.c
```diff
@@ -475,6 +476,122 @@ U32 ZSTD_insertAndFindFirstIndex(ZSTD_matchState_t* ms, const BYTE* ip) {
     return ZSTD_insertAndFindFirstIndex_internal(ms, cParams, ip, ms->cParams.minMatch);
 }
 
+void ZSTD_dedicatedDictSearch_lazy_loadDictionary(ZSTD_matchState_t* ms, const BYTE* const ip)
```
terrelln, line 479: The DDS has its own indexing scheme. I wonder if we can add `chainSize + 1` to the index, so we can distinguish chain table indices from source indices. That would allow us to fit 4-entry chains into the hash table without spilling over.

1. Index `chainSize` (or 0, whichever is faster) means no chain - we have reached the end of our chain
2. Indices `[0, chainSize)` mean this is where the chain starts. We necessarily need a size, because we could `0` terminate the chain. Generally searching for a `0` element is faster than iterating over a size.
3. Any index in the hash table `>= chainSize + 1` is a valid index. If this is the 4th element in the hash table, then there is no chain.

Or, I wonder if we could fit both a chain index and a position in the hash table. If we limit dedicatedDictSearch to dictionaries <= 128KB we have 15 bits leftover for the chain table index. That means we can fit 32K entries in our chain table. That would allow us to utilize all 4 entries in our hash table. Then the first entry who has non-zero high 15 bits is the last entry in the hash table, and that is the position in the chain table.

I'm not sure if either is worth the effort. But something along these lines might pan out and be worth trying.

lib/compress/zstd_lazy.c
```diff
@@ -475,6 +476,122 @@ U32 ZSTD_insertAndFindFirstIndex(ZSTD_matchState_t* ms, const BYTE* ip) {
...
+                             * to replace the entries pulled out of the
+                             * chainTable into the cache. This lets us reach
+                             * back further without increasing the total number
+                             * of entries in the chainTable, guaranteeing the
+                             * DDSS chain table will fit into the space
+                             * allocated for the regular one. */
+                            break;
+                        }
+                    }
+                    chainTable[chainPos++] = i;
```
terrelln, line 551: Does this possibly overwrite chain positions?

lib/compress/zstd_lazy.c
```diff
@@ -531,8 +661,92 @@ size_t ZSTD_HcFindBestMatch_generic (
...
+            U32 const chainPackedPointer = dms->hashTable[ddsIdx + bucketSize - 1];
+            U32 chainIndex = chainPackedPointer >> 8;
+            U32 const chainLength = chainPackedPointer & 0xFF;
+            U32 const chainAttempts = nbAttempts - ddsAttempt;
+            U32 const chainLimit = chainAttempts > chainLength ? chainLength : chainAttempts;
+            U32 chainAttempt;
+
+            for (chainAttempt = 0 ; chainAttempt < chainLimit; chainAttempt++) {
+                PREFETCH_L1(ddsBase + dms->chainTable[chainIndex + chainAttempt]);
+            }
```
terrelln, line 725: Ideally we'd be able to do these sooner. But we can consider that in the follow-up diff which goes into prefetching in depth.

lib/compress/zstd_lazy.c
```diff
@@ -641,27 +871,47 @@ ZSTD_compressBlock_lazy_generic(
     typedef size_t (*searchMax_f)(
                         ZSTD_matchState_t* ms,
                         const BYTE* ip, const BYTE* iLimit, size_t* offsetPtr);
-    searchMax_f const searchMax = dictMode == ZSTD_dictMatchState ?
-        (searchMethod==search_binaryTree ? ZSTD_BtFindBestMatch_dictMatchState_selectMLS
-                                         : ZSTD_HcFindBestMatch_dictMatchState_selectMLS) :
-        (searchMethod==search_binaryTree ? ZSTD_BtFindBestMatch_selectMLS
-                                         : ZSTD_HcFindBestMatch_selectMLS);
+
+    const searchMax_f searchFuncs[4][2] = {
```
Cyan4973, line 875: Please document the meaning behind table indexes `[4]` and `[2]`

lib/compress/zstd_lazy.c
```diff
@@ -641,27 +871,47 @@ ZSTD_compressBlock_lazy_generic(
...
+    };
+
+    searchMax_f const searchMax = searchFuncs[dictMode][searchMethod == search_binaryTree];
     U32 offset_1 = rep[0], offset_2 = rep[1], savedOffset=0;
 
+    const int isDMS = dictMode == ZSTD_dictMatchState;
+    const int isDDS = dictMode == ZSTD_dedicatedDictSearch;
     const ZSTD_matchState_t* const dms = ms->dictMatchState;
-    const U32 dictLowestIndex      = dictMode == ZSTD_dictMatchState ?
+    const U32 dictLowestIndex      = isDMS || isDDS ?
```
Cyan4973, line 900: minor :
since `isDMS || isDDS` seems to be present in many places,
maybe it could pre-calculated into a `const` variable (maybe `isDxS`, or something more meaningful).

It might help compiler optimization, but the more important point is to help code readability.

Cyan4973:
Only some minor comments, and I think this first PR is good to merge.

I had additional comments on potential alternative optimization techniques worth exploring, but we can keep such topic for a dedicated meeting and a following PR.

Cyan4973 (approved):

PR #2299, opened 2020-09-08
Allow environment variable to specify number of threads for compression

senhuang42:
Using `ZSTD_NUMTHREADS` as the environment variable. `-T#` supersedes this setting.

### DIFF
--- a/programs/README.md
+++ b/programs/README.md
@@ -204,14 +204,19 @@ Benchmark arguments :
 ### Passing parameters through Environment Variables
 `ZSTD_CLEVEL` can be used to modify the default compression level of `zstd`
 (usually set to `3`) to another value between 1 and 19 (the "normal" range).
-This can be useful when `zstd` CLI is invoked in a way that doesn't allow passing arguments.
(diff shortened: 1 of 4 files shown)


terrelln (approved):

Cyan4973:
very minor comment : 
I tend to favor `ZSTD_NBTHREADS` instead of `ZSTD_NUMTHREADS`.

Yeah, that's very subtle, but we have some precedent in our [existing API](https://github.com/facebook/zstd/blob/v1.4.5/lib/zstd.h#L370) naming convention, so I guess it's better to remain consistent.

To expand on this, in the code, `nbThreads` will rather be the _total_ number of threads, 
while `threadNum` will rather be the number of the _current_ thread, typically within a loop 
(like a `threadID`, but within some local context).

As I write it, I realize that this distinction makes more sense in french, 
where we have "number => nombre", which is a concept closer to a quantity, 
and "number = numero", which is closer to an ID.
In english, this is spelled "number" for both, which makes it easier to conflated the 2 topics.

gebi:
awesome thx :)!

Cyan4973 (approved):

Issue #2300, opened 2020-09-09
compressionLevel is not respected in this case


animalize:
This [test code](https://github.com/animalize/misc/blob/master/size_use_dict.c) read all files in a folder to memory, and train a dictionary with these files:
(usage: modify `PATH` to zstd/lib/compress folder, should ends with `\` or `/` depend on the system)

Then compare these two cases:

1, One shot compress, all files concatenated in a single flat buffer.
```c
    // load dictionary to context
    cdict = ZSTD_createCDict(dict_buffer, DICT_SIZE, compression_level);
    ZSTD_CCtx_refCDict(cctx, cdict);
    
    // one shot compress
    ZSTD_compressStream2(cctx, &out, &in, ZSTD_e_end);
```

2, Stream compress, file by file.
```c
    // load dictionary to context
    cdict = ZSTD_createCDict(dict_buffer, DICT_SIZE, compression_level);
    ZSTD_CCtx_refCDict(cctx, cdict);
    
    // stream compress, file by file
    in.src = &input_buffer;
    for (int i = 0; i < content_count; i++) {
        in.size = content_sizes[i];
        in.pos = 0;
        
        ZSTD_compressStream2(cctx, &out, &in, ZSTD_e_continue);
        
        in.src += content_sizes[i];
    }
    // end the frame
    in.size = 0;
    in.pos = 0;
    ZSTD_compressStream2(cctx, &out, &in, ZSTD_e_end);
```

It seems one shot compress doesn't respect `compressionLevel`, its output size is always `121086` bytes.

The output of test code:
```
Read 25 files, total read size: 687379

compressionLevel 0:
One shot compress, output size: 121086
Stream compress, output size: 120456

compressionLevel 1:
One shot compress, output size: 121086
Stream compress, output size: 150086

compressionLevel 2:
One shot compress, output size: 121086
Stream compress, output size: 136046

compressionLevel 3:
One shot compress, output size: 121086
Stream compress, output size: 120456

compressionLevel 4:
One shot compress, output size: 121086
Stream compress, output size: 119666

compressionLevel 5:
One shot compress, output size: 121086
Stream compress, output size: 109670

compressionLevel 6:
One shot compress, output size: 121086
Stream compress, output size: 102324

compressionLevel 7:
One shot compress, output size: 121086
Stream compress, output size: 100145

compressionLevel 8:
One shot compress, output size: 121086
Stream compress, output size: 97499

compressionLevel 9:
One shot compress, output size: 121086
[...]

PR #2303, opened 2020-09-10
For ZSTD_compressStream2(), let cdict take compression level priority

senhuang42:
This PR does resolves two issues brought up in #2300, discovered by @animalize:
1) When using ZSTD_compressStream2(), if a `cdict` was used during compression, that cdict's compression level was not respected. Now the `cdict` will take priority in terms of compression level, but will not change `cctx->requestedParams`.
2) Fixes `ZSTD_createCDict_byReference()` to use the overloaded `compressionLevel = 0` behavior.

No unit tests included in this PR, but tested locally (and by @animalize in #2300) to confirm that using `ZSTD_compressStream2()` does in fact take on the `cdict` compression level when compressing files that would cause a dictionary reload, and that everything else works as usual.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3469,9 +3469,12 @@ ZSTD_CDict* ZSTD_createCDict(const void* dict, size_t dictSize, int compressionL
 ZSTD_CDict* ZSTD_createCDict_byReference(const void* dict, size_t dictSize, int compressionLevel)
 {
     ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
-    return ZSTD_createCDict_advanced(dict, dictSize,
+    ZSTD_CDict* cdict = ZSTD_createCDict_advanced(dict, dictSize,
                                      ZSTD_dlm_byRef, ZSTD_dct_auto,
                                      cParams, ZSTD_defaultCMem);
+    if (cdict)
+        cdict->compressionLevel = compressionLevel == 0 ? ZSTD_CLEVEL_DEFAULT : compressionLevel;
+    return cdict;
 }
 
 size_t ZSTD_freeCDict(ZSTD_CDict* cdict)
@@ -3989,6 +3992,8 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
         FORWARD_IF_ERROR( ZSTD_initLocalDict(cctx) , ""); /* Init the local dict if present. */
         ZSTD_memset(&cctx->prefixDict, 0, sizeof(cctx->prefixDict));   /* single usage */
         assert(prefixDict.dict==NULL || cctx->cdict==NULL);    /* only one can be set */
+        if (cctx->cdict)
+            cctx->requestedParams.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
         DEBUGLOG(4, "ZSTD_compressStream2 : transparent init stage");
         if (endOp == ZSTD_e_end) cctx->pledgedSrcSizePlusOne = input->size + 1;  /* auto-fix pledgedSrcSize */
         params.cParams = ZSTD_getCParamsFromCCtxParams(
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3993,11 +3993,10 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
         ZSTD_memset(&cctx->prefixDict, 0, sizeof(cctx->prefixDict));   /* single usage */
         assert(prefixDict.dict==NULL || cctx->cdict==NULL);    /* only one can be set */
         if (cctx->cdict)
-            cctx->requestedParams.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
+            params.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
         DEBUGLOG(4, "ZSTD_compressStream2 : transparent init stage");
         if (endOp == ZSTD_e_end) cctx->pledgedSrcSizePlusOne = input->size + 1;  /* auto-fix pledgedSrcSize */
-        params.cParams = ZSTD_getCParamsFromCCtxParams(
-                &cctx->requestedParams, cctx->pledgedSrcSizePlusOne-1, 0 /*dictSize*/);
+        params.cParams = ZSTD_getCParamsFromCCtxParams(&params, cctx->pledgedSrcSizePlusOne-1, 0 /*dictSize*/);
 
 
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -3469,9 +3469,12 @@ ZSTD_CDict* ZSTD_createCDict(const void* dict, size_t dictSize, int compressionL
 ZSTD_CDict* ZSTD_createCDict_byReference(const void* dict, size_t dictSize, int compressionLevel)
 {
     ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
-    return ZSTD_createCDict_advanced(dict, dictSize,
+    ZSTD_CDict* cdict = ZSTD_createCDict_advanced(dict, dictSize,
```
Cyan4973, line 3472: minor : coding style :
prefer `ZSTD_CDict* const cdict = `
which indicates that, once created, the `cdict` pointer will never change.

lib/compress/zstd_compress.c
```diff
@@ -3469,9 +3469,12 @@ ZSTD_CDict* ZSTD_createCDict(const void* dict, size_t dictSize, int compressionL
...
                                      ZSTD_dlm_byRef, ZSTD_dct_auto,
                                      cParams, ZSTD_defaultCMem);
+    if (cdict)
+        cdict->compressionLevel = compressionLevel == 0 ? ZSTD_CLEVEL_DEFAULT : compressionLevel;
```
Cyan4973, line 3476: minor : coding style : 
I have a preference for  : 
` X = (compressionLevel == 0) ? val1 : val2;`
where the test is cleanly encapsulated between parenthesis.
At least for me, it improves clarity of the ternary expression.

lib/compress/zstd_compress.c
```diff
@@ -3989,6 +3992,8 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
         FORWARD_IF_ERROR( ZSTD_initLocalDict(cctx) , ""); /* Init the local dict if present. */
         ZSTD_memset(&cctx->prefixDict, 0, sizeof(cctx->prefixDict));   /* single usage */
         assert(prefixDict.dict==NULL || cctx->cdict==NULL);    /* only one can be set */
+        if (cctx->cdict)
+            cctx->requestedParams.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
```
Cyan4973, line 3996: This operation overwrites a parameter set by the user.
We generally try to avoid that, 
it can have surprising impacts in later usages of `cctx` since these parameters are sticky.

That's also the reason for the existence of `appliedParams`,
which is where dynamic adaptations are performed,
there they do not pollute user's explicit requests, nor survive the current session.

Cyan4973:
2 minor comments regarding the PR's presentation : 
- reference the issues that are solved by this PR (it will also automatically close them on merge)
- name the contributor which inspired these changes

lib/compress/zstd_compress.c
senhuang42: I've looked into this issue some more - it seems like this behavior also isn't consistent between `ZSTD_compress_usingCDict()` and `ZSTD_compressStream2()`. `ZSTD_compress_usingCDict()` doesn't seem to have this issue, and always will use the CDict compression level.

The reason is: in `ZSTD_compress_usingCDict()`, we essentially have:
```
ZSTD_CCtx_params params = cctx->requestedParams
[...]
params.cParams = ZSTD_getCParamsFromCDict() OR params.cParams = ZSTD_getCParams()
// where both these functions take cdict compression level into account
```
Whereas, in `ZSTD_compressStream2()`, we have
```
ZSTD_CCtx_params params = cctx->requestedParams
params.cParams = ZSTD_getCParamsFromCCtxParams(&cctx->requestedParams, cctx->pledgedSrcSizePlusOne-1, 0 /*dictSize*/));
// followed immediately (within ZSTD_resetCStream_internal) by
params.cParams = ZSTD_getCParamsFromCCtxParams(&params, pledgedSrcSize, dictSize);
// in neither case is cdict compression level taken into account
```

I guess I'm a little perplexed here - what's the point of calling `ZSTD_getCParamsFromCCtxParams()` twice back-to-back, with what are essentially the same arguments (with the exception of dictSize)? Doesn't the second call just overwrite?

Cyan4973: It's likely redundant in this case.

But `params` is also used for other purposes when `ZSTD_MULTITHREAD` is defined : 
- It makes it possible to adjust `params.nbWorkers` without modifying `cctx->requestedParams`
- It's a parameter of `ZSTDMT_initCStream_internal()`

Also, consider that calling `ZSTD_getCParamsFromCCtxParams()` twice is a very minor impact on performance, making this issue less critical to be adressed.

Nevertheless, if you believe you can improve the situation, you are welcomed.

lib/compress/zstd_compress.c
```diff
@@ -3989,10 +3992,11 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
         FORWARD_IF_ERROR( ZSTD_initLocalDict(cctx) , ""); /* Init the local dict if present. */
         ZSTD_memset(&cctx->prefixDict, 0, sizeof(cctx->prefixDict));   /* single usage */
         assert(prefixDict.dict==NULL || cctx->cdict==NULL);    /* only one can be set */
+        if (cctx->cdict)
```
senhuang42, line 3995: After some more digging, the reason for the getting the correct clevel for filesizes + dictsizes that do not qualify for reloading is that downstream in `ZSTD_compressBegin_internal()`, the CCtx is readjusted using `ZSTD_resetCCtx_usingCDict()`. 

However, in the case of reloading, the function called was `ZSTD_resetCCtx_internal()` which doesn't take into account the `cdict`.

So I think the mitigation is to adjust the `params` upstream (which is a copy of requestedParams), and later on if we need to reload the dict and call `ZSTD_resetCCtx_internal()`, this very `params` which may have the `cdict` compression level, will be copied into the `cctx->appliedParams` during `ZSTD_resetCCtx_internal()`, and in this case, we don't modify the `cctx->requestedParams`.

I'm not sure if this is exactly what you meant by using the `appliedParams`, but it does seem like the `appliedParams` are mainly touched in the `ZSTD_resetCCtx_internal()`

Cyan4973: I think that's fine.

`appliedParams` is the place where the _final_ result of parameter adjustments will be stored.

Before reaching that point, `appliedParams` could also be used as a form of "temporary storage", where intermediate adjustment results are saved.
However, this could prove messy and confusing, hence separating the 2 topics, by creating and using a dedicated `params` structure for the purpose of temporary storage and intermediate adjustment is fine, probably clearer.

senhuang42:
Are there any further follow-up/changes required on this PR to merge?

Cyan4973 (approved):

Issue #2309, opened 2020-09-15
CLI display bug: Multifile summary lines not clearing fully


senhuang42:
On current `dev` branch:

If a large update is printed in the status bar and the next update isn't quite as large, the rest isn't overwritten.
![image](https://user-images.githubusercontent.com/17059792/93234869-d74cda00-f74a-11ea-8b0a-fb41d8dfa421.png)

Similarly, the multifile compression summary may include some leftovers from the status bar.
![image](https://user-images.githubusercontent.com/17059792/93241253-1848ec80-f753-11ea-81c5-ac990a86829c.png)

PR #2310, opened 2020-09-15
Fix multifile status bar and summary - clear out extraneous characters

senhuang42:
To address #2309 

This PR will make it so that the lines printed throughout the multifile status bar update are roughly the same size each time, and adds a bit of padding at the end, so it will more successfully overwrite. We also add 40 characters of whitespace to the right of the printed summary, which would override any leftover characters from the status bar.

I'm not sure if there are better ways to do this. There are still theoretically rare/extreme cases in which we still may leave behind some leftover characters (if the number of megabytes read for a given file are more than tens of thousands, and then the next file is small). Escape codes would be a cleaner solution, but aren't portable.

Tested locally to make sure it works as intended, and that the leftover characters are overwritten with whitespace:
image

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1338,14 +1338,23 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
                                 cShare );
                 } else {   /* summarized notifications if == 2 */
                     if (fCtx->nbFilesTotal > 1) {
-                        DISPLAYLEVEL(2, "\rCompressing %u/%u files. Current source: %s ", fCtx->currFileIdx+1, fCtx->nbFilesTotal, srcFileName);
+                        size_t srcFileNameSize = strlen(srcFileName);
+                        /* Ensure that the string we print is roughly the same size each time */
+                        if (srcFileNameSize > 20) {
+                            const char* truncatedSrcFileName = srcFileName + srcFileNameSize - 17;
+                            DISPLAYLEVEL(2, "\rCompressing %2u / %2u files. Current source: ...%s ",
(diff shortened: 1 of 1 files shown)


Cyan4973:
Some general ideas : 
- when updating (and overwriting) a single line, try to keep the total character budget <= 79.
   + this will ensure that it remains a "single line". Otherwise, 80 columns screens will start to display 2+ lines, and `\r` will no longer point back to the beginning of the line.
- 79 characters is actually not that large; try to keep as much of it usable.
   + Ex1 : `current source:` => would `current:` be enough ?
   + Ex2 : `compressing` => would `compress` be enough ?
- controlling maximum length of variable large fields is key (as this PR does for `path/to/filename`)
- some fields can receive a display size controlled by a variable, to fit in the target budget (also employed in this PR)
- if line budget can be constrained <= 79 characters, then it's possible to employ some generic "line cleaner" whenever it's unclear if adding a few space characters will be enough to remove left-overs, like `DISPLAYLEVEL(2, "\r%79s\r", "");` (already used somewhere within `fileio.c`).
- rare exceptions can happen, don't over-engineer it

senhuang42:
Ah good point - the 80 character budget and the `DISPLAYLEVEL(2, "\r%79s\r", "");` is really useful!

senhuang42:
Took some of your suggestions. Now, what this PR does two things:
1) Keep the printed lines to be within 79 chars
2) Clear everything out before the final summary to avoid dangling chars from previous print (which is not necessary for decompression since it uses `DISPLAYUPDATE()` instead)

Cyan4973 (approved):

PR #2311, opened 2020-09-15
Fix Compression Parameter Derivation Bugs Introduced by DDSS Changes

felixhandte:
The regression tests discovered a couple ways in which #2276 and #2295 caused failures to correctly derive compression parameters. The regression test results now have a much smaller diff against `v1.4.6`:

```diff
--- tests/regression/results.csv
+++ results-eee51a66.csv
@@ -26,4 +26,4 @@
-silesia,     level 5,           compress cctx,                4710237
-silesia,     level 6,           compress cctx,                4660057
-silesia,     level 7,           compress cctx,                4596295
-silesia,     level 9,           compress cctx,                4543924
+silesia,     level 5,           compress cctx,                4710236
+silesia,     level 6,           compress cctx,                4660056
+silesia,     level 7,           compress cctx,                4596296
+silesia,     level 9,           compress cctx,                4543925
@@ -39 +39 @@
-silesia,     explicit params,   compress cctx,                4794666
+silesia,     explicit params,   compress cctx,                4794677
@@ -90,4 +90,4 @@
-silesia,     level 5,           zstdcli,                      4710285
-silesia,     level 6,           zstdcli,                      4660105
-silesia,     level 7,           zstdcli,                      4596343
-silesia,     level 9,           zstdcli,                      4543972
+silesia,     level 5,           zstdcli,                      4710284
+silesia,     level 6,           zstdcli,                      4660104
+silesia,     level 7,           zstdcli,                      4596344
+silesia,     level 9,           zstdcli,                      4543973
@@ -103 +103 @@
-silesia,     explicit params,   zstdcli,                      4797100
+silesia,     explicit params,   zstdcli,                      4797112
@@ -129 +129 @@
-silesia.tar, explicit params,   zstdcli,                      4822354
+silesia.tar, explicit params,   zstdcli,                      4822362
@@ -149 +149 @@
-github,      level 5 with dict, zstdcli,                      40938
+github,      level 5 with dict, zstdcli,                      40741
@@ -180,4 +180,4 @@
-silesia,     level 5,           advanced one pass,            4710237
-silesia,     level 6,           advanced one pass,            4660057
-silesia,     level 7,           advanced one pass,            4596295
-silesia,     level 9,           advanced one pass,            4543924
+silesia,     level 5,           advanced one pass,            4710236
+silesia,     level 6,           advanced one pass,            4660056
+silesia,     level 7,           advanced one pass,            4596296
+silesia,     level 9,           advanced one pass,            4543925
@@ -194 +194 @@
-silesia,     explicit params,   advanced one pass,            4797086
+silesia,     explicit params,   advanced one pass,            4797095
@@ -220 +220 @@
-silesia.tar, explicit params,   advanced one pass,            4808581
+silesia.tar, explicit params,   advanced one pass,            4808589
@@ -272,4 +272,4 @@
-silesia,     level 5,           advanced one pass small out,  4710237
-silesia,     level 6,           advanced one pass small out,  4660057
-silesia,     level 7,           advanced one pass small out,  4596295
-silesia,     level 9,           advanced one pass small out,  4543924
+silesia,     level 5,           advanced one pass small out,  4710236
+silesia,     level 6,           advanced one pass small out,  4660056
+silesia,     level 7,           advanced one pass small out,  4596296
+silesia,     level 9,           advanced one pass small out,  4543925
@@ -286 +286 @@
-silesia,     explicit params,   advanced one pass small out,  4797086
+silesia,     explicit params,   advanced one pass small out,  4797095
@@ -312 +312 @@
-silesia.tar, explicit params,   advanced one pass small out,  4808581
+silesia.tar, explicit params,   advanced one pass small out,  4808589
@@ -364,4 +364,4 @@
-silesia,     level 5,           advanced streaming,           4710237
-silesia,     level 6,           advanced streaming,           4660057
-silesia,     level 7,           advanced streaming,           4596295
-silesia,     level 9,           advanced streaming,           4543924
+silesia,     level 5,           advanced streaming,           4710236
+silesia,     level 6,           advanced streaming,           4660056
+silesia,     level 7,           advanced streaming,           4596296
+silesia,     level 9,           advanced streaming,           4543925
@@ -378 +378 @@
-silesia,     explicit params,   advanced streaming,           4797100
+silesia,     explicit params,   advanced streaming,           4797112
@@ -404 +404 @@
-silesia.tar, explicit params,   advanced streaming,           4808608
+silesia.tar, explicit params,   advanced streaming,           4808618
@@ -456,4 +456,4 @@
-silesia,     level 5,           old streaming,                4710237
-silesia,     level 6,           old streaming,                4660057
-silesia,     level 7,           old streaming,                4596295
-silesia,     level 9,           old streaming,                4543924
+silesia,     level 5,           old streaming,                4710236
+silesia,     level 6,           old streaming,                4660056
+silesia,     level 7,           old streaming,                4596296
+silesia,     level 9,           old streaming,                4543925
@@ -524,4 +524,4 @@
-silesia,     level 5,           old streaming advanced,       4710237
-silesia,     level 6,           old streaming advanced,       4660057
-silesia,     level 7,           old streaming advanced,       4596295
-silesia,     level 9,           old streaming advanced,       4543924
+silesia,     level 5,           old streaming advanced,       4710236
+silesia,     level 6,           old streaming advanced,       4660056
+silesia,     level 7,           old streaming advanced,       4596296
+silesia,     level 9,           old streaming advanced,       4543925
@@ -538 +538 @@
-silesia,     explicit params,   old streaming advanced,       4797100
+silesia,     explicit params,   old streaming advanced,       4797112
@@ -564 +564 @@
-silesia.tar, explicit params,   old streaming advanced,       4808608
+silesia.tar, explicit params,   old streaming advanced,       4808618
@@ -630,4 +630,4 @@
-github,      level 5 with dict, old streaming advanced cdict, 39158
-github,      level 6 with dict, old streaming advanced cdict, 38748
-github,      level 7 with dict, old streaming advanced cdict, 38744
-github,      level 9 with dict, old streaming advanced cdict, 38992
+github,      level 5 with dict, old streaming advanced cdict, 39159
+github,      level 6 with dict, old streaming advanced cdict, 38749
+github,      level 7 with dict, old streaming advanced cdict, 38746
+github,      level 9 with dict, old streaming advanced cdict, 38993
```

I'm still not sure why the `explicit params` case is so slightly different.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3540,8 +3540,8 @@ ZSTDLIB_API ZSTD_CDict* ZSTD_createCDict_advanced2(const void* dict, size_t dict
         ZSTD_dedicatedDictSearch_isSupported(
             cctxParams->compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
     if (!dedicatedDictSearch) {
-        ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(
-            cctxParams->compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
+        ZSTD_compressionParameters cParams = ZSTD_getCParamsFromCCtxParams(
+            cctxParams, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
         return ZSTD_createCDict_advanced(dict, dictSize,
             dictLoadMethod, dictContentType, cParams,
             customMem);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -869,9 +869,7 @@ static ZSTD_compressionParameters ZSTD_dedicatedDictSearch_getCParams(
         unsigned long long srcSizeHint,
         size_t const dictSize);
 static int ZSTD_dedicatedDictSearch_isSupported(
-        int const compressionLevel,
-        unsigned long long srcSizeHint,
-        size_t const dictSize);
+        const ZSTD_compressionParameters* cParams);
 static void ZSTD_dedicatedDictSearch_revertCParams(
         ZSTD_compressionParameters* cParams);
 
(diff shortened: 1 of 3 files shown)


terrelln:
Can you explain what the bug is? It isn't obvious to me why non-dictionary compression was effected, or how this fixes it.

lib/compress/zstd_compress.c
```diff
@@ -3498,63 +3503,63 @@ static ZSTD_CDict* ZSTD_createCDict_advanced_internal(size_t dictSize,
...
-    if (!dedicatedDictSearch) {
-        ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(
-            cctxParams->compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
-        return ZSTD_createCDict_advanced(dict, dictSize,
-            dictLoadMethod, dictContentType, cParams,
-            customMem);
+    if (ZSTD_isError( ZSTD_initCDict_internal(cdict,
+                                    dict, dictSize,
+                                    dictLoadMethod, dictContentType,
+                                    cParams, cctxParams) )) {
```
terrelln, line 3557: This probably shouldn't take both `cParams` and `cctxParams`. It be less ambiguous if we set only passed `cctxParams`, and set `cctxParams.cParams = cParams`.

lib/compress/zstd_compress.c
```diff
@@ -3498,63 +3503,63 @@ static ZSTD_CDict* ZSTD_createCDict_advanced_internal(size_t dictSize,
...
+        ZSTD_overrideCParams(&cParams, &cctxParams.cParams);
+    } else {
+        cParams = ZSTD_getCParamsFromCCtxParams(
+            &cctxParams, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
+    }
 
-        return cdict;
+    if (!ZSTD_dedicatedDictSearch_isSupported(&cParams)) {
+        /* Fall back to non-DDSS params */
+        cctxParams.enableDedicatedDictSearch = 0;
```
terrelln, line 3545: Is this the fix that changes non-dictionary compression? Since otherwise we would be calling `ZSTD_dedicatedDictSearch_revertCParams()` when the DDS wasn't in fact used, which would reduce our hash log by 2 in the working context?

felixhandte:
@terrelln, the root of that bug was that `ZSTD_createCDict_advanced2()`, when not in DDSS mode, was using `ZSTD_getCParams_internal(cctxParams->compressionLevel, ...)` to derive the cParams. This ignored any configuration of cParams in the cctxParams, looking only at the compression level. This PR switches it to use `ZSTD_getCParamsFromCCtxParams(&cctxParams, ...)`, which looks at both the cLevel and cParams. Apparently in the old API entry point a compression level gets resolved to cParams before being written into the cctxParams somewhere.

lib/compress/zstd_compress.c
felixhandte: Nah.

terrelln:
Ah I see, I was mis-remembering that there was a large impact for non-dictionary compression, but only dictionary compression regressed negligibly.

terrelln (approved):

PR #2321, opened 2020-09-22
Update documentation about repcodes in dictionaries

senhuang42:
There was a request to disallow repcodes == 0 in dictionaries. However, it seems to already be implemented (see: `zstd_compress.c: line 3048`).

But the compression format doc was not updated. Also, according to the code, in `zstd_compress.c: line 3049`, `repcode == dictContentSize` is allowed, so the documentation for that is accordingly adjusted.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -1440,7 +1440,7 @@ __`Entropy_Tables`__ : follow the same format as tables in [compressed blocks].
               Repeat distribution mode for sequence decoding.
               It's finally followed by 3 offset values, populating recent offsets (instead of using `{1,4,8}`),
               stored in order, 4-bytes __little-endian__ each, for a total of 12 bytes.
-              Each recent offset must have a value <= dictionary size, and cannot equal 0.
+              Each recent offset must have a value <= dictionary content size, and cannot equal 0.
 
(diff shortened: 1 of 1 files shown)


doc/zstd_compression_format.md
```diff
@@ -1440,7 +1440,7 @@ __`Entropy_Tables`__ : follow the same format as tables in [compressed blocks].
               Repeat distribution mode for sequence decoding.
               It's finally followed by 3 offset values, populating recent offsets (instead of using `{1,4,8}`),
               stored in order, 4-bytes __little-endian__ each, for a total of 12 bytes.
-              Each recent offset must have a value < dictionary size.
+              Each recent offset must have a value <= dictionary size, and cannot equal 0.
```
Cyan4973, line 1443: Since we are into this level of detail,
the limit is actually the dictionary _content_ size.

Cyan4973 (approved):

Cyan4973:
Thanks @senhuang42 !

FYI, any time we edit the specification in a way which is more a typo fix, and impacts / clarifies some meaning,
we also update the version number : 
https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#version
https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#version-changes

PR #2322, opened 2020-09-22
Don't let warning messages consume input from stdin

senhuang42:
`getchar()`-dependent warning prompts may swallow characters from stdin. We don't want that if stdin is our input.
The following behavior is currently:
```
$ cat <<EOF >test
> y
> Hello, world!
> EOF
$ ./zstd test
test                 :181.25%   (    16 =>     29 bytes, test.zst)             
$ ./zstd -q -d test.zst -c           // compresses fine when directly processing file
y
Hello, world!
$ ./zstd < test -o test.zst.         // 'y' in the first line is consumed when input is fed thru stdin
/*stdin*\            :192.86%   (    14 =>     27 bytes, test.zst)             
$ ./zstd -q -d test.zst -c
Hello, world!
```

Unit tests included, and checked that they fail on current `dev` branch, but work with the patch.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -91,8 +91,10 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
                                  const char* acceptableLetters, int hasStdinInput) {
     int ch, result;
 
-    if (hasStdinInput)
+    if (hasStdinInput) {
+        UTIL_DISPLAY("Stdin is an input - not proceeding.\n");
         return 1;
+    }
 
     UTIL_DISPLAY("%s", prompt);
     ch = getchar();
--- a/programs/util.c
+++ b/programs/util.c
@@ -92,7 +92,7 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
     int ch, result;
 
     if (hasStdinInput) {
-        UTIL_DISPLAY("Stdin is an input - not proceeding.\n");
+        UTIL_DISPLAY("stdin is an input - not proceeding.\n");
         return 1;
     }
 
--- a/programs/util.c
+++ b/programs/util.c
@@ -90,9 +90,6 @@ int g_utilDisplayLevel;
 int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
                                           const char* acceptableLetters) {
     int ch, result;
-    /* If input is presented via stdin, dont use prompt as it may swallow characters */
-    if (!IS_CONSOLE(stdin))
-        return 0;
 
     UTIL_DISPLAY("%s", prompt);
     ch = getchar();
--- a/programs/util.c
+++ b/programs/util.c
@@ -90,6 +90,10 @@ int g_utilDisplayLevel;
 int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
                                           const char* acceptableLetters) {
     int ch, result;
+    /* If input is presented via stdin, dont use prompt as it may swallow characters */
+    if (!IS_CONSOLE(stdin))
+        return 0;
+
     UTIL_DISPLAY("%s", prompt);
     ch = getchar();
     result = 0;
--- a/programs/util.c
+++ b/programs/util.c
@@ -88,8 +88,11 @@ UTIL_STATIC void* UTIL_realloc(void *ptr, size_t size)
 int g_utilDisplayLevel;
 
 int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
-                                          const char* acceptableLetters) {
+                                 const char* acceptableLetters, int hasStdinInput) {
     int ch, result;
+
+    if (hasStdinInput)
+        return 1;
 
     UTIL_DISPLAY("%s", prompt);
     ch = getchar();
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -309,6 +309,18 @@ test -f precompressedFilterTestDir/input.5.zst.zst
 test -f precompressedFilterTestDir/input.6.zst.zst
 println "Test completed"
 
+
+
+println "\n===>  warning prompt does not swallow characters"
+println "y" > tmpPrompt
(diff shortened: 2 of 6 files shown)


tests/playTests.sh
```diff
@@ -310,6 +310,17 @@ test -f precompressedFilterTestDir/input.6.zst.zst
 println "Test completed"
 
 
+
+println "\n===>  warning prompt for decompressing multiple files does not swallow characters"
```
terrelln, line 314: nit: Is this comment accurate? This looks like it is triggering the overwrite prompt.

senhuang42: Oops yeah - the overwrite prompt uses the same warning function as multifile decompression, I'll correct that.

programs/util.c
```diff
@@ -90,6 +90,10 @@ int g_utilDisplayLevel;
 int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
                                           const char* acceptableLetters) {
     int ch, result;
+    /* If input is presented via stdin, dont use prompt as it may swallow characters */
```
felixhandte, line 93: So there's two considerations to this. One is if input to be compressed is being piped in via stdin. In which case we want to avoid consuming from stdin. But we could test that with `!strcmp(inputFilename, stdinmark)` or whatever. Whereas this also catches the scenario where we are running in a non-interactive environment, and stdin is not a tty, even though input is not being presented through stdin. I think that this is the right behavior there as well, since trying to read from stdin will fail or hang. But it's worth calling that out as well.

tests/playTests.sh
```diff
@@ -310,6 +310,17 @@ test -f precompressedFilterTestDir/input.6.zst.zst
 println "Test completed"
 
 
+
+println "\n===>  warning prompt does not swallow characters"
+println "y" > tmpPrompt
+println "hello world" >> tmpPrompt
+zstd tmpPrompt
+zstd < tmpPrompt -o tmpPrompt.zst
```
Cyan4973, line 318: My understanding of this scenario : 
- `zstd tmpPrompt` creates `tmpPrompt.zst`.
- `zstd < tmpPrompt -o tmpPrompt.zst` tries to basically overwrite `tmpPrompt.zst`
- since there is no `-f`, it can't overwrite without confirmation
- since input is `stdin` it can't ask for overwrite confirmation
- since it can't ask for overwrite confirmation, it should automatically fail

This is not what I see : it seems this test is intended to succeed, hence it would overwrite `tmpPrompt.zst`,
which, I believe, is not correct for this use case.

Thoughts ?

programs/util.c
terrelln: Thats a good point. I think that checking `inputFilename` against `stdinmark` is the better check. People expect `yes | zstd f1 f2 -o f.zst` to work, even though `-f` would be more conventional. Also, if it doesn't work, that makes it much harder to test.

programs/util.c
```diff
@@ -90,6 +90,10 @@ int g_utilDisplayLevel;
...
+    if (!IS_CONSOLE(stdin))
```
Cyan4973, line 94: Technically, this test checks if `stdin` is the console.

This is different from the comment, which states that it tests if input is presented via `stdin`.

Example : 
`zstd FILE < /dev/null` : input is _not_ `stdin`, yet `stdin` is not the console either.

I'm fine with the proposed policy : if `stdin` is not the console, then it's not possible to ask for user confirmation via the console, hence `UTIL_requireUserConfirmation()` automatically fails. Just fix the comment.

This policy works as long as we don't support "interactive input" (aka, input defined directly on console),
such as `gzip > dest.gz` : in this scenario, input is `stdin`, yet `stdin` is still the console.

As mentioned, `zstd` doesn't support interactive input, 
so `if (!IS_CONSOLE(stdin))` is fine.
Note though that if we were to revisit that design decision to allow some interactive mode in the future, it would impact this code path.
Might be worth a comment.

terrelln: There is another thread about this above.

programs/util.c
Cyan4973: > People expect `yes | zstd f1 f2 -o f.zst` to work

It's not clear to me if this is a valid scenario.
Using `gzip` as a reference : 

```
gzip FILE
FILE.gz already exists -- do you wish to overwrite (y or n)? y

yes | gzip FILE
gzip: FILE.gz already exists -- skipping
```

terrelln: I expect tools that accept yes/no input to work with `yes`. I'm surprised when they don't. When I'm using an unfamiliar API, it can be easier to just pipe in `yes` than to try to figure out which option suppresses the warning, if any.

I think it would also be hard to test that the prompt works correctly if the input must be a console, meaning all the existing test cases we have where we `echo y | zstd` wouldn't work.

senhuang42:
We now check if stdin is one of the inputs, and conveniently we can use the `fCtx` object to store this information.
Updated with the following behavior which is what I believe we want:

```
# assuming tmp.zst exists
echo 'yes' | zstd tmp -o tmp.zst # goes through
zstd < tmp -o tmp.zst # fails immediately
echo 'yes' | zstd < tmp -o tmp.zst # fails immediately
zstd tmp - < tmp -o tmp.zst --rm # fails immediately
```

There are some more issues with displayed status info mixing with stdout - that's going to be another PR.

programs/util.c
```diff
@@ -88,8 +88,14 @@ UTIL_STATIC void* UTIL_realloc(void *ptr, size_t size)
 int g_utilDisplayLevel;
 
 int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
-                                          const char* acceptableLetters) {
+                                 const char* acceptableLetters, int hasStdinInput) {
     int ch, result;
+
+    if (hasStdinInput) {
+        UTIL_DISPLAY("Stdin is an input - not proceeding.\n");
```
Cyan4973, line 95: minot nit: 
it's `stdin`, all lowercase

Cyan4973 (approved):

PR #2326, opened 2020-09-24
Fix issues and warnings exposed by Kernel Test Robot

terrelln:
* Clean up `mem.h` API by moving unrelated stuff to `compiler.h`.
* Provide custom `mem.h` in the kernel to use preferred unaligned accessors, since `__builtin_bswap*()` isn't available on all architectures.
* Move basic types from `zstd_deps.h` to `mem.h`.
* Disable ASAN/MSAN for the kernel. I had to slightly change how ASAN/MSAN is detected. And improve `freestanding.py` to handle `#if X`.
* Wrap `customMem` xor checks in parens for readability.
* Add a missing semicolon in `zstd_deps.h` in a section that doesn't get used in the kernel.
* Add `ZSTD_COMPRESS_HEAPMODE` and use it in the kernel to avoid large stack warnings.
* Reduce stack usage of `ZSTD_compressSuperblock()` by reducing the space reserved for tables to be exactly the right amount.

See report here https://lkml.org/lkml/2020/9/23/1226. A lot of the cppcheck warnings are false positives. I've opted to only silence one warning, which was wrapping the `customMem` xor checks in parens, because that improves readability.

I will merge these changes into the `staging` branch for zstd-1.4.6.


lib/common/mem.h
```diff
@@ -18,6 +18,7 @@ extern "C" {
 /*-****************************************
 *  Dependencies
 ******************************************/
+#define ZSTD_DEPS_NEED_SWAP
 #include "zstd_deps.h"  /* size_t, ptrdiff_t, ZSTD_memcpy */
```
Cyan4973, line 22: I understand the short term requirement to get this working fine for kernel and 1.4.6,

but for the sake of code clarity beyond that goal,
I admit that after all these successive changes,
it's no longer clear to me what are the intended scope of `mem.h` and `zstd_deps.h`.

Maybe the purpose of each file is worth documenting in the header of each file,
as there is a risk for them becoming some kind of "bag of everything" after multiple edition rounds.

terrelln: We could move the int types, the swaps, and maybe mem{cpy,move,set} into `mem.h`. Then move some of the extranuous stuff, like the asan/msan poisoning, out of `mem.h`. And have the kernel provide both `zstd_deps.h` and `mem.h`.

Cyan4973: Yes,
I like the idea of each file having a clean perimeter,
and this proposal makes sense.

terrelln:
I've moved the basic types to `mem.h`, and cleaned up the `mem.h` module by moving extra stuff to `compiler.h`.

contrib/freestanding_lib/freestanding.py
```diff
@@ -261,95 +265,124 @@ def _preprocess_once(self):
...
-                        is_true = defined_value != value
-                    elif cmp == '>=':
-                        is_true = defined_value >= value
-                    elif cmp == '>':
-                        is_true = defined_value > value
+                    is_int = False
+
+                resolved = is_int
+                is_true = (defined_value != 0)
+            else:
```
terrelln, line 313: This is all just indented one more.

terrelln:
Another Kernel Test Robot failure showed that ASAN/MSAN don't work, which is not unexpected, so I had to disable ASAN/MSAN support.

lib/common/zstd_internal.h
```diff
@@ -186,6 +186,10 @@ typedef enum { set_basic, set_rle, set_compressed, set_repeat } symbolEncodingTy
 #define OffFSELog   8
 #define MaxFSELog  MAX(MAX(MLFSELog, LLFSELog), OffFSELog)
 
+#define ZSTD_MAX_HUF_HEADER_SIZE 128 /* header + <= 127 byte tree description */
+/* Each table cannot take more than #symbols * FSELog bits */
+#define ZSTD_MAX_FSE_HEADERS_SIZE ((MaxML * MLFSELog + MaxLL * LLFSELog + MaxOff * OffFSELog + 7) / 8)
```
terrelln, line 191: @Cyan4973 can you double check these numbers? I'm using them to reduce the stack usage of superblock mode, to avoid >1024 byte frame warnings.

lib/common/zstd_internal.h
```diff
@@ -186,6 +186,10 @@ typedef enum { set_basic, set_rle, set_compressed, set_repeat } symbolEncodingTy
 #define OffFSELog   8
 #define MaxFSELog  MAX(MAX(MLFSELog, LLFSELog), OffFSELog)
 
+#define ZSTD_MAX_HUF_HEADER_SIZE 128 /* header + <= 127 byte tree description */
+/* Each table cannot take more than #symbols * FSELog bits */
+#define ZSTD_MAX_FSE_HEADERS_SIZE (((MaxML + 1) * MLFSELog + (MaxLL + 1) * LLFSELog + (MaxOff + 1) * OffFSELog + 7) / 8)
```
terrelln, line 191: @Cyan4973 can you double check these numbers? I'm using them to reduce the stack usage of superblock mode, to avoid >1024 byte frame warnings.

Cyan4973: Looks good to me

terrelln:
@Cyan4973 have you had a chance to take a look at this?

Cyan4973:
Yes, this looks good to me

Cyan4973 (approved):

PR #2328, opened 2020-09-25
Allow external creation of POOLs that can be shared.

marxin:
This is a follow up of #2290 which implements sharing of threads pools.
@Cyan4973: what do you think about this approach?

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -1409,6 +1409,21 @@ ZSTDLIB_API ZSTD_CDict* ZSTD_createCDict_advanced(const void* dict, size_t dictS
                                                   ZSTD_compressionParameters cParams,
                                                   ZSTD_customMem customMem);
 
+/* ! Thread pool :
+ * These prototypes make it possible to share a thread pool among multiple compression contexts.
+ * This can limit resources for applications with multiple threads where each one uses
+ * a threaded compression mode (via ZSTD_c_nbWorkers parameter).
+ * ZSTD_createThreadPool creates a new thread pool with a given number of threads.
+ * Note that the lifetime of such pool must exist while being used.
+ * ZSTD_CCtx_refThreadPool assigns a thread pool to a context (use NULL argument value
+ * to use an internal thread pool).
+ * ZSTD_freeThreadPool frees a thread pool.
+ */
+typedef struct POOL_ctx_s ZSTD_threadPool;
+ZSTDLIB_API ZSTD_threadPool* ZSTD_createThreadPool(size_t numThreads);
+ZSTDLIB_API void ZSTD_freeThreadPool (ZSTD_threadPool* pool);
+ZSTDLIB_API size_t ZSTD_CCtx_refThreadPool(ZSTD_CCtx* cctx, ZSTD_threadPool* pool);
+
 /**
  * This API is temporary and is expected to change or disappear in the future!
  */
@@ -1423,8 +1438,6 @@ ZSTDLIB_API ZSTD_DDict* ZSTD_createDDict_advanced(const void* dict, size_t dictS
                                                   ZSTD_dictLoadMethod_e dictLoadMethod,
                                                   ZSTD_dictContentType_e dictContentType,
                                                   ZSTD_customMem customMem);
-
-
 
 /***************************************
 *  Advanced compression functions
--- a/examples/streaming_compression_thread_pool.c
+++ b/examples/streaming_compression_thread_pool.c
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) 2020, Martin Liska, SUSE, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
+
+#include <stdio.h>     // printf
+#include <stdlib.h>    // free
+#include <string.h>    // memset, strcat, strlen
+#include <zstd.h>      // presumes zstd library is installed
+#include "common.h"    // Helper functions, CHECK(), and CHECK_ZSTD()
+#include <pthread.h>
+
+typedef struct compress_args
+{
+  const char *fname;
+  char *outName;
+  int cLevel;
+#if defined(ZSTD_STATIC_LINKING_ONLY)
+  ZSTD_threadPool *pool;
+#endif
+} compress_args_t;
+
+static void *compressFile_orDie(void *data)
+{
+    compress_args_t *args = (compress_args_t *)data;
+    fprintf (stderr, "Starting compression of %s with level %d\n", args->fname, args->cLevel);
+    /* Open the input and output files. */
+    FILE* const fin  = fopen_orDie(args->fname, "rb");
+    FILE* const fout = fopen_orDie(args->outName, "wb");
+    /* Create the input and output buffers.
+     * They may be any size, but we recommend using these functions to size them.
+     * Performance will only suffer significantly for very tiny buffers.
+     */
+    size_t const buffInSize = ZSTD_CStreamInSize();
+    void*  const buffIn  = malloc_orDie(buffInSize);
+    size_t const buffOutSize = ZSTD_CStreamOutSize();
+    void*  const buffOut = malloc_orDie(buffOutSize);
+
+    /* Create the context. */
+    ZSTD_CCtx* const cctx = ZSTD_createCCtx();
+    CHECK(cctx != NULL, "ZSTD_createCCtx() failed!");
+
+#if defined(ZSTD_STATIC_LINKING_ONLY)
+    size_t r = ZSTD_CCtx_refThreadPool(cctx, args->pool);
+    CHECK(r == 0, "ZSTD_CCtx_refThreadPool failed!");
+#endif
+
+    /* Set any parameters you want.
+     * Here we set the compression level, and enable the checksum.
+     */
+    CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel, args->cLevel) );
+    CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 1) );
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 16);
+
+    /* This loop read from the input file, compresses that entire chunk,
+     * and writes all output produced to the output file.
+     */
+    size_t const toRead = buffInSize;
+    for (;;) {
+        size_t read = fread_orDie(buffIn, toRead, fin);
+        /* Select the flush mode.
+         * If the read may not be finished (read == toRead) we use
+         * ZSTD_e_continue. If this is the last chunk, we use ZSTD_e_end.
+         * Zstd optimizes the case where the first flush mode is ZSTD_e_end,
+         * since it knows it is compressing the entire source in one pass.
+         */
+        int const lastChunk = (read < toRead);
+        ZSTD_EndDirective const mode = lastChunk ? ZSTD_e_end : ZSTD_e_continue;
+        /* Set the input buffer to what we just read.
+         * We compress until the input buffer is empty, each time flushing the
+         * output.
+         */
+        ZSTD_inBuffer input = { buffIn, read, 0 };
+        int finished;
+        do {
+            /* Compress into the output buffer and write all of the output to
+             * the file so we can reuse the buffer next iteration.
+             */
+            ZSTD_outBuffer output = { buffOut, buffOutSize, 0 };
+            size_t const remaining = ZSTD_compressStream2(cctx, &output , &input, mode);
+            CHECK_ZSTD(remaining);
+            fwrite_orDie(buffOut, output.pos, fout);
+            /* If we're on the last chunk we're finished when zstd returns 0,
+             * which means its consumed all the input AND finished the frame.
+             * Otherwise, we're finished when we've consumed all the input.
+             */
+            finished = lastChunk ? (remaining == 0) : (input.pos == input.size);
+        } while (!finished);
+        CHECK(input.pos == input.size,
+              "Impossible: zstd only returns 0 when the input is completely consumed!");
+
+        if (lastChunk) {
+            break;
+        }
+    }
+
+    fprintf (stderr, "Finishing compression of %s\n", args->outName);
+
+    ZSTD_freeCCtx(cctx);
+    fclose_orDie(fout);
+    fclose_orDie(fin);
+    free(buffIn);
+    free(buffOut);
+    free(args->outName);
+
+    return NULL;
+}
+
+
+static char* createOutFilename_orDie(const char* filename)
+{
+    size_t const inL = strlen(filename);
+    size_t const outL = inL + 5;
+    void* const outSpace = malloc_orDie(outL);
+    memset(outSpace, 0, outL);
+    strcat(outSpace, filename);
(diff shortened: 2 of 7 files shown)


Cyan4973:
I believe it goes in the right direction @marxin,
and follows pretty closely the recipe suggested in #2290 .
It's also a pretty short implementation, which is a very good sign.

Now for a few comments on the currently proposed implementation : 
- Naming :  in the context of a public API, `createPool` feels not specific enough. Pool of what ? Prefer `createThreadPool`.
- Stronger type system : a `void*` type is too generic, it can be easily confused with other types. Make it an [incomplete type](https://www.refreshnotes.com/2016/07/c-basics-incomplete-type-in-c.html), which is way stronger in term of compiler guarantee, following established convention in this project, like `ZSTD_CCtx*` and `ZSTD_CDict*`. Call it `ZSTD_threadPool*`, so that it's in sync with `createThreadPool`.
- Avoid modifying existing `create` functions by adding a parameter, it's a slippery slope. For this case, adding an explicit parameter accessor feels better and follows existing practice. For example, we already have [`ZSTD_CCtx_refCDict()`](https://github.com/facebook/zstd/blob/v1.4.5/lib/zstd.h#L923), which, as the name implies, references an external `ZSTD_CDict*` to be shared by the target `CCtx`. The equivalent here could be `ZSTD_CCtx_refThreadPool()`.
- Now the most important point : ensure it actually works in a multi-threaded scenarios, with multiple contexts sharing the same `threadPool`. The `threadPool` was never tested for this scenario, since each `CCtx` currently owns its own `threadPool`. This _may_ result in more changes in `pool.c`, like adding mutexes. There is no guarantee that such changes are needed, maybe it works "as is". @terrelln might now more about this. Ultimately, nothing can replace a real test with multiple active contexts sharing the same `threadPool` in parallel, and preferably backed by a [thread sanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html).

marxin:
All right, I updated the pull request based on the comments and I've just tested compression of 4 parallel jobs with `examples/streaming_compression_thread_pool.c`:

```
gcc -g examples/streaming_compression_thread_pool.c -L ./lib/ -lzstd -DZSTD_MULTITHREAD -I lib -pthread -DZSTD_STATIC_LINKING_ONLY=1 && LD_LIBRARY_PATH=lib time  ./a.out /dev/shm/1.img /dev/shm/2.img /dev/shm/3.img /dev/shm/4.img 
Using shared thread pool
Starting compression of /dev/shm/1.img
Starting compression of /dev/shm/2.img
Starting compression of /dev/shm/3.img
Starting compression of /dev/shm/4.img
Finishing compression of /dev/shm/3.img.zst
Finishing compression of /dev/shm/1.img.zst
Finishing compression of /dev/shm/4.img.zst
Finishing compression of /dev/shm/2.img.zst
510.44user 2.31system 1:41.98elapsed 502%CPU (0avgtext+0avgdata 1711516maxresident)k
0inputs+0outputs (0major+14383minor)pagefaults 0swaps
```

Moreover, I tested the same example with TSAN for a smaller input and it works fine.
I'm planning to improve documentation and is there anything else I should fix?

lib/zstd.h
```diff
@@ -1424,7 +1424,10 @@ ZSTDLIB_API ZSTD_DDict* ZSTD_createDDict_advanced(const void* dict, size_t dictS
                                                   ZSTD_dictContentType_e dictContentType,
                                                   ZSTD_customMem customMem);
 
-
+typedef struct POOL_ctx_s ZSTD_threadPool;
+ZSTDLIB_API ZSTD_threadPool* ZSTD_createThreadPool(size_t numThreads);
+ZSTDLIB_API void ZSTD_freePool (ZSTD_threadPool* pool);
```
Cyan4973, line 1429: `createThreadPool` => `freeThreadPool`

lib/zstd.h
```diff
@@ -1424,7 +1424,10 @@ ZSTDLIB_API ZSTD_DDict* ZSTD_createDDict_advanced(const void* dict, size_t dictS
...
+ZSTDLIB_API size_t ZSTD_CCtx_refThreadPool(ZSTD_CCtx* cctx, ZSTD_threadPool* pool);
```
Cyan4973, line 1430: This is important to document.
Notably, mention that `pool == NULL` is a way to remove reference to any external threadPool.
Also, mention reliance on `pool`'s lifetime, which must always exist while being used.

Cyan4973:
This looks pretty good !

Please consider testing the feature : 
- with large files (> 30 MB). Small enough files _may_ trigger a special single-thread shortcut, as multi-threading only makes sense for large enough content which can be split across multiple jobs.
- with uneven threadPool / file workload (for example, 4 threads for a 6 concurrent file compression).
- with `tsan` (as you already did)

If it can survive that, then this implementation is in good shape.
The existing pool code may already be thread-safe after all !

marxin:
> If it can survive that, then this implementation is in good shape.
> The existing pool code may already be thread-safe after all !

All right, I've made a proper fuzzing of the functionality, see the following gist:
https://gist.github.com/marxin/3d72b3f34bbad01973ee452cc58190f7

I generated random files and used a random number of threads in a pool (with random compression level).
Then I decompressed each file and verified it's content.

A documentation entry is added in the rebased version. I'm removing the WIP marker.

marxin:
Thank you.

I guess you should accept code from external contributors ;)

marxin:
Few more questions I have:
- When do you plan to come up with the next release of `zstd`?
- How long does it take a staging feature (`ZSTD_STATIC_LINKING_ONLY`) lands to shared library build?

Cyan4973:
> When do you plan to come up with the next release of zstd?

`v1.4.6` feature set is locked,
so this feature will be part of next release, `v1.4.7`.

_Probably_ around Christmas

> How long does it take a staging feature (ZSTD_STATIC_LINKING_ONLY) lands to shared library build?

This needs a larger version change, such as `v1.5.0`.
Could be the next one after `v1.4.7`.
Potentially ~June 2021.

These are just ballpark guesstimates. Nothing is sure at this stage. Only `v1.4.6` seems more or less settled.

Cyan4973:
All right,
the proposed implementation looks good for a merge !

The last remaining issue is some licensing stuff.
The OSS policy is to offer a consistent license experience for users of the repository,
to reduce risks of unwanted surprises such as some "poison pill" contribution.
This is in essence what the CLA protects our users from.
This process enforces the presence of a selected license text and copyright header at the top of each file.
Currently, `examples/streaming_compression_thread_pool.c` has a small issue with this rule,
though it's almost fine : the license is fine, and the only missing part is having `Facebook, Inc` as part of the copyright header.
I think it's fine to keep your copyright notice too, it's just that `Facebook, Inc` must become part of this list.
To be more complete, there can be exceptions, but that's a heavier process, and it felt overkill to trigger it for an example file.

marxin:
> To be more complete, there can be exceptions, but that's a heavier process, and it felt overkill to trigger it for an example file.

Sure. Moreover, I took `examples/streaming_compression.c` as a base for the example. So it perfectly makes sense to have `Facebook, Inc.` as part of the Copyright header.

Cyan4973 (approved):

Cyan4973:
Thanks @marxin for this great contribution !

marxin:
> Thanks @marxin for this great contribution !

Thanks! I would like to thank you for the nice cooperation.

marxin:
@Cyan4973 One last missing bit can be an adjustment of the HTML documentation page (`zstd_manual.html`).
Is the file generated from `zstd.h` or should I make a separate pull request?

marxin:
> @Cyan4973 One last missing bit can be an adjustment of the HTML documentation page (`zstd_manual.html`). Is the file generated from `zstd.h` or should I make a separate pull request?

@Cyan4973 Can you please reply to this?

One more question I have: would it be possible marking the API as non-experimental for the next release? What do you think?

Cyan4973:
>  the HTML documentation page (`zstd_manual.html`). Is the file generated from `zstd.h` (...) ?

Yes, `zstd_manual.html` is generated from `zstd.h`.
It's generated with `make manual`.

> would it be possible marking the API as non-experimental for the next release?

The next "slot" to formally accept new APIs into stable will be `v1.6.0`.
There is no date set for `v1.6.0`, this can depend on priorities.
But that's not all, before that point, we would like to gather enough experience / proofs that this APIs works well in real-world scenarios.
Soonest it _may_ happen could be next year.

marxin:
> Yes, `zstd_manual.html` is generated from `zstd.h`. It's generated with `make manual`.
> 
> >

Oh, I see and it's fixed in #2822. Who runs the `make manual`? Am I supposed to do it in the pull request?

marxin:
> The next "slot" to formally accept new APIs into stable will be `v1.6.0`. There is no date set for `v1.6.0`, this can depend on priorities. But that's not all, before that point, we would like to gather enough experience / proofs that this APIs works well in real-world scenarios. Soonest it _may_ happen could be next year.

All right. So we'll have hopefully an experience from build openSUSE packages with `rpm` that leverages the thread pool.

Cyan4973:
>  Who runs the `make manual`? Am I supposed to do it in the pull request?

You certainly can, and it's not compulsory.
Sooner or later, someone will run `make all`, which includes `make manual`, and this will update the manual. This stage is guaranteed to happen before release, so the manual will be updated at one point or another.

marxin:
All right, I run `make manual` and included that in the pull request.

marxin:
After a debugging session related to `rpm` where we utilize thread pool, a noticed 2 quite significant limitations:

- when there are much more threads than the size of a thread pool, the other threads busy loop in `native_queued_spin_lock_slowpath` and other CPUs are utilized :/

- having a machine with 128 CPUs and 256GB of memory, using 128 threads, each using parallel compression with 128 threads and having pool_size == 128, the memory allocation here:
https://github.com/facebook/zstd/blob/dev/lib/compress/zstdmt_compress.c#L1296-L1302

allocates in my case ~4GB per thread (of which I have 128). Thus I see memory exhaustion.

Cyan4973:
The thread pool should sleep, rather than busy-loop, the extraneous threads beyond pool capacity.
So there's definitely something to investigate here.

Regarding memory exhaustion, we don't have a good dynamic mechanism here. At least not yet.
A suggestion could be to reduce the pool size, 
if you know your workload requires ~4GB per thread (which is definitely a lot),
maybe limit to something like ~50 threads,
this way, up to 50x4  = 200 GB will be allocated, but no more, leaving room for other tasks to run on the same system.

terrelln:
@marxin do you have a stack for the busy loop? We can probably debug without it, since it seems pretty easy to be reproduce. But, it'd definitely help speed up the resolution.

marxin:
> @marxin do you have a stack for the busy loop? We can probably debug without it, since it seems pretty easy to be reproduce. But, it'd definitely help speed up the resolution.

Sure, try this:

```
$ gcc pool.c -lzstd -O2 -DZSTD_STATIC_LINKING_ONLY
$ head -c 20M </dev/urandom > /tmp/20M
$ ./a.out
wrong arguments
usage:
./a.out POOL_SIZE THREADS LEVEL FILES
$ ./a.out 1 16 19 /tmp/20M
Using shared thread pool of size 1
Starting compression of /tmp/20M with level 19, using 16 threads
Starting compression of /tmp/20M with level 19, using 16 threads
...
```

[pool.c.txt](https://github.com/facebook/zstd/files/8386540/pool.c.txt)

marxin:
@terrelln Are you able to reproduce the busy loop?

terrelln:
Yes, it looks like this is what is happening:

In `ZSTDMT_compressStream_generic()` we call `ZSTDMT_createCompressionJob()` here:

https://github.com/facebook/zstd/blob/460780f804c12363842763b5a8afa5adc282ef29/lib/compress/zstdmt_compress.c#L1850

In `ZSTDMT_createCompressionJob()`, `POOL_tryAdd()` fails (returns 0) because the pool is full:

https://github.com/facebook/zstd/blob/460780f804c12363842763b5a8afa5adc282ef29/lib/compress/zstdmt_compress.c#L1424

We call `ZSTDMT_flushProduced()` with `blockToFlush = 1`, which is supposed to block if we can't make any forward progress:

https://github.com/facebook/zstd/blob/460780f804c12363842763b5a8afa5adc282ef29/lib/compress/zstdmt_compress.c#L1854

But in `ZSTDMT_flushProduced()` we see that `mtctx->doneJobID == mtctx->nextJobID`, so we don't block.

https://github.com/facebook/zstd/blob/460780f804c12363842763b5a8afa5adc282ef29/lib/compress/zstdmt_compress.c#L1445

Then in the caller's loop, we end up calling`ZSTDMT_compressStream_generic()` again, and repeating the same steps. Since we aren't blocking anywhere, its a busy loop.

---

This is happening because other threads have already scheduled their work on the pool, so we aren't waiting for a local job to complete, we're waiting for an external job to complete.

We switched from `POOL_add()` to `POOL_tryAdd()` to allow buffers to be flushed when adding a job. But, in the case that there is no outstanding work, we won't block in `ZSTDMT_flushProduced()`, so we need to call `POOL_add()` instead.

terrelln:
This diff should fix it, I'll put up a PR soon:

```diff
1	
diff --git i/lib/compress/zstdmt_compress.c w/lib/compress/zstdmt_compress.c
index 4ac2249b..3e1d3f77 100644
--- i/lib/compress/zstdmt_compress.c
+++ w/lib/compress/zstdmt_compress.c
@@ -1409,26 +1409,39 @@ static size_t ZSTDMT_createCompressionJob(ZSTDMT_CCtx* mtctx, size_t srcSize, ZS
             return 0;
         }
     }
 
     DEBUGLOG(5, "ZSTDMT_createCompressionJob: posting job %u : %u bytes  (end:%u, jobNb == %u (mod:%u))",
                 mtctx->nextJobID,
                 (U32)mtctx->jobs[jobID].src.size,
                 mtctx->jobs[jobID].lastJob,
                 mtctx->nextJobID,
                 jobID);
-    if (POOL_tryAdd(mtctx->factory, ZSTDMT_compressionJob, &mtctx->jobs[jobID])) {
+    if (mtctx->doneJobID == mtctx->nextJobID) {
+        /* No outsanding jobs ==> call POOL_add so we block if the pool is full.
+         * This can happen if we are using an external thread pool, and all the
+         * workers are busy on other work.
+         * If we don't block here, then we also won't block in ZSTDMT_flushProduced(),
+         * since there is no outstanding work, and we will busy loop.
+         */
+        DEBUGLOG(5, "ZSTDMT_createCompressionJob: no outstanding jobs - potentially blocking for %u", mtctx->nextJobID);
+        POOL_add(mtctx->factory, ZSTDMT_compressionJob, &mtctx->jobs[jobID]);
         mtctx->nextJobID++;
         mtctx->jobReady = 0;
     } else {
-        DEBUGLOG(5, "ZSTDMT_createCompressionJob: no worker available for job %u", mtctx->nextJobID);
-        mtctx->jobReady = 1;
+        if (POOL_tryAdd(mtctx->factory, ZSTDMT_compressionJob, &mtctx->jobs[jobID])) {
+            mtctx->nextJobID++;
+            mtctx->jobReady = 0;
+        } else {
+            DEBUGLOG(5, "ZSTDMT_createCompressionJob: no worker available for job %u", mtctx->nextJobID);
+            mtctx->jobReady = 1;
+        }
     }
     return 0;
 }
 
 
 /*! ZSTDMT_flushProduced() :
  *  flush whatever data has been produced but not yet flushed in current job.
  *  move to next job if current one is fully flushed.
  * `output` : `pos` will be updated with amount of data flushed .
  * `blockToFlush` : if >0, the function will block and wait if there is no data available to flush .
```

marxin:
> This diff should fix it, I'll put up a PR soon:

Btw. have you made a PR with this fix? Thanks.

marxin:
> 

May I please ping this?

PR #2329, opened 2020-09-25
Prevent summary updates when using stdout

senhuang42:
Note: this will have merge conflicts with #2322 (and recently merged #2310), so once #2322 is merged, I'll fix the merge conflicts in this PR.

Currently we have some issues where intermediate status summaries are interspersed with compressed or decompressed data output to `stdout`.

This PR changes the behavior so that no matter the verbosity, there will never be anything displayed/printed in the middle of a compressed or decompressed output in `stdout`. This is particularly important for decompression, if the user wants to spot check decompressed output in the terminal. Though displayed compressed output in the terminal isn't really comprehensible anyways, it's good to be consistent here too.

However, at higher verbosities, a summary may be printed afterwards, along with some other informational text before the (de)compressed output. But never in the middle. The behavior is now as follows:

image

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -331,6 +331,7 @@ struct FIO_ctx_s {
     /* file i/o info */
     int nbFilesTotal;
     int hasStdinInput;
+    int hasStdoutOutput;
 
     /* file i/o state */
     int currFileIdx;
@@ -388,6 +389,7 @@ FIO_ctx_t* FIO_createContext(void)
 
     ret->currFileIdx = 0;
     ret->hasStdinInput = 0;
+    ret->hasStdoutOutput = 0;
     ret->nbFilesTotal = 1;
     ret->nbFilesProcessed = 0;
     ret->totalBytesInput = 0;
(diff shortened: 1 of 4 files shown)


Cyan4973:
Fixed output looks good !

senhuang42:
So one consideration: `echo 'yes' | zstd tmp1 tmp2 -c --rm`

Will automatically abort with the new behavior (using `-f` would still go through, and `-v` would trigger the warning prompt).
Previously, it would provide a prompt by default. After some thought, I think this new behavior is probably what we actually want. 
Essentially, the behavior have specified that we want is:

- If the output is stdout, by default we should output/print nothing other than the (de)compressed output itself
- If the output is stdout, and -v is specified, we are allowed to print extra info _around_ the (de)compressed output itself

So based on these criteria, we should only provide the multifile warning prompt with `-v`, and otherwise go to the default action, which in this case, is bailing out. As such, the unit test that was failing is no longer representative of what we want, so I've changed it to agree with our new behavior.

Cyan4973:
I agree.

I actually would even suggest that, even with `-v`, there should be no prompt, and it would just fail immediately (though it's important to produce a meaningful error message for this case).

That being said, I'm concerned that this last suggestion could increase code complexity, notably if the decision to present the prompt or not is based on verbosity level. In which case, consider it optional.

PR #2332, opened 2020-09-29
Version 3 ldm opt

senhuang42:
Draft PR
Preliminary results regarding compressed size:

5259889920 (5GB) bytes: 5 Linux kernels concatenated: versions 4.19.148, 5.4.68, 5.8.12, 5.9-rc7, 5.8.12:
Old: Level 22 - long mode - window log = 27
622643668 → 11.8376%
Old: Level 22 - no long mode - window log = 27
624737369 → 11.8774%
New: Level 22 - long mode - window log = 27
622639116 → 11.8375%
211950592 (200MB) bytes: silesia.tar
Old: Level 22 - long mode - window log = 27
52777588 → 24.900%
Old: Level 22 - no long mode - window log = 27
52701591 → 24.865%
New: Level 22 - long mode - window log = 27
52699843 → 24.864%
New algorithm approach:

Essentially, we will use the LDM rawSeqStore given to us in a similar way to the current LDM insertion algorithm works. What this means is that essentially for each new block, the LDM sequence at position pos within the LDM seqStore is guaranteed to start at the same relative position as the block. That is, for each n bytes processed in the block, n bytes are processed and consumed in the LDM seqstore, and the member pos will point to relevant sequence.
As such, all of our calculations are purely relative within each block, and the approach is robust against issues such as window updates, etc.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2320,7 +2320,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         if (curr > ms->nextToUpdate + 384)
             ms->nextToUpdate = curr - MIN(192, (U32)(curr - ms->nextToUpdate - 384));
     }
-
+    printf("--NEW BLOCK--\n");
     /* select and store sequences */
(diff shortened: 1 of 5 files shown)


senhuang42:
Ignore - meant to PR to my fork.

PR #2334, opened 2020-09-30
[seekable_format][reproducible] ZSTD_seekable_decompress() example that hangs

mdittmer:
This change illustrates a bug in which using the seekable format library as intended hangs. Granted, the input is contrived (not a valid ZSTD archive) but the expected behaviour is an error, not a hang.


Cyan4973:
Thanks for the clear reproduction case @mdittmer .
This issue is being fixed by @senhuang42 , in #2408.

mdittmer:
Glad it's getting fixed! Thanks for your work. I'll let @senhuang42 or @Cyan4973 close this once the fix has landed.

PR #2337, opened 2020-10-01
Fix ZSTD_adjustCParams_internal() to handle dictionary logic

terrelln:
This PR is divided into several commits which are split logically. See each commit message for details. Summary included below.
* [bug] Fix dictContentType when reprocessing cdict - The cdict could get loaded in `auto` mode when it was supposed to be raw content.
* [refactor] Remove ZSTD_resetCStream_internal() - It wasn't a logical division of responsibilities and was hiding a bug.
* [minor improvement] Pass dictSize when selecting parameters - In `ZSTD_compressStream2()` we weren't passing the dictionary size when selecting compression parameters.
*  [minor improvement] Pass 0 as the content size in the DDS.
* Fix ZSTD_adjustCParams_internal() to handle dictionary logic  - Pass the compression parameter mode into `ZSTD_{get,adjust}CParams_internal()` and avoid downsizing the hashLog and chainLog too much.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3793,34 +3793,6 @@ size_t ZSTD_CStreamOutSize(void)
     return ZSTD_compressBound(ZSTD_BLOCKSIZE_MAX) + ZSTD_blockHeaderSize + 4 /* 32-bits hash */ ;
 }
 
-static size_t ZSTD_resetCStream_internal(ZSTD_CStream* cctx,
-                    const void* const dict, size_t const dictSize, ZSTD_dictContentType_e const dictContentType,
-                    const ZSTD_CDict* const cdict,
-                    ZSTD_CCtx_params params, unsigned long long const pledgedSrcSize)
-{
-    DEBUGLOG(4, "ZSTD_resetCStream_internal");
-    /* Finalize the compression parameters */
-    params.cParams = ZSTD_getCParamsFromCCtxParams(&params, pledgedSrcSize, dictSize);
-    /* params are supposed to be fully validated at this point */
-    assert(!ZSTD_isError(ZSTD_checkCParams(params.cParams)));
-    assert(!((dict) && (cdict)));  /* either dict or cdict, not both */
-
-    FORWARD_IF_ERROR( ZSTD_compressBegin_internal(cctx,
-                                         dict, dictSize, dictContentType, ZSTD_dtlm_fast,
-                                         cdict,
-                                         &params, pledgedSrcSize,
-                                         ZSTDb_buffered) , "");
-
-    cctx->inToCompress = 0;
-    cctx->inBuffPos = 0;
-    cctx->inBuffTarget = cctx->blockSize
-                      + (cctx->blockSize == pledgedSrcSize);   /* for small input: avoid automatic flush on reaching end of block, since it would require to add a 3-bytes null block to end frame */
-    cctx->outBuffContentSize = cctx->outBuffFlushedSize = 0;
-    cctx->streamStage = zcss_load;
-    cctx->frameEnded = 0;
-    return 0;   /* ready to go */
-}
-
 /* ZSTD_resetCStream():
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1055,6 +1055,38 @@ U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
     return hashLog - btScale;
 }
 
+/** ZSTD_adjustCParams_internal() :
```
Cyan4973, line 1058: nit: wrong title (`ZSTD_adjustCParams_internal` => `ZSTD_dictAndWindowLog`)

Cyan4973:
We probably need to discuss the relation of this PR with #2342

terrelln:
> We probably need to discuss the relation of this PR with #2342

@felixhandte and I have discussed a bit. I think basically we want this patch when not attaching the dictionary, and don't want it when attaching the dictionary. And when attaching a dictionary, we probably also want to do other things, like not take the `dictSize` into account when shrinking the `windowLog`. @felixhandte is going to put up a PR to handle it. Would it make sense to do it on top of this PR?

Cyan4973:
I agree with these scenarios.
Perhaps the remaining question is, are they really 2 separate topics ? Or should all these dictionary scenarios be taken care of in the same PR ?

terrelln:
So this ended up being non-trivial, and it exposed several minor inefficiencies and bugs in dictionary compression.

lib/compress/zstd_compress.c
```diff
@@ -1055,24 +1055,71 @@ U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
...
+static U32 ZSTD_dictAndWindowLog(U32 windowLog, U64 srcSize, U64 dictSize)
+{
+    const U64 maxWindowSize = 1ULL << ZSTD_WINDOWLOG_MAX;
+    /* No dictionary ==> No change */
+    if (dictSize == 0) {
+        return windowLog;
+    }
+    assert(srcSize != ZSTD_CONTENTSIZE_UNKNOWN); /* Handled in ZSTD_adjustCParams_internal() */
+    {
+        U64 const windowSize = 1ULL << windowLog;
```
Cyan4973, line 1074: minor : maybe `assert(windowLog <= ZSTD_WINDOWLOG_MAX)` to avoid shift overflow ?

lib/compress/zstd_compress.c
```diff
@@ -1055,24 +1055,71 @@ U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat)
...
+ *  the hashLog and windowLog.
+ */
+static U32 ZSTD_dictAndWindowLog(U32 windowLog, U64 srcSize, U64 dictSize)
+{
+    const U64 maxWindowSize = 1ULL << ZSTD_WINDOWLOG_MAX;
+    /* No dictionary ==> No change */
+    if (dictSize == 0) {
+        return windowLog;
+    }
+    assert(srcSize != ZSTD_CONTENTSIZE_UNKNOWN); /* Handled in ZSTD_adjustCParams_internal() */
```
Cyan4973, line 1072: maybe this condition worth mentioning in the function description (even though it's just a few lines above)

lib/compress/zstd_compress.c
```diff
@@ -1121,18 +1169,18 @@ static void ZSTD_overrideCParams(
 }
 
 ZSTD_compressionParameters ZSTD_getCParamsFromCCtxParams(
-        const ZSTD_CCtx_params* CCtxParams, U64 srcSizeHint, size_t dictSize)
+        const ZSTD_CCtx_params* CCtxParams, U64 srcSizeHint, size_t dictSize, ZSTD_cParamMode_e mode)
```
Cyan4973, line 1172: idea for discussion :
does it make sense to consider having `ZSTD_cParamMode_e mode` as a member of `CCtxParams`, rather than a separate parameter ?

Cyan4973 (approved):

Issue #2327, opened 2020-09-25
Zstd multithreaded output can depend on number of threads


terrelln:
**Describe the bug**
As reported by @animalize in Issue #2238:

When using ZSTD_e_end end directive and output buffer size >= ZSTD_compressBound() the job number is calculated by ZSTDMT_computeNbJobs() function. This function produces a different number of jobs depending on `nbWorkers`:

https://github.com/facebook/zstd/blob/b706286adbba780006a47ef92df0ad7a785666b6/lib/compress/zstdmt_compress.c#L1243-L1255

**Expected behavior**
The output of zstd multithreaded compression must be independent of the number of threads.

**Fix**
* [ ] Make `ZSTDMT_computeNbJobs()` independent of `nbWorkers`.
* [ ] Add a fuzz test that checks that the output of multithreaded zstd is always independent of the number of threads.

**Workaround**
If you need to work around this bug, don't start your streaming job with ZSTD_e_end. Pass at least one byte of input with ZSTD_e_continue before calling ZSTD_e_end, or ensure your output buffer is `< ZSTD_compressBound(inputSize)`.

Cyan4973:
It's a shortcut to say that multithreaded `zstd` does not depend on nb of threads.

Actually, the feature supported is that _streaming_ multithreaded `zstd` does not depend on nb of threads
(and that's what is used by the `zstd` CLI).

This definition makes it possible to consider another potential fix : 
do not employ the one-pass shortcut for `ZSTD_e_end ` when `nbWorkers >= 1`,
since it's the delegation to the one-pass mode which triggers this issue.

This could be less disruptive than trying to adapt the single-pass MT compressor,
which was never designed to offer this guarantee.

Another (potentially positive) side effect is that it would guarantee that streaming multithreaded compression is _always_ non-blocking, since it would no longer delegate to the blocking mode.

animalize:
> Another (potentially positive) side effect is that it would guarantee that streaming multithreaded compression is always non-blocking, since it would no longer delegate to the blocking mode.

I once wanted to propose adding a `ZSTD_compressStream3()` function, that is always blocking in multithreaded compression.

If the caller keeps checking the non-blocking progress, it's very inconvenient.

terrelln:
> This could be less disruptive than trying to adapt the single-pass MT compressor,
> which was never designed to offer this guarantee.

Yeah, that is probably easier. I had forgotten that all the jobs in the single pass MT compressor needed to be launched at once.

[...]

PR #2339, opened 2020-10-02
Fix zstdmt stability issues and clean up the zstdmt code

terrelln:
* Remove `singleBlockingThread` mode because it is already handled in `zstd_compress.c`.
* Remove the single-pass shortcut fixing Issue #2327.
* Rip out all unused functions from the zstdmt API. Delete the `zstreamtest --mt` test. Rewrite unit tests to use the zstd API.
* Rewrite the `zstreamtest` fuzzer loop to check for compression determinism. It had to be rewritten to always make maximal forward progress to ensure determinism in multithreading mode.

Bug fixes (the last 2 commits: see commit messages):
* Zstdmt would create empty compression jobs when called with `ZSTD_e_end` and when it can't get an input buffer. It would also incorrectly set `mtctx->frameEnded = 1`, which could cause other issues, like an unterminated frame. The fix is to move the switch to `ZSTD_e_flush` outside of the `if()` block.
* Zstdmt with rsyncable enabled could be non-deterministic because it would sometimes "skip" a synchronization point when the job table was full.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -884,7 +884,7 @@ static size_t ZSTDMT_expandJobsTable (ZSTDMT_CCtx* mtctx, U32 nbWorkers) {
 
 /* ZSTDMT_CCtxParam_setNbWorkers():
  * Internal use only */
-size_t ZSTDMT_CCtxParam_setNbWorkers(ZSTD_CCtx_params* params, unsigned nbWorkers)
+static size_t ZSTDMT_CCtxParam_setNbWorkers(ZSTD_CCtx_params* params, unsigned nbWorkers)
 {
     return ZSTD_CCtxParams_setParameter(params, ZSTD_c_nbWorkers, (int)nbWorkers);
 }
@@ -941,11 +941,6 @@ ZSTDMT_CCtx* ZSTDMT_createCCtx_advanced(unsigned nbWorkers, ZSTD_customMem cMem,
     (void)pool;
     return NULL;
 #endif
-}
-
-ZSTDMT_CCtx* ZSTDMT_createCCtx(unsigned nbWorkers)
-{
-    return ZSTDMT_createCCtx_advanced(nbWorkers, ZSTD_defaultCMem, NULL);
 }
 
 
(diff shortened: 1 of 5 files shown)


terrelln:
After ~400K iterations I saw some non-determinism. But it isn't reproducible. I have to figure out if it is a problem with the test or zstdmt.

terrelln:
I believe I've found all the sources of non-determinism. Running 3 x 400K tests now.

terrelln:
There is some non-determinism left. It is pretty rare. I would be okay merging the test as-is, though it may occasionally fail.

I'm running the tests with tracing enabled now. Hopefully it will eventually fail and I can get a trace of what is happening to narrow it down and fix it.

lib/compress/zstdmt_compress.c
```diff
@@ -1239,174 +1174,6 @@ static size_t ZSTDMT_computeOverlapSize(const ZSTD_CCtx_params* params)
...
-        unsigned const nbJobsMax = (unsigned)(srcSize / jobSizeTarget) + 1;
-        unsigned const nbJobsSmall = MIN(nbJobsMax, nbWorkers);
-        return (multiplier>1) ? nbJobsLarge : nbJobsSmall;
-}   }
-
-/* ZSTDMT_compress_advanced_internal() :
- * This is a blocking function : it will only give back control to caller after finishing its compression job.
- */
-static size_t
-ZSTDMT_compress_advanced_internal(
```
Cyan4973, line 1260: Early in the life of `zstdmt`, one of the benefits of this function
is that it was able to compress directly from `src` into `dst`
provided there is enough `dstCapacity` (`<= ZSTD_compressBound(srcSize)`),
thus skipping the need to use intermediate buffers,
thus saving memory, and some unnecessary `memcpy()`.

With this function removed,
is the replacement able to retain this property ?

lib/compress/zstdmt_compress.c
```diff
@@ -2076,8 +1773,16 @@ size_t ZSTDMT_compressStream_generic(ZSTDMT_CCtx* mtctx,
             mtctx->inBuff.filled += syncPoint.toLoad;
             forwardInputProgress = syncPoint.toLoad>0;
         }
-        if ((input->pos < input->size) && (endOp == ZSTD_e_end))
-            endOp = ZSTD_e_flush;   /* can't end now : not all input consumed */
+    }
+    if ((input->pos < input->size) && (endOp == ZSTD_e_end)) {
+        /* Can't end yet because the input is not fully consumed.
+            * We are in one of these cases:
+            * - Input buffer is NULL & empty
```
Cyan4973, line 1780: note : if `input` buffer is `NULL`,
we should also have : 
`input->pos == 0 == input->size`,
therefore, this scenario shouldn't pass the `if()` test.

Cyan4973:
This is a pretty good simplification.

I mostly have a question regarding the memory usage of the replacement function for one-shot compression.
Maybe a good case to employ `stableSrc` & `stableDst` directive,
though I'm not sure if they are actually available for `zstdmt`.

lib/compress/zstdmt_compress.c
terrelln: I should clarify the comment. I mean the `mtctx->inBuff`.

lib/compress/zstdmt_compress.c
terrelln: > is the replacement able to retain this property ?

No. We already weren't saving `memory` because when using the streaming API we have already allocated all of our buffers at this point. With this change, we are doing one extra memcpy for the source buffers.

terrelln:
> I mostly have a question regarding the memory usage of the replacement function for one-shot compression.
> Maybe a good case to employ stableSrc & stableDst directives,
> though I'm not sure if they are actually available for zstdmt.

The memory usage isn't impacted, since there isn't a way to avoid allocating all the buffers using the zstd API.

I think there is opportunity for `stableSrc` and `stableDst` directives, especially with `ZSTD_compress2()`. Both for zstdmt and single-thread zstd. With `stableDst` we could, for example, always take the single-threaded shortcut, even when the output is less than `ZSTD_compressBound()`, since we know the user will never provide more output size. With zstdmt, we could lazily allocate, and save memory when not needed, since we can't handle static allocation anyways.

I think that should be a follow up PR though, since this doesn't regress the current state of the world (except for 1 more memcpy() of the source).

terrelln:
I've opened issue #2353 to handle the memory saving optimizations.

Cyan4973 (approved):

PR #2341, opened 2020-10-02
Integrate long distance matches into optimal parser

senhuang42:
## Context

Using the long range matchfinder (`--long`) with high compression levels (16 and above) currently can cause regressions compared to not using the long mode at all. This is because enabling long mode will cause the matches found by the long range matchfinder to be blindly accepted. This PR changes that behavior, and makes each LDM found a candidate for the optimal parser instead.
 
## --patch-from benchmarks
This PR improves `--long` mode compression the most when `--long` mode is actually useful. `--patch-from` is a great example of `--long` mode being useful, so there are some nice benefits here. With this PR, `--patch-from` becomes even more competitive with `bsdiff` for generating small patches. The following results are fastest of three runs. 

**linux kernels tarballs (5.8 -> 5.9-rc7) : ~1GB dst**
| Name               | Conf    | Patch Size   | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|--------|-------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 5.90MB | 0.58% | 3.24 GB  | 119% | 3m:36s     |
| `zstd-dev`           | -22 -T0 | 6.55MB | 0.65% | 4.09 GB  | 99%  | 5m:54s     |
| `bsdiff`             | default | 5.18MB | 0.51% | 15.58 GB | 99%  | 12m:08s    |
| `zstd-ldm_optimized` | -19 -T0 | 5.10MB | 0.51% | 3.24 GB  | 162% | 4m:30s     |
| `zstd-ldm_optimized` | -22 -T0 | 4.26MB | 0.42% | 4.09 GB  | 99%  | 9m:20s     |

**distant linux kernels tarballs (4.4.238 -> 5.9-rc7): ~1GB dst**
| Name               | Conf    | Patch Size | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|------------|---------------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 66.90MB    | 6.62%         | 2.68 GB  | 190% | 3m:38s     |
| `zstd-dev`           | -22 -T0 | 71.85MB    | 7.11%         | 3.53 GB  | 99%  | 9m:05s     |
| `bsdiff`             | default | 111.69MB   | 11.04%        | 10.3 GB  | 99%  | 10m:01s    |
| `zstd-ldm_optimized` | -19 -T0 | 58.84MB    | 5.82%         | 2.68 GB  | 220% | 3m:49s     |
| `zstd-ldm_optimized` | -22 -T0 | 56.54MB    | 5.59%         | 3.53 GB  | 99%  | 12m:34s    |

**wordpress.tar (5.5 -> 5.5.1): ~80MB dst**
| Name               | Conf    | Patch Size | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|------------|---------------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 21.31KB    | 0.04%         | 174 MB   | 99%  | 9.8s       |
| `bsdiff`             | default | 13.67KB    | 0.03%         | 800 MB   | 99%  | 23.8s      |
| `zstd-ldm_optimized` | -19 -T0 | 18.94KB    | 0.04%         | 174 MB   | 99%  | 9.8s       |

**zstd.tar (1.4.4 -> 1.4.5): ~7MB dst**
| Name               | Conf    | Patch Size | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|------------|---------------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 91.13KB    | 1.12%         | 95 MB    | 99%  | 1.6s       |
| `bsdiff`             | default | 124.23KB   | 1.60%         | 83 MB    | 99%  | 2.3s       |
| `zstd-ldm_optimized` | -19 -T0 | 86.74KB    | 1.12%         | 95 MB    | 99%  | 1.6s       |

Of note is that using `-22 --ultra -T0` with `--patch-from` on actually can cause compressed size regressions on `dev`, compared to `-19 -T0`. This issue is no longer present this PR's branch, however.

## General compression benchmarks

The original goal of this PR was to show that there are no compressed size regression with using `--long` with the opt parser. We demonstrate that fact with the following benchmarks, and show that we can actually get some nice improvements in compression ratio too, particularly in scenarios where `--long` mode is meant to work well. The following results are the fastest of 5 runs.

**five_linux_kernels.tar (4.19.148, 5.4.68, 5.8.12, 5.9-rc7, 5.8.12) ~5GB = 5259889920 bytes**
| Name         | Conf                   | Size      | % of original | CPU% | Time    |
|--------------|------------------------|-----------|-----------------|------|---------|
| `zstd-dev` | -22 --zstd=wlog=30 -T0 | 232896847 | 4.43% | 471% | 14m:30s |
| `zstd-ldm_opt` | -22 --zstd=wlog=30 -T0 | 232896847 | 4.43% | 470% | 14m:31s |
| `zstd-dev` | -22 --long=30 -T0   | 185109024 | 3.52% | 340% | 9m:50s |
| `zstd-ldm_opt` |-22 --long=30 -T0 | 154324523 | 2.93% | 411%  | 12m:30s |

**two_linux_kernels.tar (5.8, 5.9-rc7): ~2GB**
| Name         | Conf                   | Size      | % of original | CPU% | Time    |
|--------------|------------------------|-----------|-----------------|------|---------|
| `zstd-dev`     | -22 --zstd=wlog=30 -T0 | 131944438 | 6.61%           | 195% | 10m:46s |
| `zstd-ldm_opt` | -22 --zstd=wlog=30 -T0 | 131944438 | 6.61%           | 195% | 10m:40s |
| `zstd-dev`     | -22 --long=30 -T0      | 123668734 | 6.20%           | 124% | 9m:32s  |
| `zstd-ldm_opt` | -22 --long=30 -T0      | 120585499 | 6.06%           | 163% | 10m:13s |
| `xz`           | -9 --extreme -T0       | 113720040 | 11.24%          | 951% | 3m:08s  |

`xz` at highest compression settings included here just for fun - it also uses ~8 GB memory compared to 4 GB of `zstd -22 --long=30` too. But it parallelizes the workload incredibly well.

**silesia.tar: ~200MB**

| Name         | Conf                   | Size     | Pct of original | CPU% | Time   |
|--------------|------------------------|----------|-----------------|------|--------|
| `zstd-dev`     | -22 --zstd=wlog=27 -T0 | 52701591 | 24.87%         | 100% | 98.23s |
| `zstd-ldm_opt` | -22 --zstd=wlog=27 -T0 | 52701591 |24.87%          | 100% | 98.87s |
| `zstd-dev`     | -22 --long=27 -T0      | 52777588 | 24.90%          | 100% | 102.23s |
| `zstd-ldm_opt` | -22 --long=27 -T0      | 52695282 | 24.86%          | 100% | 97.98s |

| Name         | Conf                   | Size     | Pct of original | CPU% | Time   |
|--------------|------------------------|----------|-----------------|------|--------|
| `zstd-dev`     | -19 --zstd=wlog=27 -T0 | 53058292 | 25.03%          | 100% | 74.78s |
| `zstd-ldm_opt` | -19 --zstd=wlog=27 -T0 | 53058292 | 25.03%          | 100% | 75.44s |
| `zstd-dev`     | -19 --long=27 -T0      | 53307862 | 25.15%          | 100% | 76.87s |
| `zstd-ldm_opt` | -19 --long=27 -T0      | 53032764 | 25.02%          | 100% | 74.42s |

| Name         | Conf                   | Size     | Pct of original | CPU% | Time   |
|--------------|------------------------|----------|-----------------|------|--------|
| `zstd-dev`     | -16 --zstd=wlog=23 -T0 | 55497790 | 26.18%          | 520% | 11.95s |
| `zstd-ldm_opt` | -16 --zstd=wlog=23 -T0 | 55497790 | 26.18%          | 520% | 12.07s |
| `zstd-dev`     | -16 --long=23 -T0      | 55892935 | 26.37%          | 284% | 18.01s |
| `zstd-ldm_opt` | -16 --long=23 -T0      | 55454265 | 26.16%          | 293% | 17.32s |

Particularly at higher levels (18+), there is virtually no difference in speed on `silesia.tar` when using `--long`, and a speed _gain_ on the concatenated linux kernels file. Not to mention the compression ratio gain. As such, it makes a strong argument for enabling `--long` by default on the highest levels.

### Testing

I've added some more unit tests in `playTests.sh` and a unit test in `fuzzer.c` for regressions, although I'm not sure if there might be weird cases where adding a long match candidate to the optimal parser would legitimately cause a compression ratio regression (and as a result, we shouldn't have these regression tests). Either way, it seems like ldm is already tested decently in `fuzzer.c`, so I don't think adding too many additional tests to verify correctness of compression/roundtrip-ability with ldm here would be that helpful. Letting this patch sit in a fuzzer for a bit might be a good idea though.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2336,7 +2336,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         if (curr > ms->nextToUpdate + 384)
             ms->nextToUpdate = curr - MIN(192, (U32)(curr - ms->nextToUpdate - 384));
     }
-    //printf("--NEW BLOCK--\n");
+
     /* select and store sequences */
     {   ZSTD_dictMode_e const dictMode = ZSTD_matchState_dictMode(ms);
         size_t lastLLSize;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2336,7 +2336,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         if (curr > ms->nextToUpdate + 384)
             ms->nextToUpdate = curr - MIN(192, (U32)(curr - ms->nextToUpdate - 384));
     }
-    
+
     /* select and store sequences */
     {   ZSTD_dictMode_e const dictMode = ZSTD_matchState_dictMode(ms);
         size_t lastLLSize;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2336,7 +2336,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         if (curr > ms->nextToUpdate + 384)
             ms->nextToUpdate = curr - MIN(192, (U32)(curr - ms->nextToUpdate - 384));
     }
-
+    printf("--NEW BLOCK--\n");
     /* select and store sequences */
     {   ZSTD_dictMode_e const dictMode = ZSTD_matchState_dictMode(ms);
         size_t lastLLSize;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2354,7 +2354,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
                                        src, srcSize);
             assert(zc->externSeqStore.pos <= zc->externSeqStore.size);
         } else if (zc->appliedParams.ldmParams.enableLdm) {
-            rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0, 0};
+            rawSeqStore_t ldmSeqStore = kNullRawSeqStore;
 
             ldmSeqStore.seq = zc->ldmSequences;
             ldmSeqStore.capacity = zc->maxNbLdmSequences;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2336,7 +2336,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         if (curr > ms->nextToUpdate + 384)
             ms->nextToUpdate = curr - MIN(192, (U32)(curr - ms->nextToUpdate - 384));
     }
-    printf("--NEW BLOCK--\n");
+    //printf("--NEW BLOCK--\n");
     /* select and store sequences */
     {   ZSTD_dictMode_e const dictMode = ZSTD_matchState_dictMode(ms);
         size_t lastLLSize;
@@ -2354,7 +2354,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
                                        src, srcSize);
             assert(zc->externSeqStore.pos <= zc->externSeqStore.size);
         } else if (zc->appliedParams.ldmParams.enableLdm) {
-            rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0};
+            rawSeqStore_t ldmSeqStore = {NULL, NULL, 0, 0, 0};
 
             ldmSeqStore.seq = zc->ldmSequences;
             ldmSeqStore.capacity = zc->maxNbLdmSequences;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2371,6 +2371,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
             assert(ldmSeqStore.pos == ldmSeqStore.size);
         } else {   /* not long range mode */
             ZSTD_blockCompressor const blockCompressor = ZSTD_selectBlockCompressor(zc->appliedParams.cParams.strategy, dictMode);
+            ms->ldmSeqStore = NULL;
             lastLLSize = blockCompressor(ms, &zc->seqStore, zc->blockState.nextCBlock->rep, src, srcSize);
         }
         {   const BYTE* const lastLiterals = (const BYTE*)src + srcSize - lastLLSize;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2336,7 +2336,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         if (curr > ms->nextToUpdate + 384)
             ms->nextToUpdate = curr - MIN(192, (U32)(curr - ms->nextToUpdate - 384));
     }
-
+    
     /* select and store sequences */
     {   ZSTD_dictMode_e const dictMode = ZSTD_matchState_dictMode(ms);
         size_t lastLLSize;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2354,7 +2354,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
                                        src, srcSize);
             assert(zc->externSeqStore.pos <= zc->externSeqStore.size);
         } else if (zc->appliedParams.ldmParams.enableLdm) {
-            rawSeqStore_t ldmSeqStore = {NULL, NULL, 0, 0, 0};
+            rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0, 0};
 
             ldmSeqStore.seq = zc->ldmSequences;
             ldmSeqStore.capacity = zc->maxNbLdmSequences;
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -578,7 +578,6 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     if (cParams->strategy >= ZSTD_btopt) {
         size_t lastLLSize;
         ms->ldmSeqStore = *rawSeqStore;
-        ms->ldmSeqStore.base = ms->window.base;
         lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
         /* ldm seqstore will have changed during blockCompressor() call, make sure we copy those changes */
         *rawSeqStore = ms->ldmSeqStore;
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -579,8 +579,7 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
         size_t lastLLSize;
         ms->ldmSeqStore = *rawSeqStore;
         lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
-        /* ldm seqstore will have changed during blockCompressor() call, make sure we copy those changes */
-        *rawSeqStore = ms->ldmSeqStore;
+        *rawSeqStore = ms->ldmSeqStore;  /* Persist changes to ldmSeqStore during blockCompressor() */
         return lastLLSize;
     }
 
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -562,13 +562,6 @@ static rawSeq maybeSplitSequence(rawSeqStore_t* rawSeqStore,
     return sequence;
 }
 
-static void printSeqStore(rawSeqStore_t* rawSeqStore) {
-    printf("rawSeqStore: pos: %zu\n", rawSeqStore->pos);
-    for (int i = 0; i < rawSeqStore->size; ++i) {
-        printf("pos %d (of:%u ml:%u ll: %u)\n", i, rawSeqStore->seq[i].offset, rawSeqStore->seq[i].matchLength, rawSeqStore->seq[i].litLength);
-    }
-}
-
 size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
     void const* src, size_t srcSize)
@@ -582,19 +575,13 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     BYTE const* const iend = istart + srcSize;
     /* Input positions */
     BYTE const* ip = istart;
-    //printSeqStore(rawSeqStore);
     if (cParams->strategy >= ZSTD_btopt) {
         size_t lastLLSize;
-        //printSeqStore(rawSeqStore);
-        ms->ldmSeqStore = *rawSeqStore; /* copy current seqStore */
+        ms->ldmSeqStore = *rawSeqStore;
         ms->ldmSeqStore.base = ms->window.base;
         lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
+        /* ldm seqstore will have changed during blockCompressor() call, make sure we copy those changes */
         *rawSeqStore = ms->ldmSeqStore;
-        /*if (prevBase != ms->window.base) {
-            int baseDiff = (int)(prevBase - ms->window.base);
-            printf("Bases were different, adjusting, diff = %d\n", baseDiff);
-            rawSeqStore->seq[rawSeqStore->pos].litLength += baseDiff;
-        }*/
         return lastLLSize;
     }
 
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -555,6 +555,26 @@ static rawSeq maybeSplitSequence(rawSeqStore_t* rawSeqStore,
     return sequence;
 }
 
+/* ZSTD_ldm_skipRawSeqStoreBytes():
+ * Moves forward in rawSeqStore by nbBytes, updating fields 'pos' and 'posInSequence'.
+ */
+static void ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore_t* rawSeqStore, size_t nbBytes) {
+    U32 currPos = rawSeqStore->posInSequence + nbBytes;
+    while (currPos && rawSeqStore->pos < rawSeqStore->size) {
+        rawSeq currSeq = rawSeqStore->seq[rawSeqStore->pos];
+        if (currPos >= currSeq.litLength + currSeq.matchLength) {
+            currPos -= currSeq.litLength + currSeq.matchLength;
+            rawSeqStore->pos++;
+            if (currPos == 0) {
+                rawSeqStore->posInSequence = 0;
+            }
+        } else {
+            rawSeqStore->posInSequence = currPos;
+            break;
+        }
+    }
+}
+
 size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
     void const* src, size_t srcSize)
@@ -574,7 +594,11 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
         size_t lastLLSize;
         ms->ldmSeqStore = *rawSeqStore;
         lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
-        *rawSeqStore = ms->ldmSeqStore;
+        ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore, srcSize);
+        if (rawSeqStore->pos >= rawSeqStore->size) {
+            /* If we're done with rawSeqStore, invalidate the one in matchState as well */
+            ms->ldmSeqStore.pos = rawSeqStore->size;
+        }
         return lastLLSize;
     }
 
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -27,13 +27,6 @@ void ZSTD_ldm_adjustParameters(ldmParams_t* params,
     DEBUGLOG(4, "ZSTD_ldm_adjustParameters");
     if (!params->bucketSizeLog) params->bucketSizeLog = LDM_BUCKET_SIZE_LOG;
     if (!params->minMatchLength) params->minMatchLength = LDM_MIN_MATCH_LENGTH;
-    if (cParams->strategy >= ZSTD_btopt) {
-      /* Get out of the way of the optimal parser */
-      U32 const minMatch = MAX(cParams->targetLength, params->minMatchLength);
-      assert(minMatch >= ZSTD_LDM_MINMATCH_MIN);
-      assert(minMatch <= ZSTD_LDM_MINMATCH_MAX);
-      params->minMatchLength = minMatch;
-    }
     if (params->hashLog == 0) {
         params->hashLog = MAX(ZSTD_HASHLOG_MIN, params->windowLog - LDM_HASH_RLOG);
         assert(params->hashLog <= ZSTD_HASHLOG_MAX);
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -576,6 +576,13 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     /* Input positions */
     BYTE const* ip = istart;
 
+    if (cParams->strategy >= ZSTD_btopt) {
+        size_t lastLLSize;
+        ms->ldmSeqStore = *rawSeqStore; /* copy current seqStore */
+        lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
+        return lastLLSize;
+    }
+
     DEBUGLOG(5, "ZSTD_ldm_blockCompress: srcSize=%zu", srcSize);
     assert(rawSeqStore->pos <= rawSeqStore->size);
     assert(rawSeqStore->size <= rawSeqStore->capacity);
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -575,6 +575,8 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     BYTE const* const iend = istart + srcSize;
     /* Input positions */
     BYTE const* ip = istart;
+
+    /* If using opt parser, use LDMs only as candidates rather than always accepting them */
     if (cParams->strategy >= ZSTD_btopt) {
         size_t lastLLSize;
         ms->ldmSeqStore = *rawSeqStore;
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -562,6 +562,13 @@ static rawSeq maybeSplitSequence(rawSeqStore_t* rawSeqStore,
     return sequence;
 }
 
+static void printSeqStore(rawSeqStore_t* rawSeqStore) {
+    printf("rawSeqStore: pos: %zu, bytesDiscarded: %zu\n", rawSeqStore->pos);
+    for (int i = 0; i < rawSeqStore->size; ++i) {
+        printf("(of:%u ml:%u ll: %u)\n", rawSeqStore->seq[i].offset, rawSeqStore->seq[i].matchLength, rawSeqStore->seq[i].litLength);
+    }
+}
+
 size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
     void const* src, size_t srcSize)
@@ -578,6 +585,7 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
 
     if (cParams->strategy >= ZSTD_btopt) {
         size_t lastLLSize;
+        printSeqStore(rawSeqStore);
         ms->ldmSeqStore = *rawSeqStore; /* copy current seqStore */
         const BYTE* const prevBase = (BYTE const*)ms->window.base;
         lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -573,8 +573,6 @@ static void ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore_t* rawSeqStore, size_t nbB
     if (currPos == 0 || rawSeqStore->pos == rawSeqStore->size) {
         rawSeqStore->posInSequence = 0;
     }
-    assert(rawSeqStore->posInSequence <=
-           rawSeqStore->seq[rawSeqStore->pos].litLength + rawSeqStore->seq[rawSeqStore->pos].matchLength);
 }
 
 size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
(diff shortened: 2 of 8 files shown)


lib/compress/zstd_ldm.c
```diff
@@ -576,6 +576,15 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     /* Input positions */
     BYTE const* ip = istart;
 
+    /* If using opt parser, use LDMs only as candidates rather than always accepting them */
+    if (cParams->strategy >= ZSTD_btopt) {
+        size_t lastLLSize;
+        ms->ldmSeqStore = *rawSeqStore;
+        lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
+        *rawSeqStore = ms->ldmSeqStore;  /* Persist changes to ldmSeqStore during blockCompressor() */
```
Cyan4973, line 584: Is it still required ?
If `ms->ldmSeqStore` is no longer modified, it should not be necessary.

lib/compress/zstdmt_compress.c
```diff
@@ -277,7 +277,7 @@ static size_t ZSTDMT_sizeof_seqPool(ZSTDMT_seqPool* seqPool)
 
 static rawSeqStore_t bufferToSeq(buffer_t buffer)
 {
-    rawSeqStore_t seq = {NULL, 0, 0, 0};
+    rawSeqStore_t seq = {NULL, 0, 0, 0, 0};
```
Cyan4973, line 280: `= kNullRawSeqStore;`
I guess that's the point of defining this constant.

lib/compress/zstd_opt.c
```diff
@@ -764,6 +764,180 @@ FORCE_INLINE_TEMPLATE U32 ZSTD_BtGetAllMatches (
...
+ * Adds a match if it's long enough, based on it's 'matchStartPosInBlock'
+ * and 'matchEndPosInBlock', into 'matches'. Maintains the correct ordering of 'matches'
+ */
+static void ZSTD_opt_maybeAddLdm(ZSTD_match_t* matches, U32* nbMatches,
+                                 U32 matchStartPosInBlock, U32 matchEndPosInBlock,
+                                 U32 matchOffset, U32 currPosInBlock) {
+    U32 posDiff = currPosInBlock - matchStartPosInBlock;
+    /* Note: ZSTD_match_t actually contains offCode and matchLength (before subtracting MINMATCH) */
+    U32 candidateMatchLength = matchEndPosInBlock - matchStartPosInBlock - posDiff;
+    U32 candidateOffCode = matchOffset + posDiff + ZSTD_REP_MOVE;
```
senhuang42, line 869: Just realized that I shouldn't be adding `posDiff` here, but it doesn't affect this commit since `posDiff` is guaranteed to be 0 if we proceed.

So after fixing that bug, enabling the functionality to add "extra" LDMs is a one-liner. I.e. for any ldm with `matchLength > 3`, insert all intermediate matches, decrementing `matchLength` until `matchLength == 3`. We simply remove the check for `posDiff > 0` in the condition below. Based on some brief tests just now, it does bring some _slight_ compression ratio benefits on extreme cases, at the cost of severe compression speed reduction.

**Compressing two linux kernels**
Current commit - Ratio: 6.14%, Time: 10m:46s
Adding extra LDMs - Ratio: 6.12% Time: 15m:39s

**Compressing five linux kernels**
Current commit - Ratio: 3.36%, Time: 10m:46s
Adding extra LDMs - Ratio: 3.33% Time: 36m:08s

lib/compress/zstd_compress.c
```diff
@@ -2338,7 +2338,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
                                        src, srcSize);
             assert(zc->externSeqStore.pos <= zc->externSeqStore.size);
         } else if (zc->appliedParams.ldmParams.enableLdm) {
-            rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0};
+            rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0, 0};
```
Cyan4973, line 2341: `= kNullRawSeqStore`

lib/compress/zstd_ldm.c
senhuang42: The array of `seq` in `ms->ldmSeqStore` is immutable, but we still need to keep an updated `pos` field through some blocks.

lib/compress/zstd_opt.c
Cyan4973: Thanks for these early results !
The compression ratio benefit is in line with expectation,
but the severe speed reduction is not. And is a concern.

It will be interesting to understand why that happens.
Presuming the current strategy of only presenting ldm matchLen > targetLen,
the parser is expected to accept such large matches point blank,
without even considering any further update up to the end of the long match.

senhuang42: I guess that part of the reason may be that for `-22`, the `targetLength = 999` - in this case it's pretty likely that we find a bunch of LDMs less than `999`. 

So if our LDM had length 999, we'd potentially insert and process up to a maximum of 996 additional matches over the course of the next 996 positions, and have to perform all the operations that exist in the loops that exist in the lines of code at `zstd_opt.c:1172 - 1197` calculating the relative cost and whatnot, which probably doesn't come for free.

Other than that, removing the `posDiff > 0` line of code also will use up a few more operations in the code that I added, even when rejecting the new partial LDMs, but I don't think that should be too significant since it's mostly a couple of compare operations.

Cyan4973: Trying `-22 --long` on `silesia.tar` : 
```
../lib/compress/zstd_ldm.c: ZSTD_ldm_adjustParameters
../lib/compress/zstd_ldm.c: minMatchLength : 999 ;  hashRateLog : 7  ; bucketSizeLog : 3
```

The ldm should only deliver matches which are >= 999 bytes.

That being said, if a ldm match can be started mid-flight, it can result in a candidate which has a length < 999.

It could be interesting to trace a few statistics on this topic to visualize better.

senhuang42:
I've made a few follow-up adjustments to improve compression ratio further in these last few commits suggested by @Cyan4973 : inclusion of mid-flight LDMs (`mid_flight`) and removing the `minMatch` adjustment in the ldm generator (`minmatch_rm`). I've included some benchmarks (fastest of two runs) for the additional changes compared to the current branch. On `five_linux_kernels.tar` we notice a _massive_ improvement in compression ratio with the two changes.

I noticed that `silesia.tar` was actually seeing some _tiny_ regressions in compressed size because of these two commits. I suspected that inclusion of LDMs with the same `mlen` but larger/"worse" `offCode` in the candidate `matches` array may have been hurting the ratio, even though they'd be respecting the ordering of the candidate `matches`. Perhaps because the `offCode` of LDMs are almost always huge by a large margin and likely less compressible. As a result, even with a match length equaling the current longest match, it probably doesn't belong in the 2nd or 3rd spot from the end in the `matches` array. 

I removed inclusion of these LDMs in another commit (`bubble_insertion_rm`) and compression ratio improved, particularly on `silesia.tar`, though just on the scale of a few KB.

**five_linux_kernels.tar** on `ldm_optimized`
conf | commit | size | ratio | time | cpu% |
| --- | --- | --- | --- | --- | --- |
| -22 **--long=30** | `HEAD` | 176535234 | 3.36% | 12m:20s | 389% |
| -22 **--long=30** | `+mid_flight` | 174945548 | 3.33% | 11m:39s | 391% | 
| -22 **--long=30** | `+mid_flight+minmatch_rm` | 154797141 | 2.94% | 12m:01s | 417% | 
| -22 **--long=30** | `+mid_flight+minmatch_rm+bubble_insertion_rm `| 154279903 | 2.93% | 11m:57s | 413% | 
|||||||

**silesia.tar** on `ldm_optimized`
conf | commit | size | ratio | time | cpu% |
| --- | --- | --- | --- | --- | --- |
| -22 **--long=27** | `HEAD` | 52699843 | 24.86% | 105.82s | 100% | 
| -22 **--long=27** | `+mid_flight` | 52700110 | 24.86% | 102.66s | 100% | 
| -22 **--long=27** | `+mid_flight+minmatch_rm` | 52714246 | 24.87% | 104.98s | 100% | 
| -22 **--long=27** | `+mid_flight+minmatch_rm+bubble_insertion_rm `| 52695409 | 24.86% | 114.33s | 100% | 
|||||||

lib/compress/zstd_opt.c
senhuang42: This speed regression wasn't making sense to me so I tried reproducing it today and couldn't - at which point I realized that in my quick check last Friday I must have ran it without `-T0`.

Though, that does once again bring up the separate issue of the rules for triggering multithreaded mode automatically, since this seems like a case where we'd want multithreaded mode to get triggered.

Cyan4973:
> I suspected that inclusion of LDMs with the same mlen but larger/"worse" offCode in the candidate matches array may have been hurting the ratio

The optimal parser presumes that the list of candidates
comes using ___strictly__ increasing_ match lengths.
Hence, adding a candidate with the same length but larger offset
is violating this assumption, and likely results in the parser making less optimal choices.

lib/compress/zstd_ldm.c
```diff
@@ -576,6 +569,15 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
     /* Input positions */
     BYTE const* ip = istart;
 
+    /* If using opt parser, use LDMs only as candidates rather than always accepting them */
+    if (cParams->strategy >= ZSTD_btopt) {
+        size_t lastLLSize;
+        ms->ldmSeqStore = *rawSeqStore;
```
terrelln, line 575: Why not use a pointer for the `ldmSeqStore`, since we're effectively modifying the original `rawSeqStore`?

lib/compress/zstd_opt.c
```diff
@@ -818,6 +968,17 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
     ZSTD_match_t* const matches = optStatePtr->matchTable;
     ZSTD_optimal_t lastSequence;
 
+    U32 ldmStartPosInBlock = 0;
+    U32 ldmEndPosInBlock = 0;
+    U32 ldmOffset = 0;
```
terrelln, line 973: These 3 variable are all related and modified together. And in fact the optimal parser doesn't care about these variables at all, it only matters for the LDM helper functions.

These should be grouped together into a struct. I'd probably also include the `&ms->ldmSeqStore` in this struct.

```c
typedef struct {
  rawSeqStore* seqStore;
  U32 startPosInBlock;
  U32 endPosInBlock;
  U32 offset;
} ZSTD_optLdm_t;
```

Then you might want to rename all the functions to be `ZSTD_optLdm_blah(ZSTD_optLdm_t* optLdm, ...)` to be clear it is operating on this struct.

Cyan4973:
Note : these improvements may also positively impact the `--patch-from` mode.
Would be interesting to compare with performance tests completed for the `v1.4.5` announcement of `--patch-from`.

lib/compress/zstd_opt.c
```diff
@@ -818,6 +968,17 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
...
+
+    /* Get first match from ldm seq store if long mode is enabled */
+    if (ms->ldmSeqStore.size > 0 && ms->ldmSeqStore.pos < ms->ldmSeqStore.size) {
+        ZSTD_opt_getNextLdmAndUpdateSeqStore(&ms->ldmSeqStore, &ldmStartPosInBlock,
+                                             &ldmEndPosInBlock, &ldmOffset,
+                                             (U32)(ip-istart), (U32)(iend-ip));
+    }
```
terrelln, line 980: I think these checks should be moved inside of the LDM functions, so they aren't duplicated. Another benefit is that it will minimize the amount of LDM logic that the optimal parser has, just these 4 functions calls.

E.g. Here you check `size > 0` and below you check `size != 0`. They are equivalent, but it shows that differences can crop up.

terrelln:
@senhuang42 this is looking great! Just have some minor style suggestions.

lib/compress/zstd_ldm.c
senhuang42: So apparently `*rawSeqStore` is can be something allocated on the stack or heap (depending on multithreading in `buildSeqStore()`). So if `ms->ldmSeqStore` were a pointer, we'd lose track of the `*rawSeqStore` it points to once we call `blockCompressor()` - which is why we just copy the `*rawSeqStore`.

terrelln: > So if ms->ldmSeqStore were a pointer, we'd lose track of the *rawSeqStore it points to once we call blockCompressor()

What do you mean by that?

lib/compress/zstd_ldm.c
```diff
@@ -576,6 +569,15 @@ size_t ZSTD_ldm_blockCompress(rawSeqStore_t* rawSeqStore,
...
+        lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
+        *rawSeqStore = ms->ldmSeqStore;  /* Persist changes to ldmSeqStore during blockCompressor() */
```
terrelln:
```suggestion
        ms->ldmSeqStore = rawSeqStore;
        lastLLSize = blockCompressor(ms, seqStore, rep, src, srcSize);
        ms->ldmSeqStore = NULL;
```

terrelln: If it is a pointer (const pointer if not modified), and we set reset it to `NULL` whenever it isn't active, then it is very simple to tell if LDM is active or not in the opt parser.

lib/compress/zstd_ldm.c
senhuang42: In `ZSTD_buildSeqStore()`, there are two ways to trigger the ldm codepath: 1) there exists a `zc->externSeqStore` (ldm is enabled with multithreading), or 2) ldm is enabled, and is allocated on the stack as `rawSeqStore_t ldmSeqStore = {NULL, 0, 0, 0, 0};`.

So in the second case, if `ms->ldmSeqStore` is a pointer to the `ldmSeqStore` object, then once we call `blockCompressor(ms, seqStore, rep, src, srcSize);`, because the pointer to the object itself isn't in the function arguments, it seems that by entering the new scope of `blockCompressor()`, the pointer doesn't seem to point to anything that's valid anymore. Or at least that's what I think is happening. A long time ago I did have `ms->ldmSeqStore` as a pointer and ran into this bug where the `ldmSeqStore` did not persist into the call to `blockCompressor()`.

lib/compress/zstdmt_compress.c
```diff
@@ -266,7 +266,7 @@ static void ZSTDMT_releaseBuffer(ZSTDMT_bufferPool* bufPool, buffer_t buf)
 
 /* =====   Seq Pool Wrapper   ====== */
 
-static rawSeqStore_t kNullRawSeqStore = {NULL, 0, 0, 0};
+static rawSeqStore_t kNullRawSeqStore = {NULL, 0, 0, 0, 0};
```
Cyan4973, line 269: `static const` since it's a constant

lib/compress/zstd_opt.c
```diff
@@ -817,6 +962,17 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
     ZSTD_optimal_t* const opt = optStatePtr->priceTable;
     ZSTD_match_t* const matches = optStatePtr->matchTable;
     ZSTD_optimal_t lastSequence;
+    ZSTD_optLdm_t optLdm;
+
+    if (ms->ldmSeqStore) {
+        optLdm.seqStore = *ms->ldmSeqStore;
+        optLdm.endPosInBlock = optLdm.startPosInBlock = optLdm.offset = 0;
+    } else {
+        optLdm.seqStore.size = optLdm.seqStore.pos = 0;
```
senhuang42, line 971: Could also use `kNullRawSeqStore` here.

lib/compress/zstd_opt.c
```diff
@@ -764,6 +764,151 @@ FORCE_INLINE_TEMPLATE U32 ZSTD_BtGetAllMatches (
...
+    }
+
+    DEBUGLOG(6, "ZSTD_optLdm_maybeAddMatch(): Adding ldm candidate match (offCode: %u matchLength %u) at block position=%u",
+             candidateOffCode, candidateMatchLength, currPosInBlock);
+    if (*nbMatches == 0) {
+        matches[*nbMatches].len = candidateMatchLength;
+        matches[*nbMatches].off = candidateOffCode;
+        (*nbMatches)++;
+    } else if ((candidateMatchLength >= matches[*nbMatches-1].len) && *nbMatches < ZSTD_OPT_NUM) {
+        if (candidateMatchLength == matches[*nbMatches-1].len) {
```
Cyan4973, line 877: This doesn't seem worthwhile.
This case seems very rare, and offer limited, if any, benefit.

I guess it means this could be simplified : 
only take care of the case when `candidateMatchLength` is strictly `>` longer.

This is a very minor tweak though : I don't expect either runtime nor compression to be significantly impacted.

Cyan4973:
This all looks good to me.

I also went ahead, and decided to measure the impact of this PR on `--patch-from` at high compression level (`-19`),
using 2 linux tarball (5.7->5.8) compressed as delta.

And indeed, compression ratio is significantly improved : 
current `dev` branch produces a patch of ~7.33 MB,
while this PR produces a smaller patch of ~6.31 MB.
This is > 10% additional saving, so not bad!

I tried to compare with `bsdiff` compression ratio, but that's one of those examples where `--patch-from` shines for other reasons : it completed the patch in 5mn. On the other hand, `bsdiff` has been running for 1h30mn, and finally failed, primarily due to gigantic memory requirements to produce the diff (> 16 GB). In contrast, `zstd` was fine with 1.3 GB. Still a lot, but nonetheless considerably better.

lib/compress/zstd_opt.c
senhuang42: Yeah, I measured a few hundred bytes saved on `silesia.tar` which is why I kept it in since it seemed like a more or less "free" gain, but I'm fine removing this check if making the code simpler is worth that.

Cyan4973: In my tests, the gain from this section of the code (`==` matchlength) is variable, sometimes better, sometimes worse.
For `silesia.tar`, at `-19 --long=23`, strict `>` is smaller by 3 bytes.
But indeed, at `--ultra -22 --long`, strict `>` is larger by ~150 bytes.

If you want to keep it, consider a variant which tests : `if (4*candidateOffCode < matches[*nbMatches-1].off)`

i.e. only replace the existing match by the new one if it is _noticeably_ cheaper to encode.

I tested this variant, and it seems more consistently positive,
including on `silesia.tar`, at all compression settings tested.
At `--ultra -22 --long`, it saves an additional KB.

senhuang42: Good idea! It's interesting that the long distance match finder in certain cases can even produce matches with offsets multiple times smaller than what the regular match finder produces - it's definitely a little surprising..

senhuang42: I just measured myself, and oddly couldn't replicate the same results on `silesia.tar`:

With `-22 --ultra --long`:
`candidateOffCode < matches[*nbMatches-1].off <: 52694897`
`2*candidateOffCode < matches[*nbMatches-1].off <: 52695398`
`4*candidateOffCode < matches[*nbMatches-1].off <: 52695296`
`candidateMatchLength > matches[*nbMatches-1].len : 52695282`

With `-19 --long`:
`candidateOffCode < matches[*nbMatches-1].off < : 53032567`
`2*candidateOffCode < matches[*nbMatches-1].off <: 53034038`
`4*candidateOffCode < matches[*nbMatches-1].off < : 53033516`
`candidateMatchLength > matches[*nbMatches-1].len : 53032764`

On concatenated linux kernels with `-22 --ultra --long=30`
`candidateOffCode < matches[*nbMatches-1].off < : 120561718`
`2*candidateOffCode < matches[*nbMatches-1].off <: 120563852`
`4*candidateOffCode < matches[*nbMatches-1].off <: 120555259`
`candidateMatchLength > matches[*nbMatches-1].len : 120564642`

Despite the `silesia.tar` results, it seems like on the case that is most practical (large, similar files concatenated), the `4*candidateOffCode` change is most beneficial by a fair bit.

Cyan4973: Indeed, we don't experience the same results.

I presume that's because `silesia.tar` actually doesn't exist,
the `silesia` corpus is distributed as one or several `.zip` files,
and the `.tar` archive must be rebuild locally, using whatever local version of `tar`.
Which means that 2 different systems can produce slightly different `tar` archives.
It's enough to produce measurable differences.

It wouldn't matter as much if the overall direction was always the same,
but on such tiny changes, it ends up measuring benefits on one side, and losses on the other one,
so it's confusing.

I also agree that, since results remain so close to each other, the simplified "strict `>`" version feels more appealing
(though I'm surprised that difference can be as much as 10KB for the concatenated linux kernel).

senhuang42:
More benchmarks:

## --patch-from benchmarks
This PR improves `--long` mode compression the most when `--long` mode is actually useful. `--patch-from` is a great example of `--long` mode being useful, so there are some nice benefits here. With this PR, `--patch-from` becomes even more competitive with `bsdiff` for generating small patches. The following results are fastest of three runs. 

**linux kernels tarballs (5.8 -> 5.9-rc7) : ~1GB dst**
| Name               | Conf    | Patch Size   | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|--------|-------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 5.90MB | 0.58% | 3.24 GB  | 119% | 3m:36s     |
| `zstd-dev`           | -22 -T0 | 6.55MB | 0.65% | 4.09 GB  | 99%  | 5m:54s     |
| `bsdiff`             | default | 5.18MB | 0.51% | 15.58 GB | 99%  | 12m:08s    |
| `zstd-ldm_optimized` | -19 -T0 | 5.10MB | 0.51% | 3.24 GB  | 162% | 4m:30s     |
| `zstd-ldm_optimized` | -22 -T0 | 4.26MB | 0.42% | 4.09 GB  | 99%  | 9m:20s     |

**distant linux kernels tarballs (4.4.238 -> 5.9-rc7): ~1GB dst**
| Name               | Conf    | Patch Size | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|------------|---------------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 66.90MB    | 6.62%         | 2.68 GB  | 190% | 3m:38s     |
| `zstd-dev`           | -22 -T0 | 71.85MB    | 7.11%         | 3.53 GB  | 99%  | 9m:05s     |
| `bsdiff`             | default | 111.69MB   | 11.04%        | 10.3 GB  | 99%  | 10m:01s    |
| `zstd-ldm_optimized` | -19 -T0 | 58.84MB    | 5.82%         | 2.68 GB  | 220% | 3m:49s     |
| `zstd-ldm_optimized` | -22 -T0 | 56.54MB    | 5.59%         | 3.53 GB  | 99%  | 12m:34s    |

**wordpress.tar (5.5 -> 5.5.1): ~80MB dst**
| Name               | Conf    | Patch Size | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|------------|---------------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 21.31KB    | 0.04%         | 174 MB   | 99%  | 9.8s       |
| `bsdiff`             | default | 13.67KB    | 0.03%         | 800 MB   | 99%  | 23.8s      |
| `zstd-ldm_optimized` | -19 -T0 | 18.94KB    | 0.04%         | 174 MB   | 99%  | 9.8s       |

**zstd.tar (1.4.4 -> 1.4.5): ~7MB dst**
| Name               | Conf    | Patch Size | % of new file | Peak Mem | CPU% | Total Time |
|--------------------|---------|------------|---------------|----------|------|------------|
| `zstd-dev`           | -19 -T0 | 91.13KB    | 1.12%         | 95 MB    | 99%  | 1.6s       |
| `bsdiff`             | default | 124.23KB   | 1.60%         | 83 MB    | 99%  | 2.3s       |
| `zstd-ldm_optimized` | -19 -T0 | 86.74KB    | 1.12%         | 95 MB    | 99%  | 1.6s       |

Of note is that using `-22 --ultra -T0` with `--patch-from` on actually can cause compressed size regressions on `dev`, compared to `-19 -T0`. This issue is no longer present this PR's branch, however.

## General compression benchmarks

The original goal of this PR was to show that there are no compressed size regression with using `--long` with the opt parser. We demonstrate that fact with the following benchmarks, and show that we can actually get some nice improvements in compression ratio too, particularly in scenarios where `--long` mode is meant to work well. The following results are the fastest of 5 runs.

**five_linux_kernels.tar (4.19.148, 5.4.68, 5.8.12, 5.9-rc7, 5.8.12) ~5GB = 5259889920 bytes**
| Name         | Conf                   | Size      | % of original | CPU% | Time    |
|--------------|------------------------|-----------|-----------------|------|---------|
| `zstd-dev` | -22 --zstd=wlog=30 -T0 | 232896847 | 4.43% | 471% | 14m:30s |
| `zstd-ldm_opt` | -22 --zstd=wlog=30 -T0 | 232896847 | 4.43% | 470% | 14m:31s |
| `zstd-dev` | -22 --long=30 -T0   | 185109024 | 3.52% | 340% | 9m:50s |
| `zstd-ldm_opt` |-22 --long=30 -T0 | 154324523 | 2.93% | 411%  | 12m:30s |

**two_linux_kernels.tar (5.8, 5.9-rc7): ~2GB**
| Name         | Conf                   | Size      | % of original | CPU% | Time    |
|--------------|------------------------|-----------|-----------------|------|---------|
| `zstd-dev`     | -22 --zstd=wlog=30 -T0 | 131944438 | 6.61%           | 195% | 10m:46s |
| `zstd-ldm_opt` | -22 --zstd=wlog=30 -T0 | 131944438 | 6.61%           | 195% | 10m:40s |
| `zstd-dev`     | -22 --long=30 -T0      | 123668734 | 6.20%           | 124% | 9m:32s  |
| `zstd-ldm_opt` | -22 --long=30 -T0      | 120585499 | 6.06%           | 163% | 10m:13s |
| `xz`           | -9 --extreme -T0       | 113720040 | 11.24%          | 951% | 3m:08s  |

`xz` at highest compression settings included here just for fun - it also uses ~8 GB memory compared to 4 GB of `zstd -22 --long=30` too. But it parallelizes the workload incredibly well.

**silesia.tar: ~200MB**

| Name         | Conf                   | Size     | Pct of original | CPU% | Time   |
|--------------|------------------------|----------|-----------------|------|--------|
| `zstd-dev`     | -22 --zstd=wlog=27 -T0 | 52701591 | 24.87%         | 100% | 98.23s |
| `zstd-ldm_opt` | -22 --zstd=wlog=27 -T0 | 52701591 |24.87%          | 100% | 98.87s |
| `zstd-dev`     | -22 --long=27 -T0      | 52777588 | 24.90%          | 100% | 102.23s |
| `zstd-ldm_opt` | -22 --long=27 -T0      | 52695282 | 24.86%          | 100% | 97.98s |

| Name         | Conf                   | Size     | Pct of original | CPU% | Time   |
|--------------|------------------------|----------|-----------------|------|--------|
| `zstd-dev`     | -19 --zstd=wlog=27 -T0 | 53058292 | 25.03%          | 100% | 74.78s |
| `zstd-ldm_opt` | -19 --zstd=wlog=27 -T0 | 53058292 | 25.03%          | 100% | 75.44s |
| `zstd-dev`     | -19 --long=27 -T0      | 53307862 | 25.15%          | 100% | 76.87s |
| `zstd-ldm_opt` | -19 --long=27 -T0      | 53032764 | 25.02%          | 100% | 74.42s |

| Name         | Conf                   | Size     | Pct of original | CPU% | Time   |
|--------------|------------------------|----------|-----------------|------|--------|
| `zstd-dev`     | -16 --zstd=wlog=23 -T0 | 55497790 | 26.18%          | 520% | 11.95s |
| `zstd-ldm_opt` | -16 --zstd=wlog=23 -T0 | 55497790 | 26.18%          | 520% | 12.07s |
| `zstd-dev`     | -16 --long=23 -T0      | 55892935 | 26.37%          | 284% | 18.01s |
| `zstd-ldm_opt` | -16 --long=23 -T0      | 55454265 | 26.16%          | 293% | 17.32s |

Cyan4973 (approved):

PR #2348, opened 2020-10-11
Fix dll path in case of cross-compilation

dscheg:
The result of building lib under Windows Subsystem for Linux is dlllibzstd.dll in lib folder, not libzstd.dll in lib/dll as expected.

```
zstd/lib$ make CC=x86_64-w64-mingw32-gcc TARGET_SYSTEM=Windows_NT all
compiling dynamic library 1.4.5
x86_64-w64-mingw32-gcc ... -Wl,--out-implib,dll\libzstd.dll.a -shared ... -o dll\libzstd.dll

zstd/lib$ ls *.dll
dlllibzstd.dll
```

Seems like a slash instead of a backslash should work everywhere.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -197,10 +197,10 @@ libzstd.a: $(ZSTD_OBJ)
 
 ifneq (,$(filter Windows%,$(TARGET_SYSTEM)))
 
-LIBZSTD = dll\libzstd.dll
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks @dscheg !

Issue #2350, opened 2020-10-12
compressionLevel in ZSTD_createCDict*() functions is not clamped


animalize:
When using `ZSTD_CCtx_setParameter()` function to set compressionLevel, the value is clamped:
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L559-L567
```c
    case ZSTD_c_compressionLevel : {
        FORWARD_IF_ERROR(ZSTD_cParam_clampBounds(param, &value), "");
        if (value == 0)
            CCtxParams->compressionLevel = ZSTD_CLEVEL_DEFAULT; /* 0 == default */
        else
            CCtxParams->compressionLevel = value;
        if (CCtxParams->compressionLevel >= 0) return (size_t)CCtxParams->compressionLevel;
        return 0;  /* return type (size_t) cannot represent negative values */
    }
```

But in `ZSTD_createCDict*()` functions, the compressionLevel is not clamped:
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3591-L3611
```c
ZSTD_CDict* ZSTD_createCDict(const void* dict, size_t dictSize, int compressionLevel)
{
    ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
    ZSTD_CDict* const cdict = ZSTD_createCDict_advanced(dict, dictSize,
                                                  ZSTD_dlm_byCopy, ZSTD_dct_auto,
                                                  cParams, ZSTD_defaultCMem);
    if (cdict)
        cdict->compressionLevel = (compressionLevel == 0) ? ZSTD_CLEVEL_DEFAULT : compressionLevel;
    return cdict;
}

ZSTD_CDict* ZSTD_createCDict_byReference(const void* dict, size_t dictSize, int compressionLevel)
{
    ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize);
    ZSTD_CDict* const cdict = ZSTD_createCDict_advanced(dict, dictSize,
                                     ZSTD_dlm_byRef, ZSTD_dct_auto,
                                     cParams, ZSTD_defaultCMem);
    if (cdict)
        cdict->compressionLevel = (compressionLevel == 0) ? ZSTD_CLEVEL_DEFAULT : compressionLevel;
    return cdict;
}
```

[...]

PR #2351, opened 2020-10-12
compressionLevel in ZSTD_createCDict*() functions should be clamped

animalize:
Issue https://github.com/facebook/zstd/issues/2350


animalize:
The test-cases cannot be compiled, my C language skills are not enough.

Please create another PR.

Issue #2353, opened 2020-10-12
Optimize ZSTD_compress2() with ZSTD_c_stableOutBuffer and ZSTD_stableInBuffer


terrelln:
* Add `ZSTD_c_stableOutBuffer`, which tells the compressor that the output buffer will never change.
* Add `ZSTD_c_stableInBuffer`, which tells the compressor that the input buffer will never change. This is subtle. You may think that `ZSTD_e_end` is enough. But that isn't a strong enough guarantee when streaming can't complete in one pass (because the output buffer is too small for example). Passing `ZSTD_e_end` just guarantees that you pass the same *unconsumed* bytes, it doesn't say anything about the bytes already consumed.
* Use `ZSTD_c_stableOutBuffer` to take the single-thread shortcut even when the output buffer isn't `ZSTD_compressBound(srcSize)`.
* Use `ZSTD_c_stableInBuffer` to avoid allocating a window buffer.
* Use `ZSTD_c_stableOutBuffer` to compress directly to the output buffer, even when not guaranteed to be large enough.
* Thread these optimizations through to zstdmt.
* Set both parameters in `ZSTD_compress2()` (and reset them to their original values after).

terrelln:
One consideration for `ZSTD_c_stableInBuffer`. It would be possible to allow `size` to grow. E.g. the user could add more input to the same buffer. I've chosen to not allow that, for API simplicity, but it would be possible to add.

Additionally, what should we do if `ZSTD_c_stableInBuffer` is set, but `ZSTD_e_end` is not used. Should we error? I'm thinking of requiring `ZSTD_e_end` to be set for now.

Cyan4973:
cc @terrelln : is it complete ?

terrelln:
The only thing remaining is to use this info in zstdmt. But, that doesn't need to be in this release.

PR #2354, opened 2020-10-12
Add ZSTD_c_stable{In,Out}Buffer and optimize when set

terrelln:
This completes all the tasks in Issue #2353 except hooking up zstdmt. See each commit for summaries. Each one hooks up / optimizes one thing.

This saves allocations in `ZSTD_compress2()`, and saves `memcpy()` calls when `dstCapacity < ZSTD_compressBound(srcSize)`.

Let me know what you think of this methodology. I'm happy with everything except in `ZSTD_compress2()`, where I temporarily enable stable in/out buffers, and then reset to the original values. This could be accomplished with a different flush value instead, like `ZSTD_e_singlePass`, which modifies the `appliedParams` to stable in/out only. But, then that would be part of the stable API for users.

I've added unit tests for: ZSTD_compress2(), stable in & out, only stable in, only stable out.

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -336,6 +336,12 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
  * In which case, resize it down to free some memory */
 #define ZSTD_WORKSPACETOOLARGE_MAXDURATION 128
 
+/* Controls whether the input/output buffer is buffered or stable. */
+typedef enum {
+    ZSTD_bm_buffered = 0,  /* Buffer the input/output */
+    ZSTD_bm_stable = 1     /* ZSTD_inBuffer/ZSTD_outBuffer is stable */
+} ZSTD_bufferMode_e;
+
 
 /*-*******************************************
 *  Private declarations
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1352,7 +1352,9 @@ size_t ZSTD_estimateCStreamSize_usingCCtxParams(const ZSTD_CCtx_params* params)
                 ZSTD_getCParamsFromCCtxParams(params, ZSTD_CONTENTSIZE_UNKNOWN, 0, ZSTD_cpm_noAttachDict);
         size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << cParams.windowLog);
         size_t const inBuffSize = ((size_t)1 << cParams.windowLog) + blockSize;
-        size_t const outBuffSize = ZSTD_compressBound(blockSize) + 1;
+        size_t const outBuffSize = (params->outBufferMode == ZSTD_bm_buffered)
+                ? ZSTD_compressBound(blockSize) + 1
+                : 0;
 
         return ZSTD_estimateCCtxSize_usingCCtxParams_internal(
             &cParams, &params->ldmParams, 1, inBuffSize, outBuffSize,
@@ -1598,7 +1600,9 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
         size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, windowSize);
         U32    const divider = (params.cParams.minMatch==3) ? 3 : 4;
         size_t const maxNbSeq = blockSize / divider;
-        size_t const buffOutSize = (zbuff==ZSTDb_buffered) ? ZSTD_compressBound(blockSize)+1 : 0;
+        size_t const buffOutSize = (zbuff == ZSTDb_buffered && params.outBufferMode == ZSTD_bm_buffered)
+                ? ZSTD_compressBound(blockSize) + 1
+                : 0;
         size_t const buffInSize = (zbuff==ZSTDb_buffered) ? windowSize + blockSize : 0;
         size_t const maxNbLdmSeq = ZSTD_ldm_getMaxNbSeq(params.ldmParams, blockSize);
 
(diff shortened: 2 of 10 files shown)


terrelln:
One consideration for ZSTD_c_stableInBuffer. It would be possible to allow size to grow. E.g. the user could add more input to the same buffer. I've chosen to not allow that, for API simplicity, but it would be possible to add.

Additionally, what should we do if ZSTD_c_stableInBuffer is set, but ZSTD_e_end is not used. Should we error? I'm thinking of requiring ZSTD_e_end to be set for now.

lib/compress/zstd_compress.c
```diff
@@ -4134,10 +4224,14 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
         if (cctx->cdict)
             params.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
         DEBUGLOG(4, "ZSTD_compressStream2 : transparent init stage");
-        if (endOp == ZSTD_e_end) cctx->pledgedSrcSizePlusOne = input->size + 1;  /* auto-fix pledgedSrcSize */
+        /* auto-fix pledgedSrcSize when we know we have the entire input buffer available.
+         * NOTE: It is possible to have a stable input buffer without ZSTD_e_end.
+         * E.g. If the user wants to compress with small blocks.
+         */
+        if (endOp == ZSTD_e_end || params.inBufferMode == ZSTD_bm_stable)
```
terrelln, line 4231: This actually doesn't make sense. I will revert this change. Instead we should require `ZSTD_e_end` to be used if `ZSTD_c_stableInBuffer` is set, or support growing the `in.size` in stable mode.

terrelln:
`ZSTD_c_stableInBuffer` can now only be used with `ZSTD_e_end`.

senhuang42:
This looks really interesting! 

I know that this is more API-related work and an experimental param, but I was wondering if it might make sense to enable this as an option for the CLI? Seeing as the CLI uses `compressStream2()`, but it seems pretty reasonable to expect that file i/o represents stable input and output buffers?

terrelln:
> I know that this is more API-related work and an experimental param, but I was wondering if it might make sense to enable this as an option for the CLI? Seeing as the CLI uses compressStream2(), but it seems pretty reasonable to expect that file i/o represents stable input and output buffers?

Nope! In `fileio.c` we allocate one buffer for reading (128 KB) and one buffer for writing (`ZSTD_compressBound(128KB)`). The input file is filled, we compress, flush the output buffer, and repeat. Since we're overwriting the buffer it doesn't meet the conditions required for `ZSTD_stable{In,Out}Buffer`.

lib/zstd.h
```diff
@@ -1621,6 +1625,59 @@ ZSTDLIB_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const void* pre
...
+ *
+ * WARNING: The data in the ZSTD_inBuffer in the range [dst, dst + pos) MUST
+ * not be modified during compression or you will get data corruption. This
+ * is because zstd needs to reference data in the ZSTD_inBuffer to find
+ * matches. Normally zstd maintains its own window buffer for this purpose,
+ * but passing this flag tells zstd to use the user provided buffer.
+ */
+#define ZSTD_c_stableInBuffer ZSTD_c_experimentalParam9
+
+/* ZSTD_c_stableOutBuffer
```
Cyan4973, line 1661: I was reading the definition of this parameter,
and considering : is it really about being "stable" ?

In contrast with the input buffer, it doesn't really matter that the output buffer is not flushed between invocations.
It can. Which means, the output buffer doesn't have to be "stable".

What matters is the pledge that the output buffer should be "large enough" to not require an intermediate buffer.

In which case, maybe the name of the parameter could reflect that property instead.

lib/common/zstd_internal.h
```diff
@@ -336,6 +336,12 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
  * In which case, resize it down to free some memory */
 #define ZSTD_WORKSPACETOOLARGE_MAXDURATION 128
 
+/* Controls whether the input/output buffer is buffered or stable. */
+typedef enum {
+    ZSTD_bm_buffered = 0,  /* Buffer the input/output */
+    ZSTD_bm_stable = 1     /* ZSTD_inBuffer/ZSTD_outBuffer is stable */
```
Cyan4973, line 342: The question below on the name of parameter `ZSTD_c_stableOutBuffer`
could have an impact on this name too.
In which case, `ZSTD_bm_stable` should rather insist on the _absence_ of intermediate buffer,
which is a consequence of the caller's pledge (`stable` for `input`, `largeEnough` for `output`).
`ZSTD_bm_unbuffered` comes to mind.

Cyan4973:
Mostly one opened question about the name of a (public) parameter

lib/zstd.h
terrelln: > What matters is the pledge that the output buffer should be "large enough" to not require an intermediate buffer.

Thats not quite it either. For correctness, it either:

* Has enough space remaining to not require an intermediate buffer
* Or the space remaining never increases and the user is okay with the `dstSizeTooSmall` error (so if `dstSizeTooSmall` is returned, it will always be returned).

I've opted to require stability, because that is a simpler contract. But we could loosen the requirements.

If you want to loosen the requirements, what do you think a good parameter name would be?

Cyan4973 (approved):

lib/zstd.h
naav: It's not clear to me how one is supposed to use `ZSTD_c_stableOutBuffer` with streaming compression, since the output buffer has to be flushed eventually, and hence `output.pos` reset to 0, otherwise any compression loop eventually reaches `dstSizeTooSmall`. Am I missing something? 

I rebuilt libzstd without the output buffer stability check and streaming compression with a stable output buffer works exactly as expected on some test data. I am not familiar with zstd internals so unsure whether this is the correct way of proceeding however. Happy to submit a PR otherwise.

terrelln: @naav `ZSTD_c_stableOutBuffer` is not useful for general streaming compression. It is useful in two scenarios:

1. Implementing a non-streaming interface optimally using the streaming compression interface. E.g. `ZSTD_compress2()` is implemented using `ZSTD_compressStream2()` together with `ZSTD_c_stableInBuffer` and `ZSTD_c_stableOutBuffer`.
2. It can also be useful if you're streaming the input buffer, but not streaming the output buffer. E.g. the output buffer is `mmaped()`. Though this might not be particularly useful, since the more useful buffer to mmap would be the input buffer.

PR #2357, opened 2020-10-14
Fix nondeterministic LDM behavior in multithreading

senhuang42:
@terrelln is currently working on more robust nondeterminism tests. One of the failures that showed up was in the LDM. 

The investigation revelaed that the LDM backwards match counting causes the nondeterminism. Essentially, the backwards match counting function will stop if it reaches the beginning of the circular buffer. However, we shouldn't stop if we reach the beginning - we should keep counting from the very end of the circular buffer once we reach the beginning since the backwards match might continue from the back. Note that the forward match counter already has this functionality.

The tests are yet to be merged, but if you want to check it out you can observe that if you run:
```
git remote add terrelln https://github.com/terrelln/zstd
git fetch terrelln
git checkout T77290931
cd tests
make -j zstreamtest
./zstreamtest --newapi -s5992
```
it will fail.

If you apply this patch then run the tests, it will succeed.

### DIFF
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -161,17 +161,15 @@ static size_t ZSTD_ldm_countBackwardsMatch(
  *  On reaching `pMatchBase`, start counting from mEnd */
 static size_t ZSTD_ldm_countBackwardsMatch_2segments(
             const BYTE* pIn, const BYTE* pAnchor,
-            const BYTE* pMatch, const BYTE* pMatchBase, const BYTE* mEnd)
+            const BYTE* pMatch, const BYTE* pMatchBase, const BYTE* dictStart, const BYTE* mEnd)
 {
-    const BYTE* const vBegin = MAX(pIn - (pMatch - pMatchBase), pAnchor);
-    size_t const matchLength = ZSTD_ldm_countBackwardsMatch(pIn, vBegin, pMatch, pMatchBase);
-    if (pMatch - matchLength != pMatchBase) {
+    size_t const matchLength = ZSTD_ldm_countBackwardsMatch(pIn, pAnchor, pMatch, pMatchBase);
+    if (pMatch - matchLength != pMatchBase || pMatchBase == dictStart) {
         return matchLength;
     }
     DEBUGLOG(7, "ZSTD_ldm_countBackwardsMatch_2segments: found a 2-parts backwards match (current length==%zu)", matchLength);
-    DEBUGLOG(7, "distance from pMatch to start = %zi", pMatch - pMatchBase);
     DEBUGLOG(7, "final backwards match length = %zu", matchLength + ZSTD_ldm_countBackwardsMatch(pIn - matchLength, pAnchor, mEnd, pMatchBase));
-    return matchLength + ZSTD_ldm_countBackwardsMatch(pIn - matchLength, pAnchor, mEnd, pMatchBase);
+    return matchLength + ZSTD_ldm_countBackwardsMatch(pIn - matchLength, pAnchor, mEnd, dictStart);
 }
 
 /** ZSTD_ldm_fillFastTables() :
@@ -350,7 +348,7 @@ static size_t ZSTD_ldm_generateSequences_internal(
                     }
                     curBackwardMatchLength =
                         ZSTD_ldm_countBackwardsMatch_2segments(ip, anchor, pMatch,
-                                                               lowMatchPtr, dictEnd);
+                                                               lowMatchPtr, dictStart, dictEnd);
                     curTotalMatchLength = curForwardMatchLength +
                                           curBackwardMatchLength;
                 } else { /* !extDict */
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -143,10 +143,10 @@ static void ZSTD_ldm_makeEntryAndInsertByTag(ldmState_t* ldmState,
  *  We count only bytes where pMatch >= pBase and pIn >= pAnchor. */
 static size_t ZSTD_ldm_countBackwardsMatch(
             const BYTE* pIn, const BYTE* pAnchor,
-            const BYTE* pMatch, const BYTE* pBase)
+            const BYTE* pMatch, const BYTE* pMatchBase)
 {
     size_t matchLength = 0;
-    while (pIn > pAnchor && pMatch > pBase && pIn[-1] == pMatch[-1]) {
+    while (pIn > pAnchor && pMatch > pMatchBase && pIn[-1] == pMatch[-1]) {
         pIn--;
         pMatch--;
         matchLength++;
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_ldm.c
```diff
@@ -329,8 +349,8 @@ static size_t ZSTD_ldm_generateSequences_internal(
                         continue;
                     }
                     curBackwardMatchLength =
-                        ZSTD_ldm_countBackwardsMatch(ip, anchor, pMatch,
-                                                     lowMatchPtr);
+                        ZSTD_ldm_countBackwardsMatch_2segments(ip, anchor, pMatch,
+                                                               lowMatchPtr, dictEnd);
```
Cyan4973, line 353: This likely requires another parameter like `dictBegin`

lib/compress/zstd_ldm.c
```diff
@@ -154,6 +154,26 @@ static size_t ZSTD_ldm_countBackwardsMatch(
...
+{
+    const BYTE* const vBegin = MAX(pIn - (pMatch - pMatchBase), pAnchor);
+    size_t const matchLength = ZSTD_ldm_countBackwardsMatch(pIn, vBegin, pMatch, pMatchBase);
+    if (pMatch - matchLength != pMatchBase) {
+        return matchLength;
+    }
+    DEBUGLOG(7, "ZSTD_ldm_countBackwardsMatch_2segments: found a 2-parts backwards match (current length==%zu)", matchLength);
+    DEBUGLOG(7, "distance from pMatch to start = %zi", pMatch - pMatchBase);
+    DEBUGLOG(7, "final backwards match length = %zu", matchLength + ZSTD_ldm_countBackwardsMatch(pIn - matchLength, pAnchor, mEnd, pMatchBase));
+    return matchLength + ZSTD_ldm_countBackwardsMatch(pIn - matchLength, pAnchor, mEnd, pMatchBase);
```
Cyan4973, line 174: `pMatchBase` can't be the same for both memory segments.
This is likely incorrect.

terrelln (changes requested):
You should think through the 3 cases and ensure the code works in all 3:
1. Match is in the extDict - only one count
2. Match is in the prefix & doesn't reach `base + dictLimit` - only one count
3. Match is in the prefix & reaches `base + dictLimit` - two counts

lib/compress/zstd_ldm.c
```diff
@@ -154,6 +154,26 @@ static size_t ZSTD_ldm_countBackwardsMatch(
...
+ *  Returns the number of bytes that match backwards from pMatch,
+ *  even with the backwards match spanning 2 different segments.
+ *
+ *  On reaching `pMatchBase`, start counting from mEnd */
+static size_t ZSTD_ldm_countBackwardsMatch_2segments(
+            const BYTE* pIn, const BYTE* pAnchor,
+            const BYTE* pMatch, const BYTE* pMatchBase, const BYTE* mEnd)
+{
+    const BYTE* const vBegin = MAX(pIn - (pMatch - pMatchBase), pAnchor);
+    size_t const matchLength = ZSTD_ldm_countBackwardsMatch(pIn, vBegin, pMatch, pMatchBase);
```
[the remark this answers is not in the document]
senhuang42: Yeah, you're right - I thought I was taking care of the case where the entire backward match was in extDict, but that would've been automatically taken care of in the following call to `ZSTD_ldm_countBackwardsMatch()`

lib/compress/zstd_ldm.c
```diff
@@ -154,6 +154,24 @@ static size_t ZSTD_ldm_countBackwardsMatch(
...
+static size_t ZSTD_ldm_countBackwardsMatch_2segments(
+            const BYTE* pIn, const BYTE* pAnchor,
+            const BYTE* pMatch, const BYTE* pMatchBase, const BYTE* dictStart, const BYTE* mEnd)
+{
+    size_t const matchLength = ZSTD_ldm_countBackwardsMatch(pIn, pAnchor, pMatch, pMatchBase);
+    if (pMatch - matchLength != pMatchBase || pMatchBase == dictStart) {
+        return matchLength;
+    }
+    DEBUGLOG(7, "ZSTD_ldm_countBackwardsMatch_2segments: found a 2-parts backwards match (current length==%zu)", matchLength);
+    DEBUGLOG(7, "final backwards match length = %zu", matchLength + ZSTD_ldm_countBackwardsMatch(pIn - matchLength, pAnchor, mEnd, pMatchBase));
```
terrelln, line 171: This `DEBUGLOG()` doesn't have the correct arguments. I recommend saving the result of the next line in a variable instead of calling it twice.

lib/compress/zstd_ldm.c
```diff
@@ -154,6 +154,24 @@ static size_t ZSTD_ldm_countBackwardsMatch(
...
 }
 
+/** ZSTD_ldm_countBackwardsMatch_2segments() :
+ *  Returns the number of bytes that match backwards from pMatch,
+ *  even with the backwards match spanning 2 different segments.
+ *
+ *  On reaching `pMatchBase`, start counting from mEnd */
+static size_t ZSTD_ldm_countBackwardsMatch_2segments(
+            const BYTE* pIn, const BYTE* pAnchor,
+            const BYTE* pMatch, const BYTE* pMatchBase, const BYTE* dictStart, const BYTE* mEnd)
```
terrelln, line 164: nit: `mEnd` isn't the right name for this variable. `dictEnd` maybe?

senhuang42:
Yeah, I needed to think through this a few more times, and let the automated tests tell me I was wrong on a couple things.

> 1. Match is in the extDict - only one count
In this case, `dictStart` is equal to `pMatchBase`, and we will always return the result of `ZSTD_ldm_countBackwardsMatch()`
> 2. Match is in the prefix & doesn't reach `base + dictLimit` - only one count
In this case, `pMatchBase = base+dictLimit`, and `ZSTD_ldm_countBackwardsMatch()` returns a `matchLength` such that `pMatch - matchLength > pMatchBase`, and we return early
> 3. Match is in the prefix & reaches `base + dictLimit` - two counts
In this case, `matchLength` should be exactly `pMatch - matchBase == pMatch - (base + dictLimit)`, since the call to `ZSTD_ldm_countBackwardsMatch()` should terminate because the check `pMatch > pBase` fails. So, then we should start counting backwards again, this time with our `pMatchBase == dictBase + lowLimit == dictStart` and `pMatch == dictEnd`

lib/compress/zstd_ldm.c
senhuang42: I agree, the variable names are very confusing. I'm gonna go through and change some stuff for clarity

senhuang42:
Out of curiosity, `ZSTD_count()` looks very different than `ZSTD_ldm_countBackwardsMatch()`. The code inside `ZSTD_count()` looks pretty bizzare - I'm guessing this is a super-fast version of counting matching bytes? Is there any reason that the `ZSTD_ldm_countBackwardsMatch()` doesn't have the same sort of structure?

terrelln:
> Out of curiosity, ZSTD_count() looks very different than ZSTD_ldm_countBackwardsMatch(). The code inside ZSTD_count() looks pretty bizzare - I'm guessing this is a super-fast version of counting matching bytes? Is there any reason that the ZSTD_ldm_countBackwardsMatch() doesn't have the same sort of structure?

`ZSTD_count()` is optimized to minimize branches, and to be as quick as possible. It is extremely hot during compression. `ZSTD_ldm_countBackwardsMatch()` isn't as hot, and no one has taken the time to optimize it. It could be written in a similar method as `ZSTD_count()`.

If you want to take a stab at it, you could use perf to figure out how hot `ZSTD_ldm_countBackwardsMatch()` is when LDM is enabled, and if it takes up a significant amount of CPU, write `ZSTD_ldm_countBackwardsMatch()` using the same method. If you want I can explain how `ZSTD_count()` works sometime.

terrelln (approved):
LGTM once the tests pass!

PR #2359, opened 2020-10-15
[minor] Fix some /tests/fuzz Makefile annoyances

senhuang42:
Using a clean checkout of zstd, running `make` in tests/fuzz/ will currently fail, apparently unless you also run `make regressiontest` beforehand. I'm no Makefile expert, but I usually expect `make` to just work. So, this PR does three things:

- Adds `libregression.a` to the build targets in `fuzz/Makefile`
- Make clean now properly removes all the fuzz build targets and prints the usual message
- .gitignore is updated to include all the appropriate build artifacts, as well as transient ones that might stick around if you abort in the middle of `make`.

### DIFF
--- a/tests/fuzz/.gitignore
+++ b/tests/fuzz/.gitignore
@@ -13,7 +13,11 @@ simple_round_trip
 stream_decompress
 stream_round_trip
 zstd_frame_info
+decompress_dstSize_tooSmall
+fse_read_ncount
 fuzz-*.log
+rt_lib_*
+d_lib_*
 
 # misc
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Cyan4973:
Good cleaning !

The failed test seems unrelated to this PR: it's a race condition in `zlibwrapper` build script.

terrelln (approved):
Awesome!

Generally I use `./fuzz.py build all` to build the fuzzers.

PR #2361, opened 2020-10-15
[travis CI] Change make -j all check to make -j all && make check

senhuang42:
`make -j all check` isn't really a valid way to invoke `make check`, and is the likely cause test failures like: https://travis-ci.org/github/facebook/zstd/jobs/736047280

Test: I'll do multiple re-runs of the travis CI test on this PR to see if fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,14 +76,14 @@ matrix:
 
     - name: Minimal Decompressor Macros    # ~5mn
       script:
-        - make clean
-        - make -j all check ZSTD_LIB_MINIFY=1 MOREFLAGS="-Werror"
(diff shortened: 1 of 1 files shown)


Cyan4973 (changes requested):
The variable `ZSTD_LIB_MINIFY` and `MOREFLAGS` are actually important during `make all`, since they are active at compilation time.

Cyan4973 (approved):

senhuang42:
I've re-ran the Travis job a few times, no failures, while on other PRs the test currently seems to fail. So this seems the issue being addressed was the actual cause of the failures.

PR #2362, opened 2020-10-15
Fix long distance matcher OSS-fuzz issue

senhuang42:
#2341 introduced some issues in OSS-Fuzz. This PR fixes those issues.

The root cause was that sometimes we could begin an LDM compression job without having a "clean" `rawSeqStore` to begin with - i.e., the `posInSequence` value was not set to 0 during `ZSTD_referenceExternalSequences()`, which could cause the opt parser to integrate some LDMs that weren't in the right position.

Luckily, the blame PR already included an assert made debugging this much easier: `assert(optLdm->seqStore.posInSequence <= currSeq.litLength + currSeq.matchLength)`

Now, it's not entirely clear to me what the best way is to test/catch this specific case earlier - would appreciate any suggestions

1) I don't think `assert(pos == 0 && posInSequence == 0)` in the code that follows the call to `ZSTD_referenceExternalSequences()` is useful, since `ZSTD_referenceExternalSequences()` is a simple function that sets some values. For example, similar-ish functions like `ZSTD_CCtx_setParameter(zc, ZSTD_c_compressionLevel, cLevel)` doesn't expect a check/assert like `assert(zc->compressionLevel == cLevel)` after calling them, so that type of `assert()` doesn't seem so useful. 

Perhaps if there's some point/function in the compression codepath where we do a one-time validation of all parameters in the `cctx` prior to compressing, an assert like this could belong there? I couldn't find a particular place for that though.

2) It doesn't seem like a good idea to add a unit test that tests specifically for the input that caused this issue, cause then we'd have to apply that principle in general, and have unit tests for really specific inputs that might cause other issues, which would go on into infinity.

3) The MT-specific tests in `fuzzer.c` mostly deal with higher-level parameters, and `rawSeqStore.posInSequence` seems too low-level of a parameter to directly test (it is in `compress_internal.h` after all)

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2374,7 +2374,11 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
     /* Assert that we have correctly flushed the ctx params into the ms's copy */
     ZSTD_assertEqualCParams(zc->appliedParams.cParams, ms->cParams);
     if (srcSize < MIN_CBLOCK_SIZE+ZSTD_blockHeaderSize+1) {
-        ZSTD_ldm_skipSequences(&zc->externSeqStore, srcSize, zc->appliedParams.cParams.minMatch);
+        if (zc->appliedParams.cParams.strategy >= ZSTD_btopt) {
+            ZSTD_ldm_skipRawSeqStoreBytes(&zc->externSeqStore, srcSize);
+        } else {
+            ZSTD_ldm_skipSequences(&zc->externSeqStore, srcSize, zc->appliedParams.cParams.minMatch);
+        }
         return ZSTDbss_noCompress; /* don't even attempt compression below a certain srcSize */
     }
     ZSTD_resetSeqStore(&(zc->seqStore));
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2853,6 +2853,7 @@ size_t ZSTD_referenceExternalSequences(ZSTD_CCtx* cctx, rawSeq* seq, size_t nbSe
     cctx->externSeqStore.size = nbSeq;
     cctx->externSeqStore.capacity = nbSeq;
     cctx->externSeqStore.pos = 0;
+    cctx->externSeqStore.posInSequence = 0;
     return 0;
 }
 
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -555,10 +555,7 @@ static rawSeq maybeSplitSequence(rawSeqStore_t* rawSeqStore,
     return sequence;
 }
 
-/* ZSTD_ldm_skipRawSeqStoreBytes():
- * Moves forward in rawSeqStore by nbBytes, updating fields 'pos' and 'posInSequence'.
- */
-static void ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore_t* rawSeqStore, size_t nbBytes) {
+void ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore_t* rawSeqStore, size_t nbBytes) {
     U32 currPos = (U32)(rawSeqStore->posInSequence + nbBytes);
     while (currPos && rawSeqStore->pos < rawSeqStore->size) {
         rawSeq currSeq = rawSeqStore->seq[rawSeqStore->pos];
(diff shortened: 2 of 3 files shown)


terrelln:
How does `posInSequence != 0` happen? Would it be possible to trigger the condition to have `posInSequence != 0`, then call compression, to hopefully trigger data corruption?

senhuang42:
> How does `posInSequence != 0` happen? Would it be possible to trigger the condition to have `posInSequence != 0`, then call compression, to hopefully trigger data corruption?

In each case of failure, we have the following situation:
- The LDM seqStore for the job contains sequences that total to `ldmSizeStoreSize = (all litLengths + all matchLengths)`.
- In `ZSTD_ldmBlockCompress()`, we call `ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore, srcSize)` after every call to our `blockCompressor()`, where `srcSize = blockSize`.
- I'd expect `ldmSeqStoreSize == (sum of all blockSizes)`
- However, it appears that somehow, certain inputs can produce: `(sum of all blockSizes) == ldmSeqStoreSize - 1`
- So then, over the course of the job, we effectively are calling `ZSTD_ldm_skipRawSeqStoreBytes(rawSeqStore, (sum of all blockSizes) - 1)`, which means that at the end of the job, `posInSequence` reaches the very end of the last match in the LDM seqStore, but doesn't go that one extra byte that would cause us to go to the next (non-existent) `pos`, which equals `size`, which would cause `posInSequence` to reset.
  - This is actually fine in general if we were to correctly reset `posInSequence`, since it would only occur at the end of each job

I'm still not sure how possible it is to replicate this situation with a minimal input that would fit inside a test case, since I'm not quite sure what property of the input would allow this to happen.

terrelln:
> But actually, we can have (sum of all blockSizes) == ldmSeqStoreSize - 1

How? Is this because of the last literals? It would be worth understanding this, because it doesn't quite make sense to me.

senhuang42:
> > But actually, we can have (sum of all blockSizes) == ldmSeqStoreSize - 1
> 
> How? Is this because of the last literals? It would be worth understanding this, because it doesn't quite make sense to me.

Aha! In `ZSTD_buildSeqStore()` we have that if `srcSize < MIN_CBLOCK_SIZE+ZSTD_blockHeaderSize+1`, we do not compress the block and return immediately. This was what was giving me the impression that we only compressed `ldmSeqStoreSize - 1` bytes, since I was counting the number of bytes passed into `blockCompressor()`, and that is apparently not always all of the bytes.

So in addition to resetting `posInSequence` at the end of each job, we should also call `ZSTD_ldm_skipRawSeqStoreBytes()` on blocks that are too small. One subtlety here though is that on `noCompress` blocks, we already invoke `ZSTD_ldm_skipSequences()` which actually mutates the sequences, so is not really compatible with `ZSTD_ldm_skipRawSeqStoreBytes()`, so we should invoke either or.

terrelln:
@senhuang42 could you trigger the bug like this?

1. Enable multithreading - 1 worker
2. Enable LDM
3. Enable btopt
4. Enable checksum
5. Compress once with `ZSTD_e_flush` with a job of `128KB + 5` bytes. This could be a block of all `0` to ensure the last byte is part of a sequence (not last literals).
6. Compress again with `ZSTD_e_end`. This should be normal bytes (all `0`s wouldn't expose the bug, since starting at position 5 wouldn't be corruption). Maybe uncompressible bytes.
7. Decompress and the checksum should fail to match before this patch.

You might need to fiddle with stuff a bit, but I think this will trigger the bug.

Cyan4973 (approved):

Cyan4973:
If I understand correctly,
this PR still needs to add a test case before merge ?

senhuang42:
> If I understand correctly,
> this PR still needs to add a test case before merge ?

I'm currently having some trouble reproducing the exact bug. In particular, right now I'm calling a `memset(src, 'a', 5 KB)` and `memset(src+123 KB+2, 'a', 5 KB)` and then compressing with `ZSTD_e_flush` doesn't actually seem to generate a 5K LDM at position `123 KB + 2`.

If it's fine with you, I think it'd make sense to merge this PR (to stop any more OSS-Fuzz failures) and I can put up another work-in-progress PR with the test case that I can get some comments on separately?

Cyan4973:
If I were you, I would not use `memset()`.
I would generate a noise, typically with `datagen`, and then copy a segment of noise at intended position.
Using a repeated character `aaa` introduces issues which make it a non ideal candidate for a `ldm` match finder.

That being said, I agree it can be done in a separate PR.

Issue #2358, opened 2020-10-14
Meson build fails on windows


o-mdr:
**Describe the bug**
Meson build fails on Windows platform

**To Reproduce**
Steps to reproduce the behaviour:
1. On Windows platform with MSVC tools run a developer command line (this comes with MS Visual Studio. NB: NOT vscode, but full VS)
2. Download source code of zstd (main or release 1.4.5)
3. In _developer_ console execute first time setup
```
cd build\meson
meson setup builddir
---
C:\Users\om\Downloads\zstd-dev\build\meson>meson setup builddir
The Meson build system
Version: 0.55.3
Source dir: C:\Users\om\Downloads\zstd-dev\build\meson
Build dir: C:\Users\om\Downloads\zstd-dev\build\meson\builddir
Build type: native build
Project name: zstd
Project version: DUMMY
C compiler for the host machine: cl (msvc 19.27.29111)
C linker for the host machine: link link 14.27.29111.0
C++ compiler for the host machine: cl (msvc 19.27.29111)
C++ linker for the host machine: link link 14.27.29111.0
Host machine cpu family: x86
Host machine cpu: x86
Program GetZstdLibraryVersion.py found: YES
Message: Project version is now: 1.4.6
Library m found: NO
Run-time dependency threads found: YES
Did not find pkg-config by name 'pkg-config'
Found Pkg-config: NO
Found CMake: C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.EXE (3.17.20032601)
Run-time dependency zlib found: NO (tried pkgconfig and cmake)
Run-time dependency liblzma found: NO (tried pkgconfig and cmake)
Run-time dependency liblz4 found: NO (tried pkgconfig and cmake)
Message: Enable legacy support back to version 0.5
Message: Enable multi-threading support
Windows resource compiler: Microsoft (R) Windows (R) Resource Compiler Version 10.0.10011.16384
Program ../InstallSymlink.py found: YES (C:\Program Files\Meson\meson.exe runpython C:\Users\om\Downloads\zstd-dev\build\meson\programs\../InstallSymlink.py)
Build targets in project: 5

Found ninja-1.10.0 at "C:\Program Files\Meson\ninja.EXE"
```

4. Compile code

```
meson compile -C builddir

---
C:\Users\om\Downloads\zstd-dev\build\meson>meson compile -C builddir
[...]

Issue #2314, opened 2020-09-18
Shortcut for ZSTD_compressBound() is slower when compiled with MSVC2019


animalize:
I wrote a [pyzstd](https://github.com/animalize/pyzstd) module for Python, it has a "rich memory mode" that the size of output buffer is provided by `ZSTD_compressBound()` function.

The document said it should faster, but I observed that when the code is compiled with MSVC2019, "rich memory mode" mode is slower.

This is a [test code](https://github.com/animalize/misc/blob/master/richmem.c), it test two cases:
* the size of output buffer is `ZSTD_compressBound(srcSize) - 1`
* the size of output buffer is `ZSTD_compressBound(srcSize)`

On Windows 10 64-bit, MSVC2019:
```
read 545863680 bytes.
ZSTD_compressBound() value: 547995960

ZSTD_compressBound() - 1
output size: 286842128, time: 9270 ms

ZSTD_compressBound()
output size: 286842126, time: 14669 ms
```
`ZSTD_compressBound()` output buffer size is slower a lot.

FYI, compress the same file using pyzstd module (compiled with MSVC2019):
* `ZSTD_compressBound()-1` consumes 4.9 seconds
* `ZSTD_compressBound()` consumes 6.9 seconds
* when the output buffer grows gradually, consumes 5.6 seconds

While compiled with GCC 9.3.0, `ZSTD_compressBound()` output buffer size is faster as expected:
On Windows 10 64-bit, Cygwin64 (gcc 9.3.0):
```
ZSTD_compressBound() - 1
output size: 286842128, time: 4530 ms

ZSTD_compressBound()
output size: 286842126, time: 4376 ms
```

On Windows 10 WSL2, Ubuntu 9.3.0-17ubuntu1~20.04:
```
ZSTD_compressBound() - 1
output size: 286842128, time: 4578125 ms (seems microsecond)

ZSTD_compressBound()
output size: 286842126, time: 4390625 ms (seems microsecond)
```

animalize:
I `#define DEBUGLEVEL 5`, then compare the output of two builds (MSVC2019 / Cygwin64), they are exactly same except for memory addresses.

Maybe it's not a problem of zstd code.
```
compress\zstd_compress.c: ZSTD_compressStream2, endOp=2  
compress\zstd_compress.c: ZSTD_getCParams_internal (cLevel=3) 
compress\zstd_compress.c: ZSTD_compressStream2 : transparent init stage 
[...]

PR #2363, opened 2020-10-17
meson: Deny shared build on MSVC compiler

lzutao:
Apparently, zstd has never tested on MSVC 2019 and currently fall to build on that compiler.
It would make sense to gate Meson until there is proper support for that compiler.
cc #2358

This PR also fixes some minor correctness issues.

### DIFF
--- a/build/meson/lib/meson.build
+++ b/build/meson/lib/meson.build
@@ -77,8 +77,12 @@ endif
 libzstd_c_args = []
 if cc_id == compiler_msvc
   if default_library_type != 'static'
-    libzstd_sources += [windows_mod.compile_resources(
-      join_paths(zstd_rootdir, 'build/VS2010/libzstd-dll/libzstd-dll.rc'))]
+    libzstd_sources += [
+      windows_mod.compile_resources(
+        join_paths(zstd_rootdir, 'build/VS2010/libzstd-dll/libzstd-dll.rc'),
+        include_directories: include_directories(join_paths(zstd_rootdir, 'lib')),
+      )
+    ]
     libzstd_c_args += ['-DZSTD_DLL_EXPORT=1',
       '-DZSTD_HEAPMODE=0',
       '-D_CONSOLE',
--- a/build/meson/contrib/pzstd/meson.build
+++ b/build/meson/contrib/pzstd/meson.build
@@ -19,9 +19,11 @@ pzstd_sources = [
   join_paths(zstd_rootdir, 'programs/util.c'),
 ]
 
+pzstd_cpp_args = cxx.get_supported_arguments([ '-DNDEBUG', '-Wno-shadow', '-pedantic' ])
+
 pzstd = executable('pzstd',
   pzstd_sources,
-  cpp_args: [ '-DNDEBUG', '-Wno-shadow', '-pedantic' ],
+  cpp_args: pzstd_cpp_args,
   include_directories: pzstd_includes,
   dependencies: [ libzstd_dep, thread_dep ],
   install: true)
--- a/build/meson/contrib/pzstd/meson.build
+++ b/build/meson/contrib/pzstd/meson.build
@@ -11,11 +11,14 @@ zstd_rootdir = '../../../..'
 
 pzstd_includes = include_directories(join_paths(zstd_rootdir, 'programs'),
   join_paths(zstd_rootdir, 'contrib/pzstd'))
-pzstd_sources = [join_paths(zstd_rootdir, 'programs/util.c'),
+pzstd_sources = [
   join_paths(zstd_rootdir, 'contrib/pzstd/main.cpp'),
   join_paths(zstd_rootdir, 'contrib/pzstd/Options.cpp'),
   join_paths(zstd_rootdir, 'contrib/pzstd/Pzstd.cpp'),
-  join_paths(zstd_rootdir, 'contrib/pzstd/SkippableFrame.cpp')]
+  join_paths(zstd_rootdir, 'contrib/pzstd/SkippableFrame.cpp'),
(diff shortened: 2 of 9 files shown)


build/meson/lib/meson.build
```diff
@@ -77,8 +77,12 @@ endif
 libzstd_c_args = []
 if cc_id == compiler_msvc
   if default_library_type != 'static'
-    libzstd_sources += [windows_mod.compile_resources(
-      join_paths(zstd_rootdir, 'build/VS2010/libzstd-dll/libzstd-dll.rc'))]
+    libzstd_sources += [
+      windows_mod.compile_resources(
+        join_paths(zstd_rootdir, 'build/VS2010/libzstd-dll/libzstd-dll.rc'),
+        include_directories: include_directories(join_paths(zstd_rootdir, 'lib')),
```
o-mdr, line 83: Looking at the output from `meson compile -v -C builddir`, I am not sure `include_directories` are actually passed in to the resource compiler. 

Anyone else observes this issue?

lzutao: It will, but I have MSVC compiler error about syntax: <https://github.com/lzutao/zstd/runs/1264653094#step:8:130>.

o-mdr: Interesting, I cannot get beyond compiling resources on my system, whereas in your build it goes passed it and compiles the resource just fine.

That error is not about syntax, it's a linker error, basically it's saying that there were a couple of symbols observed in functions (`ZSTD_cycleLog`, `ZSTD_XXH64`, `ZDICT_trainFromBuffer_unsafe_legacy`) that were not actually defined. Could it be a missing source/header file or compiler/linker flag?

lzutao: That was a shared build. All the missing symbols were in `zstd.lib`, which was included the the `link` command.

There was also a statically build, which has syntax error: <https://github.com/lzutao/zstd/runs/1267569335#step:8:124>.

lzutao: *I am sorry for giving the wrong like*.

o-mdr: Oh, so it is, not seen that one

terrelln:
@lzutao It looks like `min` is defined as a macro, which is messing with the calls to `std::min()`. I am fairly certain the definition is coming from one of the windows includes from `platform.h`. Defining `NOMINMAX` in `programs/platform.h`, before the windows includes, should probably fix the issue.

lzutao:
Thank you, @terrelln. Your way really fixes the statically build.
The shared build still fall with [unresolved symbols error](https://github.com/lzutao/zstd/pull/1/checks?check_run_id=1270086990#step:8:121). Hope you could take a look.

At worst case, I could disable shared build for msvc.

animalize:
Could you add `/Ob3` c compile option to MSVC2019?
> Visual Studio 2019 has a new optimization option /Ob3, which turns on aggressive inlining.
> https://docs.microsoft.com/en-us/cpp/build/reference/ob-inline-function-expansion?view=vs-2019

In zstd v1.4.5, just turn on this option, no code changed, run a test code: `6.97s` -> `4.77s`.
See issue https://github.com/facebook/zstd/issues/2314 for more information.

In the upcoming zstd v1.4.6, commit https://github.com/facebook/zstd/commit/cc88eb7594d9c70ec2440ce6122f7861dbd64af2 accelerates from `6.97s` to `4.85s`, still a bit slower than `4.77s`.

This option was added in MSVC2019, if use this option in MSVC2017, it will emit a warning:
`cl : Command line warning D9002 : ignoring unknown option '/Ob3'`

lzutao:
@animalize I could. However does the same flag used anywhere else in the other build systems?
I don't want to use some custom optimization flag for meson only.
Also adding that flag would be better in another PR.

animalize:
> @animalize I could. However does the same flag used anywhere else in the other build systems?
> I don't want to use some custom optimization flags for meson only.
> Also adding that flag would be better in another PR.

A separated PR sounds reasonable.
The flag should not be used in other build systems. This problem was discovered last month.

terrelln:
> The shared build still fall with unresolved symbols error. Hope you could take a look.

Adding `ZDICTLIB_VISIBILITY` to this declaration should fix it. 

https://github.com/facebook/zstd/blob/614e446000a69acdf322e155419e9783d9c7170c/lib/dictBuilder/zdict.c#L972

Like

```c
ZDICTLIB_VISIBILITY size_t ZDICT_trainFromBuffer_unsafe_legacy(
                            void* dictBuffer, size_t maxDictSize,
                            const void* samplesBuffer, const size_t* samplesSizes, unsigned nbSamples,
                            ZDICT_legacy_params_t params);
```

lzutao:
I need some help to resolve this error: <https://github.com/facebook/zstd/pull/2363/checks?check_run_id=1275024168#step:3:1708>

Cyan4973:
I initially thought that the topic was about the shared dynamic library,
but apparently, it's larger than that:
it also applies to a variant of `zstd` CLI which would be _linked against_ the dynamic library.

The `zstd` CLI is not designed to be linked to the standard dynamic library `libzstd`.
That's because it relies on private symbols, which are not exposed by default in the library.

For the CLI to link against a shared library, the library would need to expose basically all symbols,
the same way a static library does (`zstd` CLI links correctly against the static library).

But there is a reason we only expose selected symbols by default : the rest is considered "internal stuff", and may change at any moment. So it's not good to let applications believe they can depend on these accidental features.

I would suggest to not attempt delivering this feature through `meson`.

tesuji:
Thanks for your explanation. I think Meson had a way to disable statically linking with the library. I've been away from my computer for a while. If anyone want to patch this issue, please don't hesitate.

PR #2366, opened 2020-10-19
Enable LDM by default if window size >= 128MB and strategy uses opt parser

senhuang42:
Since the LDM can no longer cause compressed size regressions, and at worst, is only a tiny bit slower than without LDM at `btopt+` compression strategies (benchmarks in #2341), let's enable long mode by default when the window size is greater than 128 MB and we're using the `btopt+` strategies.

It seems this is about as far upstream in the compression codepath as it makes sense to do this change, and even then, MT and single-thread have to be dealt with separately, and `ZSTD_makeCCtxParamsFromCParams()` must get updated as well.

Tests:
- Unit test included in `fuzzer.c` to ensure this doesn't modify the user's requested params in the `cctx`
- Tested locally to make sure that LDM gets enabled with the specified conditions.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1542,6 +1542,12 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
 
     zc->isFirstBlock = 1;
 
+    if (params.cParams.strategy >= ZSTD_btopt && params.cParams.windowLog >= 27) {
+        /* Enable LDM by default for optimal parser and window size >= 128MB */
+        DEBUGLOG(4, "LDM enabled by default (window size >= 128MB, strategy >= btopt)");
+        params.ldmParams.enableLdm = 1;
+    }
+
     if (params.ldmParams.enableLdm) {
         /* Adjust long distance matching parameters */
         ZSTD_ldm_adjustParameters(&params.ldmParams, &params.cParams);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1542,7 +1542,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
 
     zc->isFirstBlock = 1;
 
-    if (params.cParams.strategy >= ZSTD_btopt && params.cParams.windowLog >= 27) {
+    if (params.cParams.strategy >= ZSTD_btopt && params.cParams.windowLog >= 27 && params.nbWorkers == 0) {
         /* Enable LDM by default for optimal parser and window size >= 128MB */
         DEBUGLOG(4, "LDM enabled by default (window size >= 128MB, strategy >= btopt)");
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1542,6 +1552,12 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
 
     zc->isFirstBlock = 1;
 
+    if (params.cParams.strategy >= ZSTD_btopt && params.cParams.windowLog >= 27 && params.nbWorkers == 0) {
```
terrelln, line 1555: I don't like modifying parameters inside of `ZSTD_resetCCtx_internal()`. At this point they should be fixed.

terrelln (commented):
We should be able to find one place to make this change that works for both single-threaded and multi-threaded code.

We probably don't want this logic to apply to `ZSTD_compressCCtx()` and similar functions. It should only apply for the new advanced API (all streaming & `ZSTD_compress2()`). In which case we can do it when we determine parameters in the transparent initialization stage of `ZSTD_compressStream2()`.

senhuang42:
Ah - I did the change in `ZSTD_resetCCtx_internal()` because it seemed like we wanted this change to get incorporated in all possible codepaths. If we don't want that then, yeah - can definitely just be done in only `compressStream2()`. 

I'm happy to do it just for the newer APIs, but is there any reason we don't want to include the older stuff as well?

lib/compress/zstd_compress.c
```diff
@@ -209,6 +209,16 @@ static ZSTD_CCtx_params ZSTD_makeCCtxParamsFromCParams(
     /* should not matter, as all cParams are presumed properly defined */
     ZSTD_CCtxParams_init(&cctxParams, ZSTD_CLEVEL_DEFAULT);
     cctxParams.cParams = cParams;
+
+    if (cParams.strategy >= ZSTD_btopt && cParams.windowLog >= 27 && cctxParams.nbWorkers == 0) {
```
terrelln, line 213: We don't need to check `cctxParams.nbWorkers` here. It will either always be true (in this case), or always false, since we've just initialized them.

We should move this check `cParams.strategy >= ZSTD_btopt && cParams.windowLog >= 27` to a function, so we don't need to write it twice. Otherwise the two invocations could drift.

terrelln:
> I'm happy to do it just for the newer APIs, but is there any reason we don't want to include the older stuff as well?

We could include the old stuff as well. But my thoughts are:
* There is no way to enable LDM explicitly using the old APIs.
* We want to provide people reasons to migrate to the new advanced API.
* I favor simplicity over supporting the old APIs.
* We'll have to talk to @cyan4973 about this one, but I would like to move all of the `ZSTD_compress*` functions onto the new advanced API. This would have to be carefully measured and tested, and would probably warrant a new major version, but it would improve simplicity significantly.

terrelln (approved):

PR #2367, opened 2020-10-20
faster rebuild of zstd 

Cyan4973:
Building the `zstd` CLI costs a non negligible time. It's not very long, but it's not instantaneous either.

The main reason is that several units (those in `lib/`) must be rebuilt _from source_ at every compilation.
Therefore, any change anywhere requires re-compilation of the entire `lib/` portion of `zstd`.

A better solution would be compile `*.o` object files, and link them.
And btw, that's what already happens for units in `programs/`
(though they only represent a small portion of total compilation time).
But we can't do that for units in `lib/` because, by default, `*.o` objects are stored alongside their `*.c` source.
And problem is, these object files are already created for the benefit of the static library,
which unfortunately uses different compilation flags.
Specifically, the CLI enables multithreading, while the library doesn't (by default).

Having an incompressible part in compilation time leads to substantial iteration delay when testing small code modifications, which can be annoying.

This is solved in this commit, by generating `zstd` object files locally into `programs/` dir.
Now, `zstd` CLI and `libzstd` library can employ different sets of flags, without tripping over each other.
Note that it works nicely because no 2 source files have the same name anywhere in the directory tree.

Starting from this PR, modifying a source file doesn't require recompiling the entire `lib/`, just the modified files.
The recipe is also compatible with `-j` parallel build, leading to large build time reductions on multi-core systems.

Also, the new recipe is able to track `*.h` header file dependencies.
A header file modification will also trigger recompilation of dependent `*.c` source files
and only those files.

_limitations_ : this improvement is only for `zstd` and direct derivatives, such as `zstd-release`.
Special targets, such as `zstd32`, must still rebuild from source, to avoid mixing object files employing different compilation directives. This situation is identical to current `dev`.

Also, this will not track manual changes to compilation flags, such as introducing a different `DEBUGLEVEL` for example. For the new `DEBUGLEVEL` to take effect, one needs to `make clean` first, in order to ensure that the new parameter is applied to all object files. Here also, this is not different from current situation in `dev`.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -77,17 +77,19 @@ ZSTD_LEGACY_SUPPORT ?= 5
 ZSTDLEGACY_SRC :=
 ifneq ($(ZSTD_LEGACY_SUPPORT), 0)
 ifeq ($(shell test $(ZSTD_LEGACY_SUPPORT) -lt 8; echo $$?), 0)
-	ZSTDLEGACY_SRC += $(notdir $(shell ls $(ZSTDLEGACY_DIR)/*.c | $(GREP) 'v0[$(ZSTD_LEGACY_SUPPORT)-7]'))
+	ZSTDLEGACY_SRC += $(shell ls $(ZSTDLEGACY_DIR)/*.c | $(GREP) 'v0[$(ZSTD_LEGACY_SUPPORT)-7]')
 endif
 endif
 
 # Sort files in alphabetical order for reproducible builds
 ZSTDLIB_FULL_SRC = $(sort $(ZSTDLIB_CORE_SRC) $(ZSTDLEGACY_SRC) $(ZDICT_SRC))
-ZSTDLIB_LOCAL_SRC = $(notdir $(ZSTDLIB_FULL_SRC))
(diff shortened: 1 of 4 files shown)


programs/Makefile
```diff
@@ -43,41 +46,54 @@ else
...
+ZSTDLIB_COMPRESS := $(ZSTDDIR)/compress
+ZSTDLIB_DECOMPRESS := $(ZSTDDIR)/decompress
+ZDICT_DIR := $(ZSTDDIR)/dictBuilder
+ZSTDLEGACY_DIR := $(ZSTDDIR)/legacy
+
+vpath %.c $(ZSTDLIB_COMMON)
+vpath %.c $(ZSTDLIB_COMPRESS)
+vpath %.c $(ZSTDLIB_DECOMPRESS)
+vpath %.c $(ZDICT_DIR)
+vpath %.c $(ZSTDLEGACY_DIR)
```
terrelln, line 72: Neat, didn't know about this!

terrelln (approved):
Awesome! This is a big improvement!

PR #2368, opened 2020-10-22
faster rebuild of libzstd

Cyan4973:
Similar to [yesterday's PR](https://github.com/facebook/zstd/pull/2367),
this one makes it faster to rebuild `libzstd` after some minor source code change, enabling fast iterations.

The method is also similar: link `%.o` object files, instead of compiling from `%.c` source every time.
This was already the case for the static `libzstd.a`, but not for the dynamic `libzstd.so`, which had to be rebuilt entirely from source every time.
Reason is : by default, each `%.o` is generated alongside its corresponding `%.c` source file, so there is only space for one. 
And since the dynamic and static libraries do not use the same compilation flags (specifically, the dynamic library _requires_ Position Independent Code, while static _must not_ use it), only one get to produce `%.o` object file.

This is fixed in this PR, where each library type gets its own build directory where object files are stored.
These directories are `obj/dynlib` and `obj/statlib` by default.

Build directories also store `%.d` dependency files, tracking relations between units and `%.h` header files. Therefore, modifications of header files are no longer ignored, and will trigger recompilation of dependent units, and only these ones.

Finally, build directories can be manually redirected, by setting the `BUILD_DIR` variable, which will replace `obj`.

_limitations_ : this PR does not track manual changes to compilation flags, such as introducing a different `DEBUGLEVEL` for example. For the new `DEBUGLEVEL` to take effect, one needs to invoke `make clean` first, in order to ensure that the new parameter is applied to all object files. This is not different from current situation in `dev`.
This issue can be circumvented by setting a dedicated `BUILD_DIR` for a given set of parameters.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -10,13 +10,13 @@
 
 Q = $(if $(filter 1,$(V) $(VERBOSE)),,@)
 
-# When cross-compiling from linux to windows, you might
-# need to specify this as "Windows." Fedora build fails
-# without it.
+# When cross-compiling from linux to windows,
+# you might need to specify this as "Windows."
+# Fedora build fails without it.
 #
-# Note: mingw-w64 build from linux to windows does not
-# fail on other tested distros (ubuntu, debian) even
-# without manually specifying the TARGET_SYSTEM.
+# Note: mingw-w64 build from linux to windows
+# does not fail on other tested distros (ubuntu, debian)
+# even without manually specifying the TARGET_SYSTEM.
 TARGET_SYSTEM ?= $(OS)
 
 # Version numbers
@@ -167,12 +167,7 @@ CPPFLAGS  += -DZSTD_LEGACY_SUPPORT=$(ZSTD_LEGACY_SUPPORT)
 ZSTD_FULL_OBJ  := $(ZSTD_FILES:.c=.o)
 ZSTD_LOCAL_OBJ := $(notdir $(ZSTD_FULL_OBJ))
 
-vpath %.c ./common
-vpath %.c ./compress
-vpath %.c ./decompress
-vpath %.c ./dictBuilder
-vpath %.c ./legacy
-vpath %.c ./deprecated
+vpath %.c common compress decompress dictBuilder legacy deprecated
 
 ZSTD_DYNLIB_DIR := obj/dynlib
 ZSTD_DYNLIB_OBJ := $(addprefix $(ZSTD_DYNLIB_DIR)/, $(ZSTD_LOCAL_OBJ))
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -174,7 +174,10 @@ vpath %.c ./dictBuilder
 vpath %.c ./legacy
 vpath %.c ./deprecated
 
-ZSTD_DYNLIB_OBJ := $(ZSTD_LOCAL_OBJ)
+ZSTD_DYNLIB_DIR := obj/dynlib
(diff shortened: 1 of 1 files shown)


lib/Makefile
```diff
@@ -35,28 +35,28 @@ CCVER := $(shell $(CC) --version)
...
+    # Some compilers (clang) support an even more space-optimized setting.
+    CFLAGS += -Oz
 else
-	CFLAGS += -O3
+    CFLAGS += -Os
+endif
+  CFLAGS += -fno-stack-protector -fomit-frame-pointer -fno-ident \
+            -DDYNAMIC_BMI2=0 -DNDEBUG
+else
+  CFLAGS += -O3
```
terrelln, line 54: nit: The indentation is strange.

lib/Makefile
```diff
@@ -234,6 +246,28 @@ lib : libzstd.a libzstd
 	@echo release build completed
 
 
+# Generate .h dependencies automatically
+
+DEPFLAGS = -MT $@ -MMD -MP -MF
+
+$(ZSTD_DYNLIB_DIR)/%.o : %.c $(ZSTD_DYNLIB_DIR)/%.d | $(ZSTD_DYNLIB_DIR)
+	@echo $@
+	$(Q)$(COMPILE.c) $(DEPFLAGS) $(ZSTD_DYNLIB_DIR)/$*.d $(OUTPUT_OPTION) $<
```
terrelln, line 255: Where is `$(COMPILE.c)` defined? It was defined in `programs/Makefile`, but not here.

lib/Makefile
Cyan4973: It's probably not visible on github,
but the most important part of the indentation change is that
I switched `tab` for 2 `space`.

Reason is, a command or assignment preceded by a `tab` is considered part of ongoing recipe.

This can lead to strange consequences when introducing some new recipe before an assignment.

For example, consider this segment, using a `tab` : 
```
ifeq ($(CONDITION), $(VALUE))
	CFLAGS += $(FLAG)
endif
```

Now, let's introduce a recipe just before, even if it doesn't look like so : 
```
recipe: prereq
	$(COMMAND)

# some comment
# clearly this is a different section 

ifeq ($(CONDITION), $(VALUE))
	CFLAGS += $(FLAG)
endif
```

now, `CFLAGS += $(FLAG)` is part of `recipe` !

This confusion no longer happens when using `space` instead.

lib/Makefile
Cyan4973: `$(COMPILE.c)` is a default variable of gnu `make`

lib/Makefile
terrelln: But it looks like there are some tabs leftover on lines 47 & 49. And can the nested if-statement be indented 4-spaces?

terrelln: Ah, GitHub doesn't do a great job indicating that I was referring to the whole block of lines 38-54, not this particular line.

terrelln: Oh, I see there are now 4 spaces, which just happen to be identical to a tab. Can the `ifneq` line not have leading spaces?

terrelln (approved):

lib/Makefile
Cyan4973: > Can the `ifneq` line not have leading spaces?

Do you mean, can the inner `ifneq` be indented, like the rest of the code block inside the outer `ifneq`?

I think it can.
It's just that the automatic parser didn't like it, and would mess up syntax coloring if I tried to indent the `ifneq`.
So I didn't feel it worth fighting it.

terrelln: Sounds good

PR #2373, opened 2020-10-26
Automatic detection of configuration changes

Cyan4973:
including dynamic ones.

This completes the series of changes related to `Makefile` build system.

After [last PR](https://github.com/facebook/zstd/pull/2370), the `Makefile` build system
is able to detect manual changes in build configuration,
and store cached `%.o` object files into corresponding directories.
But it would miss "dynamic" (runtime) modifications happening while applying some complex recipe,
such as `zstd` vs `zstd-nomt` or `libzstd` vs `libzstd-mt`.

This is fixed in this PR, which uses a 2 stages methodology : 
the first one merely establishes the final set of flags,
then pass them as static argument for the second stage.
Now, object cached for `libzstd` and `libzstd-mt` are stored into 2 different directories.

The price to pay is code complexity : the new `Makefile` is harder to read and modify.
It's not "horribly difficult", but a bit less straightforward.

Also, adding more targets using the 2-stages methodology is not fully "generic" : 
additional variables must be created and tracked.
So I've limited this effort to targets which can receive dynamic flag modifications.

Also : modified verbosity mechanism : now the `Makefile` is silent by default.
It makes the code slightly easier to read, by removing all those pesky `$(Q)`.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -226,7 +226,7 @@ ifeq ($(BUILD_DIR),0)
 .PHONY: libzstd.a  # not the actual recipe
 libzstd.a:
 	$(MAKE) --no-print-directory $@ \
-        BUILD_DIR=obj/conf_$(shell echo $(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) | $(HASH) | head -c 16) \
+        BUILD_DIR=obj/conf_$(shell echo $(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(ZSTD_FILES) | $(HASH) | head -c 16) \
         CPPFLAGS="$(CPPFLAGS)"
 
 else
@@ -266,7 +266,7 @@ ifeq ($(BUILD_DIR),0)
 .PHONY: $(LIBZSTD)  # not the actual recipe
 $(LIBZSTD):
 	$(MAKE) --no-print-directory $@ \
-        BUILD_DIR=obj/conf_$(shell echo $(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) | $(HASH) | head -c 16) \
+        BUILD_DIR=obj/conf_$(shell echo $(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(ZSTD_FILES) | $(HASH) | head -c 16) \
         CPPFLAGS="$(CPPFLAGS)" \
         CFLAGS="$(CFLAGS)" \
         LDFLAGS="$(LDFLAGS)"
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -15,12 +15,9 @@ default: lib-release
 $(V)$(VERBOSE).SILENT:
 
 # When cross-compiling from linux to windows,
-# you might need to specify this as "Windows."
-# Fedora build fails without it.
-#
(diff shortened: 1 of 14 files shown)


lib/Makefile
```diff
@@ -172,33 +176,28 @@ ZSTD_SUBDIR := common compress decompress dictBuilder legacy deprecated
...
 HASH ?= md5sum
 
-HAVE_HASH := $(shell echo 1 | $(HASH) > /dev/null && echo 1 || echo 0)
-ifeq ($(HAVE_HASH), 1)
-  HASH_VALUE := $(shell echo $(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(ZSTD_FILES) | $(HASH) | head -c 16)
-  HASH_DIR := conf_$(HASH_VALUE)
-else
-  $(info warning : could not find hash function to differentiate builds with different flags)
-  HASH_DIR := 0
+HASH_DIR = conf_$(shell echo $(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $(ZSTD_FILES) | $(HASH) | head -c 16)
```
terrelln, line 192: Looks like `head -c 16` isn't portable. See https://github.com/facebook/zstd/pull/2374

lib/Makefile
```diff
@@ -223,42 +222,78 @@ lib-all: all
...
+        CPPFLAGS="$(CPPFLAGS)"
+
+else
+# $(BUILD_DIR) is defined
+
+ZSTD_STATLIB_DIR := $(BUILD_DIR)/static
+ZSTD_STATLIB := $(ZSTD_STATLIB_DIR)/libzstd.a
+ZSTD_STATLIB_OBJ := $(addprefix $(ZSTD_STATLIB_DIR)/,$(ZSTD_LOCAL_OBJ))
+$(ZSTD_STATLIB): ARFLAGS = rcs
+$(ZSTD_STATLIB): | $(ZSTD_STATLIB_DIR)
```
terrelln, line 241: What does this line do?

lib/Makefile
```diff
@@ -223,42 +222,78 @@ lib-all: all
...
 
 LIBZSTD = libzstd.$(SHARED_EXT_VER)
-$(ZSTD_DYNLIB_DIR)/$(LIBZSTD): CFLAGS += -fPIC
-$(ZSTD_DYNLIB_DIR)/$(LIBZSTD): LDFLAGS += -shared -fvisibility=hidden
-$(ZSTD_DYNLIB_DIR)/$(LIBZSTD): | $(ZSTD_DYNLIB_DIR)
-$(ZSTD_DYNLIB_DIR)/$(LIBZSTD): $(ZSTD_DYNLIB_OBJ)
+$(LIBZSTD): CFLAGS += -fPIC
+$(LIBZSTD): LDFLAGS += -shared -fvisibility=hidden
+
+ifeq ($(BUILD_DIR),0)
```
terrelln, line 265: Can you add a comment somewhere explaining the 2-stage methodology, and why it is needed.

It isn't obvious to me why it is necessary.

lib/Makefile
Cyan4973: It creates the target directory if it doesn't exist

terrelln (approved):

PR #2375, opened 2020-10-28
Add a test case for LDM + opt parser with small uncompressible block

senhuang42:
OSS Fuzz exposed a bug in the LDM that was fixed in #2362 - this PR adds a test case for the specific failure: compressing with LDM + opt parser in streaming mode with a flush that contains a small, uncompressible block, followed by another compression.

Test Plan:
- Verified that this test will fail on versions of zstd prior to #2362, but passes after that PR was merged.
- Spot-checked to make sure we have the exact same issue (`rawSeqStore::posInSequence` fails to advance fully).

### DIFF
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -721,6 +721,11 @@ static int basicUnitTests(U32 const seed, double compressibility)
     DISPLAYLEVEL(3, "test%3i : LDM + opt parser with small uncompressible block ", testNb++);
     {   ZSTD_CCtx* cctx = ZSTD_createCCtx();
         ZSTD_DCtx* dctx = ZSTD_createDCtx();
+        if (!cctx || !dctx) {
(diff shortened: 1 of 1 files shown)


tests/fuzzer.c
```diff
@@ -718,6 +718,42 @@ static int basicUnitTests(U32 const seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : LDM + opt parser with small uncompressible block ", testNb++);
+    {   ZSTD_CCtx* cctx = ZSTD_createCCtx();
+        ZSTD_DCtx* dctx = ZSTD_createDCtx();
```
Cyan4973, line 723: minor : check that `cctx` and `dctx` are non `NULL`

Cyan4973 (approved):

PR #2376, opened 2020-10-28
Refine external ZSTD_Sequence API

senhuang42:
This PR modifies the public-facing `ZSTD_Sequence` API. This will make it easier to integrate with the upcoming sequence compression API.

- Removes unnecessary field `matchPos` from `ZSTD_Sequence` and adjusts member `rep` to make more sense.
- Changes `ZSTD_getSequences()` to provide last literals (rather than ignoring them).
- Improves documentation regarding the `ZSTD_Sequence` as well as some internal definitions.
- Adjusts/fixes existing related unit tests.

Test Plan:
- Unit test for sequence extraction modified to suit new definitions
- Spot checking on multiple files that even with modifications, `ZSTD_getSequences()` outputs the exact same output, with the only difference being the value of 'rep' (we no longer allow `rep == 4`).

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -341,23 +341,28 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
 *  Private declarations
 *********************************************/
 typedef struct seqDef_s {
-    U32 offset;
+    U32 offset;         /* Offset code of the sequence */
     U16 litLength;
     U16 matchLength;
 } seqDef;
 
 typedef struct {
     seqDef* sequencesStart;
-    seqDef* sequences;
+    seqDef* sequences;      /* ptr to end of sequences */
     BYTE* litStart;
-    BYTE* lit;
+    BYTE* lit;              /* ptr to end of literals */
     BYTE* llCode;
     BYTE* mlCode;
     BYTE* ofCode;
     size_t maxNbSeq;
     size_t maxNbLit;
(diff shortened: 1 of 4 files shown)


lib/common/zstd_internal.h
```diff
@@ -341,23 +341,28 @@ MEM_STATIC size_t ZSTD_limitCopy(void* dst, size_t dstCapacity, const void* src,
...
     size_t maxNbSeq;
     size_t maxNbLit;
-    U32   longLengthID;   /* 0 == no longLength; 1 == Lit.longLength; 2 == Match.longLength; */
-    U32   longLengthPos;
+
+    /* longLengthPos and longLengthID to allow us to represent either a single litLength or matchLength
+     * in the seqStore that has a value larger than U16 (if it exists). To do so, we increment
+     * the existing value of the litLength or matchLength by 0x10000. 
+     */
+    U32   longLengthID;   /* 0 == no longLength; 1 == Represent the long literal; 2 == Represent the long match; */
```
Cyan4973, line 364: not directly relevant to this specific PR, but worth noting : 
it seems this `longLengthID` field should better be an `enum` type.

Cyan4973 (approved):

senhuang42:
Some more minor changes - the existing `ZSTD_getSequences()` function actually had a bug where the lack of `ZSTD_CompressedBlockState` updates was causing the repcodes to be not quite right, so commit `536e89c` fixes that issue.

PR #2381, opened 2020-11-02
Add enum to define ZSTD_Sequence type and update sequence extraction API

senhuang42:
Currently, `ZSTD_getSequences()` emits an array of sequences with delimiters (in the form of sequences with `offset = 0, matchLength = 0` for block boundaries and last literals. We add the capability to emit sequences in a more "raw" form with no explicit block boundaries or last literals. 

This will be a "mode" of `ZSTD_Sequence` that the sequence compression API must be able to handle, so we should add the capability to extract sequences in this format via `ZSTD_getSequences()`.

This PR:
- Adds a new enum type to `zstd.h` that defines modes of sequence extracting/ingestion
- Extends the `ZSTD_getSequences()` function to support the new mode
- Updates the relevant docs

Test Plan:
- Include new mode in `fuzzer.c` unit test.
- Manual spot checking to ensure that output looks as expected.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2550,7 +2550,7 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
 }
 
 size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
-    size_t outSeqsSize, const void* src, size_t srcSize)
+    size_t outSeqsSize, const void* src, size_t srcSize, ZSTD_sequenceFormat_e format)
 {
     const size_t dstCapacity = ZSTD_compressBound(srcSize);
     void* dst = ZSTD_customMalloc(dstCapacity, ZSTD_defaultCMem);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2562,6 +2562,7 @@ size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
     seqCollector.seqStart = outSeqs;
     seqCollector.seqIndex = 0;
     seqCollector.maxSequences = outSeqsSize;
+    seqCollector.format = format;
     zc->seqCollector = seqCollector;
 
     ZSTD_compress2(zc, dst, dstCapacity, src, srcSize);
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2551,7 +2551,7 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
 }
 
 size_t ZSTD_generateSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
-    size_t outSeqsSize, const void* src, size_t srcSize, ZSTD_sequenceFormat_e format)
+                              size_t outSeqsSize, const void* src, size_t srcSize)
 {
     const size_t dstCapacity = ZSTD_compressBound(srcSize);
     void* dst = ZSTD_customMalloc(dstCapacity, ZSTD_defaultCMem);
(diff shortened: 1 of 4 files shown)


lib/zstd.h
```diff
@@ -1297,16 +1297,29 @@ ZSTDLIB_API unsigned long long ZSTD_decompressBound(const void* src, size_t srcS
  *           or an error code (if srcSize is too small) */
 ZSTDLIB_API size_t ZSTD_frameHeaderSize(const void* src, size_t srcSize);
 
+typedef enum {
+  ZSTD_sf_explicitBlockDelimiters,  /* Representation of ZSTD_Sequence contains explicit block delimiters */
+  ZSTD_sf_noBlockDelimiters         /* Representation of ZSTD_Sequence has no block delimiters, sequences only */
+} ZSTD_sequenceFormat_e;
+
```
senhuang42, line 1304: I wonder if it makes sense to create another struct like
```struct ZSTD_Sequences {
    ZSTD_Sequence* seqs,
    ZSTD_sequenceFormat_e format
}```

In order to force a tighter coupling of the concept of the format and the `ZSTD_Sequence` array. My main hesitation is that I don't want to create too many new types for this API.

File: lib/compress/zstd_compress.c [hunk 2]
```diff
@@ -2566,6 +2567,26 @@ size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
...
+    if (format == ZSTD_sf_noBlockDelimiters) {
+        size_t i = 0;
+        size_t totalSeqs = zc->seqCollector.seqIndex;
+        while(i < totalSeqs) {
+            if (seqCollector.seqStart[i].offset == 0 && seqCollector.seqStart[i].matchLength == 0) {
+                /* Merge the block boundary or last literals */
+                if (i != totalSeqs-1) {
+                    /* Add last literals to next sequence, then "delete" this sequence */
+                    seqCollector.seqStart[i+1].litLength += seqCollector.seqStart[i].litLength;
+                    ZSTD_memmove(seqCollector.seqStart+i, seqCollector.seqStart+i+1, (totalSeqs-i-1)*sizeof(ZSTD_Sequence));
```
terrelln, line 2580: This is quadratic in the worst case.

A better way to do this is to have have an input index and an output index. Something like:
```c
while (input < totalSeqs) {
  if (merge) {
    seqCollector.seqStart[output] = seqCollector.seqStart[input+1];
    seqCollector.seqStart[output].litLength += seqCollector.seqStart[input];
    output += 1;
    input += 2;
  } else {
    seqCollector.seqStart[output++] = seqCollector.seqStart[input++];
  }
}
```

senhuang42: Yeah lol, the naive approach took way too much time even just on `silesia.tar`.

lib/zstd.h
Cyan4973: That doesn't seem necessary

lib/compress/zstd_compress.c
```diff
@@ -2566,6 +2567,26 @@ size_t ZSTD_getSequences(ZSTD_CCtx* zc, ZSTD_Sequence* outSeqs,
 
     ZSTD_compress2(zc, dst, dstCapacity, src, srcSize);
     ZSTD_customFree(dst, ZSTD_defaultCMem);
+
+    if (format == ZSTD_sf_noBlockDelimiters) {
+        /* Remove all block delimiters and append them to the next sequence's literals
+         * and do not emit last literals at all
```
Cyan4973, line 2573: > and do not emit last literals at all

Well, even in `noBlockDelimiters` mode, there is still the issue of representing the last literals of the last block.

senhuang42: In `noBlockDelimiters` mode we intentionally don't represent the last literals of the last block to be consistent, and then on the sequence compression side, I can still derive the last literals of the last block even without it, just based on how much of the source buffer we've consumed, right?

Cyan4973: Sure, that can work,
though it's very easy to miss for an API user, hence it should be clearly documented that there may be a few literals left in the buffer after the last sequence, that it's expected, and must be looked for.

To be fair, if it was me, I would only support `ZSTD_sf_explicitBlockDelimiters` for the extraction API,
as it's the most complete image of what's produced (`ZSTD_sf_noBlockDelimiters` throw information away, but doesn't bring any value).

If you worry about an ability to make a round-trip test with an ingestion API in "automated block delimiter mode", I would say this is a completely different topic. Such automated mode could still accept explicit block delimiter sequences, but would simply interpret them differently, merging them with following sequence.

senhuang42: I agree that the last literals can be easily missed, so I'll make sure to clearly document this. I chose to not represent the last block last literals in `noBlockDelimiters` mode since I would like the sequences in our `ZSTD_Sequence` array to be as generic as possible, avoiding any "dummy" sequences - which is what the last literals would be if we were to represent them.

I still do kind of like the idea of allowing the specification of block delimiters or not for the extraction API - it enforces a stronger sense of symmetry between `getSequences()` and `compressSequences()`. I do agree that the `ZSTD_sf_noBlockDelimiters` only throws information away, but it seems like a nice feature to have if the user wanted to be generate sequences with both modes and compare them, in order to better visualize/understand the different expectations that `compressSequences()` might have depending on the mode.

As for testing, I will include a roundtrip unit test in my upcoming compress sequence API pull request, but I think the "real" test will come in a follow-up PR where I add a dedicated OSS Fuzzer that will generate arbitrary sequences on a literal buffer (not using the `ZSTD_getSequences()` API), and try to compress and decompress those arbitrary sequences.

senhuang42:
Discussed offline:
- Change name of `ZSTD_getSequences()` to `ZSTD_generateSequences()` for clarity, update documentation to better reflect what actually happens.
- Instead of using enum to determine sequence generation format, provide an additional function to merge block delimiters if necessary

Cyan4973 (approved):

PR #2383, opened 2020-11-03
Add ZSTD_getDictID_fromCDict

LuAPi:
I have use for this method and it seems like an obvious one to have.

I would have implemented it the same way as `ZSTD_getDictID_fromDDict` but `ZSTD_getDictID_fromDict` wasn't visible in zstd_compress.c so I just returned the stored member value instead.

I'm not sure if I should change the documentation to reflect that it returns the dict id stored in the struct rather than what I copy-paste-edited from the `ZSTD_getDictID_fromDDict` function.


terrelln:
This looks reasonable to me, thanks for the PR! The only question I have is for @Cyan4973: Do we want to put this directly into the stable API, or start it off in the unstable section?

> I'm not sure if I should change the documentation to reflect that it returns the dict id stored in the struct rather than what I copy-paste-edited from the ZSTD_getDictID_fromDDict function.

That doesn't need to be documented, since the `ZSTD_CDict` struct is opaque.

Cyan4973:
We have a policy that new prototypes start their life in the unstable section.
After at least one release, they can join the "stable" section if they seem deemed to.
For that to happen, the release number must also increase at least its second digit.

So that would be a candidate for "stable" status for `v1.5.0`.
Given how straightforward the function definition and implementation are, 
helped by clear familiarity with surrounding similar functions,
this prototype has a fairly good chance to make it.

LuAPi:
I've made a [new PR](https://github.com/facebook/zstd/pull/2385) for this since correcting my email address made a mess of things in this one. The requested changes are in that one.

PR #2387, opened 2020-11-05
[RFC] New sequence compression API

senhuang42:
**ZSTD_compressSequences()**
```
size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size_t dstSize,
                                  const ZSTD_Sequence* inSeqs, size_t inSeqsSize,
                                  ZSTD_sequenceFormat_e format);
```

Which compress an array of `ZSTD_Sequence` given a source buffer. 

ZSTD_Sequence definition
```
// Definition of ZSTD_Sequence <- what the matchfinder should provide              
typedef struct {
    unsigned int offset;      /* The offset of the match. (NOT the same as the offset code)
                               * If offset == 0 and matchLength == 0, this sequence represents the last
                               * literals in the block of litLength size.
                               */
    unsigned int litLength;   /* Literal length of the sequence. */
    unsigned int matchLength; /* Match length of the sequence. */
                              /* Note: Users of this API may provide a sequence with matchLength == litLength == offset == 0.
                               * In this case, we will treat the sequence as a marker for a block boundary.
                               */
    
    unsigned int rep;         /* Represents which repeat offset is represented by the field 'offset'.
                               * Ranges from [0, 3].
                               * 
                               * Repeat offsets are essentially previous offsets from previous sequences sorted in
                               * recency order. For more detail, see doc/zstd_compression_format.md
                               * 
                               * If rep == 0, then 'offset' does not represent a repeat offset.
                               * If rep > 0:
                               *  If litLength != 0:
                               *      rep == 1 --> offset == repeat_offset_1
                               *      rep == 2 --> offset == repeat_offset_2
                               *      rep == 3 --> offset == repeat_offset_3
                               *  If litLength == 0:
                               *      rep == 1 --> offset == repeat_offset_2
                               *      rep == 2 --> offset == repeat_offset_3
                               *      rep == 3 --> offset == repeat_offset_1 - 1
                               */
} ZSTD_Sequence;
```

This PR is generally pretty self-contained, but there are some side effects:
- The initialization stage of `ZSTD_compressStream2()` was pulled out into its own function since we re-use that initialization code pretty much exactly.
- The existing internal function named `ZSTD_compressSequences()` was renamed `ZSTD_entropyCompressSequences()` so we could use that name.

Additionally, I should note that this has only been tested with sequences provided from `ZSTD_getSequences()`. As such, the match splitting is not particularly well-tested, since even with `ZSTD_sf_noBlockDelimiters`, `ZSTD_getSequences()` tends to produce "easy" cases that probably won't need match splitting (we only split if the match length is greater than or equal to block size). Once I write the fuzzer, I'll have a better ability to test the match splitting.

Also note that this PR is rebased on top of #2381, so until that one gets merged, changes/commits from #2381 will show up in this PR as well.

**Test Plan:**
- Unit test included
- Spot checking on random files (including `silesia.tar`)
- Follow-up PR to add OSS fuzzing with arbitrary sequences.

**Needed followups**
- How to tune validation?
- Add support for using externally-provided repcodes.
- Minor refactors to block compression logic


lib/compress/zstd_compress.c
```diff
@@ -4433,6 +4464,382 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+    BYTE* op = (BYTE*)dst;
+    size_t cSize = 0;
+    size_t compressedBlocksSize = 0;
+    size_t frameHeaderSize = 0;
+
+    /* Transparent initialization stage, same as compressStream2() */
+    DEBUGLOG(3, "ZSTD_compressSequencesCCtx()");
+    assert(cctx != NULL);
+    FORWARD_IF_ERROR(ZSTD_CCtx_init_compressStream2(cctx, ZSTD_e_end, srcSize), "CCtx initialization failed");
+    if (dstCapacity < ZSTD_compressBound(srcSize)) {
```
terrelln, line 4788: Why do we have this check? No other API requires that the output buffer is `>= ZSTD_compressBound(srcSize)`.

lib/compress/zstd_compress.c
```diff
@@ -4433,6 +4464,382 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+size_t ZSTD_compressSequences(void* dst, size_t dstCapacity,
+                                  const ZSTD_Sequence* inSeqs, size_t inSeqsSize,
+                                  const void* src, size_t srcSize, int compressionLevel,
+                                  ZSTD_sequenceFormat_e format)
+{
+    size_t cSize;
+    ZSTD_CCtx* const cctx = ZSTD_createCCtx();
+
+    RETURN_ERROR_IF(!cctx, memory_allocation, "Couldn't allocate memory for CCtx!");
+    FORWARD_IF_ERROR(ZSTD_CCtx_reset(cctx, ZSTD_reset_session_and_parameters), "CCtx reset failed");
```
terrelln, line 4833: This call isn't necessary. It doesn't hurt, but it doesn't do anything.

lib/compress/zstd_compress.c
```diff
@@ -4433,6 +4464,382 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+    FORWARD_IF_ERROR(ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel, compressionLevel), "Parameter setting failed");
+
+    cSize = ZSTD_compressSequencesCCtx(cctx, dst, dstCapacity, inSeqs, inSeqsSize, src, srcSize, format);
+    FORWARD_IF_ERROR(cSize, "ZSTD_compressSequencesCCtx() failed");
```
terrelln, line 4837: We're leaking the CCtx on error with these `FORWARD_IF_ERROR()` calls.

lib/compress/zstd_compress.c
```diff
@@ -4433,6 +4464,382 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+
+    DEBUGLOG(3, "Final compressed size: %zu", cSize);
+    return cSize;
+}
+
+
+size_t ZSTD_compressSequences(void* dst, size_t dstCapacity,
+                                  const ZSTD_Sequence* inSeqs, size_t inSeqsSize,
+                                  const void* src, size_t srcSize, int compressionLevel,
+                                  ZSTD_sequenceFormat_e format)
```
terrelln, line 4827: Why do we need this function?

lib/compress/zstd_compress.c
```diff
@@ -4433,6 +4464,382 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+        if (lastBlock) {
+            break;
+        } else {
+            ip += blockSize;
+            op += cBlockSize;
+            remaining -= blockSize;
+            dstCapacity -= cBlockSize;
+            cctx->isFirstBlock = 0;
+        }
+    }
```
terrelln, line 4770: Could (some of) this code be extracted to a function and shared with other places that have similar code?

lib/compress/zstd_compress.c
senhuang42: Initially I included it as a convenience function, but seeing as this is a pretty advanced API, it doesn't make sense to have. I'll remove.

lib/compress/zstd_compress.c
senhuang42: The other place where the block compression logic looks a bit like this is in `ZSTD_compressBlock_internal()` and `ZSTD_compress_frameChunk()` - pulling the logic into a new function is definitely possible, but would require some refactoring of those functions too. This PR is already getting kind of big, so I'm nervous add more changes/points of failure. I'd be happy to add it into the list of followups for this PR though.

senhuang42:
I've updated with an overhaul to the logic so that we only do one pass through the `ZSTD_Sequence` array, with a focus on our primary use case of: 1) no block delimiters 2) recalculating repcodes.

I've also split handling of the case with no block delimiters and the case with block delimiters into two different functions, since the latter case is significantly simpler, and supporting both in the same function is a bit needlessly complicated since they rely on some different assumptions, and would slow down the processing a tiny bit.

Some things to note:
- Currently we always validate the offsets/matchLengths. We can maybe add a toggle in the cctx params to enable or disable this validation - my only hesitation is whether or not adding even more complexity to the cctxParams is worth the few clock cycles we'd save.
- I've changed the API so that it always will (re)calculate repcodes, in accordance with our primary use case. We can add back in the ability to use repcodes given by external providers in a follow-up, but doing so requires some significant work/debugging for the case of splitting repcode matches.

Cyan4973:
> Currently we always validate the offsets/matchLengths. 

If I remember correctly, the target model is to blindly trust `offsets` and `matchLengths` provided by the sequence generator.
So it should be possible to disable it, 
especially if we want to also support dictionary compression scenarios,
where offsets will go beyond the beginning of the file.

That being said, the capability to validate sequences is still useful, particularly during debugging or development.
At which point, saving cycles is not the priority.

> I've changed the API so that it always will (re)calculate repcodes

Fine, but document it. `Sequence.rep` is effectively unused during ingestion. The API should mention that clearly.

terrelln:
>> I've changed the API so that it always will (re)calculate repcodes

>Fine, but document it. Sequence.rep is effectively unused during ingestion. The API should mention that clearly.

I think that this is a limitation we will eventually want to address. Repcodes are branches in the decompressor (as it is currently implemented). So if you have e.g. 50% repcodes you have a unpredictable branch in your decoder. Someone who strongly (strongly) favors decompression speed over size might want to avoid emitting repcodes (or alternatively avoid non-repcodes as much as possible). Or they may have a speed-size tradeoff heuristic used to generate repcodes.

Or the the sequences come from an repcode-aware optimal parser, and for whatever reason it wants to use a raw offset directly (I have a hard time imagining the scenario, maybe rep2 is cheaper than rep1?).

Cyan4973:
> I think that this is a limitation we will eventually want to address.

I believe it's relatively straightforward to address in combination with explicit Block delimiters,
which is in line with the "full control" scenario you are mentioning.

I guess Sen is concerned by the combination with automatic block boundary mode, where corner case scenarios exist. But that could be addressed in some later PR.

senhuang42:
> If your point is that there is not much saving in disabling validation, which is a good point,
> then a remaining question is how to support dictionary compression scenarios,
> where offsets will go beyond the beginning of the source.
> 
> Another consequence is that sequence generation may fail as an outcome of validation,
> so this behavior should be clearly mentioned in the documentation,
> so that users don't get surprised by this outcome (error handling is still on their side).

It seems like there is perhaps some more cost than I initially expected: I've followed the docs to deal with the dictionary offset scenarios:
```
As long as the amount of data decoded from this frame is less than or equal to Window_Size, sequence commands may specify offsets longer than the total length of decoded output so far to reference back to the dictionary, even parts of the dictionary with offsets larger than Window_Size. After the total output has surpassed Window_Size however, this is no longer allowed and the dictionary is no longer accessible.
```
In this case, we have to keep track of the current theoretical position in what would be the decoded stream, based on the sequences, and then based on that position, make a decision about the offset. Actually, the only reason that `ZSTD_sequencePosition` even has the member `posInSrc` is to keep track of this in order to validate dictionary-based offsets. 
And so the cost of adding validation seems be: an extra branch(?) (whether or not `posInSrc` is greater than `windowSize`), calculating that position each block, and extra complexity in our internal `ZSTD_sequencePosition` struct. Seeing all this extra cost, I'm leaning somewhat more toward dropping the validation especially since it's mostly useful as the person developing this API in the first place.

Cyan4973:
>  it's mostly useful to me as the person developing this API in the first place, rather than providing as much value to the user.

I would actually see it as providing a value for developpers of sequence generators.

> the cost of adding validation seems be: an extra branch(?)

Keep in mind that speed is less critical when running in debug / dev mode.

I don't see this development as very difficult, but I agree it nonetheless represents additional complexity. 
And therefore, since it's not on the critical path, it could be decoupled and added at a later stage.

senhuang42:
> > the cost of adding validation seems be: an extra branch(?)
> 
> Keep in mind that speed is less critical when running in debug / dev mode.
> 
> I don't see this development as very difficult, but I agree it nonetheless represents additional complexity.
> And therefore, since it's not on the critical path, it could be decoupled and added at a later stage.

Ah right - it seems that without debug levels enabled, it looks like the function mostly gets optimized out, good to know! And yeah, it's not difficult to add, but I'm trying to bias toward simpler code if possible. 

That being said, it's simple enough that I can just keep it in this PR.

senhuang42:
Another edge case has popped up in the fuzzing that is probably worth flagging:

The decoder does not allow play well with the following case:
Suppose we have the following sequence:
- matchLength: 1000
- matchPos: 800
- offset: 1000
- litLength: 0 (for convenience)

With:
- dictSize: 300
- windowSize: 850
- blockSize: 900

On the encoder side, we are forced to split the match, since it's greater than `blockSize`. So, the first match will have:
(offset: 1000, matchLength: 100, matchPos: 800), which is valid, since the absolute position of the match (800) is less than windowSize (850), which means we have offsets reference up to `matchPos + dictSize == 1100`

But then the second half of the split match has `(offset: 1000, matchLength: 900, matchPos: 900)`. But now, the offset of 1000 is no longer valid, since matchPos (900) is greater than `windowSize` (850), which mean we are now limited to `windowSize == 850` as the maximum for our offset.

If it is the case that a "real" matchfinder would not find matches in this form, then this is an issue with the fuzzer, and we can adjust it so that we bound the matchlength when matchpos is less than windowsize.

If not, then perhaps something needs to change on the decoder validation side? If I disable `assert()` in the decoder, this case actually decodes and roundtrips just fine.

Cyan4973:
`blockSize` is not allowed to be greater than `windowSize`.

senhuang42:
> Note that `blockSize` is not allowed to be greater than `windowSize`.
> (see `Block_Maximum_Size` in https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#blocks).
> 
> That being said, the scenario described is indeed not encodable.
> The specification is pretty clear that, after `windowSize` bytes of input are consumed, the dictionary is no longer accessible, so all matches must have a maximum distance of `windowSize`.
> 
> All match finders within the reference implementation guarantee that condition, so it's reasonable that the fuzzer complies with it too, at least if its objective is to generate valid sequences.

Good to know, I'll change the fuzzer to agree with this constraint. I guess it makes sense as the matchfinder more or less works byte-by-byte, while I was thinking about it more in terms of sequences.

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+    size_t blockSize;
+    size_t compressedSeqsSize;
+    size_t remaining = srcSize;
+    ZSTD_sequencePosition seqPos = {0, 0, 0};
+    U32 repCodesBackup[ZSTD_REP_NUM];   /* If we emit block as nocompress or RLE block, the decoder will
+                                         * never "see" those sequences. So, in order for the repcode table
+                                         * to remain in sync, we need to revert the repcode table back to the
+                                         * state it was in before processing the sequences in the RLE or nocompress
+                                         * block.
+                                         */
```
senhuang42, line 4730: I believe this should fix the edge case where RLE blocks (and I'm guessing uncompressed blocks) ruin repcode history for the next three matches.

If I understand correctly, the reason that RLE blocks (and no compress blocks) are troubling to the current API is that on the decoder side, it never sees the sequences that might have been inside an RLE block or nocompress block. But the issue is that in this API we update the repcode table using those "unused" sequences anyways.

To mitigate, we keep a copy of the repcodes that this current block would start with. After `ZSTD_copySequencesToSeqStore`, the block will end up with a new repcode table in the `blockState`. If it's the case that we emit the block as RLE or noCompress, then we copy the repcodes from this backup back into the `blockState`. Does this make sense?

senhuang42: Actually, looking at my code again, I actually think this case might have already been taken care of. Since `ZSTD_confirmRepcodesAndEntropyTables()` is only called when we emit a compressed block. That means when we emit an RLE or noCompress block, we don't actually update the value of `rep` in `prevCBlock`. 

Therefore, when we begin processing the sequences for the next block, it's actually using the same starting repcode table values as the block that was emitted uncompressed, or RLE.

senhuang42:
Actually, looking at my code again, I actually think the case where we emit an RLE block should actually be fine, and my latest commit to address that is redundant. 

If I understand correctly, the concern was that RLE blocks (and no compress blocks) are troubling to the current API is that on the decoder side, it never sees the sequences that might have been inside an RLE block or nocompress block. But the issue is that in this API we update the repcode table using those "unused" sequences anyways. However, in these cases on the compression side we actually don't update the repcode table for these cases.

Since `ZSTD_confirmRepcodesAndEntropyTables()` is only called when we emit a compressed block. That means when we emit an RLE or noCompress block, we don't actually update the value of `rep` in `prevCBlock`, since we create a copy of the repcodes during `ZSTD_copySequencesToSeqStore()` rather than modifying the repcodes in-place.

Therefore, when we begin processing the sequences for the next block, it's actually using the same starting repcode table state as the uncompress block would've been using at the beginning of its processing, so RLE/uncompressed blocks should have no effect on repcode history on the compression side.

@Cyan4973 If I'm missing something and this is not what's happening, then I'm happy to just remove RLE/nocompress blocks for now.

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+
+    return bytesAdjustment;
+}
+
+/* Compress, block-by-block, all of the sequences given.
+ *
+ * Returns the cumulative size of all compressed blocks (including their headers), otherwise a ZSTD error.
+ */
+static size_t ZSTD_compressSequences_internal(void* dst, size_t dstCapacity,
+                                              ZSTD_CCtx* cctx,
```
Cyan4973, line 4716: minor : parameter order
Keep `cctx` in first position, mirroring `ZSTD_compressSequences()` interface.
`cctx` is a mutable state, and by convention should always be in first position.

Cyan4973:
There are 2 scenarios.

In the first one, the user provides repcodes explicitly.
In which case, these repcodes are expected to be encoded "as is", and their meaning should remain unaltered.
Problem is, introducing an unexpected RLE block in the block history is bound to impact repcode history, thus changing the meaning of future repcodes.

The defense here is to say that this mode is currently not supported.
Right. But this impact should be noted "somewhere", so that when this mode becomes available, this detail is not forgotten.

Now let's get back to the current scope of the function, where repcodes are recalculated internally, hence not provided by the user.
It leaves some rules to decide when to employ repcodes and which values to use.
Inserting an RLE block is now possible, _but_ it can't be a "transparent" choice that the sequence encoder is not aware of, since it impacts repcode history.

It all comes down to how you maintain the repcode history. If you maintain it yourself, and if the decision to output an RLE block is done independently from the sequence encoder, then there is a risk.
However, if your repcode history is re-synchronized after each block _from_ the block encoder, then it's by definition properly maintained, and the introduction of RLE block is not a disruptive event.

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+        size_t cBlockSize;
+        size_t additionalByteAdjustment;
+        lastBlock = remaining <= cctx->blockSize;
+        blockSize = lastBlock ? (U32)remaining : (U32)cctx->blockSize;
+        blockSeqStore = cctx->seqStore;
+        ZSTD_resetSeqStore(&blockSeqStore);
+        DEBUGLOG(4, "Working on new block. Blocksize: %zu", blockSize);
+        ZSTD_memcpy(repCodesBackup, cctx->blockState.prevCBlock->rep, sizeof(repcodes_t));
+        if (cctx->appliedParams.blockDelimiters == ZSTD_sf_noBlockDelimiters) {
+            additionalByteAdjustment = ZSTD_copySequencesToSeqStore(&blockSeqStore, &seqPos,
```
Cyan4973, line 4757: minor : since these functions use the exact same signature,
selection could also be achieved through a function pointer.

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+ * This function will attempt to scan through blockSize bytes represented by the sequences
+ * in inSeqs, storing any (partial) sequences. 
+ * 
+ * Occasionally, we may want to change the actual number of bytes we consumed from inSeqs to
+ * avoid splitting a match, or to avoid splitting a match such that it would produce a match
+ * smaller than MINMATCH. In this case, we return the number of bytes that we didn't read from this block.
+ */
+static size_t ZSTD_copySequencesToSeqStore(seqStore_t* seqStore, ZSTD_sequencePosition* seqPos,
+                                           const ZSTD_Sequence* const inSeqs, size_t inSeqsSize,
+                                           const void* src, size_t blockSize, ZSTD_CCtx* cctx) {
```
Cyan4973, line 4597: Same comment regarding parameter order : 
since `cctx` is a (mutable) state, which internals are updated by passing through this function, it should be in first position.

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+                         "Sequence validation failed");
+        DEBUGLOG(6, "Storing sequence: (of: %u, ml: %u, ll: %u)", offCode, matchLength, litLength);
+        ZSTD_storeSeq(seqStore, litLength, ip, iend, offCode, matchLength - MINMATCH);
+        ip += matchLength + litLength;
+    }
+    DEBUGLOG(5, "Ending seq: idx: %u (of: %u ml: %u ll: %u)", idx, inSeqs[idx].offset, inSeqs[idx].matchLength, inSeqs[idx].litLength);
+    assert(idx == inSeqsSize || endPosInSequence <= inSeqs[idx].litLength + inSeqs[idx].matchLength);
+    seqPos->idx = idx;
+    seqPos->posInSequence = endPosInSequence;
+    ZSTD_memcpy(cctx->blockState.nextCBlock->rep, updatedRepcodes.rep, sizeof(repcodes_t));
```
Cyan4973, line 4696: Is that the only mutable part of the `cctx` state ? It's used as a repcode history storage ?

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+ * went wrong.
+ * 
+ * This function will attempt to scan through blockSize bytes represented by the sequences
+ * in inSeqs, storing any (partial) sequences. 
+ * 
+ * Occasionally, we may want to change the actual number of bytes we consumed from inSeqs to
+ * avoid splitting a match, or to avoid splitting a match such that it would produce a match
+ * smaller than MINMATCH. In this case, we return the number of bytes that we didn't read from this block.
+ */
+static size_t ZSTD_copySequencesToSeqStore(seqStore_t* seqStore, ZSTD_sequencePosition* seqPos,
```
Cyan4973, line 4595: Since `cctx` is already passed as state, and contains the `seqStore`,
is there any value in passing `seqStore` as a specific parameter ?

Passing `seqStore` as a separate parameter implies that it's a separate storage,
while passing it as part of `cctx` implies that it _has to_ be `cctx.seqStore` for the rest of the algorithm to work properly.

lib/compress/zstd_compress.c
senhuang42: Mostly - we also will modify the entropy tables if we emit a compressed block, so the mutable part is the `blockState`.

Each time we go through a block, we will update the `blockState.nextCBlock` with the new repcodes, and only when we emit a compressed block, we'll update `blockState.prevCBlock`, using `ZSTD_confirmRepcodesAndEntropyTables()`, which updates the repcodes and entropy tables for the next block. 

If there's no compressed block emitted, `nextCBlock` still gets mutated, but the values in it are not used during sequence processing or compression, since it's effectively just storage.

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4486,400 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+     * window size. After output surpasses windowSize, we're limited to windowSize offsets again.
+     */
+    offsetBound = posInSrc > windowSize ? (size_t)windowSize : posInSrc + (size_t)dictSize;
+    RETURN_ERROR_IF(offCode > offsetBound + ZSTD_REP_MOVE, corruption_detected, "Offset too large!");
+    RETURN_ERROR_IF(matchLength < MINMATCH, corruption_detected, "Matchlength too small");
+    return 0;
+}
+
+/* Returns an offset code, given a sequence's raw offset, the ongoing repcode array, and whether litLength == 0 */
+static U32 ZSTD_finalizeOffCode(U32 rawOffset, const U32* const rep, U32 ll0) {
```
Cyan4973, line 4512: very minor (suggestion):
from a compiler standpoint, `const U32* const rep` is equivalent to `const U32 rep[3]`.
But from an interface perspective (reader), the second one conveys a more accurate intention.

lib/compress/zstd_compress.c
Cyan4973: So it's definitely a state

senhuang42:
Updated to address comments - function signatures changed, added new experimental param for validation

Cyan4973 (approved):

lib/compress/zstd_compress.c
```diff
@@ -4461,6 +4501,403 @@ size_t ZSTD_compress2(ZSTD_CCtx* cctx,
...
+            cBlockSize = ZSTD_noCompressBlock(op, dstCapacity, ip, blockSize, lastBlock);
+            FORWARD_IF_ERROR(cBlockSize, "Nocompress block failed");
+            DEBUGLOG(4, "Block too small, writing out nocompress block: cSize: %zu", cBlockSize);
+            cSize += cBlockSize;
+            ip += blockSize;
+            op += cBlockSize;
+            remaining -= blockSize;
+            dstCapacity -= cBlockSize;
+            continue;
+        }
```
terrelln, line 4798: I know that you don't currently read the `rep` field of the `ZSTD_sequence`, but what is the plan for this case when you start to respect the field. Will we just emit a compressed block, since we now allow compressed blocks to grow in size? Or should we have a way for the caller to specify that they want a raw/rle block?

terrelln (approved):
This looks good to me!

My only suggestion is that we should allow a way to specify raw/rle blocks using this API. We don't have to implement it in this PR, but we should make sure the API will be able to support it in the future.

E.g. `offset==0 && matchLength == 1` means this sequence is a raw block of `literalLength`. `offset == 0 && matchLength == 2` means RLE block of `literalLength`.

lib/compress/zstd_compress.c
senhuang42: We can definitely add capability for the caller to specify an RLE. In the mode with block delimiters, it will be entirely "strict", where the user must specify everything from last literals, block delims, repcodes, RLE/raw blocks (this can be done by overloading dummy sequences with offsets even more, like saying RLE = (of: 0, ml: rle_length) or something like that). 

I think we're converging on a model where the mode with no block delimiters will delegate to the API a lot of control of the compressed output. Since it's just arbitrary sequences we're being fed, it makes sense that we have control over that to make it fit into zstd's model.

PR #2393, opened 2020-11-12
Improve repcode handling in sequence extraction API

senhuang42:
The repcode handling logic was slightly off in the original version of this API - this PR aims to fix it by: 1) Keeping track of a dedicated repcode array, and 2) Persist the repcodes array through each block. I've aimed to make the handling logic as clear to read as possible since we will need a similar process in the sequence compression API.

Test Plan:
- Manual spot checking (roundtrips silesia.tar with compression api)
- Modify the unit test. Previously, it turns out that setting the `litProbability` field to something like 0.5 in the `RDG_genBuffer()` function would almost always cause the unit test to fail. With this PR, it no longer does.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2497,7 +2497,6 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
 
     ZSTD_Sequence* outSeqs = &zc->seqCollector.seqStart[zc->seqCollector.seqIndex];
     size_t i;
-    int repIdx;
     U32 rep[ZSTD_REP_NUM];
     U32 shouldUpdateRep;
 
@@ -2561,8 +2560,9 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
         }
         outSeqs[i].offset = rawOffset;
         if (shouldUpdateRep) {
-            for (int i = ZSTD_REP_NUM - 1; i > 0; i--) {
-                rep[i] = rep[i - 1];
+            int j;
+            for (j = ZSTD_REP_NUM - 1; j > 0; j--) {
+                rep[j] = rep[j - 1];
             }
             rep[0] = outSeqs[i].offset;
         }
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2503,7 +2503,8 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
     assert(zc->seqCollector.seqIndex + 1 < zc->seqCollector.maxSequences);
     /* Ensure we have enough space for last literals "sequence" */
     assert(zc->seqCollector.maxSequences >= seqStoreSeqSize + 1);
-    ZSTD_memcpy(rep, repStartValue, ZSTD_REP_NUM * sizeof(U32));
+    ZSTD_memcpy(rep, zc->blockState.prevCBlock->rep, ZSTD_REP_NUM * sizeof(U32));
+    DEBUGLOG(2, "%u %u %u", rep[0], rep[1], rep[2]);
 
     for (i = 0; i < seqStoreSeqSize; ++i) {
         U32 rawOffset = seqStoreSeqs[i].offset - ZSTD_REP_NUM;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2522,6 +2522,7 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
 
         /* Derive the correct offset from the repcode in seqStore_t */
         if (seqStoreSeqs[i].offset <= ZSTD_REP_NUM) {
+            outSeqs[i].rep = seqStoreSeqs[i].offset;
             if (seqStoreSeqs[i].litLength != 0) {
                 if (seqStoreSeqs[i].offset == 1) {
                     shouldUpdateRep = 0;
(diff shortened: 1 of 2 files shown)


senhuang42:
discovered some issues, fixing now

terrelln:
Can you use `repcodes_t` and `ZSTD_updateRep()` instead of re-implementing the logic? 

https://github.com/facebook/zstd/blob/7205e609a91258c9c4b4757be5dd55a1b2b8507f/lib/compress/zstd_compress_internal.h#L388-L411

senhuang42:
> Can you use `repcodes_t` and `ZSTD_updateRep()` instead of re-implementing the logic?

Oh amazing, I guess it makes sense that the logic for this was centralized somewhere already

lib/compress/zstd_compress.c
```diff
@@ -2516,24 +2518,22 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
...
-            outSeqs[i].offset = repIdx >= 0 ? outSeqs[repIdx].offset : repStartValue[-repIdx - 1];
-            if (outSeqs[i].rep == 3 && outSeqs[i].litLength == 0) {
-                --outSeqs[i].offset;
-            }
-        } else {
-            outSeqs[i].offset = seqStoreSeqs[i].offset - ZSTD_REP_NUM;
         }
+        outSeqs[i].offset = rawOffset;
+        updatedRepcodes = ZSTD_updateRep(rep, seqStoreSeqs[i].offset - 1, seqStoreSeqs[i].litLength == 0);
+        ZSTD_memcpy(rep, updatedRepcodes.rep, sizeof(repcodes_t));
```
terrelln, line 2535: It would be slightly more efficient to do:
1. Copy `rep` into `updatedRepcodes` before the loop
2. Only used `updatedRepcodes` in the loop
3. Copy `updatedRepcodes` into `rep` after the loop

lib/compress/zstd_compress.c
```diff
@@ -2516,24 +2518,22 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
...
-            assert(repIdx >= -3);
-            outSeqs[i].offset = repIdx >= 0 ? outSeqs[repIdx].offset : repStartValue[-repIdx - 1];
-            if (outSeqs[i].rep == 3 && outSeqs[i].litLength == 0) {
-                --outSeqs[i].offset;
-            }
-        } else {
-            outSeqs[i].offset = seqStoreSeqs[i].offset - ZSTD_REP_NUM;
         }
+        outSeqs[i].offset = rawOffset;
+        updatedRepcodes = ZSTD_updateRep(rep, seqStoreSeqs[i].offset - 1, seqStoreSeqs[i].litLength == 0);
```
terrelln, line 2534: Can you add a comment explaining why you need to use `offset - 1` here.

lib/compress/zstd_compress.c
```diff
@@ -2547,6 +2549,7 @@ static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
     outSeqs[i].matchLength = outSeqs[i].offset = outSeqs[i].rep = 0;
     seqStoreSeqSize++;
 
+    ZSTD_memcpy(zc->blockState.nextCBlock->rep, updatedRepcodes.rep, sizeof(repcodes_t));
```
terrelln, line 2552: We shouldn't need this line. `nextCBlock->rep` is set by the block compressor already.

An assert that they are the same would be nice though.

Cyan4973 (approved):

Cyan4973:
Seems there are still a few differences detected ?

senhuang42:
> Seems there are still a few differences detected ?

Hmm, it seems like `ZSTD_updateRep()` is used only for the optimal parser and superblocks, and the other matchfinders have their own repcode finding logic. As such, I suspect at lower compression levels, the repcode table in the `blockState` might begin to diverge from the one that we are deriving in `copyBlockSequences()`.

Cyan4973:
Indeed, faster strategies only care about the 1st (mostly) and 2nd (sometimes) repcode.
They probably to not track the 3rd one for speed consideration.

senhuang42:
> Indeed, faster strategies only care about the 1st (mostly) and 2nd (sometimes) repcode.
> They probably to not track the 3rd one for speed consideration.

I've reverted the previous commit

terrelln (approved):
> I reverted the previous commit

That should be fine for correctness. However, I wonder if it might be better not to modify the repcodes. I wouldn't naively expect that `ZSTD_copyBlockSequences()` would modify the repcodes. If something changes in how zstd handles repcodes, e.g. the block compressor expects repcode 1 or 2 == 0 if they are out of bounds, this could mess that up.

senhuang42:
> > I reverted the previous commit
> 
> That should be fine for correctness. However, I wonder if it might be better not to modify the repcodes. I wouldn't naively expect that `ZSTD_copyBlockSequences()` would modify the repcodes. If something changes in how zstd handles repcodes, e.g. the block compressor expects repcode 1 or 2 == 0 if they are out of bounds, this could mess that up.



> > I reverted the previous commit
> 
> That should be fine for correctness. However, I wonder if it might be better not to modify the repcodes. I wouldn't naively expect that `ZSTD_copyBlockSequences()` would modify the repcodes. If something changes in how zstd handles repcodes, e.g. the block compressor expects repcode 1 or 2 == 0 if they are out of bounds, this could mess that up.

Actually yeah, I like that approach - if we don't update it, the locally derived repcodes within the block might still diverge from the "truth", but at the end of each block, we have the same ending/starting point as the block compressor does.

Cyan4973 (approved):

PR #2394, opened 2020-11-16
Fuzzer for sequence compression API

senhuang42:
This PR is rebased on top of #2387.

This adds a fuzzer for the new sequence compression API. Basically, the fuzzer generates some arbitrary, valid sequences based on some literal buffer, generates the source based on the sequences, and asks `ZSTD_compressSequences()` to roundtrip this source.

What is the process for adding this to OSS-Fuzz and properly integrating it into the fuzzer libraries/systems? Seems like something needs to be uploaded/downloaded.

Followups:
- Add fuzzing for sequences with block delimiter
- Add fuzzing for externally provided arbitrary repcodes (once `ZSTD_compressSequences()` supports it)

### DIFF
--- a/tests/fuzz/sequence_compression_api.c
+++ b/tests/fuzz/sequence_compression_api.c
@@ -59,7 +59,6 @@ static char *generatePseudoRandomString(char *str, size_t size) {
     const char charset[] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJK1234567890!@#$^&*()_";
     uint32_t seed = 0;
     if (size) {
-        --size;
         for (size_t n = 0; n < size; n++) {
             int key = FUZZ_RDG_rand(&seed) % (int) (sizeof charset - 1);
             str[n] = charset[key];
--- a/tests/fuzz/sequence_compression_api.c
+++ b/tests/fuzz/sequence_compression_api.c
@@ -134,6 +134,7 @@ static size_t generateRandomSequences(FUZZ_dataProducer_t* producer,
     uint32_t nbSeqGenerated = 0;
     uint32_t litLength;
     uint32_t matchLength;
+    uint32_t matchBound;
     uint32_t offset;
     uint32_t offsetBound;
     uint32_t repCode = 0;
@@ -143,6 +144,7 @@ static size_t generateRandomSequences(FUZZ_dataProducer_t* producer,
     while (nbSeqGenerated < ZSTD_FUZZ_MAX_NBSEQ 
          && bytesGenerated < ZSTD_FUZZ_GENERATED_SRC_MAXSIZE
          && !FUZZ_dataProducer_empty(producer)) {
+        matchBound = ZSTD_FUZZ_MATCHLENGTH_MAXSIZE;
         litLength = isFirstSequence && dictSize == 0 ? FUZZ_dataProducer_uint32Range(producer, 1, literalsSizeLimit)
                                                      : FUZZ_dataProducer_uint32Range(producer, 0, literalsSizeLimit);
         bytesGenerated += litLength;
@@ -151,7 +153,16 @@ static size_t generateRandomSequences(FUZZ_dataProducer_t* producer,
         }
         offsetBound = bytesGenerated > windowSize ? windowSize : bytesGenerated + dictSize;
         offset = FUZZ_dataProducer_uint32Range(producer, 1, offsetBound);
-        matchLength = FUZZ_dataProducer_uint32Range(producer, ZSTD_MINMATCH_MIN, ZSTD_FUZZ_MATCHLENGTH_MAXSIZE);
+        if (dictSize > 0 && bytesGenerated <= windowSize) {
+            uint32_t bytesToReachWindowSize = windowSize - bytesGenerated;
+            if (bytesToReachWindowSize < ZSTD_MINMATCH_MIN) {
+                offset = FUZZ_dataProducer_uint32Range(producer, 1, windowSize);
+            } else {
+                matchBound = bytesToReachWindowSize > ZSTD_FUZZ_MATCHLENGTH_MAXSIZE ?
+                             ZSTD_FUZZ_MATCHLENGTH_MAXSIZE : bytesToReachWindowSize;
+            }
+        }
+        matchLength = FUZZ_dataProducer_uint32Range(producer, ZSTD_MINMATCH_MIN, matchBound);
         bytesGenerated += matchLength;
         if (bytesGenerated > ZSTD_FUZZ_GENERATED_SRC_MAXSIZE) {
             break;
--- a/tests/fuzz/sequence_compression_api.c
+++ b/tests/fuzz/sequence_compression_api.c
@@ -9,8 +9,10 @@
  */
 
 /**
- * This fuzz target performs a zstd round-trip test (compress & decompress),
- * compares the result with the original, and calls abort() on corruption.
+ * This fuzz target performs a zstd round-trip test by generating an arbitrary
+ * array of sequences, generating the associated source buffer, calling
+ * ZSTD_compressSequences(), and then decompresses and compares the result with
+ * the original generated source buffer.
  */
 
 #define ZSTD_STATIC_LINKING_ONLY
@@ -30,12 +32,13 @@ static void* literalsBuffer = NULL;
 static void* generatedSrc = NULL;
 static ZSTD_Sequence* generatedSequences = NULL;
 
-#define ZSTD_FUZZ_GENERATED_SRC_MAXSIZE (1 << 25) /* Allow up to 32MB generated data */
+#define ZSTD_FUZZ_GENERATED_SRC_MAXSIZE (1 << 20) /* Allow up to 1MB generated data */
 #define ZSTD_FUZZ_MATCHLENGTH_MAXSIZE (1 << 18) /* Allow up to 256KB matches */
 #define ZSTD_FUZZ_GENERATED_DICT_MAXSIZE (1 << 18) /* Allow up to a 256KB dict */
 #define ZSTD_FUZZ_GENERATED_LITERALS_SIZE (1 << 18) /* Fixed size 256KB literals buffer */
 #define ZSTD_FUZZ_MAX_NBSEQ (1 << 17) /* Maximum of 128K sequences */
 
+/* Deterministic random number generator */
 #define FUZZ_RDG_rotl32(x,r) ((x << r) | (x >> (32 - r)))
 static uint32_t FUZZ_RDG_rand(uint32_t* src)
 {
@@ -50,7 +53,7 @@ static uint32_t FUZZ_RDG_rand(uint32_t* src)
 }
 
 /* Make a pseudorandom string - this simple function exists to avoid
- * taking a dependency on datagen.h to have RDG_genBuffer(). We don't need anything fancy.
+ * taking a dependency on datagen.h to have RDG_genBuffer().
  */
 static char *generatePseudoRandomString(char *str, size_t size) {
     const char charset[] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJK1234567890!@#$^&*()_";
@@ -124,7 +127,7 @@ static size_t decodeSequences(void* dst, size_t nbSequences,
     return generatedSrcBufferSize;
 }
 
-/* Returns nb sequences generated 
+/* Returns nb sequences generated
  * TODO: Add repcode fuzzing once we support repcode match splits
  */
 static size_t generateRandomSequences(FUZZ_dataProducer_t* producer,
@@ -141,7 +144,7 @@ static size_t generateRandomSequences(FUZZ_dataProducer_t* producer,
     uint32_t isFirstSequence = 1;
     uint32_t windowSize = 1 << windowLog;
 
-    while (nbSeqGenerated < ZSTD_FUZZ_MAX_NBSEQ 
+    while (nbSeqGenerated < ZSTD_FUZZ_MAX_NBSEQ
          && bytesGenerated < ZSTD_FUZZ_GENERATED_SRC_MAXSIZE
          && !FUZZ_dataProducer_empty(producer)) {
         matchBound = ZSTD_FUZZ_MATCHLENGTH_MAXSIZE;
@@ -154,9 +157,14 @@ static size_t generateRandomSequences(FUZZ_dataProducer_t* producer,
         offsetBound = bytesGenerated > windowSize ? windowSize : bytesGenerated + dictSize;
         offset = FUZZ_dataProducer_uint32Range(producer, 1, offsetBound);
         if (dictSize > 0 && bytesGenerated <= windowSize) {
-            uint32_t bytesToReachWindowSize = windowSize - bytesGenerated;
+            /* Prevent match length from being such that it would be associated with an offset too large
+             * from the decoder's perspective. If not possible (match would be too small),
+             * then reduce the offset if necessary.
+             */
+            size_t bytesToReachWindowSize = windowSize - bytesGenerated;
             if (bytesToReachWindowSize < ZSTD_MINMATCH_MIN) {
-                offset = FUZZ_dataProducer_uint32Range(producer, 1, windowSize);
+                uint32_t newOffsetBound = offsetBound > windowSize ? windowSize : offsetBound;
+                offset = FUZZ_dataProducer_uint32Range(producer, 1, newOffsetBound);
             } else {
                 matchBound = bytesToReachWindowSize > ZSTD_FUZZ_MATCHLENGTH_MAXSIZE ?
                              ZSTD_FUZZ_MATCHLENGTH_MAXSIZE : bytesToReachWindowSize;
@@ -241,8 +249,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *src, size_t size)
         dictBuffer = generatePseudoRandomString(dictBuffer, dictSize);
     }
     /* Generate window log first so we dont generate offsets too large */
-    wLog = FUZZ_dataProducer_uint32Range(producer, ZSTD_WINDOWLOG_MIN, ZSTD_WINDOWLOG_MAX);
-    cLevel = FUZZ_dataProducer_int32Range(producer, (int)ZSTD_minCLevel, (int)ZSTD_maxCLevel);
+    wLog = FUZZ_dataProducer_uint32Range(producer, ZSTD_WINDOWLOG_MIN, ZSTD_WINDOWLOG_MAX_32);
+    cLevel = FUZZ_dataProducer_int32Range(producer, -3, 22);
 
     if (!generatedSequences) {
         generatedSequences = FUZZ_malloc(sizeof(ZSTD_Sequence)*ZSTD_FUZZ_MAX_NBSEQ);
@@ -252,7 +260,6 @@ int LLVMFuzzerTestOneInput(const uint8_t *src, size_t size)
     }
     nbSequences = generateRandomSequences(producer, ZSTD_FUZZ_GENERATED_LITERALS_SIZE, dictSize, wLog);
     generatedSrcSize = decodeSequences(generatedSrc, nbSequences, ZSTD_FUZZ_GENERATED_LITERALS_SIZE, dictBuffer, dictSize);
-
     cBufSize = ZSTD_compressBound(generatedSrcSize);
     cBuf = FUZZ_malloc(cBufSize);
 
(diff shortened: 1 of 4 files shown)


terrelln:
@senhuang42 to fix the fuzzer tests:

```
# Generate a corpus, run for a few hours
./fuzz.py libfuzzer $FUZZER_NAME
# Minimize the corpus
./fuzz.py minimize $FUZZER_NAME
# Zip the corpus
./fuzz.py zip $FUZZER_NAME
```

Then upload generated the `.zip` file to this release https://github.com/facebook/zstd/releases/tag/fuzz-corpora. It should be named `${FUZZER_NAMEE}_seed_corpus.zip`.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+
+/* Make a pseudorandom string - this simple function exists to avoid
+ * taking a dependency on datagen.h to have RDG_genBuffer(). We don't need anything fancy.
+ */
+static char *generatePseudoRandomString(char *str, size_t size) {
+    const char charset[] = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJK1234567890!@#$^&*()_";
+    if (size) {
+        --size;
+        for (size_t n = 0; n < size; n++) {
+            int key = rand() % (int) (sizeof charset - 1);
```
terrelln, line 44: We can't use `rand()` here. This introduces non-determinism. Fuzzers need be deterministic so that when the fuzzer finds a bad input, we can always reproduce it.

If you want to take this approach look at the pseudo-random number generation in `fuzzer.c` and `zstreamtest.c`.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+                              const void* literals, size_t literalsSize, const void* dict, size_t dictSize) {
+    const uint8_t* ip = literals;
+    const uint8_t* dictPtr = dict;
+    uint8_t* op = dst;
+    size_t generatedSrcBufferSize = 0;
+    size_t bytesWritten = 0;
+
+    /* Note that src is a literals buffer */
+    for (size_t i = 0; i < nbSequences; ++i) {
+        assert(generatedSequences[i].matchLength != 0);
```
terrelln, line 62: Instead of assert use [`FUZZ_ASSERT()`](https://github.com/facebook/zstd/blob/dev/tests/fuzz/fuzz_helpers.h#L44). We want to make sure that these aren't compiled out of no-debug builds.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+    uint32_t matchLength;
+    uint32_t offset;
+    uint32_t offsetBound;
+    uint32_t repCode = 0;
+    uint32_t isFirstSequence = 1;
+    uint32_t windowSize = 1 << windowLog;
+
+    while (bytesGenerated < ZSTD_FUZZ_GENERATED_SRC_MAXSIZE && !FUZZ_dataProducer_empty(producer)) {
+        litLength = isFirstSequence ? FUZZ_dataProducer_uint32Range(producer, 1, literalsSize)
+                                    : FUZZ_dataProducer_uint32Range(producer, 0, literalsSize);
```
terrelln, line 119: Shouldn't it be more like `isFirstSequence && dictSize == 0`. We don't want to miss the edge case where we have a dictionary and have 0 literals.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+#include "fuzz_helpers.h"
+#include "zstd_helpers.h"
+#include "fuzz_data_producer.h"
+
+static ZSTD_CCtx *cctx = NULL;
+static ZSTD_DCtx *dctx = NULL;
+
+#define ZSTD_FUZZ_GENERATED_SRC_MAXSIZE (1 << 25) /* Allow up to 32MB generated data */
+#define ZSTD_FUZZ_MATCHLENGTH_MAXSIZE (1 << 18) /* Allow up to 256KB matches */
+#define ZSTD_FUZZ_GENERATED_LITERALS_MAXSIZE (1 << 19) /* Allow up to 512KB literals buffer */
```
terrelln, line 32: It would be nice to avoid this constraint. One easy solution, that wouldn't use more memory, would be to allow 256KB of literals, like matches, but then do something like:

```
if (litPtr + litSize > litEnd)
  litPtr = litBegin
// literals are [litPtr, litPtr + litSize)
litPtr ++ litSize
```

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+    ZSTD_CCtx_reset(cctx, ZSTD_reset_session_and_parameters);
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 0);
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel, cLevel);
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_windowLog, wLog);
+    /* TODO: Add block delim mode fuzzing */
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_blockDelimiters, ZSTD_sf_noBlockDelimiters);
+    if (hasDict) {
+        cdict = ZSTD_createCDict(dict, dictSize, cLevel);
+        FUZZ_ASSERT(cdict);
+        ZSTD_CCtx_refCDict(cctx, cdict);
```
terrelln, line 161: You could use `FUZZ_ZASSERT(ZSTD_CCtx_loadDictionary(cctx, dict, dictSize))`. That would be more concise.

Similarly you could use `ZSTD_DCtx_loadDictionary()` below.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+    void* literalsBuffer;
+    size_t literalsSize;
+    void* dictBuffer;
+    size_t dictSize = 0;
+    unsigned hasDict;
+    unsigned wLog;
+    int cLevel;
+
+    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(src, size);
+    literalsSize = FUZZ_dataProducer_uint32Range(producer, 1, ZSTD_FUZZ_GENERATED_LITERALS_MAXSIZE);
```
terrelln, line 203: I don't think we need this. It should be fine to create a `literalsBuffer` of a fixed size once and reuse it. This function gets called many times in the same process. To make the fuzzer run a bit faster we could do:

```c
// above with the cctx definition
static void* literalsBuffer = NULL;

// Inside LLVMFuzzerTestOneInput
if (literalsBuffer == NULL) {
  literalsBuffer = generatePseudoRandomString(literalsBuffer, ZSTD_FUZZ_GENERATED_LITERALS_MAXSIZE);
  FUZZ_ASSERT(literalsBuffer);
}
```

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+    if (hasDict) {
+        dictSize = FUZZ_dataProducer_uint32Range(producer, 1, ZSTD_FUZZ_GENERATED_DICT_MAXSIZE);
+        dictBuffer = FUZZ_malloc(dictSize);
+        dictBuffer = generatePseudoRandomString(dictBuffer, dictSize);
+    }
+    // Generate window log first so we dont generate offsets too large
+    wLog = FUZZ_dataProducer_uint32Range(producer, ZSTD_WINDOWLOG_MIN, ZSTD_WINDOWLOG_MAX);
+    cLevel = FUZZ_dataProducer_int32Range(producer, (int)ZSTD_minCLevel, (int)ZSTD_maxCLevel);
+
+    generatedSequences = FUZZ_malloc(sizeof(ZSTD_Sequence)*ZSTD_FUZZ_GENERATED_SRC_MAXSIZE);
```
terrelln, line 217: Woah this is a huge allocation. Like 512MB I think. Could we put a cap on the number of sequences, like 2^17?

Also, this allocation, and `generatedSrc` can be shared across multiple calls to `LLVMFuzzerTestOneInput()`.

Also, for all allocations you should `FUZZ_ASSERT(generatedSequences != NULL)`.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,258 @@
...
+    free(generatedSequences);
+    free(generatedSrc);
+    free(literalsBuffer);
+    FUZZ_dataProducer_free(producer);
+#ifndef STATEFUL_FUZZING
+    ZSTD_freeCCtx(cctx); cctx = NULL;
+    ZSTD_freeDCtx(dctx); dctx = NULL;
+#endif
+    return 0;
+}
```
terrelln, line 258: nit: newline at end of file.

tests/fuzz/sequence_compression_api.c
senhuang42: Ah, I hadn't considered that. It seems `srand()` wouldn't be sufficient either, since apparently the behavior of `rand()` still might differ on different systems. I'll use the same rng from the datagen library.

tests/fuzz/sequence_compression_api.c
senhuang42: Yikes you're right, that's a bit excessive.
Generally should we just make all of these buffers global to reduce allocations? They're all pretty big.

Is the additional `FUZZ_ASSERT()` necessary? It looks like `FUZZ_MALLOC()` has a `FUZZ_ASSERT()` in it, unless we're allocating 0 bytes.

senhuang42:
> @senhuang42 to fix the fuzzer tests:
> 
> # Generate a corpus, run for a few hours
> ./fuzz.py libfuzzer $FUZZER_NAME

@terrelln So when I run this, eventually it crashes:
`Tested 500 files: Failure!`
When I take a look at `corpora/sequence_compression_api-crash/` I don't see anything there, and when I manually run
`./sequence_compression_api corpora/sequence_compression_api-seed/z000*`
Nothing fails, and I get:
`Tested 500 files: Success!`

Is there something else I need to do as well?

tests/fuzz/sequence_compression_api.c
terrelln: > Generally should we just make all of these buffers global to reduce allocations? 

Allocations of a fixed size can be global. If they are dynamically sized they should be local. E.g. the buffer that we decompress into should be local because we want it to be exactly sized, so ASAN can catch buffer over-runs.

> Is the additional FUZZ_ASSERT() necessary? It looks like FUZZ_MALLOC() has a FUZZ_ASSERT() in it, unless we're allocating 0 bytes.a

Oh good point, I forgot about that!

terrelln:
> Edit: Figured it out - I needed to build with LIB_FUZZING_ENGINE not pointing to libregression.a

Yeah, if you follow the readme you build the fuzzers with:

```
CC=clang CXX=clang++ ./fuzz.py build all --enable-fuzzer
```

senhuang42:
Seems like there's something else I need to upload?

`Unable to download from: https://storage.googleapis.com/zstd-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/zstd_sequence_compression_api/public.zip.`

terrelln:
Looks like the MSAN fuzzer is finding a bug.

tests/fuzz/sequence_compression_api.c
```diff
@@ -0,0 +1,302 @@
...
+             */
+            size_t bytesToReachWindowSize = windowSize - bytesGenerated;
+            if (bytesToReachWindowSize < ZSTD_MINMATCH_MIN) {
+                uint32_t newOffsetBound = offsetBound > windowSize ? windowSize : offsetBound;
+                offset = FUZZ_dataProducer_uint32Range(producer, 1, newOffsetBound);
+            } else {
+                matchBound = bytesToReachWindowSize > ZSTD_FUZZ_MATCHLENGTH_MAXSIZE ?
+                             ZSTD_FUZZ_MATCHLENGTH_MAXSIZE : bytesToReachWindowSize;
+            }
+        }
```
terrelln, line 172: This is interesting and I think worth thinking about more closely. Because I think it could be exposing a real bug or limitation in the sequence compression API that this is hiding. And is worth thinking about for the block splitter.

Imagine we are at position `windowSize - 1` have match with offset `windowSize + 1` into the dictionary, and a match length of `windowSize`. This match cannot be split in two, because any splitting would result in either a too short match, or a match with an invalid offset.

This has ramifications for the sequence compression API. The only way for this match to be valid is to split a block at the beginning of the match at `windowSize - 1` where the match starts. This is an example of a valid set of sequences that can only be compressed one way.

If the match length was `windowSize + 1` the match would not be representable in the zstd format, and the sequence compression API would either have to return an error, or produce a corrupted block. What would it do in this case?

How do we want to handle this? CC @Cyan4973

senhuang42: I think this is pretty similar what I discuss here: https://github.com/facebook/zstd/pull/2387#issuecomment-728673874 and seems to be a non-encodable sequence?

So suppose `windowSize = 10`. At `matchPos: 9`, we have a match with `(of: 11, mLen: 10)`. If I understand Yann's response to the this issue I brought up, then the encoder actually can't generate this match, since at `matchPos == 10 == windowSize`, we have consumed `windowSize` bytes, and therefore the dictionary is no longer accessible, and at this point the offset is greater than windowSize which is not valid.

terrelln: Its not entirely clear to me if the sequence at `windowSize - 1` with `offset = windowSize + 1` and `matchLength = windowSize` with a block starting at `windowSize - 1` is:

1. Valid because at the position `windowSize - 1` the offset is valid, and we only enforce at the `offset` level in the spec.
2. Invalid because after the first byte of the match the offset `windowSize + 1` is invalid, even though it only becomes invalid in the middle of the sequence.

The spec isn't super clear here:

>As long as the amount of data decoded from this frame is less than or equal to Window_Size, sequence commands may specify **offsets** longer than the total length of decoded output so far to reference back to the dictionary, even parts of the dictionary with offsets larger than Window_Size. After the total output has surpassed Window_Size however, this [**offsets longer than Window_Size**] is no longer allowed and the dictionary is no longer accessible.

It seems to imply the first condition. But, it could probably be read either way.

terrelln: What does the sequence compression API do it it comes across this case? Do we say "it is undefined behavior if your sequences are invalid and we reserve the right to produce a corrupted block", or do we check for valid sequences?

senhuang42: I also read it in the first condition at first, so it probably deserves a rewording if it's not the case. When the fuzzer originally produced a case like this, on the decoder side, attempting to decode a case like this failed at least in the debug modes where validations were turned on.

terrelln: > When the fuzzer originally produced a case like this, on the decoder side, attempting to decode a case like this failed at least in the debug modes where validations were turned on.

Did the sequence API split the sequence in two though? That would certainly be invalid.

terrelln: There are two intertwined issues:

1. What we decide about this specific case, is the sequence valid or not.
2. In general, what is the contract of the sequence compression API when it is presented invalid sequences. Do we silently produce a corrupted block, or do we return an error, or something in between?

senhuang42: > Did the sequence API split the sequence in two though? That would certainly be invalid.

No, when we might split a block in a way that produces a sequence of matchlength too small, we will avoid splitting at that point, and attempt to split at an earlier position. In this particular case, since it's the first half of the match that's too small, we'd move the end of the block to the beginning of this match.

> 2. In general, what is the contract of the sequence compression API when it is presented invalid sequences. Do we silently produce a corrupted block, or do we return an error, or something in between?

In debug levels that are high enough to trigger `RETURN_ERROR_IF()` statements, we will spit out an error for cases where offsets are not valid, or matchlengths are too small. However, in general operation, this will produce a corrupted block, since we're generally presuming the matchfinder to provide valid sequences, and only enabling validation in debug/development mode.

terrelln: `RETURN_ERROR_IF()` is always enabled.

terrelln: Which assert fails in the decoder?

senhuang42: > `RETURN_ERROR_IF()` is always enabled.
Oh yeah, you're right - I guess the full error message forwarding doesn't show up without the higher debug levels.

> Which assert fails in the decoder?
If I remember correctly: `assert((size_t)(oend - op) >= sequenceSize);` in `ZSTD_assertValidSequence()`

terrelln: Neither of those asserts should be triggering for the scenario I'm describing. In this scenario `ZSTD_dictionaryIsActive()` would return true here

https://github.com/facebook/zstd/blob/9736c497ce7db2351fc4a170da31433209499868/lib/decompress/zstd_decompress_block.c#L1085

senhuang42: Ah, I believe the exact scenario was ever so slightly different when I triggered it:

We only split matches if they're larger than `windowSize`. So suppose that your `matchlength` was greater than `windowSize`. (of: 11, mLen: 15) at `matchPos: 9` with `windowSize = blockSize = 10`. In this case, what happens is that the API will split at `pos: 9` and `pos: 19`. So that leaves us with `pos: 9 (of: 11 mlen: 10)` and `pos: 19 (of: 11 mlen: 5`. In which case, the second match wouldn't be valid, since by then we've read more than `windowSize` bytes. 

But in this case, we have `matchLength == windowSize`. So I suppose that the `assert()` would not fail here.

Cyan4973: > Imagine we are at position `windowSize - 1` have match with offset `windowSize + 1` into the dictionary

This situation is ___invalid___.

At the very moment the decoder has generated `windowSize` bytes, the dictionary becomes inaccessible,
even if this happens in the middle of a match.

> The spec isn't super clear here:

I agree.
The spec implies that the condition is applied to the first byte of the match (where the match starts), while it should be applied to the _last_ byte of the match (where the match ends).

> If the match length was `windowSize + 1`

This cannot be represented as a single match in `zstd` : `matchLength` is necessarily `<= blockSize`, which itself is necessarily `<= windowSize`.

terrelln: > This cannot be represented as a single match in zstd : matchLength is necessarily <= blockSize, which itself is necessarily <= windowSize.

Yeah, the question with `matchLength = windowSize + 1` was what would the sequence compression API do about it? What is our contract when the user presents invalid sequences?

senhuang42: I guess currently we don't validate that particular case - see: https://github.com/facebook/zstd/pull/2387/files#diff-0c010843fdcc669acc5d75bdd2d3d1bd7aaba7f869394ec85570321fc1b1ee76R4496

I can definitely add that in as another condition though. I don't think that function has any real impact on the speed, so we can leave it as-is and always validate, or I'm guessing the way we hide the validations to be just in debug levels is to add some `if defined DEBUGLEVEL` statements.

Cyan4973: > Yeah, the question with `matchLength = windowSize + 1` was what would the sequence compression API do about it? What is our contract when the user presents invalid sequences?

It depends if sequence verification is enabled or not.

If verification is enabled, the sequence should be detected as invalid, and the function should return an error code.
If verification is disabled, all sequences are presumed "valid" and blindly encoded, hence if they are not, it's UB.

From the perspective of the first target user of this API, this verification isn't useful, because their generator is presumed to always generate valid sequences.
However, I believe it should at least be possible to enable verification, even if it's not a "default" setting, because it's clearly useful during development or debug sessions.

A follow-up is that, if verification cost is basically insensible, we could make it default instead, and allow it to be disabled for some strange corner case that we can't guess now that could require it (like, for example, intentionally generating invalid encoded sequences).

terrelln: Yeah, a verification flag makes sense to me.

terrelln (approved):

tests/fuzz/sequence_compression_api.c
senhuang42: > Yeah, a verification flag makes sense to me.

What is the best mechanism for something like this, if not through `if defined (DEBUGLEVEL)` like `assertValidMatch()` in the decoder? I don't want to add to the public function signature, nor add yet another experimental param to the cctx params?

> Therefore, a sequence with matchLength = windowSize + 1 is not invalid

Right, that's what a lot of the logic in `ZSTD_copySequencesToSeqStore()` deals with - splitting these matches sanely.

Cyan4973: > I don't want to add to the public function signature, nor add yet another experimental param to the cctx params?

Don't be shy about adding an experimental param to `cctx` params.
That's why this flexible mechanism was created.
We can always remove or update experimental params later on.

PR #2395, opened 2020-11-19
10x speedup for ZSTD_isRLE()

senhuang42:
Clickbait title, but I noticed that `ZSTD_isRLE()` is unoptimized. @terrelln suggested unrolling the loop and showed me an unrolled version of `ZSTD_isRLE()` that's much faster (most of the code here he wrote!), and I've made some minor tweaks.

Using `fullbench.c` to do a microbenchmark on just the function with a buffer of all 0:
Raw input: 1000 bytes 
Old version: ~3700MB/s
New version: ~35000MB/s

Raw input: 100000 bytes
Old version: ~4500MB/s
New version: ~45000MB/s

And actual compression:
When compressing a file of all 'a', 150K, compression level 1:
Old version: ~10000MB/s
New version: ~14000MB/s

When compressing a file of all 'a', 7MB, compression level 1:
Old version: ~3000MB/s
New version: ~11000MB/s

These not really realistic files, but as expected, compression speed is a bit better on files that are just 1 character. This PR will mostly be useful because it should speed up ASAN a lot on cases that would time out in the sequence compression fuzzer.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2586,12 +2586,27 @@ size_t ZSTD_mergeBlockDelimiters(ZSTD_Sequence* sequences, size_t seqsSize) {
     return out;
 }
 
-/* Returns true if the given block is a RLE block */
-static int ZSTD_isRLE(const BYTE *ip, size_t length) {
+/* Unrolled loop to read four size_ts of input at a time. Returns 1 if is RLE, 0 if not. */
+static int ZSTD_isRLE(const BYTE* src, size_t length) {
+    const BYTE* ip = src;
+    const BYTE value = ip[0];
+    const size_t valueST = value * 0x0101010101010101ULL;
+    const size_t unrollSize = sizeof(size_t) * 4;
+    const size_t unrollMask = unrollSize - 1;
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

terrelln (approved):
Awesome!

lib/compress/zstd_compress.c
```diff
@@ -2586,12 +2586,27 @@ size_t ZSTD_mergeBlockDelimiters(ZSTD_Sequence* sequences, size_t seqsSize) {
     return out;
 }
 
-/* Returns true if the given block is a RLE block */
-static int ZSTD_isRLE(const BYTE *ip, size_t length) {
+/* Unrolled loop to read four size_ts of input at a time. Returns 1 if is RLE, 0 if not. */
+static int ZSTD_isRLE(const BYTE* src, size_t length) {
+    const BYTE* ip = src;
+    const BYTE value = ip[0];
+    const size_t valueST = value * 0x0101010101010101ULL;
```
terrelln, line 2593: You'll need to make this an explicit cast to silence the visual warnings.
```suggestion
    const size_t valueST = (size_t)((U64)value * 0x0101010101010101ULL));
```

PR #2399, opened 2020-11-24
Update cmake_minimum_required to 2.8.12

brawner:
This has to do with a deprecated policy with the recent release of cmake 3.19.1. A policy about `INTERFACE_LINK_LIBRARIES` has been deprecated for versions `< 2.8.12`. The deprecation will produce warnings if an older version is used, or the policy is set to `OLD`. 

More info about the deprecated policy:
https://cmake.org/cmake/help/latest/policy/CMP0022.html

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -7,7 +7,7 @@
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
 
-cmake_minimum_required(VERSION 2.8.9 FATAL_ERROR)
+cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
   
 # As of 2018-12-26 ZSTD has been validated to build with cmake version 3.13.2 new policies. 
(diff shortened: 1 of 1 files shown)


brawner:
It doesn't look like the test failure is related. Importantly the cmake job succeeded.

Test failure output
```
Seed = 6964

     1/     1    
    26          
    30          
    38          
    46          
    63          
    69          
    78          
    83          
    88          
   100          
   101          Error => Compression is not deterministic! (seed 6964, test nb 117, line 2329) 
make[1]: *** [test-zstream] Error 1
```

Cyan4973:
The test error is unrelated

Cyan4973 (approved):
I'm fine with the suggested change.
The main requirement for `2.8.9` compatibility was to remain compatible with base version of `cmake` in Ubuntu Precise, which is no longer maintained.

We can now move up to `2.8.12` which corresponds to the base version in Ubuntu Trusty, still in extended support mode.

PR #2401, opened 2020-11-27
Attemps to  "meson: Deny shared build on MSVC compiler"

animalize:


lib/common/xxhash.c
```diff
@@ -459,7 +459,7 @@ FORCE_INLINE_TEMPLATE U64 XXH64_endian_align(const void* input, size_t len, U64
 }
 
 
-XXH_PUBLIC_API unsigned long long XXH64 (const void* input, size_t len, unsigned long long seed)
+XXHASH_DLL_PUBLIC XXH_PUBLIC_API unsigned long long XXH64 (const void* input, size_t len, unsigned long long seed)
```
Cyan4973, line 462: Why do you want to make `XXH64` public ?

lib/compress/zstd_compress_internal.h
```diff
@@ -1173,6 +1182,6 @@ size_t ZSTD_referenceExternalSequences(ZSTD_CCtx* cctx, rawSeq* seq, size_t nbSe
 
 /** ZSTD_cycleLog() :
  *  condition for correct operation : hashLog > 1 */
-U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat);
+ZSTD_COMPRESS_DLL_PUBLIC U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat);
```
Cyan4973, line 1185: Why should `ZSTD_cycleLog()` become public ?

lib/dictBuilder/zdict.c
```diff
@@ -969,15 +969,16 @@ static size_t ZDICT_addEntropyTablesFromBuffer_advanced(
 }
 
 /* Hidden declaration for dbio.c */
-size_t ZDICT_trainFromBuffer_unsafe_legacy(
+ZDICTLIB_API size_t ZDICT_trainFromBuffer_unsafe_legacy(
```
Cyan4973, line 972: I'm surprised, normally it's enough for `ZDICTLIB_API` to be on the declaration side (`zdict.h`). It shouldn't be needed on the implementation side (`zdict.c`).

animalize:
Only the last commit is from me, it makes the CI green.
This PR is test for CI, as I said in https://github.com/lzutao/zstd/pull/2, I will close this PR.
No need to review in this one, sorry I didn't say it clearly.
@lzutao

animalize:
I will try to solve them later.

There is no need to work these days, I read some news and understand these days are very bad, rest more, these works can be done later.

animalize:
@lzutao 
It seems all changes in `.c` `.h` files can be removed.

PR #2405, opened 2020-11-30
Sequence Compression API - make validator use cctx minMatch instead of global MINMATCH

senhuang42:
The validator should be stricter - we should use the value of `minMatch` in the compression params rather than the global value of 3, since `minMatch` can actually change to higher values.

Also updates fuzzer to validate the sequences.


terrelln:
QQ: If someone violates this assumption and produces too many sequences with validation turned off, what will the sequence compression API do? Return an error or segfault/assert failure? We should ensure that even in this case we never segfault. When validation is disabled it is okay to produce an invalid compressed blob, but we should never segfault/assert.

terrelln:
Also, can you document that `ZSTD_c_minMatch` must be set to the correct lower bound? By default, the compression level will be 3, which has a fairly large `minMatch`, so this is something that people will probably run into.

senhuang42:
> we should never segfault/assert.

The assert that was failing prior to the fix was that the maximum number of sequences allowed by the `seqStore` was being exceeded, since we allocate space for the sequences based on some multiple of `blockSize/minMatch`. 

For example, if a `seqStore` allocated assuming 4 is the `minMatch`, but in reality every single match has `minMatch == 3` means we will exceed the maximum number of allowable sequences. I don't really see a clean way around this if someone violates this assumption, other than always forcing the `minMatch` of the `cctx` to be 3, which means we always will allocate enough space. It also means that the user wouldn't need to ensure that the cctx's `minMatch` matches what their generator creates. But then, that kind of takes away some of the choice on the user's end if they want to use less memory.

Other than that though, I believe it wouldn't even produce a corrupted blob, since we don't really care about what the `minMatch` is, other than for the space allocation.

terrelln:
What I mean is that it is never okay for a user input to cause a assert/segfault. So we will need to validate that the number of sequences that they pass in can fit in our seqstore, and return an error if it doesn't.

When validate sequences is off, the API contract says we are allowed to produce corrupted data if the sequences are invalid. However, we are still never allowed to crash.

terrelln (approved):

Issue #2335, opened 2020-09-30
ZSTD_seekable_decompress() can hang


mdittmer:
**Describe the bug**
`ZSTD_seekable_decompress()` can hang.

**To Reproduce**
Checkout #2334 

```
$ cd contrib/seekable_format/examples
$ make
$ ./hang
```

**Expected behavior**
Program halts, returning some ZSTD error as status code.

**Additional context**
This bug was found via [this fuzzer](https://fuchsia.googlesource.com/fuchsia/+/6c0911153a4b36d5a6ab86f74cc9d1a313d8e8ca/src/storage/blobfs/test/zstd-seekable-fuzzer.cc). The specific decompressor input is an input generated by the fuzzer to reproduce the issue.

senhuang42:
Resolved in #2408

PR #2408, opened 2020-12-02
Remove possibility of hanging when using seekable decompression

senhuang42:
Issue #2335 from @mdittmer noted that seekable decompress can hang, and provided a useful repro in #2334

The root cause is in the main loop of `ZSTD_seekable_decompress()`, which can get stuck in an infinite loop if `zs->decompressedOffset` never increases, and the initial `offset != 0`. 

Since the decompressed offset never increases, we will execute code from lines `385-395`, which resets the stream object, so when we finish decompressing the whole buffer, we will just start all over again, into infinity.

So here, I add a requirement (similar to the `ZSTD_decompressStream()` API) where after a certain number of consecutive times where no forward progress is made, we bail out. 

It's not clear to me if it's ever valid to have many consecutive calls where no progress is made, and if so, maybe we need another approach? However, this approach should also just prevent hangs in general too, since the way the loop is written in general seems not entirely safe against infinite loops, and perhaps there are other ways to trigger the bug too.

Test Plan:
- Validated change works against bug repro case, tried using some random files to make sure nothing else was affected.

### DIFF
--- a/contrib/seekable_format/tests/Makefile
+++ b/contrib/seekable_format/tests/Makefile
@@ -13,7 +13,7 @@ ZSTDLIB_PATH = ../../../lib
 ZSTDLIB_NAME = libzstd.a
 ZSTDLIB = $(ZSTDLIB_PATH)/$(ZSTDLIB_NAME)
 
-CPPFLAGS += -I../ -I../../../lib -I../../../lib/common
+CPPFLAGS += -I../ -I$(ZSTDLIB_PATH) -I$(ZSTDLIB_PATH)/common
 
 CFLAGS ?= -O3
 CFLAGS += -g
@@ -22,15 +22,15 @@ SEEKABLE_OBJS = ../zstdseek_compress.c ../zstdseek_decompress.c $(ZSTDLIB)
 
 .PHONY: default clean test
 
-default: test
+default: seekable_tests
 
 test: seekable_tests
+	./seekable_tests
 
 $(ZSTDLIB):
-	make -C $(ZSTDLIB_PATH) $(ZSTDLIB_NAME)
+	$(MAKE) -C $(ZSTDLIB_PATH) $(ZSTDLIB_NAME)
 
 seekable_tests : seekable_tests.c $(SEEKABLE_OBJS)
-	$(CC) $(CPPFLAGS) $(CFLAGS) $^ $(LDFLAGS) -o $@
 
 clean:
 	@rm -f core *.o tmp* result* *.zst \
--- a/contrib/seekable_format/tests/Makefile
+++ b/contrib/seekable_format/tests/Makefile
@@ -1,0 +1,38 @@
+# ################################################################
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under both the BSD-style license (found in the
+# LICENSE file in the root directory of this source tree) and the GPLv2 (found
+# in the COPYING file in the root directory of this source tree).
+# ################################################################
+
+# This Makefile presumes libzstd is built, using `make` in / or /lib/
+
+ZSTDLIB_PATH = ../../../lib
+ZSTDLIB_NAME = libzstd.a
+ZSTDLIB = $(ZSTDLIB_PATH)/$(ZSTDLIB_NAME)
+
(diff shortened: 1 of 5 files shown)


Cyan4973:
Would it be possible to add a test ?

senhuang42:
> Would it be possible to add a test ?

I was wondering about this - I don't see any sort of test directory/suite in `contrib/` or `contrib/seekable_format` for this whole directory - where would a test go?

Cyan4973:
It would have to be part of `contrib/seekable_format`.
I see a directory `examples`, which has been probably used as a kind of "minimal test" so far,
but it's not proper for a test targeting this fix.
So probably create a new `tests` subdirectory.

senhuang42:
Added basic unit testing for seekable fmt - this isn't integrated into the CI on github for consistency with other directories in `contrib/` that aren't using CI (and cause this is just a single test). But we could add that if we want.

Cyan4973:
Yes, please add it to CI,
otherwise this test will never be played and rot at some point.

`contrib` tests area actually played during CI,
see : https://github.com/facebook/zstd/blob/dev/Makefile#L119

They are mostly build tests, but nothing prevents from adding some runtime tests.
For example, `single_file_libs` builds and runs tests.

contrib/seekable_format/tests/Makefile
```diff
@@ -0,0 +1,38 @@
...
+.PHONY: default clean test
+
+default: test
+
+test: seekable_tests
+
+$(ZSTDLIB):
+	make -C $(ZSTDLIB_PATH) $(ZSTDLIB_NAME)
+
+seekable_tests : seekable_tests.c $(SEEKABLE_OBJS)
```
Cyan4973, line 32: minor : since `seekable_tests` is both the name of the final binary and the name of its main unit, I think this recipe could be simplified to : 
```
seekable_tests : $(SEEKABLE_OBJS)

```

with nothing else. The implicit rules should build it correctly.

contrib/seekable_format/tests/Makefile
```diff
@@ -0,0 +1,38 @@
...
+SEEKABLE_OBJS = ../zstdseek_compress.c ../zstdseek_decompress.c $(ZSTDLIB)
+
+.PHONY: default clean test
+
+default: test
+
+test: seekable_tests
+
+$(ZSTDLIB):
+	make -C $(ZSTDLIB_PATH) $(ZSTDLIB_NAME)
```
Cyan4973, line 30: `make` ==> `$(MAKE)`

(note: on FreeBSD, it's actually `gmake`, `$(MAKE)` abstracts such differences)

contrib/seekable_format/tests/Makefile
```diff
@@ -0,0 +1,38 @@
...
+CFLAGS ?= -O3
+CFLAGS += -g
+
+SEEKABLE_OBJS = ../zstdseek_compress.c ../zstdseek_decompress.c $(ZSTDLIB)
+
+.PHONY: default clean test
+
+default: test
+
+test: seekable_tests
```
Cyan4973, line 27: You could use this recipe to both build _and_ then run `seekable_tests`

contrib/seekable_format/tests/Makefile
```diff
@@ -0,0 +1,38 @@
...
+# in the COPYING file in the root directory of this source tree).
+# ################################################################
+
+# This Makefile presumes libzstd is built, using `make` in / or /lib/
+
+ZSTDLIB_PATH = ../../../lib
+ZSTDLIB_NAME = libzstd.a
+ZSTDLIB = $(ZSTDLIB_PATH)/$(ZSTDLIB_NAME)
+
+CPPFLAGS += -I../ -I../../../lib -I../../../lib/common
```
Cyan4973, line 16: minor :
reuse `ZSTDLIB_PATH` : `-I$(ZSTDLIB_PATH)` (instead of `-I../../../lib`)

contrib/seekable_format/tests/Makefile
senhuang42: Oh interesting, thanks for all the Makefile tips!

Cyan4973 (approved):

PR #2412, opened 2020-12-03
use ZSTD_CLEVEL_DEFAULT in zdict.c

animalize:
This PR is questionable.
Should use `ZSTD_CLEVEL_DEFAULT` in `zdict.c` file?

For example:
1, A man sets `ZSTD_CLEVEL_DEFAULT = 6` in his build.
2, He uses public function `ZDICT_trainFromBuffer()` to train a dictionary.
3, He gives the dictionary to others.
Then this dictionary has a non-standard level, is this ok?

### DIFF
--- a/lib/dictBuilder/zdict.c
+++ b/lib/dictBuilder/zdict.c
@@ -62,7 +62,6 @@
 
 #define NOISELENGTH 32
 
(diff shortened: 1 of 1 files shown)


terrelln (approved):
Thanks for the PR @animalize!

PR #2416, opened 2020-12-05
removed (commented) icc test from github action

Cyan4973:
currently failing, and producing noise
needs further investigation


bimbashrestha:
It looks like the `intel-oneapi` was split into different packages on ubuntu recently. We can fix this by installing both `intel-basekit` and `intel-hpckit` (this has the icc compiler now). This [commit](https://github.com/bimbashrestha/zstd/pull/56/files) fixed it on my [fork](https://github.com/bimbashrestha/zstd/pull/56/checks?check_run_id=1503980352).

I can create a PR with the fix if we want to keep icc around still?

Cyan4973:
by all means @bimbashrestha , you are very welcome !

bimbashrestha:
PR: https://github.com/facebook/zstd/pull/2417

PR #2417, opened 2020-12-05
[build] Fix icc build. Using updated ubuntu packages

bimbashrestha:
Intel-oneapi doesn't exist anymore. Instead, it looks like it's been split into specific packages. Just needed to install the new ones:)


Cyan4973 (approved):

Cyan4973:
Thanks @bimbashrestha !

Issue #2286, opened 2020-08-26
ASAN poisoning doesn't work with static context allocation


terrelln:
1. Use `ZSTD_initStaticCStream(workspace, workspaceSize)`
2. Compress
3. Reuse the same workspace in `ZSTD_initStaticDStream()`
4. Get a use-after-poison ASAN error

Can be worked around by defining `ZSTD_ASAN_DONT_POISON_WORKSPACE`

PR #2419, opened 2020-12-08
Don't ASAN-Poison Statically-Allocated Workspaces

felixhandte:
Addresses #2286.

Pretty straightforward, we keep the redzones when built with ASAN so the size estimation code still works, but disable poisoning and unpoisoning when the workspace is statically allocated. We still do MSAN poisoning though, which should be safe.

### DIFF
--- a/lib/compress/zstd_cwksp.h
+++ b/lib/compress/zstd_cwksp.h
@@ -45,6 +45,16 @@ typedef enum {
 } ZSTD_cwksp_alloc_phase_e;
 
 /**
+ * Used to describe whether the workspace is statically allocated (and will not
+ * necessarily ever be freed), or if it's dynamically allocated and we can
+ * expect a well-formed caller to free this.
+ */
+typedef enum {
+    ZSTD_cwksp_dynamic_alloc,
+    ZSTD_cwksp_static_alloc
+} ZSTD_cwksp_static_alloc_e;
+
+/**
  * Zstd fits all its internal datastructures into a single continuous buffer,
  * so that it only needs to perform a single OS allocation (or so that a buffer
  * can be provided to it and it can perform no allocations at all). This buffer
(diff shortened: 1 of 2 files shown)


terrelln (approved):

lib/compress/zstd_cwksp.h
```diff
@@ -447,24 +462,25 @@ MEM_STATIC void ZSTD_cwksp_clear(ZSTD_cwksp* ws) {
...
     ws->tableValidEnd = ws->objectEnd;
     ws->phase = ZSTD_cwksp_alloc_objects;
+    ws->isStatic = !!isStatic;
     ZSTD_cwksp_clear(ws);
     ws->workspaceOversizedDuration = 0;
     ZSTD_cwksp_assert_internal_consistency(ws);
 }
 
-MEM_STATIC size_t ZSTD_cwksp_create(ZSTD_cwksp* ws, size_t size, ZSTD_customMem customMem) {
+MEM_STATIC size_t ZSTD_cwksp_create(ZSTD_cwksp* ws, size_t size, ZSTD_customMem customMem, int isStatic) {
```
Cyan4973, line 479: This function uses `ZSTD_customMalloc()`.
It seems this should rule out `isStatic` ?

felixhandte: Oh yeah. That's a good point. I'll fix that.

Cyan4973 (approved):

Cyan4973:
As a very minor complementary comment :

note how you have to comment what `0`/`1` means (`1 /* static */`, ` 0 /* not static */`)
since the meaning of this parameter is not obvious in the calling context.
The presence of comment itself is nice, I just guess we could do even better.

Faced with some other similar situation, it's generally best to create an `enum` for this purpose,
so that the meaning of the parameter is explicitly spelled within the invocation.
The issue then is to find a good `enum` name, expressive, concise, and which avoids namespace collision.

This comment doesn't block since PR which is already validated.

Cyan4973 (approved):

Issue #2312, opened 2020-09-16
`EXEC_PREFIX` and `PREFIX` are interpreted as regexes resulting in build failures in some cases


jmesmon:
**Describe the bug**

In [lib/Makefile](https://github.com/facebook/zstd/blob/d96e98cfde66e9e20dcadcfd9ed3b82ba648adfe/lib/Makefile#L273-L290), `PCLIBDIR` and `PCINCDIR` assignments treat `EXEC_PREFIX` and `PREFIX` as regular expressions, which can result in `PCLIBDIR` (and/or `PCINDIR`) testing as empty, and then triggering the later errors.

I observed this within a oe build that had a `prefix` with a `+` in it (which is common for oe sdk packages)

**To Reproduce**
1. `make PREFIX=$HOME/xxx/x+foo`
2. it shows:

```
make[1]: Entering directory '/home/cody/d/zstd/lib'
Makefile:280: *** configured libdir (/home/cody/xxx/x+foo/lib) is outside of prefix (/home/cody/xxx/x+foo), can't generate pkg-config file.  Stop.
make[1]: Leaving directory '/home/cody/d/zstd/lib'
make: *** [Makefile:63: lib-release] Error 2
```

**Expected behavior**

Build succeeds. (ie: what happens if `+` is replaced with a normal character)

**Additional info**

In oe, I've worked around this with the `zstd_%.bbappend` fragment to set the `PCLIBDIR`/`PCINCDIR` directly:
```
export PCLIBDIR="${@oe.path.relative("${exec_prefix}", "${libdir}")}"
export PCINCDIR="${@oe.path.relative("${prefix}", "${includedir}")}"
```

jmesmon:
Also notable: the error messages here appear to print the wrong `prefix` vs `exec_prefix` value. This initially threw me off, but in my case they're set to the same value anyhow.

felixhandte:
Should be fixed in `dev`.

Issue #2192, opened 2020-06-04
.pc file is broken when generated by CMake with explicit CMAKE_INSTALL_*DIR


tobim:
**Describe the bug**
When CMake is used to build zstd, it can be configured with explicit install paths for the includes, libraries, binaries and so on. When one of these paths is absolute, the generated `libzstd.pc` contains wrong fields.

**To Reproduce**
```sh
$ mkdir build_ && cd $_
$ cmake ../build/cmake/ -DCMAKE_INSTALL_PREFIX=/tmp/usr -DCMAKE_INSTALL_INCLUDEDIR=/tmp/opt/include -DCMAKE_INSTALL_LIBDIR=/tmp/foobar/lib
$ cmake --build . --target all
$ cmake --build . --target install
$ cat /tmp/foobar/lib/pkgconfig/libzstd.pc
#   ZSTD - standard compression algorithm
#   Copyright (C) 2014-2016, Yann Collet, Facebook
#   BSD 2-Clause License (http://www.opensource.org/licenses/bsd-license.php)

prefix=/tmp/usr
exec_prefix=${prefix}
includedir=${prefix}//tmp/opt/include
libdir=${exec_prefix}//tmp/foobar/lib

Name: zstd
Description: fast lossless compression algorithm library
URL: http://www.zstd.net/
Version: 1.4.5
Libs: -L${libdir} -lzstd
Cflags: -I${includedir}
```

**Expected behavior**
```
...
prefix=/tmp/usr
exec_prefix=${prefix}
includedir=/tmp/opt/include
libdir=/tmp/foobar/lib
...
```

tobim:
CMake has `CMAKE_INSTALL_FULL_*DIR` variables that can be used instead, but the `${prefix}` parts would need to be removed from `includedir` and `libdir`.

There have been quite a few PRs regarding pkgconfig recently, and I don't know the motivation and requirements, but it seems to me that the variable preparation logic in both the Makefile and CMake based builds are too complicated.

felixhandte:
@tobim, given the apparently mutually exclusive needs we've gotten from users of our `pkg-config` file, we've struggled to find a solution that supports everyone's use cases. In particular, we were requested to always leave `${prefix}` in, in #1794. I'm not sure how we could accommodate that use case along with this at the same time. Any suggestions?

tobim:
[...]

PR #2421, opened 2020-12-09
Don't Use Regexes to Build Pkg-Config File

felixhandte:
Instead of using `sed` to manipulate the lib and include dirs, this PR causes make and cmake to use their native language features to compare the `prefix` /`exec_prefix` to the prefix of those paths, and if they match, rewrite that prefix into a reference to the `prefix`/`exec_prefix` variables.

This means that if `includedir` and `libdir` are not children of `prefix`/`exec_prefix`, we'll fall back to just emitting a whole, absolute path, rather than failing.

This PR addresses #2192 and #2312.

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -137,12 +137,33 @@ endif ()
 if (UNIX OR MINGW)
     # pkg-config
     set(PREFIX "${CMAKE_INSTALL_PREFIX}")
-    set(LIBDIR "${CMAKE_INSTALL_LIBDIR}")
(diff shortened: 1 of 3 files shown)


terrelln (approved):
It looks like cmake doesn't support setting the `exec_prefix` is that correct?

Other than that question, LGTM

felixhandte:
@terrelln, I don't see any way to do so in the [conventional cmake variables](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html). And the current code doesn't allow configuring it. If needed, we can always add that pretty easily I think.

Cyan4973 (approved):

PR #2425, opened 2020-12-14
add release dates to CHANGELOG

animalize:
CHANGELOG doesn't have date, which is a bit inconvenient for readers.

The dates are from https://github.com/facebook/zstd/releases, I checked them twice.
v0.7.0 is missing in that page, so I used v0.7.0 tag's date.


Cyan4973:
A few words of context :
by the time the `CHANGELOG` is updated,
the date of next release is generally not yet guaranteed (just around the corner).
So, given an objective of "exact" date, it's not possible to write it while updating the `CHANGELOG`.

That being said, the `CHANGELOG` could be updated after the release to reflect the effective date of latest publication.

Cyan4973 (approved):

PR #2426, opened 2020-12-15
Updated CHANGELOG for v1.4.7

Cyan4973:
@felixhandte , @terrelln , @senhuang42 ,
please have a look to make sure that all topics you want to be mentioned are effectively present.

A few additional notes : 

- the next release is presumed to be numbered `v1.4.7`.
This will automatically bring the question "what about `v1.4.6` ?".
I think we can mention that `v1.4.6` is a non-public release.
If that feels too much hassle, we could also decide to call _this_ coming release `v1.4.6`,
though we will then need another discriminant to differentiate from the kernel release.
   + Note that the next action is to update source code and documentation to `v1.4.7` if this is the preferred choice
(likely a follow-up commit in this PR).

- My bullet-point list (`v1.4.7`) is not as dense as the `v1.4.6` one.
I only care to mention changes which are visible to users (perf, cli, api, package, etc.)
while invisible stuff (improved CI and tests, internal source reorg, etc) doesn't make it to `CHANGELOG`.
We could keep it this way, or decide to unify policy for `v1.4.6` and `v1.4.7`, or even fuse these 2 paragraphs if that feels preferable.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,24 @@
+v1.4.7
+perf: stronger --long mode at high compression levels, by @senhuang42
+perf: stronger --patch-from at high compression levels, thanks to --long improvements
+perf: faster dictionary compression at medium compression levels, by @felixhandte
+perf: small speed & memory usage improvements for ZSTD_compress2(), by @terrelln
+perf: improved fast compression speeds with Visual Studio, by @animalize
+cli : Set nb of threads with environment variable ZSTD_NBTHREADS, by @senhuang42
+cli : accept decompressing files with *.zstd suffix
+cli : provide a condensed summary by default when processing multiple files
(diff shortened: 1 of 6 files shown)


animalize:
It's worthy to mention issue https://github.com/facebook/zstd/issues/2327: fix multithreaded output can depend on number of threads, by @terrelln

animalize:
> api : fix : multithreaded compression has predictable output even in special cases (see #2327) (issue not accessible from cli)

#2327 is the issue's number, it seems all numbers in CHANGLOG are PR numbers.
The PR number of this problem is https://github.com/facebook/zstd/pull/2339.

senhuang42:
Might be worth mentioning the Makefile build changes? Incremental builds and config detection are pretty nice user-facing improvements.

Issue #2428, opened 2020-12-17
1.4.7 dictionary tests fail on i586 and s390x


bmwiedemann:
**Describe the bug**
dictionary tests fail on i586 and s390x

**To Reproduce**
1. on openSUSE or Debian do
2. osc checkout Archiving/zstd && cd $_
2. osc build --vm-type=kvm -j1 --clean openSUSE_Factory i586

**Expected behavior**
tests should pass the same way they did in 1.4.5

**Screenshots and charts**
build log:
``` ===>  dictionary tests
 - test with raw dict (content only)



 - Create first dictionary
 !  Warning : data size of samples too small for target dictionary size
 !  Samples should be about 100x larger than target dictionary size
 Loading /home/abuild/rpmbuild/BUILD/zstd-1.4.7/tests/../programs/zstdcli.c...       ^M^M                                                                               ^MTrying 5 different sets of parameters
 WARNING: The maximum dictionary size 112640 is too large compared to the source size 688542! size(source)/size(dictionary) = 6.112766, but it should be >= 10! This may lead to a subpar dictionary! We recommend training on sources at least 10x, and preferably 100x the size of the dictionary!
 HUF_writeCTable error
 Failed to select dictionary
 HUF_writeCTable error
 Failed to select dictionary
 HUF_writeCTable error
 Failed to select dictionary
 HUF_writeCTable error
 Failed to select dictionary
 HUF_writeCTable error
 Failed to select dictionary

^Mdictionary training failed : Error (generic)
 make: *** [Makefile:330: test-zstd] Error 1
```

**Desktop:**
 - OS: openSUSE-Tumbleweed
 - Version 20201215
 - Compiler gcc10
 - Flags `-O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -ffat-lto-objects`
 - Build system: Makefile

**Additional context**

[How openSUSE builds](https://github.com/bmwiedemann/openSUSE/blob/master/packages/z/zstd/zstd.spec#L93)

Current build logs can be found at
https://build.opensuse.org/package/live_build_log/Archiving/zstd/openSUSE_Factory/i586

ismail:
On openSUSE we compile both the library and the tests with some flags and turns out, if you do

```
export CFLAGS="-O2 -fstack-protector"
export CXXFLAGS="-O2 -fstack-protector"
[...]

PR #2430, opened 2020-12-17
Fix alignment of scratchBuffer in HUF_compressWeights()

terrelln:
The scratch buffer must be 4-byte aligned. This causes test failures in 32-bit systems, where the stack isn't aligned.

Fixes Issue #2428.


terrelln:
`make check` in 32-bit mode is still failing, but this new failure seems like a test problem, where we're trying to run a test that allocates too much memory.

```
===>  zstd long distance matching with optimal parser compressed size tests
zstd: error 11 : Allocation error : not enough memory
```

Cyan4973:
Great fix @terrelln !

Should we also add a test, able to trigger that issue (when unfixed) ?

Cyan4973:
> make check in 32-bit mode is still failing, but this new failure seems like a test problem, where we're trying to run a test that allocates too much memory.

Indeed,
maybe we should find a way to reduce memory requirement for this test.

terrelln:
Yes, we should run `make check` in 32-bit mode without any sanitizers. I don't know any other way to trigger the bug. But, it is failing later on, so I can't add it yet.

I'm going to be away for an hour or two now, so I can't look into it immediately. When I get back I will look into it, unless someone beats me to the punch.

Cyan4973:
I'm guessing `zstd -22 --ultra` is using way too much resources, especially when combined with `stdin` streaming

lib/compress/huf_compress.c
```diff
@@ -69,7 +69,7 @@ static size_t HUF_compressWeights (void* dst, size_t dstSize, const void* weight
     U32 tableLog = MAX_FSE_TABLELOG_FOR_HUFF_HEADER;
 
     FSE_CTable CTable[FSE_CTABLE_SIZE_U32(MAX_FSE_TABLELOG_FOR_HUFF_HEADER, HUF_TABLELOG_MAX)];
-    BYTE scratchBuffer[FSE_BUILD_CTABLE_WORKSPACE_SIZE(HUF_TABLELOG_MAX, MAX_FSE_TABLELOG_FOR_HUFF_HEADER)];
+    U32 scratchBuffer[FSE_BUILD_CTABLE_WORKSPACE_SIZE_U32(HUF_TABLELOG_MAX, MAX_FSE_TABLELOG_FOR_HUFF_HEADER)];
```
Cyan4973, line 72: minor question:

I'm struggling to understand how this could be an issue _only_ for 32-bit systems ...

terrelln: I believe that either compilers or the x86-64 ABI always align all stack entries to 8 or 16 bytes. But 32-bit systems wont.

Cyan4973:
Redirecting the PR into `check32`, where a 32-bit runtime test is added, and currently fails (as expected).
The fix in this PR should make the new test pass.

lib/compress/huf_compress.c
Cyan4973: That's a good guess.
It's in line with observations that the bug disappears on enabling sanitizers.

Cyan4973 (approved):

lib/compress/huf_compress.c
bmwiedemann: s390x also failed and is a 64bit arch. So probably more related to the alignment standards applied by compilers for the arch.

PR #2438, opened 2020-12-21
try to keep libzstd.a "as is" once created

### LINKED ISSUE #2436: Parallel build of v1.4.8 can fail
**Describe the bug**

Could be race condition:
image

**To Reproduce**

`make almost -j36`

 - OS: CentOS 7 / Ubuntu 18.04
 - Compiler: gcc-9/gcc-8
 - Other relevant hardware specs: 36-thread Xeon
 - Build system: Makefile

Cyan4973:
to improve compatibility with parallel build scenarios such as `make -j allmost`.
Presumed to fix #2436

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,9 +31,9 @@ jobs:
           command: |
             make gnu90build; make clean
             make gnu99build; make clean
-            make ppc64build; make clean
-            make ppcbuild  ; make clean
-            make armbuild  ; make clean
+            make ppc64build V=1; make clean
+            make ppcbuild   V=1; make clean
+            make armbuild   V=1; make clean
             make -C tests test-legacy test-longmatch; make clean
             make -C lib libzstd-nomt; make clean
   # This step is only run on release tags.
--- a/.github/workflows/generic-dev.yml
+++ b/.github/workflows/generic-dev.yml
@@ -71,7 +71,7 @@ jobs:
         make gcc8install
         CC=gcc-8 CFLAGS="-Werror" make -j all
         make clean
-        CC=gcc-8 make -j uasan-test-zstd </dev/null
+        CC=gcc-8 make -j uasan-test-zstd </dev/null V=1
 
   gcc-6-asan-ubsan-testzstd-32bit:
     runs-on: ubuntu-latest
--- a/.github/workflows/generic-dev.yml
+++ b/.github/workflows/generic-dev.yml
@@ -84,17 +84,15 @@ jobs:
         make clean
         CC=gcc-6 make -j uasan-test-zstd32
 
-  clang-38-msan-testzstd:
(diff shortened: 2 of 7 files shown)


terrelln (approved):

terrelln (changes requested):
Tests are failing

Cyan4973:
All tests fixed.
Remaining ones are failing independently of this PR.
32-bit test is fixed in #2440.
Fuzzing + msan has been failing for a while now,
I may disable it on short term so that we stop considering "normal" to have red failures in CI.

Cyan4973:
Actually, disabling fuzzing + msan seems more difficult than anticipated,
as it seems to be a [runtime issue (time out)](https://github.com/facebook/zstd/pull/2438/checks?check_run_id=1592001475#step:5:11),
but the script controllling _all_ fuzzing jobs seems controlled from `oss-fuzz` : 
https://github.com/facebook/zstd/blob/dev/.github/workflows/main.yml#L18 ,
so there is no obvious way to fix it, nor to opt out selectively from msan tests (but keep the rest running)
and I'm reluctant about removing all `oss-fuzz` tests on the ground that the `msan` variant is flackey.

terrelln:
I'll open an issue on oss-fuzz about the msan failure.

terrelln:
It looks like `check-32bit` test is failing, but that seems unrelated to this PR I think.

terrelln (approved):

terrelln:
https://github.com/google/oss-fuzz/issues/4879

evverx:
> the script controllling all fuzzing jobs seems controlled from oss-fuzz :
https://github.com/facebook/zstd/blob/dev/.github/workflows/main.yml#L18 ,
so there is no obvious way to fix it, nor to opt out selectively from msan tests (but keep the rest running)
and I'm reluctant about removing all oss-fuzz tests on the ground that the msan variant is flackey.

I think it should be possible to remove "memory" from [the "sanitizer" list](https://github.com/facebook/zstd/blob/d4f7d75f84427c7d06903d9dbecee7033f3ad47a/.github/workflows/main.yml#L9) to turn off only Msan there:
```diff
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index 0e14345..5e5aae1 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sanitizer: [address, undefined, memory]
+        sanitizer: [address, undefined]
     steps:
     - name: Build Fuzzers (${{ matrix.sanitizer }})
       id: build
```
Though I hope the issue will be fixed soon.

Cyan4973:
That's a good point @evverx ,
we should probably do that while waiting for the issue to get fixed on `ossfuzz` side.

Issue #2275, opened 2020-08-18
Version Information for debugging


nand28:
Hi,

ZSTD  guarantees successful decompression of data compressed with older versions of zstd. But is the same guaranteed vice versa? That is if we need to compress with newer version of zstd (at one node) and decompress with older version of zstd (at another node). If that can break at any point in future, we would like to identify such cases by knowing the version with which the data is compressed. The version of zstd which is currently used for compression can be obtained via ZSTD_version() api but similarly, can we obtain the zstd version of the compressed data? Like an optional checksum appended to the compressed data, can we have an optional version number appended to the compressed data and the same is verified for compatibility during decompression?

Also, is optional checksum allowed for single step compression using context?

Thanks.

felixhandte:
Hi @nand28,

Yes, it is intended that both old-to-new and new-to-old roundtrips should always work (as long as both endpoints are >=[v0.8](https://github.com/facebook/zstd/releases/tag/v0.8.0)). The compression format is stable.

The only exception to that is that we sometimes find bugs in the decompressor and fix them (you can [search closed PRs with the `bug` tag](https://github.com/facebook/zstd/pulls?page=2&q=is%3Apr+label%3Abug), if you want to see examples).

> can we obtain the zstd version of the compressed data?

The first four bytes of the compressed frame are, in effect, a version identifier. `"\x28\xB5\x2F\xFD"` indicates that the blob uses the stable (v0.8+) format.

> Also, is optional checksum allowed for single step compression using context?

Yes, you can enable checksumming on any frame, though you may have to use a different API entry point than you are used to in order to do so. For example, `ZSTD_compressCCtx` ignores any configuration in the cctx; it only uses the cctx for efficiency. So you'd want to use something like `ZSTD_compress2()` instead:

```
ZSTD_CCtx_reset(cctx, ZSTD_reset_session_and_parameters); /* clear the cctx */
ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel, 3); /* set compression level, etc. */
ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 1); /* enable checksumming */
size_t dstSize = ZSTD_compress2(cctx, dst, dstCapacity, src, srcSize);
assert(!ZSTD_isError(dstSize));
```

See [this comment](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L236-L247) for more.

nand28:
Thanks Felix for the quick response.

[...]

PR #2439, opened 2020-12-21
Generate skippable frame API

senhuang42:
`ZSTD_writeSkippableFrame(void* dst, size_t dstCapacity, const void* src, size_t srcSize, U32 magicVariant)`

Writes a skippable frame into `dst` buffer, using `src` content, and using one of the 16 available skippable frame magic numbers, controllable with `magicVariant`. More detailed documentation in `zstd.h`.

Test Plan:
- Modify existing unit test to ensure roundtrip-ability

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2954,6 +2954,26 @@ static size_t ZSTD_writeFrameHeader(void* dst, size_t dstCapacity,
     return pos;
 }
 
+/* ZSTD_generateSkippableFrame_advanced() :
+ * Writes out a skippable frame with the specified magic number variant (16 are supported), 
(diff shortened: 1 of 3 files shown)


terrelln:
This looks good to me. Is there a particular use case that motivates adding this to the API?

senhuang42:
Was mentioned in #2275 that this might be a nice thing to have.

Cyan4973:
> `U32 magicVariant`

In the future, if we ever come to settle on a few skippable frame IDs implying something meaningful, we could use an `enum` instead.  Or in complementarity, as I think this is compatible with `U32`, so it should be fine.

Speaking of "meaningful" skippable frame IDs, maybe we could consider those from `pzstd`, and maybe the `seekable` format.

Cyan4973 (approved):

PR #2440, opened 2020-12-21
try to fix 32-bit test on github actions

Cyan4973:
for some reasons, this test fails at _installing_ 32-bit dependencies
using the exact same command that actually works in other tests !!?

It's unclear why it fails repeateadly for this test only.
Try another way to install dependencies to fix that.


Cyan4973:
Well, it seems to finally fix this annoying installation issue...
https://github.com/facebook/zstd/pull/2440/checks?check_run_id=1592020554

PR #2441, opened 2020-12-22
Fix visibility of symbols in .so

yumeyao:
There was a change of the makefile that converted compiling static/shared lib _once with multiple source files_ to _compiling each to object file and linking_. But the `-fvisibility=hidden` flag is not moved onto compiling stage by mistake. So this fixes the default visibility by moving it there.

~~Furthermore, I discovered there is an API renaming. so I also added an alias for this API for the sake of backward-compatibility.~~ 
~~Only makefile under *nix/mingw is fixed. cmake/Visual Studio project files in `build/` directory may also want this alias in the .so/.dll symbol table, but the change is not contained in this PR.~~ 
~~On MSVC this can be done via linker flag: `"/EXPORT:ZSTD_getSequences=ZSTD_generateSequences"`(mangling with prefix `_` is required on x86_32)~~

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -252,14 +252,14 @@ ifneq (,$(filter Windows%,$(TARGET_SYSTEM)))
 LIBZSTD = dll/libzstd.dll
 $(LIBZSTD): $(ZSTD_FILES)
 	@echo compiling dynamic library $(LIBVER)
-	$(CC) $(FLAGS) -DZSTD_DLL_EXPORT=1 -Wl,--out-implib,dll/libzstd.dll.a -shared $^ -o $@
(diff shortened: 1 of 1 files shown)


Cyan4973 (changes requested):
Thanks @yumeyao,

the symbol visibility issue is real, and deserves to be fixed,

but the name update of an experimental symbol is a different thing, and I wouldn't want to mix the 2 topics.
Prefer separating them.

yumeyao:
pushed a commit to remove symbol aliasing.

Also updated pr title/descript to reflect changes.

Cyan4973 (approved):

Cyan4973:
Thanks @yumeyao !

PR #2446, opened 2020-12-29
[RFC] Support references to multiple DDicts

senhuang42:
This PR adds a new experimental param `ZSTD_d_refMultipleDDicts` which controls the behavior of `ZSTD_refDDict()`.

If enabled, then multiple calls to `ZSTD_refDDict()` will persist the multiple referenced `ZSTD_DDict*` in memory rather than overwriting the last reference. Under the hood, we use a hash set w/ linear probing to keep track of the `ZSTD_DDict*`. Example usage is in the included unit test in `fuzzer.c`, but is basically as simple as:

```
ZSTD_DCtx* dctx = ZSTD_createDCtx();
ZSTD_DCtx_setParameter(dctx, ZSTD_d_refMultipleDDicts, ZSTD_rmd_refMultipleDDicts);
ZSTD_DCtx_refDDict(dctx, ddict1);
ZSTD_DCtx_refDDict(dctx, ddict2);
ZSTD_DCtx_refDDict(dctx, ddict3);
...
ZSTD_decompress...
```

The goal is no changes to the existing API, leaving existing behavior/usages of `ZSTD_refDDict()` unchanged.
Only when we call `setParameter()` with the new experimental param can behavior change.

There are still issues with this PR, but I wanted to put this up to get some initial comments to see if this was roughly the direction we wanted to head in.

Open questions/followups:
- Hash table allocation entry point (is `refDDict()` best?)
- Performance (hash table param tuning/hash function)
- API Extensions? (un-reference a dictionary)
- Add streaming unit test
- Still need to test more extensively
- Some code cleanup

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -648,6 +648,16 @@ static size_t ZSTD_decompressFrame(ZSTD_DCtx* dctx,
         ip += frameHeaderSize; remainingSrcSize -= frameHeaderSize;
     }
 
+    /* Reference DDict requested by frame if dctx references multiple ddicts */
+    if (dctx->refMultipleDDicts && dctx->ddictSet) {
+        ZSTD_DDict* frameDDict = ZSTD_DDictHashSet_getDDict(dctx->ddictSet, dctx->fParams.dictID);
+        if (frameDDict) {
+            ZSTD_clearDict(dctx);
+            dctx->ddict = frameDDict;
+            dctx->dictUses = ZSTD_use_indefinitely;
+        }
+    }
+
     /* Loop on each block */
     while (1) {
         size_t decodedSize;
@@ -1495,6 +1505,10 @@ size_t ZSTD_DCtx_refDDict(ZSTD_DCtx* dctx, const ZSTD_DDict* ddict)
     if (ddict) {
         dctx->ddict = ddict;
         dctx->dictUses = ZSTD_use_indefinitely;
+        if (dctx->refMultipleDDicts && dctx->ddictSet) {
+            assert(!dctx->staticSize);  /* Impossible: ddictSet cannot have been allocated if static dctx */
+            ZSTD_DDictHashSet_addDDict(dctx->ddictSet, ddict, ZSTD_getDictID_fromDDict(ddict), dctx->customMem);
+        }
     }
     return 0;
 }
@@ -1611,14 +1625,12 @@ size_t ZSTD_DCtx_setParameter(ZSTD_DCtx* dctx, ZSTD_dParameter dParam, int value
             return 0;
         case ZSTD_d_refMultipleDDicts:
             CHECK_DBOUNDS(ZSTD_d_refMultipleDDicts, value);
+            if (dctx->staticSize != 0) {
+                RETURN_ERROR(parameter_unsupported, "Static dctx does not support multiple DDicts!");
+            }
             dctx->refMultipleDDicts = (ZSTD_refMultipleDDicts_e)value;
-            if (dctx->refMultipleDDicts == ZSTD_d_refMultipleDicts) {
-                if (dctx->ddictSet == NULL) {
-                    if (dctx->staticSize != 0) {
-                        RETURN_ERROR(memory_allocation, "Multiple DDicts are not allowed with static dctx!");
-                    }
-                    dctx->ddictSet = ZSTD_createDDictHashSet(dctx->customMem);
-                }
+            if (dctx->refMultipleDDicts == ZSTD_d_refMultipleDicts && dctx->ddictSet == NULL) {
+                dctx->ddictSet = ZSTD_createDDictHashSet(dctx->customMem);
             } else {
                 ZSTD_freeDDictHashSet(dctx->ddictSet, dctx->customMem);
                 dctx->ddictSet = NULL;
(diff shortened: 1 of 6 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -72,6 +72,140 @@
...
+
+#define DDICT_HASHSET_MAX_LOAD_FACTOR 0.75
+#define DDICT_HASHSET_TABLE_BASE_SIZE 64
+#define DDICT_HASHSET_RESIZE_FACTOR 2
+
+/* Hash function to determine starting position of dict insertion within the table
+ * Returns an index between [0, hashSet->ddictPtrTableSize]
+ */
+static size_t ZSTD_DDictHashSet_getIndex(const ZSTD_DDictHashSet* hashSet, U32 dictID) {
+    U32 hash = XXH64(&dictID, sizeof(U32), 0);
```
senhuang42: One thing I can't quite figure out is why on certain builds/CI tests, `zstd_decompress.c` can't see `XXH64`, but can see `XXH64_digest(), etc.`? As far as I can tell, `xxhash.h` is within the graph of dependencies, and locally, it seems fine. Is there an additional macro I need to invoke?

Cyan4973: Is this related to the linux kernel test ?

senhuang42: Yeah, I do see a line about `Replacing XXH64 prefix with xxh64` in the logs, and I assume it might be related.

lib/decompress/zstd_decompress.c
```diff
@@ -72,6 +72,140 @@
 #endif
```
Cyan4973, line 73: I don't see an `#include "../common/xxhash.h"` anywhere

lib/decompress/zstd_decompress.c
```diff
@@ -72,6 +72,140 @@
...
+ * Returns 0 if successful (ddict may or may not have actually been added), or a zstd error code if something went wrong
+ */
+static size_t ZSTD_DDictHashSet_emplaceDDict(ZSTD_DDictHashSet* hashSet, const ZSTD_DDict* ddict, U32 dictID) {
+    size_t idx = ZSTD_DDictHashSet_getIndex(hashSet, dictID);
+    RETURN_ERROR_IF(hashSet->ddictPtrCount == hashSet->ddictPtrTableSize, GENERIC, "Hash set is full!");
+    DEBUGLOG(4, "Hashed index: for dictID: %u is %zu", dictID, idx);
+    while (hashSet->ddictPtrTable[idx] != NULL) {
+        /* Replace existing ddict if inserting ddict with same dictID */
+        if (ZSTD_getDictID_fromDDict(hashSet->ddictPtrTable[idx]) == dictID) {
+            DEBUGLOG(2, "DictID already exists, replacing rather than adding");
```
Cyan4973: minor : reminder : 
`DEBUGLOG(2,` is fine for debugging / developing purposes,
but the level will have to be increased before merge (likely `4` in this case).

lib/decompress/zstd_decompress.c
senhuang42: Hmm, `zstd_decompress.c` has `#include "../common/zstd_internal.h"`, which has `#include "xxhash.h"` - I assumed that that would be sufficient to include all the public symbols from `xxhash.h`?

lib/decompress/zstd_decompress.c
```diff
@@ -72,6 +72,140 @@
...
+ * Returns an index between [0, hashSet->ddictPtrTableSize]
+ */
+static size_t ZSTD_DDictHashSet_getIndex(const ZSTD_DDictHashSet* hashSet, U32 dictID) {
+    U32 hash = XXH64(&dictID, sizeof(U32), 0);
+    /* DDict ptr table size is a multiple of 2, use size - 1 as mask to get index within [0, hashSet->ddictPtrTableSize) */
+    return hash & (hashSet->ddictPtrTableSize - 1);
+}
+
+/* Adds ddict to a hashset resizing it 
+ * Returns 0 if successful (ddict may or may not have actually been added), or a zstd error code if something went wrong
```
Cyan4973: > 0 if successful (ddict may or may not have actually been added)

This feels weird on reading it.
Can the function both return a success code, and _not_ add the dictionary ?
Please describe this case if it exists.

Below, I see that a dictionary may replace an existing one if they both have same `dictID`. Is it related ?

lib/decompress/zstd_decompress.c
```diff
@@ -72,6 +72,140 @@
...
+ */
+static size_t ZSTD_DDictHashSet_emplaceDDict(ZSTD_DDictHashSet* hashSet, const ZSTD_DDict* ddict, U32 dictID) {
```
Cyan4973: interface :
is `dictID` already present within `ddict` ?
It's likely, since we have `ZSTD_getDictID_fromDDict()`.
In which case, it seems better to rely on it rather than adding a parameter.

lib/decompress/zstd_decompress.c
```diff
@@ -72,6 +72,140 @@
...
+    DEBUGLOG(4, "Hashed index: for dictID: %u is %zu", dictID, idx);
+    while (hashSet->ddictPtrTable[idx] != NULL) {
+        /* Replace existing ddict if inserting ddict with same dictID */
+        if (ZSTD_getDictID_fromDDict(hashSet->ddictPtrTable[idx]) == dictID) {
+            DEBUGLOG(2, "DictID already exists, replacing rather than adding");
+            hashSet->ddictPtrTable[idx] = ddict;
+            return 0;
+        }
+        if (idx == hashSet->ddictPtrTableSize - 1) {
+            idx = 0;
```
Cyan4973: minor (potential suggestion) : could also use a mask _if_ `ddictPtrTableSize` is a clean power of 2.

lib/decompress/zstd_decompress.c
terrelln: You should just need to change this line:

https://github.com/facebook/zstd/blob/bc0a1e4d59bc9541cb4740f08ae17d97a3fcc4b5/contrib/freestanding_lib/freestanding.py#L579

Something like: `re.compile(r"([^\w]|^)(?P<orig>XXH64)[\(_]")`

Cyan4973 (approved):

Issue #2261, opened 2020-08-05
make / cmake inconsistencies


patrickdepinguin:
**Describe the bug**
The compiler and link commands used depend on the build system used, e.g. make vs cmake (maybe same for the other build systems). This is unexpected and undesired.

For example, for libzstd, the makefile specifies -fvisibility=hidden, whereas the cmake version does not.
Similarly, the compilation of the 'programs' have many compiler warning flags which are absent for cmake.
On the other hand, the cmake system now allows to link the 'zstd' program dynamically to libzstd via -DZSTD_PROGRAMS_LINK_SHARED=ON, but this feature is not supported in the 'make' system. There does exist a 'zstd-dll' target, which does not work because the visibility of symbols in libzstd is set to hidden.

**To Reproduce**
- compare output from a clean directory for following steps:
    - cmake build/cmake/ -j1 V=1 VERBOSE=1
    - make -j1 V=1 VERBOSE=1

**Expected behavior**
I expect the same output files regardless of the build system

**Additional context**
With the upgrade from zstd 1.4.3 to 1.4.5, I notice yet another big increase in size, +500KB on libzstd. I thus wanted to benefit from ZSTD_PROGRAMS_LINK_SHARED=ON, but zstd as packaged in Buildroot is using 'make', hence I stumbled upon these inconsistencies.

patrickdepinguin:
I'm currently using this patch to support ZSTD_PROGRAMS_LINK_SHARED in the make context.

From afc368e7247abfe89250def5b7673a9ccb79e0b1 Mon Sep 17 00:00:00 2001
From: Thomas De Schampheleire <thomas.de_schampheleire@nokia.com>
Date: Wed, 5 Aug 2020 15:35:01 +0200
Subject: [PATCH] make: support ZSTD_PROGRAMS_LINK_SHARED to link zstd with
 libzstd

zstd allows building via make, cmake, meson, and others, but unfortunately
the features and flags used for these build systems are inconsistent.

The cmake version respects an option 'ZSTD_PROGRAMS_LINK_SHARED' to let the
zstd binary link dynamically with its own libzstd. Without it, the zstd
program uses static linking and thus has a big size.

This option is not provided in the 'make' system. There does exist a target
'zstd-dll' which intends to do exactly this, but it does not work out of the
box due to linking issues. These in turn are caused because the lib makefile
passes '-fvisibility=hidden' to the linker, which hides certain symbols that
the zstd program needs. The cmake-based compilation does not pass this
[...]

Issue #1680, opened 2019-07-11
Dynamic linking libzstd?


zvezdochiot:
Dynamic linking libzstd?:
```
ldd zstd | grep zstd
(none)
```
PS:
```
./zstd -V
*** zstd command line interface 32-bits v1.4.0, by Yann Collet ***
```

Cyan4973:
`zstd`, as the binary CLI program, doesn't link against the shared library `libzstd`.
It includes directly the static library `libzstd.a`.

One of the main reasons is that `zstd` enables multi-threading by default, while `libzstd` does not.

zvezdochiot:
@Cyan4973 say> zstd enables multi-threading by default, while libzstd does not.

Ok. It's a pity.

PR #2450, opened 2021-01-03
make: Add ZSTD_PROGRAMS_LINK_SHARED to dll-link zstd

cemeyer:
Add a make variable ZSTD_PROGRAMS_LINK_SHARED.  When set, libzstd
exports private symbols needed to shared-link the zstd program.  zstd is
linked against this libzstd.so.

(See issues #2261, #1680.)

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -1198,6 +1198,9 @@ size_t ZSTD_referenceExternalSequences(ZSTD_CCtx* cctx, rawSeq* seq, size_t nbSe
 
 /** ZSTD_cycleLog() :
  *  condition for correct operation : hashLog > 1 */
+#if defined(ZSTD_PROGRAMS_LINK_SHARED)
+ZSTDLIB_API
+#endif
 U32 ZSTD_cycleLog(U32 hashLog, ZSTD_strategy strat);
 
 #endif /* ZSTD_COMPRESS_H */
--- a/lib/dictBuilder/zdict.c
+++ b/lib/dictBuilder/zdict.c
@@ -968,6 +968,9 @@ static size_t ZDICT_addEntropyTablesFromBuffer_advanced(
 }
 
 /* Hidden declaration for dbio.c */
+#if defined(ZSTD_PROGRAMS_LINK_SHARED)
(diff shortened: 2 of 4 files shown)


lib/compress/zstd_compress_internal.h
```diff
@@ -1198,6 +1198,9 @@ size_t ZSTD_referenceExternalSequences(ZSTD_CCtx* cctx, rawSeq* seq, size_t nbSe
 
 /** ZSTD_cycleLog() :
  *  condition for correct operation : hashLog > 1 */
+#if defined(ZSTD_PROGRAMS_LINK_SHARED)
```
cemeyer, line 1201: This could instead be some conditionally-defined `ZSTDLIB_CLI_DLL_API` (or whatever) in lib/zstd.h.  I don't feel strongly about it; the number of instances needed will probably remain very small.

Cyan4973: We could probably remove `ZSTD_cycleLog()` dependency. It's a trivial code to copy.

lib/dictBuilder/zdict.c
```diff
@@ -968,6 +968,9 @@ static size_t ZDICT_addEntropyTablesFromBuffer_advanced(
 }
 
 /* Hidden declaration for dbio.c */
+#if defined(ZSTD_PROGRAMS_LINK_SHARED)
+ZSTDLIB_API
```
Cyan4973, line 972: Here, we could make `ZDICT_trainFromBuffer_unsafe_legacy()` function part of the "unstable"/"experimental" section of `zdict` API.

Another alternative is to not use it in `dibio.c` and use the safe variant instead.
It comes at an additional memory allocation cost, but might be worth it to remove this dependency.

Cyan4973:
In #2457, there is a proposed PR
which makes possible a `zstd` CLI linked against `libzstd` dynamic library.

The proposed method (`make zstd-dll`) is a bit different from the one suggested here,
that being said it's not settled in the marble, and could be updated for a "better" one if there is such a proposal.

cemeyer:
Both proposals work well for me.

Cyan4973:
Superseded by #257

PR #2456, opened 2021-01-07
removed CLI dependency to legacy unsafe function

Cyan4973:
making the CLI one step closer to being linkable to the `libzstd` dynamic library


Cyan4973:
superseded by #2457

PR #2462, opened 2021-01-09
cmake: use configure_file() for creating the .pc file

lazka:
Escaping in add_custom_target() seems to depend on the shell used in the cmake
generator and using Ninja on Windows, which uses cmd.exe, results in stray backslashes
in the .pc file.

Instead of going through escaping hell just use configure_file() with the existing
libzstd.pc.in file already used by the simple Makefile based build system.

This fixes the .pc file syntax when building zstd with CMake+Ninja+gcc on Windows.

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -137,7 +137,7 @@ endif ()
 if (UNIX OR MINGW)
     # pkg-config
     set(PREFIX "${CMAKE_INSTALL_PREFIX}")
-    set(EXEC_PREFIX "\\$$\{prefix}")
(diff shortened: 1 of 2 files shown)


lazka:
(re-using the libzstd.pc.in might lead to breakage if it's changed in the future without realizing that it's used in two places, so I can also copy it to the cmake dir if wanted)

terrelln:
Awesome! I don't know why the `add_custom_target()` indirection was used in the first place, but this is a great simplification! 

This was causing me huge problems when I was trying to fix the escaping. I've added a test on Linux, which is passing on this PR, so looks like it is correctly escaped for Linux too.

lazka:
Thanks!

Issue #2463, opened 2021-01-10
"stdin is a console, aborting" when trying to read from stdin


KaMyKaSii:
**Describe the bug**
I am backing up my Android device (creating a tarball) and using zstd to compress it. I am using the commands adb exec-out (to send the output of the command on the Android device to the computer) and adb exec-in (to send the input on the computer to the commamd on the Android device). The problem is that zstd (in the device) can read from stdin during the backup/compress step but refuses to read during the restore/decompress. The same works correctly when using just tar or tar + pigz to compress and decompress.

**To Reproduce**
1. Backup:
```
./adb exec-out "tar -cp /system/bin 2>/dev/null | pigz -1 -f" > system-bin.tar.gz (works)

./adb exec-out "tar -cp /system/bin 2>/dev/null | /data/adb/bin/zstd148 -3 -c" > system-bin.tar.zst (works)
```
2. Restore:

```
./adb exec-in "pigz -d -f | tar -xp -C /cache 2>/dev/null" < system-bin.tar.gz (works)

./adb exec-in "/data/adb/bin/zstd148 -d -c 2>/cache/zstdlog.txt | tar -xp -C /cache 2>/dev/null" < system-bin.tar.zst (do not work)
```
The command ends instantly. Checking the zstdlog.txt file:
`stdin is a console, aborting`

**Expected behavior**
Just like tar or tar+pigz, read from stdin correctly.

**Desktop (please complete the following information):**
 - OS: Android
 - Version: 9
 - Compiler: Clang from Android NDK r22
 - Build system "make"

felixhandte:
Ah. Interesting. I think due to the way processes are spawned on android devices from the connected debugging machine, `isatty(stdin)` is (semi-erroneously) returning true. Normally, when you redirect input into a command, `isatty(stdin)` should return false. But here I guess `stdin` is still a console, and that console is just being populated from a file at the other end.

So if you copy the archive onto the device, unpacking should work fine.

Alternatively, given that it sounds like you were already building `zstd` yourself, this problem should be really easy to work around. The offending test is here: https://github.com/facebook/zstd/blob/master/programs/zstdcli.c#L1246-L1249. You can just comment that out and rebuild and I imagine it will work.

I'm not sure if this corner case is common enough to warrant handling. Maybe `-f`/`--force` should also override this check?

terrelln:
> I'm not sure if this corner case is common enough to warrant handling. Maybe -f/--force should also override this check?

[...]

PR #2466, opened 2021-01-11
Allow Input From Console When `-f`/`--force` is Passed

felixhandte:
Addresses #2463.


Cyan4973 (approved):

Cyan4973:
minor : 
it would be great if we could also have a test validating this new ability in `playTests.sh`.
But I'm unsure if that's possible ...

terrelln (approved):

Issue #2469, opened 2021-01-13
Decompressing file doesn't create a directory


jin8495:
**Describe the bug**
First, I compressed a directory which has many files. And then I decompressed it. But the decompressed one is not a directory. It was a file. So, I cannot get in to the directory to search a specific file that I want.

**To Reproduce**
1. $ zstd -r --ultra DIRECTORY -o DIRECTORY.zst
2. $ zstd -d DIRECTORY.zst
3. decompressed one is not a directory!

**Expected behavior**
I just wanted to get my file. but '$ cd DIRECTORY' doesn't work.

**Desktop (please complete the following information):**
 - OS: ubuntu 18.04
 - Version: zstd CLI 64-bits v1.4.5

Cyan4973:
`-o FILE.zst` achieves exactly the same thing as `-c > FILE.zst`.

The second formulation is maybe clearer in what it does: it merely pushes all compressed frames into `FILE.zst`, directly appending them to each other, with no additional metadata (only the content). 
That means, it has no knowledge of what a directory is, and it doesn't even know what filenames are. This pack of concatenated frames cannot, on its own, regenerate the original directory.

Recent versions of `zstd` warn against such misunderstanding, by issuing a warning message when they detect this pattern, or even requiring the use of `-f` force mode to bypass protection.

```
zstd: WARNING: all input files will be processed and concatenated into a single output file: dir.zst
The concatenated output CANNOT regenerate the original directory tree.
```

In order to achieve your goal (saving and compacting a complete directory), it's necessary to use a proper archiver like `tar`. Then pipe the result to `zstd`. The reverse operation (`zstd -d | tar x`) will regenerate the directory fully (including all metadata, such as creation / modification time, read /write access, etc.). And it will also compress better.

See https://github.com/facebook/zstd/issues/1526 for additional details (this question is effectively a duplicate).

jin8495:
Thank you for the reply. I have already compressed my files which are stored in the directory. I think I have to throw away them ;(

terrelln:
@jin8495 you can use the tool I'm adding in PR #2473 to recover your files. You won't get any of the file names or metadata back, but it will split all the file content up correctly. Once it is merged you can run:
 
```
git clone https://github.com/facebook/zstd
cd zstd/contrib/recovery
make recover_directory
mkdir recovered/
./recover_directory DIRECTORY.zst recovered/
```

It will write each file to `recovered/N` in the order they were compressed.

PR #2473, opened 2021-01-15
[contrib][recovery] Add recovery_directory program

terrelln:
This program takes a file with concatenated zstd frames and splits the
file up by frame. E.g. if `dir.zst` has 4 frames:

```
> ./recover_directory dir.zst recovery/file
Recovering 4 files...
Recovered recovery/file0
Recovered recovery/file1
Recovered recovery/file2
Recovered recovery/file3
Complete
```

### DIFF
--- a/contrib/recovery/Makefile
+++ b/contrib/recovery/Makefile
@@ -1,0 +1,35 @@
+# ################################################################
+# Copyright (c) 2019-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under both the BSD-style license (found in the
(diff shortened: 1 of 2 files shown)


Cyan4973:
Yes ! That's thoughtful !
While it doesn't restore any directory structure, nor the filenames, at least each file content is regenerated properly, and for some simpler scenarios, it might prove good enough to retrieve a few files which matter, for situations likes #2469.

Cyan4973 (approved):

jin8495:
Thank you so much!! You saved my life XDXD

PR #2479, opened 2021-01-31
Create separate cmake exports for static/share targets

insilications:
Create separate cmake exports for static/share targets. Allow projects 
which import libzstd_static and libzstd_shared in different modules to 
escape the protection against multiple inclusion, which would fail when 
already imported targets are added once more (https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmExportFileGenerator.cxx#L974)


terrelln:
@insilications sorry this PR fell through the cracks!

We're not super familiar with cmake, so we're not exactly sure what the problem is, and how this PR fixes it. If you are still interested in getting this PR merged, could you please clarify the problem and what this PR is doing?

terrelln:
@insilications please reopen the PR if you would like to see it merged, and provide some clarifications about why you are making this change.

PR #2496, opened 2021-02-12
[trace] Minor fixes found during integration

terrelln:
* Mark `ZSTD_CCtx_getParameter()` as const
* Add `extern "C"` guards to `zstd_trace.h`


felixhandte (approved):
Kinda bothers me that you put the `const` in different places between the prototype and definition. But that's ok. :)

Cyan4973 (approved):

PR #2497, opened 2021-02-12
[lib] Set appliedParams.compressionLevel correctly

terrelln:
Forward the correct compressionLevel to the appliedParams in all cases.
It was already correct for the advanced API, so only the old single-pass
functions needed to be fixed.

This compression level is unused by the library, but is set so that the
tracing framework can consume it.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -270,6 +270,8 @@ size_t ZSTD_CCtxParams_init(ZSTD_CCtx_params* cctxParams, int compressionLevel) 
     return 0;
 }
 
+#define ZSTD_NO_CLEVEL 0
+
 size_t ZSTD_CCtxParams_init_advanced(ZSTD_CCtx_params* cctxParams, ZSTD_parameters params)
 {
     RETURN_ERROR_IF(!cctxParams, GENERIC, "NULL pointer!");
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress.c
```diff
@@ -3614,7 +3621,7 @@ size_t ZSTD_compress_usingDict(ZSTD_CCtx* cctx,
                                int compressionLevel)
 {
     ZSTD_parameters const params = ZSTD_getParams_internal(compressionLevel, srcSize, dict ? dictSize : 0, ZSTD_cpm_noAttachDict);
-    ZSTD_CCtx_params cctxParams = ZSTD_assignParamsToCCtxParams(&cctx->requestedParams, &params);
+    ZSTD_CCtx_params cctxParams = ZSTD_assignParamsToCCtxParams(&cctx->requestedParams, &params, (compressionLevel == 0) ? ZSTD_CLEVEL_DEFAULT: compressionLevel);
```
Cyan4973, line 3624: minor : `0` or `ZSTD_NO_CLEVEL` ?

Cyan4973 (approved):

lib/compress/zstd_compress.c
terrelln: I was thinking `0` because the semantic isn't that no compression level is set, but its a default. I am using `ZSTD_NO_CLEVEL` only when no compression level is set.

We could add a second macro for this meaning of `0`, but I'm not quite sure what I'd name it.

PR #2501, opened 2021-02-17
Explicitly hide static symbols

skitt:
Even with -fvisibility=hidden added to CFLAGS, any symbol which is
given a default visibility attribute ends up exported in the dynamic
library. This happens through zstd_internal.h which defines
..._STATIC_LINKING_ONLY before including various header files, and is
included for example in lib/common/pool.c.

To avoid this, this patch distinguishes static and non-static APIs;
only the latter are exported, the former are hidden.

In addition, -fvisibility=hidden is moved to CFLAGS for libzstd-mt (as
done previously for libzstd).

Signed-off-by: Stephen Kitt <steve@sk2.org>

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -71,9 +71,9 @@ DEBUGFLAGS= -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
 CFLAGS   += $(DEBUGFLAGS) $(MOREFLAGS)
 FLAGS     = $(CPPFLAGS) $(CFLAGS)
 
-CPPFLAGS_DYNLIB  = -DZSTD_MULTITHREAD # dynamic library build defaults to multi-threaded
-LDFLAGS_DYNLIB   = -pthread
-CPPFLAGS_STATLIB =                    # static library build defaults to single-threaded
+CPPFLAGS_DYNLIB  += -DZSTD_MULTITHREAD # dynamic library build defaults to multi-threaded
+LDFLAGS_DYNLIB   += -pthread
+CPPFLAGS_STATLIB +=                    # static library build defaults to single-threaded
 
 HAVE_COLORNEVER = $(shell echo a | grep --color=never a > /dev/null 2> /dev/null && echo 1 || echo 0)
 GREP_OPTIONS ?=
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1067,7 +1067,7 @@ size_t ZSTD_CCtx_setParametersUsingCCtxParams(
     return 0;
 }
 
-ZSTDLIB_API size_t ZSTD_CCtx_setPledgedSrcSize(ZSTD_CCtx* cctx, unsigned long long pledgedSrcSize)
+size_t ZSTD_CCtx_setPledgedSrcSize(ZSTD_CCtx* cctx, unsigned long long pledgedSrcSize)
 {
     DEBUGLOG(4, "ZSTD_CCtx_setPledgedSrcSize to %u bytes", (U32)pledgedSrcSize);
     RETURN_ERROR_IF(cctx->streamStage != zcss_init, stage_wrong,
@@ -1147,14 +1147,14 @@ size_t ZSTD_CCtx_loadDictionary_advanced(
     return 0;
 }
 
-ZSTDLIB_API size_t ZSTD_CCtx_loadDictionary_byReference(
+size_t ZSTD_CCtx_loadDictionary_byReference(
       ZSTD_CCtx* cctx, const void* dict, size_t dictSize)
 {
     return ZSTD_CCtx_loadDictionary_advanced(
             cctx, dict, dictSize, ZSTD_dlm_byRef, ZSTD_dct_auto);
 }
 
-ZSTDLIB_API size_t ZSTD_CCtx_loadDictionary(ZSTD_CCtx* cctx, const void* dict, size_t dictSize)
+size_t ZSTD_CCtx_loadDictionary(ZSTD_CCtx* cctx, const void* dict, size_t dictSize)
 {
     return ZSTD_CCtx_loadDictionary_advanced(
             cctx, dict, dictSize, ZSTD_dlm_byCopy, ZSTD_dct_auto);
@@ -4842,7 +4842,7 @@ ZSTD_CDict* ZSTD_createCDict_advanced(const void* dictBuffer, size_t dictSize,
         &cctxParams, customMem);
 }
 
-ZSTDLIB_API ZSTD_CDict* ZSTD_createCDict_advanced2(
+ZSTD_CDict* ZSTD_createCDict_advanced2(
         const void* dict, size_t dictSize,
         ZSTD_dictLoadMethod_e dictLoadMethod,
         ZSTD_dictContentType_e dictContentType,
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -20,19 +20,21 @@ extern "C" {
 
 
 /* =====   ZSTDLIB_API : control library symbols visibility   ===== */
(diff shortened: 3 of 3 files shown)


skitt:
... and this reveals (at least with CMake) that `pzstd` relies on dynamically accessing a number of static-only symbols (`ZSTD_getParams` etc.).

lib/Makefile
```diff
@@ -330,7 +330,8 @@ include $(wildcard $(DEPFILES))
 # Special case : building library in single-thread mode _and_ without zstdmt_compress.c
 ZSTDMT_FILES = compress/zstdmt_compress.c
 ZSTD_NOMT_FILES = $(filter-out $(ZSTDMT_FILES),$(ZSTD_FILES))
-libzstd-nomt: LDFLAGS += -shared -fPIC -fvisibility=hidden
+libzstd-nomt: CFLAGS += -fvisibility=hidden
```
Cyan4973: 👍

skitt: Thanks for taking a look ;-). I’m wondering what to do about `ZSTD_getParams` since `pztd` uses that dynamically even though it’s supposed to be a static-only symbol. Declaring it as a non-static API would fix the failing tests, but that might not be a stability promise the project is happy with!

(For context, I’m trying to reduce the amount of exported symbols to avoid [issues such as this one](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=969597) — Debian carries a symbols file, which is used to generate versioned dependencies, so that a binary can have dependencies which are as “generous” as possible; but this gets confused by `libzstd`’s symbols. Perhaps the correct solution is to drop that file entirely, and always request a version of `libzstd` equal than or greater to the version with which a given binary was compiled...)

Cyan4973:
While there are some good points in this PR,
and I appreciate the capability to hide static symbols on demand,
I don't think we are in position to hide such symbols _by default_.

Even our own applications can make intentional use of such symbols,
(the CLI is a great testing ground to flesh out a capability before it becomes part of the stable library API),
Perhaps more importantly, beyond apps present in this repository,
in the larger `zstd` ecosystem, we don't know how many other apps may depend on some of these symbols.

I don't rule out that we may one day hide static symbols in the dynamic library,
but this is not a light decision, we will have to investigate side effects, eventually find mitigations if necessary,
and it will have to be paired with a large version number upgrade.

Anyway, on short term, my recommendation is : 
- Keep the ability to hide static symbols _on demand_
- For the time being, keep static symbols _visible by default_
- Devise a set of specific tests that use hidden static symbols, and only work when static symbols are hidden. We will add that in CI and likely progressively expand from there.

lib/Makefile
Cyan4973: > Perhaps the correct solution is to (...) always request a version of `libzstd` equal than or greater to the version with which a given binary was compiled

`libzstd` library policy is to only update the "stable" section of the API when the second digit changes.
For example, in `1.4.x` series, all versions from `v1.4.0` to `v1.4.8` support the exact same set of symbols as "stable".
(Meanwhile, the "experimental" section can, and does, evolve at each iteration.)

The next update to "stable" API will happen at `v1.5.0`.
Even then, we are restrained to only _add_ new symbols, existing symbols are considered completely stable, and guaranteed to remain present in all `v1.x.y` versions of the library.

Given that, the proposed rule (always request a version of `libzstd` equal than or greater to the version with which a given binary was compiled) would work.
A slightly more generous rule would allow usage of older `libzstd` as long as the second digit matches or is greater (i.e. an application compiled with `v1.4.5` would still accept a `v1.4.0` library, but not `v1.3.6`). It's effectively the same rule, but limited to the first 2 digits.

skitt: Right, that’s what I suspected. The problem in Debian is that the default analysis results in, for example, the library package declaring that `ZSTD_minCLevel` is available since 1.3.8 (which is when it appeared in the library’s symbols) even though it’s only part of the stable API since 1.4. But that really means the problem lies in the tools used ;-).

skitt:
Should I amend this PR to only include the `Makefile` changes? As I understand it, those would be useful anyway, wouldn’t they?

lib/compress/zstd_compress.c
```diff
@@ -1039,7 +1039,7 @@ size_t ZSTD_CCtx_loadDictionary_advanced(
     return 0;
 }
 
-ZSTDLIB_API size_t ZSTD_CCtx_loadDictionary_byReference(
+ZSTDLIB_STATIC_API size_t ZSTD_CCtx_loadDictionary_byReference(
```
Cyan4973: These prefixes (`ZSTDLIB_API`, `ZSTDLIB_STATIC_API`) shouldn't be needed at this stage in `zstd_compress.c`.
They should remain confined within `zstd.h`.
So just remove them.

lib/zstd.h
```diff
@@ -20,19 +20,24 @@ extern "C" {
...
+#  define ZSTDLIB_API __declspec(dllexport) ZSTDLIB_VISIBLE
+#  define ZSTDLIB_STATIC_API
 #elif defined(ZSTD_DLL_IMPORT) && (ZSTD_DLL_IMPORT==1)
-#  define ZSTDLIB_API __declspec(dllimport) ZSTDLIB_VISIBILITY /* It isn't required but allows to generate better code, saving a function pointer load from the IAT and an indirect jump.*/
+#  define ZSTDLIB_API __declspec(dllimport) ZSTDLIB_VISIBLE /* It isn't required but allows to generate better code, saving a function pointer load from the IAT and an indirect jump.*/
+#  define ZSTDLIB_STATIC_API
 #else
-#  define ZSTDLIB_API ZSTDLIB_VISIBILITY
+#  define ZSTDLIB_API ZSTDLIB_VISIBLE
+#  define ZSTDLIB_STATIC_API ZSTDLIB_HIDDEN
```
Cyan4973: Keep the default to `VISIBLE`.
Make it possible for this value to be externally defined (`#ifndef ZSTDLIB_STATIC_API` ...).

Cyan4973:
> Should I amend this PR to only include the `Makefile` changes? As I understand it, those would be useful anyway, wouldn’t they?

I believe most of this PR is fine.
It's a good thing to have the ability to hide experimental symbols on demand.

What should be updated is that the default should remain to `VISIBLE`.
Moreover, in order to enable the "on demand" capability to hide experimental symbols,
it should be possible to change `ZSTDLIB_STATIC_API` macro value at compilation stage, directly from the command line.
And since it's a new capability, it should be testable.

lib/compress/zstd_compress.c
skitt: Done!

lib/zstd.h
skitt: Done!

skitt:
This is still missing tests.

Cyan4973:
This PR was in a relatively good shape,
and can still make it for next `zstd` release.

Current problems are : 
- it's a bit old now, and should be rebased on latest `dev` branch
- I would prefer `ZSTDLIB_STATIC_API` to be defined in the second section, in order to clearly indicate that this attribute is only meant for symbols in the "static linking only" section
- pay attention to the way static library is used, a few test programs expect access to hidden symbols to run correctly. Therefore, don't change the build rule for the static library (it's a different topic, more complex than this PR).

Would you want to complete it ?

Alternatively, your PR could be merged into a specific feature branch, in order to be finalized before reaching `dev`.

skitt:
Yes, I do intend to complete it, I’ll try to take care of it in the next few days. I’ve pulled the `libzstd-nomt` flags part out into #2628 because that’s really separate IMO. For this main part I’ll move the definition and try to add some tests.

Cyan4973:
This PR is interesting, but needs to be rebased on latest `dev` branch.
I'll take care of this action in a dedicated feature branch.

PR #2502, opened 2021-02-19
fix ubsan test errors

Cyan4973:
allows recovering from pointer overflow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -286,12 +286,11 @@ uasanregressiontest:
 msanregressiontest:
 	$(MAKE) -C $(FUZZDIR) regressiontest CC=clang CXX=clang++ CFLAGS="-O3 -fsanitize=memory" CXXFLAGS="-O3 -fsanitize=memory"
 
-# run UBsan with -fsanitize-recover=signed-integer-overflow
-# due to a bug in UBsan when doing pointer subtraction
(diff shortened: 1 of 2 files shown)


terrelln (approved):
FYI this won't work with gcc versions older than `8.0.0`, but that shouldn't be a problem.

terrelln:
I've fixed the LLVM fuzzers in https://github.com/facebook/zstd/pull/2503.

Makefile
```diff
@@ -309,10 +309,10 @@ asan32: clean
 	$(MAKE) -C $(TESTDIR) test32 CC=clang MOREFLAGS="-g -fsanitize=address"
 
 uasan: clean
-	$(MAKE) test CC=clang MOREFLAGS="-g -fno-sanitize-recover=all -fsanitize-recover=signed-integer-overflow -fsanitize=address,undefined -Werror"
+	$(MAKE) test CC=clang MOREFLAGS="-g -fno-sanitize-recover=all -fsanitize-recover=signed-integer-overflow -fsanitize-recover=pointer-overflow -fsanitize=address,undefined -Werror"
```
terrelln: Can we remove `-fsanitize-recover=signed-integer-overflow`?

Cyan4973: Yes.
At least on my local platform, it works fine.

PR #2504, opened 2021-02-20
Stop using deprecated reset?Stream functions

skitt:
These are replaced by the corresponding context resets. When
converting resetCStream, setPledgedSrcSize isn't called if the source
size is "unknown".

This helps reduce the reliance on "static only" symbols, as well as
reducing the use of deprecated functions.

Signed-off-by: Stephen Kitt <steve@sk2.org>

### DIFF
--- a/contrib/pzstd/Pzstd.cpp
+++ b/contrib/pzstd/Pzstd.cpp
@@ -274,7 +274,7 @@ static void compress(
     return;
   }
   {
-    auto err = ZSTD_resetCStream(ctx.get(), 0);
+    auto err = ZSTD_CCtx_reset(ctx.get(), ZSTD_reset_session_only);
     if (!errorHolder.check(!ZSTD_isError(err), ZSTD_getErrorName(err))) {
       return;
     }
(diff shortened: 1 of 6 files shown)


Cyan4973 (approved):

tests/zstreamtest.c
```diff
@@ -1071,19 +1071,19 @@ static int basicUnitTests(U32 seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
-    DISPLAYLEVEL(3, "test%3i : ZSTD_resetDStream() with dictionary : ", testNb++);
```
Cyan4973: I would recommend not modifying the test functions,
such as the ones included in `tests/zstreamtest.c`
which scope _includes checking static symbols_.

skitt:
I don’t think the FreeBSD failure is related to this patch:

    ld-elf.so.1: /usr/local/bin/gmd5sum: Undefined symbol "fread_unlocked@FBSD_1.6"

Cyan4973:
Nope, FreeBSD issues are unrelated.

Cyan4973 (approved):
I'm fine with the proposed changes.

I would wait for @terrelln 's feedback regarding changes in the `linux-kernel` and `pzstd` projects.

terrelln:
The changes to pzstd looks good.

I'd like to revert the changes to linux-kernel. It isn't used in user-space, so it doesn't impact the static symbol usage. And I'd rather keep the implementations in the `module.c` files straightforward.

skitt:
> I'd like to revert the changes to linux-kernel. It isn't used in user-space, so it doesn't impact the static symbol usage. And I'd rather keep the implementations in the `module.c` files straightforward.

Done, thanks for the review!

PR #2505, opened 2021-02-22
ZSTD_seekable_decompress() example that hangs.

mlindsay:
This generates a hang using the zstd_seekable format. It is contrived as it was generated using the fuzzer at https://fuchsia.googlesource.com/fuchsia/+/6c0911153a4b36d5a6ab86f74cc9d1a313d8e8ca/src/storage/blobfs/test/zstd-seekable-fuzzer.cc  but it should return an error for an invalid archive instead of hanging.

This change is just a reference for issue #2506

### DIFF
--- a/contrib/seekable_format/tests/seekable_tests.c
+++ b/contrib/seekable_format/tests/seekable_tests.c
@@ -53,6 +53,60 @@ int main(int argc, const char** argv)
     }
     printf("Success!\n");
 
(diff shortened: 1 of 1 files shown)


senhuang42:
Big thanks for noticing this and writing out this test case! We've fixed the issue in #2516 (including your test case) and added a roundtrip seekable fuzzer #2617 here.

Issue #2509, opened 2021-02-24
--output-dir-mirror fails with file names containing ".."


awused:
**Describe the bug**
--output-dir-mirror fails to compress files with names containing ".."

This also applies to directories containing ".." in their names.

**To Reproduce**
```
touch a..b
zstd a..b --output-dir-mirror output
```
Produces the error `zstd: --output-dir-mirror cannot compress 'a..b' into 'output'`

**Expected behavior**
"a..b" is compressed as "output/a..b.zst"

**Desktop (please complete the following information):**
 - OS: Fedora Linux
 - Version: 1.4.7

PR #2512, opened 2021-02-26
Detect `..` in Paths Correctly

felixhandte:
This commit addresses #2509.

As it's a `static` function, I don't know a good way to test it directly. I tested it some via `--output-dir-mirror` on the CLI, all of which seemed to work well.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -679,7 +679,27 @@ const char* UTIL_getFileExtension(const char* infilename)
 
 static int pathnameHas2Dots(const char *pathname)
 {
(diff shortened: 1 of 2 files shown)


Cyan4973:
>  I don't know a good way to test it directly. I tested it some via --output-dir-mirror on the CLI, all of which seemed to work well.

Adding a test case employing `--output-dir-mirror` looks good enough to me.

terrelln (approved):

Cyan4973 (approved):

PR #2518, opened 2021-03-03
New direct seekTable access methods

Cyan4973:
This is a follow up of #2113, incorporating several comments made on this PR, 
plus a few minor refactoring operations, both for style and more robust compilation.

Additional topics to consider : 
- The reasoning for the creation of this new abstraction layer should be better documented (__done__)
- There should be at least one automated test involving the new abstraction (__done__)
- I'm puzzled about the object's name (`ZSTD_seekTable`) which feels a bit too close to existing `ZSTD_seekable`. Basically, one letter difference. The type system should avoid mixing wrong types, but it's still a potential source of confusion. Problem is, I haven't found any better name (__no change__)
- Not directly related, but I noticed that I/O `FILE*` functions are present in the same unit as memory-only ones. Seems like they'd better be separate, as it's entirely possible for an application to only need memory access functions, without the need to drag `<stdio.h>` dependencies (__separate PR__)

This PR most likely conflicts with #2516 , though hopefully nothing too difficult to fix. (_edit_: #2516 is now merged into this PR)


senhuang42 (approved):
These changes look good to me! The roundtrip unit test is a great addition.

PR #2522, opened 2021-03-04
Reduce stack usage of ZSTD_buildCTable()

terrelln:
It is a stack high-point for some compression strategies and has an easy fix. This moves the normalized count into the entropy workspace.

### DIFF
--- a/lib/compress/zstd_compress_sequences.c
+++ b/lib/compress/zstd_compress_sequences.c
@@ -232,6 +232,11 @@ ZSTD_selectEncodingType(
     return set_compressed;
 }
 
+typedef struct {
+    S16 norm[MaxSeq + 1];
+    U32 wksp[FSE_BUILD_CTABLE_WORKSPACE_SIZE_U32(MaxSeq, MaxFSELog)];
+} ZSTD_BuildCTableWksp;
+
 size_t
 ZSTD_buildCTable(void* dst, size_t dstCapacity,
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress_sequences.c
```diff
@@ -258,19 +263,19 @@ ZSTD_buildCTable(void* dst, size_t dstCapacity,
...
         const U32 tableLog = FSE_optimalTableLog(FSELog, nbSeq, max);
         if (count[codeTable[nbSeq-1]] > 1) {
             count[codeTable[nbSeq-1]]--;
             nbSeq_1--;
         }
         assert(nbSeq_1 > 1);
-        assert(entropyWorkspaceSize >= FSE_BUILD_CTABLE_WORKSPACE_SIZE(MaxSeq, MaxFSELog));
-        FORWARD_IF_ERROR(FSE_normalizeCount(norm, tableLog, count, nbSeq_1, max, ZSTD_useLowProbCount(nbSeq_1)), "");
-        {   size_t const NCountSize = FSE_writeNCount(op, oend - op, norm, max, tableLog);   /* overflow protected */
+        RETURN_ERROR_IF(entropyWorkspaceSize < sizeof(ZSTD_BuildCTableWksp), GENERIC, "wksp too small");
```
Cyan4973: minor: if you believe that passing a too small `entropyWorkspaceSize` is something that should never happen if the code is correct, then using an `assert()` instead of a runtime error check would be appropriate here.

Cyan4973 (approved):

PR #2523, opened 2021-03-05
Add HUF_writeCTable_wksp() function

terrelln:
This saves ~700 bytes of stack space in `HUF_writeCTable()` by adding a workspace variant and using it in the `HUF_compress_internal()`.


lib/compress/huf_compress.c
```diff
@@ -732,7 +751,10 @@ static size_t HUF_compressCTable_internal(
 typedef struct {
     unsigned count[HUF_SYMBOLVALUE_MAX + 1];
     HUF_CElt CTable[HUF_SYMBOLVALUE_MAX + 1];
-    HUF_buildCTable_wksp_tables buildCTable_wksp;
+    union {
```
Cyan4973, line 754: unfortunately, nameless `union` is not a `C90` feature.

terrelln: Bummer, I'll name it then

Cyan4973 (approved):

PR #2524, opened 2021-03-05
[huf] Reduce stack usage of HUF_readDTableX2 by ~972 bytes

terrelln:
* Use `HUF_readStats_wksp()`
* Use workspace in `HUF_fillDTableX2*()`
* Clean up workspace usage to use a workspace struct
* Move a 512 byte array in `FSE_decompress_wksp()` into the workspace.
* Increase `HUF_DECOMPRESS_WORKSPACE_SIZE` by 512 bytes to accommodate the new array.

This reduces the decompression high water mark by 972 bytes.

### DIFF
--- a/lib/common/fse_decompress.c
+++ b/lib/common/fse_decompress.c
@@ -318,22 +318,29 @@ FORCE_INLINE_TEMPLATE size_t FSE_decompress_wksp_body(
 {
     const BYTE* const istart = (const BYTE*)cSrc;
     const BYTE* ip = istart;
-    short counting[FSE_MAX_SYMBOL_VALUE+1];
     unsigned tableLog;
     unsigned maxSymbolValue = FSE_MAX_SYMBOL_VALUE;
-    FSE_DTable* const dtable = (FSE_DTable*)workSpace;
+    short* counting = (short*) workSpace;
+    size_t const countingSize = sizeof(short) * (FSE_MAX_SYMBOL_VALUE + 1);
+    FSE_DTable* const dtable = (FSE_DTable*)(void*)((BYTE*)workSpace + countingSize);
+
+    DEBUG_STATIC_ASSERT((FSE_MAX_SYMBOL_VALUE + 1) % 2 == 0);
+    if (wkspSize < countingSize) return ERROR(GENERIC);
 
     /* normal FSE decoding mode */
-    size_t const NCountLength = FSE_readNCount_bmi2(counting, &maxSymbolValue, &tableLog, istart, cSrcSize, bmi2);
-    if (FSE_isError(NCountLength)) return NCountLength;
-    if (tableLog > maxLog) return ERROR(tableLog_tooLarge);
-    assert(NCountLength <= cSrcSize);
-    ip += NCountLength;
-    cSrcSize -= NCountLength;
+    {
+        size_t const NCountLength = FSE_readNCount_bmi2(counting, &maxSymbolValue, &tableLog, istart, cSrcSize, bmi2);
+        if (FSE_isError(NCountLength)) return NCountLength;
+        if (tableLog > maxLog) return ERROR(tableLog_tooLarge);
+        assert(NCountLength <= cSrcSize);
+        ip += NCountLength;
+        cSrcSize -= NCountLength;
+    }
 
     if (FSE_DECOMPRESS_WKSP_SIZE(tableLog, maxSymbolValue) > wkspSize) return ERROR(tableLog_tooLarge);
     workSpace = dtable + FSE_DTABLE_SIZE_U32(tableLog);
     wkspSize -= FSE_DTABLE_SIZE(tableLog);
+    wkspSize -= countingSize;
 
     CHECK_F( FSE_buildDTable_internal(dtable, counting, maxSymbolValue, tableLog, workSpace, wkspSize) );
 
--- a/lib/common/fse_decompress.c
+++ b/lib/common/fse_decompress.c
@@ -310,6 +310,11 @@ size_t FSE_decompress_wksp(void* dst, size_t dstCapacity, const void* cSrc, size
     return FSE_decompress_wksp_bmi2(dst, dstCapacity, cSrc, cSrcSize, maxLog, workSpace, wkspSize, /* bmi2 */ 0);
 }
 
+typedef struct {
+    short count[FSE_MAX_SYMBOL_VALUE + 1];
+} FSE_NormalizedCount;
+
+
 FORCE_INLINE_TEMPLATE size_t FSE_decompress_wksp_body(
         void* dst, size_t dstCapacity,
         const void* cSrc, size_t cSrcSize,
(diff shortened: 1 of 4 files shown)


lib/common/fse_decompress.c
```diff
@@ -318,22 +318,29 @@ FORCE_INLINE_TEMPLATE size_t FSE_decompress_wksp_body(
 {
     const BYTE* const istart = (const BYTE*)cSrc;
     const BYTE* ip = istart;
-    short counting[FSE_MAX_SYMBOL_VALUE+1];
     unsigned tableLog;
     unsigned maxSymbolValue = FSE_MAX_SYMBOL_VALUE;
-    FSE_DTable* const dtable = (FSE_DTable*)workSpace;
+    short* counting = (short*) workSpace;
+    size_t const countingSize = sizeof(short) * (FSE_MAX_SYMBOL_VALUE + 1);
+    FSE_DTable* const dtable = (FSE_DTable*)(void*)((BYTE*)workSpace + countingSize);
```
Cyan4973: minor:
since `counting` is an array of fixed size here,
we have the opportunity to declare a `struct` describing the content of `workSpace`,
in order for pointer arithmetic to be solved automatically by the type system.

lib/decompress/huf_decompress.c
```diff
@@ -608,6 +614,15 @@ static void HUF_fillDTableX2(HUF_DEltX2* DTable, const U32 targetLog,
...
 }
 
+typedef struct {
+    rankValCol_t rankVal[HUF_TABLELOG_MAX];
+    U32 rankStats[HUF_TABLELOG_MAX + 1];
+    U32 rankStart0[HUF_TABLELOG_MAX + 2];
+    sortedSymbol_t sortedSymbol[HUF_SYMBOLVALUE_MAX + 1];
+    BYTE weightList[HUF_SYMBOLVALUE_MAX + 1];
+    U32 wksp[HUF_READ_STATS_WORKSPACE_SIZE_U32];
+} HUF_ReadDTableX2_Workspace;
```
Cyan4973, line 624: :+1:

lib/decompress/huf_decompress.c
```diff
@@ -620,47 +635,32 @@ size_t HUF_readDTableX2_wksp(HUF_DTable* DTable,
...
-    sortedSymbol = (sortedSymbol_t *)workSpace + (spaceUsed32 * sizeof(U32)) / sizeof(sortedSymbol_t);
-    spaceUsed32 += HUF_ALIGN(sizeof(sortedSymbol_t) * (HUF_SYMBOLVALUE_MAX + 1), sizeof(U32)) >> 2;
-    weightList = (BYTE *)((U32 *)workSpace + spaceUsed32);
-    spaceUsed32 += HUF_ALIGN(HUF_SYMBOLVALUE_MAX + 1, sizeof(U32)) >> 2;
-
-    if ((spaceUsed32 << 2) > wkspSize) return ERROR(tableLog_tooLarge);
-
-    rankStart = rankStart0 + 1;
-    ZSTD_memset(rankStats, 0, sizeof(U32) * (2 * HUF_TABLELOG_MAX + 2 + 1));
+    HUF_ReadDTableX2_Workspace* wksp = (HUF_ReadDTableX2_Workspace*)workSpace;
```
Cyan4973: very minor : 
`HUF_ReadDTableX2_Workspace* const wksp`

lib/decompress/huf_decompress.c
```diff
@@ -620,47 +635,32 @@ size_t HUF_readDTableX2_wksp(HUF_DTable* DTable,
...
+
+    if (sizeof(*wksp) > wkspSize) return ERROR(tableLog_tooLarge);
```
Cyan4973: minor:
Is that the best error nb ?
`tableLog` only accounts for a minor part of `*wksp`,
and only indirectly, since its value is fixed at compile time,
so there is nothing the user can do about it.

lib/decompress/huf_decompress.c
```diff
@@ -608,6 +614,15 @@ static void HUF_fillDTableX2(HUF_DEltX2* DTable, const U32 targetLog,
     }
 }
 
+typedef struct {
+    rankValCol_t rankVal[HUF_TABLELOG_MAX];
+    U32 rankStats[HUF_TABLELOG_MAX + 1];
+    U32 rankStart0[HUF_TABLELOG_MAX + 2];
+    sortedSymbol_t sortedSymbol[HUF_SYMBOLVALUE_MAX + 1];
+    BYTE weightList[HUF_SYMBOLVALUE_MAX + 1];
+    U32 wksp[HUF_READ_STATS_WORKSPACE_SIZE_U32];
```
Cyan4973: Could this member be named in a slightly more specific way ?

Cyan4973:
This is a good saving, and a nice refactor, especially when employing `struct` to replace pointer arithmetic.
There are just a few minor comments left, for the most part this PR is good.

lib/decompress/huf_decompress.c
terrelln: Sure, I can change it. I was preserving the same error as before, but we can change to GENERIC.

terrelln:
@Cyan4973 is this good to go?

lib/common/fse_decompress.c
```diff
@@ -310,6 +310,11 @@ size_t FSE_decompress_wksp(void* dst, size_t dstCapacity, const void* cSrc, size
     return FSE_decompress_wksp_bmi2(dst, dstCapacity, cSrc, cSrcSize, maxLog, workSpace, wkspSize, /* bmi2 */ 0);
 }
 
+typedef struct {
+    short count[FSE_MAX_SYMBOL_VALUE + 1];
+} FSE_NormalizedCount;
```
Cyan4973: very minor :
I think I wasn't specific enough in my previous comment:
I was expecting a second `struct` member `FSE_DTable dtable[1]`.

The point was to avoid pointer arithmetic to determine `dtable`.
Thus, instead of `FSE_DTable* const dtable = (FSE_DTable*)(void*)((BYTE*)workSpace + sizeof(*ncount));`,
it would rather be `FSE_DTable* const dtable = &ncount.dtable;`.

Having a last member of a `struct` as an array of size `1` is a known trick in `C`.
As the size is not enforced once the array decays as a pointer, it can effectively represent an array of any size.
Of course, `sizeof(struct)` is no longer correct then, since it's unaware of the real size of the array.
It's also not possible to stack anything after the array.

Some people prefer using an array of size `[0]` for this purpose,
which is indeed nice to make `sizeof(struct)` represent the size of members _before_ the variable size array.
However, I read here and there that it could cause compatibility concerns with "some" compilers (though I don't know which one), hence I tend to use `[1]` to avoid such issue.
That being said, another consensus was that the size should represent the minimum expected size of the array,
thus `[0]` is relevant for an optional array that may be empty.

Cyan4973 (approved):

Cyan4973:
Yes, that's fine.
I have one last comment, although it's more a suggestion than a blocking point.
Feel free to proceed as you feel is best.

lib/common/fse_decompress.c
terrelln: Sure, I wasn't sure if 0-sized arrays were a standard compliant feature or not. But I can give it a go and see if anything in CI complains.

Issue #2491, opened 2021-02-11
Race condition allows attacker to access world-readable destination file


sdelafond:
The patches for #1630 still create the file with the default umask, before chmod'ing down to 0600, so an attacker could still open it in the meantime. inotify can for instance help automating such an attack.

zstd should either set the mode directly through open(2), or use umask(2) before creating the file.

This is Debian bug https://bugs.debian.org/982519.

carnil:
It was suggested in https://bugs.debian.org/981404#28 to do something like:

```
fd = open(dstFileName, O_WRONLY|O_CREAT|O_EXCL, 0600);
if (fd != -1)
    f = fdopen( fd, "wb" );
if (fd == -1 || f == NULL)
    DISPLAYLEVEL(1, "zstd: %s: %s\n", dstFileName, strerror(errno));
return f;
```

felixhandte:
Thanks for the report. The fix has been merged to `dev` and will go out in the next release.

PR #2525, opened 2021-03-05
Improve Setting Permissions of Created Files

felixhandte:
This PR is a follow-up to issues #1630 and #2491 (and their associated PRs #1644 and #2495).

This stack starts by adding tests to (a) codify the expected behavior (which is at present poorly defined) and (b) enforce that zstd implements that behavior.

I expect, once we've settled on the desired behavior, to refactor how zstd sets permissions on created files. In particular, I expect to switch to the `open()` + `fdopen()` pattern suggested in [the bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981404#28) and possibly remove the trailing `chmod()` (if we can get the permissions right from the start, why wait until the end to set them?).

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -124,6 +124,23 @@ case "$UNAME" in
     Darwin | FreeBSD | OpenBSD | NetBSD) MTIME="stat -f %m" ;;
 esac
 
+GET_PERMS="stat -c %a"
+case "$UNAME" in
+    Darwin | FreeBSD | OpenBSD | NetBSD) GET_PERMS="stat -f %a" ;;
+esac
+
+assertFilePermissions() {
+    STAT1=$($GET_PERMS "$1")
+    STAT2=$2
+    [ "$STAT1" = "$STAT2" ] || die "permissions on $1 don't match expected ($STAT1 != $STAT2)"
+}
+
+assertSamePermissions() {
+    STAT1=$($GET_PERMS "$1")
+    STAT2=$($GET_PERMS "$2")
+    [ "$STAT1" = "$STAT2" ] || die "permissions on $1 don't match those on $2 ($STAT1 != $STAT2)"
+}
+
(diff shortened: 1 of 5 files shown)


flx42:
Hello!

When compressing from stdin, we were a bit surprised to discover than the behavior of zstd changed on Ubuntu 20.04 as a result of CVE-2021-24031 being fixed in security package update [`1.4.4+dfsg-3ubuntu0.1`](http://changelogs.ubuntu.com/changelogs/pool/main/libz/libzstd/libzstd_1.4.4+dfsg-3ubuntu0.1/changelog). Before this fix, the output file would be created as `0666`. With the new package version, the file is created as `0600`. 

Looks like this PR restores the previous behavior when compressing from stdin, so it would be great to have for us. Our current workaround will be to manually chmod the output file, but we might also build a new release of zstd with this patch, if it gets accepted.

Cyan4973 (approved):

Cyan4973:
The fuzz sanitizer test failures are unrelated to this PR.

But there are still some issues left on Windows : 
https://ci.appveyor.com/project/YannCollet/zstd-p0yf0/builds/38127695/job/xvecnmj07lttm64l#L252

tests/playTests.sh
```diff
@@ -209,7 +209,7 @@ println "test : compress to stdout"
 zstd tmp -c > tmpCompressed
 zstd tmp --stdout > tmpCompressed       # long command format
 println "test : compress to named file"
-rm tmpCompressed
+rm -f tmpCompressed
```
Cyan4973, line 212: minor:
a variable `$RM` would be better to propagate this change throughout the script

Cyan4973 (approved):
My only comment is that
I would have replaced all these `rm -f` in `playTests.sh`
by a variable `$RM`
but that's minor.

PR #2534, opened 2021-03-12
Fix seek table descriptor check when loading

foxeng:
The `Seek_Table_Descriptor` in the seekable format's `Seek_Table_Footer` defines bits 6-2 as reserved.

This fixes the corresponding check when loading a seek table for decompression.


senhuang42 (approved):
LGTM. Thanks for fixing the CI test as well @foxeng !

Issue #2520, opened 2021-03-03
Signed releases


miyashitatouka:
I see the latest version's git tag was PGP signed by someone other than the github-actions bot. That's great. Downstream packagers really appreciate authenticity from upstream projects. I have two questions then:

- Can we count on future releases being signed by Felix Handte's key (B774B0A2C0A08A8A7C1D3EF499B5DE0B30966569)?

- If so, could we please also get signed tarball releases with this key, including one for 1.4.9?

felixhandte:
Hi! I'm surprised you noticed. This is something I did as an experiment. I wasn't sure to what extent people care about this or whether it would replace or just supplement the GitHub signature (although how valuable is that signature really?).

We generally rotate responsibility for the release process amongst the team members. So, no, we won't use my key for every future release. But if there's interest, we could probably set up a signing key for the project.

anthraxx:
There is a huge interest in this from Arch Linux and I'm sure others would benefit and be grateful about this bit of the supply chain security as well. GitHub signature are in my humble opinion the worst feature ever invented, it gives virtually zero benefit and weighs some people in false security -- signature on a project level should exclusively authenticate the developers and project owners which is the only legitimacy that *really* matters, contrary to the used platform.

A dedicated project signing key makes sense, however if it ever gets rotated please make sure to provide a chain of trust with either cleartext sigs of the old key legitimizing the new key or do cross key sigs :cat: Sorry if this got a wall of text and it all seems super obvious to you, I'm nowadays just mentioning all this because I've seen way too many projects doing this wrong and feel like it can't hurt to throw in some of my views and experience to avoid running into problems again :cat2:

miyashitatouka:
A single project key would be much appreciated!

felixhandte:
So, input would be appreciated on this topic. It seems like we have a number of options about how we could set this up.

I think the biggest question is: Is it more valuable to have the release tag itself signed? Or the tarballs? (Or both?)

Also: Is it valuable to produce and sign binaries for OS X and Windows? (This one seems less likely since it would be a lot more setup work on our end.)

anthraxx:
@felixhandte the more the merrier :cat:
Speaking from a Linux perspective:
I would say the MVP would be to sign the tarballs -- as the tarballs are the officially distributed containers that are meant for downstream projects to consume. It would therefor totally make sense to provide them in a way that the authenticity could be verified.

[...]

PR #2535, opened 2021-03-12
Add GitHub Action to Automatically Publish Release Tarballs

felixhandte:
This PR introduces a GitHub action that is triggered on release creation, which creates the release tarball, compresses it, hashes it, signs it, and attaches all of those files to the release.

Here's an example invocation on my repo:

Screenshot from 2021-03-12 17-33-48

I wanted to also generate the windows artifacts but windows is hard and I gave up.

Because this PR signs the artifacts when a key is provided, this is one step towards addressing #2520.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,33 +36,6 @@ jobs:
             make armbuild   V=1; make clean
             make -C tests test-legacy test-longmatch; make clean
             make -C lib libzstd-nomt; make clean
-  # This step is only run on release tags.
-  # It publishes the source tarball as artifacts and if the GITHUB_TOKEN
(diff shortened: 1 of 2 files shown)


terrelln:
This looks good to me! I don't have a strong preference on the tarball names. As long as we aren't changing the old names, it shouldn't be a huge deal to add `v`. But, I guess I'd very slightly prefer keeping `zstd-1.4.9`.

terrelln (approved):

mohe2015:
I think this is fine as an MVP but I really don't like it that this seems to do signing on Github's servers. The keys should be guarded as much as possible - some people even use hardware security keys.

PR #2538, opened 2021-03-16
Add memory monotonicity test over srcSize

senhuang42:
@terrelln and I were discussing about monotonicity of memory usage as source size increases, as well as compression level increases. This PR adds a test for the former case: mem usage monotonically increases w/ `srcSize`.

It actually appears that we don't currently do this. Between the 16 KB and 128 KB parameters for level 1, if the source file is > 16 KB and <= 128 KB, we actually use less memory compared to source files <= 16 KB. The parameters in [`zstd_compress.c`](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L5114) confirm this `wLog: 14 cLog: 14 hLog: 15` -> `wLog: 17 cLog: 12 hLog:13` is a decrease in memory usage.

Is this something we would want to fix? Do we also want to maintain memory usage monotonicity as compression level increases?

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1390,8 +1390,15 @@ size_t ZSTD_estimateCCtxSize_usingCParams(ZSTD_compressionParameters cParams)
 
 static size_t ZSTD_estimateCCtxSize_internal(int compressionLevel)
 {
-    ZSTD_compressionParameters const cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, 0, ZSTD_cpm_noAttachDict);
-    return ZSTD_estimateCCtxSize_usingCParams(cParams);
+    int tier = 0;
+    size_t largestSize = 0;
+    static const unsigned long long srcSizeTiers[4] = {16 KB, 128 KB, 256 KB, ZSTD_CONTENTSIZE_UNKNOWN};
+    for (; tier < 4; ++tier) {
+        /* Choose the set of cParams for a given level across all srcSizes that give the largest cctxSize */
+        ZSTD_compressionParameters const cParams = ZSTD_getCParams_internal(compressionLevel, srcSizeTiers[tier], 0, ZSTD_cpm_noAttachDict);
+        largestSize = MAX(ZSTD_estimateCCtxSize_usingCParams(cParams), largestSize);
+    }
+    return largestSize;
 }
 
 size_t ZSTD_estimateCCtxSize(int compressionLevel)
@@ -1399,6 +1406,7 @@ size_t ZSTD_estimateCCtxSize(int compressionLevel)
     int level;
     size_t memBudget = 0;
     for (level=MIN(compressionLevel, 1); level<=compressionLevel; level++) {
+        /* Ensure monotonically increasing memory usage as compression level increases */
         size_t const newMB = ZSTD_estimateCCtxSize_internal(level);
         if (newMB > memBudget) memBudget = newMB;
     }
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -3181,6 +3181,31 @@ static int basicUnitTests(U32 const seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : check compression mem usage monotonicity over srcSize : ", testNb++);
+    {
+        size_t const kSizeIncrement = 2 KB;
+        int level = -3;
+
+        for (; level <= ZSTD_maxCLevel(); ++level) {
+            size_t dictSize = 0;
+            for (; dictSize <= 256 KB; dictSize += 8 * kSizeIncrement) {
+                size_t srcSize = 2 KB;
+                size_t prevCCtxSize = 0;
+                for (; srcSize < 300 KB; srcSize += kSizeIncrement) {
+                    ZSTD_compressionParameters const cParams = ZSTD_getCParams(level, srcSize, dictSize);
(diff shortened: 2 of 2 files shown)


terrelln:
Is that the only instance of non-monotonicity for source size?

terrelln:
I think we need monotonicity in source size, at the very least least when reporting `ZSTD_estimateCCtxSize()` et al. And for simplicity, it seems like we should guarantee it in the parameters as well.

Right now if you use `ZSTD_estimateCCtxSize(1)`, you could not compress a source of <= 16 KB with `ZSTD_compressCCtx()` (unless I'm missing something).

The only reason BtrFS (which pre-allocates workspaces) isn't bitten by this is because they use streaming. When the smaller parameters are selected, the window is also shrunk enough to make up for the lack of table space.

senhuang42:
Mem usage also decreases for level 4 as we cross the 128 KB parameter boundary. 1559264 bytes -> 1297120 bytes reported by `estimateCCtxSize_usingCParams()`. These appear to be the only two.

senhuang42:
> Right now if you use `ZSTD_estimateCCtxSize(1)`, you could not compress a source of <= 16 KB with `ZSTD_compressCCtx()` (unless I'm missing something).

I think this actually might be fine? `ZSTD_estimateCCtxSize()` assumes `ZSTD_CONTENTSIZE_UNKNOWN` and therefore should select the > 256K parameter set. But definitely for the specific case tested in the unit test, generating CParams `(level 1, >16K, <=128K src)`, then creating a static CCtx based on a size estimate for that and trying to compress `<= 16K`, could go badly.

terrelln:
> I think this actually might be fine? ZSTD_estimateCCtxSize() assumes ZSTD_CONTENTSIZE_UNKNOWN and therefore should select the > 256K parameter set. But definitely for the specific case tested in the unit test, generating CParams, then creating a static CCtx based on a size estimate for that, could go badly.


https://github.com/facebook/zstd/blob/413b3198b01327e83c668ae488c6d22e64def14c/lib/compress/zstd_compress.c#L5059-L5062

https://github.com/facebook/zstd/blob/413b3198b01327e83c668ae488c6d22e64def14c/lib/compress/zstd_compress.c#L5137-L5140

When using `ZSTD_estimateCCtxSize(1)` the window size isn't counted (it is with `ZSTD_estimateCStreamSize()`). So the <= 16 KB params use more memory than the > 256 KB params for single-pass compression.

terrelln:
I think monotonicity in the table wrt source size makes sense. Here is the scenario I'm imagining:

```c++
auto cparams = ZSTD_getCParams(1, 0, 0);
auto size = ZSTD_estimateCCtxSize_usingCParams(cparams);
void* wksp = malloc(size);

...

auto params = ZSTD_getParams(1, srcSize, 0);
auto cctx = ZSTD_initStaticCCtx();
ZSTD_compress_advanced(cctx, dst, dstCapacity, src, srcSize, params);
```

This shows up in the Linux kernel in `crypto/zstd.c`. It can't "tune" the parameters for the source size because of this limitation.

https://github.com/torvalds/linux/blob/1df27313f50a57497c1faeb6a6ae4ca939c85a7d/crypto/zstd.c#L35-L36
https://github.com/torvalds/linux/blob/1df27313f50a57497c1faeb6a6ae4ca939c85a7d/crypto/zstd.c#L155

You might say, just use a compression level instead. Currently the kernel is limited to using cparams because of its API (though we could try to update it after the version update). And, I would intuitively expect that for all `level` and `srcSize` `memory(ZSTD_getCParams(level, 0, 0)) >= memory(ZSTD_getCParams(level, srcSize, 0))`.

terrelln:
So the reason why level 1 with 16KB source is better with a larger hash table seems to be branch misses. With 32K entries in the hash table, collisions are rare. So you generally either end up with a match, or end up with `< lowLimit`.

With 8K entries compression has about 5.5% missed branches. With 32K entries it has about 4.0% missed branches. With 64K entries it has 3.6% missed branches and is slightly faster than hashLog=15.

terrelln:
@senhuang42 do you have the patch for `estimateCCtxSize()` ready?

senhuang42:
> @senhuang42 do you have the patch for `estimateCCtxSize()` ready?

Thanks for the reminder - I actually was looking at this again, and I actually wasn't able to repro any issue with the existing `estimateCCtxSize(1)` - looking at the code, doesn't it take windowLog into account here or am I missing something? https://github.com/facebook/zstd/blob/515807182f16b4a94e73ad2f1d1b83cc151bdad3/lib/compress/zstd_compress.c#L1336-L1342

If `windowLog < 17`, we effectively would reduce `tokenSpace`, so in this case the 16 K params would use less mem than the >256K params (though the matchstate would still be larger) since the the increase in matchstate size from the hash and chainlog are offset by the smaller tokenSpace from small enough windowLog.

Currently, adjusting `ZSTD_estimateCCtxSize_internal()` so it assumes 16 KB, I get 211680 bytes, but for `ZSTD_CONTENTSIZE_UNKNOWN`: 576224 bytes. So it seems my patch doesn't actually do anything currently, but makes the size estimation a bit safer in case the default params got adjusted to a point where this might be an issue?

terrelln:
Didn't 128KB level 4 use more memory than 256KB level 4? In which case the looping is necessary because `tokenSpace` would be the same.

terrelln (approved):

senhuang42:
> Didn't 128KB level 4 use more memory than 256KB level 4? In which case the looping is necessary because `tokenSpace` would be the same.

128K L4 used more than 256K L4, but not more than >256K L4, so estimateCCtxSize(4) would still be fine in that case, I think.

But I think the looping would still be nice to have (row hash might introduce oddities?)

Cyan4973 (approved):

PR #2540, opened 2021-03-17
Fix dedicated dict search isSupported() requirements.

senhuang42:
Currently, asserts in `zstd_lazy.c` will fail if you try to compress with `chainLog > 24` or `hLog < cLog` with dedicated dict search.

The following command fails on `dev`:
`make -j zstd DEBUGLEVEL=1 && ./zstd [file] -D [dictFile] --zstd=clog=25,hlog=23 -7f`

This PR adds the same checks that are `assert()`ed to be true in `zstd_lazy.c` when using dedicated dict search, and includes a unit test that fails without this fix.

Though, I will note that dedicated dict search at least seems to work fine when `clog > 24` or `hLog < cLog`, so an alternative could just be to remove the asserts.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -5186,7 +5186,10 @@ static ZSTD_compressionParameters ZSTD_dedicatedDictSearch_getCParams(int const 
 static int ZSTD_dedicatedDictSearch_isSupported(
         ZSTD_compressionParameters const* cParams)
 {
-    return (cParams->strategy >= ZSTD_greedy) && (cParams->strategy <= ZSTD_lazy2);
+    return (cParams->strategy >= ZSTD_greedy)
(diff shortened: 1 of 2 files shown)


felixhandte (approved):
Cool. Looks good.

As I recall, these asserts are present for the following reasons:

### `assert(chainLog <= 24);`

In the DDSS, the hashtable layout for each entry is:

- 4 bytes: first match index
- 4 bytes: second match index
- 4 bytes: third match index
- 1 byte: number of additional matches in the chain table
- 3 bytes: index of head of chain in chain table

So we only have 24 bits to index into the chain table.

### `assert(hashLog >= chainLog)`

We need this for DDSS table construction, because we first build the chain table in the hash table, and then sort the chains into the chain table proper.

PR #2541, opened 2021-03-17
simple fix for using bit operator.

ihsinme:
good day.
It seems to me that the developer intended to use a logical operator.
so I suggest a simple fix.

### DIFF
--- a/lib/compress/zstd_compress_literals.c
+++ b/lib/compress/zstd_compress_literals.c
@@ -117,7 +117,7 @@ size_t ZSTD_compressLiterals (ZSTD_hufCTables_t const* prevHuf,
         }
     }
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
The first version makes it possible to determine unitary tests results in parallel, and then combine them, resulting in a single branch.
In contrast, the second version requires them to be serialized, resulting in (up to) 3 more branches.

For performance, the first version is generally a bit better.

That being said, at this place in the code, such performance detail should not matter much, since it's only triggered once per block.
Hence there is case to prefer clarity.

ihsinme:
good day.
maybe I don’t know your PR procedures.
Should I remind you of this PR, or are you making changes elsewhere.

Cyan4973 (approved):

PR #2546, opened 2021-03-18
[cwksp] Align all allocated "tables" and "aligneds" to 64 bytes

senhuang42:
The general strategy is:
1. When we begin allocating aligneds or tables, move the start of that section so that it is 64-byte aligned, done through `ZSTD_cwksp_internal_advance_phase()`.
2. Guarantee that the size of every table allocated in the aligned or tables section are all 64-bytes, therefore all of them will be 64-byte aligned. This is done with `ZSTD_cwksp_aligned_alloc_size()` - however, the tables don't actually need this since they're already guaranteed to be in multiples of 64 (we additionally static assert this to be true).
3. Since we want our wksp estimation to still be accurate, we also will always reserve 64+64 bytes for the purposes of aligning the beginnings of the tables and aligned. The comments in `ZSTD_cwksp_slack_space_required()` explain this mechanism.

The two public functions added: `ZSTD_cwksp_slack_space_required()` and `ZSTD_cwksp_finalize()` are meant to be generic - the former returning all space required for internal purposes, and the latter handling any final things to take care of in the wksp after we're done allocating the buffers. In both cases we currently just deal with alignment.

Test Plan:
- Unit tests, manual spot checking of 64-byte alignment (and asserts that guarantee them)

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1576,17 +1576,6 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
         ZSTD_cwksp_mark_tables_dirty(ws);
     }
 
-    /* opt parser space */
-    if ((forWho == ZSTD_resetTarget_CCtx) && (cParams->strategy >= ZSTD_btopt)) {
-        DEBUGLOG(4, "reserving optimal parser space");
-        ms->opt.litFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (1<<Litbits) * sizeof(unsigned));
-        ms->opt.litLengthFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (MaxLL+1) * sizeof(unsigned));
-        ms->opt.matchLengthFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (MaxML+1) * sizeof(unsigned));
-        ms->opt.offCodeFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (MaxOff+1) * sizeof(unsigned));
-        ms->opt.matchTable = (ZSTD_match_t*)ZSTD_cwksp_reserve_aligned(ws, (ZSTD_OPT_NUM+1) * sizeof(ZSTD_match_t));
-        ms->opt.priceTable = (ZSTD_optimal_t*)ZSTD_cwksp_reserve_aligned(ws, (ZSTD_OPT_NUM+1) * sizeof(ZSTD_optimal_t));
-    }
-
     ms->hashLog3 = hashLog3;
 
     ZSTD_invalidateMatchState(ms);
@@ -1595,7 +1584,7 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
 
     ZSTD_cwksp_clear_tables(ws);
 
-    DEBUGLOG(4, "reserving table space");
+    DEBUGLOG(5, "reserving table space");
     /* table Space */
     ms->hashTable = (U32*)ZSTD_cwksp_reserve_table(ws, hSize * sizeof(U32));
     ms->chainTable = (U32*)ZSTD_cwksp_reserve_table(ws, chainSize * sizeof(U32));
@@ -1607,6 +1596,17 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
     if (crp!=ZSTDcrp_leaveDirty) {
         /* reset tables only */
         ZSTD_cwksp_clean_tables(ws);
+    }
+
+    /* opt parser space */
+    if ((forWho == ZSTD_resetTarget_CCtx) && (cParams->strategy >= ZSTD_btopt)) {
+        DEBUGLOG(4, "reserving optimal parser space");
+        ms->opt.litFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (1<<Litbits) * sizeof(unsigned));
+        ms->opt.litLengthFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (MaxLL+1) * sizeof(unsigned));
+        ms->opt.matchLengthFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (MaxML+1) * sizeof(unsigned));
+        ms->opt.offCodeFreq = (unsigned*)ZSTD_cwksp_reserve_aligned(ws, (MaxOff+1) * sizeof(unsigned));
+        ms->opt.matchTable = (ZSTD_match_t*)ZSTD_cwksp_reserve_aligned(ws, (ZSTD_OPT_NUM+1) * sizeof(ZSTD_match_t));
+        ms->opt.priceTable = (ZSTD_optimal_t*)ZSTD_cwksp_reserve_aligned(ws, (ZSTD_OPT_NUM+1) * sizeof(ZSTD_optimal_t));
     }
 
     ms->cParams = *cParams;
@@ -1761,6 +1761,14 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
         zc->seqStore.ofCode = ZSTD_cwksp_reserve_buffer(ws, maxNbSeq * sizeof(BYTE));
         zc->seqStore.sequencesStart = (seqDef*)ZSTD_cwksp_reserve_aligned(ws, maxNbSeq * sizeof(seqDef));
 
+        FORWARD_IF_ERROR(ZSTD_reset_matchState(
+            &zc->blockState.matchState,
+            ws,
+            &params.cParams,
+            crp,
+            needsIndexReset,
+            ZSTD_resetTarget_CCtx), "");
+
         /* ldm hash table */
         if (params.ldmParams.enableLdm) {
             /* TODO: avoid memset? */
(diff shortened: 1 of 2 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1311,18 +1311,23 @@ ZSTD_sizeof_matchState(const ZSTD_compressionParameters* const cParams,
...
+      + ZSTD_cwksp_aligned_alloc_size((1<<Litbits) * sizeof(U32))
+      + ZSTD_cwksp_aligned_alloc_size((ZSTD_OPT_NUM+1) * sizeof(ZSTD_match_t))
+      + ZSTD_cwksp_aligned_alloc_size((ZSTD_OPT_NUM+1) * sizeof(ZSTD_optimal_t));
     size_t const optSpace = (forCCtx && (cParams->strategy >= ZSTD_btopt))
                                 ? optPotentialSpace
                                 : 0;
+    size_t const slackSpace = ZSTD_cwksp_slack_space_required();
+
+    /* tables are guaranteed to be sized in multiples of 64 */
+    ZSTD_STATIC_ASSERT(ZSTD_HASHLOG_MIN >= 6 && ZSTD_WINDOWLOG_MIN >= 6 && ZSTD_CHAINLOG_MIN >= 6);
```
terrelln: Technically you should be checking hash & chain logs against 4, since they allocate U32's

lib/compress/zstd_compress.c
```diff
@@ -1579,7 +1584,18 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
 
     ZSTD_cwksp_clear_tables(ws);
 
-    DEBUGLOG(5, "reserving table space");
+    /* opt parser space */
```
terrelln: Why is this moved above tables?

lib/compress/zstd_compress.c
```diff
@@ -1778,13 +1775,20 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
...
+            needsIndexReset,
+            ZSTD_resetTarget_CCtx), "");
+
+        ZSTD_cwksp_finalize(ws);
         /* Due to alignment, when reusing a workspace, we can actually consume
          * up to 3 extra bytes for alignment. See the comments in zstd_cwksp.h
          */
-        assert(ZSTD_cwksp_used(ws) >= neededSpace &&
-               ZSTD_cwksp_used(ws) <= neededSpace + 3);
+        assert(ZSTD_cwksp_used(ws) >= neededSpace && ZSTD_cwksp_used(ws) <= neededSpace + 3);
```
terrelln: Is this 3 extra bytes comment still true?

lib/compress/zstd_cwksp.h
```diff
@@ -198,29 +210,120 @@ MEM_STATIC size_t ZSTD_cwksp_alloc_size(size_t size) {
...
+    return slackSpace;
+}
+
+/**
+ * Return the number of additional bytes required to align a pointer to the given number of bytes
+ */
+MEM_STATIC size_t ZSTD_cwksp_bytes_to_align_ptr(void* ptr, const size_t alignBytes) {
+    size_t const alignBytesMask = alignBytes - 1;
+    size_t const bytes = (alignBytes - ((size_t)ptr & (alignBytesMask))) & alignBytesMask;
+    return bytes;
```
terrelln, line 241: can you add an assert that `alignBytes` is a power of two (`assert((alignBits & alignBytesMask) == 0)`).

lib/compress/zstd_cwksp.h
```diff
@@ -275,6 +378,29 @@ MEM_STATIC void* ZSTD_cwksp_reserve_internal(
...
 }
 
+/* Performs any final internal actions on the wksp before it is ready to go. */
+MEM_STATIC void ZSTD_cwksp_finalize(ZSTD_cwksp* ws) {
+    DEBUGLOG(5, "cwksp: finalizing: %zu table alignment bytes and %zu aligned alignment bytes",
+             ws->tableExtraAlignmentBytes, ws->alignedExtraAlignmentBytes);
+    if (ws->tableExtraAlignmentBytes == ZSTD_CWKSP_ALIGNMENT_BYTES) {
+        /* Due to ASAN fixed-size allocation size increase, we must always
+           perform two allocations that sum to 64 bytes. */
+        ZSTD_cwksp_reserve_internal(ws, ws->tableExtraAlignmentBytes/2, ZSTD_cwksp_alloc_tables);
```
terrelln: Could this be cleaned up by allocating a 0-sized object when `bytesToAlign == 0`?

lib/compress/zstd_compress.c
senhuang42: According to `zstd_cwksp.h` we're actually supposed to allocate all our aligneds before tables. That hadn't been enforced before (tables and aligneds had the same phase), but now that we do distinct stuff before the tables, we have to enforce the ordering.

lib/compress/zstd_compress.c
senhuang42: Nope! I guess we can be exactly precise now.

lib/compress/zstd_cwksp.h
```diff
@@ -198,29 +212,112 @@ MEM_STATIC size_t ZSTD_cwksp_alloc_size(size_t size) {
...
-             * by a larger margin than the space that will be consumed. */
-            /* TODO: cleaner, compiler warning friendly way to do this??? */
-            ws->allocStart = (BYTE*)ws->allocStart - ((size_t)ws->allocStart & (sizeof(U32)-1));
-            if (ws->allocStart < ws->tableValidEnd) {
-                ws->tableValidEnd = ws->allocStart;
+            /* Align the start of the "aligned" to 64 bytes. All aligneds are sized in multiples
+               of 64 bytes, so all subsequent aligneds will naturally be aligned to 64 bytes */
+            void* alloc;
+            size_t bytesToAlign = ZSTD_CWKSP_ALIGNMENT_BYTES - ZSTD_cwksp_bytes_to_align_ptr(ws->allocStart, ZSTD_CWKSP_ALIGNMENT_BYTES);
+            if (bytesToAlign == ZSTD_CWKSP_ALIGNMENT_BYTES) bytesToAlign = 0;
```
terrelln: nit: This would be more simply represented as `(uintptr_t)ws->allocStart & ZSTD_CWSKP_ALIGNMENT_MASK`.

lib/compress/zstd_cwksp.h
```diff
@@ -198,29 +212,112 @@ MEM_STATIC size_t ZSTD_cwksp_alloc_size(size_t size) {
...
+                ws->tableValidEnd = alloc;
+            }
+            ws->allocStart = alloc;
+#if ZSTD_ADDRESS_SANITIZER && !defined (ZSTD_ASAN_DONT_POISON_WORKSPACE)
+            alloc = (BYTE *)alloc + ZSTD_CWKSP_ASAN_REDZONE_SIZE;
+            if (ws->isStatic == ZSTD_cwksp_dynamic_alloc) {
+                __asan_unpoison_memory_region(alloc, bytesToAlign);
+            }
+#endif
+        }
```
terrelln: Why do we need thee ASAN stuff here? We are never actually going to use this memory, so we shouldn't need to unpoison it, or reserve extra redzone space, right?

lib/compress/zstd_cwksp.h
senhuang42: Nice! though iirc `uintptr_t` isn't portable enough for the CI tests.

lib/compress/zstd_cwksp.h
senhuang42: I thought it'd be nice for consistencies sake, though yeah, we aren't really supposed to access this memory - removing this will make the code a lot nicer, and we can refactor it a bit to share the common code.

terrelln (approved):
This looks good to me, but @felixhandte can you take a look?

felixhandte (changes requested):
So, as I understand it, this makes things:

```
[objects][padding][tables->][padding] ... gap? ... [padding][<-aligned][padding][buffers]
```

However, if everything in tables and aligned is a multiple of 64 bytes, which you enforce AFAICT, then you're guaranteed to not need any padding between them. So you could remove 64 bytes of overhead and just do:

```
[objects][padding1][tables->] ... gap? ... [<-aligned][padding2][buffers]
```

And `padding1 + padding2 == 64`.

senhuang42:
> ```
> [objects][padding1][tables->] ... gap? ... [<-aligned][padding2][buffers]
> ```
> 
> And `padding1 + padding2 == 64`.

@felixhandte so as discussed offline, I've changed this to just be padding in front of the tables and aligneds, rather than in front and after. This does widen our estimation error range though, in cases where we reuse a context (or if it's static) and it's too large, i.e. there is a gap between the tables and aligned that we can't fully accurately account for (is there a way to detect cwksp re-use when it hadn't been resized?)

felixhandte (commented):
It doesn't seem necessary to enforce sequencing of aligned -> table allocations. You could apply both alignment allocations when you transition from buffers -> aligned. Maybe you should back that out?

Otherwise this looks good.

senhuang42:
> It doesn't seem necessary to enforce sequencing of aligned -> table allocations. You could apply both alignment allocations when you transition from buffers -> aligned. Maybe you should back that out?

Sure thing - do we want to adjust the comment:
```
* The various types of objects must be allocated in order, so they can be
 * correctly packed into the workspace buffer. That order is:
 *
 * 1. Objects
 * 2. Buffers
 * 3. Aligned
 * 4. Tables
 *
 * Attempts to reserve objects of different types out of order will fail.
```
Or is there some other mechanism that depends on this order somewhere else? The code as-is should be allocate aligned/tables in whatever order (tables also have the "`aligned`" enum phase).

terrelln:
Ping @felixhandte

felixhandte (approved):
Yeah it might be nice to update the comment.

Then other thing we had talked about was applying the stricter used == needed assertion when the workspace is newly allocated. Not necessary though.

Looks good!

Issue #2531, opened 2021-03-11
Add a ZSTD_defaultCLevel() function


ghost:
The stable API has these two functions:
```c
int  ZSTD_minCLevel(void);  /*!< minimum negative compression level allowed */
int  ZSTD_maxCLevel(void);  /*!< maximum compression level available */
```
Does it need a `ZSTD_defaultCLevel()` function that returns the value of `ZSTD_CLEVEL_DEFAULT`?

you-me227:
> `
Duplicate of #`

ghost:
It can be used when dynamically linking to zstd.

This request is not strong, maybe very few people modify the value of `ZSTD_CLEVEL_DEFAULT`.

terrelln:
I think that very few people modify it, but since it is modifiable, and there is no other way to detect the default level, I wouldn't be against adding it.

senhuang42:
Included in #2555

PR #2555, opened 2021-03-24
Add ZSTD_defaultCLevel() function to public API

senhuang42:
Requested in #2531

Simple unit test included. I didn't see much value in changing existing `ZSTD_CLEVEL_DEFAULT` invocations to `ZSTD_defaultCLevel()`, so existing code is basically left as-is.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -1918,6 +1918,10 @@ ZSTDLIB_API size_t ZSTD_compressStream2_simpleArgs (
                       const void* src, size_t srcSize, size_t* srcPos,
                             ZSTD_EndDirective endOp);
 
+/*! ZSTD_defaultCLevel() :
+ * Returns the default compression level, specified by ZSTD_CLEVEL_DEFAULT
+ */
+ZSTDLIB_API int ZSTD_defaultCLevel(void);
 
 /***************************************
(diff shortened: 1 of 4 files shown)


lib/zstd.h
```diff
@@ -182,6 +182,7 @@ ZSTDLIB_API unsigned    ZSTD_isError(size_t code);          /*!< tells if a `siz
 ZSTDLIB_API const char* ZSTD_getErrorName(size_t code);     /*!< provides readable string from an error code */
 ZSTDLIB_API int         ZSTD_minCLevel(void);               /*!< minimum negative compression level allowed */
 ZSTDLIB_API int         ZSTD_maxCLevel(void);               /*!< maximum compression level available */
+ZSTDLIB_API int         ZSTD_defaultCLevel(void);           /*!< default compression level, specified by ZSTD_CLEVEL_DEFAULT */
```
Cyan4973: Note : this is "stable" API domain.

However simple this symbol might be, its introduction within "stable" implies `v1.5.0`.

senhuang42: Ah okay, I'll move it to experimental and we can put this on the list of stuff to move to stable for `1.5.0`

Cyan4973 (approved):

PR #2557, opened 2021-03-25
Add msan fuzztest to github actions

senhuang42:
We need an msan fuzztest - block splitter PR actually failed this, but it went undetected since our CI no longer does an msan fuzztest.

Is there any way to force `generic-dev` to run on this PR? Or must there be changes to the actual code for that to happen.


senhuang42:
going to just put this commit on the other github msan PR

Cyan4973:
`msan` fuzztests used to fail, due to some issues on ossfuzz side.

I recently noticed a message from ossfuzz stating that the issue was fixed (it was storage related).
So hopefully, we can re-enable them again now.

PR #2559, opened 2021-03-25
Add different dict modes to compression ratio regression test, update results.csv

senhuang42:
On the row-hash PR, DMS was silently broken until recently fixed/discovered - our coverage of tests on different dict modes (DMS, DDS, Copy, Load) is not particularly high. This should be a good first step in ensuring we detect such things. (The row-hash PR adds a DMS unit test in fuzzer.c as well, to check for regressions)

This PR:
- Adds ratio regression tests to clevels >= 1 for each dictMode when testing using advanced API that can honor requested parameters.
- Updates `results.csv`
   - New rows with new dictmode tests
  - Adjustments for improvements introduced with block splitter and levels 16, 19
  - Minor spelling fix
- Fixes compiler warnings when building this regression test

Testing: I tested this on row-hash with a broken DMS, and indeed the compression results were bad.


tests/regression/config.c
```diff
@@ -28,20 +28,68 @@
...
+        .param_values = PARAM_VALUES(level_##x##_param_values_dictcopy),          \
+        .use_dictionary = 1,                                                      \
+        .advanced_api_only = 1,                                                   \
+    };                                                                            \
+    config_t const level_##x##_dict_load = {                                      \
+        .name = "level " #x " with dict load",                                    \
+        .cli_args = "-" #x,                                                       \
+        .param_values = PARAM_VALUES(level_##x##_param_values_dictload),          \
+        .use_dictionary = 1,                                                      \
+        .advanced_api_only = 1,                                                   \
```
senhuang42, line 92: So I added the new configs to `LEVEL()` so we automatically get testing for all the levels we try.

An alternative approach is to have each of these just be their own config, with a fixed compression level (and if we wanted to test more clevels/strategies, we could just add more configs). But that does have the downside of not automatically hitting all the tested clevels. 

I do prefer having these run on all the compression levels, in case in the future, strategies get changed in a way that might affect the dictionary strategies.

Cyan4973: makes sense

Cyan4973 (approved):

terrelln (approved):

terrelln:
Thanks for adding this Sen! Really glad to see increased coverage of dictionary compression.

PR #2560, opened 2021-03-25
Fix repcode-related OSS-fuzz issues in block splitter

senhuang42:
OSS-Fuzz found an edge case with the block splitter: we shouldn't emit the last partition of a split block as uncompressed in any case, since we can't determine if the next 128K block has a repcode.

We already handle this for the next partitions (i.e. don't emit uncompressed if the next partition has repcode within the first three entries), but missed the case of a repcode in the first three seqs of the next "complete" 128K block. The fuzzer found this pretty quickly, so good job oss-fuzz :)

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2793,11 +2793,18 @@ static int ZSTD_maybeRLE(seqStore_t const* seqStore)
     return nbSeqs < 4 && nbLits < 10;
 }
 
-static void ZSTD_confirmRepcodesAndEntropyTables(ZSTD_CCtx* zc)
+static void ZSTD_blockState_confirmRepcodesAndEntropyTables(ZSTD_blockState_t* const bs)
 {
-    ZSTD_compressedBlockState_t* const tmp = zc->blockState.prevCBlock;
-    zc->blockState.prevCBlock = zc->blockState.nextCBlock;
-    zc->blockState.nextCBlock = tmp;
+    ZSTD_compressedBlockState_t* const tmp = bs->prevCBlock;
+    bs->prevCBlock = bs->nextCBlock;
+    bs->nextCBlock = tmp;
+}
+
+static void ZSTD_blockState_confirmEntropyTables(ZSTD_blockState_t* const bs)
+{
+    ZSTD_entropyCTables_t const tmp = bs->prevCBlock->entropy;
+    bs->prevCBlock->entropy = bs->nextCBlock->entropy;
+    bs->nextCBlock->entropy = tmp;
 }
 
 /* Writes the block header */
@@ -3216,7 +3223,7 @@ static size_t ZSTD_compressSeqStore_singleBlock(ZSTD_CCtx* zc, seqStore_t* seqSt
 
     if (zc->seqCollector.collectSequences) {
         ZSTD_copyBlockSequences(zc);
-        ZSTD_confirmRepcodesAndEntropyTables(zc);
+        ZSTD_blockState_confirmRepcodesAndEntropyTables(&zc->blockState);
         return 0;
     }
 
@@ -3234,11 +3241,9 @@ static size_t ZSTD_compressSeqStore_singleBlock(ZSTD_CCtx* zc, seqStore_t* seqSt
     } else {
         /* Always update entropy, only update repcodes when compressing entire block or last partition */
         if (updateRepcodes) {
-            ZSTD_confirmRepcodesAndEntropyTables(zc);
+            ZSTD_blockState_confirmRepcodesAndEntropyTables(&zc->blockState);
         } else {
-            ZSTD_entropyCTables_t tmp = zc->blockState.prevCBlock->entropy;
-            zc->blockState.prevCBlock->entropy = zc->blockState.nextCBlock->entropy;
-            zc->blockState.nextCBlock->entropy = tmp;
+            ZSTD_blockState_confirmEntropyTables(&zc->blockState);
         }
         writeBlockHeader(op, cSeqsSize, srcSize, lastBlock);
         cSize = ZSTD_blockHeaderSize + cSeqsSize;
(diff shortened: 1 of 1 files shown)


Cyan4973:
If I do understand properly, when a block is split, some of its "partitions" (== sub-blocks) can be emitted uncompressed, likely as a post-processing decision, once it's clear that sending them compressed is unfavorable.

Problem is, if next partition uses a repeat code within its first 3 offset codes, it probably refers to one of the cancelled offsets of previous partition, now sent uncompressed, and therefore these offset codes are no longer valid.

To avoid that problem, the code _forces_ the previous partition to be sent _compressed_, so that the offset history remains unaltered, so that the following repeat codes remain valid.

This is fine as long as all partitions are part of the same original block (128 KB), because it's a condition which can be controlled. However, if it's the last partition, the next one will be part of the next big block, which stats are not yet known. If it uses a repeat code, and the last partition was sent uncompressed, it now refers to a non-existing offset. To avoid this issue, force the last partition to be sent compressed, always, so that last 3 offsets are always well defined.

This inspires the following questions : 

- It seems that the offset history generated by the match finder on current full block is sent verbatim to next block, no matter what. 
- Another possibility could have been to send the offset history of _latest compressed partition_. This way, the last partition(s) could be emitted uncompressed, there is still a valid offset history to provide to next block.
- However this would require to _know_ the offset history at the boundaries between partitions.
    + => I presume this is the issue, the offset history is not known at the boundary between partitions.
    + I also presume it could be rebuilt, when needed, since the offsets are known and stored in the `seqStore`. However, if this list contains repeat codes, it becomes more complex (the translation from rep-code to actual offset codes is not tracked within the `seqStore`).
- Similarly, when a partition has a repeat code within its first 3 offsets codes, it relies on the existence of a repeat offset history, which would be altered if previous partition was sent uncompressed. 
- Another solution would be to translate these repeat codes into actual offset values, so that they are no longer dependent on previous offset history.
     + => I presume that the issue is that the translation from repeat code into actual offset values is not known (not stored into `seqStore`).

senhuang42:
> I also presume it could be rebuilt, when needed, since the offsets are known and stored in the seqStore. However, if this list contains repeat codes, it becomes more complex (the translation from rep-code to actual offset codes is not tracked within the seqStore).

True, this is something worth investigating, since it could improve compression ratio for these particular cases, and would be a more elegant solution than emitting an compressed block even when not advantageous to do so. I think I'll add a follow-up PR since the recomputation doesn't seem trivial. And yeah, the crux of the issue is that we don't know the offset history at any given split point, which means we'd have to recompute.

The other OSS-fuzz failure is also related to repcodes. I have been investigating for a bit, and realized that in general, these were not handled correctly to begin with. The block splitter PR had been using `nextCBlock->rep` as sort of temporary storage for the next repcodes for the current 128K block. Hence the `ZSTD_memcpy()` call to copy repcodes from `prevCBlock` to `nextCBlock` after each block. This works because we call `ZSTD_confirmRepcodesAndEntropyTables()` at each partition (though we also do unnecessary copies). But this is wrong if we have a no-compress block, since then we do _not_ call `ZSTD_confirmRepcodesAndEntropyTables()` but still do the `ZSTD_memcpy()`. The one-liner fix was to only call the `ZSTD_memcpy()` on compressed blocks, but in general this is a bit janky, so I've gone ahead and refactored (and fixed) this behavior for clarity. 

The two new commits do the following:
- Only update repcodes when we should: at the last partition, or if we decide not to split and compress in one go.
  - Note that we should not update repcodes if we emit a set of partitions that exceed 128K and re-compress, since we already would have updated the repcodes in the last partition.
  - This fixes the newer OSS-Fuzz bug
- Changes internal function signature of `ZSTD_confirmRepcodesAndEntropyTables()` to take in a `ZSTD_blockState_t*` instead of an entire `ZSTD_CCtx*`, which seems like way too big of a scope for a function that is basically just doing a swap.

senhuang42:
Updated again with repcode history update/tracking as we compress each individual partition. This means we no longer need to be able to force emit compressed blocks, which means now it's no longer possible to emit a compressed block > 128K, so we no longer need the fallback either.

lib/compress/zstd_compress.c
```diff
@@ -2793,11 +2791,11 @@ static int ZSTD_maybeRLE(seqStore_t const* seqStore)
     return nbSeqs < 4 && nbLits < 10;
 }
 
-static void ZSTD_confirmRepcodesAndEntropyTables(ZSTD_CCtx* zc)
+static void ZSTD_blockState_confirmRepcodesAndEntropyTables(ZSTD_blockState_t* const bs)
```
Cyan4973, line 2794: 👍 
Good call ! more accurate object parameter == better clarity

lib/compress/zstd_compress.c
```diff
@@ -3180,6 +3178,26 @@ static void ZSTD_deriveSeqStoreChunk(seqStore_t* resultSeqStore,
     resultSeqStore->ofCode += startIdx;
 }
 
+/**
+ * ZSTD_seqStore_updateRepcodes(): Starting from an array of initial repcodes and a seqStore,
```
Cyan4973, line 3182: 👍 
I like the way the logic is well encapsulated here

lib/compress/zstd_compress.c
```diff
@@ -3215,24 +3232,30 @@ static size_t ZSTD_compressSeqStore_singleBlock(ZSTD_CCtx* zc, seqStore_t* seqSt
...
-    } else if (cSeqsSize == 1 && canEmitRLEorNoCompress) {
+    } else if (cSeqsSize == 1) {
         cSize = ZSTD_rleCompressBlock(op, dstCapacity, *ip, srcSize, lastBlock);
         FORWARD_IF_ERROR(cSize, "RLE compress block failed");
         DEBUGLOG(4, "Writing out RLE block, size: %zu", cSize);
     } else {
-        /* Error checking and repcodes update */
-        ZSTD_confirmRepcodesAndEntropyTables(zc);
+        if (isPartition) {
+            /* We manually update repcodes if we are currently compressing a partition. Otherwise,
```
Cyan4973, line 3252: 👍 
Great comment, justifying the "why"

Cyan4973 (approved):
Great PR ! The updated logic is _way_ better

PR #2561, opened 2021-03-26
Add enum for representing long length ID

senhuang42:
As title.

This has been a perpetual minor annoyance whenever I look at the `longLengthID` and have to pause a minute to make sure that the numbers correspond to what I think they do.


Cyan4973 (approved):

Cyan4973:
Yeah, it definitely helps clarity

PR #2562, opened 2021-03-28
Stop complaining about hash tool not found

niclasr:
If build_dir is set the zstd build complains about md5sum not being found.
Fix this by checking if build_dir is set before checking and using the hash tool
just like in lib/Makefile .


terrelln:
Thanks for the PR @niclasr!

PR #2569, opened 2021-04-05
Replace invalid repcodes during block splitting

senhuang42:
OSS-Fuzz has found more edge cases with the block splitter and repcodes.

The previous patch #2560  added the capability to properly emit no-compress blocks, but we now run into the next issue of cases where we are forced to replace repcodes in the cases where they would be invalid.

But, it's not just as simple as doing a raw replacement of the repcode, since this can introduce issues later on where the repcode history becomes invalid as a result of changing one of the `seqStore`'s offsets from a repcode to an offset - maintaining a single set of repcodes is not sufficient - we must know on the compression side exactly what the decompression offset history is to be able to make correct substitutions.

Quick example: https://pastebin.com/641BkQRK

The code/comments explains the approach to resolving the issue. I believe this patch should comprehensively address any issues with repcodes + block splitter.

Test Plan:
- Check that this fixes the OSS-fuzz crash samples (and the ones from #2560 are still OK).

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3181,23 +3181,40 @@ static void ZSTD_deriveSeqStoreChunk(seqStore_t* resultSeqStore,
 }
 
 /**
- * ZSTD_seqStore_updateRepcodes(): Starting from an array of initial repcodes and a seqStore,
- * construct the final repcodes at the conclusion of compressing the seqStore, stored in dstRep.
+ * ZSTD_seqStore_resolveOffCodes() reconciles any possible divergences in offset history that may arise
+ * due to emission of RLE/raw blocks that disturb the offset history, and replaces any repcodes within
+ * the seqStore that may be invalid.
+ * 
+ * dRepcodes are updated as would be on the decompression side. cRepcodes are updated exactly in
+ * accordance with the seqStore.
  */
-static void ZSTD_seqStore_updateRepcodes(U32 dstRep[ZSTD_REP_NUM],
-                                         const U32 initialRep[ZSTD_REP_NUM],
-                                         const seqStore_t* const seqStore, U32 const nbSeq) {
-    repcodes_t updatedRepcodes;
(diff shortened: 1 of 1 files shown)


Cyan4973:
Is there a speed impact with this new method ?

senhuang42:
> Is there a speed impact with this new method ?

Probably not, we basically add a branch to the loop over each `seqStore_t` and a few small `memcpy()` calls per block.

Benchmarks seem to confirm this:
```
blocksplitter_parallel_repcodes
1#silesia.tar       : 211957760 ->  72914525 (2.907), 160.4 MB/s ,1138.9 MB/s
16#silesia.tar       : 211957760 ->  55350330 (3.829),  5.79 MB/s ,1091.4 MB/s 
17#silesia.tar       : 211957760 ->  54292138 (3.904),  4.58 MB/s ,1064.9 MB/s 
18#silesia.tar       : 211957760 ->  53419772 (3.968),  3.65 MB/s ,1029.1 MB/s 
19#silesia.tar       : 211957760 ->  52993627 (4.000),  3.03 MB/s ,1012.1 MB/s
```

```
dev
 1#silesia.tar       : 211957760 ->  72914525 (2.907), 159.1 MB/s ,1140.3 MB/s
16#silesia.tar       : 211957760 ->  55350330 (3.829),  5.76 MB/s ,1091.1 MB/s 
17#silesia.tar       : 211957760 ->  54292138 (3.904),  4.58 MB/s ,1054.2 MB/s 
18#silesia.tar       : 211957760 ->  53419772 (3.968),  3.66 MB/s ,1019.0 MB/s 
19#silesia.tar       : 211957760 ->  52993627 (4.000),  3.03 MB/s ,1003.9 MB/s
```

Cyan4973 (approved):
This looks pretty good.
The logic is simple, well explained, and general (it takes care of the entire discrepancy issue, not one corner case at a time).
I'm also impressed (and surprised) that its speed cost seems negligible.

PR #2570, opened 2021-04-05
Fix dictionary force reloading clevel selection

senhuang42:
As title - reloading the dictionary with `ZSTD_dictForceLoad` was basically broken because we would always choose the default compression level since `initLocalDict()` always sets a cdict. Now, we only override cctx clevel with cdict's clevel if the cctx had the cdict to begin with.

`results.csv` updated accordingly.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -5164,11 +5164,11 @@ static size_t ZSTD_CCtx_init_compressStream2(ZSTD_CCtx* cctx,
                                              size_t inSize) {
     ZSTD_CCtx_params params = cctx->requestedParams;
     ZSTD_prefixDict const prefixDict = cctx->prefixDict;
+    if (cctx->cdict)
+        params.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
     FORWARD_IF_ERROR( ZSTD_initLocalDict(cctx) , ""); /* Init the local dict if present. */
     ZSTD_memset(&cctx->prefixDict, 0, sizeof(cctx->prefixDict));   /* single usage */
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

senhuang42:
Investigating the surprising non determinism failures... these are locally reproducible too. I wonder if this is exposing an existing bug or is the cause itself.

Edit: Seems to be because subsequent calls of `compressStream2()` that have a `localDict` would appear to have a dedicated `cdict`, and therefore we'd end up taking that cdict's level when we shouldn't. So, we should check that there is no `localDict.cdict` as well.

lib/compress/zstd_compress.c
```diff
@@ -5164,11 +5164,11 @@ static size_t ZSTD_CCtx_init_compressStream2(ZSTD_CCtx* cctx,
                                              size_t inSize) {
     ZSTD_CCtx_params params = cctx->requestedParams;
     ZSTD_prefixDict const prefixDict = cctx->prefixDict;
+    if (cctx->cdict && !cctx->localDict.cdict)
+        params.compressionLevel = cctx->cdict->compressionLevel; /* let cdict take priority in terms of compression level */
```
terrelln, line 5168: @senhuang42 I think this is correct, but it is really confusing to have moved this above `ZSTD_initLocalDict()`. When `ZSTD_initLocalDict()` is creates a cdict, this will never run, because necessarily `cctx->cdict == nullptr` in that case. Since `ZSTD_CCtx_loadDictionary()` first calls `ZSTD_clearAllDicts()`.

I believe that leaving this line in the original location (L5170), with the additional `!cctx->localDict.cdict` check will behave identically, but is much clearer.

PR #2571, opened 2021-04-06
Move clevel override to after initLocalDict()

senhuang42:
As @terrelln noted, the check for overriding the clevel using cdict below `ZSTD_initLocalDict()` makes the logic here clearer - the behavior should be identical (basically, don't use cdict's clevel if this dictionary comes from `loadDictionary()`).

Also added a comment to explain this behavior more clearly.


Cyan4973 (approved):

terrelln (approved):
Thanks!

PR #2572, opened 2021-04-09
Adjust nb elements to prefetch in ZSTD_row_fillHashCache()

senhuang42:
I noticed an off-by-one bug in row hash. I am curious to see if OSS-fuzz will catch this.

`ZSTD_row_fillHashCache()` gets called at the beginning of each call to `ZSTD_compressBlock_lazy_generic()`.
We set `iLimit` to `iend-16` to leave enough space for `ZSTD_hashPtr()` call (8 bytes) and 8 entries to prefetch for the row matchfinder's hash cache.

Of note is the main `ZSTD_compressBlock_lazy_generic()` loop which is conditioned on `while (ip < iLimit)`. However, while within the loop, if we do extra searches (lazy, lazy2), we actually may increment `ip`, so actually, `ip` may equal `iLimit` at some point. If we don't include the byte at `iLimit` in the cache, then upon trying to fetch the corresponding hash from the hashcache, we may fetch some garbage hash value since we don't zero the `ms->hashCache` on each new compression. I don't think that zeroing it would be necessary to do if the cache is behaving correctly, but it might be an additional, cheap precaution worth considering to avoid segfaults.

The following situation is where the segfault occurred (we set everything relative to ip, let `ip = 0`):
```
Example 1:
ip = 0
iend = 23
iLimit = 23 - 16 == 7
base + nextToUpdate == ip == 0
ZSTD_row_fillHashCache():
	maxElemsToPrefetch = (iLimit - (base+nextToUpdate)) ==  (7 - 0) == 7
        // (WRONG: we should prefetch 8 elements: 0, 1, 2, 3, 4, 5, 6, 7), since iLimit == 7 might be searched from
```

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1136,16 +1136,17 @@ FORCE_INLINE_TEMPLATE void ZSTD_row_prefetch(U32 const* hashTable, U16 const* ta
 }
 
 /* ZSTD_row_fillHashCache():
- * Fill up the hash cache starting at idx, prefetching ZSTD_ROW_HASH_CACHE_SIZE entries.
+ * Fill up the hash cache starting at idx, prefetching up to ZSTD_ROW_HASH_CACHE_SIZE entries,
+ * but not beyond iLimit.
  */
 static void ZSTD_row_fillHashCache(ZSTD_matchState_t* ms, const BYTE* base,
                                    U32 const rowLog, U32 const mls,
-                                   U32 idx, const BYTE* const iend)
+                                   U32 idx, const BYTE* const iLimit)
 {
     U32 const* const hashTable = ms->hashTable;
     U16 const* const tagTable = ms->tagTable;
     U32 const hashLog = ms->rowHashLog;
-    U32 const maxElemsToPrefetch = (base + idx) >= iend ? 0 : (U32)(iend - (base + idx));
+    U32 const maxElemsToPrefetch = (base + idx) > iLimit ? 0 : (U32)(iLimit - (base + idx) + 1);
     U32 const lim = idx + MIN(ZSTD_ROW_HASH_CACHE_SIZE, maxElemsToPrefetch);
 
     for (; idx < lim; ++idx) {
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_lazy.c
```diff
@@ -1140,12 +1140,12 @@ FORCE_INLINE_TEMPLATE void ZSTD_row_prefetch(U32 const* hashTable, U16 const* ta
  */
```
Cyan4973, line 1141: The relation between `iLimit` (end of prefetching range) and `iend` (end of input buffer, not present in the function's interface) could be explained in code comment, especially if it justifies the way `iLimit` is used in the code below.

Cyan4973 (approved):

Cyan4973:
> We set iLimit to iend-16 to leave enough space for ZSTD_hashPtr() call (8 bytes) and 8 entries to prefetch for the row matchfinder's hash cache.

That's a part I may not understand.
The code seems to limit prefetching at `iLimit`, which then requires 8 bytes of input to (safely) produce the hash.
So why `iend - 16` ?

senhuang42:
> So why `iend - 16` ?

In general, we return hashes at `ip` from `ZSTD_row_nextCachedHash()`. In this function, we always return hashes from within the hash cache, then hash the position at `ZSTD_ROW_HASH_CACHE_SIZE == 8` bytes away, and store that hash in the hash cache.

So as a result, if we call `ZSTD_row_nextCachedHash()` at `ip`, we need to also hash the position `ip+8`, which means we add another 8 bytes to safely hash it, leaving us needing `ip+16` at least to be still within the input buffer.

https://github.com/facebook/zstd/blob/56421f34e4ab5c17cbac0832bd12413e4f27ecf9/lib/compress/zstd_lazy.c#L1167-L1179

So, when we prefill the hash cache at the beginning of each block, we don't need to go beyond `iend - 16`.

I will add a comment clarifying this in `ZSTD_lazy_compressBlock_generic()`
Edit: I've changed `iend - 16` to `iend - 8 - ZSTD_ROW_HASH_CACHE_SIZE` for clarity. That should make implicit the reason we need an earlier `ilimit` without needing a comment.

PR #2581, opened 2021-04-16
[1.5.0] Promote ZSTD_c_literalCompressionMode to stable params

senhuang42:
Promotes `ZSTD_c_literalCompressionMode` to stable, removes it from the experimental section, and adjusts the rest of the `experimentalParams` numberings accordingly.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -266,6 +266,14 @@ typedef enum { ZSTD_fast=1,
                          Only the order (from fast to strong) is guaranteed */
 } ZSTD_strategy;
 
+typedef enum {
+  ZSTD_lcm_auto = 0,          /**< Automatically determine the compression mode based on the compression level.
+                               *   Negative compression levels will be uncompressed, and positive compression
+                               *   levels will be compressed. */
+  ZSTD_lcm_huffman = 1,       /**< Always attempt Huffman compression. Uncompressed literals will still be
+                               *   emitted if Huffman compression is not profitable. */
+  ZSTD_lcm_uncompressed = 2   /**< Always emit uncompressed literals. */
+} ZSTD_literalCompressionMode_e;
 
 typedef enum {
 
@@ -332,6 +340,10 @@ typedef enum {
                               * The higher the value of selected strategy, the more complex it is,
                               * resulting in stronger and slower compression.
                               * Special: value 0 means "use default strategy". */
+    ZSTD_c_literalCompressionMode=108,  /* Controls how the literals are compressed (default is auto).
+                                         * The value must be of type ZSTD_literalCompressionMode_e.
+                                         * See ZSTD_literalCompressionMode_e enum definition for details.
+                                         */
 
     /* LDM mode parameters */
     ZSTD_c_enableLongDistanceMatching=160, /* Enable long distance matching.
@@ -411,7 +423,6 @@ typedef enum {
      * ZSTD_c_format
      * ZSTD_c_forceMaxWindow
      * ZSTD_c_forceAttachDict
-     * ZSTD_c_literalCompressionMode
      * ZSTD_c_targetCBlockSize
      * ZSTD_c_srcSizeHint
      * ZSTD_c_enableDedicatedDictSearch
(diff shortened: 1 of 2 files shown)


lib/zstd.h
```diff
@@ -429,16 +432,15 @@ typedef enum {
...
-     ZSTD_c_experimentalParam6=1003,
-     ZSTD_c_experimentalParam7=1004,
-     ZSTD_c_experimentalParam8=1005,
-     ZSTD_c_experimentalParam9=1006,
-     ZSTD_c_experimentalParam10=1007,
-     ZSTD_c_experimentalParam11=1008,
-     ZSTD_c_experimentalParam12=1009,
-     ZSTD_c_experimentalParam13=1010,
-     ZSTD_c_experimentalParam14=1011
+     ZSTD_c_experimentalParam5=1003,
```
Cyan4973: Changing the `enum` nb of a parameter _might_ lead to some downstream issues.

Some binders in other languages, such as Java,
may not access the `enum` definition, and use the known integer value directly.
In which case, changing the meaning of this integer can lead to downstream surprises.

Of course, this is a very niche scenario, so I don't anticipate a lot of real-life issues,
but we also have nothing to gain from introducing this disruption.

So I would keep the other numbers "as is", and accept holes in the number scheme.

_edit_ : reading about it again,
it seems you kept the integer value,
but changed instead the associated intermediate enum.
This is milder, since the above scenario (using integer value directly) is properly covered.
Yet, I fail to see a benefit from doing this update of intermediate enum. Is there one ?

senhuang42: It was more just a visual decision, so that the whole range of `ZSTD_c_experimentalParam[1-n]` is represented. But, I can see why it might just be preferable to delete it, since it shouldn't really matter to anyone `experimentalParam*` wouldn't be consecutive.

ghost:
Just take this opportunity to make a suggestion.
It's better not to use the same values in `ZSTD_cParameter` and `ZSTD_dParameter`.
For example, these two values are both `100`:
```c
typedef enum {
    ZSTD_c_compressionLevel=100,
    ...
} ZSTD_cParameter;

typedef enum {
    ZSTD_d_windowLogMax=100, 
    ...
} ZSTD_dParameter;
```
If an user incorrectly sets the decompression parameter to a compression context, it's very hard to find such bug.

In addition, please create a meta issue for v1.5.0, then the users can know changes and feedback earlier.

terrelln:
> If an user incorrectly sets the decompression parameter to a compression context, it's very hard to find such bug.

We can try to avoid it in the future, but we cannot change the values which are already stable. This will be caught in C++ source code, because an implicit conversion from `ZSTD_cParameter` to `ZSTD_dParameters` is disallowed. Additionally, passing `-Werror=enum-conversion` to clang/gcc will also raise an error for this bug.

lib/zstd.h
```diff
@@ -332,6 +332,10 @@ typedef enum {
                               * The higher the value of selected strategy, the more complex it is,
                               * resulting in stronger and slower compression.
                               * Special: value 0 means "use default strategy". */
+    ZSTD_c_literalCompressionMode=108,  /* Controls how the literals are compressed (default is auto).
+                                         * The value must be of type ZSTD_literalCompressionMode_e.
```
Cyan4973, line 344: The documentation states that `The value must be of type ZSTD_literalCompressionMode_e`.

This implies that `ZSTD_literalCompressionMode_e` must also be part of "stable" public API.

lib/zstd.h
```diff
@@ -332,6 +332,10 @@ typedef enum {
...
+                                         * See ZSTD_literalCompressionMode_t enum definition for details.
```
Cyan4973: Likely a typo:
`ZSTD_literalCompressionMode_t` => `ZSTD_literalCompressionMode_e`

ghost:
> This will be caught in C++ source code, because an implicit conversion from `ZSTD_cParameter` to `ZSTD_dParameters` is disallowed. Additionally, passing `-Werror=enum-conversion` to clang/gcc will also raise an error for this bug.

For some language bindings, this may be a problem, especially for dynamic languages.
On the whole, it's not a big problem, just a suggestion for the future.

Cyan4973 (approved):

PR #2582, opened 2021-04-16
[1.5.0] Deprecate some functions

senhuang42:
We mark some functions as deprecated for the upcoming 1.5.0 release:

Functions that take `ZSTD_parameters` as an argument:
`ZSTD_compress_advanced()`
`ZSTD_compress_usingCDict_advanced()`
`ZSTD_initCStream_advanced()`
`ZSTD_initCStream_usingCDict_advanced()`
`ZSTD_compressBegin_advanced()`
`ZSTD_compressBegin_usingCDict_advanced()`

Functions that are redundant:
`ZSTD_initCStream_srcSize()`
`ZSTD_initCStream_usingDict()`
`ZSTD_initCStream_usingCDict()`
`ZSTD_resetCStream()`
`ZSTD_DCtx_setFormat()`

In addition, we modify some build scripts (Make, CMake, VS) for `fuzzer`, `fullbench`, and `zstreamtest` to ignore deprecated declarations since we still explicitly test some of these functions.

In situations where it is appropriate, we also convert the deprecated API into the current API.

Some remaining `zbuff` related build targets/paths that were missed in #2583 get removed too.


senhuang42:
I've added `-Wno-deprecated-declarations` to the linux kernel test. I notice that `zstd_compress_module.c` doesn't really use any of the new `compress2`-type advanced API, and actually the wrapper functions translate pretty directly to this API that we're deprecating with their usage of `ZSTD_parameters`. 

So I'm not entirely sure whether it would be better to update the internals of these wrapper functions or just let them keep using this API we're deprecating? cc @terrelln

Cyan4973:
> Question: Should we also deprecate ZSTD_initCStream()?

Nope, no need.

Cyan4973:
> We update docs on setFormat() to mention that it is redundant, but don't generate a compiler warning for this one, since it is a stable API entrypoint that can modify an advanced API param.

As far as I know, `ZSTD_DCtx_setFormat()` is ___not___ part of "stable".

lib/zstd.h
```diff
@@ -35,6 +35,28 @@ extern "C" {
...
+#  elif defined(__GNUC__) && (__GNUC__ >= 3)
+#    define ZSTD_DEPRECATED(message) ZSTDLIB_API __attribute__((deprecated))
+#  elif defined(_MSC_VER)
+#    define ZSTD_DEPRECATED(message) ZSTDLIB_API __declspec(deprecated(message))
+#  else
+#    pragma message("WARNING: You need to implement ZSTD_DEPRECATED for this compiler")
+#    define ZSTD_DEPRECATED(message) ZSTDLIB_API
+#  endif
+#endif /* ZSTD_DISABLE_DEPRECATE_WARNINGS */
+
```
terrelln: nit: Can we move this to the top of the `ZSTD_STATIC_LINKING_ONLY` section?

I think that helps convey the message that we only deprecate functions in the unstable portion of our API. People who only use the stable section don't need to think about deprecation.

Cyan4973 (approved):

PR #2584, opened 2021-04-20
[1.5.0] Enable multithreading in lib build by default

senhuang42:
With 1.5.0, the default `lib` build now will be multithreaded.

- Force enable multithreading on both dynamic and static libraries by appending `-mt` to the target, e.g. `make lib-mt`.
- Force disable multithreading on both dynamic and static libraries by appending `-nomt` to the target, e.g. `make lib-nomt`.
- By default, as mentioned before, dynamic library is multithreaded, and static library is single-threaded, e.g. `make lib`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,8 @@ all32:
 	$(MAKE) -C $(PRGDIR) zstd32
 	$(MAKE) -C $(TESTDIR) all32
 
-.PHONY: lib lib-release
-lib lib-release :
+.PHONY: lib lib-release lib-mt lib-nomt
+lib lib-release lib-mt lib-nomt:
 	$(Q)$(MAKE) -C $(ZSTDDIR) $@
 
 .PHONY: zstd zstd-release
@@ -151,7 +151,6 @@ clean:
 ifneq (,$(filter $(shell uname),Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD DragonFly NetBSD MSYS_NT Haiku))
 
 HOST_OS = POSIX
-CMAKE_PARAMS = -DZSTD_BUILD_CONTRIB:BOOL=ON -DZSTD_BUILD_STATIC:BOOL=ON -DZSTD_BUILD_TESTS:BOOL=ON -DZSTD_ZLIB_SUPPORT:BOOL=ON -DZSTD_LZMA_SUPPORT:BOOL=ON -DCMAKE_BUILD_TYPE=Release
 
 HAVE_COLORNEVER = $(shell echo a | egrep --color=never a > /dev/null 2> /dev/null && echo 1 || echo 0)
 EGREP_OPTIONS ?=
@@ -357,11 +356,12 @@ lz4install:
 endif
 
 
+CMAKE_PARAMS = -DZSTD_BUILD_CONTRIB:BOOL=ON -DZSTD_BUILD_STATIC:BOOL=ON -DZSTD_BUILD_TESTS:BOOL=ON -DZSTD_ZLIB_SUPPORT:BOOL=ON -DZSTD_LZMA_SUPPORT:BOOL=ON -DCMAKE_BUILD_TYPE=Release
+
 ifneq (,$(filter MSYS%,$(shell uname)))
 HOST_OS = MSYS
 CMAKE_PARAMS = -G"MSYS Makefiles" -DCMAKE_BUILD_TYPE=Debug -DZSTD_MULTITHREAD_SUPPORT:BOOL=OFF -DZSTD_BUILD_STATIC:BOOL=ON -DZSTD_BUILD_TESTS:BOOL=ON
 endif
-
 
 #------------------------------------------------------------------------
 # target specific tests
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -8,6 +8,9 @@
 # You may select, at your option, one of the above-listed licenses.
 # ################################################################
 
+# Note: by default, the static library is built single-threaded and dynamic library is built
+# multi-threaded. It is possible to force multi or single threaded builds by appending
+# -mt or -nomt to the build target (like lib-mt for multi-threaded, lib-nomt for single-threaded).
 .PHONY: default
 default: lib-release
 
@@ -67,6 +70,10 @@ DEBUGFLAGS= -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
             -Wredundant-decls -Wmissing-prototypes -Wc++-compat
 CFLAGS   += $(DEBUGFLAGS) $(MOREFLAGS)
 FLAGS     = $(CPPFLAGS) $(CFLAGS)
+
+CPPFLAGS_DYNLIB  = -DZSTD_MULTITHREAD # dynamic library build defaults to multi-threaded
+LDFLAGS_DYNLIB   = -pthread
+CPPFLAGS_STATLIB =                    # static library build defaults to single-threaded
 
 HAVE_COLORNEVER = $(shell echo a | grep --color=never a > /dev/null 2> /dev/null && echo 1 || echo 0)
 GREP_OPTIONS ?=
@@ -222,6 +229,7 @@ all: lib
 
 
 .PHONY: libzstd.a  # must be run every time
+libzstd.a: CPPFLAGS += $(CPPFLAGS_STATLIB)
 
 ifndef BUILD_DIR
 # determine BUILD_DIR from compilation flags
@@ -238,7 +246,10 @@ ZSTD_STATLIB_OBJ := $(addprefix $(ZSTD_STATLIB_DIR)/,$(ZSTD_LOCAL_OBJ))
 $(ZSTD_STATLIB): ARFLAGS = rcs
 $(ZSTD_STATLIB): | $(ZSTD_STATLIB_DIR)
 $(ZSTD_STATLIB): $(ZSTD_STATLIB_OBJ)
-	@echo compiling static library
+  # Check for multithread flag at target execution time
+	$(if $(filter -DZSTD_MULTITHREAD,$(CPPFLAGS)),\
+    @echo compiling multi-threaded static library $(LIBVER),\
+    @echo compiling single-threaded static library $(LIBVER))
 	$(AR) $(ARFLAGS) $@ $^
 
 libzstd.a: $(ZSTD_STATLIB)
@@ -257,8 +268,9 @@ else  # not Windows
 
 LIBZSTD = libzstd.$(SHARED_EXT_VER)
 .PHONY: $(LIBZSTD)  # must be run every time
-$(LIBZSTD): CFLAGS += -fPIC -fvisibility=hidden
-$(LIBZSTD): LDFLAGS += -shared
+$(LIBZSTD): CPPFLAGS += $(CPPFLAGS_DYNLIB)
+$(LIBZSTD): CFLAGS   += -fPIC -fvisibility=hidden
+$(LIBZSTD): LDFLAGS  += -shared $(LDFLAGS_DYNLIB)
 
 ifndef BUILD_DIR
 # determine BUILD_DIR from compilation flags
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -151,7 +151,7 @@ clean:
 ifneq (,$(filter $(shell uname),Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD DragonFly NetBSD MSYS_NT Haiku))
 
 HOST_OS = POSIX
-CMAKE_PARAMS = -DZSTD_BUILD_CONTRIB:BOOL=ON -DZSTD_BUILD_STATIC:BOOL=ON -DZSTD_BUILD_TESTS:BOOL=ON -DZSTD_ZLIB_SUPPORT:BOOL=ON -DZSTD_LZMA_SUPPORT:BOOL=ON -DCMAKE_BUILD_TYPE=Release
+CMAKE_PARAMS = -DZSTD_BUILD_CONTRIB:BOOL=ON -DZSTD_BUILD_STATIC:BOOL=ON -DZSTD_BUILD_TESTS:BOOL=ON -DZSTD_ZLIB_SUPPORT:BOOL=ON -DZSTD_LZMA_SUPPORT:BOOL=ON -DZSTD_MULTITHREAD_SUPPORT:BOOL=ON -DCMAKE_BUILD_TYPE=Release
```
Cyan4973: do you mean `-DZSTD_MULTITHREAD` ?

lib/Makefile
```diff
@@ -292,16 +295,20 @@ libzstd : $(LIBZSTD)
 
 .PHONY: lib
 lib : libzstd.a libzstd
+	@if [ $(ZSTD_NO_MULTITHREAD) = 1 ]; then \
```
Cyan4973: are you sure of this syntax ?

lib/Makefile
```diff
@@ -292,16 +295,20 @@ libzstd : $(LIBZSTD)
...
+  CPPFLAGS += -DZSTD_MULTITHREAD
+  LDFLAGS  += -pthread
+endif
 
-
-# note : do not define lib-mt or lib-release as .PHONY
+# note : do not define lib-release as .PHONY
 # make does not consider implicit pattern rule for .PHONY target
-
-%-mt : CPPFLAGS += -DZSTD_MULTITHREAD
```
Cyan4973, line 300: Minor : modified scope

The purpose of `%-mt` rule is to convert any base rule (`lib`, `libzstd`, `libzstd.a`) into a multithreaded variant. Heck, even `make all` could become `make all-mt`.

By removing these lines, we make invoking these targets an error.
I get it that they become redundant,
but _maybe_ there are users who are using them right now
and it would be great that their build script still work.

Suggestion : keep it, as a simple redirector, so that invoking them still works.
`%-mt : %`
Maybe update the message to indicate that `-mt` suffix is no longer necessary, and that the target will be removed in a future version (v1.6.0?).

Makefile
senhuang42: Seems like CMake uses a different var for multithreading:
https://github.com/facebook/zstd/blob/12c045f74d922dc934c168f6e1581d72df983388/build/cmake/CMakeLists.txt#L96

lib/Makefile
senhuang42: It seems like this works? I'm not too familiar with what's acceptable in a Makefile, though I suppose another way to emit this message would be to use the Makefile syntax, and do something like:
```
ZSTD_MULTITHREAD_MSG = multithreading disabled
ifneq ($(ZSTD_NO_MULTITHREAD), 1)
  ZSTD_MULTITHREAD_MSG = multithreading enabled
endif

lib : libzstd.a libzstd
	@echo $(ZSTD_MULTITHREAD_MSG)
```

Is this better?

Makefile
Cyan4973: ok ! I was confused because `-DNAME` feels similar to `gcc` method to add preprocessor macro `NAME` on command line, 
but apparently it is the very same convention used by `cmake` to define its own variables,
hence the confusion.
Anyway, in such a case, the `cmake` build script can indeed use its own naming scheme.

The other thing that feels a bit unexpected here 
is to define `CMAKE_PARAMS` so far away from the only place which uses it.
It looks to me that this variable definition should be brought down closer to `cmakebuild`,
which is the only recipe which uses it.
(pay attention to the `ifneq (,$(filter MSYS%,$(shell uname)))` just above which also redefines `CMAKE_PARAMS`).

lib/Makefile
Cyan4973: Yes, this construction would be a bit more familiar.

Cyan4973:
We may have one remaining issue 
for applications which link to the static library `libzstd.a`.

Unless I'm mistaken, it's not possible to embed the link `-pthread` into the archive `libzstd.a`.
(in contrast, the dynamic library `libzstd.so.1` can be generated with an embedded link to `-pthread`).

A consequence is that an existing program linking to `libzstd.a` and expecting the usual "single-threaded" variant will fail the link stage, unless it adds `-pthread` to its build script.
This is in contrast with another program linking to the dynamic `libzstd.so.1`, which would still run and link exactly the same way, and wouldn't need to be aware that `libzstd.do.1` is now multithreaded.

To move this PR forward, we will have to decide something regarding the static library.
Either : 
- Keep it that way, users linking to the static library `libzstd.a` must add `-pthread` to their build script. Unmodified build scripts will fail.
- Make `libzstd.a` single-threaded by default, while `libzstd.so.1` would become multi-threaded. This way, existing build scripts would remain unaffected.
- Scrap the PR, keep everything single-threaded by default, as it is today.

While the second option may feel a bit "weird", due to creating different scopes for the static and dynamic libraries, I also note that it would make `libzstd.a` similar to "integrate `zstd` by source code" scenario, which is still "single-threaded" by default (one has to add `ZSTD_MULTITHREAD` macro to make it multi-threaded).

For discussion.

senhuang42:
Option 2 to generate a single-threaded `libzstd.a` with muli-threaded `libzstd.so.1` is an interesting option... but I think it does start to make zstd + mt somewhat confusing with all the possible combinations?

If we go with option 2 we'll have:
- `ZSTD_NO_MULTITHREAD` - new addition to disable multithreading by default. Has no effect on `libzstd.a`.
- `ZSTD_MULTITHREAD` - existing macro to enable multithreading, which would get enabled by default in the dylib build, but not the `libzstd.a` build
  - Which would need to be directly specified when building `libzstd.a` to get multithreading
- `libzstd-nomt` - another build target for with single-thread mode _and_ without zstdmt_compress.c

Cyan4973:
Yeah, that's a bit confusing.
It points to a need for better documentation.

What seems to be logical would be : 
- `make lib-mt` : both libraries (dynamic and static) are produced with multithreading enabled
- `make lib-nomt` : both libraries are produced with multithreading _disabled_
- `make lib` : default mode, dynamic library is multithreaded, static library is single-threaded.
    we will also have to justify _why_ we made this choice for _default_
    (impact on existing build scripts).
- In all cases : integration at source code level remains single-threaded by default (multi-thread requires explicit setting of build macro `ZSTD_MULTITHREAD`).

lib/Makefile
```diff
@@ -166,6 +169,14 @@ endif
 endif
 CPPFLAGS  += -DZSTD_LEGACY_SUPPORT=$(ZSTD_LEGACY_SUPPORT)
```
Cyan4973: I see a slight imbalance in the way flags are passed down the chain
between the static of dynamic libraries.
The static library depends on modifications of default flags implicitly updated (or not),
while the dynamic library entirely resets the flag variables during the final recipe.
While it works, I'm afraid such imbalance could impact maintenance (i.e. future modifications).

So here is my recommendation.
- Have a variable `CPPFLAGS_STATICLIB` the same way you already have `CPPFLAGS_DYNLIB`.
    - Initialize it to "empty"
    - Prefer to set it using `=` rather than `:=`. It's a nice convention for the reviewer to pay attention that this value will be updated later on.
- Initialize  `CPPFLAGS_DYNLIB` and `LDFLAGS_DYNLIB` to their _delta_ vs baseline (not the full line), aka:
    - `CPPFLAGS_DYNLIB = -DZSTD_MULTITHREAD`
    - `LDFLAGS_DYNLIB = -pthread`
    - Alternatively, use `CPPFLAGS_MT` to declare starting values as constant (almost the same as currently, but only the delta to baseline, not the full concatenation). In this case, use `:=` to indicate they are supposed to remain unmodified.
- In the recipe for the dynamic library : 
    - `$(LIBZSTD): CPPFLAGS += $(CPPFLAGS_DYNLIB)`   ( note the `+=`)
    - `$(LIBZSTD): LDFLAGS  += $(LDFLAGS_DYNLIB)`
- In the recipe for the static library : 
    - `$(ZSTD_STATLIB): CPPFLAGS += $(CPPFLAGS_STATICLIB)`
- Now, in `%-mt` and `%-nomt`, set these 3 complementary variables to their appropriate values.

The outcome will be similar. But the chain of flag modifications will be symmetric between the static and the dynamic libraries, which makes it easier to follow and update in the future.

lib/Makefile
```diff
@@ -8,6 +8,9 @@
 # You may select, at your option, one of the above-listed licenses.
 # ################################################################
 
+# Note: by default, the dynamic library is built single-threaded and static library is built
```
terrelln: Is this comment accurate?

lib/Makefile
senhuang42: This suggestion makes a lot of sense, and is definitely easier to reason about.

lib/Makefile
senhuang42: Good catch!

lib/Makefile
```diff
@@ -238,7 +246,11 @@ ZSTD_STATLIB_OBJ := $(addprefix $(ZSTD_STATLIB_DIR)/,$(ZSTD_LOCAL_OBJ))
 $(ZSTD_STATLIB): ARFLAGS = rcs
 $(ZSTD_STATLIB): | $(ZSTD_STATLIB_DIR)
 $(ZSTD_STATLIB): $(ZSTD_STATLIB_OBJ)
-	@echo compiling static library
+ifneq (,$(filter -DZSTD_MULTITHREAD,$(CPPFLAGS)))
+	@echo compiling multi-threaded static library $(LIBVER)
+else
+	@echo compiling single-threaded static library $(LIBVER)
+endif
```
senhuang42: Bizzare behavior I don't understand here: When I compile with `BUILD_DIR` set, somehow the expression `$(filter -DZSTD_MULTITHREAD,$(CPPFLAGS))` evaluates to empty string (this doesn't happen normally). This causes us to output `compiling single-threaded` even in mt builds.

The confusing thing is that even when I also `@echo $(CPPFLAGS)` in the same line, it clearly shows that `-DZSTD_MULTITHREAD` is in `CPPFLAGS`, and even `@echo $(filter -DZSTD_MULTITHREAD,$(CPPFLAGS))` evaluates to `-DZSTD_MULTITHREAD`. Is there something weird going on with the `SET_CACHE_DIRECTORY` thing?

Edit:
Hmm, I suppose one explanation is that `ifneq (,$(filter -DZSTD_MULTITHREAD,$(CPPFLAGS)))` gets evaluated before any of this begins? So then $(CPPFLAGS) would not contain `-DZSTD_MULTITHREAD` at the beginning of execution. Is there a proper way to make the `if` check a lazy one?

Edit2:
Changed this to the shell `if`, seems to work now with `BUILD_DIR` set.

Cyan4973: Yep, it was a question of parsing order,
with the initial formulation being part of stage 1.
Since `BUILD_DIR` skips reloading the `Makefile`,
all updates, which are part of stage 2, are ignored for stage 1 
(without `BUILD_DIR`, `Makefile` is reloaded, and variables are passed to following process,
hence "stage 2" changes become "stage 1" for the following process).

The new formulation is part of stage 3,
thus applies any change achieved during previous stages.

lib/Makefile
```diff
@@ -8,6 +8,9 @@
 # You may select, at your option, one of the above-listed licenses.
 # ################################################################
 
+# Note: by default, the static library is built single-threaded and dynamic library is built
```
Cyan4973, line 11: I think it's worthwhile to indicate something similar
in `lib/README.md`.

senhuang42: Yep! In the process of updating the readme now.

lib/README.md
```diff
@@ -19,12 +19,16 @@ The scope can be reduced on demand (see paragraph _modular build_).
...
+
 Enabling multithreading requires 2 conditions :
 - set build macro `ZSTD_MULTITHREAD` (`-DZSTD_MULTITHREAD` for `gcc`)
 - for POSIX systems : compile with pthread (`-pthread` compilation flag for `gcc`)
 
-Both conditions are automatically applied when invoking `make lib-mt` target.
+For convenience, we provide a build target to generate multi and single threaded libraries:
+- Force enable multithreading on both dynamic and static libraries by appending `-mt` to the target, e.g. `make lib-mt`.
+- Force disable multithreading on both dynamic and static libraries by appending `-nomt` to the target, e.g. `make lib-nomt`.
+- And as mentioned before, dynamic library is multithreaded, and static library is single-threaded, e.g. `make lib`.
```
Cyan4973: nit: "By default, as mentioned before, ..."

Cyan4973:
minor : PR description :
I guess `ZSTD_NO_MULTITHREAD` is no longer within the picture

Cyan4973 (approved):

PR #2586, opened 2021-04-22
Add DDS to oss fuzzer

senhuang42:
DDS wasn't getting fuzzed, this adds it (back?) to OSS-Fuzz

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3365,7 +3365,7 @@ static U32 ZSTD_resolveRepcodeToRawOffset(const U32 rep[ZSTD_REP_NUM], const U32
  * ZSTD_seqStore_resolveOffCodes() reconciles any possible divergences in offset history that may arise
  * due to emission of RLE/raw blocks that disturb the offset history, and replaces any repcodes within
  * the seqStore that may be invalid.
- * 
+ *
  * dRepcodes are updated as would be on the decompression side. cRepcodes are updated exactly in
  * accordance with the seqStore.
  */
@@ -3555,14 +3555,14 @@ static size_t ZSTD_compressBlock_splitBlock_internal(ZSTD_CCtx* zc, void* dst, s
      * may become invalid. In order to reconcile potentially invalid repcodes, we keep track of two
      * separate repcode histories that simulate repcode history on compression and decompression side,
      * and use the histories to determine whether we must replace a particular repcode with its raw offset.
-     * 
+     *
      * 1) cRep gets updated for each partition, regardless of whether the block was emitted as uncompressed
      *    or RLE. This allows us to retrieve the offset value that an invalid repcode references within
      *    a nocompress/RLE block.
      * 2) dRep gets updated only for compressed partitions, and when a repcode gets replaced, will use
      *    the replacement offset value rather than the original repcode to update the repcode history.
      *    dRep also will be the final repcode history sent to the next block.
-     * 
+     *
      * See ZSTD_seqStore_resolveOffCodes() for more details.
      */
     repcodes_t dRep;
@@ -6192,7 +6192,7 @@ static int ZSTD_dedicatedDictSearch_isSupported(
 {
     return (cParams->strategy >= ZSTD_greedy)
         && (cParams->strategy <= ZSTD_lazy2)
-        && (cParams->hashLog >= cParams->chainLog)
+        && (cParams->hashLog > cParams->chainLog)
         && (cParams->chainLog <= 24);
 }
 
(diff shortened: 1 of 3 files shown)


Cyan4973 (approved):

senhuang42:
Looks like something is not quite happy with how DDS adjusts its parameters and the increase and/or decrease of `hashLog`.

senhuang42:
Also, local testing reveals that this assert is also failing `make -j regressiontest`: https://github.com/facebook/zstd/blob/3eb3845898401c0beb09904439aee4e1f3ca8111/lib/compress/zstd_lazy.c#L531

Cyan4973:
> Also, local testing reveals that this assert is also failing `make -j regressiontest`:
> 
> https://github.com/facebook/zstd/blob/3eb3845898401c0beb09904439aee4e1f3ca8111/lib/compress/zstd_lazy.c#L531

cc @felixhandte

senhuang42:
So I _think_ the two changes should be the correct fix. If I'm not mistaken, the original seems to indicate that there was never any actual adjustment of the DDSS `hashLog`? Would this change/improve the performance of `DDSS`?

lib/compress/zstd_compress.c
```diff
@@ -6168,7 +6168,7 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
 
 static ZSTD_compressionParameters ZSTD_dedicatedDictSearch_getCParams(int const compressionLevel, size_t const dictSize)
 {
-    ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, 0, dictSize, ZSTD_cpm_createCDict);
+    ZSTD_compressionParameters cParams = ZSTD_getCParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize, ZSTD_cpm_createCDict);
```
felixhandte: Makes sense.

lib/compress/zstd_compress.c
```diff
@@ -4795,7 +4795,7 @@ ZSTDLIB_API ZSTD_CDict* ZSTD_createCDict_advanced2(
     if (cctxParams.enableDedicatedDictSearch) {
         cParams = ZSTD_dedicatedDictSearch_getCParams(
             cctxParams.compressionLevel, dictSize);
-        ZSTD_overrideCParams(&cParams, &cctxParams.cParams);
+        ZSTD_overrideCParams(&cctxParams.cParams, &cParams);
```
felixhandte: Hm. The original may be mistaken, but I don't think this is a correct fix.

This call is of the form `ZSTD_overrideCParams(dst, src)`. This had been part of a multi-step resolution of params into the local `cParams`. The intent here is to first get the default cparams for the level, and then override those params with any explicit cparams that were set in the cctxparams by the user. These resolved params are then fed back into the cctxparams (on :4812) and passed down the stack.

With this new version, we will just ignore any explicit cparams set by the user. Which might well fix the assertion failure you're seeing, if the clevel derived params are all safe, but the fuzzer is setting bad cparams that aren't getting caught and rejected.

But then the correct fix would be to correctly validate the cparam set for ddss.

Cyan4973: This detailed explanation is probably worth becoming a code comment

senhuang42: That makes sense: so the idea here is that if `hLog` is explicitly set, DDS will not override that by adding 2 to it.

But the actual issue here is that if we set `hLog = 7` explicitly, DDS will not override. But later on, `ZSTD_dedicatedDictSearch_revertCParams()` is called anyways, subtracting 2 from the `hLog`, making us end up with an invalid `hLog` of 5 in `ZSTD_resetCCtx_byAttachingCDict()` here:

https://github.com/facebook/zstd/blob/3eb3845898401c0beb09904439aee4e1f3ca8111/lib/compress/zstd_compress.c#L1996-L2016

At this point in the code, is it even possible to tell whether or not the `hLog` is an `hLog` that was explicitly set or an `hLog` that was adjusted because DDS is available? Ideally, we'd want to be able to detect the latter case, and only subtract from `hLog` then.

Maybe we can just call `ZSTD_clampCParams()` on the `ZSTD_adjustCParams_internal()` call right after, but that would still mean that the `hLog` could still get reduced when it shouldn't be.

Cyan4973: Why would the interpretation of `hlog` by DDS be different, depending on this parameter being automatically determined or explicitly set ? The way `hlog` is set should not matter.

Moreover, the _interpretation_ of `hlog` should remain local.
If DDS considers that a `hlog==7` will actually be interpreted as 5 for its own use, that's fine,
but this "interpreted value" should not be confused with the _parameter_ `hlog`.
As a consequence, this interpretation should remain private within DDS, and in no way impact any other part of the system, not even indirectly.

senhuang42:
I've pushed @felixhandte's fix for all the assert issues onto this PR in the latest commit, it should fix the issue.

terrelln:
Thanks @senhuang42, DDS coverage looks good now!

PR #2589, opened 2021-04-26
[trace] Remove default definitions of weak symbols

terrelln:
Instead of providing a default no-op implementation, check the symbols
for `NULL` before accessing them. Providing a default implementation
doesn't reliably work with dynamic linking. Depending on link order the
default implementations may not be overridden. By skipping the default
implementation, all link order issues are resolved. If the symbols
aren't provided the weak function will be `NULL`.

This is showing up as test failures in dev mode (dynamic linking) in D27778613.


senhuang42 (approved):
Makes sense, and LGTM

build/meson/lib/meson.build
```diff
@@ -22,7 +22,6 @@ libzstd_sources = [join_paths(zstd_rootdir, 'lib/common/entropy_common.c'),
   join_paths(zstd_rootdir, 'lib/common/threading.c'),
   join_paths(zstd_rootdir, 'lib/common/pool.c'),
   join_paths(zstd_rootdir, 'lib/common/zstd_common.c'),
-  join_paths(zstd_rootdir, 'lib/common/zstd_trace.c'),
```
senhuang42, line 25: Why was this removed from the meson build btw?

Cyan4973 (approved):

build/meson/lib/meson.build
terrelln: This PR deletes the file.

Issue #2591, opened 2021-04-29
A possible divide by zero bug?


yiyuaner:
Hi,  in function `ZSTD_buildBlockEntropyStats_sequences`, we have the following [code](https://github.com/facebook/zstd/blob/f8afc6657393a5833ae9303e127cc76299893b20/lib/compress/zstd_compress.c#L3089):

```
size_t const nbSeq = seqStorePtr->sequences - seqStorePtr->sequencesStart;
...
stats = ZSTD_buildSequencesStatistics(seqStorePtr, nbSeq,...)
```
If `seqStorePtr->sequences == seqStorePtr->sequencesStart`, `nbSeq` will be equal to `0` and passed to the function `ZSTD_buildSequencesStatistics`.  Eventually, `nbSeq` will be used as the divisor of a division operation without checking.

Here is the full trace for triggering this problem (with links pointing to specific lines in the code):
[`ZSTD_buildBlockEntropyStats_sequences`](https://github.com/facebook/zstd/blob/f8afc6657393a5833ae9303e127cc76299893b20/lib/compress/zstd_compress.c#L3107)   --> [`ZSTD_buildSequencesStatistics`](https://github.com/facebook/zstd/blob/f8afc6657393a5833ae9303e127cc76299893b20/lib/compress/zstd_compress.c#L2412)  --> [`ZSTD_selectEncodingType`](https://github.com/facebook/zstd/blob/f8afc6657393a5833ae9303e127cc76299893b20/lib/compress/zstd_compress_sequences.c#L207) --> [`ZSTD_entropyCost`](https://github.com/facebook/zstd/blob/f8afc6657393a5833ae9303e127cc76299893b20/lib/compress/zstd_compress_sequences.c#L89), finally leading to a division in ZSTD_entropyCost:

``
        unsigned norm = (unsigned)((256 * count[s]) / total); // here total equals to nbSeq, which may equal to zero
``

senhuang42:
Thanks for bringing this up. 

In practice, I don't believe that we can ever pass `nbSeq == 0` to `ZSTD_buildSequencesStatistics`. 
In the `ZSTD_entropyCompressSeqStore_internal()` entrypoint:
https://github.com/facebook/zstd/blob/f8afc6657393a5833ae9303e127cc76299893b20/lib/compress/zstd_compress.c#L2570-L2574

And the block splitter entry point:
[...]

PR #2592, opened 2021-04-29
Assert no division by 0 in `ZSTD_entropyCost()`, fix superblocks no sequences case

senhuang42:
Discovered in #2591

Not possible to have the division by 0, so we should `assert()` that.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2378,8 +2378,9 @@ typedef struct {
 } ZSTD_symbolEncodingTypeStats_t;
 
 /* ZSTD_buildSequencesStatistics():
- * Returns the size of the statistics for a given set of sequences, or a ZSTD error code,
- * Also modifies LLtype, Offtype, MLtype, and lastNCount to the appropriate values.
+ * Returns a ZSTD_symbolEncodingTypeStats_t, or a zstd error code in the `size` field.
+ * Modifies `nextEntropy` to have the appropriate values as a side effect.
+ * nbSeq must be greater than 0.
  *
  * entropyWkspSize must be of size at least ENTROPY_WORKSPACE_SIZE - (MaxSeq + 1)*sizeof(U32)
  */
@@ -2404,6 +2405,7 @@ ZSTD_buildSequencesStatistics(seqStore_t* seqStorePtr, size_t nbSeq,
     /* convert length/distances into codes */
     ZSTD_seqToCodes(seqStorePtr);
     assert(op <= oend);
+    assert(nbSeq != 0); /* ZSTD_selectEncodingType() divides by nbSeq */
     /* build CTable for Literal Lengths */
     {   unsigned max = MaxLL;
         size_t const mostFrequent = HIST_countFast_wksp(countWorkspace, &max, llCodeTable, nbSeq, entropyWorkspace, entropyWkspSize);   /* can't fail */
@@ -3081,6 +3083,20 @@ static size_t ZSTD_buildBlockEntropyStats_literals(void* const src, size_t srcSi
     }
 }
 
+
+/* ZSTD_buildDummySequencesStatistics():
+ * Returns a ZSTD_symbolEncodingTypeStats_t with all encoding types as set_basic,
+ * and updates nextEntropy to the appropriate repeatMode.
+ */
+static ZSTD_symbolEncodingTypeStats_t
+ZSTD_buildDummySequencesStatistics(ZSTD_fseCTables_t* nextEntropy) {
+    ZSTD_symbolEncodingTypeStats_t stats = {set_basic, set_basic, set_basic, 0, 0};
+    nextEntropy->litlength_repeatMode = FSE_repeat_none;
+    nextEntropy->offcode_repeatMode = FSE_repeat_none;
+    nextEntropy->matchlength_repeatMode = FSE_repeat_none;
+    return stats;
+}
+
 /** ZSTD_buildBlockEntropyStats_sequences() :
  *  Builds entropy for the sequences.
  *  Stores symbol compression modes and fse table to fseMetadata.
@@ -3104,10 +3120,11 @@ static size_t ZSTD_buildBlockEntropyStats_sequences(seqStore_t* seqStorePtr,
     ZSTD_symbolEncodingTypeStats_t stats;
 
     DEBUGLOG(5, "ZSTD_buildBlockEntropyStats_sequences (nbSeq=%zu)", nbSeq);
-    stats = ZSTD_buildSequencesStatistics(seqStorePtr, nbSeq,
-                                        prevEntropy, nextEntropy, op, oend,
-                                        strategy, countWorkspace,
-                                        entropyWorkspace, entropyWorkspaceSize);
+    stats = nbSeq != 0 ? ZSTD_buildSequencesStatistics(seqStorePtr, nbSeq,
+                                          prevEntropy, nextEntropy, op, oend,
+                                          strategy, countWorkspace,
+                                          entropyWorkspace, entropyWorkspaceSize)
+                       : ZSTD_buildDummySequencesStatistics(nextEntropy);
     FORWARD_IF_ERROR(stats.size, "ZSTD_buildSequencesStatistics failed!");
     fseMetadata->llType = (symbolEncodingType_e) stats.LLtype;
     fseMetadata->ofType = (symbolEncodingType_e) stats.Offtype;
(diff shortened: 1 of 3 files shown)


senhuang42:
Aha, I guess not, for superblocks.

Cyan4973:
Well, even for superblocks,
with the objective of creating inner blocks with approximately constant compressed size,
assuming that one such inner block has zero sequence, hence is all literals,
I would have expected such case to _not_ attempt to generate statistics for sequences.
Maybe something to improve there ...

Regarding the issue #2591 : what matters specifically is to ensure there is no division by zero.
Hence, the more important check is `assert(total > 0)`.

terrelln:
Yeah, looks like there is a bug in superblock mode. When it calls `ZSTD_buildBlockEntropyStats()` `nbSeq` may be zero, and it isn't checked later. But regular compression looks fine.

https://github.com/facebook/zstd/blob/10e551311369d6f0e2b0df7b7673e10324b74062/lib/compress/zstd_compress_superblock.c#L555-L560
https://github.com/facebook/zstd/blob/10e551311369d6f0e2b0df7b7673e10324b74062/lib/compress/zstd_compress.c#L3159-L3164
https://github.com/facebook/zstd/blob/10e551311369d6f0e2b0df7b7673e10324b74062/lib/compress/zstd_compress.c#L3125-L3128
https://github.com/facebook/zstd/blob/10e551311369d6f0e2b0df7b7673e10324b74062/lib/compress/zstd_compress.c#L2405

Cyan4973:
I presume the suspected bug in superblock mode will be a separate issue / PR ?

senhuang42:
Updated to fix superblocks with `nbSeq == 0`

lib/compress/zstd_compress.c
```diff
@@ -3103,6 +3104,12 @@ static size_t ZSTD_buildBlockEntropyStats_sequences(seqStore_t* seqStorePtr,
     size_t entropyWorkspaceSize = wkspSize - (MaxSeq + 1) * sizeof(*countWorkspace);
     ZSTD_symbolEncodingTypeStats_t stats;
 
+    if (nbSeq == 0) {
+        /* Do not build sequences statistics if nbSeq is 0, just return a default */
+        fseMetadata->llType = fseMetadata->mlType = fseMetadata->ofType = set_compressed;
```
Cyan4973: What are the consequences of settings these `fseMedata` to `set_compressed` ?
We know that there is no sequence, so it's not going to be used in this block.
It also will not be written into the compressed frame.

But what about the _next_ block ?
Will it consider the previous block as featuring "compressed sequences", and therefore entropy headers ?
Asking as it _might_ impact the algorithm which tries to decide between "repeating previous stats" or "issuing new stats".

Alternatively, this information might possibly be discarded later on and not even read (in which case, why setting them ?).

senhuang42: So basically `regressiontest` only passes with the default set to `set_compressed` or `set_rle`.
There are three places where the `encodingType` is read:
https://github.com/facebook/zstd/blob/e7e4b74b2b99016234e8a44b04764ab6305fe83c/lib/compress/zstd_compress_superblock.c#L193-L194
https://github.com/facebook/zstd/blob/e7e4b74b2b99016234e8a44b04764ab6305fe83c/lib/compress/zstd_compress_superblock.c#L368-L371
https://github.com/facebook/zstd/blob/e7e4b74b2b99016234e8a44b04764ab6305fe83c/lib/compress/zstd_compress_superblock.c#L405-L407

The first two cases do not ever get read, due to `nbSeq == 0`. The last case still does, so basically whatever we decide to set `fseMetadata->llType` to, it's basically to affect how this function returns. It seems like even with `nbSeq == 0`, we'd want `ZSTD_needSequenceEntropyTables()` to evaluate to true right? At least the fuzzer tests seem to indicate so.

Cyan4973: If my understanding is correct, `ZSTD_needSequenceEntropyTables()` is invoked in only a single place :  
https://github.com/facebook/zstd/blob/e7e4b74b2b99016234e8a44b04764ab6305fe83c/lib/compress/zstd_compress_superblock.c#L521

This place seems to check if there is a discrepancy between needing entropy tables, and writing entropy tables.

This feels weird to me.
If `nbSeq==0`, then there should be no need to write any sequence entropy header.
Hence, we should have `writeSeqEntropy == 0`.

But it's set to `1` unconditionally at the beginning of the function.
The only place where it could be updated to `0` is here, and that's for an unrelated reason : 
https://github.com/facebook/zstd/blob/e7e4b74b2b99016234e8a44b04764ab6305fe83c/lib/compress/zstd_compress_superblock.c#L512

This code is very complex, and difficult to follow.
It looks to me that there should be no need to write Sequences Entropy is there is no sequence.
So my first idea would be to change that initialization to `0` in this case.
Unfortunately, the "number of Sequences" is not provided in a straightforward manner,
but is accessible through `(unsigned)(send-sstart)` (discovered from the traces) .

Since the code is complex, it could be that I misinterpret the objective or hidden usages of this variable.
Given that it's then used in multiple places (`ZSTD_estimateSubBlockSize()` and `ZSTD_compressSubBlock()`) 
it feels like a dangerous change.
Worth testing, and maybe worth following what happens to this variable in these sub-functions.

senhuang42: Yeah the code is confusing.

Actually, in both `ZSTD_estimateSubBlockSize()` and `ZSTD_compressSubBlock()`, for the `nbSeq == 0` case, we don't use `writeSeqEntropy`. In both cases, we return early if `nbSeq == 0`, leaving that variable unread.

So really, I think it's just this decision of whether or not to emit uncompressed that we end up affecting. 

`writeSeqEntropy == 0` if `seqEntropyWritten = 1`, so we can think of `writeSeqEntropy == 1` as saying "entropy has not been written". So basically the check of 
` if (writeSeqEntropy && ZSTD_needSequenceEntropyTables(&entropyMetadata->fseMetadata))` is saying that if we did not write seqEntropyTables, but did need them, we should emit uncompressed. So my setting of `llType` to `set_compressed/rle` is not correct, since we're saying that we always need them.

But at the same time, setting `writeSeqEntropy = (send-sstart) == 0 ? 0 : 1;` doesn't seem to produce the correct results either, as it fails the tests.

Cyan4973: It's a pretty bad situation.
It seems we don't understand well enough what's happening in this unit,
which becomes a maintenance liability.
And since it has never worked correctly, and is not used anywhere,
I'm wondering if we would be better off without it.

senhuang42: One thing I've figured out is that the test failures are a result of the added early return in `ZSTD_buildBlockEntropyStats_sequences()` for `nbSeq==0` causing us to not run `ZSTD_buildSequencesStatistics()`, which causes us to miss out on some side effects which are apparently necessary - it doesn't actually matter what the `llType, etc.` are when `nbSeq == 0`. I'll modify the PR accordingly to see if I can get that working.

senhuang42: I've updated this with a fix that should make sense (basically `selectEncodingType()` should evaluate to `set_basic` if `nbSeq==0`, so we just set that explicitly here. Though yeah, I'm not sure whether the cost of maintaining this code is worth its general complexity, especially considering that I don't think anyone really uses it.

The added branch should be hardwired for the typical compression path - I measured no speed difference at level 1.

lib/compress/zstd_compress.c
```diff
@@ -2388,7 +2388,8 @@ ZSTD_buildSequencesStatistics(seqStore_t* seqStorePtr, size_t nbSeq,
                         const ZSTD_fseCTables_t* prevEntropy, ZSTD_fseCTables_t* nextEntropy,
                               BYTE* dst, const BYTE* const dstEnd,
                               ZSTD_strategy strategy, unsigned* countWorkspace,
-                              void* entropyWorkspace, size_t entropyWkspSize) {
+                              void* entropyWorkspace, size_t entropyWkspSize,
+                              const U32 selectEncodingTypes) {
```
Cyan4973: I'm annoyed that this `selectEncodingTypes` concept gets externalized in the function prototype.

It looks to me that it would make more sense to keep these topics separated.

What you could have is something like 
`ZSTD_symbolEncodingTypeStats_t ZSTD_buildDummyStatistics()`
and it would do the same thing as what's inside the `if (selectEncodingTypes) { } else { ... } ` statement.

It would then be selected directly within `ZSTD_buildBlockEntropyStats_sequences()`, instead of `ZSTD_buildSequencesStatistics()`, depending on `nbSeq==0` or not.

Cyan4973 (approved):

PR #2594, opened 2021-04-30
seekable decompression fixes

azat:
### Changelog:

- **seekable_format: cap the offset+len up to the last dOffset**
  This will allow to read the whole file w/o gotting corruption error if
  the offset is more then the data left in file, i.e.:

      $ ./seekable_compression seekable_compression.c 8192 | head
      $ zstd -cdq seekable_compression.c.zst | wc -c
      4737

  Before this patch:

      $ ./seekable_decompression seekable_compression.c.zst 0 10000000 | wc -c
      ZSTD_seekable_decompress() error : Corrupted block detected
      0

  After:

      $ ./seekable_decompression seekable_compression.c.zst 0 10000000 | wc -c
      4737

- **seekable_decompression: break when ZSTD_seekable_decompress() returns zero**
- **seekable_decompression_mem: break when ZSTD_seekable_decompress() returns zero**
- **seekable_format: fix from-file reading (not in-memory)**

### DIFF
--- a/contrib/seekable_format/examples/seekable_decompression.c
+++ b/contrib/seekable_format/examples/seekable_decompression.c
@@ -99,6 +99,9 @@ static void decompressFile_orDie(const char* fname, off_t startOffset, off_t end
 
     while (startOffset < endOffset) {
         size_t const result = ZSTD_seekable_decompress(seekable, buffOut, MIN(endOffset - startOffset, buffOutSize), startOffset);
+        if (!result) {
+            break;
+        }
 
         if (ZSTD_isError(result)) {
             fprintf(stderr, "ZSTD_seekable_decompress() error : %s \n",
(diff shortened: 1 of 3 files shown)


contrib/seekable_format/examples/seekable_decompression.c
```diff
@@ -99,6 +99,9 @@ static void decompressFile_orDie(const char* fname, off_t startOffset, off_t end
 
     while (startOffset < endOffset) {
         size_t const result = ZSTD_seekable_decompress(seekable, buffOut, MIN(endOffset - startOffset, buffOutSize), startOffset);
+        if (!result) {
```
Cyan4973, line 102: Nice optimization.
Is there more to it ? (does it dodge an error case ?)

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -433,6 +433,11 @@ size_t ZSTD_seekable_initAdvanced(ZSTD_seekable* zs, ZSTD_seekable_customFile sr
 
 size_t ZSTD_seekable_decompress(ZSTD_seekable* zs, void* dst, size_t len, unsigned long long offset)
 {
+    unsigned long long eos = zs->seekTable.entries[zs->seekTable.tableLen].dOffset;
```
Cyan4973: nit: this code base is very `const` friendly.
=> `unsigned long long const eos = ..`

contrib/seekable_format/examples/seekable_decompression.c
azat: >Is there more to it ?

Added the same for `seekable_decompression_mem.c`

> (does it dodge an error case ?)

In theory before fixing frame overrun and not using checksums it is possible to go to endless loop

Cyan4973 (approved):

Issue #2556, opened 2021-03-25
No progress counter when decompressing into a pipe


fleaz:
**Describe the bug**
Hi,
we use zstd for the database dumps in our development environment at work and normaly one would get this Megabyte counter which shows how much data has already been decompressed which is quite handy for large dumps because one can estimate the remaining time. Recently this progress counter disappeared. Everything is still working fine but the progress bar was a nice addition.

I tried to bisect when this problem occurred and with 1.4.4 (official Ubuntu Focal repo) I still see the progress counter, but with 1.4.8 (ondrej ppa) or directly building the dev branch myself, the counter is gone.

Best,
Felix

**To Reproduce**
1. Decrompress a SQL dump with a pipe directly into the `mysql` cli tool
```
zstd -d -c dump.sql.zst | mysql my-awesome-database
```

**Expected behavior**
I want to see the counter showing me how much data has already been decompressed.

**Screenshots and charts**
![Screenshot_2021-03-25_12:00:35](https://user-images.githubusercontent.com/2489598/112462784-bef4f200-8d61-11eb-875d-c6dd96e6469e.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu Focal in a Container

Cyan4973:
I presume that's because the output is pushed into `stdout`.
If it was pushed into a file instead, the activity notification counter would still be present

This is likely the consequence of a refactoring of `zstd` CLI output
which fixed a nb of edge cases regarding its display output.
And this resulted in the notification being disabled by default when `zstd` is used in combination with `stdout`.
I believe it's the correct default, as otherwise, the display can get messed up, with both `stdout` and `stderr` being printed interleaved,
or when multiple programs want to output something to display.

What's missing though is an ability to re-enable it on demand.
I would expect `-v` to achieve that, but it doesn't (it only prints the final result).
So either fix that, or support a new command flags, such as `--progress` (since we already have `--no-progress`).

senhuang42:
Addressed in #2595.

PR #2595, opened 2021-04-30
Add --progress flag

senhuang42:
Addressing #2556

A while ago, the way zstd interacts with stdin and stdout was refactored, removing any possibility of (de)compressed output to stdout displayed in the terminal to mix with stderr (since that's generally not what we want), so currently `zstdcli` doesn't generate a refreshing progress meter when output is going to stdout.

This PR adds that possibility back in, but now behind a new flag, `--progress` (to go along with --no-progress) that will force the progress meter to display on (de)compression.

### DIFF
--- a/programs/fileio.h
+++ b/programs/fileio.h
@@ -68,6 +68,8 @@ void FIO_freeContext(FIO_ctx_t* const fCtx);
 
 typedef struct FIO_display_prefs_s FIO_display_prefs_t;
 
+typedef enum { FIO_ps_auto, FIO_ps_never, FIO_ps_always } FIO_progressSetting_e;
+
 /*-*************************************
 *  Parameters
 ***************************************/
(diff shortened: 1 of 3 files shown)


programs/fileio.h
```diff
@@ -68,6 +68,8 @@ void FIO_freeContext(FIO_ctx_t* const fCtx);
 
 typedef struct FIO_display_prefs_s FIO_display_prefs_t;
 
+typedef enum { FIO_ps_auto, FIO_ps_never, FIO_ps_always } FIO_progressSetting_e;
```
Cyan4973, line 71: 👍

Cyan4973 (approved):
Looks good to me !

I'm wondering if there would be a way to automatically test this feature,
though I understand it's a fairly tricky topic.

Maybe redirect `stderr` to `stdout` and `grep` the result for traces (or absence of traces) of the counter ?

PR #2602, opened 2021-05-03
[LDM] Speed optimization on repetitive data

terrelln:
LDM does especially poorly on repetitive data when that data's hash happens
to have `(hash & stopMask) == 0`. Either because the `stopMask == 0` or
random chance. Optimize this case by skipping over repetitive patterns.
The detection is very simplistic, but should catch most of the offending
cases.

```
head -c 1G /dev/zero | perf stat -- ./zstd -1 -o /dev/null -v --zstd=ldmHashRateLog=1 --long
      21.187881087 seconds time elapsed

head -c 1G /dev/zero | perf stat -- ./zstd -1 -o /dev/null -v --zstd=ldmHashRateLog=1 --long
       1.208744521 seconds time elapsed
```

With this change, I see 1-5% ratio improvements for very compressible data,
like in `--patch-from` mode, because we skip adding the tails of very long
matches to our tables, which leaves room for other candidates.

It is approximately neutral for non `--patch-from` mode, since extremely long
matches aren't very common, so this corner case rarely gets triggered.

### DIFF
--- a/lib/compress/zstd_ldm.c
+++ b/lib/compress/zstd_ldm.c
@@ -57,6 +57,33 @@ static void ZSTD_ldm_gear_init(ldmRollingHashState_t* state, ldmParams_t const* 
     }
 }
 
+/** ZSTD_ldm_gear_reset()
+ * Feeds [data, data + minMatchLength) into the hash without registering any
+ * splits. This effectively resets the hash state. This is used when skipping
+ * over data, either at the beginning of a block, or skipping sections.
+ */
+static void ZSTD_ldm_gear_reset(ldmRollingHashState_t* state,
+                                BYTE const* data, size_t minMatchLength)
+{
+    U64 hash = state->rolling;
+    size_t n = 0;
+
+#define GEAR_ITER_ONCE() do {                                  \
+        hash = (hash << 1) + ZSTD_ldm_gearTab[data[n] & 0xff]; \
+        n += 1;                                                \
+    } while (0)
+    while (n + 3 < minMatchLength) {
+        GEAR_ITER_ONCE();
+        GEAR_ITER_ONCE();
+        GEAR_ITER_ONCE();
+        GEAR_ITER_ONCE();
+    }
+    while (n < minMatchLength) {
+        GEAR_ITER_ONCE();
+    }
+#undef GEAR_ITER_ONCE
+}
+
 /** ZSTD_ldm_gear_feed():
  *
  * Registers in the splits array all the split points found in the first
@@ -255,7 +282,7 @@ void ZSTD_ldm_fillHashTable(
     while (ip < iend) {
         size_t hashed;
         unsigned n;
-        
+
         numSplits = 0;
         hashed = ZSTD_ldm_gear_feed(&hashState, ip, iend - ip, splits, &numSplits);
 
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_ldm.c
```diff
@@ -57,6 +57,33 @@ static void ZSTD_ldm_gear_init(ldmRollingHashState_t* state, ldmParams_t const*
     }
 }
 
+/** ZSTD_ldm_gear_reset()
```
Cyan4973, line 60: 👍

Cyan4973:
So repetitive data is detected whenever an overlapping match happens ?
In which case, the match length is unconditionally skipped, irrespective of the produced hash value (wether `(hash & stopMask) == 0` or not) ?

To be fair, I was expecting this policy to be already in place for any match.
While I would expected such policy to scan faster, I presume it would also cost too much compression ratio ?

terrelln:
> So repetitive data is detected whenever an overlapping match happens ?
> In which case, the match length is unconditionally skipped

Exactly.

> irrespective of the produced hash value (wether (hash & stopMask) == 0 or not) ?

If it is a match, then it was inserted, so we actually know that `(hash & stopMask) == 0`.

> To be fair, I was expecting this policy to be already in place for any match.
> While I would expected such policy to scan faster, I presume it would also cost too much compression ratio ?

LDM has always inserted every position. However, switching to gear-hash made this scenario much worse. We used to use the same hash for the insertion-mask and the table-index. But with gear-hash we use `XXH64()` for the table-index. That means we have to re-hash every position that matches. It isn't a big deal when matches are somewhat infrequent, but repetitive bytes that happened to match can be a bad corner case.

Cyan4973 (approved):

lib/compress/zstd_ldm.c
```diff
@@ -447,6 +467,17 @@ static size_t ZSTD_ldm_generateSequences_internal(
...
 
             anchor = split + forwardMatchLength;
+
+            /* If there is an overlapping match skip to the end of it.
+             * No need to insert every position into our table. This
+             * is especially slow because position of a long overlapping match
+             * is inserted, every position is inserted.
+             */
+            if (mLength > (size_t)offset) {
+                ip = anchor;
```
mpu: I had a bit of trouble thinking about this because `ip` used to be constant in this loop.

It probably does not matter much but I think this may skip a bit too much. I would find the code easier to reason about if it skipped just the right amount: `ip = anchor - hashed;`

lib/compress/zstd_ldm.c
```diff
@@ -447,6 +467,17 @@ static size_t ZSTD_ldm_generateSequences_internal(
             ZSTD_ldm_insertEntry(ldmState, hash, newEntry, *params);
 
             anchor = split + forwardMatchLength;
+
+            /* If there is an overlapping match skip to the end of it.
+             * No need to insert every position into our table. This
+             * is especially slow because position of a long overlapping match
+             * is inserted, every position is inserted.
+             */
+            if (mLength > (size_t)offset) {
```
mpu: If I understand correctly, we want to save work in the `while (ip < ilimit)` loop.

If that is the case, I would instead detect the degenerate case using `anchor > ip + hashed`, then reset `ip = anchor - hashed` and break instead of continue. It makes it a bit clearer where the savings are expected and also makes it easy to see the relation with the `ip` reset and the increment line 483.

mpu (commented):
My two cents!

lib/compress/zstd_ldm.c
terrelln: Thats a good point! I actually intended to continue the `while (ip < ilimit)` loop, not the inner loop, so this is a logic bug. What you're saying is what I was intending to write. Thanks for the review!

terrelln: That isn't quite the same logic, since `mLength > offset` detects overlapping matches, whereas `anchor > ip + hashed` detects long matches.

However, I've found that `anchor > ip + hashed` improves compression ratio by a reasonable amount in some `--patch-from` use cases that I tested, where the compression ratio is very high. So it is probably a reasonable choice.

terrelln:
I've updated to the logic suggested by @mpu, but I have to do some more in-depth performance testing, since this covers more than just the overlapping-match edge case. I think it will be positive.

PR #2603, opened 2021-05-03
Bug fix & run overflow correction much more frequently in tests

terrelln:
* Fix overflow correction when `windowLog < cycleLog`. Previously, we
  got the correction wrong in this case, and our chain tables and binary
  trees would be corrupted. Now, we work as long as `maxDist` is a power
  of two, by adding `MAX(maxDist, cycleSize)` to our indices.
* When `ZSTD_WINDOW_OVERFLOW_CORRECT_FREQUENTLY` is defined to non-zero
  run overflow correction as frequently as allowed without impacting
  compression ratio.
* Enable `ZSTD_WINDOW_OVERFLOW_CORRECT_FREQUENTLY` in `fuzzer` and
  `zstreamtest` as well as all the OSS-Fuzz fuzzers. This has a 5-10%
  speed penalty at most, which seems reasonable.

This change allows both the OSS-Fuzz fuzzers and `fuzzer` and `zstreamtest` to all catch the `ZSTD_reduceIndex()` bug, as well as the assertion failure.

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -895,15 +895,55 @@ MEM_STATIC ZSTD_dictMode_e ZSTD_matchState_dictMode(const ZSTD_matchState_t *ms)
             ZSTD_noDict;
 }
 
+/* Defining this macro to non-zero tells zstd to run the overflow correction
+ * code much more frequently. This is very inefficient, and should only be
+ * used for tests and fuzzers.
+ */
+#ifndef ZSTD_WINDOW_OVERFLOW_CORRECT_FREQUENTLY
+#  ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+#    define ZSTD_WINDOW_OVERFLOW_CORRECT_FREQUENTLY 1
+#  else
+#    define ZSTD_WINDOW_OVERFLOW_CORRECT_FREQUENTLY 0
+#  endif
+#endif
+
+/**
+ * ZSTD_window_canOverflowCorrect():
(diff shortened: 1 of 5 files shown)


terrelln:
Tests failed because it caught the `ZSTD_reduceIndex()` bug, since I hadn't yet rebased onto dev.

lib/compress/zstd_compress_internal.h
```diff
@@ -895,15 +895,55 @@ MEM_STATIC ZSTD_dictMode_e ZSTD_matchState_dictMode(const ZSTD_matchState_t *ms)
             ZSTD_noDict;
 }
 
+/* Defining this macro to non-zero tells zstd to run the overflow correction
+ * code much more frequently. This is very inefficient, and should only be
+ * used for tests and fuzzers.
+ */
+#ifndef ZSTD_WINDOW_OVERFLOW_CORRECT_FREQUENTLY
```
Cyan4973, line 902: Should this macro be documented ?
I presume it's only meant to be used by maintainers.

Cyan4973 (approved):
Good design

lib/compress/zstd_compress_internal.h
terrelln: I don't think it really deserves documentation in any `README.md`, since it shouldn't be enabled in anything but tests. It is automatically enabled for fuzzers, which have a standard macro `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`.

If we wanted this to be used more widely, I think we would want to have a general debug flag like `DEBUGLOG`, except for enabling "expensive" checks.

PR #2604, opened 2021-05-03
Add threadPool unit tests to fuzzer.c

senhuang42:
This PR adds `threadPool` API tests for basic functionality. We test pool re-use and simultaneous use (I think in practice these are pretty similar scenarios). It's gated behind `ZSTD_MULTITHREAD`.

We add a dependency to `fuzzer.c` in order to give us access to `ZSTD_pthread_create/join`, (which `poolTests.c` already includes too).

### DIFF
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -42,6 +42,7 @@
 #include "timefn.h"       /* SEC_TO_MICRO, UTIL_time_t, UTIL_TIME_INITIALIZER, UTIL_clockSpanMicro, UTIL_getTime */
 /* must be included after util.h, due to ERROR macro redefinition issue on Visual Studio */
 #include "zstd_internal.h"  /* ZSTD_WORKSPACETOOLARGE_MAXDURATION, ZSTD_WORKSPACETOOLARGE_FACTOR, KB, MB */
+#include "threading.h"    /* ZSTD_pthread_create, ZSTD_pthread_join */
 
 
 /*-************************************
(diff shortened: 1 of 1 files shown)


tests/fuzzer.c
```diff
@@ -335,6 +336,128 @@ static void FUZ_decodeSequences(BYTE* dst, ZSTD_Sequence* seqs, size_t seqsSize,
...
+    ZSTD_threadPool* pool;
+    void* const CNBuffer;
+    size_t CNBuffSize;
+    void* const compressedBuffer;
+    size_t compressedBufferSize;
+    void* const decodedBuffer;
+    int err;
+} threadPoolTests_compressionJob_payload;
+
+static void* threadPoolTests_compressionJob(void* payload) {
```
Cyan4973, line 351:
```
tests/fuzzer.c:350:14: error: ‘threadPoolTests_compressionJob’ defined but not used [-Werror=unused-function]
  350 | static void* threadPoolTests_compressionJob(void* payload) {
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

May need `ZSTD_MULTITHREAD` too.

terrelln (approved):

tests/fuzzer.c
```diff
@@ -3988,6 +4114,11 @@ int main(int argc, const char** argv)
 
     if (nbTests < testNb) nbTests = testNb;
 
+    result = threadPoolTests();
```
Cyan4973: This invocation of `threadPoolTests()` could be done inside `basicUnitTests()`.
It would make it similar to other unit tests, which can be skipped on demand.

senhuang42: Sure, I've added it to the end of `basicUnitTests()`.

Cyan4973 (approved):

Issue #2043, opened 2020-03-19
Make ZSTDMT_JOBSIZE_MIN runtime configurable


romankarlstetter:
On low-end (but multicore) systems, having `ZSTDMT_JOBSIZE_MIN` set to [1MB](https://github.com/facebook/zstd/blob/66607d0eac54a99946ca14fee6af29fc525275c0/lib/compress/zstdmt_compress.h#L51) might be too large. Even for smaller buffers than 1MB, having multithreading enabled would be beneficial. Of course, I could compile with a lower value for `ZSTDMT_JOBSIZE_MIN`, but that is not very accessible for every user.

So, I thought it might be a good idea to have `ZSTDMT_JOBSIZE_MIN` configurable as a runtime parameter (like number of threads already is configurable), so that I can enable it already at smaller sizes.

As an alternative, it might be good to at least mention in the API documentation for the ZstdCompressor (I'm using the python API via `import zstandard`) that there is the limit of 1MB for parallel compression.

Here is an image illustrating the performance jump on an Intel Atom E3940 @ 1.6GHz (measurement is done via python in a jupyter notebook, compressor is initialized like that: `ZstdCompressor(zstd_level, threads=-1)`).
![2020-03-19 10_50_35-ZSTD compression throughput test](https://user-images.githubusercontent.com/1151415/77054141-ab38e200-69cf-11ea-8280-cc17205efb8e.png)

terrelln:
Related to #2071

Cyan4973:
`jobSize` is already configurable as runtime parameter.

The lower limit `ZSTDMT_JOBSIZE_MIN` has been reduced to 512 KB in #2611.
 It can also be set to any other value at compilation time if need be.

PR #2611, opened 2021-05-04
allow jobSize to be as low as 512 KB

Cyan4973:
answer #2043 .
previous lower limit was 1 MB.

Note : by default, the smallest job size, achieved at level 1, is 2 MB.
So this PR has no impact on multi-threading compression at default settings.
Smaller job sizes can be achieved by manipulating job sizes directly,
or by manually modifying window sizes to lower amounts.

#2043 requested to make the minimum modifiable at runtime,
but it doesn't make sense, since `jobSize` itself is a runtime parameter.
If there is a lower limit, that's because we believe there are side effects introduced when job sizes are too small, and they should be avoided. The value of the lower limit can be negotiated, but its existence should not.

Minimum job size can still be modified at compile time, setting build variable `ZSTDMT_JOBSIZE_MIN`.
This makes it practical for tests, in trying to assess what's a better minimum on a specific system.

Also : updated unit test to ensure that this new limit works fine
(test would fail with previous 1 MB limit).

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1240,9 +1240,8 @@ size_t ZSTDMT_initCStream_internal(
 
     if (params.rsyncable) {
         /* Aim for the targetsectionSize as the average job size. */
-        U32 const jobSizeMB = (U32)(mtctx->targetSectionSize >> 20);
-        U32 const rsyncBits = ZSTD_highbit32(jobSizeMB) + 20;
-        assert(jobSizeMB >= 1);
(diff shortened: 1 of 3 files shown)


terrelln (approved):

terrelln:
Looks like this needs to be fixed:

https://github.com/facebook/zstd/blob/12c045f74d922dc934c168f6e1581d72df983388/lib/compress/zstdmt_compress.c#L1243-L1245

IDK why it converts job size to MB, but it seems like it could just be:

```c
 U32 const rsyncBits = ZSTD_highbit32(mtctx->targetSectionSize);
```

Issue #2552, opened 2021-03-22
zstd does not accept a block device as input


kogorman:
I wanted a complete copy of a hard drive before I tried to fix a problem on it, which may or may not have had to do with any recognizable file.   Zstd refused because it wasn't a "regurlar file".  Of course it wasn't.  So I had to pipe input through dd first.  Not a big issue, but an annoyance.

Running with -8 -T0 on a core i7 it squashed down to 13% taking jsut over an hour working on a 2TB SSD so yippee for that!!

felixhandte:
This will be fixed in the 1.5.0 release. Thanks for the report!

PR #2613, opened 2021-05-04
Allow Reading from Block Devices with `--force`

felixhandte:
Addresses #2552.

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -459,6 +459,22 @@ if [ -n "$DEVNULLRIGHTS" ] ; then
     ls -las $INTOVOID | grep "rw-rw-rw-"
 fi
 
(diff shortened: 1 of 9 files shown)


tests/playTests.sh
```diff
@@ -459,6 +459,19 @@ if [ -n "$DEVNULLRIGHTS" ] ; then
     ls -las $INTOVOID | grep "rw-rw-rw-"
 fi
 
+if [ -n "$READFROMBLOCKDEVICE" ] ; then
```
Cyan4973, line 462: nit: similar to above test,
just add a little comment here
explaining why these tests are conditional, and usage of `sudo`

Cyan4973 (approved):

PR #2547, opened 2021-03-18
Refactor prefetching for the decoding loop

Cyan4973:
Following #2545,
I noticed that one field (`match`) in `seq_t` is optional,
and only used in combination with prefetching.
(This may have contributed to static analyzer failure to detect correct initialization).

I then wondered if it would be possible to rewrite the code
so that this optional part is handled directly by the prefetching code
since it's the only one needing it
rather than delegating a part as an optional workload into a distant `ZSTD_decodeSequence()`.

This resulted into this refactoring exercise
where the prefetching responsibility is better isolated into its own function
and `ZSTD_decodeSequence()` is streamlined to sequence decoding operations strictly.
Incidentally, due to better code locality,
it reduces the need to send information around,
leading to simplified interface, and smaller state structures.

***caveat*** : 
While correctness is fine,
I measured a decompression speed regression
on both my laptop (clang) and desktop (gcc)
(up to -100 MB/s on a i7-9700k @ 5 GHz).

This is unexpected, as this PR preserves the same operations, mostly moving some code around.

That being said, I measured this performance regression even when using the non-prefetching mode,
and even when disabling it entirely (with `-DZSTD_FORCE_DECOMPRESS_SEQUENCES_SHORT`),
which should not make sense since only the "long" prefetching mode has been modified.

This makes me believe that it could be an issue related to instruction alignment,
hence only indirectly related to this PR.

Nonetheless, it makes me worried about merging this PR as is.

A mitigation strategy to harness this instruction alignment issue would be welcome.

@terrelln : this proposed modification may impact the `__asm__(".p2align` strategy or values

_edit_ : actually, the decompression speed difference seems **concentrated on the short (no prefetch) version**. When I force usage of the _long_ variant,  the performance difference becomes much smaller (~3%).

_edit 2_ : measuring cycles spent in DSB (Decoded Stream Buffer) & MITE (Micro-instruction Translation Engine) with `perf`:

Decompressing `enwik9` compressed with `--ultra -22 --long=30`, thus ensuring presence of a lot of long distance matches:

| branch | DSB | MITE | comment |
| --- | --- | --- | --- |
| `dev` | 1.46 G | 2.07 G |
| this PR | 1.81 G | 1.75 G | DSB share improved, yet performance down ~-3% |

Decompressing `enwik9` compressed with `-1`, thus testing the "normal" short-distance decoder

| branch | DSB | MITE | comment |
| --- | --- | --- | --- |
| `dev` | 1.48 G | 1.26 G |
[...]

PR #2614, opened 2021-05-05
faster speed for decompressSequencesLong

Cyan4973:
by using a deeper prefetching pipeline, increased from 4 to 8 slots.

This change substantially improves decompression speed when there are long distance offsets.
example with `enwik9` compressed at level 22 :
gcc-9 : 947 -> 1039 MB/s
clang-10: 884 -> 946 MB/s

I also checked the "cold dictionary" scenario, with `largeNbDicts`,
and found a smaller benefit, around ~2%
(measurements are more noisy for this scenario).

This is a follow up from  #2547,
though it's separate because in this case, the benefits are much more clear cut.

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -1254,9 +1254,9 @@ ZSTD_decompressSequencesLong_body(
 
     /* Regen sequences */
     if (nbSeq) {
-#define STORED_SEQS 4
+#define STORED_SEQS 8
 #define STORED_SEQS_MASK (STORED_SEQS-1)
-#define ADVANCED_SEQS 4
+#define ADVANCED_SEQS STORED_SEQS
         seq_t sequences[STORED_SEQS];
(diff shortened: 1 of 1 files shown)


senhuang42:
What are the scenarios in which this prefetching might be left unused/wasted?

Cyan4973:
> What are the scenarios in which this prefetching might be left unused/wasted?

The decoder only prefetches memory regions
that are effectively going to be copied later for a match operations.

That being said, in some cases, when a match's offset is really small, 
the source region may not yet be filled (at the time prefetching is issued).

That should not matter much because it means this memory region is very fresh, 
hence likely already in L1, and would have been in L1 anyway, even without prefetching.

So we could say that, in this case, the unconditional prefetching was "useless".
However, branching on the speculative presence of a memory region within L1
is way more expensive than merely prefetching unconditionally inside the hot loop.
So it's preferable to always prefetch, even when a memory region is likely already present in L1.

terrelln (approved):

PR #2615, opened 2021-05-05
[lib] Move some ZSTD_CCtx_params off the stack

terrelln:
* Take `params` by const reference in `ZSTD_resetCCtx_internal()`.
* Add `simpleApiParams` to the CCtx and use them in the simple API
  functions, instead of creating those parameters on the stack.
* Set `DEBUGLEVEL=2` by default in `tests/Makefile` to catch compiler
  errors in `DEBUGLOG()` statements.

I think this is a good direction to move in, because we shouldn't need
to worry about adding parameters to `ZSTD_CCtx_params`, since it should
always be on the heap (unless they become absoultely gigantic).

Some `ZSTD_CCtx_params` are still on the stack in the CDict functions,
but I've left them for now, because it was a little more complex, and we
don't use those functions in stack-constrained currently.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -24,7 +24,7 @@ PRGDIR  = ../programs
 PYTHON ?= python3
 TESTARTEFACT := versionsTest
 
-DEBUGLEVEL ?= 1
+DEBUGLEVEL ?= 2
 export DEBUGLEVEL  # transmit value to sub-makefiles
 DEBUGFLAGS  = -g -DDEBUGLEVEL=$(DEBUGLEVEL)
 CPPFLAGS   += -I$(ZSTDDIR) -I$(ZSTDDIR)/common -I$(ZSTDDIR)/compress \
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1802,7 +1802,7 @@ static int ZSTD_dictTooBig(size_t const loadedDictSize)
  * note : `params` are assumed fully validated at this stage.
  */
 static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
-                                      ZSTD_CCtx_params params,
+                                      ZSTD_CCtx_params const* params,
                                       U64 const pledgedSrcSize,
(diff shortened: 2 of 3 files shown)


tests/Makefile
```diff
@@ -24,7 +24,7 @@ PRGDIR  = ../programs
 PYTHON ?= python3
 TESTARTEFACT := versionsTest
 
-DEBUGLEVEL ?= 1
+DEBUGLEVEL ?= 2
```
Cyan4973, line 27: > Set `DEBUGLEVEL=2` by default in `tests/Makefile` to catch compiler errors in `DEBUGLOG()` statements.

There (is ? / used to be ?) a CI test with `DEBUGLEVEL` set to 2 specifically for this purpose (check `DEBUGLOG()` statements).

The question is : 
should checking validity of `DEBUGLOG()` statement be the purpose of a specific CI test,
rather than default for any compilation within `tests/` ?

My concern is that traces from `DEBUGLEVEL = 2` could impact the generated binary layout more than just `assert()`.

terrelln: > There (is ? / used to be ?) a CI test with DEBUGLEVEL set to 2 specifically for this purpose (check DEBUGLOG() statements).

I added this because only Appveyor caught it.

> My concern is that traces from DEBUGLEVEL = 2 could impact the generated binary layout more than just assert().

Does it matter for test cases?

Cyan4973: > Does it matter for test cases?

I'm not sure.
I was concerned that it could make a sensible performance difference,
reducing the nb of fuzzer tests performed within a given timeframe.

It could also be that the difference is not that large.

Cyan4973: So I went ahead and tested the performance difference between `DEBUGLEVEL=1` and `DEBUGLEVEL=2` on a given test workload : `time ./fuzzer -v -i1000 -t1 -s7653`

There was a performance impact, but I believe it's small enough, < 5% (time went up from 22.3s to 23.2s).

Cyan4973 (approved):

Issue #2566, opened 2021-03-30
Questions about training dictionary functions


ghost:
Currently, the stable API has two functions for training dictionary [1]:

- ZDICT_trainFromBuffer()
- ZDICT_finalizeDictionary()

`ZDICT_finalizeDictionary()`'s doc said specifying compression level is helpful for compression [2]:
```
 * The compression level can be set in `parameters`. You should pass the
 * compression level you expect to use in production. The statistics for each
 * compression level differ, so tuning the dictionary for the compression level
 * can help quite a bit.
```

But `ZDICT_trainFromBuffer()` uses `ZSTD_CLEVEL_DEFAULT` as compression level [3], does the trained dictionary have sub-optimal effect for other compression levels?

If this is an API defect, maybe `ZDICT_finalizeDictionary()` function can be allowed to accept `NULL` as basis dictionary. In this case, it only trains the dictionary, doesn't finalize the dictionary. This will not break the API, and it's backward compatible. Then `ZDICT_trainFromBuffer()` function can be marked as obsolete.

Moreover, it would be nice to describe `ZDICT_finalizeDictionary()` function from user's perspective, for example:

- When to use this function?
- Why not re-train a dictionary with the old samples and the new samples together?
- Can the user compose a raw content dictionary by hand, and use it as a basis dictionary? For example, manually select the common parts in some JSON files. 

[1] https://github.com/facebook/zstd/blob/v1.4.9/lib/dictBuilder/zdict.h#L40-L109
[2] https://github.com/facebook/zstd/blob/v1.4.9/lib/dictBuilder/zdict.h#L80-L83
[3] https://github.com/facebook/zstd/blob/v1.4.9/lib/dictBuilder/zdict.c#L1116

ghost:
These notes come from `README.md`, `zdict.h`, format specification, GitHub issues. If put them together (such as a guide), may be it can prevent users from having their misunderstandings.

- If use pre-trained zstd dictionary, the compression ratio achievable on small data (a few KiB) improves dramatically, has best effect on data that smaller than 1 KiB.
- Zstd dictionary has negligible effect on large data (multi-MiB) compression.
- Zstd dictionary is vulnerable.
- Introduce the Dictionary ID, and the reserved ranges for future use.
- A reasonable dictionary has a size of ~100 KiB. It’s possible to select smaller or larger size, just by specifying dict_size argument.
- It’s recommended to provide a few thousands samples, though this can vary a lot.
[...]

PR #2622, opened 2021-05-06
[zdict] Add a FAQ to the top of zdict.h

terrelln:
The FAQ covers the questions asked in Issue #2566. It first covers why
you would want to use a dictionary, then what a dictionary is, and
finally it tells you how to train a dictionary, and clarifies some of
the parameters.

There is definitely more that could be said about some of the advanced
trainers, but this should be a good start.

### DIFF
--- a/lib/zdict.h
+++ b/lib/zdict.h
@@ -36,6 +36,145 @@ extern "C" {
 #  define ZDICTLIB_API ZDICTLIB_VISIBILITY
 #endif
 
+/*******************************************************************************
+ * Zstd dictionary builder
+ *
+ * FAQ
+ * ===
+ * Why should I use a dictionary?
+ * ------------------------------
+ *
+ * Zstd can use dictionaries to improve compression ratio of small data.
+ * Traditionally small files don't compress well because there is very little
+ * repetion in a single sample, since it is small. But, if you are compressing
+ * many similar files, like a bunch of JSON records that share the same
+ * structure, you can train a dictionary on ahead of time on some samples of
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):
Excellent !

lib/zdict.h
```diff
@@ -36,6 +36,146 @@ extern "C" {
...
+ *   # Benchmark levels 1-3 without a dictionary
+ *   zstd -b1e3 -r /path/to/my/files
+ *   # Benchmark levels 1-3 with a dictioanry
+ *   zstd -b1e3 -r /path/to/my/files -D /path/to/my/dictionary
+ *
+ * When should I retrain a dictionary?
+ * -----------------------------------
+ *
+ * You should retrain a dictionary when its effectiveness drops. Dictionary
+ * effectiveness drops the data you are compressing changes. Generally, we do
```
Cyan4973: nit : " ... effectiveness drops __as__ the data you are compressing changes. ..."

lib/zdict.h
```diff
@@ -36,6 +36,146 @@ extern "C" {
...
+ * expect dictionaries to "decay" over time, as your data changes, but the rate
+ * at which they decay depnds on your use case. Internally, we try to retrain
+ * a dictionary every week or so, and if the new dictionary performns
```
Cyan4973: no need to be so precise.
"Internally , we regularly retrain dictionaries, and..."

Cyan4973: nit : performns ->  performs

PR #2633, opened 2021-05-11
Avoid SIGBUS on armv6

### LINKED ISSUE #2632: ARMv6 unaligned hack ends up in bus-error if running under a 64bit kernel
the comment in the source code already points out that the unaligned access violates the C standard.
just for reference if you run this code with a 32bit userspace on a 64bit armv8/aarch64 kernel the unaligned access is no longer caught and handled by the kernel but leads directly to a bus error.
not sure if this could/should make it into the documentation somewhere

bmwiedemann:
When running armv6 userspace on armv8 hardware with a 64 bit Linux kernel,
the mode 2 caused SIGBUS (unaligned memory access).
Running all our arm builds in the build farm
only on armv8 simplifies administration a lot.

Depending on compiler and environment, this change might slow down
memory accesses (did not benchmark it). The original analysis is 6 years old and compilers might have improved since then.

Fixes #2632

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -143,9 +143,7 @@ MEM_STATIC size_t MEM_swapST(size_t in);
  * Prefer these methods in priority order (0 > 1 > 2)
  */
 #ifndef MEM_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

terrelln (approved):

Cyan4973:
Yes, that's a safer default behavior

PR #2637, opened 2021-05-12
fileio: clamp value of windowLog in patch-mode

occivink:
With small enough input files, the inferred value of fileWindowLog could
be smaller than ZSTD_WINDOWLOG_MIN.

This can be reproduced like so:
$ echo abc > small
$ echo abcdef > small2
$ zstd --patch-from small small2 -o patch
previously, this would fail with the error "zstd: error 11 : Parameter is out of bound"


Cyan4973 (approved):

senhuang42 (approved):
Thanks for the fix @occivink !

PR #2643, opened 2021-05-12
reduce ZSTDMT_NBWORKERS_MAX in 32-bit mode

Cyan4973:
now reduced to 64.

Also, restored max limit to 256 when in 64-bit mode,
since it had been reduced to 200 to give more room for 32-bit environments,
which ultimately proved not enough.


senhuang42 (approved):
Looks like this has fixed the previously uasan 32 bit tests as expected!

PR #2645, opened 2021-05-13
Add source level deprecation warning disabling to certain tests/utils

senhuang42:
As title. This adds the macro to `fuzzer`, `zstreamtest`, `fullbench`, and `Pzstd`.

Also adds `thread_dep` target to `fuzzer` in meson build, and `-Wno-deprecated-declarations` to the `pzstd` target in `meson`. For some reason, including the deprecate macro isn't sufficient in `Pzstd.h`

I think this is all of them.


Cyan4973:
> For some reason, including the deprecate macro isn't sufficient in Pzstd.h

That might be because `zstd.h` is included from `Options.h` too ?

senhuang42:
I have verified that this now builds locally. Still, there is the remaining question of what to do with all the warnings `-Wdocumentation -Wcomment`.

Cyan4973 (approved):

PR #2647, opened 2021-05-13
Remove additional flag for Visual

senhuang42:
Now that `ZSTD_DISABLE_DEPRECATE_WARNINGS` is in the source code itself, including them in vcxproj files will cause a `C4005 macro redefinition` warning. So we should remove them to avoid appveyor failures.

Test Plan:
- Appveyor


Cyan4973 (approved):

senhuang42:
Hmm, appveyor has been queued for over an hour, not sure how to force it to run.

Cyan4973:
We do have a time-length pb with Appveyor tests.
When one specific test should be run swiftly, best way is to cancel all prior runs queued before.

senhuang42:
> We do have a time-length pb with Appveyor tests.
> When one specific test should be run swiftly, best way is to cancel all prior runs queued before.

Ah, it seems like I don't have permissions to cancel any appveyor runs, but the ones on just PRs seem to be a little faster.

Cyan4973:
Appveyor tests moved towards d3634bf9

PR #2648, opened 2021-05-13
Add determinism fuzzers and fix rare determinism bugs

terrelln:
See each commit message for details.

* Fix determinism bug in the optimal parser
* Fix off-by-one error in repcode checks (not a determinism problem itself, but inefficient)
* Fix up the dictionary invalidation logic to update `lowLimit` when it is too low for the beginning of the block
* Adds determinism fuzzing to `simple_round_trip` and `dictionary_round_trip`


terrelln:
This is a non-trivial change. It isn't terribly complex, but it is not trivial. If we wanted to merge this into `1.5.0` we'd probably need to re-run the long tests and fuzzers.

I don't think it is terribly critical to merge, since these are longstanding bugs. The only one that might impact us in practice is the dictionary invalidation bug, which might be losing us a few percent of speed.

Cyan4973:
A few percent of speed, for a rare corner case, doesn't feel critical.

Cyan4973 (approved):

PR #2654, opened 2021-05-14
Initialize "potentially uninitialized" pointers.

wolfpld:
In certain configurations MSVC may enable some sort of a static analyser during compilation, which marks "unsafe" code constructs. In this case the error messages are:

```
1>...\zstd\compress\zstd_lazy.c(1399): error C4703: potentially uninitialized local pointer variable 'dmsTagRow' used
1>...\zstd\compress\zstd_lazy.c(1407): error C4703: potentially uninitialized local pointer variable 'dmsRow' used
```

Initialization and usage of both `dmsRow` and `dmsTagRow` is contained within two separate `if (dictMode == ZSTD_dictMatchState)` sections. It is not possible to enter the second (usage) section without entering the first (initialization), but the analyser fails to see this.

This change silences the "error".

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1296,8 +1296,8 @@ size_t ZSTD_RowFindBestMatch_generic (
     size_t ddsIdx;
     U32 ddsExtraAttempts; /* cctx hash tables are limited in searches, but allow extra searches into DDS */
     U32 dmsTag;
-    U32* dmsRow;
-    BYTE* dmsTagRow;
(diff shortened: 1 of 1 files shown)


felixhandte (approved):
This is a reasonable fix for an unfortunate warning.

I worry that zealously fixing these false positives creates room for actually incorrect uses in the future. Initializing these values in this way allows for *syntactically* correct uses which are nonetheless *semantically* incorrect. But I don't think there's really anything we can do about that here.

PR #2655, opened 2021-05-15
[trace] remove zstd_trace.c reference from freestanding

heitbaum:
zstd_trace.c was removed as part of PR #2589


terrelln:
Thanks for the PR @heitbaum!

PR #2657, opened 2021-05-15
ZSTD_VecMask_next: fix incorrect variable name in fallback code path

dnelson-1901:
This is a trivial typo fix found when building zstd with the tinycc compiler (which doesn't claim gcc compatibility and doesn't provide a __builtin_ctz).  The resulting binary passes "gmake check", so even though this codepath has never been used it seems to work :)


Cyan4973 (approved):

Cyan4973:
Thanks @dnelson-1901 !

senhuang42 (approved):

PR #2659, opened 2021-05-16
[lib] Fix libzstd.pc for lib-mt builds

ericonr:
Add the libzstd.pc target to the lib target in lib/Makefile, which makes
it inherit LDFLAGS_DYNLIB from the lib-mt target. This allows us to add
a Libs.private field to libzstd.pc which gets conditionally populated
with '-pthread'.

The 1.5.0 release notes mention that the static library isn't
multi-threaded by default, due to concern for people building static
binaries with libzstd:

   Now the dynamic library supports multi-threaded compression by
   default.  Note that this property is not extended to the static
   library because doing so would have impacted the build script of
   existing client applications (requiring them to add -pthread to their
   recipe), thus potentially breaking their build.

To get closer to being able to enable multi-threading for all library
builds by default, this commit makes it so that any libzstd consumer
using pkg-config gets the correct flags.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -308,7 +308,7 @@ endif  # if windows
 libzstd : $(LIBZSTD)
 
 .PHONY: lib
-lib : libzstd.a libzstd
+lib : libzstd.a libzstd libzstd.pc
 
 
 # note : do not define lib-mt or lib-release as .PHONY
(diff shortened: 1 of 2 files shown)


ericonr:
It would seem something about this broke the mingw builds:

```
Thread model: posix
gcc version 6.3.0 (x86_64-posix-seh-rev1, Built by MinGW-W64 project) 
make[1]: Entering directory '/c/projects/zstd-p0yf0/lib'
make[1]: *** No rule to make target 'libzstd.pc', needed by 'lib'.  Stop.
make[1]: *** Waiting for unfinished jobs....
```

I will try to figure out what it was, would appreciate guidance.

ericonr:
Ok, should be fixed now.

ericonr:
Travis on aarch64 fails due to Travis: https://travis-ci.com/github/facebook/zstd/jobs/516610662

> Automatic restarts limited: Please try restarting this job later or contact support@travis-ci.com.

felixhandte:
@ericonr,

Thanks for this PR! Sorry it took so long for us to get to it. It looks like you're fundamentally doing three things:

1. Adding a field to the generated pkg-config file.
2. Building `libzstd.pc` whenever the library is built.
3. Moving the `libzstd.pc` definition out of the unix-only section.

The first looks correct and important.

The last two are tricky since people may build zstd on platforms where a `.pc` file is meaningless and they may not even have `sed` installed. In those situations, we don't want this dependency.

I'm going to merge this PR into a working branch and make a few fix-ups. I'll tag you when I put it up and if you can make sure it's still addressing your needs, that would be great.

PR #2664, opened 2021-05-17
Fixes to build with MSVC with /W4

cwoffenden:
Updating to 1.5.0 no longer builds with MSVC's warning level 4 due to "potentially uninitialized" local vars. It's a false positive, since each use is wrapped in a test for `if (dictMode == ZSTD_dictMatchState)`. The quick fix is to globally ignore the warnings, silencing these:
```
error C2220: the following warning is treated as an error
warning C4701: potentially uninitialized local variable 'nextSeqStore' used
warning C4701: potentially uninitialized local variable 'ddsIdx' used
warning C4701: potentially uninitialized local variable 'ddsExtraAttempts' used
warning C4701: potentially uninitialized local variable 'dmsTag' used
warning C4701: potentially uninitialized local variable 'dmsTagRow' used
error C4703: potentially uninitialized local pointer variable 'dmsTagRow' used
warning C4701: potentially uninitialized local variable 'dmsRow' used
error C4703: potentially uninitialized local pointer variable 'dmsRow' used
```
It's not a comfortable warning to globally silence so the alternative would be to locally silence them (which is preferable to setting them to anything just to make the compiler happy).


cwoffenden:
This is doing the same as #2654 just in a different way.

felixhandte:
I've just merged the other PR. So I'm going to close this in the expectation that the underlying issue is fixed.

Thanks for the contribution though!

PR #2676, opened 2021-05-21
Optimize ZSTD_decodeSequence for clang

danlark1:
TL;DR; this patch optimizes number of cycles for clang by several percent in non bmi2 mode, e.g. simple compilation mode. I checked on clang-12, clang-11 and clang-9

Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz:

Before:

```
 Performance counter stats for './zstd -t /home/danilak/benchmark_zstd/data/silesia.tar.zst' (50 runs):

            227.71 msec task-clock                #    0.998 CPUs utilized            ( +-  0.29% )
                 1      context-switches          #    0.006 K/sec                    ( +- 14.18% )
                 0      cpu-migrations            #    0.001 K/sec                    ( +- 22.66% )
               656      page-faults               #    0.003 M/sec                    ( +-  3.85% )
       944,476,846      cycles                    #    4.148 GHz                      ( +-  0.07% )
     2,474,680,823      instructions              #    2.62  insn per cycle           ( +-  0.01% )
       261,676,884      branches                  # 1149.176 M/sec                    ( +-  0.03% )
         5,780,728      branch-misses             #    2.21% of all branches          ( +-  0.01% )

          0.228253 +- 0.000743 seconds time elapsed  ( +-  0.33% )
```

After:
```
 Performance counter stats for './zstd -t /home/danilak/benchmark_zstd/data/silesia.tar.zst' (50 runs):

            214.90 msec task-clock                #    0.997 CPUs utilized            ( +-  0.25% )
                 3      context-switches          #    0.013 K/sec                    ( +- 17.17% )
                 1      cpu-migrations            #    0.002 K/sec                    ( +- 19.99% )
               647      page-faults               #    0.003 M/sec                    ( +-  4.05% )
       892,299,745      cycles                    #    4.152 GHz                      ( +-  0.09% )
     2,503,422,821      instructions              #    2.81  insn per cycle           ( +-  0.00% )
       261,614,529      branches                  # 1217.365 M/sec                    ( +-  0.01% )
         5,782,972      branch-misses             #    2.21% of all branches          ( +-  0.01% )

          0.215486 +- 0.000562 seconds time elapsed  ( +-  0.26% )
```

Or, around 5%. GCC is stable and does not change. On server xeon I saw only around 1% performance win.

Why: Unfortunately, as all code is forced inlined, compilers have bad time in finding good opportunities

In current code prevOffset are reassigned in their order and clang generates 

```cpp
seqState->prevOffset[1] = seqState->prevOffset[0];
seqState->prevOffset[0] = offset;
```

3 times, see 3 pictures below.

2021-05-21-021537_1533x216_scrot

2021-05-21-021550_1612x329_scrot

2021-05-21-021559_1575x605_scrot 

After the change, clang manages to find common block 3, 4 and 5 for all 3 branches that end up in reassigning prevOffset[1] and prevOffset[0].

2021-05-21-015511_1525x255_scrot

2021-05-21-015523_1505x732_scrot

2021-05-21-015532_1486x386_scrot

In short, clang managed to do this

qx6oGDvqGchzAw3

I believe this is at least win for some platforms, I can check for other CPUs also but it will take time

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -970,10 +970,8 @@ ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets)
         } else {
             U32 const ll0 = (llBase == 0);
             if (LIKELY((ofBits == 0))) {
-                if (LIKELY(!ll0))
-                    offset = seqState->prevOffset[0];
-                else {
-                    offset = seqState->prevOffset[1];
+                offset = seqState->prevOffset[ll0];
+                if (UNLIKELY(ll0)) {
                     seqState->prevOffset[1] = seqState->prevOffset[0];
                     seqState->prevOffset[0] = offset;
                 }


lib/decompress/zstd_decompress_block.c
```diff
@@ -970,10 +970,8 @@ ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets)
         } else {
             U32 const ll0 = (llBase == 0);
             if (LIKELY((ofBits == 0))) {
-                if (LIKELY(!ll0))
-                    offset = seqState->prevOffset[0];
-                else {
-                    offset = seqState->prevOffset[1];
+                offset = seqState->prevOffset[ll0];
+                if (UNLIKELY(ll0)) {
```
Cyan4973, line 974: That seems incorrect.
If we take this branch, then we should have `offset = seqState->prevOffset[1];`.
This statement has disappeared in the new formulation.
Consequently, I would expect the decompression to fail (corrupted data) whenever this case happens.

Cyan4973:
FYI, early results,
I started a quick series of tests based on this PR,
unfortunately, the outcome is not favorable so far.

The only compiler that seems to register a _tiny_ benefit is `clang-11`,
all other compilers, and versions, get either the same, or slightly worse performance.

This is not final evaluation, I will need more time to properly evaluate.

In any case, whatever the outcome, it was a good suggestion worth trying out.

danlark1:
Yeah, I also noticed that clang-11 benefits the most. And on some processors there is no seen benefit at all.

Anyway, if you think that's not justified, I am ok, I have several other more global ideas but they need time :)

PR #2677, opened 2021-05-21
[CI][2/2] Migrate CI tests which (currently) fail

senhuang42:
PR #2675 is a fairly involved change to the CI. This PR simply adds another commit on top of that which migrates the tests that are failing on #2675 due to various static analyzer warnings. 

The intention is that PR #2675 can get merged first with all tests passing, then the static analyzer-related fixes will get coded and merged, and then this PR can get merged.

No need to review this until #2675 and the static analyzer fixes get merged.

This PR:
- Adds the visual 2019 build test (which didn't previously exist in AppVeyor)
- Moves static analyzer and minimal decompressor macros test from Travis to GH Actions.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -151,6 +151,23 @@ jobs:
           make gcc8install
           CC=gcc-8 CFLAGS="-Werror" make -j all
 
+  visual-2019:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        platform: [x64, Win32]
+        configuration: [Debug, Release]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Add MSBuild to PATH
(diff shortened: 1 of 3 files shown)


terrelln:
Whats the status of this PR?

senhuang42:
> Whats the status of this PR?

I've updated/rebased it so that now all it does is:

1. Add VS2019 test
2. Move minimal decompressor macros test to GH actions
3. Moving static analyzer test to GH actions isn't trivial since there are just way too many false positive warnings that aren't trivial to silence with the latest clang analyzer.

Even just doing (1, 2) requires some extra initializations to satisfy the static analyzers. So if we don't want to do that we can just abandon this PR since now the only actual novel thing it adds is a VS2019 test.

Cyan4973 (approved):

PR #2678, opened 2021-05-21
Fuzzer test with no intrinsics on S390x (big endian)

senhuang42:
Thanks to @aqrit , a bug on big-endian systems with the row-based matchfinder was discovered - the scalar fallback wouldn't work on those systems. We add a test for S390X architectures.

Adds:
- Fuzzer test case for `compress2` with forced row matchfinder for compression ratio regressions
- CI tests for:
  - S390X (big endian) fuzz test with and without intrinsics
  - Fuzz test without intrinsics on a normal LE system.

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -27,6 +27,13 @@ jobs:
     - name: OS-X test
       run: make test # make -c lib all doesn't work because of the fact that it's not a tty
 
+  no-intrinsics-fuzztest:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


senhuang42:
This has been updated to be merge-able after the changes to CI.

Cyan4973:
I see that a few tests on big endian platforms are failing,
but I guess that was precisely the point ?
The fix will be part of another PR ? or added to this one ?

senhuang42:
> The fix will be part of another PR ? or added to this one ?

I've just merged the PR with the fix into dev, and the tests now pass.

Cyan4973 (approved):

PR #2681, opened 2021-05-23
Flatten ZSTD_row_getMatchMask

aqrit:
* Remove the SIMD abstraction layer.
* Add big endian support.
* Align `hashTags` within `tagRow` to a 16-byte boundary. 
* Optimize scalar path using SWAR.
* Optimize neon path for `n == 32`
* Work around minor clang issue for NEON (https://bugs.llvm.org/show_bug.cgi?id=49577)
* Improved endian detection

Benchmark:

tl;dr: compression speed of scalar path increases by about 22%.
Scalar path is now only about 4% slower than the SSE2 path.

```
patch scalar:
 5#silesia.tar       : 211957760 ->  63789770 (3.323), 145.7 MB/s ,1152.4 MB/s 
 6#silesia.tar       : 211957760 ->  62963574 (3.366), 139.3 MB/s ,1180.3 MB/s 
 7#silesia.tar       : 211957760 ->  61467735 (3.448),  95.3 MB/s ,1258.3 MB/s 
 8#silesia.tar       : 211957760 ->  60900152 (3.480),  76.3 MB/s ,1289.6 MB/s 
 9#silesia.tar       : 211957760 ->  59914334 (3.538),  61.7 MB/s ,1303.8 MB/s 
10#silesia.tar       : 211957760 ->  59282317 (3.575),  55.1 MB/s ,1297.1 MB/s 
11#silesia.tar       : 211957760 ->  59140003 (3.584),  51.0 MB/s ,1296.1 MB/s 
12#silesia.tar       : 211957760 ->  58629417 (3.615),  38.6 MB/s ,1315.1 MB/s 
```

```
dev scalar:
 5#silesia.tar       : 211957760 ->  63789770 (3.323), 118.6 MB/s ,1154.1 MB/s 
 6#silesia.tar       : 211957760 ->  62963574 (3.366), 114.7 MB/s ,1179.7 MB/s 
 7#silesia.tar       : 211957760 ->  61467735 (3.448),  76.8 MB/s ,1257.8 MB/s 
 8#silesia.tar       : 211957760 ->  60900152 (3.480),  61.1 MB/s ,1289.0 MB/s 
 9#silesia.tar       : 211957760 ->  59914334 (3.538),  51.6 MB/s ,1303.2 MB/s 
10#silesia.tar       : 211957760 ->  59282317 (3.575),  47.3 MB/s ,1296.3 MB/s 
11#silesia.tar       : 211957760 ->  59140003 (3.584),  43.6 MB/s ,1294.8 MB/s 
12#silesia.tar       : 211957760 ->  58629417 (3.615),  31.7 MB/s ,1315.6 MB/s 

```

```
patch SSE2:
 5#silesia.tar       : 211957760 ->  63789770 (3.323), 151.3 MB/s ,1148.2 MB/s 
 6#silesia.tar       : 211957760 ->  62963574 (3.366), 143.4 MB/s ,1169.3 MB/s 
 7#silesia.tar       : 211957760 ->  61467735 (3.448), 100.8 MB/s ,1247.6 MB/s 
 8#silesia.tar       : 211957760 ->  60900152 (3.480),  81.5 MB/s ,1279.4 MB/s 
 9#silesia.tar       : 211957760 ->  59914334 (3.538),  65.5 MB/s ,1292.1 MB/s 
10#silesia.tar       : 211957760 ->  59282317 (3.575),  57.2 MB/s ,1284.2 MB/s 
11#silesia.tar       : 211957760 ->  59140003 (3.584),  52.3 MB/s ,1282.9 MB/s 
12#silesia.tar       : 211957760 ->  58629417 (3.615),  39.7 MB/s ,1301.6 MB/s
```

```
dev SSE2:
 5#silesia.tar       : 211957760 ->  63789770 (3.323), 151.1 MB/s ,1148.5 MB/s 
 6#silesia.tar       : 211957760 ->  62963574 (3.366), 143.8 MB/s ,1171.8 MB/s 
 7#silesia.tar       : 211957760 ->  61467735 (3.448),  98.2 MB/s ,1250.2 MB/s 
 8#silesia.tar       : 211957760 ->  60900152 (3.480),  79.3 MB/s ,1281.9 MB/s 
 9#silesia.tar       : 211957760 ->  59914334 (3.538),  63.1 MB/s ,1294.3 MB/s 
10#silesia.tar       : 211957760 ->  59282317 (3.575),  56.8 MB/s ,1286.5 MB/s 
11#silesia.tar       : 211957760 ->  59140003 (3.584),  51.9 MB/s ,1285.7 MB/s 
12#silesia.tar       : 211957760 ->  58629417 (3.615),  40.2 MB/s ,1304.4 MB/s 
```

benched on my desktop `i7-8700K` using `gcc version 9.3.0`

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1096,10 +1096,10 @@ ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, con
             const uint8x16_t chunk1 = vreinterpretq_u8_u16(chunk.val[1]);
             const uint8x16_t equalMask0 = vceqq_u8(chunk0, vdupq_n_u8(tag));
             const uint8x16_t equalMask1 = vceqq_u8(chunk1, vdupq_n_u8(tag));
-            const uint16x4_t pack0 = vqmovn_s16(vreinterpretq_s16_u8(equalMask0));
-            const uint16x4_t pack1 = vqmovn_s16(vreinterpretq_s16_u8(equalMask1));
-            const uint8x8_t t0 = vreinterpret_u8_s16(pack0);
-            const uint8x8_t t1 = vreinterpret_u8_s16(pack1);
+            const int8x8_t pack0 = vqmovn_s16(vreinterpretq_s16_u8(equalMask0));
+            const int8x8_t pack1 = vqmovn_s16(vreinterpretq_s16_u8(equalMask1));
+            const uint8x8_t t0 = vreinterpret_u8_s8(pack0);
+            const uint8x8_t t1 = vreinterpret_u8_s8(pack1);
             const uint8x8_t t2 = vsri_n_u8(t1, t0, 2);
             const uint8x8x2_t t3 = vuzp_u8(t2, t0);
             const uint8x8_t t4 = vsri_n_u8(t3.val[1], t3.val[0], 4);
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -902,14 +902,14 @@ FORCE_INLINE_TEMPLATE
 U32 ZSTD_rotateRight_U32(U32 const value, U32 count) {
     assert(count < 32);
     count &= 0x1F; /* for fickle pattern recognition */
-    return (value >> count) | (U32)(value << (-(S32)count & 0x1F));
+    return (value >> count) | (U32)(value << ((0U - count) & 0x1F));
 }
 
 FORCE_INLINE_TEMPLATE
 U16 ZSTD_rotateRight_U16(U16 const value, U32 count) {
     assert(count < 16);
     count &= 0x0F; /* for fickle pattern recognition */
-    return (value >> count) | (U16)(value << (-(S16)count & 0x0F));
+    return (value >> count) | (U16)(value << ((0U - count) & 0x0F));
 }
 
 /* ZSTD_row_nextIndex():
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1045,7 +1045,7 @@ ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, con
     if (rowEntries == 16) {
         const __m128i chunk = _mm_loadu_si128((const __m128i*)(const void*)src);
         const __m128i equalMask = _mm_cmpeq_epi8(chunk, _mm_set1_epi8(tag));
-        const U32 matches = (U32)_mm_movemask_epi8(equalMask);
+        const U16 matches = (U16)_mm_movemask_epi8(equalMask);
         return ZSTD_rotateRight_U16(matches, head);
     } else { /* rowEntries == 32 */
         const __m128i chunk0 = _mm_loadu_si128((const __m128i*)(const void*)&src[0]);
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -871,22 +871,6 @@ FORCE_INLINE_TEMPLATE size_t ZSTD_HcFindBestMatch_extDict_selectMLS (
 
 #define ZSTD_ROW_HASH_CACHE_MASK (ZSTD_ROW_HASH_CACHE_SIZE - 1)
 
-#if !defined(ZSTD_NO_INTRINSICS)
-#  if defined(__SSE2__) || defined(_M_AMD64) || (defined (_M_IX86) && defined(_M_IX86_FP) && (_M_IX86_FP >= 2))
-#    define ZSTD_ARCH_X86_SSE2
-#  endif
-#  if defined(__ARM_NEON)
-#    define ZSTD_ARCH_ARM_NEON
-#  endif
-#
-#
-#  if defined(ZSTD_ARCH_X86_SSE2)
-#    include <emmintrin.h>
-#  elif defined(ZSTD_ARCH_ARM_NEON)
-#    include <arm_neon.h>
-#  endif
-#endif
-
 typedef U32 ZSTD_VecMask;   /* Clarifies when we are interacting with a U32 representing a mask of matches */
 
 /* ZSTD_VecMask_next():
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1121,8 +1121,7 @@ ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, con
         if (MEM_isLittleEndian()) { /* runtime check so have two loops */
             const size_t extractMagic = (xFF / 0x7F) >> chunkSize;
             do {
-                size_t chunk;
-                memcpy(&chunk, &src[i], chunkSize);
+                size_t chunk = MEM_readST(&src[i]);
                 chunk ^= splatChar;
                 chunk = (((chunk | x80) - x01) | chunk) & x80;
                 matches <<= chunkSize;
@@ -1133,8 +1132,7 @@ ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, con
             const size_t msb = xFF ^ (xFF >> 1);
             const size_t extractMagic = (msb / 0x1FF) | msb;
             do {
-                size_t chunk;
-                memcpy(&chunk, &src[i], chunkSize);
+                size_t chunk = MEM_readST(&src[i]);
                 chunk ^= splatChar;
                 chunk = (((chunk | x80) - x01) | chunk) & x80;
                 matches <<= chunkSize;
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1060,14 +1060,15 @@ FORCE_INLINE_TEMPLATE
 ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, const U32 head, const U32 rowEntries) {
     const BYTE* const src = tagRow + ZSTD_ROW_HASH_TAG_OFFSET;
 #if defined(ZSTD_ARCH_X86_SSE2)
+    assert(ZSTD_isAligned(src, 16));
     if (rowEntries == 16) {
-        const __m128i chunk = _mm_load_si128((const __m128i*)src);
+        const __m128i chunk = _mm_load_si128((const __m128i*)(const void*)src);
         const __m128i equalMask = _mm_cmpeq_epi8(chunk, _mm_set1_epi8(tag));
         const U32 matches = (U32)_mm_movemask_epi8(equalMask);
         return ZSTD_VecMask_rotateRight(matches, head, 16);
     } else {
-        const __m128i chunk0 = _mm_load_si128((const __m128i*)&src[0]);
-        const __m128i chunk1 = _mm_load_si128((const __m128i*)&src[16]);
+        const __m128i chunk0 = _mm_load_si128((const __m128i*)(const void*)&src[0]);
+        const __m128i chunk1 = _mm_load_si128((const __m128i*)(const void*)&src[16]);
         const __m128i equalMask0 = _mm_cmpeq_epi8(chunk0, _mm_set1_epi8(tag));
         const __m128i equalMask1 = _mm_cmpeq_epi8(chunk1, _mm_set1_epi8(tag));
         const U32 lo = (U32)_mm_movemask_epi8(equalMask0);
@@ -1078,6 +1079,7 @@ ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, con
 #else
 #  if defined(ZSTD_ARCH_ARM_NEON)
     if (MEM_isLittleEndian()) {
+        assert(ZSTD_isAligned(src, 16));
         if (rowEntries == 16) {
             const uint8x16_t chunk = vld1q_u8(src);
             const uint16x8_t equalMask = vreinterpretq_u16_u8(vceqq_u8(chunk, vdupq_n_u8(tag)));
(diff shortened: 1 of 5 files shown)


aqrit:
AFAICT, the test failures are all: 
`error: cast from 'const BYTE *' (aka 'const unsigned char *') to 'const __m128i *' increases required alignment from 1 to 16 [-Werror,-Wcast-align]
        const __m128i chunk = _mm_load_si128((const __m128i*)&src[0]);`

which leaves me mystified on how any SIMD ever passed these tests...
Local tests with gcc `-Wcast-align=strict` errors out with unaligned loads and typedefs.

Do we need (or want) `__attribute__((may_alias))` or `__attribute__((aligned(16)))`, reportedly there are issues on MSVC as well?

lib/compress/zstd_lazy.c
```diff
@@ -865,204 +865,32 @@ FORCE_INLINE_TEMPLATE size_t ZSTD_HcFindBestMatch_extDict_selectMLS (
...
-        for (; j < sizeof(size_t); ++j, ++l) {
-            if (((cmp1 >> j*8) & 0xFF) == ((cmp2 >> j*8) & 0xFF)) {
-                res |= ((U32)1 << (j+i*sizeof(size_t)));
-            }
-        }
-    }
-    return res;
-}
+#if !defined(ZSTD_NO_INTRINSICS)
+#  if defined(_M_AMD64) || (defined (_M_IX86) && defined(_M_IX86_FP) && (_M_IX86_FP >= 2))
```
senhuang42: There are other places that depend on these series of defines: https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L225

It would be best to consolidate and then define these macros in `zstd_internal.h`, where `zstd_lazy.c` and `zstd_compress.c` could both access it.

Also, I believe the `defined(__SSE2__)` could be included in this statement as well.

lib/compress/zstd_lazy.c
```diff
@@ -1226,24 +1054,103 @@ void ZSTD_row_update(ZSTD_matchState_t* const ms, const BYTE* ip) {
 
 /* Returns a ZSTD_VecMask (U32) that has the nth bit set to 1 if the newly-computed "tag" matches
  * the hash at the nth position in a row of the tagTable.
- */
+ * Each row is a circular buffer beginning at the value of "head". So we must rotate the "matches" bitfield
+ * to match up with the actual layout of the entries within the hashTable */
 FORCE_INLINE_TEMPLATE
 ZSTD_VecMask ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, const U32 head, const U32 rowEntries) {
-    ZSTD_VecMask matches = 0;
+    const BYTE* const src = tagRow + ZSTD_ROW_HASH_TAG_OFFSET;
```
senhuang42, line 1042: Personally, I prefer retaining the SIMD abstraction layer, so that this `getMatchMask()` function is easier to understand without having to know what the API for various types of SIMD instructions. Is there a particular reason you want to flatten it out?

lib/compress/zstd_lazy.c
```diff
@@ -1226,24 +1054,103 @@ void ZSTD_row_update(ZSTD_matchState_t* const ms, const BYTE* ip) {
...
-        ZSTD_Vec128 expandedTags  = ZSTD_Vec128_set8(tag);
-        matches                   = ZSTD_Vec128_cmpMask8(hashes, expandedTags);
-    } else if (rowEntries == 32) {
-        ZSTD_Vec256 hashes        = ZSTD_Vec256_read(tagRow + ZSTD_ROW_HASH_TAG_OFFSET);
-        ZSTD_Vec256 expandedTags  = ZSTD_Vec256_set8(tag);
-        matches                   = ZSTD_Vec256_cmpMask8(hashes, expandedTags);
+        const __m128i chunk = _mm_load_si128((const __m128i*)(const void*)src);
+        const __m128i equalMask = _mm_cmpeq_epi8(chunk, _mm_set1_epi8(tag));
+        const U32 matches = (U32)_mm_movemask_epi8(equalMask);
+        return ZSTD_VecMask_rotateRight(matches, head, 16);
```
senhuang42: Nit: `return ZSTD_VecMask_rotateRight(matches, head, rowEntries);` instead, rather than using a nameless constant. It should be neutral in performance since it'll get inlined anyways.

lib/compress/zstd_lazy.c
```diff
@@ -1226,24 +1054,103 @@ void ZSTD_row_update(ZSTD_matchState_t* const ms, const BYTE* ip) {
...
     } else {
-        assert(0);
+        const __m128i chunk0 = _mm_load_si128((const __m128i*)(const void*)&src[0]);
+        const __m128i chunk1 = _mm_load_si128((const __m128i*)(const void*)&src[16]);
+        const __m128i equalMask0 = _mm_cmpeq_epi8(chunk0, _mm_set1_epi8(tag));
+        const __m128i equalMask1 = _mm_cmpeq_epi8(chunk1, _mm_set1_epi8(tag));
+        const U32 lo = (U32)_mm_movemask_epi8(equalMask0);
+        const U32 hi = (U32)_mm_movemask_epi8(equalMask1);
+        assert(rowEntries == 32);
+        return ZSTD_VecMask_rotateRight((hi << 16) | lo, head, 32);
```
senhuang42: Nit: within the zstd codebase, we prefer to use the
```
if (rowEntries == 16) { ... }
else if (rowEntries == 32 { ... }
else { assert(0); }
```
To basically say that we cannot have another value. We could also `assert(rowEntries == 16 || rowEntries == 32)` as well.

lib/compress/zstd_lazy.c
```diff
@@ -1226,24 +1054,103 @@ void ZSTD_row_update(ZSTD_matchState_t* const ms, const BYTE* ip) {
...
     }
-    /* Each row is a circular buffer beginning at the value of "head". So we must rotate the "matches" bitfield
-        to match up with the actual layout of the entries within the hashTable */
-    return ZSTD_VecMask_rotateRight(matches, head, rowEntries);
+#else
+#  if defined(ZSTD_ARCH_ARM_NEON)
+    if (MEM_isLittleEndian()) {
+        assert(ZSTD_isAligned(src, 16));
+        if (rowEntries == 16) {
```
senhuang42, line 1062: awesome 👍

lib/compress/zstd_lazy.c
```diff
@@ -1226,24 +1054,103 @@ void ZSTD_row_update(ZSTD_matchState_t* const ms, const BYTE* ip) {
...
+                chunk ^= splatChar;
+                chunk = (((chunk | x80) - x01) | chunk) & x80;
+                matches <<= chunkSize;
+                matches |= ((chunk >> 7) * extractMagic) >> shiftAmount;
+                i -= chunkSize;
+            } while (i >= 0);
+        }
+        matches = ~matches;
+        if (rowEntries == 16) {
+            return ZSTD_VecMask_rotateRight((U16)matches, head, 16);
```
senhuang42: Nit: return `ZSTD_VecMask_rotateRight((U16)matches, head, rowEntries);`

lib/compress/zstd_lazy.c
```diff
@@ -865,204 +865,32 @@ FORCE_INLINE_TEMPLATE size_t ZSTD_HcFindBestMatch_extDict_selectMLS (
 * (SIMD) Row-based matchfinder
 ***********************************/
 /* Constants for row-based hash */
-#define ZSTD_ROW_HASH_TAG_OFFSET 1                               /* byte offset of hashes in the match state's tagTable from the beginning of a row */
+#define ZSTD_ROW_HASH_TAG_OFFSET 16                              /* byte offset of hashes in the match state's tagTable from the beginning of a row */
```
senhuang42, line 868: I've done some quick benchmarks with 16-byte aligned reads against unaligned - unfortunately I didn't measure any performance benefit.

senhuang42 (commented):
Thanks for these contributions!! These NEON and scalar optimizations look great - I've left some comments that mostly have to do with style/code structure. I'll get around to benchmarking the improvements scalar mode at some point as well.

lib/compress/zstd_lazy.c
aqrit: AFAIK  --- For this particular access pattern, aligned loads are pointless on modern CPUs.  Though, who knows what strange CPUs will use the scalar path (ARMv5 or below)? If I understand correctly, an AMD Jaguar (PS4 or Xbox One) using an align 16-byte load at a 16-byte boundary is 3 cycles faster.

lib/compress/zstd_lazy.c
aqrit: I don't like that pattern. For example, the default case in `ZSTD_VecMask_rotateRight` is force to use an extra `test & cmov` because of this.

lib/compress/zstd_lazy.c
aqrit: I started to look into this and found that there are 180 copies of this function :-)

lib/compress/zstd_lazy.c
aqrit: I wished to use a different strategy for each instruction set. I believe the scalar version will take a performance hit if it is forced to emulate the SSE2 version.

aqrit:
Amusingly, NEON gets cheaper when `rowEntries == 64`

```
uint64_t NEON_i8x64_MatchMask (const uint8_t* ptr, uint8_t match_byte) {
    uint8x16x4_t src = vld4q_u8(ptr);
    uint8x16_t dup = vdupq_n_u8(match_byte);
    uint8x16_t cmp0 = vceqq_u8(src.val[0], dup);
    uint8x16_t cmp1 = vceqq_u8(src.val[1], dup);
    uint8x16_t cmp2 = vceqq_u8(src.val[2], dup);
    uint8x16_t cmp3 = vceqq_u8(src.val[3], dup);

    uint8x16_t t0 = vsriq_n_u8(cmp1, cmp0, 1);
    uint8x16_t t1 = vsriq_n_u8(cmp3, cmp2, 1);
    uint8x16_t t2 = vsriq_n_u8(t1, t0, 2);
    uint8x16_t t3 = vsriq_n_u8(t2, t2, 4);
    uint8x8_t t4 = vshrn_n_u16(vreinterpretq_u16_u8(t3), 4);
    return vget_lane_u64(vreinterpret_u64_u8(t4), 0);
}
```

senhuang42:
> Amusingly, NEON gets cheaper when `rowEntries == 64`
> 
> ```
> uint64_t NEON_i8x64_MatchMask (const uint8_t* ptr, uint8_t match_byte) {
>     uint8x16x4_t src = vld4q_u8(ptr);
>     uint8x16_t dup = vdupq_n_u8(match_byte);
>     uint8x16_t cmp0 = vceqq_u8(src.val[0], dup);
>     uint8x16_t cmp1 = vceqq_u8(src.val[1], dup);
>     uint8x16_t cmp2 = vceqq_u8(src.val[2], dup);
>     uint8x16_t cmp3 = vceqq_u8(src.val[3], dup);
> 
>     uint8x16_t t0 = vsriq_n_u8(cmp1, cmp0, 1);
>     uint8x16_t t1 = vsriq_n_u8(cmp3, cmp2, 1);
>     uint8x16_t t2 = vsriq_n_u8(t1, t0, 2);
>     uint8x16_t t3 = vsriq_n_u8(t2, t2, 4);
>     uint8x8_t t4 = vshrn_n_u16(vreinterpretq_u16_u8(t3), 4);
>     return vget_lane_u64(vreinterpret_u64_u8(t4), 0);
> }
> ```

That's actually good to know - we're going to also add support for 64 row entries as well.

senhuang42 (approved):
This looks good to me! @terrelln do you have any thoughts?

senhuang42:
Thanks @aqrit for this great contribution 🚀

Cyan4973:
Excellent work @aqrit !

lib/compress/zstd_lazy.c
```diff
@@ -1226,24 +1035,98 @@ void ZSTD_row_update(ZSTD_matchState_t* const ms, const BYTE* ip) {
...
+            const uint8x16_t chunk = vld1q_u8(src);
+            const uint16x8_t equalMask = vreinterpretq_u16_u8(vceqq_u8(chunk, vdupq_n_u8(tag)));
+            const uint16x8_t t0 = vshlq_n_u16(equalMask, 7);
+            const uint32x4_t t1 = vreinterpretq_u32_u16(vsriq_n_u16(t0, t0, 14));
+            const uint64x2_t t2 = vreinterpretq_u64_u32(vshrq_n_u32(t1, 14));
+            const uint8x16_t t3 = vreinterpretq_u8_u64(vsraq_n_u64(t2, t2, 28));
+            const U16 hi = (U16)vgetq_lane_u8(t3, 8);
+            const U16 lo = (U16)vgetq_lane_u8(t3, 0);
+            return ZSTD_rotateRight_U16((hi << 8) | lo, head);
+        } else { /* rowEntries == 32 */
```
Cyan4973, line 1072: @aqrit :
This `NEON` code must be designed specifically for each `rowEntries`.
Hence, we have an implementation for 16, for 32 and then for 64.
If we would like to introduce 128 or even 256, it's necessary to write another implementation.

Ignoring the issue of generating the final mask (which width directly depends on the nb of entries),
does that seem possible to rewrite the main test loop in a way which would be generic,
and scale naturally with `rowEntries` (provided it's a power of 2) ?

For an example of what I mean by "an implementation which can scale with `rowEntries`",
please have a look at : https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_lazy.c#L992 ,
which uses `SSE2`.

aqrit: 128/256 would just loop the 64 byte path.

I've done no benchmarking but... 
it seems really painful to emulate `pmovmskb` using ARMv7-A NEON.
AArch64 added the SIMD instruction `addv` which might be useful for simplifying all this.

On AArch64, is the ARMv7-A path actually faster than the scalar fallback?

PR #2684, opened 2021-05-24
Add ldm and block splitter auto-enable to old api

senhuang42:
This PR adds the auto-force-enabled ldm/block splitter to the old API, as well as the new one, and updates regressiontest.


terrelln (approved):
LGTM once you get the tests to pass. Maybe you just need a re-run?

PR #2685, opened 2021-05-26
MinGW Build Fixes

JPEWdev:
Fixes building on MinGW by
 1) Exporting the OS variable so that it is visible to sub-make
    invocations
 2) Fixing the extension handling so that make correctly identifies when
    targets need to be rebuilt. Without the correct handling, make would
    rebuild the executable targets when running `make install` because
    it couldn't find them with the correct extension.

Signed-off-by: Joshua Watt <JPEWhacker@gmail.com>


Cyan4973:
Superseded by #2687

PR #2691, opened 2021-06-01
Make regression test run on every PR

senhuang42:
As title

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,21 +84,10 @@ workflows:
   commit:
     jobs:
       # Run the tests in parallel
-      - short-tests-0:
-          filters:
-            tags:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -92,13 +92,11 @@ workflows:
...
               only: /.*/
-      # Create a branch called regression and set it to dev to force a
-      # regression test run
       - regression-test:
           filters:
-            branches:
-              only:
-                - regression
+            tags:
+              only: /.*/
```
Cyan4973: question : what does mean `only : /.*/` ?
Would it be simpler and equivalent to not have this line ?

Cyan4973 (approved):

Issue #2694, opened 2021-06-04
Compression results are hard to read if numbers are very large


scottchiefbaker:
**Is your feature request related to a problem? Please describe.**

After compressing a file in the CLI some summary text is outputted listing the size before and after compression. If the file is sufficiently large it's **very** hard to determine the file size visually. This is because the size is reported in bytes. A file that is 200 megabytes has nine digits, and a file that is 20 megabyte has eight digits. Files with large sizes like this are very hard to compare.

**Describe the solution you'd like**

Instead of reporting the file size in bytes, instead report in a more [human readable](https://www.perturb.org/display/810_Perlfunc_human_size_.html) string format.

##### Current:
```
:zstd input.bin
input.bin            : 45.57%   (200836721 => 91519541 bytes, input.bin.zst)
```

##### Proposed:
```
:zstd input.bin
input.bin            : 45.57%   (199.1M => 89.4M bytes, input.bin.zst)
```

**Describe alternatives you've considered**

Other than training myself to visually compare 10 digit numbers (unlikely) I don't see an alternative.

PR #2696, opened 2021-06-05
Make the CLI summary data use a human readable format instead of raw bytes

scottchiefbaker:
This PR changes the CLI output summary from bytes to a more human readable string format. This should address issue #2694.

**Before:**
```
:zstd -f /tmp/foo.bin 
/tmp/foo.bin         :  2.94%   (280708256 => 8264707 bytes, /tmp/foo.bin.zst) 
```

**After:**
```
:./zstd -f /tmp/foo.bin 
/tmp/foo.bin         :  2.94%   (267.7M => 7.9M, /tmp/foo.bin.zst)      
```

This PR adds a `humanSize()` function to `util.c`. I wasn't sure where best to put it, but that seemed like a good place. Then it modifies the `sprintf()` calls for outputting the summary after a file is compressed. It also modifies the `sprintf()` benchmark code.

Things I checked before landing this code:

1. Code is space indented not tabs
2. Functions and variables are camelCase
3. Variable declarations are grouped with other variables
3. Code should be C89 compatible
4. Make test runs without errors on x86_64
5. Code compiles and outputs the correct format for summary, benchmark, and multiple files

I will probably need some help modifying this to work on ARM or other platforms.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -121,8 +121,14 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
 *  Functions
 ***************************************/
 
+/*
+ * Take a size in bytes and output a human readable string. Maximum
+ * buffer size is 8 but it's usually 7. Example: "123.4G"
+*/
 char* humanSize(unsigned long long size, char* str) {
-    if (size > 1125899906842624L) {
+    if (size > 1152921504606846976L) {
+        snprintf(str, 7, "%.1fE", (float)size / 1152921504606846976L);
+	} else if (size > 1125899906842624L) {
         snprintf(str, 7, "%.1fP", (float)size / 1125899906842624L);
     } else if (size > 1099511627776L) {
         snprintf(str, 7, "%.1fT", (float)size / 1099511627776L);
@@ -134,8 +140,6 @@ char* humanSize(unsigned long long size, char* str) {
         snprintf(str, 7, "%.1fK", (float)size / 1024);
     } else if (size <= 1024) {
         snprintf(str, 7, "%lluB", size);
-    } else {
-        str[0] = '\0';
     }
 
     return str;
--- a/programs/util.c
+++ b/programs/util.c
@@ -139,7 +139,7 @@ char* humanSize(unsigned long long size, char* str) {
     } else if (size > 1024) {
         snprintf(str, 7, "%.1fK", (float)size / 1024);
     } else if (size <= 1024) {
-        snprintf(str, 7, "%lluB", size);
+        snprintf(str, 7, "%uB", (unsigned)size);
     }
 
     return str;
--- a/programs/util.c
+++ b/programs/util.c
@@ -121,7 +121,7 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
 *  Functions
 ***************************************/
 
-char* human_size(long size, char* str) {
+char* humanSize(long size, char* str) {
 	if (size > 1125899906842624L) {
 		snprintf(str, 7, "%.1fP", (float)size / 1125899906842624L);
 	} else if (size > 1099511627776L) {
--- a/programs/util.c
+++ b/programs/util.c
@@ -122,27 +122,23 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
 ***************************************/
 
 char* humanSize(unsigned long long size, char* str) {
-	/* This only works on 64 bit platforms so I commented it out for now */
-	/*
-	if (size > 1125899906842624L) {
-		snprintf(str, 7, "%.1fP", (float)size / 1125899906842624L);
-	} else if (size > 1099511627776L) {
-		snprintf(str, 7, "%.1fT", (float)size / 1099511627776L);
-	*/
+    if (size > 1125899906842624L) {
+        snprintf(str, 7, "%.1fP", (float)size / 1125899906842624L);
+    } else if (size > 1099511627776L) {
+        snprintf(str, 7, "%.1fT", (float)size / 1099511627776L);
+    } else if (size > 1073741824L) {
+        snprintf(str, 7, "%.1fG", (float)size / 1073741824L);
+    } else if (size > 1048576L) {
+        snprintf(str, 7, "%.1fM", (float)size / 1048576L);
+    } else if (size > 1024) {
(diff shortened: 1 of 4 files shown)


programs/util.c
```diff
@@ -121,6 +121,27 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
...
+    } else if (size > 1073741824L) {
+        snprintf(str, 7, "%.1fG", (float)size / 1073741824L);
+    } else if (size > 1048576L) {
+        snprintf(str, 7, "%.1fM", (float)size / 1048576L);
+    } else if (size > 1024) {
+        snprintf(str, 7, "%.1fK", (float)size / 1024);
+    } else if (size <= 1024) {
+        snprintf(str, 7, "%lluB", size);
+    } else {
+        str[0] = '\0';
```
felixhandte: This should be impossible.

programs/util.c
```diff
@@ -121,6 +121,27 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
 *  Functions
 ***************************************/
 
+char* humanSize(unsigned long long size, char* str) {
+    if (size > 1125899906842624L) {
+        snprintf(str, 7, "%.1fP", (float)size / 1125899906842624L);
```
felixhandte, line 132: `snprintf` was standardized in C99, so it's not guaranteed to be available in all the environments we want Zstd to build.

You could consider rewriting this to not require an intermediate buffer if you just output a rescaled float value and a (static) suffix string that are then passed to the `DISPLAYLEVEL()` call with `"%.1f%s"`. But maybe that would lead to poor results for small inputs ("you compressed 123.0B" feels weird).

A worst case alternative is that you just wrap this in an `#if POSIX_PLATFORM_VERSION >= 200112L` and provide a fallback `#else` implementation.

programs/util.c
```diff
@@ -121,6 +121,27 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
 *  Functions
 ***************************************/
 
+char* humanSize(unsigned long long size, char* str) {
+    if (size > 1125899906842624L) {
```
felixhandte: For the sake of completeness, include a case for exabytes? Otherwise you could theoretically end up with a value like "12345.6P" instead of the preferred "12.3E".

programs/util.h
```diff
@@ -122,6 +122,8 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg, const
 #define STRDUP(s) strdup(s)
 #endif
 
+char* humanSize(unsigned long long size, char* str);
```
felixhandte, line 129: Switch to a more descriptive name? Maybe add a little comment here describing what it does?

felixhandte (changes requested):
Hey @scottchiefbaker,

Thanks for this PR.

This makes sense to me as a nicer way to present this information. I guess there's a question of whether changing this format will break anyone trying to parse this output, but I can't really imagine why anyone would be doing that in the first place.

I included a few suggestions. Let me know what you think.

programs/util.c
scottchiefbaker: When this started it accepted negative values, and this used to catch those. I can remove it for completeness.

programs/util.c
scottchiefbaker: This is an easy add... I had to double check if 64 bit int will hold and exabyte... it does.

programs/util.h
scottchiefbaker: I'm open to naming suggestions. I will definitely add a description though, that's easy.

programs/util.c
scottchiefbaker: I see references to `fprintf()`all over the code base so I just assumed `snprintf()` was valid. `util.c.` includes `<stdio.h>` for `fprintf()`, can we rely on that, but not `snprintf()`?

C is not my forte, so I may need some guidance on this one.

scottchiefbaker:
I'm not a C expert, so I will need some help resolving the [one failing test](https://github.com/facebook/zstd/runs/2753939883).

programs/util.c
```diff
@@ -121,6 +121,31 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
...
+    } else if (size > 1099511627776L) {
+        snprintf(str, 7, "%.1fT", (float)size / 1099511627776L);
+    } else if (size > 1073741824L) {
+        snprintf(str, 7, "%.1fG", (float)size / 1073741824L);
+    } else if (size > 1048576L) {
+        snprintf(str, 7, "%.1fM", (float)size / 1048576L);
+    } else if (size > 1024) {
+        snprintf(str, 7, "%.1fK", (float)size / 1024);
+    } else if (size <= 1024) {
+        snprintf(str, 7, "%lluB", size);
```
aqrit: There are a bunch of portability issues with passing `long long` to `printf`. We know `size <= 1024` so I would just cast it to unsigned. `snprintf(str, 7, "%uB", (unsigned)size);`

scottchiefbaker: That did it! Looks like all the tests pass now.

You're a gentlemen and a scholar sir!

programs/util.c
felixhandte: As C has evolved, they've introduced new functions into the same header. C90 has `fprintf()` but not necessarily `snprintf()`. `snprintf()` was only standardized (and guaranteed to be present) in C99. Zstd restricts itself to the features guaranteed to be present in C90 (or at least has fallbacks to handle that strict case), so that we can be sure it will compile on *all* C90-conforming platforms.

programs/util.h
felixhandte: Maybe `UTIL_renderHumanReadableByteSize()`?

felixhandte (commented):
Followed up on a couple notes.

If at any point you want to hand this off, I'm happy to merge what you have into a working branch and I can take over and polish this up. I don't want you to feel like you *have* to put in any more work, especially if, by your admission, you are not all that comfortable working in C. Let me know what you wanna do!

scottchiefbaker:
@felixhandte if you want to take this and run with it that would be totally fine with me. I just wanted to get the ball rolling. I think most of the hard work is done?

scottchiefbaker:
@felixhandte when you get things massaged let me know. I'd like to see what it looks like when it's all done. I'm sure I'll have questions for you as I'd like to learn more C.

PR #2697, opened 2021-06-07
[bug] Fix entropy repeat mode bug

senhuang42:
Fixes a bug from OSS-Fuzz that occurs in the block splitter: the change to the `blockState` entropy repeat mode was erroneously happening _before_ `ZSTD_blockState_confirmRepcodesAndEntropyTables()` and would get overwritten.

Testing - this was never caught by the OSS fuzzer when it was running on the advanced API. But it was caught within a few days of adding block splitter to the old API fuzzers.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3332,6 +3332,7 @@ static size_t ZSTD_estimateBlockSize(const BYTE* literals, size_t litSize,
  */
 static size_t ZSTD_buildEntropyStatisticsAndEstimateSubBlockSize(seqStore_t* seqStore, const ZSTD_CCtx* zc) {
     ZSTD_entropyCTablesMetadata_t entropyMetadata;
+    DEBUGLOG(6, "ZSTD_buildEntropyStatisticsAndEstimateSubBlockSize()");
     FORWARD_IF_ERROR(ZSTD_buildBlockEntropyStats(seqStore,
                     &zc->blockState.prevCBlock->entropy,
                     &zc->blockState.nextCBlock->entropy,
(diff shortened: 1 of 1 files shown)


terrelln:
> Testing - this was never caught by the OSS fuzzer when it was running on the advanced API. But it was caught within a few days of adding block splitter to the old API fuzzers.

I haven't looked at the PR yet, but why is this the case? Is it because the fuzzer that caught it doesn't use the advanced API, and we need to extend its coverage?

senhuang42:
> I haven't looked at the PR yet, but why is this the case? Is it because the fuzzer that caught it doesn't use the advanced API, and we need to extend its coverage?

The fuzzer that caught it is `dictionary_round_trip`, which uses `compress_usingDict`. The `dictionary_stream_round_trip` fuzzer does use block splitter, but with `compressStream2`. So maybe it's the case that this repeat mode issue only surfaces on single-shot dict compression? But then again I'm not entirely sure why that would be the case for this particular issue.

terrelln:
> The fuzzer that caught it is dictionary_round_trip, which uses compress_usingDict

Could we extend it to use the advanced API 1/2 the time? E.g. the same way `simple_round_trip` does.

senhuang42:
> Could we extend it to use the advanced API 1/2 the time? E.g. the same way `simple_round_trip` does.

I was wrong - `dictionary_round_trip` only uses the old API 1/16 times, `compress2` otherwise. https://github.com/facebook/zstd/blob/8a3bdfaa7bb2944dbed1ce9dede39f512d6da3f9/tests/fuzz/dictionary_round_trip.c#L37

terrelln:
Do you think there is a reason why the fuzzer didn't catch it until we added the old API? Or do you think it is just random chance?

senhuang42:
> Do you think there is a reason why the fuzzer didn't catch it until we added the old API? Or do you think it is just random chance?

I had investigated a bit, and it's still not really clear to me how this ends up happening, though I'll keep looking. Random chance being the reason seems somewhat unlikely since the old API is only hit with 1/16th of the frequency, and the old API only uses block splitter with high compression levels.

terrelln (approved):

Cyan4973 (approved):

PR #2698, opened 2021-06-07
Fix --progress flag to properly control progress display and default …

binhdvo:
…progress display on when using -v

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -2155,6 +2155,7 @@ FIO_decompressZstdFrame(FIO_ctx_t* const fCtx, dRess_t* ress, FILE* finput,
         ZSTD_inBuffer  inBuff = { ress->srcBuffer, ress->srcBufferLoaded, 0 };
         ZSTD_outBuffer outBuff= { ress->dstBuffer, ress->dstBufferSize, 0 };
         size_t const readSizeHint = ZSTD_decompressStream(ress->dctx, &outBuff, &inBuff);
+        const int displayLevel = (!fCtx->hasStdoutOutput || g_display_prefs.progressSetting == FIO_ps_always) ? 1 : 2;
         if (ZSTD_isError(readSizeHint)) {
             DISPLAYLEVEL(1, "%s : Decoding error (36) : %s \n",
                             srcFileName, ZSTD_getErrorName(readSizeHint));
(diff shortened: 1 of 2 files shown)


programs/fileio.c
```diff
@@ -2165,21 +2165,20 @@ FIO_decompressZstdFrame(FIO_ctx_t* const fCtx, dRess_t* ress, FILE* finput,
...
-                size_t srcFileNameSize = strlen(srcFileName);
-                if (srcFileNameSize > 18) {
-                    const char* truncatedSrcFileName = srcFileName + srcFileNameSize - 15;
-                    DISPLAYUPDATE(2, "\rDecompress: %2u/%2u files. Current: ...%s : %u MB...    ",
-                                    fCtx->currFileIdx+1, fCtx->nbFilesTotal, truncatedSrcFileName, (unsigned)((alreadyDecoded+frameSize)>>20) );
-                } else {
-                    DISPLAYUPDATE(2, "\rDecompress: %2u/%2u files. Current: %s : %u MB...    ",
-                                fCtx->currFileIdx+1, fCtx->nbFilesTotal, srcFileName, (unsigned)((alreadyDecoded+frameSize)>>20) );
-                }
+        int displayLevel = (!fCtx->hasStdoutOutput || g_display_prefs.progressSetting == FIO_ps_always) ? 1 : 2;
```
terrelln: nit: `const` & you'll have to move this to the top of the `while` loop scope, or add a new scope.

terrelln (commented):
This looks good to me, once the build warning is fixed!

I think there is a follow up though, for both compression & decompression:

```
zstd -c --progress /path/to/file > /dev/null
zstd -dc --progress /path/to/file > /dev/null
```

In both of these cases, no newline is printed at the end of compression. Normally, the progress is replaced with a status line, like these commands:

```
zstd /path/to/file -o /dev/null
zstd -d /path/to/file -o /dev/null
```

I think we should either:

1. Add a `\n` after the last progress message.
2. Also print the final status line with `--progress`.

I don't have a super strong preference, but I would go with (2) unless we have an explicit reason why we chose not to.

programs/fileio.c
```diff
@@ -1612,7 +1612,8 @@ FIO_compressFilename_internal(FIO_ctx_t* const fCtx,
         U64 const timeLength_ns = UTIL_clockSpanNano(timeStart);
         double const timeLength_s = (double)timeLength_ns / 1000000000;
         double const cpuLoad_pct = (cpuLoad_s / timeLength_s) * 100;
-        DISPLAYLEVEL(4, "%-20s : Completed in %.2f sec  (cpu load : %.0f%%)\n",
+        DISPLAYLEVEL((g_display_prefs.progressSetting == FIO_ps_always ? 1 : 4),
```
Cyan4973: This seems separate from the `--progress` notification issue.
This message is more part of the final status.

And I don't believe that this additional information (time spent) belongs to default, which should remain restricted to 1 line.
We may have room for discussion in determining if it's more appropriate for level `3` (`-v`) or `4` (`-vv`),
but it should certainly not be part of `2` (default).

binhdvo: This was added per Nick's comment suggesting we add the final status line when using --progress to overwrite the last progress message; it will still remain at level 4 in absence of the progress flag

Cyan4973: The final status line is just the line specifying the final size and compression ratio.
Information about timing and speed are considered supplementary (and do constitute an additional line).

binhdvo: Ah got it, will update

Cyan4973 (approved):

PR #2702, opened 2021-06-09
Format File Sizes Human-Readable in the CLI

### LINKED ISSUE #2694: Compression results are hard to read if numbers are very large

felixhandte:
This PR extends @scottchiefbaker's #2696. It switches zstd's CLI output to printing human-readable representations of file sizes, rather than full-precision integers.

This table shows how this PR formats various sizes in comparison to `ls -lh`. There are some differences, but in general I prefer this formatting over `ls`'s, since this provides more consistent 3-4 digits of precision and rounds-to-nearest rather than always rounding-up.

   Size    |  `zstd`  | `ls -lh`
---------- | -------- | --------
1          | 1 B       | 1
12         | 12 B      | 12
123        | 123 B     | 123
1234       | 1.21 KiB  | 1.3K
12345      | 12.1 KiB  | 13K
123456     | 121 KiB   | 121K
1234567    | 1.18 MiB  | 1.2M
12345678   | 11.8 MiB  | 12M
123456789  | 118 MiB   | 118M
1234567890 | 1.15 GiB  | 1.2G
999        | 999 B     | 999
1000       | 1000 B    | 1000
1001       | 1001 B    | 1001
1023       | 1023 B    | 1023
1024       | 1.000 KiB | 1.0K
1025       | 1.00 KiB  | 1.1K
999999     | 977 KiB   | 977K
1000000    | 977 KiB   | 977K
1000001    | 977 KiB   | 977K
1023999    | 1000 KiB  | 1000K
1024000    | 1000 KiB  | 1000K
1024001    | 1000 KiB  | 1001K
1048575    | 1024 KiB  | 1.0M
1048576    | 1.000 MiB | 1.0M
1048577    | 1.00 MiB  | 1.1M

<details>
<summary>Repro Instructions:</summary>

```
for N in 1 12 123 1234 12345 123456 1234567 12345678 123456789 1234567890 999 1000 1001 1023 1024 1025 999999 1000000 1000001 1023999 1024000 1024001 1048575 1048576 1048577; do
  head -c $N /dev/urandom > r$N
done
./zstd -i1 -b1 -S r1 r12 r123 r1234 r12345 r123456 r1234567 r12345678 r123456789 r1234567890 r999 r1000 r1001 r1023 r1024 r1025 r999999 r1000000 r1000001 r1023999 r1024000 r1024001 r1048575 r1048576 r1048577
```
</details>

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -121,8 +121,14 @@ int UTIL_requireUserConfirmation(const char* prompt, const char* abortMsg,
 *  Functions
 ***************************************/
 
+/*
+ * Take a size in bytes and output a human readable string. Maximum
+ * buffer size is 8 but it's usually 7. Example: "123.4G"
+*/
 char* humanSize(unsigned long long size, char* str) {
-    if (size > 1125899906842624L) {
+    if (size > 1152921504606846976L) {
+        snprintf(str, 7, "%.1fE", (float)size / 1152921504606846976L);
+	} else if (size > 1125899906842624L) {
         snprintf(str, 7, "%.1fP", (float)size / 1125899906842624L);
     } else if (size > 1099511627776L) {
         snprintf(str, 7, "%.1fT", (float)size / 1099511627776L);
@@ -134,8 +140,6 @@ char* humanSize(unsigned long long size, char* str) {
         snprintf(str, 7, "%.1fK", (float)size / 1024);
     } else if (size <= 1024) {
         snprintf(str, 7, "%lluB", size);
-    } else {
-        str[0] = '\0';
     }
 
     return str;
--- a/programs/util.c
+++ b/programs/util.c
@@ -311,7 +311,7 @@ UTIL_HumanReadableSize_t UTIL_makeHumanReadableSize(U64 size) {
          * values that exceed the integral precision of a double. */
         if (size >= (1ull << 53)) {
             hrs.value = (double)size / (1ull << 20);
-            hrs.suffix = " MB";
+            hrs.suffix = " MiB";
             /* At worst, a double representation of a maximal size will be
              * accurate to better than tens of kilobytes. */
             hrs.precision = 2;
@@ -324,22 +324,22 @@ UTIL_HumanReadableSize_t UTIL_makeHumanReadableSize(U64 size) {
         /* In regular mode, scale sizes down and use suffixes. */
         if (size >= (1ull << 60)) {
             hrs.value = (double)size / (1ull << 60);
-            hrs.suffix = " EB";
+            hrs.suffix = " EiB";
         } else if (size >= (1ull << 50)) {
             hrs.value = (double)size / (1ull << 50);
-            hrs.suffix = " PB";
+            hrs.suffix = " PiB";
         } else if (size >= (1ull << 40)) {
             hrs.value = (double)size / (1ull << 40);
-            hrs.suffix = " TB";
+            hrs.suffix = " TiB";
         } else if (size >= (1ull << 30)) {
             hrs.value = (double)size / (1ull << 30);
-            hrs.suffix = " GB";
+            hrs.suffix = " GiB";
         } else if (size >= (1ull << 20)) {
             hrs.value = (double)size / (1ull << 20);
-            hrs.suffix = " MB";
+            hrs.suffix = " MiB";
         } else if (size >= (1ull << 10)) {
             hrs.value = (double)size / (1ull << 10);
-            hrs.suffix = " KB";
+            hrs.suffix = " KiB";
         } else {
             hrs.value = (double)size;
             hrs.suffix = " B";
--- a/programs/util.c
+++ b/programs/util.c
@@ -305,23 +305,23 @@ U64 UTIL_getFileSizeStat(const stat_t* statbuf)
 
 UTIL_HumanReadableSize_t UTIL_makeHumanReadableSize(U64 size) {
     UTIL_HumanReadableSize_t hrs;
-    if (size >= (1L << 60)) {
-        hrs.value = (float)size / (1L << 60);
+    if (size >= (1ull << 60)) {
+        hrs.value = (float)size / (1ull << 60);
         hrs.suffix = "E";
-    } else if (size >= (1L << 50)) {
-        hrs.value = (float)size / (1L << 50);
+    } else if (size >= (1ull << 50)) {
+        hrs.value = (float)size / (1ull << 50);
         hrs.suffix = "P";
-    } else if (size >= (1L << 40)) {
-        hrs.value = (float)size / (1L << 40);
+    } else if (size >= (1ull << 40)) {
+        hrs.value = (float)size / (1ull << 40);
         hrs.suffix = "T";
-    } else if (size >= (1L << 30)) {
-        hrs.value = (float)size / (1L << 30);
+    } else if (size >= (1ull << 30)) {
+        hrs.value = (float)size / (1ull << 30);
         hrs.suffix = "G";
-    } else if (size >= (1L << 20)) {
-        hrs.value = (float)size / (1L << 20);
+    } else if (size >= (1ull << 20)) {
+        hrs.value = (float)size / (1ull << 20);
         hrs.suffix = "M";
-    } else if (size >= (1L << 10)) {
-        hrs.value = (float)size / (1L << 10);
+    } else if (size >= (1ull << 10)) {
+        hrs.value = (float)size / (1ull << 10);
         hrs.suffix = "K";
     } else {
         hrs.value = (float)size;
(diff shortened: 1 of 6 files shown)


programs/util.c
```diff
@@ -303,6 +303,43 @@ U64 UTIL_getFileSizeStat(const stat_t* statbuf)
     return (U64)statbuf->st_size;
 }
 
+UTIL_HumanReadableSize_t UTIL_makeHumanReadableSize(U64 size) {
+    UTIL_HumanReadableSize_t hrs;
+    if (size >= (1L << 60)) {
+        hrs.value = (float)size / (1L << 60);
```
Cyan4973: should it rather be `1ULL << 60` ?

Cyan4973:
I like the proposed new formatting.
Ready to validate once tests pass.

senhuang42:
I think the human readable format is generally better so this is great!

My only concern is that sometimes the full-precision ints are useful in development for spotting very small changes in compressed size when using `-b#e#` or even in general. I could also just compress the file and check the compressed result to spot these changes, but this gets a bit annoying when I'm trying out multiple levels. But then again, this human-readable thing was meant to improve the experience in the first place, so I think it's a valid enough concern.

Is there any easy way we can make it toggle back to raw bytes with `-v`, depending on `g_displayLevel` or something?

felixhandte:
@senhuang42, good point. I was thinking about whether `-v` should expand these back to full precision but I couldn't think of a motivating use case. The one you provide is a good one that I missed.

The problem is that I'm not sure how I would best accomplish that. It wouldn't be enough to cancel the scaling and suffix. The value is currently stored as float, but even if it were a double, it would still lose the ability to carry full precision for files over 2^53 bytes in size. So to maintain full precision over the range of file sizes it would have to be transported as a `uint64_t`. But then it's unclear how to dispatch it in the format string...

You'd have to go back to preparing the value in a separate buffer. Which we could do, I guess. @Cyan4973, do you think that's worth doing?

felixhandte:
Actually I think it's not so hard to return to working with our own buffer. I'll play around with that.

Cyan4973:
Yes, I believe that full-precision of size for benchmarking / measurement purposes is a good use case.

Also : 
> it would still lose the ability to carry full precision for files over 2^53 bytes in size

This represents 8 PB. It feels like an acceptable limitation.
Reducing maximum accuracy to `KB` level in this case also seems an acceptable mitigation.

felixhandte:
Ok I've made a few changes (refer to commit messages). Let me know if you think these make sense. I chose to bump full precision display to require double-verbose because that way single-verbose still gets you human-readable display of each compression when processing multiple files.

Cyan4973 (approved):

Cyan4973:
> I chose to bump full precision display to require double-verbose because that way single-verbose still gets you human-readable display of each compression when processing multiple files.

I agree, that seems a good choice.

aqrit:
bike shedding: megabyte `MB` or mebibyte `MiB`

felixhandte:
The new changes include the `--list` command:

```
$ ./zstd -l *.zst
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
     1      0     977 KiB       977 KiB  1.000  XXH64  r1000000.zst
     1      0     977 KiB       977 KiB  1.000  XXH64  r1000001.zst
     1      0    1014   B      1000   B  0.986  XXH64  r1000.zst
     1      0    1015   B      1001   B  0.986  XXH64  r1001.zst
     1      0    1000 KiB      1000 KiB  1.000  XXH64  r1023999.zst
     1      0    1.01 KiB      1023   B  0.986  XXH64  r1023.zst
     1      0    1000 KiB      1000 KiB  1.000  XXH64  r1024000.zst
     1      0    1000 KiB      1000 KiB  1.000  XXH64  r1024001.zst
     1      0    1.01 KiB     1.000 KiB  0.987  XXH64  r1024.zst
     1      0    1.01 KiB      1.00 KiB  0.987  XXH64  r1025.zst
     1      0    1.00 MiB      1024 KiB  1.000  XXH64  r1048575.zst
     1      0    1.00 MiB     1.000 MiB  1.000  XXH64  r1048576.zst
     1      0    1.00 MiB      1.00 MiB  1.000  XXH64  r1048577.zst
     1      0    1.15 GiB      1.15 GiB  1.000  XXH64  r1234567890.zst
     1      0     118 MiB       118 MiB  1.000  XXH64  r123456789.zst
     1      0    11.8 MiB      11.8 MiB  1.000  XXH64  r12345678.zst
     1      0    1.18 MiB      1.18 MiB  1.000  XXH64  r1234567.zst
     1      0     121 KiB       121 KiB  1.000  XXH64  r123456.zst
     1      0    12.1 KiB      12.1 KiB  0.999  XXH64  r12345.zst
     1      0    1.22 KiB      1.21 KiB  0.989  XXH64  r1234.zst
     1      0     136   B       123   B  0.904  XXH64  r123.zst
     1      0      25   B        12   B  0.480  XXH64  r12.zst
     1      0      14   B         1   B  0.071  XXH64  r1.zst
     1      0     977 KiB       977 KiB  1.000  XXH64  r999999.zst
     1      0    1013   B       999   B  0.986  XXH64  r999.zst
----------------------------------------------------------------- 
    28      0    2.56 GiB      2.56 GiB  1.000  XXH64  28 files
```

As well as in-progress compression in various ways:

```
$ ./zstd -f -9 r*
Compress: 35/50 files. Current: r123456789 Read:   108 MiB /   118 MiB ==> 100%
```

```
$ ./zstd -f -v -9 r*
*** zstd command line interface 64-bits v1.5.0, by Yann Collet ***
r1                   :1400.00%   (     1   B =>     14   B, r1.zst)            .00% 
r1000                :101.40%   (  1000   B =>   1014   B, r1000.zst)          
r1000000             :100.00%   (   977 KiB =>    977 KiB, r1000000.zst)       
r1000001             :100.00%   (   977 KiB =>    977 KiB, r1000001.zst)       
...
r12345678            :100.00%   (  11.8 MiB =>   11.8 MiB, r12345678.zst)      0.00% 
(L9) Buffered :  10.5 MiB - Consumed :  21.5 MiB - Compressed :  21.5 MiB => 100.00%
```

```
$ ./zstd -f -vv -9 r*
*** zstd command line interface 64-bits v1.5.0, by Yann Collet ***
r1                   :1400.00%   (     1   B =>     14   B, r1.zst)            00.00% 
r1                   : Completed in 0.00 sec  (cpu load : 96%)
r1000                :101.40%   (  1000   B =>   1014   B, r1000.zst)          
r1000                : Completed in 0.00 sec  (cpu load : 98%)
r1000000             :100.00%   (1000000   B => 1000037   B, r1000000.zst)     
r1000000             : Completed in 0.02 sec  (cpu load : 103%)
...
r123456789           :100.00%   (123456789   B => 123459629   B, r123456789.zst)  => 100.00% 
r123456789           : Completed in 1.21 sec  (cpu load : 118%)
(L9) Buffered :10616832   B - Consumed :1134559232   B - Compressed :1134585210   B => 100.00%
```

Cyan4973:
Thanks @felixhandte ! Looks like a great improvement !

scottchiefbaker:
Let me start with: Thanks for working on this and getting merged so quickly.

Small nitpick though:

```
:./zstd -b1 -e10 -f /tmp/foo.bin
 1#foo.bin           : 268 MiB -> 6.98 MiB (38.36), 2002.5 MB/s, 5363.6 MB/s 
 2#foo.bin           : 268 MiB -> 6.82 MiB (39.25), 1916.1 MB/s, 5268.9 MB/s 
 3#foo.bin           : 268 MiB -> 7.87 MiB (34.02), 1619.2 MB/s, 4759.0 MB/s 
 4#foo.bin           : 268 MiB -> 7.87 MiB (34.00), 1571.1 MB/s, 4753.0 MB/s 
```

We have conflicting data types? MiB for the before/after but MB/s for the rate? I would expect those to be the same.

terrelln:
> We have conflicting data types? MiB for the before/after but MB/s for the rate? I would expect those to be the same.

The speeds are also in `MiB/s`, they just use `MB/s` to mean that. We should switch them to print `MiB/s` as well.

Cyan4973:
Actually, speeds are indeed in `MB/s`.
We have to keep it that way, in order to generate results comparable between versions.

scottchiefbaker:
If that is indeed the case, then why don't we change the new output to also be in MB instead of MiB?

Since we're introducing the new human string format, no one is expecting it to be any format, let's make that MB. That way both units are the same.

PR #2703, opened 2021-06-09
Add support for --long-param flag, fix #2104

### LINKED ISSUE #2104: Fast Mode UX is Poor
Every recent benchmark of ZSTD that I can find skips "fast mode" ([lzbench][lzbench], [squash][squash], [peazip][peazip], ***[zstd][zstd]***).

This is partly because most of the material about ZSTD was written before the negative ranges were added.  People just don't know this and you need to advertise this capability.  Maybe you could break your benchmark table and graphs into 3 ranges, like is done [here](https://engineering.fb.com/core-data/zstandard/):

image

The other issue is that users don't read manuals.  While I understand hiding the highest compression levels behind `--ultra` due to spiking client memory consumption, the `--fast=[#]` is just there to accommodate CLI formatting convention and hides the higher speed levels from end users:

> No extra compression settings were used to fine tune the compression, neither for Brotli nor for Zstandard, compression was defined exclusively setting the compression level parameter. Both algorithms features plenty fine-tuning options, which are out of the scope of this benchmark.
> -[PeaZIP Blog][peazip]

I would strongly suggest the following:

- [ ]  Accept negative integers (i.e. `--5` and maybe `- -5`) for the compression level.
- [ ]  Deprecate (but probably never actually remove) the `--fast` flag to discourage this being seen as fine tuning.
- [ ] Update all [documentation](https://github.com/facebook/zstd/blob/4a4018870f48ae859a5fb730ef4ff1e1bd79e6ba/programs/zstd.1) to state the full range of levels.
- [ ] Highlight the full range on the website and readme.md, perhaps by breaking up the benchmarks into fast, normal, and ultra sections.  Make sure to explicitly state the range being from -5 to 22.

[lzbench]: https://github.com/inikep/lzbench/issues/68
[squash]: https://github.com/quixdb/squash/issues/251
[peazip]: https://www.peazip.org/fast-compression-benchmark-brotli-zstandard.html
[zstd]: https://github.com/facebook/zstd/issues/2098

binhdvo:
--long-param allows negative compression levels to be specified as "fast" compression levels.  Minimum and maximum compression levels are still enforced.  Fixes #2104

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -206,6 +206,7 @@ static void usage_advanced(const char* programName)
     DISPLAYOUT( "--ultra : enable levels beyond %i, up to %i (requires more memory) \n", ZSTDCLI_CLEVEL_MAX, ZSTD_maxCLevel());
     DISPLAYOUT( "--long[=#]: enable long distance matching with given window log (default: %u) \n", g_defaultMaxWindowLog);
     DISPLAYOUT( "--fast[=#]: switch to very fast compression levels (default: %u) \n", 1);
+    DISPLAYOUT( "--long-param=#: specify compression level, accepts negative values as fast compression levels \n");
     DISPLAYOUT( "--adapt : dynamically adapt compression level to I/O conditions \n");
     DISPLAYOUT( "--[no-]row-match-finder : force enable/disable usage of fast row-based matchfinder for greedy, lazy, and lazy2 strategies \n");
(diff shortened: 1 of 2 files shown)


Cyan4973:
little trick : 
document the related issue number in the PR description (ex: "fix #2104")
This will help cross-referencing, and will also automatically close the issue once the PR is merged.

PR #2705, opened 2021-06-09
Add support for negative values in advanced flags

binhdvo:
Fixes command line flag from https://github.com/facebook/zstd/pull/2703

Adds --long-level flag which behaves as -# except supporting negative values as fast compression levels.

--long- is intended to be the prefix for any future parameters that should have a version which accepts negative values.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -354,6 +354,23 @@ static unsigned readU32FromChar(const char** stringPtr) {
     return result;
 }
 
+/*! readIntFromChar() :
+ * @return : signed integer value read from input in `char` format.
+ *  allows and interprets K, KB, KiB, M, MB and MiB suffix.
+ *  Will also modify `*stringPtr`, advancing it to position where it stopped reading.
+ *  Note : function will exit() program if digit sequence overflows */
+static int readIntFromChar(const char** stringPtr) {
+    static const char errorMsg[] = "error: numeric value overflows 32-bit int";
+    int sign = 1;
+    unsigned result;
+    if (**stringPtr=='-') {
+        (*stringPtr)++;
+        sign = -1;
+    }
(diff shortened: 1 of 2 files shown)


Cyan4973 (changes requested):
Let's discuss a rewording of this task for improved clarity

programs/zstdcli.c
```diff
@@ -547,8 +564,8 @@ static ZDICT_fastCover_params_t defaultFastCoverParams(void)
 static unsigned parseAdaptParameters(const char* stringPtr, int* adaptMinPtr, int* adaptMaxPtr)
 {
     for ( ; ;) {
-        if (longCommandWArg(&stringPtr, "min=")) { *adaptMinPtr = (int)readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
-        if (longCommandWArg(&stringPtr, "max=")) { *adaptMaxPtr = (int)readU32FromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
+        if (longCommandWArg(&stringPtr, "min=")) { *adaptMinPtr = (int)readIntFromChar(&stringPtr); if (stringPtr[0]==',') { stringPtr++; continue; } else break; }
```
Cyan4973: nit : no need for the `(int)` cast, since it's already the return type of `readIntFromChar()`.

tests/playTests.sh
```diff
@@ -1519,6 +1519,7 @@ then
     println "\n===>   adaptive mode "
     roundTripTest -g270000000 " --adapt"
     roundTripTest -g27000000 " --adapt=min=1,max=4"
+    roundTripTest -g27000000 " --adapt=min=-2,max=-1"
```
Cyan4973, line 1522: 👍

Cyan4973 (approved):

PR #2708, opened 2021-06-11
Add API for fetching skippable frame content

binhdvo:
Provides functions for navigating frames in memory and identifying and extracting content from skippable frames.  Streaming mode left out of scope for now.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -380,6 +380,19 @@ unsigned ZSTD_isFrame(const void* buffer, size_t size)
     return 0;
 }
 
+/*! ZSTD_isSkippableFrame() :
+ *  Tells if the content of `buffer` starts with a valid Frame Identifier for a skippable frame.
+ *  Note : Frame Identifier is 4 bytes. If `size < 4`, @return will always be 0.
+ */
+unsigned ZSTD_isSkippableFrame(const void* buffer, size_t size)
+{
+    if (size < ZSTD_FRAMEIDSIZE) return 0;
+    {   U32 const magic = MEM_readLE32(buffer);
+        if ((magic & ZSTD_MAGIC_SKIPPABLE_MASK) == ZSTD_MAGIC_SKIPPABLE_START) return 1;
+    }
+    return 0;
+}
+
 /** ZSTD_frameHeaderSize_internal() :
  *  srcSize must be large enough to reach header size fields.
  *  note : only works for formats ZSTD_f_zstd1 and ZSTD_f_zstd1_magicless.
@@ -503,7 +516,6 @@ size_t ZSTD_getFrameHeader(ZSTD_frameHeader* zfhPtr, const void* src, size_t src
     return ZSTD_getFrameHeader_advanced(zfhPtr, src, srcSize, ZSTD_f_zstd1);
 }
 
-
 /** ZSTD_getFrameContentSize() :
  *  compatible with legacy mode
  * @return : decompressed size of the single frame pointed to be `src` if known, otherwise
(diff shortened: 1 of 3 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -544,6 +556,59 @@ static size_t readSkippableFrameSize(void const* src, size_t srcSize)
...
+
+    /* check input validity */
+    RETURN_ERROR_IF(!ZSTD_isSkippableFrame(src, srcSize), frameParameter_unsupported, "");
+    if (skippableFrameSize < ZSTD_SKIPPABLEHEADERSIZE || skippableFrameSize > srcSize)
+        return ZSTD_CONTENTSIZE_ERROR;
+    if (skippableContentSize > dstCapacity)
+        return ZSTD_CONTENTSIZE_ERROR;
+
+    /* deliver payload */
+    ZSTD_memcpy(dst, (const BYTE *)src + ZSTD_SKIPPABLEHEADERSIZE, skippableContentSize);
```
terrelln: This will need to handle `dst == NULL` and `skippableContentSize == 0`, since it is UB to call memcpy() with `NULL.

lib/decompress/zstd_decompress.c
```diff
@@ -544,6 +556,59 @@ static size_t readSkippableFrameSize(void const* src, size_t srcSize)
...
+    *magicVariant = magicNumber - ZSTD_MAGIC_SKIPPABLE_START;
```
terrelln: nit: Do we want to allow people to pass `magicVariant == NULL` if they don't care about it?

lib/zstd.h
```diff
@@ -1441,6 +1441,37 @@ ZSTDLIB_API size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size
...
+ *
+ * The parameter magicVariant will receive the magicVariant that was supplied when the frame was written,
+ * i.e. magicNumber - ZSTD_MAGIC_SKIPPABLE_START
+ *
+ * Returns an error if destination buffer is not large enough, if the source size is not representable
+ * with a 4-byte unsigned int, or if the frame is not skippable.
+ *
+ * @return : number of bytes written or a ZSTD error.
+ */
+ZSTDLIB_API unsigned long long ZSTD_readSkippableFrame(void* dst, size_t dstCapacity, unsigned* magicVariant,
```
terrelln: nit: A ZSTD error sum type (error | size) cannot be an `unsigned long long`. This will have to return a `size_t`.

It is fine to return a `size_t` because we know the data is contiguous, so a `size_t` is large enough.

binhdvo: I think this caused one of the CI tests to fail before, let me try again

lib/decompress/zstd_decompress.c
```diff
@@ -544,6 +556,59 @@ static size_t readSkippableFrameSize(void const* src, size_t srcSize)
...
+ */
+unsigned long long ZSTD_findFrameSize(const void *src, size_t srcSize)
+{
+    ZSTD_frameHeader zfh;
+    if (ZSTD_getFrameHeader(&zfh, src, srcSize) != 0)
+        return ZSTD_CONTENTSIZE_ERROR;
+    if (zfh.frameType == ZSTD_skippableFrame) {
+        return readSkippableFrameSize(src, srcSize);
+    } else {
+        return ZSTD_findFrameCompressedSize(src, srcSize);
```
Cyan4973: Isn't `ZSTD_findFrameCompressedSize()` already able to provide the size of a skippable frame ?

binhdvo: Ah, I thought that was only for compressed frames; removed the new function.

lib/decompress/zstd_decompress.c
```diff
@@ -544,6 +556,38 @@ static size_t readSkippableFrameSize(void const* src, size_t srcSize)
...
+
+    /* check input validity */
+    RETURN_ERROR_IF(!ZSTD_isSkippableFrame(src, srcSize), frameParameter_unsupported, "");
+    RETURN_ERROR_IF(skippableFrameSize < ZSTD_SKIPPABLEHEADERSIZE || skippableFrameSize > srcSize, srcSize_wrong, "");
+    RETURN_ERROR_IF(skippableContentSize > dstCapacity, dstSize_tooSmall, "");
+
+    /* deliver payload */
+    if (skippableContentSize > 0  && dst != NULL)
+        ZSTD_memcpy(dst, (const BYTE *)src + ZSTD_SKIPPABLEHEADERSIZE, skippableContentSize);
+    if (magicVariant != NULL)
```
Cyan4973, line 585: Does the API allow `magicVariant == NULL` ?
In which case, the documentation should mention it and tell what happens in this case.
If not, an `assert()` here would be more appropriate.
Finally, an alternative design could be to return a `struct` that combine the 2 results (`size_t` bytes written, and `unsigned` magic variant), making them clearly return values, and avoiding potential confusion with regards to caller-side allocation, in+out values, etc.

Cyan4973:
Almost there !
I like the consistency of this proposal, which is a clear mirror of existing `readSkippableFrame()`.
Let's just have a discussion about function's return type, and I guess it will be good to go.

lib/zstd.h
```diff
@@ -1441,6 +1441,27 @@ ZSTDLIB_API size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size
...
 
+/*! ZSTD_readSkippableFrame() :
+ * Retrieves a zstd skippable frame containing data given by src, and writes it to dst buffer.
+ *
+ * The parameter magicVariant will receive the magicVariant that was supplied when the frame was written,
+ * i.e. magicNumber - ZSTD_MAGIC_SKIPPABLE_START.  This can be NULL if the caller is not interested
+ * in the magicVariant.
+ *
+ * Returns an error if destination buffer is not large enough, if the source size is not representable
+ * with a 4-byte unsigned int, or if the frame is not skippable.
```
Cyan4973: minor comment : 

> if the source size is not representable * with a 4-byte unsigned int

I think the binary format does not allow such scenario. Skippable frame content size _must_ be represented as a 4-bytes unsigned int.

Cyan4973 (approved):

PR #2711, opened 2021-06-17
Limit --ultra compression level in 32 bit systems and update warning …

binhdvo:
Limits max --ultra compression level in 32 bit systems to 21 to avoid issues with contiguous memory allocation.  Also updates tests, documentation, and warning message accordingly.

### DIFF
--- a/build/meson/tests/meson.build
+++ b/build/meson/tests/meson.build
@@ -186,10 +186,6 @@ test('test-zstream-1',
   zstreamtest,
   args: ['-v', ZSTREAM_TESTTIME] + FUZZER_FLAGS,
   timeout: 240)
-test('test-zstream-2',
-  zstreamtest,
-  args: ['-mt', '-t1', ZSTREAM_TESTTIME] + FUZZER_FLAGS,
-  timeout: 120)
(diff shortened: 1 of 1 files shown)


tests/fuzzer.c
```diff
@@ -1978,7 +1978,7 @@ static int basicUnitTests(U32 const seed, double compressibility)
             int const rowLevelEnd = 8;
 
             DISPLAYLEVEL(3, "test%3i : flat-dictionary efficiency test : \n", testNb++);
-            assert(maxLevel == 22);
+            assert(maxLevel == (sizeof(void *) == 8 ? 22 : 21));
```
Cyan4973: I'm somewhat embarrassed that the logic of max compression level differentiated by system width must be duplicated here. Though on this topic, I don't have a clearly better solution to propose.

Rather, I'm wondering if this `assert()` is actually valid. What if we want to add a level `23` in the future ? Why should it fail this `assert()` ? I don't see how this `assert()` pre-condition makes the rest of the test more "valid".

Cyan4973 (approved):

PR #2712, opened 2021-06-18
[WIP] Allocate Huge Pages for the Compression Workspace

felixhandte:
This PR is experimental and is not suitable for merge. This is a demonstrator / testbench for the effects of backing the `cwksp` with huge pages.

A quick experiment showed the following performance improvement (huge page speeds on the left):

```
$ ./zstd -b1e19
 1: 9.54 MiB -> 3.01 MiB (3.170),  336.6 MB/s vs 336.6 MB/s
 2: 9.54 MiB -> 2.99 MiB (3.194),  242.4 MB/s vs 242.1 MB/s
 3: 9.54 MiB -> 3.08 MiB (3.095),  128.7 MB/s vs 115.4 MB/s
 4: 9.54 MiB -> 3.19 MiB (2.988),  117.6 MB/s vs 110.0 MB/s
 5: 9.54 MiB -> 3.13 MiB (3.046),   82.5 MB/s vs  75.5 MB/s
 6: 9.54 MiB -> 3.13 MiB (3.047),   80.8 MB/s vs  76.3 MB/s
 7: 9.54 MiB -> 3.12 MiB (3.058),   69.6 MB/s vs  65.2 MB/s
 8: 9.54 MiB -> 3.11 MiB (3.064),   62.5 MB/s vs  61.9 MB/s
 9: 9.54 MiB -> 3.16 MiB (3.022),   56.1 MB/s vs  53.6 MB/s
10: 9.54 MiB -> 3.19 MiB (2.987),   52.2 MB/s vs  49.0 MB/s
11: 9.54 MiB -> 3.20 MiB (2.979),   48.0 MB/s vs  41.7 MB/s
12: 9.54 MiB -> 3.20 MiB (2.980),   46.7 MB/s vs  37.7 MB/s
13: 9.54 MiB -> 3.20 MiB (2.980),   16.1 MB/s vs  13.0 MB/s
14: 9.54 MiB -> 3.20 MiB (2.981),   11.5 MB/s vs  9.80 MB/s
15: 9.54 MiB -> 3.20 MiB (2.982),   9.17 MB/s vs  8.00 MB/s
16: 9.54 MiB -> 2.94 MiB (3.243),   8.26 MB/s vs  8.00 MB/s
17: 9.54 MiB -> 3.00 MiB (3.183),   4.44 MB/s vs  3.89 MB/s
18: 9.54 MiB -> 3.00 MiB (3.177),   3.72 MB/s vs  2.35 MB/s
19: 9.54 MiB -> 3.00 MiB (3.179),   2.82 MB/s vs  2.03 MB/s
```

Which is pretty attractive!

## Running This

You may want to first run e.g.,

```
echo 128 | sudo tee /proc/sys/vm/nr_hugepages
```

In order to attempt to get the kernel to reserve a bunch hugepages in advance. Otherwise if your system has fairly fragmented memory the `mmap()` call will simply fail.

The recommended `make` arguments to play around with this are `DEBUGLEVEL=2 MOREFLAGS="-DZSTD_CWKSP_ALLOC_HUGEPAGE=1`. That way you'll get a message printed if the `mmap()` fails.

While running `./zstd -b3` you should then observe something like:

```
$ cat /proc/meminfo
[...]
HugePages_Total:     128
HugePages_Free:      127
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
```

This indicates you're successfully using a hugepage!

## Future Directions

To be discussed.

### DIFF
--- a/lib/compress/zstd_cwksp.h
+++ b/lib/compress/zstd_cwksp.h
@@ -16,9 +16,6 @@
 ***************************************/
 #include "../common/zstd_internal.h"
 
-#if defined (__cplusplus)
-extern "C" {
-#endif
 
 /*-*************************************
 *  Constants
@@ -39,6 +36,29 @@ extern "C" {
 /* Set our tables and aligneds to align by 64 bytes */
 #define ZSTD_CWKSP_ALIGNMENT_BYTES 64
 
+
+/* This macro controls whether Zstd will try to allocate huge pages to back the
+ * cwksp.
+ */
+#ifndef ZSTD_CWKSP_ALLOC_HUGEPAGE
+#if defined(__x86_64__) && defined(__linux__)
+#define ZSTD_CWKSP_ALLOC_HUGEPAGE 1
+#else
+#define ZSTD_CWKSP_ALLOC_HUGEPAGE 0
+#endif
+#endif
+
+#if ZSTD_CWKSP_ALLOC_HUGEPAGE
+#include <errno.h>
+#include <string.h>
+#include <sys/mman.h>
+#endif
+
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
 /*-*************************************
 *  Structures
 ***************************************/
@@ -55,6 +75,7 @@ typedef enum {
  */
 typedef enum {
     ZSTD_cwksp_dynamic_alloc,
+    ZSTD_cwksp_mmap_alloc,
     ZSTD_cwksp_static_alloc
 } ZSTD_cwksp_alloc_type_e;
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Great results @felixhandte !

In term of design, 
I was expecting this mod to impact `zstd`'s custom allocation path,
as defined here : https://github.com/facebook/zstd/blob/dev/lib/common/zstd_common.c#L54,
replacing the generic invocation of `malloc()` ([here](https://github.com/facebook/zstd/blob/dev/lib/common/zstd_deps.h#L43))
by your system specific invocation of `mmap()`.

(As a bonus, such an approach makes it possible to test without modifying `libzstd`, just by providing [`customMem`](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L1578) functions).

It would extend the benefits of huge pages beyond `CCtx` workspace to any place where the code needs to allocate.
More complex strategies, such as using `malloc` instead of `mmap` when the amount of data to allocate is small, or using `malloc` as backup when `mmap` is too brittle due to memory fragmentation, could be implemented there too.
Though mixing `malloc` and `mmap` comes at the expense of a more complex `ZSTD_free` operation, since the correct function (`munmap` or `free`) must then be invoked on the correct memory segment. That's exactly what the custom allocator state is for, but it sure is more complex.

Another strategy could be to invoke the new `mmap`-based custom allocator selectively for some objects only (the `CCtx` for example, via `ZSTD_createCCtx()`).

felixhandte:
Pursuing a different approach (doing huge-page customization in caller framework rather than in zstd itself).

PR #2724, opened 2021-07-09
pzstd: fix linking for static builds

jonringer:
When trying to build statically, you get the following error:
```
[ 69%] Linking CXX executable pzstd
x86_64-unknown-linux-musl-ld: cannot find -llibzstd_shared
```

This is to correct the build when shared libraries aren't available. (e.g. `BUILD_SHARED=OFF`)

With changes:
```
[100%] Linking CXX executable pzstd
[100%] Built target pzstd
```


jonringer:
doing an `else if` may be more proper. But there's already an xor check on BUILD_SHARED and BUILD_STATIC

jonringer:
failing test seems to be from a fuzz test getting stalled

Cyan4973 (approved):

terrelln (approved):

PR #2725, opened 2021-07-10
Add support for MCST LCC compiler

makise-homura:
I'd like to introduce support for MCST LCC compiler based on EDG front end. It is primarily used to build software for computers based on Russian [Elbrus](https://en.wikipedia.org/wiki/Elbrus-8S) CPUs.

The main two fixes I introduce here, are:
* there is no `optimize("no-tree-vectorize")` support in said compiler;
* it generates a `-Wmain` warning (or error on `-Werror`) while testing zstd because of `argc` is specified as `const` in `main()` of `zstdcli`.

Contribution guidelines followed:
* CLA is signed.
* Tests (both `shortest` and `test`) performed successfully, both on x86_64 with gcc-9 and e2k (Elbrus) with lcc 1.25.
* Travis build is [OK](https://travis-ci.com/github/makise-homura/zstd).
* AppVeyor [build](https://ci.appveyor.com/project/makise-homura/zstd) failed in mingw tasks: looks like it isn't properly configured (`The system cannot find the path specified.` while preparing the build environment).
* Static analyzer (`scan-build` from `clang-tools-10`) found 59 bugs, but somehow 71 were found for current `dev` upstream branch. No one of them, by the way, related to parts of code I modified.

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -150,7 +150,8 @@
 }
 
 /* vectorization
- * older GCC (pre gcc-4.3 picked as the cutoff) uses a different syntax */
+ * older GCC (pre gcc-4.3 picked as the cutoff) uses a different syntax,
+ * and some compilers, like Intel ICC and MCST LCC, do not support it at all. */
 #if !defined(__INTEL_COMPILER) && !defined(__clang__) && defined(__GNUC__) && !defined(__LCC__)
 #  if (__GNUC__ == 4 && __GNUC_MINOR__ > 3) || (__GNUC__ >= 5)
 #    define DONT_VECTORIZE __attribute__((optimize("no-tree-vectorize")))
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -79,13 +79,6 @@
 #  define UNUSED_ATTR
 #endif
 
-/* CONST_ARGC tells the compiler correct const argc specificator for -Wmain. */
-#if !defined(__LCC__)
-#  define CONST_ARGC const
-#else
-#  define CONST_ARGC
-#endif
-
 /* force no inlining */
 #ifdef _MSC_VER
 #  define FORCE_NOINLINE static __declspec(noinline)
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -79,6 +79,13 @@
 #  define UNUSED_ATTR
 #endif
 
+/* CONST_ARGC tells the compiler correct const argc specificator for -Wmain. */
+#if !defined(__LCC__)
+#  define CONST_ARGC const
+#else
+#  define CONST_ARGC
+#endif
+
 /* force no inlining */
 #ifdef _MSC_VER
 #  define FORCE_NOINLINE static __declspec(noinline)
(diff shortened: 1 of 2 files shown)


programs/zstdcli.c
```diff
@@ -741,7 +741,7 @@ typedef enum { zom_compress, zom_decompress, zom_test, zom_bench, zom_train, zom
 # define MAXCLEVEL  ZSTD_maxCLevel()
 #endif
 
-int main(int const argCount, const char* argv[])
+int main(int CONST_ARGC argCount, const char* argv[])
```
Cyan4973: I'm puzzled by this proposed change. 
I get it that it's a workaround for elbrus compiler. And yes, it's correct.
I just wonder if it wouldn't be simpler to just remove the `const` statement.
This would avoid having to maintain yet another specificity.

lib/common/compiler.h
```diff
@@ -151,7 +158,7 @@
 
 /* vectorization
  * older GCC (pre gcc-4.3 picked as the cutoff) uses a different syntax */
-#if !defined(__INTEL_COMPILER) && !defined(__clang__) && defined(__GNUC__)
+#if !defined(__INTEL_COMPILER) && !defined(__clang__) && defined(__GNUC__) && !defined(__LCC__)
```
Cyan4973, line 155: Could you please add a comment explaining the presence of this test ?
This is for future contributors, that may only access the code, not the PR.

Cyan4973:
I agree with the general objective of this PR.
I made a couple of comments, mostly focused on improving maintainability.

Speaking of which : would there be any way to test this capability (elbrus support) in (public) CI ?
The objective would be to ensure that this capability remains supported in the future.
Otherwise, it's too easy for any future contributor to break it without noticing it.
That being said, I get it that elbrus compiler is not exactly commonplace.
So this is more a "bonus" topic.

programs/zstdcli.c
makise-homura: Yes, that was the first of my ideas--like why would you declare scalar local variable (or function parameter) as `const`? That may be useful only to capture some of programming mistakes (still there's nothing wrong in treating `argc` as not `const`, e.g. in loops like `for(int i = 0; argc; --argc, ++i) do_something_with(argv[i]);`), but still there's nothing to do with actual code flow. I then found commit a49417b and there was no explanation why `argCount` was changed to `const`; so I preferred not touching the part, reason of which I don't clearly understand. But if you're ok with that, I've just removed `const` at all.

lib/common/compiler.h
makise-homura: Added a comment in 3cd085c.

makise-homura:
> would there be any way to test this capability (elbrus support) in (public) CI ?

@Cyan4973 I'm not sure it is widely available for now out-of-the-box. But there is SSH access to several elbrus machines available (registration is not automated yet, but you may just send me a public key and username, and I will create users for you on this machines; you may do it right here, or contact me by Discord `makise-homura#8793` or in Telegram group `https://t.me/elbrus_gensokyo`). These machines are under NAT, so they have access to internet, but aren't visible from there directly (unless there's ssh tunnel is open).
If it is enough for it, then you may ask me for access, and configure some CI worker there to include it into CI system.

P.S. I'll be on vacation since Jul 16 till Aug 6, so this option won't be available this time (also I won't be able to do commits and answer here until I'm back); but I can do it today or tomorrow, or in August after I'm back.

Cyan4973 (approved):

Cyan4973:
>  I can do it today or tomorrow, or in August after I'm back.

No need to rush this topic. We can certainly wait for August.

The difficulty will be in automating this test, and making sure that it doesn't generate too much noise in CI results, such as failures for reasons different than the code itself. Better take the time to do it well.

makise-homura:
Ok, let's wait for August then. Feel free to ask me for any assistance regarding this if needed.

PR #2732, opened 2021-07-26
[HUF] Improve Huffman sorting algorithm

senhuang42:
This PR improves the current huffman sorting strategy, mostly on small blocks, by doing the following:

- Decoupling bucketing and sorting into separate steps.
- Skip attempts to sort buckets that are 0 or 1-sized. (e.g. silesia/dickens has a lot of 0-count symbols).
- Use 128 buckets that are split into distinct count buckets that don't need to be sorted, and the typical log2 bucketing
  - This reduces the number of buckets we actually need to sort to at most 11, and reduces the avg nb of elements per bucket.
- Use quicksort

Using @terrelln's nifty benchmarking tool, we see the following improvements:
|     Benchmark     | Config  |  Dataset   | Ratio (dev) | Ratio (huf_sort_improvement) | Ratio (huf_sort_improvement - dev) | Speed MB/s (dev) | Speed MB/s (huf_sort_improvement) | Speed MB/s (huf_sort_improvement - dev) |
|-------------------|---------|------------|-------------|------------------------------|------------------------------------|------------------|-----------------------------------|-----------------------------------------|
| compress          | level_1 | silesia    | 2.88        | 2.88                         | -0.0%                              | 347.03           | 347.49                            | 0.1%                                    |
| compress          | level_1 | silesia_4k | 2.31        | 2.31                         | 0.0%                               | 183.02           | 200.93                            | 9.8%                                    |
| compress_literals | level_1 | silesia    | 1.28        | 1.28                         | -0.0%                              | 755.42           | 780.19                            | 3.3%                                    |
| compress_literals | level_1 | silesia_4k | 1.30        | 1.30                         | 0.0%                               | 233.50           | 330.14                            | 41.4%                                   |
| compress_literals | level_7 | silesia    | 1.11        | 1.11                         | -0.0%                              | 677.14           | 746.20                            | 10.2%                                   |

- Nearly 10% speedup for 4KB block compression, 40% literal compression speedup.
- Neutral e2e compression speed on 128K blocks, with improvements to literal compression.
- Very small perturbations to compressed size (since quicksort is not stable).

### DIFF
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -428,22 +428,122 @@ static U32 HUF_setMaxHeight(nodeElt* huffNode, U32 lastNonNull, U32 maxNbBits)
 }
 
 typedef struct {
-    U32 base;
-    U32 curr;
+    U16 base;
+    U16 curr;
 } rankPos;
 
 typedef nodeElt huffNodeTable[HUF_CTABLE_WORKSPACE_SIZE_U32];
 
-#define RANK_POSITION_TABLE_SIZE 32
+/* Number of buckets available for HUF_sort() */
+#define RANK_POSITION_TABLE_SIZE 128
 
 typedef struct {
   huffNodeTable huffNodeTbl;
   rankPos rankPosition[RANK_POSITION_TABLE_SIZE];
 } HUF_buildCTable_wksp_tables;
 
+/* RANK_POSITION_DISTINCT_COUNT_CUTOFF == Cutoff point in HUF_sort() buckets for which we use log2 bucketing.
+ * Strategy is to use as many buckets as possible for representing distinct
+ * counts while using the remainder to represent all counts up to HUF_BLOCKSIZE_MAX
+ * using log2 bucketing.
+ * 
+ * To satisfy this requirement for 128 buckets, we can do the following:
+ * Let buckets 0-114 represent distinct counts of [0, 114]
+ * Let buckets 115 to 126 represent counts of [115, HUF_BLOCKSIZE_MAX]. (the final bucket 127 must remain empty)
+ * 
+ * Note that we don't actually need 17 buckets (assuming 2^17 maxcount) for log2 bucketing since
+ * the first few buckets in the log2 bucketing representation are already covered by the distinct count bucketing.
+ */
+#define RANK_POSITION_LOG_BUCKETS_BEGIN (RANK_POSITION_TABLE_SIZE - 1) - BIT_highbit32(HUF_BLOCKSIZE_MAX) - 1
+#define RANK_POSITION_DISTINCT_COUNT_CUTOFF RANK_POSITION_LOG_BUCKETS_BEGIN + BIT_highbit32(RANK_POSITION_LOG_BUCKETS_BEGIN)
+
+/* Return the appropriate bucket index for a given count. See definition of
+ * RANK_POSITION_DISTINCT_COUNT_CUTOFF for explanation of bucketing strategy.
+ */
+static U32 HUF_getIndex(U32 const count) {
+    return (count < RANK_POSITION_DISTINCT_COUNT_CUTOFF)
+        ? count
+        : BIT_highbit32(count) + RANK_POSITION_LOG_BUCKETS_BEGIN;
+}
+
+/* Helper swap function for HUF_quickSortPartition() */
+static void HUF_swapNodes(nodeElt* a, nodeElt* b) {
+	nodeElt tmp = *a;
+	*a = *b;
+	*b = tmp;
+}
+
+/* Returns 0 if the huffNode array is not sorted by descending count */
+UNUSED_ATTR
+static int HUF_isSorted(nodeElt huffNode[], U32 const maxSymbolValue1) {
+    U32 i;
+    for (i = 1; i < maxSymbolValue1; ++i) {
+        if (huffNode[i].count > huffNode[i-1].count) {
+            return 0;
+        }
+    }
+    return 1;
+}
+
+/* Insertion sort by descending order */
+HINT_INLINE void HUF_insertionSort(nodeElt huffNode[], int const low, int const high) {
+    int i;
+    int const size = high-low+1;
+    huffNode += low;
+    for (i = 1; i < size; ++i) {
+        nodeElt const key = huffNode[i];
+        int j = i - 1;
+        while (j >= 0 && huffNode[j].count < key.count) {
+            huffNode[j + 1] = huffNode[j];
+            j--;
+        }
+        huffNode[j + 1] = key;
+    }
+}
+
+/* Pivot helper function for quicksort. */
+static int HUF_quickSortPartition(nodeElt arr[], int const low, int const high) {
+    /* Simply select rightmost element as pivot. "Better" selectors like
+     * median-of-three don't experimentally appear to have any benefit.
+     */
+    U32 const pivot = arr[high].count;
+    int i = low - 1;
+    int j = low;
+    for ( ; j < high; j++) {
+        if (arr[j].count > pivot) {
+            i++;
+            HUF_swapNodes(&arr[i], &arr[j]);
+        }
+    }
+    HUF_swapNodes(&arr[i + 1], &arr[high]);
+    return i + 1;
+}
+
+/* Classic quicksort by descending with partially iterative calls
+ * to reduce worst case callstack size. 
+ */
+static void HUF_simpleQuickSort(nodeElt arr[], int low, int high) {
+    int const kInsertionSortThreshold = 8;
+    if (high - low < kInsertionSortThreshold) {
+        HUF_insertionSort(arr, low, high);
+        return;
+    }
+    while (low < high) {
+        int const idx = HUF_quickSortPartition(arr, low, high);
+        if (idx - low < high - idx) {
+            HUF_simpleQuickSort(arr, low, idx - 1);
+            low = idx + 1;
+        } else {
+            HUF_simpleQuickSort(arr, idx + 1, high);
+            high = idx - 1;
+        }
+    }
+}
+
 /**
  * HUF_sort():
  * Sorts the symbols [0, maxSymbolValue] by count[symbol] in decreasing order.
+ * This is a typical bucket sorting strategy that uses either quicksort or insertion sort to sort each bucket.
  *
  * @param[out] huffNode       Sorted symbols by decreasing count. Only members `.count` and `.byte` are filled.
  *                            Must have (maxSymbolValue + 1) entries.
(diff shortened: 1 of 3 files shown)


lib/compress/huf_compress.c
```diff
@@ -367,67 +367,161 @@ static U32 HUF_setMaxHeight(nodeElt* huffNode, U32 lastNonNull, U32 maxNbBits)
...
+    for (n = RANK_POSITION_DISTINCT_COUNT_CUTOFF; n < RANK_POSITION_TABLE_SIZE - 1; ++n) {
+        U32 const bucketSize = rankPosition[n].curr-rankPosition[n].base;
+		U32 const bucketStartIdx = rankPosition[n].base;
+        if (bucketSize <= 1) {
+            continue;
+        } else if (bucketSize <= 128) {
+            HUF_insertionSort(huffNode + bucketStartIdx, bucketSize);
+        } else {
+            HUF_simpleQuickSort(huffNode + bucketStartIdx, 0, bucketSize-1);
+        }
```
terrelln: I'd recommend just using a single sort function. Generally, fast quick sorts will look at array size, and for small array sizes will fall back to insertion sort. So the quick sort could detect arrays of < K (K ~5 maybe) elements and fall back to insertion sort.

senhuang42: So I think it's a net positive change, but it's hard to tell, really. 6 elements experimentally seems like an okay cutoff point. With using insertion sort within quicksort for calls of less than 6 elements, we see the following (cfeccb43 being the version with your suggestion):

|     Benchmark     | Config  |   Dataset   | Ratio (5accf234) | Ratio (cfeccb43) | Ratio (cfeccb43 - 5accf234) | Speed MB/s (5accf234) | Speed MB/s (cfeccb43) | Speed MB/s (cfeccb43 - 5accf234) |
|-------------------|---------|-------------|------------------|------------------|-----------------------------|-----------------------|-----------------------|----------------------------------|
| compress          | level_1 | enwik7      | 2.43             | 2.43             | -0.0%                       | 260.90                | 259.36                | -0.6%                            |
| compress          | level_1 | enwik7_1k   | 1.75             | 1.75             | 0.0%                        | 115.55                | 117.16                | 1.4%                             |
| compress          | level_1 | silesia     | 2.88             | 2.88             | 0.0%                        | 349.45                | 347.77                | -0.5%                            |
| compress          | level_1 | silesia_16k | 2.62             | 2.62             | -0.0%                       | 247.51                | 249.12                | 0.7%                             |
| compress          | level_1 | silesia_4k  | 2.31             | 2.31             | 0.0%                        | 204.78                | 205.35                | 0.3%                             |
| compress          | level_3 | enwik7      | 2.79             | 2.79             | 0.0%                        | 144.61                | 142.33                | -1.6%                            |
| compress          | level_3 | enwik7_1k   | 1.75             | 1.75             | -0.0%                       | 87.10                 | 87.85                 | 0.9%                             |
| compress          | level_3 | silesia     | 3.18             | 3.18             | 0.0%                        | 190.72                | 187.68                | -1.6%                            |
| compress          | level_3 | silesia_16k | 2.67             | 2.67             | 0.0%                        | 182.81                | 183.10                | 0.2%                             |
| compress          | level_3 | silesia_4k  | 2.35             | 2.35             | -0.0%                       | 154.32                | 154.96                | 0.4%                             |
| compress          | level_7 | enwik7      | 3.05             | 3.05             | 0.0%                        | 66.42                 | 66.32                 | -0.1%                            |
| compress          | level_7 | enwik7_1k   | 1.79             | 1.79             | -0.0%                       | 52.59                 | 53.06                 | 0.9%                             |
| compress          | level_7 | silesia     | 3.45             | 3.45             | -0.0%                       | 87.01                 | 86.72                 | -0.3%                            |
| compress          | level_7 | silesia_16k | 2.80             | 2.80             | -0.0%                       | 51.72                 | 52.20                 | 0.9%                             |
| compress          | level_7 | silesia_4k  | 2.42             | 2.42             | 0.0%                        | 62.42                 | 62.49                 | 0.1%                             |
| compress_literals | level_1 | enwik7      | 1.49             | 1.49             | -0.0%                       | 783.73                | 778.03                | -0.7%                            |
| compress_literals | level_1 | enwik7_1k   | 1.46             | 1.46             | 0.0%                        | 264.67                | 271.91                | 2.7%                             |
| compress_literals | level_1 | silesia     | 1.30             | 1.30             | 0.0%                        | 604.35                | 604.73                | 0.1%                             |
| compress_literals | level_1 | silesia_16k | 1.29             | 1.29             | 0.0%                        | 558.75                | 557.81                | -0.2%                            |
| compress_literals | level_1 | silesia_1k  | 1.28             | 1.28             | 0.0%                        | 265.02                | 272.08                | 2.7%                             |
| compress_literals | level_1 | silesia_4k  | 1.29             | 1.29             | 0.0%                        | 276.64                | 283.87                | 2.6%                             |
| compress_literals | level_3 | enwik7      | 1.34             | 1.34             | 0.0%                        | 640.60                | 630.37                | -1.6%                            |
| compress_literals | level_3 | enwik7_1k   | 1.40             | 1.40             | -0.0%                       | 201.50                | 205.74                | 2.1%                             |
| compress_literals | level_3 | silesia     | 1.17             | 1.17             | -0.0%                       | 499.56                | 501.60                | 0.4%                             |
| compress_literals | level_3 | silesia_16k | 1.21             | 1.21             | 0.0%                        | 492.30                | 492.93                | 0.1%                             |
| compress_literals | level_3 | silesia_1k  | 1.23             | 1.23             | 0.0%                        | 223.39                | 228.90                | 2.5%                             |
| compress_literals | level_3 | silesia_4k  | 1.24             | 1.24             | 0.0%                        | 239.28                | 244.97                | 2.4%                             |
| compress_literals | level_7 | enwik7      | 1.29             | 1.29             | 0.0%                        | 567.40                | 557.35                | -1.8%                            |
| compress_literals | level_7 | enwik7_1k   | 1.39             | 1.39             | -0.0%                       | 189.78                | 193.95                | 2.2%                             |
| compress_literals | level_7 | silesia     | 1.15             | 1.15             | -0.0%                       | 461.80                | 464.98                | 0.7%                             |
| compress_literals | level_7 | silesia_16k | 1.21             | 1.21             | -0.0%                       | 477.96                | 478.90                | 0.2%                             |
| compress_literals | level_7 | silesia_1k  | 1.22             | 1.22             | 0.0%                        | 215.32                | 220.73                | 2.5%                             |
| compress_literals | level_7 | silesia_4k  | 1.23             | 1.23             | 0.0%                        | 232.33                | 237.94                | 2.4%                             |


The main issue just seems to be that e2e compression on normal sized blocks seems to be consistently slower by a bit. It doesn't really make sense that that should happen since literals compression is typically still faster by a bit, so that might just be mostly noise, but it happening 6/6 times we compress 128K blocks is worrying.

lib/compress/huf_compress.c
```diff
@@ -367,67 +367,161 @@ static U32 HUF_setMaxHeight(nodeElt* huffNode, U32 lastNonNull, U32 maxNbBits)
 }
 
 typedef struct {
-    U32 base;
-    U32 curr;
+    U16 base;
+    U16 curr;
```
Cyan4973, line 432: note : accessing 32-bit values _tends_ to be faster than accessing 16-bit values on modern cpu architectures.

Not sure if this impact is large enough to be measurable though, but it's easy enough to test.

lib/compress/huf_compress.c
senhuang42: Update: 32 elements actually seems like a better breakpoint. I've compared 8, 16, and 32 element thresholds. You can see the raw data here: https://pastebin.com/raw/jWd7Gu8R

I will try and see if some of the insertion sort optimizations in glibc's `qsort()` provide any additional benefit

lib/compress/huf_compress.c
senhuang42: Interesting, I'm actually seeing (slightly) better performance when using 16-bit values.

Cyan4973: Well, there is also a benefit, which is a smaller memory footprint of the associated array, resulting in reduced L1 cache occupation.
So the cache benefit might outpace the access deficit (`movzwl` vs `movl`).

In any case, it's the measurement which matters, and ensuring that the 16-bit has better (or equivalent) speed is the right move.

tests/regression/results.csv
```diff
@@ -11,10 +11,10 @@ silesia.tar,                        level 6,                            compress
 silesia.tar,                        level 7,                            compress simple,                    4613541
 silesia.tar,                        level 9,                            compress simple,                    4555426
 silesia.tar,                        level 13,                           compress simple,                    4491764
-silesia.tar,                        level 16,                           compress simple,                    4356834
-silesia.tar,                        level 19,                           compress simple,                    4264392
```
senhuang42, line 15: Investigating why this regression happens (and compress simple and compress cctx is the only ones that have these large regressions).

terrelln: `uncompressed literals optimal` is also regressing. I suspect that you maybe are writing beyond the end of your workspace?

lib/compress/huf_compress.c
```diff
@@ -368,67 +368,173 @@ static U32 HUF_setMaxHeight(nodeElt* huffNode, U32 lastNonNull, U32 maxNbBits)
...
+#define RANK_POSITION_LOG_BUCKETS_BEGIN (RANK_POSITION_TABLE_SIZE - 1) - BIT_highbit32(HUF_BLOCKSIZE_MAX) - 1
+#define RANK_POSITION_DISTINCT_COUNT_CUTOFF RANK_POSITION_LOG_BUCKETS_BEGIN + BIT_highbit32(RANK_POSITION_LOG_BUCKETS_BEGIN)
+
+/* Return the appropriate bucket index for a given count. See definition of
+ * RANK_POSITION_DISTINCT_COUNT_CUTOFF for explanation of bucketing strategy.
+ */
+static U32 HUF_getIndex(U32 const count) {
+    return (count < RANK_POSITION_DISTINCT_COUNT_CUTOFF)
+        ? count
+        : BIT_highbit32(count+1) + RANK_POSITION_LOG_BUCKETS_BEGIN;
```
terrelln: You shouldn't need a `+1` here, right? Since we know the count isn't zero.

senhuang42: When I did this, I noticed between a 0.4-0.6% slowdown on level 1, 4KB block silesia.tar so I left it in, so it seems like a maybe a code alignment issue? I guess it would be better to not leave it in just for the sake of that perturbation.

tests/regression/results.csv
senhuang42: It seems to just have been that I never ran `make clean` in the regression test when I still was on the revision with the correctness bug in the insertion sort.

tests/regression/results.csv
```diff
@@ -2,19 +2,19 @@ Data,                               Config,                             Method,
 silesia.tar,                        level -5,                           compress simple,                    6738593
 silesia.tar,                        level -3,                           compress simple,                    6446372
 silesia.tar,                        level -1,                           compress simple,                    6186042
-silesia.tar,                        level 0,                            compress simple,                    4861425
```
Cyan4973, line 5: My understanding of this PR is that it improves speed of the Huffman Sorting stage,
resulting in better compression, visible on small blocks.

However, regression tests show that it also (slightly) alters compression ratio results,
suggesting that there is more to this PR than just improving the speed of the sorting stage.

Can you explain this effect ?

senhuang42: Now that we use a hybrid quicksort/insertion sort instead of raw insertion sort, the sorting algorithm is not stable, so symbols with the same frequency might shift around into different positions than they used to be in just a raw insertion sort. It doesn't matter for correctness though (since huffman just theoretically cares about the frequency), and is still deterministic.

Cyan4973: > is still deterministic

I think that's the important point. As long as it's deterministic, I'm fine with the very small impact.

terrelln: Since the counts are exactly the same for symbols that are shuffled, we would expect the difference in compressed size to come from the Huffman header, not the stream. I guess from weights appearing in different order in the FSE compressed weights.

I wonder if there is some (small) opportunity to gain compression ratio by using a heuristic to predict which which symbol to prefer to give a higher weight to, to get a better FSE ratio. E.g. prefer to give higher weights to smaller symbols.

Cyan4973: Yeah, I also expect the difference to come from the FSE encoding of the Huffman header.

The problem here is that the succession of state values is rather chaotic, and therefore doesn't lend itself well to any kind of "cheap heuristic" prediction.
If left with only brute-force to discover the smallest succession of states, and considering the tiny differences at stake, the efforts looks outsized compared to the benefit.

Cyan4973 (approved):

terrelln (approved):

Issue #2739, opened 2021-08-03
zstd 1.5.0 no longer preserves modification time


0x9fff00:
**Describe the bug**
In zstd 1.4.9 and earlier, the modification time of a compressed file was the same as the uncompressed file, but in 1.5.0 this is no longer the case and the modification time is set to the compression time.

**To Reproduce**
1. Create a file: `touch testfile`
2. Compress it: `zstd testfile`
3. Compare the modification times: `stat testfile{,.zst}`

Using zstd 1.5.0, testfile and testfile.zst have different modification times.

**Expected behavior**
The modification times would be the same, like in zstd 1.4.9. gzip, bzip2 and xz also have this behavior.

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Version: zstd 1.5.0
 - Compiler: gcc 11.1.0
 - Flags: `CFLAGS="-march=native -mtune=native -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection"`, `CXXFLAGS="$CFLAGS -Wp,-D_GLIBCXX_ASSERTIONS"`, `LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"`
 - Build system: Makefile

**Additional context**
`git bisect` shows that b87f97b3ea5a51eb884a62849cf752187bfb8004 is the first commit with this issue. The associated pull request (#2525) only mentions file permissions and not modification times.

terrelln:
Thanks for the bug report!

Looks like we're missing a test case that checks that we don't change the `mtime`. We'll have to add that when we fix this issue.

PR #2742, opened 2021-08-04
Set mtime on Output Files

felixhandte:
In changing how permission bits were set on created files, #2525 introduced a regression, no longer updating the mtime on output files. This PR fixes this regression, by calling `utime()` on destination files.

This addresses #2739.

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -597,11 +597,10 @@ fi
 
 println "\n===>  zstd created file timestamp tests"
 datagen > tmp
-sleep 1 # could also use touch -t but I don't know how portable that is.
+touch -m -t 200001010000.00 tmp
 println "test : copy mtime in file -> file compression "
 zstd -f tmp -o tmp.zst
 assertSameMTime tmp tmp.zst
-sleep 1 # could also use touch -t but I don't know how portable that is.
 println "test : copy mtime in file -> file decompression "
 zstd -f -d tmp.zst -o tmp.out
 assertSameMTime tmp.zst tmp.out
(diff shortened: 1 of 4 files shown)


tests/playTests.sh
```diff
@@ -124,6 +124,13 @@ case "$UNAME" in
     Darwin | FreeBSD | OpenBSD | NetBSD) MTIME="stat -f %m" ;;
 esac
 
+assertSameMTime() {
```
Cyan4973, line 127: 👍

tests/playTests.sh
```diff
@@ -588,6 +595,18 @@ if [ -n "$READFROMBLOCKDEVICE" ] ; then
     rm -f tmp.img tmp.img.zst tmp.img.copy
 fi
 
+println "\n===>  zstd created file timestamp tests"
+datagen > tmp
+sleep 1 # could also use touch -t but I don't know how portable that is.
```
Cyan4973: note : 
each `sleep 1` adds 1 full second to this test.
This may not sound like much, but within the context of `playTests.sh`, which is run very often, even locally, we try to keep the test set "relatively" fast, and since there are a lot of tests to run , each test is supposed to only cost a fraction of a second.
When it doesn't, we generally move the test to the "long tests" category.

Possible mitigation : GNU `sleep` is supposed to accept floating point format.
So `sleep 0.01` would do.
However, I'm also not sure how portable that is.

felixhandte: The comparison we're making is at 1-second resolution. :(

But maybe it's worth trying the `touch` route to set an artificial mtime.

Cyan4973 (approved):

terrelln (approved):

PR #2746, opened 2021-08-09
meson fixups

eli-schwartz:
Two small changes:
- with regard to #2261, fix linking libzstd.so with private symbol visibility, emulating the Makefile build with zstd-dll, which currently only "works" because the meson.build incorrectly failed to set private visibility
- use more idiomatic, warning-free ways to configure meson to use -Wextra -pedantic -Werror

### DIFF
--- a/build/meson/contrib/pzstd/meson.build
+++ b/build/meson/contrib/pzstd/meson.build
@@ -18,7 +18,7 @@ pzstd_sources = [join_paths(zstd_rootdir, 'programs/util.c'),
   join_paths(zstd_rootdir, 'contrib/pzstd/SkippableFrame.cpp')]
 pzstd = executable('pzstd',
   pzstd_sources,
-  cpp_args: [ '-DNDEBUG', '-Wno-shadow', '-pedantic', '-Wno-deprecated-declarations' ],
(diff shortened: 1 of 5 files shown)


eli-schwartz:
-Werror fails because `make clangbuild` fails but isn't tested in CI. I guess assuming that use of -Werror in the other build systems is worth emulating for parity, was a bad assumption...

EDIT: also this is generally impossible if you build the tests too.

eli-schwartz:
In the end I had to kind of rip out the current shared library linkage in tests/ because I don't think this was ever going to work. So all the test binaries now link to libtestcommon.a, a convenience library shared between test programs which also includes a copy of all .o files from libzstd. Since they are .o files, the fact that they're built with `-fvisibility=hidden` doesn't matter, and the private symbols successfully link.

eli-schwartz:
New in this force-push is commit ee3355df30752456481541d6e120f60fd1f81e8f, which should actually slightly speed up building by sharing some object files which are built identically for multiple test binaries, and can now be shared.

eli-schwartz:
Please let this be the last push...  if pthread is a dependency of libzstd, it needs to be added to the convenience library too. But this worked flawlessly locally... because I'm running a development version of glibc which has moved libpthread into libc.so and is no longer required to be separately linked, lol.

Cyan4973 (approved):

PR #2747, opened 2021-08-13
Add AIX support in Makefile

Helflym:
This PR adds two patches for AIX. 
One for adding the general support of AIX inside Makefile. 
One to remove the test dealing with /dev/full for AIX.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -212,7 +212,11 @@ ifeq ($(UNAME), Darwin)
   SHARED_EXT_VER = $(LIBVER).$(SHARED_EXT)
   SONAME_FLAGS = -install_name $(LIBDIR)/libzstd.$(SHARED_EXT_MAJOR) -compatibility_version $(LIBVER_MAJOR) -current_version $(LIBVER)
 else
-  SONAME_FLAGS = -Wl,-soname=libzstd.$(SHARED_EXT).$(LIBVER_MAJOR)
+  ifeq ($(UNAME), AIX)
+    SONAME_FLAGS =
+  else
+    SONAME_FLAGS = -Wl,-soname=libzstd.$(SHARED_EXT).$(LIBVER_MAJOR)
+  endif
   SHARED_EXT = so
   SHARED_EXT_MAJOR = $(SHARED_EXT).$(LIBVER_MAJOR)
   SHARED_EXT_VER = $(SHARED_EXT).$(LIBVER)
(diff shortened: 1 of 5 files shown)


lib/Makefile
```diff
@@ -212,7 +212,11 @@ ifeq ($(UNAME), Darwin)
   SHARED_EXT_VER = $(LIBVER).$(SHARED_EXT)
   SONAME_FLAGS = -install_name $(LIBDIR)/libzstd.$(SHARED_EXT_MAJOR) -compatibility_version $(LIBVER_MAJOR) -current_version $(LIBVER)
 else
-  SONAME_FLAGS = -Wl,-soname=libzstd.$(SHARED_EXT).$(LIBVER_MAJOR)
+  ifeq ($(UNAME), AIX)
+    SONAME_FLAGS =
```
Cyan4973, line 216: That part feels weird.
What happens to dynamic libraries in the context of AIX ?

Helflym: AIX linker doesn't simply understand -soname flag. GCC when calling the linker will pass the right arguments as long as "-shared" is provided. The "-o" flag is enough to have the correct soname.

Cyan4973:
Thanks for this proposed PR.
It looks good to me.

Note though that the `CLA` signing stage is mandatory before proceeding to merge the PR.

lib/Makefile
```diff
@@ -374,7 +378,7 @@ clean:
 #-----------------------------------------------------------------------------
 # make install is validated only for below listed environments
 #-----------------------------------------------------------------------------
-ifneq (,$(filter $(UNAME),Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS Haiku))
+ifneq (,$(filter $(UNAME),Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS Haiku AIX))
```
Cyan4973, line 381: tangentially related :
It seems we have the same list of `uname` identifiers in 3 different places (for the purpose of controlling `make install` compatibility).
We could probably consolidate this list in one place and distribute it, making its maintenance and evolution a bit easier.
cc @felixhandte .

Helflym:
> Note though that the `CLA` signing stage is mandatory before proceeding to merge the PR.

Just to give an update, I'm still waiting for the CLA to be signed by my company.

PR #2753, opened 2021-08-23
[easy] Fix zstd bench error message

senhuang42:
I remember this confusing me the first time I saw it, so should probably fix it.
Prior to this fix, if you ran
```
./zstd -b1 nonexistent_file
Not enough memory; testing 0 MB only...
Error 12 : not enough memory 
```
Now the message is more accurate
```
Error 9 : Error loading files
```


Cyan4973 (approved):

Cyan4973:
Thanks @senhuang42 ! This is indeed a small yet welcome improvement !

Issue #2756, opened 2021-08-28
m68k-unknown-linux-gnu: zstd-1.4.9 test suite fails badly


akhuettel:
I've been trying to build zstd on m68k-unknown-linux-gnu (since it is a dependency of the Gentoo package manager). (More precisely, a qemu-user chroot, but that should not have too much impact here.)

Unfortunately, while the build itself runs fine and the command line utility seems to start, the results in the zstd-1.4.9 testsuite look rather messy.
(1.5.0 does not work at all, bailing out with a "generic error").

I've attached the test log. Advice on how serious this is would be appreciated...

[build.log](https://github.com/facebook/zstd/files/7071186/build.log)

Cyan4973:
We have never tested `zstd` on such a platform.
And to be honest, I'm impressed `v1.4.9` runs so well on this first attempt.

Beyond that, it's too difficult to debug. There are myriads of things that could go wrong at runtime, especially in the CLI which features a number of system calls beyond the library itself (I expect `libzstd` to be very portable, but the CLI application is another story).

To ensure `libzstd` and `zstd` work on `m68k`, we would need to add this target to our CI.
This _might_ be possible if it's an available target of `qemu-static`.

chewi:
It is indeed a user (and system) target of QEMU.

Cyan4973:
OK, I could reproduce a qemu-m68k config locally,
and indeed, there is a problem at some point during the test : 

```
../lib/compress/zstd_compress.c: Building LL table 
../lib/compress/zstd_compress_sequences.c: Selected set_compressed 
../lib/compress/zstd_compress_sequences.c: ZSTD_buildCTable (dstCapacity=113538) 
../lib/compress/zstd_compress_sequences.c:281: ERROR!: forwarding error in FSE_buildCTable_wksp(nextCTable, wksp->norm, max, tableLog, wksp->wksp, sizeof(wksp->wksp)): Error (generic): 
../lib/compress/zstd_compress.c: ZSTD_buildCTable for LitLens failed 
../lib/compress/zstd_compress.c:2654: ERROR!: forwarding error in stats.size: Error (generic): ZSTD_buildSequencesStatistics failed!
../lib/compress/zstd_compress.c:2711: ERROR!: forwarding error in cSize: Error (generic): ZSTD_entropyCompressSeqStore_internal failed
```

So it's not especially trivial, and I'm unable to tell what's wrong just from the traces, but at least, it's reproducible. So there's a likely way forward.

[...]

PR #2758, opened 2021-08-29
added qemu tests

Cyan4973:
This PR adds basic correctness test (`make check`) on several non-x86 platforms emulated by `qemu` : 
- ARM
- ARM64
- PPC
- PPC64LE
- S390X
- MIPS
- M68K

This PR is inspired by the work of @t-mat on [xxHash](https://github.com/Cyan4973/xxHash/commit/1528d7c88352c7b23618746e184d4cff97604445).
Several tested platforms feature big-endian cpus, which is a good check. They all seem to work correctly, but the last one (`m68k`) is designed to not report errors yet, since it has ongoing [reported alignment issues](https://github.com/facebook/zstd/issues/2756) (reproduced locally).

### DIFF
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -78,12 +78,11 @@ size_t FSE_buildCTable_wksp(FSE_CTable* ct,
     U32 const maxSV1 = maxSymbolValue+1;
 
     U16* cumul = (U16*)workSpace;   /* size = maxSV1 */
-    FSE_FUNCTION_TYPE* tableSymbol = (FSE_FUNCTION_TYPE*)(cumul + (maxSV1+1));  /* size = tableSize */
-    BYTE* spread = tableSymbol + tableSize; /* size = tableSize */
+    FSE_FUNCTION_TYPE* const tableSymbol = (FSE_FUNCTION_TYPE*)(cumul + (maxSV1+1));  /* size = tableSize */
 
     U32 highThreshold = tableSize-1;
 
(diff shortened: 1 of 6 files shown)


senhuang42 (approved):

terrelln (approved):

lib/compress/fse_compress.c
```diff
@@ -205,26 +206,16 @@ size_t FSE_buildCTable_wksp(FSE_CTable* ct,
...
-        }
-    }
+    }   }
 #endif
 
     return 0;
 }
 
-#ifndef ZSTD_NO_UNUSED_FUNCTIONS
-size_t FSE_buildCTable(FSE_CTable* ct, const short* normalizedCounter, unsigned maxSymbolValue, unsigned tableLog)
```
danlark1, line 216: Is FSE_buildCTable expected to be deleted from .c file but not from the header file? I believe it's expected to stay here :)

Cyan4973: It should be deleted from `*.h` file too.

What is happening is that we stop trying to keep parity between the stand-alone `libfse` version and the one integrated into `zstd`. All entry points that are not employed in `zstd` will be deleted in the near future, making this variant effectively no longer extractable.

What prompted the deletion of this specific entry point is that it was wrong and thus misleading : the amount of stack memory it allocates is no longer correct, as the underlying invoked function has changed. Using this function "as is" would lead to stack memory corruption. Yet, this issue was missed, because this entry is simply never used and never tested. 

Since it is not a sane situation, we are moving out from there, and making the maintenance workload more focused on what matters for `zstd`.

PR #2760, opened 2021-08-31
fixes regression of #2692, proper 64 row entry support for lazy

sebres:
**reserved**

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1797,7 +1797,7 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
             if (ms->tagTable) ZSTD_memset(ms->tagTable, 0, tagTableSize);
         }
         {   /* Switch to 32-entry rows if searchLog is 5 (or more) */
-            U32 const rowLog = MAX(MIN(cParams->searchLog, 6), 4);
+            U32 const rowLog = MAX(MIN(cParams->searchLog, 4), 6);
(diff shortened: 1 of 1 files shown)


senhuang42:
I don't think this is quite right.

Originally:
If `slog = 12`.
Min(12, 6) = 6
Max(6, 4) = 6.
`slog = 6`

If `slog = 1`.
Min(1, 6) = 1.
Max(1, 4) = 4.
`slog = 4`

If `slog = 5`
Min(5, 6) = 5
Max(5, 4) = 5
`slog = 5`.

Now:
If `slog = 5`.
Min(5, 4) = 4
Max(4, 6) = 6.
`slog = 6` <- this is incorrect.

sebres:
Yeah, my mistake... already seen... (it is always 6 now for slog 4..6).. Sorry for noise.
I put it to draft firstly, because strange but the regression is away (e. g. it is faster now in -T4)... I'll look further.

sebres:
FWIW, this does not show the regression at level 6/7 (and 8/9) to me anymore:
```diff
-            U32 const rowLog = MAX(MIN(cParams->searchLog, 6), 4);
+            U32 const rowLog = cParams->searchLog < 5 ? 4 : 6;
```
but it looks like some secondary reason for that, primarily the increase of `slog` still the issue, which significantly slowing down MT-mode for greedy/lazy strategies. So one need to investigate deeper. 
This one is closed now as a mistake.

PR #2761, opened 2021-09-01
Add 8 bytes to FSE_buildCTable wksp

senhuang42:
Writes into the `spread` table in `FSE_buildCTable_wksp()` may overwrite by up to 8 bytes, so enlarge the wksp by 8 bytes to account for this.

### DIFF
--- a/lib/common/fse.h
+++ b/lib/common/fse.h
@@ -336,8 +336,9 @@ size_t FSE_buildCTable_rle (FSE_CTable* ct, unsigned char symbolValue);
 /* FSE_buildCTable_wksp() :
  * Same as FSE_buildCTable(), but using an externally allocated scratch buffer (`workSpace`).
  * `wkspSize` must be >= `FSE_BUILD_CTABLE_WORKSPACE_SIZE_U32(maxSymbolValue, tableLog)` of `unsigned`.
+ * See FSE_buildCTable_wksp() for breakdown of workspace usage.
  */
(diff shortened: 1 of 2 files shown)


lib/common/fse.h
```diff
@@ -336,8 +336,9 @@ size_t FSE_buildCTable_rle (FSE_CTable* ct, unsigned char symbolValue);
 /* FSE_buildCTable_wksp() :
  * Same as FSE_buildCTable(), but using an externally allocated scratch buffer (`workSpace`).
  * `wkspSize` must be >= `FSE_BUILD_CTABLE_WORKSPACE_SIZE_U32(maxSymbolValue, tableLog)` of `unsigned`.
+ * See FSE_buildCTable_wksp() for breakdown of workspace usage.
  */
-#define FSE_BUILD_CTABLE_WORKSPACE_SIZE_U32(maxSymbolValue, tableLog) (((maxSymbolValue + 2) + (1ull << (tableLog)))/2)
+#define FSE_BUILD_CTABLE_WORKSPACE_SIZE_U32(maxSymbolValue, tableLog) (((maxSymbolValue + 2) + (1ull << (tableLog)))/2 + 8/sizeof(unsigned) /* additional 8 bytes for potential table overwrite */)
```
terrelln: nit: It is somewhat odd to use both 8 and `sizeof(unsigned)`, since 8 comes from a 64-bit write.

Either just add `8 / 2`, or `sizeof(U64) / sizeof(U32)`. Especially because you can include `mem.h` in `fse.h` now, since it is no longer a public header.

terrelln (approved):

Cyan4973:
And apparently no "stack smashing" issue on `s390x` detected this time. Good sign.

PR #2763, opened 2021-09-01
Improve compile speed and binary size in `opt`

senhuang42:
Remove basically all inlining/templating in opt.
I measure no impact to end-to-end compression speed at all (levels 13-16, `silesia.tar`).

```
macOS with gcc

dev:
zstdcli binary size: 1703976
time to compile opt.o: ~8.5s

opt_compiletime:
zstdcli binary size: 1458352  (14.5% reduction)
time to compile opt.o: ~1.8s
```

The bigger bottleneck is still lazy, which will require a bit more work to reduce without speed regressions.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -253,11 +253,10 @@ static U32 ZSTD_litLengthPrice(U32 const litLength, const optState_t* const optP
  * Provides the cost of the match part (offset + matchLength) of a sequence
  * Must be combined with ZSTD_fullLiteralsCost() to get the full cost of a sequence.
  * optLevel: when <2, favors small offset for decompression speed (improved cache efficiency) */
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

terrelln (approved):

Cyan4973:
> I measure no impact to end-to-end compression speed at all (levels 13-16, silesia.tar).

I checked that statement.
To begin with, `zstd_opt` only starts at level 16 for large files, so obviously levels 13-15 will remain identical after this change.

Then, from level 16 to 22, I measured a -5 to -10% speed drop across a few scenarios on my Ubuntu station (using `gcc-9.3`).

I also measured the mentioned binary size reduction, as well as the sharp compilation time reduction.

So now it's a matter of trade-off, since it's not completely free.

PR #2771, opened 2021-09-08
Improve optimal parser performance on small data

### LINKED ISSUE #2765: Compression ratio example
Hello,

first time working with zstd, I'm testing it and found a little problem.

I have 3 files: "part1.bin", "part2.bin" and a combined "bothparts.bin".

part1 is compressed to 500 bytes, part2 to 30 bytes, so I expected
the combined file to be compressed to around 530, but it is 1k bytes.

Don't know if you guys are aware of this... the files are here:
http://users.beotel.net/~gwh/zstd/zstd.zip

(both rar and zip are good with these example files)

Cyan4973:
This investigation was driven by #2765 .

In essence, this user sample contains a file which, depending on early choices made by the optimal parser, end up compressing or not the data, resulting in large relative differences (+100%).

This is a case of self-reinforcing local optimal. In order to find the shortest path (i.e. which leads to better compression), the algorithm needs to evaluate the "cost" of its choices. For this sample to work, the algorithm must consider small matches as better than literals. If it does, its weight evaluation will favor future choices of the same kind. If not, it will keep considering them too expensive.

One of the known issues of the optimal parser is that it doesn't know which stats to start from. Over time, the process tends to self-regulate, but in the early stages, there is not enough statistics collected yet.
The initially selected approach was to start from a blank pages, with all choices essentially equivalent.
This is what this PR modifies.

Initial statistics are modified this way : 
- Nudge literal lengths slightly towards less or no literals : this small change to initial conditions has an outsized impact, featuring > 100 KB savings on `silesia.tar` cut into small blocks. It's not clear why, I suspect it merely corrects another side effect which tends to favor more literals than it should. This change, all by itself, fix #2765.
- Nudge offset codes towards short distances and repeat codes : this one has a small positive effect. I was expecting more impact. It only improves `silesia.tar` by a few dozens of KB. That might be because, due to the total cost including supplementary bits, the algorithm might already be biased enough ( in the general case ) in favor of small offsets and repeat code. Still, since the impact was consistently positive, this change was kept. It also fixes, all by itself, #2765, proving there is more than one way to fix that one.
- Revisit statistics update : this point is supposed to be mostly important for large files, since it impacts the transport of statistics between blocks. But it also impacts `btultra2`, which compress a small block twice, the first time merely to collect statistics for the second time. The issue is that the general update algorithm was presumed to be invoked once every 128 KB block. On small inputs, it's invoked twice in a much shorter timeframe, having no time to collect a "full block" of statistics. The issue is that the rescaling factor is static, and will squash down statistics the same way, no matter if they are numerous or not. In this new method, the rescaling is a bit more dynamic, trying to maintain a certain baseline budget between consecutive blocks, and adapting the squashing depending on actual runtime amounts. One issue is that there is no "one size fits all". Files with homogeneous entropy prefer slow adaptation, hence larger inter-block correlations, while files with heterogeneous sections prefer fast adaptation. I presume a more optimal solution should adapt the update rate depending on how likely statistics are to be representative of the following section. It's a topic for a full paper. For the time being, the simple heuristics selected in this PR will do (See benchmarks below).
- Revisit early literals estimation : when starting a new frame from scratch, the algorithm takes some clues from actual statistics in the samples. It seems it wasn't such a good idea after all. Potential reasons are : there is a large difference between left-over from LZ sequences, and raw bytes directly present in the file, the most recurrent bytes might be present too often and overwhelm statistics unfairly, and possibly the remaining nb of bytes is sometimes too small to be compressed. In all of these cases, literals will be considered cheaper than they actually will be in the final compressed block. Here also, this topic should deserve a full revisit. But for this PR, the simple heuristic selected improves compression ratio enough.

So what are the benefits ? Let's do some benchmark : 

<details>
<summary> `silesia.tar`, cut into 2 KB blocks : </summary>

Level | File | dev | PR | diff | diff % | comment |  
-- | -- | -- | -- | -- | -- | -- | --
11 | silesia.tar -B2K | 95169105 | 95000131 | -168974 | -0.18% | opt | mml=4
12 | silesia.tar -B2K | 93413153 | 93017601 | -395552 | -0.42% |   | mml=3
13 | silesia.tar -B2K | 93108397 | 92691659 | -416738 | -0.45% | ultra |  
14 | silesia.tar -B2K | 92972963 | 92569885 | -403078 | -0.43% |   |  
15 | silesia.tar -B2K | 92950860 | 92547104 | -403756 | -0.43% |   |  
16 | silesia.tar -B2K | 92628560 | 92401081 | -227479 | -0.25% | ultra2 |  
17 | silesia.tar -B2K | 92516103 | 92288346 | -227757 | -0.25% |   |  
18 | silesia.tar -B2K | 92513639 | 92285819 | -227820 | -0.25% |   |  
19 | silesia.tar -B2K | 92513546 | 92285727 | -227819 | -0.25% |   |  
20 | silesia.tar -B2K | 92511223 | 92284095 | -227128 | -0.25% |   |  
21 | silesia.tar -B2K | 92510965 | 92283782 | -227183 | -0.25% |   |  
22 | silesia.tar -B2K | 92510917 | 92283698 | -227219 | -0.25% |   |  
</details>

This round is consistently positive. It's even better when small matches of length 3 can be selected (levels 12+). The benefit is less pronounced on reaching `ultra2` levels. The generous interpretation is that `ultra2` was already designed to compensate for the "early statistics" problem, so it benefits less from this new round of improvement. Still, it benefits a little, which is almost surprising, and points at a self-reinforcing effect.

OK, but maybe 2 KB is a too favorable scenario, and the new statistics were designed for this one use case ? Let's try some bigger 16 KB blocks then.

<details>
<summary> `silesia.tar`, cut into 16 KB blocks : </summary>

Level | File | dev | PR | diff | diff % | comment |  
-- | -- | -- | -- | -- | -- | -- | --
11 | silesia.tar -B16K | 75619806 | 75488340 | -131466 | -0.17% | opt | mml=4
12 | silesia.tar -B16K | 73648745 | 73392013 | -256732 | -0.35% |   | mml=3
13 | silesia.tar -B16K | 73257726 | 72967672 | -290054 | -0.40% | ultra |  
14 | silesia.tar -B16K | 73032188 | 72730261 | -301927 | -0.41% |   |  
15 | silesia.tar -B16K | 72968198 | 72663133 | -305065 | -0.42% |   |  
16 | silesia.tar -B16K | 72800467 | 72641294 | -159173 | -0.22% | ultra2 |  
17 | silesia.tar -B16K | 72577872 | 72413473 | -164399 | -0.23% |   |  
18 | silesia.tar -B16K | 72572119 | 72407890 | -164229 | -0.23% |   |  
19 | silesia.tar -B16K | 72571968 | 72407720 | -164248 | -0.23% |   |  
20 | silesia.tar -B16K | 72572244 | 72409779 | -162465 | -0.22% |   |  
21 | silesia.tar -B16K | 72571829 | 72409500 | -162329 | -0.22% |   |  
22 | silesia.tar -B16K | 72571211 | 72408727 | -162484 | -0.22% |   |  

</details>

Well, with bigger blocks, the impact seems a little less pronounced, but it's still there, proving initial stats remain useful after 2 KB.

OK, fine for small blocks. What about large blocks ? Full 128 KB ?

<details>
<summary> `silesia.tar`, cut into 128 KB blocks : </summary>

Level | File | dev | PR | diff | diff % | comment |  
-- | -- | -- | -- | -- | -- | -- | --
13 | silesia.tar -B128K | 65656825 | 65611779 | -45046 | -0.07% | opt | mml=4
14 | silesia.tar -B128K | 63675123 | 63462802 | -212321 | -0.33% |   | mml=3
15 | silesia.tar -B128K | 63282894 | 63059738 | -223156 | -0.35% |   |  
16 | silesia.tar -B128K | 63023353 | 62799390 | -223963 | -0.36% | ultra |  
17 | silesia.tar -B128K | 63007738 | 62783036 | -224702 | -0.36% |   |  
18 | silesia.tar -B128K | 63003547 | 62780536 | -223011 | -0.35% |   |  
19 | silesia.tar -B128K | 62798322 | 62743773 | -54549 | -0.09% | ultra2 |  
20 | silesia.tar -B128K | 62789783 | 62736999 | -52784 | -0.08% |   |  
21 | silesia.tar -B128K | 62785135 | 62739262 | -45873 | -0.07% |   |  
22 | silesia.tar -B128K | 62783801 | 62736795 | -47006 | -0.07% |   |  

</details> 

Note that, for these sizes, `zstd_btopt` only starts at level 13.
Anyway, the benefit is now clearly reduced. But reassuringly, it's still globally positive, so it's still worth it.

Does this change impact full-length files, and if yes how ?

<details>
<summary> `silesia.tar` whole file : </summary>

Level | File | dev | PR | diff | diff % | comment |  
-- | -- | -- | -- | -- | -- | -- | --
16 | silesia.tar | 55351684 | 55321640 | -30044 | -0.05% | opt | mml=4
17 | silesia.tar | 54284712 | 54275326 | -9386 | -0.02% |   | mml=3
18 | silesia.tar | 53420100 | 53427159 | 7059 | 0.01% | ultra |  
19 | silesia.tar | 52986703 | 52991722 | 5019 | 0.01% | ultra2 |  
20 | silesia.tar | 52577641 | 52574471 | -3170 | -0.01% |   |  
21 | silesia.tar | 52481750 | 52480144 | -1606 | 0.00% |   |  
22 | silesia.tar | 52460817 | 52458096 | -2721 | -0.01% |   |  

</details>

Well, barely. This time, size differences are barely significant. And note that they are not always positive. But wether they are or not, it doesn't matter, as the compression ratio is essentially equivalent.

OK, but let's try some different file now. Maybe `calgary.tar`, which is a collection of small files of different types appended together, resulting in rapidly changing statistics. How does it behave with this new update policy ? 

<details>
<summary> `calgary.tar` whole file : </summary>

Level | File | dev | PR | diff | diff % | comment |  
-- | -- | -- | -- | -- | -- | -- | --
16 | calgary.tar | 881311 | 881780 | 469 | 0.05% | opt | mml=4
17 | calgary.tar | 873376 | 874742 | 1366 | 0.16% |  | mml=3
18 | calgary.tar | 861665 | 863272 | 1607 | 0.19% | ultra
19 | calgary.tar | 859732 | 860927 | 1195 | 0.14% | ultra2
20 | calgary.tar | 859732 | 860927 | 1195 | 0.14% |  
21 | calgary.tar | 859598 | 860842 | 1244 | 0.14% |  
22 | calgary.tar | 859587 | 860789 | 1202 | 0.14% |  

</details>

Well, this time, it's rather negative. But thankfully, by very little.
This is a case where statistics update would benefit from being more actively changed, since 2 consecutive internal files can be very different (text, image, db, etc.). It could be fixed with a faster update rate, but alas, other use cases (like `silesia.tar`) would then suffer. This is a situation where selecting a "good enough" middle ground heuristic is all we can do before moving on to some more complex algorithm.

Anyway, an important point : it's not _always_ a win. This change is more targeted at small blocks (~2 KB), for which it's _generally_ a win. For larger files, it's less clear, but the impact is also less pronounced.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -79,21 +79,6 @@ static void ZSTD_setBasePrices(optState_t* optPtr, int optLevel)
 }
 
 
-/* ZSTD_downscaleStat() :
- * reduce all elements in table by a factor 2^(ZSTD_FREQ_DIV+malus)
- * return the resulting sum of elements */
-static U32 ZSTD_downscaleStat(unsigned* table, U32 lastEltIndex, int malus)
-{
-    U32 s, sum=0;
-    DEBUGLOG(5, "ZSTD_downscaleStat (nbElts=%u)", (unsigned)lastEltIndex+1);
-    assert(ZSTD_FREQ_DIV+malus > 0 && ZSTD_FREQ_DIV+malus < 31);
-    for (s=0; s<lastEltIndex+1; s++) {
-        table[s] = 1 + (table[s] >> (ZSTD_FREQ_DIV+malus));
-        sum += table[s];
-    }
-    return sum;
-}
-
 static U32 sum_u32(const unsigned table[], size_t nbElts)
 {
     size_t n;
@@ -102,6 +87,31 @@ static U32 sum_u32(const unsigned table[], size_t nbElts)
         total += table[n];
     }
     return total;
+}
+
+static U32 ZSTD_downscaleStats(unsigned* table, U32 lastEltIndex, U32 shift)
+{
+    U32 s, sum=0;
+    DEBUGLOG(5, "ZSTD_downscaleStats (nbElts=%u, shift=%u)", (unsigned)lastEltIndex+1, (unsigned)shift);
+    assert(shift < 30);
+    for (s=0; s<lastEltIndex+1; s++) {
+        table[s] = 1 + (table[s] >> shift);
+        sum += table[s];
+    }
+    return sum;
+}
+
+/* ZSTD_scaleStats() :
+ * reduce all elements in table is sum too large
+ * return the resulting sum of elements */
+static U32 ZSTD_scaleStats(unsigned* table, U32 lastEltIndex, U32 logTarget)
+{
+    U32 const prevsum = sum_u32(table, lastEltIndex+1);
+    U32 const factor = prevsum >> logTarget;
+    DEBUGLOG(5, "ZSTD_scaleStats (nbElts=%u, target=%u)", (unsigned)lastEltIndex+1, (unsigned)logTarget);
+    assert(logTarget < 30);
+    if (factor <= 1) return prevsum;
+    return ZSTD_downscaleStats(table, lastEltIndex, ZSTD_highbit32(factor));
 }
 
 /* ZSTD_rescaleFreqs() :
@@ -185,7 +195,7 @@ ZSTD_rescaleFreqs(optState_t* const optPtr,
             if (compressedLiterals) {
                 unsigned lit = MaxLit;
                 HIST_count_simple(optPtr->litFreq, &lit, src, srcSize);   /* use raw first block to init statistics */
-                optPtr->litSum = ZSTD_downscaleStat(optPtr->litFreq, MaxLit, 1);
+                optPtr->litSum = ZSTD_downscaleStats(optPtr->litFreq, MaxLit, 8);
             }
 
             {   unsigned const baseLLfreqs[MaxLL+1] = {
@@ -219,10 +229,10 @@ ZSTD_rescaleFreqs(optState_t* const optPtr,
     } else {   /* new block : re-use previous statistics, scaled down */
 
         if (compressedLiterals)
-            optPtr->litSum = ZSTD_downscaleStat(optPtr->litFreq, MaxLit, 1);
-        optPtr->litLengthSum = ZSTD_downscaleStat(optPtr->litLengthFreq, MaxLL, 0);
-        optPtr->matchLengthSum = ZSTD_downscaleStat(optPtr->matchLengthFreq, MaxML, 0);
-        optPtr->offCodeSum = ZSTD_downscaleStat(optPtr->offCodeFreq, MaxOff, 0);
+            optPtr->litSum = ZSTD_scaleStats(optPtr->litFreq, MaxLit, 10);
+        optPtr->litLengthSum = ZSTD_scaleStats(optPtr->litLengthFreq, MaxLL, 9);
+        optPtr->matchLengthSum = ZSTD_scaleStats(optPtr->matchLengthFreq, MaxML, 9);
+        optPtr->offCodeSum = ZSTD_scaleStats(optPtr->offCodeFreq, MaxOff, 9);
     }
 
     ZSTD_setBasePrices(optPtr, optLevel);
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -14,7 +14,6 @@
 
 
 #define ZSTD_LITFREQ_ADD    2   /* scaling factor for litFreq, so that frequencies adapt faster to new stats */
-#define ZSTD_FREQ_DIV       4   /* log factor when using previous stats to init next stats */
 #define ZSTD_MAX_PRICE     (1<<30)
 
 #define ZSTD_PREDEF_THRESHOLD 1024   /* if srcSize < ZSTD_PREDEF_THRESHOLD, symbols' cost is assumed static, directly determined by pre-defined distributions */
@@ -1253,28 +1252,7 @@ size_t ZSTD_compressBlock_btopt(
 }
 
 
-/* used in 2-pass strategy */
-static U32 ZSTD_upscaleStat(unsigned* table, U32 lastEltIndex, int bonus)
-{
-    U32 s, sum=0;
-    assert(ZSTD_FREQ_DIV+bonus >= 0);
-    for (s=0; s<lastEltIndex+1; s++) {
-        table[s] <<= ZSTD_FREQ_DIV+bonus;
-        table[s]--;
-        sum += table[s];
-    }
-    return sum;
-}
 
-/* used in 2-pass strategy */
-MEM_STATIC void ZSTD_upscaleStats(optState_t* optPtr)
-{
-    if (ZSTD_compressedLiterals(optPtr))
-        optPtr->litSum = ZSTD_upscaleStat(optPtr->litFreq, MaxLit, 0);
-    optPtr->litLengthSum = ZSTD_upscaleStat(optPtr->litLengthFreq, MaxLL, 0);
-    optPtr->matchLengthSum = ZSTD_upscaleStat(optPtr->matchLengthFreq, MaxML, 0);
-    optPtr->offCodeSum = ZSTD_upscaleStat(optPtr->offCodeFreq, MaxOff, 0);
-}
 
 /* ZSTD_initStats_ultra():
  * make a first compression pass, just to seed stats with more accurate starting values.
@@ -1305,8 +1283,6 @@ ZSTD_initStats_ultra(ZSTD_matchState_t* ms,
     ms->window.lowLimit = ms->window.dictLimit;
     ms->nextToUpdate = ms->window.dictLimit;
 
-    /* re-inforce weight of collected statistics */
-    ZSTD_upscaleStats(&ms->opt);
 }
 
 size_t ZSTD_compressBlock_btultra(
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -204,7 +204,7 @@ ZSTD_rescaleFreqs(optState_t* const optPtr,
                     1, 1, 1, 1, 1, 1, 1, 1,
                     1, 1, 1, 1
                 };
-                memcpy(optPtr->litLengthFreq, baseLLfreqs, sizeof(baseLLfreqs)); optPtr->litLengthSum = sum_u32(baseLLfreqs, MaxLL+1);
+                ZSTD_memcpy(optPtr->litLengthFreq, baseLLfreqs, sizeof(baseLLfreqs)); optPtr->litLengthSum = sum_u32(baseLLfreqs, MaxLL+1);
             }
 
             {   unsigned ml;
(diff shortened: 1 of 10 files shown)


programs/benchzstd.c
```diff
@@ -431,10 +429,9 @@ BMK_benchMemAdvancedNoAlloc(
         dctxprep.dictBuffer = dictBuffer;
         dctxprep.dictBufferSize = dictBufferSize;
 
-        hr_isize = UTIL_makeHumanReadableSize((U64) srcSize);
-
         DISPLAYLEVEL(2, "\r%70s\r", "");   /* blank line */
-        DISPLAYLEVEL(2, "%2s-%-17.17s : %.*f%s -> \r", marks[markNb], displayName, hr_isize.precision, hr_isize.value, hr_isize.suffix);
+        assert(srcSize < UINT_MAX);
+        DISPLAYLEVEL(2, "%2s-%-17.17s :%10u -> \r", marks[markNb], displayName, (unsigned)srcSize);
```
felixhandte: With the existing code, you can see full sizes with `-vv`. Does that seem insufficient? In general I like that this is human-readable formatted.

Cyan4973: Ah yes, I noticed this capability.
It wasn't enough unfortunately, as it was breaking scripts depending on previous format.
Moreover, and more importantly, I simply believe that this is not the right place for the condensed quantity format. Benchmarking is about accuracy. I typically always need the exact amount, in order to clearly appreciate changes, even when they are small.

The new "human readable" format is more suitable for general information status.

senhuang42: Yeah, I will say more often than not I have found myself benchmarking something and then having to re-do it once i realize I forgot `-vv`. At least for me, I'd always prefer to have it on.

felixhandte: Makes sense.

Cyan4973:
The `make benchmarking` [test error](https://github.com/facebook/zstd/runs/3550044688#step:3:298) is a bit obscure to me. Is it a case of incorrect parsing ?

> `IndexError: list index out of range`

terrelln:
> Is it a case of incorrect parsing ?

It seems so, probably a change to the `benchzstd.c` output format that is breaking `automated_benchmarking.py`.

lib/compress/zstd_opt.c
```diff
@@ -174,36 +194,44 @@ ZSTD_rescaleFreqs(optState_t* const optPtr,
...
 
         if (compressedLiterals)
-            optPtr->litSum = ZSTD_downscaleStat(optPtr->litFreq, MaxLit, 1);
-        optPtr->litLengthSum = ZSTD_downscaleStat(optPtr->litLengthFreq, MaxLL, 0);
-        optPtr->matchLengthSum = ZSTD_downscaleStat(optPtr->matchLengthFreq, MaxML, 0);
-        optPtr->offCodeSum = ZSTD_downscaleStat(optPtr->offCodeFreq, MaxOff, 0);
+            optPtr->litSum = ZSTD_scaleStats(optPtr->litFreq, MaxLit, 12);
+        optPtr->litLengthSum = ZSTD_scaleStats(optPtr->litLengthFreq, MaxLL, 11);
+        optPtr->matchLengthSum = ZSTD_scaleStats(optPtr->matchLengthFreq, MaxML, 11);
+        optPtr->offCodeSum = ZSTD_scaleStats(optPtr->offCodeFreq, MaxOff, 11);
```
terrelln, line 234: Should the `logTarget` maybe be based on the block size?

I would expect smaller blocks to want to have a smaller history.

terrelln:
I wonder if `btultra2` would do better if you ran `btlazy2` as the first parser instead of the optimal parser. I know we've talked about it, but I don't remember if you've tried it. That would help solve the bad literals pricing, and would maybe help favor shorter literal/match lengths. Although I know that btlazy2 does have more literals than other parsers, due to its laziness.

Cyan4973:
> I wonder if btultra2 would do better if you ran btlazy2 as the first parser instead of the optimal parser. 

It's a good question. It's probably worth an investigation.

But note that it's not clear if it's obviously better.
I hear the argument that `btlazy2` doesn't depend on prior statistics, and therefore will not be pushed one way or another depending on wrong initial statistics.
But then, it introduces other issues. At the very least, I would expect statistics from `btlazy2` to be somewhat off compared to those produced by a first `btultra` pass. For example, `btlazy2` doesn't look for (and never selects) match length 3. It also only considers `rep0`, like all `lazy` parsers. And as you already mentioned, it tends to produce more literals. So obviously all these cases will be incorrectly represented in final statistics. 
It's probably possible to attempt some kind of "blind fix" to partially compensate these bias, but it won't solve all cases.  For example, in #2765, all matches are length 3. Therefore, `btlazy2` would find no match at all. Difficult to extrapolate initial statistics from there.

Extending the topic, I've been considering some form of "light" initial evaluation for `btultra`, like for example `greedy`. This might help, compared to "default" initial statistics, at the cost of new dependency. `btultra2` would build up from there.

As can be guessed, the most important point here is that these investigations cost time. And time is the scarcest resource there is.

This PR doesn't "terminate" this topic, it mainly brings it to light. I'm opened to any complementary PR that would improve these initial results.

lib/compress/zstd_opt.c
Cyan4973: The update policy could be refined even further. But note that the topic of "block size" has many sub-cases, that probably deserve separate policies.

For example, is the "small block" also the "only block" ?
(I presume that's what you meant.)
In which case, the first pass of `btultra2` would produce stats which are already scaled to the size of the block. If there are only a few sequences, then the stats will only contain a few elements, way below the `logTarget` threshold. In which case, they won't be up-scaled (stats are only _down-scaled_ when they need to). So there is a form of built-in adaptation for small inputs in this process.

But if the "small block" is the N-th block in a long stream, and is also expected to be followed by other blocks of variable size, the situation is different. I would guess that a "stream level" `logTarget` would feel more appropriate.

This can certainly be analyzed even more.
I believe the update policy is one of these places where some non-negligible gains can still be produced.
And my expectation is that a better update policy should be dynamic, reacting to the "probability" that historic statistics may or may not match the statistics of the following block.

As can be guessed though, this is a fairly complex topic, which would require a dedicated (and time-consuming) study.

Cyan4973:
> It seems so, probably a change to the benchzstd.c output format that is breaking automated_benchmarking.py.

I guess that's because `benchzstd` now outputs its regular results into `stdout` instead of `stderr` (which is still used to output error messages). The python script seems to be designed to intercept and parse `stderr`.

A trivial replacement of `stderr` by `stdout` doesn't do the trick though ...

_edit_ : indeed, `output` is empty, so the attempt at parsing the floating point value fails. Still, I don't get it why it's still empty after replacing `stderr` by `stdout`.

_edit 2_ : get it, that's because the script is downloading and comparing several versions of `zstd`; therefore, some are outputting into `stdout`, and some into `stderr`, making neither of these choices sufficient. 
Onto a fix.

Cyan4973:
All remaining issues fixed

terrelln (approved):

Taefive55 (commented):
#2815

lib/compress/zstd_opt.c
```diff
@@ -174,36 +194,44 @@ ZSTD_rescaleFreqs(optState_t* const optPtr,
...
 
-            {   unsigned of;
-                for (of=0; of<=MaxOff; of++)
-                    optPtr->offCodeFreq[of] = 1;
+            {   unsigned const baseOFCfreqs[MaxOff+1] = {
+                    6, 2, 1, 1, 2, 3, 4, 4,
+                    4, 3, 2, 1, 1, 1, 1, 1,
+                    1, 1, 1, 1, 1, 1, 1, 1,
+                    1, 1, 1, 1, 1, 1, 1, 1
+                };
```
terrelln, line 221: This is the change that makes the majority of the difference

PR #2779, opened 2021-09-15
Fix NCountWriteBound

senhuang42:
OSS-Fuzz discovered an error in FSE_writeNCount().

Problem distribution:
```
(tableLog == 5, maxSymbolValue == 10)
[1,0,1,0,1,0,1,0,1,9,18]
```

Write: 6, 6, 5, 5, 5, 5, 5, 5, 5, 5, 1 bits
Total: 53 bits

But, main loop prior to flush currently writes 8 bytes rather than 7. Why?
BitCount initialized at 4, and first two symbols require an additional bit each, so we actually need 4 + 2 + 53 = 59 bits.

Currently calculated bound = 55 bits ((10+1)*5). So we need to add 4 + 2 to this bound.

### DIFF
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -221,7 +221,11 @@ size_t FSE_buildCTable_wksp(FSE_CTable* ct,
 ****************************************************************/
 size_t FSE_NCountWriteBound(unsigned maxSymbolValue, unsigned tableLog)
 {
-    size_t const maxHeaderSize = (((maxSymbolValue+1) * tableLog) >> 3) + 3;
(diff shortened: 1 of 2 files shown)


Cyan4973:
Fix looks good !

Does that seem possible to add one test case that triggers this situation (and fails without this PR) ?

terrelln (commented):
Looks good once a test case is added!

senhuang42:
Test case added (fails when fuzzer.c is built with ASAN)

Cyan4973 (approved):

terrelln (approved):

PR #2780, opened 2021-09-15
Reduce stack usage of block splitter

senhuang42:
This PR moves some of the block splitter stuff that was using too much stack space

I see no speed delta for levels 16-19 (have block splitter enabled by default) on `enwik7` with `gcc`.

```
contrib/linux-kernel test
old: Maximum stack size: 2760
new: Maximum stack size: 1848
```

Todo: Fix fuzzer errors

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1506,7 +1506,7 @@ size_t ZSTD_estimateCCtxSize_usingCCtxParams(const ZSTD_CCtx_params* params)
     ZSTD_compressionParameters const cParams =
                 ZSTD_getCParamsFromCCtxParams(params, ZSTD_CONTENTSIZE_UNKNOWN, 0, ZSTD_cpm_noAttachDict);
     ZSTD_paramSwitch_e const useRowMatchFinder = ZSTD_resolveRowMatchFinderMode(params->useRowMatchFinder,
-                                                                                         &cParams);
+                                                                               &cParams);
 
     RETURN_ERROR_IF(params->nbWorkers > 0, GENERIC, "Estimate CCtx size is supported for single-threaded compression only.");
     /* estimateCCtxSize is for one-shot compression. So no buffers should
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1959,6 +1960,12 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
         zc->outBuffSize = buffOutSize;
         zc->outBuff = (char*)ZSTD_cwksp_reserve_buffer(ws, buffOutSize);
 
+        /* block splitter ctx */
+        if (ZSTD_CParams_useBlockSplitter(&params->cParams)) {
+            /* Silence -Wcast-align with cast to void* */
+            zc->blockSplitCtx = (ZSTD_blockSplitCtx*)(void*)ZSTD_cwksp_reserve_buffer(ws, sizeof(ZSTD_blockSplitCtx));
```
terrelln: `-Wcast-align` has a point here. You'll need correct alignment here or certain platforms will fail.

This probably needs to go in the `ZSTD_cwksp_reserve_object()` space, or if that doesn't work just put it directly into the CCtx.

senhuang42:
Rebased on top of #2788. For review, just consider the last commit in the stack.

lib/compress/zstd_compress.c
senhuang42: Yeah you're right. I think this will just rest directly in the CCtx. If I want to save space when we're not using block splitting, a wksp object doesn't seem to work properly with the different cctx reuse cases, since objects are assumed to remain the same throughout compressions.

If I wanted to make it work somehow, it would have to be the same as the other objects in that they stay the same throughout different cctx re-uses. And if we have to do that anyways, we may as well just keep it in the cctx. Though, this does mean we will just have to eat the >1K additional heap cost.

terrelln: > Though, this does mean we will just have to eat the >1K additional heap cost.

That should be fine, thats much less than the total CCtx allocation. E.g. we have 500KB space for sequences, which is way overkill in most cases. So there is lower hanging fruit.

terrelln (approved):
Looks good to me!

Before merging, you should rebase on top of `dev`, so you don't get any extra commits in the diff.

Cyan4973 (approved):

PR #2783, opened 2021-09-17
[build] Add support for ASM files in Make + CMake

terrelln:
* Extract out common portion of `lib/Makefile` into `lib/libzstd.mk`.
  Most relevantly, the way we find library files.
* Use `lib/libzstd.mk` in the other Makefiles instead of repeating the
  same code.
* Add a test `tests/test-variants.sh` that checks that the builds of
  `make -C programs allVariants` are correct, and run it in Actions.
* Adds support for ASM files in the CMake build.

The Meson build is not updated because it lists every file in zstd,
and supports ASM off the bat, so the Huffman ASM commit will just add
the ASM file to the list.

The Visual Studios build is not updated because I'm not adding ASM
support to Visual Studios yet.


Cyan4973 (approved):

terrelln:
Looks like some of these changes are triggering a compiler bug in an aarch64 cross compiler :(

terrelln:
I've attempted to move the various cross compile builds to the QEMU matrix in GitHub actions, and am hoping they have a different version of the compiler that doesn't have the bug.

Cyan4973 (approved):

terrelln:
I'm going to disable the aarch64 test for now. I've opened Issue #2785 to track it.

PR #2790, opened 2021-09-21
[contrib][linux] Fix build after introducing ASM HUF implementation

solbjorn:
This mostly fixes kernel build after introducing ASM HUF implementation, but also adds the ability to build the kernel with recently landed `CONFIG_WERROR` while in-kernel ZSTD library still uses the functions deprecated in 1.5.0.

Compile and run time tested on Linux 5.15-rc2 with Clang 12.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ dictionary
 NUL
 
 # Build artefacts
+contrib/linux-kernel/linux/
 projects/
 bin/
 .buckd/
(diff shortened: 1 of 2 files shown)


terrelln:
Thanks for the PR! The `linux.mk` changes look good. The macro changes seem right, but I have to think them through to be sure. Looks like we're missing a test case for `!DYNAMIC_BMI2` with BMI2 enabled, I'll add that.

terrelln:
@solbjorn I'm going to fix the macro issues in PR #2791, where I also add a test case that covers them.

Once that is merged (later today or tomorrow), can you rebase on top of the latest dev branch, and just include the config changes?

solbjorn:
@terrelln, sure, great!

solbjorn:
Updated.

terrelln (approved):

terrelln:
Thanks for the PR @solbjorn!

PR #2793, opened 2021-09-22
Minor documentation update

Cyan4973:
notably to emphasize usage of `-r` to circumvent shell expansion limitation.

`ronn-ng` was used to rebuild the `man` pages,
resulting in minor visual differences (most notably the bullet in bulleted-list).


programs/zstdless.1
```diff
@@ -1,14 +0,0 @@
-.
```
terrelln, line 1: Why is the contents of this file deleted?

Cyan4973: Good point, it shouldn't.
_edit_ : fixed.

terrelln (approved):

PR #2796, opened 2021-09-23
[lib] Make lib compatible with `-Wfall-through` excepting legacy

terrelln:
Switch to a macro `ZSTD_FALLTHROUGH;` instead of a comment. On supported
compilers this uses an attribute, otherwise it becomes a comment.

This is necessary to be compatible with clang's `-Wfall-through`, and
gcc's `-Wfall-through=2` which don't support comments. Without this the
linux build emits a bunch of warnings.

Also add a test to CI to ensure that we don't regress.


Cyan4973:
This feature was also recently added to `xxhash`,
and I remember it wasn't completely trivial to ensure it would work correctly with a broad range of compilers.
The macro initialization ended up being a bit complex : 
https://github.com/Cyan4973/xxHash/blob/dev/xxhash.h#L598
That being said, it works now. Might be useful for inspiration.

terrelln:
It looks like [this section](https://github.com/Cyan4973/xxHash/blob/df5d63917e0ea553065e9b39429ded3a520b3de9/xxhash.h#L615-L627) can be simplified with `__has_attribute(__fallthrough__)` as I'm doing in zstd. Clang always supported it, and gcc started supporting `__has_attribute()` before it supported `__fallthrough__`.

We can definitely add on the section that uses the C/C++ language attribute `[[fallthrough]]` though.

Cyan4973:
Apparently, the macro determination code in `xxhash` uses `__has_c_attribute()`, instead of `__has_attribute()`.
Maybe because it's looking for `fallthrough`, instead of `__fallthrough__`.

terrelln:
> Apparently, the macro determination code in xxhash uses __has_c_attribute(), instead of __has_attribute().

`__has_c_attribute()` is to detect support for `[[fallthrough]]`, where `__has_attribute()` detects support for `__attribute__((__fallthrough__))`.

lib/common/compiler.h
```diff
@@ -224,6 +224,14 @@
 #  define __has_feature(x) 0
 #endif
 
+#ifndef ZSTD_FALLTHROUGH
+# if __has_attribute(__fallthrough__)
+#  define ZSTD_FALLTHROUGH __attribute__((__fallthrough__))
+# else
+#  define ZSTD_FALLTHROUGH /* fall-through */
```
Cyan4973: I presume this will _not_ add a comment where this macro is invoked ?

Cyan4973 (approved):

PR #2800, opened 2021-09-24
Fix a C89 error in msvc

ghost:
This causes msvc2012 to fail to compile 64 bit build.

### DIFF
--- a/lib/common/entropy_common.c
+++ b/lib/common/entropy_common.c
@@ -43,8 +43,14 @@ static U32 FSE_ctz(U32 val)
     assert(val != 0);
     {
 #   if defined(_MSC_VER)   /* Visual */
-        unsigned long r;
-        return _BitScanForward(&r, val) ? (unsigned)r : 0;
+        if (val != 0) {
+            unsigned long r;
+            _BitScanForward(&r, val);
+            return (unsigned)r;
+        } else {
+            /* Should not reach this code path */
+            __assume(0);
+        }
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* GCC Intrinsic */
         return __builtin_ctz(val);
 #   elif defined(__ICCARM__)    /* IAR Intrinsic */
--- a/lib/common/entropy_common.c
+++ b/lib/common/entropy_common.c
@@ -43,7 +43,7 @@ static U32 FSE_ctz(U32 val)
     assert(val != 0);
     {
 #   if defined(_MSC_VER)   /* Visual */
-        unsigned long r=0;
+        unsigned long r;
         return _BitScanForward(&r, val) ? (unsigned)r : 0;
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* GCC Intrinsic */
         return __builtin_ctz(val);
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -358,7 +358,7 @@ MEM_STATIC U32 ZSTD_highbit32(U32 val)   /* compress, dictBuilder, decodeCorpus 
 #       if STATIC_BMI2 == 1
             return _lzcnt_u32(val)^31;
 #       else
-            unsigned long r=0;
+            unsigned long r;
             return _BitScanReverse(&r, val) ? (unsigned)r : 0;
 #       endif
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* GCC Intrinsic */
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -391,7 +391,7 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 #           if STATIC_BMI2
                 return _tzcnt_u64(val);
 #           else
-                unsigned long r = 0;
+                unsigned long r;
                 return _BitScanForward64( &r, (U64)val ) ? (unsigned)(r >> 3) : 0;
 #           endif
 #       elif defined(__GNUC__) && (__GNUC__ >= 4)
@@ -409,7 +409,7 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 #       endif
     } else { /* 32 bits */
 #       if defined(_MSC_VER)
-            unsigned long r=0;
+            unsigned long r;
             return _BitScanForward( &r, (U32)val ) ? (unsigned)(r >> 3) : 0;
 #       elif defined(__GNUC__) && (__GNUC__ >= 3)
             return (__builtin_ctz((U32)val) >> 3);
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -358,8 +358,14 @@ MEM_STATIC U32 ZSTD_highbit32(U32 val)   /* compress, dictBuilder, decodeCorpus 
 #       if STATIC_BMI2 == 1
             return _lzcnt_u32(val)^31;
 #       else
-            unsigned long r;
-            return _BitScanReverse(&r, val) ? (unsigned)r : 0;
+            if (val != 0) {
+                unsigned long r;
+                _BitScanReverse(&r, val);
+                return (unsigned)r;
+            } else {
+                /* Should not reach this code path */
+                __assume(0);
+            }
 #       endif
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* GCC Intrinsic */
         return __builtin_clz (val) ^ 31;
@@ -391,8 +397,14 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 #           if STATIC_BMI2
                 return _tzcnt_u64(val);
 #           else
-                unsigned long r;
-                return _BitScanForward64( &r, (U64)val ) ? (unsigned)(r >> 3) : 0;
+                if (val != 0) {
+                    unsigned long r;
+                    _BitScanForward64(&r, (U64)val);
+                    return (unsigned)(r >> 3);
+                } else {
+                    /* Should not reach this code path */
+                    __assume(0);
+                }
 #           endif
 #       elif defined(__GNUC__) && (__GNUC__ >= 4)
             return __builtin_ctzll((U64)val);
@@ -409,8 +421,14 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 #       endif
     } else { /* 32 bits */
 #       if defined(_MSC_VER)
-            unsigned long r;
-            return _BitScanForward( &r, (U32)val ) ? (unsigned)(r >> 3) : 0;
+            if (val != 0) {
+                unsigned long r;
+                _BitScanForward(&r, (U32)val);
+                return (unsigned)(r >> 3);
+            } else {
+                /* Should not reach this code path */
+                __assume(0);
+            }
 #       elif defined(__GNUC__) && (__GNUC__ >= 3)
             return (__builtin_ctz((U32)val) >> 3);
 #       else
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -385,10 +385,10 @@ MEM_STATIC U32 ZSTD_highbit32(U32 val)   /* compress, dictBuilder, decodeCorpus 
 }
 
 /**
- * Computes CTZ on a U64.
- * This will be slow on 32-bit mode, and on unsupported compilers.
- * If you need this function to be fast (because it is hot) expand
- * support.
+ * Counts the number of trailing zeros of a `size_t`.
+ * Most compilers should support CTZ as a builtin. A backup
+ * implementation is provided if the builtin isn't supported, but
+ * it may not be terribly efficient.
  */
 MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 {
@@ -400,7 +400,7 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
                 if (val != 0) {
                     unsigned long r;
                     _BitScanForward64(&r, (U64)val);
-                    return (unsigned)(r >> 3);
+                    return (unsigned)r;
                 } else {
                     /* Should not reach this code path */
                     __assume(0);
@@ -424,13 +424,13 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
             if (val != 0) {
                 unsigned long r;
                 _BitScanForward(&r, (U32)val);
-                return (unsigned)(r >> 3);
+                return (unsigned)r;
             } else {
                 /* Should not reach this code path */
                 __assume(0);
             }
 #       elif defined(__GNUC__) && (__GNUC__ >= 3)
-            return (__builtin_ctz((U32)val) >> 3);
+            return __builtin_ctz((U32)val);
 #       else
             static const int DeBruijnBytePos[32] = {  0,  1, 28,  2, 29, 14, 24,  3,
                                                      30, 22, 20, 15, 25, 17,  4,  8,
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -143,10 +143,16 @@ MEM_STATIC unsigned BIT_highbit32 (U32 val)
     {
 #   if defined(_MSC_VER)   /* Visual */
 #       if STATIC_BMI2 == 1
-		return _lzcnt_u32(val) ^ 31;
+            return _lzcnt_u32(val) ^ 31;
 #       else
-		unsigned long r;
-		return _BitScanReverse(&r, val) ? (unsigned)r : 0;
+            if (val != 0) {
+                unsigned long r;
+                _BitScanReverse(&r, val);
+                return (unsigned)r;
+            } else {
+                /* Should not reach this code path */
+                __assume(0);
+            }
 #       endif
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* Use GCC Intrinsic */
         return __builtin_clz (val) ^ 31;
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -145,7 +145,7 @@ MEM_STATIC unsigned BIT_highbit32 (U32 val)
 #       if STATIC_BMI2 == 1
 		return _lzcnt_u32(val) ^ 31;
 #       else
-		unsigned long r = 0;
+		unsigned long r;
 		return _BitScanReverse(&r, val) ? (unsigned)r : 0;
 #       endif
 #   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* Use GCC Intrinsic */
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -672,7 +672,7 @@ static unsigned ZSTD_NbCommonBytes (size_t val)
 #           if STATIC_BMI2
                 return _tzcnt_u64(val) >> 3;
 #           else
-                unsigned long r = 0;
+                unsigned long r;
                 return _BitScanForward64( &r, (U64)val ) ? (unsigned)(r >> 3) : 0;
 #           endif
 #       elif defined(__GNUC__) && (__GNUC__ >= 4)
@@ -690,7 +690,7 @@ static unsigned ZSTD_NbCommonBytes (size_t val)
 #       endif
         } else { /* 32 bits */
 #       if defined(_MSC_VER)
-            unsigned long r=0;
+            unsigned long r;
             return _BitScanForward( &r, (U32)val ) ? (unsigned)(r >> 3) : 0;
 #       elif defined(__GNUC__) && (__GNUC__ >= 3)
             return (__builtin_ctz((U32)val) >> 3);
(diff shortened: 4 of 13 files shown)


ghost:
According to this article:
https://docs.microsoft.com/en-us/cpp/intrinsics/bitscanforward-bitscanforward64

`_BitScanForward()` is available in 32-bit build.
Can it be used for 32-bit build? 

(in msvc, `long`/`unsigned long` are 4-byte integers)

Cyan4973:
I think `_BitScanForward64()` is, or used to be, a pb for 32-bit Visual builds, as it could not be linked.

I'm not sure to understand what is being fixed here.

ghost:
This PR fixes a C89 error, variables (`r` in this site) must be declared at the beginning.
msvc2012 only support C89, so it can't be compiled.

A possible improvement is that 32-bit build can use `_BitScanForward()`. 
Moreover, searching `_BitScanForward64()` on GitHub, it seems that no need to initialize `r` to 0.

ghost:
If do such improvements, this PR can be closed, fix the C89 problem together with the improvements.

Cyan4973:
> it seems that no need to initialize r to 0.

There is no need to.
But, if I do remember correctly, a static analyzer complained about that, and that's why it was initialized.
I think that's because `_BitScanForward64()` can fail if `val == 0`.
But since then, the code has been improved to detect this case and `return 0` when that happens,
so there should be no need to initialize `r` anymore.

Not that I would expect such change to make any difference, as I'm pretty sure good compilers can properly deduct the behavior of `r`.

ghost:
> I'm pretty sure good compilers can properly deduct the behavior of r.

In msvc2019 release build, `r = 0` is not optimized away in some sites.
Worse, it is written to a stack variable that will not be used.

```asm
...
; 1279 :         for (; (matches > 0) && (nbAttempts > 0); --nbAttempts, matches &= (matches - 1)) {

	mov	r8, QWORD PTR row$1$[rsp]
	test	r9, r9
	je	$LN1950@ZSTD_RowFi
	mov	r14d, DWORD PTR dictMode$1$[rsp]
	mov	esi, DWORD PTR rowMask$1$[rsp]
	mov	ebx, DWORD PTR dictLimit$1$[rsp]
	mov	rdx, QWORD PTR base$1$[rsp]
	mov	r12d, DWORD PTR lowLimit$1$[rsp]
$LL13@ZSTD_RowFi:
	test	r13d, r13d
	je	SHORT $LN2055@ZSTD_RowFi

; 885  :     return _BitScanForward64(&r, val) ? (U32)r : 0; /* _BitScanForward64 not defined outside of x86/64 */

	bsf	rax, r9
	mov	ecx, ebp
	mov	DWORD PTR r$85[rsp], ebp    ; r = 0 (ebp is 0)
	cmovne	ecx, eax

; 1280 :             U32 const matchPos = (head + ZSTD_VecMask_next(matches)) & rowMask;

	add	ecx, r10d
	and	ecx, esi

...
```
But in other sites, `r = 0` is optimized away:
```asm
...
; 1346 :             for (; (matches > 0) && (nbAttempts > 0); --nbAttempts, matches &= (matches - 1)) {

	test	r9, r9
	je	$LN2058@ZSTD_RowFi
	mov	esi, DWORD PTR rowMask$1$[rsp]
	mov	rdx, QWORD PTR dmsRow$1$[rsp]
$LL19@ZSTD_RowFi:
	test	r13d, r13d
	je	SHORT $LN1955@ZSTD_RowFi

; 885  :     return _BitScanForward64(&r, val) ? (U32)r : 0; /* _BitScanForward64 not defined outside of x86/64 */

	bsf	rax, r9
	mov	ecx, ebp
	cmovne	ecx, eax

; 1347 :                 U32 const matchPos = (head + ZSTD_VecMask_next(matches)) & rowMask;

	add	ecx, r11d
	and	ecx, esi
...
```

Cyan4973:
OK, let's remove that useless initialization then.

ghost:
I'll investigate more, hope it can be done in a day or two.

ghost:
If remove `r = 0` for `_BitScanReverse*` and `_BitScanForward*` functions, dozens of stack variable assignments have been eliminated.
There are a few performance improvements on i5-4570, see below table.

It seems the files in `decompress` folder don't have this problem, but the decompression speed is also improved, don't know why. I have tried several times, The speeds are like the table.

There is no change (C/D speeds) on r5-3600x.

If enable [`/sdl`](https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks?view=msvc-160) option (check uninitialized local variable), no error is reported. Maybe other static analyzers can also handle these functions correctly. 

```
                  c_speed      d_speed         c_speed      d_speed
 1#silesia.tar  (325.0 MB/s,  774.8 MB/s) -> (332.1 MB/s,  782.4 MB/s)
 2#silesia.tar  (257.8 MB/s,  699.2 MB/s) -> (261.3 MB/s,  705.5 MB/s)
 3#silesia.tar  (184.7 MB/s,  682.2 MB/s) -> (185.8 MB/s,  686.8 MB/s)
 4#silesia.tar  (170.4 MB/s,  667.5 MB/s) -> (172.8 MB/s,  671.4 MB/s)
 5#silesia.tar   (93.1 MB/s,  685.3 MB/s) ->  (95.3 MB/s,  692.3 MB/s)
 6#silesia.tar   (64.7 MB/s,  693.6 MB/s) ->  (65.4 MB/s,  696.2 MB/s)
 7#silesia.tar   (56.3 MB/s,  725.3 MB/s) ->  (56.9 MB/s,  727.9 MB/s)
 8#silesia.tar   (45.7 MB/s,  734.5 MB/s) ->  (46.9 MB/s,  739.6 MB/s)
 9#silesia.tar   (41.0 MB/s,  742.1 MB/s) ->  (40.9 MB/s,  748.1 MB/s)
10#silesia.tar   (35.4 MB/s,  733.2 MB/s) ->  (35.6 MB/s,  736.6 MB/s)
11#silesia.tar   (27.3 MB/s,  740.2 MB/s) ->  (27.8 MB/s,  743.6 MB/s)
12#silesia.tar   (19.1 MB/s,  745.4 MB/s) ->  (19.4 MB/s,  754.6 MB/s)
13#silesia.tar  (10.26 MB/s,  754.2 MB/s) -> (10.28 MB/s,  757.2 MB/s)
14#silesia.tar   (8.29 MB/s,  760.7 MB/s) ->  (8.36 MB/s,  765.8 MB/s)
15#silesia.tar   (6.48 MB/s,  753.4 MB/s) ->  (6.50 MB/s,  757.2 MB/s)
16#silesia.tar   (4.93 MB/s,  726.9 MB/s) ->  (4.89 MB/s,  733.5 MB/s)
17#silesia.tar   (3.93 MB/s,  704.6 MB/s) ->  (3.91 MB/s,  707.5 MB/s)
```

ghost:
If the tested operand is `0`, the `bsf`/`bsf` instructions may return an undefined value **in theory**:
https://stackoverflow.com/a/43386587/15717325
https://stackoverflow.com/a/54509783/15717325
> **Do note that they (`bsf`/`bsf`) have an output dependency on the destination register, which they leave unmodified if the input was zero**. AMD documents this behaviour, Intel implements it in hardware but documents it as an "undefined" result, so unfortunately compilers won't take advantage of it and human programmers maybe should be cautious.


`gcc` doc:

> Built-in Function: `int __builtin_clz (unsigned int x)`
> Returns the number of leading 0-bits in x, starting at the most significant bit position. **If x is 0, the result is undefined.**
>
> Built-in Function: `int __builtin_ctz (unsigned int x)`
> Returns the number of trailing 0-bits in x, starting at the least significant bit position. **If x is 0, the result is undefined.**
(https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html)

Zstd doesn't check `0` when using these `gcc` functions, so the `gcc` build of zstd relies on an undefined behavior.
But unlike `msvc`, the code generated by `gcc` looks like this, both operands are the same, then `0` is fine. 
```asm
	movq	-8(%rbp), %rax
	bsrl	%eax, %eax
	xorl	$31, %eax
	sarl	$3, %eax
```

Cyan4973:
I believe the "bad" case (`val == 0`) simply cannot happen, because we never call this function in this case.
But that's something that some static analyzer may miss, even with the presence of a corresponding `assert(val != 0)`, that they simply ignore on the ground that it's not enabled in "release" mode. That's what leads to this branch `_BitScanReverse(&r, val) ? (unsigned)r : 0;`, which is correct, but overkill, and may lead to performance degradation since it's a hot function (although I would expect a clever enough compiler to be able to use Constant Propagation to determine that one of these branches is useless and avoid the performance penalty). 
Anyway, avoiding the overly cautious variable initialization already brings a little bit of performance, and that's good news.

Cyan4973 (approved):

ghost:
`msvc` has a [`__assume(0)`](https://docs.microsoft.com/en-us/cpp/intrinsics/assume) statement, used to tell the compiler unreachable code path. For example:
```c
    if (val != 0) {
        unsigned long r;
        BitScanReverse(&r, val);
        return (unsigned)r;
    } else {
        __assume(0);
    }
```
Then the `if (val != 0)` check will be optimized away.
`gcc` has a similar function [`__builtin_unreachable()`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html)

Such an `UNREACHABLE()` macro may be used to make assertion in debug build, and optimize the check in release build:
```c
#ifdef NDEBUG
#   if defined(_MSC_VER)
#      define UNREACHABLE()   __assume(0)
#   elif defined(__GNUC__)
#      define UNREACHABLE()   __builtin_unreachable()
#else
#   define UNREACHABLE()      assert(0)
#endif

    if (val != 0) {
        unsigned long r;
        _BitScanReverse(&r, val);
        return (unsigned)r;
    } else {
        UNREACHABLE();
    }
```
This is a code style of solving such problems.

I see it optimizes 3 instructions to 1 `bsr` instruction in `msvc`. I will benchmark later.

ghost:
This time, turn off the anti-virus software, and change the power management to performance mode.

```
i5-4570, silesia.tar
           dev branch            remove r=0 (this PR)    further remove condition check   WSL2(gcc-9.3.0)
 1#   324.0 MB/s, 775.4 MB/s    332.0 MB/s, 782.5 MB/s    332.5 MB/s, 783.1 MB/s        376.1 MB/s, 1246.7 MB/s
 2#   258.1 MB/s, 699.8 MB/s    262.1 MB/s, 706.1 MB/s    263.0 MB/s, 707.2 MB/s        287.6 MB/s, 1095.3 MB/s
 3#   184.7 MB/s, 681.3 MB/s    185.9 MB/s, 688.1 MB/s    187.2 MB/s, 688.3 MB/s        204.7 MB/s, 1048.9 MB/s
 4#   170.7 MB/s, 663.8 MB/s    173.0 MB/s, 675.2 MB/s    174.5 MB/s, 672.5 MB/s        188.3 MB/s, 1015.6 MB/s
 5#    93.3 MB/s, 683.9 MB/s     94.8 MB/s, 688.9 MB/s     96.3 MB/s, 691.9 MB/s        110.4 MB/s, 1039.2 MB/s
 6#    64.7 MB/s, 692.8 MB/s     65.5 MB/s, 698.7 MB/s     66.0 MB/s, 695.6 MB/s         73.7 MB/s, 1043.8 MB/s
 7#    56.5 MB/s, 723.3 MB/s     56.7 MB/s, 728.1 MB/s     57.2 MB/s, 725.7 MB/s         64.0 MB/s, 1098.1 MB/s
 8#    45.6 MB/s, 732.5 MB/s     46.8 MB/s, 738.1 MB/s     47.2 MB/s, 741.9 MB/s         52.4 MB/s, 1116.4 MB/s
 9#    40.7 MB/s, 739.5 MB/s     41.3 MB/s, 747.3 MB/s     42.2 MB/s, 751.1 MB/s         45.4 MB/s, 1124.7 MB/s
10#    35.4 MB/s, 729.7 MB/s     35.6 MB/s, 737.5 MB/s     36.3 MB/s, 739.7 MB/s         38.7 MB/s, 1092.6 MB/s
11#    27.4 MB/s, 740.0 MB/s     27.7 MB/s, 745.3 MB/s     28.1 MB/s, 748.0 MB/s         29.9 MB/s, 1107.1 MB/s
12#    19.1 MB/s, 744.8 MB/s     19.5 MB/s, 754.5 MB/s     19.2 MB/s, 753.2 MB/s         20.8 MB/s, 1119.8 MB/s
13#   10.25 MB/s, 751.8 MB/s    10.27 MB/s, 759.0 MB/s    10.32 MB/s, 760.0 MB/s        10.33 MB/s, 1130.2 MB/s
14#    8.35 MB/s, 759.5 MB/s     8.31 MB/s, 759.3 MB/s     8.40 MB/s, 765.1 MB/s         8.37 MB/s, 1139.6 MB/s
15#    6.48 MB/s, 756.4 MB/s     6.49 MB/s, 759.5 MB/s     6.51 MB/s, 759.0 MB/s         6.47 MB/s, 1124.0 MB/s
16#    4.82 MB/s, 725.8 MB/s     4.88 MB/s, 730.7 MB/s     4.94 MB/s, 734.4 MB/s         5.08 MB/s, 1078.4 MB/s
17#    3.86 MB/s, 700.1 MB/s     3.91 MB/s, 707.7 MB/s     3.95 MB/s, 709.7 MB/s         4.05 MB/s, 1016.4 MB/s

r5-3600x, silesia.tar
           dev branch             remove r=0 (this PR)     further remove condition check
 1#   474.1 MB/s, 1055.7 MB/s    470.7 MB/s, 1055.3 MB/s    472.3 MB/s, 1060.7 MB/s
 2#   403.4 MB/s, 970.6 MB/s     397.1 MB/s, 967.7 MB/s     400.2 MB/s, 971.8 MB/s
 3#   290.1 MB/s, 948.4 MB/s     290.5 MB/s, 945.0 MB/s     289.6 MB/s, 951.3 MB/s
 4#   265.6 MB/s, 933.0 MB/s     265.1 MB/s, 928.0 MB/s     264.9 MB/s, 939.5 MB/s
 5#   128.0 MB/s, 949.0 MB/s     130.8 MB/s, 945.3 MB/s     131.2 MB/s, 954.9 MB/s
 6#   111.3 MB/s, 960.6 MB/s     114.2 MB/s, 955.3 MB/s     114.9 MB/s, 970.2 MB/s
 7#    81.4 MB/s, 994.2 MB/s      85.2 MB/s, 988.9 MB/s      85.8 MB/s, 1002.7 MB/s
 8#    69.2 MB/s, 1020.3 MB/s     71.7 MB/s, 1016.3 MB/s     72.4 MB/s, 1032.0 MB/s
 9#    63.5 MB/s, 1019.3 MB/s     61.8 MB/s, 1014.1 MB/s     65.1 MB/s, 1029.1 MB/s
10#    58.3 MB/s, 1032.7 MB/s     56.8 MB/s, 1026.4 MB/s     59.5 MB/s, 1043.3 MB/s
11#    44.6 MB/s, 1048.2 MB/s     46.2 MB/s, 1030.9 MB/s     47.4 MB/s, 1050.6 MB/s
12#    29.6 MB/s, 1062.7 MB/s     30.5 MB/s, 1057.9 MB/s     30.1 MB/s, 1076.9 MB/s
13#    16.2 MB/s, 1067.6 MB/s     16.2 MB/s, 1062.7 MB/s     16.2 MB/s, 1081.0 MB/s
14#    13.1 MB/s, 1082.0 MB/s     13.3 MB/s, 1076.3 MB/s     13.2 MB/s, 1096.1 MB/s
15#    9.67 MB/s, 1080.7 MB/s     9.82 MB/s, 1074.7 MB/s     9.75 MB/s, 1096.6 MB/s
16#    7.23 MB/s, 1017.2 MB/s     7.21 MB/s, 1001.7 MB/s     7.26 MB/s, 1020.2 MB/s
17#    5.57 MB/s, 986.2 MB/s      5.58 MB/s, 982.4 MB/s      5.61 MB/s, 1000.4 MB/s
```

Cyan4973:
OK, so it seems very minor, but still consistently positive for Visual.
I think it's worth adopting.

ghost:
I'm going to sleep, let's do it tomorrow.
Just updated the benchmark for i5-4570+gcc9.3 in the last column of the table, the latest `msvc2019` is slower than `gcc9.3`.

Cyan4973:
> the latest msvc2019 is slower than gcc9.3

Yes, we are aware of the vast performance difference between these compilers.
Truth be told, since we don't use Visual in our daily environment, a part of this performance difference can be attributed directly to this lack of focus.

ghost:
The above commit can optimize away the condition checks.
I didn't modify the files in `legacy` folder, so review burden can be reduced a little.

**edit**: The above commit didn't fix #2807, it just left the original behavior.

Cyan4973 (approved):

ghost:
I will not force-push approved commits, so please review commit after `approved` message.

**edit:** `#error` is a C89 directive.
http://port70.net/~nsz/c/c89/c89-draft.html#3.8.5

lib/common/zstd_internal.h
```diff
@@ -391,34 +395,38 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
...
+            }
 #       elif defined(__GNUC__) && (__GNUC__ >= 3)
-            return (__builtin_ctz((U32)val) >> 3);
+            return __builtin_ctz((U32)val);
 #       else
-            static const int DeBruijnBytePos[32] = {  0,  1, 28,  2, 29, 14, 24,  3,
-                                                     30, 22, 20, 15, 25, 17,  4,  8,
-                                                     31, 27, 13, 23, 21, 19, 16,  7,
-                                                     26, 12, 18,  6, 11,  5, 10,  9 };
-            return DeBruijnBytePos[((U32)((val & -(S32)val) * 0x077CB531U)) >> 27];
```
terrelln: Please leave this code in. This function is actually very cold, only called 4x per block.

lib/common/entropy_common.c
```diff
@@ -43,8 +43,14 @@ static U32 FSE_ctz(U32 val)
     assert(val != 0);
     {
 #   if defined(_MSC_VER)   /* Visual */
-        unsigned long r=0;
-        return _BitScanForward(&r, val) ? (unsigned)r : 0;
+        if (val != 0) {
```
terrelln, line 46: We've decided to no longer enforce separation between FSE & zstd. So this can just call `ZSTD_countTrailingZeros()` now.

So you can just `#include "zstd_internal.h"` and use that function.

lib/common/zstd_internal.h
```diff
@@ -391,34 +395,38 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
...
+                if (val != 0) {
+                    unsigned long r;
+                    _BitScanForward64(&r, (U64)val);
+                    return (unsigned)r;
+                } else {
+                    /* Should not reach this code path */
+                    __assume(0);
+                }
 #           endif
+#       elif defined(_MSC_VER) && !defined(_WIN64)
```
terrelln: In this case we're already in 64-bit mode, `MEM_64bits()`. We can definitely use the fallback implementation here, if for some reason `!defined(_WIN64)`.

lib/common/zstd_internal.h
```diff
@@ -379,10 +385,8 @@ MEM_STATIC U32 ZSTD_highbit32(U32 val)   /* compress, dictBuilder, decodeCorpus
 }
 
 /**
- * Computes CTZ on a U64.
- * This will be slow on 32-bit mode, and on unsupported compilers.
- * If you need this function to be fast (because it is hot) expand
- * support.
```
terrelln, line 385: I see that this comment is a bit confusing. Its maybe a bit out of date, because I changed it to a `size_t` instead of a `U64`, so the slowness on 32-bit mode should be resolved. Maybe change it to:

```c
/**
 * Counts the number of trailing zeros of a `size_t`.
 * Most compilers should support CTZ as a builtin. A backup
 * implementation is provided if the builtin isn't supported, but
 * it may not be terribly efficient.
 */
```

terrelln (changes requested):
Please fix the comments, then this should be good to merge.

lib/common/entropy_common.c
ghost: `FSE_ctz(U32 val)` is for `U32` type:
```c
U32 bitStream;
...
int repeats = FSE_ctz(~bitStream | 0x80000000) >> 1;
```
While `ZSTD_countTrailingZeros(size_t val)` is for `size_t` type, there is an unnecessary cast in 64-bit build.

So the force-pushed last commit only fixes `>> 3` bug.

terrelln (approved):

lib/common/zstd_internal.h
```diff
@@ -391,8 +397,14 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 #           if STATIC_BMI2
                 return _tzcnt_u64(val);
 #           else
-                unsigned long r = 0;
-                return _BitScanForward64( &r, (U64)val ) ? (unsigned)(r >> 3) : 0;
+                if (val != 0) {
+                    unsigned long r;
+                    _BitScanForward64(&r, (U64)val);
```
terrelln, line 402: This is the bug fix.

lib/common/zstd_internal.h
```diff
@@ -409,10 +421,16 @@ MEM_STATIC unsigned ZSTD_countTrailingZeros(size_t val)
 #       endif
     } else { /* 32 bits */
 #       if defined(_MSC_VER)
-            unsigned long r=0;
-            return _BitScanForward( &r, (U32)val ) ? (unsigned)(r >> 3) : 0;
+            if (val != 0) {
+                unsigned long r;
+                _BitScanForward(&r, (U32)val);
```
terrelln, line 426: And here.

PR #2801, opened 2021-09-24
[NOCOMMIT] Revert "Huffman ASM"

terrelln:
This reverts commit a5f2c45528032ed20c33e0f8cd2c163a800a0017.

See if Appveyor speeds up.


Cyan4973:
Appveyor CI ticks in at 34 min 25 sec,
which is no better than other recent test sessions.

The increase from ~22mn to ~34mn is likely related to something else then.

PR #2802, opened 2021-09-24
[contrib][linux] Fix -Wundef inside Linux kernel tree

solbjorn:
Commit d7ef97a013b5
("[build] Fix oss-fuzz build with the dataflow sanitizer") broke
build inside Linux-kernel after 'import', as it no longer can
conditionally remove ZSTD_MEMORY_SANITIZER definition from
the #if DEF_A || DEF_B block. This emits -Wundef warning which
can be treated as error.
Split this preprocessor condition into two separate conditions
to fix this.

### DIFF
--- a/lib/decompress/huf_decompress.c
+++ b/lib/decompress/huf_decompress.c
@@ -47,11 +47,13 @@
  * Disable when MSAN is enabled.
  */
 #if defined(__linux__) || defined(__linux) || defined(__APPLE__)
-# if ZSTD_MEMORY_SANITIZER || ZSTD_DATAFLOW_SANITIZER
+# if ZSTD_MEMORY_SANITIZER
+#  define HUF_ASM_SUPPORTED 0
+# elif ZSTD_DATAFLOW_SANITIZER
 #  define HUF_ASM_SUPPORTED 0
 # else
 #  define HUF_ASM_SUPPORTED 1
-#endif
+# endif
 #else
 # define HUF_ASM_SUPPORTED 0
 #endif


solbjorn:
Got 15 minutes late after my previous PR was merged (wanted to append this commit), meh.

lib/decompress/huf_decompress.c
```diff
@@ -47,11 +47,15 @@
...
 #  define HUF_ASM_SUPPORTED 0
-# else
+# endif
+# if ZSTD_DATAFLOW_SANITIZER
+#  define HUF_ASM_SUPPORTED 0
+# endif
+# ifndef HUF_ASM_SUPPORTED
 #  define HUF_ASM_SUPPORTED 1
-#endif
+# endif
```
terrelln:
```c
# if ZSTD_MEMORY_SATITIZER
#  define HUF_ASM_SUPPORTED 0
# elif ZSTD_DATAFLOW_SANITIZER
#. define HUF_ASM_SUPPORTED 0
# else
#. define HUF_ASM_SUPPORTED 1
# endif
```

terrelln (commented):
Looks good once you change to `elif`. I will also follow up by adding `-Wundef -Werror` and other flags to our CI for the linux build.

Just curious though, what are you using the linux kernel build for?

solbjorn:
@terrelln, I just import the latest dev snapshot every time I build the kernel for my home PC -- I use BtrFS transparent compression and want to both make sure I have the best performance and test the recent changes 😸

solbjorn:
Updated, #elif variant gets correctly unifdefed. Thanks!

Cyan4973 (approved):

terrelln (approved):

PR #2805, opened 2021-09-26
Smaller code with disabled features

nolange:
Hello,

Since zstd is everywhere nowadays, it would be time to get it into projects that pragmatically aim at code-size.
The aim of this series if to allow the developer a relatively easy way of dropping unused code, and the compiler to
propagate constants, partially specialize functions and remove a bigger junk of then dead code.

Since I expect some discussions covering all of those changes I put them into one pull request.


Cyan4973 (approved):

Cyan4973:
This first batch of modifications looks relatively straightforward to me.

felixhandte (approved):
Looks good to me.

PR #2809, opened 2021-09-28
Limit train samples

### LINKED ISSUE #2745: zstd train-cover chokes if sample size is "too large"
**Describe the bug**

Training on a big sample set, first zstd attempts to load everything into memory, then it notices memory is not enough and only loads part of it, and then fails anyway because the sample size is over 4GB.

```
> zstd --train-cover split -r --maxdict 262144 -o split-zstd-dict/dictionary
Not enough memory; training on 14562 MB only...
Trying 82 different sets of parameters
Total samples size is too large (14562 MB), maximum size is 4095 MB
Failed to initialize context
dictionary training failed : Src size is incorrect
```

**To Reproduce**

- Sample size: ~1.5m files, 32KB/file
- Run the command in the listing above

**Expected behavior**

It would make much more sense to at least just load 4GB worth of samples, print a warning, ignore the rest and train on that.

(Ideally, training could proceed incrementally and not require loading all samples in memory at the same time. This would obviously also make the problem described here go away)

 - OS: Mac (x86-64)
 - Version zstd 1.5.0 (via homebrew)

stanjo74:
- Limit loaded samples to 2GB.
- Fixes https://github.com/facebook/zstd/issues/2745 
- Support for --block-size option which enables chunking - the sample files are broken up into chunks and each chunk is a training sample.
- A sample is limited to 128KB.
- Rewrote the sample loading logic to be more expressive.

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -40,6 +40,11 @@
 /*-*************************************
 *  Constants
 ***************************************/
+/**
+* There are 32bit indexes used to ref samples, so limit samples size to 4GB
+* on 64bit builds.
+* For 32bit builds 1 GB is an arbitrary memory limit.
+*/
 #define COVER_MAX_SAMPLES_SIZE (sizeof(size_t) == 8 ? ((unsigned)-1) : ((unsigned)1 GB))
 #define COVER_DEFAULT_SPLITPOINT 1.0
 
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -43,7 +43,9 @@
 /**
 * There are 32bit indexes used to ref samples, so limit samples size to 4GB
 * on 64bit builds.
-* For 32bit builds 1 GB is an arbitrary memory limit.
+* For 32bit builds we choose 1 GB.
+* Most 32bit platforms have 2GB user-mode addressable space and we allocate a large
+* contiguous buffer, so 1GB is already a high limit.
 */
 #define COVER_MAX_SAMPLES_SIZE (sizeof(size_t) == 8 ? ((unsigned)-1) : ((unsigned)1 GB))
 #define COVER_DEFAULT_SPLITPOINT 1.0
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -145,15 +145,15 @@ static unsigned DiB_loadFiles(void* buffer, size_t* bufferSizePtr,
         DISPLAYUPDATE(2, "Loading %s...       \r", fileNamesTable[fileIndex]);
 
         /* Load the first chunk of data from the file */
-        size_t const chunkSize = targetChunkSize > 0 ?
+        size_t const headSize = targetChunkSize > 0 ?
                                     MIN(fileSize, targetChunkSize) :
                                     MIN(fileSize, SAMPLESIZE_MAX );
-        if (totalDataLoaded + chunkSize > *bufferSizePtr)
+        if (totalDataLoaded + headSize > *bufferSizePtr)
             break;
 
         size_t fileDataLoaded = fread(
-            buff+totalDataLoaded, 1, chunkSize, f );
-        if (fileDataLoaded != chunkSize)
+            buff+totalDataLoaded, 1, headSize, f );
+        if (fileDataLoaded != headSize)
             EXM_THROW(11, "Pb reading %s", fileNamesTable[fileIndex]);
         sampleSizes[nbSamplesLoaded++] = fileDataLoaded;
         totalDataLoaded += fileDataLoaded;
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -115,7 +115,7 @@ static int DiB_loadFiles(
     void* buffer, size_t* bufferSizePtr,
     size_t* sampleSizes, int sstSize,
     const char** fileNamesTable, int nbFiles,
-    int targetChunkSize, int displayLevel )
+    size_t targetChunkSize, int displayLevel )
 {
     char* const buff = (char*)buffer;
     size_t totalDataLoaded = 0;
@@ -126,7 +126,7 @@ static int DiB_loadFiles(
     assert(targetChunkSize <= SAMPLESIZE_MAX);
 
     while ( nbSamplesLoaded < sstSize && fileIndex < nbFiles ) {
-        S64 fileDataLoaded;
+        size_t fileDataLoaded;
         S64 const fileSize = DiB_getFileSize(fileNamesTable[fileIndex]);
         if (fileSize <= 0) /* skip if zero-size or file error */
             continue;
@@ -137,28 +137,24 @@ static int DiB_loadFiles(
         DISPLAYUPDATE(2, "Loading %s...       \r", fileNamesTable[fileIndex]);
 
         /* Load the first chunk of data from the file */
-        {
-            int const headSize = targetChunkSize > 0 ?
-                                        (int)MIN(fileSize, targetChunkSize) :
-                                        (int)MIN(fileSize, SAMPLESIZE_MAX );
-            if (totalDataLoaded + headSize > *bufferSizePtr)
-                break;
-
-            fileDataLoaded = fread( buff+totalDataLoaded, 1, headSize, f );
-            if (fileDataLoaded != headSize)
-                EXM_THROW(11, "Pb reading %s", fileNamesTable[fileIndex]);
-        }
-        sampleSizes[nbSamplesLoaded++] = (size_t)fileDataLoaded;
-        totalDataLoaded += (size_t)fileDataLoaded;
+        fileDataLoaded = targetChunkSize > 0 ?
+                            (size_t)MIN(fileSize, (S64)targetChunkSize) :
+                            (size_t)MIN(fileSize, SAMPLESIZE_MAX );
+        if (totalDataLoaded + fileDataLoaded > *bufferSizePtr)
+            break;
+        if (fread( buff+totalDataLoaded, 1, fileDataLoaded, f ) != fileDataLoaded)
+            EXM_THROW(11, "Pb reading %s", fileNamesTable[fileIndex]);
+        sampleSizes[nbSamplesLoaded++] = fileDataLoaded;
+        totalDataLoaded += fileDataLoaded;
 
         /* If file-chunking is enabled, load the rest of the file as more samples */
         if (targetChunkSize > 0) {
             while( fileDataLoaded < fileSize && nbSamplesLoaded < sstSize ) {
-                int const chunkSize = (int)MIN(fileSize-fileDataLoaded, targetChunkSize);
+                size_t const chunkSize = MIN(fileSize-fileDataLoaded, targetChunkSize);
                 if (totalDataLoaded + chunkSize > *bufferSizePtr) /* buffer is full */
                     break;
 
-                if (fread( buff+totalDataLoaded, 1, chunkSize, f ) != (unsigned)chunkSize)
+                if (fread( buff+totalDataLoaded, 1, chunkSize, f ) != chunkSize)
                     EXM_THROW(11, "Pb reading %s", fileNamesTable[fileIndex]);
                 sampleSizes[nbSamplesLoaded++] = chunkSize;
                 totalDataLoaded += chunkSize;
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -49,6 +49,7 @@
 static const size_t g_maxMemory = (sizeof(size_t) == 4) ? (2 GB - 64 MB) : ((size_t)(512 MB) << sizeof(size_t));
 
 #define NOISELENGTH 32
+#define MAX_SAMPLES_SIZE (2 GB) /* training dataset limited to 2GB */
 
 
 /*-*************************************
@@ -272,6 +273,10 @@ int DiB_trainFromFiles(const char* dictFileName, unsigned maxDictSize,
                            FASTCOVER_MEMMULT;
     size_t const maxMem =  DiB_findMaxMem(fs.totalSizeToLoad * memMult) / memMult;
     size_t loadedSize = (size_t) MIN ((unsigned long long)maxMem, fs.totalSizeToLoad);
+
+    /* Limit the total size of the training samples to 2GB */
+    if (loadedSize > MAX_SAMPLES_SIZE)
+        loadedSize = MAX_SAMPLES_SIZE;
     void* const srcBuffer = malloc(loadedSize+NOISELENGTH);
     int result = 0;
 
@@ -296,7 +301,9 @@ int DiB_trainFromFiles(const char* dictFileName, unsigned maxDictSize,
 
     /* init */
     if (loadedSize < fs.totalSizeToLoad)
-        DISPLAYLEVEL(1, "Not enough memory; training on %u MB only...\n", (unsigned)(loadedSize >> 20));
+        DISPLAYLEVEL(1, "Trainig samples set too large (%u MB); training on %u MB only...\n",
+            (unsigned)(fs.totalSizeToLoad >> 20),
+            (unsigned)(loadedSize >> 20));
 
     /* Load input buffer */
     DISPLAYLEVEL(3, "Shuffling input files\n");
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -96,7 +96,7 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 static S64 DiB_getFileSize (const char * fileName)
 {
     U64 const fileSize = UTIL_getFileSize(fileName);
-    return (fileSize == UTIL_FILESIZE_UNKNOWN) ? -1 : fileSize;
+    return (fileSize == UTIL_FILESIZE_UNKNOWN) ? -1 : (S64)fileSize;
 }
 
 /* ********************************************************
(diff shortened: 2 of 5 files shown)


stanjo74:
The first commit d758afe limits the samples size to 2GB.
The rest of the commits are optional rewrite of the sample-loading logic and are optional.

lib/dictBuilder/cover.c
```diff
@@ -40,6 +40,11 @@
 /*-*************************************
 *  Constants
 ***************************************/
+/**
+* There are 32bit indexes used to ref samples, so limit samples size to 4GB
+* on 64bit builds.
+* For 32bit builds 1 GB is an arbitrary memory limit.
```
Cyan4973: That's a good thing to add some comments to clarify the intent of the code ! 👍 

A word of comment on the "1 GB limit for 32-bit" : 
it's worth knowing that we typically use a few "large" allocations, which require continuous memory segments.
In 32-bit address space, this quickly becomes a problem, depending on the amount of fragmentation before reaching the allocation point. 
Moreover, some 32-bit systems (Windows) limit user space to 2 GB. Even Linux will typically reserve the end of the address space for kernel, leaving just the rest to user space.
So we need some margin.

programs/dibio.c
```diff
@@ -88,6 +92,20 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 #undef MIN
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 
+/**
+  Returns the size of a file.
+  If error returns 0. Zero filesize or error is same for us.
+  Emit warning when the file is inaccessible or zero size.
+*/
+static size_t Dib_getFileSize (const char * fileName)
```
Cyan4973: nit: `DiB_` prefix

programs/dibio.c
```diff
@@ -49,21 +49,25 @@
...
 
 #define NOISELENGTH 32
+#define MAX_SAMPLES_SIZE (2 GB) /* training dataset limited to 2GB */
 
 
 /*-*************************************
 *  Console display
 ***************************************/
+#define DISPLAY_LEVEL_DEFAULT 2
+static int g_displayLevel = DISPLAY_LEVEL_DEFAULT;
```
Cyan4973: I kind of remember that I tried to limit usage of global variables in this file.
The annoying part is that functions which display messages on console need to receive variables like `displayLevel`, and propagation becomes a load on their parameter list.

As a beneficial side-effect, it's less easy now to "forget" synchronizing `displayLevel` between `zstd` modules, which all tend to have their own `g_displayLevel` that needs to be updated. We can also quickly notice which functions need to display something and which ones do not.

Not that it matters a lot, but reducing the amount of global variables felt somewhat cleaner, and now that the work is done, why reverting back to usage of global variables ?

programs/dibio.c
```diff
@@ -102,46 +120,74 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
-                pos += readSize;
-                sampleSizes[nbLoadedChunks++] = toLoad;
-                remainingToLoad -= targetChunkSize;
-                if (nbLoadedChunks == sstSize) { /* no more space left in sampleSizes table */
-                    fileIndex = nbFiles;  /* stop there */
+    char * buff = (char*)buffer;
+    size_t totalDataLoaded = 0;
+    int nbSamplesLoaded = 0;
+    int fileIndex = 0;
+    FILE * f = NULL;
```
Cyan4973, line 124: Why would you need `f` outside of the loop ?

Edit : I see, the new code can `break` from the loop.

programs/dibio.c
```diff
@@ -102,46 +120,74 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+        if (fileSize == 0)
+            continue;
+
+        f = fopen( fileNamesTable[fileIndex], "rb");
+        if (f == NULL)
+            EXM_THROW(10, "zstd: dictBuilder: %s %s ", fileNamesTable[fileIndex], strerror(errno));
+        DISPLAYUPDATE(2, "Loading %s...       \r", fileNamesTable[fileIndex]);
+
+        /* Load the first chunk of data from the file */
+        size_t const headSize = targetChunkSize > 0 ?
```
Cyan4973: `C90` requires all declarations to be done at the beginning of a block.

Note that it's allowed and even recommended to create sub-blocks just for the purpose to declaring _and terminating_ new variables. The _terminating_ part is actually the more important : the less _leaky_ is a variable lifetime, the better.

programs/dibio.c
```diff
@@ -88,6 +92,20 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 
+/**
+  Returns the size of a file.
+  If error returns 0. Zero filesize or error is same for us.
+  Emit warning when the file is inaccessible or zero size.
+*/
+static size_t Dib_getFileSize (const char * fileName)
+{
+    size_t fileSize = UTIL_getFileSize(fileName);
```
Cyan4973: minor : this code base is extremely `const` friendly. I would even call it `const` fanatic.

Any mutable variable is immediately frown upon, and should ask the question : could I do it differently ? Sometimes, there's really no other way, but in many circumstances, it's just a matter of thinking the sequence differently.

For this case, I would suggest: `if (fileSize == UTIL_FILESIZE_UNKNOWN) return 0;`. This remove the need to make `fileSize` mutable.

programs/dibio.c
stanjo74: Yes, there are tradeoffs. I sense you want to do away with the global variable and pass displayLevel as parameter. I can make that change.

programs/dibio.c
stanjo74: Yes, because the loop breaks. We can do away with `break` but the loop condition becomes a lot more cryptic.

programs/dibio.c
stanjo74: Yes, I'm taking the scoping approach throughout the code.

programs/dibio.c
stanjo74: Agree. Was thinking about changing it like you suggested and forgot.

lib/dictBuilder/cover.c
stanjo74: Will add more to the comment, thank you for the explanation.

programs/dibio.c
```diff
@@ -102,46 +115,74 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
-            if (toLoad > *bufferSizePtr-pos) break;
-            {   size_t const readSize = fread(buff+pos, 1, toLoad, f);
-                if (readSize != toLoad) EXM_THROW(11, "Pb reading %s", fileName);
-                pos += readSize;
-                sampleSizes[nbLoadedChunks++] = toLoad;
-                remainingToLoad -= targetChunkSize;
-                if (nbLoadedChunks == sstSize) { /* no more space left in sampleSizes table */
-                    fileIndex = nbFiles;  /* stop there */
-                    break;
+    char * buff = (char*)buffer;
```
terrelln: nit: Why remove the `const`? I don't see it modified anywhere.

style nit: We don't put a space before the `*` of pointers, so `char* const`.

programs/dibio.c
```diff
@@ -88,6 +89,18 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
 #undef MIN
 #define MIN(a,b)    ((a) < (b) ? (a) : (b))
 
+/**
+  Returns the size of a file.
+  If error returns 0. Zero filesize or error is same for us.
+  Emit warning when the file is inaccessible or zero size.
```
terrelln: nit: This comment is out of date, or I'm missing something, I don't see any warnings emitted when this function returns 0.

programs/dibio.c
```diff
@@ -102,46 +115,74 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+        if (fileSize == 0)
+            continue;
+
+        f = fopen( fileNamesTable[fileIndex], "rb");
+        if (f == NULL)
+            EXM_THROW(10, "zstd: dictBuilder: %s %s ", fileNamesTable[fileIndex], strerror(errno));
+        DISPLAYUPDATE(2, "Loading %s...       \r", fileNamesTable[fileIndex]);
+
+        /* Load the first chunk of data from the file */
+        size_t fileDataLoaded = 0;
```
terrelln: This will have to be moved to the top of the loop, or start a new scope.

programs/dibio.c
```diff
@@ -102,46 +115,74 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+            if (fileDataLoaded != headSize)
+                EXM_THROW(11, "Pb reading %s", fileNamesTable[fileIndex]);
+        }
+        sampleSizes[nbSamplesLoaded++] = fileDataLoaded;
+        totalDataLoaded += fileDataLoaded;
+
+        /* If file-chunking is enabled, load the rest of the file as more samples */
+        if (targetChunkSize > 0) {
+            while( fileDataLoaded < fileSize && nbSamplesLoaded < sstSize ) {
+                size_t chunkDataLoaded = 0;
```
terrelln: nit: Don't initialize this. e.g. `size_t chunkDataLoaded;`. That way the compiler will warn you if any branches fail to initialize the variable.

Also, why do you need this variable? If you just don't create a new scope, you can use `chunkSize` directly. You'll just need to check the result of `fread()` with a separate variable.

programs/dibio.c
```diff
@@ -102,46 +115,74 @@ static UTIL_time_t g_displayClock = UTIL_TIME_INITIALIZER;
...
+                {
+                    size_t const chunkSize = MIN(fileSize-fileDataLoaded, targetChunkSize);
+                    if (chunkSize == 0) /* no more to read */
+                        break;
```
terrelln: I don't think this condition is possible, right?

programs/dibio.c
```diff
@@ -235,22 +275,46 @@ typedef struct {
...
+      if (chunkSize > 0)
+      {
+        // TODO: is there a minimum sample size? Can we have a 1-byte sample?
+        fs.nbSamples += ((fileSize + chunkSize-1) / chunkSize) * chunkSize;
+        fs.totalSizeToLoad += fileSize;
+      }
+      else {
+      /* the case where one file is one sample */
+        if (fileSize > SAMPLESIZE_MAX) {
+          /* flag excessively large smaple files */
```
terrelln: nit: spelling `sample`.

programs/dibio.c
```diff
@@ -260,18 +324,31 @@ int DiB_trainFromFiles(const char* dictFileName, unsigned maxDictSize,
...
-                        coverParams ? coverParams->zParams.notificationLevel :
-                        fastCoverParams ? fastCoverParams->zParams.notificationLevel :
-                        0;   /* should never happen */
+    fileStats fs;
     void* const dictBuffer = malloc(maxDictSize);
-    fileStats const fs = DiB_fileStats(fileNamesTable, nbFiles, chunkSize, displayLevel);
+
+    int const displayLevel = params ? params->zParams.notificationLevel :
+        coverParams ? coverParams->zParams.notificationLevel :
+        fastCoverParams ? fastCoverParams->zParams.notificationLevel : 0;
```
terrelln, line 324: Ugh this is ugly. Nothing to really do about it now, but I've never been happy with the way we pass parameters to the dictionary builder... It would be nice to find the time to improve it, but it has never been a priority.

programs/dibio.c
```diff
@@ -260,18 +324,31 @@ int DiB_trainFromFiles(const char* dictFileName, unsigned maxDictSize,
...
+
+    /* Shuffle input files before we start assessing how much sample date to load.
+       The purpose of the shuffle is to pick random samples when the sample
+       set is larger than what we can load in memory. */
```
terrelln: nit: That isn't the only purpose. We also shuffle to improve training, because there are some biases that can be introduced when samples show up in a "sorted" order, that are mitigated by shuffling.

programs/dibio.c
```diff
@@ -260,18 +324,31 @@ int DiB_trainFromFiles(const char* dictFileName, unsigned maxDictSize,
...
+    DISPLAYLEVEL(3, "Shuffling input files\n");
+    DiB_shuffle(fileNamesTable, nbFiles);
+
+    /* Figure out how much sample data to load with how many samples */
+    fs = DiB_fileStats(fileNamesTable, nbFiles, chunkSize, displayLevel);
+
     size_t* const sampleSizes = (size_t*)malloc(fs.nbSamples * sizeof(size_t));
     size_t const memMult = params ? MEMMULT :
```
terrelln: You have declaration after statement here.

terrelln (commented):
Overall, I love the simplification to file loading, just some style nits.

programs/dibio.c
stanjo74: correct, it's already enforced by `while( fileDataLoaded < fileSize`

programs/dibio.c
stanjo74: Why is `notificationLevel ` a property of the training algorithm? Should it be a property of the process?

stanjo74: Should we refactor all modules to use a shared g_displayLevel and routines? I started going in this direction, but @Cyan4973 made an argument that passing the display level around as function parameter makes it clear who needs to print. We should discuss and unify the approach as both arguments have merit.

programs/dibio.c
stanjo74: This is interesting, I'd like to know more about these biases. I'm not sure what to add to the comment as this point.

stanjo74:
One observation is `size_t* sampleSizes` array - we don't need 64bits to track a sample size that is limited to 128KB. I think we should change it to `int` and save some memory.
Which raises the philosophical question of `signed/unsigned int` to store cardinal values. My personal preference is `int` - change my mind.

programs/dibio.c
```diff
@@ -223,58 +256,97 @@ static void DiB_saveDict(const char* dictFileName,
...
+          /* Limit to the first SAMPLESIZE_MAX (128kB) of the file */
+          DISPLAYLEVEL(3, "Sample file '%s' is too large, limiting to %ukB",
+              fileNamesTable[n], SAMPLESIZE_MAX >> 10);
+        }
+        fs.nbSamples += 1;
+        fs.totalSizeToLoad += MIN(fileSize, SAMPLESIZE_MAX);
+      }
     }
-    DISPLAYLEVEL(4, "Preparing to load : %u KB \n", (unsigned)(fs.totalSizeToLoad >> 10));
+    DISPLAYLEVEL(4, "Training files are %lldKB, %d samples\n", fs.totalSizeToLoad >> 10, fs.nbSamples);
```
stanjo74: Looks like MinGW doesn't like %lld. How do we go about this for cross-platform?

stanjo74:
Tested 32KB chunking with 40K files of 112KB and got this output.
The number of samples reported along the chain is suspicious - will track.

```
Shuffling input files
Training files are 4458176KB, **160002 samples**
Training samples set too large (4353 MB); training on 2048 MB only...
loaded 2097127KB total data, **75265 nb samples**
Trying 82 different sets of parameters
d=6
Training on **120001 samples** of total size 2147458979
```

programs/dibio.c
terrelln: > Should we refactor all modules to use a shared g_displayLevel and routines?

This cannot be done for the code in `lib/`, which the dictBuilder resides in. The code in `lib/` must be thread-safe, so no global variables. Although it is a bit strange that the dictBuilder prints to output at all, none of our other library code does that.

I don't think that it makes sense to change the `notificationLevel` parameter on its own. A larger refactor is needed, and is beyond the scope of this PR. I think that the dictBuilder should ultimately not print to `stderr` at all, and should communicate its progress / status through other means, like a callback & better error codes.

programs/dibio.c
terrelln: Maybe: "The purpose of the shuffle is to avoid bias in the trainer due to sorted files, and to pick random samples ..."

I'd be happy to explain more how the dictionary builder works, and how sorting can introduce bias, but is probably too verbose to put into a comment. It would be worth a meeting + writing up a doc, so others can read it too.

programs/dibio.c
terrelln: Just cast it to an unsigned and us `%u`. If somehow it doesn't fit, then we don't really care if the print is wrong.

terrelln:
> One observation is size_t* sampleSizes array - we don't need 64bits to track a sample size that is limited to 128KB. I think we should change it to int and save some memory.

The dictionary builder API in `lib/dictBuilder/` requires a `size_t const*`. The function `ZDICT_trainFromBuffer()` is stable and its API cannot be changed. It can be deprecated, and maybe eventually deleted a few years later, but never changed.

We could change the other functions, but then it would be inconsistent with `ZDICT_trainFromBuffer()`, and would break our users without a real gain. But the `128KB` limit is part of the `programs/` limitation, not part of the `lib/dictBuilder` limitation. The API can accept files of any size, although passing very large files doesn't really make sense.

> Which raises the philosophical question of signed/unsigned int to store cardinal values. My personal preference is int - change my mind.

I always use `size_t` or `int`. `unsigned` never really makes sense, as it just makes the job harder for the compiler. In cases where I know overflow isn't an issue, I tend to prefer `int` as well. Generally as loop variables.

But, zstd does deal with sizes > 2 GB in many cases, so we have to be careful to use `size_t` there. And we only use `size_t` for counts when we know it is referring to contiguous memory, like buffers. When we're dealing with larger numbers, generally the frame content size (how large the data decompresses into), we make sure to use `uint64_t`, so 32-bit mode works with sizes > 4GB.

stanjo74:
> Tested 32KB chunking with 40K files of 112KB and got this output. The number of samples reported along the chain is suspicious - will track.

Fixed the bug with 9eb56a3. The estimated number of samples was passed rather than the loaded number of samples.

programs/dibio.c
Cyan4973: > I think that the dictBuilder should ultimately not print to `stderr` at all, and should communicate its progress / status through other means, like a callback & better error codes.

I fully agree.
`dictBuilder` is part of `lib`, it should be refactored to not depend on `<stdio>` (in `release` mode).
And yes, a callback is likely more appropriate to communicate progression status.
This is indeed a much more complex topic, not to be addressed here.

Cyan4973 (approved):

PR #2813, opened 2021-10-04
Enhance streaming_compression examples.

marxin:
Add level argument to the first test and be more verbose about
used compression level and number of threads.

### DIFF
--- a/examples/streaming_compression.c
+++ b/examples/streaming_compression.c
@@ -15,9 +15,12 @@
 #include <zstd.h>      // presumes zstd library is installed
 #include "common.h"    // Helper functions, CHECK(), and CHECK_ZSTD()
 
-
-static void compressFile_orDie(const char* fname, const char* outName, int cLevel)
+static void compressFile_orDie(const char* fname, const char* outName, int cLevel,
+                               int nbThreads)
 {
+    fprintf (stderr, "Starting compression of %s with level %d, using %d threads\n",
+             fname, cLevel, nbThreads);
+
     /* Open the input and output files. */
     FILE* const fin  = fopen_orDie(fname, "rb");
     FILE* const fout = fopen_orDie(outName, "wb");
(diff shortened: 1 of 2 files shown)


examples/streaming_compression.c
```diff
@@ -15,9 +15,12 @@
 #include <zstd.h>      // presumes zstd library is installed
 #include "common.h"    // Helper functions, CHECK(), and CHECK_ZSTD()
 
+#define THREADS 4
```
senhuang42: I don't think a macro is necessary here. You can pass a `const int nbThreads = 4` from `main()` and add an `nbThreads` arg to `compressFile_orDie`. That way it's more consistent with the change to `cLevel` getting passed from main as well (and in fact, you could add `nbThreads` as an additional arg to the CLI too).

examples/streaming_compression.c
```diff
@@ -106,17 +109,22 @@ int main(int argc, const char** argv)
 {
     const char* const exeName = argv[0];
 
-    if (argc!=2) {
+    if (argc!=3) {
```
senhuang42: Nit: I'd prefer if we were able to keep the usage as `exeName FILE` with the optionality of specifying `exeName FILE LEVEL` and maybe even `exeName FILE LEVEL NUM_THREADS`, and default back to `cLevel = 1` and `nbThreads=4` or something.

examples/streaming_compression_thread_pool.c
```diff
@@ -56,7 +58,7 @@ static void *compressFile_orDie(void *data)
      */
     CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_compressionLevel, args->cLevel) );
     CHECK_ZSTD( ZSTD_CCtx_setParameter(cctx, ZSTD_c_checksumFlag, 1) );
-    ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, 16);
+    ZSTD_CCtx_setParameter(cctx, ZSTD_c_nbWorkers, THREADS);
```
senhuang42: Nit: ditto for the unnecessary macro - prefer using local `const` variable.

senhuang42 (changes requested):
Overall I like the improvement to the example! Just some minor suggestions.

marxin:
Thank you for the comments, I made the suggested changes. I like your suggestions.

marxin:
@senhuang42 May I please ping this?

senhuang42 (approved):
lgtm, thanks for the update!

senhuang42:
I see some tests are failing, can you update the base repo on top of which this branch sits to the latest dev branch and then force push onto this PR again?

marxin:
Sure, I've just done that.

Cyan4973 (approved):

PR #2820, opened 2021-10-08
[binary-tree] Fix underflow of nbCompares

terrelln:
Fix underflow of `nbCompares` by switching to a `for` loop, which is
clearer, and consistent with the hash-chain and row-based match finders.

Also, add asserts that underflow isn't happening, which were easily
triggered by the OSS-Fuzz fuzzers before the fix was added.

Clang performance is neutral to 2-3% worse, and gcc performance is
neutral to 3-5% positive. So overall it is about a wash.

The bug was reported by Dan Carpenter and found by Smatch static
checker.

https://lore.kernel.org/all/20211008063704.GA5370@kili/

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -93,7 +93,7 @@ ZSTD_insertDUBT1(const ZSTD_matchState_t* ms,
     assert(curr >= btLow);
     assert(ip < iend);   /* condition for ZSTD_count */
 
-    while (nbCompares-- && (matchIndex > windowLow)) {
+    for (; nbCompares && (matchIndex > windowLow); --nbCompares) {
         U32* const nextPtr = bt + 2*(matchIndex & btMask);
         size_t matchLength = MIN(commonLengthSmaller, commonLengthLarger);   /* guaranteed minimum nb of common bytes */
         assert(matchIndex < curr);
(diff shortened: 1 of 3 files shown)


terrelln:
@Cyan4973 is there a reason why the binary tree loops are doing `nbCompares + 1` iterations?

lib/compress/zstd_lazy.c
```diff
@@ -93,7 +93,7 @@ ZSTD_insertDUBT1(const ZSTD_matchState_t* ms,
     assert(curr >= btLow);
     assert(ip < iend);   /* condition for ZSTD_count */
 
-    while (nbCompares-- && (matchIndex > windowLow)) {
+    while ((nbCompares-- > 0) && (matchIndex > windowLow)) {
```
Cyan4973: In the original implementation, `while (nbCompares--)`, the loop will stop when `nbCompares == 0`.
Granted, `nbCompares` underflows right after that,
but it's not really a problem since the loop is exited, and the variable is never used afterwards.

Another potential fix could have been to use `for(; nbCompares ; nbCompares--)`,
which would exit the loop without triggering the underflowing `--` operation.

terrelln: Ah, I miscounted. I thought that `while (nbCompares--)` was doing `n+1` iterations, which is why I didn't use a for loop. I think I will switch to the for loop, since it is more clear.

Cyan4973:
> is there a reason why the binary tree loops are doing nbCompares + 1 iterations?

Well, it probably doesn't matter much, but no, I don't see any specific reason.
I also do not see where this happens.
Could you point at a specific place in the code where you think there are `nbCompares + 1` iterations ?

Cyan4973 (approved):

terrelln:
> Could you point at a specific place in the code where you think there are nbCompares + 1 iterations ?

Yeah, it doesn't happen. I made a mistake when counting the number of iterations.

I've switched to using a `for` loop instead, and keeping them as `U32`. This is consistent with the other places that use `nbCompares`, and is more clear.

I'm also seeing some non-negligible perf wins of ~3-5%.

PR #2826, opened 2021-10-21
Fix a warning in MSVC

ghost:
This warning appears when using `/arch:AVX2` option.

`zstd_compress_internal.h(673,44): warning C4244: “return”: from “unsigned __int64” to “unsigned int”，possible loss of data`

Tested with msvc 2015/2019/2022, 32/64 bit builds.


ghost:
For some reason, close this issue temporarily.

ghost:
Let's close it again, I'm thinking about a refactoring method, here may save an assembly instruction.

Change like this:
```diff
- static unsigned ZSTD_NbCommonBytes (size_t val)
+ static size_t   ZSTD_NbCommonBytes (size_t val)
```

PR #2829, opened 2021-10-25
minor : change build macro to ZSTD_DECODER_INTERNAL_BUFFER

Cyan4973:
and reworded the doc to make the topic more accessible for potential users (focus on the visible symptoms).

One _potential_ shortcoming is that I've not seen any minimum amount required for `ZSTD_LITBUFFEREXTRASIZE` to work correctly. Now that's a value potentially controlled by the user, should there be such a minimum ? cc @binhdvo .

### DIFF
--- a/lib/decompress/zstd_decompress_internal.h
+++ b/lib/decompress/zstd_decompress_internal.h
@@ -110,10 +110,11 @@ typedef struct {
 #  define ZSTD_DECODER_INTERNAL_BUFFER  (1 << 16)
 #endif
 
+#define ZSTD_LBMIN 64
 #define ZSTD_LBMAX (128 << 10)
 
-/* extra buffer, compensates amount of dst required to store litBuffer */
-#define ZSTD_LITBUFFEREXTRASIZE  MIN(ZSTD_DECODER_INTERNAL_BUFFER, ZSTD_LBMAX)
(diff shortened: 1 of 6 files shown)


terrelln (approved):

lib/decompress/zstd_decompress_internal.h
```diff
@@ -106,10 +106,15 @@ typedef struct {
     size_t ddictPtrCount;
 } ZSTD_DDictHashSet;
 
-#ifndef ZSTD_LITBUFFEREXTRASIZE
-#define ZSTD_LITBUFFEREXTRASIZE    (1 << 16) /* extra buffer reduces amount of dst required to store litBuffer */
+#ifndef ZSTD_DECODER_INTERNAL_BUFFER
+#  define ZSTD_DECODER_INTERNAL_BUFFER  (1 << 16)
 #endif
 
+#define ZSTD_LBMAX (128 << 10)
```
binhdvo, line 114: I can set a 64 byte minimum in a followup; is it sensible to add it here in the same manner?

binhdvo (approved):

lib/decompress/zstd_decompress_internal.h
Cyan4973: I'll do it in this PR.

PR #2831, opened 2021-10-26
Refactor tzcnt/lzcnt code

ghost:
Using GitHub's Split mode to review this PR is better.
This way there will be no wordy similar code everywhere.

* add `BIT_ctz32`/`BIT_ctz64`/`BIT_clz32`/`BIT_clz64` inline functions.
* change the type of `ZSTD_NbCommonBytes()` from `unsigned` to `size_t`. And move it to `common/zstd_internal.h`, so it can be shared by `compress` and `dictBuilder`.

`BIT_c[tl]z64` return `U64`, since `x86_64`/`aarch64`/`RISC-V64 (Bitmanip)` CPUs put the result in a 64-bit register.

The return type of gcc `__builtin_c[tl]zll(unsigned long long)` functions is `int`, there is an unnecessary type cast in `x86_64` build. Before this PR (gcc-11.1):
```asm
# lib/compress/zstd_compress_internal.h:774:             size_t const diff = MEM_readST(pMatch) ^ MEM_readST(pIn);
	xorq	%r8, %rcx	# _645, diff
# lib/compress/zstd_compress_internal.h:685:             return (__builtin_ctzll((U64)val) >> 3);
	rep bsfq	%rcx, %rcx	# diff, tmp668
# lib/compress/zstd_compress_internal.h:685:             return (__builtin_ctzll((U64)val) >> 3);
	sarl	$3, %ecx	#, tmp669
# lib/compress/zstd_compress_internal.h:776:             pIn += ZSTD_NbCommonBytes(diff);
	movslq	%ecx, %rcx	# tmp669, tmp670
# lib/compress/zstd_compress_internal.h:776:             pIn += ZSTD_NbCommonBytes(diff);
	addq	%rcx, %rdx	# tmp670, pIn
# lib/compress/zstd_compress_internal.h:777:             return (size_t)(pIn - pStart);
	subq	%rsi, %rdx	# _582, _678
```
After this PR, it removes the unnecessary cast, but there seems to be no performance change:
```asm 
# lib/compress/zstd_compress_internal.h:731:             size_t const diff = MEM_readST(pMatch) ^ MEM_readST(pIn);
	xorq	%r11, %r10	# _386, diff
# lib/compress/../common/bitstream.h:195:         asm ("rep bsfq %1, %0"
	rep bsfq %r10, %r10	# diff, ret
# lib/compress/zstd_compress_internal.h:672:                 return BIT_ctz64((U64)val) >> 3;
	shrq	$3, %r10	#, tmp418
# lib/compress/zstd_compress_internal.h:733:             pIn += ZSTD_NbCommonBytes(diff);
	addq	%r10, %rax	# tmp418, pIn
# lib/compress/zstd_compress_internal.h:734:             return (size_t)(pIn - pStart);
	subq	%rdi, %rax	# _60, _418
```

`aarch64`+`gcc` doesn't have this unnecessary cast, it may be a `x86_64` + `gcc` bug.
`aarch64` asm:
```asm
// lib/compress/zstd_compress_internal.h:678:             size_t const diff = MEM_readST(pMatch) ^ MEM_readST(pIn);
	eor	x0, x2, x10	// diff, _260, _262
// lib/compress/../common/bitstream.h:203:     return (U64) __builtin_ctzll(val);
	rbit	x0, x0	// _267, diff
	clz	x0, x0	// _267, _267
// lib/compress/zstd_compress_internal.h:680:             pIn += ZSTD_NbCommonBytes(diff);
	add	x0, x8, x0, lsr 3	// pIn, pIn, _267,
// lib/compress/zstd_compress_internal.h:681:             return (size_t)(pIn - pStart);
	sub	x0, x0, x4	// _249, pIn, _44
```


lib/compress/zstd_lazy.c
```diff
@@ -880,38 +880,20 @@ typedef U64 ZSTD_VecMask;   /* Clarifies when we are interacting with a U64 repr
...
+            return BIT_ctz32(leastSignificantWord);
         }
     } else {
-        return (U32)__builtin_ctzll(val);
+        return (U32) BIT_ctz64(val);
     }
-#   else
-    /* Software ctz version: http://aggregate.org/MAGIC/#Trailing%20Zero%20Count
-     * and: https://stackoverflow.com/questions/2709430/count-number-of-bits-in-a-64-bit-long-big-integer
-     */
```
ghost, line 909: If let `BIT_ctz32`/`BIT_ctz64` use software implementation, compare to this software ctz code:

- 32-bit 10.22x faster
- 64-bit 2.10x faster

Tested on gcc-11.1, so I remove this ctz code.

Cyan4973:
Trying to catch up on this topic.
What's the expected outcome of this PR ?
Cleaner assembly generation ?

ghost:
> Trying to catch up on this topic. What's the expected outcome of this PR ? Cleaner assembly generation ?

It creates an abstract layer, avoid similar code everywhere, convenient for development.

For example, `ZSTD_VecMask_next()` doesn't have 32-bit MSVC code path currently, then it will fall to very slow software implementation.
https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_lazy.c#L771-L803

After checking "IAR C/C++ Development Guide",  `__CTZ()` doesn't exist, only `__CLZ()` is available. This code In [entropy_common.c](https://github.com/facebook/zstd/blob/dev/lib/common/entropy_common.c#L56-L57), such mistake can be avoided to some degree.
```c
#   elif defined(__ICCARM__)    /* IAR Intrinsic */
        return __CTZ(val);
```

There is a doubt, `BIT_c[tl]z64` return `U32` or `U64`, but the asm code seems to be the same. 
* return `U64` (This PR): https://godbolt.org/z/Yf9W8538Y
* return `U32` (no U32->U64 cast for `__builtin_c[tl]zll`): https://godbolt.org/z/rjz3ocrG1

terrelln:
I'm going to put up a PR with some refactoring. But, I'm going to hold it for after the release, since it is a relatively big change.

ghost:
IMO this PR is easy to review.
The software implementations were copied from the current code, just comparing whether they have changed is fine.

I used to write a small program to verify these software implementations, ensure they have the same result as the corresponding `__builtin_c[tl]z(ll)?` functions.

I only wrote `BIT_clz64()`'s  software implementation, but it's not used in practice, and the logic is very simple.

By the way, I also changed the software implemented's table type from 4-byte integer to 1-byte integer, this will save memory a little, and maybe also be good for cache.

terrelln:
@animalize I want to take a slightly different approach. I really like the refactoring to reduce code duplication. However, I also want to make sure the software backups are testable.

I'm doing similar work in one of our other projects. So I am planning on sharing the code between the two.

terrelln:
And the fact that I'm putting it off until after our release doesn't reflect anything on the quality / review-ability of the code. It's just a somewhat larger change to the core of the library. And while it's important for ease of maintenance, there isn't a strong reason it needs to be in the 1.5.1 release. So I'd rather merge it after the release, and get more testing in the dev branch, rather than jam it in 1 week before a release.

ghost:
I'm glad to hear that, especially Linux supports some non-mainstream compilers and hardwares, it will be more secure.
I close this PR.

ghost:
> So I'd rather merge it after the release, and get more testing in the dev branch, rather than jam it in 1 week before a release.

I understand.

PR #2839, opened 2021-11-01
Improvements to verbose mode output

### LINKED ISSUE #2834: zstd verbose output could give more info
Here is the output from `xz -9 -vv FILE`:
```
xz: Filter chain: --lzma2=dict=64MiB,lc=3,lp=0,pb=2,mode=normal,nice=64,mf=bt4,depth=0
xz: 674 MiB of memory is required. The limiter is disabled.
xz: Decompression will need 65 MiB of memory.
/home/user/tmp/FILE (1/1)
  100 %     781.0 KiB / 9,367.9 KiB = 0.083                   0:02             
```

The extra bits that zstd is missing are, in order of importance IMO:
* Decompression memory requirements (rough number, it's not precise)
* Filter chain, i.e. the command line parameters that reproduce level 9 compression; It's super easy to copy-paste them and experiment.
* Compression memory requirements

These are printed in advance, so if the user sees unreasonable values he can abort with Ctrl-C.

ghost:
**Draft PR, still in progress**

When completed, this will close #2834.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1332,6 +1332,7 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
     FILE* const dstFile = ress.dstFile;
     U64 compressedfilesize = 0;
     ZSTD_EndDirective directive = ZSTD_e_continue;
+    U64 pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN;
 
     /* stats */
     ZSTD_frameProgression previous_zfp_update = { 0, 0, 0, 0, 0, 0 };
@@ -1348,12 +1349,25 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
 
     /* init */
     if (fileSize != UTIL_FILESIZE_UNKNOWN) {
+        pledgedSrcSize = fileSize;
         CHECK(ZSTD_CCtx_setPledgedSrcSize(ress.cctx, fileSize));
     } else if (prefs->streamSrcSize > 0) {
       /* unknown source size; use the declared stream size */
+      pledgedSrcSize = prefs->streamSrcSize;
       CHECK( ZSTD_CCtx_setPledgedSrcSize(ress.cctx, prefs->streamSrcSize) );
     }
 
+    {
+        int windowLog;
+        UTIL_HumanReadableSize_t windowSize;
+        CHECK(ZSTD_CCtx_getParameter(ress.cctx, ZSTD_c_windowLog, &windowLog));
+        if (windowLog == 0) {
+            const ZSTD_compressionParameters cParams = ZSTD_getCParams(compressionLevel, fileSize, 0);
+            windowLog = cParams.windowLog;
+        }
+        windowSize = UTIL_makeHumanReadableSize(MAX(1ULL, MIN(1ULL << windowLog, pledgedSrcSize)));
+        DISPLAYLEVEL(4, "Decompression will require %.*f%s of memory\n", windowSize.precision, windowSize.value, windowSize.suffix);
+    }
     (void)srcFileName;
 
     /* Main compression loop */


lib/zstd.h
```diff
@@ -1053,6 +1053,7 @@ ZSTDLIB_API size_t ZSTD_DCtx_refPrefix(ZSTD_DCtx* dctx,
  *  Note that object memory usage can evolve (increase or decrease) over time. */
 ZSTDLIB_API size_t ZSTD_sizeof_CCtx(const ZSTD_CCtx* cctx);
 ZSTDLIB_API size_t ZSTD_sizeof_DCtx(const ZSTD_DCtx* dctx);
+ZSTDLIB_API size_t ZSTD_sizeof_DCtxWindow(const ZSTD_DCtx* dctx);
```
Cyan4973: No new library symbol is allowed at "stable" status directly at creation.
When a new symbol is needed, it starts its life as "experimental" (below section)

lib/decompress/zstd_decompress.c
```diff
@@ -225,6 +225,8 @@ size_t ZSTD_sizeof_DCtx (const ZSTD_DCtx* dctx)
            + dctx->inBuffSize + dctx->outBuffSize;
 }
 
+size_t ZSTD_sizeof_DCtxWindow(const ZSTD_DCtx* dctx) { return dctx->fParams.windowSize; }
```
Cyan4973: I don't think there is a need for a new library symbol to deliver this information.
Look at `ZSTD_getFrameHeader()`.

Cyan4973 (changes requested):

Cyan4973:
Generally speaking, for a CLI feature, we try to keep all code modifications at CLI level only (directory `programs/`).
Library territory (directory `lib/`) is supposed to be off-limit, except in special circumstances where the feature is really not possible or too inconvenient without additional library support.

Cyan4973:
This PR delivers a feature which displays the window size of the currently decompressed frame if the CLI has received verbosity level 4 or more (aka command `-vv`). A few points:

- This is different from requested feature at #2834, which asks for this information at _compression_ time instead. It doesn't make the feature bad, just different from the request, and therefore cannot be used to close afore-mentioned issue.
- With #2834, the request is to imitate `xz -vv` during compression. However, there is nothing equivalent during decompression : `unxz -vv` doesn't provide any information regarding window size. While this is not necessarily a blocker, it introduces a risk of goal definition : how should the feature behave, since there is no model to just copy ?
- Here comes the kicker : while many files consist of a single frame, there are cases of files which consist of multiple appended frames. Each frame is entitled to its window size. How does the feature work in this case ? Define it, and test this more complex case if you plan to deliver this feature.

ghost:
> This PR delivers a feature which displays the window size of the currently decompressed frame if the CLI has received verbosity level 4 or more (aka command `-vv`). A few points:
> 
> * This is different from requested feature at [zstd verbose output could give more info #2834](https://github.com/facebook/zstd/issues/2834), which asks for this information at _compression_ time instead. It doesn't make the feature bad, just different from the request, and therefore cannot be used to close afore-mentioned issue.
> * With [zstd verbose output could give more info #2834](https://github.com/facebook/zstd/issues/2834), the request is to imitate `xz -vv` during compression. However, there is nothing equivalent during decompression : `unxz -vv` doesn't provide any information regarding window size. While this is not necessarily a blocker, it introduces a risk of goal definition : how should the feature behave, since there is no model to just copy ?
> * Here comes the kicker : while many files consist of a single frame, there are cases of files which consist of multiple appended frames. Each frame is entitled to its window size. How does the feature work in this case ? Define it, and test this more complex case if you plan to deliver this feature.

Thanks for the clarification, and for bringing that more complex case to light, I was definitely unaware of that possibility. I do believe this feature was requested though (the first bulletpoint in the issue says: "Decompression memory requirements (rough number, it's not precise)").

Cyan4973:
> I do believe this feature was requested though (the first bulletpoint in the issue says: "Decompression memory requirements (rough number, it's not precise)").

It's possible to know the decompression memory requirement at compression time.
That's actually useful, because user creating the compressed document might realize that they are requiring too much resources to the decompression side in order to later access said document, and decide to alter parameters.

It also corresponds to the information delivered by `xz -vv` (third line). Example (`xz -vv enwik8`): 
```
xz: Filter chain: --lzma2=dict=8MiB,lc=3,lp=0,pb=2,mode=normal,nice=64,mf=bt4,depth=0
xz: 94 MiB of memory is required. The limiter is disabled.
xz: Decompression will need 9 MiB of memory.
```

ghost:
> > I do believe this feature was requested though (the first bulletpoint in the issue says: "Decompression memory requirements (rough number, it's not precise)").
> 
> It's possible to know the decompression memory requirement at compression time. That's actually useful, because user creating the compressed document might realize that they are requiring too much resources to the decompression side in order to later access said document, and decide to alter parameters.
> 
> It also corresponds to the information delivered by `xz -vv` (third line). Example (`xz -vv enwik8`):
> 
> ```
> xz: Filter chain: --lzma2=dict=8MiB,lc=3,lp=0,pb=2,mode=normal,nice=64,mf=bt4,depth=0
> xz: 94 MiB of memory is required. The limiter is disabled.
> xz: Decompression will need 9 MiB of memory.
> ```

Understood, sorry for the misunderstanding.

Cyan4973:
The updated PR looks good to me.

We could work on merging this one, then you can continue the topic by delivering more value in another PR.

programs/fileio.c
```diff
@@ -1346,14 +1346,25 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
 
     DISPLAYLEVEL(6, "compression using zstd format \n");
 
+    size_t pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN;
```
Cyan4973: This code base uses `C90` convention, which only allows declaration of variables at the beginning of blocks.

Depending on the lifetime of the variable, you may either add the variable declaration at the top of current block,
or create a new sub-block specifically for this variable.

The second method is actually pretty good whenever the variable has a short lifespan.
This is for example the case for `int windowLog` below.

ghost:
> The updated PR looks good to me.
> 
> We could work on merging this one, then you can continue the topic by delivering more value in another PR.

Sure, or I could add more self-contained commits to this branch to reduce the number of merge commits that end up on `dev` and thereby avoid polluting the git history. Whichever you prefer.

Cyan4973:
Whichever method you select is good to me.

If you have a short delay between 2 consecutive commits, it seems better to continue stacking commits in the same PR.
But if delays grow too much between consecutive commits, it becomes better to merge what is ready to be merged, and give freedom to restart the work later without the pressure to "not forget" previous work.

ghost:
Alright, in that case if I can get the next change in by the end of the day today I'll keep it in this PR, and otherwise I'll start a new one.

programs/fileio.c
```diff
@@ -1332,6 +1332,7 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
     FILE* const dstFile = ress.dstFile;
     U64 compressedfilesize = 0;
     ZSTD_EndDirective directive = ZSTD_e_continue;
+    size_t pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN;
```
Cyan4973: `size_t` is a memory-related length, expressed as a 32-bit value on 32-bit systems, and 64-bit on 64-bit systems.

However, `ZSTD_CONTENTSIZE_UNKNOWN` is always a 64-bit value.

There is a type mismatch on 32-bit.

I believe you want a 64-bit type for `pledgedSrcSize`, as it's the type expected by the `zstd` API for this field.

Cyan4973 (changes requested):
one last type mismatch to fix

programs/fileio.c
ghost: Oh, okay I was inferring `size_t` from `prefs->streamSrcSize` being a `size_t`, but given that `fileSize` is a `U64` I see that it makes sense to go with the wider type. Will fix.

Cyan4973 (approved):

Cyan4973:
Thanks @Svetlitski-FB !

PR #2846, opened 2021-11-03
Use unused functions to appease Visual Studio

senhuang42:
Will need to be merged before #2845 to check that tests pass.
Unfortunately, Visual Studio doesn't have a good way of specifying unused static functions. It seems only versions below 2015 complain about this.

The alternative to this is to suppress warning C4505 for these files:
```
#ifdef _MSC_VER    /* Visual Studio */
#  pragma warning(disable : 4505)        /* disable: C4505: unused function */
#endif
```

To the beginning of `huf_compress.c` and `zstd_compress.c`.

### DIFF
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -471,8 +471,7 @@ static void HUF_swapNodes(nodeElt* a, nodeElt* b) {
 }
 
 /* Returns 0 if the huffNode array is not sorted by descending count */
-UNUSED_ATTR
-static int HUF_isSorted(nodeElt huffNode[], U32 const maxSymbolValue1) {
+MEM_STATIC int HUF_isSorted(nodeElt huffNode[], U32 const maxSymbolValue1) {
     U32 i;
     for (i = 1; i < maxSymbolValue1; ++i) {
         if (huffNode[i].count > huffNode[i-1].count) {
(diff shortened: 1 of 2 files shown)


binhdvo (approved):

Cyan4973:
I'm surprised this is the first time this warning would get noticed.
I would expect there are many other static functions throughout the code that end up being unused, depending on compilation directives.

Cyan4973:
For static symbols that may end up being unused (notably in included header files) we use `MEM_STATIC` instead of `static`. I would expect this method to continue to work fine even on Visual.

terrelln:
If the functions are truly unused, we could delete them too.

Cyan4973:
I believe `HUF_isSorted()` is a test function, which is only used as part of an `assert()`, 
and therefore calling it disappears when `assert` is disabled.

Strange though that the warning doesn't fire up on release `gcc`/`clang` builds.

senhuang42:
Actually we should just remove `ZSTD_convertBlockSequencesToSeqStore()`. That stuff is purely for the WIP hardware acceleration efforts, which is tracked in the `hardware_acceleration_integration` branch. So no need to pollute the main `dev` branch with this stuff when it isn't being used yet (and it will still be preserved for the `hardware_acceleration_integration` branch).

Cyan4973 (approved):

PR #2847, opened 2021-11-05
Display command line parameters with concrete values in verbose mode

ghost:
Continue addressing #2834

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1809,11 +1809,11 @@ void FIO_displayCompressionParameters(const FIO_prefs_t* prefs) {
     DISPLAY("%s", INDEX(rowMatchFinderOptions, prefs->useRowMatchFinder));
     DISPLAY("%s", prefs->rsyncable ? " --rsyncable" : "");
     if (prefs->streamSrcSize)
-        DISPLAY(" --stream-size=%lu", prefs->streamSrcSize);
+        DISPLAY(" --stream-size=%u", (unsigned) prefs->streamSrcSize);
     if (prefs->srcSizeHint)
         DISPLAY(" --size-hint=%d", prefs->srcSizeHint);
     if (prefs->targetCBlockSize)
-        DISPLAY(" --target-compressed-block-size=%lu", prefs->targetCBlockSize);
+        DISPLAY(" --target-compressed-block-size=%u", (unsigned) prefs->targetCBlockSize);
     DISPLAY("%s", INDEX(compressLiteralsOptions, prefs->literalCompressionMode));
     DISPLAY(" --memory=%u", prefs->memLimit ? prefs->memLimit : 128 MB);
     DISPLAY(" --threads=%d", prefs->nbWorkers);
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1782,6 +1782,48 @@ FIO_compressFilename_srcFile(FIO_ctx_t* const fCtx,
     return result;
 }
 
+static const char* checked_index(const char* options[], size_t length, size_t index) {
+    assert(index < length);
+    return options[index];
+}
+
+#define INDEX(options, index) checked_index((options), sizeof(options)  / sizeof(char*), (index))
+
+void FIO_displayCompressionParameters(FIO_prefs_t* const prefs) {
+    static const char* formatOptions[5] = {ZSTD_EXTENSION, GZ_EXTENSION, XZ_EXTENSION,
+                                           LZMA_EXTENSION, LZ4_EXTENSION};
+    static const char* sparseOptions[3] = {" --no-sparse", "", " --sparse"};
+    static const char* checkSumOptions[3] = {" --no-check", "", " --check"};
+    static const char* rowMatchFinderOptions[3] = {"", " --no-row-match-finder", " --row-match-finder"};
+    static const char* compressLiteralsOptions[3] = {"", " --compress-literals", " --no-compress-literals"};
+
+    assert(g_display_prefs.displayLevel >= 4);
+
+    DISPLAY("--format=%s", formatOptions[prefs->compressionType]);
+    DISPLAY("%s", INDEX(sparseOptions, prefs->sparseFileSupport));
+    DISPLAY("%s", prefs->dictIDFlag ? "" : " --no-dictID");
+    DISPLAY("%s", INDEX(checkSumOptions, prefs->checksumFlag));
+    DISPLAY(" --block-size=%d", prefs->blockSize);
+    if (prefs->adaptiveMode)
+        DISPLAY(" --adapt=min=%d,max=%d", prefs->minAdaptLevel, prefs->maxAdaptLevel);
+    DISPLAY("%s", INDEX(rowMatchFinderOptions, prefs->useRowMatchFinder));
+    DISPLAY("%s", prefs->rsyncable ? " --rsyncable" : "");
+    if (prefs->streamSrcSize)
+        DISPLAY(" --stream-size=%lu", prefs->streamSrcSize);
+    if (prefs->srcSizeHint)
+        DISPLAY(" --size-hint=%d", prefs->srcSizeHint);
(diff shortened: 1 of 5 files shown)


ghost:
Ah, it seems that some of the tests are failing due to using different `CFLAGS` than those that are present in the `cmake` build, I will address these.

Cyan4973:
I believe the intention of #2834 is to know the parameters used in "default" circumstances, when they are not specified on command line.

For example : 

```
xz -kf -vv FILE
xz: Filter chain: --lzma2=dict=8MiB,lc=3,lp=0,pb=2,mode=normal,nice=64,mf=bt4,depth=0
xz: 94 MiB of memory is required. The limiter is disabled.
xz: Decompression will need 9 MiB of memory.
```

Note how `xz` is not provided any compression parameter on command line,
yet the `-vv` verbose command is providing detailed compression parameters in use.

The intention is to reproduce this level of details,
and then, possibly, extend it with additional capabilities that `zstd` supports but `xz` doesn't.
But we'll have to look at this second part, and pay attention to information clutter : 
we don't want to use precious console space to display mass amount of details of negligible importance.

For reference, currently, using this PR, I'm getting : 
```
./zstd -kf -vv FILE
*** zstd command line interface 64-bits v1.5.1, by Yann Collet ***
--format=.zst --no-sparse --block-size=0 --memory=134217728 --threads=1 --content-size
Decompression will require 39607 B of memory
```

The intention would rather be to generate a line like (for example): 
```
--zstd=wlog=23,clog=23,hlog=22,slog=6,mml=3,tlen=48,strat=6
```
and then possibly extend it with additional options if they are useful.

ghost:
So more or less what `--show-default-cparams` is presently doing, but in a format that is directly compatible with the command line args?

Cyan4973:
> So more or less what `--show-default-cparams` is presently doing, but in a format that is directly compatible with the command line args?

Yes, pretty close.

One expected difference though is that these parameters should be the ones actually used by the compressor.
So it's not "just" the "default" parameters, but if any of these parameters end up being adjusted,
this should be reflected on screen.

Cyan4973:
From a high level perspective, this PR does the expected job.
Now, a few tactical considerations : 

- The display logic is split into 2 parts, between `fileio.c`/`FIO_displayCompressionParameters()`, and `zstdcli.c`/`main()`.
That does not seem a good idea. Consider: 
   + move any display logic out of `main()` into its own function.
      * For consistency, this should be applied to `showDefaultCParams` too, even if it's not your fault that its display logic is currently within `main()`
    + For which reason is `displayCompressionParameters()` positioned within `fileio.c` ? In absence of any specific reason, it seems it should rather be present within `zstdcli.c`, since `fileio.c` is more about functions related to opening / reading / writing files and pipes.
- Could you imagine a way to test this new feature ? Not necessarily a complex / exhaustive test, but at least something that can check the basics, able to trigger an error signal if, in some future, the feature is erroneously cancelled by some future contribution.

ghost:
The reason `displayCompressionParameters()` is in `fileio.c` is that from `zstdcli.c`, `FIO_prefs_t` is an opaque `struct` (i.e. an incomplete type) and thus its fields cannot be accessed. I could remedy this by exposing the `struct` definition in the header file, but it looks like it is deliberately `typedef`'d right now to maintain an abstraction barrier, so I assumed you'd want to keep it that way.

programs/fileio.h
```diff
@@ -108,6 +108,7 @@ void FIO_setExcludeCompressedFile(FIO_prefs_t* const prefs, int excludeCompresse
 void FIO_setAllowBlockDevices(FIO_prefs_t* const prefs, int allowBlockDevices);
 void FIO_setPatchFromMode(FIO_prefs_t* const prefs, int value);
 void FIO_setContentSize(FIO_prefs_t* const prefs, int value);
+void FIO_displayCompressionParameters(FIO_prefs_t* const prefs);
```
Cyan4973: minor: it's probably `const FIO_prefs_t*`, because no content within `prefs` is modified by calling this function.

Cyan4973:
> The reason `displayCompressionParameters()` is in `fileio.c` is that from `zstdcli.c`, `FIO_prefs_t` is an opaque `struct` (i.e. an incomplete type) and thus its fields cannot be accessed. I could remedy this by exposing the `struct` definition in the header file, but it looks like it is deliberately `typedef`'d right now to maintain an abstraction barrier, so I assumed you'd want to keep it that way.

Yeah, these are good points.

With both angles having pros and cons, there is no "obvious" better solution.
So okay, let's keep it that way, with `displayCompressionParameters()` remaining within `fileio.c`.

programs/zstdcli.c
```diff
@@ -1411,25 +1411,33 @@ int main(int argCount, const char* argv[])
           assert(ZSTD_NB_STRATEGIES == strategyBounds.upperBound);
           (void)strategyBounds; }
 
-        if (showDefaultCParams) {
+        if (showDefaultCParams || g_displayLevel >= 4) {
             size_t fileNb;
             for (fileNb = 0; fileNb < (size_t)filenames->tableSize; fileNb++) {
                 unsigned long long fileSize = UTIL_getFileSize(filenames->fileNames[fileNb]);
                 const size_t dictSize = dictFileName != NULL ? (size_t)UTIL_getFileSize(dictFileName) : 0;
                 const ZSTD_compressionParameters cParams = ZSTD_getCParams(cLevel, fileSize, dictSize);
```
Cyan4973: This will give you the _default_ compression parameters, for a given compression level and input size.
But if any of these parameters is altered, using for example an advanced command line parameter, this won't be captured by this invocation, and therefore won't be faithfully reported.

ghost:
@Cyan4973 I noticed some GitHub actions are failing due to `size_t` not matching `printf` formatting directives because it is defined differently on different platforms. Normally I'd resolve this by using the `%zu` directive, but that was introduced with C99, and I remember you stating that the codebase adheres to C90, so I'm not sure how you want to address this.

Cyan4973:
Indeed, `%zu` would be the correct directive, but unfortunately is not `C90` compliant.

We generally solve this situation by using `%u` instead.
This, in turn, requires casting the value to display with `(unsigned)val`.

programs/fileio.c
```diff
@@ -1782,6 +1782,54 @@ FIO_compressFilename_srcFile(FIO_ctx_t* const fCtx,
     return result;
 }
 
+#ifdef __GNUC__
+static const char* checked_index(const char* options[], size_t length __attribute__((unused)), size_t index)
```
Cyan4973: minor: for such a warning, prefer adding `(void)length;` somewhere in the code below.
It's a portable way to express that a variable may not be used and it's intentional.
It avoids usage of compiler-specific extensions.

tests/test_verbose_output.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e -u -o pipefail
+
+make -j -C .. CFLAGS=-O1 zstd
+empty_file=$(mktemp)
+trap 'rm -f $empty_file' EXIT
+../zstd -vv "$empty_file" 2>&1 | \
+grep -q -E -- "--zstd=wlog=[[:digit:]]+,clog=[[:digit:]]+,hlog=[[:digit:]]+,\
```
Cyan4973: great !

Cyan4973 (changes requested):
Great PR @Svetlitski-FB !
I believe it's in good shape, and only have one minor request : 

Could you merge your test, in `tests/test_verbose_output.sh`
as part of `test/playTests.sh` ?
It seems to be a fast enough test, so it can be played as part of the baseline test suite which is always triggered (`make check`), typically near https://github.com/facebook/zstd/blob/dev/tests/playTests.sh#L769 to remain on-topic.

Cyan4973 (approved):

Cyan4973:
Thanks for delivering this new feature @Svetlitski-FB !

PR #2850, opened 2021-11-09
Fix Determinism Bug: Avoid Reducing Indices to Reserved Values

felixhandte:
Previously, if an index was equal to `reducerValue + 1`, it would get remapped
during index reduction to 1 i.e. `ZSTD_DUBT_UNSORTED_MARK`. This can affect
the parsing of the input slightly, by causing tree nodes to be nullified when
they otherwise wouldn't be. This hardly matters from a correctness or
efficiency perspective, but it does impact determinism.

So this Pull Request changes index reduction to avoid mapping indices to
collide with `ZSTD_DUBT_UNSORTED_MARK`.

I am somewhat concerned that `ZSTD_reduceTable_internal()` will be slower
now. I'm not sure how important the speed of this is in practive, but it
looks like it's written to be auto-vectorized, and I'm not sure the new
version is similarly vectorizable.

Credit to OSS-Fuzz for discovery.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2332,12 +2332,17 @@ ZSTD_reduceTable_internal (U32* const table, U32 const size, U32 const reducerVa
     for (rowNb=0 ; rowNb < nbRows ; rowNb++) {
         int column;
         for (column=0; column<ZSTD_ROWSIZE; column++) {
+            U32 newVal;
             if (preserveMark && table[cellNb] == ZSTD_DUBT_UNSORTED_MARK) {
+                /* This write is pointless, but is required(?) for the compiler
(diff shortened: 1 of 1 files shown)


Cyan4973:
Indeed, if I do remember correctly, the previous version was designed to be auto-vectorized, the outcome of which was observed using `godbolt`.

That being said, the index reducer is hardly a performance bottleneck, so even if this part takes a little bit longer, it shouldn't be materially sensible in any scenario.

Cyan4973 (approved):

felixhandte:
I've put up a new version that does a superfluous write, but that helps the compiler vectorize the loop (https://godbolt.org/z/Kq4oc139n).

PR #2858, opened 2021-11-14
Support Swift Package Manager

cntrump:


lib/include/zstd.h
```diff
@@ -0,0 +1 @@
+../zstd.h
```
Cyan4973: ?
This seems a strange content for a `*.h` header file ?

cntrump: Because SPM needs a separated `publicHeadersPath`, so I create a symbol link to `../zstd.h`

Cyan4973: What do you call a "symbol link" ?
Is that a specific SPM concept, or is it the standard unix' `ln -s` link ?

cntrump: Yes, It is a symbol link created by `ln -s ../zstd.h zstd.h`

Package.swift
```diff
@@ -0,0 +1,36 @@
...
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "zstd",
+            path: "lib",
+            sources: [ "common", "compress", "decompress", "dictBuilder" ],
+            publicHeadersPath: "include",
```
Cyan4973: What about `publicHeadersPath: "."` (or equivalent) ?
Is that forbidden or meaningless ?

cntrump: I tried, it can't work

```
$ swift build

'zstd' ./zstd: error: target 'zstd' has invalid header layout: umbrella header found at './zstd/lib/zstd.h', but directories exist next to it: ./zstd/lib/common, ./zstd/lib/compress, ./zstd/lib/decompress, ./zstd/lib/deprecated, ./zstd/lib/dictBuilder, ./zstd/lib/dll, ./zstd/lib/include, ./zstd/lib/legacy; consider removing them
```

And finally, I figure it out. Using a `module.modulemap` replace symbol link for public header.

And renamed lib name `zstd` to `libzstd`

Cyan4973 (approved):

Cyan4973:
Thanks for adding this Swift Package support !

Issue #2816, opened 2021-10-08
Zstd should try to draw the progress bar only on TTY


jimis:
I run zstd v1.4.4 as a cron job, and I get a progress bar output even when the stdout and stderr are not tty:
```
zstd --rm -19 $FILENAME  2>&1 | logger
```

Zstd tries to draw the progress bar, as a result I see this in the journal:
```
[386B blob data]
```

It would be best to imply `-q` when the stdout/stderr are not a tty. Or at least print `\n` instead of `\r` as a separator between the lines.

terrelln:
I think implying `-q` when `stderr` is not a TTY would make sense.

PR #2884, opened 2021-11-29
Imply -q when stderr is not a tty

binhdvo:
This avoids progress bar updates when inappropriate for stderr, to address https://github.com/facebook/zstd/issues/2816

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -683,7 +683,7 @@ FIO_openDstFile(FIO_ctx_t* fCtx, FIO_prefs_t* const prefs,
         }
 #endif
         if (!prefs->overwrite) {
-            if (g_display_prefs.displayLevel <= 1) {
+            if (g_display_prefs.displayLevel == 0) {
                 /* No interaction possible */
                 DISPLAY("zstd: %s already exists; not overwritten  \n",
                         dstFileName);
@@ -1870,7 +1870,7 @@ int FIO_compressMultipleFilenames(FIO_ctx_t* const fCtx,
     /* init */
     assert(outFileName != NULL || suffix != NULL);
     if (outFileName != NULL) {   /* output into a single destination (stdout typically) */
-        if (FIO_removeMultiFilesWarning(fCtx, prefs, outFileName, 1 /* displayLevelCutoff */)) {
+        if (FIO_removeMultiFilesWarning(fCtx, prefs, outFileName, 0 /* displayLevelCutoff */)) {
             FIO_freeCResources(&ress);
             return 1;
         }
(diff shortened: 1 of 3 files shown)


programs/zstdcli.c
```diff
@@ -1163,6 +1164,11 @@ int main(int argCount, const char* argv[])
         UTIL_refFilename(filenames, argument);
     }
 
+    /* If stderr is not a TTY, imply -q */
+    if (!IS_CONSOLE(stderr) && !quieted) {
```
Cyan4973: If the intention is to track that `g_displayLevel` is still in "default" mode,
testing `g_displayLevel == DISPLAY_LEVEL_DEFAULT` here
skips the need to create, track and test an additional variable.

senhuang42:
The tests that fail here seem to be related to PR #2322. Changing the `g_displayLevel` is unfortunately fraught and the mechanisms for deciding what should be displayed are convoluted - if the tests failures are not descriptive enough, there's a good amount of discussion in #2322 for what the intended behavior should be, which might help with determining the correct fix for this.

terrelln:
We should do this in the same place we handle quieting for stdout:

https://github.com/facebook/zstd/blob/d7e17363751974dc1ad10785deb4170b23bee0ec/programs/zstdcli.c#L1405

I'm not sure if it would be better to move the stdout one up, or this one down though.

programs/fileio.c
```diff
@@ -683,7 +683,7 @@ FIO_openDstFile(FIO_ctx_t* fCtx, FIO_prefs_t* const prefs,
         }
 #endif
         if (!prefs->overwrite) {
-            if (g_display_prefs.displayLevel <= 1) {
+            if (g_display_prefs.displayLevel == 0) {
```
Cyan4973: As the comment implies, the goal here is to take over decision when prompt interactions are disabled.
`-q` disables user interactions.
It's necessary for some script to not "hang" waiting for user input.

So the previous threshold `<= 1` was valid.

The condition could be rewritten differently, in order to be more understandable.
But it's nonetheless correct.

Cyan4973 (changes requested):
`-q` implies disabling user prompt.
The existing `displayLevel` cutoff thresholds are correct.

programs/fileio.c
binhdvo: We  have some CI tests, and I could imagine some scripts use this behavior, where stderr is not a tty, however we pipe in "y" or "yes" to continue execution.  With this change, those are now considered `-q` and fail if we disable user interactions.  I suppose in those use cases, we could just require that `-f` be used instead and we could remove those test cases.  Or, we could change this change so that rather than fully assuming `-q` when stderr is tty, we instead leave the display level alone and more specifically target the progress bar and suppress it.

terrelln (approved):

PR #2885, opened 2021-11-30
Limit `ZSTD_maxCLevel` to 21 for 32-bit binaries.

yoniko:
Adjusted the max level for 32-bit binaries to be 21 instead of 22.
Added CI test to check 32-bit binaries in a memory constrained environment.
The purpose of this change is to reduce the number of failures created by memory constrained 32-bit platforms.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -30,6 +30,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: make check on 32-bit
+      env:
+        CHECK_CONSTRAINED_MEM: true
       run: |
         sudo apt update
         APT_PACKAGES="gcc-multilib" make apt-install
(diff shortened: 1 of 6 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks @yoniko , I think this covers the topic pretty well.

I also really like the new test !

ghost:
If someone uses compression level 22 to compress small files, will this change reduce the compression rate?

Cyan4973:
Yes of course, at least theoretically, since level 21 is obviously less powerful than level 22 (though not by much).
But in practice, chances are that level 22 will crash a 32-bit system anyway, due to exhaustion of address space. So there might not be anything to compare too.

Furthermore, someone who absolutely wants stronger compression parameters can take matters into its own hand, and manually specify the compression parameters they wish, advanced compression parameters remain applied as requested.

ghost:
If there are no complaints from many users, and it's mainly to prevent CI failure, maybe no need to modify it. Just my two cents.

Issue #2865, opened 2021-11-16
[test] Test that libzstd.so has the no exec stack bit set 


terrelln:
Add a test case to detect the build bug fixed in PR #2857 so we don't regress in the future.

PR #2886, opened 2021-11-30
[test] Test that the exec-stack bit isn't set on libzstd.so

terrelln:
Tests that libzstd.so doesn't have the exec-stack bit set using
readelf. If the stack is marked executable systemd will refuse
to link against zstd. We now test that it isn't set on every PR.

Adds a test for PR #2857
Fixes Issue #2865

### DIFF
--- a/tests/libzstd_builds.sh
+++ b/tests/libzstd_builds.sh
@@ -23,14 +23,19 @@ mustBeAbsent() {
 
 # default compilation : all features enabled - no zbuff
 $ECHO "testing default library compilation"
-CFLAGS= make -C $DIR/../lib libzstd.a > $INTOVOID
+CFLAGS= make -C $DIR/../lib libzstd libzstd.a > $INTOVOID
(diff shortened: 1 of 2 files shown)


tests/libzstd_builds.sh
```diff
@@ -23,14 +23,19 @@ mustBeAbsent() {
...
-CFLAGS= make -C $DIR/../lib libzstd.a > $INTOVOID
+CFLAGS= make -C $DIR/../lib libzstd libzstd.a > $INTOVOID
 nm $DIR/../lib/libzstd.a | $GREP "\.o" > tmplog
 isPresent "zstd_compress.o"
 isPresent "zstd_decompress.o"
 isPresent "zdict.o"
 isPresent "zstd_v07.o"
 mustBeAbsent "zbuff_compress.o"
-$RM $DIR/../lib/libzstd.a tmplog
+rm tmplog
```
Cyan4973: minor : consistency : `rm` -> `$RM`

Cyan4973 (approved):

PR #2892, opened 2021-12-01
[CircleCI] Fix short-tests-0

### LINKED ISSUE #2785: [CI] Work around compiler bug and re-enable aarch64build
https://github.com/facebook/zstd/pull/2783 triggered an internal compiler error in the aarch64 compiler. E.g. this [run](https://app.circleci.com/pipelines/github/facebook/zstd/3169/workflows/214e6e68-b827-468c-b364-0b2d0a619f91/jobs/11255). I haven't been able to fix it, so I've just disabled the test for now. We have aarch64 coverage of the main binary in GitHub Actions [here](https://github.com/facebook/zstd/blob/dev/.github/workflows/dev-short-tests.yml#L243). What we're missing is the build of the other binaries covered by `make allzstd`.

Alternatively, `make allzstd` could be covered by the QEMU tests in GitHub Actions. But, I had some issues with linking in the naive attempt, so I haven't done it yet.

terrelln:
short-tests-0 were silently failing. I think because of the `&& make clean` construction. Switch to `;` instead.

Fix the reported build errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,12 +15,13 @@ jobs:
           name: Test
           command: |
             ./tests/test-license.py
-            cc -v; CFLAGS="-O0 -Werror -pedantic" make all && make clean
-            make c99build         && make clean
-            make c11build         && make clean
-            make -j regressiontest&& make clean
-            make shortest         && make clean
-            make cxxtest          && make clean
+            cc -v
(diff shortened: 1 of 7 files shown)


terrelln:
So the aarch64 compiler error is still there. But, it turns out that our CircleCI tests are failing silently. Likely because they don't have a trailing `&&` after `make clean`.

Cyan4973 (approved):

Issue #2789, opened 2021-09-20
[Huffman Assembly] Expand supported platforms


terrelln:
The x86-64 Huffman assembly implementation is currently allow-listed to work on:
* Linux: `defined(__linux__) || defined(__linux)`
* MacOS: `defined(__APPLE__)`

The macro to check this is `HUF_ASM_SUPPORTED` defined in `huf_decompress.c`.

I've selected a very restricted set of platforms because I want to be sure that it is well tested before enabling it. Any platform that shares a calling convention with Linux should just work, but I want to be cautious of details, like working with interrupt handlers & debuggers. It should be easy to add BSD to this list, if someone wants to test that everything works as expected.

The assembly currently doesn't work on Windows because it has a different calling convention. And I had a really hard time getting the detection macros right. So if someone wants to take that on, that would be great.

Tasks:
* [ ] Test and enable support for x86-64 assembly on BSD.
* [ ] Add support for Windows in the x86-64 assembly & enable assembly in the Visual Studios build.
* [ ] Add support for assembly in the amalgamated build (`build/single_file_libs`). Probably by translating the `.S` file into inline assembly.
* [ ] Add an aarch64 assembly implementation.

ghost:
`huf_decompress_amd64.S` is written in [AT&T syntax](https://en.wikipedia.org/wiki/X86_assembly_language#Syntax).
Gcc's [GAS (GNU Assembler)](https://en.wikipedia.org/wiki/GNU_Assembler) supports AT&T syntax.

Visual Studio's [MASM (Microsoft Macro Assembler)](https://en.wikipedia.org/wiki/Microsoft_Macro_Assembler) only supports `Intel syntax`, so it may need to be rewritten to use it on Windows.

FYI, [isa-l](https://github.com/intel/isa-l) uses [NASM](https://en.wikipedia.org/wiki/Netwide_Assembler) to compile assembly code (`Intel syntax`), then `isa-l` supports gcc, clang, icc and VC compiler.

`GAS` also supports `Intel syntax`, maybe it's better to switch to `Intel syntax` in the long run.
> Many x86 assemblers use Intel syntax, including NASM, FASM, MASM, TASM, and YASM.
> GAS, which originally used AT&T syntax, has supported both syntaxes since version 2.10 via the .intel_syntax directive.

Of course, this needs to be decided based on the work of your team.

gvanem:
I have an error while building for `x86` and hit this issue. Basically, I just did:
```
cd build\cmake
[...]

PR #2893, opened 2021-12-01
[asm] Share portability macros and restrict ASM further

terrelln:
Move portability macros to `lib/common/portability_macros.h`. This file
only contains platform/feature detection (e.g. 0/1 macros). This file is
shared between C and ASM code, so it cannot include any C code.

Rename `HUF_` ASM macros to be `ZSTD_` prefixed, and move to the new
header.

Restrict `ZSTD_ASM_SUPPORTED` to `__GNUC__`, because we need the GAS
assembler.

Finally, only include the ASM code if we are actually going to use it.
This disables it on all Windows platforms, which should resolve the
problem brought up in Issue #2789.

### DIFF
--- a/build/single_file_libs/zstd-in.c
+++ b/build/single_file_libs/zstd-in.c
@@ -44,7 +44,7 @@
 #endif
 #define ZSTD_TRACE 0
 /* TODO: Can't amalgamate ASM function */
-#define HUF_DISABLE_ASM 1
+#define ZSTD_DISABLE_ASM 1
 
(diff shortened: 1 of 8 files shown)


Cyan4973 (approved):

Cyan4973:
Is there any need to add any CI test to detect issues like #2789 ?

terrelln:
I'll look into adding a CMake based Windows CI.

I'll put it in a separate PR, with a failing CI. Then rebase this on top of that.

PR #2895, opened 2021-12-02
Makefile: sort all wildcard file list expansions

kanavin:
Otherwise the order is non-deterministic and breaks
reproducible builds.


felixhandte:
I'll wait until the builds all succeed, but this looks good to me!

Cyan4973 (approved):

terrelln (approved):

PR #2896, opened 2021-12-02
Zstandard compiles and run on m68k cpus

### LINKED ISSUE #2756: m68k-unknown-linux-gnu: zstd-1.4.9 test suite fails badly

Cyan4973:
fix #2756

### DIFF
--- a/lib/common/mem.h
+++ b/lib/common/mem.h
@@ -81,26 +81,6 @@ extern "C" {
 
 
 /*-**************************************************************
-*  Alignment check
-*****************************************************************/
-
-/* C11 support */
-#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
-
-#include <stdalign.h>
-# define MEM_ALIGN_COND(T) alignof(T)
-
-#elif defined(__GNUC__) || defined(_MSC_VER)
-
-# define MEM_ALIGN_COND(T) __alignof(T)
-
-#else
-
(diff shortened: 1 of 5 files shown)


lib/common/mem.h
```diff
@@ -80,6 +80,26 @@ extern "C" {
...
+#include <stdalign.h>
+# define MEM_ALIGN_COND(T) alignof(T)
+
+#elif defined(__GNUC__) || defined(_MSC_VER)
+
+# define MEM_ALIGN_COND(T) __alignof(T)
+
+#else
+
+# define MEM_ALIGN_COND(T) 1  /* most likely incorrect, just to pass tests */
```
terrelln: Won't this break some of the alignment in `huf_compress.c` on non-supported compilers?

Could we maybe default to `MIN(roundUpToPow2(sizeof(T)), sizeof(size_t))`?

terrelln (approved):

Issue #2767, opened 2021-09-06
Problem with macro deprecated


Jamaika1:
GCC 12.0.0 20210826 X86_64 Windows  10
```
zbuff_compress.c: In function 'ZBUFF_compressInit_advanced':
zbuff_compress.c:76:5: warning: 'ZSTD_initCStream_advanced' is deprecated [-Wdeprecated-declarations]
   76 |     return ZSTD_initCStream_advanced(zbc, dict, dictSize, params, pledgedSrcSize);
      |     ^~~~~~
In file included from zbuff.h:189,
                 from zbuff_compress.c:17:
zstd.h:2216:8: note: declared here
 2216 | size_t ZSTD_initCStream_advanced(ZSTD_CStream* zcs,
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
zbuff_compress.c: In function 'ZBUFF_compressInitDictionary':
zbuff_compress.c:82:5: warning: 'ZSTD_initCStream_usingDict' is deprecated [-Wdeprecated-declarations]
   82 |     return ZSTD_initCStream_usingDict(zbc, dict, dictSize, compressionLevel);
      |     ^~~~~~
In file included from zbuff.h:189,
                 from zbuff_compress.c:17:
zstd.h:2196:8: note: declared here
 2196 | size_t ZSTD_initCStream_usingDict(ZSTD_CStream* zcs,
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

Cyan4973:
Not sure which application you are compiling,
but we would like to invite you to no longer use `ZBUFF_`,
which is an old API that has been deprecated a long time ago.
Prefer using stable APIs, which are fully supported now and in the future, and offer all the capabilities of `ZBUFF_` and more.

terrelln:
We no longer build zbuff as part of the library, unless deprecated APIs have been explicitly enabled. And we're planning on removing the source code from the library at some point in the future. So I strongly recommend you move away from the API.

However, if you are blocked by this in the short term, you can use `-Wno-deprecated-declarations` to silence the warning. But again, this will only work for so long, because we will be deleting this code.

Jamaika1:
I use ZSTDLIB_VISIBILITY=ZSTDLIB_API=/"/"
Adding ZBUFF_ doesn't fix anything

felixhandte:
The linked PR should fix this issue.

PR #2897, opened 2021-12-02
Avoid Using Deprecated Functions in Deprecated Code

felixhandte:
`lib/deprecated` is no longer built by zstd's bundled build files. However,
users may try to build these files when they import the source tree into
their own build systems. And if they have `-Wdeprecated-declarations` on,
this can produce warnings.

This PR migrates these files away from using deprecated declarations.

This addresses #2767.

### DIFF
--- a/lib/deprecated/zbuff_compress.c
+++ b/lib/deprecated/zbuff_compress.c
@@ -15,6 +15,7 @@
 ***************************************/
 #define ZBUFF_STATIC_LINKING_ONLY
 #include "zbuff.h"
+#include "error_private.h"
 
 
 /*-***********************************************************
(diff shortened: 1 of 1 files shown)


lib/deprecated/zbuff_compress.c
```diff
@@ -72,14 +72,36 @@ size_t ZBUFF_compressInit_advanced(ZBUFF_CCtx* zbc,
                                    const void* dict, size_t dictSize,
                                    ZSTD_parameters params, unsigned long long pledgedSrcSize)
 {
+    size_t ret;
     if (pledgedSrcSize==0) pledgedSrcSize = ZSTD_CONTENTSIZE_UNKNOWN;  /* preserve "0 == unknown" behavior */
-    return ZSTD_initCStream_advanced(zbc, dict, dictSize, params, pledgedSrcSize);
+    ret = ZSTD_CCtx_reset(zbc, ZSTD_reset_session_only);
+    ret = ZSTD_isError(ret) ? ret : ZSTD_CCtx_setPledgedSrcSize(zbc, pledgedSrcSize);
```
Cyan4973: minor : a macro would be helpful here to reduce repetition

felixhandte: Sure. I can use `FORWARD_IF_ERROR()` if we're comfortable taking an internal dependency on zstd. Otherwise I can gin up my own.

Cyan4973: It's reasonable to depend on `lib/common` for this code

lib/deprecated/zbuff_compress.c
```diff
@@ -15,6 +15,7 @@
 ***************************************/
 #define ZBUFF_STATIC_LINKING_ONLY
 #include "zbuff.h"
+#include "error_private.h"
```
Cyan4973: note : I believe we have decided to use relative `#include` path in this repository, so that should be `../common/error_private.h`

Cyan4973 (approved):

PR #2904, opened 2021-12-04
COPYING

Kactuskactus:


Cyan4973:
This is not a text that can be freely edited.
This license text comes from the GNU foundation, and shall be reproduced exactly identically.

PR #2905, opened 2021-12-05
add test case

15596858998:
add test case

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -184,7 +184,7 @@ fi
 
 
 println "\n===>  simple tests "
-datagen
+
 datagen > tmp
 zstd -h
 zstd -H
@@ -793,6 +793,8 @@ cat hello.tmp world.tmp > helloworld.tmp
 zstd -c hello.tmp > hello.zst
 zstd -c world.tmp > world.zst
 zstd -c hello.tmp world.tmp > helloworld.zst
+zstd -dc helloworld.zst > result.tmp
+$DIFF helloworld.tmp result.tmp
 cat hello.zst world.zst > helloworld.zst
 zstd -dc helloworld.zst > result.tmp
 cat result.tmp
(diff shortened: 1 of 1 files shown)


tests/playTests.sh
```diff
@@ -184,10 +184,16 @@ fi
 
 
 println "\n===>  simple tests "
-
+datagen
```
terrelln: Why did you add this call to `datagen`? This will output a ton of garbage to `stdout`.

tests/playTests.sh
```diff
@@ -785,6 +792,7 @@ println "world!" > world.tmp
 cat hello.tmp world.tmp > helloworld.tmp
 zstd -c hello.tmp > hello.zst
 zstd -c world.tmp > world.zst
+zstd -c hello.tmp world.tmp > helloworld.zst
```
terrelln, line 795: In order for this to be a useful test, you would want to decompress it and diff it. E.g. add the lines:

```
zstd -dc helloworld.zst > result.tmp
$DIFF helloworld.tmp result.tmp
```

tests/playTests.sh
```diff
@@ -1549,6 +1565,7 @@ if [ -n "$hasMT" ]
 then
     println "\n===>   adaptive mode "
     roundTripTest -g270000000 " --adapt"
+    roundTripTest -g270000000 " --adapt=" && die "invalid compression parameter"
```
terrelln: This shouldn't be a `roundTripTest`.

Try moving this down after line 1573, and doing:

```
zstd --adapt= tmp && die "invalid compression parameter"
```

terrelln (changes requested):

tests/playTests.sh
15596858998: To test whether the function of the Datagen without parameters by default is normal.

tests/playTests.sh
15596858998: ok，I'll try it.

tests/playTests.sh
15596858998: It's like this.

tests/playTests.sh
felixhandte: That's already tested by the following line.

15596858998:
@terrelln

terrelln (approved):

PR #2909, opened 2021-12-06
Change zstdless behavior to align with zless

binhdvo:
Flags for zstdless will now be passed to less instead of zstd.

Not for immediate merge; this will change the behavior for zstdless and would be better to bake in dev for awhile rather than going into the impending release.


programs/zstdless
```diff
@@ -1,2 +1,6 @@
 #!/bin/sh
-zstdcat "$@" | less
+
+# TODO: Address quirks and bugs tied to old versions of less, provide a mechanism to pass flags directly to zstd
+
+export LESSOPEN="|zstd -cdfq %s"
```
felixhandte: I think this should be

```
export LESSOPEN="|-zstd -cdfq %s"
```

I.e., we should add a dash so that it's also invoked on stdin (e.g., `... | zstd | zstdless`).

felixhandte (commented):
This definitely looks like the right approach! I have one nit.

Maybe this is outside the scope of this PR, but it makes me uncomfortable that we have no tests for `zstdless`. I see we have at least a couple for `zstdgrep` and `zstdcat` in `playTest.sh`. Maybe we could add a couple for `zstdless`?

programs/zstdless
```diff
@@ -1,2 +1,6 @@
...
+less "$@"
```
felixhandte: Also potentially we could use the `exec` command here (`exec less "$@"`), which would save us a process creation.

felixhandte:
Ok, looks good in terms of applying the requested changes. Did we decide adding tests is outside the scope of this PR? It would be nice to at least have a trivial few.

Cyan4973:
We don't need complex tests, but basic ones would be welcome.

That being said, I don't think they should be part of `playTests.sh`, because it would make this test script now dependent on `zstdless`, and there are situations where it's not desirable.

Probably making  a new small `tests/Makefile` entry for this purpose would be appropriate, something like `test-zstdless` for example. Whatever is simpler.

terrelln:
> That being said, I don't think they should be part of playTests.sh, because it would make this test script now dependent on zstdless, and there are situations where it's not desirable.

@Cyan4973  What scenarios are you thinking of? We do already use `zstdgrep` in this script.

Cyan4973:
> > That being said, I don't think they should be part of playTests.sh, because it would make this test script now dependent on zstdless, and there are situations where it's not desirable.
> 
> @Cyan4973 What scenarios are you thinking of? We do already use `zstdgrep` in this script.

`playTests.sh` is mostly associated to `make check`,
which is the "minimal set of fast tests for `zstd` CLI".
I can imagine `make check` being run on systems which only feature `zstd`, but no other derivative tools such as `zstdless` and `zstdgrep`.

Therefore, the comment I made regarding `zstdless` tests is also applicable for `zstdgrep` tests .

I presume these tests are currently present in `playTests.sh` because there was no other established solution nor requirements at the time.

Refactoring tests for `zstdless` and `zstdgrep` could be considered a separate task, in order to not mix topics.

binhdvo:
Added some very basic tests, I will expand on these and move both zstdgrep and zstdless tests in a followup.

Cyan4973 (approved):

tests/playTests.sh
```diff
@@ -322,6 +323,15 @@ ZCAT=./zstdcat "$ZSTDGREP" 2>&1 "1234" tmp_grep_bad.zst && die "Should have fail
 ZCAT=./zstdcat "$ZSTDGREP" 2>&1 "1234" tmp_grep_bad.zst | grep "No such file or directory" || true
 rm -f tmp_grep*
 
+println "\n===> zstdless tests"
+rm -f tmp_less
+echo "1234" > tmp_less
+zstd -f tmp_less
+lines=$("$ZSTDLESS" 2>&1 tmp_less.zst | wc -l)
```
felixhandte: why not a test of the form `[ "$($ZSTDLESS ...)" = "1234" ]`? The zstd-ed encoding likely doesn't have any newlines in it and so will probably report as one line as well.

programs/zstdless
felixhandte: It looks like we lost the changes that had been made here?

felixhandte (changes requested):

binhdvo:
Updated and also added environment variable to specify zstd path, necessary for some of the testing platforms

tests/playTests.sh
```diff
@@ -322,6 +323,19 @@ ZCAT=./zstdcat "$ZSTDGREP" 2>&1 "1234" tmp_grep_bad.zst && die "Should have fail
...
 
+println "\n===> zstdless tests"
+ln -sf "$ZSTD_BIN" zstd
+rm -f tmp_less*
+echo "1234" > tmp_less
+zstd -f tmp_less
+lines=$(ZSTD=./zstd "$ZSTDLESS" 2>&1 tmp_less.zst | wc -l)
+test 1 -eq $lines
+ZSTD=./zstd "$ZSTDLESS" -f tmp_less.zst > tmp_less_regenerated
+cat tmp_less_regenerated
```
felixhandte: Nit: `playTests.sh` is already very spammy, but this doesn't seem like a useful thing to print (assuming they're going to match).

tests/playTests.sh
```diff
@@ -322,6 +323,19 @@ ZCAT=./zstdcat "$ZSTDGREP" 2>&1 "1234" tmp_grep_bad.zst && die "Should have fail
...
+$DIFF tmp_less tmp_less_regenerated
+ZSTD=./zstd "$ZSTDLESS" 2>&1 tmp_less_bad.zst | grep "No such file or directory" || true
```
felixhandte: Shouldn't this be `|| die "..."`? You're trying to assert that you got this error message, right?

felixhandte (approved):
Cool. A couple nits. I guess you'd said you were planning to follow up with more thorough tests? The major thing that seems to be missing is passing flags to `zstdless` and seeing that they go to `less` and not `zstd`.

felixhandte (changes requested):
Oops, looks like [Circle CI isn't happy](https://app.circleci.com/pipelines/github/facebook/zstd/4059/workflows/a22aad8c-2cbb-4657-ab23-29baeae903b9/jobs/13327):

```
/home/circleci/project/tests/../programs/zstdless: 8: exec: less: not found
```

I guess it's reasonable to wrap all the tests in an

```
if [ -n "$(which less)"]; then
    ...
fi
```

binhdvo:
Test issues fixed

Cyan4973 (approved):

felixhandte (approved):
Ship it!

PR #2910, opened 2021-12-06
Reject Irregular Dictionary Files

felixhandte:
I hadn't seen #2890, so I wrote my own version. I like this approach a little
better, since it does an explicit check for a regular file, rather than
passing a magic value.

Addresses #2874.


terrelln (approved):

terrelln:
Looks good to me, except that the MinGW tests are failing. I wonder if the stat is failing?

Cyan4973 (approved):

PR #2911, opened 2021-12-07
Determine whether the pointer is empty before using it

15596858998:


terrelln:
C90 doesn't allow mixed declarations and code. I've put up PR #2916 to fix the issue, because we want to get it in before our next release.

Thanks for the PR and the report!

PR #2912, opened 2021-12-07
Fix Up #2659; Build libzstd.pc Whenever Building the Lib on Unix

felixhandte:
This PR builds on #2659 to add `-pthread` to `libzstd.pc` when the library is built with multithreading. It also promotes the `.pc` file to be built whenever the library is built (on Unix--unlike the original PR which builds it everywhere).

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -175,7 +175,7 @@ endif  # if windows
 libzstd : $(LIBZSTD)
 
 .PHONY: lib
-lib : libzstd.a libzstd libzstd.pc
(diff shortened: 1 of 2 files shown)


felixhandte:
Curiously, none of the GitHub Actions have run... Not sure why that is.

Cyan4973 (approved):

terrelln (approved):

ericonr:
@felixhandte thank you for your fix and for merging it!

Issue #2918, opened 2021-12-09
No message is displayed when -T is used to decompress the package.


15596858998:
I use the multi-thread function of -T to decompress the file, you can see that the decompression is successful, but we do not support multi-thread decompression, this will cause a misunderstanding to users.
![image](https://user-images.githubusercontent.com/33825201/145392081-502927ea-5321-44da-b16d-516b60ae8fab.png)
I think there's a need to do some processing here.

yoniko:
Thank you for reporting.
Issue was fixed in https://github.com/facebook/zstd/pull/3208.

PR #2920, opened 2021-12-09
Fix the problem of successful decompression using -T parameter

15596858998:
#2918

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -1353,6 +1353,7 @@ int main(int argCount, const char* argv[])
     }
 
     if (nbWorkers) {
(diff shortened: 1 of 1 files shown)


programs/zstdcli.c
```diff
@@ -1352,6 +1354,13 @@ int main(int argCount, const char* argv[])
         }
     }
 
+    if (useTparam) {
+        if (operation == zom_decompress) {
+            DISPLAY("error : can't use -T in decomrpession mode \n");
```
felixhandte, line 1359: Nit: please fix the spelling.

felixhandte (commented):
At most, this should be a warning.

I also think you can probably just look at `nbWorkers > 1` rather than define a new variable.

PR #2922, opened 2021-12-11
x32 compatibility

Cyan4973:
commit 1 : added x32 compatibility test

note : this test fails, due to ongoing issues with x32

commit 2 : fix the x32 issue


terrelln (approved):
Looks like the correct solution to me! Thanks for the fix!

PR #2925, opened 2021-12-13
Allow user to specify memory limit for dictionary training

embg:
Follow-up to https://github.com/facebook/zstd/pull/2809.

### DIFF
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -309,7 +309,7 @@ static fileStats DiB_fileStats(const char** fileNamesTable, int nbFiles, size_t 
 int DiB_trainFromFiles(const char* dictFileName, size_t maxDictSize,
                        const char** fileNamesTable, int nbFiles, size_t chunkSize,
                        ZDICT_legacy_params_t* params, ZDICT_cover_params_t* coverParams,
-                       ZDICT_fastCover_params_t* fastCoverParams, int optimize)
+                       ZDICT_fastCover_params_t* fastCoverParams, int optimize, unsigned memLimit)
 {
     fileStats fs;
     size_t* sampleSizes; /* vector of sample sizes. Each sample can be up to SAMPLESIZE_MAX */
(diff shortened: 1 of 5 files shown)


terrelln:
Looks good to me, except for `zstd.1`! If you're having a hard time getting `ronn-ng` to work, you can always revert the `zstd.1` changes, and we can regenerate it later.

programs/dibio.c
```diff
@@ -341,6 +341,11 @@ int DiB_trainFromFiles(const char* dictFileName, size_t maxDictSize,
         /* Limit the size of the training data to 2GB */
         /* TODO: there is opportunity to stop DiB_fileStats() early when the data limit is reached */
         loadedSize = (size_t)MIN( MIN((S64)maxMem, fs.totalSizeToLoad), MAX_SAMPLES_SIZE );
+        if (memLimit != 0) {
+            DISPLAYLEVEL(2, "!  Warning : setting manual memory limit for dictionary training data at %u MB \n",
+                (unsigned)(memLimit / (1 MB)));
```
Cyan4973, line 346: Tricky side effect:
if `loadedSize < memLimit`, the reported value might be incorrect.
Prefer updating `loadedSize` first, and then use the value to print the report.

tests/playTests.sh
```diff
@@ -1029,6 +1029,9 @@ then
 fi
 rm -f tmp* dictionary
 
+println "- Test --memory for dictionary compression"
+datagen -g12M -P90 > tmpCorpusHighCompress
+zstd --train -B2K tmpCorpusHighCompress -o tmpDictHighCompress --memory=10K && die "Dictionary training should fail : --memory too low (10K)"
```
Cyan4973, line 1034: It seems you are trying to generate a side effect (training failure) to prove that `--memory` is active.

I was wondering if it would be possible to consider a test that checks a successful action of `--memory`, for example by `grep`ing `stderr` and looking for the warning message that `--memory` will trigger ?

embg: Good idea, I will add a second test.

programs/dibio.c
embg: My reasoning was that it would be confusing for the user to report `loadedSize` as the limit. For example, suppose the total size of dictionary samples is 1 MB and the user sets `memLimit` at 10 MB. Then it may be confusing to see a message saying "the manual memory limit has been set at 1 MB".

Cyan4973: ah ok, you focus on the manual setting itself, rather than the effective limit.
Yes, that's valid.

embg:
Added a test which greps the stderr, and reverted zstd.1 changes (could not get ronn-ng working on Mac).

Cyan4973 (approved):

PR #2926, opened 2021-12-14
Fix performance degradation with -m32

binhdvo:
Addresses performance issues caused by gcc failing to properly inline ZSTD_memmove when used inside ZSTD_copy16 macro.  This change (from the original ZSTD_memcpy) was necessary because changes to the dctx handling now make overlap in memory a possibility.  That original change was introduced in 6a7ede3dfccbf3e0a5928b4224a039c260dcff72 and had a negative impact of about 15% when compiling with the -m32 flag and using gcc.

Performance when applying this change to that commit is reduced to a 5% drop for -m32 using gcc.  Performance improvement of this fix when applied to the current HEAD and using -m32 is 10%.

Performance testing indicates the loss was primarily due to this inlining behavior and not alignment issues.  Continuing to try different approaches to see if there is a way to recoup the remaining performance loss.

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -188,8 +188,13 @@ static void ZSTD_copy16(void* dst, const void* src) {
     vst1q_u8((uint8_t*)dst, vld1q_u8((const uint8_t*)src));
 #elif defined(ZSTD_ARCH_X86_SSE2)
     _mm_storeu_si128((__m128i*)dst, _mm_loadu_si128((const __m128i*)src));
-#else
+#elif defined(__clang__)
     ZSTD_memmove(dst, src, 16);
+#else
+    /* ZSTD_memmove is not inlined properly by gcc */
+    BYTE copy16_buf[16];
+    ZSTD_memcpy(copy16_buf, src, 16);
+    ZSTD_memcpy(dst, copy16_buf, 16);
 #endif
 }
 #define COPY16(d,s) { ZSTD_copy16(d,s); d+=16; s+=16; }


lib/common/zstd_internal.h
```diff
@@ -188,8 +188,13 @@ static void ZSTD_copy16(void* dst, const void* src) {
     vst1q_u8((uint8_t*)dst, vld1q_u8((const uint8_t*)src));
 #elif defined(ZSTD_ARCH_X86_SSE2)
     _mm_storeu_si128((__m128i*)dst, _mm_loadu_si128((const __m128i*)src));
-#else
+#elif defined(__clang)
     ZSTD_memmove(dst, src, 16);
+#else
+    /* ZSTD_memmove is not inlined properly by gcc */
+    uint8_t copy16_buf[16];
```
terrelln: nit: `BYTE` not `uint8_t`.

terrelln:
What does the x86-64 performance look like with this change?

lib/common/zstd_internal.h
```diff
@@ -188,8 +188,13 @@ static void ZSTD_copy16(void* dst, const void* src) {
     vst1q_u8((uint8_t*)dst, vld1q_u8((const uint8_t*)src));
 #elif defined(ZSTD_ARCH_X86_SSE2)
     _mm_storeu_si128((__m128i*)dst, _mm_loadu_si128((const __m128i*)src));
-#else
+#elif defined(__clang)
```
terrelln: This should be `__clang__` not `__clang`.

If x86-64 perf is not neutral, you could make this `elif defined(__clang__) || defined(__x86_64__) || defined(_M_X64)`.

binhdvo:
> What does the x86-64 performance look like with this change?

It was neutral as compared to the current HEAD

ghost:
There is a SSE2 path:
```c
#elif defined(ZSTD_ARCH_X86_SSE2)
    _mm_storeu_si128((__m128i*)dst, _mm_loadu_si128((const __m128i*)src));
```

~~But even use `-m32 -msse2` options, gcc doesn't use that path:~~
https://godbolt.org/z/Yz3qsaca7

~~ Maybe `ZSTD_ARCH_X86_SSE2` macro can be improved to support 32bit+SSE2.~~

**edit**: `-m32 -msse2` shoudl enable `ZSTD_ARCH_X86_SSE2` macro.
It seems clang use SSE2 by default, gcc doesn't.

Cyan4973:
> 
> ~~ Maybe `ZSTD_ARCH_X86_SSE2` macro can be improved to support 32bit+SSE2.~~
> 

`-m32 -msse2` enables `ZSTD_ARCH_X86_SSE2`.
But `-m32` alone doesn't (and shouldn't, since `x86` cpus without `sse2` exist, in contrast with `x64`).
Verified for both `gcc` and `clang`.

Cyan4973 (approved):

PR #2928, opened 2021-12-14
Refactored exec sequence

yoniko:


yoniko:
Opened by mistake, still premature

PR #2929, opened 2021-12-14
simplify SSE implementation of row_lazy match finder

Cyan4973:
The refactor is proposed mostly for maintenance convenience.

Performance wise, there is very little change,
I measure it slightly faster for slog 3 & 4 (~+2%)
but neutral or very slightly negative for slot 5 & 6 (< 1 %)
on both `gcc` and `clang`.
It feels like an okay side effect
(we want to improve speed more for faster modes).

But the real goal was to reduce the amount of intrinsic code.

This approach could not be easily extended to `NEON` because the current implementation is significantly different between the 16, 32, and 64 entries.


felixhandte (approved):
Looks correct to me.

PR #2930, opened 2021-12-14
Reduce tables to 8bit

nolange:


felixhandte (approved):
Looks like this saves 4KB. Cool!

I looked at the disassembled diff on amd64 and confirmed that this just replaces `mov` instructions with `movzbl`, so it should otherwise be a total no-op.

PR #2931, opened 2021-12-15
fixed incorrect rowlog initialization

Cyan4973:
the variable has only very limited usage,
being used only once at the beginning of the block, and only for prefetching,
hence, with no impact on final compression ratio, the error remained hidden.


felixhandte (approved):
So the intended change is that the rowLog can also be 6?

I guess that's right, since looking at your previous diff there are specializations for 16, 32, and 64.

LGTM.

Cyan4973:
> So the intended change is that the rowLog can also be 6?
> 
> I guess that's right, since looking at your previous diff there are specializations for 16, 32, and 64.

Yep, exactly.

PR #2934, opened 2021-12-16
rebalance lazy compression levels

Cyan4973:
This is a proposed update to rebalance compression levels.
It makes levels stronger than then used to be in `v1.5.0`, and closer to (though generally still faster than) `v1.4.9`, which achieves a more gradual scale towards level 13 (the first `btlazy2` level).

It's still difficult to create a "good" level 12, because `rowhash` is just very fast, so there's only so much that can be slowed down by increasing parameters to the max.

note: having looked at `256K` table, there is no obvious update, the current table already tries to be balanced, and there is still a large gap between strongest `lazy2` and fastest `btlazy2`, simply because `lazy2` is very fast, even when pushed to its maximum.


Cyan4973:
arf, tons of updates to do in the regression test ... :(

Cyan4973:
After extensive testings on `x64`, `x86` and `aarch64`, this new map of compression levels produces a more "regular" speed / compression ratio curve on these systems (than the map currently defined in `dev`).

felixhandte (approved):
Discussed my concerns about memory monotonicity offline. Looks good to me.

This is definitely more regular:

Screenshot from 2021-12-20 14-32-00

PR #2935, opened 2021-12-17
Remove code duplication in exec sequence

yoniko:
Removed code duplication in exec sequence and consolidated the flow for split / non-split buffers.
Maintained performance benefits of separate functions by moving most of the code into inlined templates that are inlined into the specialized functions with a constant choosing the correct behavior.

I inspected the binary to make sure compilation works as expected on clang. Also ran benchmarks on levels 1-15 on clang-12, gcc-7, gcc-8, gcc-9 and gcc-10 with synthesized data. Benchmarks showed performance is on par with dev.

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -854,7 +854,7 @@ size_t ZSTD_execSequenceEndInner(BYTE* op,
                             BYTE* const oend, const BYTE* const oend_w, seq_t sequence,
                             const BYTE** litPtr, const BYTE* const litLimit,
                             const BYTE* const prefixStart, const BYTE* const virtualStart, const BYTE* const dictEnd,
-                            const int isSplitLitBuffer)
+                            const unsigned isSplitLitBuffer)
 {
     BYTE* const oLitEnd = op + sequence.litLength;
     size_t const sequenceLength = sequence.litLength + sequence.matchLength;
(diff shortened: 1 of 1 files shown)


binhdvo:
Great, thanks for refactoring this, will take a look shortly.  For benchmarking, it would be worth using some real data (silesia/calgary/canterbury).  Also using a high compression level on a large file (like enwik8) in order to exercise ZSTD_decompressSequencesLong_body which also uses these functions.

yoniko:
Good call. I can definitely see a regression here with enwiki8.
Looking into it.

lib/decompress/zstd_decompress_block.c
```diff
@@ -891,16 +889,34 @@ size_t ZSTD_execSequenceEnd(BYTE* op,
...
+ * to be optimized for many small sequences, since those fall into ZSTD_execSequence().
+ */
+FORCE_NOINLINE
+size_t ZSTD_execSequenceEnd(BYTE* op,
+    BYTE* const oend, const BYTE* const oend_w, seq_t sequence,
+    const BYTE** litPtr, const BYTE* const litLimit,
+    const BYTE* const prefixStart, const BYTE* const virtualStart, const BYTE* const dictEnd)
+{
+    return ZSTD_execSequenceEndInner(op, oend, oend_w, sequence, litPtr, litLimit, prefixStart,
+                              virtualStart, dictEnd, 0);
```
binhdvo, line 917: It would be good to either make isSplitLitBuffer an enum, or add a small comment at each of the references so that it's clear what 0/1 means.

lib/decompress/zstd_decompress_block.c
```diff
@@ -849,25 +849,17 @@ static void ZSTD_safecopyDstBeforeSrc(BYTE* op, BYTE const* ip, ptrdiff_t length
...
- * NOTE: This function needs to be fast for a single long sequence, but doesn't need
- * to be optimized for many small sequences, since those fall into ZSTD_execSequence().
- */
-FORCE_NOINLINE
-size_t ZSTD_execSequenceEnd(BYTE* op,
-    BYTE* const oend, seq_t sequence,
-    const BYTE** litPtr, const BYTE* const litLimit,
-    const BYTE* const prefixStart, const BYTE* const virtualStart, const BYTE* const dictEnd)
+FORCE_INLINE_TEMPLATE
+size_t ZSTD_execSequenceEndInner(BYTE* op,
```
binhdvo, line 853: ZSTD_execSequenceEnd_body and ZSTD_execSequence_body would match naming of other similar functions

yoniko:
This refactor creates some performance penalties in gcc-9. Closing the PR.

PR #2936, opened 2021-12-17
Add Kiwix to software list

data-man:
Kiwix uses https://github.com/openzim/libzim and offers terabytes of compressed Wikipedia data.
So I think Kiwix merits a place on the list. :smiley:


data-man:
Thanks, you are fast as Zstd! :)

PR #2937, opened 2021-12-18
Fixed compilation under MSVC

mikex86:
MSVC requires ASM_MASM instead of ASM.

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -6,7 +6,7 @@
 # LICENSE file in the root directory of this source tree) and the GPLv2 (found
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
-if (WIN32)
+if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
 project(libzstd C ASM_MASM)
 else()
 project(libzstd C ASM)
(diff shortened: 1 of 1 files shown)


Cyan4973:
Hi @mikex86, 
thanks for the suggestion,

Just to clarify the objectives : 
we already have "Visual Studio" tests in CI,
and they have been running fine so far.
What is this PR improving ?
And how to check that the improvement is active ?

mikex86:
The error I was receiving on build was:
```
CMake Error in libraries/zstd/build/cmake/lib/CMakeLists.txt:
  MSVC_RUNTIME_LIBRARY value 'MultiThreadedDebugDLL' not known for this ASM
  compiler.
```
After further examination, this seems to have been the result of variable manipulation by another submodule in my project (https://github.com/KhronosGroup/Vulkan-Loader).
For those encountering this issue, `set(CMAKE_MSVC_RUNTIME_LIBRARY "")` seems to be another solution to the problem.
The reason why I thought this was required is that building with MSVC clang-cl worked just fine,
which utilized a GNU-like assembler and was thus not affected by the issue.
The fact that setting the language to ASM_MASM seems to have miraculously fixed it, made me believe I had a complete understanding of the issue.
This is not an issue of zstd and I am sorry for having made conclusions too quickly.

Issue #2933, opened 2021-12-16
Add Tools to Scale Compression Resources on Constrained Systems


felixhandte:
**Is your feature request related to a problem? Please describe.**
Developers who want to use zstd in its higher compression modes face difficult choices when their application is shipped to and runs on highly disparate devices. On high powered server-class hardware, level 19 (or even higher) may be appropriate, but the same level on a constrained device (e.g., a Raspberry Pi) may take unconscionably long to complete--or even crash!

**Describe the solution you'd like**
We should offer users of zstd tools to scale down the resource consumption of compression based on the constraints of the system. The three relevant constraints are probably time, memory, and threads. The most pressing is probably memory consumption, since it can lead to crashes.

There are a number of ways to indicate these constraints and to reconcile them with the overall compression intent. My expectation is that the most straightforward is for users to select the compression level etc. that they'd like to use in the absence of constraints and then supply any constraints. It would then be zstd's job to reconcile them if they conflict. E.g.:

```c
ZSTD_compressionParameters ZSTD_selectCParamsMatchingConstraints(int cLevel, size_t inputSize, size_t maxMem) {
  ZSTD_compressionParameters cParams;
  do {
    /* cctx doesn't get any smaller after -1 */
    cParams = ZSTD_getCParams(cLevel, inputSize, 0);
    size_t size = ZSTD_estimateCStreamSize_usingCParams(cParams);
    if (size <= maxMem) {
      break;
    }
  } while (--cLevel >= -1);
  return cParams;
}
```

A follow-up topic would be whether zstd could determine these constraints itself (like `-T0` does for threads).

**Describe alternatives you've considered**
Users have the tools to make these choices themselves. However, it would put significant burden on them to have to understand how compression parameters map to memory usage etc.

**Additional context**
This came up on the [ubuntu mailing list](https://lists.ubuntu.com/archives/ubuntu-devel/2021-December/041726.html) in the context of rebuilding the initramfs.

felixhandte:
In short, I am proposing that to start we add a CLI flag `--max-mem`. Ideally, it could take arguments of the form `x%`, and would limit zstd to using that percentage of total system memory.

julian-klode:
Here I am wondering if it's possible for the non-memory case to do adaptive compression wherein we specify a target file size / compression ratio and a target time, and it then adjusts settings as it processes blocks to reach a (pareto?) optimum. But that might be too weird.

waveform80:
[...]

PR #2940, opened 2021-12-20
Revert "Limit `ZSTD_maxCLevel` to 21 for 32-bit binaries."

Cyan4973:
Reverts facebook/zstd#2885

There is a need for users of `zstd` CLI to ensure the application doesn't crash in 32-bit environment,
but it's a more complete need and more complex to achieve than just disabling level 22.
We'll revisit how to achieve this goal in a future PR.


felixhandte (approved):
Followups can potentially be tracked in #2933.

Issue #2248, opened 2020-07-16
multiple rules generate `zstd.lib`


tobq:
Building on windows with cmake > ninja > clang (> msvc) leads to the duplicate generation error

> cmake-build-debug> cmake -G Ninja ..

```
-- The C compiler identification is Clang 10.0.0 with GNU-like command-line
-- The CXX compiler identification is Clang 10.0.0 with GNU-like command-line
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/LLVM/bin/clang.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
[...]
-- CMAKE_INSTALL_LIBDIR: lib
-- ZSTD_LEGACY_SUPPORT not defined!
-- ZSTD_MULTITHREAD_SUPPORT is enabled
-- Configuring done
-- Generating done
CMake Error:
  Running

   'C:/STANDALONE-BINARIES/ninja.exe' '-C' 'cmake-build-debug' '-t' 'cleandead'

  failed with:

   ninja: error: build.ninja:6103: multiple rules generate libs/zstd/build/cmake/lib/zstd.lib [-w dupbuild=err]

CMake Error:
  Running

   'C:/STANDALONE-BINARIES/ninja.exe' '-C' 'cmake-build-debug' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:6103: multiple rules generate libs/zstd/build/cmake/lib/zstd.lib [-w dupbuild=err]

CMake Generate step failed.  Build files cannot be regenerated correctly.
```

> CMakeLists.txt

```
[...]
add_subdirectory(libs/zstd/build/cmake)
target_link_libraries(${TargetName} PRIVATE libzstd_static)
```

tobq:
@Cyan4973 This will be due to an unclean build. Clang is stricter in terms of duplicate generations - other libraries build fine.

jannkoeker:
This likely happens due to this piece of code:
https://github.com/facebook/zstd/blob/32a8443b5c0235c7569fe29bfeec59b677665778/build/cmake/lib/CMakeLists.txt#L108-L113
[...]

PR #2877, opened 2021-11-24
Add detection when compiling with Clang and Ninja under Windows

jannkoeker:
According to the [Clang documentation](https://clang.llvm.org/docs/MSVCCompatibility.html) Clang creates MSVC-compatible libraries by default when compiling for Windows. This includes the names of the resulting library files.
This fixes the ninja error `multiple rules generate lib/zstd.lib [-w dupbuild=err]` for Clang under Windows.

Fixes #2248 and propably #2770

PR #2947, opened 2021-12-22
Fix zstd-static output name with MINGW/Clang

MehdiChinoune:


Cyan4973 (approved):
This seems to match the intention of the script,
which specifies that this library name should only be used with MSVC.

I'm surprised we can't rely solely on the `MSVC` detection, even when using the `clang` back end, but hey, that's a different problem.

MehdiChinoune:
The problem is with #2877 where they tried to fix the issue #2248 which was the user fault who expect to get the same library name with clang in gnu mode, so instead of telling the user to use clang-cl which should be detected as MSVC they introduced this weird condition!

PR #2951, opened 2021-12-23
meson: fix MSVC support

eli-schwartz:
Regression from commit a5f2c45528032ed20c33e0f8cd2c163a800a0017. It is not possible to unconditionally add the asm sources, since not all compilers understand the .s file extension.

Specifically for meson, only compilers inheriting from the GNU mixin will allow a .s file at configure time.

zstd doesn't support asm for MSVC for the same basic reason; if/when Windows asm support is added, it would involve preprocessing with nasm, most likely.

### DIFF
--- a/build/meson/lib/meson.build
+++ b/build/meson/lib/meson.build
@@ -37,7 +37,6 @@ libzstd_sources = [join_paths(zstd_rootdir, 'lib/common/entropy_common.c'),
   join_paths(zstd_rootdir, 'lib/compress/zstd_opt.c'),
   join_paths(zstd_rootdir, 'lib/compress/zstd_ldm.c'),
   join_paths(zstd_rootdir, 'lib/decompress/huf_decompress.c'),
-  join_paths(zstd_rootdir, 'lib/decompress/huf_decompress_amd64.S'),
   join_paths(zstd_rootdir, 'lib/decompress/zstd_decompress.c'),
   join_paths(zstd_rootdir, 'lib/decompress/zstd_decompress_block.c'),
   join_paths(zstd_rootdir, 'lib/decompress/zstd_ddict.c'),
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Discovered while attempting to update the meson WrapDB entry for zstd. We have Windows CI, and it failed.

Cyan4973:
2 comments : 
- Maybe we need to upgrade our CI test suite with a test illustrating this issue.
- The fix seems entirely `meson` build oriented, meaning any other build system will face the same issue, and will need to be fixed accordingly. I wonder if we could find a source-code level solution, that would avoid being build-system dependent.

eli-schwartz:
Fundamentally, testing this would mean "attempt to build using MSVC" I guess.

My understanding was that your build matrix already tests cmake using the MSVC compiler and VS solution files backend, but does not test meson on windows except using mingw-gcc (and the ninja backend).

I don't know what a source code solution would be, to the fact that you have source code files that are only valid inputs to certain compilers. Asm isn't portable. The GNU Compiler Collection includes drivers for many programming languages, including C/C++ but also fortran, ada, objc, Java, golang, and, of relevance here, ***gas***. GNU as. An asm dialect specific to GNU, which clang also implements, but MSVC doesn't. Nor, one presumes, do most alternative C compilers unless they happen to inherit from gcc or clang.

You cannot pass an invalid source file to a compiler that doesn't accept it. I don't see any way of working around that anywhere other than a build system, because it is the build system that passes files to the compiler.

And, one alternative to writing GNU asm is to write nasm.exe-compatible asm, which works everywhere independent of compiler. Of course, that is because you need to use the nasm compiler instead. This will require even more build system modifications, not "source code" modifications, but at least it will be the same build system modifications to each build system.

eli-schwartz:
Regarding #2957 I do not have any clue why the current CI that runs cmake tests on VS + MSVC passed, but conda forge failed. I had originally assumed that it must work in cmake because I dunno, maybe cmake automatically ignores incompatible file extensions. I'm not a heavy cmake user so I wouldn't know. :D

The VS solution files presumably don't need to be updated since they can never include the asm source file to begin with (and don't need to do so conditionally).

eli-schwartz:
Is there anything more that you'd like me to do in this PR?

Cyan4973:
> Is there anything more that you'd like me to do in this PR?

Probably not, but since we have an ongoing action with regards to `asm` files, I wanted to couple the fate of this PR with it.
cc @terrelln .

terrelln (approved):
This looks good to me.

The file itself can be parsed as valid C (or anything that supports the C preprocessor) when [`ZSTD_ENABLE_ASM_X86_64_BMI2`](https://github.com/facebook/zstd/blob/a0b9520e38c459408330bfdfb62f666f331d3bed/lib/common/portability_macros.h#L95-L129) isn't `1`, because after the C preprocessor runs, the file will be empty.

Is it meson that is erroring, because it doesn't recognize the extension and doesn't pass it to the compiler, or is it the compiler that is erroring?

eli-schwartz:
The former.

terrelln:
@eli-schwartz I put up #2972, as I realized a small problem after I merged your PR.

Thanks the the PR!

PR #2954, opened 2021-12-23
storeSeq & mlBase : clarity refactoring

Cyan4973:
changed `ZSTD_storeSeq()` interface to accept `matchLength` instead of `mlBase` as parameter.

The new interface feels more logical, and removes the need to do `- MINMATCH` at every call site.
The new contract is checked with an `assert()`.

Also : changed `seqDef.matchLength` into `seqDef.mlBase`
since it more accurately describes what is stored in this field (`== matchLength - MINMATCH`).

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -285,9 +285,9 @@ typedef enum {
 *  Private declarations
 *********************************************/
 typedef struct seqDef_s {
-    U32 offset;         /* offset == rawOffset + ZSTD_REP_NUM, or equivalently, offCode + 1 */
+    U32 offset;    /* offset == rawOffset + ZSTD_REP_NUM, or equivalently, offCode + 1 */
     U16 litLength;
-    U16 matchLength;
+    U16 mlBase;    /* mlBase == matchLength - MINMATCH */
 } seqDef;
 
 /* Controls whether seqStore has a single "long" litLength or matchLength. See seqStore_t. */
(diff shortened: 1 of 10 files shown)


Cyan4973:
There is a second PR coming, focused on `offCode` sum-type parameter which combines `repCode` and `offset`.

Unfortunately, the sum-type numeric representation is a leaky abstractions, with many subtle consequences throughout the code. Even when keeping ambitions low (i.e. not change behavior, just make the code clearer), the resulting PR is still complex, updating a lot of lines across many files.

That's why this PR, focused on `matchLength` / `mlBase` is presented first, separately, as it's relatively easier to review on its own.

_note_ : the lone failing test is just a CI issue, `travisci` seems to have difficulties with all hw `aarch64` jobs lately.

lib/common/zstd_internal.h
```diff
@@ -285,9 +285,9 @@ typedef enum {
 *  Private declarations
 *********************************************/
 typedef struct seqDef_s {
-    U32 offset;         /* offset == rawOffset + ZSTD_REP_NUM, or equivalently, offCode + 1 */
+    U32 offset;    /* offset == rawOffset + ZSTD_REP_NUM, or equivalently, offCode + 1 */
     U16 litLength;
-    U16 matchLength;
+    U16 mlBase;    /* mlBase == matchLength - MINMATCH */
```
felixhandte, line 290: Would a more natural name be `mlCode`, by analogy to `offCode`?

felixhandte (approved):
Looks good to me!

lib/common/zstd_internal.h
Cyan4973: `mlCode` is already taken,
it relates to the FSE code actually entropy-encoded.
`mlBase` is the starting value that leads to later transformation `mlBase => mlCode + mlBits`.

PR #2962, opened 2021-12-29
Refactor offset+repcode sumtype 

Cyan4973:
The format of the `offset` parameter used by `ZSTD_storeSeq()` interface is weird.
To begin, with it's not an `offset`, it's actually a sum type, combining `offset` | `repcode`.

The numerical format of this sum-type is fully exposed to callers of `ZSTD_storeSeq()`, which consist of many units within the `zstd` project. They must know this numerical format for the parameter to make sense.

Due to historical reasons, this format is as follows : 
- 0,1,2 : represent repcodes 1,2,3 (respectively)
- 3+ : represent real offsets, but with a +2 (offset 1 ==> 3)

Not only do all call sites must know and use this numerical format directly,
but due to its pervasive presence in many parts of the code,
it has also become the de-facto numerical format of the sum-type for a growing set of other usages,
such as integration of 3rd-party match finders.

And the funny part is : it's not even the format actually stored in the `seqStore`, called `offBase`,
which is slightly different (+1), making this format effectively transient.

This PR tries to break this invisible dependency 
by using instead a centralized set of macros,
`STORE_OFFSET()` and `STORE_REPCODE()`,
to convert from raw `repcode` and `offset` values  into the sumtype required by `ZSTD_storeSeq()`.

Because the numerical format is also used for other purposes in other functions,
this set is complemented by accessor macros,
such as `STORED_IS_OFFSET()`,
so that future evolutions of the format can happen while preserving correctness of these other functions.

The end result of this PR is functionally equivalent to current version in `dev` branch,
with the intermediate transient format still present exactly as it were.
However, with all interactions with this transient numerical format now under control,
it makes it possible possible to later update this format,
for example by removing the intermediate transient representation to target directly the `offBase` format.
This would make the life of this field similar to existing `matchLength => mlBase => mlCode + mlBits`.

Such a second step would be the topic of a later PR, as this one is already complex enough as it is.

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -688,7 +688,7 @@ MEM_STATIC repcodes_t
 ZSTD_newRep(U32 const rep[ZSTD_REP_NUM], U32 const offBase_minus1, U32 const ll0)
 {
     repcodes_t newReps;
-    memcpy(&newReps, rep, sizeof(newReps));
+    ZSTD_memcpy(&newReps, rep, sizeof(newReps));
     ZSTD_updateRep(newReps.rep, offBase_minus1, ll0);
     return newReps;
 }
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -586,6 +586,7 @@ ZSTD_safecopyLiterals(BYTE* op, BYTE const* ip, BYTE const* const iend, BYTE con
 #define STORED_IS_REPCODE(o) ((o) < ZSTD_REP_NUM)
 #define STORED_OFFSET(o)  (assert(STORED_IS_OFFSET(o)), (o)-ZSTD_REP_MOVE)
 #define STORED_REPCODE(o) (assert(STORED_IS_REPCODE(o)), (o)+1)  /* returns ID 1,2,3 */
+#define STORED_TO_OFFBASE(o) ((o)+1)
 
 /*! ZSTD_storeSeq() :
  *  Store a sequence (litlen, litPtr, offCode and matchLength) into seqStore_t.
@@ -639,7 +640,7 @@ ZSTD_storeSeq(seqStore_t* seqStorePtr,
     seqStorePtr->sequences[0].litLength = (U16)litLength;
 
     /* match offset */
-    seqStorePtr->sequences[0].offBase = offBase_minus1 + 1;
+    seqStorePtr->sequences[0].offBase = STORED_TO_OFFBASE(offBase_minus1);
 
     /* match Length */
     assert(matchLength >= MINMATCH);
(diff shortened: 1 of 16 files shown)


lib/compress/zstd_compress_internal.h
```diff
@@ -600,14 +577,30 @@ static void ZSTD_safecopyLiterals(BYTE* op, BYTE const* ip, BYTE const* const ie
     while (ip < iend) *op++ = *ip++;
 }
 
+#define ZSTD_REP_MOVE     (ZSTD_REP_NUM-1)
+#define STORE_REPCODE_1 STORE_REPCODE(1)
+#define STORE_REPCODE_2 STORE_REPCODE(2)
+#define STORE_REPCODE_3 STORE_REPCODE(3)
+#define STORE_REPCODE(r) (assert((r)>=1), assert((r)<=3), (r)-1)
+#define STORE_OFFSET(o)  (assert((o)>0), o + ZSTD_REP_MOVE)
```
felixhandte, line 585: Note that this pattern means evaluating the argument must not have any side effects.

lib/compress/zstd_compress_internal.h
```diff
@@ -600,14 +577,30 @@ static void ZSTD_safecopyLiterals(BYTE* op, BYTE const* ip, BYTE const* const ie
...
+#define STORED_IS_OFFSET(o)  ((o) > ZSTD_REP_MOVE)
+#define STORED_IS_REPCODE(o) ((o) <= ZSTD_REP_MOVE)
+#define STORED_OFFSET(o)  (assert(STORED_IS_OFFSET(o)), (o)-ZSTD_REP_MOVE)
```
felixhandte, line 588: `STORE_...` and `STORED_...` are easily confused. Better antonyms of `STORE_...` might be `LOAD_...` or `RETRIEVE_...` or `DECODE_...`.

felixhandte (commented):
Overall this makes sense to me. I guess things will improve further in the next PR. I don't love the naming of the macros.

Cyan4973:
> Overall this makes sense to me. I guess things will improve further in the next PR. I don't love the naming of the macros.

The macro names will indeed be changed in next PR,
they will likely be clearer, it will also make more sense to fine-tune naming in this following PR.

lib/compress/zstd_compress_internal.h
Cyan4973: Agreed.
Initially, only the `STORE_*()` macros were supposed to exist.
The other ones (`STORED_*()`) were added later, when realizing that the numeric format of the sum type was relied upon in various parts of the code base. So this was a quick way to bring these usages under control. And then, the nb of `STORED_*()` macros kept increasing...

It's more a temporary situation.
In the following PR, macro names will be updated significantly.

lib/compress/zstd_compress_internal.h
Cyan4973: Yes, this is a known side effect of macros.
Indeed, these macros shall only be used with scalar variables, not invoking function calls.
And it happens this condition is well respected, throughout the whole code base.

For more defensive properties, a possible alternative would be to swap these macros with `inline` functions,
although this second solution introduces its own set of potential casting issues.

felixhandte (approved):

Issue #2963, opened 2021-12-29
libs built with executable stack on non amd64 architectures


pixelb:
Due to the effectively empty huf_decompress_amd64.S on non amd64 architectures,
the empty assembly file is still significant, and causes libs to be built with an executable stack.
This triggers various security issues, most obvious being failure to boot on non x86_64 fedora systems,
due to systemd not loading as it sees the dependent libzstd.so.1 as a security issue (which it is).

There were a few solutions discussed at https://bugzilla.redhat.com/2035802

A possible fix may be just to move the "non executable stack" directive outside the ifdef in the huf_decompress_amd64.S file

jpalus:
You beat me to it by 5 minutes :)

Can confirm that today 2 of my arm 32-bit boards (armv6hl and armv7hnl) did not boot after upgrade of zstd to 1.5.1.

For reference:
```
a5f2c45528032ed20c33e0f8cd2c163a800a0017 is the first bad commit
commit a5f2c45528032ed20c33e0f8cd2c163a800a0017
Author: Nick Terrell <terrelln@fb.com>
Date:   Fri Sep 17 11:43:04 2021 -0700

    Huffman ASM
```

PR #2964, opened 2021-12-30
Mark Huffman Decoder Assembly `noexecstack` on All Architectures

felixhandte:
Apparently, even when the assembly file is empty (because `ZSTD_ENABLE_ASM_X86_64_BMI2` is false), it still is marked as possibly needing an executable stack and so the whole library is marked as such. This commit applies a simple patch for this problem by moving the noexecstack indication outside the macro guard.

In addition, this PR ships flags to the Makefile-driven builds to explicitly request a non-executable stack. This ought to be redundant, but it appears on at least one platform (MIPS) to succeed where the source code note fails.

This PR builds on #2857.

This PR addresses #2963.

### DIFF
--- a/lib/decompress/huf_decompress_amd64.S
+++ b/lib/decompress/huf_decompress_amd64.S
@@ -3,7 +3,7 @@
 /* Stack marking
  * ref: https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
  */
-#if defined(__linux__) && defined(__ELF__)
+#if defined(__ELF__)
 .section .note.GNU-stack,"",%progbits
 #endif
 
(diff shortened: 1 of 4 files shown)


Cyan4973 (approved):

lib/decompress/huf_decompress_amd64.S
```diff
@@ -1,14 +1,14 @@
 #include "../common/portability_macros.h"
 
-#if ZSTD_ENABLE_ASM_X86_64_BMI2
-
 /* Stack marking
  * ref: https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
  */
 #if defined(__linux__) && defined(__ELF__)
 .section .note.GNU-stack,"",%progbits
```
terrelln, line 7: We may also want a `&& defined(__GNUC__)`

PR #2965, opened 2021-12-30
Converge sumtype (offset | repcode) numeric representation towards offBase

Cyan4973:
This completes #2962 , by ensuring that the numeric representation of the sumtype `(offset | repcode)` used at the interface of `ZSTD_storeSeq()` (and elsewhere) is the same as `offBase`, as stored in the `seqStore`.
This skips one transient numeric representation, hopefully simplifying the pipeline.

It's a good opportunity to rename all these transformations to better reflect their position in the pipeline, aka : 
`offset | repcode` => `offBase` => `offCode` + `offBits`
following the model previously established for `matchLength`.

This PR does not impact the compressed output.
The first commit left a subtle minor difference in `btlazy2` parser, resulting in rare minor differences in compressed output. This was later fixed in the last commit, so now there is no difference, even for `btlazy2`.

Also :
fixed `PLATFORM_POSIX_VERSION` within `zstdseek_decompress.c`.
This macro was used there without being defined.
This triggered warning / errors on my workstation during tests.

### DIFF
--- a/contrib/seekable_format/zstdseek_decompress.c
+++ b/contrib/seekable_format/zstdseek_decompress.c
@@ -24,12 +24,63 @@
 #endif
 
 /* ************************************************************
+*  Detect POSIX version
+*  PLATFORM_POSIX_VERSION = 0 for non-Unix e.g. Windows
+*  PLATFORM_POSIX_VERSION = 1 for Unix-like but non-POSIX
+*  PLATFORM_POSIX_VERSION > 1 is equal to found _POSIX_VERSION
+*  Value of PLATFORM_POSIX_VERSION can be forced on command line
+***************************************************************/
+#ifndef PLATFORM_POSIX_VERSION
+
+#  if (defined(__APPLE__) && defined(__MACH__)) || defined(__SVR4) || defined(_AIX) || defined(__hpux) /* POSIX.1-2001 (SUSv3) conformant */ \
+     || defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)  /* BSD distros */
+     /* exception rule : force posix version to 200112L,
+      * note: it's better to use unistd.h's _POSIX_VERSION whenever possible */
+#    define PLATFORM_POSIX_VERSION 200112L
+
+/* try to determine posix version through official unistd.h's _POSIX_VERSION (http://pubs.opengroup.org/onlinepubs/7908799/xsh/unistd.h.html).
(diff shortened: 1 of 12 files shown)


Cyan4973:
ping
this should be ready for merge, now that `v1.5.2` is published

Cyan4973:
ping 2

contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -23,13 +23,64 @@
 #  endif
 #endif
 
+/* ************************************************************
```
terrelln, line 26: Is this change intended? It isn't mentioned in the summary.

terrelln:
Can you please also mention what the first commit, "fixed minor compression difference in btlazy2" is fixing?

It isn't clear what the problem is. And I didn't realize that this PR  had a functional change until I looked at that commit title, I assumed it was all refactoring.

contrib/seekable_format/zstdseek_decompress.c
Cyan4973: Yes, 
`PLATFORM_POSIX_VERSION` is used a bit below
but it's not defined anywhere,
so it's silently `== 0` by default, always,
or with the right set of compiler flags, it generates a warning, or an error.

Hence, I define it here, using the same definition as in `platform.h`.

Cyan4973:
> Can you please also mention what the first commit, "fixed minor compression difference in btlazy2 " is fixing?
> 
> It isn't clear what the problem is. And I didn't realize that this PR had a functional change until I looked at that commit title, I assumed it was all refactoring.

The PR, as a whole, is "just" a refactoring.
But the first commit introduced a very subtle compression difference in `btlazy2` mode, at line 330.
It's an off-by-one difference in the way cost of offsets are evaluated.
It barely ever triggers, but with billions of attempts, it did happen to change the selection outcome from time to time.
The last commit fixes that, so that the PR as a whole does generate exactly the same compressed outcome as before, including for `btlazy2`.

terrelln:
Can you add that to the PR summary please?

Cyan4973:
> Can you add that to the PR summary please?

done

contrib/seekable_format/zstdseek_decompress.c
terrelln: Weird... I've been refreshing this page, but GitHub didn't show me this comment until now.

Can you also mention this in the PR summary please? Then the PR LGTM.

Cyan4973: done

terrelln (approved):

PR #2969, opened 2022-01-04
fix performance issue in scenario #2966 (part 1)

### LINKED ISSUE #2966: MUCH slower compression speeds using version 1.5.1
After updating to version 1.5.1 of the library I noticed that compression speeds were much worse compared to version 1.4.9 at the same compression level. In my benchmarks for my use case, which is compressing streaming network data, zlib now beats zstd with regard to speed and output size. 

To reproduce, build the attached source file first with version 1.4.9 used as a DLL, check the output, and then do the same with version 1.5.1 and check the output after running the program.

On my PC running Windows 10 and using Visual Studio 2022 to build, I get the following results with compressionlevel `10`:

with zstd 1.4.9: duration: **512 ms**, output size: 422 bytes
with zstd 1.5.1: duration: **71266 ms**, output size: 422 bytes

Even with compression level `1` there is a difference; 1.5.1 is slower than 1.4.9 albeit not by much compared to compression level `10`.

Where does this huge difference in duration come from?

[Source.zip](https://github.com/facebook/zstd/files/7794771/Source.zip)

Cyan4973:
Re-employing a compression state across multiple successive compression jobs
shall minimize the amount of allocation and initialization required.

This mostly matters in situations where initialization is an overwhelming task compared to compression itself.
This can happen when the amount to compress is small,
while the compression state was expecting it to be much larger,
aka, streaming mode without the benefit of a `srcSize` hindsight.
In which case, `zstd` can't rely on its automatic resource resizing capability, and initialization cost can only be made acceptable by amortizing it over a series of compressions.

This diff restores a lean-initialization optimization trick that used to be present up to `v1.4.9` and was lost in commit 980f3bbf8354edec0ad32b4430800f330185de6a .

The following measurement is taken on a core `i7-9700K` (turbo disabled) with `fullbench -b41`, using `geldings.txt` as sample (a small text file). The test corresponds to a scenario using `ZSTD_compressStream()` without the benefit of knowing the small sample size beforehand.

| level | v1.5.1 | this PR | comment
| --- | ---:| ---:| --- |
| 1 | 101 MB/s | 113 MB/s |
| 2 |  67 MB/s | 112 MB/s |
| 3 |  31 MB/s |  94 MB/s |
| 4 |  14 MB/s |  93 MB/s |

Note how important is this lean-initialization optimization for small data as the compression level increases, hence the amount of resources used increases too.

Note that this PR does not completely fix #2966,
since another heavy initialization, specific to `rowHash` mode,
is also present (and wasn't in `v1.4.9`).
This second issue will be fixed in a separate commit.

For information :
| level | v1.5.1 | this PR | comment
| --- | ---:| ---:| --- |
| 4 |  14 MB/s |  93 MB/s | `dfast`
| 5 |   9 MB/s |  20 MB/s | `rowHash`
| ... | ... | ... | `rowHash`
|12 | 0.4 MB/s | 1.2 MB/s | `rowHash`
|13 | 0.6 MB/s |  38 MB/s | `btlazy2`

### DIFF
--- a/lib/compress/zstd_cwksp.h
+++ b/lib/compress/zstd_cwksp.h
@@ -243,12 +243,14 @@ MEM_STATIC size_t ZSTD_cwksp_bytes_to_align_ptr(void* ptr, const size_t alignByt
 
 /**
  * Internal function. Do not use directly.
- * Reserves the given number of bytes within the aligned/buffer segment of the wksp, which
- * counts from the end of the wksp. (as opposed to the object/table segment)
+ * Reserves the given number of bytes within the aligned/buffer segment of the wksp,
+ * which counts from the end of the wksp (as opposed to the object/table segment).
  *
  * Returns a pointer to the beginning of that space.
  */
-MEM_STATIC void* ZSTD_cwksp_reserve_internal_buffer_space(ZSTD_cwksp* ws, size_t const bytes) {
+MEM_STATIC void*
+ZSTD_cwksp_reserve_internal_buffer_space(ZSTD_cwksp* ws, size_t const bytes)
+{
     void* const alloc = (BYTE*)ws->allocStart - bytes;
     void* const bottom = ws->tableEnd;
     DEBUGLOG(5, "cwksp: reserving %p %zd bytes, %zd bytes remaining",
@@ -260,6 +262,8 @@ MEM_STATIC void* ZSTD_cwksp_reserve_internal_buffer_space(ZSTD_cwksp* ws, size_t
         ws->allocFailed = 1;
         return NULL;
     }
+    /* the area is reserved from the end of wksp.
+     * If it overlaps with tableValidEnd, it voids guarantees on values' range */
     if (alloc < ws->tableValidEnd) {
         ws->tableValidEnd = alloc;
     }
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_cwksp.h
```diff
@@ -295,13 +301,13 @@ MEM_STATIC size_t ZSTD_cwksp_internal_advance_phase(
...
                 DEBUGLOG(5, "reserving table alignment addtl space: %zu", bytesToAlign);
-                RETURN_ERROR_IF(end > ws->workspaceEnd, memory_allocation,
+                RETURN_ERROR_IF(objectEnd > ws->workspaceEnd, memory_allocation,
                                 "table phase - alignment initial allocation failed!");
-                ws->objectEnd = end;
-                ws->tableEnd = end;
-                ws->tableValidEnd = end;
+                ws->objectEnd = objectEnd;
+                ws->tableEnd = objectEnd;
+                if (ws->tableEnd > ws->tableValidEnd) ws->tableValidEnd = objectEnd;
```
felixhandte: I believe this is the single line of change that actually changes the behavior of the code and that everything else is just formatting/naming/comments. (Correct me if I'm wrong!)

I guess this makes sense.

Nit: the phrasing is a bit awkward. Normally I would expect an expression of this form to be `if (a > b) b = a;`, but here you have `if (a > b) b = c;` where it happens to be the case that `a == c`. I think this would be clearer as:

```
if (objectEnd > ws->tableValidEnd) {
  ws->tableValidEnd = objectEnd;
}
```

felixhandte (approved):
Ok, I believe I understand the functional change.

(It would be helpful in the future to separate functional and aesthetic changes into their own commits.)

terrelln (approved):

PR #2974, opened 2022-01-05
Lazy parameters adaptation (part 1 - ZSTD_c_stableInBuffer)

Cyan4973:
This is merely a way to "prove" that lazy parameters adaptation can work,
though this implementation only makes it happen when `ZSTD_c_stableInBuffer` is enabled.
`ZSTD_c_stableInBuffer` pushes the responsibility of buffering to the user,
making lazy parameters adaptation simpler to achieve.

As a (favorable) side effect, it extends the nb of use cases supported by `ZSTD_c_stableInBuffer`.
Specifically, this parameter is now compatible with "appending" scenarios,
where more data is continuously added after compression has started,
instead of being restricted to presenting the entire input immediately at start.

It also makes `ZSTD_c_stableInBuffer` compatible with the regular streaming interface
aka `initCStream`, `compressStream`, `flushStream` and `endStream`
(was previously limited to `ZSTD_compressStream2( , , , zstd_e_end)` only).

### DIFF
--- a/lib/common/error_private.c
+++ b/lib/common/error_private.c
@@ -38,6 +38,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(tableLog_tooLarge): return "tableLog requires too much memory : unsupported";
     case PREFIX(maxSymbolValue_tooLarge): return "Unsupported max Symbol Value : too large";
     case PREFIX(maxSymbolValue_tooSmall): return "Specified maxSymbolValue is too small";
+    case PREFIX(stabilityCondition_notRespected): return "pledged buffer stability condition is not respected";
     case PREFIX(dictionary_corrupted): return "Dictionary is corrupted";
     case PREFIX(dictionary_wrong): return "Dictionary mismatch";
     case PREFIX(dictionaryCreation_failed): return "Cannot create Dictionary from provided samples";
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -177,12 +177,9 @@ size_t ZSTD_freeCCtx(ZSTD_CCtx* cctx)
     if (cctx==NULL) return 0;   /* support free on NULL */
     RETURN_ERROR_IF(cctx->staticSize, memory_allocation,
                     "not compatible with static CCtx");
-    {
-        int cctxInWorkspace = ZSTD_cwksp_owns_buffer(&cctx->workspace, cctx);
+    {   int cctxInWorkspace = ZSTD_cwksp_owns_buffer(&cctx->workspace, cctx);
         ZSTD_freeCCtxContent(cctx);
-        if (!cctxInWorkspace) {
-            ZSTD_customFree(cctx, cctx->customMem);
-        }
+        if (!cctxInWorkspace) ZSTD_customFree(cctx, cctx->customMem);
     }
     return 0;
 }
@@ -4487,6 +4484,7 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
 #define ZSTD_USE_CDICT_PARAMS_DICTSIZE_MULTIPLIER (6ULL)
 
 /*! ZSTD_compressBegin_internal() :
+ * Assumption : either @dict OR @cdict (or none) is non-NULL, never both
  * @return : 0, or an error code */
 static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
                                     const void* dict, size_t dictSize,
@@ -4567,11 +4565,11 @@ size_t ZSTD_compressBegin_advanced(ZSTD_CCtx* cctx,
                                             &cctxParams, pledgedSrcSize);
 }
 
-size_t ZSTD_compressBegin_usingDict(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, int compressionLevel)
+size_t
+ZSTD_compressBegin_usingDict(ZSTD_CCtx* cctx, const void* dict, size_t dictSize, int compressionLevel)
 {
     ZSTD_CCtx_params cctxParams;
-    {
-        ZSTD_parameters const params = ZSTD_getParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize, ZSTD_cpm_noAttachDict);
+    {   ZSTD_parameters const params = ZSTD_getParams_internal(compressionLevel, ZSTD_CONTENTSIZE_UNKNOWN, dictSize, ZSTD_cpm_noAttachDict);
         ZSTD_CCtxParams_init_internal(&cctxParams, &params, (compressionLevel == 0) ? ZSTD_CLEVEL_DEFAULT : compressionLevel);
     }
     DEBUGLOG(4, "ZSTD_compressBegin_usingDict (dictSize=%u)", (unsigned)dictSize);
@@ -5648,6 +5646,12 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
 
     /* transparent initialization stage */
     if (cctx->streamStage == zcss_init) {
+        if ( (cctx->requestedParams.inBufferMode == ZSTD_bm_stable) /* input is presumed stable, across invocations */
+          && (endOp == ZSTD_e_continue)                             /* more to come */
+          && (input->pos < ZSTD_BLOCKSIZE_MAX) ) {                  /* not even reached one block yet */
+            cctx->expectedInBuffer = *input;
+            return (ZSTD_BLOCKSIZE_MAX - input->pos);               /* don't do anything : allows lazy compression parameters adaptation */
+        }
         FORWARD_IF_ERROR(ZSTD_CCtx_init_compressStream2(cctx, endOp, input->size), "CompressStream2 initialization failed");
         ZSTD_setBufferExpectations(cctx, output, input);    /* Set initial buffer expectations now that we've initialized */
     }
@@ -6159,6 +6163,7 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size_t dstCapaci
 /*======   Finalize   ======*/
 
 /*! ZSTD_flushStream() :
+*   Note : not compatible with ZSTD_c_stableInBuffer
  * @return : amount of data remaining to flush */
 size_t ZSTD_flushStream(ZSTD_CStream* zcs, ZSTD_outBuffer* output)
 {
@@ -6169,7 +6174,9 @@ size_t ZSTD_flushStream(ZSTD_CStream* zcs, ZSTD_outBuffer* output)
 
 size_t ZSTD_endStream(ZSTD_CStream* zcs, ZSTD_outBuffer* output)
 {
-    ZSTD_inBuffer input = { NULL, 0, 0 };
+    ZSTD_inBuffer const nullInput = { NULL, 0, 0 };
+    int const stableInput = (zcs->appliedParams.inBufferMode == ZSTD_bm_stable);
+    ZSTD_inBuffer input = stableInput ? zcs->expectedInBuffer : nullInput;
     size_t const remainingToFlush = ZSTD_compressStream2(zcs, output, &input, ZSTD_e_end);
     FORWARD_IF_ERROR( remainingToFlush , "ZSTD_compressStream2 failed");
     if (zcs->appliedParams.nbWorkers > 0) return remainingToFlush;   /* minimal estimation */
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -5316,9 +5316,14 @@ size_t ZSTD_initCStream(ZSTD_CStream* zcs, int compressionLevel)
 
 static size_t ZSTD_nextInputSizeHint(const ZSTD_CCtx* cctx)
 {
-    size_t hintInSize = cctx->inBuffTarget - cctx->inBuffPos;
-    if (hintInSize==0) hintInSize = cctx->blockSize;
-    return hintInSize;
+    if (cctx->appliedParams.inBufferMode == ZSTD_bm_stable) {
+        return cctx->blockSize - cctx->stableIn_notConsumed;
+    }
+    assert(cctx->appliedParams.inBufferMode == ZSTD_bm_buffered);
+    {   size_t hintInSize = cctx->inBuffTarget - cctx->inBuffPos;
+        if (hintInSize==0) hintInSize = cctx->blockSize;
+        return hintInSize;
+    }
 }
 
 /** ZSTD_compressStream_generic():
@@ -5329,16 +5334,23 @@ static size_t ZSTD_compressStream_generic(ZSTD_CStream* zcs,
                                           ZSTD_inBuffer* input,
                                           ZSTD_EndDirective const flushMode)
 {
-    const char* const istart = (const char*)input->src;
+    const char* const istart = (assert(input != NULL), (const char*)input->src);
     const char* const iend = (istart != NULL) ? istart + input->size : istart;
     const char* ip = (istart != NULL) ? istart + input->pos : istart;
-    char* const ostart = (char*)output->dst;
+    char* const ostart = (assert(output != NULL), (char*)output->dst);
     char* const oend = (ostart != NULL) ? ostart + output->size : ostart;
     char* op = (ostart != NULL) ? ostart + output->pos : ostart;
     U32 someMoreWork = 1;
 
     /* check expectations */
-    DEBUGLOG(5, "ZSTD_compressStream_generic, flush=%i", (int)flushMode);
+    DEBUGLOG(5, "ZSTD_compressStream_generic, flush=%i, srcSize = %zu", (int)flushMode, input->size - input->pos);
+    assert(zcs != NULL);
+    if (zcs->appliedParams.inBufferMode == ZSTD_bm_stable) {
+        assert(input->pos >= zcs->stableIn_notConsumed);
+        input->pos -= zcs->stableIn_notConsumed;
+        ip -= zcs->stableIn_notConsumed;
+        zcs->stableIn_notConsumed = 0;
+    }
     if (zcs->appliedParams.inBufferMode == ZSTD_bm_buffered) {
         assert(zcs->inBuff != NULL);
         assert(zcs->inBuffSize > 0);
@@ -5347,8 +5359,10 @@ static size_t ZSTD_compressStream_generic(ZSTD_CStream* zcs,
         assert(zcs->outBuff !=  NULL);
         assert(zcs->outBuffSize > 0);
     }
-    assert(output->pos <= output->size);
+    if (input->src == NULL) assert(input->size == 0);
     assert(input->pos <= input->size);
+    if (output->dst == NULL) assert(output->size == 0);
+    assert(output->pos <= output->size);
     assert((U32)flushMode <= (U32)ZSTD_e_end);
 
     while (someMoreWork) {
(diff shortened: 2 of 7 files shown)


lib/compress/zstd_compress.c
```diff
@@ -6156,17 +6163,23 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size_t dstCapaci
 /*======   Finalize   ======*/
 
 /*! ZSTD_flushStream() :
+*   Note : not compatible with ZSTD_c_stableInBuffer
```
terrelln: Is this comment accurate?

lib/compress/zstd_compress.c
```diff
@@ -5645,7 +5644,15 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
...
     if (cctx->streamStage == zcss_init) {
-        FORWARD_IF_ERROR(ZSTD_CCtx_init_compressStream2(cctx, endOp, input->size), "CompressStream2 initialization failed");
+        size_t const inputSize = input->size - input->pos;  /* no obligation to start from pos==0 */
+        if ( (cctx->requestedParams.inBufferMode == ZSTD_bm_stable) /* input is presumed stable, across invocations */
+          && (endOp == ZSTD_e_continue)                             /* no flush requested, more input to come */
+          && (inputSize < ZSTD_BLOCKSIZE_MAX) ) {                   /* not even reached one block yet */
+            /* just wait, allows lazy adaptation of compression parameters */
+            cctx->expectedInBuffer = *input;
+            return ZSTD_FRAMEHEADERSIZE_MIN(cctx->requestedParams.format);  /* at least some header to produce */
+        }
```
terrelln: I don't think that this is correct. I get that the use case of streaming + stable input is somewhat contrived, and that it isn't super useful in practice. But...

Zstd guarantees to make forward progress. Either consume some input or produce some output. This breaks that contract, since `input` is not modified.

A loop constructed like this should not infinite loop:

```
while (!(input empty || output empty || finished))
  ZSTD_compressStream(cctx, &output, &input);
```

Instead, we should save the initial `input->pos`, and update `input->pos = input->size`.

Additionally, we shouldn't directly set `cctx->expectedInBuffer`, `ZSTD_setBufferExpectations(cctx, output, input)` should be used instead. The call on line 5656 can probably be moved up to handle both lazy & non-lazy adaptation, but only for the first call of lazy adaptation.

Lastly, since we are now modifying the input, we should make sure that we are calling `FORWARD_IF_ERROR(ZSTD_checkBufferStability(cctx, output, input, endOp), "invalid buffers");` in the lazy-adaptation mode.

terrelln (changes requested):

lib/compress/zstd_compress.c
Cyan4973: nope, that's a leftover, to be fixed

felixhandte:
I think we probably want to hold this until after the release, yeah?

Cyan4973:
> I think we probably want to hold this until after the release, yeah?

It's not critical for the release.

lib/compress/zstd_compress.c
Cyan4973: > Zstd guarantees to make forward progress. 

Forward progress was indeed one of my concerns when writing that code.

Some nuances here : there are limits to forward progress "guarantees" that `zstd` provides, even with current code.
As a perhaps obvious example, if a user continuously presents an empty input, then there is nothing to do. Same thing when a user continuously presents an output buffer with no more available space to flush into, then the streaming interface cannot progress. In both cases, repetitively calling `ZSTD_compressStream()` will just loop forever.

But it's okay, I guess we can classify above scenarios as user errors, that users should be able to debug, while the scenario you advocate for is different. What you mean is that, if it is possible to make some forward progresses, aka there is room in the output buffer and there is some input to ingest, then `zstd` should make some forward progress.
In this instance, instead of returning immediately without doing anything, `zstd` should fake progress by pretending having consumed the input, even though it hasn't, so that the suggested loop condition, which relies on forward progress, can exit.

I think that's fair, so that's what latest commit in this PR provides.

Cyan4973: > we shouldn't directly set `cctx->expectedInBuffer`, `ZSTD_setBufferExpectations(cctx, output, input)` should be used instead.

`ZSTD_setBufferExpectations()` relies on `cctx->appliedParams` which is _not initialized yet_ since at this stage `ZSTD_CCtx_init_compressStream2()` has not been invoked yet.

Cyan4973:
ping, this should be ready for review and merge now

Cyan4973:
ping 2

lib/compress/zstd_compress.c
```diff
@@ -5645,8 +5644,29 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
...
+            if (cctx->savedInPosPlusOne) {  /* not the first time */
+                /* check stable source guarantees */
+                assert(input->src == cctx->expectedInBuffer.src);
+                assert(input->pos == cctx->expectedInBuffer.size);
+            }
+            /* keep track of first position */
+            if (cctx->savedInPosPlusOne == 0) cctx->savedInPosPlusOne = input->pos + 1;
+            cctx->expectedInBuffer = *input;
+            /* pretend input was consumed, to give a sense forward progress */
+            input[0].pos = input[0].size;
```
terrelln: Why not `input->pos = input->size`? It seems strange to access it as an array.

lib/compress/zstd_compress.c
```diff
@@ -5645,8 +5644,29 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
 
     /* transparent initialization stage */
     if (cctx->streamStage == zcss_init) {
-        FORWARD_IF_ERROR(ZSTD_CCtx_init_compressStream2(cctx, endOp, input->size), "CompressStream2 initialization failed");
-        ZSTD_setBufferExpectations(cctx, output, input);    /* Set initial buffer expectations now that we've initialized */
+        size_t const inputSize = input->size - input->pos;  /* no obligation to start from pos==0 */
+        size_t const totalInputSize = (cctx->savedInPosPlusOne == 0) ? inputSize : input->size - (cctx->savedInPosPlusOne - 1);
+        if ( (cctx->requestedParams.inBufferMode == ZSTD_bm_stable) /* input is presumed stable, across invocations */
+          && (endOp == ZSTD_e_continue)                             /* no flush requested, more input to come */
+          && (totalInputSize < ZSTD_BLOCKSIZE_MAX) ) {              /* not even reached one block yet */
```
terrelln, line 5678: Once one of these conditions is false, do we need to reset `input->pos` back to `cctx->expectedInBuffer.pos`?

Does this code work with `ZSTD_compressStream2(cctx, output, input, ZSTD_e_flush)` & `ZSTD_compressStream2(cctx, output, input, ZSTD_e_end)`? It looks buggy to me, or I am not understanding what is going on?

Either way, could you please add a test case using `ZSTD_compressStream2()`?

Cyan4973: This code is run every time `ZSTD_compressStream2()` is invoked.
The specific part within the `if()` condition only triggers when `ZSTD_c_stableInBuffer` is set to `1`,
and there is a whole section within `zstreamtest` dealing with this scenario, starting line 830,
including `ZSTD_compressStream2(cctx, output, input, ZSTD_e_end)`.
The more relevant test starts at line 875 : it uses `ZSTD_compressStream()`, with `ZSTD_compressFlush()` and `ZSTD_compressEnd()`, all of which redirect to `ZSTD_compressStream2()`. It also partially load, before flushing.
These tests currently pass properly.

We can certainly consider adding more tests if these tests are not enough.
I'm trying to figure out which scenario isn't already tested and would fail with this code ?

terrelln: > I'm trying to figure out which scenario isn't already tested and would fail with this code ?

```c
ZSTD_inBuffer in = {src, 0, 100};
ZSTD_compressStream2(cctx, out, in, ZSTD_e_continue);
// in.pos = 100 now
in.size = 200;
ZSTD_compressStream2(cctx, out, in, ZSTD_e_end);
```

I believe in that scenario, the first 100 bytes of the input will be skipped.

> and there is a whole section within zstreamtest dealing with this scenario, starting line 830,
> including ZSTD_compressStream2(cctx, output, input, ZSTD_e_end).

But I don't believe that there are any tests with `ZSTD_e_continue` & changing the buffer size.

The only tests with `ZSTD_e_continue` and `ZSTD_e_flush` are tests to make sure that they are disallowed. And after this change, they should work, not fail.

And they don't execute this code at all, because the input size is large.

https://github.com/facebook/zstd/blob/70df5de1b2fc291a58c1b3199dde11b353ccad8f/tests/zstreamtest.c#L873-L889

Cyan4973:
OK, so having investigated the issue in more details, it's more complex than initially anticipated. And the necessity to "pretend" consuming all input data (even though it's not) doesn't help.

The scenario suggested by @terrelln does indeed fail, as predicted. This scenario was part of the test suite, but only the error codes were checked. A full roundtrip comparison exposes the missing bytes.

This one use case can be fixed. But then, other cases appear, and it becomes more complex.

What about : 
- ingest ~200 bytes
- flush that
- ingest 130 KB, continue mode

After the `flush()` operation, a frame must be started, and the initial 200 bytes of input must be effectively consumed (the interface was only "pretending" to have consumed it previously).
OK, but then what about the second `ingest` operation ?
Well, now it depends if we want the behavior to be equivalent to `buffered` mode, which I presume we want.
In this case, we should not ingest the full `130 KB`, as it would be equivalent to `flush` mode, resulting in a different output. Only full blocks, aka `128 KB`, can be ingested and compressed. The next 2 KB must not be ingested.
But on the other hand, the interface needs to "pretend" that the full input was consumed, as a way to allow the user to test this condition and not be stuck into an infinite loop.
So now, this situation needs to be continuously tracked. This goes beyond just the first block.

Now, this can be solved too. It's just that the topic goes beyond "late adaptation of compression parameters", it becomes "how to allow `continue` mode in combination with `ZSTD_stableInBuffer`", which is quite an extension of scope. This is acceptable if the complexity is low, but not if it impacts too much code in the core library.

So I'll have to spend a bit of time to figure out how to do that without impacting too much the library. And if not, then maybe just drop it. This was more a "late parameter adaptation demo" anyway.

Cyan4973:
Good news, the new version shipped this morning seems to fix the issues reported above.
As importantly, it does so without increasing complexity.
The new formulation looks actually simpler to read and explain, and is more general.

lib/compress/zstd_compress.c
```diff
@@ -5316,30 +5314,41 @@ size_t ZSTD_initCStream(ZSTD_CStream* zcs, int compressionLevel)
...
+    char* const oend = (ostart != NULL) ? ostart + output->size : ostart;
+    char* op = (ostart != NULL) ? ostart + output->pos : ostart;
     U32 someMoreWork = 1;
 
     /* check expectations */
-    DEBUGLOG(5, "ZSTD_compressStream_generic, flush=%u", (unsigned)flushMode);
+    DEBUGLOG(5, "ZSTD_compressStream_generic, flush=%i, srcSize = %zu", (int)flushMode, input->size - input->pos);
+    assert(zcs != NULL);
+    if (zcs->appliedParams.inBufferMode == ZSTD_bm_stable) {
+        assert(input->pos >= zcs->stableIn_notConsumed);
```
terrelln, line 5349: This should be a `RETURN_ERROR_IF()` right? Since we're validating user input, not library assumptions.

lib/compress/zstd_compress.c
```diff
@@ -5645,8 +5669,27 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
 
     /* transparent initialization stage */
     if (cctx->streamStage == zcss_init) {
-        FORWARD_IF_ERROR(ZSTD_CCtx_init_compressStream2(cctx, endOp, input->size), "CompressStream2 initialization failed");
```
terrelln, line 5650: As you point out in the comment `inputSize = input->size - input->pos`. So this is actually a bug, and we'll get the `pledgedSrcSize` wrong.

Can you add a test case for the scenario where `input->pos` starts off as non-zero? It should fail before the PR.

lib/compress/zstd_compress.c
```diff
@@ -5659,6 +5702,13 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
             ZSTDMT_updateCParams_whileCompressing(cctx->mtctx, &cctx->requestedParams);
             cctx->cParamsChanged = 0;
         }
+        if (cctx->stableIn_notConsumed) {
+            assert(cctx->appliedParams.inBufferMode == ZSTD_bm_stable);
+            /* some early data was skipped - make it available for consumption */
+            assert(input->pos >= cctx->stableIn_notConsumed);
```
terrelln, line 5710: This should also be a `RETURN_ERROR_IF()`.

lib/compress/zstd_compress.c
```diff
@@ -5645,8 +5669,27 @@ size_t ZSTD_compressStream2( ZSTD_CCtx* cctx,
...
-        ZSTD_setBufferExpectations(cctx, output, input);    /* Set initial buffer expectations now that we've initialized */
+        size_t const inputSize = input->size - input->pos;  /* no obligation to start from pos==0 */
+        size_t const totalInputSize = inputSize + cctx->stableIn_notConsumed;
+        if ( (cctx->requestedParams.inBufferMode == ZSTD_bm_stable) /* input is presumed stable, across invocations */
+          && (endOp == ZSTD_e_continue)                             /* no flush requested, more input to come */
+          && (totalInputSize < ZSTD_BLOCKSIZE_MAX) ) {              /* not even reached one block yet */
+            if (cctx->stableIn_notConsumed) {  /* not the first time */
+                /* check stable source guarantees */
+                assert(input->src == cctx->expectedInBuffer.src);
+                assert(input->pos == cctx->expectedInBuffer.size);
```
terrelln: These should be `RETURN_ERROR_IF()` since we're validating user input.

lib/compress/zstd_compress.c
```diff
@@ -6159,14 +6209,19 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size_t dstCapaci
  * @return : amount of data remaining to flush */
 size_t ZSTD_flushStream(ZSTD_CStream* zcs, ZSTD_outBuffer* output)
 {
-    ZSTD_inBuffer input = { NULL, 0, 0 };
+    ZSTD_inBuffer const nullInput = { NULL, 0, 0 };
+    int const stableInput = (zcs->appliedParams.inBufferMode == ZSTD_bm_stable);
+    ZSTD_inBuffer input = stableInput ? zcs->expectedInBuffer : nullInput;
+    input.size = input.pos; /* do not ingest more input during flush */
```
terrelln, line 6222: Can this be an `assert(input.size == input.pos)` instead?

I believe that we should always have consumed the entire input, is that correct?

lib/compress/zstd_compress.c
```diff
@@ -6159,14 +6209,19 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* const cctx, void* dst, size_t dstCapaci
...
     return ZSTD_compressStream2(zcs, output, &input, ZSTD_e_flush);
 }
 
 
 size_t ZSTD_endStream(ZSTD_CStream* zcs, ZSTD_outBuffer* output)
 {
-    ZSTD_inBuffer input = { NULL, 0, 0 };
+    ZSTD_inBuffer const nullInput = { NULL, 0, 0 };
+    int const stableInput = (zcs->appliedParams.inBufferMode == ZSTD_bm_stable);
+    ZSTD_inBuffer input = stableInput ? zcs->expectedInBuffer : nullInput;
```
terrelln: nit: Can you move this into a small helper function that is shared between `ZSTD_flushStream()` and `ZSTD_endStream()`?

tests/zstreamtest.c
```diff
@@ -870,26 +872,37 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+            outBuf.dst = (char*)(compressedBuffer)+cSize;
+            outBuf.size = ZSTD_compressBound(inputSize);
+            outBuf.pos = 0;
+            CHECK_Z( ZSTD_compressStream(cctx, &outBuf, &inBuf) );
+            inBuf.size = 200;
+            CHECK_Z( ZSTD_compressStream(cctx, &outBuf, &inBuf) );
+            CHECK_Z( ZSTD_flushStream(cctx, &outBuf) );
+            inBuf.size = inputSize;
+            CHECK_Z( ZSTD_compressStream(cctx, &outBuf, &inBuf) );
+            CHECK(ZSTD_endStream(cctx, &outBuf) != 0, "compression should be successful and fully flushed");
```
terrelln: nit: Can you also test `ZSTD_compressStream2()`, since it isn't going through the exact same code-path as `ZSTD_{compress,flush,end}Stream()` is?

terrelln (commented):
This PR looks good to me, just need to change some asserts to `RETURN_ERROR_IF()`, and a few more tests.

The approach seems right, and simplifies the usage of `ZSTD_c_stableInBuffer`.

lib/compress/zstd_compress.c
Cyan4973: This function `ZSTD_compressStream_generic()` is invoked after `ZSTD_checkBufferStability()`, 
which should have already validated that `ZSTD_c_stableInBuffer` condition is correctly respected, 
aka `input->pos` is stable and must not have been directly manipulated by the user.

So if we reach that `assert()` condition, it looks to me that this would be a programmatic error, not (just) user input.

lib/compress/zstd_compress.c
Cyan4973: We won't get `pledgedSrcSize` wrong because it's based on `totalInputSize`, which correctly counts from initial `pos`, wether it is 0 or non-0, by summing the `inputSize` across invocations.

Point taken for the `input->pos != 0` start condition test.

terrelln: I'm talking about the previous version of the code. It is buggy before your patch, and correct after.

You fixed the bug, so I would love to see the test case, so it doesn't accidentally get re-introduced.

lib/compress/zstd_compress.c
Cyan4973: Same idea, `ZSTD_checkBufferStability()` is invoked just before this paragraph, and should already take care of user input issues. So if we nonetheless hit this `assert()` at this position, seems like it's a programmatic error.

lib/compress/zstd_compress.c
Cyan4973: Yes, in this case, these checks are front line, they are effectively validating user input. They should return errors.

lib/compress/zstd_compress.c
Cyan4973: Even with `ZSTD_c_stableInBuffer`, it's possible to meet a situation where some input has not been consumed.

This happens if output buffer size is not large enough, in which case, the streaming state machine will be stopped during a flush stage, with some input not yet consumed.
After that `ZSTD_flushStream()` does not present `inBuffer` anymore, so it's not possible to update `inBuffer->pos`. `ZSTD_flushStream()` can only flush what's left in the output buffer, not ingest a bit more. 
This situation is a bit different from `ZSTD_compressStream2(cctx, outbuf, inbuf, ZSTD_e_flush)`, which allows ingesting a bit more input as part of the flush operation.

tests/zstreamtest.c
Cyan4973: done

lib/compress/zstd_compress.c

Cyan4973:
I believe final commit answers all remaining issues.

lib/common/error_private.c
```diff
@@ -38,6 +38,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(tableLog_tooLarge): return "tableLog requires too much memory : unsupported";
     case PREFIX(maxSymbolValue_tooLarge): return "Unsupported max Symbol Value : too large";
     case PREFIX(maxSymbolValue_tooSmall): return "Specified maxSymbolValue is too small";
+    case PREFIX(stabilityCondition_notRespected): return "pledged buffer stability condition is not respected";
```
terrelln, line 41: `ZSTD_checkBufferStability()` is currently using `srcBuffer_wrong` and `dstBuffer_wrong`. If you're going to add this code, can you please switch them over too?

terrelln (approved):
LGTM, just one minor nit about consistency of the new error code

lib/common/error_private.c
Cyan4973: Yes, good idea

PR #2975, opened 2022-01-05
Async write for decompression

yoniko:
Changes:
- Added `--[no-]asyncio` flag for CLI decompression.
- Replaced dstBuffer in decompression with a pool of write jobs.
- Added an ability to execute write jobs in a separate thread.
- Added an ability to wait (join) on all jobs in a thread pool (queue and running).
- Fixed bug in multiframe decompression in lz4.

Benchmarked total runtime (real time) of decompression on 3 machines, these are the runtime improvements in %:
```
flag                     --asyncio  --no-asyncio  improvement
input_file      machine
enwik8.zst      desktop      0.168         0.211    20.379147
                macbook      0.296         0.556    46.762590
                server       0.195         0.241    19.087137
silesia.tar.zst desktop      0.285         0.388    26.546392
                macbook      0.634         0.982    35.437882
                server       0.337         0.439    23.234624
```
Notes:
- user & system time / cpu utilization is mostly the same and might even be higher, but total time is reduced because we wait less time on IO
- Impact might be even bigger on bigger files as we also measure initialization here

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -96,9 +96,7 @@ static void* POOL_thread(void* opaque) {
             /* If the intended queue size was 0, signal after finishing job */
             ZSTD_pthread_mutex_lock(&ctx->queueMutex);
             ctx->numThreadsBusy--;
-            if (ctx->queueSize == 1) {
-                ZSTD_pthread_cond_signal(&ctx->queuePushCond);
-            }
+            ZSTD_pthread_cond_signal(&ctx->queuePushCond);
             ZSTD_pthread_mutex_unlock(&ctx->queueMutex);
         }
     }  /* for (;;) */
@@ -188,6 +186,17 @@ void POOL_free(POOL_ctx *ctx) {
     ZSTD_customFree(ctx->queue, ctx->customMem);
     ZSTD_customFree(ctx->threads, ctx->customMem);
     ZSTD_customFree(ctx, ctx->customMem);
+}
+
+/*! POOL_joinJobs() :
+ *  Waits for all queued jobs to finish executing.
+ */
+void POOL_joinJobs(POOL_ctx* ctx) {
+    ZSTD_pthread_mutex_lock(&ctx->queueMutex);
+    while(!ctx->queueEmpty || ctx->numThreadsBusy > 0) {
+        ZSTD_pthread_cond_wait(&ctx->queuePushCond, &ctx->queueMutex);
+    }
+    ZSTD_pthread_mutex_unlock(&ctx->queueMutex);
 }
 
 void ZSTD_freeThreadPool (ZSTD_threadPool* pool) {
@@ -329,6 +338,11 @@ void POOL_free(POOL_ctx* ctx) {
     assert(!ctx || ctx == &g_poolCtx);
     (void)ctx;
 }
+
+void POOL_joinJobs(POOL_ctx* ctx){
+    assert(!ctx || ctx == &g_poolCtx);
+    (void)ctx;
+};
 
 int POOL_resize(POOL_ctx* ctx, size_t numThreads) {
     (void)ctx; (void)numThreads;
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -342,7 +342,7 @@ void POOL_free(POOL_ctx* ctx) {
 void POOL_joinJobs(POOL_ctx* ctx){
     assert(!ctx || ctx == &g_poolCtx);
     (void)ctx;
-};
+}
 
 int POOL_resize(POOL_ctx* ctx, size_t numThreads) {
     (void)ctx; (void)numThreads;
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -34,6 +34,8 @@
 #include <limits.h>     /* INT_MAX */
 #include <signal.h>
 #include "timefn.h"     /* UTIL_getTime, UTIL_clockSpanMicro */
+#include "../lib/common/pool.h"
+#include "../lib/common/threading.h"
 
 #if defined (_MSC_VER)
 #  include <sys/stat.h>
@@ -325,6 +327,7 @@ struct FIO_prefs_s {
     /* IO preferences */
     U32 removeSrcFile;
     U32 overwrite;
+    U32 asyncIO;
 
     /* Computation resources preferences */
     unsigned memLimit;
@@ -395,6 +398,7 @@ FIO_prefs_t* FIO_createPreferences(void)
     ret->literalCompressionMode = ZSTD_ps_auto;
     ret->excludeCompressedFiles = 0;
     ret->allowBlockDevices = 0;
+    ret->asyncIO = 0;
     return ret;
 }
 
@@ -556,6 +560,10 @@ void FIO_setPatchFromMode(FIO_prefs_t* const prefs, int value)
 void FIO_setContentSize(FIO_prefs_t* const prefs, int value)
 {
     prefs->contentSize = value != 0;
+}
+
+void FIO_setAsyncIOFlag(FIO_prefs_t* const prefs, unsigned value) {
+    prefs->asyncIO = value;
 }
 
 /* FIO_ctx_t functions */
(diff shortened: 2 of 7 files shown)


terrelln:
> Added --[no-]asyncio flag for CLI decompression.

This should probably support compression too. And be overridden if the number of threads is explicitly set with `-T`.

yoniko:
That's a good point.
Taking it a bit further, perhaps it'd be better to just consolidate with `--single-threaded`?
The only problem there would be that the `--single-threaded` has an opposite default than `--asyncio` has at the moment (we wanted to have it off by default as its still experimental).

terrelln:
Yeah, I think we should combine the functionality with `--single-thread`. `--no-asyncio` and `--single-thread` should just be aliases.

We can default compression to `--asyncio` and decompression to `--no-asyncio`, until we're ready to enable it by default.

Currently, `--single-thread` only applies when `-T#` is set to `0` or unset. It also doesn't apply if `ZSTD_NBTHREADS` is not zero. I don't think we actually want `ZSTD_NBTHREADS` to override `--single-thread`, but that is open for discussion.

Point being, we should think carefully about the interactions between `--single-thread`, `--[no-]asyncio`, `-T#`, and `ZSTD_NBTHREADS`. And make sure that we have a consistent story that we're happy with. And that it can be extended to multi-threaded decompression, if we ever add that.

lib/common/pool.c
```diff
@@ -188,6 +186,17 @@ void POOL_free(POOL_ctx *ctx) {
...
+/*! POOL_joinJobs() :
+ *  Waits for all queued jobs to finish executing.
+ */
+void POOL_joinJobs(POOL_ctx* ctx) {
+    ZSTD_pthread_mutex_lock(&ctx->queueMutex);
+    while(!ctx->queueEmpty || ctx->numThreadsBusy > 0) {
+        ZSTD_pthread_cond_wait(&ctx->queuePushCond, &ctx->queueMutex);
+    }
+    ZSTD_pthread_mutex_unlock(&ctx->queueMutex);
+}
```
terrelln, line 200: This, and the rest of the pool changes, looks correct to me.

programs/fileio.c
```diff
@@ -1999,16 +2007,105 @@ int FIO_compressMultipleFilenames(FIO_ctx_t* const fCtx,
 /* **************************************************************************
  *  Decompression
  ***************************************************************************/
+#define DECOMPRESSION_MAX_WRITE_JOBS    (10)
+
+typedef struct {
+    POOL_ctx* writerPool;
+    ZSTD_pthread_mutex_t writeJobsMutex;
+    void* jobs[DECOMPRESSION_MAX_WRITE_JOBS];
+    volatile int availableWriteJobs;
```
terrelln: `availableWriteJobs` does not need to be volatile.

But can you add a newline before `writeJobsMutex` and after `availableWriteJobs`, and a comment explaining that the mutex protects `jobs` and `availableWriteJobs`?

programs/fileio.c
```diff
@@ -2147,6 +2244,82 @@ FIO_fwriteSparseEnd(const FIO_prefs_t* const prefs, FILE* file, unsigned storedS
     }   }
 }
 
+static void FIO_writePoolSetDstFile(write_pool_ctx_t *ctx, FILE* dstFile) {
+    assert(ctx!=NULL);
+    // We can change the dst file only if we have finished writing
```
terrelln: nit: These need to be `/* */` comments. `//` comments are C99.

programs/fileio.c
```diff
@@ -1999,16 +2007,105 @@ int FIO_compressMultipleFilenames(FIO_ctx_t* const fCtx,
...
+        EXM_THROW(101, "Allocation error : not enough memory");
+    job->buffer = buffer;
+    job->bufferSize = ZSTD_DStreamOutSize();
+    job->usedBufferSize = 0;
+    job->dstFile = NULL;
+    job->ctx = ctx;
+    return job;
+}
+
+static void FIO_writePoolCreateThreadPool(write_pool_ctx_t *ctx, const FIO_prefs_t *prefs) {
```
terrelln: nit: naming: Can you name the functions `FIO_WritePool_createThreadPool()`? Or just `WritePool_createThreadPool()` since they're all `static`. That makes it clear its the function `createThreadPool()` for the object/concept/group `WritePool`.

programs/fileio.c
```diff
@@ -1999,16 +2007,105 @@ int FIO_compressMultipleFilenames(FIO_ctx_t* const fCtx,
...
+        ctx->writerPool = POOL_create(1, DECOMPRESSION_MAX_WRITE_JOBS - 2);
+        if (!ctx->writerPool)
+            EXM_THROW(103, "Failed creating writer thread pool");
+#else
+        DISPLAYLEVEL(2, "Note : asyncio decompression is disabled (lack of multithreading support) \n");
+#endif
+    }
+}
+
+static write_pool_ctx_t* FIO_writePoolCreate(FIO_prefs_t* const prefs) {
```
terrelln: `WritePool_create()`

programs/fileio.c
```diff
@@ -2147,6 +2244,82 @@ FIO_fwriteSparseEnd(const FIO_prefs_t* const prefs, FILE* file, unsigned storedS
     }   }
 }
 
+static void FIO_writePoolSetDstFile(write_pool_ctx_t *ctx, FILE* dstFile) {
```
terrelln: `WritePool_setDstFile()`, etc.

Cyan4973:
> Currently, `--single-thread` only applies when `-T#` is set to `0` or unset. It also doesn't apply if `ZSTD_NBTHREADS` is not zero.

`ZSTD_NBTHREADS` is only meant to be a "default" value for `-T#`, therefore it should be overwritten by any explicit command, including `-T#` or `--single-thread`.

> Currently, `--single-thread` only applies when `-T#` is set to `0` or unset.

This seems strange to me, likely an overlook.
I believe it should supersede `-T#`,
and there is probably a point to output a warning message in case `--single-thread` is combined with `-T2`.

Another candidate policy could be "last one wins", meaning that `--single-thread -T2` actually means `-T2`, and `-T2 --single-thread` actually means `--single-thread`. Here also, a warning message would be welcomed to underline the discrepancy.

Lastly, a last policy could be to refuse to work in presence of a discrepancy.

I tend to prefer the first policy (`--single-thread` has priority), but this is certainly opened to discussion.

programs/fileio.c
```diff
@@ -2147,6 +2244,82 @@ FIO_fwriteSparseEnd(const FIO_prefs_t* const prefs, FILE* file, unsigned storedS
...
+
+static void FIO_writePoolQueueWriteJob(write_job_t *job) {
+    write_pool_ctx_t* ctx = job->ctx;
+    if(ctx->writerPool)
+        POOL_add(ctx->writerPool, FIO_WritePoolWriteJobExecute, job);
+    else
+        FIO_WritePoolWriteJobExecute(job);
+}
+
+static void FIO_writePoolQueueWriteJobAndGetNextAvailable(write_job_t **job) {
```
terrelln: For all the functions you intend to be "public", e.g. not the helper functions only used by the other `WritePool` functions, could you add documentation? You could also add docs for the helpers, but that is less necessary.

terrelln (commented):
Overall this is looking great!

Could you also add some benchmark figures to the PR summary?

terrelln:
> I tend to prefer the first policy (--single-thread has priority), but this is certainly opened to discussion.

I agree that `--single-thread` should certainly have priority over `ZSTD_NBTHREADS`. I could see it going either way for `-T#`, and don't have a strong preference, as long as we are consistent.

If we decide that `--single-thread` should override `-T#`, should `--no-asyncio` also override `-T#`?

Cyan4973:
> If we decide that `--single-thread` should override `-T#`, should `--no-asyncio` also override `-T#`?

If we decide that `--single-thread` == `--no-asyncio`, then yes.

There is a catch though.
`--single-thread` produces a compressed output different from `-T1`.
if `--no-asyncio` == `--single-thread`, then it will have the same impact, 
which has consequences from a reproducibility perspective.
Yet, while it's reasonable to explain that `--single-thread` impacts the compression process, hence impacts the compressed output, it's more difficult to use the same argument for `--no-asyncio`, which should only impact I/O.

Thoughts ?

Cyan4973:
Could you add a test featuring multi-frames files ?
It's currently a supported scenario, and it's possibly one of the difficult scenarios for `--async-io`.

Bonus point for files with multiple heterogeneous frames (also currently supported).

yoniko:
I've actually tested this manually, but it makes sense to add an automated test. Will do it.

yoniko:
Updating here that we've spoken about `--single-thread` and `--[no-]asyncio` and decided to leave the two uncoupled as they have two different underlining meanings:

- `-T#` and `--single-thread` control the threading mode for compression which directly affects the result of compression itself. Thus the `--single-thread` flag's importance is in compressing in a single-threaded mode and not about having the cli actually run in a single thread.
- `--[no-]asyncio` is about having asyncio interfaces for (de)compression. These asyncio interfaces could work with a single threaded compression as well (if available).

tests/playTests.sh
```diff
@@ -1568,6 +1568,44 @@ elif [ "$longCSize19wlog23" -gt "$optCSize19wlog23" ]; then
...
+fi
+if [ $LZMAMODE -eq 1 ]; then
+  testAsyncIO lzma
+  addTwoFrames lzma
+fi
+if [ $LZ4MODE -eq 1 ]; then
+  testAsyncIO lz4
+  addTwoFrames lz4
+fi
+cat tmp_uncompressed | shasum > tmp2
```
Cyan4973: We use `$MD5SUM` to designate the checksum program,
which generally translates into `md5sum`, but not always (freeBSD uses `gmd5sum`,  macos uses `md5 -r`, etc.).

yoniko: Thank you, pushed a fix.

programs/fileio.c
yoniko: Fixed

programs/fileio.c
yoniko: Fixed

programs/fileio.c
yoniko: Fixed

programs/fileio.c
yoniko: Fixed

Cyan4973:
I like the new code path.
I believe it fits well in current design, and even clarifies the processing pattern.
Evidence is that it can be generalized with minimal effort to all supported decoders.

The last issue is about the `meson` build script,
which I suspect is related to meson's contributors insistence of building the CLI against the shared library,
even though it is a scenario we explicitly do not support.
This PR makes the situation a bit worse in this regard, since it makes the CLI depend even more on non-public components of `libzstd`, which is fine as long as CLI is linked to the static library, but not the dynamic one.

It may seem a unrelated secondary topic, but we'll nonetheless need a solution to this situation before the merge.

yoniko:
`meson` build fixed and Travis finishes successfully.
I did need to include some more files from `lib/common` in the building of the binary.

PR #2987, opened 2022-01-10
Prepare v1.5.2

felixhandte:


felixhandte:
Yep, `man programs/zstd.1` looks fine on my system. So I think the changes are all immaterial.

Depending what we end up merging, we may want to add more to the `CHANGELOG`. I really only included the headliners.

Cyan4973:
> Yep, `man programs/zstd.1` looks fine on my system. So I think the changes are all immaterial.
> 
> Depending what we end up merging, we may want to add more to the `CHANGELOG`. I really only included the headliners.

I'm fine with a `CHANGELOG` which only contains "material" items featuring at least one user-perceptible impact.

Cyan4973 (approved):

PR #2989, opened 2022-01-10
Reduce -T1 compression memory usage

embg:
Instead of allocating temporary output buffers for the single worker, why not write directly into the final output buffer? In addition to saving memory on at least two temporary buffers, it also saves a `memcpy`, although with preliminary benchmarks I haven't been able to demonstrate a performance impact. In my preliminary benchmark, the savings from this PR would be 2 MB out of 13 MB peak memory usage on silesia.

Note: this requires that the output buffer is larger than `ZSTD_compressBound(mtctx->targetSectionSize)`. It could be exposed as an option for the CLI, which could supply a sufficiently large output buffer. CI tests are currently failing on the PR because the output buffer supplied by the CLI right now is too small.


embg:
After trying to pass asan for a few hours, realized there are some major issues with this approach. For example, as `output->pos` increases, the `dstBuffer` for workers becomes arbitrarily small, even if the full output buffer is large. All the solutions I can think of to address the issues add too much complexity (given that there is no perf win). Closing PR.

Cyan4973:
note : 
consider enabling all our supported CI on your github account,
this will allow you to test the same suite as here directly from your local fork
without a need to start a PR to witness the tests' outcome.

PR #2998, opened 2022-01-13
Avoid updating timestamps when the destination is stdout

### LINKED ISSUE #2997: decompressing files tries to modify a /*stdout*\ file
**Describe the bug**
After upgrading to zstd-1.5.1, decompressing files with `zstd -dfc` shows it trying to modify a `/*stdout*\` file.  zstd-1.5.0 does not do this.

**To Reproduce**
1. git clone zstd repo
2. run `MOREFLAGS=-static make` to build it
3. run `strace -eutimensat ./programs/zstd -kdfc deutex-5.2.2.tar.zst >/dev/null`
```
utimensat(AT_FDCWD, "/*stdout*\\", [UTIME_NOW, {tv_sec=1609297863, tv_nsec=0} /* 2020-12-29T22:11:03-0500 */], 0) = -1 ENOENT (No such file or directory)
```

Bisecting the v1.5.0 & v1.5.1 tags says commit 9cd6c1ff4d56fc74a2cbdfd9bcc82a64e0fe4bb7 is to blame.

floppym:
Fixes: 9cd6c1ff4d56fc74a2cbdfd9bcc82a64e0fe4bb7
Fixes: https://github.com/facebook/zstd/issues/2997

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1679,6 +1679,7 @@ static int FIO_compressFilename_dstFile(FIO_ctx_t* const fCtx,
     if (ress.dstFile == NULL) {
         int dstFilePermissions = DEFAULT_FILE_PERMISSIONS;
         if ( strcmp (srcFileName, stdinmark)
+          && strcmp (dstFileName, stdoutmark)
           && UTIL_stat(srcFileName, &statbuf)
           && UTIL_isRegularFileStat(&statbuf) ) {
             dstFilePermissions = statbuf.st_mode;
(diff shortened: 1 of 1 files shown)


terrelln:
The fix looks reasonable to me, but I'll let @felixhandte review.

We should also add a test case for this.

@Cyan4973, @felixhandte: Does this need to be in v1.5.2?

floppym:
> We should also add a test case for this.

I'm not sure how to go about that, so help would be appreciated.

Cyan4973:
> 
> @Cyan4973, @felixhandte: Does this need to be in v1.5.2?

Yes, I would be favorable to that.

terrelln (approved):

terrelln (approved):

PR #2999, opened 2022-01-14
Update CI documentation

embg:
Travis CI has [ended](https://news.ycombinator.com/item?id=25338983) free usage for OSS. Additionally, the instructions for setting up Travis (and possibly the other services) are outdated. Since we have moved almost all of our CI onto Github Actions (which don't require any setup for forks), I think we should only require Github Actions to pass before submitting PRs against the main repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,8 +68,8 @@ Our contribution process works in three main stages:
             ```
 2. Code Review and CI tests
     * Ensure CI tests pass:
-        * Before sharing anything to the community, make sure that all CI tests pass on your local fork.
-        See our section on setting up your CI environment for more information on how to do this.
+        * Before sharing anything to the community, create a pull request in your own fork against the dev branch
+        and make sure that all GitHub Actions CI tests pass. See the Continuous Integration section below for more information.
         * Ensure that static analysis passes on your development machine. See the Static Analysis section
         below to see how to do this.
     * Create a pull request:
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks fine for the general message,
 
but it seems we are missing an opportunity to educate our users.
The previous version of `CONTRIBUTING.md` was providing detailed instructions about how to subscribe to external CI services, in order to increase test coverage. In the proposed update, they are simply gone, with no replacement.

I could imagine that some of these instructions have become obsolete, and some CI may no longer be possible (TravisCI notably). But they are not all useless. We shouldn't go as far as discouraging users from adding other CI beyond Github Actions.

To counter the risk of obsolescence of local doc, try to find a documentation from the CI provider itself, and link to it.
The link might also die over time, though it's an acceptable limit to what it's reasonable to provide.

CONTRIBUTING.md
```diff
@@ -339,57 +357,6 @@ counter `L1-dcache-load-misses`
 
 TODO
 
-
-## Setting up continuous integration (CI) on your fork
```
Cyan4973, line 343: I think this passage is still valid, and should have a replacement.

I understand that it's less important, since Github Action is the main CI system now, but we should use this opportunity to educate our users (or future contributors, including from Meta).

What I would expect in this part : 
- A list of CI used to validate a PR.
- For each one : 
  - why they exist (what do they cover that other CI do not)
  - where is the corresponding test script (within the repository)
  - how to subscribe (link mostly, keep that on the provide side, since it tends to change over time)

I'm available for additional background information should you wish to.

embg: Force-pushed for the first and third bullet, will add second bullet shortly.

Cyan4973 (approved):

Issue #2232, opened 2020-07-08
Support ZSTD_estimateCCtxSize for multi-threaded compression


marxin:
It would be handy to have a memory estimation also in the multi-threaded compression mode:
https://github.com/facebook/zstd/blob/2cdd33ae16915f738613143496e963c5a1be353d/lib/zstd.h#L1290

PR #3000, opened 2022-01-16
Documentation and minor refactor to clarify MT memory management

embg:
Clarifies some misunderstandings I had while working on https://github.com/facebook/zstd/issues/2232. In particular:
* `ZSTDMT_initCStream_internal` can reuse a `ZSTDMT_CCtx` from a previous compression, and memory usage depends on the previous compression's parameters.
* The buffer pool size upper bound only applies to `mtctx->bufPool`. For `mtctx->seqPool`, the true bound is lower.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -102,9 +102,8 @@ typedef struct ZSTDMT_bufferPool_s {
     buffer_t bTable[1];   /* variable size */
 } ZSTDMT_bufferPool;
 
-static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(unsigned nbWorkers, ZSTD_customMem cMem)
+static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(unsigned maxNbBuffers, ZSTD_customMem cMem)
 {
-    unsigned const maxNbBuffers = 2*nbWorkers + 3;
     ZSTDMT_bufferPool* const bufPool = (ZSTDMT_bufferPool*)ZSTD_customCalloc(
         sizeof(ZSTDMT_bufferPool) + (maxNbBuffers-1) * sizeof(buffer_t), cMem);
     if (bufPool==NULL) return NULL;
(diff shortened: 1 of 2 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -102,9 +102,8 @@ typedef struct ZSTDMT_bufferPool_s {
     buffer_t bTable[1];   /* variable size */
 } ZSTDMT_bufferPool;
 
-static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(unsigned nbWorkers, ZSTD_customMem cMem)
+static ZSTDMT_bufferPool* ZSTDMT_createBufferPool(ZSTD_customMem cMem, unsigned maxNbBuffers)
```
Cyan4973: why changing the order of arguments ?

Cyan4973:
The additional comments are welcome, and correct.

I just have one question about an included modification of parameters order.

lib/compress/zstdmt_compress.c
embg: Originally I just added maxNbBuffers to the end, and I only removed nbWorkers later after the compiler warned me it was unused. I updated the PR to keep the original order of arguments.

Cyan4973 (approved):

Cyan4973:
Look good to me !

lib/compress/zstdmt_compress.c
```diff
@@ -263,6 +261,16 @@ static void ZSTDMT_releaseBuffer(ZSTDMT_bufferPool* bufPool, buffer_t buf)
     ZSTD_customFree(buf.start, bufPool->cMem);
 }
 
+/* We need 2 output buffers per worker since each dstBuff must be flushed after it is released.
+ * The 3 additional buffers are as follows:
+ *   1 buffer for input loading
+ *   1 buffer for "next input" when submitting current one
+ *   1 buffer stuck in queue */
+#define BUF_POOL_MAX_NB_BUFFERS(nbWorkers) 2*nbWorkers + 3
```
felixhandte, line 269: Hmmm. This is potentially dangerous. Macro expressions and parameters should always be carefully wrapped (either with parentheses or a `do { ... } while (0)` as appropriate). A safe way to write this would be `(2 * (nbWorkers) + 3)`. Without the parentheses, using this macro like `BUF_POOL_MAX_NB_BUFFERS(foo) * 2` or `BUF_POOL_MAX_NB_BUFFERS(foo + 1)` will produce incorrect results. 

You're only using it directly as an argument to functions, with a direct argument, so this PR avoids hitting these landmines. But it may still be worth fixing?

embg: Good catch, will put up a fix

Cyan4973: We don't need `do {...} while(0)` in this case, as it's just a mathematical expression and we need it to return a value,
but definitely, enclosing the variable inside brackets `(nbWorkers)` is a good code-hardening practice.

Issue #2924, opened 2021-12-13
single_files_libs/combine.sh very slow, ~3 minutes.


averms:
**Describe the bug**
The amalgamation script is very slow, running it through Bash takes around 3 minutes for `zstd-in.c`.

**To Reproduce**
1. `cd build/single_file_libs && time create_single_file_library.sh`

**Expected behavior**
Faster amalgamation.

**Solution**

I wrote an [alternative in Python](https://github.com/averms/ccombine). It is 135x faster so it takes around 1 second. If you don't  want to take an (optional) build dependency on Python, perhaps the shell script could be refactored somehow. There is a comment in the source suggesting that speed was sacrificed for POSIX compliance but the script isn't compliant with POSIX because it uses `local`.

Cyan4973:
Moving to Python is something we consider, especially for such a remarkable speed improvement.
The plan is to keep both versions (at least in short term), so that environments without python dependency can still run the shell script as backup.

cwoffenden:
I also rewrote this in Python3 (original author of the script here) but due to the required platform support (at the time) this shell script was the option that worked everywhere. The original was version in Bash, and was much quicker, but the CI for older platforms meant plain ol' `sh` or nothing.

Cyan4973:
I believe we should keep the `sh` version around, precisely because it can be used in more systems with less dependencies.
But it doesn't prevent the existence of faster variants alongside the `sh` one, that could even become default if it feels reasonable in term of dependencies (and `python` is a reasonable dependency).

cwoffenden:
Fixed in #3005

PR #3005, opened 2022-01-17
Use faster Python script to amalgamate

cwoffenden:
As mentioned in #2924 the `sh` amalgamation script is extremely, unusably slow. This faster Python version completes almost instantly. TL;DR where it used to take over 6.5 _minutes_ for the full library it now takes a 1/3 of a second (on this old, old Mac Pro). Testing on an M1 Mac it takes **0.095 seconds** to run the same.

When developing the original single file decoder a Bash script was used, which completed in reasonable time, but didn't work with Zstd's CI on older OS/machine combos. This Python version was the working replacement but never supported all of the command line flags (I'll add them in the next days if there's interest, before calling this done).

Some numbers before:
```
carl@Betty single_file_libs % time ./create_single_file_decoder.sh
./create_single_file_decoder.sh  115.85s user 138.85s system 150% cpu 2:49.02 total
carl@Betty single_file_libs % time ./create_single_file_library.sh
./create_single_file_library.sh  279.35s user 322.86s system 154% cpu 6:29.05 total
```
And after:
```
carl@Betty single_file_libs % time ./create_single_file_decoder.sh
./create_single_file_decoder.sh  0.12s user 0.04s system 95% cpu 0.174 total
carl@Betty single_file_libs % time ./create_single_file_library.sh
./create_single_file_library.sh  0.20s user 0.06s system 97% cpu 0.270 total
```
The creation scripts first test that Python 3.8 or higher is installed, and if not the original shell script is run instead.

If there's interest in taking this I'll spend a few hours to make sure all the flags are supported and tested. As it stands it doesn't support excluding, which isn't used by Zstd anyway (it is in other projects).

### DIFF
--- a/build/single_file_libs/combine.py
+++ b/build/single_file_libs/combine.py
@@ -4,25 +4,30 @@
 # 
 # Author: Carl Woffenden, Numfum GmbH (this script is released under a CC0 license/Public Domain)
 
-import argparse, io, os, re, sys
+import argparse, re, sys
 
 from pathlib import Path
+from typing import Any, List, Optional, Pattern, Set, TextIO
 
 # Set of file roots when searching (equivalent to -I paths for the compiler).
-roots = set()
+roots: Set[Path] = set()
 
 # Set of (canonical) file Path objects to exclude from inlining (and not only
 # exclude but to add a compiler error directive when they're encountered).
-excludes = set()
+excludes: Set[Path] = set()
 
 # Set of (canonical) file Path objects to keep as include directives.
-keeps = set()
+keeps: Set[Path] = set()
 
-# Set of file Path objects previously inlined.
-found = set()
+# Whether to keep the #pragma once directives (unlikely, since this will result
+# in a warning, but the option is there).
+keep_pragma: bool = False
 
-# Destination file TextIOBase object (or stdout if no output file was supplied).
-destn = None
+# Destination file object (or stdout if no output file was supplied).
+destn:TextIO = sys.stdout
+
+# Set of file Path objects previously inlined (and to ignore if reencountering).
+found: Set[Path] = set()
 
 # Compiled regex Patern to handle the following type of file includes:
 # 
@@ -44,7 +49,7 @@ destn = None
 # file (whether it's a valid comment or not, since anything after the quoted
 # string is ignored)
 # 
-include_regex = re.compile(r'^\s*#\s*include\s*"(.+?)"')
+include_regex: Pattern = re.compile(r'^\s*#\s*include\s*"(.+?)"')
 
 # Simple tests to prove include_regex's cases.
 # 
@@ -73,7 +78,7 @@ def test_match_include() -> bool:
 # 
 # Ignoring commented versions, same as include_regex.
 # 
-pragma_regex = re.compile(r'^\s*#\s*pragma\s*once\s*')
+pragma_regex: Pattern = re.compile(r'^\s*#\s*pragma\s*once\s*')
 
 # Simple tests to prove pragma_regex's cases.
 # 
@@ -93,8 +98,11 @@ def text_match_pragma() -> bool:
 # for a match, followed by the list of 'root' paths, returning a valid Path in
 # canonical form. If no match is found None is returned.
 # 
-def resolve_include(parent: Path, file: str) -> Path:
-	found = parent.joinpath(file).resolve();
+def resolve_include(file: str, parent: Optional[Path] = None) -> Optional[Path]:
+	if (parent):
+		found = parent.joinpath(file).resolve();
+	else:
+		found = Path(file)
 	if (found.is_file()):
 		return found
 	for root in roots:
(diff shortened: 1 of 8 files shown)


Cyan4973:
Thanks @cwoffenden :+1:  ,
I'm sure there will be quite some interest in this faster version :wink:

cwoffenden:
This is done and tidied. I limited running the Python version to 3.8 (and higher) though it _may_ work on earlier versions (testing was only done on 3.8). The Python is as Python-like as I know how (I'm not a Python guy), has the type annotations and has been passed through `mypy` static type checking.

I changed the amalgamation scripts to use the `-x` flag to exclude `zstd_legacy.h` since both the decoder and full libraries build without `ZSTD_LEGACY_SUPPORT`, and this lets us remove 50kB from the resulting generated C (and if the macro is ever enabled an `#error` directive is inserted to show why it fails). I also tested when using the `-k` flag to not inline `zstd.h` to keep the library as separate header and implementation, and that also works well.

And it's fast, even on my 2013 Mac Pro:
```
carl@Betty single_file_libs % time ./combine.py -r ../../lib -x legacy/zstd_legacy.h -o zstddeclib.c zstddeclib-in.c
./combine.py -r ../../lib -x legacy/zstd_legacy.h -o zstddeclib.c   0.11s user 0.03s system 96% cpu 0.145 total
```
Consistently less than 0.2s vs nearly 3m previously (for the decoder lib), so over a thousand times faster.

cwoffenden:
Added bonus: the Python script also works on Windows. In theory the shell script did, with the Git shell support, for example, but would take hours. It takes longer than on macOS on the same machine, completing in 0.6s:
```
Measure-Command {python3 combine.py -r ../../lib -x legacy/zstd_legacy.h -o zstddeclib.c zstddeclib-in.c}
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 594
```

cwoffenden:
This has had further testing and edge cases fixed. I can now call it done.

build/single_file_libs/combine.py
```diff
@@ -0,0 +1,234 @@
...
+
+# Set of (canonical) file Path objects to keep as include directives.
+keeps: Set[Path] = set()
+
+# Whether to keep the #pragma once directives (unlikely, since this will result
+# in a warning, but the option is there).
+keep_pragma: bool = False
+
+# Destination file object (or stdout if no output file was supplied).
+destn:TextIO = sys.stdout
```
gyscos:
```suggestion
destn: TextIO = sys.stdout
```

build/single_file_libs/combine.py
```diff
@@ -0,0 +1,234 @@
...
+# We don't try to catch errors since the compiler will do this (and the code is
+# expected to be valid before processing) and we don't care what follows the
+# file (whether it's a valid comment or not, since anything after the quoted
+# string is ignored)
+# 
+include_regex: Pattern = re.compile(r'^\s*#\s*include\s*"(.+?)"')
+
+# Simple tests to prove include_regex's cases.
+# 
+def test_match_include() -> bool:
```
gyscos, line 80: It's often good practice to have all constants at the top of the file, and then all the functions, rather than intertwining these.

Might be worth moving test functions to a separate file as well.

build/single_file_libs/combine.py
```diff
@@ -0,0 +1,234 @@
...
+resolve_excluded_files(args.keep,    keeps,    args.input.parent)
+keep_pragma = args.pragma;
+
+# Then recursively process the input file
+try:
+    if (args.output):
+        destn = args.output
+    add_file(args.input)
+finally:
+    if (not destn):
```
gyscos: In which case would you expect `destn` to evaluate to false, but still be able to call `close` on it?

build/single_file_libs/combine.py
cwoffenden: I've never been happy with where this was but since it was test code I wanted to keep it near what it tests. I've moved it further down rather than moving it out (I didn't really want a single-file generator to but built of multiple files!).

build/single_file_libs/combine.py
cwoffenden: Ah, good catch. This got badly refactored from `if (destn is not None):`.

Cyan4973:
minor question :

It seems the new recipe needs to explicitly exclude `zstd_legacy.h`,
while that wasn't required before.
Any explanation why ?

cwoffenden:
The input file always stripped out certain features, primarily to reduce the size when this was just the decompressor alone, e.g.:
```
#define ZSTD_LEGACY_SUPPORT 0
#define ZSTD_STRIP_ERROR_STRINGS
#define ZSTD_TRACE 0
```
In the original version of tool there wasn't the `-x` option, so all files were inlined, used or not.

I can add it back in, if you'd prefer? It removed 50k-ish from the generated source, which the compiler skips over because those parts are currently if'd out.

Cyan4973:
I think it's a pretty good reason.
I believe it just needs to be documented, as otherwise people will just not understand why it's there, and may miss this optimization in some future update.

cwoffenden:
I made the `#error` inserted into the generated code more detailed, and also added docs for the `ZSTD_LEGACY_SUPPORT` flag. I tidied the instructions and copyright year whilst I was in there.

Cyan4973 (approved):

Cyan4973:
Thanks @cwoffenden !

PR #2970, opened 2022-01-04
Improved huffman depth optimization

binhdvo:
Not ready for merge yet; this has been tested across samples from silesia and canterbury.  There are small gains in compression ratio as expected, however there are also losses, so it is not a consistent benefit.  The accuracy of the huffman size estimation needs to be improved to ensure that neighboring depths are only chosen when beneficial.  Most likely primarily with the estimation of the FSE header within the compression of the huffman header.

PR #3019, opened 2022-01-21
More traces to improved debugging of literals compression

Cyan4973:
While investigating the sequence compression API,
it emerged a weird corner case, featuring poorer than expected performance.
This investigation required additional traces for proper analysis, in the literals compression process, and notably the Huffman part.

-------------

_edit_ : the below section is no longer valid, only additional traces and some refactoring remain, no functional change.

This is the topic of the 1st commit.

The issue was an abnormally bad compression ratio for a specific histogram of literals (`[4, 257, 257, 1, 1, .... , 1]`).
The solution was to be less brutal in the automatic resizing of the tree depth.
For the specific example investigated, compressed output became more compact, from 601 to 441 bytes, which is rather significant.

While the impact is large for this special case,
in more generic situations, I expect this update to produce no change, or very small benefits.
I've tested it on several public corpus samples at several compression level settings, and it followed the expected outcome : very small improvement, notably for low levels, or no impact at all, generally for high levels.
The same outcome is reflected in regression test results.

### DIFF
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -745,10 +745,10 @@ HUF_buildCTable_wksp(HUF_CElt* CTable, const unsigned* count, U32 maxSymbolValue
     nonNullRank = HUF_buildTree(huffNode, maxSymbolValue);
 
     /* determine and enforce maxTableLog */
-    /* Loosen target when maxNbBits is within limits.
+    /* Loosen target when maxNbBits is already within limits.
      * A harsh rebalancing can be bad for compression ratio
      * while a mild one tends to be better */
-    while (maxNbBits < HUF_TABLELOG_MAX) {
+    while (maxNbBits < HUF_TABLELOG_DEFAULT) {
         size_t const nbSTL = HUF_nbSymbolsTooLarge(huffNode, maxSymbolValue, maxNbBits);
         #define HUF_NB_NODES_TO_FIX_MAX 32
         if (nbSTL < HUF_NB_NODES_TO_FIX_MAX) /* heuristic */
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -272,7 +272,7 @@ size_t HUF_readCTable (HUF_CElt* CTable, unsigned* maxSymbolValuePtr, const void
 
 U32 HUF_getNbBitsFromCTable(HUF_CElt const* CTable, U32 symbolValue)
 {
-    const HUF_CElt* ct = CTable + 1;
+    const HUF_CElt* const ct = CTable + 1;
     assert(symbolValue <= HUF_SYMBOLVALUE_MAX);
     return (U32)HUF_getNbBits(ct[symbolValue]);
 }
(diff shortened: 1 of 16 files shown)


lib/compress/huf_compress.c
```diff
@@ -677,28 +705,56 @@ static void HUF_buildCTableFromTree(HUF_CElt* CTable, nodeElt const* huffNode, i
...
+    /* Loosen target when maxNbBits is already within limits.
+     * A harsh rebalancing can be bad for compression ratio
+     * while a mild one tends to be better */
+    while (maxNbBits < HUF_TABLELOG_DEFAULT) {
+        size_t const nbSTL = HUF_nbSymbolsTooLarge(huffNode, maxSymbolValue, maxNbBits);
+        #define HUF_NB_NODES_TO_FIX_MAX 32
+        if (nbSTL < HUF_NB_NODES_TO_FIX_MAX) /* heuristic */
+            break;
+        maxNbBits++;
+    }
```
terrelln: This is breaking the fuzzer, because it expects that if it passes the `maxNbBits` to `HUF_buildCTable_wksp()`, it won't build a table with a larger table log. And it picks a random tableLog between the minimum possible and 12.

I guess the table log gets shrunk too far by `HUF_optimalTableLog()`. Would it make sense to move the call to `HUF_optimalTableLog()` into this function? Or replace `HUF_optimalTableLog()` with the logic here, and shrink where applicable, instead of growing?

I think logically, we should only be modifying the table log in one place anyway. So merging this wit h`HUF_optimalTableLog()` makes sense.

Cyan4973: Thought about that too.
The issue is that `HUF_optimalTableLog()` doesn't have enough information to make such choice. `HUF_optimalTableLog()` only receives `srcSize` and `maxSymbolValue` to make a decision.
In this PR, the proposed refinement can happen thanks to access to `huffNode`, which is a sorted list of symbols with already decided `nbBits` (no maximum enforced yet). This information is not available anywhere before, so it can't be used before that point, and `HUF_optimalTableLog()` is run before even entering this function.

There are multiple options starting from there, with more or less refactoring cost.

One that comes to mind is that the test is simply no longer appropriate.
What it tests is that, if the huffman compressor is instructed to use a low (yet achievable) `tableLog`, then it will always respect the limit, no matter the cost to compression ratio.
However, if we accept the premisses that the huffman codec within `zstd` is no longer "generic" but designed solely for `zstd`, then some of its existing properties are no longer relevant. In this instance, `HUF_buildCTable_wksp()` does no longer guarantee to respect the maximum `tableLog` provided, though it does guarantee to remain below `tableLog==11`, which is a limit specific to `zstd` proper. The test could simply adjust to the new contract.

That being said, one issue I see with this design is that it's no longer possible to "force" `HUF_buildCTable_wksp()` to use an explicit low `tableLog`, because the function is free to decide on its own that it's too low, and fix it to use its own limit. Yet, even the new heuristic, while better informed than `HUF_optimalTableLog()`, is still just an heuristic, and could be wrong. A more thorough analysis might actually discover that it's indeed better to use a smaller table. Or it might want to use a smaller table for other reasons while accepting a compression ratio cost (speed of table construction on the decoding side for example).

This could be fixed by adding _yet another_ parameter, which states if the provided `tableLog` limit is "hard" or "soft". I generally avoid adding parameters, because it makes the contract ever more complex to use, but sometimes, it may nonetheless be the better solution.

Cyan4973:
Trying to progress on the issues presented by the new design.

An idea could be to keep the contract of `HUF_buildCTable_wksp()` in place, with `maxNbBits` still being strictly enforced, as it is today.
Instead, the caller is expected to pass the maximum defined by the format of the decoder restriction (`11` for literals), and let `HUF_buildCTable_wksp()` do the adaptation to lower values internally. Effectively, it means the caller should no longer use `HUF_optimalTableLog()` (though it could _still_ do it, it would just reduce the margins left for `HUF_buildCTable_wksp()` to adjust).

Note that `HUF_buildCTable_wksp()` was always allowed to return with a lower `maxNbBits` than the one requested at its interface. For example, maybe there are only a few symbols, so the tree cannot be deep, even in degenerated cases. So this new policy doesn't break the contract.

**The Good** : 
- No change to current contract
- All existing tests pass as intended

**The Acceptable** : 
- change the zstd source code to no longer invoke `HUF_optimalTableLog()` before `HUF_buildCTable_wksp()`

**The Worrying** : 
- In order to reproduce the same behavior as currently, it would be necessary to invoke `HUF_optimalTableLog()` from _within_  `HUF_buildCTable_wksp()`. Problem is, one of the required argument, `srcSize`, is unknown to `HUF_buildCTable_wksp()`. This could be solved by either passing the argument to `HUF_buildCTable_wksp()`, thus killing the first property of this proposal (contract would be changed), or by re-calculating `srcSize` from the histogram, thus costing cpu.
   + Another alternative could be to _not_ use `HUF_optimalTableLog()`, and find another heuristic, which doesn't need `srcSize`. This would lead to different results. There results would have to be better, or at least neutral, with current version
- If the invocation _absolutely wants_ to control `tableLog`, it can only control its _maximum_ value. But `HUF_buildCTable_wksp()` is still free to lower `tableLog` it feels so. Since it's just an heuristic, it could make an unwise choice.
   + This suggests that future attempt to control `tableLog` "right", as in "better selected than the heuristic", should be implemented at the same position in the code, within  `HUF_buildCTable_wksp()`, and not outside of it.

Cyan4973:
OK, let's wait for #2970, to avoid confusion between the 2 efforts.

Or, alternatively, I could make this PR solely about traces, and not introduce any change in the huffman depth heuristic. This should make it a no-brainer to merge.

terrelln:
> Or, alternatively, I could make this PR solely about traces, and not introduce any change in the huffman depth heuristic. This should make it a no-brainer to merge.

I think it would make sense to break it into 2 PRs.
* One with the additional traces we can merge now
* One with the improved depth heuristic. Which we can wait for #2970, and decide what to do.

Cyan4973:
The last commit removes the update to the huffman depth heuristic,
so now this PR is mostly refactoring,
adding a few traces, changing a few comments.

We'll revisit the question of improved huffman depth heuristic after #2970 lands.

terrelln (approved):

PR #3022, opened 2022-01-22
AsyncIO compression part 2 - added async read and asyncio to compression code

yoniko:
PR 2/2 for asyncio compression.
This PR builds on the previous PR and adds:
- `ReadPool` - a thread pool based reader for async input.
- AsyncIO code paths to compression in zstdcli (fileio.c).

Benchmarking from a single desktop, shows this branch and dev. "No cache" means I used `dd` to remove file from fs caching before each run. `multiple.zst` is a 111MB file with multiple frames in it.
```
flag                                        --asyncio  --no-asyncio
input_file      comment
enwik8.zst      decompression                   0.168         0.206
                decompression no cache          0.167         0.213
                decompression no cache dev      0.172         0.214
multiple.zst    decompression                   0.888         1.076
                decompression no cache          0.896         1.100
                decompression no cache dev      0.929         1.112
silesia.tar.zst decompression                   0.273         0.368
                decompression no cache          0.276         0.379
                decompression no cache dev      0.292         0.388


flag                                  --asyncio  --asyncio --fast  --no-asyncio  --no-asyncio --fast
input_file  comment
enwik8      compression                   0.749             0.420         0.783                0.456
            compression no cache          0.751             0.420         0.797                0.470
            compression no cache dev      0.814             0.470         0.802                0.470
silesia.tar compression                   1.218             0.648         1.281                0.713
            compression no cache          1.227             0.646         1.294                0.749
            compression no cache dev      1.316             0.752         1.324                0.752
```


Cyan4973:
Maybe rebase on top of `dev` after `Part 1` get merged,
this way only changes related to `Part 2` will be shown in the `diff` to review.

programs/fileio.c
```diff
@@ -1033,6 +1034,8 @@ FIO_compressGzFrame(const cRess_t* ress,  /* buffers & handlers are used, but no
             EXM_THROW(79, "zstd: %s: deflateEnd error %d \n", srcFileName, ret);
     }   }
     *readsize = inFileSize;
+    AIO_WritePool_releaseIoJob(writeJob);
+    AIO_WritePool_sparseWriteEnd(ress->writeCtx);
```
terrelln, line 1038: We don't want to be doing sparse writes here, since we should basically never have long runs of zeros in a compressed file.

Do we always create the compression-side `writeCtx` with `sparseFileSupport == 0`?

yoniko:
> Maybe rebase on top of `dev` after `Part 1` get merged, this way only changes related to `Part 2` will be shown in the `diff` to review.

Merged, changes now belong only to part 2.

programs/fileio.c
yoniko: We create the `writeCtx` with the default `sparseFileSupport` for compression, which afaik is the global default (1 for all archs apart from MacOS).
This can be fixed by updating `sparseFileSupport` to 0 in the case of compression. Do you think it will make any measurable impact?

programs/fileio_asyncio.c
```diff
@@ -363,3 +365,223 @@ void AIO_WritePool_free(WritePoolCtx_t* ctx) {
...
+    if(ctx->base.threadPool) {
+        ZSTD_pthread_cond_signal(&ctx->jobCompletedCond);
+        ZSTD_pthread_mutex_unlock(&ctx->base.ioJobsMutex);
+    }
+}
+
+/* AIO_ReadPool_findNextWaitingOffsetCompletedJob:
+ * Looks through the completed jobs for a job matching the waitingOnOffset and returns it,
+ * if job wasn't found returns NULL.
+ * IMPORTANT: assumes ioJobsMutex is locked. */
```
terrelln, line 399: nit: Maybe add `_locked()` to the name?

programs/fileio_asyncio.c
```diff
@@ -363,3 +365,223 @@ void AIO_WritePool_free(WritePoolCtx_t* ctx) {
...
+static IOJob_t* AIO_ReadPool_findNextWaitingOffsetCompletedJob(ReadPoolCtx_t *ctx) {
+    IOJob_t *job = NULL;
+    int i;
+    for (i=0; i<ctx->completedJobsCount; i++) {
+        job = (IOJob_t *) ctx->completedJobs[i];
+        if (job->offset == ctx->waitingOnOffset) {
+            ctx->completedJobs[i] = ctx->completedJobs[--ctx->completedJobsCount];
+            return job;
+        }
+    }
```
terrelln, line 413: You're writing it this way so there isn't a dependency on which order the jobs finish correct?

In a single-threaded thread pool you could replace this with a fixed-size queue. But I see that you would want to write it this way so you could generalize to something like io_uring. Could you add a comment explaining why you are choosing this method?

programs/fileio_asyncio.c
```diff
@@ -363,3 +365,223 @@ void AIO_WritePool_free(WritePoolCtx_t* ctx) {
...
+        return;
+    }
+    job->usedBufferSize = fread(job->buffer, 1, job->bufferSize, job->file);
+    if(job->usedBufferSize < job->bufferSize) {
+        if(ferror(job->file)) {
+            EXM_THROW(37, "Read error");
+        } else if(feof(job->file)) {
+            ctx->reachedEof = 1;
+        } else
+        EXM_THROW(37, "Unexpected short read");
```
terrelln: nit: Please enclose the `else` in brackets & indent.

programs/fileio_asyncio.c
```diff
@@ -363,3 +365,223 @@ void AIO_WritePool_free(WritePoolCtx_t* ctx) {
...
+        job = AIO_ReadPool_getNextCompletedJob(ctx);
+        if(job == NULL)
+            break;
+        srcBufferOffsetFromBase = ctx->srcBuffer - ctx->srcBufferBase;
+        srcBufferRemainingSpace = ctx->srcBufferBaseSize - (srcBufferOffsetFromBase + ctx->srcBufferLoaded);
+        if(job->usedBufferSize > srcBufferRemainingSpace) {
+            memmove(ctx->srcBufferBase, ctx->srcBuffer, ctx->srcBufferLoaded);
+            ctx->srcBuffer = ctx->srcBufferBase;
+        }
+        memcpy(ctx->srcBuffer + ctx->srcBufferLoaded, job->buffer, job->usedBufferSize);
```
terrelln: Is there something you can do to avoid these `memcpy()`?

Do they actually happen in the normal case? E.g. if zstd is consuming the entire buffer do we ever hit this case?

If they don't happen for zstd, and only happen for other formats, then it probably isn't a big deal.

programs/fileio_asyncio.h
```diff
@@ -113,6 +128,44 @@ WritePoolCtx_t* AIO_WritePool_create(FIO_prefs_t* const prefs, size_t bufferSize
...
+/* AIO_ReadPool_consumeBytes:
+ * Consumes byes from srcBuffer's beginning and updates srcBufferLoaded accordingly. */
+void AIO_ReadPool_consumeBytes(ReadPoolCtx_t *ctx, size_t n);
+
+/* AIO_ReadPool_fillBuffer:
+ * Makes sure buffer has at least n bytes loaded (as long as n is not bigger than the initalized bufferSize).
+ * Returns if srcBuffer has at least n bytes loaded or if we've reached the end of the file.
+ * Return value is the number of bytes added to the buffer.
+ * Note that srcBuffer might have up to 2 times bufferSize bytes. */
+size_t AIO_ReadPool_fillBuffer(ReadPoolCtx_t *ctx, size_t n);
```
terrelln, line 158: We may be able to simplify this a little bit by saying you MUST consume the entire source buffer. And instead of storing it in the `ReadPoolCtx_t`, return the buffer from the `AIO_ReadPool_fillBuffer()` function.

Maybe that way we could avoid some memcpy's, or just simplify the API.

yoniko: The problem here is that you might have multiple frames in a single file, and those frames can theoretically be different formats as well. That means that we can't assume the whole buffer is consumed.

A possible API here would be `consume` and `read` where `read` would return a pointer to a buffer and a buffer size. That'd be cleaner as we won't expose `->srcBuffer` to the user.
The downsize is that kind of change would require more changes to `fileio.c`. As this is a big patch already I elected to avoid those, but willing to rethink it.

terrelln: > The problem here is that you might have multiple frames in a single file, and those frames can theoretically be different formats as well. That means that we can't assume the whole buffer is consumed.

Good point!

> A possible API here would be consume and read where read would return a pointer to a buffer and a buffer size. That'd be cleaner as we won't expose ->srcBuffer to the user.
> The downsize is that kind of change would require more changes to fileio.c. As this is a big patch already I elected to avoid those, but willing to rethink it.

I have a slight preference for returning a buffer pointer. That might also allow us to avoid some `memcpy()`. If we can give pointers directly to buffers that the jobs read into.

E.g. have one extra input buffer. And when we need to read the next job, swap the job's buffer with the current input buffer. That way the job can continue with its new buffer, and the input can read directly from the job's old buffer w/o a memcpy.

I'm sure theres different ways to achieve that as well, and maybe those are cleaner. Just a thought.

programs/fileio_asyncio.c
yoniko: Two possible ways to avoid `memcpy` (at the cost of code complexity) are: 
1. Continue holding the `IOJob_t` and referencing the buffer it owns. This doesn't remove the need for memcpy completely as there are cases where the code would ask for more bytes than are currently available and so we will need to coalesce two buffers. For that we could have another buffer allocated that will be used only when strickly needed.
2. Implement a ring buffer and have each `IOJob_t` reference a portion of this ring buffer.

The `memcpy/memove` bothered me, but I believe these two solutions increase the complexity of the code without actually improving performance. Memcpy should be cheap compared to our other actions (compression / decompression and IO). Am I wrong to assume memcpy is relatively cheap?

terrelln: > Am I wrong to assume memcpy is relatively cheap?

Relatively cheap, yes. Negligible, no.

Decompression runs at > 1GB/s. So adding an extra memcpy can be a 10% slowdown.

Compression can run at a few hundred MB/s, or GBs/s with multithreading. So an extra memcpy can be a 5% slowdown.

Definitely worth benchmarking.

Could you have jobs+1 buffers, and reserve 1 buffer for reading?  When you're finished reading, swap that buffer with the next jobs buffer?

programs/fileio.c
terrelln: I'd guess yes. But it would be worth benchmarking.

There is also no reason to do it, so setting it to 0 in the case of compression makes sense to me. And compression doesn't currently use sparse writes, so there is no change in behavior.

programs/fileio_asyncio.c
yoniko: Fixed

programs/fileio_asyncio.c
yoniko: You are correct.
Added a comment.

programs/fileio_asyncio.c
yoniko: Done

programs/fileio.c
yoniko: Fixed, now setting to 0 in `zstcli.c` for compression.

programs/fileio_asyncio.c
yoniko: > Decompression runs at > 1GB/s. So adding an extra memcpy can be a 10% slowdown.

The memcpy can happen only for input, not for output which should have a lower rate.
Other than that, up to now each such memcpy/memove was instead an `fread` which is a memcpy at the best case, but might also involve a syscall (depending on buffering). While the memcpy can have a non negligible cost, it is definitely cheaper than the previous implementation.

>  Could you have jobs+1 buffers, and reserve 1 buffer for reading? When you're finished reading, swap that buffer with the next jobs buffer?

As mentioned on our second thread this is problematic as there are some cases the user would `consume` some bytes leaving us with a lower number of bytes available than required in the next read. In such a cases we'd need to take some bytes from the current buffer and joint them with bytes from the next buffer. This can only be possible using `memcpy` or if the buffers are all a part of one continuous ring buffer (which also has edge cases toward the end of the buffer, unless we use some mmap tricks) .

yoniko: In any case, I will give it a try, and we can judge it afterwards.

terrelln: > In such a cases we'd need to take some bytes from the current buffer and joint them with bytes from the next buffer

Not necessarily. You could loosen the contract that you always return the requested # of bytes. All compressors should have the ability to consume the input, even if it is smaller than the expected size.

So in the case where you have leftover bytes, instead of joining with the next job, just return a shorter read. Then when its consumed the tail of the current buffer, grab the next job's buffer.

yoniko: Wouldn't that increase chances of memcpy to internal decompressor buffers though?

yoniko: > Relatively cheap, yes. Negligible, no.

You are right there. Perf shows 1-3.5% time spent on memcpy.
Also interesting is that `ZSTD_(de)compressStream` spends even more time on `memcpy`. I wonder if this is related to the code getting less data than expected.

yoniko:
Added benchmarking information and removed most of the memcpy calls.

programs/fileio.c
```diff
@@ -1130,13 +1143,15 @@ FIO_compressLz4Frame(cRess_t* ress,
     LZ4F_preferences_t prefs;
     LZ4F_compressionContext_t ctx;
 
+    IOJob_t *writeJob =AIO_WritePool_acquireJob(ress->writeCtx);
```
Cyan4973: minor (coding style): 
if `writeJob` pointer never changes after initialization, you should `const` it : `IOJob_t* const writeJob = ...`.
It sends a signal to the reviewer (or the next contributor) that this pointer will never change its value later on.
When applied consistently, its _absence_ becomes a signal that this pointer _will_ change in the future, and should therefore receive extra attention.

programs/fileio_asyncio.c
```diff
@@ -363,3 +359,251 @@ void AIO_WritePool_free(WritePoolCtx_t* ctx) {
...
+    if (ctx->base.threadPool)
+        ZSTD_pthread_mutex_unlock(&ctx->base.ioJobsMutex);
+    return job;
+}
+
+
+/* AIO_ReadPool_executeReadJob:
+ * Executes a read job synchronously. Can be used as a function for a thread pool. */
+static void AIO_ReadPool_executeReadJob(void* opaque){
+    IOJob_t* job = (IOJob_t*) opaque;
```
Cyan4973: minor : it looks to me that this pointer never changes, so it should be `IOJob_t* const`

tests/playTests.sh
```diff
@@ -186,7 +186,7 @@ fi
 
 println "\n===>  simple tests "
 
-datagen > tmp
+datagen -g500K > tmp
```
Cyan4973: any specific reason for this change ?

Cyan4973:
This is a good rewrite.
I believe it contributes positively to the readability of the source code.
It does so while providing new features, and improving performance measurably.
Great PR.

Cyan4973 (approved):
Only some very minor comments on coding style

tests/playTests.sh
yoniko: Yes.
One of the following tests checks that we fail with "-M2K". This test started failing (i.e decompression succeeded) when I introduced the changes in the branch.
Reason being that I increased the CLI's buffer sizes (and changed buffer usage pattern) resulting in less mallocs during decompression. Increasing the data size forces the lib to allocate space for the window. (**note this is as far as I currently understand the code, might have missed something in this explanation).

Cyan4973: OK, 
so maybe it is this one `-M2K` test (or series of tests) that needs to generate and use its own specific source,
rather than impacting a generic sample used by many other tests.

programs/fileio.c
yoniko: It actually does change later on with 
```
AIO_WritePool_enqueueAndReacquireWriteJob(&writeJob);
```

programs/fileio_asyncio.c
yoniko: Added `const`s throughout the new patch, I believe I found all occurrences that it could apply to.

tests/playTests.sh
yoniko: Done

programs/fileio.c
```diff
@@ -978,51 +972,58 @@ FIO_compressGzFrame(const cRess_t* ress,  /* buffers & handlers are used, but no
...
+            size_t const availBefore = strm.avail_in;
+            ret = deflate(&strm, Z_NO_FLUSH);
+            AIO_ReadPool_consumeBytes(ress->readCtx, availBefore - strm.avail_in);
         }
-        ret = deflate(&strm, Z_NO_FLUSH);
+
         if (ret != Z_OK)
-            EXM_THROW(72, "zstd: %s: deflate error %d \n", srcFileName, ret);
-        {   size_t const cSize = ress->dstBufferSize - strm.avail_out;
+        EXM_THROW(72, "zstd: %s: deflate error %d \n", srcFileName, ret);
```
terrelln: nit: This needs indentation

programs/fileio.c
```diff
@@ -1146,27 +1161,27 @@ FIO_compressLz4Frame(cRess_t* ress,
 #if LZ4_VERSION_NUMBER >= 10600
     prefs.frameInfo.contentSize = (srcFileSize==UTIL_FILESIZE_UNKNOWN) ? 0 : srcFileSize;
 #endif
-    assert(LZ4F_compressBound(blockSize, &prefs) <= ress->dstBufferSize);
+    assert(LZ4F_compressBound(blockSize, &prefs) <= writeJob->bufferSize);
 
     {
         size_t readSize;
```
terrelln: nit: Please rename this variable, it is super close to the parameter `U64* readsize`.

programs/fileio.c
```diff
@@ -1430,14 +1441,14 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
         }  /* while ((inBuff.pos != inBuff.size) */
     } while (directive != ZSTD_e_end);
 
-    if (ferror(srcFile)) {
-        EXM_THROW(26, "Read error : I/O error");
-    }
```
terrelln, line 1435: Based on my understanding of the code, we should be checking `ferror()` whenever we call `fread()` in the current implementation. No matter if we use `--async-io` or `--no-async-io`. Is that correct?

I just want to verify because this check is absolutely critical.

And by moving it to the `fread()`, we now get coverage for other formats.

programs/fileio.c
```diff
@@ -1919,54 +1923,39 @@ static dRess_t FIO_createDResources(FIO_prefs_t* const prefs, const char* dictFi
...
+        assert(writeSize <= writeJob->bufferSize);
+        memcpy(writeJob->buffer, ress->readCtx->srcBuffer, writeSize);
+        writeJob->usedBufferSize = writeSize;
+        AIO_WritePool_enqueueAndReacquireWriteJob(&writeJob);
+        AIO_ReadPool_consumeBytes(ress->readCtx, writeSize);
     }
-    assert(feof(finput));
-
-    AIO_fwriteSparseEnd(prefs, foutput, storedSkips);
+    assert(ress->readCtx->reachedEof);
```
terrelln, line 1954: This can't happen, right? Assuming the code is correct.

programs/fileio_asyncio.c
```diff
@@ -127,15 +129,23 @@ void AIO_fwriteSparseEnd(const FIO_prefs_t* const prefs, FILE* file, unsigned st
...
 
 static IOJob_t *AIO_IOPool_createIoJob(IOPoolCtx_t *ctx, size_t bufferSize) {
-    void *buffer;
-    IOJob_t *job;
-    job = (IOJob_t*) malloc(sizeof(IOJob_t));
-    buffer = malloc(bufferSize);
+    IOJob_t* const job  = (IOJob_t*) malloc(sizeof(IOJob_t));
+    void* const buffer = malloc(bufferSize);
     if(!job || !buffer)
     EXM_THROW(101, "Allocation error : not enough memory");
```
terrelln: nit: Indentation

programs/fileio_asyncio.c
```diff
@@ -363,3 +367,255 @@ void AIO_WritePool_free(WritePoolCtx_t* ctx) {
...
+size_t AIO_ReadPool_fillBuffer(ReadPoolCtx_t* ctx, size_t n) {
+    IOJob_t *job;
+    int useCoalesce = 0;
+    assert(n <= ctx->base.jobBufferSize);
+
+    /* We are good, don't read anything */
+    if (ctx->srcBufferLoaded >= n)
+        return 0;
+
+    /* We still have bytes loaded, but enough to satisfy caller. We need to get the next job
```
terrelln: nit: "but not enough"

programs/fileio.c
yoniko: > Based on my understanding of the code, we should be checking ferror() whenever we call fread() in the current implementation. No matter if we use --async-io or --no-async-io. Is that correct?

Yes, the flows are almost same for both cases. The difference is that the async code queues the execution function to the worker thread while the sync version directly calls the execution function (see `AIO_ReadPool_executeReadJob` and `AIO_IOPool_enqueueJob`).

programs/fileio_asyncio.c
yoniko: Fixed

programs/fileio_asyncio.c
yoniko: Fixed. 
I think these annoying indentation issues happened when I split my original branch into two and while copy-pasting some code my editor decided that no indentation is the correct indentation. :(

programs/fileio.c
yoniko: Actually the code wasn't correct, but it didn't happen because 64k is less than our in size (128k). I fixed it.

As it is now, it shouldn't trigger and might be redundant (I just replaced the previous assert because it was easy).
We should only exit the loop if `ress->readCtx->srcBufferLoaded==0`.
Since we fill the buffer right before this is checked, it should happen only if we reached the EOF or encountered a read error. 
Read errors, in turn, will panic before we get to this point.

programs/fileio.c
yoniko: Good point!
Just removed it as it was actually useless here.

programs/fileio.c
yoniko: Fixed.

PR #3023, opened 2022-01-22
fix sequence compression API in Explicit Delimiter mode

Cyan4973:
In Explicit Delimiter mode, the Sequence compression API was only working when 
the frame has only a single block
or when all blocks are full (except the last one).

This fix makes is possible to select block boundaries wherever the caller wants them.
It still requires that data conforms to `zstd` specification, aka block sizes still have a maximum.

The fuzzer capability has been extended to test the explicit delimiter mode too.


terrelln:
Looks like theres an assert in the `tests/fuzz/sequence_compression_api.c` fuzzer that needs to be fixed.

lib/common/error_private.c
```diff
@@ -27,7 +27,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(version_unsupported): return "Version not supported";
     case PREFIX(frameParameter_unsupported): return "Unsupported frame parameter";
     case PREFIX(frameParameter_windowTooLarge): return "Frame requires too much memory for decoding";
-    case PREFIX(corruption_detected): return "Corrupted block detected";
+    case PREFIX(corruption_detected): return "Input corruption detected";
```
terrelln: nit: Maybe "Data corruption detected"? It feels more natural than "Input corruption"

terrelln (approved):
LGTM once the tests pass

lib/compress/zstd_compress.c
```diff
@@ -6029,10 +6078,15 @@ ZSTD_compressSequences_internal(ZSTD_CCtx* cctx,
...
+        size_t compressedSeqsSize;
         size_t cBlockSize;
         size_t additionalByteAdjustment;
-        lastBlock = remaining <= cctx->blockSize;
-        blockSize = lastBlock ? (U32)remaining : (U32)cctx->blockSize;
+        size_t blockSize = determine_blockSize(cctx->appliedParams.blockDelimiters,
+                                        cctx->blockSize, remaining,
+                                        inSeqs, inSeqsSize, seqPos);
+        U32 const lastBlock = (blockSize == remaining);
+        assert(blockSize <= remaining);
```
terrelln: Looks like this assert is failing.

Maybe it just needs to be moved below the `FORWARD_IF_ERROR()` check?

Cyan4973: Yes, that's exactly the issue, 
`blockSize` can be an error code, so this `assert()` is only valid after the `FORWARD_IF_ERROR()`.

Cyan4973:
The misplaced `assert()` was the only issue affecting the library.
The bulk of the work (and changes) was in the fuzz tester itself, in order to make it compatible with the Explicit Delimiter mode. As a nice side-effect, it allowed observing that the library was correctly rejecting all sort of bad conditions, from incorrect sequences, to invalid blocks, to incomplete frames, etc.
So I think this is finally good to go.

PR #3027, opened 2022-01-25
fix for -r on empty directory 

brailovich:
fix for -r on empty directory resulted in zstd waiting for input from stdin. this fix makes zstd exit without erroring
but printing a warning message explaining that there were no files or non-empty directories to process.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -1366,8 +1366,8 @@ int main(int argCount, const char* argv[])
        was a number of empty directories. In this case
        stdin and stdout should not be used */
        if (nbInputFileNames > 0 ){
-        DISPLAYLEVEL(2, "please provide correct input file(s) or non-empty directories -- ignored \n");
-        CLEAN_RETURN(2);
+        DISPLAYLEVEL(1, "please provide correct input file(s) or non-empty directories -- ignored \n");
+        CLEAN_RETURN(0);
        }
        UTIL_refFilename(filenames, stdinmark);
     }
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -834,6 +834,7 @@ int main(int argCount, const char* argv[])
     size_t streamSrcSize = 0;
     size_t targetCBlockSize = 0;
     size_t srcSizeHint = 0;
+    size_t nbInputFileNames = 0;
     int dictCLevel = g_defaultDictCLevel;
     unsigned dictSelect = g_defaultSelectivityLevel;
 #ifndef ZSTD_NODICT
@@ -1256,6 +1257,8 @@ int main(int argCount, const char* argv[])
         }
     }
 
+    nbInputFileNames = filenames->tableSize; /* saving number of input files */
+    
     if (recursive) {  /* at this stage, filenameTable is a list of paths, which can contain both files and directories */
         UTIL_expandFNT(&filenames, followLinks);
     }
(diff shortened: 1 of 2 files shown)


terrelln:
Thanks for the PR @brailovich!

I definitely agree with the behavior you want. We definitely don't want to use stdin when it is a TTY. There is a small question of whether we want to use stdin when it isn't a TTY. However, gzip doesn't. So I think that this behavior is fine.

programs/zstdcli.c
```diff
@@ -1358,7 +1361,17 @@ int main(int argCount, const char* argv[])
 #endif
 
     /* No input filename ==> use stdin and stdout */
-    if (filenames->tableSize == 0) UTIL_refFilename(filenames, stdinmark);
+    if (filenames->tableSize == 0) {
+      /* It is possible that the input
+       was a number of empty directories. In this case
+       stdin and stdout should not be used */
+       if (nbInputFileNames > 0 ){
+        DISPLAYLEVEL(2, "please provide correct input file(s) or non-empty directories -- ignored \n");
```
terrelln: This should be `DISPLAYLEVEL(1, ` since this is an error.

programs/zstdcli.c
```diff
@@ -1358,7 +1361,17 @@ int main(int argCount, const char* argv[])
...
+        CLEAN_RETURN(2);
```
terrelln: This should also be `CLEAN_RETURN(1);`

Cyan4973: Actually, if we want the CLI to return without erroring in this scenario, it should be `CLEAN_RETURN(0)`.

Cyan4973:
Agreed with @terrelln comments,
this PR proposes a good behavior for the CLI.

Consider adding a test, to ensure this behavior will be preserved across future changes by future committers.

programs/zstdcli.c
brailovich: thanks, it should be CLEAN_RETURN(0)

brailovich:
tests added, return values changed.

tests/playTests.sh
```diff
@@ -390,6 +390,18 @@ println "\n===>  recursive mode test "
 # combination of -r with empty list of input file
 zstd -c -r < tmp > tmp.zst
 
+# combination of -r with empty folder
+mkdir -p tmpEmptyDir
+zstd -r tmpEmptyDir 2>tmplog2
+if [grep "aborting" tmplog2]; then
```
Cyan4973: minor `sh` error here with `[` and `]`

brailovich: thanks, updated!

tests/playTests.sh
```diff
@@ -390,6 +390,18 @@ println "\n===>  recursive mode test "
 # combination of -r with empty list of input file
 zstd -c -r < tmp > tmp.zst
 
+# combination of -r with empty folder
+mkdir -p tmpEmptyDir
+zstd -r tmpEmptyDir 2>tmplog2
+if [ grep "aborting" tmplog2 ]; then
```
Cyan4973: Seems this construction is still problematic for `sh` compatibility.

I'm no `sh` expert, and this specific construction is not used in our test script, 
but I can recommend an existing idiom that has worked well so far : 

```
zstd -r tmpEmptydir 2>&1 | grep "aborting" && die "Should not abort on empty directory"
```

This would be equivalent, but you could go for something even simpler : 

```
zstd -r tmpEmptydir 
```

this second test only relies on return code being successful, which, I believe, is the goal of the test ?

brailovich: thanks! my initial thought was to go with just zstd -r tmpEmptydir  but I thought that the test should be more verbose. if this is acceptable, that's the best solution, thanks!

Cyan4973 (approved):
Looks good to me ! Thanks @brailovich for your contribution !

tests/playTests.sh
brailovich: btw, "if then" sh syntax allows cleaning tmp dirs/files before exiting.

PR #3030, opened 2022-01-27
Print zlib/lz4/lzma library versions in verbose version output

terrelln:
Knowing the version of zlib/lz4/lzma we're linking against is very
useful for debugging issues with those libraries, so print it out in the
verbosity 4 version output.

Also print this information at the top of `playTests.sh`.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -637,6 +637,11 @@ static void printVersion(void)
     #endif
         DISPLAYOUT("\n");
         if (g_displayLevel >= 4) {
+            /* library versions */
+            DISPLAYOUT("zlib version %s\n", FIO_zlibVersion());
+            DISPLAYOUT("lz4 version %s\n", FIO_lz4Version());
+            DISPLAYOUT("lzma version %s\n", FIO_lzmaVersion());
+
             /* posix support */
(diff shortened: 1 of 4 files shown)


programs/zstdcli.c
```diff
@@ -637,6 +637,11 @@ static void printVersion(void)
     #endif
         DISPLAYOUT("\n");
         if (g_displayLevel >= 4) {
+            /* library versions */
+            DISPLAYOUT("zlib version %s\n", FIO_zlibVersion());
```
Cyan4973, line 641: Do I understand that a library version is displayed for each and every supported external codec,
irrespective of the fact that `zstd` was linked to them or not ?
(and display "Unsupported" in this last case)

terrelln: Yup. We could change that behavior by gating each of these behind an `ifdef` check. I figured that since this is in `zstd -vvV` output, there is not need to make it super pretty. I'm okay either way though, so lmk if you want me to change it.

Cyan4973: It's not urgent indeed. This is a debug print, I think some verbosity is acceptable.

Cyan4973 (approved):

PR #3020, opened 2022-01-21
Add new CLI testing platform

terrelln:
The first 3 commits fix bugs found by the CLI testing platform and fix

The 4th commit adds the new CLI testing platform that I'm proposing.
See `tests/cli-tests/README.md` [here](https://github.com/terrelln/zstd/tree/cli-tests/tests/cli-tests).

The final commit hooks the CLI test suite into `make test`, but not
`make check`. This runs in CI everywhere that `make test` runs.

PR #3031, opened 2022-01-27
Move zstdless and zstdgrep tests into separate script

binhdvo:
Refactors CLI tests for zstdless and zstdgrep out of playTests.sh and adds some tests for passing flags to less through zstdless


tests/playToolTests.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/sh
```
felixhandte, line 1: I'm concerned that this preamble should stay in sync with the one in `playTests.sh` but may fall out of sync as `playTests.sh` will likely get more attention than this narrower-purpose/smaller script. And this setup stuff is largely to support exotic/non-standard platforms, where it's easy for us to blindly miss something.

A safer approach is probably to share this code in some shared file that can be included from both scripts (via `source` or `.`, so that it effects the caller's environment).

felixhandte (commented):
Looks good except as noted.

Cyan4973:
To be coordinated with #3020

PR #3035, opened 2022-01-29
Macos playtest envvars fix

yoniko:
playtests.sh: fix for a bug in macos' /bin/sh that persists temporary env vars when introduced before function calls

### DIFF
--- a/tests/cli-tests/run.py
+++ b/tests/cli-tests/run.py
@@ -278,7 +278,8 @@ class TestCase:
         Returns the environment to be used for the
         test subprocess.
         """
-        env = copy.copy(os.environ)
+        # We want to omit ZSTD cli flags so tests will be consistent across environments
+        env = {k: v for k, v in os.environ.items() if not k.startswith("ZSTD")}
         for k, v in self._opts.env.items():
             self._vlog(f"${k}='{v}'")
             env[k] = v
--- a/tests/cli-tests/run.py
+++ b/tests/cli-tests/run.py
@@ -283,6 +283,7 @@ class TestCase:
         for k, v in self._opts.env.items():
             self._vlog(f"${k}='{v}'")
             env[k] = v
+        return env
 
     def _launch_test(self) -> None:
         """Launch the test subprocess, but do not join it."""
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -210,6 +210,11 @@ zstd -c --fast=0 tmp > $INTOVOID && die "--fast must not accept value 0"
 println "test : too large numeric argument"
 zstd --fast=9999999999 -f tmp  && die "should have refused numeric value"
 println "test : set compression level with environment variable ZSTD_CLEVEL"
+
+if [ -n "${ZSTD_CLEVEL}" ]; then
+  # Save the original value because macos /bin/sh would actually override the envvar's value in the following tests
+  ORIGINAL_ZSTD_CLEVEL=$ZSTD_CLEVEL
+fi
 ZSTD_CLEVEL=12  zstd -f tmp # positive compression level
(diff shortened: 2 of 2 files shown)


tests/playTests.sh
```diff
@@ -220,6 +225,14 @@ ZSTD_CLEVEL=3a7 zstd -f tmp # malformed env var, warn and revert to default sett
 ZSTD_CLEVEL=50000000000 zstd -f tmp # numeric value too large, warn and revert to default setting
 println "test : override ZSTD_CLEVEL with command line option"
 ZSTD_CLEVEL=12  zstd --fast=3 -f tmp # overridden by command line option
+# restore original value or unset
+if [ -n "${ORIGINAL_ZSTD_CLEVEL}" ]; then
+  ZSTD_CLEVEL=$ORIGINAL_ZSTD_CLEVEL
+else
+  unset ZSTD_CLEVEL
```
Cyan4973: If the goal is just to ensure that no environment variables bleed between tests, they maybe just this line : 
`unset ZSTD_CLEVEL` 
would be enough ?

I don't see a good reason to preserve previous value of `ZSTD_CLEVEL` and re-generate it for future tests.
Ideally, we would like each test to be as autonomous as possible, to better control test conditions.
This `playtests.sh` script doesn't always respect this rule, sometimes to better re-employ resources, but I believe it would be preferable to lean towards "more autonomous tests".

Cyan4973:
To be coordinated with changes in #3020

tests/playTests.sh
yoniko: We can definitely do that, but if we do we should probably start the script with an unset so the environment variables will be disabled for the previous tests as well. Will get to it later today.

tests/cli-tests/run.py
```diff
@@ -280,6 +280,9 @@ def _test_environment(self) -> typing.Dict[str, str]:
         """
         env = copy.copy(os.environ)
         for k, v in self._opts.env.items():
+            if k.startswith("ZSTD"):
+                # We want to omit ZSTD cli flags so tests will be consistent across environments
+                continue
```
terrelln: This will remove the flags that we add, not the existing environment variables. E.g. `$ZSTD_BIN`.

Instead, we should remove elements from `os.environ` that start with `ZSTD`. But leave everything in `self._opts.env`.

terrelln: See

https://github.com/facebook/zstd/blob/b9566fc558830de2a3fd486961b674774c17b74e/tests/cli-tests/bin/zstd#L1-L7

And

https://github.com/facebook/zstd/blob/b9566fc558830de2a3fd486961b674774c17b74e/tests/cli-tests/run.py#L659-L666

yoniko: Correct, it was my bad for not paying enough attention. Pushed a possible fix now.

tests/cli-tests/run.py
```diff
@@ -278,7 +278,8 @@ def _test_environment(self) -> typing.Dict[str, str]:
         Returns the environment to be used for the
         test subprocess.
         """
-        env = copy.copy(os.environ)
+        # We want to omit ZSTD cli flags so tests will be consistent across environments
+        env = {k: v for k, v in os.environ.items() if not k.startswith("ZSTD")}
```
terrelln, line 282: Ohhhhh I figured out why its broken... This function is missing a `return env`... Adding that fixes it.

It only works before because apparently `copy.copy(os.environ)` doesn't really copy it. And modifying `env` added stuff to `os.environ`.

Python is fun.

yoniko: Hah, guess we came to this realization at the same time :)
Tests are running now.

PR #3036, opened 2022-01-30
fixed bug 44168

Cyan4973:
discovered by oss-fuzz

It's a bug in the test itself :
`ZSTD_compressBound()` provides an upper bound of the compressed size
but only works for data compressed "normally".
In situations where many flushes are forcefully introduced,
this creates many more blocks than expectation,
each of which has a potential to increase the compressed size by 3 more bytes.
In extreme cases (lots of small incompressible blocks), the expansion can go beyond `ZSTD_compressBound()`.

This situation is similar when using the `ZSTD_compressSequences()` API with Explicit Block Delimiters.
In which case, each explicit delimiter acts like a flush() order.
When this API is exploited by a fuzzer, it's possible to generate scenarios like the one described above,
with tons of incompressible blocks of small sizes, thus reaching beyond `ZSTD_compressBound()`.

fix : in Explicit Block Delimiters mode, use a (much) larger compressed bound, to account for above scenario.


Cyan4973:
Thinking a bit more about it, I don't think it's the best fix.

Increasing `dstCapacity` will make the fuzzer less likely to generate a scenario which overflows `dstCapacity`. But since it's a fuzzer, it will over time find a way towards that goal.
The issue here is that there is _no limit_ to the possible expansion factor in the test with "Explicit Delimiters". That's because this scenario is allowed to generate empty blocks (each one using 3 bytes). So even with a `dstCapacity` as a large multiple of `srcSize`, a fuzzer could, at least in theory, find a way to consume all `dstCapacity` irrespective of original `srcSize`.

There are 2 ways to solve this issue : 
- Change the nature of the test, so that a `dstCapacity_tooSmall` error is now a possible outcome of the test. 
   - So far, the current test, which has been refurbished for the "Explicit Delimiter" mode, presumes that it's always possible to compress the set of sequences, and then it's always possible to decompress it and compare, realizing a round-trip check. 
   - If it's possible for the compression stage to fail, it means no decompression and no roundtrip.
   - Yet, it could still be considered a "valid" and "successful" test, on the ground that the API correctly detected and dealt with an under-sized `dstCapacity`.
- Use `nbSeqs` as a way to determine `dstCapacity` : since we know that each block can add up to 3 additional bytes, we could calculate a pessimistic bound by adding 3 more bytes per sequence. This way, even if the input is full of empty blocks, it's accounted for.

The 2nd method seems a clever way to preserve the test as it was originally designed, by making the compression stage always successful, thus always allowing a validation round-trip.

However, bugs like #3040 makes me think that the 1st method is actually a better choice. That's because without a scenario testing the ability of the API to deal with undersized `dstCapacity`, it would not be possible to generate, hence to detect, bugs like oss-44239 (fixed in #3040).

So I think I'll update this PR to reflect the new goal.

Cyan4973:
I believe this PR is ready for review and merge

terrelln (approved):

PR #3039, opened 2022-01-31
Meson fixups for Windows

eli-schwartz:
Three changes:
- MSVC did not build, because the resource compiler (`*.rc` files) was not instructed where to find zstd.h
- fix building the testsuite binaries on Windows, which doesn't have a separate libm
- console UI fixes for the valgrind wrapper

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,8 +92,8 @@ matrix:
         - make -C tests versionsTest
 
     # meson dedicated test
-    - name: Xenial (Meson + clang)    # ~15mn
-      dist: bionic
+    - name: Focal (Meson + clang)    # ~15mn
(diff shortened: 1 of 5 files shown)


eli-schwartz:
Aside: https://github.com/mesonbuild/wrapdb/actions/runs/1770447908 (This is after I backport the fixes.)

The build itself is green on meson+MSVC, but the test is red:
```
3/8 test-zstream-3           FAIL             3.94s   (exit status 3221225477 or signal 3221225349 SIGinvalid)
```

eli-schwartz:
Alright, let's try to actually fix that meson CI on travis...

Cyan4973 (approved):

PR #3041, opened 2022-01-31
Fix broken Travi CI build - installing pip for python 3.6

yoniko:
Pip install script no longer supports python3.6 by default, switched to a script that does.


Cyan4973 (approved):
Thanks @yoniko , this fixes an annoying CI issue.

PR #3042, opened 2022-02-01
Fix required decompression memory usage reported by -vv + --long

### LINKED ISSUE #2968: Minor -vv issue.
I use   `tar -I 'zstd -T0 --long -vv' -cf `   command and decompression only 2 MiB required, instead --long (default) 128 MiB.

```
*** zstd command line interface 64-bits v1.5.1, by Yann Collet ***
Note: 4 physical core(s) detected 
--zstd=wlog=21,clog=16,hlog=17,slog=1,mml=5,tlen=0,strat=2
--format=.zst --block-size=0 --memory=134217728 --threads=4 --content-size
Using stdin for input 
Using stdout for output 
Sparse File Support is automatically disabled on stdout ; try --sparse 
Decompression will require 2097152 B of memory
/*stdin*\            : Completed in 1.63 sec  (cpu load : 198%)                B => 98.47% 
```

u1f35c:
The use of --long alters the window size internally in the underlying
library (lib/compress/zstd_compress.c:ZSTD_getCParamsFromCCtxParams),
which changes the memory required for decompression. This means that the
reported requirement from the zstd binary when -vv is specified is
incorrect.

A full fix for this would be to add an API call to be able to retrieve
the required decompression memory from the library, but as a
lighterweight fix we can just take account of the fact we've enabled
long mode and update our verbose output appropriately.

Fixes #2968

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1301,8 +1301,13 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
         UTIL_HumanReadableSize_t windowSize;
         CHECK(ZSTD_CCtx_getParameter(ress.cctx, ZSTD_c_windowLog, &windowLog));
         if (windowLog == 0) {
-            const ZSTD_compressionParameters cParams = ZSTD_getCParams(compressionLevel, fileSize, 0);
-            windowLog = cParams.windowLog;
+            if (prefs->ldmFlag) {
+                /* If long mode is set without a window size libzstd will set this size internally */
+                windowLog = ZSTD_WINDOWLOG_LIMIT_DEFAULT;
+            } else {
+                const ZSTD_compressionParameters cParams = ZSTD_getCParams(compressionLevel, fileSize, 0);
+                windowLog = cParams.windowLog;
+            }
         }
         windowSize = UTIL_makeHumanReadableSize(MAX(1ULL, MIN(1ULL << windowLog, pledgedSrcSize)));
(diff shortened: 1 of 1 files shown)


programs/fileio.c
```diff
@@ -1300,7 +1300,10 @@ FIO_compressZstdFrame(FIO_ctx_t* const fCtx,
         int windowLog;
         UTIL_HumanReadableSize_t windowSize;
         CHECK(ZSTD_CCtx_getParameter(ress.cctx, ZSTD_c_windowLog, &windowLog));
-        if (windowLog == 0) {
+        if (prefs->ldmFlag) {
+            /* If long mode is set libzstd will set this window size internally */
+            windowLog = ZSTD_WINDOWLOG_LIMIT_DEFAULT;
```
Cyan4973: `--long` mode can be activated with a parameter.
For example, `--long=30` means "use a 1 GB window size".
The shorter `--long` is equivalent to `--long=27`.

This means we can't guarantee that the window size is necessarily `ZSTD_WINDOWLOG_LIMIT_DEFAULT` when the `--long` mode is activated, it could be a different value if an explicit parameter was passed.

u1f35c: Thanks, yes, there's an override via `ZSTD_overrideCParams` that I missed, so the ldmFlag check should only be performed if the windowLog ends up as 0.

Cyan4973 (approved):

Cyan4973:
Thanks @u1f35c !

PR #3045, opened 2022-02-02
Refactor bitwise intrinsics

### LINKED ISSUE #3067: ZSTD_NbCommonBytes implementation is incorrect for Big-Endian MSVC
[These](https://github.com/facebook/zstd/blob/970460f67d8641f38d4954683e95887d97d3e388/lib/compress/zstd_compress_internal.h#L761) [lines](https://github.com/facebook/zstd/blob/970460f67d8641f38d4954683e95887d97d3e388/lib/compress/zstd_compress_internal.h#L782) should be changed to `(r ^ 63) >> 3` and  `(r ^ 31) >> 3`, respectively. I found the bug while adding tests for bitwise intrinsics in https://github.com/facebook/zstd/pull/3045. That PR fixes the bug (as well as removing the Big-Endian MSVC specific code, which is probably not used by anyone).

embg:
zstd currently has multiple nearly-identical implementations of several bitwise intrinsics, such as countTrailingZeros, highbit32, and NbCommonBytes. This PR unifies the implementations and moves them into a new file, `lib/common/bits.h` (with the exception of `MEM_swap32` and `MEM_swap64`, which have to stay in `mem.h` to avoid circular dependency). Additionally, this PR pulls out the software fallback implementations for each operation into a separate function to allow for unit testing against the intrinsic implementations for each platform.

Before I take the PR out of draft mode, I will add a file with those unit tests. For now, I am putting up the refactoring portion of the PR (which is complete) so I can finalize that code to allow for benchmarking, which is required to merge this change. I will work on the unit tests while I wait for review on the refactoring portion, which I expect will take some time.

Prior work in this area: https://github.com/facebook/zstd/pull/2831


lib/decompress/huf_decompress.c
```diff
@@ -263,7 +264,7 @@ static size_t HUF_initRemainingDStream(BIT_DStream_t* bit, HUF_DecompressAsmArgs
 
     /* Construct the BIT_DStream_t. */
     bit->bitContainer = MEM_readLE64(args->ip[stream]);
-    bit->bitsConsumed = ZSTD_countTrailingZeros((size_t)args->bits[stream]);
+    bit->bitsConsumed = ZSTD_countTrailingZeros64(args->bits[stream]);
```
embg, line 268: @terrelln is this safe? My reasoning is that `args->bits[stream]` has type `U64`, so the cast is unnecessary on 64-bit and could be problematic on 32-bit, unless I'm misunderstanding the purpose?

If the `size_t` does need to stay around, I can branch on `MEM_64bit()` and add a call to `ZSTD_countTrailingZeros32`. Just let me know what's appropriate here.

Cyan4973: I note that `bitContainer` is of type `size_t`,
so it seems to me that the issue is the previous `MEM_readLE64()`.

I presume a `-Wconversion` warning flag would have caught such silent downsizing.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+        return (unsigned)__builtin_ctz(val);
+#   elif defined(__ICCARM__)    /* IAR Intrinsic */
+        return __CTZ(val);
+#   else
+        return ZSTD_countTrailingZeros32_fallback(val);
+#   endif
+}
+
+MEM_STATIC unsigned ZSTD_countTrailingZeros64_fallback(U64 val)
+{
```
embg: [Benchmarking](https://gist.github.com/embg/728f102cf4387919b7c3fccafc170c48) favored this fallback implementation over the one from `ZSTD_VecMask_next`.

lib/decompress/huf_decompress.c
embg: Update from Nick: this code only runs on x86-64. So my change is safe.

Cyan4973: > Update from Nick: this code only runs on x86-64. So my change is safe.

It works, yes.
But as many things, it works fine today, and thanks to complex environmental guarantees, that are easy to forget.
Such knowledge can be lost in time, and some future committer may inadvertently trigger a bug there.
So part of the job is to make such future mistake more difficult to happen.

I can imagine 2 ways : 
- Stay on `(size_t)`, because it's the official type for this field. It happens to be 64-bit, but that's secondary (the code would still be technically correct for 32-bit too). You can leave a comment about the field width to improve understanding for future reviewers.
- Move firmly into `64-bit`, but in this case, `STATIC_ASSERT` that `sizeof(size_t)==8`. This will ensure that this code cannot be compiled in 32-bit mode, where this assumption would be wrong. It also justifies why a 64-bit type is used here instead of the expected `size_t`.

Note that some compiler might still not like the `U64 == size_t` equivalence, even when they are both 64-bit, on the ground that they are different types. As usual for this type of warnings, it depends. Compilers can change their mind, even between versions.

terrelln: Removing the cast is the correct decision.

It was added because `ZSTD_countTrailingZeros()` accepted a `size_t`. But, now that we have a `U64` version we can use it directly.


Now the  question is the previous line `bitContainer = MEM_readLE64()`. It could instead be `MEM_readLEST()`. However, this code would be completely broken if `sizeof(size_t) != 8`, because we're grabbing `bitsConsumed` from a 64-bit container.


This code is ifdefed behind `ZSTD_ENABLE_ASM_X86_64_BMI2`, so we could add a `ZSTD_STATIC_ASSERT(sizeof(size_t) == 8) `in this function.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+#ifndef ZSTD_BITS_H
+#define ZSTD_BITS_H
+
+#include "mem.h"
+
+MEM_STATIC unsigned ZSTD_highbit32_fallback(U32 val) {
+    static const U32 DeBruijnClz[32] = {  0,  9,  1, 10, 13, 21,  2, 29,
+                                         11, 14, 16, 18, 22, 25,  3, 30,
+                                          8, 12, 20, 28, 15, 17, 24,  7,
+                                         19, 27, 23,  6, 26,  5,  4, 31 };
```
terrelln: nit: add an `assert(val != 0);`

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+#   endif
+    }
+}
+
+MEM_STATIC unsigned ZSTD_countTrailingZeros32_fallback(U32 val)
+{
+    static const int DeBruijnBytePos[32] = {  0,  1, 28,  2, 29, 14, 24,  3,
+                                             30, 22, 20, 15, 25, 17,  4,  8,
+                                             31, 27, 13, 23, 21, 19, 16,  7,
+                                             26, 12, 18,  6, 11,  5, 10,  9 };
```
terrelln: nit: `assert(val != 0);`

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+MEM_STATIC unsigned ZSTD_countTrailingZeros64_fallback(U64 val)
+{
+    static const int DeBruijnBytePos[64] = {  0,  1,  2,  7,  3, 13,  8, 19,
+                                              4, 25, 14, 28,  9, 34, 20, 56,
+                                              5, 17, 26, 54, 15, 41, 29, 43,
+                                             10, 31, 38, 35, 21, 45, 49, 57,
+                                             63,  6, 12, 18, 24, 27, 33, 55,
+                                             16, 53, 40, 42, 30, 37, 44, 48,
+                                             62, 11, 23, 32, 52, 39, 36, 47,
+                                             61, 22, 51, 46, 60, 50, 59, 58 };
```
terrelln: nit: `assert(val != 0);`

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+            return r;
+        } else { /* 32 bits */
+            if (!(val>>16)) { r=2; val>>=8; } else { r=0; val>>=24; }
+            r += (!val);
+            return r;
+        }
+    }
+}
+
+MEM_STATIC unsigned ZSTD_NbCommonBytes(size_t val)
```
terrelln: nit: Add a `ZSTD_countLeadingZeros{32, 64}()` function.

Then you can implement `ZSTD_NbCommonBytes()` without any `#ifdef` code or fallbacks.

The only difference is that the software fallback would need one extra `>> 3` instruction. But I don't think that is a big deal.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+                } else {
+                    /* Should not reach this code path */
+                    __assume(0);
+                }
+#           elif defined(__GNUC__) && (__GNUC__ >= 3)
+                return (unsigned)(__builtin_ctz((U32)val) >> 3);
+#           else
+                return ZSTD_NbCommonBytes_fallback(val);
+#           endif
+        }
```
terrelln: E.g. the LE branch would be this.

```suggestion
        if (MEM_64bits()) {
            return ZSTD_countTrailingZeros64((U64)val) >> 3;
        } else { /* 32 bits */
            return ZSTD_countTrailingZeros32((U32)val) >> 3;
        }
```

And the BE branch would be the same but with `ZSTD_countLeadingZeros*()`.

lib/compress/zstd_lazy.c
```diff
@@ -1202,7 +1165,7 @@ size_t ZSTD_RowFindBestMatch(
 
         /* Cycle through the matches and prefetch */
         for (; (matches > 0) && (nbAttempts > 0); --nbAttempts, matches &= (matches - 1)) {
-            U32 const matchPos = (head + ZSTD_VecMask_next(matches)) & rowMask;
+            U32 const matchPos = (head + ZSTD_countTrailingZeros64(matches)) & rowMask;
```
terrelln: nit: I would keep the `ZSTD_VecMask_next()` function and just have it wrap `ZSTD_countTrailingZeros64()`. The reason is because that function name actually conveys something, and the documentation is useful.

terrelln (commented):
This is looking great!

Just some minor nits + the unit tests and it will be good to go!

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+        if (MEM_64bits()) {
+            static const int DeBruijnBytePos[64] = { 0, 0, 0, 0, 0, 1, 1, 2,
+                                                     0, 3, 1, 3, 1, 4, 2, 7,
+                                                     0, 2, 3, 6, 1, 5, 3, 5,
+                                                     1, 3, 4, 4, 2, 5, 6, 7,
+                                                     7, 0, 1, 2, 3, 3, 4, 6,
+                                                     2, 6, 5, 5, 3, 4, 5, 6,
+                                                     7, 1, 2, 4, 6, 4, 4, 5,
+                                                     7, 2, 6, 5, 7, 6, 7, 7 };
+            return DeBruijnBytePos[((U64)((val & -(long long)val) * 0x0218A392CDABBD3FULL)) >> 58];
```
aqrit:
```
const U64 m = 0x0101010101010101ULL;
val ^= val - 1;
return (unsigned)(((U64)((val & (m - 1)) * m)) >> 56);
```

embg: @aqrit Can we address this in a separate PR? So this can be closer to a pure refactor?

aqrit: > Then you can implement ZSTD_NbCommonBytes() without any #ifdef code or fallbacks.
The only difference is that the software fallback would need one extra `>> 3` instruction. But I don't think that is a big deal.

terrelln is advocating for refactoring away this function? So it seemed like a good time to bring it up :-(

terrelln: @aqrit is there a scenario you care about that uses this fallback code?

aqrit: no.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+                unsigned long r;
+                _BitScanReverse(&r, val);
+                return (unsigned)r;
+            } else {
+                /* Should not reach this code path */
+                __assume(0);
+            }
+#       endif
+#   elif defined(__GNUC__) && (__GNUC__ >= 3)   /* GCC Intrinsic */
+        return (unsigned)__builtin_clz (val) ^ 31;
```
ghost: `__builtin_c[tl]z(ll)?` function is available in gcc 3.4+, here is 3.0+.
https://en.wikipedia.org/wiki/Find_first_set#Tool_and_library_support

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+            return _lzcnt_u32(val)^31;
+#       else
+            if (val != 0) {
+                unsigned long r;
+                _BitScanReverse(&r, val);
+                return (unsigned)r;
+            } else {
+                /* Should not reach this code path */
+                __assume(0);
+            }
```
ghost, line 42: This can be in this way, https://godbolt.org/ shows the same assembly code generated. 
```c
        unsigned long r;
        _BitScanReverse(&r, val);
        return (U32) r;
```
Other MSVC code can also be simplified like this.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+            _BitScanForward(&r, val);
+            return (unsigned)r;
+        } else {
+            /* Should not reach this code path */
+            __assume(0);
+        }
+#   elif defined(__GNUC__) && (__GNUC__ >= 3)
+        return (unsigned)__builtin_ctz(val);
+#   elif defined(__ICCARM__)    /* IAR Intrinsic */
+        return __CTZ(val);
```
ghost: After checking "IAR C/C++ Development Guide" (get via Google), `__CTZ()` doesn't exist, only `__CLZ()` is available.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+#       else
+            if (val != 0) {
+                unsigned long r;
+                _BitScanForward64(&r, val);
+                return (unsigned)r;
+            } else {
+                /* Should not reach this code path */
+                __assume(0);
+            }
+#       endif
```
ghost, line 105: 32-bit MSVC code path can't use intrinsics, IMO #2831's mode is easy & clear.

embg: This preserves the behavior of ZSTD_VecMask_next, which is the only callsite for countTrailingZeros64 when zstd is compiled in 32-bit mode. So this PR should not regress performance for 32-bit MSVC. 

I'd definitely be curious if the 32-bit path in this function for GNUC (which uses 32-bit intrinsics to calculate countTrailingZero64) can be replicated for MSVC. If you have any ideas there, I'd welcome a separate PR with those changes. You would probably be able to measure a non-trivial performance win on 32-bit MSVC for compression levels 5+.

lib/common/bits.h
```diff
@@ -0,0 +1,244 @@
...
+#define ZSTD_BITS_H
+
+#include "mem.h"
+
+MEM_STATIC unsigned ZSTD_highbit32_fallback(U32 val) {
+    static const U32 DeBruijnClz[32] = {  0,  9,  1, 10, 13, 21,  2, 29,
+                                         11, 14, 16, 18, 22, 25,  3, 30,
+                                          8, 12, 20, 28, 15, 17, 24,  7,
+                                         19, 27, 23,  6, 26,  5,  4, 31 };
+    val |= val >> 1;
```
Cyan4973: minor : 
(and mostly just an argument in favor of defensive programming).
I note you haven't added an `assert(val != 0)` here,
probably because it's already in `ZSTD_highbit32()` below.

But the point of adding `assert()` is not just that it works, today,
but that it's more difficult, in the future, to make modifications or bad usage of this function.

In this case, if you don't `assert(val != 0)`,
it implicitly means that this function could work with `val == 0`, 
irrespective of the fact that it's never called in this configuration ... yet.
And by the way, it superficially does : if `val == 0`, this function cleanly returns `0`, so it could seem that everything works as intended. There will be consequences in later parts of the code, and it might take a while to get back to the source of the issue.

Recommendation : even though the function is never called with `val == 0`, it should `assert()` it inside its own body, as the function could be re-employed in some different context in the future.
Also, you may want to add a code comment to make this condition clear, or explain why it exists (which is something the `assert()` alone cannot do).

lib/common/bits.h
terrelln: @embg I think we discussed just making this `__GNUC__ >= 4` for simplicity? Maybe the 3 is a typo?

Or add a helper macro `#define ZSTD_GNUC_SUPPORTS_CLZ_CTZ (defined(__GNUC__) && (__GNUC__ >= 4 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 4)))`

lib/common/bits.h
embg: Agreed, I added this assert on my own fork -- will force-push here once I get that update to pass CI.

lib/common/bits.h
embg: I looked at the approach you took in your PR and I see what you're talking about now. In your software fallback for countTrailingZeros64, you insert two calls to countTrailingZeros32. This improves performance for MSVC in 32-bit mode. I will adopt your approach and do the same thing -- this will improve the performance of VecMask_next compared to the current state of the code. Thanks for pointing this out!

ghost:
Another refactoring solution is to create an `intrinsics.h` file (not `bits.h`), and put all intrinsics wrappers in it.

For example, some architectures can use 16-byte memcpy:
https://github.com/facebook/zstd/blob/dev/lib/common/zstd_internal.h#L186-L189

Maybe SIMD wrappers can be put in it as well.
`intrinsics.h` will be very friendly to support new architectures in the future.

embg:
> Another refactoring solution is to create an `intrinsics.h` file (not `bits.h`), and put all intrinsics wrappers in it.
> 
> For example, some architectures can use 16-byte memcpy: https://github.com/facebook/zstd/blob/dev/lib/common/zstd_internal.h#L186-L189
> 
> Maybe SIMD wrappers can be put in it as well. `intrinsics.h` will be very friendly to support new architectures in the future.

@animalize Thank you for the suggestion. I think including intrinsics beyond bitwise functions would go beyond the scope of this PR -- indeed, I think it is difficult to say where we would even draw the line with a refactor like that, since some intrinsics are used inline in larger #ifdef blocks (such as x86-64 blocks in zstd_lazy.c). It's not obvious to me whether replacing all of those inline calls with wrapper functions is worthwhile.

lib/decompress/huf_decompress.c
embg: Was unable to add `ZSTD_STATIC_ASSERT(sizeof(size_t) == 8)` because this code is gated at runtime (not compile time) for x32 ABI. Added a runtime assert to make sure we are still validating the precondition.

lib/common/bits.h
```diff
@@ -34,16 +37,11 @@ MEM_STATIC unsigned ZSTD_highbit32(U32 val)   /* compress, dictBuilder, decodeCo
...
 #       else
-            if (val != 0) {
-                unsigned long r;
-                _BitScanReverse(&r, val);
-                return (unsigned)r;
-            } else {
-                /* Should not reach this code path */
-                __assume(0);
-            }
+            unsigned long r;
```
embg: Note: I removed `if (val != 0)` for consistency with the GNUC branch (there is an `assert(val != 0)` at the top of the function). If this is unsafe for some reason (we are worried about security when the assert is compiled out or something), please let me know and I'll add a commit reverting the change.

Cyan4973: While I believe the `assert()` is correct for this situation,
I also suspect that the `if (val != 0)` was added by a 3rd party contributor
because without it, some windows static analyzer would complain in production mode (essentially assuming the `val == 0` is a possible scenario).
Mentioning it because the issue might come back after this change.

embg: @Cyan4973  Makes sense, I'll revert the change. It only affects Win32 anyway.

Cyan4973: > @Cyan4973 Makes sense, I'll revert the change. 

Not necessarily. 
I think the message here is "how can we find a way to convince the static analyzer that this situation doesn't happen".
The `if (val !=0)` is one of them, but as you mentioned, it's bound to have an impact on performance.

To study that, you will have to find a static analyzer that complains about it, if that still exists (it was likely a false positive, and a newer version may no longer draw this conclusion). The most important one I can think of is the Visual Studio embedded static analyzer, which is not so bad. I even believe we used to enable it in CI for Visual compilation tests in debug mode.

embg:
Addressed all comments and added tests. Also improved bits.h a bit (re-implemented ZSTD_highbit32 so it just wraps ZSTD_countLeadingZeros32). Once I get a stamp I'll just need to benchmark and then should be good to merge!

lib/common/bits.h
embg: I think for this PR it makes sense to keep the current implementation (just to keep the scope of this PR manageable). Since it only affects Win32, and the performance cost is just a single 100%-predictable branch, it may not be worth the time to figure out a less expensive way to appease the static analyzers. But if someone wants to look into this it would definitely be welcome.

lib/common/bits.h
```diff
@@ -0,0 +1,171 @@
...
+    assert(val != 0);
+    {
+        unsigned result = 0;
+        while ((val & 0x80000000) == 0) {
+            result++;
+            val <<= 1;
+        }
+        return result;
+    }
+}
```
terrelln: Is there a better fallback for this function?

It should be possible because `BIT_highbit32()` has a seemingly more efficient fallback.

I don't think it is critical, and could be done in a follow-up. But it would be nice to not regress, if anyone uses it.

tests/fuzzer.c
```diff
@@ -3375,6 +3375,62 @@ static int basicUnitTests(U32 const seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : testing bitwise instrinsics PR#3045: ", testNb++);
+    {
+        U32 seed_32 = seed == 0 ? seed + 1 : seed; // these intrinsics are undefined on 0
+        U64 seed_64 = (U64)seed * 0x87654321; // decent 64-bit distribution
```
terrelln: nit: Use `FUZ_rand()`

```
U32 rand32 = FUZ_rand(&seed);
if (rand32 == 0) rand32 = 1;
U32 rand64 = ((U64)FUZ_rand(&seed) << 32) | FUZ_rand(&seed);
if (rand64 == 0) rand64 = 1;
```

tests/fuzzer.c
```diff
@@ -3375,6 +3375,62 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        U32 lowbit_only_32 = 1;
+        U64 lowbit_only_64 = 1;
+        U32 highbit_only_32 = (U32)1 << 31;
+        U64 highbit_only_64 = (U64)1 << 63;
+
+        /* Test ZSTD_countTrailingZeros32 */
+        CHECK_EQ(ZSTD_countTrailingZeros32(lowbit_only_32), 0u);
+        CHECK_EQ(ZSTD_countTrailingZeros32(highbit_only_32), 31u);
+        CHECK_EQ(ZSTD_countTrailingZeros32(seed_32), ZSTD_countTrailingZeros32_fallback(seed_32));
```
terrelln: I'd also love to see these fallback equality tests for a loop of values. E.g. just run over values 1-1000 in a loop, and make sure we're returning the right values.

terrelln (commented):
This looks amazing, and is a welcome cleanup + bug fix!

Just 1 question and a few nits about the tests.

lib/common/bits.h
embg: Good point, I'll just use the highbit fallback with a ^31 added to the end.

lib/common/bits.h
embg: Yup I changed it to >= 4 in the most recent version :)

embg:
## Benchmarking
I tested levels 1-19 on silesia.tar on an i7-9700K CPU. I compiled with gcc and with clang. My results are available [here](https://gist.github.com/embg/3ef8b119a2ff6486dd29979fd362b960) and generally don't seem to show a significant difference (on informal inspection) between embg/zstd/dev and embg/zstd/asm_refactor. 

When I ran the gcc test I noticed that asm_refactor results seemed to be consistently 1-2% higher than dev. To see if this effect was real, I did 15 interleaved iterations of silesia.tar at level 3 with gcc compilation. I used [this tool](https://www.estimationstats.com) to produce [estimation plots](https://www.estimationstats.com/#/background) for compression and decompression. The black vertical bar at the right is a 95% confidence interval for the mean difference across re-samples of the data.

The plots show no significant difference between dev and asm_refactor (at least at n=15 iterations). These data provide sufficient evidence to merge (in my view) but if we want to obtain smaller confidence intervals, I am happy to re-run the experiments with larger n (will just have to modify the benchmarking tool a bit to print out CSV data).

Decompression:
output-onlinepngtools (1)

Compression (forgot y-axis label, it should be "compression speed (MB/s)"):
output-onlinepngtools

embg:
Replaced XOR with subtraction following https://github.com/facebook/zstd/issues/3067#issuecomment-1042170165. Re-ran benchmark for gcc and got the same results as before: https://gist.github.com/embg/5618fb1a0af5d453b6feee32364ed31b

terrelln (approved):
LGTM!

Cyan4973 (approved):

Cyan4973:
Great result presentation !

embg:
> It's a minor detail, but we could probably have this `31 - X` operation done directly into the `DeBruijnClz` lookup table.

@davidbolvansky Will look into this in a few weeks, if benchmarks show benefit from XOR we can go back to that implementation. Moving work into the DeBruijn table can eliminate some of the subtractions but not all.

PR #3050, opened 2022-02-04
Select legacy level for cmake

shadchin:
I want to be able to override the default value from the command line, for example `cmake -DZSTD_LEGACY_SUPPORT:BOOL=ON -DZSTD_LEGACY_LEVEL:INT=3`


Cyan4973 (approved):
Yes, that seems a useful capability

terrelln (approved):

PR #3052, opened 2022-02-07
Keep original file if -c or --stdout is given

dirkmueller:
Set removeSrcFile back to false when -c or --stdout is used to improve
compatibility with gzip(1) behavior.

gzip(1) is removing the original file on compression unless --stdout or
/-c is used. zstd is defaulting to keep the file unless --rm is used or
when it is called via a gzip symlink, in which it is removing by
default. Specifying -c/--stdout turns this behavior off.

### DIFF
--- a/programs/zstd.1
+++ b/programs/zstd.1
@@ -1,5 +1,5 @@
 .
-.TH "ZSTD" "1" "January 2022" "zstd 1.5.2" "User Commands"
+.TH "ZSTD" "1" "February 2022" "zstd 1.5.2" "User Commands"
 .
 .SH "NAME"
 \fBzstd\fR \- zstd, zstdmt, unzstd, zstdcat \- Compress or decompress \.zst files
@@ -165,7 +165,7 @@ Additionally, this can be used to limit memory for dictionary training\. This pa
 \fB\-f\fR, \fB\-\-force\fR: disable input and output checks\. Allows overwriting existing files, input from console, output to stdout, operating on links, block devices, etc\.
 .
 .IP "\(bu" 4
-\fB\-c\fR, \fB\-\-stdout\fR: write to standard output (even if it is the console)
+\fB\-c\fR, \fB\-\-stdout\fR: write to standard output (even if it is the console); keep original files unchanged\.
 .
 .IP "\(bu" 4
 \fB\-\-[no\-]sparse\fR: enable / disable sparse FS support, to make files with many zeroes smaller on disk\. Creating sparse files may save disk space and speed up decompression by reducing the amount of disk I/O\. default: enabled when output is into a file, and disabled when output is stdout\. This setting overrides default and can force sparse mode over stdout\.
--- a/tests/cli-tests/compression/basic.sh
+++ b/tests/cli-tests/compression/basic.sh
@@ -24,6 +24,10 @@ zstd -c file       | zstd -t
 zstd --stdout file | zstd -t
 println bob | zstd | zstd -t
 
+# Test keeping input file when compressing to stdout in gzip mode
+$ZSTD_SYMLINK_DIR/gzip -c file       | zstd -t ; test -f file
+$ZSTD_SYMLINK_DIR/gzip --stdout file | zstd -t ; test -f file
+
 # Test --rm
 cp file file-rm
 zstd --rm file-rm; zstd -t file-rm.zst
--- a/programs/zstd.1.md
+++ b/programs/zstd.1.md
@@ -212,7 +212,7 @@ the last one takes effect.
     disable input and output checks. Allows overwriting existing files, input
     from console, output to stdout, operating on links, block devices, etc.
 * `-c`, `--stdout`:
-    write to standard output (even if it is the console)
+    write to standard output (even if it is the console); keep original files unchanged.
 * `--[no-]sparse`:
     enable / disable sparse FS support,
     to make files with many zeroes smaller on disk.
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -178,7 +178,7 @@ static void usage_advanced(const char* programName)
     DISPLAYOUT( "Advanced arguments : \n");
     DISPLAYOUT( " -V     : display Version number and exit \n");
 
-    DISPLAYOUT( " -c     : write to standard output (even if it is the console) \n");
+    DISPLAYOUT( " -c     : write to standard output (even if it is the console), keep original file \n");
 
     DISPLAYOUT( " -v     : verbose mode; specify multiple times to increase verbosity \n");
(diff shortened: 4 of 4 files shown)


programs/zstd.1
```diff
@@ -165,7 +165,7 @@ Additionally, this can be used to limit memory for dictionary training\. This pa
 \fB\-f\fR, \fB\-\-force\fR: disable input and output checks\. Allows overwriting existing files, input from console, output to stdout, operating on links, block devices, etc\.
 .
 .IP "\(bu" 4
-\fB\-c\fR, \fB\-\-stdout\fR: write to standard output (even if it is the console)
```
Cyan4973, line 168: We typically do our `man` modifications into `zstd.1.md`, which then is used to automatically generate `zstd.1`.
I recommend you ported your doc change onto `zstd.1.md`, otherwise it will be lost next time `zstd.1` is regenerated.

Cyan4973:
Thanks @dirkmueller, I believe that's a good suggestion.

One missing aspect in this PR could be to add a test validating the new behavior, and ensuring it remains respected in future changes.
That's something we could also do ourselves if you don't feel comfortable to.

programs/zstd.1
dirkmueller: good catch, I did not notice that. moved the change to .md and regenerated .1.

dirkmueller:
@Cyan4973 Thank you for the quick review. I added a testcase to `tests/cli-tests/compression/basic.sh`. let me know if that works for you.

tests/cli-tests/compression/basic.sh
```diff
@@ -24,6 +24,10 @@ zstd -c file       | zstd -t
 zstd --stdout file | zstd -t
 println bob | zstd | zstd -t
 
+# Test preservation in gzip mode
+(exec -a gzip zstd -c file)       | zstd -t
```
Cyan4973: I believe `exec -a` is a `bash` thing, not a `sh` one.
We try to maintain `sh` compatibility to improve portability of the test suite.

Possible workaround: create a `ln -s` symlink ?

tests/cli-tests/compression/basic.sh
```diff
@@ -24,6 +24,10 @@ zstd -c file       | zstd -t
...
+(exec -a gzip zstd --stdout file) | zstd -t
```
Cyan4973: These tests are titled "preservation in `gzip` mode".
However, both tests only check that output of `--stdout` or `-c` get properly decoded,
not explicitly that the original file is preserved.
I believe it could be paired with an explicit `test -f` test.

tests/cli-tests/compression/basic.sh
dirkmueller: that appears to be the only way unless one of the methods listed here is more acceptable: https://unix.stackexchange.com/questions/250681/is-there-a-posix-way-of-setting-zeroth-argument-of-a-target-application

terrelln: @dirkmueller I just fixed this in PR #3055

Can you please add this test to a new test case in `zstd-symlinks/gzip.sh`?

Then the equivalent test case would be:

```
#!/bin/sh
set -e
ln -s $(which zstd) ./gzip

./gzip -c file | zstd -t
test -f file
./gzip --stdout file | zstd -t
test -f file

```

You'd also need to add this line to `zstd-symlinks/setup`:

```
datagen > file
```

terrelln: Also, you should test that the file still exists explicitly, with `test -f file`.

tests/cli-tests/compression/basic.sh
dirkmueller: added `test -f` to make it more explicit.

tests/cli-tests/compression/basic.sh
```diff
@@ -24,6 +24,12 @@ zstd -c file       | zstd -t
 zstd --stdout file | zstd -t
 println bob | zstd | zstd -t
 
+# Test keeping input file when compressing to stdout in gzip mode
+ln -s "$(type -p zstd)" ./gzip
+./gzip -c file       | zstd -t ; test -f file
```
Cyan4973: minor :
if you want to test that both sides of the test are completing fine, prefer `&&` instead of `;` separator.

```
> false ; echo this is successful
this is successful
> false ; true
> echo $?
0

> false && echo this is successful

> false && true
> echo $?
1
```

dirkmueller: Please test with `set -e` like these test scripts are using. In my experience the behavior of that with `&&` is odd - `&&` similar to `if` and `while` will temporarily turn off errexit mode and simply hide the error. try this example script:

```
$ cat x.sh 
set -e

false | false && true && echo it detected the failure
true && echo "did it really find the false?"
$ sh x.sh ; echo $?
did it really find the false?
0
```

The behavior we want that the first false return aborts the script (because if zstd -c fails for whatever reason, test -f will also remain trivially true)

dirkmueller:
actually I realize after https://github.com/facebook/zstd/commit/169f8c11fff94c7cb2280efb1b5ae3c8fb3a7c5c I can remove the gzip symlinking logic and just call `gzip` directly.

terrelln:
@dirkmueller You still need the symlink logic. I believe that this is calling the system gzip.

I've avoided putting the zstd gzip symlink in `cli-tools/bin` so that it is still possible to call the system gzip.

dirkmueller:
@terrelln you're right. I've discovered `$ZSTD_SYMLINK_DIR` and changed it to use that. how about that?

terrelln (commented):
Sorry for the delay, and thanks for the ping!

@Cyan4973 are you happy if `zstd --rm -c` keeps the file, and `zstd -c --rm` removes the file?

If Yann is okay with that, can you also add a test to codify that behavior? E.g.

```
# Test -c overrides --rm
zstd --rm -c file | zstd -t; test -f file

# Test --rm overrides -c
cp file file-rm
zstd -c --rm file-rm | zstd -t; test ! -f file-rm
```

Then this PR looks good to me!

dirkmueller:
@terrelln tests updated accordingly. I think this behavior is no different than using `--rm` and  `-k` . I was a bit hesitant to hardcode this behavior in tests as afaik no user is relying on this behavior. we can still change the semantics. 

However now the tests cover that case as you suggested.

dirkmueller:
hey y'all, I'd like to move this patch forward. anything I can do to get another round of review? @terrelln

Cyan4973 (approved):
This looks good to me.
It's indeed proper behavior when combined with `gzip` symlink.

dirkmueller:
@terrelln  are you good with the change as well?

terrelln (approved):

terrelln:
Thanks for the PR @dirkmueller!

Issue #3037, opened 2022-01-30
better gzip(1) command line option compatibility: Support -l for gzip files


dirkmueller:
**Is your feature request related to a problem? Please describe.**
I'm experimenting with the "gzip" symlink on zstd binary and notice that there these compatibility options missing:

-n  --no-name When compressing, do not save the original file name and time stamp by default. (The original name is always saved if the name had to be truncated.)

-l --list For each compressed file, list the following fields:
compressed size: size of the compressed file uncompressed size: size of the uncompressed file ratio: compression ratio (0.0% if unknown) uncompressed_name: name of the uncompressed file

While listing is available as an option, it just aborts with an error message:

```
zstd --format=gzip -l *gz
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
File "test-0.1.0.tar.gz" not compressed by zstd
```

**Describe the solution you'd like**

provide more drop-in compatibility when used via a symlink to "gzip" with the gzip command line options, as documented
here (and elsewhere): https://linux.die.net/man/1/gzip

terrelln:
The only suggestion left from this pr is support for `-l` with gzip files. I think that is a reasonable request, but it isn't super high priority.

dirkmueller:
Might take a look at it over the holidays.

PR #3059, opened 2022-02-08
Implement more gzip compatibility (#3037)

dirkmueller:
-n --no-name is the current behavior already, so we can implement
this as a noop.

--best is an alias for -9 in gzip

add basic cli tests.

### DIFF
--- a/programs/zstd.1
+++ b/programs/zstd.1
@@ -217,6 +217,17 @@ If input directory contains "\.\.", the files in this directory will be ignored\
 .
 .IP "" 0
 .
+.SS "gzip Operation modifiers"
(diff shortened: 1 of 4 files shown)


felixhandte (commented):
These options should be documented in the CLI help text and in the man page. Can you add descriptions?

(If you'd prefer we do that part, we can add them in a follow-up PR.)

dirkmueller:
@felixhandte Okay, wasn't sure as these options are just for better gzip compatibility and not exposed in standard zstd cli. anyway added.

dirkmueller:
is there anything I can do to move this forward?

Cyan4973 (approved):
This looks good to me

felixhandte (approved):
Thanks for the reminder. Let's ship it!

PR #3075, opened 2022-02-23
Use helper function for bit manipulations.

TocarIP:
We already have BIT_getLowerBits, so use it. Benefits are 2fold:
1) Somewhat cleaner code
2) We are now using bzhi instructions, when available. Performance
delta is too small for microbenchmarks, but avoiding load still helps
larger applications, by reducing data cache pressure.

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -162,6 +162,16 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


embg (approved):
Looks good thanks for your contribution!

@terrelln should I move BIT_getLowerBits into `bits.h` and add tests? (in a separate PR)

ghost:
I still recommend to consider putting all intrinsics wrappers into one file, which is very friendly to porting to new architectures.

After zstd is used by Linux kernel, it will be used on many platforms. This preparation can be done in advance.

PR #3078, opened 2022-03-01
More checks on numeric input for `zstd` CLI

### LINKED ISSUE #3070: `zstd` CLI accepts bogus values for numeric parameters
**Describe the bug**

CLI parameters which are intended to accept numerical values (including those which take on optional suffix to denote the magnitude of the value) do not check that the values passed are well-formed.

**To reproduce**

For example, try running the lastest version of `zstd` as shown below:

```sh
zstd --memory=32LB file.txt
```

**Expected behavior**

I expected for `zstd` to reject the value passed for `--memory` since "LB" isn't a meaningful unit (perhaps the user made a typo and intended to say "KB"). This issue extends to even further malformed inputs, for instance the below example is also accepted without raising an error:

```sh
zstd --memory=hello file.txt
```

IAL32:
Closes https://github.com/facebook/zstd/issues/3070

- Added check for a possible invalid combination of numeric values in `zstd` CLI
- Does not perfectly cover all invalid combinations, but it's better than before
- Added some tests
- Slightly changes `--memory`, `--memlimit` and `--memlimit-decompress` to only be used with `=` at the end of the parameter
- `memLimit` type changed from `unsigned` to `size_t` to be more in line with `srcSizeHint`, as both refer to memory values

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -312,23 +312,56 @@ static void errorOut(const char* msg)
     DISPLAY("%s \n", msg); exit(1);
 }
 
+/*! nextCharValidCheck() :
+ * @return 0 if next char maybe valid.
+ *  next char is necessarily one of the accepted arguments.
+ *  This method does not assure absolute validity of the argument, but covers
+ *  a large portion of wrong cases.
+ * @return 1 if next char is not an accepted character.
+*/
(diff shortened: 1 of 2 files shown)


Cyan4973:
Issue fixed in #3268

PR #3080, opened 2022-03-01
Enable STATIC_BMI2 for gcc/clang

TocarIP:
Some usage (e.g. BIT_getLowerBit) uses without checking for MSVC,
so enabling for clang gives a small performance boost.


Cyan4973 (approved):

embg:
The CI failures are probably unrelated, and https://github.com/facebook/zstd/pull/3082 should fix them. If you re-run checks on this PR those jobs shouldn't fail again.

Issue #3011, opened 2022-01-19
Migrate some Travis CI tests to Github Actions


embg:
Many of our Travis CI tests run on special hardware (like power pc) and cannot be migrated to GitHub actions (except possibly via QEMU, which isn't exactly equivalent). However, we have some Travis CI tests which run on AMD64 architecture. It's likely that at least some can be migrated to GitHub Actions, which moves us closer to our goal of eliminating dependency on Travis CI.

Follow-up to this issue: it's likely possible to migrate many Circle CI tests as well.

PR #3082, opened 2022-03-02
Fix CI failures by adding apt-get update to Github Actions

embg:
Following advice from [here](https://github.com/actions/virtual-environments/issues/675) and [here](https://serverfault.com/questions/1070642/ubuntu-package-revision-not-found-in-github-actions), add `sudo apt-get update` to hopefully avoid this CI error:
```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/glibc/libc6-dev-i386_2.31-0ubuntu9.2_amd64.deb  404  Not Found [IP: 52.250.76.244 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/glibc/libc6-x32_2.31-0ubuntu9.2_amd64.deb  404  Not Found [IP: 52.250.76.244 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/glibc/libc6-dev-x32_2.31-0ubuntu9.2_amd64.deb  404  Not Found [IP: 52.250.76.244 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
make[1]: *** [Makefile:337: apt-install] Error 100
make: *** [Makefile:355: libc6install] Error 2
make[1]: Leaving directory '/home/runner/work/zstd/zstd'
Error: Process completed with exit code 2.
```
The first time I ran CI on this commit (on a [PR](https://github.com/embg/zstd/pull/12) in embg/zstd) I got a weird failure on a different job, but the second time everything passed. I propose we land this for now (if it passes CI on facebook/zstd), and if the issue persists there are some other things in the first link above that we can try.

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -156,7 +156,6 @@ jobs:
       shell: 'script -q -e -c "bash {0}"'
       run: |
         sudo apt-get -qqq update
-        make arminstall
         make valgrindinstall
         make -C tests valgrindTest
         make clean
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -75,6 +75,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: ASan + UBSan + Test Zstd, 32bit mode
       run: |
+        sudo apt-get -qqq update
         make libc6install
         make -j uasan-test-zstd32 V=1
 
@@ -88,6 +89,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: gcc-8 + ASan + UBSan + Fuzz Test
       run: |
+        sudo apt-get -qqq update
         make gcc8install
         CC=gcc-8 FUZZER_FLAGS="--long-tests" make clean uasan-fuzztest
 
@@ -97,6 +99,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: ASan + UBSan + Fuzz Test 32bit
       run: |
+        sudo apt-get -qqq update
         make libc6install
         CFLAGS="-O3 -m32" FUZZER_FLAGS="--long-tests" make uasan-fuzztest
 
@@ -152,6 +155,8 @@ jobs:
     - name: valgrind + fuzz test stack mode    # ~ 7mn
       shell: 'script -q -e -c "bash {0}"'
       run: |
+        sudo apt-get -qqq update
+        make arminstall
         make valgrindinstall
         make -C tests valgrindTest
         make clean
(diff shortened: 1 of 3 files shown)


embg:
Well looks like Travis is failing... will investigate tomorrow.

Edit: [this link](https://travis-ci.community/t/then-sudo-apt-get-update-failed-public-key-is-not-available-no-pubkey-6b05f25d762e3157-in-ubuntu-xenial/1728) should fix the issue, will test tomorrow. Expect some spam from this PR since I can't test this on embg/zstd (Travis CI doesn't give free credits for public repos anymore).

(note to self: Travis debug instructions [here](https://docs.travis-ci.com/user/running-build-in-debug-mode/#legacy-repositories))

Edit: also used [this link](https://techoverflow.net/2018/04/30/how-to-fix-apt-keyexpired-1515625755/) to fix Travis (there were multiple keys that had to be updated).

embg:
Fixed most of the Travis issues, but there is one that apparently [can only be fixed](https://travis-ci.community/t/gpg-errors-with-ubuntugis/8608/2) by moving away from Trusty. So there are two options:
* Move away from Trusty
* Move the "apt-get update" out of the Makefile and have it be specific to Github Actions only (i.e. have Github Actions run "apt-get update" before "make arminstall"). **This is a hack that would allow us to continue using Trusty**, but should only be a temporary solution since we should really always run "apt-get update" before "apt-get install".

Here is the failing Travis job: https://app.travis-ci.com/github/facebook/zstd/jobs/561969471

Cyan4973:
The failing `travis CI` test is a `qemu` `ARM` 32-bit test.
This is the kind of test we should no longer need on Travis, and that we should replace with an equivalent test on Github Action.
The reason it's attached to `Trusty` is likely because we couldn't find a correct combination of flags to correctly compile `zstd` on this platform, resulting in something working fine implicitly on `Trusty` but failing on later platforms due to some changes of default, or something like that.

Bottom line : we would be better off removing that test from `travis CI`, and replace it with an equivalent test on Github Action, that _would not need to be run on `Trusty`_. What matters is `qemu`, and `ARM` 32-bit.

Oh, and btw, it seems we already have such a test : 
https://github.com/facebook/zstd/runs/5410450999?check_suite_focus=true

Just check it's "equivalent enough" to remove the `Trusty` one from `travis CI`.

embg:
> Just check it's "equivalent enough" to remove the `Trusty` one from `travis CI`.

Sounds good, so for the long term solution I'll get rid of the Trusty test as part of https://github.com/facebook/zstd/issues/3011 (sometime this half). Will land the hack fix for now to unbreak CI, and replace with permanent fix once https://github.com/facebook/zstd/issues/3011 is closed.

felixhandte (approved):
Ship it.

PR #3084, opened 2022-03-03
fix the assertion in readLinesFromFile

xry111:
When the file is not terminated by endline, readLineFromFile will append
a '\0' for the last line.  In this case pos + lineLength == dstCapacity.


Cyan4973 (approved):

embg:
The CI failures are probably unrelated, and https://github.com/facebook/zstd/pull/3082 should fix them. If you re-run checks on this PR those jobs shouldn't fail again.

xry111:
> The CI failures are probably unrelated, and #3082 should fix them. If you re-run checks on this PR those jobs shouldn't fail again.

So rebased and force-pushed, waiting for the result...

Cyan4973 (approved):

PR #2749, opened 2021-08-17
Pipelined Implementation of ZSTD_fast (~+5% Speed)

felixhandte:
This PR introduces a new implementation of the `ZSTD_fast` parser for single-segment compressions. This new match-finder achieves up to 5% speed improvements, and improves compression slightly on average.

## Description

If you squint hard enough (and ignore repcodes), the search operation at any given position is broken into 4 stages:

1. Hash (map position to hash value via input read)
2. Lookup (map hash val to index via hashtable read)
3. Load (map index to value at that position via input read)
4. Compare (determine whether a match exists)

Each of these steps involves a memory read at an address which is computed from the previous step. This means that for each position, these steps must be sequenced and their latencies are cumulative.

Originally, `ZSTD_fast` simply did each step sequentially:

```
Pos | Time -->
----|-----------------
N   | 1234
N+1 |     1234
N+2 |         1234
N+3 |             1234
```

In #1562, @terrelln changed the implementation to work on two positions at a time:

```
Pos | Time -->
----|-----------------
N   | 1 2 3 4
N+1 |  1 2 3 4
N+2 |         1 2 3 4
N+3 |          1 2 3 4
```

This PR changes to a different strategy of parallelizing work, and does approximately the following:

```
R = Repcode Read & Compare
H = Hash Position
T = HashTable Lookup
M = Match Read & Compare

Pos | Time -->
----+-------------------
N   | ... M
N+1 | ...   TM
N+2 |    R H   T M
N+3 |         H    TM
N+4 |           R H   T M
N+5 |                H   ...
N+6 |                  R ...
```

This is very much analogous to the pipelining of execution in a CPU, and has the same benefits and drawbacks. This approach appears to more successfully parallelize read latencies.

However, just like a CPU, we have to dump the pipeline when we find a match (take a branch). When this happens, we throw away our current state, record the match, and then do the following prep to re-enter the loop:

```
Pos | Time -->
----+-------------------
N   | H T
N+1 |  H
```

This is also the work we do at the beginning to enter the loop initially.

In addition to this broad rearchitecture, various implementation details are tweaked to coax the best performance possible. This includes:

- Rather than recalculate the next step size every time we need it, we keep track of it and simply increment it when we fail to find a match within a certain number of searches. This is equivalent to the previous approach, but slightly less work.
- This also gives us a good infrequent checkpoint outside the core loop which ends up being a good opportunity to do some prefetching of the input.
[...]

PR #3086, opened 2022-03-04
Software pipeline for ZSTD_compressBlock_fast_dictMatchState (+5-6% compression speed)

embg:
## Summary
Using similar techniques as https://github.com/facebook/zstd/pull/2749, improves compression speed by 5-6% on a dataset of HTML headers between 0 and 8KB (the dataset has ratio 2.32 w/o dictionary, 3.41 with dictionary). This PR doesn't specifically address the cold dictionary scenario, which is something I'd like to explore in the future, but it does improve cold dictionary performance (as well as hot dictionary performance, see plots below).

## Plan to merge
@Cyan4973 , @felixhandte , and I are in agreement regarding which measurements are required to merge this change. Because the full suite of measurements will take some work to produce, I am only including two experiments for now. Once I get all nits and feedback resolved, and have a final version of the code ready to merge, I will do the full suite and post results in this thread.

## Ratio regression
~~This change regresses compression ratio by 0.1% on the html8K dataset (4350234 bytes -> 4354095 bytes). This is because we no longer search the very last position. It may be possible to fix this without killing the perf win -- however I think @felixhandte struggled to do this for noDict, so I didn't attempt it for now. Let me know if this regression is a blocker.~~

**EDIT:** [6cba817](https://github.com/facebook/zstd/pull/3086/commits/6cba817c8f2eff8a387b17033a02ec88eab4648e) fixes the ratio regression (at least on html8K) with no performance cost! Thanks @felixhandte for pointing out this fix.

It is still possible that this PR will cause zstd not to search the last position -- this would occur in a situation where the matchfinder is in accelerated search mode when it hits the end of the input. However, across the entire html8K dataset, this change in the parsing strategy did not regress compressed size by even a single byte. So there is essentially no ratio tradeoff with this PR.

## Preliminary results
<details>
<summary>Cold dictionary</summary>
Screen Shot 2022-03-04 at 2 25 30 PM
</details>
<details>
<summary>Hot dictionary</summary>
Screen Shot 2022-03-04 at 2 32 39 PM
</details>

## Final results
I benchmarked on a Intel(R) Xeon(R) D-2191A CPU @ 1.60GHz CentOS machine with core isolation and turbo disabled. I measured all combinations of the following variables:
* Dataset: html8K vs. github-users
* Compiler: gcc 11.1.1 vs. clang 12.0.1
* Dictionary temperature: hot vs. warm vs. cold
    * Hot = 1 CDict, 0.1MB working set size (L2 cache). Warm = 100 CDicts, 8-15MB working set size (roughly L3 cache). Cold = 10000 CDicts, >1GB working set size (main memory).
* Dictionary size: 4K vs 32K vs 110K

I modified the [largeNbDicts](https://github.com/facebook/zstd/tree/dev/contrib/largeNbDicts) benchmark to append the max speed over 10 iterations to a CSV file. I ran each scenario 10 times interleaved ([bash script](https://gist.github.com/embg/a2b570e5180a64257e5f16c67826cb96)) to eliminate bias from background processes competing for main memory / L3 bandwidth. I then used the [bootstrap method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.bootstrap.html) to compute confidence intervals for the percent increase in speed, `(optimized / dev) - 1` ([python script](https://gist.github.com/embg/10cc75c9639382d715cf570e4cd292cc)).

### Summary of results
**Regressions**

_All scenarios are speed-positive except github-users with 4K dict_. The latter scenarios have small regressions, between 0.2% and 1.6% depending on dict temperature. This is due to the extremely high compression ratio on github-users (which reduced the win for all github-users dict sizes). Every time a match is found, the pipeline has to reset, killing the wins from pipelining when there are no matches found. **Most real-world datasets have much lower compression ratio, and most real-world dictionaries are much larger than 4K.** So the regression only affects an extremely unrealistic use-case.

Again: html8K was speed-positive across all scenarios. github-users with 32K and 110K dicts were speed-positive across all scenarios. This regression is only for github-users with 4K dict size.

**Wins**

_All html8K scenarios are 5%+ wins on gcc_. For html8K with 32K dictionary size, the gcc wins are 7%+ for all temperatures. The largest win I measured was `9.8% (± 0.2%)` on html_8K/cold/110K with gcc.

clang is speed positive across all scenarios, but wins are more like 1-3% (this makes sense, since I only measured gcc in my exploration).

github-users wins for 32K dict and 110K dict are also around 1-3%. As discussed above, this is due to the extremely high compression ratio of github-users, which is not representative of most real-world data.

### Full data
* [Percent wins for all scenarios](https://gist.github.com/embg/1b8244d709a0bd60a7f78e57244a4334)
* [Percent wins + raw speeds for all scenarios](https://gist.github.com/embg/c46019a12efda519088e4f063388f990)
    * Note: gcc and clang speeds for a single scenario can be compared, since those experiments were interleaved. Speeds across different scenarios cannot be compared, since those measurements were not interleaved (thus the background noise could have been different).
    
## Rejected optimizations
* **L1 prefetching the dictMatchIndex / matchIndex**. Together with pipelining, this was speed-positive for hot dict, but the win was very small. It basically killed the win from pipelining.
    * Theory: L1 prefetching tied up Line Fill Buffers, of which there are only 12 on Skylake Server, delaying more urgent reads / writes.
* **Various L2 prefetches**. These were speed-positive for cold dict, but huge regression for hot dict. I cannot land them without adding cold-dict detection for compression, which I plan to do in a future PR. L2 prefetches are thus out of scope for this PR.
* **[Optimize pipeline initialization after a match is found](https://github.com/embg/zstd/pull/13/commits/ba87837f8631af81b0156825193b99b61cdcf5fa)**. This ended up killing about half of the total pipelining win on html8K. Perhaps the extra branches outweighed the win from more extensive pipelining?
* **Replace ZSTD_count_2segments with ZSTD_count**. For some reason this was a clear loss. Not sure why, it should have been a small win. Didn't spend time to investigate. 

## Future work
I did not try deeper pipelines, or combining pipelining with loop unrolling (as we have in noDict). It is completely possible that either of those directions could lead to bigger wins. Since I hit the noDict wins with a less complex pipeline (and even exceeded them in some important scenarios), I did not try these more complex optimizations.

## Conclusion
There are big wins for html8K, and every realistic scenario I tested is speed-positive. Even for an unrealistic worst-case scenario (github-hot-4K-gcc), the speed regression was only 1.6%. The ratio regression has been addressed. This PR is ready to merge!

### DIFF
--- a/lib/compress/zstd_fast.c
+++ b/lib/compress/zstd_fast.c
@@ -401,14 +401,7 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
 
     /* loop variables */
     const BYTE* ip0 = istart;
-    const BYTE* ip1;
-    size_t hash0;
-    size_t hash1;
-    size_t dictHash; /* inside the do{} loop, this is the hash for ip1 */
-    U32 matchIndex;
-    U32 dictMatchIndex;
-    size_t mLength=0; /* initialize to avoid compiler warning, assert != 0 later on */
-    U32 curr;
+    const BYTE* ip1 = ip0 + stepSize;
 
     /* if a dictionary is still attached, it necessarily means that
      * it is within window size. So we just check it. */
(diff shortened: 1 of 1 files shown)


Cyan4973:
> This change regresses compression ratio by 0.1% on the html8K dataset

0.1% ratio loss for ~5% speed win
is a good enough ratio for fast compression modes.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
     /* dictMatchState repCode checks don't currently handle repCode == 0
      * disabling. */
     assert(offset_1 <= dictAndPrefixLength);
     assert(offset_2 <= dictAndPrefixLength);
 
-    /* Main Search Loop */
-    while (ip < ilimit) {   /* < instead of <=, because repcode check at (ip+1) */
+    /* Outer search loop */
+    assert(stepSize >= 1);
+    while (ip1 < ilimit) {   /* < instead of <=, because repcode check at (ip+1) */
```
felixhandte: Can't this be `<=`? The repcode check is still at `ip0 + 1`, not at `ip1 + 1`.

Would this alleviate the ratio regression?

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
-            offset_2 = offset_1;
-            offset_1 = offset;
-            ZSTD_storeSeq(seqStore, (size_t)(ip-anchor), anchor, iend, OFFSET_TO_OFFBASE(offset), mLength);
+
+            /* Prepare for next iteration */
+            dictMatchIndex = dictHashTable[dictHash];
+            matchIndex = hashTable[hash1];
+            ip0 = ip1;
+            ip1 = ip1 + ((ip1 - anchor) >> kSearchStrength) + stepSize;
+            if (ip1 >= ilimit) goto _cleanup;
```
felixhandte: Same question I guess, can this be `>`?

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+                                   base + repIndex;
+            size_t hash1 = ZSTD_hashPtr(ip1, hlog, mls);
+            dictHash = ZSTD_hashPtr(ip1, dictHLog, mls);
+            hashTable[hash0] = curr;   /* update hash table */
+
+            if (((U32) ((prefixStartIndex - 1) - repIndex) >=
+                 3) /* intentional underflow : ensure repIndex isn't overlapping dict + prefix */
+                && (MEM_read32(repMatch) == MEM_read32(ip0 + 1))) {
+                const BYTE *const repMatchEnd = repIndex < prefixStartIndex ? dictEnd : iend;
+                mLength = ZSTD_count_2segments(ip0 + 1 + 4, repMatch + 4, iend, repMatchEnd, prefixStart) + 4;
```
felixhandte, line 453: Possible (unrelated) experiment: unlike extDict (streaming) where it's reasonable to want to capture matches that span the prefix and the input, that doesn't really seem likely in DMS, where the prefix is not expected to continue smoothly into the input. So it might be cheaper to just do a `ZSTD_count` for what I would expect would be no loss of ratio.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
-                 && (ip[-1] == match[-1])) { ip--; match--; mLength++; } /* catch up */
-            offset_2 = offset_1;
-            offset_1 = offset;
-            ZSTD_storeSeq(seqStore, (size_t)(ip-anchor), anchor, iend, OFFSET_TO_OFFBASE(offset), mLength);
+
+            /* Prepare for next iteration */
+            dictMatchIndex = dictHashTable[dictHash];
+            matchIndex = hashTable[hash1];
+            ip0 = ip1;
+            ip1 = ip1 + ((ip1 - anchor) >> kSearchStrength) + stepSize;
```
felixhandte: We had talked about rewriting this expression a la bc768bc. Did you give that a try?

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+            if (ip1 >= ilimit) goto _cleanup;
+            curr = (U32)(ip0 - base);
+            hash0 = hash1;
         }
 
         /* match found */
-        ip += mLength;
-        anchor = ip;
+        assert(mLength);
```
felixhandte, line 509: Isn't `hash1` always calculated if you arrive here? If so, you may want to write `hashTable[hash1] = ip1` here like in 39f2491, since you've already done the work of hashing it. You may well see a (very small) ratio win.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+                    anchor = ip0;
                     continue;
                 }
                 break;
             }
         }
+
+        /* Prepare for next iteration */
+        assert(ip0 == anchor);
+        ip1 = ip0 + stepSize;
```
felixhandte, line 543: So did you try hoisting up the loop re-entry prep work into this stuff? How'd that turn out?

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,125 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+
+        /* Inner search loop */
+        while (1) {
+            const BYTE *match = base + matchIndex;
+            const U32 repIndex = curr + 1 - offset_1;
+            const BYTE *repMatch = (repIndex < prefixStartIndex) ?
+                                   dictBase + (repIndex - dictIndexDelta) :
+                                   base + repIndex;
+            size_t hash1 = ZSTD_hashPtr(ip1, hlog, mls);
+            dictHash = ZSTD_hashPtr(ip1, dictHLog, mls);
```
felixhandte: So you've moved the hashing one iteration earlier, and the hashtable reads up to the end of the previous iteration. Which is pretty close to the layout of #2749, so I guess that makes sense.

I am still curious what results you have to suggest that that's the optimal layout here though, and that a deeper or shallower pipeline isn't better. Maybe we can discuss. Or ideally, you would add a little write-up to the PR why you think this is the right approach.

felixhandte (commented):
Overall this looks really solid! Well done.

I left some comments and questions, which we can discuss sometime next week.

lib/compress/zstd_fast.c
embg: Good idea, I will try this.

lib/compress/zstd_fast.c
embg: Yes, unfortunately this killed about half of the overall win. Here is the exact code I measured: [ba87837](https://gist.github.com/embg/e626f080a03dd914db74960ffcd73bd4). I think the extra branch is why it didn't improve performance (I tested both warm and cold dict), but I didn't actually look at the perf counters to confirm.

lib/compress/zstd_fast.c
embg: Sorry I wasn't clear about this in the summary -- I don't know if this is optimal. I started with a very simple pipeline, made some modifications to get it to 5%+ win across warm and cold dict (matching the win from #2749), and as soon as I hit that target I put up this PR. Let's discuss offline how far we want to iterate before landing.

lib/compress/zstd_fast.c
embg: I hit 5% across warm and cold dict before I had a chance to try this. It's a very simple change so I can definitely measure it and include in this PR if it is a win (will do the same for your ZSTD_count proposal).

lib/compress/zstd_fast.c
embg: I'm worried about the cost of the write, but this is something I can measure very quickly so I'll try it out along with the other suggestions. Since it's non-dependent it might be free, and if so would be nice to have a small ratio win.

lib/compress/zstd_fast.c
embg: Ah good point. I think it will reduce the regression but not eliminate it (because of search acceleration in the inner loop). Will measure and update the PR summary if the regression is reduced.

lib/compress/zstd_fast.c
embg: I tried this out and got a small improvement on html8K compressed size (about -0.3%). However, I measured a regression in compressed size on github-users (about +0.4%). So I am not going to include this change.

I actually expected github-users to be a *bigger* win than html8K because there are so many matches. Perhaps the extra writes are causing collisions which outweigh the positive improvement.

lib/compress/zstd_fast.c
embg: Tried it out and it's a small performance regression. Will update PR summary with details.

lib/compress/zstd_fast.c
embg: This is a small perf win, will push commit to this branch in a moment. Details will be added to PR summary.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,134 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+                step++;
+                nextStep += kStepIncr;
+            }
+            ip0 = ip1;
+            ip1 = ip1 + step;
+            if (ip1 > ilimit) goto _cleanup;
+
+            curr = (U32)(ip0 - base);
+            hash0 = hash1;
         }
```
Cyan4973: minor : when there are giant nested loops, like here,
it is useful to add a comment at the end of the block
just to help the reader determine
what this end of block corresponds to.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,134 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+        const BYTE* nextStep = ip0 + kStepIncr;
+
+        /* Inner search loop */
+        while (1) {
+            const BYTE *match = base + matchIndex;
+            const U32 repIndex = curr + 1 - offset_1;
+            const BYTE *repMatch = (repIndex < prefixStartIndex) ?
+                                   dictBase + (repIndex - dictIndexDelta) :
+                                   base + repIndex;
+            size_t hash1 = ZSTD_hashPtr(ip1, hlog, mls);
```
Cyan4973: It seems this variable is not modified after initialization.
In such a case, `const` it.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,134 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
-                continue;
-            } else {
-                /* found a dict match */
-                U32 const offset = (U32)(curr-dictMatchIndex-dictIndexDelta);
-                mLength = ZSTD_count_2segments(ip+4, dictMatch+4, iend, dictEnd, prefixStart) + 4;
-                while (((ip>anchor) & (dictMatch>dictStart))
-                     && (ip[-1] == dictMatch[-1])) {
-                    ip--; dictMatch--; mLength++;
+        size_t hash0 = ZSTD_hashPtr(ip0, hlog, mls);
+        size_t dictHash = ZSTD_hashPtr(ip0, dictHLog, mls);
```
Cyan4973: This `dictHash` is only used once, in order to calculate `dictMatchIndex` just after.

Then, the name `dictHash` is re-employed in the following "inner search loop",
but they don't correspond anymore,
since the "inner search loop"'s `dictHash` is actually initialized at the beginning of the loop.
So they should be different variables.

Limiting variable's lifetime is an important part of keeping the code base readable.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,134 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+
+        /* Inner search loop */
+        while (1) {
+            const BYTE *match = base + matchIndex;
+            const U32 repIndex = curr + 1 - offset_1;
+            const BYTE *repMatch = (repIndex < prefixStartIndex) ?
+                                   dictBase + (repIndex - dictIndexDelta) :
+                                   base + repIndex;
+            size_t hash1 = ZSTD_hashPtr(ip1, hlog, mls);
+            dictHash = ZSTD_hashPtr(ip1, dictHLog, mls);
```
Cyan4973: This `dictHash` is only used once below, within the scope of "inner search loop",
and it's always initialized at the beginning of the loop,
so it's essentially different from the `dictHash` declared one level above.

Separate them.
Make this one `const`.

lib/compress/zstd_fast.c
```diff
@@ -415,101 +416,134 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
+        U32 dictMatchIndex = dictHashTable[dictHash];
+        U32 matchIndex = hashTable[hash0];
+        U32 curr = (U32)(ip0 - base);
+        size_t step = stepSize;
+        const size_t kStepIncr = 1 << kSearchStrength;
+        const BYTE* nextStep = ip0 + kStepIncr;
+
+        /* Inner search loop */
+        while (1) {
+            const BYTE *match = base + matchIndex;
```
Cyan4973: very minor : 
in this code base, note the preference for `const Type* var`, instead of `const Type *var` (position of the `*`).

Cyan4973:
As a minor coding style comment : 
a lot of variables have shifted from being `const` to being mutable in this PR.

This may seem a "minor event" (just one keyword difference), but it's not : in most cases, bugs involve mutable variables. When a variable is `const`, I know it doesn't change after initialization, and during a code review, I can mentally "forget" it after initialization. When it's mutable, it's a state, and I need to track its every change, to be sure the system behaves as intended. Multiply the nb of mutable variables, and the nb of states to keep track of becomes unwieldy.

This is a balance though. I'm not saying that mutable variables are forbidden. They are necessary. But because they add a "burden cost" to maintenance, adding one should be pondered with caution, and additional efforts to keep them `const` immutable are worthwhile (whenever applicable).

In this PR, I only found 2 cases where variables could be made `const`, so the vast majority remain mutable. No magical wand here. I'm merely using this box to emphasize the benefits of `const` for code maintenance.

lib/compress/zstd_fast.c
embg: Ya my editor did this to a bunch of lines, I manually fixed most of them. Will fix this one too. I will look into how to get the editor to stop formatting like this on paste.

lib/compress/zstd_fast.c
embg: This is a remnant from earlier prefetching experiments, where I pulled this out to prefetch the dictMatchIndex later in the function (using dictHash as a single variable). Thanks for pointing out that without the prefetches, it should really now be two variables.

felixhandte (approved):
LGTM!

Cyan4973 (approved):

embg:
Updated PR summary with results from full benchmarking suite and discussion of good and bad future directions for this work.

PR #3092, opened 2022-03-09
[doc] Add decompressor errata document

terrelln:
Add a document that lists the known bugs in the decoder where valid
frames are rejected, along with the version that they were fixed.


terrelln:
These are the only bugs I can remember, but let me know if there were any I missed.

Cyan4973 (approved):
Great documentation !

PR #3093, opened 2022-03-09
Handle newer less versions in zstdless testing

dirkmueller:
Newer less versions appear to have changed how stderr
and stdout are showing error messages. hardcode the
expected behavior to make the tests pass with any less version.
    
Also set locale to C so that the strings are matching.


terrelln (approved):

terrelln:
Thanks for the fix @dirkmueller!!

PR #3096, opened 2022-03-12
put </body> </html> in correct order

tansy:
In `doc/zstd-manual.html` closing tags are I opposite order - `</html>` precedes `</body>` which is, at least formally, incorrect. It will be rendered fine as they are very lays tags but they should be in correct order.


Cyan4973:
The issue is that `zstd-manual.html` is automatically generated.
So any direct fix into this file is short-lived, and will be automatically overwritten at next general update.

I presume the only way to fix that "cleanly" would be to fix the manual generator.

tansy:
> zstd-manual.html is automatically generated

I suspected that but couldn't find it. Well, now you know what to look for.

PR #2921, opened 2021-12-10
Stagger Stepping in Negative Levels

felixhandte:
#2749 returned the negative levels to pre-#1562 behavior. I.e., skipping `stepSize` positions every iteration. This PR emulates #1562-like stepping by applying the `stepSize` skip every other position.

This PR addresses #2827.

## Benchmarks

`silesia.tar` on `gcc-10`:

Screenshot from 2021-12-13 17-14-03

`silesia.tar` on `clang-13`:

Screenshot from 2021-12-13 17-14-15

## Status

I believe this PR is ready for merge.

PR #3114, opened 2022-04-19
Software pipeline for ZSTD_compressBlock_fast_extDict (+4-9% compression speed)

embg:
## Summary
Using similar techniques as https://github.com/facebook/zstd/pull/2749 and https://github.com/facebook/zstd/pull/3086, improves level-1 compression speed by 4-9% on a dataset of HTML headers between 8 and 16KB (the dataset has ratio 3.03 w/o dictionary, 4.92 with a 110KB dictionary). Dictionary compression speed improvements hold across gcc/clang and a variety of dictionary temperatures and sizes. Additionally, compression speed for level-1 streaming compression (i.e. `zstd --single-thread silesia.tar`) is improved by 4-5%.

Compression ratio is also slightly improved. For example, the compressed size of silesia.tar is reduced by 0.1%.

## Preliminary results
I have added final measurements below, preliminary measurements are still here if you are interested:
<details>

### Dictionary compression
I used the experimental setup described in https://github.com/facebook/zstd/pull/3086 (see the section titled "Final Results"). On the 4KB dictionary, speed improvements are 4-5% across all temperatures and compilers. On the larger dictionaries (more realistic), speed improvements are 6-9% across all temperatures and compilers.

<details>
<summary>Dictionary compression speed improvements at level 1 (preliminary)</summary>

```
/Users/embg/results0415/html_16K/cold/4K
	gcc11: 5.4% (± 0.0%)
	clang12: 4.3% (± 0.1%)

/Users/embg/results0415/html_16K/cold/32K
	gcc11: 8.5% (± 0.0%)
	clang12: 6.2% (± 0.1%)

/Users/embg/results0415/html_16K/cold/110K
	gcc11: 9.1% (± 0.1%)
	clang12: 6.6% (± 0.1%)

/Users/embg/results0415/html_16K/warm/4K
	gcc11: 5.6% (± 0.1%)
	clang12: 4.5% (± 0.1%)

/Users/embg/results0415/html_16K/warm/32K
	gcc11: 8.4% (± 0.1%)
	clang12: 6.4% (± 0.2%)

/Users/embg/results0415/html_16K/warm/110K
	gcc11: 8.2% (± 0.9%)
	clang12: 6.9% (± 1.0%)

/Users/embg/results0415/html_16K/hot/4K
	gcc11: 5.8% (± 0.1%)
	clang12: 4.6% (± 0.1%)

/Users/embg/results0415/html_16K/hot/32K
	gcc11: 8.4% (± 0.1%)
	clang12: 6.1% (± 0.1%)

/Users/embg/results0415/html_16K/hot/110K
	gcc11: 7.9% (± 0.1%)
	clang12: 5.7% (± 0.1%)
```
</details>

There were small improvements in compression ratio at all dictionary sizes. The compressed size of the full html16K dataset was reduced as follows: 0.18% with the 110KB dictionary, 0.22% with the 32KB dictionary, and 0.52% with the 4KB dictionary.

### Streaming compression
Compression speed was measured via `time ./zstd_bin -1 --single-thread silesia.tar`. I timed each binary a few times interleaved and took the best time for each one. I tested the same compilers as for dictionary compression (gcc11 and clang12) and used the same setup (same CPU, core isolation, turbo disabled).

|         | 460780f (dev) | 3536262 (this PR) | % improvement |
|---------|---------------|-------------------|---------------|
| gcc11   | 1.49 (s) user          | 1.41 (s) user             | 5.4%          |
| clang12 | 1.47 (s) user         | 1.41  (s) user          | 4.1%          |

The compressed size of silesia.tar is reduced by 0.17%. 
</details>

### Dictionary compression
I used the experimental setup described in https://github.com/facebook/zstd/pull/3086 (see the section titled "Final Results"). The final dictionary perf numbers are very close to the preliminary numbers, except the clang win is about 1% higher now on most scenarios. Gcc numbers are almost identical.

My follow-up commits on this PR did not touch ratio so that is unchanged. See the Preliminary Results section for a discussion of the ratio impacts.

<details>
<summary>Dictionary compression speed improvements at level 1 (final)</summary>

```	
results0502/html_16K/cold/4K
	gcc11: 5.5% (± 0.1%)
	clang12: 5.4% (± 0.1%)

results0502/html_16K/cold/32K
	gcc11: 8.5% (± 0.1%)
	clang12: 6.7% (± 0.1%)

results0502/html_16K/cold/110K
	gcc11: 9.2% (± 0.1%)
	clang12: 7.2% (± 0.1%)

results0502/html_16K/warm/4K
	gcc11: 5.6% (± 0.1%)
	clang12: 5.6% (± 0.1%)

results0502/html_16K/warm/32K
	gcc11: 8.5% (± 0.2%)
	clang12: 6.7% (± 0.2%)

results0502/html_16K/warm/110K
	gcc11: 8.5% (± 0.7%)
	clang12: 7.8% (± 1.3%)

results0502/html_16K/hot/4K
	gcc11: 5.7% (± 0.1%)
	clang12: 5.5% (± 0.0%)

results0502/html_16K/hot/32K
	gcc11: 8.4% (± 0.1%)
	clang12: 6.3% (± 0.2%)

results0502/html_16K/hot/110K
	gcc11: 8.0% (± 0.1%)
	clang12: 5.9% (± 0.1%)
```

</details>

### Streaming compression
User time numbers are for `zstd <LEVEL> --single-thread /mnt/ramdisk/silesia.tar -o /dev/null`

I used core isolation and disabled turbo. Benchmarking method was roughly “run many many times interleaved, take the lowest number that I’ve seen at least twice for each binary”. Speed variations were around 3% due to background load on the server. 

**Everything is speed AND ratio neutral/positive except fast level 3, which trades 2.7% speed regression for 2.4% ratio improvement** (very good tradeoff IMO). Also worth noting that fast level 2 has a significant ratio improvement for no cost in speed. Fast level 1 has a significant speed improvement but no meaningful change in ratio.

| Level:   | 460780 gcc11           | ac371b gcc11           | 460780 clang12         | ac371b clang12         |
|----------|------------------------|------------------------|------------------------|------------------------|
| -1       | 1.49 (s), 34.69% ratio | 1.41 (s), 34.63% ratio | 1.45 (s), 34.69% ratio | 1.39 (s), 34.63% ratio |
| --fast=1 | 1.35 (s), 41.10% ratio | 1.29 (s), 41.04% ratio | 1.33 (s), 41.10% ratio | 1.27 (s), 41.04% ratio |
| --fast=2 | 1.18 (s), 43.56% ratio | 1.19 (s), 42.68% ratio | 1.17 (s), 43.56% ratio | 1.17 (s), 42.68% ratio |
| --fast=3 | 1.08 (s), 45.77% ratio | 1.11 (s), 44.66% ratio | 1.06 (s), 45.77% ratio | 1.10 (s), 44.66% ratio |

## What didn't work
* **A simpler pipeline with only two positions (ip0/ip1)**. This is what I tried initially, following my DMS pipeline in https://github.com/facebook/zstd/pull/3086. gcc wins were well over 5%, but unfortunately clang speed regressed a few %. I tried a couple approaches to fix the clang regression (documented [here](https://github.com/embg/zstd/pull/16)), but none of them worked. In the end, I decided to rewrite the pipeline closer to @felixhandte's noDict implementation, and this fixed the clang regression.
* **Optimizing backwards match length calculation**. The current implementation calculates backwards match length byte by byte. This creates a ton of branch mispredictions whenever the backwards match length is non-zero. I tried to rewrite this code to look back 8 bytes at a time, reducing branch mispredictions. Unfortunately, the increased overhead seems to have slightly outweighed the benefit from better branch prediction, with speed regressions around 1%. My commits are saved [here](https://github.com/embg/zstd/pull/20).

## Future work
Compression speed could be optimized for cold dictionaries by prefetching the dictionary content starting from dictEnd. The hash table is warm because this is extDict, but in a cold dictionary situation the dictionary content is still cold. I did not pursue this optimization because my code does not have cold dictionary gating. I am not sure how prevalent cold extDict compression is, depending on the prevalence this might be worth pursuing.

### DIFF
--- a/lib/compress/zstd_fast.c
+++ b/lib/compress/zstd_fast.c
@@ -587,7 +587,7 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
     U32* const hashTable = ms->hashTable;
     U32 const hlog = cParams->hashLog;
     /* support stepSize of 0 */
-    U32 const stepSize = cParams->targetLength + !(cParams->targetLength) + 1;
+    size_t const stepSize = hasStep ? (cParams->targetLength + !(cParams->targetLength) + 1) : 2;
     const BYTE* const base = ms->window.base;
     const BYTE* const dictBase = ms->window.dictBase;
     const BYTE* const istart = (const BYTE*)src;
@@ -625,13 +625,17 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
     const BYTE* nextStep;
     const size_t kStepIncr = (1 << (kSearchStrength - 1));
 
-    (void)hasStep; /* not currently specialized on whether it's accelerated */
-
     DEBUGLOG(5, "ZSTD_compressBlock_fast_extDict_generic (offset_1=%u)", offset_1);
 
     /* switch to "regular" variant if extDict is invalidated due to maxDistance */
     if (prefixStartIndex == dictStartIndex)
         return ZSTD_compressBlock_fast(ms, seqStore, rep, src, srcSize);
+
+    {   U32 const curr = (U32)(ip0 - base);
+        U32 const maxRep = curr - dictStartIndex;
+        if (offset_2 >= maxRep) offset_2 = 0;
+        if (offset_1 >= maxRep) offset_1 = 0;
+    }
 
     /* start each op */
 _start: /* Requires: ip0 */
@@ -655,51 +659,44 @@ _start: /* Requires: ip0 */
     idxBase = idx < prefixStartIndex ? dictBase : base;
 
     do {
-        U32 mval; /* src or dict value at match idx */
+        {   /* load repcode match for ip[2] */
+            U32 const current2 = (U32)(ip2 - base);
+            U32 const repIndex = current2 - offset_1;
+            const BYTE* const repBase = repIndex < prefixStartIndex ? dictBase : base;
+            U32 rval;
+            if ( ((U32)(prefixStartIndex - repIndex) >= 4) /* intentional underflow */
+                 & (offset_1 > 0) ) {
+                rval = MEM_read32(repBase + repIndex);
+            } else {
+                rval = MEM_read32(ip2) ^ 1; /* guaranteed to not match. */
+            }
 
-        /* load repcode match for ip[2] */
-        const U32 current2 = (U32)(ip2 - base);
-        const U32 repIndex = current2 - offset_1;
-        const BYTE* const repBase = repIndex < prefixStartIndex ? dictBase : base;
-        U32 rval;
-        assert(offset_1 > 0);
-        if ( ( ((U32)(prefixStartIndex - repIndex) >= 4) /* intentional underflow */
-               & (offset_1 < current2 - dictStartIndex) ) ) {
-            rval = MEM_read32(repBase + repIndex);
-        } else {
-            rval = MEM_read32(ip2) ^ 1; /* guaranteed to not match. */
-        }
+            /* write back hash table entry */
+            current0 = (U32)(ip0 - base);
+            hashTable[hash0] = current0;
 
-        /* write back hash table entry */
-        current0 = (U32)(ip0 - base);
-        hashTable[hash0] = current0;
+            /* check repcode at ip[2] */
+            if (MEM_read32(ip2) == rval) {
+                ip0 = ip2;
+                match0 = repBase + repIndex;
+                matchEnd = repIndex < prefixStartIndex ? dictEnd : iend;
+                assert((match0 != prefixStart) & (match0 != dictStart));
+                mLength = ip0[-1] == match0[-1];
+                ip0 -= mLength;
+                match0 -= mLength;
+                offcode = REPCODE1_TO_OFFBASE;
+                mLength += 4;
+                goto _match;
+        }   }
 
-        /* check repcode at ip[2] */
-        if (MEM_read32(ip2) == rval) {
-            ip0 = ip2;
-            match0 = repBase + repIndex;
-            matchEnd = repIndex < prefixStartIndex ? dictEnd : iend;
-            assert((match0 != prefixStart) & (match0 != dictStart));
-            mLength = ip0[-1] == match0[-1];
-            ip0 -= mLength;
-            match0 -= mLength;
-            offcode = REPCODE1_TO_OFFBASE;
-            mLength += 4;
-            goto _match;
-        }
+        {   /* load match for ip[0] */
+            U32 const mval = idx >= dictStartIndex ? MEM_read32(idxBase + idx) : MEM_read32(ip0) ^ 1; /* guaranteed not to match */
 
-        /* load match for ip[0] */
-        if (idx >= dictStartIndex) {
-            mval = MEM_read32(idxBase + idx);
-        } else {
-            mval = MEM_read32(ip0) ^ 1; /* guaranteed to not match. */
-        }
-
-        /* check match at ip[0] */
-        if (MEM_read32(ip0) == mval) {
-            /* found a match! */
-            goto _offset;
-        }
+            /* check match at ip[0] */
+            if (MEM_read32(ip0) == mval) {
+                /* found a match! */
+                goto _offset;
+        }   }
 
         /* lookup ip[1] */
         idx = hashTable[hash1];
@@ -718,18 +715,14 @@ _start: /* Requires: ip0 */
         current0 = (U32)(ip0 - base);
         hashTable[hash0] = current0;
 
-        /* load match for ip[0] */
-        if (idx >= dictStartIndex) {
-            mval = MEM_read32(idxBase + idx);
-        } else {
-            mval = MEM_read32(ip0) ^ 1; /* guaranteed to not match. */
-        }
+        {   /* load match for ip[0] */
+            U32 const mval = idx >= dictStartIndex ? MEM_read32(idxBase + idx) : MEM_read32(ip0) ^ 1; /* guaranteed not to match */
 
-        /* check match at ip[0] */
-        if (MEM_read32(ip0) == mval) {
-            /* found a match! */
-            goto _offset;
-        }
+            /* check match at ip[0] */
+            if (MEM_read32(ip0) == mval) {
+                /* found a match! */
+                goto _offset;
+        }   }
 
         /* lookup ip[1] */
         idx = hashTable[hash1];
@@ -760,8 +753,8 @@ _cleanup:
      * them. So let's not. */
 
     /* save reps for next block */
-    rep[0] = offset_1;
-    rep[1] = offset_2;
+    rep[0] = offset_1 ? offset_1 : rep[0];
+    rep[1] = offset_2 ? offset_2 : rep[1];
 
     /* Return the last literals size */
     return (size_t)(iend - anchor);
@@ -808,11 +801,10 @@ _match: /* Requires: ip0, match0, offcode, matchEnd */
         hashTable[ZSTD_hashPtr(base+current0+2, hlog, mls)] = current0+2;  /* here because current+2 could be > iend-8 */
         hashTable[ZSTD_hashPtr(ip0-2, hlog, mls)] = (U32)(ip0-2-base);
 
-        assert(offset_2 > 0);
         while (ip0 <= ilimit) {
             U32 const repIndex2 = (U32)(ip0-base) - offset_2;
             const BYTE* const repMatch2 = repIndex2 < prefixStartIndex ? dictBase + repIndex2 : base + repIndex2;
-            if ( (((U32)((prefixStartIndex-1) - repIndex2) >= 3) & (offset_2 <= (U32)(ip0-base) - dictStartIndex))  /* intentional overflow */
+            if ( (((U32)((prefixStartIndex-1) - repIndex2) >= 3) & (offset_2 > 0))  /* intentional underflow */
                  && (MEM_read32(repMatch2) == MEM_read32(ip0)) ) {
                 const BYTE* const repEnd2 = repIndex2 < prefixStartIndex ? dictEnd : iend;
                 size_t const repLength2 = ZSTD_count_2segments(ip0+4, repMatch2+4, iend, repEnd2, prefixStart) + 4;
@@ -829,6 +821,11 @@ _match: /* Requires: ip0, match0, offcode, matchEnd */
     goto _start;
 }
 
+ZSTD_GEN_FAST_FN(extDict, 4, 1)
+ZSTD_GEN_FAST_FN(extDict, 5, 1)
+ZSTD_GEN_FAST_FN(extDict, 6, 1)
+ZSTD_GEN_FAST_FN(extDict, 7, 1)
+
 ZSTD_GEN_FAST_FN(extDict, 4, 0)
 ZSTD_GEN_FAST_FN(extDict, 5, 0)
 ZSTD_GEN_FAST_FN(extDict, 6, 0)
@@ -839,16 +836,30 @@ size_t ZSTD_compressBlock_fast_extDict(
         void const* src, size_t srcSize)
 {
     U32 const mls = ms->cParams.minMatch;
-    switch(mls)
-    {
-    default: /* includes case 3 */
-    case 4 :
-        return ZSTD_compressBlock_fast_extDict_4_0(ms, seqStore, rep, src, srcSize);
-    case 5 :
-        return ZSTD_compressBlock_fast_extDict_5_0(ms, seqStore, rep, src, srcSize);
-    case 6 :
-        return ZSTD_compressBlock_fast_extDict_6_0(ms, seqStore, rep, src, srcSize);
-    case 7 :
-        return ZSTD_compressBlock_fast_extDict_7_0(ms, seqStore, rep, src, srcSize);
+    assert(ms->dictMatchState == NULL);
+    if (ms->cParams.targetLength > 1) {
+        switch (mls) {
+            default: /* includes case 3 */
+            case 4 :
+                return ZSTD_compressBlock_fast_extDict_4_1(ms, seqStore, rep, src, srcSize);
+            case 5 :
+                return ZSTD_compressBlock_fast_extDict_5_1(ms, seqStore, rep, src, srcSize);
+            case 6 :
+                return ZSTD_compressBlock_fast_extDict_6_1(ms, seqStore, rep, src, srcSize);
+            case 7 :
+                return ZSTD_compressBlock_fast_extDict_7_1(ms, seqStore, rep, src, srcSize);
+        }
+    } else {
+        switch (mls) {
+            default: /* includes case 3 */
+            case 4 :
+                return ZSTD_compressBlock_fast_extDict_4_0(ms, seqStore, rep, src, srcSize);
+            case 5 :
+                return ZSTD_compressBlock_fast_extDict_5_0(ms, seqStore, rep, src, srcSize);
+            case 6 :
+                return ZSTD_compressBlock_fast_extDict_6_0(ms, seqStore, rep, src, srcSize);
+            case 7 :
+                return ZSTD_compressBlock_fast_extDict_7_0(ms, seqStore, rep, src, srcSize);
+        }
     }
 }
--- a/lib/compress/zstd_fast.c
+++ b/lib/compress/zstd_fast.c
@@ -603,7 +603,6 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
     const BYTE* const iend = istart + srcSize;
     const BYTE* const ilimit = iend - 8;
     U32 offset_1=rep[0], offset_2=rep[1];
-    U32 offsetSaved = 0;
 
     const BYTE* ip0 = istart;
     const BYTE* ip1;
@@ -616,7 +615,6 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
     size_t hash1; /* hash for ip1 */
     U32 idx; /* match idx for ip0 */
     const BYTE* idxBase; /* base pointer for idx */
-    U32 mval; /* src or dict value at match idx */
 
     U32 offcode;
     const BYTE* match0;
@@ -657,13 +655,13 @@ _start: /* Requires: ip0 */
     idxBase = idx < prefixStartIndex ? dictBase : base;
 
     do {
-        /* load repcode match for ip[2]*/
+        U32 mval; /* src or dict value at match idx */
+
+        /* load repcode match for ip[2] */
         const U32 current2 = (U32)(ip2 - base);
         const U32 repIndex = current2 - offset_1;
         const BYTE* const repBase = repIndex < prefixStartIndex ? dictBase : base;
         U32 rval;
-
-        /* load repcode match for ip[2] */
         assert(offset_1 > 0);
         if ( ( ((U32)(prefixStartIndex - repIndex) >= 4) /* intentional underflow */
                & (offset_1 < current2 - dictStartIndex) ) ) {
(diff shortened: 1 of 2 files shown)


lib/compress/zstd_fast.c
```diff
@@ -613,75 +633,200 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
...
-            const BYTE* const repMatchEnd = repIndex < prefixStartIndex ? dictEnd : iend;
-            size_t const rLength = ZSTD_count_2segments(ip+1 +4, repMatch +4, iend, repMatchEnd, prefixStart) + 4;
-            ip++;
-            ZSTD_storeSeq(seqStore, (size_t)(ip-anchor), anchor, iend, REPCODE1_TO_OFFBASE, rLength);
-            ip += rLength;
-            anchor = ip;
+        U32 rval;
+        assert(offset_1 > 0);
+        if ( ( ((U32)(prefixStartIndex - repIndex) >= 4) /* intentional underflow */
+               & (offset_1 < current2 - dictStartIndex) ) ) {
```
embg: Note: `offset_1 == current2 - dictStartIndex` is surprisingly possible in streaming compression, since the offset can be carried over from a previous noDict block. The fuzzers actually found a case where this happens. We need strict inequality to satisfy an assertion below.

More broadly, this predicate should be pulled into a helper function and documented (along with the same predicate in all other extDict / DMS matchfinders). I will do that in June or July. @terrelln

Cyan4973: It seems this condition is `unlikely()` to be false ?

lib/compress/zstd_fast.c
```diff
@@ -613,75 +633,200 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
...
+        current0 = (U32)(ip0 - base);
+        hashTable[hash0] = current0;
+
+        /* check repcode at ip[2] */
+        if (MEM_read32(ip2) == rval) {
+            ip0 = ip2;
+            match0 = repBase + repIndex;
+            matchEnd = repIndex < prefixStartIndex ? dictEnd : iend;
+            assert((match0 != prefixStart) & (match0 != dictStart));
+            mLength = ip0[-1] == match0[-1];
```
Cyan4973: Minor : coding style : 
I would recommend to add parenthesis here : 
`mLength = (ip0[-1] == match0[-1]);`

This makes the separation between test and assignment a bit easier to see.

Cyan4973:
Great results @embg !

You'll just have to update the regression tests using the new (improved) compression ratio of this PR.

I also have a few minor questions left, but that shouldn't change the outcome much. This PR is in good shape.

embg:
> Great results @embg !
> 
> You'll just have to update the regression tests using the new (improved) compression ratio of this PR.
> 
> I also have a few minor questions left, but that shouldn't change the outcome much. This PR is in good change.

Thanks! @felixhandte  and I looked at the regression test numbers and it seems like the negative levels might have some ratio regressions (haven't looked closely yet to see if they are serious or not). This is consistent with what Felix encountered while doing the noDict pipeline. I am going to produce graphs like Felix did in https://github.com/facebook/zstd/pull/2921 to see if the regression for negative levels is an acceptable speed-ratio tradeoff (that's how Felix was able to land the noDict pipeline).

lib/compress/zstd_fast.c
Cyan4973: My understanding is that this test went from `<=` to `<` because there is an unconditional look-back-by-one-byte later on. This should probably be mentioned. Alternatively, the "unconditional" nature of the "look-back-by-one-byte" operation could be revisited.

Cyan4973: Also, presuming that the case `offset_1 == current2 - dictStartIndex` can be taken care of,
I wonder if it is even possible for `offset_1 > current2 - dictStartIndex`,
and if it is not, should it be replaced by an `assert()`?

lib/compress/zstd_fast.c
Cyan4973: Looking back by one single byte is achieved here.
It's unconditional.
Why ? 
An `unlikely()` test could have been possible here, with probably minimal performance impact.

Also : it only looks back by one byte only, not 2+.
Here also, why limited to 1 ?

lib/compress/zstd_fast.c
```diff
@@ -613,75 +633,200 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
...
+            ip0 -= mLength;
+            match0 -= mLength;
+            offcode = REPCODE1_TO_OFFBASE;
+            mLength += 4;
+            goto _match;
+        }
+
+        /* load match for ip[0] */
+        if (idx >= dictStartIndex) {
+            mval = MEM_read32(idxBase + idx);
```
Cyan4973: minor : `mval` assignment could be ternary.
Also : `mval` lifetime should be smaller, local.
both combined : `mval` could be `const` for the (short) duration of its lifetime.

lib/compress/zstd_fast.c
felixhandte: It's designed this way to guarantee that we never accidentally emit a repcode with `litlen == 0` and unintentially change the meaning of the repcode from the one we intended. We could explore changing the structure of this check to be more thorough, but I think we should treat that as out-of-scope of this round of optimizations. I think with the reduced timeframe in which we want to land refactors of `{fast, dfast} * {extDict, dictMatchState}`, we should focus on a fairly direct translation of the `noDict` refactors.

lib/compress/zstd_fast.c
```diff
@@ -613,75 +633,200 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
...
+        } else {
+            mval = MEM_read32(ip0) ^ 1; /* guaranteed to not match. */
+        }
+
+        /* check match at ip[0] */
+        if (MEM_read32(ip0) == mval) {
```
Cyan4973: What about combining this branch with the previous one ?

felixhandte: For reasons I didn't try to understand when I was working on the `noDict` version, I found that this construct was faster.

felixhandte: I think it's because it gets assembled to something like:

```
int ival = MEM_read32(ip0);
int mval = ival ^ 1;
if (idx >= dictStartIndex) {
  mval = MEM_read32(idxBase + idx);
}
if (mval == ival) {
  goto _offset;
}
```

which I guess is pleasant for the CPU.

felixhandte: Ok kind of pointless investigation but I was curious:

Original C:

```
/* load match for ip[0] */
if (idx >= prefixStartIndex) {
    mval = MEM_read32(base + idx);
} else {
    mval = MEM_read32(ip0) ^ 1; /* guaranteed to not match. */
}

/* check match at ip[0] */
if (MEM_read32(ip0) == mval) {
    /* found a match! */
    goto _offset;
}
```

`gcc-11 -O3 -DNDEBUG`:

```
mov    (%rdi),%edx        // edx /* ival */ = *rdi /* ip0 */;
mov    %edx,%ecx          // ecx /* mval */ = edx /* ival */;
xor    $0x1,%ecx          // ecx /* mval */ ^= 1;
cmp    -0x74(%rsp),%eax   // flags = cmp(prefixStartIndex, eax /* idx */);
jb     3935c              // if (flags.less_than) {
mov    %eax,%ecx          //   ecx /* mval */ = eax /* idx */;
mov    (%r14,%rcx,1),%ecx //   ecx /* mval */ = r14 /* base */[rcx /* mval */]; }
cmp    %edx,%ecx          // flags = cmp(edx /* ival */, ecx /* mval */);
je     393a1              // if (flags.equal) goto _offset;
```

Interpretation:

```
ival = *ip0;
mval = ival;
mval ^= 1;
if (prefixStartIndex < idx) {
  mval = idx;
  mval = base[mval];
}
if (ival == mval) goto _offset;
```

The `mval = idx` is weird... But ok.

lib/compress/zstd_fast.c
embg: I think `offset_1 > current2 - dictStartIndex` is possible in streaming compression if using a sufficiently large dictionary. A very large offset could be carried over into a future extDict block which would have a smaller dictionary (because the "dictionary" in the future block is really just part of the window).

However, this case (along with the case `offset_1 == current2 - dictStartIndex`) could be addressed by checking the offset at the top of the extDict function, before entering the search loop. Once in the search loop, it's impossible to revert to that state. I'll try this in addition to marking the remaining condition UNLIKELY and see if there are any gains. If so, I can carry over this optimization to the other dictionary matchfinders.

lib/compress/zstd_fast.c
embg: I just followed the noDict approach here -- given the timeframe to land this, I'm not going to touch the mval assignment logic (based on Felix's analysis in the other comment thread it seems like the current code is generating pretty good assembly).

Edit: sorry I see what you're saying now, you're just proposing to put it in a block and make it const using a ternary. OK, I'll try that -- shouldn't really change codegen.

lib/compress/zstd_fast.c
Cyan4973: > I think offset_1 > current2 - dictStartIndex is possible in streaming compression if using a sufficiently large dictionary

Yes, that's a very good point.
And the following optimization suggestion is very good too.

lib/compress/zstd_fast.c
Cyan4973: > I see what you're saying now, you're just proposing to put it in a block and make it const using a ternary.

Yes, exactly.

embg:
Added a [commit](https://github.com/facebook/zstd/pull/3114/commits/809f65291266de966ba4262220992f5f7e7903a0) addressing comments so far:
* Optimizes the repcode predicate (checks for invalid offsets before entering the loop)
* Adds an optimized variant for the case `hasStep == 0`
* Fixes some cosmetic nits (scope of mval etc)

Comments I didn't address:
* Adding an `UNLIKELY()` macro -- will try this when I pull out the repcode check in a documented helper function (separate PR, June or July)
* Didn't add comments explaining the change in the repcode predicate to exclude `match0 == prefixStart`. Will explain this in the helper function which will have documentation for this predicate across all dictionary matchfinders.

Remaining steps to land:
* Address any remaining comments (@felixhandte  @Cyan4973  are there any?)
* Re-run previous measurements
* Measure speed-ratio tradeoff at negative levels, and if it is acceptable update the regression test csv file

Cyan4973:
> Comments I didn't address:
> 
> * Adding an `UNLIKELY()` macro -- will try this when I pull out the repcode check in a documented helper function (separate PR, June or July)

My bad,
I though this macro already existed within `zstd` code base, and one could just employ it.
But I `grep`ed and only found it as part of sub-module `xxhash`, not `zstd` proper.

No need to add that as part of this PR.

Cyan4973:
> * Address any remaining comments (@felixhandte  @Cyan4973  are there any?)

Only waiting for the documentation on negative compression levels, and then an update of the regression `csv` file.

Other than that, this PR seems in good shape and good to go.

lib/compress/zstd_fast.c
```diff
@@ -587,11 +587,10 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
     U32* const hashTable = ms->hashTable;
     U32 const hlog = cParams->hashLog;
     /* support stepSize of 0 */
-    U32 const stepSize = cParams->targetLength + !(cParams->targetLength);
+    size_t const stepSize = hasStep ? (cParams->targetLength + !(cParams->targetLength) + 1) : 2;
```
felixhandte: As discussed, make sure you benchmark this. It's only useful if it actually improves perf, and if it doesn't we shouldn't double all the varieties.

lib/compress/zstd_fast.c
```diff
@@ -605,85 +604,228 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
...
-                    ZSTD_storeSeq(seqStore, 0 /*litlen*/, anchor, iend, REPCODE1_TO_OFFBASE, repLength2);
-                    hashTable[ZSTD_hashPtr(ip, hlog, mls)] = current2;
-                    ip += repLength2;
-                    anchor = ip;
-                    continue;
-                }
-                break;
-    }   }   }
+        {   /* load match for ip[0] */
+            U32 const mval = idx >= dictStartIndex ? MEM_read32(idxBase + idx) : MEM_read32(ip0) ^ 1; /* guaranteed not to match */
```
felixhandte: Nit: although zstd doesn't have an official style guide, we try to respect some kind of line length convention (80 chars?). Wrap this?

lib/compress/zstd_fast.c
```diff
@@ -605,85 +604,228 @@ static size_t ZSTD_compressBlock_fast_extDict_generic(
...
+        ip0 = ip1;
+        ip1 = ip2;
+        ip2 = ip3;
+
+        /* write back hash table entry */
+        current0 = (U32)(ip0 - base);
+        hashTable[hash0] = current0;
+
+        {   /* load match for ip[0] */
+            U32 const mval = idx >= dictStartIndex ? MEM_read32(idxBase + idx) : MEM_read32(ip0) ^ 1; /* guaranteed not to match */
```
felixhandte: Nit: same.

felixhandte (commented):
Overall this looks solid! (I guess it's not surprising I would be fine with a refactor that mirrors one I wrote...)

As discussed, please provide final benchmark numbers.

lib/compress/zstd_fast.c
embg: Hardcoding hasStep by itself improves clang12 compression speed by 0.2% and doesn't affect gcc11. However, combining hasStep hardcoding with repcode-safety hardcoding improves gcc11 compression speed by 0.4% and clang12 by 0.9%. This is significant enough to justify a variant IMO. Here's the code I tested: https://gist.github.com/embg/e79fa67ab63db1e9634ce63e1af0aebe

The same idea can also be transferred to noDict in a separate PR, since it also has repcode safety checks which are unnecessary 99%+ of the time. The number of variants doesn't need to be increased since the "common case" can be folded into a single variant `(hasStep == 0) & repcodeOffsetsAreSafe`.
```
/home/embg/zstd/largeNbDicts_2c0540_clang12
110.0
109.9
109.9
/home/embg/zstd/largeNbDicts_2c0540_gcc11
106.3
106.3
106.3
/home/embg/zstd/largeNbDicts_a438e8e_clang12
110.9
110.9
111.0
/home/embg/zstd/largeNbDicts_a438e8e_gcc11
106.7
106.7
106.8
```

embg:
Addressed @felixhandte's comments in [518cb83](https://github.com/facebook/zstd/pull/3114/commits/518cb83833074d304dfcaa93cfc16039ea4683c8). The main update is hardcoding the repcode safety check in the variant where `hasStep == 0` (not that those two things are not directly connected, but they are both the "common scenario", so they should share one variant). See my measurements [here](https://github.com/facebook/zstd/pull/3114#discussion_r859147257) justifying the variant.

Will do full measurements and update the regression csv file tomorrow.

Edit: "directly" -> "not directly", typo

embg:
It seems like my last commit (with full implementation of the repcode variant + cosmetic stuff) regressed perf pretty significantly:
<details>
<summary>GCC</summary>

```
././largeNbDicts_353626_gcc11
148.7
148.5
148.6

././largeNbDicts_2820ef_gcc11
148.7
148.3
148.4

././largeNbDicts_809f65_gcc11
148.2
148.3
148.3

././largeNbDicts_518cb8_gcc11
146.4
146.5
146.4
```

</details>

<details>
<summary>Clang</summary>

```
././largeNbDicts_353626_clang12
150.5
150.5
150.6

././largeNbDicts_2820ef_clang12
150.6
150.7
150.6

././largeNbDicts_809f65_clang12
150.4
150.2
150.2

././largeNbDicts_518cb8_clang12
149.3
149.3
149.4
```

</details>

I will revert that now and put up a new commit tomorrow removing the hasStep variant (since the cost of the variant was only justified using numbers for hasStep + repcode).

Cyan4973:
I'm trying to understand the evolution of performance for negative compression levels,
and there seems to be something off.

When looking at the [updated regression test figures](https://github.com/facebook/zstd/pull/3114/files#diff-f7c7da2456f7b22b8e5f0a5e336c6cbab8dc7c4381c766a9ece1b28953694912),
a picture emerges for the compression ratio of negative compression levels : 
- `--fast=1` is similar (though not identical), typically very slightly worse (does not matter)
- the new `--fast=3` has a compression ratio roughly on par with old `--fast=5`, hence "worse", which doesn't matter if the speed of the new `--fast=3` is increases to become similar to the old `--fast=5`
- the new `--fast=5` is much worse. There is no "old" `--fast` level to compare to, but the gap is substantial. Here also, it's not a problem if speed increases by a similar amount.

Problem is, when looking at the report at the top of this PR, the story described is completely different.
I quote : 
> Everything is speed AND ratio neutral/positive except fast level 3, which trades 2.7% speed regression for 2.4% ratio improvement 

So, that's a completely different picture.

How do we reconcile ?

terrelln:
I only see that regression in the `results.csv` for the github files. If you look at silesia, you see neutral to positive ratio.

Since GitHub is so compressible, it could be very sensitive to small changes to the positions searched. If you look at `github.tar` level -5 vs. level -5 w/ dictionary, we can see that the dictionary still wins a little. I think it is likely that what is happening is that we have perturbed the positions we've searched slightly, and suddenly we're mostly searching at "useless" positions instead of "useful" positions. Though that is just a theory.

We should just be sure to measure compression ratio on several datasets to make sure that the github example is an outlier, and not the general case.

embg:
> How do we reconcile ?

@Cyan4973 Sorry for the miscommunication -- that quote is referring to the table below it, which only covers silesia (following the same ratio/speed tradeoff estimation method as https://github.com/facebook/zstd/pull/2921). I'll add a discussion of `results.csv` to the PR summary later today.

html16K and silesia are both ratio-positive at all levels (silesia we can see in results.csv, html16K I measured myself). github is also positive at level 1, but has a major regression at level -5 (also minor regressions at levels -1 and -2).

Why is github special? As @terrelln pointed out, it has extremely long matches, making it very sensitive to hash table collisions. Especially for dictionary compression, adding additional hash table writes can overwrite long matches in the dictionary, trading them for much less valuable indices in the prefix.

And in this PR we do add an [additional hash table write](https://github.com/facebook/zstd/pull/3114/files#diff-405263d039e2f37da95827b0724ff759a64cf00d4ef6e557acd1fcb869ffbd58R800), following @felixhandte's noDict PR. I did a quick experiment deleting that line of code, and the github dataset goes from major regression to major ratio win. However, github.tar is still a regression, though smaller than before.

Here are three `results.csv` files corresponding to dev, this PR with the extra write, and this PR without the extra write: https://gist.github.com/embg/064f088a0bf7e5585f6114f2ef5219d2 (all three files are in one gist). I used a regex to filter extraneous lines: `grep -E '(level 1|level \-[0-9]+)( |,)' results.csv`.

Deleting that write is unlikely to affect performance (since it's outside the hot loop), so this is just a question of which CSV file you prefer :). Just let me know if you want me to delete that extra hash table write and I'll quickly redo perf measurements, otherwise I'll land as-is.

embg:
@Cyan4973  Here are the numbers for html16K at negative compression levels with a 32K dictionary (this is for the code as-is, not the modification I mentioned in my previous comment):
* `--fast=1`: 25517077 bytes opt, 25554192 bytes dev
* `--fast=3`: 27725969 bytes opt, 31154011 bytes dev
* `--fast=5`: 30530273 bytes opt, 33781945 bytes dev

This is an 8-9% ratio win at levels -3 and -5. html16K is more representative of real-world data since it has a more realistic compression ratio (3-4x at level 1 depending on the dictionary size IIRC), so I would give this win more weight than the github loss.

Based on these numbers my vote would be to land the current version, but I understand if you prefer to focus on mitigating the github ratio regression instead (by deleting that extra hash table write). Just let me know which approach you prefer!

Cyan4973:
OK, that's fine. I think we are ready to accept that `github.tar` is a fairly specific scenario, with more impact in the randomness of searches during negative compression levels, which is not representative of other use cases.

There is a remaining issue in a FreeBSD test, but it seems unrelated to this PR.

Cyan4973 (approved):

felixhandte (approved):
Looks good to ship to me!

embg:
Thanks @felixhandte  @Cyan4973 for the detailed reviews! I will fix the FreeBSD infra failure sometime next week. That job passed on all commits before I updated `results.csv` so the failure is clearly unrelated, therefore I'll go ahead and merge.

nadavrot:
Nice!

Issue #3090, opened 2022-03-09
spec: 0-sized compressed blocks permitted by spec


klauspost:
**Describe the bug**

The current [decompression code](https://github.com/facebook/zstd/blob/dev/lib/decompress/zstd_decompress_block.c#L125) fails if the remainder of a block is sent to the literal decoder with only 2 bytes.

With 0 literals and 0 sequences that condition can be met.

While it is easy to argue that a block with such encoding is pointless (it should just be a raw block), I don't see any mention of it in the spec, and as far as I can tell it is conforming to spec. Ofc. I might have missed it.

Example: [2274d31e0d569fe9e31bedc4f9fddd9c9f114c2f.zst.gz](https://github.com/facebook/zstd/files/8215237/2274d31e0d569fe9e31bedc4f9fddd9c9f114c2f.zst.gz)

Cyan4973:
Interesting, we have a `harness` test program which is supposed to generate all sorts of valid blocks, and is a good tool to validate compliance of decoders.
It should have been able to generate such empty block.
We probably need to update this test program if it doesn't.

terrelln:
Thanks for the report!

We'll fix this bug in the next release.

We've had other bugs in the decoder that rejected valid frames. I will put up a PR shortly with a document that lists all the known cases where we've rejected a valid frame, an example frame, and the version we fixed the issue.

terrelln:
Actually, this block used to be invalid. Older versions of the spec had this clause:

> A Compressed_Block has the extra restriction that Block_Size is always strictly less than the decompressed size. If this condition cannot be respected, the block must be sent uncompressed instead (Raw_Block).

We removed that clause because the decoder didn't enforce it. And because it made certain topics (namely block splitting) a lot simpler. Since there are cases where you'd want to emit a small block that had large headers,  but you'd then re-use those headers in later blocks to amortize the cost.

However, it looks like we missed the edge case where compressed block is 0 sized.

Cyan4973:
> Actually, this block used to be invalid.

That explains why it wasn't part of `harness`. 
Indeed, subtle consequence of a small change !

terrelln:
This frame is the simplest test case. The only difference from the one you provided is that yours has the `Unused_Bit` of the frame header set to 1.

```
28b5 2ffd 2000 1500 0000 00
```

terrelln:
The one liner fix is to change `MIN_CBLOCK_SIZE` from `3` to `2`.

[...]

PR #3118, opened 2022-04-26
Intial commit to address 3090. Added support to decompress empty block.

udayanbapat:
As discussed with @terrelln, added a fix to address issue 3090 (https://github.com/facebook/zstd/issues/3090) to decompress minimum length 0 size block. As part of change MIN_CBLOCK_SIZE was changed from 3 to 2. However, since this macro was also used on the compression side, the logic was corrected by compensating with additional one. For clarity purposes, I have put a TODO comment and kept the logic as ..+1+1" instead of just "...+2".


udayanbapat:
I tested the fix manually as well as running 'make test' and 'make check' the newly added test passes correctly.

lib/decompress/zstd_decompress_block.c
```diff
@@ -234,10 +234,12 @@ size_t ZSTD_decodeLiteralsBlock(ZSTD_DCtx* dctx,
                     break;
                 case 1:
                     lhSize = 2;
+                    RETURN_ERROR_IF(srcSize<3, corruption_detected, "srcSize >= MIN_CBLOCK_SIZE == 2; here we need lhSize+1 = 3");
```
terrelln: We only read 2 bytes in this case.

```suggestion
```

lib/decompress/zstd_decompress_block.c
```diff
@@ -234,10 +234,12 @@ size_t ZSTD_decodeLiteralsBlock(ZSTD_DCtx* dctx,
...
                     litSize = MEM_readLE16(istart) >> 4;
                     break;
                 case 3:
                     lhSize = 3;
+                    RETURN_ERROR_IF(srcSize<4, corruption_detected, "srcSize >= MIN_CBLOCK_SIZE == 2; here we need lhSize+1 = 4");
```
terrelln: We only read 3 bytes in this case. You can see that we do our next bounds check on line 251, and don't read any other bytes before then.

```suggestion
                    RETURN_ERROR_IF(srcSize<3, corruption_detected, "srcSize >= MIN_CBLOCK_SIZE == 2; here we need lhSize = 3");
```

lib/decompress/zstd_decompress_block.c
```diff
@@ -280,12 +282,13 @@ size_t ZSTD_decodeLiteralsBlock(ZSTD_DCtx* dctx,
                     break;
                 case 1:
                     lhSize = 2;
+                    RETURN_ERROR_IF(srcSize<3, corruption_detected, "srcSize >= MIN_CBLOCK_SIZE == 2; here we need lhSize+1 = 3");
```
terrelln, line 284: This is correct. We have `+1` here because of the read of `istart[lhSize]` on lines 300 and 305. But in `set_basic` (above) we don't have the `+1`.

terrelln (changes requested):

lib/decompress/zstd_decompress_block.c
udayanbapat: Thanks!

terrelln (approved):

terrelln:
Thanks for the PR @udayanbapat!

Issue #3119, opened 2022-04-27
"./zstreamtest --newapi" test case fails on Windows


eli-schwartz:
In the Meson WrapDB integration CI, we try to run the build && test on all 3 major OSes. On Windows, one of the tests fails. Logs: https://github.com/mesonbuild/wrapdb/runs/6086663524?check_suite_focus=true

Relevant bits:

```
3/8 test-zstream-3           FAIL            21.14s   (exit status 3221225477 or signal 3221225349 SIGinvalid)
>>> MALLOC_PERTURB_=124 D:\a\wrapdb\wrapdb\_build\subprojects\zstd-1.5.2\build/meson/tests\zstreamtest.exe --newapi -t1 -T90s --no-big-tests
------------------------------------- 8< -------------------------------------
stderr:
Starting zstream tester (64-bits, 1.5.2)

Seed = 1643


     1/     1    
    34          
    37          
    41          
    45          
    47          
    60          
    78          
    82          
    95          
   102          
   132          
   133          
   134          
   143          
   145          
   150          
   159          
   171          
   196          
   201          
   202          
   209          
   216          
   225          
   249          
   257          
   263          
   267          
   272          
   275          
   287          
   297          
   310          
   313          
   318          
   339          
   348          
   356          
   362          
   374          
   381          
   413          
   414          
   433          
   458          
   484          
   491          
------------------------------------------------------------------------------
```

It is dying of a strange signal apparently. I've grepped around the source tree, and --newapi is only exercised by the Makefile and by the meson.build, and neither one is getting run on Windows-based CI jobs.

terrelln:
@eli-schwartz are you able to reproduce the error consistently?

eli-schwartz:
My only windows test environment is GitHub actions. In that environment I could consistently produce these results while doing debug iteration. (This is the same debug iteration in which I discovered the valgrind issue I was facing etc. in my other PR.)

eli-schwartz:
It was also reproducible at the end of January (again in the WrapDB github actions CI), as I mentioned this exact problem here: https://github.com/facebook/zstd/pull/3039#issuecomment-1025331166

Cyan4973:
I've tried to reproduce it.
On `macos` or Linux, I was unable to reproduce the same issue.

On Windows + Mingw64, I could finally reproduce it.
[...]

PR #3120, opened 2022-04-27
Meson test fixups

eli-schwartz:
- don't make running the valgrind tests mandatory
- mark known-broken test on Windows, to work around #3119

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -178,6 +178,30 @@ jobs:
           make clean
           CC=clang MOREFLAGS="-Werror -Wimplicit-fallthrough -O0" make -C lib -j libzstd.a ZSTD_LEGACY_SUPPORT=0
 
+  meson-linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install packages
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install build-essential python3-pip ninja-build liblz4-dev
+          pip install --pre meson
+      - name: Build with Meson
+        run: |
+          meson setup \
+            --buildtype=debugoptimized \
+            -Db_lundef=false \
+            -Dauto_features=enabled \
+            -Dbin_programs=true \
+            -Dbin_tests=true \
+            -Dbin_contrib=true \
+            -Ddefault_library=both \
+            build/meson builddir
+          ninja -C builddir/
+          meson test -C builddir/ --print-errorlogs
+          meson install -C builddir --destdir staging/
+
   cmake-visual-2019:
     runs-on: windows-2019
     strategy:
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -202,6 +202,27 @@ jobs:
           meson test -C builddir/ --print-errorlogs
           meson install -C builddir --destdir staging/
 
+  meson-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install packages
+        run: pip install --pre meson
+      - name: Initialize the MSVC dev command prompt
+        uses: ilammy/msvc-dev-cmd@674ff850cbd739c402260838fa45b7114f750570
+      - name: Configure with Meson
+        run: |
+          meson setup build/meson/ builddir/ -Dbin_tests=true
+      - name: Build with Meson
+        run: |
+          ninja -C builddir/
+      - name: Test with Meson
+        run: |
+          meson test -C builddir/ --print-errorlogs
+      - name: Install with Meson
+        run: |
+          meson install -C builddir --destdir staging/
+
   cmake-visual-2019:
     runs-on: windows-2019
     strategy:
--- a/build/meson/tests/meson.build
+++ b/build/meson/tests/meson.build
@@ -192,6 +192,8 @@ test('test-zstream-1',
 test('test-zstream-3',
   zstreamtest,
   args: ['--newapi', '-t1', ZSTREAM_TESTTIME] + FUZZER_FLAGS,
+  # --newapi dies on Windows with "exit status 3221225477 or signal 3221225349 SIGinvalid"
+  should_fail: host_machine_os == os_windows,
   timeout: 120)
 test('test-longmatch', longmatch, timeout: 36)
 test('test-invalidDictionaries', invalidDictionaries) # should be fast
--- a/build/meson/tests/meson.build
+++ b/build/meson/tests/meson.build
@@ -134,13 +134,15 @@ checkTag = executable('checkTag',
 # =============================================================================
 
 if tests_supported_oses.contains(host_machine_os)
-  valgrind_prog = find_program('valgrind', ['/usr/bin/valgrind'], required: true)
+  valgrind_prog = find_program('valgrind', ['/usr/bin/valgrind'], required: false)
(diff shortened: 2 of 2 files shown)


eli-schwartz:
After manually installing these fixes into the Meson WrapDB, I finally got both Windows and macOS (new additions to our test matrix) to pass CI: https://github.com/mesonbuild/wrapdb/runs/6089540648?check_suite_focus=true

build/meson/tests/meson.build
```diff
@@ -183,6 +185,8 @@ test('test-zstream-1',
 test('test-zstream-3',
   zstreamtest,
   args: ['--newapi', '-t1', ZSTREAM_TESTTIME] + FUZZER_FLAGS,
+  # --newapi dies on Windows with "exit status 3221225477 or signal 3221225349 SIGinvalid"
+  should_fail: host_machine_os == os_windows,
```
Cyan4973, line 196: what if the issue is fixed, and `zstreamtest --newapi` now completes successfully ?
would it be considered a test failure due to `should_fail` ?

eli-schwartz: This PR changes its status to EXPECTEDFAIL, which is a success status. When the bug is fixed, it will start producing an UNEXPECTEDPASS status, which is considered a failure and causes the overall test results to be negative.

Cyan4973: > This PR changes its status to EXPECTEDFAIL, which is a success status. When the bug is fixed, it will start producing an UNEXPECTEDPASS status, which is considered a failure and causes the overall test results to be negative.

Better fix the issue then

eli-schwartz: Yep, once that other PR is merged I will re-check this and drop the obsolete commit.

eli-schwartz:
Is there anything else I should do about this?

As far as discussion over the second commit goes, it seems the attempt to actually fix the issue isn't getting anywhere. I think it's fine to add a "known broken" label to the testsuite.

Cyan4973:
I don't remember all the details,
it seems that : 
- On Windows + MinGW, and likely Windows + Visual, `zstreamtest --newpi` fails sometimes. But it's not reliably reproducible, meaning it fails at different places between consecutive runs on the same exact test.
- This seems to points at a race condition in the multi-threading code. And it's possibly related to the shim `pthread` translation layer for Windows, since there is no such issue when using `pthread` directly, as in Windows + Cygwin combination for example, which works fine.
- The proposal in this PR is to hide the issue by disabling this test, or making its result non-blocking, when run on Windows.

I think it's fair to say that I would largely prefer to have the issue fixed.
Problem is, the fix is not yet found, and it's a hard problem because the issue is very difficult to observe.
And a previous fix attempt, in #3121, failed to reach the objective.

Given a choice, I would still prefer to fix the issue.
Problem is, finding the time to look into it is difficult, our team is tied by other priorities.
We can still make this item a priority objective for the next release, but it doesn't tell when someone will take care of it.
It just imply that it's probably going to happen this year.

eli-schwartz:
> The proposal in this PR is to hide the issue by disabling this test, or making its result non-blocking, when run on Windows.

To be clear, the proposal is to make it be non-blocking, but visibly reported:

```
passing tests, yay : XXX
failing tests, phew: 0
known broken? :/   : 1
```

This is something that the Test Anything protocol calls "todo" tests: https://testanything.org/tap-specification.html#todo-tests
> These tests represent a feature to be implemented or a bug to be fixed and act as something of an executable “things to do” list.

Automake's ["Generalities about testing"](https://www.gnu.org/software/automake/manual/html_node/Generalities-about-Testing.html) mentions:
> It’s not uncommon, especially during early development stages, that some tests fail for known reasons, and that the developer doesn’t want to tackle these failures immediately (this is especially true when the failing tests deal with corner cases).

It's not uncommon to make a test failure be non-blocking when you've temporarily exhausted your efforts at solving it (and you have a tracking issue for it, *and* the testsuite flags it as a FIXME). If the test failure doesn't indicate a critical release blocker, then you want the testsuite status to be useful -- i.e. it should be non-blocking by default, and only raise critical errors if new or unknown issues are introduced.

Cyan4973:
OK, this will be discussed with the team, so that we can provide a feedback on this PR in a reasonable delay.

eli-schwartz:
Alright, thanks for looking into it. :)

Cyan4973:
Follow up : @terrelln is going to have a look at this topic.
It requires a working Windows machine, with development tools, so it's gonna take a little time, but hopefully nothing too long.
The hope is that a fix can be found quickly, on the ground that the problem can be reproduced pretty easily.
But if not, we'll revisit this proposal of having a non-blocking test while waiting for a (longer term) fix.

eli-schwartz:
Any update on this? :)

eli-schwartz:
@terrelln did you ever have a chance to investigate this?

yoniko:
> @terrelln did you ever have a chance to investigate this?

I'll take a look at this today / tomorrow.

yoniko:
An update: I've reproduced the problem and found the root cause.
There's indeed a bug in the windows threading translation layer.
Will put up a PR with a fix soon.

eli-schwartz:
Fantastic! Thank you for your investigation. Looking forward to being able to pass tests on the next zstd release. :)

eli-schwartz:
Added CI improvements on top. This also includes #3368 which is needed to configure the contrib directory so that CI runs, although that PR is much more easily merged.

.github/workflows/dev-short-tests.yml
```diff
@@ -178,6 +178,51 @@ jobs:
...
+          meson install -C builddir --destdir staging/
+
+  meson-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install packages
+        run: pip install --pre meson
+      - name: Initialize the MSVC dev command prompt
+        uses: ilammy/msvc-dev-cmd@v1
```
yoniko: We should specify a commit hash rather than a tag, especially when depending on third-party actions.

yoniko (changes requested):
Looks good, my only feedback is that I think it'd be better to pin the third-party action to a specific commit.
After making the change I think it'd be a good idea to merge this and then merge a PR that fixes the issue and removes the expected failure for windows.

.github/workflows/dev-short-tests.yml
eli-schwartz: Fair enough. :) I tend to avoid third-party ones too, at least for workflows that have secrets! but I don't usually bother with hash pinning since I either trust the action source or don't give it secrets -- and I copied this from a workflow that doesn't have secrets.

I'll make this change when rebasing (soon).

yoniko: Yeah, it's probably safe, but at the same time it's a better practice to pin to specific hashes and it's a change we plan on doing with other dependencies as well.
In any case, sounds good.

eli-schwartz: Done.

yoniko:
Looks good, although a bit unstable as the windows test might actually pass sometimes.
Merging so we can merge the other PR that solves the bug and would fix the CI to expect to pass.

eli-schwartz:
Oops, yeah, I think I grabbed a hash for the msvc-dev-cmd action, from the wrong branch.

eli-schwartz:
Honestly, I really wish it was possible to target semantically meaningful versions of an action, like I originally did, and specify a checksum to verify, instead of abusing commit hashes. But GitHub doesn't support that. :(

PR #3127, opened 2022-05-10
Correct and clarify repcode offset history logic

embg:
# Summary
In zstd, repcode offsets are passed to each block from the previous `Compressed_Block`. The compression and decompression side need to maintain identical repcode offset histories to prevent data corruption.

My last PR (https://github.com/facebook/zstd/pull/3114) introduced a bug in `ZSTD_compressBlock_fast_extDict` which caused those histories to fall out of sync. This bug was found by OSS fuzz, which was able to trigger data corruption by encoding a repcode match using an incorrect repcode offset (passed incorrectly from a previous block).

This PR fixes that bug and goes further, fixing a latent issue in the existing code (pre-#3114) which @terrelln and I discovered while addressing the fuzzer bug.

# The latent bug
The existing `offsetSaved` logic in fast, doublefast, and lazy noDict breaks the repcode offset history in a more subtle way than https://github.com/facebook/zstd/pull/3114. The value of `offset_1` which is passed to the next block is always correct, but `offset_2` can be incorrect if both offsets are invalid going into the block and no matches are found. This is because there is only one `offsetSaved` variable, but two offsets to save.

Luckily, this **cannot produce data corruption** due to the specifics of how `offset_2` is used in those matchfinders. Still, I have been able to construct an [input](https://gist.github.com/embg/4f445b51c362ed1dbd6267b13d1e4f11) for streaming compression (without any dictionary) which passes an incorrect `offset_2` from the second-to-last block to the final block.

Even without producing real corruption, this is undesirable; the correctness of the repcode offset history logic shouldn't depend on contingent factors regarding how `offset_2` is used in practice. 

# Solutions
This PR addresses the above problems in three ways:
* Adds repcode offset history preservation to `fast_extDict`, fixing the immediate OSS fuzz issue.
* Removes `offsetSaved` logic from the DMS matchfinders, since they don't actually use it. (Probably because in zstd, the whole dictionary is considered to be in the window if any byte of it is in the window).
* Splits `offsetSaved` into two variables `offsetSaved1` and `offsetSaved2` with rotation from 1 -> 2 when necessary.

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -175,7 +175,7 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
         } while (ip1 <= ilimit);
 
 _cleanup:
-        /* If rep_offset1 started invalid (offsetSaved1 > 0) and became valid (rep_offset1 > 0),
+        /* If offset_1 started invalid (offsetSaved1 > 0) and became valid (offset_1 > 0),
          * rotate saved offsets. */
         offsetSaved2 = ((offsetSaved1 > 0) & (offset_1 > 0)) ? offsetSaved1 : offsetSaved2;
 
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -175,9 +175,9 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
         } while (ip1 <= ilimit);
 
 _cleanup:
-        /* If offset_1 started invalid (offsetSaved1 > 0) and became valid (offset_1 > 0),
-         * rotate saved offsets. */
-        offsetSaved2 = ((offsetSaved1 > 0) & (offset_1 > 0)) ? offsetSaved1 : offsetSaved2;
+        /* If offset_1 started invalid (offsetSaved1 != 0) and became valid (offset_1 != 0),
+         * rotate saved offsets. See comment in ZSTD_compressBlock_fast_noDict for more context. */
+        offsetSaved2 = ((offsetSaved1 != 0) && (offset_1 != 0)) ? offsetSaved1 : offsetSaved2;
 
         /* save reps for next block */
         rep[0] = offset_1 ? offset_1 : offsetSaved1;
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -67,7 +67,7 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
     const BYTE* const iend = istart + srcSize;
     const BYTE* const ilimit = iend - HASH_READ_SIZE;
     U32 offset_1=rep[0], offset_2=rep[1];
-    U32 offsetSaved = 0;
+    U32 offsetSaved1 = 0, offsetSaved2 = 0;
 
     size_t mLength;
     U32 offset;
@@ -100,8 +100,8 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
         U32 const current = (U32)(ip - base);
         U32 const windowLow = ZSTD_getLowestPrefixIndex(ms, current, cParams->windowLog);
         U32 const maxRep = current - windowLow;
-        if (offset_2 > maxRep) offsetSaved = offset_2, offset_2 = 0;
-        if (offset_1 > maxRep) offsetSaved = offset_1, offset_1 = 0;
+        if (offset_2 > maxRep) offsetSaved2 = offset_2, offset_2 = 0;
+        if (offset_1 > maxRep) offsetSaved1 = offset_1, offset_1 = 0;
     }
 
     /* Outer Loop: one iteration per match found and stored */
@@ -175,9 +175,13 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
         } while (ip1 <= ilimit);
 
 _cleanup:
+        /* If rep_offset1 started invalid (offsetSaved1 > 0) and became valid (rep_offset1 > 0),
+         * rotate saved offsets. */
+        offsetSaved2 = ((offsetSaved1 > 0) & (offset_1 > 0)) ? offsetSaved1 : offsetSaved2;
+
         /* save reps for next block */
-        rep[0] = offset_1 ? offset_1 : offsetSaved;
-        rep[1] = offset_2 ? offset_2 : offsetSaved;
+        rep[0] = offset_1 ? offset_1 : offsetSaved1;
+        rep[1] = offset_2 ? offset_2 : offsetSaved2;
 
         /* Return the last literals size */
         return (size_t)(iend - anchor);
@@ -275,7 +279,6 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
     const BYTE* const iend = istart + srcSize;
     const BYTE* const ilimit = iend - HASH_READ_SIZE;
     U32 offset_1=rep[0], offset_2=rep[1];
-    U32 offsetSaved = 0;
 
     const ZSTD_matchState_t* const dms = ms->dictMatchState;
     const ZSTD_compressionParameters* const dictCParams = &dms->cParams;
@@ -461,8 +464,8 @@ _match_stored:
     }   /* while (ip < ilimit) */
 
     /* save reps for next block */
-    rep[0] = offset_1 ? offset_1 : offsetSaved;
-    rep[1] = offset_2 ? offset_2 : offsetSaved;
+    rep[0] = offset_1;
+    rep[1] = offset_2;
 
     /* Return the last literals size */
     return (size_t)(iend - anchor);
--- a/lib/compress/zstd_fast.c
+++ b/lib/compress/zstd_fast.c
@@ -117,7 +117,7 @@ ZSTD_compressBlock_fast_noDict_generic(
 
     U32 rep_offset1 = rep[0];
     U32 rep_offset2 = rep[1];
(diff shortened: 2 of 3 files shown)


Cyan4973:
Great explanation @embg !

Any performance measurement, to check if the changes produce any impact ?

embg:
> Great explanation @embg !
> 
> Any performance measurement, to check if the changes produce any impact ?

I figured performance measurement isn't necessary since I only touched code that executes at the very top and very bottom of the matchfinders (outside the hot loops).

Cyan4973:
> I figured performance measurement isn't necessary since I only touched code that executes at the very top and very bottom of the matchfinders (outside the hot loops).

You would be surprised !
Sometimes, modifications that seem to be locked outside of the main loop
actually messes up register allocation inside the main loop,
resulting in measurable impact where none was expected.

embg:
> > I figured performance measurement isn't necessary since I only touched code that executes at the very top and very bottom of the matchfinders (outside the hot loops).
> 
> You would be surprised ! Sometimes, modifications that seem to be locked outside of the main loop actually messes up register allocation inside the main loop, resulting in measurable impact where none was expected.

Understood -- would `zstd -b1 -e9 silesia.tar` be sufficient here?

Cyan4973:
> would `zstd -b1 -e9 silesia.tar` be sufficient here?

Yes

lib/compress/zstd_double_fast.c
```diff
@@ -175,9 +175,13 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
         } while (ip1 <= ilimit);
 
 _cleanup:
+        /* If offset_1 started invalid (offsetSaved1 > 0) and became valid (offset_1 > 0),
+         * rotate saved offsets. */
+        offsetSaved2 = ((offsetSaved1 > 0) & (offset_1 > 0)) ? offsetSaved1 : offsetSaved2;
```
terrelln:
```suggestion
        offsetSaved2 = ((offsetSaved1 > 0) && (offset_1 > 0)) ? offsetSaved1 : offsetSaved2;
```

terrelln:
```suggestion
        offsetSaved2 = ((offsetSaved1 > 0) && (offset_1 > 0)) ? offsetSaved1 : offsetSaved2;
```

embg:
@Cyan4973 Measurements look good! Mostly identical speeds, a few changes < 1% in both directions at various levels. Seems like noise to me since the changes go both ways, in some cases for levels which use the same matchfinder (e.g. levels 1 and 2 with clang12).

<details>
<summary> Measurements for gcc11 and clang12 </summary>

```

[embg@dev3100.frc1 ~/zstd] $BENCHMARK_PY run -- ./zstd_22875e_gcc11 -b1 -e9 ../silesia.tar
 1#silesia.tar       : 211975168 ->  73410303 (x2.888),  180.4 MB/s,  618.0 MB/s
 2#silesia.tar       : 211975168 ->  69490292 (x3.050),  149.2 MB/s,  559.2 MB/s
 3#silesia.tar       : 211975168 ->  66508266 (x3.187),  110.5 MB/s,  537.0 MB/s
 4#silesia.tar       : 211975168 ->  65313180 (x3.246),   96.6 MB/s,  522.7 MB/s
 5#silesia.tar       : 211975168 ->  62967059 (x3.366),   56.1 MB/s,  527.1 MB/s
 6#silesia.tar       : 211975168 ->  61468353 (x3.449),   39.9 MB/s,  557.4 MB/s
 7#silesia.tar       : 211975168 ->  60446005 (x3.507),   33.8 MB/s,  559.0 MB/s
 8#silesia.tar       : 211975168 ->  59919804 (x3.538),   26.8 MB/s,  573.2 MB/s
 9#silesia.tar       : 211975168 ->  59288102 (x3.575),   24.9 MB/s,  562.2 MB/s
[embg@dev3100.frc1 ~/zstd] $BENCHMARK_PY run -- ./zstd_8bf32d_gcc11 -b1 -e9 ../silesia.tar
 1#silesia.tar       : 211975168 ->  73410303 (x2.888),  180.9 MB/s,  617.8 MB/s
 2#silesia.tar       : 211975168 ->  69490292 (x3.050),  148.6 MB/s,  561.1 MB/s
 3#silesia.tar       : 211975168 ->  66508266 (x3.187),  110.2 MB/s,  538.0 MB/s
 4#silesia.tar       : 211975168 ->  65313180 (x3.246),   96.6 MB/s,  524.0 MB/s
 5#silesia.tar       : 211975168 ->  62967059 (x3.366),   56.7 MB/s,  527.4 MB/s
 6#silesia.tar       : 211975168 ->  61468353 (x3.449),   40.0 MB/s,  558.2 MB/s
 7#silesia.tar       : 211975168 ->  60446005 (x3.507),   34.0 MB/s,  560.1 MB/s
 8#silesia.tar       : 211975168 ->  59919804 (x3.538),   26.9 MB/s,  574.7 MB/s
 9#silesia.tar       : 211975168 ->  59288102 (x3.575),   25.2 MB/s,  563.3 MB/s
[embg@dev3100.frc1 ~/zstd] $BENCHMARK_PY run -- ./zstd_22875e_clang12 -b1 -e9 ../silesia.tar
 1#silesia.tar       : 211975168 ->  73410303 (x2.888),  181.4 MB/s,  601.0 MB/s
 2#silesia.tar       : 211975168 ->  69490292 (x3.050),  152.8 MB/s,  540.0 MB/s
 3#silesia.tar       : 211975168 ->  66508266 (x3.187),  108.7 MB/s,  517.0 MB/s
 4#silesia.tar       : 211975168 ->  65313180 (x3.246),   95.4 MB/s,  502.4 MB/s
 5#silesia.tar       : 211975168 ->  62967059 (x3.366),   58.4 MB/s,  505.5 MB/s
 6#silesia.tar       : 211975168 ->  61468353 (x3.449),   41.3 MB/s,  536.7 MB/s
 7#silesia.tar       : 211975168 ->  60446005 (x3.507),   34.7 MB/s,  536.9 MB/s
 8#silesia.tar       : 211975168 ->  59919804 (x3.538),   27.0 MB/s,  552.6 MB/s
 9#silesia.tar       : 211975168 ->  59288102 (x3.575),   25.5 MB/s,  541.7 MB/s
[embg@dev3100.frc1 ~/zstd] $BENCHMARK_PY run -- ./zstd_8bf32d_clang12 -b1 -e9 ../silesia.tar
 1#silesia.tar       : 211975168 ->  73410303 (x2.888),  182.7 MB/s,  601.4 MB/s
 2#silesia.tar       : 211975168 ->  69490292 (x3.050),  150.9 MB/s,  540.0 MB/s
 3#silesia.tar       : 211975168 ->  66508266 (x3.187),  110.7 MB/s,  519.6 MB/s
 4#silesia.tar       : 211975168 ->  65313180 (x3.246),   95.5 MB/s,  502.9 MB/s
 5#silesia.tar       : 211975168 ->  62967059 (x3.366),   58.0 MB/s,  503.3 MB/s
 6#silesia.tar       : 211975168 ->  61468353 (x3.449),   41.3 MB/s,  538.1 MB/s
 7#silesia.tar       : 211975168 ->  60446005 (x3.507),   34.6 MB/s,  537.1 MB/s
 8#silesia.tar       : 211975168 ->  59919804 (x3.538),   27.2 MB/s,  554.1 MB/s
 9#silesia.tar       : 211975168 ->  59288102 (x3.575),   25.4 MB/s,  536.8 MB/s
 ```
 
 </details>

File: lib/compress/zstd_double_fast.c [hunk 2]
```diff
@@ -175,9 +175,13 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
         } while (ip1 <= ilimit);
 
 _cleanup:
+        /* If offset_1 started invalid (offsetSaved1 > 0) and became valid (offset_1 > 0),
+         * rotate saved offsets. */
```
terrelln: Please add a more complete explaination. You can just add it to one match finder, e.g. fast, and then add a brief summary in the other match finders, and point to the full explaination.

Maybe something like this, but feel free to re-word as you see fit.

```suggestion

        /* When the repcodes are outside of the prefix, we set them to zero before the loop.
         * When the offsets are still zero, we need to restore them after the block to have a correct
         * repcode history. If only one offset was invalid, it is easy. The tricky case is when both
         * offsets were invalid. We need to figure out which offset to refill with. If both offsets
         * are zero they are in the same order. If both offsets are non-zero, we won't restore the
         * offsets from `offsetSaved[12]`. If only one is zero, we need to decide which offset to
         * restore. If offset_1 is non-zero, then offset_2 must be offsetSaved1. It is impossible for
         * offset_2 to be non-zero.
         * So if offset_1 started invalid (offsetSaved1 > 0) and became valid (offset_1 > 0), then
         * set rep[0] = offset_1 and rep[1] = offsetSaved1.
         */
```

lib/compress/zstd_double_fast.c
embg: Good suggestion, since `offsetSaved1 == 0` 99.999% of the time we don't need to waste time evaluating `offset_1 > 0`.

lib/compress/zstd_double_fast.c
embg: This explanation is great, I'll add it to fast and point the other comments there. I'll just add a small clarifiction around "It is impossible for offset_2 to be non-zero" since it took me a second to understand what you meant. At first I didn't realize that was scoped underneath "If only one is zero" :)

Cyan4973 (approved):

terrelln (approved):

PR #3129, opened 2022-05-10
ZSTD_fast_noDict: Avoid Safety Check When Writing `ip1` into Table

felixhandte:
This commit avoids checking whether a hashtable write is safe in two of the
three match-found paths in `ZSTD_compressBlock_fast_noDict_generic`. This pro-
duces a ~1% speed-up in compression.

This is a branch I've wanted to kill for a long time. It's very rarely useful.
Recent discussion with @embg brought this topic up again, and I finally had an
idea for how to do it (mostly).

A comment in the code describes why we can skip this check in the other two
paths (the repcode check and the first match check in the unrolled loop).

A downside is that in the new position where we make this check, we have not
yet computed `mLength`. We therefore have to avoid writing *possibly* dangerous
positions, rather than the old check which only avoids writing *actually*
dangerous positions. This leads to a miniscule loss in ratio (remember that
this scenario can only been triggered in very negative levels or under incomp-
ressibility acceleration).

### DIFF
--- a/lib/compress/zstd_fast.c
+++ b/lib/compress/zstd_fast.c
@@ -183,7 +183,9 @@ _start: /* Requires: ip0 */
             offcode = REPCODE1_TO_OFFBASE;
             mLength += 4;
 
-            /* first write next hash table entry; we've already calculated it */
+            /* First write next hash table entry; we've already calculated it.
+             * This write is known to be safe because the ip1 is before the
+             * repcode (ip2). */
             hashTable[hash1] = (U32)(ip1 - base);
 
             goto _match;
@@ -200,7 +202,9 @@ _start: /* Requires: ip0 */
         if (MEM_read32(ip0) == mval) {
             /* found a match! */
 
-            /* first write next hash table entry; we've already calculated it */
+            /* First write next hash table entry; we've already calculated it.
+             * This write is known to be safe because the ip1 == ip0 + 1, so
+             * we know we will resume searching after ip1 */
             hashTable[hash1] = (U32)(ip1 - base);
 
             goto _offset;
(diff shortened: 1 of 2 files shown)


lib/compress/zstd_fast.c
```diff
@@ -288,9 +305,7 @@ ZSTD_compressBlock_fast_noDict_generic(
     anchor = ip0;
 
     /* write next hash table entry */
-    if (ip1 < ip0) {
-        hashTable[hash1] = (U32)(ip1 - base);
-    }
+    hashTable[hash1] = (U32)(ip1 - base);
```
Cyan4973: If I understand correctly, when `step > 4`, then `hashTable[hash1] = 0;`, so it's removing a potentially useful past reference from the hash table ?

felixhandte: Correct. So this does affect levels -5 and beyond. But otherwise it has basically no effect, since that situation is very rare otherwise.

Cyan4973: What would be the consequences of _not_ having this hash table write operation ?

felixhandte: So I thought about this some more and came up with a better way to do it!

To answer your question, we would gain some speed and lose some ratio. Here's a quick benchmark comparison:

Comp | Version | Speed | Ratio
--|--|--|--
gcc-11 | no write | 348.2 | x2.876
gcc-11 | `dev` | 345.7 | x2.887
gcc-11 | this PR | 348.7 | x2.887
clang-13 | no write | 346.1 | x2.876
clang-13 | `dev` | 343.9 | x2.887
clang-13 | this PR | 348.5 | x2.887

Cyan4973 (approved):

nadavrot:
Cool optimization @felixhandte !

PR #3145, opened 2022-05-26
common: apply two stage copy to aarch64

JunHe77:
On aarch64 ZSTD_wildcopy uses a simple loop to do
16B based memory copy. There is existing optimized
two stage copy that can achieve better performance.
By applying this to aarch64 it is also observed ~1%
uplift in silesia corpus.

Change-Id: Ic1253308e7a8a7df2d08963ba544e086c81ce8be

### DIFF
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -235,12 +235,6 @@ void ZSTD_wildcopy(void* dst, const void* src, ptrdiff_t length, ZSTD_overlap_e 
          * one COPY16() in the first call. Then, do two calls per loop since
          * at that point it is more likely to have a high trip count.
          */
-#ifdef __aarch64__
-        do {
-            COPY16(op, ip);
-        }
-        while (op < oend);
-#else
         ZSTD_copy16(op, ip);
         if (16 >= length) return;
         op += 16;
@@ -250,7 +244,6 @@ void ZSTD_wildcopy(void* dst, const void* src, ptrdiff_t length, ZSTD_overlap_e 
             COPY16(op, ip);
         }
         while (op < oend);
-#endif
     }
 }
 


Cyan4973:
By `aarch64`, could you provide some details about the exact type of cpu / soc this patch has been benchmarked on ?

Cyan4973:
By `aarch64`, could you provide some details about the exact type of cpu / soc this patch has been benchmarked on ?

JunHe77:
Hi, @Cyan4973 , the result have been benchmarked on the Arm N1/A72/A57 platforms and observed similar uplift.

Cyan4973:
I can't remember why this code was added here.

It could be that, with `aarch64` being merely an instruction set under which so many different SoCs are built, maybe some of them (non `N1`) would prefer the first loop.
However, it's pretty hard to confirm / test (couldn't find a test platform where this holds true).

From what I can see, the second formulation just separates the first branch from later ones, so that it can have its own statistics (as opposed to being merged with other loop iterations). Such a construction is expected to be _rather good_ in the context of `wildcopy`, essentially distinguishing small copies from larger ones. This should translate into almost always better performance, except maybe for specific systems with rather poor branch predictors.

So I'm gonna make an educated guess here and state that this PR seems tends to improve the situation, on top of simplifying it by removing a weird and poorly documented corner case.

Cyan4973:
I can't remember why this code was added here.

It could be that, with `aarch64` being merely an instruction set under which so many different SoCs are built, maybe some of them (non `N1`) would prefer the first loop.
However, it's pretty hard to confirm / test (couldn't find a test platform where this holds true).

From what I can see, the second formulation just separates the first branch from later ones, so that it can have its own statistics (as opposed to being merged with other loop iterations). Such a construction is expected to be _rather good_ in the context of `wildcopy`, essentially distinguishing small copies from larger ones. This should translate into almost always better performance, except maybe for specific systems with rather poor branch predictors.

So I'm gonna make an educated guess here and state that this PR seems tends to improve the situation, on top of simplifying it by removing a weird and poorly documented corner case.

Cyan4973 (approved):

JunHe77:
Thank you @Cyan4973 for the review. I used to check the log info of that change (969ba4f2b92441eb), but could not find the context of it. It looks like this is designed for compression. With my benchmarks here, I didn't find regression in compression with removing the aarch64 specific part on N1/A72/A57.

terrelln:
Thanks for the PR @JunHe77!

PR #3148, opened 2022-05-30
simple fix

ihsinme:


Cyan4973 (approved):
Thanks @ihsinme ,
indeed this is a more logical order

PR #3152, opened 2022-06-02
"Short cache" optimization for level 1-4 DMS (+5-30% compression speed)

embg:
## TLDR
This PR increases dictionary compression speed for small inputs (< 8KB at levels 1-2, < 16KB at levels 3-4) by 5-30%. The win is especially large when dictionaries are cold, i.e. in L3 cache or main memory.

## Description of the optimization
Short cache is a change to fast/dfast CDict hashtable construction which allows the corresponding matchfinders to avoid unnecessary memory loads. A picture is worth 2^10 words:

short_cache_diagram

The circle at the bottom of the diagram is where matchfinders use a dictionary index (loaded from the CDict hashtable) to load some position from the dictionary and compare it to the current position. Short cache allows us to prevent that load with high probability when the current position and dictionary position *do not match*. This is the common case in nearly all scenarios, and will usually prevent an L2 or even L3 cache miss.

How do we prevent the load? When the CDict hashtable is constructed, we insert an 8-bit independent hash (called a "tag") into the lower bits of each entry. We can do that since dictionary indices are less than 24 bits (this PR adds code to guarantee that), so we can pack an index and tag into a single `U32`. In the matchfinder loop, we unpack the index and tag from the CDict hashtable, compute a tag at the current position, and only load the dictionary position *if the tags match*.

## Preliminary win measurements
<details>

Level 1:
results0518c

Level 3:
newplot (2)

</details>

## Final win measurements
The final measurements are comparable to the preliminary ones. Note that fast/hot/gcc/110K goes from 0% to -1% (this is on top of my previous fast DMS pipeline which increased the speed of that scenario by more than 5%). I didn't find any other regressions relative to dev. Regressions from preliminary to final seem to roughly cancel out improvements, and I don't think it's worth digging into which are measurement noise vs real changes. These are good win graphs and I think we should land them.

Benchmarking environment (machine, core isolation, etc) was the same as for https://github.com/facebook/zstd/pull/3086 -- see that PR for details.

Level 1:
newplot (7)

Level 3:
newplot (6)

## Regression
The tags added by short cache need to be removed from the CDict hashtable when it is copied into the CCtx for extDict compression. This turns a normal memcpy into a vectorized shift-and-write loop. In a microbenchmark, I found that SSE2 shift-and-write is 2x slower than memcpy, while AVX2 is 1.5x slower. Note that we only pay this cost when using extDict for dictionary compression. Streaming compression is not affected.

I measured the regression for level2 and level4 extDict. I chose those levels because they use at least 2x larger hashtables than the much more common levels 1 and 3. Thus, level2 and level4 upper-bound the regression (I confirmed this with measurements on level1 and level3).

The AVX2 numbers are for binaries compiled with `-march=core-avx2`. I will add AVX2 dynamic dispatch before our next open source release to ensure that binaries not compiled with `-march=core-avx2` can still benefit from those instructions when available.

newplot (13)

---
- [x] Respond to initial code review
- [x] Update the PR description with a summary of the changes and more information on measurements.
- [ ] Add discussion of the (negligible) ratio impact.
- [x] Add documentation to the code itself (see marked location in `zstd_compress_internal.h`).
- [x] Add measurements to the PR description regarding the `byCopyingCDict` regression.

Out of scope (will be a separate PR):
- ~~Add AVX2 dynamic dispatch for the (minor) performance regression in `ZSTD_resetCCtx_byCopyingCDict`.~~

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2131,9 +2131,17 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
                                                             : 0;
         size_t const hSize =  (size_t)1 << cdict_cParams->hashLog;
 
-        ZSTD_memcpy(cctx->blockState.matchState.hashTable,
-               cdict->matchState.hashTable,
-               hSize * sizeof(U32));
+        if (cctx->appliedParams.cParams.strategy == ZSTD_fast){
+            size_t i;
+            for (i=0; i < hSize; i++) {
+                cctx->blockState.matchState.hashTable[i] = cdict->matchState.hashTable[i] >> 8; // TODO make this reference macro
+            }
+        } else {
+            ZSTD_memcpy(cctx->blockState.matchState.hashTable,
+                cdict->matchState.hashTable,
+                hSize * sizeof(U32));
+        }
+
         /* Do not copy cdict's chainTable if cctx has parameters such that it would not use chainTable */
         if (ZSTD_allocateChainTable(cctx->appliedParams.cParams.strategy, cctx->appliedParams.useRowMatchFinder, 0 /* forDDSDict */)) {
             ZSTD_memcpy(cctx->blockState.matchState.chainTable,
@@ -4205,7 +4213,7 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
                                          ZSTD_cwksp* ws,
                                          ZSTD_CCtx_params const* params,
                                          const void* src, size_t srcSize,
-                                         ZSTD_dictTableLoadMethod_e dtlm)
+                                         ZSTD_dictTableLoadMethod_e dtlm, U32 const forCCtx)
 {
     const BYTE* ip = (const BYTE*) src;
     const BYTE* const iend = ip + srcSize;
@@ -4252,7 +4260,7 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
     switch(params->cParams.strategy)
     {
     case ZSTD_fast:
-        ZSTD_fillHashTable(ms, iend, dtlm);
+        ZSTD_fillHashTable(ms, iend, dtlm, forCCtx);
         break;
     case ZSTD_dfast:
         ZSTD_fillDoubleHashTable(ms, iend, dtlm);
@@ -4421,7 +4429,8 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
                                       ZSTD_CCtx_params const* params,
                                       const void* dict, size_t dictSize,
                                       ZSTD_dictTableLoadMethod_e dtlm,
-                                      void* workspace)
+                                      void* workspace,
+                                      U32 const forCCtx)
 {
     const BYTE* dictPtr = (const BYTE*)dict;
     const BYTE* const dictEnd = dictPtr + dictSize;
@@ -4439,7 +4448,7 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
     {
         size_t const dictContentSize = (size_t)(dictEnd - dictPtr);
         FORWARD_IF_ERROR(ZSTD_loadDictionaryContent(
-            ms, NULL, ws, params, dictPtr, dictContentSize, dtlm), "");
+            ms, NULL, ws, params, dictPtr, dictContentSize, dtlm, forCCtx), "");
     }
     return dictID;
 }
@@ -4455,7 +4464,8 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
                          const void* dict, size_t dictSize,
                                ZSTD_dictContentType_e dictContentType,
                                ZSTD_dictTableLoadMethod_e dtlm,
-                               void* workspace)
+                               void* workspace,
+                               U32 const forCCtx)
 {
     DEBUGLOG(4, "ZSTD_compress_insertDictionary (dictSize=%u)", (U32)dictSize);
     if ((dict==NULL) || (dictSize<8)) {
@@ -4467,13 +4477,13 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
 
     /* dict restricted modes */
     if (dictContentType == ZSTD_dct_rawContent)
-        return ZSTD_loadDictionaryContent(ms, ls, ws, params, dict, dictSize, dtlm);
+        return ZSTD_loadDictionaryContent(ms, ls, ws, params, dict, dictSize, dtlm, forCCtx);
 
     if (MEM_readLE32(dict) != ZSTD_MAGIC_DICTIONARY) {
         if (dictContentType == ZSTD_dct_auto) {
             DEBUGLOG(4, "raw content dictionary detected");
             return ZSTD_loadDictionaryContent(
-                ms, ls, ws, params, dict, dictSize, dtlm);
+                ms, ls, ws, params, dict, dictSize, dtlm, forCCtx);
         }
         RETURN_ERROR_IF(dictContentType == ZSTD_dct_fullDict, dictionary_wrong, "");
         assert(0);   /* impossible */
@@ -4481,7 +4491,7 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
 
     /* dict as full zstd dictionary */
     return ZSTD_loadZstdDictionary(
-        bs, ms, ws, params, dict, dictSize, dtlm, workspace);
+        bs, ms, ws, params, dict, dictSize, dtlm, workspace, forCCtx);
 }
 
 #define ZSTD_USE_CDICT_PARAMS_SRCSIZE_CUTOFF (128 KB)
@@ -4524,11 +4534,11 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
                         cctx->blockState.prevCBlock, &cctx->blockState.matchState,
                         &cctx->ldmState, &cctx->workspace, &cctx->appliedParams, cdict->dictContent,
                         cdict->dictContentSize, cdict->dictContentType, dtlm,
-                        cctx->entropyWorkspace)
+                        cctx->entropyWorkspace, 1)
               : ZSTD_compress_insertDictionary(
                         cctx->blockState.prevCBlock, &cctx->blockState.matchState,
                         &cctx->ldmState, &cctx->workspace, &cctx->appliedParams, dict, dictSize,
-                        dictContentType, dtlm, cctx->entropyWorkspace);
+                        dictContentType, dtlm, cctx->entropyWorkspace, 1);
         FORWARD_IF_ERROR(dictID, "ZSTD_compress_insertDictionary failed");
         assert(dictID <= UINT_MAX);
         cctx->dictID = (U32)dictID;
@@ -4832,7 +4842,8 @@ static size_t ZSTD_initCDict_internal(
         {   size_t const dictID = ZSTD_compress_insertDictionary(
                     &cdict->cBlockState, &cdict->matchState, NULL, &cdict->workspace,
                     &params, cdict->dictContent, cdict->dictContentSize,
-                    dictContentType, ZSTD_dtlm_full, cdict->entropyWorkspace);
+                    dictContentType, ZSTD_dtlm_full, cdict->entropyWorkspace,
+                    params.cParams.strategy == ZSTD_fast ? 0 : 1);
             FORWARD_IF_ERROR(dictID, "ZSTD_compress_insertDictionary failed");
             assert(dictID <= (size_t)(U32)-1);
             cdict->dictID = (U32)dictID;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2131,7 +2131,7 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
                                                             : 0;
         size_t const hSize =  (size_t)1 << cdict_cParams->hashLog;
 
-        if (cctx->appliedParams.cParams.strategy == ZSTD_fast){
+        if (cctx->appliedParams.cParams.strategy == ZSTD_fast || cctx->appliedParams.cParams.strategy == ZSTD_dfast){ // TODO clean up this logic
             size_t i;
             for (i=0; i < hSize; i++) {
                 cctx->blockState.matchState.hashTable[i] = cdict->matchState.hashTable[i] >> 8; // TODO make this reference macro
@@ -2144,9 +2144,16 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
 
         /* Do not copy cdict's chainTable if cctx has parameters such that it would not use chainTable */
         if (ZSTD_allocateChainTable(cctx->appliedParams.cParams.strategy, cctx->appliedParams.useRowMatchFinder, 0 /* forDDSDict */)) {
-            ZSTD_memcpy(cctx->blockState.matchState.chainTable,
-               cdict->matchState.chainTable,
-               chainSize * sizeof(U32));
+            if (cctx->appliedParams.cParams.strategy == ZSTD_fast || cctx->appliedParams.cParams.strategy == ZSTD_dfast){ // TODO clean up this logic
+                size_t i;
+                for (i=0; i < chainSize; i++) {
+                    cctx->blockState.matchState.chainTable[i] = cdict->matchState.chainTable[i] >> 8; // TODO make this reference macro
+                }
+            } else {
+                ZSTD_memcpy(cctx->blockState.matchState.chainTable,
+                            cdict->matchState.chainTable,
+                            chainSize * sizeof(U32));
+            }
         }
         /* copy tag table */
         if (ZSTD_rowMatchFinderUsed(cdict_cParams->strategy, cdict->useRowMatchFinder)) {
@@ -4263,7 +4270,7 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
         ZSTD_fillHashTable(ms, iend, dtlm, forCCtx);
         break;
     case ZSTD_dfast:
-        ZSTD_fillDoubleHashTable(ms, iend, dtlm);
+        ZSTD_fillDoubleHashTable(ms, iend, dtlm, forCCtx);
         break;
 
     case ZSTD_greedy:
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -275,6 +275,12 @@ static ZSTD_paramSwitch_e ZSTD_resolveEnableLdm(ZSTD_paramSwitch_e mode,
     return (cParams->strategy >= ZSTD_btopt && cParams->windowLog >= 27) ? ZSTD_ps_enable : ZSTD_ps_disable;
 }
 
+/* Returns 1 if compression parameters are such that CDict hashtable and chaintable indices are tagged.
+ * If so, the tags need to be removed in ZSTD_resetCCtx_byCopyingCDict. */
+static int ZSTD_CDictIndicesAreTagged(const ZSTD_compressionParameters* const cParams) {
+    return cParams->strategy == ZSTD_fast || cParams->strategy == ZSTD_dfast;
+}
+
 static ZSTD_CCtx_params ZSTD_makeCCtxParamsFromCParams(
         ZSTD_compressionParameters cParams)
 {
@@ -2131,10 +2137,11 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
                                                             : 0;
         size_t const hSize =  (size_t)1 << cdict_cParams->hashLog;
 
-        if (cctx->appliedParams.cParams.strategy == ZSTD_fast || cctx->appliedParams.cParams.strategy == ZSTD_dfast){ // TODO clean up this logic
+        if (ZSTD_CDictIndicesAreTagged(&cctx->appliedParams.cParams)){
             size_t i;
-            for (i=0; i < hSize; i++) {
-                cctx->blockState.matchState.hashTable[i] = cdict->matchState.hashTable[i] >> 8; // TODO make this reference macro
+            for (i = 0; i < hSize; i++) {
+                U32 const taggedIndex = cdict->matchState.hashTable[i];
+                cctx->blockState.matchState.hashTable[i] = taggedIndex >> ZSTD_SHORT_CACHE_TAG_BITS;
             }
         } else {
             ZSTD_memcpy(cctx->blockState.matchState.hashTable,
@@ -2144,15 +2151,16 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
 
         /* Do not copy cdict's chainTable if cctx has parameters such that it would not use chainTable */
         if (ZSTD_allocateChainTable(cctx->appliedParams.cParams.strategy, cctx->appliedParams.useRowMatchFinder, 0 /* forDDSDict */)) {
-            if (cctx->appliedParams.cParams.strategy == ZSTD_fast || cctx->appliedParams.cParams.strategy == ZSTD_dfast){ // TODO clean up this logic
+            if (ZSTD_CDictIndicesAreTagged(&cctx->appliedParams.cParams)){
                 size_t i;
-                for (i=0; i < chainSize; i++) {
-                    cctx->blockState.matchState.chainTable[i] = cdict->matchState.chainTable[i] >> 8; // TODO make this reference macro
+                for (i = 0; i < chainSize; i++) {
+                    U32 const taggedIndex = cdict->matchState.chainTable[i];
+                    cctx->blockState.matchState.chainTable[i] = taggedIndex >> ZSTD_SHORT_CACHE_TAG_BITS;
                 }
             } else {
                 ZSTD_memcpy(cctx->blockState.matchState.chainTable,
-                            cdict->matchState.chainTable,
-                            chainSize * sizeof(U32));
+                    cdict->matchState.chainTable,
+                    chainSize * sizeof(U32));
             }
         }
         /* copy tag table */
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -4842,8 +4842,7 @@ static size_t ZSTD_initCDict_internal(
         {   size_t const dictID = ZSTD_compress_insertDictionary(
                     &cdict->cBlockState, &cdict->matchState, NULL, &cdict->workspace,
                     &params, cdict->dictContent, cdict->dictContentSize,
-                    dictContentType, ZSTD_dtlm_full, cdict->entropyWorkspace,
-                    params.cParams.strategy == ZSTD_fast ? 0 : 1);
+                    dictContentType, ZSTD_dtlm_full, cdict->entropyWorkspace, 0);
             FORWARD_IF_ERROR(dictID, "ZSTD_compress_insertDictionary failed");
             assert(dictID <= (size_t)(U32)-1);
             cdict->dictID = (U32)dictID;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -4228,7 +4228,8 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
                                          ZSTD_cwksp* ws,
                                          ZSTD_CCtx_params const* params,
                                          const void* src, size_t srcSize,
-                                         ZSTD_dictTableLoadMethod_e dtlm, U32 const forCCtx)
+                                         ZSTD_dictTableLoadMethod_e dtlm,
+                                         ZSTD_tableFillPurpose_e tfp)
 {
     const BYTE* ip = (const BYTE*) src;
     const BYTE* const iend = ip + srcSize;
@@ -4275,10 +4276,10 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
     switch(params->cParams.strategy)
     {
     case ZSTD_fast:
-        ZSTD_fillHashTable(ms, iend, dtlm, forCCtx);
+        ZSTD_fillHashTable(ms, iend, dtlm, tfp);
         break;
     case ZSTD_dfast:
-        ZSTD_fillDoubleHashTable(ms, iend, dtlm, forCCtx);
+        ZSTD_fillDoubleHashTable(ms, iend, dtlm, tfp);
         break;
 
     case ZSTD_greedy:
@@ -4445,7 +4446,7 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
                                       const void* dict, size_t dictSize,
                                       ZSTD_dictTableLoadMethod_e dtlm,
                                       void* workspace,
-                                      U32 const forCCtx)
+                                      ZSTD_tableFillPurpose_e tfp)
 {
     const BYTE* dictPtr = (const BYTE*)dict;
     const BYTE* const dictEnd = dictPtr + dictSize;
@@ -4463,7 +4464,7 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
     {
         size_t const dictContentSize = (size_t)(dictEnd - dictPtr);
         FORWARD_IF_ERROR(ZSTD_loadDictionaryContent(
-            ms, NULL, ws, params, dictPtr, dictContentSize, dtlm, forCCtx), "");
+            ms, NULL, ws, params, dictPtr, dictContentSize, dtlm, tfp), "");
     }
     return dictID;
 }
@@ -4480,7 +4481,7 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
                                ZSTD_dictContentType_e dictContentType,
                                ZSTD_dictTableLoadMethod_e dtlm,
                                void* workspace,
-                               U32 const forCCtx)
+                               ZSTD_tableFillPurpose_e tfp)
 {
     DEBUGLOG(4, "ZSTD_compress_insertDictionary (dictSize=%u)", (U32)dictSize);
     if ((dict==NULL) || (dictSize<8)) {
@@ -4492,13 +4493,13 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
 
     /* dict restricted modes */
     if (dictContentType == ZSTD_dct_rawContent)
-        return ZSTD_loadDictionaryContent(ms, ls, ws, params, dict, dictSize, dtlm, forCCtx);
+        return ZSTD_loadDictionaryContent(ms, ls, ws, params, dict, dictSize, dtlm, tfp);
 
     if (MEM_readLE32(dict) != ZSTD_MAGIC_DICTIONARY) {
         if (dictContentType == ZSTD_dct_auto) {
             DEBUGLOG(4, "raw content dictionary detected");
             return ZSTD_loadDictionaryContent(
-                ms, ls, ws, params, dict, dictSize, dtlm, forCCtx);
+                ms, ls, ws, params, dict, dictSize, dtlm, tfp);
         }
         RETURN_ERROR_IF(dictContentType == ZSTD_dct_fullDict, dictionary_wrong, "");
         assert(0);   /* impossible */
@@ -4506,7 +4507,7 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
 
     /* dict as full zstd dictionary */
     return ZSTD_loadZstdDictionary(
-        bs, ms, ws, params, dict, dictSize, dtlm, workspace, forCCtx);
+        bs, ms, ws, params, dict, dictSize, dtlm, workspace, tfp);
 }
 
 #define ZSTD_USE_CDICT_PARAMS_SRCSIZE_CUTOFF (128 KB)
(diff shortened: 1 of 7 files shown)


lib/compress/zstd_double_fast.c
```diff
@@ -312,8 +359,12 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
         U32 offset;
         size_t const h2 = ZSTD_hashPtr(ip, hBitsL, 8);
         size_t const h = ZSTD_hashPtr(ip, hBitsS, mls);
-        size_t const dictHL = ZSTD_hashPtr(ip, dictHBitsL, 8);
-        size_t const dictHS = ZSTD_hashPtr(ip, dictHBitsS, mls);
+        size_t const dictHashAndTagL = ZSTD_hashPtr(ip, dictHBitsL + ZSTD_SHORT_CACHE_TAG_BITS, 8);
+        size_t const dictHashAndTagS = ZSTD_hashPtr(ip, dictHBitsS + ZSTD_SHORT_CACHE_TAG_BITS, mls);
+        U32 const dictMatchIndexAndTagL = dictHashLong[dictHashAndTagL >> ZSTD_SHORT_CACHE_TAG_BITS];
+        U32 const dictMatchIndexAndTagS = dictHashSmall[dictHashAndTagS >> ZSTD_SHORT_CACHE_TAG_BITS];
+        size_t const dictTagsMatchL = (dictMatchIndexAndTagL & ZSTD_SHORT_CACHE_TAG_MASK) == (dictHashAndTagL & ZSTD_SHORT_CACHE_TAG_MASK);
```
embg: TODO: split this across more lines

lib/compress/zstd_compress_internal.h
```diff
@@ -745,27 +745,27 @@ ZSTD_count_2segments(const BYTE* ip, const BYTE* match,
  *  Hashes
  ***************************************/
 static const U32 prime3bytes = 506832829U;
-static U32    ZSTD_hash3(U32 u, U32 h) { return ((u << (32-24)) * prime3bytes)  >> (32-h) ; }
+static U32    ZSTD_hash3(U32 u, U32 h) { assert(h <= 32); return ((u << (32-24)) * prime3bytes)  >> (32-h) ; }
```
felixhandte, line 749: We probably shouldn't leave these asserts here.

lib/compress/zstd_compress_internal.h
```diff
@@ -1264,6 +1264,24 @@ MEM_STATIC void ZSTD_debugTable(const U32* table, U32 max)
...
+/* Short Cache */
+
+typedef enum {
+    /* TODO: this is a good place to document short cache */
+    ZSTD_tfp_forCCtx = 0,
+    ZSTD_tfp_forCDict = 1
+} ZSTD_tableFillPurpose_e;
+
+#define ZSTD_SHORT_CACHE_TAG_BITS 8
+#define ZSTD_SHORT_CACHE_TAG_MASK ((1u << ZSTD_SHORT_CACHE_TAG_BITS) - 1)
```
felixhandte, line 1290: One note: it looks to me like there's actually no reason that the index/tag split needs to be byte-aligned. I guess maybe it means we can skip the mask operation with a `movb` instead. I'm not sure how important that is.

I'm just wondering if there's a motivation for adjusting the split point. Is that something you thought about?

lib/compress/zstd_double_fast.c
```diff
@@ -11,8 +11,43 @@
...
+{
+    const ZSTD_compressionParameters* const cParams = &ms->cParams;
+    U32* const hashLarge = ms->hashTable;
+    U32  const hBitsL = cParams->hashLog;
+    U32  const mls = cParams->minMatch;
+    U32* const hashSmall = ms->chainTable;
+    U32  const hBitsS = cParams->chainLog;
+    const BYTE* const base = ms->window.base;
+    const BYTE* ip = base + ms->nextToUpdate;
+    const BYTE* const iend = MIN((const BYTE*)end, base + (1u << (32 - ZSTD_SHORT_CACHE_TAG_BITS))) - HASH_READ_SIZE;
```
felixhandte: I'm not sure I understand. You're limiting the dict fill to the first 16MB of the dict right? But for very large dictionaries we'd rather limit it to the last 16MB.

Well, actually for very large dictionaries we'd probably rather not do short cache. But I guess that's not an option as implemented.

You might want to hook into the logic here?:

https://github.com/facebook/zstd/blob/v1.5.2/lib/compress/zstd_compress.c#L4224-L4229

lib/compress/zstd_compress_internal.h
```diff
@@ -1264,6 +1264,24 @@ MEM_STATIC void ZSTD_debugTable(const U32* table, U32 max)
...
+
+/* Helper function for ZSTD_fillHashTable and ZSTD_fillDoubleHashTable */
+MEM_STATIC void writeTaggedIndex(U32* const hashTable, size_t hashAndTag, U32 index) {
```
felixhandte: This symbol should still have a `ZSTD_` prefix, even if it is static  / internal.

lib/compress/zstd_double_fast.c
```diff
@@ -312,8 +359,12 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
         U32 offset;
         size_t const h2 = ZSTD_hashPtr(ip, hBitsL, 8);
         size_t const h = ZSTD_hashPtr(ip, hBitsS, mls);
-        size_t const dictHL = ZSTD_hashPtr(ip, dictHBitsL, 8);
-        size_t const dictHS = ZSTD_hashPtr(ip, dictHBitsS, mls);
+        size_t const dictHashAndTagL = ZSTD_hashPtr(ip, dictHBitsL + ZSTD_SHORT_CACHE_TAG_BITS, 8);
+        size_t const dictHashAndTagS = ZSTD_hashPtr(ip, dictHBitsS + ZSTD_SHORT_CACHE_TAG_BITS, mls);
+        U32 const dictMatchIndexAndTagL = dictHashLong[dictHashAndTagL >> ZSTD_SHORT_CACHE_TAG_BITS];
+        U32 const dictMatchIndexAndTagS = dictHashSmall[dictHashAndTagS >> ZSTD_SHORT_CACHE_TAG_BITS];
```
felixhandte, line 365: I see you're now unconditionally doing these reads where before they were only issued if they working ctx search failed. I guess that fits with the pipelining theme... But yeah I'm slightly curious how you arrived at this choice.

lib/compress/zstd_double_fast.c
```diff
@@ -312,8 +359,12 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
         U32 offset;
         size_t const h2 = ZSTD_hashPtr(ip, hBitsL, 8);
         size_t const h = ZSTD_hashPtr(ip, hBitsS, mls);
-        size_t const dictHL = ZSTD_hashPtr(ip, dictHBitsL, 8);
-        size_t const dictHS = ZSTD_hashPtr(ip, dictHBitsS, mls);
+        size_t const dictHashAndTagL = ZSTD_hashPtr(ip, dictHBitsL + ZSTD_SHORT_CACHE_TAG_BITS, 8);
+        size_t const dictHashAndTagS = ZSTD_hashPtr(ip, dictHBitsS + ZSTD_SHORT_CACHE_TAG_BITS, mls);
```
felixhandte: These should be `U32`s, right? Well ok, I guess the max hashLog is like 30 and if you add 8 it spills over... but if we're limiting the dict content to 16MB there's no point in having gigantic tables. So maybe it's worth capping the hashLog for CDicts for these levels?

Even if you don't do that, using a `size_t` is going to do funky things on a 32-bit platform...

lib/compress/zstd_double_fast.c
```diff
@@ -312,8 +359,12 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
...
+        U32 const dictMatchIndexAndTagL = dictHashLong[dictHashAndTagL >> ZSTD_SHORT_CACHE_TAG_BITS];
+        U32 const dictMatchIndexAndTagS = dictHashSmall[dictHashAndTagS >> ZSTD_SHORT_CACHE_TAG_BITS];
+        size_t const dictTagsMatchL = (dictMatchIndexAndTagL & ZSTD_SHORT_CACHE_TAG_MASK) == (dictHashAndTagL & ZSTD_SHORT_CACHE_TAG_MASK);
+        size_t const dictTagsMatchS = (dictMatchIndexAndTagS & ZSTD_SHORT_CACHE_TAG_MASK) == (dictHashAndTagS & ZSTD_SHORT_CACHE_TAG_MASK);
```
felixhandte: These are `bool`s, they don't need to be a `size_t`. Should probably just be `int`s?

lib/compress/zstd_double_fast.c
```diff
@@ -396,7 +447,7 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
                 }
             } else {
                 /* check dict long +1 match */
-                U32 const dictMatchIndexL3 = dictHashLong[dictHLNext];
+                U32 const dictMatchIndexL3 = dictHashLong[dictHLNext] >> ZSTD_SHORT_CACHE_TAG_BITS;
```
felixhandte: Looks like you're not doing the tag comparison for the next position search? How come?

lib/compress/zstd_double_fast.c
```diff
@@ -11,8 +11,43 @@
...
+    /* Always insert every fastHashFillStep position into the hash tables.
+     * Insert the other positions into the large hash table if their entry
+     * is empty.
+     */
+    for (; ip + fastHashFillStep - 1 <= iend; ip += fastHashFillStep) {
+        U32 const curr = (U32)(ip - base);
+        U32 i;
+        for (i = 0; i < fastHashFillStep; ++i) {
+            size_t const smHashAndTag = ZSTD_hashPtr(ip + i, hBitsS + ZSTD_SHORT_CACHE_TAG_BITS, mls);
+            size_t const lgHashAndTag = ZSTD_hashPtr(ip + i, hBitsL + ZSTD_SHORT_CACHE_TAG_BITS, 8);
```
felixhandte: Nit: You can just do the `+ ZSTD_SHORT_CACHE_TAG_BITS` where you initialize `hBitsS` and `hBitsL` right? Might be a bit cleaner.

lib/compress/zstd_double_fast.c
felixhandte: Also same kind of comment. Would it be cleaner overall to do the `+ TAG_BITS` adjustment when we initialize these variables rather than at every use?

lib/compress/zstd_compress.c
```diff
@@ -2131,14 +2137,31 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
...
         size_t const hSize =  (size_t)1 << cdict_cParams->hashLog;
 
-        ZSTD_memcpy(cctx->blockState.matchState.hashTable,
-               cdict->matchState.hashTable,
-               hSize * sizeof(U32));
+        if (ZSTD_CDictIndicesAreTagged(&cctx->appliedParams.cParams)){
+            size_t i;
+            for (i = 0; i < hSize; i++) {
+                U32 const taggedIndex = cdict->matchState.hashTable[i];
+                cctx->blockState.matchState.hashTable[i] = taggedIndex >> ZSTD_SHORT_CACHE_TAG_BITS;
```
felixhandte: Nit: I have a slight general preference towards decomposing the arithmetic and the memory operation. E.g., this would maybe better be:

```
U32 const taggedIndex = cdict->matchState.hashTable[i];
U32 const index = taggedIndex >> ZSTD_SHORT_CACHE_TAG_BITS;
cctx->blockState.matchState.hashTable[i] = index;
```

lib/compress/zstd_compress.c
```diff
@@ -4421,7 +4445,8 @@ static size_t ZSTD_loadZstdDictionary(ZSTD_compressedBlockState_t* bs,
                                       ZSTD_CCtx_params const* params,
                                       const void* dict, size_t dictSize,
                                       ZSTD_dictTableLoadMethod_e dtlm,
-                                      void* workspace)
+                                      void* workspace,
+                                      ZSTD_tableFillPurpose_e tfp)
```
felixhandte: This arg is probably more at home next to `dtlm`, rather than having `workspace` between them?

lib/compress/zstd_compress.c
```diff
@@ -4455,7 +4480,8 @@ ZSTD_compress_insertDictionary(ZSTD_compressedBlockState_t* bs,
                          const void* dict, size_t dictSize,
                                ZSTD_dictContentType_e dictContentType,
                                ZSTD_dictTableLoadMethod_e dtlm,
-                               void* workspace)
+                               void* workspace,
+                               ZSTD_tableFillPurpose_e tfp)
```
felixhandte: Same.

felixhandte (commented):
Ok I only got a chance to look at the dfast version for now. I have a few questions. It might be worth finding a time to chat about them? Monday maybe?

You should also benchmark a couple negative levels, since this effects them as well.

lib/compress/zstd_compress_internal.h
embg: I think there is a good reason to leave them. It validates that `SHORT_CACHE_TAG_BITS + mls` is not too large (in case someone decides to increase it in the future).

The main reason I can see to take them out would be debug performance. Is that why you think we should delete them? Or a different reason?

lib/compress/zstd_compress_internal.h
embg: I went with 8 bits because that's what rowHash uses. There's no reason to increase the number of bits, since the collision probability is already less than 1% (so we don't pay branch mispredict tax from collisions). The only reason to reduce the number of bits would be to support dictionaries larger than 16MB, which we don't care about.

If we want to take short cache out of CDict-world and into the main hashtables, then there would be a stronger reason to reduce the number of bits. I have tried to set everything up so we can just change the macro value and everything will propagate automatically in case that does happen.

lib/compress/zstd_double_fast.c
embg: Good point, will change so it uses the last 16MB.

lib/compress/zstd_double_fast.c
embg: I think this fits with the general pattern we've talked about of hoisting memory above logic. When the reads are inside branches, the compiler can't safely move them up.

Also, because DMS runs on small inputs, the ctx search will fail most of the time, so most of the time the reads will happen anyway. And because doublefast DMS is not pipelined, there is likely load bandwidth sitting unused, so it's probably not too expensive anyway.

However, the answer ultimately comes down to "I tried this and it's fast" :)

felixhandte: Makes sense.

lib/compress/zstd_compress_internal.h
felixhandte: Indeed my concern is debug performance. Maybe that's not an important concern.

lib/compress/zstd_double_fast.c
embg: `ZSTD_hashPtr` calls in `zstd_fast.c` and `zstd_double_fast.c` currently all use `size_t`. If we want to change that, it should be a separate PR and applied uniformly across all matchfinders.
```
-        size_t const dictHL = ZSTD_hashPtr(ip, dictHBitsL, 8);
-        size_t const dictHS = ZSTD_hashPtr(ip, dictHBitsS, mls);
+        size_t const dictHashAndTagL = ZSTD_hashPtr(ip, dictHBitsL + ZSTD_SHORT_CACHE_TAG_BITS, 8);
+        size_t const dictHashAndTagS = ZSTD_hashPtr(ip, dictHBitsS + ZSTD_SHORT_CACHE_TAG_BITS, mls);
```

embg: Sure, I can add the tag bits when initializing `dictHBitsL`.

lib/compress/zstd_double_fast.c
embg: I tried this and it was a couple % win for gcc11 across the board, but nuked the clang12 win. One clang12 scenario went from +45% to +30%. So I figured I'd leave it out. I can show you the numbers offline though and we can re-evaluate if you want.

Here's the commit I tested: https://github.com/embg/zstd/commit/db8033796d3b9c42251091b37bbb5cda178dd419

lib/compress/zstd_compress_internal.h
embg: I'm not sure how to value the safety win against the debug perf hit. @Cyan4973 can you weigh in?

terrelln: I'm not concerned about debug performance. We've never considered it important so far.

lib/compress/zstd_double_fast.c
terrelln: Fast hash tables will not effectively index more than 16MB. You'd need a O(16M) entry hash table. If someone wanted to index a larger table, they'd need to either use LDM, or a higher compression level.

But yeah, we definitely want the last 16MB.

terrelln:
Before merging you'll want to squash this into one commit, so we don't end up with many non-building commits in the repo.

You can either do that manually, or use one of the "squash" merge methods.

lib/compress/zstd_double_fast.c
embg: @terrelln @felixhandte I discussed offline with Felix and he raised concerns about non-standard usecases. E.g. seekable compression where every 10th frame is used as a dictionary for the other 9 frames. You could imagine someone creating a 16MB CDict for this. 

Nick, you're saying that because fast/dfast hashtables are < 1MB, they can't store 16MB worth of positions, so we don't need to worry. Is my understanding correct?

lib/compress/zstd_compress_internal.h
terrelln: Keeping 8-bits allows you to access the tag with a sub-register. E.g. if the index+tag is in `%rax` then the tag is `%al`. And you can `cmp %al, %bl` to compare about needing to mask.

The compiler should translate the mask + comparison into a comparison of the low byte sub-register.

Now, I don't know if that matters for performance, or if a mask would be just as fast. Just pointing out that there is a difference.

lib/compress/zstd_double_fast.c
embg: Discussed offline with @felixhandte . Unless we change the hashLog cap, we need to use `U64` here rather than `size_t` so the upper bits aren't cut off on 32-bit platforms. This won't affect 64-bit performance, and shouldn't have a large impact on 32-bit (which is lower priority for us).

I think we should just do this rather than messing with the hashLog cap since it is the easiest solution and won't affect 64-bit perf. I'll fix this for ZSTD_fast as well.

terrelln: @felixhandte all of hash functions return `size_t`.

However, good point about the large hash log. We need to make sure that the dictionaries hash log is <= 24.

Elliot added asserts for that, but we need to make sure that the dictionaries hash log is <= 24. We need to do the same for the row based match finder. I was able to trigger the asserts with the row-based match finder.

terrelln: E.g. `./zstd -o /dev/null -7 < ~/datasets/enwik7 --zstd=wlog=29,hlog=29`

lib/compress/zstd_double_fast.c
terrelln: Yeah, an `int` would best convey the meaning of a bool.

Though this is entirely for readability, the compiler shouldn't care.

lib/compress/zstd_compress_internal.h
```diff
@@ -745,27 +745,27 @@ ZSTD_count_2segments(const BYTE* ip, const BYTE* match,
...
 static size_t ZSTD_hash6Ptr(const void* p, U32 h) { return ZSTD_hash6(MEM_readLE64(p), h); }
 
 static const U64 prime7bytes = 58295818150454627ULL;
-static size_t ZSTD_hash7(U64 u, U32 h) { return (size_t)(((u  << (64-56)) * prime7bytes) >> (64-h)) ; }
+static size_t ZSTD_hash7(U64 u, U32 h) { assert(h <= 64); return (size_t)(((u  << (64-56)) * prime7bytes) >> (64-h)) ; }
 static size_t ZSTD_hash7Ptr(const void* p, U32 h) { return ZSTD_hash7(MEM_readLE64(p), h); }
 
 static const U64 prime8bytes = 0xCF1BBCDCB7A56463ULL;
-static size_t ZSTD_hash8(U64 u, U32 h) { return (size_t)(((u) * prime8bytes) >> (64-h)) ; }
+static size_t ZSTD_hash8(U64 u, U32 h) { assert(h <= 64); return (size_t)(((u) * prime8bytes) >> (64-h)) ; }
```
terrelln, line 769: These should also be `assert(h <= 32)` I think. Since we generally allow the user to control the MLS, we need to ensure that we always work with the 4-byte hash which is a U32 multiplication.

lib/compress/zstd_double_fast.c
terrelln: I don't like that we're not using the tag here.

I understand that you saw a clang perf regression. But it just doesn't make any sense not to use a tag here. You should try again, and see if you can get clang to do better.

If you absolutely cannot, we may consider leaving the tag comparison out. But we'd need to add a comment explaining why, so the next person doesn't scratch their head wondering why we're not using the tag.

But I strongly, strongly, prefer to use the tag comparison.

lib/compress/zstd_compress_internal.h
embg: Oh, I just realized that `h` is known at compile time anyway :)

lib/compress/zstd_double_fast.c
terrelln: You could try not doing the dictionary load until we hit this else branch, but still using the tag.

lib/compress/zstd_fast.c
```diff
@@ -463,8 +505,11 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
     while (ip1 <= ilimit) {   /* repcode check at (ip0 + 1) is safe because ip0 < ip1 */
         size_t mLength;
         size_t hash0 = ZSTD_hashPtr(ip0, hlog, mls);
-        const size_t dictHash0 = ZSTD_hashPtr(ip0, dictHLog, mls);
-        U32 dictMatchIndex = dictHashTable[dictHash0];
+
+        size_t const dictHashAndTag0 = ZSTD_hashPtr(ip0, dictHLog + ZSTD_SHORT_CACHE_TAG_BITS, mls);
+        U32 dictMatchIndexAndTag = dictHashTable[dictHashAndTag0 >> ZSTD_SHORT_CACHE_TAG_BITS];
+        size_t dictTagsMatch = (dictMatchIndexAndTag & ZSTD_SHORT_CACHE_TAG_MASK) == (dictHashAndTag0 & ZSTD_SHORT_CACHE_TAG_MASK);
```
terrelln: nit: const. Also the 2nd line should be simplified. I'd recommend just adding a helper function to compare the hash/index tags.

lib/compress/zstd_fast.c
```diff
@@ -490,26 +535,33 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
...
-                        ip0--;
-                        dictMatch--;
-                        mLength++;
-                    } /* catch up */
-                    offset_2 = offset_1;
-                    offset_1 = offset;
-                    ZSTD_storeSeq(seqStore, (size_t) (ip0 - anchor), anchor, iend, OFFSET_TO_OFFBASE(offset), mLength);
-                    break;
+                    /* To replicate extDict parse behavior, we only use dict matches when the normal matchIndex is invalid */
+                    if (matchIndex <= prefixStartIndex) {
```
terrelln, line 555: It seems weird to put this condition after checking that we have a match.

Maybe the compiler generates slightly better code in this case for some dumb reason, but it seems like clearly the wrong order.

Could you do `dictTagsMatch && matchIndex <= prefixStartIndex` on line 540?

lib/compress/zstd_double_fast.c
terrelln: Yes

The fast/dfast match finders will never be good at finding matches > 16 MB in the past. So there is no reason to support it IMO.

We have a great match finder for that purpose, LDM, and if someone wants that behavior they should use that.

lib/compress/zstd_compress_internal.h
embg: Can you explain the last bit? `u` and `prime8bytes` are both `U64`, right?

lib/compress/zstd_double_fast.c
embg: To clarify, the clang numbers were still very good -- just not as good. And gcc was improved across the board. So I think it's fine to just cherry-pick https://github.com/embg/zstd/commit/db8033796d3b9c42251091b37bbb5cda178dd419.

lib/compress/zstd_fast.c
embg: Which variable should be const? Agree on simplifying the dictTagsMatch logic.

lib/compress/zstd_fast.c
embg: It's really unfortunate that clang doesn't like `dictTagsMatch && matchIndex <= prefixStartIndex` (or the reverse). I tried one permutation [here](https://github.com/facebook/zstd/pull/3152/commits/a243a5343dfdbb14ed08a216e2661429438aeaf2) (tried the other permutations locally) and it turned out to be a regression for some clang scenarios.

I think the problem is that if you combine the branches like that, you have to add a branch for the ctx search: 
```
if (dictTagsMatch && matchIndex <= prefixStartIndex) {
    // check dict match
} else if (matchIndex > prefixStartIndex) {
    // check ctx match
}
```
The extra branch appears to mess up codegen.

The other option is to split it into two nested branches:
```
if (matchIndex <= prefixStartIndex) {
    if (dictTagsMatch) { 
        // check dict match
    }
} else {
    // check ctx match
}
```
Either clang or gcc didn't like that (don't remember which, it was one of about a dozen setups I tested locally without pushing to github).

I can definitely add a comment explaining the situation though, agree it's unclear.

lib/compress/zstd_fast.c
terrelln: shouldn't both be? Or are they modified later?

lib/compress/zstd_fast.c
embg: In a separate PR (not anytime soon, but someday) it would be interesting to test whether we can just remove the `matchIndex <= prefixStartIndex` check entirely. There's no fundamental reason why we have to parse the same as extDict.

lib/compress/zstd_compress_internal.h
terrelln: I mean that we always use `ZSTD_hashPtr()`. Which dispatches to one of the hashes. Some of the hashes only support `h <= 32`.

Maybe the best course of action is to `assert(h <= 32)` in `ZSTD_hashPtr()`.

lib/compress/zstd_fast.c
embg: One more point of clarification: we can set up the branches any way we want to get large cold dict wins. The hyper-tuning is just about preventing hot dict regressions, because we really want to avoid gating short cache if possible.

lib/compress/zstd_compress_internal.h
embg: Ok I understand. My preference would be to leave it here because it's closer to where we actually use the condition (easier to understand immediately why the assert is needed). But if you think it's better in `ZSTD_hashPtr` I'm happy to move it there.

lib/compress/zstd_fast.c
embg: Yes, modified at the end of the loop:
```
            /* Prepare for next iteration */
            dictMatchIndexAndTag = dictHashTable[dictHashAndTag1 >> ZSTD_SHORT_CACHE_TAG_BITS];
            dictTagsMatch = (dictMatchIndexAndTag & ZSTD_SHORT_CACHE_TAG_MASK) == (dictHashAndTag1 & ZSTD_SHORT_CACHE_TAG_MASK);
```

lib/compress/zstd_compress_internal.h
terrelln: We can leave the `h <= 64` assert here, but also add an `h <= 32` assert to `ZSTD_hashPtr()` and explain why in a comment.

lib/compress/zstd_compress_internal.h
Cyan4973: Agreed, such a simple `assert()` is fine with regards to debug performance.
Plus, if `h` is static and the compiler can understand it, it will probably remove the `assert()` anyway.

embg:
See changes since last review by @felixhandte and @terrelln [here](https://github.com/facebook/zstd/compare/a1063e1436ef4053e9cddf5f9799938a906a2fd0..a040b1dcdb1453a4cb5ed93b1d3746c2b6fdf959). Responded to all feedback:
* Cherry-pick [add short cache to ip+1 long search](https://github.com/embg/zstd/pull/29/commits/07f2096e8bc54cc46e8118e4ec3a85c115f22ada)
* Truncate dictionaries over 16 MB in ZSTD_loadDictionaryContent
* Move tag size into hashBits variables
* Added ZSTD_ prefix to writeTaggedIndex
* Added assert hBits <= 32 to ZSTD_hashPtr
* Scale hashLog down to 24 in ZSTD_adjustCParams_internal
* Split memory and arithmetic in ZSTD_resetCCtx_byCopyingCDict
* Moved ZSTD_tableFillPurpose_e next to ZSTD_dictTableLoadMethod_e
* Pull out tag comparison logic into ZSTD_comparePackedTags
* Convert dictTagsMatch boolean variables from size_t to int

See PR summary for remaining blockers.

lib/compress/zstd_compress.c
```diff
@@ -4214,22 +4247,33 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
...
-        U32 const maxDictSize = ZSTD_CURRENT_MAX - ZSTD_WINDOW_START_INDEX;
-        /* We must have cleared our windows when our source is this large. */
-        assert(ZSTD_window_isEmpty(ms->window));
-        if (loadLdmDict)
-            assert(ZSTD_window_isEmpty(ls->window));
+    {   /* Ensure large dictionaries can't cause index overflow */
+        U32 maxDictSize;
+        int const CDictTaggedIndices = ZSTD_CDictIndicesAreTagged(&params->cParams);
+        if (CDictTaggedIndices && tfp == ZSTD_tfp_forCDict) {
+            /* TODO comment */
```
terrelln: Do this todo

lib/compress/zstd_compress_internal.h
```diff
@@ -1264,6 +1268,31 @@ MEM_STATIC void ZSTD_debugTable(const U32* table, U32 max)
 
 #endif
 
+/* Short Cache */
+
+typedef enum {
+    /* TODO: this is a good place to document short cache */
```
terrelln: Yeah, this enum is generic and not really related to short cache except, but right after it would be a good place.

lib/compress/zstd_compress_internal.h
```diff
@@ -1264,6 +1268,31 @@ MEM_STATIC void ZSTD_debugTable(const U32* table, U32 max)
...
+
+/* Helper function for ZSTD_fillHashTable and ZSTD_fillDoubleHashTable */
+MEM_STATIC void ZSTD_writeTaggedIndex(U32* const hashTable, size_t hashAndTag, U32 index) {
+    size_t const hash = hashAndTag >> ZSTD_SHORT_CACHE_TAG_BITS;
+    U32 const tag = (U32)(hashAndTag & ZSTD_SHORT_CACHE_TAG_MASK);
+    assert(index >> (32 - ZSTD_SHORT_CACHE_TAG_BITS) == 0);
+    hashTable[hash] = (index << ZSTD_SHORT_CACHE_TAG_BITS) | tag;
+}
+
+/* Helper function for short cache matchfinders */
```
terrelln: nit: Also explain what these functions do in the comments.

lib/compress/zstd_compress.c
```diff
@@ -2131,14 +2144,33 @@ static size_t ZSTD_resetCCtx_byCopyingCDict(ZSTD_CCtx* cctx,
...
+            for (i = 0; i < hSize; i++) {
+                U32 const taggedIndex = cdict->matchState.hashTable[i];
+                U32 const index = taggedIndex >> ZSTD_SHORT_CACHE_TAG_BITS;
+                cctx->blockState.matchState.hashTable[i] = index;
+            }
+        } else {
+            ZSTD_memcpy(cctx->blockState.matchState.hashTable,
+                cdict->matchState.hashTable,
+                hSize * sizeof(U32));
+        }
```
terrelln: nit: Can you extract this out into a function that copies a table. E.g.

```
void ZSTD_copyCDictTableIntoCCtx(ZSTD_CCtx const* cctx, U32* dst, U32 const* src, size_t size);
```

And use it here and below.

lib/compress/zstd_compress.c
```diff
@@ -4214,22 +4247,33 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
...
-        assert(ZSTD_window_isEmpty(ms->window));
-        if (loadLdmDict)
-            assert(ZSTD_window_isEmpty(ls->window));
+    {   /* Ensure large dictionaries can't cause index overflow */
+        U32 maxDictSize;
+        int const CDictTaggedIndices = ZSTD_CDictIndicesAreTagged(&params->cParams);
+        if (CDictTaggedIndices && tfp == ZSTD_tfp_forCDict) {
+            /* TODO comment */
+            maxDictSize = (1u << (32 - ZSTD_SHORT_CACHE_TAG_BITS)) - ZSTD_WINDOW_START_INDEX;
+            assert(maxDictSize + ZSTD_WINDOW_START_INDEX < ZSTD_CURRENT_MAX);
```
terrelln: nit: This assert won't really work, since they are U32s, and could easily wrap around, follow the comment below to be correct by construction.

lib/compress/zstd_compress.c
```diff
@@ -4214,22 +4247,33 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
...
+            assert(!loadLdmDict);
+        } else {
+            /* Allow the dictionary to set indices up to exactly ZSTD_CURRENT_MAX.
+             * Dictionaries right at the edge will immediately trigger overflow
+             * correction, but I don't want to insert extra constraints here.
+             */
+            maxDictSize = ZSTD_CURRENT_MAX - ZSTD_WINDOW_START_INDEX;
```
terrelln: nit: Don't make this an else.

Start with `maxDictSize = ZSTD_CURRENT_MAX - ZSTD_WINDOW_START_INDEX`. Then if short-cache, set it to `MIN(maxDictSize, shortCacheMaxDictSize)`.

The current construction assumes that `shortCacheMaxDictSize < ZSTD_CURRENT_MAX - ZSTD_WINDOW_START_INDEX`. Re-writing it this way avoids that assumption.

terrelln (commented):
This is looking good!

lib/compress/zstd_compress.c
embg: This will make it easier to measure future improvements from AVX2 and/or PREFETCHW. (And it's cleaner). Good idea.

lib/compress/zstd_double_fast.c
```diff
@@ -11,8 +11,43 @@
...
+            size_t const lgHashAndTag = ZSTD_hashPtr(ip + i, hBitsL, 8);
+            if (i == 0) {
+                ZSTD_writeTaggedIndex(hashSmall, smHashAndTag, curr + i);
+            }
+            if (i == 0 || hashLarge[lgHashAndTag >> ZSTD_SHORT_CACHE_TAG_BITS] == 0) {
+                ZSTD_writeTaggedIndex(hashLarge, lgHashAndTag, curr + i);
+            }
+            /* Only load extra positions for ZSTD_dtlm_full */
+            if (dtlm == ZSTD_dtlm_fast)
+                break;
```
felixhandte, line 46: From a cursory inspection, it seems like `dtlm == ZSTD_dtlm_full` when `tfp == ZSTD_tfp_forCDict` and `dtlm == ZSTD_dtlm_fast` when `tfp == ZSTD_tfp_forCCtx`. Is that true? If so, we can certainly remove this branch, and maybe we can actually deduplicate these enums through the stack?

lib/compress/zstd_fast.c
felixhandte: Man I dunno this seems janky to me. You're adding a branch and it's weird that the dict search is coming ahead of the working ctx search. What about:

```
if (matchIndex >= prefixStartIndex) {
  if (MEM_read32(match) == MEM_read32(ip0)) {
    // found ctx match
  }
} else if (dictTagsMatch) {
  if (MEM_read32(dictMatch) == MEM_read32(ip0)) {
    // found dict match
  }
}

```

embg: I think I measured this and it regressed some scenario but I don't remember for sure. Will measure and report back, would be great if we can clean this up.

lib/compress/zstd_double_fast.c
embg: You are correct. @terrelln suggested to have two enums since they are orthogonal concepts (even though in our current codebase they are 100% correlated).

Discussed offline: I'll assert the current correlation in both implementations, with a comment saying it's ok to delete the asserts if they weren't triggered by accident.

embg:
Responded to all feedback from @terrelln. Will respond to feedback from @felixhandte tomorrow (just have to measure the proposed change to fast_DMS and fix a nit). See PR summary for remaining blockers.

lib/compress/zstd_compress.c
```diff
@@ -2109,6 +2109,22 @@ ZSTD_resetCCtx_byAttachingCDict(ZSTD_CCtx* cctx,
     return 0;
 }
 
+static void ZSTD_copyCDictTableIntoCCtx(U32* dst, U32 const* src, size_t tableSize,
+                                        ZSTD_compressionParameters const* cParams) {
```
embg, line 2113: typo, too many spaces for some reason

lib/compress/zstd_fast.c
embg: Here are the numbers. The colors aren't consistent between graphs, be careful to look at each legend (sorry about that, not sure why plotly didn't order things the same way for each one).

It looks like your version is a few % win for clang and a few % loss for gcc. In the compiler=gcc and DictSize=4K scenario, your version takes the win to 0 or slightly negative. However, these numbers are *on top* of the already-landed fast DMS pipeline, so it's ok if some scenarios are slightly negative.

newplot (10)
newplot (11)
newplot (12)

embg: Discussed offline: the proposed change *shouldn't* affect perf... but of course in the real world we know our compilers can be a little picky sometimes :) @felixhandte and I decided to land as-is for now, and at some point we'll see if we can figure out why GCC doesn't like the less janky version.

embg:
Responded to all feedback. Will start fuzzers on my devserver and update PR summary with extDict regression graphs.

terrelln (approved):

felixhandte (approved):
Ship it!

embg:
Fuzzers didn't find anything (100 seconds * 10 workers for each target). Merging!

nadavrot:
Cool optimization. I love the chart.

PR #2722, opened 2021-07-07
Assembly implementation of 4X1 & 4X2 Huffman

terrelln:
Assembly in `huf_decompress_x86_bmi2.S`.

* `proba32.dat` is 1MB of random symbols in `[0, 32)`. Zstd compresses as all literals and uses 4X2 to decompress.
* `proba128.dat` is 1MB of random symbols in `[0, 128)`. Zstd compresses as all literals and uses 4X1 to decompress.
* All files are compressed with level 1 to get the most # of literals.

The assembly is wrapped by a C function that handles the initialization & ending conditions. Once the streams get close to the end, the C takes over and uses `HUF_decompress1X*()` to finish, similarly to the normal 4X1 and 4X2 loops.

| File | gcc-11 | gcc-11 asm | gcc-11 speedup | clang-12 | clang-12 asm | clang-12 speedup |
| --- | --- | --- | --- | --- | --- | --- |
| proba32.dat | 1746 | 2357 | +35% | 1449 | 2354 | +62% |
| proba128.dat | 944 | 1318 | +39% | 906 | 1320 | +45% |
| silesia.tar | 1159 | 1244 | +7.3% | 1131 | 1246 | +10.1% |
| dickens | 1064 | 1160 | +9.0% | 1036 | 1169 | +12.8% |
| mozilla | 1000 | 1064 | +6.4% | 988 | 1071 | +8.4% |
| mr | 1264 | 1344 | +6.3% | 1203 | 1348 | +12.0% |
| nci | 1786 | 1855 | +3.8% | 1771 | 1819 | +2.7% |
| ooffice | 876 | 989 | +12.9% | 814 | 995 | +22.2% |
| osdb | 1209 | 1302 | +7.6% | 1216 | 1300 | +6.9% |
| reymont | 1073 | 1124 | +4.7% | 1069 | 1134 | +6.1% |
| samba | 1439 | 1529 | +6.2% | 1418 | 1525 | +7.5% |
| sao | 862 | 1017 | +18.0% | 868 | 1017 | +17.1% |
| webster | 1113 | 1183 | +6.3% | 1088 | 1189 | +9.3% |
| xml | 1796 | 1851 | +3.1% | 1753 | 1826 | +4.2% |
| x-ray | 1042 | 1280 | +22.8% | 877 | 1276 | +45.5% |

PR #3155, opened 2022-06-07
Add fast huf_dec with generic C and tuned aarch64 assembly

JunHe77:
This includes implementations of a generic C version of [fast decode](https://github.com/facebook/zstd/pull/2722) and a tuned 4x1 assembly version for Arm.
For silesia, observed 3.9% for sao, ~2% for mozilla/ooffice/osdb/x-ray.
As the author of the original algorithm, could you pls help to review this, @terrelln ? Thanks a lot.

### DIFF
--- a/lib/common/portability_macros.h
+++ b/lib/common/portability_macros.h
@@ -114,6 +114,14 @@
 #  define ZSTD_ASM_SUPPORTED 0
 #endif
 
+#if !defined(ZSTD_DISABLE_ASM) &&                                 \
+    ZSTD_ASM_SUPPORTED &&                                         \
+    defined(__aarch64__)
+#  define ZSTD_ENABLE_ASM_AARCH64 1
+#else
+# define ZSTD_ENABLE_ASM_AARCH64 0
+#endif
+
 /**
  * Determines whether we should enable assembly for x86-64
  * with BMI2.
--- a/lib/decompress/huf_decompress.c
+++ b/lib/decompress/huf_decompress.c
@@ -57,17 +57,26 @@
 #endif
 #define HUF_ASM_DECL HUF_EXTERN_C
 
+#ifndef HUF_ENABLE_FAST_DEC
+#  if (defined(__x86_64__) || defined(__aarch64__))
+#    define HUF_ENABLE_FAST_DEC 1
+#  else
+#    define HUF_ENABLE_FAST_DEC 0
+#  endif
+#endif
+
 #if DYNAMIC_BMI2 || (ZSTD_ENABLE_ASM_X86_64_BMI2 && defined(__BMI2__))
 # define HUF_NEED_BMI2_FUNCTION 1
 #else
 # define HUF_NEED_BMI2_FUNCTION 0
 #endif
 
-#if !(ZSTD_ENABLE_ASM_X86_64_BMI2 && defined(__BMI2__))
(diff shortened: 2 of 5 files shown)


terrelln:
Whats the difference between the C version and the ASM version?

It looks like the gains for the aarch64 version are smaller than the gains for the x86-64 version, which could make sense because aarch64 has way less register pressure, and that was the main constraint on x86-64.

For those smaller gains, I'd be a bit hesitant about merging an ASM implementation. But would be more open to a generic C version.

JunHe77:
Thanks for the comments, @terrelln . I understand the maintenance effort for an assembly implementation. Following is the change comparison between C and ASM for silesia at L8. Pls check.

|data file|ASM/C on N1|ASM/C on A72|
|---------|---------------|----------------|
|dickens|-0.19%|1.22%|
|mozilla|1.17%|0.33%|
|mr|0.21%|-1.38%|
|nci|0.29%|0.76%|
|ooffice|1.01%|-0.12%|
|osdb|0.92%|0.54%|
|reymont|0.11%|1.27%|
|samba|0.19%|1.51%|
|sao|2.32%|1.87%|
|webster|0.17%|0.42%|
|x-ray|0.97%|1.42%|
|xml|-0.21%|0.16%|

In 5 of 12 cases, ASm version achieves 1%+ better performance (~2% for sao on both N1 and A72).

terrelln:
Is the C version the current zstd code, or the fast decode C you've written? If it is the latter, what is the difference between zstd's code and the tuned C?

JunHe77:
The C version is a kind of "rewritten asm in C" of [huf_decompress_amd64.S](https://github.com/facebook/zstd/blob/dev/lib/decompress/huf_decompress_amd64.S). I started the porting on Arm with C, then hand tuned asm version for both 4x1 and 4x2.
For C version, both 4x1 and 4x2 showed uplift on Arm. While further boosts are observed on certain test data with 4x1_asm, no obvious boost is found with 4x2_asm. That's why there is only 4x1 asm version in this PR. 😄

JunHe77:
Hi @terrelln, anything I need to follow for this PR? Thanks.

terrelln (commented):
Thanks for the PR @JunHe77, and sorry for the delay in review!

I'm working on a PR for C decoders, similar to what you added here. I was able to eek out a bit more speed (on x86-64, will measure aarch64 as well), and clean up the code a little bit.

I will put my PR up next week, at which point please feel free to benchmark your C / aarch64 ASM implementations against the PR. If there are meaningful gains, we can look into merging.

JunHe77:
Well received, @terrelln . Will do the benchmark once your PR is ready. Thanks.

terrelln:
Hi @JunHe77, I've just merged PR #3449.

I've found that on my M1 chip those loops perform at least as well as your C versions. But I would be happy to accept patches to the fast C loops if you can find a faster variant.

If the assembly implementation is still significantly faster than the fast C variant, I would be willing to accept the aarch64 assembly implementation. But it would have to be disabled by default, and require the definition of a macro in the build process `ZSTD_ENABLE_AARCH64_ASM` to use.

The reason it would have to be disabled by default, is that we have no way to continuously fuzz the code. All of our fuzzers run on x86-64 and i386. We want to minimize the amount of code that isn't fuzzed in zstd. I trust that your assembly is correct, but I don't feel comfortable including any non-trivial code in zstd that isn't fuzzed.

If we merge it disabled by default, and later oss-fuzz adds support for aarch64, we could start fuzzing that code and switch it to enabled by default.

PR #3160, opened 2022-06-13
Fix big endian ARM NEON path

danlark1:
It is not using the NEON acceleration but the bit grouping was applied


embg (approved):

embg:
The Travis CI failure is not related to this PR, so I'll go ahead and merge this.

PR #3161, opened 2022-06-14
[contrib] largeNbDicts bugfix + improvements

embg:
These are local changes I've been cherry-picking on builds for a while. Figured it's time to put them on GitHub.
* Fixes a bug in the memory estimation found by @terrelln 
* Fixes a bug which prevented `forceAttach/forceCopy` from working properly (`ZSTD_compress_usingCDict -> ZSTD_compress2`)
* Prints results to a csv file with the executable name
    * The idea is you bake the commit hash and compiler into the executable name. (This is my setup).
    * largeNbDicts will append the best speed to the CSV file, so you can run multiple times and then get statistics. I like to interleave runs between different versions to correct for background load changes.
    * I have a benchmark runner script which creates a directory tree of CSV files for different scenarios, and then a python script which aggregates all the statistics. Will put them on GitHub when I have time :)

### DIFF
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -577,11 +577,12 @@ typedef struct {
     cdict_collection_t dictionaries;
 } compressInstructions;
 
-compressInstructions createCompressInstructions(cdict_collection_t dictionaries)
+compressInstructions createCompressInstructions(cdict_collection_t dictionaries, ZSTD_CCtx_params* cctxParams)
 {
     compressInstructions ci;
     ci.cctx = ZSTD_createCCtx();
     CONTROL(ci.cctx != NULL);
+    ZSTD_CCtx_setParametersUsingCCtxParams(ci.cctx, cctxParams);
     ci.nbDicts = dictionaries.nbCDict;
     ci.dictNb = 0;
     ci.dictionaries = dictionaries;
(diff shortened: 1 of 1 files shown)


contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -703,6 +704,22 @@ static int benchMem(slice_collection_t dstBlocks,
...
+    char* csvFileName = malloc(strlen(exeName) + 5);
+    strcpy(csvFileName, exeName);
+    strcat(csvFileName, ".csv");
+    FILE* csvFile = fopen(csvFileName, "r");
+    if (!csvFile) {
+        csvFile = fopen(csvFileName, "wt");
+        assert(csvFile);
+        fprintf(csvFile, "%s\n", exeName);
+    } else {
+        csvFile = fopen(csvFileName, "at");
```
terrelln, line 719: Does this leak a `FILE` handle?

Does opening with `"at"` not work for empty files? The spec says the files is created if it doesn't exist.

embg: I have it set up to write the executable name when the file doesn't already exist. That's why I have a branch. Agree that it's leaking the first `FILE` handle from `fopen(csvFileName, "r")` though, nice catch.

embg:
* Fixed `FILE` handle leak
* Fixed a bug which prevented forceAttach/forceCopy from working properly (ZSTD_compress_usingCDict -> ZSTD_compress2).

contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -703,6 +706,23 @@ static int benchMem(slice_collection_t dstBlocks,
...
+    strcpy(csvFileName, exeName);
+    strcat(csvFileName, ".csv");
+    FILE* csvFile = fopen(csvFileName, "r");
+    if (!csvFile) {
+        csvFile = fopen(csvFileName, "wt");
+        assert(csvFile);
+        fprintf(csvFile, "%s\n", exeName);
+    } else {
+        free(csvFileName);
+        csvFile = fopen(csvFileName, "at");
```
terrelln: You've freed the wrong thing
```suggestion
        fclose(csvFile);
        csvFile = fopen(csvFileName, "at");
```

embg: ah whoops thanks

PR #3165, opened 2022-06-19
programs/timefn: Don't use clock() on Unix when C < 11 to improve multi-thread benchmark

### LINKED ISSUE #3163: Strange performance and scalability issues with some of the build systems
**Describe the bug**

After reading a reading a recent [Phoronix benchmark](https://www.phoronix.com/scan.php?page=article&item=hp-devone-linux&num=3) (a bit down the page) I decided to investigate why Arch Linux was so much slower (10-20x) for zstd performance. It turned out that something is wrong with some of the build systems included with zstd!

When zstd is built with the cmake or meson build systems there is negative scaling with the number of threads, while when building with the Makefile in the top level directory, there is positive scaling with the number of threads.

**To Reproduce**
1. Build with build system you want to investigate. One of:
    * Plain `make`
    * `mkdir build && cmake ../zstd-1.5.2/build/cmake/ && make`
    * `meson setup builddir && cd builddir && ninja`
2. Test the resulting binary on a large file, I used the [FreeBSD image](https://download.freebsd.org/releases/amd64/amd64/ISO-IMAGES/13.1/FreeBSD-13.1-RELEASE-amd64-memstick.img) as this is what Phoronix Test Suite used, albeit from a older version that I can't find. I can however reproduce the same issue with the linked file. Use the following pair of benchmark commands and compare scaling:
  * `path/to/zstd -T1 -b4 path/to/FreeBSD-13.1-RELEASE-amd64-memstick.img`
  * `path/to/zstd -T6 -b4 path/to/FreeBSD-13.1-RELEASE-amd64-memstick.img` (adjust -T6 based on the number of cores you have)

*Note!* I see the same pattern at other compression levels such as 6 and 8, not just 4. So that value doesn't really appear to matter, as long as it is consistent of course.

**Expected behavior**
I expect that all build systems should result in binaries with roughly the same behaviour. Performance and scaling should be similar.

**Actual results**

The output below has been abbreviated for clarity, repeated command lines has been elided only showing the output. Three runs for each combination of program and flags has been performed. As can be seen the results are relatively consistent run-to-run (at least consistent enough given the huge discrepancies).

1. CMake

```console
$ programs/zstd -T1 -b4 ~/Downloads/FreeBSD-13.1-RELEASE-amd64-memstick.img
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1108.5 MB/s, 4999.1 MB/s
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1152.9 MB/s, 5006.7 MB/s
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1102.1 MB/s, 4978.5 MB/s

$ programs/zstd -T6 -b4 ~/Downloads/FreeBSD-13.1-RELEASE-amd64-memstick.img
4#md64-memstick.img :1172165120 -> 781637623 (x1.500),  717.0 MB/s, 4940.0 MB/s
4#md64-memstick.img :1172165120 -> 781637623 (x1.500),  759.3 MB/s, 4893.5 MB/s
4#md64-memstick.img :1172165120 -> 781637623 (x1.500),  697.5 MB/s, 4869.6 MB/s
```

2. Meson

```console
$ programs/zstd -T1 -b4 ~/Downloads/FreeBSD-13.1-RELEASE-amd64-memstick.img
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1097.0 MB/s, 5029.3 MB/s
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1098.2 MB/s, 4970.2 MB/s
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1117.8 MB/s, 4952.6 MB/s

$ programs/zstd -T6 -b4 ~/Downloads/FreeBSD-13.1-RELEASE-amd64-memstick.img
4#md64-memstick.img :1172165120 -> 781637623 (x1.500),  735.0 MB/s, 4982.8 MB/s
4#md64-memstick.img :1172165120 -> 781637623 (x1.500),  758.6 MB/s, 4966.9 MB/s
4#md64-memstick.img :1172165120 -> 781637623 (x1.500),  727.9 MB/s, 4949.7 MB/s
```

3. Makefile

```console
$ ./zstd -T1 -b4 ~/Downloads/FreeBSD-13.1-RELEASE-amd64-memstick.img
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1118.2 MB/s, 4971.0 MB/s
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1105.4 MB/s, 4931.3 MB/s
4#md64-memstick.img :1172165120 -> 781156418 (x1.501), 1150.6 MB/s, 4930.1 MB/s

$ ./zstd -T6 -b4 ~/Downloads/FreeBSD-13.1-RELEASE-amd64-memstick.img
4#md64-memstick.img :1172165120 -> 781637623 (x1.500), 3518.0 MB/s, 4898.2 MB/s
4#md64-memstick.img :1172165120 -> 781637623 (x1.500), 3486.3 MB/s, 4917.0 MB/s
4#md64-memstick.img :1172165120 -> 781637623 (x1.500), 3528.1 MB/s, 4900.8 MB/s
```

**Analysis of results**

For CMake and Meson: it can be seen that the performance goes down between 1 thread and 6 threads: ~1100 MB/s to ~700 MB/s.

For plain make, the performance goes up between 1 thread and 6 threads: ~1100 MB/s to ~3500 MB/s.

Decompression speed (the second value) does not seem to vary significantly across the experiments however.

 - OS: Arch Linux
 - Version 1.5.2 (upstream tarball)
 - Compiler: GCC 12.1.0
 - Flags: Defaults for each build system, though I tested with some basics such as -O2, but it did not affect the overall behaviour.
 - Other relevant hardware specs: AMD Ryzen 5 5600X 6-Core Processor
 - Build system: Multiple ones, that is the whole point of this bug

**Additional context**

jbeich:
Fixes https://github.com/facebook/zstd/issues/3163
See also https://github.com/freebsd/freebsd-src/commit/6882d53b7fb8
See also https://github.com/aosp-mirror/platform_bionic/commit/78e9ebc3b968

C90 aka `clock()` fallback being suboptimal seems to be known issue:
https://github.com/facebook/zstd/blob/b7b7edb3a3017ac8e16d7eb2dbede45168560c58/programs/timefn.c#L138-L143

### DIFF
--- a/programs/timefn.c
+++ b/programs/timefn.c
@@ -8,6 +8,7 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
+#define _POSIX_C_SOURCE	200809L /* clock_gettime */
 
 /* ===  Dependencies  === */
 
@@ -82,10 +83,8 @@ PTime UTIL_getSpanTimeNano(UTIL_time_t clockStart, UTIL_time_t clockEnd)
 }
 
 
-/* C11 requires timespec_get, but FreeBSD 11 lacks it, while still claiming C11 compliance.
-   Android also lacks it but does define TIME_UTC. */
 #elif (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* C11 */) \
-    && defined(TIME_UTC) && !defined(__ANDROID__)
+    || defined(CLOCK_MONOTONIC) /* POSIX.1-2001 (optional) */
 
 #include <stdlib.h>   /* abort */
 #include <stdio.h>    /* perror */
@@ -95,10 +94,17 @@ UTIL_time_t UTIL_getTime(void)
     /* time must be initialized, othersize it may fail msan test.
      * No good reason, likely a limitation of timespec_get() for some target */
     UTIL_time_t time = UTIL_TIME_INITIALIZER;
+#if defined(CLOCK_MONOTONIC)
+    if (clock_gettime(CLOCK_MONOTONIC, &time) != 0) {
+        perror("timefn::clock_gettime(CLOCK_MONOTONIC)");
+        abort();
+    }
+#else
     if (timespec_get(&time, TIME_UTC) != TIME_UTC) {
         perror("timefn::timespec_get");
         abort();
     }
+#endif
     return time;
 }
 
--- a/programs/timefn.h
+++ b/programs/timefn.h
@@ -55,10 +55,8 @@ extern "C" {
     typedef PTime UTIL_time_t;
     #define UTIL_TIME_INITIALIZER 0
 
-/* C11 requires timespec_get, but FreeBSD 11 lacks it, while still claiming C11 compliance.
-   Android also lacks it but does define TIME_UTC. */
 #elif (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* C11 */) \
-    && defined(TIME_UTC) && !defined(__ANDROID__)
+    || defined(CLOCK_MONOTONIC) /* POSIX.1-2001 (optional) */
 
     typedef struct timespec UTIL_time_t;
     #define UTIL_TIME_INITIALIZER { 0, 0 }


heftig:
I just came to the same conclusion. The "bad performance" is a measurement error in zstdcli, not an actual performance issue. The libzstd produced with c11 and c99 are identical.

Cyan4973:
I think this patch goes in the right direction. I'm actually surprised we don't use `clock_gettime()` already. Maybe it used to be present and was later removed, or maybe I confuse with `lz4`. Either way, it's a welcome addition, clearly better than `clock()`.

minor :
for a benchmark function, which only cares about relative differences between 2 time measurements, I would have rather expected `CLOCK_MONOTONIC`.

I understand that `CLOCK_REALTIME` is closer in meaning to `TIME_UTC`, but the main reason to use `TIME_UTC` with `timespec_get()` is that it's the only base clearly defined in the `C11` standard.

_edit_ : I just checked, and indeed `clock_gettime()` used to be present. Unsure why it was removed later on. Maybe to remove the nb of time measurement variants, possibly.

programs/timefn.c
```diff
@@ -95,10 +93,18 @@ UTIL_time_t UTIL_getTime(void)
     /* time must be initialized, othersize it may fail msan test.
      * No good reason, likely a limitation of timespec_get() for some target */
     UTIL_time_t time = UTIL_TIME_INITIALIZER;
+    /* Android < 10 lacks timespec_get but does define TIME_UTC. */
+#if defined(TIME_UTC) && !defined(__ANDROID__)
     if (timespec_get(&time, TIME_UTC) != TIME_UTC) {
         perror("timefn::timespec_get");
         abort();
     }
+#else
```
Cyan4973, line 102: Note : What happens in the following theoretical scenario ? 
- C11, aka `__STDC_VERSION__ >= 201112L`
- but `TIME_UTC` is *not* defined, or `__ANDROID__` is defined
- yet `CLOCK_REALTIME` is not defined either.

I would expect a compilation error on `clock_gettime(CLOCK_REALTIME,` line.

Cyan4973:
Note : 
I believe this patch doesn't fix https://github.com/facebook/zstd/issues/3163 .

It does fix `-std=gnu99` for example, but `-std=c99` is too strong and does not define posix-only symbols, such as `clock_gettime()`. 

It's not necessarily a problem for *this* PR, which provides `clock_gettime()` as a time measurement backup for compliant systems.

It just shows there is a bit more to do to address #3163 specifically.

programs/timefn.c
jbeich: A bit unrealistic:
- Missing `TIME_UTC` implies incomplete C11 support
- `__ANDROID__` implies Unix (POSIX) thus `CLOCK_REALTIME`

Fuchsia-based Android can be such but I don't know much.

Cyan4973: Unexpected, sure, but unfortunately, experience learned us that it unfortunately happens. `C11` without `timespec_get()` is a thing, even if it should not.

So we have 2 strategies : 
- let the problem there, and react later when it happens
- preemptively avoid the situation

It looks to me that option 2 is not too hard to obtain :  
just keep the `&& defined(TIME_UTC) && !defined(__ANDROID__)` part of the test above, make it part of the first `C11` test, using parenthesis.

This way, afterwards, if `(defined(TIME_UTC) && !defined(__ANDROID__))` is `false`, then it necessarily implies that the `#else` is fine.

jbeich:
> It does fix `-std=gnu99` for example, but `-std=c99` is too strong and does not define posix-only symbols, such as `clock_gettime()`.

`gnu` in `-std=gnu99` covers [GNU compiler extensions](https://gcc.gnu.org/onlinedocs/gcc/C-Extensions.html) that can be used on non-GNU or freestanding systems.

I don't have a Linux system around but `<features.h>` suggests a mix of POSIX.1-2008 + BSD + SysV features are available by default:
```c
/* If nothing (other than _GNU_SOURCE and _DEFAULT_SOURCE) is defined,
   define _DEFAULT_SOURCE.  */
#if (defined _DEFAULT_SOURCE					\
     || (!defined __STRICT_ANSI__				\
	 && !defined _ISOC99_SOURCE && !defined _ISOC11_SOURCE	\
	 && !defined _ISOC2X_SOURCE				\
	 && !defined _POSIX_SOURCE && !defined _POSIX_C_SOURCE	\
	 && !defined _XOPEN_SOURCE))
# undef  _DEFAULT_SOURCE
# define _DEFAULT_SOURCE	1
#endif
...
/* If none of the ANSI/POSIX macros are defined, or if _DEFAULT_SOURCE
   is defined, use POSIX.1-2008 (or another version depending on
   _XOPEN_SOURCE).  */
#ifdef _DEFAULT_SOURCE
# if !defined _POSIX_SOURCE && !defined _POSIX_C_SOURCE
#  define __USE_POSIX_IMPLICITLY	1
# endif
# undef  _POSIX_SOURCE
# define _POSIX_SOURCE	1
# undef  _POSIX_C_SOURCE
# define _POSIX_C_SOURCE	200809L
#endif
```

jbeich:
Ah, I see: `-std=c99` has `__STRICT_ANSI__`.

programs/timefn.c
jbeich: It maybe easier to reason when C11 conditional is inverted i.e., prefer `CLOCK_MONOTONIC`. C standards (unlike C++ standards) evolve very slowly and the result is often meh (e.g., C11 threads that barely anyone uses). The extra portability provided by C11 is offset by extra shenanigans of C11 library implementations.

jbeich:
> It does fix `-std=gnu99` for example, but `-std=c99` is too strong and does not define posix-only symbols, such as `clock_gettime()`.

Addressed via file-local `_POSIX_C_SOURCE`. BSD libc doesn't hide POSIX declarations when using `-std=c99`, so it's a bit of a blind fix.

programs/timefn.c
```diff
@@ -8,6 +8,7 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
+#define _POSIX_C_SOURCE	200809L /* clock_gettime */
```
Cyan4973, line 11: I believe this should rather be defined in `timefn.h`, just before the `#include <time.h>`,
otherwise, `UTIL_time_t` declaration in `timefn.h` will be out of sync with `timefn.c`.

_edit_ : although, note the following compilation issue : 
```
../programs/timefn.h:22:0: error: "_POSIX_C_SOURCE" redefined [-Werror]
```

Cyan4973:
With #3163 out of the way, let's now focus on this PR.
It extends the nb of systems that can benefit from a correct timer, so it goes in the right direction.
There are a couple of minor details I would differently,
so instead of nagging you about that,
I'll just merge this PR into a feature branch, and finish off the details there.

PR #3168, opened 2022-06-20
Restore ability to use clock_gettime() on posix systems

Cyan4973:
When `timespec_get()` was unavailable,
`timefn` unit would fall back to `clock_t`,
which is unfortunately not good enough to measure speed in multi-threading scenarios.

@jbeich proposes a patch which re-enable `clock_gettime()` as an alternative for posix systems.
(I can't remember why it was removed...)
This extends the nb of platforms that can use a timer function compatible with multi-threading scenarios.

### DIFF
--- a/programs/timefn.c
+++ b/programs/timefn.c
@@ -8,6 +8,7 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
+#define _POSIX_C_SOURCE	200809L /* clock_gettime */
 
 /* ===  Dependencies  === */
 
@@ -82,10 +83,8 @@ PTime UTIL_getSpanTimeNano(UTIL_time_t clockStart, UTIL_time_t clockEnd)
 }
 
 
-/* C11 requires timespec_get, but FreeBSD 11 lacks it, while still claiming C11 compliance.
-   Android also lacks it but does define TIME_UTC. */
 #elif (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* C11 */) \
-    && defined(TIME_UTC) && !defined(__ANDROID__)
+    || defined(CLOCK_MONOTONIC) /* POSIX.1-2001 (optional) */
 
 #include <stdlib.h>   /* abort */
 #include <stdio.h>    /* perror */
@@ -95,10 +94,17 @@ UTIL_time_t UTIL_getTime(void)
     /* time must be initialized, othersize it may fail msan test.
      * No good reason, likely a limitation of timespec_get() for some target */
     UTIL_time_t time = UTIL_TIME_INITIALIZER;
+#if defined(CLOCK_MONOTONIC)
+    if (clock_gettime(CLOCK_MONOTONIC, &time) != 0) {
+        perror("timefn::clock_gettime(CLOCK_MONOTONIC)");
+        abort();
+    }
+#else
     if (timespec_get(&time, TIME_UTC) != TIME_UTC) {
         perror("timefn::timespec_get");
         abort();
     }
+#endif
     return time;
 }
 
--- a/programs/timefn.c
+++ b/programs/timefn.c
@@ -81,7 +81,7 @@ PTime UTIL_getSpanTimeNano(UTIL_time_t clockStart, UTIL_time_t clockEnd)
 }
 
 
-#elif TIMEFN_HAS_TIMESPEC_GET \
+#elif defined(TIMEFN_HAS_TIMESPEC_GET) \
     || defined(CLOCK_MONOTONIC) /* POSIX.1-2001 (optional) */
 
 #include <stdlib.h>   /* abort */
@@ -92,7 +92,7 @@ UTIL_time_t UTIL_getTime(void)
     /* time must be initialized, othersize it may fail msan test.
      * No good reason, likely a limitation of timespec_get() for some target */
     UTIL_time_t time = UTIL_TIME_INITIALIZER;
-#if TIMEFN_HAS_TIMESPEC_GET
+#if defined(TIMEFN_HAS_TIMESPEC_GET)
     /* favor timespec_get() as first choice for c11 targets */
     if (timespec_get(&time, TIME_UTC) != TIME_UTC) {
         perror("timefn::timespec_get");
--- a/programs/timefn.c
+++ b/programs/timefn.c
@@ -8,8 +8,6 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
-#define _POSIX_C_SOURCE	200809L /* clock_gettime */
-
 /* ===  Dependencies  === */
 
 #include "timefn.h"
@@ -83,7 +81,7 @@ PTime UTIL_getSpanTimeNano(UTIL_time_t clockStart, UTIL_time_t clockEnd)
 }
 
 
-#elif (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* C11 */) \
+#elif TIMEFN_HAS_TIMESPEC_GET \
     || defined(CLOCK_MONOTONIC) /* POSIX.1-2001 (optional) */
 
 #include <stdlib.h>   /* abort */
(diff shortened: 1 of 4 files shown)


Cyan4973:
arf ...

> If the defined operator appears as a result of a macro expansion, the C standard says the behavior is undefined

[source](https://gcc.gnu.org/onlinedocs/gcc-8.4.0/cpp/Defined.html)

Cyan4973:
Still this issue : 

```
../programs/timefn.h:22: error: "_POSIX_C_SOURCE" redefined [-Werror]
   22 | #define _POSIX_C_SOURCE 200809L /* clock_gettime */
```

Not sure if it's really possible to force a specific posix level.
This happens in the `c99` test, which might force a specific level of its own.

Cyan4973:
Interestingly, if I comment out `_POSIX_C_SOURCE`,
`-std=c99` compilation mode works fine,
but as a consequence, I see a mixture of `clock_t` and `CLOCK_MONOTONIC`,
depending on which unit includes `timefn.h`.

```
In file included from benchfn.c:20:                                                                                                                                                                                                                                                                             │·
timefn.h:80:6: warning: #warning "uses C90 clock_t" [-Wcpp]                                                                                                                                                                                                                                                     │·
   80 |     #warning "uses C90 clock_t"                                                                                                                                                                                                                                                                         │·
      |      ^~~~~~~                                                                                                                                                                                                                                                                                            │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/dibio.o                                                                                                                                                                                                                                                            │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/fileio.o                                                                                                                                                                                                                                                           │·
In file included from benchzstd.c:31:                                                                                                                                                                                                                                                                           │·
timefn.h:73:6: warning: #warning "uses CLOCK_MONOTONIC" [-Wcpp]                                                                                                                                                                                                                                                 │·
   73 |     #warning "uses CLOCK_MONOTONIC"                                                                                                                                                                                                                                                                     │·
      |      ^~~~~~~                                                                                                                                                                                                                                                                                            │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/fileio_asyncio.o                                                                                                                                                                                                                                                   │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/timefn.o                                                                                                                                                                                                                                                           │·
In file included from dibio.c:31:                                                                                                                                                                                                                                                                               │·
timefn.h:73:6: warning: #warning "uses CLOCK_MONOTONIC" [-Wcpp]                                                                                                                                                                                                                                                 │·
   73 |     #warning "uses CLOCK_MONOTONIC"                                                                                                                                                                                                                                                                     │·
      |      ^~~~~~~                                                                                                                                                                                                                                                                                            │·
In file included from fileio.c:36:                                                                                                                                                                                                                                                                              │·
timefn.h:73:6: warning: #warning "uses CLOCK_MONOTONIC" [-Wcpp]                                                                                                                                                                                                                                                 │·
   73 |     #warning "uses CLOCK_MONOTONIC"                                                                                                                                                                                                                                                                     │·
      |      ^~~~~~~                                                                                                                                                                                                                                                                                            │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/util.o                                                                                                                                                                                                                                                             │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/zstdcli.o                                                                                                                                                                                                                                                          │·
In file included from timefn.c:13:                                                                                                                                                                                                                                                                              │·
timefn.h:80:6: warning: #warning "uses C90 clock_t" [-Wcpp]                                                                                                                                                                                                                                                     │·
   80 |     #warning "uses C90 clock_t"                                                                                                                                                                                                                                                                         │·
      |      ^~~~~~~                                                                                                                                                                                                                                                                                            │·
CC obj/conf_63ad40fbee40d6d23986334cfd96df82/zstdcli_trace.o                                                                                                                                                                                                                                                    │·
In file included from fileio_common.h:21,                                                                                                                                                                                                                                                                       │·
                 from fileio_asyncio.c:23:                                                                                                                                                                                                                                                                      │·
timefn.h:73:6: warning: #warning "uses CLOCK_MONOTONIC" [-Wcpp]                                                                                                                                                                                                                                                 │·
   73 |     #warning "uses CLOCK_MONOTONIC"                                                                                                                                                                                                                                                                     │·
      |      ^~~~~~~                                                                                                                                                                                                                                                                                            │·
In file included from zstdcli_trace.c:16:                                                                                                                                                                                                                                                                       │·
timefn.h:80:6: warning: #warning "uses C90 clock_t" [-Wcpp]                                                                                                                                                                                                                                                     │·
   80 |     #warning "uses C90 clock_t"                                                                                                                                                                                                                                                                         │·
      |      ^~~~~~~
```

This can't be good.
The unit is necessarily compiled once, with one mode, or the other, not both.
I also noticed wide measurement variations, nowhere near as stable as `timespec_get()` nor `clock_t` only.
This might be a reason.

Cyan4973:
This is actually a very problematic issue.
Having `_POSIX_C_SOURCE` defined differently depending on #include pattern, resulting in different interpretation of `timefn.h` interface, is a big source of trouble.

I think I fixed it, ensuring that `_POSIX_C_SOURCE` is always set through `platform.h`, which is itself always included first, before any standard library,

but that's fragile. In the future, any contributor may change or update the #include order in one of the affected files, this will probably be silent and seem harmless, but would then result in a sibtle change of interface interpretation, resulting in different `typedef` definition, resulting in wrong objects, at best perceived through wrong time measurements...
Worse, there is no automated test to detect such discrepancy.

programs/timefn.c
```diff
@@ -95,10 +92,19 @@ UTIL_time_t UTIL_getTime(void)
     /* time must be initialized, othersize it may fail msan test.
      * No good reason, likely a limitation of timespec_get() for some target */
     UTIL_time_t time = UTIL_TIME_INITIALIZER;
+#if defined(TIMEFN_HAS_TIMESPEC_GET)
+    /* favor timespec_get() as first choice for c11 targets */
```
embg, line 96: For testing: could we add something like `asm("\n# Using timespec_get")` and `asm("\n# Using clock_gettime")` to each branch of the ifdef? Then we could grep the generated assembly for those strings and ensure only one string is included.

(I'm not sure exactly how to insert a comment or symbol into the assembly, took the above code from https://stackoverflow.com/a/28493396/3154996. I think it is definitely possible, though).

Cyan4973: That would add some code specifically for this test.
It would have to be enabled through some dedicated setting (like a macro) since we don't want that for the general case.

Then indeed it would be possible to couple that generation with an external tool that grep the outcome and ensure it's consistent. This `asm("# message")` thing seems to only leave a comment in the generated assembly. So that means accessing the generated assembly and grep it.

That's some work, but it definitely looks plausible.

embg: It seems like we can get text into the actual binary like this: `asm volatile(".global message \n message:")` ([godbolt](https://godbolt.org/z/4crhcYYK9)). I can test if this works for our situation and if so put up a PR which does it (gated to `DEBUGLEVEL >= 2` or something). Agree that getting the generated assembly would require some work, hopefully that's not necessary.

Edit: it might be a problem for linking if it is inserted in multiple places, if so I can add a macro specifically for this test.

Cyan4973: Yes, this would be a welcome test capability.
As it stands, this PR is stuck as long as there is no way to reliably test `timefn` interface and ensure it's interpreted the same way in all units that include it.

Cyan4973:
Another possibility  : 
change the `timefn` API in order to use opaque types,
so that the type interpretation is done solely within `timefn.c`, 
and can't be different on user side depending on their `#include` order.

This is obviously more work, and requires refactoring `timefn` API as well as all existing calling sites.
Once completed though, it will eliminate this fragility on interpreting the return type the same way from all call sites.

Cyan4973:
Replaced by #3423

PR #3170, opened 2022-06-20
removed gnu99 statement from meson recipe

Cyan4973:
this shouldn't be necessary.
Answers #3163 .

### DIFF
--- a/build/meson/meson.build
+++ b/build/meson/meson.build
@@ -12,7 +12,10 @@ project('zstd',
   ['c', 'cpp'],
   license: ['BSD', 'GPLv2'],
   default_options : [
-    'c_std=gnu99',
+    # There shouldn't be any need to force a C standard convention for zstd
+    # but in case one would want that anyway, this can be done here.
+    # 'c_std=gnu99',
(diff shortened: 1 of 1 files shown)


terrelln (approved):

build/meson/meson.build
```diff
@@ -12,8 +12,11 @@ project('zstd',
   ['c', 'cpp'],
   license: ['BSD', 'GPLv2'],
   default_options : [
-    'c_std=gnu99',
-    'cpp_std=c++11',
+    # There shouldn't be any need to force a C or C++ standard convention
+    # but in case one would want that anyway,
+    # this can be done here.
+    # 'c_std=gnu99',
+    # 'cpp_std=c++11',
```
terrelln: Actually we do need to keep this line, since pzstd requires c++ >= 11.

terrelln (changes requested):

build/meson/meson.build
Cyan4973: changed

terrelln (approved):

PR #3173, opened 2022-06-21
Update Copyright Comments

felixhandte:
T106302328 wants us to update to "`Meta Platforms, Inc. and affiliates`".

In addition, I cleaned up some lagging year ranges, rewriting their ends to "`present`". If there's a reason it should be an explicit year though, I can switch to "`2022`".

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -2,7 +2,7 @@ BSD License
 
 For Zstandard software
 
-Copyright (c) 2016-present, Meta Platforms, Inc. and affiliates. All rights reserved.
+Copyright (c) 2016-2022, Meta Platforms, Inc. and affiliates. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
 are permitted provided that the following conditions are met:
--- a/LICENSE
+++ b/LICENSE
@@ -2,7 +2,7 @@ BSD License
 
 For Zstandard software
 
-Copyright (c) 2016-present, Facebook, Inc. All rights reserved.
+Copyright (c) 2016-present, Meta Platforms, Inc. and affiliates. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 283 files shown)


Cyan4973:
It seems we were told at some point that `present` wasn't a valid year, and that an actual date was necessary.

terrelln:
> It seems we were told at some point that present wasn't a valid year, and that an actual date was necessary.

That is not the latest guidance. We should move everything to a yearless copyright.

terrelln:
Just `Copyright (c) Meta Platforms, Inc. and affiliates.` is fine. https://www.internalfb.com/intern/wiki/Open_Source/Licenses/

JunHe77:
Sorry for just jumping in for a related copyright question. 😄 
Is a new file, like the new .S file in #3155 required to share copyright with Meta? I saw the check in [test-licens.py](https://github.com/facebook/zstd/blob/dev/tests/test-license.py#L117-L121), but [exceptions](https://github.com/facebook/zstd/blob/dev/tests/test-license.py#L60-L67) also exist.

Tachi107:
> In addition, I cleaned up some lagging year ranges, rewriting their ends to "`present`". If there's a reason it should be an explicit year though, I can switch to "`2022`".

I don't want to mess with corporation policies, but I'd recommend giving [this blog post](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change) a quick read - in short routinely bumping copyright years isn't really that worth it :)

Cyan4973:
There is no more reference of any year in the copyright header.
Is that a wanted outcome ?

felixhandte:
@Cyan4973, that seems to be [the guidance](https://www.internalfb.com/intern/wiki/Open_Source/Licenses/).

Cyan4973 (approved):

PR #3174, opened 2022-06-21
Update README.md for fuzzers

embg:
Took me a minute to figure this out, so thought I'd add it to the docs.

### DIFF
--- a/tests/fuzz/README.md
+++ b/tests/fuzz/README.md
@@ -57,7 +57,7 @@ for target in $(./fuzz.py list); do
     ./fuzz.py libfuzzer $target -jobs=10 -workers=10 -max_total_time=1000 || break;
 done
 ```
-Alternatively, you can fuzz all targets in parallel:
+Alternatively, you can fuzz all targets in parallel, using one core per target:
 ```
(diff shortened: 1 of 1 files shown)


tests/fuzz/README.md
```diff
@@ -50,6 +50,16 @@ See the help of the relevant command for options.
 Flags not parsed by `fuzz.py` are passed to the fuzzing engine.
 The command used to run the fuzzer is printed for debugging.
 
+Here's a helpful command to fuzz all targets in parallel,
+stopping only if a bug is found:
+```
+for target in $(./fuzz.py list); do
+    ./fuzz.py libfuzzer $target -jobs=10 -workers=10 -max_total_time=1000 || break;
```
terrelln, line 57: This doesn't run all the fuzzers in parallel. It runs one fuzzer at a time.

I'd suggest something like:

```
./fuzz.py list | xargs -P$(wc -l ./fuzz.py list) -I__ sh -c "./fuzz.py libfuzzer __ 2>&1 | tee __.log"
```

But double check that the command works first :)

embg: Yes that's what I meant, sorry. I'll test/add your command and correct the docs on my command. There's tradeoffs in both directions: running one fuzzer on all cores allows them to share interesting inputs, so you get super-linear scaling (at least according to the docs, idk how big the impact is in practice). On the other hand running on all cores reduces the worst case time-to-bug since you make progress on all jobs in parallel.

terrelln (approved):

PR #3177, opened 2022-06-23
Add prefetchCDictTables CCtxParam (+10-20% cold dict compression speed)

embg:
# Summary
* Adds a new CCtxParam (prefetchCDictTables)
* Exposes it in the largeNbDicts benchmark
* Adds it to fuzzers

# Description of the optimization
In some situations, zstd uses CDict tables in-place rather than copying them into the working context. (See docs on ZSTD_dictAttachPref_e for details). In such situations, compression speed is seriously impacted when CDict tables are "cold" (outside CPU cache).

This PR adds a CCtxParam (prefetchCDictTables) which instructs zstd to prefetch CDict tables when they are used in-place (specifically in level 1-4 dictMatchState matchfinders). For sufficiently small inputs, the cost of the prefetch will outweigh the benefit. For sufficiently large inputs, zstd will by default memcpy() CDict tables into the working context, so there is no need to prefetch. This parameter is targeted at a middle range of input sizes, where a prefetch is cheap enough to be useful but memcpy() is too expensive.

The exact range of input sizes where this makes sense is best determined by careful experimentation (see below for measurements on one particular machine / dataset which demonstrate 10-20% wins for a particular working set size and input size). Rather than enabling this param for all inputs, the code which calls `ZSTD_compress2()` should use a size cutoff (tuned via experimentation) to select the best prefetch strategy for each input.

# Measurements
I measured the effect of this param on the HTML dataset. I benchmarked on a Intel(R) Xeon(R) D-2191A CPU @ 1.60GHz machine with core isolation and turbo disabled. 

newplot (18)

We can see that the param is harmful for level3 *even in the cold CDict scenario* if the inputs are small enough (0-8K). For larger inputs (8-16K) at the same level, we see up to 20% wins. This demonstrates the need for selective application of this param.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -576,6 +576,11 @@ ZSTD_bounds ZSTD_cParam_getBounds(ZSTD_cParameter param)
         bounds.upperBound = 1;
         return bounds;
 
+    case ZSTD_c_prefetchCDictTables:
+        bounds.lowerBound = (int)ZSTD_ps_auto;
+        bounds.upperBound = (int)ZSTD_ps_disable;
+        return bounds;
+
     default:
         bounds.error = ERROR(parameter_unsupported);
         return bounds;
@@ -640,6 +645,7 @@ static int ZSTD_isUpdateAuthorized(ZSTD_cParameter param)
     case ZSTD_c_useBlockSplitter:
     case ZSTD_c_useRowMatchFinder:
     case ZSTD_c_deterministicRefPrefix:
+    case ZSTD_c_prefetchCDictTables:
     default:
         return 0;
     }
@@ -695,6 +701,7 @@ size_t ZSTD_CCtx_setParameter(ZSTD_CCtx* cctx, ZSTD_cParameter param, int value)
     case ZSTD_c_useBlockSplitter:
     case ZSTD_c_useRowMatchFinder:
     case ZSTD_c_deterministicRefPrefix:
+    case ZSTD_c_prefetchCDictTables:
         break;
 
     default: RETURN_ERROR(parameter_unsupported, "unknown parameter");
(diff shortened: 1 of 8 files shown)


lib/compress/zstd_double_fast.c
```diff
@@ -345,6 +345,13 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
     /* if a dictionary is attached, it must be within window range */
     assert(ms->window.dictLimit + (1U << cParams->windowLog) >= endIndex);
 
+    if (ms->prefetchCDictTables) {
+        size_t const hashTableSize = ((size_t)1) << dictCParams->hashLog;
+        size_t const chainTableSize = ((size_t)1) << dictCParams->chainLog;
```
felixhandte: It seems clearer to me if you do the `* sizeof(U32)` here rather than in the `PREFETCH_AREA()` call.

lib/compress/zstd_fast.c
```diff
@@ -500,6 +500,11 @@ size_t ZSTD_compressBlock_fast_dictMatchState_generic(
      * when translating a dict index into a local index */
     assert(prefixStartIndex >= (U32)(dictEnd - dictBase));
 
+    if (ms->prefetchCDictTables) {
+        size_t const hashTableSize = ((size_t)1) << dictCParams->hashLog;
```
felixhandte: Same.

lib/compress/zstd_compress.c
```diff
@@ -2913,6 +2928,7 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
                                                                                     zc->appliedParams.useRowMatchFinder,
                                                                                     dictMode);
             ms->ldmSeqStore = NULL;
+            ms->prefetchCDictTables = zc->appliedParams.prefetchCDictTables == ZSTD_ps_enable;
```
felixhandte: I don't think this is really the right place for this. Shouldn't it go in like `ZSTD_resetCCtx_internal()`?

felixhandte (commented):
LGTM. Just a couple nits.

lib/compress/zstd_compress_internal.h
```diff
@@ -331,6 +332,9 @@ struct ZSTD_CCtx_params_s {
 
     /* Internal use, for createCCtxParams() and freeCCtxParams() only */
     ZSTD_customMem customMem;
+
+    /* Controls prefetching in some dictMatchState matchfinders */
+    ZSTD_paramSwitch_e prefetchCDictTables;
```
felixhandte, line 341: One more thing: add to this comment that the behavior is controlled from the cctx ms, and that this parameter has no effect in the cdict ms.

felixhandte (approved):
Assuming tests pass, ship it!

PR #3179, opened 2022-06-29
Prepare v1.5.3

embg:
Updated CHANGELOG, bumped version, ran `make manual`.


felixhandte:
Ah but looks like this doesn't update the [man page](https://github.com/facebook/zstd/blob/dev/programs/zstd.1). I guess you do that with `make -C programs zstd.1`? (But you likely don't have the required tools installed. Let me know if you want me to do that part.)

embg:
> Ah but looks like this doesn't update the [man page](https://github.com/facebook/zstd/blob/dev/programs/zstd.1). I guess you do that with `make -C programs zstd.1`? (But you likely don't have the required tools installed. Let me know if you want me to do that part.)

Luckily was able to run my machine :) pushed a new commit with that change.

felixhandte (approved):
LGTM!

Assuming CI passes, ship it!

embg:
A couple jobs are down due to unrelated infra failures (will take a look next week if it doesn't resolve itself over the long weekend). Going to go ahead and land since this is a safe change and everything else passed.

gvollant:
Hello,
What is the status of 1.5.3 ? Latest current tag seem 1.5.2

regards
Gilles Vollant

felixhandte:
Hi @gvollant,

v1.5.3 was an internal release we cut for use at Meta, in order to take advantage of some recent optimizations. We do not intend to ship it publicly, since we did not go through the extended validation work involved in a public release, and it was in other ways not fully polished.

However, we do plan to prepare a v1.5.4 fairly soon, which will be fully baked and of course include all of the changes in v1.5.3.

PR #3180, opened 2022-06-30
Fix ZSTD_BUILD_TESTS=ON with MSVC

nocnokneo:
Fixes the following compilation error when building with MSVC with tests enabled:

    Command line error D8021 : invalid numeric argument '/Wno-deprecated-declarations'


Cyan4973 (approved):
I would have expected MSVC to need something equivalent to avoid compilation warnings when compiling deprecated symbols, but maybe it doesn't matter (or maybe we removed deprecated symbols from tests too).
Anyway, removing this GCC flag from MSVC compilation is surely a good idea.

embg:
Some MSVC tests are currently disabled in CI, but will be enabled before open source release, so I think it's OK to merge now.

PR #3184, opened 2022-07-06
zstd -lv <file> to show dictID

htnhan:


programs/fileio.c
```diff
@@ -2887,6 +2889,7 @@ displayInfo(const char* inFileName, const fileInfo_t* info, int displayLevel)
         DISPLAYOUT("# Zstandard Frames: %d\n", info->numActualFrames);
         if (info->numSkippableFrames)
             DISPLAYOUT("# Skippable Frames: %d\n", info->numSkippableFrames);
+        DISPLAYOUT("DictID: %u\n", info->dictID);
```
felixhandte, line 2897: Ok, so. To be *very* pedantic, a file can be composed of multiple frames, as counted just above. You could imagine that those different frames could have different `DictID`s written into their headers. That's such an impossibly rare case that I don't think it's worth supporting. But it's possibly a case that's worth detecting and erroring on? Because otherwise this display here will be misleading, by printing only the last dict id.

programs/fileio.c
```diff
@@ -2775,6 +2776,7 @@ FIO_analyzeFrames(fileInfo_t* info, FILE* const srcFile)
                 }
                 ERROR_IF(ZSTD_getFrameHeader(&header, headerBuffer, numBytesRead) != 0,
                         info_frame_error, "Error: could not decode frame header");
+                info->dictID = header.dictID;
```
felixhandte: To my comment below, maybe just print a warning here and then clear `info->dictID` if `info->dictID && header.dictID && info->dictID != header.dictID`?

felixhandte (changes requested):
Looks good! I have one concern that it would be great if you could address.

felixhandte (approved):
Great! Thanks for your contribution!

PR #3187, opened 2022-07-08
T119975957

### LINKED ISSUE #3128: Consider Removing `ZSTD_copy*Ctx()` Functions
They are effectively useless and their presence is misleading.

At the minimum we should mark them deprecated with an explanation of why they don't do what you might expect.

mileshu:
[T124890272] Consider Removing ZSTD_copy*Ctx() Functions #3128

The discussion for this task is here: https://github.com/facebook/zstd/issues/3128.

They are effectively useless and their presence is misleading.

At the minimum we should mark them deprecated with an explanation of why they don't do what you might expect.

Signed-off-by: Miles HU <yuanpu@fb.com>


yoniko:
Hi @wayayayay,
Thank you for contributing.
There are some adjustments that will be required for this PR to be merge-able:

1. Please give the PR a proper title and description. Please also reference the issue (https://github.com/facebook/zstd/issues/3128) in the description so it will be clear the two are related.
2. Follow the process described in https://github.com/facebook/zstd/blob/dev/CONTRIBUTING.md. 
Make sure your tests pass locally and on your CI.
3. Make sure to be thorough, for example you've remove `ZSTD_copyCCtx`'s body, but there are still references to this function including a declaration in the header file `zstd.h` (line 2440).

@felixhandte - are you sure removing the functions is the correct thing to do and not just marking them as deprecated? in any case I feel some context is missing in the original issue that might help to clarify the needed actions.

Cyan4973:
> 
> @felixhandte - are you sure removing the functions is the correct thing to do and not just marking them as deprecated? in any case I feel some context is missing in the original issue that might help to clarify the needed actions.

That's a good point. If we want to remain as cautious as we are used to with regards to `zstd`'s public API,
these symbols should rather be marked deprecated first. Removing them then shall wait for a more important version number increase, such as `1.6.0`.

mileshu:
Will add deprecated tag for now, remove them in next release.

PR #3197, opened 2022-07-13
Clarify benchmark chunking docstring

embg:
I spoke with a user today who thought `-B#` controls the zstd block size in benchmarks, rather than controlling the chunk size (to split a large compression into many independent small compressions). This leads to severe misinterpretation of results, because the latter interpretation affects compression ratio very differently than the former. I think the proposed doc update can prevent this misunderstanding going forward.

### DIFF
--- a/programs/README.md
+++ b/programs/README.md
@@ -209,7 +209,7 @@ Benchmark arguments :
  -b#    : benchmark file(s), using # compression level (default: 3)
  -e#    : test all compression levels successively from -b# to -e# (default: 1)
  -i#    : minimum evaluation time in seconds (default: 3s)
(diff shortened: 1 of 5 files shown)


Cyan4973:
Green light for the renaming.

The only issue is that this PR seems to completely erase the content of `zstd.1`.

embg:
> Green light for the renaming.
> 
> The only issue is that this PR seems to completely erase the content of `zstd.1`.

It seems like I have the wrong version of `ronn` set up on my laptop. Fixed the issue, should be good to merge now.

terrelln (approved):

PR #3205, opened 2022-07-19
[largeNbDicts] Fix decompression segfault; Add additional benchmark metrics

zhuhan0:
Benchmarking decompression results in a segfault in `createCompressInstructions`
because `cctxParams` is NULL. 

```
$ ./largeNbDicts -1 --nbDicts=2 -d -D ~/benchmarks/html/html_8_16K.32K.dict ~/benchmarks/html/html_8_16K/*
loading 7450 files...
created src buffer of size 83.4 MB
split input into 7450 blocks
loading dictionary /home/zhuhan/benchmarks/html/html_8_16K.32K.dict
compressing at level 1 without dictionary : Ratio=3.03  (28827863 bytes)
compressed using a 32768 bytes dictionary : Ratio=4.28  (20410262 bytes)
generating 2 dictionaries, using 0.1 MB of memory
[1]    2425316 segmentation fault (core dumped)  ./largeNbDicts -1 --nbDicts=2 -d -D ~/benchmarks/html/html_8_16K.32K.dict
```

Skip running that function if we are not benching
compression.

### DIFF
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -19,7 +19,7 @@
 /*---  Dependencies  ---*/
 
 #include <stddef.h>   /* size_t */
-#include <stdlib.h>   /* malloc, free, abort */
+#include <stdlib.h>   /* malloc, free, abort, qsort*/
 #include <stdio.h>    /* fprintf */
 #include <limits.h>   /* UINT_MAX */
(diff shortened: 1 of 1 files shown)


zhuhan0:
Added two more commits:
1. Print more metrics into the csv file.
2. Add an option to print out median speed.

embg (approved):
Looks good, thanks for adding these features!

PR #3208, opened 2022-07-20
Add warning when multi-thread decompression is requested

tomcwang:
address https://github.com/facebook/zstd/issues/2918

When user pass in argument for both decompression and multi-thread, print a warning message
to indicate that multi-threaded decompression is not supported.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -1236,6 +1236,9 @@ int main(int argCount, const char* argv[])
             DISPLAYLEVEL(3, "Note: %d physical core(s) detected \n", nbWorkers);
         }
     }
+    if ((operation==zom_decompress) && (!singleThread) && (nbWorkers > 1)) {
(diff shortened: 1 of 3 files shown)


yoniko:
Thank you @tomcwang .
I see the check is inserted after `-T0` is handled.
That means that a warning will be printed on decompression with `-T0` on a multicore system.
Can you move the check to before the `-T0` case is handled and `nbWorkers` are assigned?

tomcwang:
> Thank you @tomcwang . I see the check is inserted after `-T0` is handled. That means that a warning will be printed on decompression with `-T0` on a multicore system. Can you move the check to before the `-T0` case is handled and `nbWorkers` are assigned?

makes sense, I moved the check to before nbWorker is assigned, warning is not thrown for -T0 anymore.

yoniko:
Thank you.
One last thing - can you please add tests to `cli-tests/compression/multi-threaded.sh`?

yoniko (approved):
Looks good!

PR #3209, opened 2022-07-20
[largeNbDicts] Second try at fixing decompression segfault to always create compressInstructions

zhuhan0:
Summary:
Freeing an uninitialized pointer is undefined behavior. This caused a segfault
when compiling the benchmark with Clang -O3 and benching decompression.

Test Plan:
make and run

### DIFF
--- a/contrib/largeNbDicts/largeNbDicts.c
+++ b/contrib/largeNbDicts/largeNbDicts.c
@@ -582,7 +582,8 @@ compressInstructions createCompressInstructions(cdict_collection_t dictionaries,
     compressInstructions ci;
     ci.cctx = ZSTD_createCCtx();
     CONTROL(ci.cctx != NULL);
-    ZSTD_CCtx_setParametersUsingCCtxParams(ci.cctx, cctxParams);
+    if (cctxParams)
+      ZSTD_CCtx_setParametersUsingCCtxParams(ci.cctx, cctxParams);
     ci.nbDicts = dictionaries.nbCDict;
     ci.dictNb = 0;
     ci.dictionaries = dictionaries;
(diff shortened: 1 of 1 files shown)


contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -697,7 +697,7 @@ static int benchMem(slice_collection_t dstBlocks, slice_collection_t srcBlocks,
             BMK_createTimedFnState(total_time_ms, ms_per_round);
 
     decompressInstructions di = createDecompressInstructions(ddictionaries);
-    compressInstructions ci;
+    compressInstructions ci = {0};
```
Cyan4973: The diagnosis is correct, but this solution could trigger some warnings on some compilers, depending on compilation flags, because `compressInstructions` is a multi-fields `struct`.

A safer way to achieve the same outcome could be to use `memset(&ci, 0, sizeof(ci));`.

A potentially proper but also more complex way could be to wonder why we don't just use the constructor `createCompressInstructions()` all the time, ensuring `ci` is correctly formed. But that would require reasoning about the whole program intent & implementation, hence can be more challenging.

zhuhan0: Thanks for the review. In [this commit](https://github.com/facebook/zstd/commit/d0c88afe6d5c0a29274df735a044154c6bf6deaa) I changed the program to call `createCompressInstructions` only when we're benching compression, because that function calls `ZSTD_CCtx_setParametersUsingCCtxParams` and `cctxParams` is NULL during decompression, resulting in a segfault. I think I can change this to always call `createCompressInstructions` and check if `cctxParams` is NULL in that function instead.

Cyan4973: Sounds good to me.

zhuhan0:
Hi @Cyan4973 does this look good to merge?

Cyan4973:
Yes,
there was one CI test error left, but I believe it's unrelated to this PR.

Cyan4973 (approved):

zhuhan0:
Thanks. Could you help merge this too? I don't have write access.

PR #3214, opened 2022-07-25
Issue #3070

GeoffMStaceyWrk:
Fix for https://github.com/facebook/zstd/issues/3070

Notes:
In cases where the inputs are not well formed in terms of suffix (or non numerical prefix) the program will warn and use a default value. 

As default this can be found very simply by checking for a null value, but there was an edge case on testing where the "zstd=" command doesn't use white space (which means we need to check both /0 and ',' as well). This does mean that inputs such as "--memory=3,000KB" won't work. This is assumed satisfactory. 

There is a potential follow up task here where very small values are allowed since they follow a valid syntax - ie "--memory=15", but if there is a minimum we need, that may be useful to warn the user about, and clamp input in that way. 

Have added more tests to the playTests script, which checks for a warning and that we reset default behavior on poorly formed input, but that we don't warn for valid input. 

Testing:
Have verified that the #3070 issue is indeed fixed, and that the above added tests pass now, but did not in the earlier dev branch.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -346,9 +346,22 @@ static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
         (*stringPtr)++;  /* skip `K` or `M` */
         if (**stringPtr=='i') (*stringPtr)++;
         if (**stringPtr=='B') (*stringPtr)++;
+
     }
-    *value = result;
-    return 0;
+
+    //Only well formed entries will be 0, or , (specifically for zstd compound)
+    if (**stringPtr == 0 || **stringPtr == ',') {
+
+        *value = result;
+        return 0;
+
+    } else {
+
+        *value = 0;
+        return 2;
+
+    }
+
 }
 
 /*! readU32FromChar() :
@@ -357,9 +370,21 @@ static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
  *  Will also modify `*stringPtr`, advancing it to position where it stopped reading.
  *  Note : function will exit() program if digit sequence overflows */
 static unsigned readU32FromChar(const char** stringPtr) {
-    static const char errorMsg[] = "error: numeric value overflows 32-bit unsigned int";
+
     unsigned result;
-    if (readU32FromCharChecked(stringPtr, &result)) { errorOut(errorMsg); }
+    const char* startChar;
+    int returnVal;
+
+    startChar = *stringPtr;
+
+    returnVal = readU32FromCharChecked(stringPtr, &result);
+
+    if (returnVal == 1) {
+        errorOut("error: numeric value overflows 32-bit unsigned int"); 
+    }
+    else if (returnVal == 2) {
+        DISPLAY("warning: poorly formed parameter of %s, resetting to default \n", startChar);
+    }
     return result;
 }
 
@@ -369,14 +394,27 @@ static unsigned readU32FromChar(const char** stringPtr) {
  *  Will also modify `*stringPtr`, advancing it to position where it stopped reading.
  *  Note : function will exit() program if digit sequence overflows */
 static int readIntFromChar(const char** stringPtr) {
-    static const char errorMsg[] = "error: numeric value overflows 32-bit int";
+    
     int sign = 1;
     unsigned result;
+    const char* startChar;
+    int returnVal;
+    
     if (**stringPtr=='-') {
         (*stringPtr)++;
         sign = -1;
     }
-    if (readU32FromCharChecked(stringPtr, &result)) { errorOut(errorMsg); }
+
+    startChar = *stringPtr;
+    
+    returnVal = readU32FromCharChecked(stringPtr, &result);
+    
+    if (returnVal == 1) {
+        errorOut("error: numeric value overflows 32-bit int");
+    }
+    else if (returnVal == 2) {
+        DISPLAY("warning: poorly formed parameter of %s, resetting to default \n", startChar);
+    }
     return (int) result * sign;
 }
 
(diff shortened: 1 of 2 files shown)


programs/zstdcli.c
```diff
@@ -369,14 +394,27 @@ static unsigned readU32FromChar(const char** stringPtr) {
...
     }
-    if (readU32FromCharChecked(stringPtr, &result)) { errorOut(errorMsg); }
+
+    startChar = *stringPtr;
+    
+    returnVal = readU32FromCharChecked(stringPtr, &result);
+    
+    if (returnVal == 1) {
+        errorOut("error: numeric value overflows 32-bit int");
+    }
```
yoniko, line 414: The overflow check here isn't good enough as signed and unsigned integers have different ranges.
While I understand this is the behavior that was implemented before, it would be good to augment the overflow checks with comparisons to `INT_MAX` and `INT_MIN`. 
An easy way to implement these checks would be to convert to i64 and compare the 64 bit value.

programs/zstdcli.c
```diff
@@ -346,9 +346,22 @@ static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
...
+    if (**stringPtr == 0 || **stringPtr == ',') {
+
+        *value = result;
+        return 0;
+
+    } else {
+
+        *value = 0;
+        return 2;
+
```
yoniko, line 362: We can use this opportunity to reduce code duplication in this file.
Instead of implementing the same logic in both `readU32FromCharChecked` and `readSizeTFromCharChecked` we can use the fact that `size_t` will always be at least as big a `U32`.
Then we can have `readU32FromCharChecked` call `readSizeTFromCharChecked` and just check for U32 overflow.

programs/zstdcli.c
```diff
@@ -409,8 +449,20 @@ static int readSizeTFromCharChecked(const char** stringPtr, size_t* value)
...
+    //Only well formed entries will be 0 or , (specifically for zstd compound)
+    if (**stringPtr == 0 || **stringPtr == ',') {
+
+        *value = result;
+        return 0;
+    
+    }else{
+
+        *value = 0;
```
yoniko, line 461: This is an error, in such a case we don't want to change `*value`.

programs/zstdcli.c
```diff
@@ -409,8 +449,20 @@ static int readSizeTFromCharChecked(const char** stringPtr, size_t* value)
...
+        return 2;
```
yoniko, line 462: Please add documentation for this case.

programs/zstdcli.c
```diff
@@ -409,8 +449,20 @@ static int readSizeTFromCharChecked(const char** stringPtr, size_t* value)
...
-    *value = result;
-    return 0;
+
+    //Only well formed entries will be 0 or , (specifically for zstd compound)
+    if (**stringPtr == 0 || **stringPtr == ',') {
+
+        *value = result;
+        return 0;
+    
+    }else{
```
yoniko, line 459: This makes sense but I'm not 100% sure.
@felixhandte  / @terrelln would appreciate your inputs here.

programs/zstdcli.c
```diff
@@ -419,9 +471,20 @@ static int readSizeTFromCharChecked(const char** stringPtr, size_t* value)
...
+
+    startChar = *stringPtr;
+    
+    returnVal = readSizeTFromCharChecked(stringPtr, &result);
+    if (returnVal == 1) { 
+        errorOut("error: numeric value overflows size_t");
+    }
+    else if (returnVal == 2) {
+        DISPLAY("warning: poorly formed parameter of %s, resetting to default \n", startChar);
+    }
```
yoniko, line 487: If we encounter an error we want to `errorOut`.

yoniko (changes requested):
Thank you for the PR. I appreciate the added tests to playtest.
Please see comments, some issues were not introduced by you, but this is a good opportunity to fix those as well.

programs/zstdcli.c
GeoffMStaceyWrk: Hi @yoniko - agreed, that would be a useful addition. Can we raise it as a separate task though please since it is out of scope for this issue.

programs/zstdcli.c
GeoffMStaceyWrk: @yoniko - interesting point - can we assume that there aren't any 16 bit processors knocking around which will use this software (I'd assume something _very_ esoteric!). I did mull merging these - what are the 2 use cases that led to there being 2 functions? Both are unsigned integers - but one is (possibly) clamped to smaller than word size?

programs/zstdcli.c
GeoffMStaceyWrk: Understood - so error out similar to overflow?

programs/zstdcli.c
GeoffMStaceyWrk: No worries, can do

programs/zstdcli.c
GeoffMStaceyWrk: @yoniko - not a problem, can I quickly clarify what bit you mean?

yoniko: Function has documentation for return values, please add documentation for this new value.

programs/zstdcli.c
yoniko: Yes, but this is the internal function that doesn't error out. Even so, lets keep behavior consistent in the case of errors and not assign anything to `*value` on error.

programs/zstdcli.c
yoniko: Yes, one supports up to 32bit while the other supports up to size_t (32 or 64 bit).

Looks like the size_t function was added at https://github.com/facebook/zstd/commit/834d06b6 and was probably due to someone wanting to assign sizes that are larger than 4GB. I don't really like this previous patch and I think it creates some inconsistencies in the cli, but we shouldn't address those now.

You can assume size_t is at least 32 bit (I believe the codebase wouldn't work if size_t is 16 bit), but you can add a static_assert just in case.

programs/zstdcli.c
yoniko: Yes, this is not a must here.

PR #3215, opened 2022-07-27
Fix small file passthrough

### LINKED ISSUE #3210: Passthrough on short files fails
**Describe the bug**
When trying to use the decompression passed through functionality on files of 0, 1, or 2 bytes the decompressor fails to read enough bytes to understand the context and quits instead of just copying the bytes to the output. This means if you have small files and are trying to use `zstdcat` for all of them, the very small files will fail and not be written to the output.

**To Reproduce**
```fish
echo | zstd -df
echo 1 | zstd -df
echo 12 | zstd -df
echo 123 | zstd -df
```
```fish
echo | zstdcat
echo 1 | zstdcat
echo 12 | zstdcat
echo 123 | zstdcat
```
output
```
zstd: /*stdin*\: unknown header 
zstd: /*stdin*\: unknown header 
zstd: /*stdin*\: unknown header 
123
```

**Expected behavior**
Zstd and gzip support decompression passthrough where the bytes will simply be copied to the output as-is when the following conditions are met:
1. if a file is present that is not in a known format
2. the `-f` flag is present
3. the output will be written to stdout

example
```fish
❯ echo hello there | zstd -df
hello there
```

gzip on the failure cases produces the correct output
```fish
echo | gzip -df
echo 1 | gzip -df
echo 12 | gzip -df
echo 123 | gzip -df
```
```

1
12
123
```

**Screenshots and charts**
Linux x86 - self installed from repo

cgbur:
Fixes passthrough on files with fewer than 4 magic bytes by checking if passthrough is enabled before exiting with an error. If force is enabled and destination is `stdout`, the bytes are copied as-is to the output.

Changes:
- add a check for passthrough if we read fewer than 4 bytes (`zstd` magic length)
- add a test to ensure passthrough works on short files

fixes: #3210
review @terrelln

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -2346,7 +2346,10 @@ static int FIO_decompressFrames(FIO_ctx_t* const fCtx,
             break;   /* no more input */
         }
         readSomething = 1;   /* there is at least 1 byte in srcFile */
-        if (ress.readCtx->srcBufferLoaded < toRead) {
+        if (ress.readCtx->srcBufferLoaded < toRead) { /* not enough input to check magic number */
+            if ((prefs->overwrite) && !strcmp (dstFileName, stdoutmark)) {  /* pass-through mode */
+                return FIO_passThrough(&ress);
+            }
             DISPLAYLEVEL(1, "zstd: %s: unknown header \n", srcFileName);
             return 1;
         }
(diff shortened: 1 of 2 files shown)


terrelln (approved):
Awesome! Thanks for the PR!

PR #3219, opened 2022-07-29
[AIX] Fix Compiler Flags and Bugs on AIX to Pass All Tests

qiongsiwu:
This PR fixes: 

- Code that triggers compiler warnings (unused argument and function prototype warnings)
- The `-s` compiler flag that is no longer supported by newer versions of `clang`. It is replaced by `-Wl,-s`. 
- `malloc` of size zero returns 0 on AIX, causing the `dict-builder/no-inputs.sh` test case to fail.


terrelln (approved):
Thanks for the PR @qiongsiwu! Looks good to me!

PR #3226, opened 2022-07-29
fix issue #3144

zengyijing:
When the value of argument long is not set, the current argument parsing logic doesn't set windowLog to the default value g_defaultMaxWindowLog and windowLog remains 0. As a result, when calling printActualCParams(), it prints out a wrong value got from ZSTD_getCParams().


terrelln:
I think this fix looks good, but please add a test to the `tests/cli-tests/verbose-wlog.sh`.

You should add a shell script to `tests/cli-tests/compression` that compresses a file and prints the extra verbose information. You can just write the output to `-o $INTOVOID`. E.g.

```
#!/bin/sh

set -e

. "$COMMON/platform.sh"

zstd < file -vv -19 -o file.19.zst
zstd -vv -l file.19.zst

zstd < file -vv -19 --long -o file.19.long.zst
zstd -vv -l file.19.long.zst
```

If you run `./tests/cli-tests/run.py --preserve`, it will output the stderr to `tests/cli-tests/scratch/compression/verbose-wlog/stderr`. You can ask the tool to expect a particular stderr with a glob by adding a file `tests/cli-tests/compression/verbose-wlog.sh.stderr.glob`

```
...
*wlog=23*
...
Window Size: 8388608 B*
...
*wlog=27*
Window Size: 134217728 B*
...
```

See the [README](https://github.com/facebook/zstd/blob/dev/tests/cli-tests/README.md) for docs on the cli-tests.

terrelln (changes requested):
See above

zengyijing:
@terrelln I've added the test as you suggested. Please take a look. Thanks!

terrelln (approved):
Thanks for the PR!

PR #3230, opened 2022-08-01
Add description for ZSTD_decompressStream and ZSTD_initDStream

### LINKED ISSUE #3229: Docs lack ZSTD_initDStream and ZSTD_decompressStream signature in the streaming decompression section
**Describe the bug**

[Docs for 1.5.2](https://raw.githack.com/facebook/zstd/release/doc/zstd_manual.html#Chapter9) have only `ZSTD_DStreamInSize`/`ZSTD_DStreamOutSize` in the Streaming decompression functions section.

**To Reproduce**

1. See the link above.

**Expected behavior**

Signatures to be present in docs

grossws:
With that these functions become visible in generated docs.

Fixes #3229

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -823,13 +823,31 @@ ZSTDLIB_API size_t ZSTD_freeDStream(ZSTD_DStream* zds);  /* accept NULL pointer 
 
 /*===== Streaming decompression functions =====*/
 
-/* This function is redundant with the advanced API and equivalent to:
+/*! ZSTD_initDStream() :
+ * Initialize/reset DStream state for new decompression operation.
+ * Call before new decompression operation using same DStream.
  *
+ * Note : This function is redundant with the advanced API and equivalent to:
  *     ZSTD_DCtx_reset(zds, ZSTD_reset_session_only);
  *     ZSTD_DCtx_refDDict(zds, NULL);
  */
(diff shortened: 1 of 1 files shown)


grossws:
FYI I've just signed ICLA, not sure if @facebook-github-bot will update status automatically

lib/zstd.h
```diff
@@ -828,8 +828,23 @@ ZSTDLIB_API size_t ZSTD_freeDStream(ZSTD_DStream* zds);  /* accept NULL pointer
  *     ZSTD_DCtx_reset(zds, ZSTD_reset_session_only);
  *     ZSTD_DCtx_refDDict(zds, NULL);
  */
-ZSTDLIB_API size_t ZSTD_initDStream(ZSTD_DStream* zds);
-
+ZSTDLIB_API size_t ZSTD_initDStream(ZSTD_DStream* zds); /**< Start new decompression operation */
```
felixhandte: Rather than adding an additional comment, you should modify the comment block above this function so that it appears in the documentation (I guess by adding a first line like `/*! ZSTD_initDStream() :`).

felixhandte (changes requested):

lib/zstd.h
grossws: Ok, I'll update the PR. Thanks for the suggestion

grossws: I've updated the PR

felixhandte (approved):

PR #3241, opened 2022-08-11
restore combine.sh bash performance while still sticking to POSIX

wahern:
POSIX sh `case` statement is just as fast Bash's `=~` extended test operator. It takes a glob pattern, though, so this patch uses it to short-circuit before falling back to grep. This patch effectively restores the original Bash performance, at least as compared with macOS /bin/bash and using the three older Bash regex tests. (It's negligibly slower on account of the #include and #pragma tests being called through a function rather than inline within the if statement.)

I experimented with POSIX expr(1), which takes BREs, but it's not a built-in on FreeBSD, macOS, or OpenBSD /bin/sh and wasn't any faster than invoking grep.

fnmatch function signature comes from Rich Felker's http://www.etalabs.net/sh_tricks.html

### DIFF
--- a/build/single_file_libs/combine.sh
+++ b/build/single_file_libs/combine.sh
@@ -2,11 +2,7 @@
 
 # Tool to bundle multiple C/C++ source files, inlining any includes.
 # 
-# Note: this POSIX-compliant script is many times slower than the original bash
-# implementation (due to the grep calls) but it runs and works everywhere.
-# 
 # TODO: ROOTS, FOUND, etc., as arrays (since they fail on paths with spaces)
-# TODO: revert to Bash-only regex (the grep ones being too slow)
 # 
 # Author: Carl Woffenden, Numfum GmbH (this script is released under a CC0 license/Public Domain)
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks great !

cc @cwoffenden

cwoffenden:
This looks good!

(I'll give it a quick test on some old Unixes at the weekend)

Cyan4973 (approved):

Cyan4973 (approved):

cwoffenden:
Resurrecting this... I eventually got round to booting up an old Mac (Leopard on PPC) and out-of-the-box the shell script fails, but it's _always_ failed on there because the `grep` implementation is broken (and I put a test for said old Mac to abort).

Anyway, replacing the system `grep` with the one from MacPorts (`ggrep`) runs and amalgamating the entire library takes 2m20, which is much better than the previous version (on the same machine the Python implementation runs in 600ms). 

Part of the reason it still takes so long, at least on this machine, is because the resulting `zstd.c` is 5MB vs the approx 2MB of the Python version, due to paths not being canonically resolved (so lots of `..` means files are multiply added).

I'm not sure it's worth the time to test for ggrep, since anyone installing MacPorts will have probably added Python, but I think it worth verifying on other systems that the amalgamated library isn't larger than needs be.

I _think_ it breaks because of changes to how the headers are included. The Python version runs `resolve()` on every include, to make sure there's only one of each:

https://github.com/facebook/zstd/blob/fba704f8bb00ffffd1ee5cf1d35113a9e08b2c0c/build/single_file_libs/combine.py#L113

(After writing all of this I feel a bug report would've been better...)

Cyan4973:
Indeed, a `python` dependency is acceptable for this task.
It's general enough to be present on most systems.
Plus, amalgamation doesn't have to happen on "target" systems, it's enough that one development system can do it.

PR #3242, opened 2022-08-13
Document pass-through behavior

### LINKED ISSUE #3238: Document pass-through behavior in `zstd.1` and `--help`
Document pass-through:
- [X] `--help`
- [x] `man`

Document the always-on pass-through behavior when the `-f` flag is present and when writing to stdout during decompression:
- [x] `--help`
- [x] `man`

Open question: should the current pass-through be documented in the `-f` flag or is it too nuanced to go there. Perhaps all of the information should be in the pass-through flag documentation?

I will have time this weekend to open a pull request if there's no objection.

_also as a note to whoever fixes this theres a misplaced `\n` in the [help string for pass-through](https://github.com/facebook/zstd/blob/1b249cf0752fccf35858f2af8400ca6379e3781a/programs/zstdcli.c#L274)_

cgbur:
Adds documentation to help and man pages for legacy pass-through behavior when force is set and destination is stdout. Documents --pass-through in man pages.

fixes: #3238

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -154,7 +154,8 @@ static void usage(FILE* f, const char* programName)
 #endif
     DISPLAY_F(f, "  -f, --force          disable input and output checks. Allows overwriting existing files,\n");
     DISPLAY_F(f, "                       input from console, output to stdout, operating on links,\n");
-    DISPLAY_F(f, "                       block devices, etc.\n");
+    DISPLAY_F(f, "                       block devices, etc. During decompression and when the output\n");
+    DISPLAY_F(f, "                       destination is stdout, pass-through unrecognized formats as-is.\n");
     DISPLAY_F(f, "      --rm             remove source file(s) after successful de/compression\n");
     DISPLAY_F(f, "  -k, --keep           preserve source file(s) (default) \n");
 #ifdef ZSTD_GZCOMPRESS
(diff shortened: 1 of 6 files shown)


programs/zstdcli.c
```diff
@@ -271,7 +272,7 @@ static void usage_advanced(const char* programName)
             exeNameMatch(programName, ZSTD_GZCAT)) {
             passThroughDefault = "enabled";
         }
-        DISPLAYOUT("     --[no-]pass-through : passes through uncompressed files as-is (default: %s\n)", passThroughDefault);
+        DISPLAYOUT("     --[no-]pass-through : passes through uncompressed files as-is (default: %s)\n", passThroughDefault);
```
terrelln, line 275: Nice catch!

terrelln (approved):
Thanks for the PR!

programs/zstdless.1
```diff
@@ -1,9 +1,14 @@
-.TH "ZSTDLESS" "1" "April 2022" "zstd 1.5.2" "User Commands"
+.
+.TH "ZSTDLESS" "1" "August 2022" "zstd 1.5.3" "User Commands"
```
embg, line 2: @terrelln Is it normal that ronn-ng added whitespace to the man files?

cgbur: I deleted all the man pages and re-ran `make man`. Note that the current zstd man pages were created with https://github.com/rtomayko/ronn NOT the newer, maintained one https://github.com/apjanke/ronn-ng. The only one on AUR is `ronn-ng` but the output that is generated is very different and didn't want to blow up the diff. I think switching may be an option but expect a big diff as the `.` in new lines is not there.

PR #3252, opened 2022-08-28
build(cmake): improve pkg-config generation

Tachi107:
With this patch the pkg-config generation when using the CMake build
system is improved in the following ways:

- Libs.private is now filled when needed
- The JoinPaths module is now used to join paths, leading to simpler code
- The .pc file is always generated, regardless of the platform, as it can also be consumed on Windows

Here's how the .pc file is affected by these changes, in comparison to the one generated with the official Makefiles:

```diff
$ diff -s lib/libzstd.pc build/cmake/build-old/lib/libzstd.pc
15c15
< Libs.private: -pthread
---
> Libs.private:

$ diff -s lib/libzstd.pc build/cmake/build-new/lib/libzstd.pc
Files lib/libzstd.pc and build/cmake/build-new/lib/libzstd.pc are
identical
```

### DIFF
--- a/build/cmake/CMakeModules/JoinPaths.cmake
+++ b/build/cmake/CMakeModules/JoinPaths.cmake
@@ -1,0 +1,23 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+# Windows not supported
+function(join_paths joined_path first_path_segment)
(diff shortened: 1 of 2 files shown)


Cyan4973:
This looks good to me.

It reminds me, it's a bit regrettable that we don't seem to have good CI tests for `pkg-config` generation, that would avoid potential future regressions from future contributions. Though that's a concern legitimately outside of the scope of this PR.

Cyan4973 (approved):

Tachi107:
Maybe something like https://github.com/simdjson/simdjson/pull/1767#issuecomment-1199916435 ?

Cyan4973:
> Maybe something like [simdjson/simdjson#1767 (comment)](https://github.com/simdjson/simdjson/pull/1767#issuecomment-1199916435) ?

Yes, something like that

PR #3256, opened 2022-09-07
streamline `make clean` list maintenance

Cyan4973:
When creating a new `Makefile` target to build,
it's also necessary to update the `clean` target,
which purpose is to remove built targets when they are present.

This process is simple, but it's also easy to forget : since there is a large distance between the position in the `Makefile` where the new built target is added, and the place where the list of files to `clean` is defined. Moreover, the list of files becomes pretty long over time, hence it's difficult to visually ensure that all built targets are present there, or that no old target (no longer produced) is no longer in the list

This PR tries to improve this process by adding a `CLEAN` variable. Now, when a new built target is added to the `Makefile`, it should preceded by :
```
CLEAN += newTarget
newTarget:
<TAB> ...recipe...
```

This new requirement is somewhat similar to `.PHONY: newTarget` for non-built targets.

This new method offers the advantage of locality : there is no separate place in the file to maintain a list of files to clean. This makes maintenance of `make clean` easier.

Also : 
changed `valgrindTest` into `test-valgrind`,
to follow the same convention of all other test targets defined in `tests/Makefile`.

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -157,7 +157,7 @@ jobs:
       run: |
         sudo apt-get -qqq update
         make valgrindinstall
-        make -C tests valgrindTest
+        make -C tests test-valgrind
         make clean
         make -C tests test-fuzzer-stackmode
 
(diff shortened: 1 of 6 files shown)


terrelln (approved):
Should we also do this for the other makefiles?

Cyan4973:
> Should we also do this for the other makefiles?

Probably, 
I wanted to have the principles reviewed before generalizing.

PR #3258, opened 2022-09-08
Fix undefined behavior in ZSTD_decompressStream()

daniellerozenblit:
## TLDR
This PR fixes two instances of undefined behavior (0 offset to Null pointer) in ZSTD_decompressStream(). Also it adds some additional clang ubsan CI jobs, in addition to existing gcc ubsan jobs.

## Reproduce Bug
Run `CC=clang make uasan-test-zstream` with clang11 or higher. Make sure to add the test case that was added in this PR (in tests/zstreamtest.c) since it exposes one of these errors.

## Benchmarking
This bug fix does not seem to negatively impact decompression performance.

I benchmarked on an Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz machine with core isolation and turbo disabled. I measured decompression time for silesia.tar (203M), compiling with GCC11 and clang15. I ran each scenario 5 times interleaved to eliminate bias from background processes. 

### Clang Results
original: 0m0.472s = 430M/s
this PR: 0m0.469s = 433M/s

This PR increased compression speed by around 0.698%, probably due to random noise.

### GCC Results
original: 0m0.466s = 436M/s
this PR: 0m0.466s = 436M/s

This PR did not change compression speed (0.00% decrease).

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -50,6 +50,13 @@ jobs:
     - name: thread sanitizer zstreamtest
       run: CC=clang ZSTREAM_TESTTIME=-T3mn make tsan-test-zstream
 
+  ubsan-zstreamtest:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: undefined behavior sanitizer zstreamtest
+      run: CC=clang make uasan-test-zstream
+
   # lasts ~15mn
   tsan-fuzztest:
     runs-on: ubuntu-latest
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -76,6 +76,15 @@ jobs:
         make gcc8install
         CC=gcc-8 make -j uasan-test-zstd </dev/null V=1
 
+  clang-asan-ubsan-testzstd:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Test Zstd
+      run: |
+        sudo apt-get -qqq update
+        CC=clang make -j uasan-test-zstd </dev/null V=1
+
   gcc-asan-ubsan-testzstd-32bit:
     runs-on: ubuntu-latest
     steps:
@@ -85,6 +94,16 @@ jobs:
         sudo apt-get -qqq update
         make libc6install
         make -j uasan-test-zstd32 V=1
+
+  clang-asan-ubsan-testzstd-32bit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Test Zstd, 32bit mode
+      run: |
+        sudo apt-get -qqq update
+        make libc6install
+        CC=clang make -j uasan-test-zstd32 V=1
 
     # Note : external libraries must be turned off when using MSAN tests,
     # because they are not msan-instrumented,
@@ -100,6 +119,15 @@ jobs:
         make gcc8install
         CC=gcc-8 FUZZER_FLAGS="--long-tests" make clean uasan-fuzztest
 
+  clang-asan-ubsan-fuzz:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Fuzz Test
+      run: |
+        sudo apt-get -qqq update
+        CC=clang FUZZER_FLAGS="--long-tests" make clean uasan-fuzztest
+
   gcc-asan-ubsan-fuzz32:
     runs-on: ubuntu-latest
     steps:
@@ -110,12 +138,29 @@ jobs:
         make libc6install
         CFLAGS="-O3 -m32" FUZZER_FLAGS="--long-tests" make uasan-fuzztest
 
+  clang-asan-ubsan-fuzz32:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Fuzz Test 32bit
+      run: |
+        sudo apt-get -qqq update
+        make libc6install
+        CC=clang CFLAGS="-O3 -m32" FUZZER_FLAGS="--long-tests" make uasan-fuzztest
+
   asan-ubsan-regression:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
     - name: ASan + UBSan + Regression Test
       run: make -j uasanregressiontest
+
+  clang-ubsan-regression:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Regression Test
+      run: CC=clang make -j uasanregressiontest
 
   msan-regression:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -2179,7 +2179,9 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
         case zdss_flush:
             {   size_t const toFlushSize = zds->outEnd - zds->outStart;
                 size_t const flushedSize = ZSTD_limitCopy(op, (size_t)(oend-op), zds->outBuff + zds->outStart, toFlushSize);
-                op += flushedSize;
+
+                op = op ? op + flushedSize : op;
```
Cyan4973: If `op == NULL` here,
then it was already `NULL` at previous line,
where it's used within `(oend-op)` operation,
which already does not seem proper.

Therefore, if `op==NULL`, my suggestion would be to skip the whole flush operation,
which is understandable : there is simply no room to flush anything.

In which case, the whole block could become conditional, starting with a `if (op != NULL)`.
When `op==NULL`, it jumps straight to the end: `someMoreWork = 0; break;`.

The point is less about speed optimization, than about clarity of situation : 
inside the `{ ... }` code block, there is now a guarantee that `op != NULL`.

lib/decompress/zstd_decompress.c
```diff
@@ -2059,7 +2059,7 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
                     if (ZSTD_isError(decompressedSize)) return decompressedSize;
                     DEBUGLOG(4, "shortcut to single-pass ZSTD_decompress_usingDDict()")
                     ip = istart + cSize;
-                    op += decompressedSize;
+                    op = op ? op + decompressedSize : op;
```
Cyan4973: Detailing the thought process :

That such a scenario can even be triggered in `ubsan` tests feels strange initially: 
for execution to reach this code position, it needs to pass the test : 
`(oend-op) >= zds->fParams.frameContentSize`.
So, if there is any content, then it's required that `op != NULL`.

However, thinking about edge cases,
it's indeed possible that `frameContentSize==0`.
This is an empty frame (yet still a valid frame).
It's obviously not an expected outcome in general, the point is that it _can_ happen, it's allowed.
In which case, it becomes technically possible to "decompress" such an empty frame into `NULL`.

Now, there are more philosophical questions associated to this scenario : 
Is it normal to ask to decompress a frame into a `NULL` buffer ?
Should it be an instant fail ? or should it succeed in the specific case where `frameContentSize==0` ?

Note that this code block is just an "optimization" : it triggers single-pass mode when space conditions are met, which requires less workload than full-featured streaming mode.
Which means that skipping this block is not an "end" sentence : it just requires to employ the more complex streaming mode.

Therefore, one possible idea here could be to add a `op != NULL` condition to enter this code block,
so that `op==NULL` is no longer a possible scenario to support inside the code block
(it must employ the more general streaming code path).
However, such a solution does not appear "obviously" better than the ternary operator you selected.

I would say, the only piece of information missing is why the scenario (`op==NULL`) can technically happen here, justifying the existence of this ternary operator to any future contributor reviewing the code.

lib/decompress/zstd_decompress.c
```diff
@@ -2177,21 +2177,24 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
...
-                      && (zds->outStart + zds->fParams.blockSizeMax > zds->outBuffSize) ) {
+                    && (zds->outStart + zds->fParams.blockSizeMax > zds->outBuffSize) ) {
                         DEBUGLOG(5, "restart filling outBuff from beginning (left:%i, needed:%u)",
                                 (int)(zds->outBuffSize - zds->outStart),
                                 (U32)zds->fParams.blockSizeMax);
                         zds->outStart = zds->outEnd = 0;
                     }
                     break;
-            }   }
+            }}
```
embg: Minor nit: could you leave space between these closing brackets, like there was before?

lib/decompress/zstd_decompress.c
embg: Agree that a very brief comment would be useful here.

.github/workflows/dev-long-tests.yml
```diff
@@ -69,6 +76,15 @@ jobs:
         make gcc8install
         CC=gcc-8 make -j uasan-test-zstd </dev/null V=1
 
+  clang-asan-ubsan-testzstd:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Test Zstd
+      run: |
+        sudo apt-get -qqq update
```
embg: I don't think you need the `apt-get` command. It's used in the gcc-8 test because gcc8 is installed via apt by `make gcc8install`. If you are using the clang that's already installed in the CI environment, then I think you're good to delete the `apt-get`.

However, you should check the CI logs and make sure that clang is indeed present without requiring installation.

.github/workflows/dev-long-tests.yml
```diff
@@ -93,6 +109,15 @@ jobs:
         make gcc8install
         CC=gcc-8 FUZZER_FLAGS="--long-tests" make clean uasan-fuzztest
 
+  clang-asan-ubsan-fuzz:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: clang + ASan + UBSan + Fuzz Test
+      run: |
+        sudo apt-get -qqq update
```
embg: Same

lib/decompress/zstd_decompress.c
```diff
@@ -2177,21 +2177,24 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
                 break;
             }
         case zdss_flush:
-            {   size_t const toFlushSize = zds->outEnd - zds->outStart;
+            if (op != NULL) {
```
embg, line 2180: I like that you blocked off the whole flush operation, rather than just using a ternary on `op += flushedSize`. It's much more clear this way, and I actually didn't consider this approach when I looked at the task :)

tests/Makefile
```diff
@@ -183,6 +183,10 @@ zstreamtest_tsan : CFLAGS += -fsanitize=thread
 zstreamtest_tsan : $(ZSTREAMFILES)
 	$(LINK.c) $(MULTITHREAD) $^ -o $@$(EXT)
 
+zstreamtest_ubsan : CFLAGS += -fsanitize=undefined
```
embg, line 186: Watch out for merge conflicts with Yann's PR: https://github.com/facebook/zstd/pull/3256

tests/zstreamtest.c
```diff
@@ -548,6 +548,34 @@ static int basicUnitTests(U32 seed, double compressibility)
     {   size_t const ret = ZSTD_decompressStream(zd, &outBuff, &inBuff);
         if (ret != 0) goto _output_error;
     }
+
+    {
```
embg, line 554: Please add a title for this test like the previous one. You should probably edit the title of the previous one as well, so it's clear what scenarios are being tested in both cases.

embg (changes requested):
This PR looks great! Just a few very minor nits, and we should be good to land.

Cyan4973 (approved):

PR #3262, opened 2022-09-09
Add sequenceBound(srcSize) method

daniellerozenblit:
This PR adds a new method `ZSTD_sequenceBound(size_t srcSize)` which returns the upper bound of the number of sequences that can be generated from a buffer of `srcSize` bytes.

Note that this is calculated using `ZSTD_MINMATCH_MIN`, which is the minimum possible length of a match. It may be more accurate to get the minimum match length from the CCtx, but this is the safest option. Consider a scenario where someone calls ZSTD_sequenceBound(cctx, srcSize), change minMatch, and then call generateSequences –– in this case, the output buffer might be too small and overflow.

### DIFF
--- a/contrib/seqBench/seqBench.c
+++ b/contrib/seqBench/seqBench.c
@@ -22,9 +22,7 @@ int main(int argc, char *argv[]) {
     fread(inBuf, inBufSize, 1, f);
     fclose(f);
 
-    // Should work fine for this benchmark, but we really need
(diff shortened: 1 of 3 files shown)


embg:
LGTM, thanks for fleshing out the sequence compression API!

embg (approved):

Cyan4973 (approved):

PR #3263, opened 2022-09-12
ZSTD_decompressStream() fuzz fix

daniellerozenblit:
This PR changes some of the logic in a previous PR [(#3258)](https://github.com/facebook/zstd/pull/3258) in response to a failing oss-fuzz test.

In the previous PR, we skip the flush operation in the case of a `op=NULL`. This causes an assertion failure later on when checking for forward progress.

Instead of skipping the entire flush operation, we now simply use a ternary operator to increment `op` when it is non-null.

Note: These were the input and output buffer parameters in ZSTD_decompressStream that caused the test failure.
Screen Shot 2022-09-12 at 6 02 16 PM

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -2178,23 +2178,23 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
             }
         case zdss_flush:
             {
-            size_t const toFlushSize = zds->outEnd - zds->outStart;
-            size_t const flushedSize = ZSTD_limitCopy(op, (size_t)(oend-op), zds->outBuff + zds->outStart, toFlushSize);
+                size_t const toFlushSize = zds->outEnd - zds->outStart;
+                size_t const flushedSize = ZSTD_limitCopy(op, (size_t)(oend-op), zds->outBuff + zds->outStart, toFlushSize);
 
-            op = op ? op + flushedSize : op;
+                op = op ? op + flushedSize : op;
 
-            zds->outStart += flushedSize;
-            if (flushedSize == toFlushSize) {  /* flush completed */
-                zds->streamStage = zdss_read;
-                if ( (zds->outBuffSize < zds->fParams.frameContentSize)
-                    && (zds->outStart + zds->fParams.blockSizeMax > zds->outBuffSize) ) {
-                    DEBUGLOG(5, "restart filling outBuff from beginning (left:%i, needed:%u)",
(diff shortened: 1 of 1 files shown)


lib/decompress/zstd_decompress.c
```diff
@@ -2177,24 +2177,24 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
...
-                    }
-                    break;
-            }   }
+                    DEBUGLOG(5, "restart filling outBuff from beginning (left:%i, needed:%u)",
+                            (int)(zds->outBuffSize - zds->outStart),
+                            (U32)zds->fParams.blockSizeMax);
+                    zds->outStart = zds->outEnd = 0;
+                }
+                break;
+            } }
```
Cyan4973: The indentation seems off.
There shouldn't be a need for 2 closing brackets grouped here.
Likely the block wasn't started with correct indentation.

Cyan4973:
> This PR changes some of the logic in a previous PR [(#3258)](https://github.com/facebook/zstd/pull/3258) in response to a failing oss-fuzz test.
> 
> In the previous PR, we skip the flush operation in the case of a `op=NULL`. This causes an assertion failure later on when checking for forward progress.
> 
> Instead of skipping the entire flush operation, we now simply use a ternary operator to increment `op` when it is non-null.

The explanation is interesting.
I want to dig more.

The issue is stated to be related to forward progress.
Indeed, the library tries to ensure maximum forward progress at each invocation.
An important goal is to avoid busy loops (useless invocations with no work done).

However, if the destination buffer is `NULL`, by definition, no flush is possible.
Skipping the flush operation doesn't change that fact.

So where could the difference come from ?
The only case I can imagine, where skipping makes a difference compared to not skipping, is if `toFlushSize==0`.
In which case, the state machine moves to `zdss_read` stage, instead of stopping compression operations.

But is it possible to reach `case zdss_flush:` 
with `toFlushSize==0` ?
I assume it was not, but maybe there are scenarios where that's actually possible ?

Note that the proposed fix likely works correctly,
the purpose here is more to ensure we understand the code's behavior properly.

daniellerozenblit:
@Cyan4973 I looked further into the particular test case that failed and it does seem that `toFlushSize == 
flushedSize == 0`. I added the input and output buffer information to the PR description.

I am not sure exactly why this is the case, though I will try to look into it more.

Cyan4973:
The specific scenario mentioned in the first post
looks like an empty frame,
aka a valid `zstd` frame, which contains a single block, which is empty.

Cyan4973 (approved):

PR #3267, opened 2022-09-20
ci: test pkg-config file

Tachi107:
As mentioned in https://github.com/facebook/zstd/pull/3252#issuecomment-1251733791, this patch adds a CI job that builds and installs libzstd on the job runner, and then compiles a sample binary linking against the installed library; the needed build flags are passed by invoking pkg-config.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -410,6 +410,22 @@ jobs:
       run: |
         sde-external-9.0.0-2021-11-07-lin/sde -cet -cet-raise 0 -cet-endbr-exe -cet-stderr -cet-abort -- ./zstd -b3
 
+  pkg-config:
+    runs-on: ubuntu-latest
+    container:
+      image: debian:testing
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-short-tests.yml
```diff
@@ -410,6 +410,22 @@ jobs:
...
+  pkg-config:
+    runs-on: ubuntu-latest
+    container:
+      image: debian:testing
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          apt -y update
+          apt -y install --no-install-recommends gcc libc6-dev make pkg-config
```
Tachi107, line 422: I've added `--no-install-recommends` just to speed up the job by a few seconds.
It skips the installation of things like manpages that are not needed on a headless build server.

Cyan4973 (approved):

Cyan4973:
Thanks for adding this useful CI test @Tachi107 !

The lone CI error is unrelated to this PR, and was recently fixed by @yoniko in #3272

Tachi107:
Glad to help!

PR #3268, opened 2022-09-20
Fix for `zstd` CLI accepts bogus values for numeric parameters

ctkhanhly:
Related Issue to #3070 

Notes:
- Originally, I was gonna throw an error in `readU32FromCharChecked` but this affects at least `init_cLevel`

Testing:
- Created a test `memlimit.sh`

@yoniko


tests/cli-tests/basic/memlimit.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
Cyan4973, line 1: Good tests ! 
👍

programs/zstdcli.c
```diff
@@ -1013,8 +1013,16 @@ int main(int argCount, const char* argv[])
...
                 if (longCommandWArg(&argument, "--threads")) { NEXT_UINT32(nbWorkers); continue; }
-                if (longCommandWArg(&argument, "--memlimit")) { NEXT_UINT32(memLimit); continue; }
-                if (longCommandWArg(&argument, "--memory")) { NEXT_UINT32(memLimit); continue; }
+                if (longCommandWArg(&argument, "--memory") || longCommandWArg(&argument, "--memlimit")) {
+
+                    const char* memStrPtr;
+                    NEXT_FIELD(memStrPtr);
+                    memLimit = readU32FromChar(&memStrPtr);
+                    if(*memStrPtr != 0) {
+                        errorOut("error: only numeric values with optional suffices K, KB, KiB, M, MB, MiB are allowed");
```
Cyan4973: nit: `suffices` => `suffixes`

yoniko: Typo, plural form of suffix is suffixes.

Cyan4973:
Indeed, the solution cannot be done directly within `readU32FromCharChecked()`,
because this function is used to read unsigned integers from the command line
in both _short command_ (typically one letter) and `--long-command=` scenarios.
Erroring on wrong suffixes only makes sense in the case of the `--long-command=` scenario
because _short_ commands are allowed to stack up (for example `-bi7S` is valid).

The `--long-command=` scenario happens in 6 places,
and the logic is abstracted behind the macro `NEXT_UINT32()`.

I believe your fix is correct, but it's written directly to serve `--memory=` and `--memlimit=`, and therefore limited to these cases.

If you update `NEXT_UINT32()` with your corrected logic, you would fix all current and future usages of numeric parameters in `--long-command=` scenarios.

yoniko (changes requested):
Thank you @ctkhanhly!
Tests and logic look good.
There's one change that I'd like us to do here - there are other fields that use `NEXT_UINT32` to read sizes and which use the same logic for parsing suffixes.
It'd be better to change the `NEXT_UINT32` macro and so apply the same validation on those fields as well.

yoniko:
@ctkhanhly - if you have some time it'd also be useful to get the same treatment for the fields that directly use `readSizeTFromChar`.
The solution I'd opt for in this case is to create a `NEXT_TSIZE` that'd be the same as `NEXT_UINT32` only it'd call `readSizeTFromChar` instead of `readU32FromChar` and use it for these fields.

programs/zstdcli.c
```diff
@@ -1013,8 +1013,16 @@ int main(int argCount, const char* argv[])
                 }
 #endif
                 if (longCommandWArg(&argument, "--threads")) { NEXT_UINT32(nbWorkers); continue; }
-                if (longCommandWArg(&argument, "--memlimit")) { NEXT_UINT32(memLimit); continue; }
-                if (longCommandWArg(&argument, "--memory")) { NEXT_UINT32(memLimit); continue; }
+                if (longCommandWArg(&argument, "--memory") || longCommandWArg(&argument, "--memlimit")) {
```
Cyan4973: minor (for information): 
writing the branch this way makes the following (obviously wrong) command valid : 
`--memory=--memlimit=`
because `&argument` is updated twice.

Cyan4973:
I see a remaining test issue in `mingw-long-test` (CI).

It looks completely unrelated to this PR.
Maybe the compiler version changed for example, introducing a new warning which is flagged as an error in this test.

Anyway, I think you can safely ignore it, we'll deal with it separately.

programs/zstdcli.c
ctkhanhly: I'll fix this!

programs/zstdcli.c
ctkhanhly: my bad @@

ctkhanhly: This is fixed!

Cyan4973 (approved):

yoniko (approved):
Looks great, thank you!

Cyan4973:
Thanks @ctkhanhly , great bootcamp contribution !

Issue #3269, opened 2022-09-21
Enable the OpenSSF Scorecard Github Action


joycebrum:
Hello, I'm working on behalf of Google and the [Open Source Security Foundation][ossf] to help essential open-source projects improve their supply-chain security. Given relevance that the Zstandard has on countless projects, the OpenSSF has identified it as one of the 100 most critical open source projects.

**Is your feature request related to a problem? Please describe.**
According to [Open Source Security and Risk Analysis Report](https://www.synopsys.com/software-integrity/resources/analyst-reports/open-source-security-risk-analysis.html?intcmp=sig-blog-ossra1), 84% of all codebases have at least one vulnerability, with an average of 158 per codebase. The majority have been in the code for more than 2 years and have documented solutions available.

Even in large tech companies, the tedious process of reviewing code for vulnerabilities falls down the priority list, and there is little insight into known vulnerabilities and solutions that companies can draw on.

That’s where the OpenSSF tool called [Scorecards][sc] is helping. Its focus is to understand the security posture of a project and assess the risks that the dependencies could introduce.

**Describe the solution you'd like**
Scorecards runs dozens of automated security checks to help maintainers better understand their project's supply-chain security posture. It is developed by the OpenSSF, [in partnership with GitHub][sc-blog].

To simplify maintainers' lives, the OpenSSF has also developed the [Scorecard GitHub Action][sc-gha]. It is very lightweight and runs on every change to the repository's main branch. The results of its checks are available on the project's [security dashboard](https://github.com/facebook/zstd/security), and include suggestions on how to solve any issues (see examples in additional context). The Action does not run or interact with any workflows, but merely parses them to identify possible vulnerabilities. This Action has been adopted by 1800+ projects already.

Zstandard already follow many of the Scorecard recommended best practices and criterias for a greater security, such as not having any binary artifacts, CI-Tests, code review, Fuzzing, etc. Although, there are still some criterias that would need to be improved to achieve a good level of security. In these cases, the Scorecard Github Action can help on diagnosing and proposing solutions.  

Would you be interested in a PR which adds this Action? Optionally, it can also publish your results to the OpenSSF REST API, which allows a [badge][badge] with the project's score to be added to its README.

**Additional context**

![Code scanning dashboard with multiple alerts, including Code-Review and Token-Permissions][img-security]

![Detail of a Token-Permissions alert, indicating the specific file and remediation steps][img-detail]

[...]

PR #3270, opened 2022-09-21
[RFC] Create CodeQL Workflow

felixhandte:
Creates the workflow suggested by #3269.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,74 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "dev", gh-pages ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "dev" ]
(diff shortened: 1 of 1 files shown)


felixhandte:
Looks like this produced 0 alerts...: https://github.com/facebook/zstd/security/code-scanning?query=is%3Aopen+pr%3A3270

Not totally sure I've done this right. Alternatively, maybe our code is perfect.

Cyan4973:
I'm a bit lost here, as #3269 is about Google OpenSSF ScoreCard,
while this PR seems more about Github CodeQL Action.

Both seem valid and interesting, but they seem different, complementary maybe.

Cyan4973:
> Looks like this produced 0 alerts...: https://github.com/facebook/zstd/security/code-scanning?query=is%3Aopen+pr%3A3270
> 
> Not totally sure I've done this right. Alternatively, maybe our code is perfect.

Well, if this tool only throws alerts when it's pretty sure to have found something worth investigating,
then it's not too far from expectation.
After all, we are already using a lot of bug mitigation techniques in this code base.

And if that's the reason, it's a good thing :  
I very much prefer a tool which alert level can be trusted, and therefore acted upon,
as opposed to tools which send too many alerts in the hope that maybe one of them might be interesting, 
thus destroying the value of the alert itself.

Now, of course, as a first test, it opens the possibility that it might simply be not working at all.
To cover this risk, try to enable this workflow on another code base,
one that you know should (or make up to) trigger some kind of CodeQL warning because it's obviously wrong.

felixhandte:
So indeed, trying it on a PR on my fork with obviously incorrect code failed to produce any alerts: https://github.com/felixhandte/zstd/security/code-scanning?query=is%3Aopen+pr%3A10+. I'm not going to invest more effort into this now.

PR #3276, opened 2022-09-28
meson: make backtrace dependency on execinfo

neheb:
musl libc for example has no such header.

### DIFF
--- a/build/meson/programs/meson.build
+++ b/build/meson/programs/meson.build
@@ -42,7 +42,8 @@ endif
 
 export_dynamic_on_windows = false
 # explicit backtrace enable/disable for Linux & Darwin
-if not use_backtrace
+execinfo = cc.has_header('execinfo.h', required: use_backtrace)
+if not execinfo.found()
   zstd_c_args += '-DBACKTRACE_ENABLE=0'
 elif use_debug and host_machine_os == os_windows  # MinGW target
   zstd_c_args += '-DBACKTRACE_ENABLE=1'
--- a/build/meson/meson_options.txt
+++ b/build/meson/meson_options.txt
@@ -14,7 +14,7 @@ option('legacy_level', type: 'integer', min: 0, max: 7, value: '5',
   description: 'Support any legacy format: 7 to 1 for v0.7+ to v0.1+')
 option('debug_level', type: 'integer', min: 0, max: 9, value: 1,
   description: 'Enable run-time debug. See lib/common/debug.h')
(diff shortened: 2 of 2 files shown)


build/meson/programs/meson.build
```diff
@@ -42,7 +42,8 @@ endif
 
 export_dynamic_on_windows = false
 # explicit backtrace enable/disable for Linux & Darwin
-if not use_backtrace
+execinfo = cc.has_header('execinfo.h', required: use_backtrace)
+if execinfo.found()
```
eli-schwartz: This changed from `not use_backtrace` to `execinfo.found()` which seems inverted?

neheb: Yeah mistake

build/meson/programs/meson.build
```diff
@@ -42,7 +42,8 @@ endif
 
 export_dynamic_on_windows = false
 # explicit backtrace enable/disable for Linux & Darwin
-if not use_backtrace
+execinfo = cc.has_header('execinfo.h', required: use_backtrace)
+if not execinfo.found()
```
terrelln, line 46:
```suggestion
if not use_backtrace or not execinfo.found()
```

terrelln (changes requested):
This could enable backtraces when `use_backtrace` is false. Otherwise it looks good to me.

eli-schwartz:
`use_backtrace` cannot be false anymore, it was changed away from a boolean.

It's now a feature option with the default value of "disabled" which means that when used in the `required:` kwarg, it logs that it is skipping the check "because feature backtrace is disabled".

terrelln:
@eli-schwartz Thanks for the info! I checked out the source code and it looks like you're correct.

I previously looked at the  [documentation](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerhas_header) for `has_header()`, and I saw that:

> required: If set to true, Meson will halt if the header check fails.

This seems to me that if `required = disabled` then `execinfo.found()` could still return true. So it seems the documentation is out of date. Likely it wasn't updated when required was changed to accept a `feature`.

terrelln (approved):
Thanks for the PR!

eli-schwartz:
> So it seems the documentation is out of date. Likely it wasn't updated when required was changed to accept a `feature`.

Actually, it's not described anywhere else it is accepted either.

I've reworded the docs slightly, a new version of the website is in progress.

PR #3279, opened 2022-10-03
Rename {clean-all,uninstall} targets  -> zstd_{clean-all,uninstall}

jbms:
This avoids target namespace pollution when including zstd as a sub-project.


yoniko:
Hi @jbms ,
Thank you for your contribution, I'm not a CMake expert but I don't feel comfortable with the solution - it changes an interface that other users may depend on.
Additionally, it seems redundant to have the library name as part of a prefix for the command.
I think a better solution would be something similar to gflag's solution, where they don't add the commands if included as a subproject.

They allow the user to set a flag if the project is a subproject and if it doesn't exist, they autodetect: https://github.com/gflags/gflags/blob/master/CMakeLists.txt#L134.
I think a similar solution would be significantly better.

yoniko:
Added an issue to implement the flag based solution (#3468).
Closing this PR.

PR #3282, opened 2022-10-05
programs/Makefile: fix zstd-pgo target

ldv-alt:
Apparently, zstd-pgo target was broken at least for gcc compiler since the commit that introduced BUILD_DIR for programs subdirectory. Starting with that commit, the object directory used for generating profile data is different from the object directory used for producing the result based on that profile data, and this of course does not work with gcc which uses the object directory to store the profile data. Fortunately, gcc detects this and complains by issuing a warning for each object file, e.g.

  ../lib/common/zstd_common.c:83:1: warning: '/path/to/zstd/programs/obj/conf_3da26c95555a8e4d4ab71a33da334472/zstd_common.gcda' profile count data file not found [-Wmissing-profile]

Fix this by forwarding HASH_DIR so that the same BUILD_DIR is used both for generating and using profile data.

Fixes: dd2449695 ("programs/zstd also automatically generate object dir per conf")
Co-authored-by: Ilya Kurdyukov <ilyakurdyukov@basealt.ru>

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -232,17 +232,21 @@ zstd-dll : zstd
 ## zstd-pgo: zstd executable optimized with PGO.
 .PHONY: zstd-pgo
 zstd-pgo :
-	$(MAKE) clean
-	$(MAKE) zstd MOREFLAGS=-fprofile-generate
+	$(MAKE) clean HASH_DIR=$(HASH_DIR)
+	$(MAKE) zstd HASH_DIR=$(HASH_DIR) MOREFLAGS=-fprofile-generate
 	./zstd -b19i1 $(PROFILE_WITH)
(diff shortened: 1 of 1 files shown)


Cyan4973:
Hi Dmitry,

seems like Ilya already provided the same patch at https://github.com/facebook/zstd/pull/3281 ?

ldv-alt:
> Hi Dmitry,
> 
> seems like Ilya already provided the same patch at #3281 ?

I and Ilya discussed the issue beforehand, I haven't noticed that he has already submitted a fix.
With regards to the patch, looks like the edition submitted by Ilya misses that `$(BUILD_DIR)/zstd` has to be removed as well.

Cyan4973:
equivalent patch already merged in #3281

PR #3289, opened 2022-10-13
Don't attempt playTests.sh cmake test if running on Windows.

nmoinvaz:
Disable this test on Windows because it runs a bash script. I'm not sure of a way to easily determine if the shell is bash, so I have just disabled it on Windows.

### DIFF
--- a/build/cmake/tests/CMakeLists.txt
+++ b/build/cmake/tests/CMakeLists.txt
@@ -96,13 +96,14 @@ add_test(NAME zstreamtest COMMAND zstreamtest ${ZSTD_ZSTREAM_FLAGS})
 AddTestFlagsOption(ZSTD_PLAYTESTS_FLAGS "$ENV{PLAYTESTS_FLAGS}"
     "Semicolon-separated list of flags to pass to the playTests.sh test")
 add_test(NAME playTests COMMAND sh -c "\"${TESTS_DIR}/playTests.sh\" ${ZSTD_PLAYTESTS_FLAGS}")
-if (ZSTD_BUILD_PROGRAMS)
+find_program(UNAME uname) # Run script only in unix shell environments
+if (ZSTD_BUILD_PROGRAMS AND UNAME)
     set_property(TEST playTests APPEND PROPERTY ENVIRONMENT
         "ZSTD_BIN=$<TARGET_FILE:zstd>"
         "DATAGEN_BIN=$<TARGET_FILE:datagen>"
         )
 else()
-    message(STATUS "Disabling playTests.sh test because ZSTD_BUILD_PROGRAMS is not enabled")
+    message(STATUS "Disabling playTests.sh test because requirements not met")
     set_tests_properties(playTests PROPERTIES DISABLED YES)
 endif()
 


Cyan4973:
It seems there is already a property flag to decide running `playTests.sh` or not.
Maybe you could re-use this mechanism ?

Also, I'm not sure if "Windows" is a good enough discriminant for such a decision.
For example, running a bash shell on Windows works fine when using environments like mingw and cygwin, so they should not be impacted by this decision.
Instead, what you may mean is "Windows + Visual Studio + MS Command Prompt".
I'm concerned that `WIN32` doesn't capture this difference.

Finally, maybe we need a CI test to ensure this property is running as intended and continues to run as intended in the future, when future contributors change some part of the code base.

nmoinvaz:
> It seems there is already a property flag to decide running playTests.sh or not.

Are you referring to `ZSTD_PLAYTESTS_FLAGS`? That appears to be just flags to pass into the shell script. The shell script will get executed even if no `ZSTD_PLAYTESTS_FLAGS` are passed.

As it currently sits this cmake tests would fail to run on Windows, using `cmd` or `PowerShell` or `Windows Terminal`.

Cyan4973:
What about:
```
set_tests_properties(playTests PROPERTIES DISABLED YES)
```

nmoinvaz:
I don't think that is it... I'm not sure how I would enable/disable that. I could do a check for `uname`?

```cmake
find_program(UNAME uname)
if (UNAME)
  // Must be shell terminal
else()
  // Must not be
endif()
```

Cyan4973:
> I could do a check for `uname`?

`uname` is likely safer.
I would expect this command to offer a clearer distinction between Microsoft environments, like cmd and powershell, and posix ones like msys2 and cygwin.

nmoinvaz:
Signed.

nmoinvaz:
Added check for uname.

Cyan4973 (approved):

nmoinvaz:
There is another problem, that playTest.sh is triggered when the project is added with `EXCLUDE_FROM_ALL`.
```
add_subdirectory(${ZSTD_SOURCE_DIR}/build/cmake ${ZSTD_BINARY_DIR} EXCLUDE_FROM_ALL)
```
The problem is that `datagen`  and apps are not built with `EXCLUDE_FROM_ALL`.

nmoinvaz:
Also is it weird that `add_test` still happens for `playTest` even if `ZSTD_BUILD_PROGRAMS` is OFF? Seems like the list of tests could be built as a list and playTest added conditionally. Instead of creating the test and then disabling it..

nmoinvaz:
I fixed the logic around the uname detection, but don't have a solution for detecting if datagen will be built.

nmoinvaz:
> Seems like the list of tests could be built as a list and playTest added conditionally. Instead of creating the test and then disabling it..

Looks like using `DISABLED` might be better because the output looks like:
```
        Start 300: playTests
300/301 Test #300: playTests ........................***Not Run (Disabled)   0.00 sec
```

Cyan4973:
> I fixed the logic around the uname detection, but don't have a solution for detecting if datagen will be built.

Does it matter if `datagen` is built ?
I presume the only issue is about running the shell script.

nmoinvaz:
> Does it matter if datagen is built?

Yes, it does matter because `playTests.sh` depends on `datagen` and runs it. If include `zstd` using `add_subdirectory(zstd EXCLUDE_FROM_ALL)` then `datagen` will not be built but `playTests.sh` will be run in CTest. 

> I presume the only issue is about running the shell script.

Correct. Now this PR makes it so that `playTest.sh` only runs in unix environment. This is different from the `datagen` issue. It looks like the problem with `EXCLUDE_FROM_ALL` and CTest has been run into before. I have documented your use case [here](https://gitlab.kitware.com/cmake/cmake/-/issues/20212#note_1271697).

terrelln:
Thanks for the PR, it LGTM!

Cyan4973 (approved):

PR #3294, opened 2022-10-18
[contrib][linux-kernel] Generate SPDX license identifiers

### LINKED ISSUE #3293: Linux kernel: generate SPDX license identifiers
Since now the generation of the Linux zstd code is fully automated, it would be nice to prefix the generated files with the `SPDX` license identifiers as needed.

ojeda:
Add a `--spdx` option to the freestanding script (and use it from the `linux-kernel` `Makefile`) to prefix files with a line like (for `.c` files):

    // SPDX-License-Identifier: GPL-2.0+ OR BSD-3-Clause

or (for `.h` and `.S` files):

    /* SPDX-License-Identifier: GPL-2.0+ OR BSD-3-Clause */

Given the style of the line to be used depends on the extension, a simple `sed` insert command would not work.

It also skips the file if an existing SPDX line is there, as well as raising an error if an unexpected `SPDX-License-Identifier` string appears anywhere else in the file, as well as for unexpected file extensions.

I double-checked that all currently generated files appear to be licensed as expected with:

    grep -LRF 'This source code is licensed under both the BSD-style license (found in the'  linux/lib/zstd
    grep -LRF 'LICENSE file in the root directory of this source tree) and the GPLv2 (found' linux/lib/zstd

but somebody knowledgeable on the licensing of the project should double-check this is the intended case.

Fixes: https://github.com/facebook/zstd/issues/3293
Signed-off-by: Miguel Ojeda <ojeda@kernel.org>

### DIFF
--- a/contrib/freestanding_lib/freestanding.py
+++ b/contrib/freestanding_lib/freestanding.py
@@ -431,7 +431,7 @@ class Freestanding(object):
             external_xxhash: bool, xxh64_state: Optional[str],
             xxh64_prefix: Optional[str], rewritten_includes: [(str, str)],
             defs: [(str, Optional[str])], replaces: [(str, str)],
-            undefs: [str], excludes: [str], seds: [str],
+            undefs: [str], excludes: [str], seds: [str], spdx: bool,
     ):
         self._zstd_deps = zstd_deps
         self._mem = mem
@@ -446,6 +446,7 @@ class Freestanding(object):
         self._undefs = undefs
         self._excludes = excludes
         self._seds = seds
+        self._spdx = spdx
 
     def _dst_lib_file_paths(self):
         """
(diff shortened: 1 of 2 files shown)


terrelln:
Yeah, all of our files share the same license. We have a test to make sure that they all have the a consistent copyright + license.

This looks great! Thanks for the PR @ojeda! I will back port it to the changes I'm making on top of the v1.5.2 release for the next kernel update.

ojeda:
> Yeah, all of our files share the same license. We have a test to make sure that they all have the a consistent copyright + license.

That is a nice test to have indeed.

> This looks great! Thanks for the PR @ojeda! I will back port it to the changes I'm making on top of the v1.5.2 release for the next kernel update.

My pleasure! After I sent you that email, I thought I could give it a go :)

Issue #3275, opened 2022-09-27
Update linux kernel to latest zstd (from 1.4.10)


kdave:
Hi, people of reddit asked (https://old.reddit.com/r/kernel/comments/xp2o53/why_is_the_version_of_zstd_in_the_kernel_outdated/) about updating the linux port of zstd to newer version. This has been promised back then when 1.4.10 was synced a year ago and I'd be glad to see an update (or even a regular update in each kernel release) as btrfs is using zstd and any performance improvement is most welcome.

There's automation support (contrib/linux-kernel) so generating the patch itself should not take too much time. As there are more things to do like review and benchmarking it's not the last step but at least the preliminary version can be added to `linux-next` tree. The final pull request sent to Linus may or may not happen depending on the testing results.

IMHO neglecting the regular updates is more "expensive", like it was with the 1.4.10 update that took about a year and a lot of convincing. The linux-next tree is really convenient as it does not pose a huge risk for users and helps to catch bugs early.

Natrox:
Reddit OP here, I have a version of Linux 5.19 with upstream Zstd. If you need help with testing, integration, anything at all - feel free to ask. I can create a preliminary patch on `linux-next` if it helps you.

felixhandte:
@kdave thanks for flagging this.

@Natrox, awesome!

@terrelln is the relevant authority for this stuff, so we should defer to him. He's out on vacation at the moment, but I'll make sure he sees this next week.

terrelln:
We are planning the 1.5.4 zstd release later this year. I will plan on updating the kernel with that release.

However, we don't expect too much more development before the release. So I will put up a patch to `linux-next` for testing with the current `dev` branch in the next 2 weeks. That will allow us to work out any kinks before the 1.5.4 release.

Natrox:
@terrelln Thank you for your confirmation! Is there any way I can assist with pre-release testing when the time comes?
Please let me know, I would be happy to run tests whenever necessary.

nolange:
> We are planning the 1.5.4 zstd release later this year. I will plan on updating the kernel with that release.

Did 1.5.3 happen, or do you have the convention that uneven means unstable?

Natrox:
> > We are planning the 1.5.4 zstd release later this year. I will plan on updating the kernel with that release.
> 
> Did 1.5.3 happen, or do you have the convention that uneven means unstable?

Sorry to chime in, but I saw it so I figured I would give you a little info in case you did not know.
[...]

PR #3295, opened 2022-10-20
[lazy] Use switch instead of indirect function calls.

terrelln:
Use a switch statement to select the search function instead of an indirect function call. This results in a sizable performance win.

This PR is a modification of the approach taken in PR #2828. When I measured performance for that commit, it was neutral. However, I now see a performance regression on gcc, but still neutral on clang. I'm measuring on the same platform, but with newer compilers. The new approach beats both the current dev branch and the baseline before PR #2828 was merged.

This PR is necessary for Issue #3275, to update zstd in the kernel. Without this PR there is a large regression in greedy - btlazy2 compression speed. With this PR it is about neutral.

gcc version: 12.2.0
clang version: 14.0.6
dataset: silesia.tar

| Compiler | Level | Dev Speed (MB/s) | PR Speed (MB/s) | Delta  |
|----------|-------|------------------|-----------------|--------|
| gcc      |     5 |            102.6 |           113.4 | +10.5% |
| gcc      |     7 |             66.6 |            71.7 | +7.6% |
| gcc      |     9 |             51.5 |            56.6 | +9.9% |
| gcc      |    13 |             14.3 |            14.3 |  +0.0% |
| clang    |     5 |            108.1 |           117.8 |  +8.9% |
| clang    |     7 |             68.5 |            75.0 |  +9.4% |
| clang    |     9 |             53.2 |            58.1 |  +9.2% |
| clang    |    13 |             14.3 |            14.8 |  +3.4% |

The binary size stays just about the same for clang and gcc, measured using the `size` command:

| Compiler | Branch | Text    | Data | BSS | Total   |
|----------|--------|---------|------|-----|---------|
| gcc      | dev    | 1127950 | 3312 | 280 | 1131542 |
| gcc      | PR     | 1133134 | 2512 | 280 | 1135926 |
| clang    | dev    | 1046254 | 3256 | 216 | 1049726 |
| clang    | PR     | 1048486 | 2296 | 216 | 1050998 |

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -165,6 +165,12 @@
 #define UNLIKELY(x) (x)
 #endif
 
+#if __has_builtin(__builtin_unreachable) || (defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5)))
+#  define ZSTD_UNREACHABLE { assert(0), __builtin_unreachable(); }
+#else
+#  define ZSTD_UNREACHABLE { assert(0); }
+#endif
+
 /* disable warnings */
 #ifdef _MSC_VER    /* Visual Studio */
 #  include <intrin.h>                    /* For Visual 2005 */
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1317,14 +1317,10 @@ size_t ZSTD_RowFindBestMatch(
 }
 
 
-typedef size_t (*searchMax_f)(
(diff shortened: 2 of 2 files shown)


terrelln:
Not quite sure whats going on with the VS compiler errors, still have to debug.

Cyan4973:
So benefits are not limited to kernel land, they also translate into faster speed in user land ?
Impressive.

terrelln:
> So benefits are not limited to kernel land, they also translate into faster speed in user land ?

Yup!

Cyan4973 (approved):

PR #3304, opened 2022-10-30
Rewrite check *bufStart condition

GermanAizek:
Rewriting logic `bufStart` pointer validation condition and micro-optimization with `strlen()` function. For `DYNAMIC_BMI2` macro, it was meant to use a function with the `*_bmi2` postfix. After review PR code, feedback about it.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -693,8 +693,11 @@ static int UTIL_prepareFileList(const char *dirName,
                 ptrdiff_t newListSize = (*bufEnd - *bufStart) + LIST_SIZE_INCREASE;
                 assert(newListSize >= 0);
                 *bufStart = (char*)UTIL_realloc(*bufStart, (size_t)newListSize);
-                *bufEnd = *bufStart + newListSize;
-                if (*bufStart == NULL) { free(path); closedir(dir); return 0; }
+                if (*bufStart != NULL) {
+                    *bufEnd = *bufStart + newListSize;
+                } else {
+                    free(path); closedir(dir); return 0;
+                }
             }
             if (*bufStart + *pos + pathLength < *bufEnd) {
                 memcpy(*bufStart + *pos, path, pathLength + 1);  /* with final \0 */


lib/common/entropy_common.c
```diff
@@ -200,7 +200,7 @@ BMI2_TARGET_ATTRIBUTE static size_t FSE_readNCount_body_bmi2(
         short* normalizedCounter, unsigned* maxSVPtr, unsigned* tableLogPtr,
         const void* headerBuffer, size_t hbSize)
 {
-    return FSE_readNCount_body(normalizedCounter, maxSVPtr, tableLogPtr, headerBuffer, hbSize);
+    return FSE_readNCount_body_bmi2(normalizedCounter, maxSVPtr, tableLogPtr, headerBuffer, hbSize);
```
embg: The original code here is actually correct :) `BMI2_TARGET_ATTRIBUTE` tells the compiler to use BMI2 instructions when compiling this function. But the source code of the function doesn't change.

The reason you're seeing failures in CI is that you're defining `FSE_readNCount_body_bmi2()` in terms of itself, i.e. you have an infinite recursion.

programs/util.c
```diff
@@ -386,7 +386,7 @@ static size_t readLineFromFile(char* buf, size_t len, FILE* file)
     assert(!feof(file));
     if ( fgets(buf, (int) len, file) == NULL ) return 0;
     {   size_t linelen = strlen(buf);
-        if (strlen(buf)==0) return 0;
+        if (buf[0]=='\0') return 0;
```
embg: I think we can rely on the compiler to make this optimization, so I prefer the original code here since it's more readable.

programs/util.c
```diff
@@ -693,8 +693,8 @@ static int UTIL_prepareFileList(const char *dirName,
                 ptrdiff_t newListSize = (*bufEnd - *bufStart) + LIST_SIZE_INCREASE;
                 assert(newListSize >= 0);
                 *bufStart = (char*)UTIL_realloc(*bufStart, (size_t)newListSize);
-                *bufEnd = *bufStart + newListSize;
-                if (*bufStart == NULL) { free(path); closedir(dir); return 0; }
+                if (*bufStart != NULL) *bufEnd = *bufStart + newListSize;
+                else { free(path); closedir(dir); return 0; }
```
embg: I think this is a useful change since performing arithmetic on a NULL pointer is UB. Can you please add curly braces around `*bufEnd = *bufStart + newListSize;`?

programs/zstdcli.c
```diff
@@ -130,8 +130,9 @@ static void checkLibVersion(void)
    */
 static int exeNameMatch(const char* exeName, const char* test)
 {
-    return !strncmp(exeName, test, strlen(test)) &&
-        (exeName[strlen(test)] == '\0' || exeName[strlen(test)] == '.');
+    size_t len = strlen(test);
+    return !strncmp(exeName, test, len) &&
+        (exeName[len] == '\0' || exeName[len] == '.');
```
embg: The compiler will take care of this optimization, so I don't think we need to change this code.

programs/zstdcli.c
```diff
@@ -1037,7 +1038,7 @@ int main(int argCount, const char* argv[])
                 if (longCommandWArg(&argument, "--size-hint")) { NEXT_TSIZE(srcSizeHint); continue; }
                 if (longCommandWArg(&argument, "--output-dir-flat")) {
                     NEXT_FIELD(outDirName);
-                    if (strlen(outDirName) == 0) {
+                    if (outDirName[0] == '\0') {
```
embg: Same as above :)

programs/zstdcli.c
```diff
@@ -1053,7 +1054,7 @@ int main(int argCount, const char* argv[])
 #ifdef UTIL_HAS_MIRRORFILELIST
                 if (longCommandWArg(&argument, "--output-dir-mirror")) {
                     NEXT_FIELD(outMirroredDirName);
-                    if (strlen(outMirroredDirName) == 0) {
+                    if (outMirroredDirName[0] == '\0') {
```
embg: Same as above :)

embg (changes requested):
Thanks for the PR! I like your change related to `*bufStart == NULL`, but for the other changes I prefer the original code. In many cases the compiler will take care of the optimizations you're doing manually.

If you revert those other changes, happy to merge!

embg:
Here's a godbolt link showing that gcc won't actually call `strlen` for `strlen(str) > 0`: https://godbolt.org/z/zerrbcorr

GermanAizek:
> Here's a godbolt link showing that gcc won't actually call `strlen` for `strlen(str) > 0`: https://godbolt.org/z/zerrbcorr

I followed link and made sure that there was no optimization, and readability code suffered. I will revert changes and leave only with NULL checking.

GermanAizek:
@embg, check this commit https://github.com/facebook/zstd/commit/1b7f3c4dc732e5d0569b330deb84a0a6e4cb416d

embg:
Looks good, you need to fix two things:
* Sign the CLA: https://code.facebook.com/cla/individual
* Add curly braces around both branches of the if statement, like this:
```
if (*bufStart != NULL) {
    *bufEnd = *bufStart + newListSize;
} else {
    free(path); closedir(dir); return 0;
}
```

GermanAizek:
@embg, everything is ready for PR merge.

embg (approved):

embg:
Thanks for the contribution! Will merge after CI passes.

PR #3311, opened 2022-11-04
Move cmake files to the project top level folder

bsergean:
I know this will require some serious discussion as this makes cmake 'more official', but this helps me with using [CPM](https://github.com/cpm-cmake/CPM.cmake), which is a cmake package manager, to build zstd.


eli-schwartz:
Please move the meson files to the top level folder as well. This would help me with Meson, which has a built-in package manager that requires the meson files to be at the top level: https://mesonbuild.com/Wrap-dependency-system-manual.html

bsergean:
@eli-schwartz maybe you could make a PR in my fork, I'm not sure on how to test meson. Is it just one file ?

Also I wonder if this will create more friction to do so, I don't have a strong opinion either way, but it's more of a decision for the zstd maintainers as they will have to deal with that (cmake + meson) in the long run.

Tachi107:
This would make it look like CMake is officially supported, even if it's not. If the files get moved, maybe it would make sense to add a warning in the CMake script?

bsergean:
The protocol buffer library did something similar recently, their cmake files were nested under a sub-folder, and they moved it to the top folder. But they also implemented what you suggest to keep things stable. They kept a tiny CMake file at the old location (nested) that print a warning(“this file has been moved”), and ‘include/impoirt the file from the new location.

eli-schwartz:
> @eli-schwartz maybe you could make a PR in my fork, I'm not sure on how to test meson. Is it just one file ?

done

bsergean:
Andrea, for what it’s worth, could we say something in the doc or in the ‘other’ build systems (cmake + meson) that those aren’t ‘official … even if they are present in the top folder ?

Eli, just merged your changes.

Cyan4973:
> The protocol buffer library did something similar recently, their cmake files were nested under a sub-folder, and they moved it to the top folder. But they also implemented what you suggest to keep things stable. They kept a tiny CMake file at the old location (nested) that print a warning(“this file has been moved”), and ‘include/impoirt the file from the new location.

Could it be done the other way around ?
Aka a small "shim" `cmake` file at root that just redirects to the "real" one within `build/cmake` ?

bsergean:
Probably, hopefully. I'll give it a try when I get a chance.

In protobuf they use an include with a .. to reference the parent directory, so possibly we could do the include the other way around.

https://github.com/protocolbuffers/protobuf/blob/main/cmake/CMakeLists.txt

Cyan4973:
Is this topic still active ?

Providing a small `cmake` file at root that redirects to the project within `build/cmake/` is a valid option.

bsergean:
Sorry, I'm still very interested in this ; we use a fork in our buildsystem that uses the code from the PR, and using forks is bad long term. However my priorities have changed and I need to focus on something else ...

bsergean:
I just tried to use this small cmake that include the other one, in the top folder:

```
cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
project(zstd)
include(build/cmake/CMakeLists.txt)
```

But I'm getting errors about cmake helper functions which can't be found:

```
CMake Error at build/cmake/CMakeLists.txt:32 (include):
  include could not find requested file:

    GetZstdLibraryVersion
Call Stack (most recent call first):
  CMakeLists.txt:5 (include)


CMake Error at build/cmake/CMakeLists.txt:33 (GetZstdLibraryVersion):
  Unknown CMake command "GetZstdLibraryVersion".
Call Stack (most recent call first):
  CMakeLists.txt:5 (include)
```

Cyan4973:
It seems that the project of providing a small "shim" `cmake` file at top level folder, redirecting to `build/cmake/`, has encountered difficulties. Shall we close it, or is there still hope that it can happen ?

bsergean:
Sorry I haven’t had time to look into this. When I tried it wasn’t obvious, but to be honest I didn’t try very hard. Someone more cmake aware might know better if this is feasible though. Maybe that imaginary person will find this ticket one day and comment here.

Feel free to close or keep open, I know what we need to do now, so if I ever find the need for this again I will start with the ’shim’ option.

PR #3320, opened 2022-11-18
Fix for MSVC C4267 warning on ARM64 (which becomes error C2220 with /WX)

cwoffenden:
I've only seen this on the ARM64 toolchain under the new ARM64 version of VS 2022 (running on a shiny new Windows Dev Kit 2023), and didn't see it cross-compiling for ARM64 from x64:
```
warning C4267: '-=': conversion from 'size_t' to 'int', possible loss of data
warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
```
The change is minimal: `i` can't be a `size_t` because it needs to be signed, but changeing `chunkSize` to `int` is fine (it's only ever going to be 32 or 64 (or possibly somewhere between 16 to 128 on the more exotic systems).

(I can't verify the warning on x64 to ARM64 until next week but it's such a minor unintrusive change with known values)

### DIFF
--- a/lib/compress/zstd_lazy.c
+++ b/lib/compress/zstd_lazy.c
@@ -1078,7 +1078,7 @@ ZSTD_row_getMatchMask(const BYTE* const tagRow, const BYTE tag, const U32 headGr
     }
 # endif /* ZSTD_ARCH_ARM_NEON */
     /* SWAR */
-    {   const size_t chunkSize = sizeof(size_t);
+    {   const int chunkSize = sizeof(size_t);
         const size_t shiftAmount = ((chunkSize * 8) - chunkSize);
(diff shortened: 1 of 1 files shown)


cwoffenden:
Just to add, with the latest VS 2022, x86 host cross compiling to Windows ARM64 has the same warning without this fix.

Cyan4973:
cc @embg :
do we have ways to force compilation of the row matchfinder with the SWAR code path ?
If yes, is this code path already tested in CI ?

aqrit:
Fuzzer test with no intrinsics https://github.com/facebook/zstd/pull/2678

shakeebshams (commented):
LGTM

Cyan4973 (approved):

PR #3327, opened 2022-12-01
meson: get version up front

eli-schwartz:
Run the scraper command to establish the project version immediately, rather than wait for the build to be configured. This simplifies the code and ensures that project introspection works correctly.


Cyan4973 (approved):

Cyan4973:
I like it. It indeed feels more logical.

PR #3333, opened 2022-12-07
Block-Level Sequence Producer API

embg:
This PR introduces an API for external block-level sequence producers to plug into zstd. The user provides a function pointer and state object for the external sequence producer, and zstd will call it to generate sequences for each block. Entropy compression of sequences still remains entirely within the library's internal functions.

Potential applications of the API include hardware-accelerated sequence producers and sequence producers specialized to particular types of data.

There are some subtleties around fallback, sequence validation, memory ownership, etc. Users should read all of the documentation added by this PR to zstd.h before using the API. Note: that documentation has been updated in subsequent PRs, so make sure to look at a recent commit.

An example program is provided (see contrib/externalSequenceProducer) which demonstrates how to use the API with a simple LZ parser.

---
Note: the original version of this PR used the term "External Matchfinder API". The above summary has been updated to use the new term "Block-Level Sequence Producer API", but the code in this PR still uses old symbol names. Updated symbol names were introduced to the code in https://github.com/facebook/zstd/pull/3484.

### DIFF
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-present, Yann Collet, Facebook, Inc.
+ * Copyright (c) Yann Collet, Facebook, Inc.
  * All rights reserved.
  *
  * This source code is licensed under both the BSD-style license (found in the
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -30,7 +30,7 @@ int main(int argc, char *argv[]) {
     int simpleExternalMatchState = 0xdeadbeef;
 
     // Here is the crucial bit of code!
-    ZSTD_refExternalMatchFinder(
+    ZSTD_registerExternalMatchFinder(
         zc,
         &simpleExternalMatchState,
         simpleExternalMatchFinder
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -5,7 +5,7 @@
 #define ZSTD_STATIC_LINKING_ONLY
 #include "zstd.h"
 #include "zstd_errors.h"
-#include "matchfinder.h" // simpleExternalMatchFinder, simpleExternalMatchStateDestructor
+#include "matchfinder.h" // simpleExternalMatchFinder
 
 int main(int argc, char *argv[]) {
     size_t res;
@@ -35,17 +35,17 @@ int main(int argc, char *argv[]) {
 
     FILE *f = fopen(argv[1], "rb");
     fseek(f, 0, SEEK_END);
-    long srcSize = ftell(f);
+    long const srcSize = ftell(f);
     fseek(f, 0, SEEK_SET);
 
     char *src = malloc(srcSize + 1);
     fread(src, srcSize, 1, f);
     fclose(f);
 
-    size_t dstSize = ZSTD_compressBound(srcSize);
+    size_t const dstSize = ZSTD_compressBound(srcSize);
     char *dst = malloc(dstSize);
 
-    size_t cSize = ZSTD_compress2(zc, dst, dstSize, src, srcSize);
+    size_t const cSize = ZSTD_compress2(zc, dst, dstSize, src, srcSize);
 
     if (ZSTD_isError(cSize)) {
         printf("ERROR: %s\n", ZSTD_getErrorName(cSize));
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -32,6 +32,13 @@ int main(int argc, char *argv[]) {
         return 1;
     }
 
+    res = ZSTD_CCtx_setParameter(zc, ZSTD_c_useExternalMatchfinder, 1);
+
+    if (ZSTD_isError(res)) {
+        printf("ERROR: %s\n", ZSTD_getErrorString(res));
+        return 1;
+    }
+
     FILE *f = fopen(argv[1], "rb");
     fseek(f, 0, SEEK_END);
     long srcSize = ftell(f);
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -17,7 +17,7 @@ int main(int argc, char *argv[]) {
 
     ZSTD_CCtx* zc = ZSTD_createCCtx();
 
-    int simpleExternalMatchState = 0xdeadbeef; // @nocommit
+    int simpleExternalMatchState = 0xdeadbeef;
 
     // Here is the crucial bit of code!
     ZSTD_refExternalMatchFinder(
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -26,7 +26,7 @@ int main(int argc, char *argv[]) {
         simpleExternalMatchFinder
     );
 
-    res = ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchfinderFallback, 1);
+    res = ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchFinderFallback, 1);
 
     if (ZSTD_isError(res)) {
         printf("ERROR: %s\n", ZSTD_getErrorName(res));
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -1,3 +1,13 @@
+/*
+ * Copyright (c) 2018-present, Yann Collet, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -19,10 +19,12 @@ int main(int argc, char *argv[]) {
 
     ZSTD_CCtx* zc = ZSTD_createCCtx();
 
+    int simpleExternalMatchState = 0xdeadbeef; // @nocommit
+
     // Here is the crucial bit of code!
     res = ZSTD_registerExternalMatchFinder(
         zc,
-        NULL,
+        &simpleExternalMatchState,
         simpleExternalMatchFinder,
         simpleExternalMatchStateDestructor
     );
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -1,0 +1,64 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define ZSTD_STATIC_LINKING_ONLY
+#include "zstd.h"
+#include "matchfinder.h" // simpleExternalMatchfinder
+
+const size_t ZSTD_LEVEL = 1;
+
+int main(int argc, char *argv[]) {
+    ZSTD_CCtx* zc = ZSTD_createCCtx();
+
+    // Here is the crucial bit of code!
+    ZSTD_registerExternalMatchfinder(zc, simpleExternalMatchfinder);
+
+    if (argc != 2) {
+        printf("Usage: exampleMatchfinder <file>\n");
+        return 1;
+    }
+
+    FILE *f = fopen(argv[1], "rb");
+    fseek(f, 0, SEEK_END);
+    long srcSize = ftell(f);
+    fseek(f, 0, SEEK_SET);
+
+    char *src = malloc(srcSize + 1);
+    fread(src, srcSize, 1, f);
+    fclose(f);
+
+    size_t dstSize = ZSTD_compressBound(srcSize);
+    char *dst = malloc(dstSize);
+
+    size_t cSize = ZSTD_compress2(zc, dst, dstSize, src, srcSize);
+
+    if (ZSTD_isError(cSize)) {
+        printf("ERROR: %lu\n", cSize);
+        return 1;
+    }
+
+    char *val = malloc(srcSize);
+    size_t res = ZSTD_decompress(val, srcSize, dst, cSize);
+
+    if (ZSTD_isError(res)) {
+        printf("ERROR: %lu\n", cSize);
+        return 1;
+    }
+
+    if (memcmp(src, val, srcSize) == 0) {
+        printf("Compression and decompression were successful!\n");
+        printf("Original size: %lu\n", srcSize);
+        printf("Compressed size: %lu\n", cSize);
+        return 0;
+    } else {
+        printf("ERROR: input and validation buffers don't match!\n");
+        for (int i = 0; i < srcSize; i++) {
+            if (src[i] != val[i]) {
+                printf("First bad index: %d\n", i);
+                break;
+            }
+        }
+        return 1;
+    }
+}
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -20,24 +20,11 @@ int main(int argc, char *argv[]) {
     int simpleExternalMatchState = 0xdeadbeef; // @nocommit
 
     // Here is the crucial bit of code!
-    res = ZSTD_registerExternalMatchFinder(
+    ZSTD_refExternalMatchFinder(
         zc,
         &simpleExternalMatchState,
-        simpleExternalMatchFinder,
-        simpleExternalMatchStateDestructor
+        simpleExternalMatchFinder
     );
-
-    if (ZSTD_isError(res)) {
-        printf("ERROR: %s\n", ZSTD_getErrorName(res));
-        return 1;
-    }
-
-    res = ZSTD_CCtx_setParameter(zc, ZSTD_c_useExternalMatchfinder, 1);
-
-    if (ZSTD_isError(res)) {
-        printf("ERROR: %s\n", ZSTD_getErrorName(res));
-        return 1;
-    }
 
     res = ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchfinderFallback, 1);
 
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -41,6 +41,13 @@ int main(int argc, char *argv[]) {
         return 1;
     }
 
+    res = ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchfinderFallback, 1);
+
+    if (ZSTD_isError(res)) {
+        printf("ERROR: %s\n", ZSTD_getErrorName(res));
+        return 1;
+    }
+
     FILE *f = fopen(argv[1], "rb");
     fseek(f, 0, SEEK_END);
     long srcSize = ftell(f);
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -4,7 +4,8 @@
 
 #define ZSTD_STATIC_LINKING_ONLY
 #include "zstd.h"
-#include "matchfinder.h" // simpleExternalMatchfinder
+#include "zstd_errors.h"
+#include "matchfinder.h" // simpleExternalMatchFinder, simpleExternalMatchStateDestructor
 
 const size_t ZSTD_LEVEL = 1;
 
@@ -12,7 +13,12 @@ int main(int argc, char *argv[]) {
     ZSTD_CCtx* zc = ZSTD_createCCtx();
 
     // Here is the crucial bit of code!
-    ZSTD_registerExternalMatchfinder(zc, simpleExternalMatchfinder);
+    ZSTD_registerExternalMatchfinder(
+        zc,
+        NULL,
+        simpleExternalMatchFinder,
+        simpleExternalMatchStateDestructor
+    );
 
     if (argc != 2) {
         printf("Usage: exampleMatchfinder <file>\n");
@@ -34,7 +40,7 @@ int main(int argc, char *argv[]) {
     size_t cSize = ZSTD_compress2(zc, dst, dstSize, src, srcSize);
 
     if (ZSTD_isError(cSize)) {
-        printf("ERROR: %lu\n", cSize);
+        printf("ERROR: %s\n", ZSTD_getErrorString(cSize));
         return 1;
     }
 
@@ -42,7 +48,7 @@ int main(int argc, char *argv[]) {
     size_t res = ZSTD_decompress(val, srcSize, dst, cSize);
 
     if (ZSTD_isError(res)) {
-        printf("ERROR: %lu\n", cSize);
+        printf("ERROR: %s\n", ZSTD_getErrorString(res));
         return 1;
     }
 
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -28,14 +28,14 @@ int main(int argc, char *argv[]) {
     );
 
     if (ZSTD_isError(res)) {
-        printf("ERROR: %s\n", ZSTD_getErrorString(res));
+        printf("ERROR: %s\n", ZSTD_getErrorName(res));
         return 1;
     }
 
     res = ZSTD_CCtx_setParameter(zc, ZSTD_c_useExternalMatchfinder, 1);
 
     if (ZSTD_isError(res)) {
-        printf("ERROR: %s\n", ZSTD_getErrorString(res));
+        printf("ERROR: %s\n", ZSTD_getErrorName(res));
         return 1;
     }
 
@@ -54,7 +54,7 @@ int main(int argc, char *argv[]) {
     size_t cSize = ZSTD_compress2(zc, dst, dstSize, src, srcSize);
 
     if (ZSTD_isError(cSize)) {
-        printf("ERROR: %s\n", ZSTD_getErrorString(cSize));
+        printf("ERROR: %s\n", ZSTD_getErrorName(cSize));
         return 1;
     }
 
@@ -64,7 +64,7 @@ int main(int argc, char *argv[]) {
     ZSTD_freeCCtx(zc);
 
     if (ZSTD_isError(res)) {
-        printf("ERROR: %s\n", ZSTD_getErrorString(res));
+        printf("ERROR: %s\n", ZSTD_getErrorName(res));
         return 1;
     }
 
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -7,8 +7,6 @@
 #include "zstd_errors.h"
 #include "matchfinder.h" // simpleExternalMatchFinder, simpleExternalMatchStateDestructor
 
-const size_t ZSTD_LEVEL = 1;
-
 int main(int argc, char *argv[]) {
     size_t res;
 
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -10,18 +10,25 @@
 const size_t ZSTD_LEVEL = 1;
 
 int main(int argc, char *argv[]) {
+    size_t res;
+
+    if (argc != 2) {
+        printf("Usage: exampleMatchfinder <file>\n");
+        return 1;
+    }
+
     ZSTD_CCtx* zc = ZSTD_createCCtx();
 
     // Here is the crucial bit of code!
-    ZSTD_registerExternalMatchfinder(
+    res = ZSTD_registerExternalMatchFinder(
         zc,
         NULL,
         simpleExternalMatchFinder,
         simpleExternalMatchStateDestructor
     );
 
-    if (argc != 2) {
-        printf("Usage: exampleMatchfinder <file>\n");
+    if (ZSTD_isError(res)) {
+        printf("ERROR: %s\n", ZSTD_getErrorString(res));
         return 1;
     }
 
@@ -45,7 +52,9 @@ int main(int argc, char *argv[]) {
     }
 
     char *val = malloc(srcSize);
-    size_t res = ZSTD_decompress(val, srcSize, dst, cSize);
+    res = ZSTD_decompress(val, srcSize, dst, cSize);
+
+    ZSTD_freeCCtx(zc);
 
     if (ZSTD_isError(res)) {
         printf("ERROR: %s\n", ZSTD_getErrorString(res));
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -50,13 +50,23 @@ int main(int argc, char *argv[]) {
 
     FILE *f = fopen(argv[1], "rb");
     assert(f);
-    fseek(f, 0, SEEK_END);
-    long const srcSize = ftell(f);
-    fseek(f, 0, SEEK_SET);
+    {
+        int const ret = fseek(f, 0, SEEK_END);
+        assert(ret == 0);
+    }
+    size_t const srcSize = ftell(f);
+    {
+        int const ret = fseek(f, 0, SEEK_SET);
+        assert(ret == 0);
+    }
 
     char* const src = malloc(srcSize + 1);
-    fread(src, srcSize, 1, f);
-    fclose(f);
+    {
+        size_t const ret = fread(src, srcSize, 1, f);
+        assert(ret == 1);
+        int const ret2 = fclose(f);
+        assert(ret2 == 0);
+    }
 
     size_t const dstSize = ZSTD_compressBound(srcSize);
     char* const dst = malloc(dstSize);
@@ -78,9 +88,9 @@ int main(int argc, char *argv[]) {
         printf("Compressed size: %lu\n", cSize);
     } else {
         printf("ERROR: input and validation buffers don't match!\n");
-        for (int i = 0; i < srcSize; i++) {
+        for (size_t i = 0; i < srcSize; i++) {
             if (src[i] != val[i]) {
-                printf("First bad index: %d\n", i);
+                printf("First bad index: %zu\n", i);
                 break;
             }
         }
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -19,7 +19,7 @@
 #include "matchfinder.h" // simpleExternalMatchFinder
 
 #define CHECK(res)                                      \
-do  {                                                   \
+do {                                                    \
     if (ZSTD_isError(res)) {                            \
         printf("ERROR: %s\n", ZSTD_getErrorName(res));  \
         return 1;                                       \
@@ -76,7 +76,6 @@ int main(int argc, char *argv[]) {
         printf("Compression and decompression were successful!\n");
         printf("Original size: %lu\n", srcSize);
         printf("Compressed size: %lu\n", cSize);
-        return 0;
     } else {
         printf("ERROR: input and validation buffers don't match!\n");
         for (int i = 0; i < srcSize; i++) {
@@ -89,4 +88,8 @@ int main(int argc, char *argv[]) {
     }
 
     ZSTD_freeCCtx(zc);
+    free(src);
+    free(dst);
+    free(val);
+    return 0;
 }
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -61,6 +61,7 @@ int main(int argc, char *argv[]) {
     }
 
     char* const src = malloc(srcSize + 1);
+    assert(src);
     {
         size_t const ret = fread(src, srcSize, 1, f);
         assert(ret == 1);
@@ -70,6 +71,7 @@ int main(int argc, char *argv[]) {
 
     size_t const dstSize = ZSTD_compressBound(srcSize);
     char* const dst = malloc(dstSize);
+    assert(dst);
 
     size_t const cSize = ZSTD_compress2(zc, dst, dstSize, src, srcSize);
     CHECK(cSize);
--- a/contrib/externalMatchfinder/main.c
+++ b/contrib/externalMatchfinder/main.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) Yann Collet, Facebook, Inc.
+ * Copyright (c) Yann Collet, Meta Platforms, Inc.
  * All rights reserved.
  *
  * This source code is licensed under both the BSD-style license (found in the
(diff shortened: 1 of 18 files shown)


contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) Yann Collet, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
```
Cyan4973, line 10: Could you add a short description of what this file is about, what are its objectives ?

contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,90 @@
...
+int main(int argc, char *argv[]) {
+    size_t res;
+
+    if (argc != 2) {
+        printf("Usage: exampleMatchfinder <file>\n");
+        return 1;
+    }
+
+    ZSTD_CCtx* zc = ZSTD_createCCtx();
```
Cyan4973: note: 
declaring variables on the go is not `C90` compatible.
This is not necessarily a pb, `contrib/` is allowed to ship non-C90 code (in contrast to `lib/`),
just a point to pay attention to in the build system.

lib/compress/zstd_compress.c
```diff
@@ -2928,10 +3028,57 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
-        } else {   /* not long range mode */
+        } else if (zc->appliedParams.useExternalMatchFinder) {
+            assert(
+                zc->externalMatchCtx.seqBufferCapacity >= ZSTD_sequenceBound(srcSize)
+            );
+            assert(zc->externalMatchCtx.mFinder != NULL);
+
+            {   size_t const nbExternalSeqs = (zc->externalMatchCtx.mFinder)(
+                    zc->externalMatchCtx.mState,
+                    zc->externalMatchCtx.seqBuffer,
```
GarenJian-Intel, line 3067: Could you memset all values of array zc->externalMatchCtx.seqBuffer to 0 before passing it to matchfinder?

embg: This is difficult because normally the matchfinder will only use a tiny fraction of the seqBuffer. So there could be a large performance cost if we memset the entire buffer.

Why do you need to memset to zero?

contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,90 @@
...
+    int simpleExternalMatchState = 0xdeadbeef;
+
+    // Here is the crucial bit of code!
+    ZSTD_refExternalMatchFinder(
+        zc,
+        &simpleExternalMatchState,
+        simpleExternalMatchFinder
+    );
+
+    res = ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchFinderFallback, 1);
```
Cyan4973: minor : coding style

`res` is only used immediately after, to test its value, and output an error if need be.

It's unrelated to the later set `res` at decompression time.

Prefer to limit variable's lifetime,
this makes it more clear that the value set in `res` is not meant to be-reused later on.
This can avoid future maintenance issues,
such as forgetting to set `res` later (like at decompression time), but still test it, 
misleadingly believing it represents the result of the most recent action.

The easiest way to reduce `res` lifetime is to create a new `{ }` block scope.

Marginally more complex, you may encapsulate the whole "test, write error and return" logic in a macro,
higher investment but this may even make the code's logic clearer.

contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,90 @@
...
+
+    if (ZSTD_isError(res)) {
+        printf("ERROR: %s\n", ZSTD_getErrorName(res));
+        return 1;
+    }
+
+    FILE *f = fopen(argv[1], "rb");
```
Cyan4973, line 51: `fopen()` may fail.
Since it's just a test, an `assert()` can be fine enough

contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,90 @@
...
+    fseek(f, 0, SEEK_END);
+    long const srcSize = ftell(f);
+    fseek(f, 0, SEEK_SET);
+
+    char *src = malloc(srcSize + 1);
```
Cyan4973: `malloc()` may fail.

contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,90 @@
...
+
+    if (ZSTD_isError(cSize)) {
+        printf("ERROR: %s\n", ZSTD_getErrorName(cSize));
+        return 1;
+    }
+
+    char *val = malloc(srcSize);
+    res = ZSTD_decompress(val, srcSize, dst, cSize);
+
+    ZSTD_freeCCtx(zc);
```
Cyan4973: minor: it feels strange to free `zc` here, in the middle of decompression.
I would recommend either to free it at end of compression (just 2 lines above),
or at end of function

contrib/externalMatchfinder/matchfinder.c
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) Yann Collet, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
```
Cyan4973, line 10: Definitely worth having a short introduction paragraph to describe what this unit is about (simple external matchfinder for test and illustration purposes).

contrib/externalMatchfinder/matchfinder.c
```diff
@@ -0,0 +1,75 @@
...
+    (void)dictSize;
+    (void)outSeqsCapacity;
+    (void)compressionLevel;
+
+    {   int i;
+        for (i=0; i < HSIZE; i++) {
+            hashTable[i] = BADIDX;
+    }   }
+
+    while (ip + 4 < iend) {
```
Cyan4973: minor: `(ip + MLS < iend)` seems proper.

tests/zstreamtest.c
```diff
@@ -1777,6 +1777,91 @@ static int basicUnitTests(U32 seed, double compressibility)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : External matchfinder API: ", testNb++);
```
Cyan4973, line 1837: 👍

tests/zstreamtest.c
```diff
@@ -1777,6 +1777,91 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+    {
+        size_t const dstBufSize = ZSTD_compressBound(CNBufferSize);
+        BYTE* dstBuf = (BYTE*)malloc(ZSTD_compressBound(dstBufSize));
```
Cyan4973: `malloc()` may fail

tests/zstreamtest.c
```diff
@@ -1777,6 +1777,91 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+                EMF_SMALL_ERROR,
+            };
+            size_t const EMF_numFailureCases = 3;
+
+            /* Test external matchfinder success scenarios */
+            for (testCaseId = 0; testCaseId < EMF_numSuccessCases; testCaseId++) {
+                externalMatchState = EMF_successCases[testCaseId];
+                ZSTD_CCtx_reset(zc, ZSTD_reset_session_only);
+                CHECK_Z(ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchFinderFallback, enableFallback));
+                res = ZSTD_compress2(zc, dstBuf, dstBufSize, CNBuffer, CNBufferSize);
```
Cyan4973, line 1880: same comment regarding `res` lifetime,
limit it to current block.

tests/zstreamtest.c
```diff
@@ -1777,6 +1777,91 @@ static int basicUnitTests(U32 seed, double compressibility)
...
+
+            /* Test compression with external matchfinder + empty src buffer */
+            externalMatchState = EMF_ZERO_SEQS;
+            ZSTD_CCtx_reset(zc, ZSTD_reset_session_only);
+            CHECK_Z(ZSTD_CCtx_setParameter(zc, ZSTD_c_enableMatchFinderFallback, enableFallback));
+            res = ZSTD_compress2(zc, dstBuf, dstBufSize, CNBuffer, 0);
+            CHECK(ZSTD_isError(res), "EMF: Compression error: %s", ZSTD_getErrorName(res));
+            CHECK(ZSTD_decompress(checkBuf, checkBufSize, dstBuf, res) != 0, "EMF: Empty src round trip failed!");
+        }
+
```
Cyan4973, line 1916: Should you have a test for an `RLE` block specifically ?

lib/compress/zstd_compress.c
```diff
@@ -2013,6 +2036,14 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
             zc->ldmState.loadedDictEnd = 0;
         }
 
+        /* reserve space for block-level external sequences */
+        if (params->useExternalMatchFinder) {
+            size_t const maxNbExternalSeq = ZSTD_sequenceBound(ZSTD_BLOCKSIZE_MAX);
```
Cyan4973: optimization : 
`blockSize` is `<= ZSTD_BLOCKSIZE_MAX`.
Prefer employing `blockSize` since it's available,
it will reduce memory budget.
This seems especially useful if the main use case is about small data.

lib/compress/zstd_compress.c
```diff
@@ -1512,6 +1529,11 @@ static size_t ZSTD_estimateCCtxSize_usingCCtxParams_internal(
 
     size_t const cctxSpace = isStatic ? ZSTD_cwksp_alloc_size(sizeof(ZSTD_CCtx)) : 0;
 
+    size_t const maxNbExternalSeq = ZSTD_sequenceBound(ZSTD_BLOCKSIZE_MAX);
```
Cyan4973: Same comment here, regarding `blockSize <= ZSTD_BLOCKSIZE_MAX`, leading to a leaner memory budget.

lib/compress/zstd_compress.c
```diff
@@ -2856,6 +2887,49 @@ void ZSTD_resetSeqStore(seqStore_t* ssPtr)
     ssPtr->longLengthType = ZSTD_llt_none;
 }
 
+/* ZSTD_postProcessExternalMatchFinderResult() :
+ * Validates and post-processes sequences obtained through the external matchfinder API.
```
Cyan4973: doc: could you describe (at high level) what `post-process` means ?

lib/compress/zstd_compress.c
```diff
@@ -2856,6 +2887,49 @@ void ZSTD_resetSeqStore(seqStore_t* ssPtr)
...
+    if (emptySrc) {
+        ZSTD_memset(&outSeqs[0], 0, sizeof(ZSTD_Sequence));
+        return 1;
+    } else {
+        ZSTD_Sequence const lastSeq = outSeqs[nbExternalSeqs - 1];
+
+        /* Check if lastSeq is a block delimiter, append one if not */
+        if (lastSeq.offset == 0 && lastSeq.matchLength == 0) {
+            return nbExternalSeqs;
+        } else {
```
Cyan4973: minor : indentation 
Since previous statement is an early exit,
you don't even need an `else { ... }` block here.

lib/compress/zstd_compress.c
```diff
@@ -2856,6 +2887,49 @@ void ZSTD_resetSeqStore(seqStore_t* ssPtr)
...
+    RETURN_ERROR_IF(
+        nbExternalSeqs == 0 && !emptySrc,
+        externalMatchFinder_failed,
+        "External matchfinder produced zero sequences for a non-empty src buffer!"
+    );
+
+    if (emptySrc) {
+        ZSTD_memset(&outSeqs[0], 0, sizeof(ZSTD_Sequence));
+        return 1;
+    } else {
```
Cyan4973: same here, no need for `else`,
this is compulsory course of action after passing previous early exit.

However, since you need to create a variable, you won't save an indentation level here.

lib/compress/zstd_compress.c
```diff
@@ -2856,6 +2887,49 @@ void ZSTD_resetSeqStore(seqStore_t* ssPtr)
...
+        "External matchfinder returned error code %lu",
+        (unsigned long)nbExternalSeqs
+    );
+
+    RETURN_ERROR_IF(
+        nbExternalSeqs == 0 && !emptySrc,
+        externalMatchFinder_failed,
+        "External matchfinder produced zero sequences for a non-empty src buffer!"
+    );
+
```
Cyan4973, line 2940: This method seems about validating that the output of the external matchfinder makes sense.

Another potential test that seems to make sense here
is to check that `sum(matchlen) + sum(litlen) == srcSize;`.
This is a heavy test, so that's only something we would want in debug mode, aka `assert()`.

embg: How can I gate the test so it only runs in debug mode? Is there an ifdef?

embg: Looks like this is the idiom?

```
#if defined(DEBUGLEVEL) && (DEBUGLEVEL>=2)
// code
#endif
```

Cyan4973: Yes, that method works.

Another method is to rely on DCE (Dead Code Elimination),
and you could have schematically something like `assert(test());`

For this case, you could write almost exactly `assert( sum(matchlen) + sum(litlen) == srcSize )`
or even `assert( sum(seqLen) == srcSize )`
provided you could provide the associated `sum()`.

lib/compress/zstd_compress.c
```diff
@@ -2903,6 +2977,24 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         }
         if (zc->externSeqStore.pos < zc->externSeqStore.size) {
             assert(zc->appliedParams.ldmParams.enableLdm == ZSTD_ps_disable);
+
```
Cyan4973, line 3015: I'm somewhat lost here.
What's the state when entering this condition ?

From the following action, `ZSTD_ldm_blockCompress()`, it seems that `LDM` is enabled.
But the previous `assert()` seems contradictory ? 
`assert(zc->appliedParams.ldmParams.enableLdm == ZSTD_ps_disable);`

Anyway, the following tests imply that this is not supported if `useExternalMatchFinder` is enabled.

And that's the first source of confusion.
It seems the real cause for error is  `useExternalMatchFinder`, not `useBlockSplitter`.
`useBlockSplitter` is employed to be more specific about the error message.
So I would naturally reverse the test order : `useExternalMatchFinder` first (now we know that it's going to be an error), `useBlockSplitter` second (to select which error message exactly).

It's also unclear to me what happens when a high compression mode is selected : it's implied that `useBlockSplitter` is going to be enabled by default in this case. So does that mean an automatic failure if `useExternalMatchFinder` is enabled too ?

lib/compress/zstd_compress.c
```diff
@@ -2914,6 +3006,14 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
         } else if (zc->appliedParams.ldmParams.enableLdm == ZSTD_ps_enable) {
             rawSeqStore_t ldmSeqStore = kNullRawSeqStore;
```
Cyan4973, line 3034: I'm still lost here.
LDM seems more clearly enabled in this block.
So what was happening in the previous one ?

lib/compress/zstd_compress.c
Cyan4973: `seqBuffer` should be write only, meaning its initial state should not matter, hence not be relied upon.
If you believe you need this condition, could you explain why ?

lib/compress/zstd_compress.c
```diff
@@ -2928,10 +3028,57 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+                    srcSize == 0  /* emptySrc */
+                );
+
+                if (!ZSTD_isError(nbPostProcessedSeqs)) {
+                    ZSTD_sequencePosition seqPos = {0,0,0};
+                    ZSTD_copySequencesToSeqStoreExplicitBlockDelim(
+                        zc, &seqPos, zc->externalMatchCtx.seqBuffer, nbPostProcessedSeqs, src, srcSize
+                    );
+                    ms->ldmSeqStore = NULL;
+                    lastLLSize = 0;
```
Cyan4973: why is `lastLLSize` automatically `= 0` here ?
Shouldn't it depend on the last sequence ?

embg: No, because we require `sum(matchlen) + sum(litlen) == srcSize`

lib/compress/zstd_compress.c
embg: I don't want to run the sum function unless we're in debug mode... and we don't have lambdas in C :) So I expect the `#ifdef` method is easier here.

lib/compress/zstd_compress.c
Cyan4973: > No, because we require `sum(matchlen) + sum(litlen) == srcSize`

OK, you mean, that's the requirement for _the input of_ `ZSTD_copySequencesToSeqStoreExplicitBlockDelim()`, and this method is already taking care of the last "incomplete" sequence (all literals), which is actually named a Block Delimiter in this case, by copying these last literals into the literals portion of the `seqStore`.

That makes sense.
I think it would also help if these properties were mentioned in the documentation of the function, into `zstd_compress_internal.h`.

It then justifies that you want to opt out of a general mechanism which adds the last literals to the `seqStore`, because it was already done inside `ZSTD_copySequencesToSeqStoreExplicitBlockDelim()`,  _in contrast with_ the "regular" software machfinders.
And btw, you could clearly opt out by returning from the function right away, thus making clear that no further parts of the function are applicable to this code path.

lib/compress/zstd_compress.c
embg: I asked about this offline a while back.

@terrelln:
> It is used by multi-threaded LDM
> We generate the LDM sequences externally because it is a serial process, then inject them into the cctx so the normal block finding can happen in parallel.

The discussion a while back also indicated that the block splitter can trigger this branch.

I'm personally not familiar with why this branch exists, so I disabled the external matchfinder API across the board here. If the block-splitter indeed triggers this branch, then the external matchfinder API will fail compression at high compression levels.

We could add some logic during cParam validation to disable block splitting if the external matchfinder is enabled or vice versa. How would you like to proceed?

lib/compress/zstd_compress.c
embg: No idea :) I disabled the external matchfinder in both of the `ZSTD_ldm_blockCompress` branches.

lib/compress/zstd_compress.c
embg: Yup, it's a constraint on the input of ZSTD_copySequencesToSeqStoreExplicitBlockDelim() as well as the output of the external matchfinder function.

embg: My only concern on the early return is that someone might add more code to the end of the function, and forget to remove the early return... Agree on updating the docs though.

lib/compress/zstd_compress.c
```diff
@@ -6527,3 +6659,19 @@ ZSTD_parameters ZSTD_getParams(int compressionLevel, unsigned long long srcSizeH
     if (srcSizeHint == 0) srcSizeHint = ZSTD_CONTENTSIZE_UNKNOWN;
     return ZSTD_getParams_internal(compressionLevel, srcSizeHint, dictSize, ZSTD_cpm_unknown);
 }
+
+void ZSTD_refExternalMatchFinder(
```
daweiq: It seems to me a regsiter function for extenralMatherFinder, Shall we call it ZSTD_regExternalMatchFinder?

Cyan4973: Sounds like a good idea @daweiq.
We could even employ the complete name, `ZSTD_registerExternalMatchFinder()`, for improved clarity.

lib/compress/zstd_compress.c
Cyan4973: Yeah, in current situation, setting an external matchfinder with a high compression level is an automatic compression failure.
This might be acceptable on the short term, but we certainly don't want that to remain "long term", aka it can't remain the situation at release time.

When it comes to having a cleaner situation for release, multiple solutions are possible, and I don't have strong opinion on any of them. 
It starts from making external matchfinder mode compatible with ldm and block splitter, though it certainly requires some work.
A shorter work around could be, as you describe, to automatically change these parameters to "disable" when external matchfinder is set. 
If we go this route, another potential option could be to _ignore_ these parameters when an external matchfinder is set.

Any of these solutions is probably better than an automatic failure when combining high compression levels with external match finder. Each one comes with its own set of trade offs, so there is room for discussion.

lib/compress/zstd_compress.c
embg: Sounds good to me, will change to ZSTD_registerExternalMatchFinder()

tests/zstreamtest.c
embg: On your suggestion, I implemented a [test](https://github.com/embg/zstd/commit/4cd75486cd8a7fd34c3dbb83248abb31dda79543) for RLE handling. It made me realize that we actually can’t have RLE be completely invisible to external matchfinder libraries. The test revealed that even with the mitigations in this PR + additional mitigations, an adversarial matchfinder may compress 10x worse than zstd on a 10MB constant input:

```
test 75 : External matchfinder API: Error => EMF: RLE compressed size doesn't match, 5603 != 339 (seed 3173672436, test nb 76, line 1947) 
```

Context: an external match finder, for example one which inherited DEFLATE’s maximum match length, may produce a large number of sequences for an RLE block. This breaks our [heuristics](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3086) for [detecting](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3869) RLE blocks.

I thought simply disabling that heuristic when `useExternalMatchFinder == 1` is sufficient to handle RLE properly for any external match finder (disabling the heuristic means we always fall back to ZSTD_isRLE(), which doesn’t give false negatives).

**The problem is that zstd [will never emit the first block as RLE](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L3591).** So a match finder which relies entirely on zstd to post-process bad parses of constant blocks will perform really badly on the first block, even if we can post-process into RLE for subsequent blocks.

I wonder if instead of tweaking our heuristics to accommodate those bad parses, we should instead encourage match finder implementors (in zstd.h docs) to do their own detection for constant blocks, and ensure they are producing a small number of sequences for such blocks.

Therefore, I would like to remove anything related to RLE from this patch, except for a note in zstd.h informing implementors that they need to handle it themselves. The existing RLE mitigations in this patch would be great, *if* we didn’t have that exception about the first block of a frame :)

@Cyan4973 thoughts?

embg: Discussed offline -- post-processing bad parses (not just for constant blocks, but in general) is a future topic we want to work on, but it's out of scope for this PR. Post processing should also be gated by a cParam, so the application can control whether sequences are rewritten by zstd.

lib/compress/zstd_compress.c
GarenJian-Intel: No need to set it to zero now. I met a bug in my qatmatchfinder implementation, so I set it to zero. Now I fixed my bug.

lib/compress/zstd_compress.c
embg: I tested, and it turns out this branch is not related to the block splitter: https://gist.github.com/embg/56d42da63eb2eeb86259eeefb0d6b2a4

Therefore, the existing code will not fail at high compression levels due the the block splitter being enabled. Compression will only fail if LDM is enabled (either single-thread or multi-thread LDM).

lib/compress/zstd_compress.c
embg: I discussed this with @terrelln. The previous branch is for multi-thread LDM specifically. It's in a separate branch because LDM is a serial step in multi-thread compression. The process for a child CCtx to integrate the sequences from that serial step is a little different than the process where a single CCtx does both the LDM and normal matchfinding.

The previous branch has `assert(zc->appliedParams.ldmParams.enableLdm == ZSTD_ps_disable);` because we want to make sure the *child* CCtx isn't using the single-thread LDM code path. However, the parent ZSTDMT_CCtx will have LDM enabled when that branch runs.

In terms of this PR, what this all means is that we can safely fail compression in these branches when the external matchfinder is enabled, and it will *only* affect LDM compression. Other params won't be impacted, e.g. block splitter.

lib/compress/zstd_compress.c
Cyan4973: Good news !

lib/compress/zstd_compress.c
embg: It turns out that Sen's code provides for sequence validation (both offsets and matchLens + litLens):
* https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L6004
* https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L6021

So we probably don't want to duplicate that code in `ZSTD_postProcessExternalMatchFinderResult`. I'll just make sure that debug builds have the validateSequences cParam enabled.

embg: Note: given that Sen's validation checks that `sum(matchLens) + sum(litLens) == srcSize` and only reads up to the block delimiter, we don't need any special check for `matchLen == 0` in postProcessSequences.

If the sequences returned by the external matchfinder have an "early" or "accidental" block delimiter which comes before the real end of the sequences, they will fail the summation test (since Sen's code will stop reading early).

contrib/externalMatchfinder/main.c
embg: Added contrib/externalMatchfinder/README.md

contrib/externalMatchfinder/matchfinder.c
embg: Added contrib/externalMatchfinder/README.md

lib/zstd.h
```diff
@@ -2674,6 +2690,120 @@ ZSTDLIB_STATIC_API size_t ZSTD_decompressBlock(ZSTD_DCtx* dctx, void* dst, size_
...
+ *
+ * If the user-provided function returns an error, zstd will either fall back
+ * to an internal matchfinder or fail the compression operation. The user can
+ * choose between the two behaviors by setting the
+ * ZSTD_c_enableMatchFinderFallback cParam. Fallback compression will follow any
+ * other cParam settings, such as compression level, the same as in a normal
+ * compression operation.
+ *
+ * The user shall instruct zstd to use a particular ZSTD_externalMatchFinder_F
+ * function by calling ZSTD_refExternalMatchFinder(cctx, externalMatchState,
```
embg: nit: ref -> register

contrib/externalMatchfinder/main.c
```diff
@@ -0,0 +1,105 @@
...
+    char* const src = malloc(srcSize + 1);
+    {
+        size_t const ret = fread(src, srcSize, 1, f);
+        assert(ret == 1);
+        int const ret2 = fclose(f);
+        assert(ret2 == 0);
+    }
+
+    size_t const dstSize = ZSTD_compressBound(srcSize);
+    char* const dst = malloc(dstSize);
```
Cyan4973, line 73: minor : 
should test that `dst != NULL`.

Note : 
if you are concerned about the nb of places where `malloc()` can fail,
you can have a local `malloc_or_abort()` function
which would automatically abort if `malloc()` fails,
freeing you from thinking about testing `malloc()`'s return value every time.

tests/zstreamtest.c
```diff
@@ -1834,6 +1834,97 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : External matchfinder API: ", testNb++);
+    {
+        size_t const dstBufSize = ZSTD_compressBound(CNBufferSize);
+        BYTE* dstBuf = (BYTE*)malloc(ZSTD_compressBound(dstBufSize));
```
Cyan4973: very minor :  
`BYTE* const dstBuf = `
This statement guarantees that `dstBuf` (as a pointer value) will not be updated after its initialization.

tests/zstreamtest.c
```diff
@@ -1834,6 +1834,97 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+    DISPLAYLEVEL(3, "test%3i : External matchfinder API: ", testNb++);
+    {
```
Cyan4973, line 1838: Coding style : 
It seems you could write this entire unit test into its own function,
and then invoke this newly created test function from here .

I know most of the existing code base in this unit is not written using this style,
but this is probably and cleaner way to proceed, more maintainable.

lib/compress/zstd_compress.c
```diff
@@ -1492,11 +1520,12 @@ static size_t ZSTD_estimateCCtxSize_usingCCtxParams_internal(
...
         const size_t buffInSize,
         const size_t buffOutSize,
-        const U64 pledgedSrcSize)
+        const U64 pledgedSrcSize,
+        int useExternalMatchFinder)
 {
     size_t const windowSize = (size_t) BOUNDED(1ULL, 1ULL << cParams->windowLog, pledgedSrcSize);
     size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, windowSize);
-    U32    const divider = (cParams->minMatch==3) ? 3 : 4;
+    U32    const divider = (cParams->minMatch==3 || useExternalMatchFinder) ? 3 : 4;
```
Cyan4973: minor : the logic employed to process `divider` seems re-used in another function later on.
It seems it would be better to factor it into a named function.

lib/compress/zstd_compress.c
```diff
@@ -1542,7 +1577,7 @@ size_t ZSTD_estimateCCtxSize_usingCCtxParams(const ZSTD_CCtx_params* params)
      * be needed. However, we still allocate two 0-sized buffers, which can
      * take space under ASAN. */
     return ZSTD_estimateCCtxSize_usingCCtxParams_internal(
```
Cyan4973, line 1585: Note : 
I find it strange that `ZSTD_estimateCCtxSize_usingCCtxParams_internal()` doesn't actually use a `const ZSTD_CCtx_params*` object as parameter.
Instead, it uses a long and growing list of parameters, some of them extracted selectively from `const ZSTD_CCtx_params*`.

It seems that accepting a `const ZSTD_CCtx_params*` main parameter would cut down some of this complexity, that this PR makes a little worse by adding a `useExternalMatchFinder` parameter.

That being said, I understand that refactoring `ZSTD_estimateCCtxSize_usingCCtxParams_internal()`, an existing function, is not an obvious part of this PR.
But maybe this could be a legitimate follow up.

embg: If it's trivial I'll do it in this PR, otherwise I'll file an issue to track this work.

Cyan4973 (approved):
Just a few minor comments to suggest, nothing actually blocking.

Generally speaking, I find the amount of changes to the core `libzstd` library to be relatively modest, despite the new proposed feature being actually quite powerful. This is a pretty good sign.

lib/zstd.h
```diff
@@ -2674,6 +2690,120 @@ ZSTDLIB_STATIC_API size_t ZSTD_decompressBlock(ZSTD_DCtx* dctx, void* dst, size_
...
+  const void* src, size_t srcSize,
+  const void* dict, size_t dictSize,
+  int compressionLevel,
+  size_t windowSize
+);
+
+/*! ZSTD_registerExternalMatchFinder() :
+ * Instruct zstd to use an external matchfinder function.
+ *
+ * The externalMatchState must be initialized by the caller, and the caller is
```
GarenJian-Intel, line 2817: There are several compression APIs available to users. Could you address which APIs are supported and which are not?

embg: Thanks for your question! After writing the answer below, I realized that the "Limitations" section of the API docs can use some additional details. I have updated the docs along the lines of the answer below: https://github.com/facebook/zstd/pull/3333/commits/bdaf0617209f1ad15a40003fd03b1cbd26d511e1

Please let me know if you have any other questions that are not answered by the docs. It's super helpful to have your feedback in this area since it's impossible for me to look at the docs with fresh eyes.

I am probably going to land the PR today, but feel free to continue leaving comments after it lands. We can always land a second PR to update the docs if we need to.

---

All compression APIs are supported. There are only two limitations:

* The `ZSTD_c_enableLongDistanceMatching` cParam is not supported. **Compression will fail if it is enabled and the user tries to compress with an external matchfinder.**
    * Note that `enableLongDistanceMatching` is enabled by default in some cases (see its documentation for details). The user needs to be careful not to use the external matchfinder in those cases.

* History buffers are not supported. Concretely, zstd will always pass `dictSize == 0` to the external matchfinder (for now). This has two implications:
    * Dictionaries are not supported. Compression will **not** fail if the user references a dictionary, but the dictionary won't have any effect.
    * Stream history is not supported. All compression APIs, including streaming APIs, work with the external matchfinder, but the external matchfinder won't get any history from the previous block.

Long-term, we plan to overcome both limitations. There's no technical reason why we can't. It's purely a question of engineering effort.

tests/zstreamtest.c
embg: Unfortunately, this isn't possible without a larger refactor because of how the `CHECK` and `CHECK_Z` macros are defined. For example, they depend on the `_output_error` label which is defined inside `basicUnitTests()`.

Compile errors when I tried to move the test into its own function: https://gist.github.com/embg/cc39f0a37bddb57cebbace40374026ca

lib/compress/zstd_compress.c
embg: Filed an issue. I think it's a great one for new contributors to take on! https://github.com/facebook/zstd/issues/3404

tests/zstreamtest.c
Cyan4973: Indeed, I had to refactor these macros in `fuzzer.c` in order to write new unit tests as independent functions.
It's not a difficult effort, but I agree that it's a _different_ effort, and is better kept as a separate topic.
Maybe a follow up.

embg:
Changes since review:
* Rebase onto upstream/dev
* Additional docs on API limitations: https://github.com/facebook/zstd/pull/3333/commits/8052b104e5158dde469f097841c60ea00e676d00
* Fix minor @Cyan4973 nits: https://github.com/facebook/zstd/pull/3333/commits/1e60543110265617d8d2366f5418163f56162925
* Refactor maxNbSeq calculation into a helper function: https://github.com/facebook/zstd/pull/3333/commits/49cd2e8e7980b12919f9d4cba0820d3ac49ae745
* Fix copyright: https://github.com/facebook/zstd/pull/3333/commits/241f2a77530965a9116782c896e9ad9ab15d0da3

CI is passing except for a single test, which seems unrelated to this PR (and is failing on other open PRs).

I'm going to merge! :)

lib/zstd.h
GarenJian-Intel: Thanks, I'll do some test. If there are any questions, we can communicate the details by email.

nadavrot:
Congrats on landing this @embg. I am excited about the use of hardware-accelerated matchfinders in zstd.

abalib:
External Sequence Producer idea is brilliant.  Although processing the source in 128KB chunks is limiting in terms of compression ratio and possibly hardware performance.

As an alternative, I did see in contrib/seqBench this 
```
ZSTD_generateSequences(zc, seqs, seqsSize, inBuf, inBufSize);
ZSTD_CCtx_setParameter(zc, ZSTD_c_blockDelimiters, ZSTD_sf_explicitBlockDelimiters);
size_t outBufSize = ZSTD_compressSequences(zc, outBuf, inBufSize, seqs, seqsSize, inBuf, inBufSize);
```
https://github.com/facebook/zstd/blob/dev/contrib/seqBench/seqBench.c#L29

Here, one could call back the external producer instead of ZSTD_generateSequences(), and the caller is not limited to the 128KB block size.  I verified this experimentally.  I concatenated the same file twice and compressed it using seqBench.c.
The last 6 lines show multiple blocks present and offsets reaching as far back as 200,000 > 128K

```
zstd/contrib/seqBench$ head -c 200000 junk >> junk1
zstd/contrib/seqBench$ head -c 200000 junk >> junk1
 ./seqBench junk1 
LL      ML      OFFS    REP
9       91      1       1
14      13      8       0
...
0       22      19457   0
1       4       19457   1
0       13      572     0
2       62144   200000  0
0       0       0       0
1       131071  200000  1
0       0       0       0
1       6783    200000  1
0       0       0       0

```

embg:
Hi @adalib, thanks for reaching out! I would love to learn more about why you are interested in this API and the use-cases you are targeting. Feel free to reply here or email me at `[my github username]@meta.com`.

You are correct that `ZSTD_compressSequences()` can be used with externally-generated sequences. This is the approach taken by [Intel's QATzip](https://github.com/intel/QATzip/blob/master/utils/qzstd.c#L250).

The motivation for block-level offload is that it integrates with existing compression APIs such as `ZSTD_compress2()` and `ZSTD_compressStream2()`. This is particularly important for streaming compression, which is impossible with `ZSTD_compressSequences()`. But even for small compressions which don't require streaming, maintaining compatibility with the common APIs used in production is an important feature. That's why we added the block-level API, which is used by [Intel's zstd plugin](https://github.com/intel/QAT-ZSTD-Plugin).

> processing the source in 128KB chunks is limiting in terms of compression ratio

So, the block-level API does support offsets larger than 128KB. External sequence producer functions are passed a [`windowSize` parameter](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L2798) and are allowed to produce any offset which is compatible with that history window. The precise requirements on sequences returned by an external callback are given [here](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L2725-L2735).

The missing component is access to the actual content of the history window. In the future, that will be provided by the [`dict` and `dictSize` parameters](https://github.com/facebook/zstd/blob/dev/lib/zstd.h#L2702-L2705). Currently we pass in `NULL` as the dictionary buffer, but we could instead provide access to the previous `windowSize` bytes of history.

This feature is on our long-term roadmap, it simply hasn't been implemented yet. If there is a real-world need for this feature we can add it in the near future.

> ...and possibly hardware performance.

Your concern is absolutely valid. Breaking the input into 128KB chunks may require more back-and-forth communication with the hardware. This is a trade-off we made to maintain compatibility with existing APIs. `ZSTD_compressSequences()` is an option to potentially use hardware more efficiently, but that API also has significant downsides (as discussed above).

Please let me know if you have any further questions!

Issue #3239, opened 2022-08-10
1 GiB file bug


ip7z:
zstd 1.5.2 doesn't reduce the Windows Size from 2 GiB to 1 GiB, if file size is exactly 1 GiB.

**To Reproduce**
```
FSUtil File CreateNew 1024 0x40000000
zstd --long=31 -1 --single-thread --no-content-size -f 1024
zstd -l -v 1024.zst
```
result:
```
Window Size: 2.00 GiB (2147483648 B)
```
**Expected behavior**
```
Window Size: 1.00 GiB
```
**Source code lines**
the bug probably in function `ZSTD_adjustCParams_internal()` in `zstd_compress.c`:
```
    const U64 maxWindowResize = 1ULL << (ZSTD_WINDOWLOG_MAX-1);
   ... 
    /* resize windowLog if input is small enough, to use less memory */
    if ( (srcSize < maxWindowResize)
```

and fixed version probably must use `<=` instead of `<`:
```
    if ( (srcSize <= maxWindowResize)
```

embg:
Thanks for the report! I will put up a fix before the next release.

hbina:
I think the original author wanted to be extra safe with the addition that follows. I believe something like this is enough?

```git
hbina@akarin:~/git/zstd$ git diff
diff --git a/lib/compress/zstd_compress.c b/lib/compress/zstd_compress.c
index 59d441b2..44693bf7 100644
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1369,8 +1369,8 @@ ZSTD_adjustCParams_internal(ZSTD_compressionParameters cPar,
     }
 
     /* resize windowLog if input is small enough, to use less memory */
-    if ( (srcSize < maxWindowResize)
-      && (dictSize < maxWindowResize) )  {
+    if ( (srcSize <= maxWindowResize)
+      && (dictSize <= maxWindowResize) )  {
         U32 const tSize = (U32)(srcSize + dictSize);
         static U32 const hashSizeMin = 1 << ZSTD_HASHLOG_MIN;
         U32 const srcLog = (tSize < hashSizeMin) ? ZSTD_HASHLOG_MIN :
 ```

PR #3345, opened 2022-12-13
Fix Window Resizing Edge Case

daniellerozenblit:
This PR fixes the bug addressed in #3239.

I have implemented the fix proposed and have validated that the issue no longer occurs.

```
dd if=/dev/zero of=testFile  bs=1024M  count=1

./zstd --long=31 -1 --single-thread --no-content-size -f ~/testFile
~/testFile :  0.00%   ( 1.000 GiB =>   32.0 KiB, /home/drozenblit/testFile.zst) 

./zstd -l -v ~/testFile.zst
Window Size: 1.000 GiB (1073741824 B)
```


Cyan4973:
Would it be possible to add an associated test in `playTests.sh` ?

terrelln:
I'd recommend adding a test to `cli-tests` compression tests, should be a bit easier than in `playTests.sh`.

https://github.com/facebook/zstd/tree/dev/tests/cli-tests/compression
https://github.com/facebook/zstd/tree/dev/tests/cli-tests/README.md

tests/cli-tests/compression/window-resize.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+datagen -g1G > file
+zstd --long=31 -1 --single-thread --no-content-size -f file
+zstd -l -v file.zst
+
+# We want to ignore stderr (its outputting "*** zstd command line interface
```
Cyan4973, line 6: One way to silence `stderr` could be `2> /dev/null` (redirects `stderr` to `/dev/null`)

tests/cli-tests/compression/window-resize.sh.stderr.glob
```diff
@@ -0,0 +1 @@
+...
```
Cyan4973: minor : I'm wondering what happens if this file is not provided, or if it is replaced by the `*.ignore` suffix variant.
This could be a more explicit way to express that `stderr` output is not significant.
cc @terrelln .

daniellerozenblit: Using the `.ignore` suffix seems to work! I can switch to that, I like that it is a bit more explicit than redirection.

Cyan4973 (approved):

PR #3058, opened 2022-02-08
seekable_format no zstd header compressing empty string to stream

yhoogstrate:
Dear zstd maintainers,

I believe I ran into a small bug in zstd_seekable. When compressing an empty string, the output of the compressed data does not include the zstd seekable header:

```
// Example 1, compressing "test"

char *data_in = (char *) malloc_or_die(255 * sizeof(char));
data_in = "test\0";

char *data_out = (char *) malloc_or_die(255 * 255 * sizeof(char));

ZSTD_seekable_CStream *s = ZSTD_seekable_createCStream();
size_t const initResult = ZSTD_seekable_initCStream(s, 1, 1, 1024 * 1024);

ZSTD_inBuffer input = { data_in, strlen(data_in), 0 };
ZSTD_outBuffer output = { data_out, 255*255, 0 };

size_t toRead = ZSTD_seekable_compressStream(s, &output, &input);
size_t const remainingToFlush = ZSTD_seekable_endStream(s, &output);

printf("size compressed output: %i\n",output.pos);

ZSTD_seekable_freeCStream(s);
```

```
// Example 2, compressing "" (empty string)

char *data_in = (char *) malloc_or_die(255 * sizeof(char));
data_in = "\0";

char *data_out = (char *) malloc_or_die(255 * 255 * sizeof(char));

ZSTD_seekable_CStream *s = ZSTD_seekable_createCStream();
size_t const initResult = ZSTD_seekable_initCStream(s, 1, 1, 1024 * 1024);

ZSTD_inBuffer input = { data_in, strlen(data_in), 0 };
ZSTD_outBuffer output = { data_out, 255*255, 0 };

size_t toRead = ZSTD_seekable_compressStream(s, &output, &input);
size_t const remainingToFlush = ZSTD_seekable_endStream(s, &output);

printf("size compressed output: %i\n",output.pos);

ZSTD_seekable_freeCStream(s);
```

results in the following two compressed strings:

```
test 1:

00000000: 28b5 2ffd 0048 2100 0074 6573 745e 2a4d  (./..H!..test^*M
00000010: 1815 0000 000d 0000 0004 0000 0039 8167  .............9.g
00000020: db01 0000 0080 b1ea 928f                 ..........
```

```
test 2:

00000000: 5e2a 4d18 0900 0000 0000 0000 80b1 ea92  ^*M.............
00000010: 8f                                       .
```

[...]

PR #3346, opened 2022-12-13
Seekable format empty string

daniellerozenblit:
This PR is a followup to a [previous PR](https://github.com/facebook/zstd/pull/3058) by @yhoogstrate. The goal of the previous PR was to fix a small bug in `seekable_format` that led to the omission of the ZSTD MAGIC when compressing an empty string.

I have validated the bug fix with the test case provided, but have created this PR to fix some minor issues (forgetting to free malloc'd memory in the test case).

### DIFF
--- a/contrib/seekable_format/tests/seekable_tests.c
+++ b/contrib/seekable_format/tests/seekable_tests.c
@@ -189,18 +189,18 @@ int main(int argc, const char** argv)
 
     printf("Test %u - check ZSTD magic in compressing empty string: ", testNb++);
     { // compressing empty string should return a zstd header
-        char *data_in = (char *) malloc(255 * sizeof(char));
-        assert(data_in != NULL);
-        data_in = "\0";
-
-        char *data_out = (char *) malloc(255 * 255 * sizeof(char));
-        assert(data_out != NULL);
+        size_t const capacity = 255;
+        char* inBuffer = malloc(capacity);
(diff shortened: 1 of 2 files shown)


terrelln:
This looks good to me once CI passes!

Only question is do you know if the seekable format tests run in CI?

Cyan4973:
> Only question is do you know if the seekable format tests run in CI?

Good point,
Seekable format tests can be triggered directly,
or as part of `contrib/` tests,
which itself can also be triggered as part `make test-all`.

But ... looking at the list of tests on Github Actions,
it's unclear to me if any of these tests are run.
I'm sure they used to be part of CI,
but it might have been in the TravisCI era, 
and they could have been forgotten in the transition towards Github Actions ...

_edit_ : I can't find any trace of `contrib/` tests within `.github/` workflows,
so we actually might have to add this test.

_edit 2_: Actually, `make contrib` is also part of `make all`,
and we do have one `make all` test,
and it runs the seekable format tests : 
https://github.com/facebook/zstd/actions/runs/3688718449/jobs/6243837843#step:3:228

So that's fine !

yhoogstrate:
wow thanks @daniellerozenblit to carry this forwards ! Good to see the test works and the contrib tests are included in the CI. Sorry I missed the free after malloc 😳

embg (approved):

Issue #3353, opened 2022-12-14
Adaptive parameter updates using --adapt don't get triggered if --no-progress is used


terrelln:
This command doesn't trigger adaptive parameter updating

```
zstd --adapt --no-progress
```

PR #3354, opened 2022-12-14
Separate parameter adaption from display update rate

terrelln:
Split the logic for parameter adaption from the logic to update the display rate.
This decouples the two updates, so changes to display updates don't affect
parameter adaption.

Also add a test case that checks that parameter adaption actually happens.

This fixes Issue https://github.com/facebook/zstd/issues/3353, where --adapt is broken when --no-progress is passed.

Note: This is based on top of PR #2984. So only the 3rd commit in the stack is relevant.


Cyan4973 (approved):
That's a welcomed decoupling ! 
It makes the logic much clearer.

Issue #3350, opened 2022-12-14
File compressed with wlog=25 gets corrupted


tansy:
**Describe the bug**
File compressed with wlog=25 gets corrupted

**To Reproduce**
I tried to compress silesia.tar with wlog=25. I went fine but when tested i reported error and decompression to stdout failed in the middle of the file.
```
$ zstd -k -19 --zstd=wlog=25 silesia.tar
$  zstd -t silesia.tar.zst 
silesia.tar.zst      : 84.9 MiB...     silesia.tar.zst : Decoding error (36) : Corrupted block detected 
```
I uploaded these files to the cloud. You can check them here:
Original, compressed: [silesia.tar.lz](https://files.fm/f/v3fweyzun)
Corrupted zst: [silesia.tar.zst](https://files.fm/f/54wqbp3ef)

**Expected behavior**
To be able to decompress compressed file.

**Desktop (please complete the following information):**
 - OS: linux
 - Version: zstd-v1.5.2
 - Compiler: gcc
 - Flags: default
 - Build system: Makefile

Just tested brand new dev branch with the same result.

Cyan4973:
This is surprising.

For information, I tested the exact same command, using the `dev` branch,
`zstd -k -19 --zstd=wlog=25 silesia.tar`
and there was no such problem, everything round-tripped properly.

Your report implies that you have observed this problem multiple times, that it's reproducible ?

_edit_ : I can confirm that the file provided as `silesia.tar.zst` is corrupted,
but I can't reproduce the same file on my side.

tansy:
I will drop you binaries later.

> Your report implies that you have observed this problem multiple times, that it's reproducible?

Yes, I compiled dev branch with gcc and clang with the same result.
You can reproduce it as well - other binaries (windows) from [1.5.2 release](https://github.com/facebook/zstd/releases/tag/v1.5.2
) share the same "feature".

I would compile windows version but my wine has trouble with these functions \`InitializeConditionVariable', \`SleepConditionVariableCS' on that computer.

terrelln:
This is definitely unexpected... I'll take a look at the compressed file and see if I see anything obviously wrong.

In the mean time, can you provide more information about your system? Anything you can think of would help, but at least:
* Filesystem the data is stored on
* Compiler versions
* OS version: `uname -a` output
* CPU: `lscpu` output

Cyan4973:
I could test with the exact `silesia.tar` produced by `silesia.tar.lz`,
[...]

PR #3361, opened 2022-12-15
Fix corruption that rarely occurs in 32-bit mode with wlog=25

terrelln:
Fix an off-by-one error in the compressor that emits corrupt blocks if:

* Zstd is compiled in 32-bit mode
* The windowLog == 25 exactly
* An offset of 2^25-3, 2^25-2, 2^25-1, or 2^25 is emitted
* The bitstream had 7 bits leftover before writing the offset

This bug has been present since before v1.0, but wasn't able to easily be triggered, since until somewhat recently zstd wasn't able to find matches that were within 128KB of the window size.

Add a test case, and fix 2 bugs in `ZSTD_compressSequences()`:
* The `ZSTD_isRLE()` check was incorrect. It wouldn't produce corruption, but it could waste CPU and not emit RLE even if the block was RLE
* One windowSize was `1 << windowLog`, not `1u << windowLog`

Thanks to @tansy for finding the issue, and giving us a reproducer!

Fixes Issue #3350.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2624,7 +2624,7 @@ ZSTD_entropyCompressSeqStore_internal(seqStore_t* seqStorePtr,
                                 void* entropyWorkspace, size_t entropyWkspSize,
                           const int bmi2)
 {
-    const int longOffsets = cctxParams->cParams.windowLog > STREAM_ACCUMULATOR_MIN;
+    const int longOffsets = cctxParams->cParams.windowLog >= STREAM_ACCUMULATOR_MIN;
     ZSTD_strategy const strategy = cctxParams->cParams.strategy;
     unsigned* count = (unsigned*)entropyWorkspace;
     FSE_CTable* CTable_LitLength = nextEntropy->fse.litlengthCTable;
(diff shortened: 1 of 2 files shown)


terrelln:
A follow up is to add fuzzers that could catch this issue:

1. Run our fuzzers in 32-bit mode in OSS-Fuzz
2. Add a fuzzer which uses the `ZSTD_compressSequences()` API to generate valid fuzzed sequences, and check that they round trip. This would allow us to fuzz very large offsets without requiring the fuzzer generate many many MB of input data.

lib/compress/zstd_compress.c
```diff
@@ -6256,7 +6256,7 @@ ZSTD_compressSequences_internal(ZSTD_CCtx* cctx,
 
         if (!cctx->isFirstBlock &&
             ZSTD_maybeRLE(&cctx->seqStore) &&
-            ZSTD_isRLE((BYTE const*)src, srcSize)) {
+            ZSTD_isRLE(ip, blockSize)) {
```
Cyan4973, line 6259: Good catch !

Cyan4973 (approved):

PR #3363, opened 2022-12-15
[api][visibility] Make the visibility macros more consistent

### LINKED ISSUE #3359: Inconsistency: ZSTDLIB_VISIBLE vs ZSTDLIB_VISIBILITY
## Describe the bug

In order to properly set symbol visibility, **zstd** uses some macros `ZSTDLIB_API`, `ZDICTLIB_API` and `ZSTDERRORLIB_API`.
Those get their values through some other macros:
* `ZSTDLIB_API` gets its value from `ZSTDLIB_VISIBLE`,
* `ZDICTLIB_API` from `ZDICTLIB_VISIBILITY` and
* `ZSTDERRORLIB_API` from `ZSTDERRORLIB_VISIBILITY`.

As you can see, the naming is not consistent, one macro ends in the adjective `VISIBLE` while the other two end in the noun `VISIBILITY`.

This alone is confusing and might lead to errors. (I tripped over this when I pre-defined the macro `ZSTDLIB_VISIBILITY`, similar to the macros `ZDICTLIB_VISIBILITY` and `ZSTDERRORLIB_VISIBILITY`, in order to modify symbol visibility.)

**And in fact, you seem to have run into this mistake, too: The `Makefile` in `contrib/linux-kernel/Makefile` makes exactly the same error and tries to replace the first occurrence of `ZSTDLIB_VISIBILITY` which does not exist.**

## A clear and concise description of what the bug is.

The bug is an inconsistency in naming that might lead to errors and which already lead to an error in the provided `contrib/linux-kernel/Makefile`.

## Expected behavior/fix

All occurrences of `ZSTDLIB_VISIBLE` in the code should probably be replaced by `ZSTDLIB_VISIBILITY`.  
Alternatively, the occurrence of `ZSTDLIB_VISIBILITY` in `contrib/linux-kernel/Makefile` needs to be replaced by `ZSTDLIB_VISIBLE`.

## Screenshots and charts

N/A

## Desktop (please complete the following information)

Not really relevant, but here you go:

 - OS: Ubuntu Linux
 - Version: 20.04
 - Compiler: clang/gcc
 - Flags: N/A
 - Other relevant hardware specs: N/A
 - Build system: CMake

## Additional context

As explained above, this is especially important if one wants to modify symbol visibility.

terrelln:
1. Follow the scheme introduced in PR #2501 for both `zdict.h` and `zstd_errors.h`.
2. If the `*_VISIBLE` macro isn't set, but the `*_VISIBILITY` macro is, use that. Also make this change for `zstd.h`, since we probably shouldn't have changed that macro name without backward compatibility in the first place.
3. Change all references to `*_VISIBILITY` to `*_VISIBLE`.
4. Add `ZDICTLIB_STATIC_API` to be consistent with `zstd.h`, this is a no-op change by default.
5. Document these macros in `lib/README.md`.

Fixes #3359.


Cyan4973:
OK, so switch to `*_VISIBLE`, only keep `*_VISIBILITY` for backward compatibility purposes.

Also :  we should probably document these build macros,
since they influence the scope of `libzstd` library,
and are meant to be employed by 3rd party library builders.
Such documentation should be part of `lib/README.md`.

terrelln:
Update README

Cyan4973 (approved):

PR #3364, opened 2022-12-16
Windows MT layer bug fixes

yoniko:
Fixes two bugs in the Windows thread / pthread translation layer:
1. If threads are resized the threads' `ZSTD_pthread_t` might move while the worker still holds a pointer into it (see more details in #3120).
2. The join operation was waiting for a thread and then return its `thread.arg` as a return value, but since the `ZSTD_pthread_t thread` was passed by value it would have a stale `arg` that wouldn't match the thread's actual return value.

This fix changes the `ZSTD_pthread_join` API and removes support for returning a value. This means that we are diverging from the `pthread_join` API and this is no longer just an alias. In the future, if needed, we could return a Windows thread's return value using `GetExitCodeThread`, but as this path wouldn't be excised in any case, it's preferable to not add it right now.

### DIFF
--- a/lib/common/threading.c
+++ b/lib/common/threading.c
@@ -37,7 +37,7 @@ int g_ZSTD_threading_useless_symbol;
 static unsigned __stdcall worker(void *arg)
 {
     ZSTD_pthread_t* const thread = (ZSTD_pthread_t*) arg;
-    thread->arg = thread->start_routine(thread->arg);
+    thread->start_routine(thread->arg);
     return 0;
 }
 
@@ -55,7 +55,7 @@ int ZSTD_pthread_create(ZSTD_pthread_t* thread, const void* unused,
         return 0;
 }
 
-int ZSTD_pthread_join(ZSTD_pthread_t thread, void **value_ptr)
+int ZSTD_pthread_join(ZSTD_pthread_t thread)
 {
     DWORD result;
 
@@ -66,7 +66,6 @@ int ZSTD_pthread_join(ZSTD_pthread_t thread, void **value_ptr)
 
     switch (result) {
     case WAIT_OBJECT_0:
-        if (value_ptr) *value_ptr = thread.arg;
         return 0;
     case WAIT_ABANDONED:
         return EINVAL;
--- a/lib/common/threading.c
+++ b/lib/common/threading.c
@@ -44,15 +44,23 @@ typedef struct {
 
 static unsigned __stdcall worker(void *arg)
 {
-    ZSTD_thread_params_t* const thread_param = (ZSTD_thread_params_t*)arg;
-    void* (*start_routine)(void*) = thread_param->start_routine;
-    void* thread_arg = thread_param->arg;
+    void* (*start_routine)(void*);
+    void* thread_arg;
 
-    /* Signal main thread that we are running and do not depend on its memory anymore */
-    ZSTD_pthread_mutex_lock(&thread_param->initialized_mutex);
-    thread_param->initialized = 1;
-    ZSTD_pthread_mutex_unlock(&thread_param->initialized_mutex);
-    ZSTD_pthread_cond_signal(&thread_param->initialized_cond);
+    /* Inialized thread_arg and start_routine and signal main thread that we don't need it
+     * to wait any longer.
+     */
+    {
+        ZSTD_thread_params_t*  thread_param = (ZSTD_thread_params_t*)arg;
+        thread_arg = thread_param->arg;
+        start_routine = thread_param->start_routine;
+
+        /* Signal main thread that we are running and do not depend on its memory anymore */
+        ZSTD_pthread_mutex_lock(&thread_param->initialized_mutex);
+        thread_param->initialized = 1;
+        ZSTD_pthread_cond_signal(&thread_param->initialized_cond);
+        ZSTD_pthread_mutex_unlock(&thread_param->initialized_mutex);
+    }
 
     start_routine(thread_arg);
 
@@ -63,25 +71,34 @@ int ZSTD_pthread_create(ZSTD_pthread_t* thread, const void* unused,
             void* (*start_routine) (void*), void* arg)
 {
     ZSTD_thread_params_t thread_param;
-    int error = 0;
     (void)unused;
+
     thread_param.start_routine = start_routine;
     thread_param.arg = arg;
     thread_param.initialized = 0;
+    *thread = NULL;
 
     /* Setup thread initialization synchronization */
-    error |= ZSTD_pthread_cond_init(&thread_param.initialized_cond, NULL);
-    error |= ZSTD_pthread_mutex_init(&thread_param.initialized_mutex, NULL);
-    if(error)
+    if(ZSTD_pthread_cond_init(&thread_param.initialized_cond, NULL)) {
+        /* Should never happen on Windows */
         return -1;
-    ZSTD_pthread_mutex_lock(&thread_param.initialized_mutex);
+    }
+    if(ZSTD_pthread_mutex_init(&thread_param.initialized_mutex, NULL)) {
+        /* Should never happen on Windows */
+        ZSTD_pthread_cond_destroy(&thread_param.initialized_cond);
+        return -1;
+    }
 
     /* Spawn thread */
     *thread = (HANDLE)_beginthreadex(NULL, 0, worker, &thread_param, 0, NULL);
-    if (!thread)
+    if (!thread) {
+        ZSTD_pthread_mutex_destroy(&thread_param.initialized_mutex);
+        ZSTD_pthread_cond_destroy(&thread_param.initialized_cond);
         return errno;
+    }
 
     /* Wait for thread to be initialized */
+    ZSTD_pthread_mutex_lock(&thread_param.initialized_mutex);
     while(!thread_param.initialized) {
         ZSTD_pthread_cond_wait(&thread_param.initialized_cond, &thread_param.initialized_mutex);
     }
--- a/lib/common/threading.c
+++ b/lib/common/threading.c
@@ -34,35 +34,72 @@ int g_ZSTD_threading_useless_symbol;
 
 /* ===  Implementation  === */
 
+typedef struct {
+    void* (*start_routine)(void*);
+    void* arg;
+    int initialized;
+    ZSTD_pthread_cond_t initialized_cond;
+    ZSTD_pthread_mutex_t initialized_mutex;
+} ZSTD_thread_params_t;
+
 static unsigned __stdcall worker(void *arg)
 {
-    ZSTD_pthread_t* const thread = (ZSTD_pthread_t*) arg;
-    thread->start_routine(thread->arg);
+    ZSTD_thread_params_t* const thread_param = (ZSTD_thread_params_t*)arg;
+    void* (*start_routine)(void*) = thread_param->start_routine;
+    void* thread_arg = thread_param->arg;
+
+    /* Signal main thread that we are running and do not depend on its memory anymore */
+    ZSTD_pthread_mutex_lock(&thread_param->initialized_mutex);
+    thread_param->initialized = 1;
+    ZSTD_pthread_mutex_unlock(&thread_param->initialized_mutex);
+    ZSTD_pthread_cond_signal(&thread_param->initialized_cond);
+
+    start_routine(thread_arg);
+
     return 0;
 }
 
 int ZSTD_pthread_create(ZSTD_pthread_t* thread, const void* unused,
             void* (*start_routine) (void*), void* arg)
 {
+    ZSTD_thread_params_t thread_param;
+    int error = 0;
     (void)unused;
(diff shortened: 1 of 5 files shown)


yoniko:
@eli-schwartz - can you verify that this PR solves the issues observed in #3120?

eli-schwartz:
Unfortunately I still get the issue. I'm using this branch to test with, since it has a couple other fixes (particularly the valgrind issue, but I discovered another bug that was merged today): https://github.com/eli-schwartz/zstd/commits/pr3121

```
3/8 test-zstream-3           FAIL            14.01s   (exit status 3221225477 or signal 3221225349 SIGinvalid)
>>> MALLOC_PERTURB_=136 D:\a\wrapdb\wrapdb\_build\subprojects\zstd-pr3121\build/meson/tests\zstreamtest.exe --newapi -t1 -T90s --no-big-tests
------------------------------------- 8< -------------------------------------
stderr:
Starting zstream tester (64-bits, 1.5.3)

Seed = 3089


     1/     1    
     2          
    16          
    32          
    49          
    59          
    66          
    74          
    78          
    87          
   114          
   117          
   123          
   129          
   132          
   150          
   151          
   161          
   168          
   173          
   182          
   191          
   197          
   203          
   204          
   214          
   221          
   223          
   231          
   234          
   237          
   252          
   263          
   273          
   274          
   310          
------------------------------------------------------------------------------
```

eli-schwartz:
Aside: the git commit message seems to be a bit off. It doesn't have a one-line subject separated by a blank line from the extended body, so git's own paragraph detection treats that quite oddly when e.g. rebasing that commit into a selected list of commits to test. (Also, lines don't wrap.)
screenshot

yoniko:
> Unfortunately I still get the issue. I'm using this branch to test with, since it has a couple other fixes (particularly the valgrind issue, but I discovered another bug that was merged today): https://github.com/eli-schwartz/zstd/commits/pr3121

I've tested again and verified that locally I'm not seeing crashes after the patch is applied.
I'm going to try and compile with meson on Windows, but it will looks like the correct way forward is to produce a CI that crashes and that I can apply patches to so we can have a source of truth for the state of the tests.

eli-schwartz:
Hmm okay, but how should we do that? Add a CI job that reports failure and causes `dev` to go red? #3120 did the reverse and marked the failure as a TODO, so that plus my CI modifications would allow testing this in CI because any change that fixes the crash would make CI fail with a meson "test unexpectedly started passing" type of message. Or by reverting the TODO mark, the CI would fail if the fix is incomplete.

yoniko:
If we just add a PR with the CI and not merge it it will run just for that CI.
I can follow up by rebasing this PR on top of the PR with CI - if this PR really solves the issue we will have one red PR and one green one.

EDIT: Perhaps we can use #3120 as a base, I'll give it a try.

eli-schwartz:
> EDIT: Perhaps we can use #3120 as a base, I'll give it a try.

Try basing on the first 6 of the 7 commits now there. The last one is the TODO marker. With the issue still present but the 7th commit also there, the `dev-short-tests / meson-windows` job passes. The tests log:
```
3/8 test-zstream-3           EXPECTEDFAIL     2.46s   (exit status 3221225477 or signal 3221225349 SIGinvalid)

[...]

Ok:                 7   
Expected Fail:      1   
```

yoniko:
So I merged the two PRs into one (#3370) and I'm seeing the same thing as you -- it's passing although I expected it to fail with UNEXPECTEDPASS.
When replicating the exact same commands the action runs on my Windows VM I'm getting a different result.

EDIT: I have managed to generate more crashes in local environment, seems  to trigger more when pinning the process to one core -- suggests a race condition.

EDIT2: I think I know where the race condition is, I'll test it later today.

terrelln (changes requested):
Ok, it took me a second to understand the problem, because I expected `pthread_t` to be un-movable. But it looks like pthreads doesn't require that the `pthread_t*` is valid for the lifetime of the worker.

Which means that this fix is insufficient. By the time `worker(void *arg)` is called, the `pthread_t` may have already been moved.

We need to do something like:

```c
typedef struct {
    void* (*start_routine)(void*);
    void* arg;
} ZSTD_pthread_state;

typedef struct {
    HANDLE handle;
    ZSTD_pthread_state* state;
} ZSTD_pthread_t;

static unsigned __stdcall worker(void *arg)
{
    ZSTD_pthread_state* const state = (ZSTD_pthread_state*) state;
    state->start_routine(state->arg);
    return 0;
}

int ZSTD_pthread_create(ZSTD_pthread_t* thread, const void* unused,
            void* (*start_routine) (void*), void* arg)
{
    (void)unused;
    thread->state = malloc(sizeof(ZSTD_pthread_state));
    if (!thread->state) {
        thread->handle = NULL;
        return ENOMEM;
    }
    thread->state->arg = arg;
    thread->state->start_routine = start_routine;
    thread->handle = (HANDLE) _beginthreadex(NULL, 0, worker,thread->state, 0, NULL);
    if (!thread->handle) {
        free(thread->state);
        thread->state = NULL;
        return errno;
    }
    return 0;
}

int ZSTD_pthread_join(ZSTD_pthread_t thread)
{
    DWORD result;

    if (!thread.handle) {
        assert(!thread.state);
        return 0;
    }
    result = WaitForSingleObject(thread.handle, INFINITE);
    CloseHandle(thread.handle);
    free(thread.state);
    // ...
}
```

yoniko:
@terrelln - yes, this is the race condition that I was talking about.
I implemented a different solution that doesn't depend on malloc but needs a lock.
It's currently being tested in #3370.

yoniko:
See two latest commits - one fixes the bug and the other updates the meson tests to expect to pass on Windows.

lib/common/threading.c
```diff
@@ -34,39 +34,75 @@ int g_ZSTD_threading_useless_symbol;
...
+    ZSTD_thread_params_t* const thread_param = (ZSTD_thread_params_t*)arg;
+    void* (*start_routine)(void*) = thread_param->start_routine;
+    void* thread_arg = thread_param->arg;
+
+    /* Signal main thread that we are running and do not depend on its memory anymore */
+    ZSTD_pthread_mutex_lock(&thread_param->initialized_mutex);
+    thread_param->initialized = 1;
+    ZSTD_pthread_mutex_unlock(&thread_param->initialized_mutex);
+    ZSTD_pthread_cond_signal(&thread_param->initialized_cond);
+
```
terrelln, line 64: nit: Set `thread_param = NULL` to clarify that this memory is now invalid.

lib/common/threading.c
```diff
@@ -34,39 +34,75 @@ int g_ZSTD_threading_useless_symbol;
...
+    thread_param.start_routine = start_routine;
+    thread_param.arg = arg;
+    thread_param.initialized = 0;
+
+    /* Setup thread initialization synchronization */
+    error |= ZSTD_pthread_cond_init(&thread_param.initialized_cond, NULL);
+    error |= ZSTD_pthread_mutex_init(&thread_param.initialized_mutex, NULL);
+    if(error)
+        return -1;
+    ZSTD_pthread_mutex_lock(&thread_param.initialized_mutex);
```
terrelln: Why lock this mutex here? I don't see any reason to hold it while creating the thread. So lets minimize the scope and move it to line 84.

lib/common/threading.c
```diff
@@ -34,39 +34,75 @@ int g_ZSTD_threading_useless_symbol;
...
+
+    /* Spawn thread */
+    *thread = (HANDLE)_beginthreadex(NULL, 0, worker, &thread_param, 0, NULL);
+    if (!thread)
         return errno;
```
terrelln, line 97: We need to unlock (until it is moved below the thread creation), and destroy the mutex/cond in this case.

lib/common/threading.c
```diff
@@ -34,39 +34,75 @@ int g_ZSTD_threading_useless_symbol;
...
-    if (!thread->handle)
+    thread_param.start_routine = start_routine;
+    thread_param.arg = arg;
+    thread_param.initialized = 0;
+
+    /* Setup thread initialization synchronization */
+    error |= ZSTD_pthread_cond_init(&thread_param.initialized_cond, NULL);
+    error |= ZSTD_pthread_mutex_init(&thread_param.initialized_mutex, NULL);
+    if(error)
+        return -1;
```
terrelln: I'd recommend against this pattern here. We need to destroy whichever one we initialized correctly.

lib/common/threading.c
```diff
@@ -34,39 +34,75 @@ int g_ZSTD_threading_useless_symbol;
...
     (void)unused;
-    thread->arg = arg;
-    thread->start_routine = start_routine;
-    thread->handle = (HANDLE) _beginthreadex(NULL, 0, worker, thread, 0, NULL);
-
-    if (!thread->handle)
+    thread_param.start_routine = start_routine;
+    thread_param.arg = arg;
+    thread_param.initialized = 0;
+
```
terrelln, line 80: nit: Set `*thread = NULL` here so when we error on line `75`, we won't run the destructor. Otherwise we will be reading uninitialized memory in the destructor.

terrelln (changes requested):
Overall strategy looks good, just needs some fixes in error conditions.

lib/common/threading.c
yoniko: You are correct, don't know why it's there, may have moved stuff and missed it.

lib/common/threading.c
yoniko: These two functions can't really fail, so we will never hit this error condition, it's just there because we have a return value (to align with pthreads) and we have to handle the return value to not get compilation errors.
Still, I'll add cleanup for for compatability.

lib/common/threading.c
yoniko: These two functions can't really fail, so we will never hit this error condition, it's just there because we have a return value (to align with pthreads) and we have to handle the return value to not get compilation errors.
Still, I'll add cleanup for for compatibility.

Cyan4973: Could an `assert()` be used if only to document that this part of the code path, or this condition, can never be reached ?

yoniko: I've added a comment and handled the errors in any case. If you think an assert would be better let me know.

lib/common/threading.c
yoniko: Added a specific scope in which `thread_param` is defined so it's invalid afterwards.

lib/common/threading.c
yoniko: Good point, added

lib/common/threading.c
```diff
@@ -34,39 +34,92 @@ int g_ZSTD_threading_useless_symbol;
...
+        /* Should never happen on Windows */
+        ZSTD_pthread_cond_destroy(&thread_param.initialized_cond);
+        return -1;
+    }
+
+    /* Spawn thread */
+    *thread = (HANDLE)_beginthreadex(NULL, 0, worker, &thread_param, 0, NULL);
+    if (!thread) {
+        ZSTD_pthread_mutex_unlock(&thread_param.initialized_mutex);
+        ZSTD_pthread_mutex_destroy(&thread_param.initialized_mutex);
```
terrelln, line 95: Also need to destroy the condition variable.

terrelln (changes requested):
LGTM, just need to destroy the condition variable if creating the thread fails.

terrelln:
Edit: Wrong PR

lib/common/threading.c
```diff
@@ -34,39 +34,92 @@ int g_ZSTD_threading_useless_symbol;
...
+    {
+        ZSTD_thread_params_t*  thread_param = (ZSTD_thread_params_t*)arg;
+        thread_arg = thread_param->arg;
+        start_routine = thread_param->start_routine;
+
+        /* Signal main thread that we are running and do not depend on its memory anymore */
+        ZSTD_pthread_mutex_lock(&thread_param->initialized_mutex);
+        thread_param->initialized = 1;
+        ZSTD_pthread_mutex_unlock(&thread_param->initialized_mutex);
+        ZSTD_pthread_cond_signal(&thread_param->initialized_cond);
```
terrelln: Sorry, I just realized this issue.

It is incorrect to signal the condition variable after we release the lock. `ZSTD_pthread_cond_wait()` is allowed to spuriously wake up. So we could get this execution:

Main thread: Launches the worker thread & waits on the condition variable.
Worker thread: Sets initialized to 1 & unlocks the mutex
Main thread: Spuriously wakes, sees that it is initialized, and destroys the condition variable
Worker thread: Signals the already destroyed condition variable.

So we just need to signal the CV before we unlock the mutex. This is allowed, and can actually generally more efficient, because threading libraries optimize this use case.

terrelln (changes requested):

lib/common/threading.c
yoniko: Good point, will make the change.

terrelln (approved):
LGTM

PR #3368, opened 2022-12-16
meson: fix broken commit that broke the build

eli-schwartz:
In commit 031de3c69ccbf3282ed02fb49369b476730aeca8 some code was added that returned a boolean, but was treated as if it returned a dependency object. This wasn't tested and could not work. Moreover, zstd no longer built at all unless the entire programs directory was disabled and not even evaluated.

Fix the return type checking.

Also:
- fix warning for using too-new features

### DIFF
--- a/build/meson/meson.build
+++ b/build/meson/meson.build
@@ -26,7 +26,7 @@ project('zstd',
   version: run_command(
     find_program('GetZstdLibraryVersion.py'), '../../lib/zstd.h',
     check: true).stdout().strip(),
-  meson_version: '>=0.48.0')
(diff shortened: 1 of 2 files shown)


eli-schwartz:
This should have been caught by Travis CI, but apparently Travis is no longer enabled. And that was where the historic tests of Meson were being done.

As mentioned in another issue, I have prepared additions to the CI to test Meson -- they currently rely on building other fixes, so I'll PR those once the other fixes go in.

yoniko (approved):

yoniko:
Looks good, merging.

PR #3370, opened 2022-12-16
Fix windows mt thread resize bug with CI - draft

yoniko:


yoniko:
This is just a draft PR that merges the added Meson CI PR with the Windows MT fixes PR.
It exists so we can see if the CI performs as expected with both changes (meson windows should break here job).

yoniko:
No longer needed, closing.

PR #3374, opened 2022-12-17
Fix Meson-Windows CI test by pulling a more recent version msvc-dev-cmd

yoniko:
Recently merge meson-windows CI test fails on a dependency issue.
Updating pinned hash to latest version remedies this issue.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -209,7 +209,7 @@ jobs:
       - name: Install packages
         run: pip install --pre meson
       - name: Initialize the MSVC dev command prompt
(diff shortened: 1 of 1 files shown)


yoniko:
@eli-schwartz looks like I was overly eager to merge your PR when in fact it was broken, this should fix the issue, please take a look.

Cyan4973 (approved):

eli-schwartz:
That's entirely possible, as I didn't even test the hash update (I did test the `@v1` version). I didn't have any time yesterday to do more than push the update. Turns out I think I copied the hash from the head of the wrong branch though.

yoniko:
That's fine and thank you for contributing.
I was under the impression that the test failed due to a sporadic issue and decided to merge it, only later I realized it broke CI.

Issue #3204, opened 2022-07-18
I'm not 100% sure that block splitter is working properly


thatch:
**Describe the bug**

This is not a correctness issue.  It simply appear to be leaving a lot of ratio unattained when faced with real-world data.

Side notes:
- zstd is zstd 1.5.2-3 from archlinux
- python-zstandard is 0.18.0 using backend_c
- block splitter being an experimental param, I don't think I can enable separately in the zstd CLI or in the [python-zstandard compression params](https://github.com/indygreg/python-zstandard/blob/main/c-ext/compressionparams.c).  I am using level 19 with the assumption that it is always on with that level...

from https://github.com/facebook/zstd/issues/2832
uncompressed.bin https://gist.github.com/thatch/853de8089e17462f2e500a7f87c76736
zstd -19: 179956 bytes
trivial subdivision in python: 129343 bytes

from https://github.com/facebook/zstd/issues/3014
test.bin https://gist.github.com/thatch/c2c2cc7304208423543edc38f2b7ed76
zstd -19: 2688424 bytes
trivial subdivision in python: 1021806 bytes

**Expected behavior**
Ratio to be as good (or a little better) than my trivial code.

For trivial random input with blocks of known entropy, it does appear to work.  I have some evidence that this affects ELF files, which isn't super surprising because they likely have 4k-aligned sections.

terrelln:
Thanks for the report! We know our block splitter is pretty trivial right now, and hope to find some time to work on it in the future. This case in particular is probably an interaction between block splitting, and the optimal parser, that tries to make decisions based on what it estimates the costs of particular matches are. So the block splitting changes the costs, but we don't re-parse based on the newly split blocks.

> trivial subdivision in python: 1021806 bytes

How do you sub-divide in Python? I want to make sure that we can exactly reproduce both the good & bad cases, so we can investigate exactly what is going on to cause this huge regression.

thatch:
> How do you sub-divide in Python?

Not very easily, unfortunately.  See https://gist.github.com/thatch/c2e8aa3d34a7a548687f43813f9cb788

Usage: `python compress_subdivide.py -19 input_filename.bin`

After reading 128KB, it tries that as well as 64+64, and 32+32+32+32, etc (uniformly sized within a 128KB block).  In initial testing there appeared to be a single minimum [sizes appeared to look like concave up], hence the "break" which avoids the worst of the performance penalty.

[...]

PR #3376, opened 2022-12-18
Block splitter : minor reformatting

Cyan4973:
After reading #3204, I wanted to have a look at the block splitter to understand what's going on.
This exercise lead to a (rather large) number of minor changes, which initially were cosmetic only, but eventually added a few correctness updates. They don't change the behavior of the functionality, but hopefully improve maintenance.

List of changes: 
- Shortened many long lines
- Conversion from `offset` to `offbase` was done manually, instead of employing the dedicated macros
- `const` correctness : `seqStore` was passed as a mutable buffer in scenarios where it was read only. The pb ran multiple levels deep, so several functions where controlled and their prototype updated to reflect this property.
- Several variables and pointers could be made `const`, improving the understanding of what happens to them. Some variables could be removed, others could be moved to a more restrictive scope
- Several fixes for `-Wconversion` warning level (note: not activated by default on `zstd` code base)
- Some `assert()` were added, to better document what's the expected state of the system.
- Added a few code comments

Only minor stuff, but I figure it was still valuable enough for maintenance to be merged.
None of these changes impact the behavior of the program. The source code is stricter, but it still does the same thing.

A side effect of this exercise is that I have a better understanding of the block splitter now.

The estimation logic, which judges if a splitting decision is beneficial or not, looks fine and seems generally re-usable.

The splitting decision though is still pretty raw, blindly mechanical.
Hence, that's where improvements could be inserted.

That being said, adding complexity at this stage is also likely going to impact speed.
So it might become desirable to feature multiple "levels" of block splitting.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -2496,11 +2496,13 @@ typedef struct {
  * entropyWkspSize must be of size at least ENTROPY_WORKSPACE_SIZE - (MaxSeq + 1)*sizeof(U32)
  */
 static ZSTD_symbolEncodingTypeStats_t
-ZSTD_buildSequencesStatistics(seqStore_t* seqStorePtr, size_t nbSeq,
-                        const ZSTD_fseCTables_t* prevEntropy, ZSTD_fseCTables_t* nextEntropy,
-                              BYTE* dst, const BYTE* const dstEnd,
-                              ZSTD_strategy strategy, unsigned* countWorkspace,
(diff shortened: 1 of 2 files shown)


terrelln:
> The splitting decision though is still pretty raw, blindly mechanical.
> Hence, that's where improvements could be inserted.

Agreed, we definitely need better splitting logic, which can hopefully be (mostly) compression-algorithm agnostic.

terrelln (approved):

PR #3375, opened 2022-12-18
Try to build the contrib directory in Windows CI

eli-schwartz:
This is just going to fail. PR submitted as a demonstration. The logs look like this:
```
[82/90] Compiling C++ object contrib/pzstd/pzstd.exe.p/.._.._.._.._contrib_pzstd_Pzstd.cpp.obj
FAILED: contrib/pzstd/pzstd.exe.p/.._.._.._.._contrib_pzstd_Pzstd.cpp.obj 
"cl" "-Icontrib\pzstd\pzstd.exe.p" "-Icontrib\pzstd" "-I..\build\meson\contrib\pzstd" "-I..\programs" "-I..\contrib\pzstd" "-I..\lib" "-I..\lib\common" "-I..\lib\compress" "-I..\lib\decompress" "-I..\lib\dictBuilder" "-I..\lib\legacy" "-DNDEBUG" "/MD" "/nologo" "/showIncludes" "/utf-8" "/Zc:__cplusplus" "/W4" "/EHsc" "/std:c++14" "/permissive-" "/O2" "/Gw" "/D_UNICODE" "/DUNICODE" "/MP" "/Fdcontrib\pzstd\pzstd.exe.p\.._.._.._.._contrib_pzstd_Pzstd.cpp.pdb" /Focontrib/pzstd/pzstd.exe.p/.._.._.._.._contrib_pzstd_Pzstd.cpp.obj "/c" ../contrib/pzstd/Pzstd.cpp
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(124): error C2589: '(': illegal token on right side of '::'
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(126): note: see reference to class template instantiation 'pzstd::Range<Iter>' being compiled
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(124): error C3878: syntax error: unexpected token '(' following 'expression'
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(124): note: error recovery skipped: '( ( identifier'
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(124): error C2760: syntax error: ')' was unexpected here; expected ';'
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(124): error C3878: syntax error: unexpected token ')' following 'jump_statement'
D:\a\zstd\zstd\contrib\pzstd\utils/Range.h(124): note: error recovery skipped: ') <'
[...]

PR #3380, opened 2022-12-19
[pzstd] Fixes for Windows build

terrelln:
* Add `Portability.h` to fix min/max issues
* Fix conversion warnings
* Assert that windowLog <= 23, which is currently always the case.
  This could be loosened, but we aren't looking to add new functionality.

Fixes on top of PR https://github.com/facebook/zstd/pull/3375 by @eli-schwartz, which added Windows CI for contrib & programs.

### DIFF
--- a/contrib/pzstd/utils/Portability.h
+++ b/contrib/pzstd/utils/Portability.h
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) 2016-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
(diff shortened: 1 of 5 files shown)


terrelln:
Looks like CI is passing now.

contrib/pzstd/utils/Portability.h
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+#pragma once
+
+#include <algorithm>
+
+// Required for windows, which defines min/max, but we want the std:: version.
+
+#ifdef min
+#undef min
```
Cyan4973, line 15: very minor : 
I believe you can `#undef` unconditionally, whether `min` is already defined or not.
This is explicitly allowed by the standard,
and I believe no compiler warning will be triggered.

Cyan4973 (approved):

PR #3381, opened 2022-12-20
[docs] Clarify dictionary loading documentation

### LINKED ISSUE #3246: Discrepancy in manual for dictionary loading
In `zstd_manual.html:872`, describing `ZSTD_DCtx_loadDictionary`:
> The dictionary remains valid for all future frames, until explicitly invalidated

However, in `ZSTD_CCtx_loadDictionary_advanced` (`zstd_compress.c:1123` which `ZSTD_DCtx_loadDictionary` wraps around), this seems not to be true:
``` cpp
size_t ZSTD_CCtx_loadDictionary_advanced(
        ZSTD_CCtx* cctx, const void* dict, size_t dictSize,
        ZSTD_dictLoadMethod_e dictLoadMethod, ZSTD_dictContentType_e dictContentType)
{
    RETURN_ERROR_IF(cctx->streamStage != zcss_init, stage_wrong,
                    "Can't load a dictionary when ctx is not in init stage.");
    DEBUGLOG(4, "ZSTD_CCtx_loadDictionary_advanced (size: %u)", (U32)dictSize);
    ZSTD_clearAllDicts(cctx);  /* in case one already exists */
```

It seems that if adictionary is already used within a context, when loading a new one, the previous dictionary gets cleared from the context. I assumed that it would be retained in the context based on the manual, since by loading the new dictionary, I am not explicitly invalidating the old dictionary. Am I correct in thinking this?

terrelln:
Reinforce that loading a new dictionary clears the current dictionary. Except for the multiple-ddict mode.

Fixes #3246.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -990,8 +990,9 @@ ZSTDLIB_API unsigned ZSTD_getDictID_fromFrame(const void* src, size_t srcSize);
  * @result : 0, or an error code (which can be tested with ZSTD_isError()).
  *  Special: Loading a NULL (or 0-size) dictionary invalidates previous dictionary,
  *           meaning "return to no-dictionary mode".
- *  Note 1 : Dictionary is sticky, it will be used for all future compressed frames.
- *           To return to "no-dictionary" situation, load a NULL dictionary (or reset parameters).
+ *  Note 1 : Dictionary is sticky, it will be used for all future compressed frames,
+ *           until parameters are reset, a new dictionary is loaded, or the dictionary
+ *           is explicitly invalidated by loading a NULL dictionary.
  *  Note 2 : Loading a dictionary involves building tables.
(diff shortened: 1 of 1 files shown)


lib/zstd.h
```diff
@@ -1039,9 +1040,9 @@ ZSTDLIB_API size_t ZSTD_CCtx_refPrefix(ZSTD_CCtx* cctx,
                                  const void* prefix, size_t prefixSize);
 
 /*! ZSTD_DCtx_loadDictionary() : Requires v1.4.0+
- *  Create an internal DDict from dict buffer,
- *  to be used to decompress next frames.
- *  The dictionary remains valid for all future frames, until explicitly invalidated.
+ *  Create an internal DDict from dict buffer, to be used to decompress al future frames.
```
Cyan4973: nit: al => all

Cyan4973 (approved):

PR #3384, opened 2022-12-20
Pin actions/checkout Dependency to Specific Commit Hash

felixhandte:
It's a bit silly, because if we can't trust GitHub, what are we doing here? But OSSF complains about it, so let's fix it.


Cyan4973:
Why would the "versions-compatibility" test fail on this PR ?
Weird, I assume it's unrelated.

terrelln:
It's unrelated, I'm looking into it. Not really sure why it is failing now.

terrelln (approved):

Issue #3383, opened 2022-12-20
Improve help/usage (`-h`) formatting


jonpalmisc:
**Is your feature request related to a problem? Please describe.**
The help output is rather difficult to parse visually and is somewhat unconventional compared to many other programs. I think giving it a makeover would be beneficial.

**Describe the solution you'd like**
I'd like to see the help output rewritten/reformatted so that it is easier to read and parse.

**Describe alternatives you've considered**
I could be more tolerant of the current output.

**Additional context**
I'm actually interested in tackling this problem myself, but wanted to create an issue first to:
- see if this is something others are interested in; and
- find out if the current help formatting is "set in stone" or not open to changes before I get ahead of myself.

Please advise&mdash;if changes to the help output are welcome, I'd happily PR changes that I think are reasonable.

Cyan4973:
Yes, this is a good idea.

jonpalmisc:
I will start working on these changes. Thanks!

Cyan4973:
#3385 merged

PR #3385, opened 2022-12-20
Improve help/usage (`-h`, `-H`) formatting

jonpalmisc:
Per our brief discussion in #3383, I've gone ahead and re-formatted the CLI help output. For convenience, I've attached text files containing the before/after of the help output.

Some main points with respect to the help formatting are:

- Flag/option descriptions now use proper capitalization and punctuation.
- Flag/options descriptions are now consistently aligned and formatted.
- Some non-technical references to 'zstd' have be rewritten to 'Zstandard'; the rationale here is that this output is for humans, and in instances where the compression algorithm rather than the CLI tool itself is being referred to, I think this makes sense.
- Some flag/option descriptions have been rephrased slightly (or significantly) to what I think is better phrasing. Once again, this is opinionated and I'm happy to discuss changes.
- Probably some more things I'm forgetting.

I've also cleaned up the "welcome message" a bit in a separate commit; I expect this is an opinionated change and I did it in a separate commit so it can be easily reverted.

Please let me know your thoughts.

[ZstdAfter.txt](https://github.com/facebook/zstd/files/10272548/ZstdAfter.txt)
[ZstdBefore.txt](https://github.com/facebook/zstd/files/10272549/ZstdBefore.txt)

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -52,12 +52,12 @@
 /*-************************************
 *  Constants
 **************************************/
-#define COMPRESSOR_NAME "zstd command line interface"
+#define COMPRESSOR_NAME "Zstandard CLI"
 #ifndef ZSTD_VERSION
 #  define ZSTD_VERSION "v" ZSTD_VERSION_STRING
 #endif
 #define AUTHOR "Yann Collet"
-#define WELCOME_MESSAGE "*** %s %i-bits %s, by %s ***\n", COMPRESSOR_NAME, (int)(sizeof(size_t)*8), ZSTD_VERSION, AUTHOR
+#define WELCOME_MESSAGE "*** %s (%i-bit) %s, by %s ***\n", COMPRESSOR_NAME, (int)(sizeof(size_t)*8), ZSTD_VERSION, AUTHOR
 
 #define ZSTD_ZSTDMT "zstdmt"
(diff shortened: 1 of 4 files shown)


yoniko:
Hi @jonpalmisc,
I've only given it a brief look but it looks great from what I've seen.
Please note that there's a test that checks the help message and it's failing, so this PR will need to make the same changes to the expected help message in the tests, please see `basic/help.sh*` files for the test's details.

Cyan4973:
I agree, this is a good update.

@jonpalmisc , if you have issues with the testing logic (which must be fixed before merge), 
I could also take over from here and modify the tests so they correspond to the new output format.

jonpalmisc:
Hey guys, thanks for the feedback; glad to hear these changes are a step in the right direction. I'll take a look at fixing that test.

jonpalmisc:
Looks to me like those two tests that were failing before pass now. Please let me know if you spot any other issues.

I see the 'versions-compatibility' test is still failing but it seems like that is failing on most recent PRs; given that wasn't mentioned earlier, I assume that's a known/irrelevant issue?

Cyan4973:
Yes, that's unrelated.
Thanks @jonpalmisc !

Cyan4973 (approved):

PR #3387, opened 2022-12-21
[wip] Try to debug version test

terrelln:
No fix yet, just trying to debug what is going on.

### DIFF
--- a/tests/test-zstd-versions.py
+++ b/tests/test-zstd-versions.py
@@ -20,6 +20,7 @@ import sys
 import subprocess
 from subprocess import Popen, PIPE
 
+
 repo_url = 'https://github.com/facebook/zstd.git'
 tmp_dir_name = 'tests/versionsTest'
 make_cmd = 'make'


terrelln:
This is definitely a bug in `v0.7.3`. Though I don't quite know whats going on yet. I will look into it a little bit more, then try just skipping `v0.7.3` in our tests.

Cyan4973:
`v0.7.3` could very well be buggy, but it's surprising that its associated test only starts to fail now.
The test was expected to remain the same.
What could have changed recently ?

terrelln:
On versions 0.7.3 - 0.8.0 (and possibly higher, I haven't tested yet), this line is failing:

https://github.com/facebook/zstd/blob/75ed1a815ef9695a2ba5afeb7f8e3f06d9c732a6/programs/fileio.c#L329

The error is: `Operation not authorized at current processing stage`

E.g. https://github.com/facebook/zstd/actions/runs/3752019678/jobs/6373711031

Cyan4973:
`ZBUFF_compressInit_advanced()` could fail,
for example, it could fail allocation.

That being said, the error message is more specific, 
so it should not be an allocation error.

Inside this function, `ZSTD_compressBegin_advanced()` is invoked,
which itself invokes `ZSTD_compress_insertDictionary()`,
which is active in the failing scenario.
Yet, I fail to see any of these functions being able to trigger the `stage_wrong` error...

terrelln:
well it is randomly passing again, so I guess I'll close this PR

PR #3388, opened 2022-12-21
Proposed modifications for level 1 to 4

skandau:
improve compression ratio on level 1-4
[zstd-1.5.2.zip](https://github.com/facebook/zstd/files/10278176/zstd-1.5.2.zip)


Cyan4973:
Hi @skandau 

My understanding is that your work is said to improve levels 1-4 on large files,
both compression ratio and speed.

I would expect that to mean that the impacted files are `zstd_fast.c` and `zstd_dfast.c` (primarily).

However, none of these files are updated in this PR.
Is it your expectation ? Or is it a side effect of unfamiliarity of the `diff` process ?

Cyan4973:
Btw, even though the term `kernel` is employed, and there is a file in the `linux-kernel` directory which is updated,
I believe it's a misnomer, and this PR is effectively unrelated to the Linux Kernel topic.
Therefore, I can take it over from here.

Cyan4973:
For clarity :
the PR itself is bogus.

The actual modification to review is in the zip file, which contains a modified version of zstd v1.5.2.

Cyan4973:
@skandau : 
I'm looking at your proposed change, in [`zstd-1.5.2.zip`](https://github.com/facebook/zstd/files/10278176/zstd-1.5.2.zip).

Focusing first on `zstd_fast.c`, which is used for levels 1 and 2.
The file must be modified to fit into `v1.5.4`, but that's not too complex to do.

Your initial claim on encode.su is that your modification improves both compression ratio and speed.

From what I can observe, this is not the benchmarked result.
Instead, I observe a very small compression ratio increase, but at a relatively important compression speed loss.
Measured on a M1 Pro laptop using `silesia.tar`:
Level 1 : (dev) 605 MB/s => 540 MB/s (your modification)
Level 2 : (dev) 460 MB/s => 420 MB/s (your modification)

So that's a ~-10% speed cost. Compression ratio is improved, but by a very small amount (<1%).
To say it bluntly, this is a not a worthwhile trade off.

Analyzing the source code, the only impactful modification worth mentioning is 
essentially pushing more reference points into the hash table, after each match,
which translates into increased opportunities to find matches (aka higher compression ratio)
but also increased workload (aka lower compression speed).
So the result seems completely logical.

Looking again at your claims on encode.su,
it seems you have released multiple variations,
from v1 to v5.
I assume you have shipped the latest one in your message, aka v5.
Reading results displayed at : https://encode.su/threads/2119-Zstandard?p=77456&viewfull=1#post77456 ,
it seems v3 would be a much better trade off.
It claims improving both compression ratio **and speed**,
while later variants v4 and then v5 lean more heavily towards compression ratio at the expense of speed.

So I would suggest shipping v3 for evaluation, as it may feature a more desirable trade off.

skandau:
hhmmmm here is zstd152modif_v6..... i think the compression speed is fast like v3 in encode.su
[zstd-1.5.2modif_v6.zip](https://github.com/facebook/zstd/files/10289722/zstd-1.5.2modif_v6.zip)

Cyan4973:
I tested the v6 variant,
and results were similar to initial proposal, if only marginally faster.

Looking at the code, the most important modification (inserting more positions into the hash table) looks identical,
so it's no surprise that the speed isn't changed much.

Not sure if v6 is really comparable to v3,
but from the benchmarks, the performance of v6 is really close to v5.

Not sure either what v3 was nor if you can reproduce v3.

skandau:
i have limited inserting more positions into the hash table so the speed should be faster than v5 or v6.
[zstd152_v3.zip](https://github.com/facebook/zstd/files/10291019/zstd152_v3.zip)

Cyan4973:
Indeed, this `v3` version looks much better, at least when analyzing the `fast` mode.

The compression speed of this variant is similar to the current `dev` branch, so this seems to be a neutral axis.

Compression ratio is improved, by a very little bit. Nothing huge, but still, that's on the positive side.

This is achieved, once again, by updating the hash table with more references, although the number of pushed references is much lower in this v3 variant, hence the speed cost is also much lower, to the point of being almost negligible if benchmark measurements are correct.

So this is a variant that's worth considering, I'll see what can be done (especially make sure I understand and can reproduce the same changes, and that they don't introduce any out-of-bound scenarios).

On the other side, I would highly recommend you to rebase your work on the current `dev` branch.
This would make your work more in line with current code base.
It shouldn't be too difficult, as long as you remember what are the changes you made.

The one last thing I'm curious about is that your initial claim was that this version was faster than reference `v1.5.2`, unfortunately I couldn't measure any significant speed difference on my laptop. That being said, there are many variations that could explain such a difference, from the compiler version, to the compiler flag, to the exact cpu type being used, to a different reference, etc.
I would be curious to know if you can reproduce the speed gains claimed vs `v1.5.2` when modifications are rebased from `dev`.

skandau:
I have compiled zstd1.5.2 using zstd.sln that has reserved in build folder and i have used zstd.exe from zstd152-win-x64.zip. I will attach binary of zstd152 original and modification.

skandau:
The compression ratio on webster file is about ~300kb compared to original I think it is big result... Hhmmm

Cyan4973:
> The compression ratio on webster file is about ~300kb compared to original I think it is big result... Hhmmm

I'm getting ~175 KB on `webster` at level 2,
that's still good, just less than 300 KB.

However, it doesn't work so well at level 1,
where it loses by 40 KB to `dev` branch on `webster`.

Generally speaking, the "trick" of pushing more information into the hash table
works better at level 2 than level 1,
by virtue of level 2 featuring a much larger hash table.

Unfortunately, both levels use the same algorithm,
we have to find a common policy workable for both of them.

skandau:
What is the meaning of numbers in clevel.h? How about if we set the equal number for level 1 and level 2 ? I guess we can get smaller size for level 1 too.

skandau:
Or how about if we set larger the size of hash table for level 1 ?

skandau:
[zstd152_v3a.zip](https://github.com/facebook/zstd/files/10295132/zstd152_v3a.zip)
[zstd152ori.zip](https://github.com/facebook/zstd/files/10295137/zstd152ori.zip)
here is the binary of zstd152 original and zstd152v3a. @cyan please run and timer it on each files in silesia corpus. i have improved level 1 too so it compress better compared to zstd152 original version by changing the number in clevel.h

skandau:
here is the compression time between zstd152ori and zstd152v3a in my laptop when compress webster file.
zstd152oriVszstd152v3a

skandau:
i just changed one number from 7 to 6 in clevels.h for level 1

skandau:
i just changed column L from 7 to 6 for level 1, in clevels.h and the result is surprisingly. i attach the source code of zstd152v3a
[zstd152v3a.zip](https://github.com/facebook/zstd/files/10298556/zstd152v3a.zip)

skandau:
@Cyan4973 I accidentally press readytoreview button. How to solve this?

skandau:
just experimenting with clevels.h and zstd_lazy.c, i have improved level 5 upto 13 but level 14 and 15 is worsen. here is zstd152_v3b
[zstd152_v3b_source.zip](https://github.com/facebook/zstd/files/10299383/zstd152_v3b_source.zip)
[zstd152_v3b_binary.zip](https://github.com/facebook/zstd/files/10299384/zstd152_v3b_binary.zip)

skandau:
here is zstd152_v3c binary and source that can compress better than zstd152 original from level 1 to level 7 (level 5 - 7 modify clevels.h, levels 1 modify clevels.h and zstd_fast.c, level 2 modify zstd_fast.c, level 3 and 4 modify zstd_doublefast.c). i attach too the compression result on level 1, level 5, level 6, and level 7 in .txt file
[zstd152v3c_source.zip](https://github.com/facebook/zstd/files/10299638/zstd152v3c_source.zip)
[zstd152v3c_binary.zip](https://github.com/facebook/zstd/files/10299640/zstd152v3c_binary.zip)
[testsilesia_ori_vs_v3c.txt](https://github.com/facebook/zstd/files/10299642/testsilesia_ori_vs_v3c.txt)

skandau:
this is a benchmark result zstd152 ori with zstd152v3c using built in benchmark on zstandard
 
benchmarkzstd152ori_vs_zstd152v3c
level 2 and level 1 zstd152v3c compresion speed is faster than original, but level 3-7 is slower than original.

PR #3392, opened 2022-12-21
[cmake] Add noexecstack to compiler/linker flags

### LINKED ISSUE #3125: Test playTests fails on ppc64 (openSUSE Tumbleweed)
I was trying to use cmake for building the RPM for openSUSE and the test `playTests` fails on ppc64. Here is the [log of the test](https://github.com/facebook/zstd/files/8645346/test_ppc64.log) and here is the [full build log](https://github.com/facebook/zstd/files/8645339/build_ppc64.log.txt), from https://build.opensuse.org/package/live_build_log/home:jaimeMF:branches:Archiving/zstd/openSUSE_Factory_PowerPC/ppc64 . The test pass on  ppc64le and ppc ( https://build.opensuse.org/package/show/home:jaimeMF:branches:Archiving/zstd ).

I don't have a ppc machine, but I algo get the error using qemu.

terrelln:
Add `noexecstack` cflags and ldflags to cmake. We already use these flags in our Makefile starting in PR #2964.

This is needed to pass PPC64 tests with ELFv1, because gcc will omit this note by default, unless the noexecstack flags are passed. That means that zstd on ppc64 wasn't broken before this PR, just our test was. This makes cmake a bit more zealous in setting noexecstack.

Fixes #3125.

### DIFF
--- a/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
+++ b/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
@@ -1,7 +1,8 @@
 include(CheckCXXCompilerFlag)
 include(CheckCCompilerFlag)
+include(CheckLinkerFlag)
 
-function(EnableCompilerFlag _flag _C _CXX)
+function(EnableCompilerFlag _flag _C _CXX _LD)
     string(REGEX REPLACE "\\+" "PLUS" varname "${_flag}")
     string(REGEX REPLACE "[^A-Za-z0-9]+" "_" varname "${varname}")
     string(REGEX REPLACE "^_+" "" varname "${varname}")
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
```diff
@@ -30,33 +38,39 @@ macro(ADD_ZSTD_COMPILATION_FLAGS)
         # EnableCompilerFlag("-std=c99" true false)   # Set C compiation to c99 standard
         if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND MSVC)
             # clang-cl normally maps -Wall to -Weverything.
-            EnableCompilerFlag("/clang:-Wall" true true)
+            EnableCompilerFlag("/clang:-Wall" true true flase)
```
jaimeMF, line 41: Shouldn't it be `false` instead of `flase`?

Cyan4973: Good catch @jaimeMF !

_edit_ : fixed in 6640377783e73211d1afd440550f8587dd9de75c

PR #3394, opened 2022-12-22
[cli-tests] Test file stat read/write

terrelln:
* Add `--trace-file-stat` to the `zstd` cli to trace reading/writing a file's metadata
* Add `--set-exact-output` to the `cli-tests` runner to aid in updating a `.exact` output
* Add tests using `--trace-file-stat`


Cyan4973 (approved):
Great new capability !

PR #3395, opened 2022-12-22
[tests] Remove deprecated function from longmatch.c test

terrelln:
Thanks to @eli-schwartz for pointing it out!

We should maybe consider adding a helper function for applying `ZSTD_parameters` and `ZSTD_compressionParameters` to a context. That would aid the transition to the new API in situations like this.


Cyan4973:
> We should maybe consider adding a helper function for applying `ZSTD_parameters` and `ZSTD_compressionParameters` to a context. That would aid the transition to the new API in situations like this.

Indeed, I thought we already had one. But maybe it's only used internally. To be checked...

Cyan4973 (approved):

PR #3398, opened 2022-12-22
spec update : require minimum nb of literals for 4-streams mode

### LINKED ISSUE #3316: Protocol question
when jump table field is exist what should be the smallest regenerated_size value? because if we take for example regenerated_size==5 than it could not split evenly in the 3 streams but regenerated_size 4 comply the spec formula.

what about regenerated_size=0?

Cyan4973:
Reported by @shulib :
the specification for 4-streams mode
doesn't work when the amount of literals to compress is 5 bytes. Extending it, it also doesn't work for sizes 1 or 2.

This patch updates the specification and the implementation to require a minimum of 6 literals to trigger or accept the 4-streams mode.

The impact is expected to be none over the ecosystem :
the 4-streams mode is never triggered for such small quantity of literals anyway, since it would be wasteful (it costs ~7.3 bytes more than single-stream mode). An informal lower limit is set at ~256 bytes,
so the technical minimum is very far from this limit.
On the other side, the decoder was already protected against scenario, and would already fail cleanly if presented with such request.

This is just meant for completeness of the specification.

Fix #3316

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -435,7 +435,7 @@ They can be decoded first, and then copied during [Sequence Execution],
 or they can be decoded on the flow during [Sequence Execution].
 
 Literals can be stored uncompressed or compressed using Huffman prefix codes.
-When compressed, an optional tree description can be present,
+When compressed, a tree description may optionally be present,
 followed by 1 or 4 streams.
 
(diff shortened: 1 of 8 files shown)


felixhandte (approved):
LGTM.

Do we also want to submit an erratum to the RFC?

Cyan4973:
> Do we also want to submit an erratum to the RFC?

Yes, that would be a good follow up.

PR #3403, opened 2022-12-28
ZSTD_CCtx_setCParams

Cyan4973:
Inspired by @terrelln  in #3395,
offer a new capability to set all parameters defined in a `ZSTD_compressionParameters` structure with a single symbol invocation to improve user code brevity.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -664,7 +664,7 @@ size_t ZSTD_CCtx_setParameter(ZSTD_CCtx* cctx, ZSTD_cParameter param, int value)
         if (ZSTD_isUpdateAuthorized(param)) {
             cctx->cParamsChanged = 1;
         } else {
-            RETURN_ERROR(stage_wrong, "can only set params in ctx init stage");
+            RETURN_ERROR(stage_wrong, "can only set params in cctx init stage");
     }   }
 
     switch(param)
@@ -1103,8 +1103,15 @@ size_t ZSTD_CCtx_setParametersUsingCCtxParams(
 
 size_t ZSTD_CCtx_setCParams(ZSTD_CCtx* cctx, ZSTD_compressionParameters cparams)
 {
-    FORWARD_IF_ERROR(ZSTD_checkCParams(cparams), "");
+    DEBUGLOG(4, "ZSTD_CCtx_setCParams");
     assert(cctx != NULL);
+    if (cctx->streamStage != zcss_init) {
+        /* All parameters in @cparams are allowed to be updated during MT compression.
+         * This must be signaled, so that MT compression picks up the changes */
+        cctx->cParamsChanged = 1;
+    }
+    /* only update if parameters are valid */
+    FORWARD_IF_ERROR(ZSTD_checkCParams(cparams), "");
     cctx->requestedParams.cParams = cparams;
     return 0;
 }
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -346,10 +346,13 @@ size_t ZSTD_CCtxParams_init(ZSTD_CCtx_params* cctxParams, int compressionLevel) 
 #define ZSTD_NO_CLEVEL 0
 
 /**
- * Initializes the cctxParams from params and compressionLevel.
+ * Initializes `cctxParams` from `params` and `compressionLevel`.
  * @param compressionLevel If params are derived from a compression level then that compression level, otherwise ZSTD_NO_CLEVEL.
  */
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1098,6 +1101,14 @@ size_t ZSTD_CCtx_setParametersUsingCCtxParams(
     return 0;
 }
 
+size_t ZSTD_CCtx_setCParams(ZSTD_CCtx* cctx, ZSTD_compressionParameters cparams)
+{
+    FORWARD_IF_ERROR(ZSTD_checkCParams(cparams), "");
+    assert(cctx != NULL);
```
embg, line 1107: Are [these checks](https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L1091-L1095) needed here?

lib/zstd.h
```diff
@@ -1740,17 +1740,22 @@ ZSTDLIB_STATIC_API size_t ZSTD_checkCParams(ZSTD_compressionParameters params);
...
 ZSTDLIB_STATIC_API
 size_t ZSTD_compress_advanced(ZSTD_CCtx* cctx,
-                                          void* dst, size_t dstCapacity,
-                                    const void* src, size_t srcSize,
-                                    const void* dict,size_t dictSize,
-                                          ZSTD_parameters params);
+                              void* dst, size_t dstCapacity,
+                        const void* src, size_t srcSize,
+                        const void* dict,size_t dictSize,
+                              ZSTD_parameters params);
```
embg, line 1758: Did you intend to update the indentation here?

embg (changes requested):
LGTM, just one small nit and one potential issue that might require changes (hence, I am requesting changes). Feel free to land if changes aren't required.

lib/zstd.h
```diff
@@ -1740,17 +1740,22 @@ ZSTDLIB_STATIC_API size_t ZSTD_checkCParams(ZSTD_compressionParameters params);
  *  This function never fails (wide contract) */
 ZSTDLIB_STATIC_API ZSTD_compressionParameters ZSTD_adjustCParams(ZSTD_compressionParameters cPar, unsigned long long srcSize, size_t dictSize);
 
+/*! ZSTD_CCtx_setCParams() :
+ *  Set all parameters provided within @cparams into the working @cctx
+ * @return 0 on success, or an error code (can be checked with ZSTD_isError()) */
```
embg, line 1745: It would be good to document that the user will lose out on some cParam validation: https://github.com/facebook/zstd/blob/89342d1e076e86e59d143ce2fa3b8e8af6f65b03/lib/compress/zstd_compress.c#L670-L944

lib/zstd.h
Cyan4973: Yes

lib/compress/zstd_compress.c
Cyan4973: Yes, some are,
the direct equivalent to reproduce would be : https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L659

And, if the same logic is present multiple times, it would be better to factor it into a common function.

_edit_ : actually, the logic is slightly different.
This method is updating a specific bunch of parameters,
and they all belong to a category which is allowed to change during MT compression.

Btw, more parameters should be allowed to change during MT compression, but that's a completely different topic. See #3406 for an optional follow-up.

embg (approved):

embg:
Hm... I just thought of a possible issue. You are assuming all the parameters in `ZSTD_compressionParameters` are authorized for mid-frame updates. However, isn't windowLog not authorized?

Cyan4973:
That's a very good point @embg .
`windowLog` can't change during compression.
And I actually assumed that windowlog's stickiness was taken care of somewhere later in the pipeline.

That's indeed the case : 
https://github.com/facebook/zstd/blob/dev/lib/compress/zstdmt_compress.c#L1072

That being said, I agree this situation is not clear for users, so adding a code comment to specify this behavior would be helpful.

embg:
> That's indeed the case : https://github.com/facebook/zstd/blob/dev/lib/compress/zstdmt_compress.c#L1072

Thanks for the pointer. Another question for my own understanding: how does this work for single-threaded compression? E.g. if ZSTD_CCtx_setCParams() is called in the middle of a frame, during streaming compression?

Cyan4973:
> Another question for my own understanding: how does this work for single-threaded compression? E.g. if ZSTD_CCtx_setCParams() is called in the middle of a frame, during streaming compression?

If (allowed) parameters are changed during single-threaded streaming compression, actually nothing changes for the current compression session.

Changes are merely registered into `requestedParameters`, while ongoing compression uses `appliedParameters`.

PR #3408, opened 2023-01-04
Deprecate advanced streaming functions

### LINKED ISSUE #3405: Mark v1.5.x deprecated functions as deprecated
We have a number of comments like 

> Note : this prototype will be marked as deprecated and generate compilation warnings on reaching v1.5.x

where the prototypes aren't actually marked as deprecated. For example: https://github.com/facebook/zstd/blob/90597d78eadadae7acf3633ab4d81ef7c4e6057f/lib/zstd.h#L2445-L2447

Filing an issue to remind myself to fix this.

embg:
These functions were scheduled to be deprecated in v1.5.x. Fixes #3405

@Cyan4973: should we deprecate `ZSTD_initDStream` as well?

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1656,7 +1656,8 @@ size_t ZSTD_initDStream_usingDict(ZSTD_DStream* zds, const void* dict, size_t di
 size_t ZSTD_initDStream(ZSTD_DStream* zds)
 {
     DEBUGLOG(4, "ZSTD_initDStream");
-    return ZSTD_initDStream_usingDDict(zds, NULL);
+    FORWARD_IF_ERROR(ZSTD_DCtx_reset(zds, ZSTD_reset_session_only));
+    return ZSTD_DCtx_refDDict(zds, NULL);
 }
 
 /* ZSTD_initDStream_usingDDict() :
@@ -1664,6 +1665,7 @@ size_t ZSTD_initDStream(ZSTD_DStream* zds)
  * this function cannot fail */
 size_t ZSTD_initDStream_usingDDict(ZSTD_DStream* dctx, const ZSTD_DDict* ddict)
 {
+    DEBUGLOG(4, "ZSTD_initDStream_usingDDict");
     FORWARD_IF_ERROR( ZSTD_DCtx_reset(dctx, ZSTD_reset_session_only) , "");
     FORWARD_IF_ERROR( ZSTD_DCtx_refDDict(dctx, ddict) , "");
     return ZSTD_startingInputLength(dctx->format);
@@ -1674,6 +1676,7 @@ size_t ZSTD_initDStream_usingDDict(ZSTD_DStream* dctx, const ZSTD_DDict* ddict)
  * this function cannot fail */
 size_t ZSTD_resetDStream(ZSTD_DStream* dctx)
 {
+    DEBUGLOG(4, "ZSTD_resetDStream");
     FORWARD_IF_ERROR(ZSTD_DCtx_reset(dctx, ZSTD_reset_session_only), "");
     return ZSTD_startingInputLength(dctx->format);
 }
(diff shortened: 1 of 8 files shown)


Cyan4973:
> @Cyan4973: should we deprecate `ZSTD_initDStream` as well?

`ZSTD_initDStream()` is part of the stable API,
I believe it's good to stay.

lib/decompress/zstd_decompress.c
```diff
@@ -1656,14 +1656,16 @@ size_t ZSTD_initDStream_usingDict(ZSTD_DStream* zds, const void* dict, size_t di
 size_t ZSTD_initDStream(ZSTD_DStream* zds)
 {
     DEBUGLOG(4, "ZSTD_initDStream");
-    return ZSTD_initDStream_usingDDict(zds, NULL);
+    FORWARD_IF_ERROR(ZSTD_DCtx_reset(zds, ZSTD_reset_session_only), "");
+    return ZSTD_DCtx_refDDict(zds, NULL);
```
terrelln: This changes the return value, it should be `ZSTD_startingInputLength(zds->format)`

tests/zstreamtest.c
```diff
@@ -126,6 +126,39 @@ static U32 FUZ_rand(U32* seedPtr)
...
+    {   const size_t refRes = ZSTD_DCtx_refDDict(zds, ddict);
+        if (ZSTD_isError(refRes)) return refRes;   }
+    return ZSTD_startingInputLength(dctx->format);
+}
+#define ZSTD_initDStream_usingDict(zds, dict, dictSize) (
+    ZSTD_initDStream_usingDict_helper(zds, dict, dictSize)
+)
+#define ZSTD_initDStream_usingDDict(zds, dict, dictSize) (
+    ZSTD_initDStream_usingDDict_helper(zds, ddict)
+)
```
terrelln: You're missing `\`, this shouldn't compile.

Also, you need to put parens around each use of a macro argument.

terrelln (changes requested):

tests/zstreamtest.c
embg: Oh, duh. Thanks.

embg:
Converting back to draft status to fix lz4-threadpool-libs test

embg:
Ready for review. I decided to `#define ZSTD_DISABLE_DEPRECATE_WARNINGS` in `zbuff.h` to suppress the warning for `ZBUFF_decompressInitDictionary()`.

I originally tried this:
```
 size_t ZBUFF_decompressInitDictionary(ZBUFF_DCtx* zbd, const void* dict, size_t dictSize)
 {
+    {   const size_t resetRet = ZSTD_DCtx_reset(zbd, ZSTD_reset_session_only);
+        if (ZSTD_isError(resetRet)) return resetRet;   }
+    {   const size_t loadRet = ZSTD_DCtx_loadDictionary(zbd, dict, dictSize);
+        if (ZSTD_isError(loadRet)) return loadRet;   }
+    return ZSTD_startingInputLength(zbd->format);
-    return ZSTD_initDStream_usingDict(zbd, dict, dictSize);
 }
```
The problem is that `ZBUFF_DCtx` is an opaque struct, so we can't access `zbd->format`. Rather than including non-public zstd headers to get the definition of the struct, I felt it's cleaner to just suppress the deprecation warning. I'm open to other solutions, though.

Cyan4973:
> Ready for review. I decided to `#define ZSTD_DISABLE_DEPRECATE_WARNINGS` in `zbuff.h` to suppress the warning for `ZBUFF_decompressInitDictionary()`.

`zbuff` itself is deprecated.
Do you need to `#define ZSTD_DISABLE_DEPRECATE_WARNINGS` in `zbuff.h` , or could it be defined in `zbuff.c` instead ?
This would avoid propagation of this constant.

embg:
> Do you need to `#define ZSTD_DISABLE_DEPRECATE_WARNINGS` in `zbuff.h` , or could it be defined in `zbuff.c` instead ? This would avoid propagation of this constant.

@Cyan4973 Good point, moved it to `zbuff_decompress.c`.

embg:
Rebased. There seems to be an unrelated CI issue with the mingw tests, will hold off landing this until I figure that out.

Cyan4973:
> Rebased. There seems to be an unrelated CI issue with the mingw tests, will hold off landing this until I figure that out.

The last PR to be merged was #3410. Could it be related ?

embg:
@cyan I confirmed that the CI failure is not related to the changes from this PR. (See https://github.com/facebook/zstd/issues/3420 for details). Since all other tests are passing, can I land?

Cyan4973 (approved):

terrelln (approved):

PR #3417, opened 2023-01-10
Fix fuzzing with ZSTD_MULTITHREAD

danlark1:
At Google we fuzz zstd without ZSTD_MULTITHREAD but we want inputs to be as much as reproducible. It allows us to test new fuzzing methods for our fuzz team internally and have more horsepower to find bugs


terrelln (approved):
Looks good to me, thanks @danlark1!

danlark1:
Don't merge for now, please :)

```cpp
    case ZSTD_c_nbWorkers :
#ifndef ZSTD_MULTITHREAD
        // Must be zero.
        RETURN_ERROR_IF(value!=0, parameter_unsupported, "not compiled with multithreading");
        return 0;
#else
        FORWARD_IF_ERROR(ZSTD_cParam_clampBounds(param, &value), "");
        CCtxParams->nbWorkers = value;
        return CCtxParams->nbWorkers;
#endif
```

**UPD** now should be good, thanks

terrelln (approved):

PR #3418, opened 2023-01-11
Fuzz on maxBlockSize

daniellerozenblit:
This PR extends our fuzz testing in response to a recently found bug in our repcode update logic (#3127). This bug occurred in scenarios where both `offset_1` and `offset_2` were invalid going into a block, and no matches were found. This bug was latent and luckily could not cause corruption, but we want to ensure that our fuzzer tests scenarios where both repcodes are invalidated to protect against these sorts of edge cases.

In order to test these scenarios, we want to generate cases with several blocks where block-size < window-size. It is currently not possible to do this in our non-streaming tests without generating large inputs. Testing with random flushing in streaming mode (which we currently do) should generate these scenarios, but we want to extend this to our non-streaming tests.

This PR includes a new cctx parameter, `maxBlockSize` which can be used to override the default `ZSTD_BLOCKSIZE_MAX` (128KB). This parameter is currently only meant to be used for testing purposes and is bounded by [1KB, `ZSTD_BLOCKSIZE_MAX`].


lib/compress/zstd_compress.c
```diff
@@ -288,6 +288,15 @@ static int ZSTD_resolveExternalSequenceValidation(int mode) {
 #endif
 }
 
+/* Sets the default maximum block size. */
+static size_t ZSTD_resolveMaxBlockSize(size_t maxBlockSize) {
```
Cyan4973, line 292: I presume you mean "defaultMaxBlockSize" ?

Cyan4973:
Can you imagine one (or several) test which would be able to check that `maxBlockSize` works as intended ?

lib/compress/zstd_compress.c
daniellerozenblit: I actually intended to follow the naming scheme of the previous "resolution" methods for the cctx parameters –– essentially just "resolving" the `maxBlockSize` value to the default if no value was present.

I think it's reasonable to keep the naming scheme for consistency, but the method annotation is definitely inaccurate. "Resolve maxBlockSize to the default if no value is present" might be more accurate?

Cyan4973: You're right, consistency matters, and I missed it was part of a larger list of `ZSTD_resolve*()` symbols.

OK for the updated annotation.

tests/zstreamtest.c
```diff
@@ -1925,6 +1925,49 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+        CHECK_Z(ZSTD_decompress(checkBuf, srcSize, dst2, size2));
+        CHECK(memcmp(src, checkBuf, srcSize) != 0, "Corruption!");
+
+        /* Compressed output should not be equal */
+        assert(memcmp(dst1, dst2, dstSize) != 0);
+        assert(size1 - size2 == 4); /* We add another RLE block with header + character */
+        ZSTD_freeCCtx(cctx);
+        free(checkBuf);
+    }
+    DISPLAYLEVEL(3, "OK \n");
```
Cyan4973, line 2025: :+1: Great !

One more test I would like to see is what happens when `maxBlockSize > windowSize`.
In such case, we want to be sure that the code will not produce blocks larger than `windowSize`.
I think your code is fine and will respect this condition.
The purpose of the test is to ensure it remains this way, a future modification could break this condition.

lib/compress/zstd_compress.c
```diff
@@ -964,6 +982,11 @@ size_t ZSTD_CCtxParams_setParameter(ZSTD_CCtx_params* CCtxParams,
         CCtxParams->enableMatchFinderFallback = value;
         return CCtxParams->enableMatchFinderFallback;
 
+    case ZSTD_c_maxBlockSize:
+        BOUNDCHECK(ZSTD_c_maxBlockSize, value);
```
terrelln: What happens if you set the value to `0`?

I think we can leave the bounds as-is, but also allow setting to zero to use the default maxBlockSize. And we should also call that out in the docs.

lib/zstd.h
```diff
@@ -2067,6 +2070,13 @@ ZSTDLIB_STATIC_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const vo
  * documentation (below) before setting this parameter. */
 #define ZSTD_c_enableMatchFinderFallback ZSTD_c_experimentalParam17
 
+/*  ZSTD_c_maxBlockSize
+ *
+ *  Default is ZSTD_BLOCKSIZE_MAX.
+ *
```
terrelln: Please add a little more detail here. Say what this parameter controls, what the min/max are (and why), and explain what setting to 0 means.

lib/zstd.h
```diff
@@ -141,6 +141,7 @@ ZSTDLIB_API const char* ZSTD_versionString(void);
 
 #define ZSTD_BLOCKSIZELOG_MAX  17
 #define ZSTD_BLOCKSIZE_MAX     (1<<ZSTD_BLOCKSIZELOG_MAX)
+#define ZSTD_BLOCKSIZE_MAX_MIN (1 << 10) /* The minimum valid max blocksize. Maximum blocksizes smaller than this make compressBound() inaccurate. */
```
terrelln: Please move this down below into the static linking only section

Putting this constant here means that it will always be present, and we don't like to do that until it has baked in the unstable section.

terrelln (commented):
Just one minor nit in the zstd.h header, otherwise looks good!

lib/compress/zstd_compress.c
```diff
@@ -1944,7 +1972,7 @@ static size_t ZSTD_resetCCtx_internal(ZSTD_CCtx* zc,
     }
 
     {   size_t const windowSize = MAX(1, (size_t)MIN(((U64)1 << params->cParams.windowLog), pledgedSrcSize));
-        size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, windowSize);
+        size_t const blockSize = MIN(params->maxBlockSize, windowSize);
```
terrelln, line 1976: Can you add an assert that `params->maxBlockSize != 0` up by the group of asserts on lines 1964-1966?

lib/compress/zstd_compress.c
```diff
@@ -1651,7 +1679,7 @@ size_t ZSTD_estimateCStreamSize_usingCCtxParams(const ZSTD_CCtx_params* params)
     RETURN_ERROR_IF(params->nbWorkers > 0, GENERIC, "Estimate CCtx size is supported for single-threaded compression only.");
     {   ZSTD_compressionParameters const cParams =
                 ZSTD_getCParamsFromCCtxParams(params, ZSTD_CONTENTSIZE_UNKNOWN, 0, ZSTD_cpm_noAttachDict);
-        size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, (size_t)1 << cParams.windowLog);
```
terrelln, line 1654: I think `params->maxBlockSize` could be equal to zero here, you have to resolve it.

lib/compress/zstd_compress.c
```diff
@@ -1546,10 +1573,11 @@ static size_t ZSTD_estimateCCtxSize_usingCCtxParams_internal(
         const size_t buffInSize,
         const size_t buffOutSize,
         const U64 pledgedSrcSize,
-        int useExternalMatchFinder)
+        int useExternalMatchFinder,
+        size_t maxBlockSize)
 {
     size_t const windowSize = (size_t) BOUNDED(1ULL, 1ULL << cParams->windowLog, pledgedSrcSize);
-    size_t const blockSize = MIN(ZSTD_BLOCKSIZE_MAX, windowSize);
+    size_t const blockSize = MIN(maxBlockSize, windowSize);
```
terrelln: I think `maxBlockSize` could be equal to zero here, you have to resolve it.

terrelln (changes requested):
Noticed some bugs we'll have to fix. They're not your fault, our `CCtxParams` have grown significantly, and now a lot of them need to be "resolved" from some default value. And this resolution has to happen in many different places. We need to refactor the cctx params, so we can centralize all the logic for resolving & copying parameters.

Can you add a test case for the estimation functions that sets the `maxBlockSize` to `0`? I think that we don't actually have any tests for these functions, because they should be failing right now.

You can add them to the test cases for the other static cctx functions here: https://github.com/facebook/zstd/blob/dev/tests/fuzzer.c#L1646

Basically you'd:
* Create some `ZSTD_CCtxParams`, using the default values
* Call `ZSTD_estimateCCtxSize_usingCCtxParams()`
* Pass a buffer of that size to `ZSTD_initStaticCCtx()`
* Do a compression & decompress the data to make sure it round trips
* Do the same thing but with `ZSTD_estimateCStreamSize_usingCCtxParams()`
* Do the same tests, but with an explicitly set `maxBlockSize`

You should write the test case before making the fix, so you can make sure that the test case exposes the bugs.

daniellerozenblit:
I've added some tests for `ZSTD_estimateCCtxSize_usingCCtxParams()` and `ZSTD_estimateCStreamSize_usingCCtxParams()` and have confirmed that they fail before resolving `maxBlockSize` in the respective methods.

tests/zstreamtest.c
```diff
@@ -1925,6 +1925,105 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+            /* maxBlockSize = 1KB, windowLog = 10 */
+            CHECK_Z(ZSTD_CCtx_setParameter(cctx, ZSTD_c_maxBlockSize, 1u << 10));
+            CHECK_Z(ZSTD_CCtx_setParameter(cctx, ZSTD_c_windowLog, 10));
+            size1 = ZSTD_compress2(cctx, dst1, dstSize, src, srcSize);
+
+            if (ZSTD_isError(size1)) goto _output_error;
+            CHECK_Z(ZSTD_decompress(checkBuf, srcSize, dst1, size1));
+            CHECK(memcmp(src, checkBuf, srcSize) != 0, "Corruption!");
+
+            /* maxBlockSize = 3KB, windowLog = 10 */
```
Cyan4973, line 1977: 👍

terrelln (approved):
Awesome!

PR #3419, opened 2023-01-11
fix root cause of #3416

### LINKED ISSUE #3416: dictionary_stream_round_trip corruption
**Describe the bug**

We found a fuzzer corruption with https://github.com/facebook/zstd/commit/5434de01e21672cdd3ac111a99a969d8c5079297

```
dictionary_stream_round_trip.c: 196: Assertion: `!FUZZ_memcmp(src, rBuf, size)' failed. Corruption!
```

Attaching a file and DEBUGLOG 10

This is the fuzzer file
[testcase-4882119846133760-6f37f712c5input.txt](https://github.com/facebook/zstd/files/10372501/testcase-4882119846133760-6f37f712c5input.txt)

This is DEBUGLOG 10 without 5434de01e21672cdd3ac111a99a969d8c5079297

[out_good.txt](https://github.com/facebook/zstd/files/10372543/out_good.txt)

This is DEBUGLOG 10 with 5434de01e21672cdd3ac111a99a969d8c5079297

[out_bad.txt](https://github.com/facebook/zstd/files/10372548/out_really_bad.txt)

**To Reproduce**
1. Downloads data attached
2. Run `CC=clang-15 CXX=clang++-15 python3 fuzz.py build dictionary_stream_round_trip`
3. `./dictionary_stream_round_trip testcase-4882119846133760-6f37f712c5input.txt`

We found this on our cluster but I was not able to reproduce it right away. For some reason fuzz data provider set param 1011 differently:

In the left picture is what I got from cluster, in the right what I got from local

2023-01-09-110350_1636x184_scrot

It's probably because of ZSTD_MULTITHREAD define. We fuzz it without it

After I set

```diff
--- a/tests/fuzz/zstd_helpers.c
+++ b/tests/fuzz/zstd_helpers.c
@@ -96,7 +96,7 @@ void FUZZ_setRandomParameters(ZSTD_CCtx *cctx, size_t srcSize, FUZZ_dataProducer
     setRand(cctx, ZSTD_c_nbWorkers, 0, 2, producer);
     setRand(cctx, ZSTD_c_rsyncable, 0, 1, producer);
 #endif
-    setRand(cctx, ZSTD_c_useRowMatchFinder, 0, 2, producer);
+    setRand(cctx, ZSTD_c_useRowMatchFinder, 2, 2, producer);
     setRand(cctx, ZSTD_c_enableDedicatedDictSearch, 0, 1, producer);
     setRand(cctx, ZSTD_c_forceMaxWindow, 0, 1, producer);
     setRand(cctx, ZSTD_c_literalCompressionMode, 0, 2, producer);
```

I was able to reproduce locally with clang-15 as well

**Expected behavior**
Fuzz passes

**Screenshots and charts**

 - OS: Linux
 - Compiler: clang-15
 - Flags: fuzz flags
 - Build system: Make

**Additional context**
Hope it reproduces right away

Cyan4973:
A minor update in 5434de0 changed a `<=` into a `<`, and as an indirect consequence allowed compression attempt of literals when there are only 6 literals to compress (previous limit was effectively 7 literals).

This is not in itself a problem, as the threshold is merely an heuristic, but it emerged a bug that has always been there, and was just never triggered so far due to the previous limit. This bug would make the literal compressor believes that all literals are the same symbol, but for the exact case where `nbLiterals==6`, plus a pretty wild combination of several exceptional conditions, this outcome could be false, resulting in data corruption.

Replaced the blind heuristic by an actual test for all limit cases. Even if the threshold is changed again in the future, the detection of Repeated Literal mode will remain reliable.

fix #3416

### DIFF
--- a/lib/compress/zstd_compress_literals.c
+++ b/lib/compress/zstd_compress_literals.c
@@ -24,9 +24,9 @@ static size_t showHexa(const void* src, size_t srcSize)
     const BYTE* const ip = (const BYTE*)src;
     size_t u;
     for (u=0; u<srcSize; u++) {
-        RAWLOG(6, " %02X", ip[u]); (void)ip;
+        RAWLOG(5, " %02X", ip[u]); (void)ip;
     }
-    RAWLOG(6, " \n");
+    RAWLOG(5, " \n");
     return srcSize;
 }
 
@@ -65,12 +65,26 @@ size_t ZSTD_noCompressLiterals (void* dst, size_t dstCapacity, const void* src, 
     return srcSize + flSize;
 }
 
+static int allBytesIdentical(const void* src, size_t srcSize)
+{
+    assert(srcSize >= 1);
+    assert(src != NULL);
+    {   const BYTE b = ((const BYTE*)src)[0];
+        size_t p;
+        for (p=1; p<srcSize; p++) {
+            if (((const BYTE*)src)[p] != b) return 0;
+        }
+        return 1;
+    }
+}
+
 size_t ZSTD_compressRleLiteralsBlock (void* dst, size_t dstCapacity, const void* src, size_t srcSize)
 {
     BYTE* const ostart = (BYTE*)dst;
     U32   const flSize = 1 + (srcSize>31) + (srcSize>4095);
 
-    (void)dstCapacity;  /* dstCapacity already guaranteed to be >=4, hence large enough */
+    assert(dstCapacity >= 4); (void)dstCapacity;
+    assert(allBytesIdentical(src, srcSize));
 
     switch(flSize)
     {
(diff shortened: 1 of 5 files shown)


lib/compress/zstd_compress_literals.c
```diff
@@ -172,9 +187,10 @@ size_t ZSTD_compressLiterals (
             return ZSTD_noCompressLiterals(dst, dstCapacity, src, srcSize);
     }   }
     if (cLitSize==1) {
-        ZSTD_memcpy(nextHuf, prevHuf, sizeof(*prevHuf));
-        return ZSTD_compressRleLiteralsBlock(dst, dstCapacity, src, srcSize);
-    }
+        if ((srcSize >= 8) || allBytesIdentical(src, srcSize)) {
```
terrelln, line 195: Please add a comment as to why this check is necessary.

terrelln (approved):
Can you please add a test case that would trigger this bug?

The fix LGTM

Cyan4973:
> Can you please add a test case that would trigger this bug?
> 
> The fix LGTM

This is a really hard test to define.
We would probably need some golden files, both for the content and the dictionary.
None of them are naturally exported from the fuzzer (only the sequence of bits which generate those artifacts).

terrelln:
> This is a really hard test to define.
> We would probably need some golden files, both for the content and the dictionary.
> None of them are naturally exported from the fuzzer (only the sequence of bits which generate those artifacts).

Ok, it'd be great if we had it, but if it is too hard, we have the fuzzers, and we know that they will quickly catch any regressions.

lib/compress/zstd_compress_literals.c
Cyan4973: done

PR #3422, opened 2023-01-13
[tests] Fix version test determinism

terrelln:
The dictionary source files were taken from the `dev` branch before this
commit, which could introduce non-determinism on PR jobs. Instead take
the sources from the PR checkout.

This PR also adds stderr logging, and verbose output for the jobs that
are failing, to help catch the failure if it occurs again.

### DIFF
--- a/tests/test-zstd-versions.py
+++ b/tests/test-zstd-versions.py
@@ -29,8 +29,19 @@ test_dat_src = 'README.md'
 test_dat = 'test_dat'
 head = 'vdevel'
 dict_source = 'dict_source'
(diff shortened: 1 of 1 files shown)


terrelln:
Alright, I've run the version test ~250 times and have seen no failures. It seems to be either consistently passing, or consistently failing. So I will resume debugging if it starts failing again on PRs, and hope I can repro it then.

terrelln:
Still haven't been able to reproduce the failure, but this PR now makes the tests more deterministic, and should help us debug if the failure shows up again.

Cyan4973 (approved):

Issue #3240, opened 2022-08-10
LLVM mingw bug AddressSanitizer: use-after-poison on address


calvin2021y:
**Describe the bug**
when build with https://github.com/mstorsjo/llvm-mingw and CMAKE,  zstd make app crash with `AddressSanitizer: use-after-poison on address`

**To Reproduce**
1.  download llvm mingw for ubuntu
2.  build zstd with cmake
3.  create libuv thread and use ZSTD_compress
4. app late crash with `AddressSanitizer: use-after-poison on address`

remove the only one function call from `ZSTD_compress` fix the problem.

**Expected behavior**
expect it work without crash.

**Desktop (please complete the following information):**
 - OS: win10
 - Compiler llvm mingw clang
 - Flags  `-O1 -g`
 - Build system  cmake  `-DBUILD_SHARED_LIBS=OFF -DZSTD_BUILD_SHARED=OFF -DZSTD_MULTITHREAD_SUPPORT=OFF -DZSTD_BUILD_PROGRAMS=OFF `

calvin2021y:
remove ZSTD, or change ZSTD into lz4 both fix the problem.

embg:
Thanks for the report! Could you share the exact code you're using to call `ZSTD_compress`? (You could create a GitHub Gist of that code and share that link with me).

calvin2021y:
https://gist.github.com/calvin2021y/955910b52c46773a846dcad856a01fcd

felixhandte:
@calvin2021y, can you share a little more about this issue? I don't see `-fsanitize=address` in the list of arguments you're providing to the build.

PR #3424, opened 2023-01-13
Disable Custom ASAN/MSAN Poisoning on MinGW Builds

felixhandte:
Addresses #3240.

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -292,7 +292,7 @@
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


felixhandte:
As discussed, I'm not convinced this is a real issue. Hopefully @calvin2021y will respond with more information.

terrelln (approved):
Looks correct to me, I'll let you decide whether you want to merge or not

PR #3429, opened 2023-01-17
Fix MSYS2 symlink issues in CI

### LINKED ISSUE #3420: mingw-short-test failing on dev
Clang mingw jobs are currently failing on dev ([example job](https://github.com/facebook/zstd/actions/runs/3903773171/jobs/6670357664)). I'm investigating here: https://github.com/embg/zstd/pull/51. It may be related to https://github.com/facebook/zstd/pull/3410, however all jobs (including mingw-short-test) passed on that PR before it landed. So I'm not sure yet what's broken.

Please ignore mingw-short-test until I land a fix.

embg:
Recently [some mingw jobs](https://github.com/embg/zstd/actions/runs/3915429703/jobs/6693598399#step:7:4) have failed because they are running with 
```
env:
    MSYS: winsymlinks:nativestrict
```
I'm not sure which upstream dependency is inserting `winsymlinks:nativestrict`, since it's not in our job spec. But it breaks `make check`, because it instructs MSYS2 to emulate symlinks in a way that only works when running as Administrator, or if Window's "developer mode" has been enabled. Our script runs with a lower privilege level and can't open those symlinks, which causes `make check` to fail.

This PR has a couple cosmetic changes, but the functional change is
```
-   make check
+   MSYS="" make check
```
Setting `MSYS=""` instructs MSYS2 to replace symlinks with file copies. That is, `ln -s` becomes `cp`. This is the default on MSYS2 and all our jobs ran with `MSYS=""` prior to the recent change coming from upstream.

If we want to test "real" symlinks, we would need to run the job as Administrator, or enable "developer mode". I have [reached out](https://github.com/orgs/community/discussions/44268) to see if this is possible, but for now I think `MSYS=""` is an acceptable trade-off.

More information on MSYS2 symlink support: https://dev.to/hakonhagland/handling-of-symlinks-on-windows-perl-msys2-cygwin-52h3

Closes https://github.com/facebook/zstd/issues/3420

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -384,6 +384,7 @@ jobs:
   mingw-short-test:
     runs-on: windows-latest
     strategy:
+      fail-fast: false  # 'false' means Don't stop matrix workflows even if some matrix failed.
       matrix:
         include: [
           { compiler: gcc, msystem: MINGW32, cflags: "-Werror"},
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):
Nice find @embg !

Cyan4973:
I don't remember we ever decided to set `winsymlinks:nativestrict`,
so I suppose this was set automatically by some github update, maybe dependabot, or maybe just the mingw image.
In which case, I would guess we were not the only ones impacted.

embg:
> I don't remember we ever decided to set `winsymlinks:nativestrict`, so I suppose this was set automatically by some github update, maybe dependabot, or maybe just the mingw image. In which case, I would guess we were not the only ones impacted.

@Cyan4973 Yes. It may have been fixed upstream over the long weekend, since [at least one recent job](https://github.com/facebook/zstd/actions/runs/3927389527/jobs/6713975844#step:8:10) is not running with `winsymlinks:nativestrict`. So the value of this PR may only be protection from future issues, in case this crops up again.

PR #3432, opened 2023-01-18
Fix CLI Handling of Permissions and Ownership (Again)

### LINKED ISSUE #2946: compressed file ownership inconsistent with that of uncompressed file
Using the zstd command (version 1.5.1) to compress a file, and the compressed file does not have the same owner:group as that of the input file. File permissions are consistent, though.

In comparison, gzip, xz, and zstd-1.4.4 retain both permission and ownership.

After looking at https://github.com/facebook/zstd/issues/2739 it seem that the problem of losing file ownership was also introduced by https://github.com/facebook/zstd/pull/2525

A casual look at https://github.com/facebook/zstd/pull/2525 reveals multiple other issues, such as no error checking for fdopen and unnecessary use of _open and fdopen because fopen with mode "wb" is sufficient.

I'd suggest to revert https://github.com/facebook/zstd/pull/2525 for now to make things work, then refine it if necessary.

A closer look shows that https://github.com/facebook/zstd/pull/2525 is trying to solve an invalid bug report at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981404#28
Being world readable is not a bug but a feature. Processes often run at umask 022 in a secure directory, and so world readable files cannot be actually read by others. If this is not the case, then the file is either readable by others intentionally, or the problem can be avoided by changing umask to e.g. 077 . open+fdopen does not resolve the problem (if it is a problem at all). umask does. The time window being one flaw. Another one might be:

Suppose a file owned by user1:group1, permission 0660, with open+fdopen it can get ownership user1:group2, permission 0660.
It's better to umask 0777 then call the zstd command.
The umask command need not be called by the zstd program. It can be issued by the user before running zstd:
$ (umask 0777; zstd file_to_be_compressed.txt)

```
$ ls -l foo
-rw-r--r-- 1 10001 20010 0 Dec 22 15:14 foo
$ xz -k foo
$ gzip -k foo
$ (umask 0777; zstd foo)
foo                  :1300.00%   (     0 =>     13 bytes, foo.zst)             
$ ls -l foo*
-rw-r--r-- 1 10001 20010  0 Dec 22 15:14 foo
-rw-r--r-- 1 10001 20010 24 Dec 22 15:14 foo.gz
-rw-r--r-- 1 10001 20010 32 Dec 22 15:14 foo.xz
-rw-r--r-- 1 10001 20010 13 Dec 22 15:14 foo.zst
$ zstd --version
*** zstd command line interface 64-bits v1.4.4, by Yann Collet ***
$ 
```

felixhandte:
This PR switches the `zstd` CLI's handling of output files' ownership and permissions back (again). This PR brings `zstd`'s behavior back into alignment with `gzip`'s, without introducing any gaps where output files have incorrectly loose permissions.

This PR addresses #2946. Previous issues and PRs on this topic include #2739, #2525, #2491 and #2495, and #1630 and #1644.

When the input is not a regular file, we create the output file with default `0666` permissions (modulated by the `umask`), and then don't change anything at the end of the operation.

When the input *is* a regular file, we create it with `0600` (u+rw) permissions, and then copy the input's ownership, permissions, and timestamps to the output file at the end of the operation. E.g.:

```sh
$ ls -lAh ~/prog/silesia/silesia.tar; \
  ./zstd -q -5 ~/prog/silesia/silesia.tar -o silesia.tar.zst & \
  sleep 1; \
  ls -lAh silesia.tar.zst; \
  wait; \
  ls -lAh silesia.tar.zst
-rw-r----- 1 felixh users 203M Aug 18  2021 /home/felixh/prog/silesia/silesia.tar
[1] 4035425
-rw------- 1 felixh users 25M Jan 18 14:05 silesia.tar.zst
[1]+  Done                    ./zstd -q -5 ~/prog/silesia/silesia.tar -o silesia.tar.zst
-rw-r----- 1 felixh users 61M Aug 18  2021 silesia.tar.zst
```

We also leverage a trick found in `gzip` to avoid a different permissions/ownership race by setting the correct group, then permissions, then user. Otherwise there would be a window of time in which incorrectly lax permissions might be exposed to the wrong user or group. (It's not meaningful for there to be incorrectly lax user permissions exposed to the user that zstd is running as / the file is created as, since an attacker running as that user would already be able to `chmod()` the file to make it user r/w/x-able. But without this 3-step process, it would be possible for incorrectly lax permissions to be applied to the group the file is created as, or the user the file ends up being owned by.)

Finally, this PR refactors how we stat files slightly, to reduce the count of `stat()` calls we make. A further improvement might be to switch to doing `fstat()` calls.

Adding testing for this PR is a challenge. We have reasonably good coverage of checking that timestamps and mode bits are copied over. But it's difficult to extend that to testing that the user and group are copied over correctly. (1) We can't rely on particular other users/groups existing across all the platforms on which we want to test. (2) Zstd will mostly fail to `chown()` the output file anyways:

> Only a privileged process (Linux: one with the `CAP_CHOWN` capability) may change the owner of a file.  The owner of a file may change the group of the file to any group of which that owner is a member.
> \- `chown(2)` man page

I have locally verified that it does correctly propagate ownership when invoked with `sudo`.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -185,7 +185,7 @@ int UTIL_isRegularFileStat(const stat_t* statbuf)
 int UTIL_chmod(char const* filename, const stat_t* statbuf, mode_t permissions)
 {
     stat_t localStatBuf;
-    UTIL_TRACE_CALL("UTIL_chmod(%s, %u)", filename, (unsigned)permissions);
+    UTIL_TRACE_CALL("UTIL_chmod(%s, %#4o)", filename, (unsigned)permissions);
     if (statbuf == NULL) {
         if (!UTIL_stat(filename, &localStatBuf)) {
             UTIL_TRACE_RET(0);
--- a/programs/util.c
+++ b/programs/util.c
@@ -272,11 +272,15 @@ int UTIL_isDirectory(const char* infilename)
 
 int UTIL_isDirectoryStat(const stat_t* statbuf)
 {
+    int ret;
+    UTIL_TRACE_CALL("UTIL_isDirectoryStat()");
 #if defined(_MSC_VER)
-    return (statbuf->st_mode & _S_IFDIR) != 0;
+    ret = (statbuf->st_mode & _S_IFDIR) != 0;
 #else
-    return S_ISDIR(statbuf->st_mode) != 0;
+    ret = S_ISDIR(statbuf->st_mode) != 0;
 #endif
+    UTIL_TRACE_RET(ret);
+    return ret;
 }
 
 int UTIL_compareStr(const void *p1, const void *p2) {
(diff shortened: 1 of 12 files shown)


programs/util.c
```diff
@@ -185,7 +185,7 @@ int UTIL_isRegularFileStat(const stat_t* statbuf)
 int UTIL_chmod(char const* filename, const stat_t* statbuf, mode_t permissions)
 {
     stat_t localStatBuf;
-    UTIL_TRACE_CALL("UTIL_chmod(%s, %u)", filename, (unsigned)permissions);
+    UTIL_TRACE_CALL("UTIL_chmod(%s, %#4o)", filename, (unsigned)permissions);
```
Cyan4973, line 188: First time I see this formating instruction...

Cyan4973 (approved):

PR #3434, opened 2023-01-18
Huffman refactor

terrelln:
The first commit deletes all unused Huffman functions.

The second commit replaces boolean parameters (`bmi2`, `suspectUncompressible`, `preferRepeat`, `optimalDepth`) with a `flags` bit field. This makes it easier to add flags, and makes the parameter list shorter. I've also added another flag `disableAsm` which disables the assembly decompressor, which is useful for fuzzing. I've renamed the functions to remove the `_bmi2()` suffix, since the first commit deleted all the non-bmi2 variants.

In a stacked PR, I will be adding a fast C variant of the decoders, and will add another flag `disableFast`, which will allow us to opt out of the fast decoder for fuzzing. This is the motivation for switching to the `flags`.


Cyan4973:
The first part is a welcome reduction in code size, so it's basically good to go.

For the second part, I wonder sometimes if the new multi-purposes `flags` field makes it more difficult for the compiler to enable constant propagation, which was actually a goal in some cases, in order to generate specialized function.

I think `bmi2` specialization is the most important one that comes to mind, though there might be others.

If a constant propagation case is missing, this should be visible in benchmarks, assuming we can produce the relevant benchmark test which triggers the target code path.

terrelln:
> If a constant propagation case is missing, this should be visible in benchmarks, assuming we can produce the relevant benchmark test which triggers the target code path.

None of these flags are used in constant propagation, or in hot code. They are all passed in through the public interface, so we don't know the value at compile time.

Specifically the BMI2 flag just picks which version of the function we call.

I benchmarked the code before and after with these 2 commands, and don't see any changes:

```
zstd -b1e1 ~/datasets/silesia.tar
zstd -b1e1 ~/datasets/silesia.tar --zstd=tlen=131072 --compress-literals
```

Cyan4973 (approved):

lib/common/huf.h
```diff
@@ -156,30 +77,44 @@ typedef U32 HUF_DTable;
...
+     * If set: Sample the input and check if the sample is uncompressible, if it is then don't attempt to compress.
+     * If unset: Always histogram the entire input.
+     */
+    HUF_flags_suspectUncompressible = (1 << 3),
+    /**
+     * If set: Don't use assembly implementations
+     * If unset: Allow using assembly implementations
+     */
+    HUF_flags_disableAsm = (1 << 4),
+} HUF_flags_e;
```
terrelln, line 111: The flags that we are replacing are all listed here.

The bmi2/disableAsm flags select which decoding loop to call, but aren't actually passed down to the hot decoding loop.

The others are all used during table building, but aren't used during inlining or constant prop.

PR #3435, opened 2023-01-18
added c89 build test to CI

Cyan4973:
This test ensures that `libzstd` and the `zstd` CLI can be compiled with `c89`/`c90` compilers,
with the additional requirement of `long long` type and variadic macros support.

Fixed a few `-pedantic` warnings that appeared in the process.

The test initially attempted to compile `tests/` tools with the same requirement,
but this is a strong requirement, which makes the code more difficult to write at times,
so it feels only appropriate to accept this restriction for `libzstd` and the `zstd` CLI proper.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -47,6 +47,15 @@ jobs:
         APT_PACKAGES="gcc-multilib" make apt-install
         CFLAGS="-mx32 -O1 -fstack-protector" make check V=1
 
+  build-c89:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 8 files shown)


Cyan4973:
`decodecorpus.c` uses `snprintf()`, which is `c99`, and is therefore not available when compiled in strict `c90` mode.

Reducing the strict iso `c90` requirement to `libzstd` and `zstd` CLI only.

terrelln:
Needs to be rebased

Cyan4973:
> Needs to be rebased

done

Cyan4973:
only `version-compatibility` test is failing now

terrelln (approved):

PR #3436, opened 2023-01-19
[versions-test] Work around bug in dictionary builder for older versions 

terrelln:
Older versions of zstandard have a bug in the dictionary builder, that
can cause dictionary building to fail. The process still exits 0, but
the dictionary is not created.

For reference, the bug is that it creates a dictionary that starts with
the zstd dictionary magic, in the process of writing the dictionary header,
but the header isn't fully written yet, and zstd fails compressions in
this case, because the dictionary is malformated. We fixed this later on
by trying to load the dictionary as a zstd dictionary, but if that fails
we fallback to content only (by default).

The fix is to:
1. Make the dictionary determinsitic by sorting the input files.
   Previously the bug would only sometimes occur, when the input files
   were in a particular order.
2. If dictionary creation fails, fallback to the `head` dictionary.

### DIFF
--- a/tests/test-zstd-versions.py
+++ b/tests/test-zstd-versions.py
@@ -85,18 +85,23 @@ def get_git_tags():
     return tags
 
 
-def create_dict(tag, dict_source_path):
+def create_dict(tag, dict_source_path, fallback_tag=None):
     dict_name = 'dict.' + tag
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):
Great investigation !

PR #3437, opened 2023-01-19
Fuzz the Block-Level Sequence Producer API

embg:
Adds external matchfinder registration to `FUZZ_setRandomParameters()`. This allows the fuzzers to catch 3/4 of the bugs from https://github.com/facebook/zstd/pull/3433. (The 4th bug is [captured](https://github.com/facebook/zstd/blob/dev/tests/zstreamtest.c#L1928-L1933) in zstreamtest.c). I ran the fuzzers overnight and did not find any new crashes, only timeouts and one OOM (these don't indicate real bugs in zstd).

### DIFF
--- a/tests/fuzz/zstd_helpers.c
+++ b/tests/fuzz/zstd_helpers.c
@@ -17,6 +17,7 @@
 #include "fuzz_helpers.h"
 #include "zstd.h"
 #include "zdict.h"
+#include "matchfinder.h"
 
 const int kMinClevel = -3;
 const int kMaxClevel = 19;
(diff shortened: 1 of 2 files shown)


tests/fuzz/zstd_helpers.c
```diff
@@ -122,6 +134,12 @@ void FUZZ_setRandomParameters(ZSTD_CCtx *cctx, size_t srcSize, FUZZ_dataProducer
     if (FUZZ_dataProducer_uint32Range(producer, 0, 1) == 0) {
       setRand(cctx, ZSTD_c_targetCBlockSize, ZSTD_TARGETCBLOCKSIZE_MIN, ZSTD_TARGETCBLOCKSIZE_MAX, producer);
     }
+
+    if (FUZZ_dataProducer_uint32Range(producer, 0, 10) == 0) {
```
terrelln: Nit: Lets enable if == 1. When the fuzzer runs out of input, it will always return 0. I'd rather that case not enable the external match finder.

```suggestion
    if (FUZZ_dataProducer_uint32Range(producer, 0, 10) == 1) {
```

terrelln (approved):
Looks good, just one minor suggestion

PR #3439, opened 2023-01-20
Fix sequence validation and seqStore bounds check

daniellerozenblit:
While working on the Sequence Compression API fuzzer, I found bugs in in `ZSTD_validateSequence()` as well as in `ZSTD_copySequencesToSeqStoreExplicitBlockDelim()` that allow one additional `seqDef` to be written to the `seqStore`, past the designated memory allocation.

#### Bug 1
The validation method currently only checks that the `matchLength` of each sequence is greater than the *global* minimum match length of 3. This is not accurate in cases where the `minMatch` in the cctx does not properly correspond with the true minimum match length of the sequences.

In most cases, this is fine. However, we currently allocate less memory for `minMatch > 3` than for `minMatch = 3`. This means that `ZSTD_validateSequence()` does not always properly validate sequences given the memory constraints determined by `minMatch` in the cctx.

#### Fix
Fix: See changes in `ZSTD_validateSequence()`

We now loosely check that match lengths respect the cctx's `minMatch`. Meaning, we check that all match lengths are greater than 3 when `minMatch` is greater than 3. This is enough to validate that `maxNbSeq` is large enough to fit all sequences.

#### Bug 2
`ZSTD_validateSequence()` is optional, so we need to protect agains these cases even when it is not enabled. We have asserts that catch overwrites to the `seqStore`, but these are not run in the production build. There is also a check before `ZSTD_storeSeq()` to ensure that we have not already passed the memory limit. However, this allows us to write one additional sequence past `MaxNbSeq` before throwing an error.

We are able to roundtrip without error in cases where we only write one sequence past the given memory limit.

#### Fix
Fix: See changes in `ZSTD_copySequencesToSeqStoreExplicitBlockDelim()` and `ZSTD_copySequencesToSeqStoreNoBlockDelim()`. I have changed these checks to return an error if we have already reached `maxNbSeq`, so that we do not attempt to write an additional sequence.

### DIFF
--- a/lib/common/error_private.c
+++ b/lib/common/error_private.c
@@ -53,6 +53,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(dstBuffer_wrong): return "Destination buffer is wrong";
     case PREFIX(srcBuffer_wrong): return "Source buffer is wrong";
     case PREFIX(externalMatchFinder_failed): return "External matchfinder returned an error code";
+    case PREFIX(invalid_external_sequences): return "External sequences are not valid";
     case PREFIX(maxCode):
     default: return notErrorCode;
     }
--- a/lib/common/error_private.c
+++ b/lib/common/error_private.c
@@ -53,7 +53,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(dstBuffer_wrong): return "Destination buffer is wrong";
     case PREFIX(srcBuffer_wrong): return "Source buffer is wrong";
     case PREFIX(externalMatchFinder_failed): return "External matchfinder returned an error code";
-    case PREFIX(invalid_external_sequences): return "External sequences are not valid";
+    case PREFIX(externalSequences_invalid): return "External sequences are not valid";
     case PREFIX(maxCode):
     default: return notErrorCode;
     }
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -6166,10 +6166,10 @@ ZSTD_validateSequence(U32 offCode, U32 matchLength, U32 minMatch,
      * window size. After output surpasses windowSize, we're limited to windowSize offsets again.
      */
     size_t const offsetBound = posInSrc > windowSize ? (size_t)windowSize : posInSrc + (size_t)dictSize;
-    RETURN_ERROR_IF(offCode > OFFSET_TO_OFFBASE(offsetBound), corruption_detected, "Offset too large!");
-    RETURN_ERROR_IF(matchLength < MINMATCH, corruption_detected, "Matchlength too small");
+    size_t const matchLenLowerBound = (minMatch == 3 || useExternalMatchFinder) ? 3 : 4;
+    RETURN_ERROR_IF(offCode > OFFSET_TO_OFFBASE(offsetBound), invalid_external_sequences, "Offset too large!");
     /* Validate maxNbSeq is large enough for the given matchLength and minMatch */
-    RETURN_ERROR_IF(!useExternalMatchFinder && minMatch >= 4 && matchLength < 4, corruption_detected, "Matchlength too small for the minMatch");
+    RETURN_ERROR_IF(matchLength < matchLenLowerBound, invalid_external_sequences, "Matchlength too small for the minMatch");
     return 0;
 }
 
@@ -6226,7 +6226,7 @@ ZSTD_copySequencesToSeqStoreExplicitBlockDelim(ZSTD_CCtx* cctx,
                                                 cctx->appliedParams.cParams.windowLog, dictSize, cctx->appliedParams.useExternalMatchFinder),
                                                 "Sequence validation failed");
         }
-        RETURN_ERROR_IF(idx - seqPos->idx >= cctx->seqStore.maxNbSeq, memory_allocation,
+        RETURN_ERROR_IF(idx - seqPos->idx >= cctx->seqStore.maxNbSeq, invalid_external_sequences,
                         "Not enough memory allocated. Try adjusting ZSTD_c_minMatch.");
         ZSTD_storeSeq(&cctx->seqStore, litLength, ip, iend, offBase, matchLength);
         ip += matchLength + litLength;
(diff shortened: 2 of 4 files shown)


Cyan4973 (approved):

Cyan4973:
I presume there is no change in performance when sequence validation is _not_ enabled ?

lib/compress/zstd_compress.c
```diff
@@ -6168,6 +6168,8 @@ ZSTD_validateSequence(U32 offCode, U32 matchLength,
     size_t const offsetBound = posInSrc > windowSize ? (size_t)windowSize : posInSrc + (size_t)dictSize;
     RETURN_ERROR_IF(offCode > OFFSET_TO_OFFBASE(offsetBound), corruption_detected, "Offset too large!");
     RETURN_ERROR_IF(matchLength < MINMATCH, corruption_detected, "Matchlength too small");
```
embg: Can we combine the two matchLength checks? Something like
```
size_t const matchLenLowerBound = (minMatch == 3 || useExternalMatchFinder) ? 3 : 4
RETURN_ERROR_IF(matchLength < matchLenLowerBound, corruption_detected, "Matchlength too small for the minMatch");
```

tests/zstreamtest.c
```diff
@@ -2066,6 +2066,90 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+            /* Test with sequence validation */
+            CHECK_Z(ZSTD_CCtx_setParameter(cctx, ZSTD_c_minMatch, 5));
+            CHECK_Z(ZSTD_CCtx_setParameter(cctx, ZSTD_c_blockDelimiters, ZSTD_sf_explicitBlockDelimiters));
+            CHECK_Z(ZSTD_CCtx_setParameter(cctx, ZSTD_c_validateSequences, 1));
+
+            cSize = ZSTD_compressSequences(cctx, dst, dstSize,
+                                   sequences, kNbSequences,
+                                   src, srcSize);
+
+            CHECK(!ZSTD_isError(cSize), "Should throw an error"); /* maxNbSeq is too small and an assert will fail */
```
embg, line 2099: Can you add a second check to ensure it's the expected error code? Example: https://github.com/facebook/zstd/blob/64963dcbd6162c52ba9273bb55d78c7a442b12f4/tests/zstreamtest.c#L1895-L1898

tests/zstreamtest.c
```diff
@@ -2066,6 +2066,90 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+
+            free(sequences);
+        }
+
+        ZSTD_CCtx_reset(cctx, ZSTD_reset_session_and_parameters);
+
+        { /* Test case with two additional sequences */
```
embg, line 2121: Does this test case also produce an error when validation is turned off?

embg (changes requested):
Great job finding and fixing these bugs in the sequence compression API!

Looks good, I just have a couple minor comments.

daniellerozenblit:
> I presume there is no change in performance when sequence validation is _not_ enabled ?

I ran some benchmarks and there doesn't appear to be any change in performance.

tests/zstreamtest.c
daniellerozenblit: Added! Note: validation is by default turned on in debug mode so I could not add this second check for the tests where validation was turned off (since a different error is produced depending on the debug level).

Cyan4973: Note: 
`debug` mode typically means `DEBUGLEVEL=1`, which enables `assert()`.
As a consequence, the program may `abort()` on some detected conditions, while it would not in `release` mode.

I don't think we ever had (so far) a difference in error code depending on `debug` vs `release`.
This would be a first, and indeed makes tests more complex to set up, since the outcome depends on `debug` vs `release` status.

I believe it's better to avoid this complication, and revert to the simpler idea that `debug` can `abort()`, but does not change produced error codes.

embg: @Cyan4973 @daniellerozenblit We just need to align these two error codes:
* https://github.com/facebook/zstd/blob/a0e3f2e4c0ff214323f7c9bab03d61a6a3124149/lib/compress/zstd_compress.c#L6342
* https://github.com/facebook/zstd/blob/a0e3f2e4c0ff214323f7c9bab03d61a6a3124149/lib/compress/zstd_compress.c#L6172

Perhaps we can add a new error code, something like "invalid external sequences"? Or move both of them to `corruption_detected`?

`memory_allocation` really seems wrong, since it's normally used elsewhere in the codebase for situations where memory allocation fails, but in this case we're not calling `malloc` or anything like that.

daniellerozenblit: I agree @embg that it would be best to align these error codes, since they are signaling the same issue.

Any preference @Cyan4973 between creating a new error code or using `corruption_detected` for both?

Cyan4973: I think it's valid to create a new error code for a more accurate description of the issue.

lib/common/error_private.c
```diff
@@ -53,6 +53,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(dstBuffer_wrong): return "Destination buffer is wrong";
     case PREFIX(srcBuffer_wrong): return "Source buffer is wrong";
     case PREFIX(externalMatchFinder_failed): return "External matchfinder returned an error code";
+    case PREFIX(invalid_external_sequences): return "External matchfinder returned an error code";
```
embg: Can you please change this error message to something like "External sequences are not valid"?

embg (approved):
LGTM!

lib/common/error_private.c
```diff
@@ -53,6 +53,7 @@ const char* ERR_getErrorString(ERR_enum code)
     case PREFIX(dstBuffer_wrong): return "Destination buffer is wrong";
     case PREFIX(srcBuffer_wrong): return "Source buffer is wrong";
     case PREFIX(externalMatchFinder_failed): return "External matchfinder returned an error code";
+    case PREFIX(invalid_external_sequences): return "External sequences are not valid";
```
Cyan4973: minor: even error codes have a naming scheme!
Looking at existing ones, the pattern is typically `topic_qualifier`.

For this case, it would be something like : 
`externalSequences_invalid`.

Cyan4973 (approved):

PR #3450, opened 2023-01-24
disable --rm on -o command

Cyan4973:
Up to now, it's possible to combine `--rm` with `-o` 
concatenating all processed frames into a single output file or device, 
while simultaneously deleting input.
Since such a concatenated result will not be able to regenerate input file names nor directory structure,
it is a destructive operation, which may result in permanent damage.

This situation has been partially dampened by adding a warning message and a confirmation prompt,
though both can be circumvented by adding the `-f` flag.

Update the policy, to disable `--rm` when `-o` is used in combination with multiple inputs.
This makes it more similar to `-c` (aka `stdout`) convention, itself derived from `gzip`.

The `-o` behavior is altered as follows : 
- With only 1 input : no change, can be combined with `--rm`
- With multiple inputs : automatically disable `--rm`, display a warning message (level 2) if it was set
- With multiple inputs : display a warning message about the impossibility to reverse operation, and prompt for confirmation.
- With multiple inputs and `-f` : do not prompt, just warn and proceed
- With multiple inputs and `-q` but not `-f` : do not prompt, explain and fail immediately

Adjusted and added test cases accordingly

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -840,9 +840,21 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
  */
 static int FIO_multiFilesConcatWarning(const FIO_ctx_t* fCtx, FIO_prefs_t* prefs, const char* outFileName, int displayLevelCutoff)
 {
-    if (fCtx->hasStdoutOutput) assert(prefs->removeSrcFile == 0);
+    if (fCtx->hasStdoutOutput) {
+        if (prefs->removeSrcFile)
+            /* this should not happen ; hard fail, to protect user's data
+             * note: this should rather be an assert(), but we want to be certain that user's data will not be wiped out in case it nonetheless happen */
+            EXM_THROW(43, "It's not allowed to remove input files when processed output is piped to stdout. "
+                "This scenario is not supposed to be possible. "
+                "This is a programming error. File an issue for it to be fixed.");
+    }
     if (prefs->testMode) {
-        assert(prefs->removeSrcFile == 0);
+        if (prefs->removeSrcFile)
+            /* this should not happen ; hard fail, to protect user's data
+             * note: this should rather be an assert(), but we want to be certain that user's data will not be wiped out in case it nonetheless happen */
+            EXM_THROW(43, "Test mode shall not remove input files! "
+                 "This scenario is not supposed to be possible. "
+                 "This is a programming error. File an issue for it to be fixed.");
         return 0;
     }
 
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -612,7 +612,7 @@ FIO_openDstFile(FIO_ctx_t* fCtx, FIO_prefs_t* const prefs,
         if (!prefs->overwrite) {
             if (g_display_prefs.displayLevel <= 1) {
                 /* No interaction possible */
-                DISPLAY("zstd: %s already exists; not overwritten  \n",
+                DISPLAYLEVEL(1, "zstd: %s already exists; not overwritten  \n",
                         dstFileName);
                 return NULL;
             }
@@ -723,7 +723,7 @@ int FIO_checkFilenameCollisions(const char** filenameTable, unsigned nbFiles) {
 
     filenameTableSorted = (const char**) malloc(sizeof(char*) * nbFiles);
     if (!filenameTableSorted) {
-        DISPLAY("Unable to malloc new str array, not checking for name collisions\n");
+        DISPLAYLEVEL(1, "Allocation error during filename collision checking \n");
         return 1;
     }
 
@@ -740,7 +740,7 @@ int FIO_checkFilenameCollisions(const char** filenameTable, unsigned nbFiles) {
     prevElem = filenameTableSorted[0];
     for (u = 1; u < nbFiles; ++u) {
         if (strcmp(prevElem, filenameTableSorted[u]) == 0) {
-            DISPLAY("WARNING: Two files have same filename: %s\n", prevElem);
+            DISPLAYLEVEL(2, "WARNING: Two files have same filename: %s\n", prevElem);
         }
         prevElem = filenameTableSorted[u];
     }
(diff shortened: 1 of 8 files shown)


programs/zstdcli.c
```diff
@@ -1480,24 +1480,24 @@ int main(int argCount, const char* argv[])
...
-        DISPLAY("error : can't use --patch-from=# on multiple files \n");
+        DISPLAYLEVEL(1, "error : can't use --patch-from=# on multiple files \n");
         CLEAN_RETURN(1);
     }
 
-    /* No status message in pipe mode (stdin - stdout) */
+    /* No status message by default when output is stdout */
     hasStdout = outFileName && !strcmp(outFileName,stdoutmark);
-    if ((hasStdout || !UTIL_isConsole(stderr)) && (g_displayLevel==2)) g_displayLevel=1;
+    if (hasStdout && (g_displayLevel==2)) g_displayLevel=1;
```
terrelln, line 1501: Yeah this is the line that is causing issues in the test. It is producing logs in the cli-tests when we didn't expect them because stderr is not a console.

When `!UTIL_isConsole(stderr)` we can instead set the `--no-progress` flag. Though we should make sure that using `--progress` will still re-enable progress messages.

Can you separate this change into a separate PR? Since this could change the CLI tests output. And otherwise the CLI tests should be unaffected by this PR.

Cyan4973: I have an issue with some tests though.

When `displayLevel==2` (default), the CLI expects interactive mode, and will sometimes prompt for confirmation.
There are a few tests which expect that to happen, and pipe a `y` into `stdin` to continue the operation successfully.

It happens here: when a user wants to concatenate multiple frames into a single output,
the CLI warns that it's probably not a good idea, and then asks the user for confirmation of this operation.

However, with `displayLevel==1`, this is interpreted as `--quiet`, in which case no prompt ever happens, and instead it's an instant failure. Since the test was expecting "success" (by providing the `y` to the prompt), the test now fails.

With the `!UTIL_isConsole(stderr)` automatic change of verbosity, CLI tests behave one way locally, and differently on Github Actions. That's a problem, tests require a unified behavior.

There are probably several ways to get around this problem. 

For example, one could simply avoid testing interactive prompt scenarios, by forcing such scenarios to be run exclusively with `-q` or `-v`.
Of course, such a strategy would make the test suite blind to interactive scenarios.

So I was more in favor of trying to "fix" this behavior, since the intention was just to remove the status updates from `stderr` when `stderr` is a log. But if it triggers too many complications, then maybe the answer will be to test less scenarios...

terrelln: I see what you're trying to test, it makes sense to combine it into this PR then.

programs/zstdcli.c
```diff
@@ -1480,24 +1480,27 @@ int main(int argCount, const char* argv[])
...
+
+    /* when stderr is not the console, do not pollute it with status updates */
+    if (!UTIL_isConsole(stderr)) FIO_setProgressSetting(FIO_ps_never);
```
terrelln: This doesn't quite work, because this breaks the `--progress` flag. When stderr is not a console, `--progress` won't work. Which is probably not the intended behavior.

terrelln: I think you basically want:

```c
if (!UTIL_isConsole(stderr) && FIO_getProgressSetting() == FIO_ps_auto) FIO_setProgressSetting(FIO_ps_never);
```

Though you'll probably have to add the function `FIO_getProgressSetting()`.

Cyan4973: Well, it _was_ the intended behavior, or rather the one I intended,
but indeed, existing tests are expecting something different.
This behavior is not explicitly defined, we end up rebuilding its definition from test results.

My understanding is that the current tests expect that `--progress` command can force writing status updates into `stderr` even when it's not the console. Not sure when that is useful, but since I don't have any strong opinion about that behavior, let's keep it.

That will require an additional variable within `zstdcli.c`, to register commands instead of pushing actions immediately, but that seems doable.

_edit_: seen your 2nd comment after redacting the answer. Indeed, I made a similar change.
Longer term, it opens up the topic: "where is the right place to adjust all these settings?",
as sometimes, it's unclear if this should be done in `zstdcli.c` or `fileio.c`.
But that's a rabbit hole. For the time being, let's focus on the pb at hand.

terrelln:
The remaining CLI test failures are benign, and can be updated with:

```
./tests/cli-tests/run.py --set-exact-output
```

Basically, they're just saying that we are calling `UTIL_isConsole(stderr)` one extra time when the output is stdout.

programs/fileio.c
```diff
@@ -823,41 +823,59 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
...
-            }
-            DISPLAYLEVEL(2, "The concatenated output CANNOT regenerate original file names nor directory structure. \n")
-            if (prefs->removeSrcFile) {
-                assert(fCtx->hasStdoutOutput == 0); /* not possible : never erase source files when output == stdout */
-                error = (g_display_prefs.displayLevel > displayLevelCutoff) && UTIL_requireUserConfirmation("This is a destructive operation. Proceed? (y/n): ", "Aborting...", "yY", fCtx->hasStdinInput);
-            }
-        }
+    if (fCtx->hasStdoutOutput) assert(prefs->removeSrcFile == 0);
+    if (prefs->testMode) {
+        assert(prefs->removeSrcFile == 0);
```
terrelln: nit: Because this is a dangerous operation, and not in the hot loop, I wonder if we want to use `EXM_THROW()` if `prefs->removeSrcFile`, and add a message saying that it is a programming error, and to open an issue. So that it is enabled in prod builds.

programs/fileio.c
```diff
@@ -823,41 +823,59 @@ static void FIO_adjustMemLimitForPatchFromMode(FIO_prefs_t* const prefs,
...
+    }
+    DISPLAYLEVEL(2, "The concatenated output CANNOT regenerate original file names nor directory structure. \n")
+
+    /* multi-input into single output : --rm is not allowed */
+    if (prefs->removeSrcFile) {
+        DISPLAYLEVEL(2, "Since it's a destructive operation, input files will not be removed. \n");
+        prefs->removeSrcFile = 0;
+    }
+
+    if (fCtx->hasStdoutOutput) return 0;
```
terrelln, line 879: Why are we not prompting if we have stdout as an output? I think the same behavior as `-o` for `-c` would be useful.

Cyan4973: Indeed, it would seem logical to have `-o` and `-c` sharing the same behavior.
But for `-c`, we are following the established behavior of `gzip`, and `gzip` features no prompt in this case.
`-o` on the other hand only exists for `zstd`, so we can select a more protective behavior.

terrelln (approved):

PR #3451, opened 2023-01-24
Bugfix redzone unpoisoning

yoniko:
Bug fix redzones by unpoisoning only the intended buffer and not the followup redzone.


felixhandte (approved):
Nice find!

PR #3460, opened 2023-01-27
fix long offset resolution

daniellerozenblit:
We currently determine whether or not `longOffsets` are present in a set of input sequences by checking if the `windowLog` is larger than `STREAM_ACCUMULATOR_MIN`.

This does not account for inputs with a dictionary, where it is legal for offsets to be larger than the window size. In these cases, we don't properly account for large offsets when flushing our bit accumulator in `ZSTD_encodeSequences_body()`.

This PR introduces a `largeOffsets` field in the `ZSTD_symbolEncodingTypeStats_t` which records whether or not large offsets are present in the sequences. This value of this field is finalized in `ZSTD_seqToCodes()`.

I ran some benchmarks for zstd level 1 & 3 and there does not appear to be any measurable change in compression speed.

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -74,7 +74,7 @@ jobs:
       run: |
         sudo apt-get -qqq update
         make libc6install
-        CC=clang make -C tests test-zstream32 FUZZERFLAGS="--big-tests"
+        CC=clang make -C tests test-zstream32 FUZZER_FLAGS="--big-tests"
 
   # lasts ~23mn
   gcc-8-asan-ubsan-testzstd:
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -65,6 +65,17 @@ jobs:
     - name: thread sanitizer fuzztest
       run: CC=clang make tsan-fuzztest
 
+
+  big-tests-zstreamtest32:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag=v3
+    - name: zstream tests in 32bit mode, with big tests
+      run: |
+        sudo apt-get -qqq update
+        make libc6install
+        CC=clang make -C tests test-zstream32 FUZZERFLAGS="--big-tests"
+
   # lasts ~23mn
   gcc-8-asan-ubsan-testzstd:
     runs-on: ubuntu-latest
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -299,6 +299,7 @@ typedef struct {
     BYTE*  ofCode;
     size_t maxNbSeq;
     size_t maxNbLit;
+    BYTE* longOffsets;
 
     /* longLengthPos and longLengthType to allow us to represent either a single litLength or matchLength
      * in the seqStore that has a value larger than U16 (if it exists). To do so, we increment
--- a/lib/common/zstd_internal.h
+++ b/lib/common/zstd_internal.h
@@ -299,7 +299,6 @@ typedef struct {
     BYTE*  ofCode;
     size_t maxNbSeq;
     size_t maxNbLit;
-    BYTE* longOffsets;
 
     /* longLengthPos and longLengthType to allow us to represent either a single litLength or matchLength
      * in the seqStore that has a value larger than U16 (if it exists). To do so, we increment
@@ -347,7 +346,7 @@ typedef struct {
 } ZSTD_frameSizeInfo;   /* decompress & legacy */
 
 const seqStore_t* ZSTD_getSeqStore(const ZSTD_CCtx* ctx);   /* compress & dictBuilder */
-void ZSTD_seqToCodes(const seqStore_t* seqStorePtr);   /* compress, dictBuilder, decodeCorpus (shouldn't get its definition from here) */
+int ZSTD_seqToCodes(const seqStore_t* seqStorePtr);   /* compress, dictBuilder, decodeCorpus (shouldn't get its definition from here) */
 
 /* custom memory allocation functions */
 void* ZSTD_customMalloc(size_t size, ZSTD_customMem customMem);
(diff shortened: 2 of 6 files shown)


lib/common/zstd_internal.h
```diff
@@ -299,6 +299,7 @@ typedef struct {
     BYTE*  ofCode;
     size_t maxNbSeq;
     size_t maxNbLit;
+    BYTE* longOffsets;
```
terrelln: This doesn't need to be a pointer.

```suggestion
    int longOffsets;
```

terrelln (changes requested):

lib/common/zstd_internal.h
daniellerozenblit: There are a bunch of places I have to un-const if I want to change this (I think).

terrelln: Makes sense, then instead of storing it in the `seqStore`, we can have `ZSTD_seqToCodes()` return that value, and store it in the `ZSTD_symbolEncodingTypeStats_t` returned by `ZSTD_buildSequencesStatistics()`.

tests/zstreamtest.c
```diff
@@ -2222,6 +2222,66 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+        /* Test large offset, small window size*/
+        {
+            size_t srcSize = 21;
+            void* const src = CNBuffer;
+            size_t dstSize = ZSTD_compressBound(srcSize);
+            void* const dst = compressedBuffer;
+            size_t const kNbSequences = 4;
+            ZSTD_Sequence* sequences = malloc(sizeof(ZSTD_Sequence) * kNbSequences);
+            void* const checkBuf = malloc(srcSize);
+            const size_t largeDictSize = 1 << 30;
```
terrelln: Can you reproduce this with a `largeDictSize = 1 << 29`?

tests/zstreamtest.c
```diff
@@ -2222,6 +2222,66 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
     }
     DISPLAYLEVEL(3, "OK \n");
 
+
+    DISPLAYLEVEL(3, "test%3i : Testing large offset with small window size: ", testNb++);
+    {
```
terrelln, line 2227: We should only run this test when `--big-tests` is passed.

```suggestion
    if (bigTests) {
```

terrelln (commented):
Looks good to me! I will also benchmark this PR to make sure I also don't see regressions.

Thanks for adding the test! We might need to try to reduce the memory usage of the test a little bit, otherwise it could be flaky in 32-bit builds, because we could run out of address space.

tests/zstreamtest.c
daniellerozenblit: It looks like `--big-tests` is false by default if the tests are built in 32-bit, but this test is specifically for 32-bit. Should we add a `--big-tests` flag to force enable big tests?

terrelln: Ah good catch - yeah lets add that flag, and add a CI job that runs zstreamtest with `--big-test` in CI.

lib/compress/zstd_compress.c
```diff
@@ -2559,26 +2559,30 @@ static void ZSTD_reduceIndex (ZSTD_matchState_t* ms, ZSTD_CCtx_params const* par
...
     assert(nbSeq <= seqStorePtr->maxNbSeq);
     for (u=0; u<nbSeq; u++) {
         U32 const llv = sequences[u].litLength;
+        U32 const ofCode = ZSTD_highbit32(sequences[u].offBase);
         U32 const mlv = sequences[u].mlBase;
         llCodeTable[u] = (BYTE)ZSTD_LLcode(llv);
-        ofCodeTable[u] = (BYTE)ZSTD_highbit32(sequences[u].offBase);
+        ofCodeTable[u] = (BYTE)ofCode;
         mlCodeTable[u] = (BYTE)ZSTD_MLcode(mlv);
+        longOffsets |= (ofCode >= STREAM_ACCUMULATOR_MIN);
```
terrelln: I did some benchmarking, and looked at the generated assembly, and lets go with this approach.

The assembly generated by using an if condition is slightly cleaner, gcc & clang use `cmov`. Additionally, it hides the condition behind `MEM_32bits()`, so in 64-bit code the compiler is sure to optimize it out.

```suggestion
        assert(!(MEM_64bits() && ofCode >= STREAM_ACCUMULATOR_MIN));
        if (MEM_32bits() && ofCode >= STREAM_ACCUMULATOR_MIN)
            longOffsets = 1;
```

This is neutral in both 64-bit mode and 32-bit mode. And in 64-bit mode, there is no instructions generated for this branch.

.github/workflows/dev-long-tests.yml
```diff
@@ -65,6 +65,17 @@ jobs:
...
+
+  big-tests-zstreamtest32:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag=v3
+    - name: zstream tests in 32bit mode, with big tests
+      run: |
+        sudo apt-get -qqq update
+        make libc6install
+        CC=clang make -C tests test-zstream32 FUZZERFLAGS="--big-tests"
```
terrelln:
```suggestion
        CC=clang make -C tests test-zstream32 FUZZER_FLAGS="--big-tests"
```

tests/zstreamtest.c
```diff
@@ -2222,6 +2222,66 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+        /* Test large offset, small window size*/
+        {
+            size_t srcSize = 21;
+            void* const src = CNBuffer;
+            size_t dstSize = ZSTD_compressBound(srcSize);
+            void* const dst = compressedBuffer;
+            size_t const kNbSequences = 4;
+            ZSTD_Sequence* sequences = malloc(sizeof(ZSTD_Sequence) * kNbSequences);
+            void* const checkBuf = malloc(srcSize);
+            const size_t largeDictSize = 1 << 25;
```
terrelln, line 2240: Oh nice, you got it to reproduce with `1 << 25`!

In that case we don't need to hide it behind `bigTests`. But I still like having one 32-bit test in CI that does run with `--big-tests`. That should help improve our 32-bit coverage a bit, without making our users tests flaky.

terrelln (approved):
Awesome, thanks!

PR #3461, opened 2023-01-27
Fix invalid assert in 32-bit decoding

terrelln:
The assert is only correct for valid sequences, so disable it for everything execpt round trip fuzzers.


Cyan4973:
Why do we need this `assert()` for fuzzers ?

terrelln:
> Why do we need this assert() for fuzzers ?

We don't absolutely need it, but I figure leaving it in is helpful to understand the code in the non-corrupted case.

And if the round trip fuzzers hit this assert, then we've violated an assumption. Presumably we could get corruption if this occurred, but the assert firing would be clearer.

terrelln:
`FUZZING_ASSERT_VALID_SEQUENCE` is only defined when we are guaranteed to only be decompressing valid sequences.

Cyan4973 (approved):

Issue #3456, opened 2023-01-25
Vectorize `blockSize_explicitDelimiter()`


embg:
The function [blockSize_explicitDelimiter()](https://github.com/facebook/zstd/blob/321490cd5b9863433b3d44816d04012874e5ecdb/lib/compress/zstd_compress.c#L6427-L6447) is used to validate that `sum(matchLen) + sum(litLen)` for external sequences adds up to the correct `blockSize`. The hot loop is:
```
    while (spos < inSeqsSize) {
        end = (inSeqs[spos].offset == 0);
        blockSize += inSeqs[spos].litLength + inSeqs[spos].matchLength;
        if (end) {
            if (inSeqs[spos].matchLength != 0)
                return 42; // error
            break;
        }
        spos++;
    }
```

Godbolt indicates that clang15 compiles this to a naive scalar loop: https://godbolt.org/z/osjE5KeP6. gcc12 yields a similar result. uiCA gives the naive loop a throughput of 3 cycles/sequence, which seems pretty slow to me: https://bit.ly/401i7ZD.

I think we can do significantly better. For example, this loop could be vectorized by loading one `ZSTD_Sequence` struct into a SSE vector (or two `ZSTD_Sequence` structs into an AVX vector) and performing 2 (or 4) parallel sums. Skylake Xeon can do 2 (load+VPADD)/cycle (according to uops.info), so with multiple accumulators and unrolling we could get up to 4 (or 8) parallel sums. Of course, we'd need to be careful about 32-bit overflow, but even with vectorized overflow checks this solution should be much faster.

PR #3465, opened 2023-01-30
Guard against invalid sequences from external sequence producers

embg:
### Overview
* Fixes two bugs in the external matchfinder API found by @terrelln. The bugs arise in scenarios where the external matchfinder produces invalid sequences.
* Fixes a bug I found in the external matchfinder API tests and extends the tests to cover the two bugs Nick found.
* ~~Adds a new CI job to run `make test` with `DEBUGLEVEL=0`.~~
* ~~Changes the debug level at which sequence validation is always on from `DEBUGLEVEL=2` to `DEBUGLEVEL=1`.~~ Edit: makes `DEBUGLEVEL` independent from `validateSequences`
* Adds `validateSequences` to the fuzzer
* Migrates most (hopefully all) sequence validation errors to the same error code.

### Bug 1
The return code of `ZSTD_copySequencesToSeqStoreExplicitBlockDelim()` [is not checked](https://github.com/facebook/zstd/blob/b3b43f2893fa03da3b8004b449a7ec590f0e1e5b/lib/compress/zstd_compress.c#L3162). This means if sequence validation is turned on, and it raises an error, that error is not propagated.
* Fix: add `FORWARD_IF_ERROR()`
* Test: the `EMF_INVALID_OFFSET` testcase exercises this code path

### Bug 2
When using an external matchfinder, there is no protection against external sequences with `litLen > srcSize`. This causes an assert failure in debug builds and a **buffer overflow** in release builds. 

The `ZSTD_compressSequences()` API is not vulnerable to this bug because [it calls](https://github.com/facebook/zstd/blob/b3b43f2893fa03da3b8004b449a7ec590f0e1e5b/lib/compress/zstd_compress.c#L6507-L6516) `determine_blockSize()` before ingesting external sequences. `determine_blockSize()` [enforces](https://github.com/facebook/zstd/blob/b3b43f2893fa03da3b8004b449a7ec590f0e1e5b/lib/compress/zstd_compress.c#L6456-L6471) `sum(matchLen) + sum(litLen) + lastLiterals <= srcSize`.
* Fix: call `determine_blockSize()` before copying sequences from external matchfinder to the seqStore. Note that it is renamed by this PR to `ZSTD_inferBlockSizeFromExternalSequences()`.
* Test: the `EMF_INVALID_MATCHLEN`, `EMF_INVALID_LITLEN`, and `EMF_INVALID_LAST_LITS` exercise this code path.
* Note 1: I ran a performance test using `contrib/seqBench`. The overhead of `ZSTD_inferBlockSizeFromExternalSequences()` reduces compression speed by about 8%, assuming an extremely fast external matchfinder. I filed a [long-term issue](https://github.com/facebook/zstd/issues/3456) to bring the overhead to effectively zero.
* Note 2: it isn't strictly necessary to check each `matchLen`, but it's almost free if we are already checking each `litLen`, so I don't think it's worth the extra complexity to eliminate the `matchLen` check from `ZSTD_inferBlockSizeFromExternalSequences()` at this time. It may be interesting to remove it when other bottlenecks have been further optimized.

### Bug 3
[This for loop](https://github.com/facebook/zstd/blob/b3b43f2893fa03da3b8004b449a7ec590f0e1e5b/tests/zstreamtest.c#L1876) in `zstreamtest.c` has an off-by-one error in the loop condition. The error prevents tests from running with `enableFallback = 1`.
* Fix: change `<` to `<=`
* Test: `make test`

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3042,7 +3042,7 @@ static size_t ZSTD_postProcessExternalMatchFinderResult(
  * except it doesn't check for a block delimiter to end summation.
  * Removing the early exit allows the compiler to auto-vectorize (https://godbolt.org/z/cY1cajz9P).
  * This function can be deleted and replaced by determine_blockSize after we resolve issue #3456. */
-static size_t ZSTD_fastSequenceLengthSum(ZSTD_Sequence* seqBuf, size_t seqBufSize) {
+static size_t ZSTD_fastSequenceLengthSum(ZSTD_Sequence const* seqBuf, size_t seqBufSize) {
     size_t matchLenSum, litLenSum, i;
     matchLenSum = 0;
     litLenSum = 0;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -279,7 +279,7 @@ static ZSTD_paramSwitch_e ZSTD_resolveEnableLdm(ZSTD_paramSwitch_e mode,
 
 /* Enables validation for external sequences in debug builds. */
 static int ZSTD_resolveExternalSequenceValidation(int mode) {
-#if defined(DEBUGLEVEL) && (DEBUGLEVEL>=2)
+#if defined(DEBUGLEVEL) && (DEBUGLEVEL>=1)
     (void)mode;
     return 1;
 #else
@@ -3159,8 +3159,21 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
                 /* Return early if there is no error, since we don't need to worry about last literals */
                 if (!ZSTD_isError(nbPostProcessedSeqs)) {
                     ZSTD_sequencePosition seqPos = {0,0,0};
-                    ZSTD_copySequencesToSeqStoreExplicitBlockDelim(
-                        zc, &seqPos, zc->externalMatchCtx.seqBuffer, nbPostProcessedSeqs, src, srcSize
+                    size_t externalBlockSize = ZSTD_inferBlockSizeFromExternalSequences(
+                                                   ZSTD_sf_explicitBlockDelimiters,
+                                                   srcSize, srcSize,
+                                                   zc->externalMatchCtx.seqBuffer, nbPostProcessedSeqs,
+                                                   seqPos
+                                               );
+                    FORWARD_IF_ERROR(externalBlockSize, "Block size inferred from external sequences is too large!");
+                    assert(externalBlockSize <= srcSize);  /* enforced by ZSTD_inferBlockSizeFromExternalSequences() */
+                    FORWARD_IF_ERROR(
+                        ZSTD_copySequencesToSeqStoreExplicitBlockDelim(
+                            zc, &seqPos,
+                            zc->externalMatchCtx.seqBuffer, nbPostProcessedSeqs,
+                            src, externalBlockSize
+                        ),
+                        "Failed to copy external sequences to seqStore!"
                     );
                     ms->ldmSeqStore = NULL;
                     DEBUGLOG(5, "Copied %lu sequences from external matchfinder to internal seqStore.", (unsigned long)nbExternalSeqs);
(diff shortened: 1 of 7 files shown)


terrelln (approved):

lib/compress/zstd_compress.c
```diff
@@ -279,7 +279,7 @@ static ZSTD_paramSwitch_e ZSTD_resolveEnableLdm(ZSTD_paramSwitch_e mode,
 
 /* Enables validation for external sequences in debug builds. */
 static int ZSTD_resolveExternalSequenceValidation(int mode) {
-#if defined(DEBUGLEVEL) && (DEBUGLEVEL>=2)
+#if defined(DEBUGLEVEL) && (DEBUGLEVEL>=1)
```
Cyan4973: The purpose of `DEBUGLEVEL==1` is to enable `assert()`.

Enabling `assert()` should never change the behavior of the program.
`assert()` are only meant to catch programming errors that should never happen.
They are not meant to "validate user input" for example.

Here we have a subtle cousin, when an assert-related constant (`DEBUGLEVEL`) is employed to select "validate user input". So what would have been `asan` crash is now a valid "return error" scenario. This is a change of behavior, and I believe it's not correct.

This is pretty subtle, but has consequences, such as test policies, so we probably need to discuss this.

embg: ~~I think keeping it at `>= 2` is reasonable.~~

Edit: discussed offline, `DEBUGLEVEL` should be independent from `validateSequences` to keep in line with existing testing setup.

terrelln: @embg Another option would be to add an `assert(sequencesValid())` after the normal sequence validation check.

Basically add an `assert(!ZSTD_isError(ZSTD_validateSequences(...)));` here

https://github.com/facebook/zstd/blob/2f74507bbd0e8aca907a2842a311cc24ec815cf2/lib/compress/zstd_compress.c#L6276

Since it is UB to pass invalid sequences without validating them, we are within our right to assert that they are valid after we do the check.

embg: I think the assert could be added in a separate PR. For now, will just go back to the status quo before the external matchfinder API landed, with `validateSequences` independent from `DEBUGLEVEL`.

Edit: the reason why I hesitate to add it in this PR is that `validateSequences()` depends on a state that is only updated if `cctx->appliedParams.validateSequences == 1`, so it's not completely trivial to add the assert.

embg: I'll add `validateSequences` to the fuzzer, though, so that when we expose the fuzzing interface for external matchfinders we'll be able to catch invalid sequence bugs.

embg:
* Removed any connection between `DEBUGLEVEL` and `ZSTD_c_validateSequences`.
* Reduced the overhead of buffer overflow protection from 8% regression to less than 2% regression (measured using perf + `contrib/externalMatchfinder`).

lib/compress/zstd_compress.c
```diff
@@ -3042,6 +3036,23 @@ static size_t ZSTD_postProcessExternalMatchFinderResult(
     }
 }
 
+/* ZSTD_fastSequenceLengthSum() :
+ * Returns sum(litLen) + sum(matchLen) + lastLits for *seqBuf*.
+ * Similar to another function in zstd_compress.c (determine_blockSize),
+ * except it doesn't check for a block delimiter to end summation.
+ * Removing the early exit allows the compiler to auto-vectorize (https://godbolt.org/z/cY1cajz9P).
+ * This function can be deleted and replaced by determine_blockSize after we resolve issue #3456. */
+static size_t ZSTD_fastSequenceLengthSum(ZSTD_Sequence* seqBuf, size_t seqBufSize) {
```
Cyan4973: minor : the array is read-only, so it can be `const ZSTD_Sequence*`.

lib/compress/zstd_compress.c
```diff
@@ -3042,6 +3036,23 @@ static size_t ZSTD_postProcessExternalMatchFinderResult(
...
+    size_t matchLenSum, litLenSum, i;
+    matchLenSum = 0;
+    litLenSum = 0;
+    for (i = 0; i < seqBufSize; i++) {
+        litLenSum += seqBuf[i].litLength;
+        matchLenSum += seqBuf[i].matchLength;
```
Cyan4973, line 3051: I presume focusing on `LitLen` only doesn't make a sensible difference performance wise ?

Cyan4973:
Looks good !
My only comment is about making `ZSTD_fastSequenceLengthSum()` `const` correct,
once this minor thing is updated, it's good to go !

lib/compress/zstd_compress.c
embg: It's very fast either way. But also, I think we do need both -- otherwise isn't it possible that `ip` will overflow `src`? https://github.com/facebook/zstd/blob/dev/lib/compress/zstd_compress.c#L6279

lib/compress/zstd_compress.c
embg: Thanks, good catch

lib/compress/zstd_compress.c
Cyan4973: Good point !

PR #3471, opened 2023-01-31
Reduce external sequence producer API overhead by 25%

embg:
Adds a cctxParam to disable repcode search during external sequence parsing (only in explicit block delim mode for now, since that's where we currently care about parsing speed). For external matchfinders which don't explicitly search for repcode matches, this sacrifices less than 1% compression ratio on silesia.tar.

In general, the compression ratio trade-off is matchfinder- and data-dependent. Users should benchmark against their own data to determine if the trade-off is worth it. I have enabled by default below compression level 10, because the speed improvement is so great that I imagine few practical use-cases would gain enough ratio from disabling to justify disabling.

In the future, we might be able to use SIMD to run repcode search much faster, and change the trade-off such that enabling at low levels makes sense.

Overall external matchfinder API overhead (non-external-matchfinder compression CPU) is currently about 50% inside `ZSTD_copySequencesToSeqStoreExplicitBlockDelim()`, so this PR reduces overall overhead by about 25% (see perf numbers below).

Before:
Screenshot 2023-01-31 at 4 11 35 PM

After:
Screenshot 2023-01-31 at 4 12 56 PM

cc @daweiq @GarenJian-Intel

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -290,6 +290,15 @@ static size_t ZSTD_resolveMaxBlockSize(size_t maxBlockSize) {
     }
 }
 
+static ZSTD_paramSwitch_e ZSTD_resolveExternalRepcodeSearch(ZSTD_paramSwitch_e value, int cLevel) {
+    if (value != ZSTD_ps_auto) return value;
+    if (cLevel < 10) {
+        return ZSTD_ps_disable;
+    } else {
+        return ZSTD_ps_enable;
+    }
+}
+
 /* Returns 1 if compression parameters are such that CDict hashtable and chaintable indices are tagged.
  * If so, the tags need to be removed in ZSTD_resetCCtx_byCopyingCDict. */
 static int ZSTD_CDictIndicesAreTagged(const ZSTD_compressionParameters* const cParams) {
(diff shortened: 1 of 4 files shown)


lib/compress/zstd_compress.c
```diff
@@ -290,6 +290,15 @@ static size_t ZSTD_resolveMaxBlockSize(size_t maxBlockSize) {
     }
 }
 
+static ZSTD_paramSwitch_e ZSTD_resolveFastExternalSequenceParsing(ZSTD_paramSwitch_e value, int cLevel) {
+    if (value != ZSTD_ps_auto) return value;
+    if (cLevel < 10) {
```
terrelln, line 295: Normally we base these types of decisions on the `strategy`, because that still works even if the user requests explicit parameters instead of a level. E.g. you could do `strategy < ZSTD_lazy2`.

Cyan4973: One of the problems for this specific scenario is that `strategy` is effectively what is being replaced by the external match finder.

lib/compress/zstd_compress.c
```diff
@@ -6296,6 +6333,29 @@ ZSTD_copySequencesToSeqStoreExplicitBlockDelim(ZSTD_CCtx* cctx,
         ZSTD_storeSeq(&cctx->seqStore, litLength, ip, iend, offBase, matchLength);
         ip += matchLength + litLength;
     }
+
+    /* If we skipped repcode search while parsing, we need to update repcodes now */
+    assert(idx >= startIdx);
+    if (externalRepSearch == ZSTD_ps_disable && idx != startIdx) {
```
Cyan4973, line 6340: Did you mean `if (externalRepSearch == ZSTD_ps_enable &&`

embg: No, I changed it after renaming the cctxParam. Disable = don't search for external repcodes (since the cctxParam is now called `ZSTD_c_searchForExternalRepcodes`).

Cyan4973: Right,
it's the special finalizer, and it's only triggered if repcode substitution was disabled,
because it's implied that offset history is already maintained during repcode substitution.

Cyan4973: I believe this test implies that, when reaching that `if ()` statement, `externalRepSearch != ZSTD_ps_auto`.
And I'm sure it is, but this condition is ensured elsewhere, so this is a long distance correlation.
This is probably worth an `assert()`.

Cyan4973 (approved):

lib/compress/zstd_compress.c
embg: Will add the assert before merging!

PR #3472, opened 2023-02-02
Fix 32-bit decoding with large dictionary

terrelln:
The 32-bit decoder could corrupt the regenerated data by using regular offset mode when there were actually long offsets. This is because we were only considering the window size in the calculation, not the dictionary size. So a large dictionary could allow longer offsets.

Fix this in two ways:
1. Instead of looking at the window size, look at the total referencable bytes in the history buffer. Use this in the comparison instead of the window size. Additionally, we were comparing against the wrong value, it was too low. Fix that by computing exactly the maximum offset for regular sequence decoding.
2. If it is possible that we have long offsets due to (1), then check the offset code decoding table, and if the decoding table's maximum number of additional bits is no more than STREAM_ACCUMULATOR_MIN, then we can't have long offsets.

This gates us to be using the long offsets decoder only when we are very likely to actually have long offsets.

Note that this bug only affects the decoding of the data, and the original compressed data, if re-read with a patched decoder, will correctly regenerate the orginal data. Except that the encoder also had the same issue previously.

This fixes both the open OSS-Fuzz issues.

Credit to OSS-Fuzz

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -1170,7 +1170,7 @@ ZSTD_updateFseStateWithDInfo(ZSTD_fseState* DStatePtr, BIT_DStream_t* bitD, U16 
 }
 
 /* We need to add at most (ZSTD_WINDOWLOG_MAX_32 - 1) bits to read the maximum
- * offset bits. But we can only read at most (STREAM_ACCUMULATOR_MIN_32 - 1)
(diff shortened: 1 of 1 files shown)


Cyan4973:
arf, I just meant to approve the PR, but got it merged instead...
Anyway, I believe it's good to go.

terrelln:
> arf, I just meant to approve the PR, but got it merged instead...

No worries! Its a good thing, because that means today's fuzzer builds picked up the fix.

PR #3476, opened 2023-02-06
fix oss-fuzz case 55714

Cyan4973:
This issue, discovered by oss-fuzz, can only happen with the following conditions : 
- Support for legacy format v0.3 enabled (note: support for format 0.3 is disabled by default)
- compiled in 32-bit mode (not reproduced in 64-bit mode)

that is to say, this scenario is very unlikely to be reproducible in the wild.
Fixing it nonetheless for completeness.

The fix is trivial, though it's likely going to cost some performance. But at this stage, it's fair to say we don't care about the performance of the deprecated legacy decoder v0.3 anymore. There shouldn't be any data left using this old short lived legacy format.

### DIFF
--- a/lib/legacy/zstd_v01.c
+++ b/lib/legacy/zstd_v01.c
@@ -1735,6 +1735,7 @@ static size_t ZSTD_execSequence(BYTE* op,
 
     if (endMatch > oend) return ERROR(dstSize_tooSmall);   /* overwrite beyond dst buffer */
     if (litEnd > litLimit) return ERROR(corruption_detected);   /* overRead beyond lit buffer */
+    if (sequence.matchLength > (size_t)(*litPtr-op)) return ERROR(dstSize_tooSmall);  /* overwrite literal segment */
 
     /* copy Literals */
     ZSTD_memmove(op, *litPtr, sequence.litLength);   /* note : v0.1 seems to allow scenarios where output or input are close to end of buffer */
--- a/lib/legacy/zstd_v01.c
+++ b/lib/legacy/zstd_v01.c
@@ -1720,20 +1720,25 @@ static size_t ZSTD_execSequence(BYTE* op,
     static const int dec32table[] = {0, 1, 2, 1, 4, 4, 4, 4};   /* added */
     static const int dec64table[] = {8, 8, 8, 7, 8, 9,10,11};   /* subtracted */
     const BYTE* const ostart = op;
+    BYTE* const oLitEnd = op + sequence.litLength;
     const size_t litLength = sequence.litLength;
     BYTE* const endMatch = op + litLength + sequence.matchLength;    /* risk : address space overflow (32-bits) */
     const BYTE* const litEnd = *litPtr + litLength;
 
-    /* check */
+    /* checks */
+    size_t const seqLength = sequence.litLength + sequence.matchLength;
+
+    if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
+    if (sequence.litLength > (size_t)(litLimit - *litPtr)) return ERROR(corruption_detected);
+    /* Now we know there are no overflow in literal nor match lengths, can use pointer checks */
+    if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
+
     if (endMatch > oend) return ERROR(dstSize_tooSmall);   /* overwrite beyond dst buffer */
-    if (litEnd > litLimit) return ERROR(corruption_detected);
-    if (sequence.matchLength > (size_t)(*litPtr-op))  return ERROR(dstSize_tooSmall);    /* overwrite literal segment */
+    if (litEnd > litLimit) return ERROR(corruption_detected);   /* overRead beyond lit buffer */
 
     /* copy Literals */
-    if (((size_t)(*litPtr - op) < 8) || ((size_t)(oend-litEnd) < 8) || (op+litLength > oend-8))
-        memmove(op, *litPtr, litLength);   /* overwrite risk */
-    else
-        ZSTD_wildcopy(op, *litPtr, litLength);
+    ZSTD_memmove(op, *litPtr, sequence.litLength);   /* note : v0.1 seems to allow scenarios where output or input are close to end of buffer */
+
     op += litLength;
     *litPtr = litEnd;   /* update for next sequence */
 
--- a/lib/legacy/zstd_v03.c
+++ b/lib/legacy/zstd_v03.c
@@ -2706,18 +2706,24 @@ static size_t ZSTD_execSequence(BYTE* op,
     const BYTE* const litEnd = *litPtr + sequence.litLength;
 
     /* checks */
-    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);   /* last match must start at a minimum distance of 8 from oend */
+    size_t const seqLength = sequence.litLength + sequence.matchLength;
+
+    if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
+    if (sequence.litLength > (size_t)(litLimit - *litPtr)) return ERROR(corruption_detected);
+    /* Now we know there are no overflow in literal nor match lengths, can use the pointer check */
+    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);
+    if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
+
     if (oMatchEnd > oend) return ERROR(dstSize_tooSmall);   /* overwrite beyond dst buffer */
     if (litEnd > litLimit) return ERROR(corruption_detected);   /* overRead beyond lit buffer */
 
     /* copy Literals */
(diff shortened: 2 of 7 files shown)


lib/legacy/zstd_v03.c
```diff
@@ -2711,7 +2711,7 @@ static size_t ZSTD_execSequence(BYTE* op,
     if (litEnd > litLimit) return ERROR(corruption_detected);   /* overRead beyond lit buffer */
 
     /* copy Literals */
-    ZSTD_wildcopy(op, *litPtr, sequence.litLength);   /* note : oLitEnd <= oend-8 : no risk of overwrite beyond oend */
+    ZSTD_memmove(op, *litPtr, sequence.litLength);   /* note : used to be wildCopy, changed to fix a bug in 32-bit mode (oss-fuzz case 55714) */
```
terrelln: I don't think this fix is sufficient. We're supposed to have 8 extra bytes in the litBuffer, so this should be valid as long as `litEnd <= litLimit`.

I think the problem we're running into is that the limit checks should be re-written to avoid overflow. We don't care about speed, so we should do the obviously correct thing.

The offset check also looks janky, so we should clean that up as well.

E.g.

```c
size_t const seqLength = sequence.litLength + sequence.matchLength;

if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
if (sequence.litLength > (size_t)(litLimit -*litPtr)) return ERROR(corruption_detected);
/* Now we know we don't have any overflow in literal / match lengths, can use the pointer check for oend_8*/
if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);
if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
```

This bug is present in every version of the legacy decoder, so we'll have to fix every one.

lib/legacy/zstd_v01.c
```diff
@@ -1720,20 +1720,25 @@ static size_t ZSTD_execSequence(BYTE* op,
...
+    size_t const seqLength = sequence.litLength + sequence.matchLength;
+
+    if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
+    if (sequence.litLength > (size_t)(litLimit - *litPtr)) return ERROR(corruption_detected);
+    /* Now we know there are no overflow in literal nor match lengths, can use pointer checks */
+    if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
+
     if (endMatch > oend) return ERROR(dstSize_tooSmall);   /* overwrite beyond dst buffer */
-    if (litEnd > litLimit) return ERROR(corruption_detected);
-    if (sequence.matchLength > (size_t)(*litPtr-op))  return ERROR(dstSize_tooSmall);    /* overwrite literal segment */
```
terrelln, line 1730: I think that we need to keep this check, because the literals are put in the output buffer, but `oend` isn't reduced.

lib/legacy/zstd_v04.c
```diff
@@ -2828,13 +2828,20 @@ static size_t ZSTD_execSequence(BYTE* op,
...
-    /* check */
-    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);   /* last match must start at a minimum distance of 8 from oend */
+    /* checks */
+    size_t const seqLength = sequence.litLength + sequence.matchLength;
+
+    if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
+    if (sequence.litLength > (size_t)(litLimit - *litPtr)) return ERROR(corruption_detected);
+    /* Now we know there are no overflow in literal nor match lengths, can use pointer checks */
+    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);
+    if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
```
terrelln: Looks like we have extDict starting in v0.4. But the offset check below looks correct, so we can just delete this starting in this version.
```suggestion
```

lib/legacy/zstd_v05.c
```diff
@@ -3182,13 +3182,20 @@ static size_t ZSTDv05_execSequence(BYTE* op,
...
-    /* check */
-    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);   /* last match must start at a minimum distance of 8 from oend */
+    /* checks */
+    size_t const seqLength = sequence.litLength + sequence.matchLength;
+
+    if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
+    if (sequence.litLength > (size_t)(litLimit - *litPtr)) return ERROR(corruption_detected);
+    /* Now we know there are no overflow in literal nor match lengths, can use pointer checks */
+    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);
+    if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
```
terrelln:
```suggestion
```

lib/legacy/zstd_v06.c
```diff
@@ -3322,13 +3322,20 @@ static size_t ZSTDv06_execSequence(BYTE* op,
...
-    /* check */
-    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);   /* last match must start at a minimum distance of 8 from oend */
+    /* checks */
+    size_t const seqLength = sequence.litLength + sequence.matchLength;
+
+    if (seqLength > (size_t)(oend - op)) return ERROR(dstSize_tooSmall);
+    if (sequence.litLength > (size_t)(litLimit - *litPtr)) return ERROR(corruption_detected);
+    /* Now we know there are no overflow in literal nor match lengths, can use pointer checks */
+    if (oLitEnd > oend_8) return ERROR(dstSize_tooSmall);
+    if (sequence.offset > (U32)(oLitEnd - base)) return ERROR(corruption_detected);
```
terrelln:
```suggestion
```

terrelln (changes requested):
Just a few small changes left. And I think we also need to add the same checks to `zstd_v07.c`

terrelln:
@Cyan4973 the version compatibility test did catch the bug in `v0.5` in the dictionary decompression tests. We don't test `v0.4` and earlier in those tests though.

Cyan4973:
`v0.4` legacy decoder and above can be easily tested, using `zstd -t`.
The update (published above) is now able to decode to decode valid frames.

Below `v0.3`, this is more complex though, due to the lack of streaming capability.
For these cases, I modified the benchmark module, to trigger the one-pass decompression function and replace the size-query functions.

With this modified benchmark module (not published here), I could verify that the fixes for legacy decoders `v0.1`, `v0.2` and `v0.3` are still able to decode valid frames (generated with the corresponding `zstd` versions).

terrelln:
@Cyan4973 do we also need to fix `v0.7`?

terrelln:
Otherwise the PR looks good!

Cyan4973:
> @Cyan4973 do we also need to fix `v0.7`?

Well, it's not clear if that would be useful.
Previous versions `v0.4` `v0.5` `v0.6` have been _pre-emptively_ fixed on the ground of similarities with `v0.3`, but we are not even sure if there was really a problem.
On the other hand, `execSequence` in `v0.7` looks significantly different.
And it also has been subject to much more intensive fuzzing (than `v0.3`).

Cyan4973:
> > @Cyan4973 do we also need to fix `v0.7`?
> 
> Well, it's not clear if that would be useful, `execSequence` in `v0.7` looks significantly different and has been subject to much more intensive fuzzing up to now.

terrelln:
@Cyan4973 I see the same potential 32-bit pointer overflows in v0.7:

https://github.com/facebook/zstd/blob/df21aceaa72cb81428e67f08ff1f86ab7fa98814/lib/legacy/zstd_v07.c#L3555-L3556

> v0.7 has been subject to much more intensive fuzzing up to now.

We've built all of our OSS-Fuzz fuzzers with all legacy versions since 2019. But we haven't run any 32-bit fuzzers, where this issue could occur, until just now.

terrelln:
> we are not even sure if there was really a problem

There is definitely a pointer overflow bug in 32-bit mode. We have a similar fix in dev, but I guess we didn't apply it to all of our legacy decoders:

https://github.com/facebook/zstd/blob/df21aceaa72cb81428e67f08ff1f86ab7fa98814/lib/decompress/zstd_decompress_block.c#L991

The bug in v0.3 is occurring because of a pointer overflow. It happens that v0.3 allows very large literal lengths (in the fuzzed example it is `1965917`), which makes it easier to happen. But AFAIK there is no reason that the output or literals pointer cannot be within 128 KB of the end of the address space.

I'm also wondering if we need to guard against the end of the literals buffer being within 128KB of the end of the address space in dev. We can consider that in a separate PR though.

terrelln:
Versions `v0.4` and `v0.5` definitely have the same bug, because they both allow literal/match lengths up to ~2^24.

`v0.6` onwards do cap lengths at ~2^17, so aren't as susceptible, but we should definitely still fix it.

Cyan4973:
OK, checks of `v0.7` have been modified in a way which should be independent of address space overflow.

terrelln (approved):

PR #3484, opened 2023-02-08
Rename "External Matchfinder" to "Block-Level Sequence Producer"

embg:
"External matchfinder" was an inaccurate name, since the user-provided callback *selects* which match to use at each position as well as *finding* multiple matches for each position. The updated term we are using is "sequence provider".

Original PR introducing the old term (title of the old PR has been updated): https://github.com/facebook/zstd/pull/3333

Here are the commands I used to migrate names in `lib/` and `tests/`:
```
find ./ -type f \( -name "*.c" -o -name "*.h" \) -exec sed -i 's/externalMatchState/sequenceProducerState/g' {} \;
find ./ -type f \( -name "*.c" -o -name "*.h" \) -exec sed -i 's/ExternalMatchState/SequenceProducerState/g' {} \;
find ./ -type f \( -name "*.c" -o -name "*.h" \) -exec sed -i 's/externalMatchFinder/sequenceProducer/g' {} \;
find ./ -type f \( -name "*.c" -o -name "*.h" \) -exec sed -i 's/ExternalMatchFinder/SequenceProducer/g' {} \;
find ./ -type f \( -name "*.c" -o -name "*.h" \) -exec sed -i 's/ZSTD_EXTERNAL_MATCHFINDER_ERROR/ZSTD_SEQUENCE_PRODUCER_ERROR/g' {} \;
find ./ -type f \( -name "*.c" -o -name "*.h" \) -exec sed -i 's/ZSTD_c_enableMatchFinderFallback/ZSTD_c_enableSeqProducerFallback/g' {} \;
```

TODOs:

- [x] Update all public headers and `contrib/externalMatchfinder`
- [x] Update fuzzer makefile (references contrib)
- [x] Open an issue to track migration of internal names / comments in `lib/` and `tests/`. Not a release blocker for 1.5.4 since it's all internal.


lib/zstd.h
```diff
@@ -2113,19 +2113,19 @@ ZSTDLIB_STATIC_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const vo
...
  * Allowed values are 0 (disable) and 1 (enable). The default setting is 0.
  *
- * Controls whether zstd will fall back to an internal matchfinder if an
- * external matchfinder is registered and returns an error code. This fallback is
- * block-by-block: the internal matchfinder will only be called for blocks where
- * the external matchfinder returns an error code. Fallback compression will
- * follow any other cParam settings, such as compression level, the same as in a
+ * Controls whether zstd will fall back to an internal sequence producer if an
+ * external sequence producer is registered and returns an error code. This fallback
+ * is block-by-block: the internal sequence produer will only be called for blocks
```
embg:
```suggestion
 * is block-by-block: the internal sequence producer will only be called for blocks
```

lib/zstd.h
```diff
@@ -565,7 +565,7 @@ typedef enum {
  *                  They will be used to compress next frame.
  *                  Resetting session never fails.
  *  - The parameters : changes all parameters back to "default".
- *                  This also removes any reference to any dictionary or external matchfinder.
+ *                  This also removes any reference to any dictionary or external sequence provider.
```
embg:
```suggestion
 *                  This also removes any reference to any dictionary or external sequence producer.
```

lib/zstd.h
```diff
@@ -2870,90 +2870,89 @@ ZSTDLIB_STATIC_API size_t ZSTD_insertBlock    (ZSTD_DCtx* dctx, const void* bloc
...
+ * The external sequence producer can be "cleared" by registering a NULL function pointer.
+ * This removes all limitations described above in the "LIMITATIONS" section of the API docs.
  *
- * The user is strongly encouraged to read the full API documentation (above)
- * before calling this function. */
+ * The user is strongly encouraged to read the full API documentation (above) before calling
+ * this function. */
 ZSTDLIB_STATIC_API void
-ZSTD_registerExternalMatchFinder(
+ZSTD_registerExternalSequenceProducer(
```
Cyan4973: We should cut it to `SequenceProducer`, removing the `External` from the name.
This would be applied to below parameter names too.

Cyan4973 (approved):

Issue #3324, opened 2022-11-24
zstd decompression using "--patch-from"  cost memory too high (when the dict is a big file)


rongekuta:
**Is your feature request related to a problem? Please describe.**

`zstd -d -M128MB --patch-from=d:/code-shared/apks/wzry_old.apk  d:/code-shared/apks/wzry.diff -o d:/code-shared/apks/wzry_new_d.apk  `

when i use zstd (--patch-from) to decompress a patch file using a big dict, i find "-M#" option is not valid, the big dict is load to memory all, and cost too many memory.  i search code for reason, and find:
https://github.com/facebook/zstd/blob/v1.5.2/programs/fileio.c#L2019

these code cause "-M" memory limit is not work under "--patch-from"  mode

**Describe the solution you'd like**
decompression under a big dict with "--patch-from" is also be  limited memory

**Describe alternatives you've considered**
the big dict also be cache from file system, not load to memory totally

**Additional context**

JiapengLi:
a workaround is to use swap to bypass this limitation. 

This is what i do the test it under a docker container.:

Host enable a 4G swap area:

```bash
dd if=/dev/zero of=/swapfile bs=1M count=4096
mkswap /swapfile 
swapon /swapfile
swapon --show
free
```

docker compose file to creat a container
```yaml
version: "3"
services:
  zstd:
    restart: always
    image: some-image-contain-zstd
    container_name: zstd
    init: true
    privileged: true
    network_mode: host
    ulimits:
      nofile:
        soft: 64000
        hard: 64000
    deploy:
      resources:
        limits:
          memory: 512M
        reservations:
          memory: 128M
    memswap_limit: 6G
```

to test

```bash
docker exec -it zstd bash
> zstd -vv --memory=some_valueM --patch-from file1 file1-2.diff -o file2
```

It is very recommended to use zstd patch-from under docker if your file is very large, this can avoid zstd drain your memory and make bad things happen.

PR #3486, opened 2023-02-09
Mmap large dictionaries in patch-from mode

daniellerozenblit:
# TLDR
This PR is a response to issue #3324, which requests a lower memory version of zstd `--patch-from` for large dictionaries.

This PR offers a partial solution by memory-mapping large dictionaries (i.e. dictionary files larger than the user set memory limit) rather than mallocing and copying them entirely into memory. This feature can also be triggered manually, using the flag `--mmap-dict` or disabled using the flag `--no-mmap-dict`.

This PR additionally offers some FIO speed improvements by avoiding a superfluous system call to check the dictionary file size for `--patch-from` as well as avoiding `memcpy` for all FIO dictionary loading. This is most significant for `--patch-from`, where dictionaries often can be upwards of `1GB` in size, but should offer benefits for all FIO dictionary operations. 

# Benchmarking
I benchmarked on an Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz with core isolation and turbo disabled. For speed measurements, I ran each scenario five times interleaved and chose the highest result.

### Memory Usage
The memory saved by using a memory-mapped dictionary is highly dependent on the scenario. I did not see reduction in max RSS when creating the patch because my environment was not memory constrained and we access all parts of the dictionary during compression. During decompression, memory reduction is highly dependent on the patch size. We see greater savings in cases with larger diffs. Additionally, we see larger differences in max RSS when patching from a larger -> smaller file than when patching from a larger -> smaller file.

Screenshot 2023-03-08 at 6 16 46 PM

### Compression Speed
There are some improvements in compression / patch-creation speed for lower compression levels when using `mmap` vs `malloc` for dictionary creation. These speed improvements taper out at higher compression levels.

Screenshot 2023-03-09 at 10 32 29 AM

### Decompression Speed
There are some improvements in decompression speed across compression levels when using `mmap` vs `malloc`. I benchmarked`--patch-from` decompression with both `--mmap-dict` and `--no-mmap-dict` on the kernel source tree tarball (6.0 -> 6.2, which are ~1GB) and found that decompression speed improved by ~40% across a range of compression levels. 

Screenshot 2023-03-09 at 11 31 03 AM

## FIO Speed Improvements
We previously avoided `memcpy`-ing dictionaries for `--patch-from` compression. This PR additionally avoids calling `memcpy` on dictionary buffers for `--patch-from` decompression, as well as all non-patch-from compression and decompression calls.

I benchmarked standard `--patch-from` decompression (without `mmap`) on the kernel source tree tarball (6.0 -> 6.2, which are ~1GB) and found that decompression speed improved significantly (by ~50-60%) across a range of compression levels. 
Screenshot 2023-03-09 at 12 08 26 PM


programs/fileio.c
```diff
@@ -973,10 +1042,20 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
      * because of memLimit check inside it */
     if (prefs->patchFromMode) {
+        U64 const dictSize = UTIL_getFileSize(dictFileName);
         unsigned long long const ssSize = (unsigned long long)prefs->streamSrcSize;
-        FIO_adjustParamsForPatchFromMode(prefs, &comprParams, UTIL_getFileSize(dictFileName), ssSize > 0 ? ssSize : maxSrcFileSize, cLevel);
+        mmapDict = prefs->patchFromMode && dictSize > prefs->memLimit;
+        FIO_adjustParamsForPatchFromMode(prefs, &comprParams, dictSize, ssSize > 0 ? ssSize : maxSrcFileSize, cLevel);
+    }
+
+    mmapDict = 1;
```
daniellerozenblit: Just doing this for testing

programs/fileio.c
```diff
@@ -973,10 +1041,19 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
     /* need to update memLimit before calling createDictBuffer
      * because of memLimit check inside it */
     if (prefs->patchFromMode) {
+        U64 const dictSize = UTIL_getFileSize(dictFileName);
```
daniellerozenblit: We should reorganize this so that we aren't stat-ing a file multiple times for patch-from.

jluebbe:
On 32 bit systems, `mmap()` of files approaching 1GB can be difficult. Would it be possible to (probably optionally) `read()` the necessary data from the `--patch-from` file as needed? The performance would obviously be lower, but it would scale to patching very large files (such as disk images).

Cyan4973:
> On 32 bit systems, `mmap()` of files approaching 1GB can be difficult. Would it be possible to (probably optionally) `read()` the necessary data from the `--patch-from` file as needed? The performance would obviously be lower, but it would scale to patching very large files (such as disk images).

It's complex enough to deserve being a separate development effort.

programs/fileio.c
```diff
@@ -2043,43 +2131,65 @@ int FIO_compressMultipleFilenames(FIO_ctx_t* const fCtx,
...
+    size_t dictBufferSize;
     ZSTD_DStream* dctx;
     WritePoolCtx_t *writeCtx;
     ReadPoolCtx_t *readCtx;
+    int mmapDict;
 } dRess_t;
 
 static dRess_t FIO_createDResources(FIO_prefs_t* const prefs, const char* dictFileName)
 {
+    int mmapDict = prefs->mmapDict;
```
Cyan4973: minor: I struggled a bit to understand what the `mmapDict` variable is.
It turns out, it's not the mmapped dictionary itself, but just a flag, that selects using `mmap` or not to load the reference content.

Naming is hard. Please consider a name that's closer to the role of the variable. For example, `use_mmap` comes to mind.

programs/zstdcli.c
```diff
@@ -984,6 +986,7 @@ int main(int argCount, const char* argv[])
                 if (longCommandWArg(&argument, "--adapt=")) { adapt = 1; if (!parseAdaptParameters(argument, &adaptMin, &adaptMax)) { badusage(programName); CLEAN_RETURN(1); } continue; }
                 if (!strcmp(argument, "--single-thread")) { nbWorkers = 0; singleThread = 1; continue; }
                 if (!strcmp(argument, "--format=zstd")) { suffix = ZSTD_EXTENSION; cType = FIO_zstdCompression; continue; }
+                if (!strcmp(argument, "--mmap-dict")) { mmapDict = 1; continue; }
```
Cyan4973: Could it be possible to have a reverse `--no-mmap-dict` command ?

programs/fileio.c
```diff
@@ -2043,43 +2131,65 @@ int FIO_compressMultipleFilenames(FIO_ctx_t* const fCtx,
...
 
     ress.writeCtx = AIO_WritePool_create(prefs, ZSTD_DStreamOutSize());
     ress.readCtx = AIO_ReadPool_create(prefs, ZSTD_DStreamInSize());
-
     return ress;
 }
 
 static void FIO_freeDResources(dRess_t ress)
 {
+    if (!ress.mmapDict) {
```
Cyan4973: minor defensive programming ideas : 

In this case, `mmapDict` is effectively an `enum`-like value, that tells if `dictBuffer` has been allocated with `malloc()` or `mmap()`, in order to invoke the correct `free()` function.

- Reduce confusion between `ress.mmapDict` which tells the nature of the buffer (for proper freeing), from `prefs->mmapDict` which is a command to allocate the buffer. I would use 2 different names to ensure they are not confused.
-  I would imagine `dictBuffer`, `dictBufferSize`, and `dictBufferType` to be grouped into a structure to form a kind of "object", that can then be passed to a dedicated free function. And conversely, the "object" would be created from a dedicated function. This would keep the mmap-or-not logic purely internal.

Cyan4973:
The code looks good, I only have some fairly minor comments.

The code should result in some memory savings when triggered by the proper scenario. For example, using `--patch-from` with a 1 GB reference file.
Could you document the memory benefits, and eventually the performance costs, from using this patch, in the description of the PR ?

programs/fileio.c
```diff
@@ -769,20 +769,24 @@ static size_t FIO_createDictBufferMMap(void** bufferPtr, const char* fileName, F
...
 }
 #endif
 
+static void FIO_freeDict(const FIO_Dict_t* dict) {
+    if (dict->dictBufferType == FIO_mallocDict) {
+        free(dict->dictBuffer);
+    } else if (dict->dictBufferType == FIO_mmapDict)  {
+        FIO_munmap(dict->dictBuffer, dict->dictBufferSize);
+    }
+}
```
Cyan4973, line 791: minor: I would add an `else assert(0);` at the end, 
because such a case is supposed to never happen,
unless the `FIO_Dict_t` code is changed in the future.

Cyan4973 (approved):
This looks great !

Just a minor defensive-programming comment, and it's basically good to go !

Dimiturpenev (commented):
No

Dimiturpenev:
> # TLDR
> This PR is a response to issue #3324, which requests a lower memory version of zstd `--patch-from` for large dictionaries.
> 
> This PR offers a partial solution by memory-mapping large dictionaries (i.e. dictionary files larger than the user set memory limit) rather than mallocing and copying them entirely into memory. This feature can also be triggered manually, using the flag `--mmap-dict` or disabled using the flag `--no-mmap-dict`.
> 
> This PR additionally offers some FIO speed improvements by avoiding a superfluous system call to check the dictionary file size for `--patch-from` as well as avoiding `memcpy` for all FIO dictionary loading. This is most significant for `--patch-from`, where dictionaries often can be upwards of `1GB` in size, but should offer benefits for all FIO dictionary operations. 
> 
> # Benchmarking
> I benchmarked on an Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz with core isolation and turbo disabled. For speed measurements, I ran each scenario five times interleaved and chose the highest result.
> 
> ## Memory-Mapped Dictionaries For Patch-From
> ### Memory Usage
> The memory saved by using a memory-mapped dictionary is highly dependent on the scenario. I did not see reduction in max RSS when creating the patch because my environment was not memory constrained and we access all parts of the dictionary during compression. During decompression, memory reduction is highly dependent on the patch size. We see greater savings in cases with larger diffs. Additionally, we see larger differences in max RSS when patching from a larger -> smaller file than when patching from a larger -> smaller file.
> 
> Screenshot 2023-03-08 at 6 16 46 PM
> 
> ### Compression Speed
> There are some improvements in compression / patch-creation speed for lower compression levels when using `mmap` vs `malloc` for dictionary creation. These speed improvements taper out at higher compression levels.
> 
> Screenshot 2023-03-09 at 10 32 29 AM
> 
> ### Decompression Speed
> There are some improvements in decompression speed across compression levels when using `mmap` vs `malloc`. I benchmarked`--patch-from` decompression with both `--mmap-dict` and `--no-mmap-dict` on the kernel source tree tarball (6.0 -> 6.2, which are ~1GB) and found that decompression speed improved by ~40% across a range of compression levels. 
> 
> Screenshot 2023-03-09 at 11 31 03 AM
> 
> ## FIO Speed Improvements
> We previously avoided `memcpy`-ing dictionaries for `--patch-from` compression. This PR additionally avoids calling `memcpy` on dictionary buffers for `--patch-from` decompression, as well as all non-patch-from compression and decompression calls.
> 
> I benchmarked standard `--patch-from` decompression (without `mmap`) on the kernel source tree tarball (6.0 -> 6.2, which are ~1GB) and found that decompression speed improved significantly (by ~50-60%) across a range of compression levels. 
> Screenshot 2023-03-09 at 12 08 26 PM
>

programs/fileio.c
Sindhuja365: Ok

PR #3491, opened 2023-02-10
Github Action to generate Win64 artifacts

Cyan4973:


Cyan4973:
Unfortunately, this workflows produces `msys2` binaries, not `mingw64` ones.
To be updated.

Cyan4973:
OK, now this workflow produces `mingw64` binaries and correctly interpret `-static`,
so the produced binaries can be run on any Windows system
(tested on a local Windows 10 laptop).

PR #3494, opened 2023-02-10
Document `pzstd -p`

bmwiedemann:
Document `pzstd -p` in man-page
so that users get at least the same information as from `pzstd --help`


Cyan4973:
`pzstd` is not part of the standard `zstd` package. It's an additional tool.
Maybe it could have its own `man` page, but I don't think it's a good idea to mix it within the `zstd.1` man page.
cc @terrelln .

Cyan4973:
Closing.
If we get to create a `man` page for `pzstd`, it will be a separate one from `zstd`.

PR #3495, opened 2023-02-10
Fix all MSVC warnings

### LINKED ISSUE #3489: MSVC with AVX2 compile option, C4244 warnings.
If build with AVX2 instruction set, MSVC has dozens of C4244 warnings:
`warning C4244: 'return': conversion from 'unsigned __int64' to 'unsigned int', possible loss of data`

```c
MEM_STATIC unsigned ZSTD_countTrailingZeros64(U64 val)
{
    assert(val != 0);
#   if defined(_MSC_VER) && defined(_WIN64)
#       if STATIC_BMI2 == 1
            return _tzcnt_u64(val);
#       else
```
Adding `(unsinged)` before `_tzcnt_u64(val)` should fixes this issue.

embg:
* Fix https://github.com/facebook/zstd/issues/3489
* Add a test to reproduce https://github.com/facebook/zstd/issues/3489.
    * I confirmed that the issue reproduces in CI before fixing: https://github.com/embg/zstd/actions/runs/4147168689/jobs/7173726044#step:5:9
* Fix an incorrect MSVC 2019 compiler warning
* Remove a deprecated MSVC 2022 option ([context](https://learn.microsoft.com/en-us/cpp/build/reference/gm-enable-minimal-rebuild?view=msvc-170&viewFallbackFrom=vs-2017)), which was triggering a warning.
* Add `/warnaserror` to our `msbuild` CI jobs so we don't have these issues in future releases.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -259,24 +259,34 @@ jobs:
     strategy:
       matrix:
         include: [
-          { name: "VS 2022 x64 Debug", platform: x64, configuration: Debug, toolset: v143, runner: "windows-2022"},
-          { name: "VS 2022 Win32 Debug", platform: Win32, configuration: Debug, toolset: v143, runner: "windows-2022"},
-          { name: "VS 2022 x64 Release", platform: x64, configuration: Release, toolset: v143, runner: "windows-2022"},
-          { name: "VS 2022 Win32 Release", platform: Win32, configuration: Release, toolset: v143, runner: "windows-2022"},
-          { name: "VS 2019 x64 Release", platform: Win32, configuration: Release, toolset: v142, runner: "windows-2019"},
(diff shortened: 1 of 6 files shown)


embg:
It looks like my second commit, adding `/warnaserror` (the msbuild equivalent of `-Werror`), exposed a warning in the MSVC 2019 build. Will investigate ASAP.

embg:
The MSVC 2019 build warnings were introduced by https://github.com/facebook/zstd/pull/3449. The errors are all caused by `memcpy` operations from `BYTE const * *` to `BYTE const * *`. The compiler incorrectly believes the `const` qualifiers don't match.

This warning doesn't show up in the MSVC 2022 build, so I think it is a bug in MSVC 2019.

Example warning (which is an error now that we have `/warnaserror`): https://github.com/facebook/zstd/actions/runs/4147381724/jobs/7174208209#step:4:207

Fixed by adding `void*` casts to the destination of each offending `memcpy`.

Cyan4973 (approved):

Issue #3492, opened 2023-02-10
Cant build zstd-dll with 1.5.4


nolange:
**Describe the bug**

**To Reproduce**

This is on debian testing x86

``` bash
make -j9 PREFIX=/usr ZSTD_LEGACY_SUPPORT=0 HAVE_ZLIB=0 HAVE_LZMA=0 HAVE_LZ4=0 MOREFLAGS="-Os" HAVE_THREAD=1 -C lib libzstd.pc-mt-release lib-mt-release
make -j9 PREFIX=/usr ZSTD_LEGACY_SUPPORT=0 HAVE_ZLIB=0 HAVE_LZMA=0 HAVE_LZ4=0 MOREFLAGS="-Os" HAVE_THREAD=1 -C programs zstd-dll
```

**Expected behavior**

zstd library and program being build (like it worked with 1.5.2)

**Actutal behavior**

```
LINK obj/conf_2f7cbea6a959ba80ad7e06b8fde79417/zstd
[...]

PR #3496, opened 2023-02-10
Fix zstd-dll build missing dependencies

yoniko:
Fixes zstd-dll build (#3492 ):
- Adds pool.o and threading.o dependency to the zstd-dll target
- Moves custom allocation functions into header to avoid needing to add dependency on common.o
- Adds test target for zstd-dll
- Adds github workflow for test zstd-dll


lib/common/zstd_internal.h
```diff
@@ -24,6 +24,8 @@
 #include "mem.h"
 #include "debug.h"                 /* assert, DEBUGLOG, RAWLOG, g_debuglevel */
 #include "error_private.h"
+#define ZSTD_DEPS_NEED_MALLOC
```
yoniko: @Cyan4973 - Is there any issue with adding this to `zstd_internal.h`?

Cyan4973: Well, `zstd_internal.h` is used everywhere within `libzstd`, so this change can impact a lot of places.
I'm not sure if I can understand all the implications.

Cyan4973:
This PR manages to get `zstd-dll` recipe working by adding 2 units, `threading.c` and `pool.c`.
This is an improvement over a naive solution which requires 4 units, including `zstd_common.c` and `error_private.c`.

To achieve that goal, this PR moves `customMalloc()` and co from `zstd_common.c` to `zstd_internal.h`.
This operation requires making `zstd_deps.h` a dependency of `zstd_internal.h`, and therefore, transitively, a dependency of almost all source files inside `libzstd`.

Now, I'm not exactly pleased about this dependency story.
But on the other hand, I don't think we lose too much in this PR specifically.

`zstd_common.c` is not terribly specific, 
moving functions from there to `zstd_internal.h` doesn't improve the situation, but it does not make it significantly worse either.

Bundling more definitions within `zstd_internal.h` will transitively increase compilation load time of all units depending on it. But the increased workload is likely very light, so it's unlikely to have a big impact.

All in all, it seems this is an acceptable trade off.

Longer term, I'm not pleased to see `zstd_internal.h` employed as a "bag of everything" which transitively includes a ton of other stuffs. Initially, this file was designed to hold common information that should not be duplicated between compression and decompression, such as the Frame ID number. It has largely deviated from this original scope, and I think this doesn't respect the `libzstd` model of minimal dependency at each step. Hence, I'll be pleased to see this bundle gradually separated into more specific packets that are included only when needed.

But that's a topic for a different effort, and since this PR is focused on `zstd-dll`, it wouldn't be fair to address that here.

yoniko:
@Cyan4973 - all tests pass. Please see most recent changes, if it looks good to you we can merge.

Cyan4973 (approved):

PR #3497, opened 2023-02-11
Use correct types in LZMA comp/decomp

Hello71:
Bytef and uInt are zlib types, not available when zlib is disabled

Fixes: 1598e6c634ac ("Async write for decompression")
Fixes: cc0657f27d81 ("AsyncIO compression part 2 - added async read and asyncio to compression code (#3022)")


Cyan4973:
Thanks @Hello71 ,

I wonder if we should consider adding a test with `lzma` linking too,
since apparently this issue evaded CI attention.

yoniko (approved):

PR #3498, opened 2023-02-11
Add 64-bit detection for LoongArch

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:

### DIFF
--- a/programs/platform.h
+++ b/programs/platform.h
@@ -41,6 +41,7 @@ extern "C" {
   || defined __x86_64__s || defined _M_X64                                                                          /* x86 64-bit */    \
   || defined __arm64__ || defined __aarch64__ || defined __ARM64_ARCH_8__                                           /* ARM 64-bit */    \
   || (defined __mips  && (__mips == 64 || __mips == 4 || __mips == 3))                                              /* MIPS 64-bit */   \
+  || defined __loongarch64                                                                                      /* LoongArch 64-bit */  \
   || defined _LP64 || defined __LP64__ /* NetBSD, OpenBSD */ || defined __64BIT__ /* AIX */ || defined _ADDR64 /* Cray */               \
   || (defined __SIZEOF_POINTER__ && __SIZEOF_POINTER__ == 8) /* gcc */
 #  if !defined(__64BIT__)


Cyan4973:
Is there any `LoongArch` core for `qemu` that could be added to CI,
so that we can test that this PR works,
and ensure that the architecture remains supported in the future (in case a contribution would break such support) ?

_edit_ : mmmh, according to [this article](https://www.phoronix.com/news/QEMU-7.1-Released), support for LoongArch in `qemu` is _very_ recent, so that might not be practical for CI (yet)...

ilyakurdyukov:
```
  || defined _LP64 || defined __LP64__ /* NetBSD, OpenBSD */ || defined __64BIT__ /* AIX */ || defined _ADDR64 /* Cray */               \
  || (defined __SIZEOF_POINTER__ && __SIZEOF_POINTER__ == 8) /* gcc */
```

I want to note that good compilers should use some of these macros. So if your compiler doesn't define any of these macros, you should file a bug report for that compiler so you don't have to add your new architecture to every whitelist of 64-bit architectures.

xry111:
GCC has `__LP64__` for 64-bit LoongArch target:

```
$ lscpu | grep Arch
Architecture:        loongarch64
$ echo __LP64__ | cpp -nostdinc
# 0 "<stdin>"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "<stdin>"
1
```

ilyakurdyukov:
> GCC has `__LP64__` for 64-bit LoongArch target:

Then this patch is not needed. It looks like some Chinese people want to add a reference to LoongArch to every project, even if it's absolutely unnecessary.

xry111:
And `__loongarch64` is deprecated (see table 15 in [the doc](https://loongson.github.io/LoongArch-Documentation/LoongArch-toolchain-conventions-EN.html)) because we don't really know if it means `__loongarch_lp64` (the ABI) or `__loongarch_grlen == 64` (the HW register).

I think we would add a deprecate warning for it into GCC considering many developers don't RTFM.

Cyan4973:
OK, in essence, this does not seem a useful macro to add

PR #3499, opened 2023-02-11
contrib/pzstd/Makefile: drop -std=c++11 to fix build of the test

ldv-alt:
It's not just strange to require an older version of the language standard that the compiler supports by default, it's also blocks building the test which requires a newer version of the language standard nowadays:

  In file included from /usr/include/gtest/gtest-message.h:57,
                   from /usr/include/gtest/gtest-assertion-result.h:46,
                   from /usr/include/gtest/gtest.h:64,
                   from utils/test/ScopeGuardTest.cpp:11:
  /usr/include/gtest/internal/gtest-port.h:270:2: error: #error C++ versions less than C++14 are not supported.


Cyan4973:
cc @terrelln

felixhandte:
We've landed a different fix to this issue, so I'm going to close this PR. Thanks for reporting!

PR #3505, opened 2023-02-14
CI Workflow for external compressors dependencies

yoniko:
Implemented CI workflow for testing compilation with external compressors and without them. This serves as a sanity check to avoid any code dependencies on libraries that may not always be present. (Reference: https://github.com/facebook/zstd/pull/3497 for a bug fix related to this issue.)

Note: the new job added here is expected to fail as https://github.com/facebook/zstd/pull/3497 isn't merged yet.
This will give us confirmation that it does find the issue.
After #3497 this will be rebased to make sure the job doesn't fail anymore.


Cyan4973 (approved):
Great new test !

Issue #3502, opened 2023-02-13
"window-resize.sh" from "cli-tests" (zstd-1.5.4) fail on 32-bit targets.


ilyakurdyukov:
**Describe the bug**

It looks like "window-resize.sh" doesn't work on 32-bit architectures due to lack of memory?

```
FAIL: compression/window-resize.sh
FAIL: compression/window-resize.sh.check_exit
Exit code mismatch! Expected 0 but got 1
FAIL: compression/window-resize.sh.check_stdout
stdout does not match!
> diff expected actual
---
< Window Size: 1.000 GiB (1073741824 B)
```

**To Reproduce**
I got this issue when building for "i586" and "armh" targets (both 32-bit) for a linux distro, don't know if I can easily reproduce this.

zwenna:
The problem seems to be that window-resize.sh invokes zstd with the --long=31 option which is not valid on 32-bit architectures, and zstd complains about that.

zstd: error 11 : Parameter is out of bound

I have not tested whether changing the option to "long=30" lets the test succeed, but I expect it will.

ilyakurdyukov:
Well actually it's easy to reproduce:

```
make CFLAGS="-O2 -m32"
make -C tests test-cli-tests CFLAGS="-O2 -m32"
```

For example, on Ubuntu with "gcc-multilib" installed.

ilyakurdyukov:
"window-resize.sh" stderr:
```
zstd: error 11 : Parameter is out of bound 
*** Zstandard CLI (32-bit) v1.5.4, by Yann Collet ***
Error : file.zst is not a file 
rm: cannot remove 'file.zst': No such file or directory
```

ilyakurdyukov:
> I have not tested whether changing the option to "long=30" lets the test succeed, but I expect it will.

You're right, the test passes with "--long=30".

Issue #3501, opened 2023-02-13
The new "cli-tests" from 1.5.4 fail if zstd is built with HAVE_ZLIB=0.


ilyakurdyukov:
**Describe the bug**

Please make proper tests that won't fail if something is built with non-default options:

```
FAIL: compression/basic.sh
FAIL: compression/gzip-compat.sh
FAILED 2 / 37 tests!
```

"basic.sh" stderr:
```
# Test keeping input file when compressing to stdout in gzip mode
$ZSTD_SYMLINK_DIR/gzip -c file       | zstd -t ; test -f file
zstd: error 20 : zstd: file: file cannot be compressed as gzip (zstd compiled without ZSTD_GZCOMPRESS) -- ignored 
```

"gzip-compat.sh" stderr:
```
# Test gzip specific compression option
$ZSTD_SYMLINK_DIR/gzip --fast file ; $ZSTD_SYMLINK_DIR/gzip -d file.gz
zstd: error 20 : zstd: file: file cannot be compressed as gzip (zstd compiled without ZSTD_GZCOMPRESS) -- ignored 
```

**To Reproduce**
```
make HAVE_ZLIB=0
make -C tests test-cli-tests HAVE_ZLIB=0
```

Cyan4973:
Indeed, these tests should first ensure that the tested binary features `gzip` support before proceeding.

PR #3509, opened 2023-02-14
Fix cli-tests issues

daniellerozenblit:
Fix for issues #3501 and #3502. 

All tests now seem to pass for `make -C tests test-cli-tests CFLAGS="-O2 -m32"` and `make -C tests test-cli-tests HAVE_ZLIB=0`.


Cyan4973:
Thanks @daniellerozenblit ! Great fix !

Another point : it seems we have been missing this issue, because we were lacking the relevant test to detect it.

Could you add it to this PR too ?
Thanks!

Cyan4973 (approved):

PR #3510, opened 2023-02-14
Don't require CMake 3.18 or later

### LINKED ISSUE #3500: CMake: bump min CMake version to 3.18 or remove usage of `check_linker_flag()`
https://github.com/facebook/zstd/pull/3392 (part of zstd 1.5.4) checks linker flags in CMake build by relying on https://cmake.org/cmake/help/latest/module/CheckLinkerFlag.html, a function available since CMake 3.18. But `cmake_minimum_required()` of main CMakeLists is still 2.8.12.
Is it intentional to use such recent feature? If it is, `cmake_minimum_required()` should be updated accordingly. Otherwise it's possible to achieve the same checks by using `check_source_compiles()` and therefore avoid constraining CMakeLists of zstd to such recent CMake version.

kou:
fix #3500

CMake 3.18 or later was required by #3392. Because it uses `CheckLinkerFlag`. But requiring CMake 3.18 or later is a bit aggressive. Because Ubuntu 20.04 LTS still uses CMake 3.16.3: https://packages.ubuntu.com/search?keywords=cmake

This change disables `-z noexecstack` check with old CMake. This will not break any existing users. Because users who need `-z noexecstack` must already use CMake 3.18 or later.

### DIFF
--- a/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
+++ b/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
@@ -1,6 +1,15 @@
 include(CheckCXXCompilerFlag)
 include(CheckCCompilerFlag)
-include(CheckLinkerFlag)
+# VERSION_GREATER_EQUAL requires CMake 3.7 or later.
+# https://cmake.org/cmake/help/latest/command/if.html#version-greater-equal
+if (CMAKE_VERSION VERSION_LESS 3.18)
+    set(ZSTD_HAVE_CHECK_LINKER_FLAG false)
+else ()
+    set(ZSTD_HAVE_CHECK_LINKER_FLAG true)
+endif ()
+if (ZSTD_HAVE_CHECK_LINKER_FLAG)
+    include(CheckLinkerFlag)
+endif()
 
 function(EnableCompilerFlag _flag _C _CXX _LD)
     string(REGEX REPLACE "\\+" "PLUS" varname "${_flag}")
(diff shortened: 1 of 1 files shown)


botanegg:
Instead of just disable this flag it can be done by `try_compile` like this:
https://github.com/mctools/ncrystal/blob/a263fb0efa4550239ef9fe7bb22e63d951d56206/CMakeLists.txt#L201-L216

kou:
The shim implementation will increase maintenance cost a bit.

kou:
The shim implementation will increase maintenance cost a bit.
Do we need to implement it?

The author of #3392 uses `CheckLinkerFlag`. It seems that users who need this feature use newer CMake. So the shim implementation may not be needed for real users.

If the shim implementation is needed for this project, I can do it.

terrelln (approved):
Thanks for the PR @kou!

terrelln:
I think it is reasonable to simply disable the extra ldflags when `check_linker_flags()` is not available, since the only flags we currently use are optional.

I think we should just add a comment to the function to clarify that the ldflags don't get added for old cmake versions.

build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
```diff
@@ -20,7 +29,11 @@ function(EnableCompilerFlag _flag _C _CXX _LD)
         endif ()
     endif ()
     if (_LD)
-        CHECK_LINKER_FLAG(C ${_flag} LD_FLAG_${varname})
+        if (ZSTD_HAVE_CHECK_LINKER_FLAG)
```
terrelln, line 36: Can you just add a comment to this function that says that it doesn't add ldflags for older cmake versions? In case that we need to add other ldflags that are non-optional later.

kou: It makes sense. I've add a comment:

```cmake
        # We never add a linker flag with CMake < 3.18. We will
        # implement CHECK_LINKER_FLAG() like feature for CMake < 3.18
        # or require CMake >= 3.18 when we need to add a required
        # linker flag in future.
```

Does this make sense?

terrelln (approved):
Thanks @kou!

kou:
@terrelln Thanks! Do you have a plan to release a new version with this change soon? (If it's done, apache/arrow can use the latest Zstandard. But it's not a blocker. So apache/arrow can wait a new version.)

PR #3511, opened 2023-02-15
Fix Permissions on Publish Release Artifacts Job

felixhandte:
Publishing release artifacts requires the `contents` permission, as documented by: https://docs.github.com/en/rest/overview/permissions-required-for-github-apps.

Tested by authoring a release on my fork.

### DIFF
--- a/.github/workflows/publish-release-artifacts.yml
+++ b/.github/workflows/publish-release-artifacts.yml
@@ -10,8 +10,7 @@ permissions: read-all
 jobs:
   publish-release-artifacts:
     permissions:
-      contents: read # to fetch code (actions/checkout)
-      actions: write # to attach binaries to release artifacts (skx/github-action-publish-binaries)
+      contents: write # to fetch code and upload artifacts
 
     runs-on: ubuntu-latest
     if: startsWith(github.ref, 'refs/tags/')


Cyan4973 (approved):

.github/workflows/publish-release-artifacts.yml
```diff
@@ -10,8 +10,7 @@ permissions: read-all
 jobs:
   publish-release-artifacts:
     permissions:
-      contents: read # to fetch code (actions/checkout)
-      actions: write # to attach binaries to release artifacts (skx/github-action-publish-binaries)
+      contents: write # to fetch code and upload artifacts
```
Cyan4973, line 13: It's more a question at this point,
I'm just curious to notice that
a local permission setting `contents: write` within `publish-release-artifacts`
seems to overwrite, or at least give more rights than the global `permissions: read-all` just above.

I was wondering if this would allow an included script or action to sneakily "update" permissions in its own code even when permissions are set to `read-all` at the beginning.

felixhandte: I believe not--I believe the included action can't modify the job's permission.

A PR could change the permissions though. Supposedly we are protected from that though, as PRs from (first-time?) contributors require approval before actions are run. So we can inspect to make sure they aren't messing with the action definitions.

PR #3516, opened 2023-02-20
fullbench with two files 

dloidolt:
When benchmarking two files with fullbench, the second file will not be benchmarked because the benchNb has not been reset to zero.
Screenshot 2023-02-20 at 16 55 30


Cyan4973 (approved):

Cyan4973:
Thanks @dloidolt, it's nice simple fix that works well.

PR #3517, opened 2023-02-22
[bug-fix] Fix rare corruption bug affecting the block splitter

terrelln:
TLDR: This is a very rare bug affecting the block splitter. Only levels using the optimal parser (levels 13 and above depending on the source size), or users who explicitly enable the block splitter, are affected.

The block splitter confuses sequences with literal length == 65536 that use a repeat offset code. It interprets this as literal length == 0 when deciding the meaning of the repeat offset, and corrupts the repeat offset history. This is benign, merely causing suboptimal compression performance, if the confused history is flushed before the end of the block, e.g. if there are 3 consecutive non-repeat code sequences after the mistake. It also is only triggered if the block splitter decided to split the block.

All that to say: This is a rare bug, and requires quite a few conditions to trigger. However, the good news is that if you have a way to validate that the decompressed data is correct, e.g. you've enabled zstd's checksum or have a checksum elsewhere, the original data is very likely recoverable. So if you were affected by this bug please reach out.

The fix is to remind the block splitter that the literal length is actually 64K. The test case is a bit tricky to set up, but I've managed to reproduce the issue.

Thanks to @danlark1 for alerting us to the issue and providing us a reproducer!


tests/fuzzer.c
```diff
@@ -1286,6 +1286,64 @@ static int basicUnitTests(U32 const seed, double compressibility)
...
+        U32 const compressibleLenU32 = 32 * 1024 / 4;
+        U32 const blockSizeU32 = 128 * 1024 / 4;
+        U32 const litLenU32 = 64 * 1024 / 4;
+        U32* data = (U32*)malloc(srcSize);
+        size_t dSize;
+
+        if (data == NULL || cctx == NULL) goto _output_error;
+
+        /* Generate data without any matches */
+        RDG_genBuffer(data, srcSize, 0.0, 0.01, 2654435761U);
```
embg, line 1301: Is this sensitive to the seed `2654435761U`, in the sense that future implementations of `RDG_genBuffer` may not generate exactly zero matches for this seed? If so, perhaps we should use a [deBruijn sequence generator](https://gist.github.com/embg/4f445b51c362ed1dbd6267b13d1e4f11)?

Cyan4973 (approved):
Great fix @terrelln ! That's indeed a pretty rare corner case.

danlark1:
Thanks a lot, am I right it was introduced with zstd 1.5.0 when blocksplitter was released?

terrelln:
> Thanks a lot, am I right it was introduced with zstd 1.5.0 when blocksplitter was released?

Correct, the first affected version was zstd v1.5.0.

tests/fuzzer.c
terrelln: @embg yeah it would make sense to use that, and to move this to a CLI test with a golden file. So changes to `RDG_genBuffer()` don't affect the test.

embg (approved):

georgthegreat:
@terrelln, could you tag a release version with this fix?

Cyan4973:
A `v1.5.5` release version will be produced in March including this fix.

rationalex:
@terrelln could you please share with us a snippet that reproduces this bug? I'd like to use it to check whether another zstd implementation (https://github.com/klauspost/compress/tree/master/zstd to be specific) has the same issue.

Issue #3525, opened 2023-03-06
ZSTD_initStaticCDict does not set level and it results in use of uninit value


danlark1:
**Describe the bug**

ZSTD_initStaticCDict does not set compression level

**To Reproduce**

While running under msan, we found an error

```
==9312==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7f4749a338a5 in ZSTD_compressBegin_usingCDict_internal zstd/compress/zstd_compress.c:5503:9
    #1 0x7f4749a339e3 in ZSTD_compress_usingCDict_internal zstd/compress/zstd_compress.c:5548:5
    #2 0x7f4749a339e3 in ZSTD_compress_usingCDict zstd/compress/zstd_compress.c:5574:12
    #3 0x7f474a860012 in basicUnitTests zstd/tests/fuzzer.c:2596:29
    #4 0x7f474a8505b2 in main zstd/tests/fuzzer.c:4734:18
```

This is the error

```cpp
ZSTD_CCtxParams_init_internal(&cctxParams, &params, cdict->compressionLevel);
```

cdict->compressionLevel was not set during initialization

**Expected behavior**

No msan error

**Desktop (please complete the following information):**
 - OS: Linux
 - Compiler: clang

**Additional context**
Should be easy to fix

yoniko:
Thank you for reporting @danlark1.

`cdict->compressionLevel` is indeed not set in all cases which should be fixed.
After experimentation I managed to reproduce by compiling with `-O0`, so I believe the reason we haven't caught it before is because the compiler is inlining / optimizing things away.

I'll add initialization and verify that we cover this case in CI.

yoniko:
Fixed in #3257.
@danlark1 - please verify fix works for you.

danlark1:
It works, thanks!

PR #3527, opened 2023-03-07
Add initialization of clevel to static cdict (#3525)

yoniko:
- Initializes clevel in `ZSTD_CCtxParams_init`
- Adds CI workflow for msan fuzzers runs without optimization (`-O0`)
- Fixes Makefile to correctly pass on user defined `MOREFLAGS` and `FUZZER_FLAGS` in cases they have been overwritten

UPDATE: CI failed without the fix, fix is now enabled

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -186,6 +186,16 @@ jobs:
     - name: MSan + Regression Test
       run: make -j msanregressiontest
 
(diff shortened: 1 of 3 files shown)


yoniko:
`clang-msan-fuzz-unoptimized` has failed as expected, verifying that the new CI flow does catch the issue. Going to amend the fix to make sure all CI tests pass now.

Cyan4973 (approved):

PR #3529, opened 2023-03-07
Add init once memory (#3528)

### LINKED ISSUE #3528: Row hash tag space initialization speed regression
This issue will be used to track the work that has been started in #2971 and #3426 and further work.

**Context**
Row hash is a fast SIMD-based hash used by various strategies in Zstd.
Other than the normal hash entries it requires an additional space for tags that are hash based and allow further filtration of entries in a bucket.

When streaming data of unknown size (for example, using `ZSTD_compressStream`) we don't have a good way to choose a hashlog and so we pick a large one. This, in turn, makes it so we need to initialize a large tag space.
This creates a noticeable regression when compressing small inputs.

A few attempts have been made to fix this, #2971 just removes the initialization but is problematic with Valgrind and might introduce another regression due to the consecutive compressions getting "false positives" from previous compressions' tags.

#3426 expands on #2971 and introduces memory regions that have been initialized at least once (thus not triggering Valgrind) and salts the hash to avoid collisions. However, it was rather complex.

**Intended solution**
Break down #3426 into multiple PRs and possibly remove some of the functionality introduced there.
This is a grandfather issue so we can tie the broken-down PRs together.

yoniko:
Part 1 of #3528 

- Adds memory type that is guaranteed to have been initialized at least once in the workspace's lifetime.
- Changes tag space in row hash to be based on init once memory.
- Unaligned buffers are now allocated after aligned buffers to allow aligned init once buffers to be allocated at the end of the workspace.

Update: now maintains the usage of unaligned buffers

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1653,9 +1653,9 @@ static size_t ZSTD_estimateCCtxSize_usingCCtxParams_internal(
     size_t const windowSize = (size_t) BOUNDED(1ULL, 1ULL << cParams->windowLog, pledgedSrcSize);
     size_t const blockSize = MIN(ZSTD_resolveMaxBlockSize(maxBlockSize), windowSize);
     size_t const maxNbSeq = ZSTD_maxNbSeq(blockSize, cParams->minMatch, useSequenceProducer);
-    size_t const tokenSpace = ZSTD_cwksp_alloc_size(WILDCOPY_OVERLENGTH + blockSize)
+    size_t const tokenSpace = ZSTD_cwksp_aligned_alloc_size(WILDCOPY_OVERLENGTH + blockSize)
                             + ZSTD_cwksp_aligned_alloc_size(maxNbSeq * sizeof(seqDef))
-                            + 3 * ZSTD_cwksp_alloc_size(maxNbSeq * sizeof(BYTE));
+                            + 3 * ZSTD_cwksp_aligned_alloc_size(maxNbSeq * sizeof(BYTE));
     size_t const entropySpace = ZSTD_cwksp_alloc_size(ENTROPY_WORKSPACE_SIZE);
     size_t const blockStateSpace = 2 * ZSTD_cwksp_alloc_size(sizeof(ZSTD_compressedBlockState_t));
     size_t const matchStateSize = ZSTD_sizeof_matchState(cParams, useRowMatchFinder, /* enableDedicatedDictSearch */ 0, /* forCCtx */ 1);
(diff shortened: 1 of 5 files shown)


yoniko:
Note: I explore reducing the scope of changes in this path by maintaining the buffer space.
Update: PR update to maintain buffer space.

lib/compress/zstd_cwksp.h
```diff
@@ -310,7 +331,7 @@ ZSTD_cwksp_internal_advance_phase(ZSTD_cwksp* ws, ZSTD_cwksp_alloc_phase_e phase
                 ws->tableEnd = objectEnd;  /* table area starts being empty */
                 if (ws->tableValidEnd < ws->tableEnd) {
                     ws->tableValidEnd = ws->tableEnd;
-        }   }   }
+                }   }   }
```
Cyan4973: nit: doesn't seem to be the proper bracket alignment

lib/compress/zstd_cwksp.h
```diff
@@ -367,6 +388,36 @@ MEM_STATIC BYTE* ZSTD_cwksp_reserve_buffer(ZSTD_cwksp* ws, size_t bytes)
...
+    size_t const alignedBytes = ZSTD_cwksp_align(bytes, ZSTD_CWKSP_ALIGNMENT_BYTES);
+    void* ptr = ZSTD_cwksp_reserve_internal(ws, alignedBytes, ZSTD_cwksp_alloc_aligned_init_once);
+    assert(((size_t)ptr & (ZSTD_CWKSP_ALIGNMENT_BYTES-1))== 0);
+    if(ptr && ptr < ws->initOnceStart) {
+        /* We assume the memory following the current allocation is either:
+         * 1. Not usable as initOnce memory (end of workspace)
+         * 2. Another initOnce buffer that has been allocated before (and so was previously memset)
+         * 3. An ASAN redzone, in which case we don't want to write on it
+         * For these reasons it should be fine to not explicitly zero every byte up to ws->initOnceStart.
+         * Note that we assume here tha MSAN and ASAN cannot run in the same time. */
```
Cyan4973: nit: we assume here **that**

lib/compress/zstd_compress.c
```diff
@@ -1932,6 +1932,17 @@ ZSTD_reset_matchState(ZSTD_matchState_t* ms,
         ZSTD_cwksp_clean_tables(ws);
     }
 
+    if (ZSTD_rowMatchFinderUsed(cParams->strategy, useRowMatchFinder)) {
+        /* Row match finder needs an additional table of hashes ("tags") */
+        size_t const tagTableSize = hSize * sizeof(U16);
```
Cyan4973: Investigate if we really need `U16` space for the tags table.

lib/compress/zstd_cwksp.h
yoniko: Fixed

lib/compress/zstd_cwksp.h
yoniko: Fixed

lib/compress/zstd_compress.c
yoniko: For now we need it

yoniko:
@Cyan4973 - see CR fixes

Cyan4973 (approved):

PR #3531, opened 2023-03-07
Clarify dstCapacity requirements

### LINKED ISSUE #3524: zstd 1.5.4 compression error less than or equal to 4 bytes src by the ZSTD_compress function
if compressing less than or equal to 4 bytes the function will return exceptions

version zstd 1.5.4 
Windows 10 Pro 21H2  19044.2604
CPU:  i7-6700K
Compiler: Visual Studio 2022 (v143)
here is a sample code
```c++
int main() {
   
	const size_t srcSize = 4;
	char src[srcSize] = {0x0 ,0x0 ,0x0 ,0x0 };
	

	size_t dstSize = srcSize + 12 + max(1, (srcSize + 12) / 1000);
	char* dst = new char[dstSize];
	
	size_t const countSize = ZSTD_compress((void*)dst, dstSize, (const void*)src, srcSize, 3);
	if (ZSTD_isError(countSize)) {
		assert(!ZSTD_isError(countSize));
	}
}
```

terrelln:
Clarify `dstCapacity` requirements for single-pass functions.

Fixes #3524.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -148,7 +148,8 @@ ZSTDLIB_API const char* ZSTD_versionString(void);
 ***************************************/
 /*! ZSTD_compress() :
  *  Compresses `src` content as a single zstd compressed frame into already allocated `dst`.
- *  Hint : compression runs faster if `dstCapacity` >=  `ZSTD_compressBound(srcSize)`.
+ *  NOTE: Providing `dstCapacity >= ZSTD_compressBound(srcSize)` guarantees that zstd will have
+ *        enough space to successfully compress the data.
  *  @return : compressed size written into `dst` (<= `dstCapacity),
  *            or an error code if it fails (which can be tested using ZSTD_isError()). */
 ZSTDLIB_API size_t ZSTD_compress( void* dst, size_t dstCapacity,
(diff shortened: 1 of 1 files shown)


lib/zstd.h
```diff
@@ -578,7 +579,8 @@ ZSTDLIB_API size_t ZSTD_CCtx_reset(ZSTD_CCtx* cctx, ZSTD_ResetDirective reset);
  *  Should cctx hold data from a previously unfinished frame, everything about it is forgotten.
  *  - Compression parameters are pushed into CCtx before starting compression, using ZSTD_CCtx_set*()
  *  - The function is always blocking, returns when compression is completed.
- *  Hint : compression runs faster if `dstCapacity` >=  `ZSTD_compressBound(srcSize)`.
+ *  NOTE: Providing `dstCapacity >= ZSTD_compressBound(srcSize)` guarantees that zstd will have
+ *        enough space to successfully compress the data, though it is possible it fails for other reasons.
```
Cyan4973, line 583: > though it is possible it fails for other reasons

What other reasons do you have in mind ?
Memory allocation failures ?
Providing `dstCapacity >= ZSTD_compressBound(srcSize)` should be enough to avoid #3524.

terrelln: I mean `ZSTD_compress2()` can let you specify a lot of options. E.g. compression could fail to allocate a thread. We could run out of memory.

I only mentioned this in `ZSTD_compress2()`, and skipped it in `ZSTD_compress()`, but I just mean that you still have to check the error code.

Cyan4973 (approved):

PR #3533, opened 2023-03-07
Introduce salt into row hash (#3528 part 2)

### LINKED ISSUE #3528: Row hash tag space initialization speed regression

yoniko:
Part 2 of #3528 

Adds hash salt that helps to avoid regressions where consecutive compressions use the same tag space with similar data (running `zstd -b5e7 enwik8 -B128K` reproduces this regression).

Note: based on #3528 and includes it, see only diffs introduced here by selecting only the changes from last commit, or by viewing https://github.com/yoniko/zstd/pull/13

### DIFF
--- a/.circleci/images/primary/Dockerfile
+++ b/.circleci/images/primary/Dockerfile
@@ -1,4 +1,4 @@
-FROM circleci/buildpack-deps:focal
+FROM circleci/buildpack-deps@sha256:f6f10c11b7b8ccfd4f4a5b830c3256803604ce61292b60cb22e26b12f62b0e8c
 
 RUN sudo dpkg --add-architecture i386
 RUN sudo apt-get -y -qq update
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -186,6 +186,16 @@ jobs:
     - name: MSan + Regression Test
       run: make -j msanregressiontest
 
+  clang-msan-fuzz-unoptimized:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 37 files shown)


yoniko:
PR performance seems to be on PR with previous versions' performance for the samples I've tested.
See previous results [here](https://docs.google.com/spreadsheets/d/14HC1oDjOr5-dKydhBvoeU7oXxcpVtQW5TdXIQVrhp_k/edit?usp=sharing).

yoniko:
@Cyan4973 - I've added some more entropy to the salt, let me know what you think

yoniko:
@Cyan4973 - is there any other change you'll like to see here?
If not, I'll merge.

Cyan4973 (approved):

PR #3543, opened 2023-03-10
Reduce RowHash's tag space size by x2

yoniko:
Allocate half the memory for tag space, which means that we get one less slot for an actual tag (needs to be used for next position index).
The results is a slight loss in compression ratio (up to 0.2%) and some regressions/improvements to speed depending on level and sample. In turn, we get to save 16% of the hash table's space (5 bytes per entry instead of 6 bytes per entry).

Update:
CR comments fixed.
Benchmark in [spreadsheet](https://docs.google.com/spreadsheets/d/1dvgVlxLdyIkDR57gcXnfcKghIm7l53dgAMnb3QA-byU/edit?usp=sharing).

Update 2:
Another interesting benchmark is `fullbench -b41 -l9 enwik8.1k` which repeatedly uses `ZSTD_compressStream` to recompress 1024 bytes out of enwik8. Since #3528 hasn't been deployed yet, we can observe that this PR makes the benchmark twice as fast for small data sizes as the majority of time is spent in the tag space intiailization.


lib/compress/zstd_lazy.c
```diff
@@ -804,9 +803,10 @@ U16 ZSTD_rotateRight_U16(U16 const value, U32 count) {
  * value to reflect the update. Essentially cycles backwards from [0, {entries per row})
```
terrelln: You should update this comment

lib/compress/zstd_lazy.c
```diff
@@ -903,7 +903,7 @@ FORCE_INLINE_TEMPLATE void ZSTD_row_update_internalImpl(ZSTD_matchState_t* ms,
         U32 const pos = ZSTD_row_nextIndex(tagRow, rowMask);
```
terrelln, line 902: You should update the comments on lines 901-902

lib/compress/zstd_lazy.c
```diff
@@ -888,7 +888,7 @@ FORCE_INLINE_TEMPLATE void ZSTD_row_update_internalImpl(ZSTD_matchState_t* ms,
                                                         U32 const rowMask, U32 const useCache)
 {
     U32* const hashTable = ms->hashTable;
-    U16* const tagTable = ms->tagTable;
+    U8* const tagTable = ms->tagTable;
```
terrelln: You should use `BYTE` here and elsewhere

lib/compress/zstd_lazy.c
```diff
@@ -903,7 +903,7 @@ FORCE_INLINE_TEMPLATE void ZSTD_row_update_internalImpl(ZSTD_matchState_t* ms,
         U32 const pos = ZSTD_row_nextIndex(tagRow, rowMask);
 
         assert(hash == ZSTD_hashPtr(base + updateStartIdx, hashLog + ZSTD_ROW_HASH_TAG_BITS, mls));
-        ((BYTE*)tagRow)[pos + ZSTD_ROW_HASH_TAG_OFFSET] = hash & ZSTD_ROW_HASH_TAG_MASK;
+        ((BYTE*)tagRow)[pos] = hash & ZSTD_ROW_HASH_TAG_MASK;
```
terrelln: Can get rid of the `BYTE*` cast now.

terrelln (commented):
Generally looks good!

I did a quick benchmark and see just about neutral performance, +- a few percent based on the compiler. Also a small ratio loss, but it seems reasonable. I'll let you do the full benchmarks, just wanted to verify that the speed looked good on my machine too.

yoniko:
@terrelln - did you benchmark on a mac or on an x86 server?

lib/compress/zstd_lazy.c
yoniko: I thought I did, thank you for noticing!

yoniko:
Fixed CR comments and added a spreadsheet for benchmarks.
Compression ratio changes are less than 0.01% on average, there's a regression of up to 5% for levels 5-7 on Skylake but a performance increase for higher levels.

I've tried reducing this regression, but don't have a good solution at the moment and unless it's critical I'll not continue.

terrelln:
> @terrelln - did you benchmark on a mac or on an x86 server?

x86

terrelln (approved):
Awesome!

shubhamchandak94:
Hi, can this be reflected in the comment at https://github.com/facebook/zstd/blame/25822342be59d831bad65426ae51f5cc22157b09/lib/compress/zstd_lazy.c#L1116-L1118

Thanks!

Issue #2189, opened 2020-06-02
Speedup scan speed for '--patch-from' via rolling hashes


RubenKelevra:
IPFS has the ability to dedup blocks between different types of files. This functionality is based on a rolling hash algorithm.

You can either select rabin or buzzhash for this task (in IPFS). Rabin is kind of slow, but buzzhash is quite fast.

The rolling hash would allow to 'prescan' both files, get some cut marks and run some fast cryptographic hash algorithm over the chunks, like blake2b.

I think both operations are much cheaper than pattern matching. This way you can skip all pattern matching attempts which are on both sides (A and B) inside the known equal blocks.

The first layer of patching would just generate a lengths+offset+move triple, which can copy the blocks from the original file into a sparse file as first patching operation.

The pattern matching rules could be used on top of that, completing the gaps of the output file.

_Originally posted by @RubenKelevra in https://github.com/facebook/zstd/issues/2063#issuecomment-616705733_

terrelln:
`--patch-from` does already using rolling hashes!

However, we also use our regular match finders, so end up inserting the whole file anyway.

One idea would be to only insert the "end" of the dictionary into our normal match finders. Basically only the portion that we expect would be reasonably indexed by our normal hash tables (e.g. like 4 * (1 << max(hashLog, chainLog))). Then let our LDM mode handle the rest.

PR #3545, opened 2023-03-10
patch-from speed optimization

daniellerozenblit:
# TLDR
This PR is a response to issue #2189, which requests a speedier version of `--patch-from` compression. 

This PR offers a solution by only loading the suffix of the dictionary into our normal match finders, rather than the entire dictionary. We continue to load the entire "dictionary" into our LDM match finders.

 We only load the portion into our normal match finders that can reasonably be indexed by our hash tables, 8 * (1 << max(hashLog, chainLog)). Note that the 8 here is an arbitrary multiplier that shows good results (I also experimented with 4 as a multiplier, which was slightly faster but seemed to perform worse on the zstd regression test). This feature is disabled for strategies `>= ZSTD_btultra`.

This optimization offers great improvements on compression speed, with very minimal  increase in patch size.

Credit and thanks to @terrelln for the optimization idea.

# Benchmarking
I benchmarked on an Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz with core isolation and turbo disabled. 

I benchmarked `--patch-from` compression on the linux kernel tree tarball v6.0 -> v6.2. For speed measurements, I ran each scenario five times interleaved and chose the highest result.

# Compression Speed
There are significant improvements in compression / patch-creation speed across a range of compression levels. These speed improvements are especially present at higher compression levels (i.e. ~617.6% increase at compression level 15).

Screenshot 2023-03-10 at 1 04 52 PM

# Patch Size
There is some increase in patch size across compression levels. This increase is more prevalent at higher compression levels, but is still fairly minimal (i.e. ~0.47% increase at compression level 15).

Screenshot 2023-03-10 at 9 17 47 AM


lib/compress/zstd_compress.c
```diff
@@ -4678,22 +4679,30 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
...
 
-    if (loadLdmDict) {
+    if (loadLdmDict) { /* Load the entire dict into LDM matchfinders. */
         ZSTD_window_update(&ls->window, src, srcSize, /* forceNonContiguous */ 0);
         ls->loadedDictEnd = params->forceWindow ? 0 : (U32)(iend - ls->window.base);
+        ZSTD_ldm_fillHashTable(ls, ip, iend, &params->ldmParams);
+    }
+
+    /* If the dict is larger than we can reasonably index in our tables, only load the suffix. */
+    {   U32 maxDictSize = 8U * (1U << MIN(MAX(params->cParams.hashLog, params->cParams.chainLog), 29));
```
Cyan4973: This is a bit too much. 
`8 * (1<<29)` is equivalent to `1<<32`, so this will not end well on 32-bit values.
`1<<31` (2 GB) shall be the max, it is (currently) our limit window size anyway.
`8 * MIN(MAX(),28)` should do the trick.

Cyan4973:
Great results @daniellerozenblit !

Is it practical to test this patch with compression level `--ultra -22` ?

The trade-off achieved here is very good, and most users will be glad to trade a little bit of compression ratio for a lot of speed. Furthermore, if they really want more compression, they could still increase the compression level, and get better ratio, for even more speed (could be shown effectively with speed/ratio graph).

But users of level `22` are typically ready to sacrifice speed to get the best possible compression ratio. And a 0.5% ratio regression would be a lot for them.
However, I suspect that your PR actually doesn't impact this case. Yet, it would be better to measure it.

lib/compress/zstd_compress.c
daniellerozenblit: Oops, silly mistake! Thanks for catching this.

Cyan4973:
1) Could we get some benchmark results from the latest changes for levels 1, 19 and 22 please ?
(Note : not necessarily in graphs, just giving numbers would be enough).

2) Could you test multithreading too (even if it's just `-T2`) ?
The point is to show that these optimizations remain perfectly valid with multi-threading enabled.

At this stage, we just want to assert the results, and show that they are globally positive.

The code looks good to me, changes are sufficiently simple to be properly reviewed.

terrelln:
> Could you test multithreading too (even if it's just -T2) ?

And also explicit single threading if you are using the CLI with `--single-thread`

terrelln:
The code LGTM once we have the benchmark results for the mentioned scenarioes!

daniellerozenblit:
> > Could you test multithreading too (even if it's just -T2) ?
> 
> And also explicit single threading if you are using the CLI with `--single-thread`

I wasn't benchmarking with any explicit `--single-thread` or `T#` flags. Should I benchmark the single-thread scenario as well?

daniellerozenblit:
## Additional Benchmarking
The results described earlier seem consistent across multiple threading scenarios: `--single-thread`, `-T2`, and the default scenario. However, for `--single-thread`, there is a higher loss of compression ratio for the optimization.
### Single Thread
For single threaded compression, speed improvements are consistent with previous results. However, there is some increased loss in compression ratio across compression levels (~4%).
Screenshot 2023-03-14 at 3 56 50 PM
### T2
Results for `-T2` are very consistent with the default scenario. There even appears to be some increase in the speed improvements.
Screenshot 2023-03-14 at 3 58 29 PM

### Levels 1, 19, 22
There are no significant changes in neither compression ratio nor speed for levels 1, 19, 22 in the default scenario.
Screenshot 2023-03-14 at 3 52 44 PM

Cyan4973 (approved):

terrelln:
Awesome!

PR #3551, opened 2023-03-14
Provide an interface for fuzzing sequence producer plugins

embg:
Fuzz-testing for the external sequence producer API was introduced in PR #3437. However, the setup in #3437 only allows fuzzers to exercise the implementation of the API on the libzstd side (the code in the core zstd library which interacts with a plugin). This PR allows us to fuzz the plugin code itself, which is a requirement for us to use plugins (such as hardware accelerators) in production.

Here is an example showing how this new interface allows us to link plugin code into the fuzzer build:
```
$ make corpora
$ make -C seq_prod_fuzz_example/
$ python3 ./fuzz.py build all --custom-seq-prod=seq_prod_fuzz_example/example_seq_prod.o --enable-fuzzer --enable-asan --enable-ubsan --cc clang --cxx clang++
$ python3 ./fuzz.py libfuzzer simple_round_trip
$ python3 ./fuzz.py libfuzzer stream_round_trip
$ python3 ./fuzz.py libfuzzer dictionary_round_trip
$ # etc...
```

### DIFF
--- a/tests/fuzz/fuzz_third_party_seq_prod.h
+++ b/tests/fuzz/fuzz_third_party_seq_prod.h
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) Yann Collet, Meta Platforms, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under both the BSD-style license (found in the
+ * LICENSE file in the root directory of this source tree) and the GPLv2 (found
+ * in the COPYING file in the root directory of this source tree).
+ * You may select, at your option, one of the above-listed licenses.
+ */
+
+#ifndef EXAMPLE_SEQ_PROD_H
+#define EXAMPLE_SEQ_PROD_H
+
+#define ZSTD_STATIC_LINKING_ONLY
+#include "zstd.h"
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* *** INTERFACE FOR FUZZING THIRD-PARTY SEQUENCE PRODUCER PLUGINS ***
+ * Fuzz-testing for the external sequence producer API was introduced in PR #3437.
+ * However, the setup in #3437 only allows fuzzers to exercise the implementation of the
+ * API itself (the code in the core zstd library which interacts with your plugin).
+ *
+ * This header defines an interface for plugin authors to link their code into the fuzzer
+ * build. Plugin authors can provide an object file implementing the symbols below,
+ * and those symbols will replace the default ones provided by #3437.
+ *
+ * To fuzz your plugin, follow these steps:
+ *   - Build your object file with a recent version of clang. Building with gcc is not supported.
+ *   - Build your object file using appropriate flags for fuzzing. For example:
+ *     `-g -fno-omit-frame-pointer -fsanitize=undefined,address,fuzzer`
+ *   - Build the fuzzer binaries with options corresponding to the flags you chose. Use --custom-seq-prod= to pass in your object file:
+ *     `./fuzz.py build all --enable-fuzzer --enable-asan --enable-ubsan --cc clang --cxx clang++ --custom-seq-prod=your_object.o`
+ *
+ * An example implementation of this header is provided at tests/fuzz/seq_prod_fuzz_example/.
+ * Use these commands to fuzz with the example code:
+ *   $ make corpora
+ *   $ make -C seq_prod_fuzz_example/
+ *   $ python3 ./fuzz.py build all --enable-fuzzer --enable-asan --enable-ubsan --cc clang --cxx clang++ --custom-seq-prod=seq_prod_fuzz_example/example_seq_prod.o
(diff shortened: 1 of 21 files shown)


tests/fuzz/seq_prod_fuzz_example/Makefile
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) Yann Collet, Meta Platforms, Inc.
+ * All rights reserved.
```
GarenJian-Intel: Typo.
Use # to comment in Makefile instead of /* */

embg: Thanks :) copy-pasted without thinking

tests/fuzz/seq_prod_fuzz_example/Makefile
```diff
@@ -0,0 +1,18 @@
+/*
```
Cyan4973: Is this Makefile run in one of the CI tests ?

tests/fuzz/fuzz_third_party_seq_prod.h
```diff
@@ -0,0 +1,115 @@
...
+  } while (0)
+#else
+#define FUZZ_SEQ_PROD_SETUP
+#endif
+
+#ifdef FUZZ_THIRD_PARTY_SEQ_PROD
+#define FUZZ_SEQ_PROD_TEARDOWN                                                 \
+  do {                                                                         \
+    FUZZ_ASSERT(FUZZ_freeSeqProdState(FUZZ_seqProdState) == 0);                \
+    FUZZ_ASSERT(FUZZ_seqProdTearDown() == 0);                                  \
```
GarenJian-Intel, line 106: Could we call FUZZ_seqProdTearDown only at the end of the whole fuzz test instead of calling it at the end of every round_trip test.
From now on, the latency of icp_sal_userStop is too big(seconds) though we will push driver to optimize it. So call it after every round_trip test will greatly slow down the running of the whole fuzz test.

GarenJian-Intel: Call setup at the beginning of the fuzz test and call teardown at the end of the fuzz test.

embg: The fuzzing loop runs forever, until either the process crashes or the user kills the process. There currently isn't any way to clean up resources at the end of the process. 

Can we just leave `FUZZ_seqProdTearDown()` empty?

tests/fuzz/seq_prod_fuzz_example/Makefile
embg: I don't think the benefit of having this in CI is worth the cost of adding another job. Happy to discuss offline.

tests/fuzz/fuzz_third_party_seq_prod.h
terrelln: > There currently isn't any way to clean up resources at the end of the process.

I mean the only thing you could do is a signal handler. But you could set that up with the current hooks, by registering it on the first call to `FUZZ_seqProdSetup()`.

But yeah, the simpler thing would just to leak any resources, since the fuzzer is called in a loop that runs for millions of iterations, and generally finishes when it crashes.

tests/fuzz/fuzz_third_party_seq_prod.h
```diff
@@ -0,0 +1,115 @@
...
+/* These macros are internal helpers. You do not need to worry about them. */
+#ifdef FUZZ_THIRD_PARTY_SEQ_PROD
+#define FUZZ_SEQ_PROD_SETUP                                                    \
+  do {                                                                         \
+    FUZZ_ASSERT(FUZZ_seqProdSetup() == 0);                                     \
+    FUZZ_seqProdState = FUZZ_createSeqProdState();                             \
+    FUZZ_ASSERT(FUZZ_seqProdState != NULL);                                    \
+  } while (0)
+#else
+#define FUZZ_SEQ_PROD_SETUP
```
terrelln: Nit: Make these function-like, to indicate that they should be treated like a function call 
```suggestion
#define FUZZ_SEQ_PROD_SETUP() 
```

terrelln (approved):
Looks good. Can you also add the documentation about how to build & run the fuzzer in `seq_prod_fuzz_example/README.md`?

GarenJian-Intel:
Hi @embg, now only simple_round_trip is enabled, will other fuzzers be enabled in this PR or other new PRs? If other fuzzers are ready, could you @ me so that I can try them ASAP?

embg:
@GarenJian-Intel Added support for the following fuzzers:
```
block_round_trip
decompress_dstSize_tooSmall
dictionary_decompress
dictionary_loader
dictionary_round_trip
dictionary_stream_round_trip
raw_dictionary_round_trip
sequence_compression_api
simple_compress
simple_round_trip
stream_round_trip
```

Also, responded to all nits from @terrelln. Merging after CI passes.

Issue #2835, opened 2021-10-28
How do I get --patch-from to work when using zstd as a library?


wspl:
I found out that the implementation of the `--patch-from` is placed in the "programs" folder and it is not distributed with the static or dynamic libraries. This means that if I don't use the cli program, I wouldn't be able to use the `--patch-from` function directly. 

But I don't know much about how zstd works, how should I configure the compress or decompress parameters to make the `--patch-from` works in my own program?

Cyan4973:
The core library API equivalent to `--patch-from` are `ZSTD_compress_usingDict()` and `ZSTD_decompress_usingDict()`.
`--patch-from` additionally modifies some advanced parameters, like `windowLog` and enables the `ldm` mode.
In short, whatever `--patch-from` does can be done with only the library, but `--patch-from` packages in a single command a number of advanced parameters that must be done manually when using the library directly.

wspl:
> The core library API equivalent to `--patch-from` are `ZSTD_compress_usingDict()` and `ZSTD_decompress_usingDict()`. `--patch-from` additionally modifies some advanced parameters, like `windowLog` and enables the `ldm` mode. In short, whatever `--patch-from` does can be done with only the library, but `--patch-from` packages in a single command a number of advanced parameters that must be done manually when using the library directly.

@Cyan4973  I use a zstd Rust Wrapper (https://github.com/gyscos/zstd-rs) that uses `ZSTD_compress_usingDict` for file compression internally, and when I compressed the files through this library, I found that the resulting diff file was much larger than the one produced by `--patch-from` of cli.

Also, compressing a small file with `zstd -D` produces a much larger file than `--patch-from`.

Cyan4973:
As mentioned, `--patch-from` is a combination of dictionary mode **+** a number of well selected parameters.
If the rust wrapper gives you access to the full stable API, then it should be possible to reproduce the same thing.
But that requires understanding advanced parameters.

If what you are asking for is a dedicated API to replicate what is achieved with command `--patch-from` at library level, without the need to understand parameters, then be aware that it will take quite some time before it gets accessible through the rust wrapper.

wspl:
@Cyan4973 My guess is the same. So I found out by reading the code that the `windowLog` parameter might need to be set, so I set a value as calculated in the source code, and it didn't work. This is what I found by reading the pull request for `--patch-from` (https://github.com/facebook/zstd/pull/1940), I don't understand if I need to set some other parameters? Can you improve some hint for this piece?

[...]

PR #3553, opened 2023-03-15
added documentation for LDM + dictionary compatibility

Cyan4973:
As mentioned by @reyqn in https://github.com/facebook/zstd/issues/2835#issuecomment-1456228611 , the Long Distance Mode works best with dictionaries loaded with `ZSTD_CCtx_refPrefix()`.
LDM is effectively incompatible with `ZSTD_CDict`, and by extension with `ZSTD_CCtx_loadDictionary()`, so results are disappointing when trying to combine them.

Added documentation to nudge users towards `ZSTD_CCtx_refPrefix()` when they want to use a dictionary as large "reference image" which requires LDM for proper indexing.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -1020,9 +1020,11 @@ ZSTDLIB_API unsigned ZSTD_getDictID_fromFrame(const void* src, size_t srcSize);
  * Advanced dictionary and prefix API (Requires v1.4.0+)
  *
  * This API allows dictionaries to be used with ZSTD_compress2(),
- * ZSTD_compressStream2(), and ZSTD_decompressDCtx(). Dictionaries are sticky, and
- * only reset with the context is reset with ZSTD_reset_parameters or
- * ZSTD_reset_session_and_parameters. Prefixes are single-use.
+ * ZSTD_compressStream2(), and ZSTD_decompressDCtx().
+ * Dictionaries are sticky, they remain valid when same context is re-used,
+ * they only reset when the context is reset
+ * with ZSTD_reset_parameters or ZSTD_reset_session_and_parameters.
+ * In contrast, Prefixes are single-use.
  ******************************************************************************/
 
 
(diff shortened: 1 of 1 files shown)


terrelln (approved):

ghost:
May I ask, why `prefix` is only used once?
If there is an API that allows `cctx` to use a `prefix` infinitely, what negative effects does it have? 

From `ZSTD_CCtx_refPrefix()` doc:
>  *  Reference a prefix (single-usage dictionary) for next compressed frame.
> *  A prefix is **only used once**. Tables are discarded at end of frame (ZSTD_e_end).

Cyan4973:
A `prefix` must be loaded into the match finder tables.
The match finder tables are then mutated during the rest of the compression process,
so the initial "state" of the match finder tables, after loading the `prefix`, is effectively lost.

Therefore, using a `prefix` a second time requires loading its content into the match finder tables again. This is a non-trivial cost.

This situation is in contrast with `CDict`, which are created once, and are then immutable,
allowing their usage by any number of `CCtx` afterwards,
without any initialization cost.

Employing a `prefix` rather than a full-feature `CDict` makes sense when it's only going to be used once.

ghost:
Thanks for your explaination.

It seems this sentence is a bit misleading, may be "regarded as a prefix" rather than "called a prefix".
https://github.com/facebook/zstd/blob/3e0550ee5279735693b01464ede7cb1ec22fe6b7/lib/zstd.h#L904

I have another question, if a prefix is loaded as `ZSTD_dct_fullDict`, how is it different from a dictionary? https://github.com/facebook/zstd/blob/3e0550ee5279735693b01464ede7cb1ec22fe6b7/lib/zstd.h#L1895-L1898

Cyan4973:
> I have another question, if a prefix is loaded as `ZSTD_dct_fullDict`, how is it different from a dictionary?

It will still be loaded directly into the match search tables, and therefore the initial state will be lost during the compression process. So it's only suitable when used once.

The point of `ZSTD_dct_fullDict` is to explicitly tell the compressor to expect *only* a well-formed trained dictionary, with a conformant header. If not, it will error out.

The difference with a default `auto` mode is that, in `auto` mode, if the dictionary doesn't contain a well-formed header, it default to `raw` mode, where the entire input is considered as "raw content", and no statistics are present (and no dictionary ID is present). This could be a missed opportunity to detect an error scenario earlier.

Issue #3507, opened 2023-02-14
Undefined behavior in `ZSTD_decompress` for certain `dest` values when `dstCapacity` is zero


squeek502:
**Describe the bug**
When `dstCapacity` is zero, in theory the `dest` pointer should be irrelevant--any pointer value should be allowed (or, alternatively, the docs should specify what is and isn't allowed). However, if `ZSTD_decompress` is called with `dstCapacity=0` and `dest` having a value of e.g. `0xffffffffffffffff`, it will hit UB and may trip an assertion. I'm unsure exactly what range of pointer values will trigger this, but the maximum pointer value definitely does.

**To Reproduce**
1. Inputs that can trigger the UB: [zstd-zero-dest-capacity.zip](https://github.com/facebook/zstd/files/10750794/zstd-zero-dest-capacity.zip) (found when fuzzing a Zig implementation of a zstd decompressor). I'm not sure the input matters too much, the uncompressed size just needs to be zero; some inputs are the same as those in #3506
2. Compile zstd with clang via `make lib CFLAGS="-fsanitize=undefined -fPIC"`
3. Modify `examples/simple_decompression.c` with the following patch:

```diff
-    void* const rBuff = malloc_orDie((size_t)rSize);
+    CHECK(rSize == 0, "Uncompressed size must be zero");
+    void* const rBuff = (void*)(-1); // 0xffffffffffffffff
```

4. Compile examples with `make LDFLAGS="-fsanitize=undefined"`
5. Run one of the inputs through `simple_decompression`, e.g. `./simple_decompression 'id:000000,sig:04,src:000193,time:1035023,execs:294994,op:havoc,rep:4'`

Example of the output with `DEBUGLEVEL=10`:

```
$ ./simple_decompression 'id:000000,sig:04,src:000193,time:1035023,execs:294994,op:havoc,rep:4'
.//decompress/zstd_decompress.c: ZSTD_getFrameHeader_advanced: minInputSize = 5, srcSize = 32 
.//decompress/zstd_decompress.c: ZSTD_decompressMultiFrame 
.//decompress/zstd_decompress.c: reading magic number FD2FB528 (expecting FD2FB528) 
.//decompress/zstd_decompress.c: ZSTD_decompressFrame (srcSize:32) 
.//decompress/zstd_decompress.c: ZSTD_getFrameHeader_advanced: minInputSize = 5, srcSize = 9 
[...]

Issue #3506, opened 2023-02-14
Undefined behavior in `ZSTD_decompress` for certain inputs


squeek502:
**Describe the bug**
When compiled with clang and `-fsanitize=undefined`, `ZSTD_decompress` will hit undefined behavior with certain inputs:

```
/decompress/zstd_decompress_block.c:1696:45: runtime error: applying non-zero offset 79823856 to null pointer
```

**To Reproduce**
1. Inputs that trigger the UB: [zstd-non-zero-offset-inputs.zip](https://github.com/facebook/zstd/files/10728643/zstd-non-zero-offset-inputs.zip) (found when fuzzing a Zig implementation of a zstd decompressor)
2. Compile zstd with clang via `make lib CFLAGS="-fsanitize=undefined -fPIC"`
3. Compile examples with `make LDFLAGS="-fsanitize=undefined"`
4. Run one of the inputs through `simple_decompression`, e.g. `./simple_decompression 'id:000032,sig:06,src:000673,time:3423388,execs:797038,op:havoc,rep:2'`

Example of the output with `DEBUGLEVEL=10`
```
.//decompress/zstd_decompress.c: ZSTD_getFrameHeader_advanced: minInputSize = 5, srcSize = 39 
.//decompress/zstd_decompress.c: ZSTD_decompressMultiFrame 
.//decompress/zstd_decompress.c: reading magic number FD2FB528 (expecting FD2FB528) 
.//decompress/zstd_decompress.c: ZSTD_decompressFrame (srcSize:39) 
.//decompress/zstd_decompress.c: ZSTD_getFrameHeader_advanced: minInputSize = 5, srcSize = 9 
.//decompress/zstd_decompress_block.c: ZSTD_decompressBlock_internal (size : 20) 
.//decompress/zstd_decompress_block.c: ZSTD_decodeLiteralsBlock 
.//decompress/zstd_decompress_block.c: ZSTD_decodeLiteralsBlock : cSize=1, nbLiterals=0 
.//decompress/zstd_decompress_block.c: ZSTD_decodeSeqHeaders 
.//decompress/zstd_decompress_block.c: ZSTD_getLongOffsetsShare: (tableLog=5) 
.//decompress/zstd_decompress_block.c: ZSTD_decompressSequencesLong 
.//decompress/zstd_decompress_block.c: ZSTD_initFseState : val=5 using 5 bits 
.//decompress/zstd_decompress_block.c: ZSTD_initFseState : val=24 using 5 bits 
[...]

PR #3555, opened 2023-03-15
Check that `dest` is valid for decompression

daniellerozenblit:
This PR is a response to issues #3506 and #3507.

It adds some checks to ensure that the `dest` buffer passed to `ZSTD_decompressBlock` are valid. In particular, these check that if the `dstCapacity == 0` then the number of sequences is 0 and that the pointer passed to `dest` is sufficiently small that any permitted `match + literal` length can be added to it without overflowing. 

This PR additionally adds some fuzzing to target these issues. Specifically, we pass a random `dest` value when fuzzing decompression on a buffer of size 0.


contrib/linux-kernel/test/static_test
yoniko, line 1: Was this file included on purpose? (same question for `contrib/linux-kernel/test/test`)

daniellerozenblit: oops, nope!

lib/decompress/zstd_decompress_block.c
```diff
@@ -2125,6 +2125,9 @@ ZSTD_decompressBlock_internal(ZSTD_DCtx* dctx,
         srcSize -= seqHSize;
 
         RETURN_ERROR_IF(dst == NULL && nbSeq > 0, dstSize_tooSmall, "NULL not handled");
+        RETURN_ERROR_IF(dstCapacity == 0 && nbSeq > 0, dstSize_tooSmall, "dstCapacity == 0, but nbSeq is non-zero");
```
terrelln: nit: We can combine this check with the line `2127`. `dst` may only be `NULL` if `dstCapacity == 0`.

lib/decompress/zstd_decompress_block.c
```diff
@@ -2125,6 +2125,9 @@ ZSTD_decompressBlock_internal(ZSTD_DCtx* dctx,
...
+        RETURN_ERROR_IF(MEM_64bits() && (uptrval)dst + dstCapacity > (uptrval)-1 - (1 << 20), dstSize_tooSmall,
```
terrelln: @Cyan4973 what do you think of requiring `uintptr_t` here? We already do the same in LZ4.

terrelln (commented):
Generally looks good, just want to see Yann's opinion on the `uintptr_t`.

lib/decompress/zstd_decompress_block.c
Cyan4973: I'm generally not a fan of `uintptr_t`, because it's an optional type, 
and we learned the hard way that this type breaks on some platforms.

What about the following alternative test ? 

```
#define MAX_ADDRESS (const char*)(size_t)(-1)
#define SAFE_DISTANCE  (1 << 20)

if (MAX_ADDRESS - (const char*)dst < SAFE_DISTANCE) { ... } 
```

lib/decompress/zstd_decompress_block.c
daniellerozenblit: Would it be possible for `dst == NULL` and `dstCapacity == 0` but `nbSeq > 0`? And would this be caught elsewhere?

lib/decompress/zstd_decompress_block.c
```diff
@@ -2124,7 +2124,9 @@ ZSTD_decompressBlock_internal(ZSTD_DCtx* dctx,
         ip += seqHSize;
         srcSize -= seqHSize;
 
-        RETURN_ERROR_IF(dst == NULL && nbSeq > 0, dstSize_tooSmall, "NULL not handled");
+        RETURN_ERROR_IF((dst == NULL || dstCapacity == 0) && nbSeq > 0, dstSize_tooSmall, "NULL not handled");
+        RETURN_ERROR_IF(MEM_64bits() && (size_t)(-1) - (size_t)dst < (size_t)(1 << 20), dstSize_tooSmall,
```
Cyan4973: I'll take it,

but note that this formulation essentially replaces `uintptr_t` by `size_t`,
which implies that `sizeof(size_t) == sizeof(void*)`.
This is often true, for "common" platforms, but not always.
See [this comment](https://github.com/lz4/lz4/pull/578#issuecomment-424240197) for an example.

In this specific example, where `sizeof(void*)==16` while `sizeof(size_t)==4`,
the proposed test will trigger for any address close enough to a multiple of 4 GB.

To be fair, this is a hard issue to fix "properly" (i.e. without UB), so that's why I accept this fix proposal.
Evaluating the difference between 2 addresses which are not part of the same memory segment is UB to begin with (even if it works fine in a single-plane memory addressing architecture). That makes it hard to check the distance of an address from the end of its address space.

For the same reason, a dangling pointer that randomly points anywhere in memory is also UB, even if there is a second variable that states `size == 0`. So we are trying to fix at runtime a user-side UB. No wonder it's difficult to fix without triggering UB.

daniellerozenblit: Yeah I'm not sure of the fully correct way to solve this.

As you said, checking the distance between these two addresses is UB regardless. 

I guess the major concern with my solution is that on some unusual platforms where `sizeof(size_t) == 8` but `sizeof(void*) != 8` we may produce a lot of false positives. Would it be better to not check at all?

Cyan4973: > I guess the major concern with my solution is that on some unusual platforms where `sizeof(size_t) == 8` but `sizeof(void*) != 8` we may produce a lot of false positives. Would it be better to not check at all?

That's indeed the question.
It seems that your check `dstCapacity==0` should catch the issues described in #3507, and possibly #3506.
In which case, maybe that's enough ?

daniellerozenblit: Correct, the first check catches all of the cases in https://github.com/facebook/zstd/issues/3507 and https://github.com/facebook/zstd/issues/3506.

Cyan4973: So maybe the first check is enough to fix the situation.

The second test is fine as long as `sizeof(size_t) == sizeof(void*)`, 
maybe that condition could be checked before triggering the second test.

Cyan4973 (approved):

Cyan4973:
I'll merge it now, so that there is enough time testing before the release

PR #3556, opened 2023-03-16
Fix patch-from speed optimization

daniellerozenblit:
This PR is a followup to a previous patch-from speed optimization (#3545).

It ensures that the window size takes into account the full, un-truncated dictionary, in the normal matchfinders.


terrelln:
The code looks good!

How did you test that it doesn't regression compression speed? E.g. that this fix is still achieving the speedup we were aiming for.

daniellerozenblit:
> The code looks good!
> 
> How did you test that it doesn't regression compression speed? E.g. that this fix is still achieving the speedup we were aiming for.

I ran some benchmarks for the lower compression levels (levels 5, 7, 9, 12, 15) and speeds were comparable to my previous benchmarks. There was also no change in the regression tests.

terrelln (approved):

PR #3557, opened 2023-03-17
mmap for windows

daniellerozenblit:
This PR is a follow-up to #3486. It adds memory-mapped dictionaries for Windows.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -687,13 +687,13 @@ static void FIO_getDictFileStat(const char* fileName, stat_t* dictFileStat) {
     }
 }
 
-/*  FIO_initDictMalloc() :
+/*  FIO_setDictMalloc() :
  *  allocates a buffer, pointed by `dict->dictBuffer`,
  *  loads `filename` content into it, up to DICTSIZE_MAX bytes.
  * @return : loaded size
  *  if fileName==NULL, returns 0 and a NULL pointer
  */
-static size_t FIO_initDictMalloc(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
+static size_t FIO_setDictBufferMalloc(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
 {
     FILE* fileHandle;
     U64 fileSize;
@@ -740,7 +740,7 @@ static void FIO_munmap(FIO_Dict_t* dict)
     dict->dictBuffer = NULL;
     dict->dictBufferSize = 0;
 }
-static size_t FIO_initDictMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
+static size_t FIO_setDictBufferMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
 {
     int fileHandle;
     U64 fileSize;
@@ -769,6 +769,7 @@ static size_t FIO_initDictMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs
     }
 
     *bufferPtr = mmap(NULL, (size_t)fileSize, PROT_READ, MAP_PRIVATE, fileHandle, 0);
+    if (*bufferPtr==NULL) EXM_THROW(34, "%s", strerror(errno));
 
     close(fileHandle);
     return (size_t)fileSize;
@@ -782,7 +783,7 @@ static void FIO_munmap(FIO_Dict_t* dict)
     dict->dictBuffer = NULL;
     dict->dictBufferSize = 0;
 }
-static size_t FIO_initDictMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
+static size_t FIO_setDictBufferMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
 {
     HANDLE fileHandle, mapping;
     U64 fileSize;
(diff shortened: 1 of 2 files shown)


programs/fileio.c
```diff
@@ -1043,7 +1085,7 @@ static void FIO_adjustParamsForPatchFromMode(FIO_prefs_t* const prefs,
 static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
                                     const char* dictFileName, unsigned long long const maxSrcFileSize,
                                     int cLevel, ZSTD_compressionParameters comprParams) {
-    int useMMap = prefs->mmapDict == ZSTD_ps_enable;
+    int useMMap = prefs->mmapDict == ZSTD_ps_enable || 1;
```
daniellerozenblit: Doing this for testing.

programs/fileio.c
```diff
@@ -733,20 +734,15 @@ static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName, FIO_p
 
 #if (PLATFORM_POSIX_VERSION > 0)
 #include <sys/mman.h>
-static void* FIO_mmap(size_t fileSize, int fileHandle)
+static int FIO_munmap(const FIO_Dict_t* dict)
```
Cyan4973: It feels strange to use `const` here,
because the action is not just a simple "read",
it is effectively modifying the underlying state, 
therefore I would prefer `FIO_Dict_t* dict` type.

You could then use the mutability to set `dictBuffer=NULL` and `dictBufferSize=0` after the `munmap()` operation,
thus ensuring the `unmap` operation is properly tracked and there is no dangling pointer left behind.

programs/fileio.c
```diff
@@ -770,27 +766,73 @@ static size_t FIO_createDictBufferMMap(void** bufferPtr, const char* fileName, F
...
 
-    *bufferPtr = FIO_mmap((size_t)fileSize, fileHandle);
+    *bufferPtr = mmap(NULL, (size_t)fileSize, PROT_READ, MAP_PRIVATE, fileHandle, 0);
 
     close(fileHandle);
     return (size_t)fileSize;
 }
+#elif defined(_MSC_VER) || defined(_WIN32)
+#include <windows.h>
+static int FIO_munmap(const FIO_Dict_t* dict)
```
Cyan4973: same comment here, prefer `FIO_Dict_t* dict` as argument type.

programs/fileio.c
```diff
@@ -770,27 +766,73 @@ static size_t FIO_createDictBufferMMap(void** bufferPtr, const char* fileName, F
         }
     }
 
-    *bufferPtr = FIO_mmap((size_t)fileSize, fileHandle);
+    *bufferPtr = mmap(NULL, (size_t)fileSize, PROT_READ, MAP_PRIVATE, fileHandle, 0);
```
Cyan4973, line 771: minor : check that `fileSize` is not `> (U64)SIZE_MAX`, it's a prerequisite for safe casting to `(size_t)`.

(note: this situation could happen for files > 4 GB on 32-bit systems).

_edit_ : already checked above with `dictSizeMax`

programs/fileio.c
```diff
@@ -770,27 +766,73 @@ static size_t FIO_createDictBufferMMap(void** bufferPtr, const char* fileName, F
...
-static size_t FIO_createDictBufferMMap(void** bufferPtr, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
+static size_t FIO_createDictBufferMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
 {
-   return FIO_createDictBuffer(bufferPtr, fileName, prefs, dictFileStat);
+   return FIO_createDictBuffer(dict, fileName, prefs, dictFileStat);
 }
-static void FIO_munmap(void* buffer, size_t bufferSize) {
-   (void)bufferSize;
-   free(buffer);
+static void FIO_munmap(const FIO_Dict_t* dict) {
```
Cyan4973: same comment here, prefer `FIO_Dict_t* dict` as argument type.

programs/fileio.c
```diff
@@ -733,20 +734,15 @@ static size_t FIO_createDictBuffer(void** bufferPtr, const char* fileName, FIO_p
...
-    (NULL, (size_t)fileSize, PROT_READ, MAP_PRIVATE, fileHandle, 0);
+    return munmap(dict->dictBuffer, dict->dictBufferSize);
 }
-static int FIO_munmap(void* buffer, size_t bufferSize)
-{
-    return munmap(buffer, bufferSize);
-}
-/* We might want to also do mapping for windows */
-static size_t FIO_createDictBufferMMap(void** bufferPtr, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
+static size_t FIO_createDictBufferMMap(FIO_Dict_t* dict, const char* fileName, FIO_prefs_t* const prefs, stat_t* dictFileStat)
```
Cyan4973: discussion on naming convention : 

When I read a function name like `PREFIX_createType()`, I expect it to return a `Type*` pointer object. 
That's not the case here.

On a related note, if the object is allocated beforehand (on stack for example), I would consider `PREFIX_initType(Type* v, ...)` for the function name. 

I note that this situation is slightly different : you want to set member `dict->dictBuffer` with the outcome of `mmap` (and I presume `FIO_Dict_t` is larger than this member), _and_ return the underlying file size.

In which case, maybe `FIO_setDictBufferMMap()` ?

programs/fileio.c
```diff
@@ -1065,13 +1127,8 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
     }
 
-    ress.dict.dictBufferType = (useMMap && !forceNoUseMMap) ? FIO_mmapDict : FIO_mallocDict;
-
-    if (ress.dict.dictBufferType == FIO_mallocDict) {
-        ress.dict.dictBufferSize = FIO_createDictBuffer(&ress.dict.dictBuffer, dictFileName, prefs, &ress.dictFileStat);   /* works with dictFileName==NULL */
-    } else {
-        ress.dict.dictBufferSize = FIO_createDictBufferMMap(&ress.dict.dictBuffer, dictFileName, prefs, &ress.dictFileStat);
-    }
+    dictBufferType = 1 || (useMMap && !forceNoUseMMap) ? FIO_mmapDict : FIO_mallocDict;
```
daniellerozenblit: Testing, make sure to change before merge.

programs/fileio.c
```diff
@@ -1065,13 +1128,8 @@ static cRess_t FIO_createCResources(FIO_prefs_t* const prefs,
...
 
-    ress.dict.dictBufferType = (useMMap && !forceNoUseMMap) ? FIO_mmapDict : FIO_mallocDict;
-
-    if (ress.dict.dictBufferType == FIO_mallocDict) {
-        ress.dict.dictBufferSize = FIO_createDictBuffer(&ress.dict.dictBuffer, dictFileName, prefs, &ress.dictFileStat);   /* works with dictFileName==NULL */
-    } else {
-        ress.dict.dictBufferSize = FIO_createDictBufferMMap(&ress.dict.dictBuffer, dictFileName, prefs, &ress.dictFileStat);
-    }
+    dictBufferType = (useMMap && !forceNoUseMMap) ? FIO_mmapDict : FIO_mallocDict;
+    ress.dict.dictBufferSize = FIO_initDict(&ress.dict, dictFileName, prefs, &ress.dictFileStat, dictBufferType);   /* works with dictFileName==NULL */
```
Cyan4973: The return value of `FIO_initDict()` is used to initialize `ress.dict.dictBufferSize`.

Since the stated goal of `FIO_initDict()` is to initialize `ress.dict`,
I would rather expect `FIO_initDict()` to directly set `ress.dict.dictBufferSize` as part of the initialization.

`FIO_initDict()` should rather return a success/error code.

Cyan4973 (approved):
Looks good !

Please squash commits before merging

PR #3559, opened 2023-03-18
Fix a bug in the CLI tests newline processing, then simplify it further

ppentchev:
Hi,

First of all, thanks a lot for writing and maintaining zstd!

What do you think about the attached trivial changes? The first one fixes a type mismatch bug: an element of a bytes array is an int value, so it cannot be appended to a bytes array. The second one uses `.split(..., maxsplit=1)` to simplify the line splitting code.

Thanks again, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/tests/cli-tests/run.py
+++ b/tests/cli-tests/run.py
@@ -114,20 +114,12 @@ def pop_line(data: bytes) -> typing.Tuple[typing.Optional[bytes], bytes]:
     if data == b'':
         return (None, data)
 
-    newline_idx = data.find(b"\n")
-    if newline_idx == -1:
-        end_idx = len(data)
-    else:
-        end_idx = newline_idx + 1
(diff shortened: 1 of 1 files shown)


ppentchev:
The assertion failure in https://github.com/facebook/zstd/actions/runs/4457266947/jobs/7835977518#step:5:1222 does not seem related to the changes in the Python wrapper analyzing the programs' output.

yoniko:
Thank you for contributing, the changes look good to me.

> The assertion failure in https://github.com/facebook/zstd/actions/runs/4457266947/jobs/7835977518#step:5:1222 does not seem related to the changes in the Python wrapper analyzing the programs' output.

Yes, it's unrelated and is most likely due to the bug fixed in #3556.

yoniko (approved):

terrelln (approved):

PR #3568, opened 2023-03-22
Add instructions for building Universal2 on macOS via CMake

rickmark:
This includes instructions for building as a Universal2 binary.

We could consider making this default for CMake builds, which would make homebrew install fat versions of zstd by default preventing numerous issues when using the binary on M1 / AppleSilicon by making both the arm64 and x86_64 versions availiable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,18 @@ to create `zstd` binary, and `libzstd` dynamic and static libraries.
 
 By default, `CMAKE_BUILD_TYPE` is set to `Release`.
 
+#### Support for Fat (Universal2) Output
+
+`zstd` can be built and installed with support for both Apple Silicon (M1/M2) as well as Intel by using CMake's Universal2 support.  
+To perform a Fat/Universal2 build and install use the following commands:
+
+```bash
+cmake -B build-cmake-debug -S build/cmake -G Ninja -DCMAKE_OSX_ARCHITECTURES="x86_64;x86_64h;arm64;arm64e"
+cd build-cmake-debug
+ninja build
+sudo ninja install
+```
+
 ### Meson
 
 A Meson project is provided within [`build/meson`](build/meson). Follow
--- a/README.md
+++ b/README.md
@@ -154,6 +154,18 @@ to create `zstd` binary, and `libzstd` dynamic and static libraries.
 
 By default, `CMAKE_BUILD_TYPE` is set to `Release`.
 
+#### Support for Fat (Universal2) Output
+
+`zstd` can be built and installed with support for both Apple Silicon (M1/M2) as well as Intel by using CMake's Universal2 support.  
+To perform a Fat/Universal2 build and install use the following commands:
+
+```bash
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
The instructions look correct to me.

Worth mentioning though: on my local laptop (M1), the generated binaries crash immediately.
But this could be due to any number of external reasons, maybe some strict security policy which refuses fat binaries, or maybe the local compiler lacks something. Anyway, it's not clear what the problem could be on my local computer.

If it works fine on another M1 macos laptop, then I believe it's still worth documenting.

rickmark:
Just discovered, while macOS will successfully build arm64e, it will not run those binaries without some kernel switch as the ABI is not fixed.  This affects the bin but not the lib, but out of simplicity I simply removed the arm64e bits.

The open question is do we make this configuration default for CMAKE, which would make `brew install zstd` FAT (I don't think we ought to since that's not how homebrew has worked historically?)  Current discussion is here: https://github.com/Homebrew/brew/issues/10307

README.md
```diff
@@ -154,6 +154,18 @@ to create `zstd` binary, and `libzstd` dynamic and static libraries.
...
 
+#### Support for Fat (Universal2) Output
+
+`zstd` can be built and installed with support for both Apple Silicon (M1/M2) as well as Intel by using CMake's Universal2 support.  
+To perform a Fat/Universal2 build and install use the following commands:
+
+```bash
+cmake -B build-cmake-debug -S build/cmake -G Ninja -DCMAKE_OSX_ARCHITECTURES="x86_64;x86_64h;arm64"
+cd build-cmake-debug
+ninja build
```
Cyan4973: note: `ninja build` will fail here : `ninja: error: unknown target 'build'`.

I think what you mean is `ninja` (only), this will employ the default target, which is the release build of both the library and CLI.

Cyan4973 (changes requested):

Cyan4973:
Yes, this fixed the issue. Now the produced fat binary will run fine on my local M1 laptop.

> The open question is do we make this configuration default for CMAKE

I don't think there is any need to rush to that conclusion. Maybe later,
For the time being, the instructions look clear to me. It seems `brew` could simply adopt the same solution to generate fat binaries on its side, without the need to change the build recipe for everyone.

rickmark:
> Yes, this fixed the issue. Now the produced fat binary will run fine on my local M1 laptop.
> 
> > The open question is do we make this configuration default for CMAKE
> 
> I don't think there is any need to rush to that conclusion. Maybe later, For the time being, the instructions look clear to me. It seems `brew` could simply adopt the same solution to generate fat binaries on its side, without the need to change the build recipe for everyone.

I think for now it might be wise to make "fat builds" default, since it's not a huge library and will improve quality of life for anyone who depends on the library (as both will be available).  If a particular CMake is too old, it will simply ignore it.

PR #3569, opened 2023-03-22
Disable linker flag detection on MSVC/ClangCL.

### LINKED ISSUE #3522: Building 1.5.4 with clang-cl and lld-link fails because faulty linker checks in CMake.
**Describe the bug**

Build 1.5.4 with clang-cl/lld-link on Windows with CMake you get this:

```
11:19:41 lld-link: warning : ignoring unknown argument '-z'
11:19:41 lld-link: error : could not open 'noexecstack': no such file or directory
```

**To Reproduce**

`cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_LINKER=lld-link ..\build\cmake && ninja`

**Expected behavior**
The problem is the check here:
https://github.com/facebook/zstd/blob/dev/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake#L74

This is in the block for Clang, but checks linker flags by calling the compiler. On Windows the compiler is not used as the linker like on Unix systems.

It needs to be a check around this to correctly detect lld-link or re-tool the linker checks to use `CMAKE_LINKER` instead to perform these checks.

**Screenshots and charts**
n/a

 - OS: Windows 11
 - Compiler: Clang-cl 15.0.7
 - Build system: CMake

tru:
This fixes compilation with clang-cl on Windows. There
is a bug in cmake so that check_linker_flag() doesn't give
the correct result when using link.exe/lld-link.exe.

Details in CMake's gitlab: https://gitlab.kitware.com/cmake/cmake/-/issues/22023

Fixes #3522

### DIFF
--- a/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
+++ b/build/cmake/CMakeModules/AddZstdCompilationFlags.cmake
@@ -33,7 +33,12 @@ function(EnableCompilerFlag _flag _C _CXX _LD)
         # implement CHECK_LINKER_FLAG() like feature for CMake < 3.18
         # or require CMake >= 3.18 when we need to add a required
         # linker flag in future.
-        if (ZSTD_HAVE_CHECK_LINKER_FLAG)
+        #
+        # We also skip linker flags check for MSVC compilers (which includes
+        # clang-cl) since currently check_linker_flag() doesn't give correct
+        # results for this configuration,
+        # see: https://gitlab.kitware.com/cmake/cmake/-/issues/22023
+        if (ZSTD_HAVE_CHECK_LINKER_FLAG AND NOT MSVC)
             CHECK_LINKER_FLAG(C ${_flag} LD_FLAG_${varname})
         else ()
             set(LD_FLAG_${varname} false)


Cyan4973 (approved):

Cyan4973:
The fix looks fine,
but I wonder if we could add, or update, a test,
that would trigger this problem without this fix.
This way, we would ensure that this property remains respected in the future.

I suspected we used to have a Windows clang-cl test on Appveyor,
but somehow it was lost in the transition towards Github Actions.

tru:
When you say test - do you mean a build bot in this case that would build zstd with clang-cl? I am not very familiar with the github actions workers and configurations, maybe someone else could help with that?

Cyan4973:
> When you say test - do you mean a build bot in this case that would build zstd with clang-cl? I am not very familiar with the github actions workers and configurations, maybe someone else could help with that?

I'll look into it

Cyan4973:
I added a `Clang-CL` test in #3579, and it fails at link stage, as expected.

Now, I wonder if the exact way it fails is the expected one, as I don't recognise the gitlab issue : 
https://github.com/facebook/zstd/actions/runs/4549108695/jobs/8020846731?pr=3579

```
lld-link : warning : ignoring unknown argument '-z' [D:\a\zstd\zstd\build\cmake\build\programs\zstd.vcxproj]
lld-link : error : could not open 'noexecstack': no such file or directory [D:\a\zstd\zstd\build\cmake\build\programs\zstd.vcxproj]
```

If yes, then it means that merging this PR, then rebasing #3579 on top of it, should fix the problem.

tru:
Great! Thanks for the CI build, that hopefully means that our configuration will continue to work in the future!

PR #3570, opened 2023-03-23
[easy] minor doc update for --rsyncable

### LINKED ISSUE #3567: please be more clear about rsyncable performance impact in manpage
manpage tells on --rsyncable:

       •   --rsyncable  :  zstd  will  periodically  synchronize the compression state to make the 
       compressed file more rsync-friendly. There is a negligible impact to compression ratio, and the
           faster compression levels will see a small compression speed hit. This feature does not work 
           with --single-thread. You probably don´t want to use it with long  range  mode,  since  it
           will decrease the effectiveness of the synchronization points, but your milage may vary.
           
           
"the faster compression levels will see a small compression speed hit"  is nothing i can confirm, please have a look at https://bugzilla.proxmox.com/show_bug.cgi?id=4605

proxmox is using --rsyncable as a default for zstd compression , which unfortunately can't be disabled.

the data i collected does not confirm that there is just a "small compression speed hit"

it's quite big difference even with default compression settings, so that should at least be documented, if not fixable.

also see https://github.com/facebook/zstd/issues/3097

i'm using v1.4.8 on debian 11

Cyan4973:
rephrase the speed impact, as suggested by @devZer0.
fix #3567

### DIFF
--- a/programs/zstd.1.md
+++ b/programs/zstd.1.md
@@ -177,9 +177,10 @@ the last one takes effect.
         (i.e. 4096), and by setting a large `--zstd=chainLog=`.
 * `--rsyncable`:
     `zstd` will periodically synchronize the compression state to make the
-    compressed file more rsync-friendly. There is a negligible impact to
-    compression ratio, and the faster compression levels will see a small
-    compression speed hit.
+    compressed file more rsync-friendly.
+    There is a negligible impact to compression ratio,
+    and a potential impact to compression speed, perceptible at higher speeds,
+    for example when combining `--rsyncable` with many parallel worker threads.
     This feature does not work with `--single-thread`. You probably don't want
     to use it with long range mode, since it will decrease the effectiveness of
     the synchronization points, but your mileage may vary.


devZer0:
>    There is a negligible impact to compression ratio, and a small impact to compression speed,    
>    only perceptible at higher speeds, typically when combining `--rsyncable` with many parallel 
>     worker threads.
    
thank you - but - sorry -  i don't want to be dogmatic, nor do i want to be pedantic ,  rude or a nagger - but even in this optimzed form, what's being claimed here simply isn't true. 

otherwise, my perception of "negligible impact" is at least very different from your's.

let me show the difference again:

```
# dd if=/dev/zero of=zeroes.dat bs=1024k count=4096
4096+0 records in
4096+0 records out
4294967296 bytes (4.3 GB, 4.0 GiB) copied, 2.66998 s, 1.6 GB/s

# dd if=/dev/urandom of=random.dat bs=1024k count=4096
4096+0 records in
4096+0 records out
4294967296 bytes (4.3 GB, 4.0 GiB) copied, 20.9213 s, 205 MB/s

# time /root/zstd/zstd -c random.dat >/dev/null

real	0m4.699s
user	0m7.190s
sys	0m2.999s

# time /root/zstd/zstd -c zeroes.dat >/dev/null

real	0m2.433s
user	0m3.518s
sys	0m2.477s

# time /root/zstd/zstd -c random.dat --rsyncable >/dev/null

real	0m19.598s
user	0m25.092s
sys	0m4.188s

# time /root/zstd/zstd -c zeroes.dat --rsyncable >/dev/null

real	0m18.160s
user	0m20.175s
sys	0m3.692s

Cyan4973:
> my perception of "negligible impact" is at least very different from your's.

The expression "negligible impact" is associated to the concept of "compression ratio", as in the sentence : 

> There is a negligible impact to compression ratio

If you were to report the compression ratio of the 2 corner case scenarios used as illustration, you would see that this statement holds up pretty well.

devZer0:
you are right, i mixed that up. sorry for that. 

and thank you for fixing the docs!

PR #3574, opened 2023-03-27
[contrib/pzstd] Select `-std=c++11` When Default is Older

felixhandte:
Rather than remove the flag entirely, as proposed in #3499, this commit uses the newest C++ standard the compiler supports. This retains the selection of using only standardized features (excluding GNU extensions) and keeps the recency requirements of the codebase explicit.

Tested with various versions of `g++` and `clang++`.


Cyan4973:
What benefit is there in selecting a C++ standard newer than C++11 ?

felixhandte:
As noted in #3499, newer gtest versions require >=C++14 (and pzstd tries to find a system-installed gtest for its tests).

Cyan4973 (approved):
nice!

PR #3576, opened 2023-03-28
Couple tweaks to improve decompression speed with clang PGO compilation

zhuhan0:
- Adding inline hint to `BIT_reloadDStream` to improve register allocation down the pipeline.
- Remove clang-only `__builtin_expect`s from `ZSTD_decodeSequence` to lean on clang PGO profile data instead.

Please see the individual commits for some analysis details.

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -396,7 +396,7 @@ MEM_STATIC BIT_DStream_status BIT_reloadDStreamFast(BIT_DStream_t* bitD)
  *  This function is safe, it guarantees it will not read beyond src buffer.
  * @return : status of `BIT_DStream_t` internal register.
  *           when status == BIT_DStream_unfinished, internal register is filled with at least 25 or 57 bits */
-MEM_STATIC BIT_DStream_status BIT_reloadDStream(BIT_DStream_t* bitD)
+MEM_STATIC FORCE_INLINE_ATTR BIT_DStream_status BIT_reloadDStream(BIT_DStream_t* bitD)
 {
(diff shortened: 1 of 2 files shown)


Cyan4973:
The explanations look good to me, now a couple of observation : 

- For patch 1 : I note this is considered desirable for `clang` with `pgo`. But what's the impact for `gcc` without `pgo` ?
Some benchmarking would be welcome.

- For patch 2 : I note that the removed branch was recently added by @danlark1 . I'm curious about his own perspective on this proposed modification.

zhuhan0:
> The explanations look good to me, now a couple of observation :
> 
> * For patch 1 : I note this is considered desirable for `clang` with `pgo`. But what's the impact for `gcc` without `pgo` ?
>   Some benchmarking would be welcome.
> * For patch 2 : I note that the removed branch was recently added by @danlark1 . I'm curious about his own perspective on this proposed modification.

Gotcha. Here're some numbers for patch 1.

clang 15 thinLTO + PGO (15 runs average)
698.83 => 721.87 (+3.3%)

gcc 11 (15 runs average)
728.17 => 728.95 (neutral)

When I measured patch 2 alone there was ~2% win but it disappeared after combined with patch 1. They likely both influence the register allocator to spill less hot code, which is the main driver here. But we expect removing branch hints and relying on PGO will give better result in production because the branch probabilities will be specialized based on the workload.

danlark1:
Not going to block the patch, we checked and PGO for us gives around 2.8%. Second commit is neutral

Thanks!

Cyan4973 (approved):

zhuhan0:
@Cyan4973 I don't have write permission. Could you help merge this?

Cyan4973:
Don't worry, it will be merged

PR #3577, opened 2023-03-28
updated version number to v1.5.5

Cyan4973:
also : updated man pages

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,22 @@
+v1.5.5 (Apr 2023)
+fix: fix rare corruption bug affecting the high compression mode, reported by @danlark1 (#3517, @terrelln)
+perf: improve mid-level compression speed (#3529, #3533, #3543, @yoniko and #3552, @terrelln)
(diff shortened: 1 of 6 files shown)


CHANGELOG
```diff
@@ -1,3 +1,21 @@
+v1.5.5 (Mar 2023)
+fix: fix rare corruption bug affecting the high compression mode (#3517) by @terrelln
```
ghost: Intentionally finding such corruption bug is great, I would suggest to add discoverer's name/id. It seems he is a testing team member.

Cyan4973: Discoverer's name is planned to be part of the release note.

But indeed, it could be added to the changelog too.

gaby: @Cyan4973 Is this getting released today or moved to April?

Cyan4973: It's coming

danlark1: Not really a testing team member. Just the owner of ZSTD sync at Google and adviser on APIs, etc :)

PR #3578, opened 2023-03-29
pinned ossfuzz action version to specific hash

Cyan4973:
this is labelled "good practice" by "Code Scanning".

But I note that the version tracked is @master, 
so I'm not sure if this is _intentionally_ a floating version,
in which case, pinning to a specific hash is not helpful.


Cyan4973:
@terrelln : curious what you think about this one.

terrelln:
OSS-Fuzz doesn't have any releases... Probably best just to explicitly use master, and comment that they don't maintain releases. Otherwise, I'm not sure what dependabot will do. It might spam us with PRs every day.

PR #3579, opened 2023-03-29
added a Clang-CL Windows test to CI

Cyan4973:
If I understand correctly,
this should trigger the issue notified in #3569.


Cyan4973:
Indeed, the `Clang-CL` test fails at link stage, as announced by #3569.

If that's correct, then merging #3569 and rebasing this PR on top of it should fix the issue.

Cyan4973:
And yes, indeed, the new `Clang-CL` Windows test pass with #3569 fix.

That's one more build scenario that is now checked by GA CI.

Issue #3597, opened 2023-04-05
Release missing win32 zip files. zstd-v1.5.4-win32.zip and zstd-v1.5.5-win32.zip


Kim-SSi:
**Describe the bug**
The 1.5.4 and 1.5.5 releases are missing the win32 zip files,
zstd-v1.5.4-win32.zip and zstd-v1.5.5-win32.zip

**To Reproduce**
Go to Assets on the Release pages
https://github.com/facebook/zstd/releases/tag/v1.5.5
https://github.com/facebook/zstd/releases/tag/v1.5.4

**Expected behavior**
1.5.2 has a zstd-v1.5.2-win32.zip

Cyan4973:
We lost our ability to generate `win32` artifacts in the migration to Github Actions.
We'll see if we can make it happen again, though it will take some time, so no promise.

Another option could be to generate these artifacts manually, but this requires a properly setup Windows computer.

Kim-SSi:
@Cyan4973 
I have opened a PR #3600 with changes to get the win32 artifacts building using GitHub Actions.

Kim-SSi:
PR merged fixed this

PR #3600, opened 2023-04-05
Add win32 to windows-artifacts.yml

Kim-SSi:
Add back win32 builds
In reference to issue https://github.com/facebook/zstd/issues/3597

Run showing the build of zstd-*-win32.zip and zstd-*-win64.zip:
https://github.com/Kim-SSi/zstd/actions/runs/4615547751

### DIFF
--- a/.github/workflows/windows-artifacts.yml
+++ b/.github/workflows/windows-artifacts.yml
@@ -49,3 +49,39 @@ jobs:
       with:
         path: ${{ github.workspace }}/zstd-${{ github.ref_name }}-win64.zip
         name: zstd-${{ github.ref_name }}-win64.zip
+
+  windows-32-artifacts:
+    # see https://ariya.io/2020/07/on-github-actions-with-msys2
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3 # tag=v3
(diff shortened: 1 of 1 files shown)


Cyan4973:
I'm surprised that modifying `msystem:` is all it takes to generate valid 32-bit applications.
I *thought* I had already tried that and failed (i.e. generated binaries were *still* 64-bit), but apparently, this outcome wasn't correct ...

More complex: as a consequence, the new recipe for `win32` binaries is very close to the `win64` one, to the point that it could be achieved via a matrix. That would be better for code maintenance (less duplication).
Is that something you would be able to achieve ?

Kim-SSi:
@Cyan4973 
I did look at using matrix but then you need to also have the different install lines and the win32/win64 used at the end also need to be set somewhere.
Do you know a good way to handle the win32/win64 in the yml?

Cyan4973:
I believe I know a few ways,
but if it's not trivial to do now,
then it's probably better to take this PR as is, and update the matrix later as a code optimization.

Cyan4973 (approved):

Kim-SSi:
I will try to see if I can get it to work

Kim-SSi:
It turns out there is an example of this on https://github.com/msys2/setup-msys2, that can be changed to handle this problem.
I have updated the file to use strategy.matrix.
Results:
https://github.com/Kim-SSi/zstd/actions/runs/4616355226

Cyan4973 (approved):
Looks good ! Thanks @Kim-SSi !

PR #3614, opened 2023-04-13
add makefile entry to build fat binary on macos

mredig:
I had the need for a cli binary that runs on both Intel and ARM Macs. I figured I'd contribute my success back to the main project in hopes that it'll help someone else too.

I admit I have little familiarity with makefiles, so if there are any changes needed for style or anything, I'm happy to oblige. All I can say is that it works and I feel like I got close to what would be desired for makefile style presented.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,15 @@ uninstall:
 travis-install:
 	$(MAKE) install PREFIX=~/install_test_dir
 
+.PHONY: clangbuild-darwin-fat
+clangbuild-darwin-fat: clean
+	clang -v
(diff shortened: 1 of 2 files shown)


Cyan4973:
Note : the following entry on fat binaries was recently added to the documentation : 
https://github.com/facebook/zstd#support-for-fat-universal2-output

This PR seems to target the same objective, though in a different way.

mredig:
Haha that's an incredible coincidence: the history shows that that was added about the same time that I developed my solution, I only now had a chance to clean it up for a PR. 

I have no idea which solution would be considered more robust, but my PR doesn't have any extra requirements for ninja or cmake, just the same basics as what's already needed for other parts of the project. 

I'll defer to others with more knowledge on the topic. 😄

Cyan4973 (approved):

PR #2751, opened 2021-08-20
Reduce size of dctx by reutilizing dst buffer

binhdvo:
WIP, this round of optimizations has gotten performance much closer to parity, though it has introduced a checksum error in the 270MB file test I'm still tracking down.  This however hasn't affected the smaller size tests; benchmarks indicate that in some cases we now see performance improvements on top of the memory reduction due to the improved cache behavior.  However there's other cases, at low file sizes and high compressibility, where we are still about 1% behind parity.

<details>
<summary>Benchmark</summary>

old performance

./tests/fullbench -b2 -B1000 -P0
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 1000 bytes :                                                  
 2#decompress                   :  4987.5 MB/s  (    1000) 
./tests/fullbench -b2 -B1000 -P10
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 1000 bytes :                                                  
 2#decompress                   :   612.0 MB/s  (    1000) 
./tests/fullbench -b2 -B1000 -P50
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 1000 bytes :                                                  
 2#decompress                   :   585.8 MB/s  (    1000) 
./tests/fullbench -b2 -B1000 -P90
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 1000 bytes :                                                  
 2#decompress                   :  2597.8 MB/s  (    1000) 
./tests/fullbench -b2 -B1000 -P100
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 1000 bytes :                                                  
 2#decompress                   :  2635.7 MB/s  (    1000) 
./tests/fullbench -b2 -B10000 -P0
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 10000 bytes :                                                 
 2#decompress                   : 36167.5 MB/s  (   10000) 
./tests/fullbench -b2 -B10000 -P10
*** Zstandard speed analyzer 1.5.0 64-bits, by Yann Collet (Aug 19 2021) ***
 Sample 10000 bytes :                                                 
 2#decompress                   :  1292.4 MB/s  (   10000) 
[...]

PR #3616, opened 2023-04-14
Reduce streaming decompression memory by (128KB - blockSizeMax)

terrelln:
The split literals buffer patch (PR #2751) increased streaming decompression memory by 64KB (shrunk lit buffer from 128KB to 64KB, and added 128KB). This patch shrinks the fixed 128KB buffer to the actual blockSizeMax. So for windowSize = 64KB, this patch saves 64KB.

The original patch just used `ZSTD_BLOCKSIZE_MAX` for simplicity, but that is wasteful when the maximum block size is smaller.

RFC: A follow up to this PR is to allow configuring `ZSTD_d_maxBlockSize` to match `ZSTD_c_maxBlockSize`. This flag would reject frames with blocks larger than the configured `ZSTD_d_maxBlockSize`. For example, if the `windowSize = 64KB`, and `maxBlockSize = 4KB`, then we'd save `2 * (64KB - 4KB) = 120KB` from the `ZSTD_DCtx` on top of the 64KB this patch would save.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -265,6 +265,7 @@ static void ZSTD_initDCtx_internal(ZSTD_DCtx* dctx)
 #endif
     dctx->noForwardProgress = 0;
     dctx->oversizedDuration = 0;
+    dctx->isFrameDecompression = 1;
 #if DYNAMIC_BMI2
     dctx->bmi2 = ZSTD_cpuSupportsBmi2();
 #endif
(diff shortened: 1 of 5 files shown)


terrelln:
This PR should be relatively safe, because it isn't doing anything that complex. We are keeping all the same assumptions, just reducing `ZSTD_BLOCKSIZE_MAX` to `blockSizeMax`.

The savings were less than I was thinking, because trying to do anything more gets complex. But, it combines nicely with PR #3617, and together we get big memory savings for use cases that force a small block size. E.g. for streaming 4KB packets over the network.

Cyan4973 (approved):

PR #3620, opened 2023-04-17
[doc] add decoder errata paragraph

Cyan4973:
for compressed blocks of size exactly 128 KB
which used to be disallowed by the spec
but have become allowed in more recent version of the spec.

While this limitation is fixed in decoders `v1.5.4`+, implementers should refrain from generating such block with their custom encoder as they could be misinterpreted as corrupted by older decoder versions.


doc/decompressor_errata.md
```diff
@@ -6,12 +6,29 @@ Each entry will contain:
...
 
+Compressed block with a size of exactly 128 KB
+------------------------------------------------
+
+**Last affected version**: v1.5.2
+
+**Affected decompressor component(s)**: Library & CLI
+
+**Produced by the reference compressor**: No
+
```
terrelln, line 23: Can you please provide an example frame?

Since it is 128KB to produce it, maybe just add an example to https://github.com/facebook/zstd/tree/dev/tests/golden-decompression and provide a link here.

Adding the file there also automatically enables it as a test case.

terrelln (approved):
LGTM once you link an example file.

Issue #2939, opened 2021-12-20
Allow Build-Time Exclusion of Compression Components


felixhandte:
**Is your feature request related to a problem? Please describe.**
Zstandard has grown into a large tool that tries to serve a wide range of compression use cases. In order to achieve the best performance possible across this spectrum, Zstd ships a number of matchfinders optimized for different points in the speed-ratio curve (`ZSTD_fast`, `ZSTD_dfast`, `ZSTD_greedy`, `ZSTD_lazy`, `ZSTD_lazy2`, `ZSTD_btlazy`, and `ZSTD_btopt`). Each of those strategies are then instantiated many times, for different memory layouts (single-segment, split-segment, and dict-match-state) and for different minimum match lengths (4, 5, 6, or 7). All of these implementations end up producing a pretty fat binary.

Similarly, though less egregiously, the huffman encoder implements 4 different modes (1X1, 1X2, 4X1, 4X2).

A number of use cases have appeared which want to use Zstandard but where binary size is a significant cost. E.g., busybox, mobile apps, embedded systems, browsers.

**Describe the solution you'd like**
We should create tooling--a set of macros they can define--that allows integrators to select a subset of zstd's compression implementations to include. This would allow space-constrained integrators to select only the functionality they need and exclude the rest. I expect this can potentially allow a huge reduction in binary size (>75%?).

Zstd will need to be robust to these exclusions and gracefully degrade when configurations are requested which have not been compiled in.

PR #3623, opened 2023-04-19
Allow Build-Time Exclusion of Individual Compression Strategies

felixhandte:
This PR adds a whole bunch of macro guards that allow build-time removal of the implementations of individual compression strategies. This represents the biggest axis of build-time configurability we want to give users in order to be able to produce small libraries that are still capable of compression. (Other future axes might include excluding dictionary support, row hash, DDSS, LDM, etc.) Trial compilations show possible savings on the order of 160KB, i.e. >50% of the library size, if all non-fast strategies are excluded.

This PR does not allow the exclusion of `ZSTD_fast`.

This PR currently resolves situations where the user has selected a strategy that isn't available (either by setting a compression level or by explicit cparams) by finding the next-highest strategy that is available and silently downgrading to that, perturbing the other cparams only to resolve the overloaded meaning of `targetLength` between `ZSTD_fast` and the high-ratio strategies.

This PR addresses #2939.


felixhandte:
I've been running the following test:

```
for EXCLUDE_DFAST in "-DZSTD_EXCLUDE_DFAST_BLOCK_COMPRESSOR" ""; do
  for EXCLUDE_GREEDY in "-DZSTD_EXCLUDE_GREEDY_BLOCK_COMPRESSOR" ""; do
    for EXCLUDE_LAZY in "-DZSTD_EXCLUDE_LAZY_BLOCK_COMPRESSOR" ""; do
      for EXCLUDE_LAZY2 in "-DZSTD_EXCLUDE_LAZY2_BLOCK_COMPRESSOR" ""; do
        for EXCLUDE_BTLAZY2 in "-DZSTD_EXCLUDE_BTLAZY2_BLOCK_COMPRESSOR" ""; do
          for EXCLUDE_BTOPT in "-DZSTD_EXCLUDE_BTOPT_BLOCK_COMPRESSOR" ""; do
            for EXCLUDE_BTULTRA in "-DZSTD_EXCLUDE_BTULTRA_BLOCK_COMPRESSOR" ""; do
              echo -- "$EXCLUDE_DFAST $EXCLUDE_GREEDY $EXCLUDE_LAZY $EXCLUDE_LAZY2 $EXCLUDE_BTLAZY2 $EXCLUDE_BTOPT $EXCLUDE_BTULTRA"

              make -C .. clean
              make clean
              make -j fuzzer MOREFLAGS="-Wall -Wextra -Werror $EXCLUDE_DFAST $EXCLUDE_GREEDY $EXCLUDE_LAZY $EXCLUDE_LAZY2 $EXCLUDE_BTLAZY2 $EXCLUDE_BTOPT $EXCLUDE_BTULTRA"

              ./fuzzer -T1m
            done;
          done;
        done;
      done;
    done;
  done;
done;
```

This should validate that there aren't any symbols whose guards are incorrect (either too strict in which case it will have an undefined symbol error or too loose in which case it will error about an unused symbol). So far everything looks good.

Such a test is not practical to ship in the CI, at least as it currently stands. But as the number of possible build modes grows, we will need to maintain some kind of coverage of them. So I'm not really sure what the testing story should be.

lib/compress/zstd_compress.c
```diff
@@ -3280,22 +3333,27 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+                            zc->appliedParams.cParams.strategy,
+                            zc->appliedParams.useRowMatchFinder,
+                            dictMode);
                     ms->ldmSeqStore = NULL;
                     DEBUGLOG(
                         5,
                         "External sequence producer returned error code %lu. Falling back to internal parser.",
                         (unsigned long)nbExternalSeqs
                     );
+                    RETURN_ERROR_IF(blockCompressor == NULL, parameter_combination_unsupported, "Got NULL block compressor!");
```
Cyan4973: Question :
is this outcome `blockCompressor == NULL` possible,
meaning there is a combination of parameters that makes this situation reachable in production,
or is it essentially impossible, because it's part of the contract, and caller of `ZSTD_buildSeqStore()` are in charge of respecting it ?

lib/compress/zstd_compress.c
```diff
@@ -4760,12 +4818,17 @@ static size_t ZSTD_loadDictionaryContent(ZSTD_matchState_t* ms,
         ZSTD_fillHashTable(ms, iend, dtlm, tfp);
         break;
     case ZSTD_dfast:
+#ifndef ZSTD_EXCLUDE_DFAST_BLOCK_COMPRESSOR
```
Cyan4973, line 4819: What happens in this case ?
It seems that the dictionary content gets silently skipped, and the user will just never get matches into the dictionary, but I suspect everything continues to work normally, so the impact can remain invisible.
Would it be better to return an error ?

Cyan4973:
That's much better.
I still find a few files like `zstd_lazy.h` a bit difficult to read, and wonder if the declarations could be re-organized to make the `#ifdef` sections a bit easier to follow, but even then, that's miles better than it was before, when the `#ifdef` where distributed across many other files.

lib/compress/zstd_compress.c
felixhandte: The first version of this functionality was designed around failing compressions that selected an unavailable compressor. I then changed the functionality to resolve unavailable selections down to an available implementation (in `adjustCParams()`). An `assert()` in `selectBlockCompressor()` remains active, validating that we never fail to resolve the cparams to an available implementation.

I though it might be harmless to leave these checks in here, but I guess it's not worth it. I'll go ahead and remove these.

lib/compress/zstd_compress.c
felixhandte: FIxed to assert that it's not called if excluded.

lib/compress/zstd_compress.c
Cyan4973: > I though it might be harmless to leave these checks in here, but I guess it's not worth it. I'll go ahead and remove these.

If by "harmless" you mean "no performance impact", then it's probably correct.

The comment was more targeted at understanding the code during a review: 
if there is an `if` branch, it's implied that both branches *could* be taken, given the proper circumstances.
Whereas, if it's an `assert()`, it means that this situation can never happen, and the code must ensure it.
This has consequences on the calling side (what invariants must be checked by the caller), on the contract documentation, and on the follow-up actions for the caller.

I don't remember the specifics, but I think I had in mind something along this line : " if `blockCompressor` can be `NULL` here, then the function returns immediately here, but then, what are the consequences for the rest of the system ? Are they properly propagated ?".
Whereas, if it's not possible to happen, then there is no consequence to worry about.

Cyan4973:
It looks in pretty good shape !
Let me know when you think this PR is no longer in _draft_ mode.

lib/compress/zstd_compress.c
felixhandte: Indeed. For that reason, I've removed them.

.github/workflows/dev-short-tests.yml
```diff
@@ -348,6 +348,7 @@ jobs:
         make clean && make check MOREFLAGS="-Werror -DHUF_FORCE_DECOMPRESS_X2 -DZSTD_FORCE_DECOMPRESS_SEQUENCES_LONG"
         make clean && make -j all MOREFLAGS="-Werror -DZSTD_NO_INLINE -DZSTD_STRIP_ERROR_STRINGS"
         make clean && make check MOREFLAGS="-Werror -DZSTD_NO_INLINE -DZSTD_STRIP_ERROR_STRINGS"
+        make clean && make check ZSTD_LIB_EXCLUDE_COMPRESSORS_DFAST_AND_UP=1 MOREFLAGS="-Werror"
```
Cyan4973, line 351: So, the only test shipped is this one, which excludes all compressors except the fast one ?

felixhandte: Yes. In the future we could consider supporting/testing more combinations, but I think this is what is of interest right now.

I'll add some documentation in lib/README.md.

felixhandte:
Following up on the dfast size test I said I'd run: for a plausible aarch64 build, excluding everything but fast compiles to 112KB. Including dfast on top of that adds 20KB (+18%). So I'm not sure how attractive that will be to folks. But I've added the option.

lib/compress/zstd_lazy.c
```diff
@@ -2101,57 +2149,71 @@ size_t ZSTD_compressBlock_lazy_extDict_generic(
...
+#endif
 
+#ifndef ZSTD_EXCLUDE_LAZY2_BLOCK_COMPRESSOR
 size_t ZSTD_compressBlock_lazy2_extDict_row(
         ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
         void const* src, size_t srcSize)
 {
     return ZSTD_compressBlock_lazy_extDict_generic(ms, seqStore, rep, src, srcSize, search_rowHash, 2);
 }
+#endif
```
Cyan4973, line 2180: Question :
I see a lot of `#ifdef` blocks in this unit, sometimes using the same trigger condition.

Would it make sense to regroup some of these (fairly shorts) implementations in order to reduce the nb of `#ifdef`  blocks ?
Intuitively, it seems it would help for maintenance.

lib/compress/zstd_opt.c
```diff
@@ -1438,34 +1451,43 @@ size_t ZSTD_compressBlock_btultra2(
...
+#endif
 
+#ifndef ZSTD_EXCLUDE_BTULTRA_BLOCK_COMPRESSOR
 size_t ZSTD_compressBlock_btultra_extDict(
         ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
         const void* src, size_t srcSize)
 {
     return ZSTD_compressBlock_opt2(ms, seqStore, rep, src, srcSize, ZSTD_extDict);
 }
+#endif
```
Cyan4973, line 1484: similar question here, regarding a reduction in nb of `#ifdef` blocks.

Cyan4973:
That's pretty good !
I have one last question, and I think it's good to go !

Cyan4973 (approved):

PR #3636, opened 2023-05-01
Get zstd working with ARM64EC on Windows

dunhor:
When building for ARM64EC, the [compiler provides definitions](https://techcommunity.microsoft.com/t5/windows-os-platform-blog/getting-to-know-arm64ec-defines-and-intrinsic-functions/ba-p/2957235) for `_M_X64`, `_M_AMD64`, and `_M_ARM64EC`. The existing code in zlib will see the definition of `_M_X64` and then include `mmintrin.h`, which then fails compilation on this line:
```c++
#if (defined(_M_ARM64) || defined(_M_ARM64EC)) && !defined(__INTRIN_H_)
#error this header should only be included through <intrin.h>
#endif
```
One option would be to instead include `intrin.h` - which seems to work - however for ARM64EC, `_mm_prefetch` gets defined as `__noop((a), (b))` which results in "unused local variable" warnings. This seems like the more appropriate solution.

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -112,7 +112,7 @@
 #  define PREFETCH_L1(ptr)  (void)(ptr)  /* disabled */
 #  define PREFETCH_L2(ptr)  (void)(ptr)  /* disabled */
 #else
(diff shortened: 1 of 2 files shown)


terrelln:
Thanks for the PR @dunhor, this looks good to me!

PR #3645, opened 2023-05-13
fix clobbering RBX register in LTO mode within ZSTD_cpuid

virt00l:
Just rewrite assembly to using std intrinsics for all compilers. It doing right thing by default (e.g. preserving aftermentioned register).

Bug actually was caught on certain Japan's console in LTO mode (PIC is also turned on) but I suspect that same would be true for other platforms as well (e e.g. Linux+PIC+LTO)


virt00l:
Summited CLA as well

virt00l:
Sorry, I get a bit confused and accidently closed this PR.
Created new one - https://github.com/facebook/zstd/pull/3646

PR #3658, opened 2023-05-25
Fix UBSAN issue (zero addition to NULL)

terrelln:
Fix UBSAN issue that came up internally.


yoniko (approved):
Looks good

Cyan4973 (changes requested):
It seems that there is a problem with the issue itself: 

if execution enters this branch, it means that the caller specifically set the `ZSTD_bm_stable` property for the input buffer. This implies, [per the documentation](https://github.com/facebook/zstd/blob/v1.5.5/lib/zstd.h#L2016), that the input buffer doesn't change during the streaming operation.
Consequently, providing `NULL` as input buffer in the middle of streaming operation is in direct violation with this contract.

_Edit_ : I incorrectly assumed this was a streaming use case,
but this code path can also be taken via `ZSTD_compress2()`,
by receiving an empty buffer as input, which is then allowed to be `NULL`.
That's a different scenario, because it's one we have accepted as valid (in contrast with the one described above).

Cyan4973 (approved):

PR #3665, opened 2023-06-02
Fix Intel Xcode builds with assembly

### LINKED ISSUE #3622: Asm code is not compiling under Xcode
I have added zstd library to my project as subdirectory using CMake. And when I configure and try to compile using Xcode it shows error in huff_decompress asm code. Problem appeared from 1.5.4 release version.

1. Add zstd as subdirectory in CMake
2. Configure Xcode project (cmake -G Xcode...)
3. Open configured .xcodeproj and start program compilation
4. See error in huf_decompress_amd64.S "Expected identifier or '('" at line 37

Expected behavior:
Program compiling without errors.

Desktop:
 - OS: macOS
 - Version Ventura 13.3.1
 - Compiler clang Xcode 14.3
 - Flags -O0
 - Build system CMake

gjasny:
When forcing the source file language to `C`, Xcode enforces the file to be compiled as `C` by appending `-x c` to the compiler command line.

For now try to limit the damage and only enforce the language if the ASM and C compilers differ.

Reproducer (CMake `3.26.4`, Xcode `14.3`):
```
cmake -S build/cmake -B _b -GXcode -DCMAKE_OSX_ARCHITECTURES=x86_64
cmake --build _b
```

Fix: #3622

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -105,7 +105,9 @@ endif ()
 # Our assembly expects to be compiled by a C compiler, and is only enabled for
 # __GNUC__ compatible compilers. Otherwise all the ASM code is disabled by
 # macros.
-set_source_files_properties(${Sources} PROPERTIES LANGUAGE C)
+if(NOT CMAKE_ASM_COMPILER STREQUAL CMAKE_C_COMPILER)
+    set_source_files_properties(${Sources} PROPERTIES LANGUAGE C)
+endif()
 
 macro (add_definition target var)
     if (NOT ("${${var}}" STREQUAL ""))


gjasny:
I signed the CLA.

Cyan4973:
The fix looks simple enough to me.
It's just a pity we don't have a CI test that would produce the error you've been experiencing,
this would ensure that the issue is really fixed, and more importantly, that it will remain fixed in future modifications.
That being said, some tests are more difficult to deploy in CI, and for XCode specifically, I have no idea.

@terrelln , any comment on this PR ?

Cyan4973 (approved):

cfnptr:
All checks passed. Looks like we can merge this fix?

Cyan4973:
All checks passed _because_ there is no check that exercises the code path that's added in this commit.

That being said, no proposal emerged to create a relevant test, and after so much time, I doubt it ever will.
So I'll let it pass, and merge,
keeping in mind that an untested property is a property that might be silently lost in a future update.

cfnptr:
@Cyan4973, I think we can add an Xcode runner to catch this kind of bugs. Apple allows building Xcode projects from the command line: https://developer.apple.com/library/archive/technotes/tn2339/_index.html.

Something like this:
1. cmake -G Xcode -DCMAKE_BUILD_TYPE=Release -S build/cmake/ -B build-xcode/
2. xcodebuild -project build-xcode/zstd.xcodeproj -scheme ALL_BUILD build

PR #3669, opened 2023-06-05
fixed decoder behavior when nbSeqs==0 is encoded using 2 bytes

Cyan4973:
The sequence section starts with a number, which tells how many sequences are present in the section. If this number if `0`, the section ends immediately.

The number `0` can be represented using the 1 byte or the 2 bytes formats. That's because the 2-bytes format fully overlaps the 1 byte format.

However, when `0` is represented using the 2-bytes format, the decoder was expecting the sequence section to continue, and was looking for FSE tables instead of ending the section immediately, which is incorrect.
Fixed this behavior, in both the reference decoder and the educational decoder.

In practice, this scenario never happens, because the encoder always selects the 1-byte format to represent `0`, since this is more efficient. So this is mostly for completeness of the specification.

Completed the fix with a new golden sample for tests, a clarification of the specification, and a new decoder errata paragraph.

Credit to @ip7z for detecting this issue.


terrelln (approved):

ip7z:
There was `srcSize != 1` check in old code:
```
    if (!nbSeq) {
        *nbSeqPtr=0;
        RETURN_ERROR_IF(srcSize != 1, srcSize_wrong, "");
```
But new fixed code doesn't check `srcSize` anymore for `if (nbSeq == 0)`  case.
Do you want to allow `(srcSize > 2)` cases?

Cyan4973:
Right,
I wasn't completely sure if the `srcSize` field was providing the exact size of the Sequences section, or just the rest of the input buffer.
Thinking about it again, it must be the exact size of the Sequences section, otherwise the previous test would have failed.

Note that we don't necessarily guarantee that we will detect _all_ sources of corruption on the spot, since there are actually many classes of corruption which are effectively undetectable, except by the final checksum.

But when it's easy to detect and report, then we should do so. And I believe we are in such a case here. So let's merge a fix.

PR #3681, opened 2023-06-19
pzstd: use # directly for the test c++ snippet

kanavin:
\043 is not portable between shells: dash expands it to #, bash does not. As far as I see, using # works totally fine.


kanavin:
Make isn't cooperative either, 4.3 is fine with #, 4.2 wants backslash-#. Let's just read the test snippet from an external file.

kanavin:
Superseded by #3682

PR #3682, opened 2023-06-19
pzstd: use c++14 without conditions

kanavin:
Doing this check with a direct c++ snippet is prone to portability problems:

- \043 is not portable between shells: dash expands it to #, bash does not;

- using # directly works with make 4.3 but does not with make 4.2.

Let's just use the c++ version that covers both the code and the gtest.

### DIFF
--- a/contrib/pzstd/Makefile
+++ b/contrib/pzstd/Makefile
@@ -37,11 +37,8 @@ CFLAGS   += -Wno-deprecated-declarations
 PZSTD_INC  = -I$(ZSTDDIR) -I$(ZSTDDIR)/common -I$(PROGDIR) -I.
 GTEST_INC  = -isystem googletest/googletest/include
 
-# If default C++ version is older than C++11, explicitly set C++11, which is the
-# minimum required by the code.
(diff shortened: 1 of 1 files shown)


kanavin:
Another problem with the test as it's written is that when the compiler fails for any other reason than the condition set in the code snippet, that is still interpreted as 'does not default to c+11 or later' and the build continues, instead of being aborted.

terrelln:
clang-6.0.0 and gcc-6.1 made C++14 the default, and they added support for `-std=c++14` in clang-3.5 and gcc-4.9.0.

I think we can take this PR for now, because then we will work work with clang-3.5 and gcc-4.9.0. When, inevitably, gtest requires C++17 or newer, we should be able to just drop the `-std=c++14` entirely.

If someone needs pzstd to compile with an older compiler, I recommend:
1. Dropping pzstd support
2. If that is not possible, set `PZSTD_CXX_STD=-std=c++11` during `make` via a command line argument

https://gcc.godbolt.org/z/erzYbezjv

PR #3688, opened 2023-06-26
Hide ASM symbols on Apple platforms

nidhijaju:
ASM symbols were hidden on platforms using the ELF binary format, but
macOS/iOS uses the Mach-O binary format, not ELF. In addition, the name of
the assembler directive used is different.

This PR hides the ASM symbols on Apple platforms.

### DIFF
--- a/lib/common/portability_macros.h
+++ b/lib/common/portability_macros.h
@@ -68,6 +68,8 @@
 /* Mark the internal assembly functions as hidden  */
 #ifdef __ELF__
 # define ZSTD_HIDE_ASM_FUNCTION(func) .hidden func
+#elif defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


nidhijaju:
@felixhandte or @Cyan4973, would you be able to help review this? Thank you!

Cyan4973 (approved):

nidhijaju:
Hi, just wondering if there is anything else that needs to be done, or if this looks good, can you help merge this?
(This is blocking being able to build zstd in Chromium on Mac.)

Cyan4973:
We typically provide a bit of time for others to chime in and provide some additional feedback if need be,

but since there were none, and the patch is relatively simple, I think it's good to merge now.

nidhijaju:
Got it, thank you so much!

PR #3700, opened 2023-07-18
Save one byte on the frame epilogue

jacob-greenfield:
The regular frame epilogue is only 3 bytes; this reduces the destination capacity check from 4 bytes to 3 and changes `MEM_writeLE32` to `MEM_writeLE24`.


Cyan4973 (approved):
This is a nice little optimization!

terrelln (approved):

PR #3712, opened 2023-08-01
Fixed typo

alexsifivetw:
According to 
https://github.com/alexsifivetw/zstd/blob/dev/tests/Makefile#L343

I believe that it's a typo.


Cyan4973:
Thanks for the fix.
It looks good to me.

Please note though that signing the CLA is mandatory for a merge.

PR #3713, opened 2023-08-01
Enable RISC-V build & test

alexsifivetw:
Enable RISC-V build & test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,7 @@ jobs:
             make ppcbuild   V=1; make clean
             make armbuild   V=1; make clean
             make aarch64build V=1; make clean
+            make riscv64build V=1; make clean
             make -C tests test-legacy test-longmatch; make clean
             make -C lib libzstd-nomt; make clean
   # This step should only be run in a cron job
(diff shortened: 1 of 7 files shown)


alexsifivetw:
Seems like the docker image doesn’t run risc-v toolchain & qemu installation. Can any one help me?

https://github.com/facebook/zstd/blob/afcd5da6d4d6c469c536be9f2e72a390c196f49b/.circleci/images/primary/Dockerfile#L10-L16

Cyan4973:
Dependencies in general, and on `docker` images specifically, are to be avoided when practical.
I would rather prefer to rely on standard `apt` packages for CI.
I could get it to run `risc-v` tests properly in [another project](https://github.com/Cyan4973/xxHash/blob/dev/.github/workflows/ci.yml#L386).

alexsifivetw:
> I would rather prefer to rely on standard apt packages for CI.

I agree. We can now use apt to install `riscv-gcc-11`, but only gcc 13 and later versions support RISC-V RVV.
We use `wget` to get latest RISC-V toolchain & QEMU because of our plane for RVV-optimization for `zstd`. 

We could use `apt` to install toolchain, and use `__riscv_v` macro to protect RVV-optimized codes.
This means that the RVV-optimized code would only be enabled when the compiler supports RVV.

However, get RVV-supported toolchain & qemu via `apt` could take long time.

alexsifivetw:
Do you accept installing the toolchain via PPA?

Cyan4973:
> Do you accept installing the toolchain via PPA?

It depends on the PPA,
but in this particular case,
I'm not sure if there is such a need.

RISC-V `qemu` tests are working fine on Github CI using only regular `apt` in another project :
https://github.com/Cyan4973/xxHash/blob/dev/.github/workflows/ci.yml#L386

alexsifivetw:
Hi, I apologize for the late reply.
RISC-V is a young and modern platform. The RISC-V spec, intrinsics, toolchain ... etc, change frequently. Therefore, we need to use the nightly prebuilt toolchain and install it using `wget`.
Would you accept this `wget` method and update the recipe for Docker image?

Cyan4973:
Let's keep it around.
I'll look at it when I get the time.

PR #3720, opened 2023-08-13
Updated Makefiles for full MSYS2 and Cygwin installation and testing …

QBos07:
…support.

They are Linux like environments under Windows and have all the tools needed to support staged installation and testing.

Beware: this only affects the make build system.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -263,7 +263,7 @@ clean:
 # valgrind tests validated only for some posix platforms
 #----------------------------------------------------------------------------------
 UNAME := $(shell uname)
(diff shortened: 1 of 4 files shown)


tests/Makefile
```diff
@@ -287,11 +287,15 @@ ifneq (,$(filter MINGW% MSYS%,$(UNAME)))
   HOST_OS = MSYS
 endif
 
+ifneq (,$(filter CYGWIN%,$(UNAME)))
+  HOST_OS = CYGWIN
```
Cyan4973: minor: isn't cygwin a `posix` environment ?

QBos07: Kinda, but in this scenario it should be.

Cyan4973 (approved):

PR #3737, opened 2023-08-24
Fix & refactor Huffman repeat tables for dictionaries

terrelln:
The Huffman repeat mode checker assumed that the CTable was zeroed in the region `[maxSymbolValue + 1, 256)`. This assumption didn't hold for tables built in the dictionaries, because it didn't go through the same codepath.

Since this code was originally written, we added a header to the CTable that specifies the `tableLog`. Add `maxSymbolValue` to that header, and check that the table's `maxSymbolValue` is at least the block's `maxSymbolValue`.

This solution is cleaner because we write this header for every CTable we build, so it can't be missed in any code path.

This bug was introduced in https://github.com/facebook/zstd/pull/3731 and did not make it into any release. It would also only be triggered by dictionaries that Zstd would previously reject, so I don't expect any real world impact.

Credit to OSS-Fuzz


Cyan4973 (approved):
Good PR, operations on the header are more expressive with this patch.

PR #3754, opened 2023-09-12
fix version comments for updated actions/checkout

Cyan4973:
following #3749


Cyan4973:
Superseded by https://github.com/facebook/zstd/pull/3926

PR #3765, opened 2023-09-20
fix #3764

### LINKED ISSUE #3764: memory leak in function 'UTIL_mirrorSourceFilesDirectories'
The memory leak is reported by a static analyzer tool developed at [CAST](https://www.linkedin.com/company/cast-center)
memory is allocated [here](https://github.com/facebook/zstd/blob/1c007e64b85248b989aa7a9bea50897aae8fdb41/programs/util.c#L1262) may left w/o free in case of [exit](https://github.com/facebook/zstd/blob/1c007e64b85248b989aa7a9bea50897aae8fdb41/programs/util.c#L62) from macro ['CONTROL'](https://github.com/facebook/zstd/blob/1c007e64b85248b989aa7a9bea50897aae8fdb41/programs/util.c#L1269C13-L1269C13)

Cyan4973:
This answer a static analysis tool, in #3764,
which complains that an allocated buffer is not `free()` just before `exit()`.

In general, this requirement is not necessary, because invoking `exit()` terminates the process, and makes the OS reclaim all its memory.

I could articulate this additional requirement in a "not too heavy" way with the use of a new macro, `CONTROL_EXIT()`. 
But "not too heavy" is still a form of maintenance burden: whenever the code is modified, by adding, removing or changing some of these buffers, it requires some manual coordination with exit points, which is easy to let go wrong. 
Besides, I wouldn't be surprised if there were some more complex scenarios left, typically across multiple levels of functions, where a call to `exit()` is made while some other buffers, inaccessible from the function, are still allocated. Tackling such issues would require a very different approach, typically forbidding the use of `exit()`, which was meant to simplify code maintenance by reducing the nb and complexity of error paths. 
I question the need to make the code more complex to read and maintain, just to tackle some largely theoretical issue with no practical impact on target platforms.

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -53,12 +53,17 @@ extern "C" {
  * It's designed for failures that may happen rarely,
  * but we don't want to maintain a specific error code path for them,
  * such as a malloc() returning NULL for example.
- * Since it's always active, this macro can trigger side effects.
  */
-#define CONTROL(c)  {         \
+#define CONTROL(c)  CONTROL_EXIT(c, {})
+
+/* CONTROL_EXIT:
+ * Same as CONTROL, but can also run a last action before exit()
+ */
+#define CONTROL_EXIT(c, _action) { \
     if (!(c)) {               \
-        UTIL_DISPLAYLEVEL(1, "Error : %s, %i : %s",  \
+        UTIL_DISPLAYLEVEL(1, "Error : %s, %i : %s not respected",  \
                           __FILE__, __LINE__, #c);   \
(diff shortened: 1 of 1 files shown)


terrelln (approved):

Cyan4973:
It seems we have some issues with the `x32` test these days.
```
../programs/zstd: Exec format error
```

Note sure what's going on.
This test error is unrelated to this PR, since it's also failing for other PRs.
But it's an annoying red signal in our CI, and will likely deserve a fix.

PR #3768, opened 2023-09-21
Fix conditional compilation when targeting fuzzers

GregTho:
ZSTD_assertValidSequence is only called when both
FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION and
FUZZING_ASSERT_VALID_SEQUENCE are defined. Match those conditions for the function's definition. Additionally, ZSTD_assertValidSequence only calls ZSTD_dictionaryIsActive when DEBUGLEVEL >= 1, so only define it under that condition.


terrelln:
Superseded by #3770.

Issue #3762, opened 2023-09-18
Possible performance regressions on some CPUs after #3449 (C fast loops)


iksaif:
**Describe the bug**
It seems that https://github.com/facebook/zstd/pull/3449 introduced a performance regression on modern CPUs, this is particularly problematic on aarch64/arm64 since those don't have an ASM function and will always use the fast loop functions.

**To Reproduce**
```
wget https://github.com/facebook/zstd/releases/download/v1.1.3/github_users_sample_set.tar.zst
zstd -d github_users_sample_set.tar.zst
tar xf github_users_sample_set.tar
wget https://mattmahoney.net/dc/silesia.zip
unzip silesia.zip
tar -cf silesia.tar dickens mozilla mr nci ooffice osdb reymont samba sao webster x-ray xml

CFLAGS="-O2 -march=native" make -j8; ./zstd -b1 -r github ; ./zstd -b1e1 --compress-literals --zstd=tlen=131072 silesia.tar
```

On Darwin: (Apple clang version 14.0.3 (clang-1403.0.22.14.1))
On Linux: (gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0)

**Expected behavior**
Performance should not degrade between v1.5.2 and v1.5.4

**Screenshots and charts**

* b1: `zstd -b1 -r github`
* b2: `zstd -b1e1 --compress-literals --zstd=tlen=131072 silesia.tar`

* 🟢 - best performances for this machine
* 🔴 - worst performances for this machine

Machine | Version | github | silesia |
-- | -- | -- | --
m1 | v1.5.2 | 🟢 529.5 MB/s | 🔴 722.5 MB/s
m1 | v1.5.4 | 🔴 412.3 MB/s | 🟢 723.4 MB/s
m1 | v1.5.4 + disableFast | 🟢 529.4 MB/s | 🟢 722.4 MB/s
m1 | dev | 🔴 410.6 MB/s | 🟢 722.7 MB/s
c7g Graviton 3 | 51b123d5f732955b2079dffe00b5f197dcc8c206 | 🟢 483.1 MB/s | 🟢 1128.2 MB/s
c7g | v1.5.2 | 🟢 436.1 MB/s | 🟢 1122.5 MB/s
c7g | v1.5.4 | 🔴 349.6 MB/s | 🔴 703.7 MB/s
c7g | v1.5.4 + disableFast | 🟢 436.7 MB/s | 🟢 1125.5 MB/s
[...]

PR #3778, opened 2023-09-28
decompress: add -DZSTD_DISABLE_FAST_C_LOOP to disable fast C loops

iksaif:
https://github.com/facebook/zstd/issues/3762 seems to show that it doesn't perform as well as we though it would in many cases. It makes sense to at least allow users to disable them at buildtime and runtime.


terrelln:
`HUF_DISABLE_FAST_DECODE` does this, it is documented in `lib/decompress/README.md`

PR #3786, opened 2023-10-08
Remove FlexArray pattern from ZSTDMT

Cyan4973:
This is directly inspired by @ebiggers 's #3785,
mentioning that `C90`'s flexArray pattern is now becoming incompatible with modern versions of `ubsan`.

This PR removes 2 out of 3 offending patterns, both located within `ZSTDMT_`,
and replaces them by a more traditional external allocation pattern,
which is well controlled within their respective Pool
(user code of these abstractions remains completely unchanged).

Another place where the flexArray pattern is used is in `fse`, 
it will likely require some dedicated attention and another PR for a fix.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -350,15 +350,16 @@ typedef struct {
     int totalCCtx;
     int availCCtx;
     ZSTD_customMem cMem;
-    ZSTD_CCtx* cctx[1];   /* variable size */
+    ZSTD_CCtx** cctxs;
 } ZSTDMT_CCtxPool;
 
-/* note : all CCtx borrowed from the pool should be released back to the pool _before_ freeing the pool */
+/* note : all CCtx borrowed from the pool must be reverted back to the pool _before_ freeing the pool */
 static void ZSTDMT_freeCCtxPool(ZSTDMT_CCtxPool* pool)
 {
     int cid;
     for (cid=0; cid<pool->totalCCtx; cid++)
-        ZSTD_freeCCtx(pool->cctx[cid]);  /* note : compatible with free on NULL */
+        ZSTD_freeCCtx(pool->cctxs[cid]);  /* note : compatible with free on NULL */
+    ZSTD_customFree(pool->cctxs, pool->cMem);
     ZSTD_pthread_mutex_destroy(&pool->poolMutex);
     ZSTD_customFree(pool, pool->cMem);
 }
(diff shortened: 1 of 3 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -373,14 +374,19 @@ static ZSTDMT_CCtxPool* ZSTDMT_createCCtxPool(int nbWorkers,
     assert(nbWorkers > 0);
     if (!cctxPool) return NULL;
     if (ZSTD_pthread_mutex_init(&cctxPool->poolMutex, NULL)) {
+        ZSTDMT_freeCCtxPool(cctxPool);
+        return NULL;
+    }
+    cctxPool->totalCCtx = nbWorkers;
+    cctxPool->cctxs = (ZSTD_CCtx**)ZSTD_customCalloc(nbWorkers * sizeof(ZSTD_CCtx*), cMem);
+    if (!cctxPool->cctxs) {
         ZSTD_customFree(cctxPool, cMem);
```
ebiggers: Missing call to ZSTD_pthread_mutex_destroy

lib/compress/zstdmt_compress.c
```diff
@@ -373,14 +374,19 @@ static ZSTDMT_CCtxPool* ZSTDMT_createCCtxPool(int nbWorkers,
     assert(nbWorkers > 0);
```
ebiggers, line 381: `(nbWorkers-1)*sizeof(ZSTD_CCtx*)` isn't needed in the line above anymore

lib/compress/zstdmt_compress.c
```diff
@@ -404,14 +410,15 @@ static size_t ZSTDMT_sizeof_CCtxPool(ZSTDMT_CCtxPool* cctxPool)
     {   unsigned const nbWorkers = cctxPool->totalCCtx;
         size_t const poolSize = sizeof(*cctxPool)
                                 + (nbWorkers-1) * sizeof(ZSTD_CCtx*);
```
ebiggers: `(nbWorkers-1) * sizeof(ZSTD_CCtx*)` isn't needed anymore

ebiggers (commented):
Thanks!  By the way, the flex array in `fse_decompress.c` was introduced by commit 634bfd339f4c, so you could consider reverting that.

lib/compress/zstdmt_compress.c
Cyan4973: Thanks,
indeed, I actually meant `ZSTDMT_freeCCtxPool()` here

terrelln (approved):

PR #3794, opened 2023-10-18
Linking fix for Android with min SDK ver less than 24

vok1980:


Cyan4973:
reference for `fseeko` : 
https://www.man7.org/linux/man-pages/man3/fseeko.3.html

Cyan4973:
It seems it would be possible to test the availability of `fseeko()` directly within `cmake`:

```cmake
include(CheckFunctionExists)
check_function_exists(fseeko HAVE_FSEEKO)

if(NOT HAVE_FSEEKO)
  add_definitions(-DHAVE_NOT_FSEEKO)
endif()
```

Such a variant would be more generic than just testing for specific Android versions within `cmake`.

Cyan4973:
Alternatively, it seems it would be possible to detect Android version directly within the `C` preprocessor, using `#if defined(__ANDROID_API__) && __ANDROID_API__ < 24`.
Such a method would have the advantage of being independent of the build system, therefore not being restricted to active `cmake` participation.

Victor-C-Zhang:
@Cyan4973 I wonder if this is still relevant. The minimum supported CMake version is now 3.10, which was released in late November 2018. This is after Android API level 28 was released and almost 3 years after the last release without `fseeko`. If the issue is `fseeko` availability, that's a different story, but this is available in libc and only old versions of Android's proprietary `bionic` libc lack this function. Maybe it's reasonable to expect new developers to develop for newer models. After all, <2% of Android devices worldwide still run Android 6.

Cyan4973:
I don't understand the comment about the relation between this PR and the `cmake` version ?
I don't see anything in this PR related to `cmake` version.

The main comment made on this PR was that `cmake` does not seem necessary, and the logic could be entirely source code driven, using `__ANDROID_API__` build macro.

The issue seems to be `fseeko` availability, which seems to be lacking for Android below version 24. Version 24 was released in 2015, so that's almost 10 years old. It's indeed reasonable to classify such operating systems as "old".

Now, another criteria is patch complexity. If the patch requires multiple changes in multiple places, its complexity is high, so it must be really worth it, and supporting some old OS versions > 10 years old may not reach that bar.
On the other hand, if it's trivial (one `#define` change for example), then I'm okay supporting old / rare systems. It's part of the experience of having universal access to `zstd` format on most computing devices.

This PR is pretty close to being simple. It requires 3 `#if` condition changes , because `fseeko` is defined in 3 different places. So not perfect, but really close to "simple", because it's the same change.
The biggest issue is how to test such a modification.

OP wasn't responsive to earlier comments, and did not even complete CLA, so I guess we can consider this PR dead, and open a new one.
I think the new PR could be entirely source code driven, without any needed modification for `cmake`.

Victor-C-Zhang:
A fix was landed in #4229. Closing this PR as resolved. Feel free to re-open as necessary.

PR #3806, opened 2023-10-31
Correct FSE probability bit consumption in specification

elasota:
The current verbiage states that a decoder "must" read the number of bits required for the maximum representable value but it actually may read 1 bit less than that, per the second bullet point.

It does not appear to overconsume bits when it needs fewer than that, which would have implications for decoder behavior if it did.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -1083,7 +1083,8 @@ It depends on :
   Presuming an `Accuracy_Log` of 8,
   and presuming 100 probabilities points have already been distributed,
   the decoder may read any value from `0` to `256 - 100 + 1 == 157` (inclusive).
-  Therefore, it may read up to `log2sup(157) == 8` bits.
+  Therefore, it may read up to `log2sup(157) == 8` bits, where `log2sup(N)`
+  is the smallest integer `T` that satisfies `(1 << T) > N`.
 
 - Value decoded : small values use 1 less bit :
   __example__ :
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):
Yes, I think it's a more accurate description.

Cyan4973:
Note that, for this PR to get merged, the CLA is compulsory.

elasota:
Added a definition of the "log2sup" function, which is currently missing.  I have further changes to propose to make this section more descriptive instead of example-based like the rest of the spec is, but I'll do that in a separate PR.

doc/zstd_compression_format.md
```diff
@@ -1083,7 +1083,8 @@ It depends on :
   Presuming an `Accuracy_Log` of 8,
   and presuming 100 probabilities points have already been distributed,
   the decoder may read any value from `0` to `256 - 100 + 1 == 157` (inclusive).
-  Therefore, it must read `log2sup(157) == 8` bits.
+  Therefore, it may read up to `log2sup(157) == 8` bits, where `log2sup(N)`
+  is the smallest integer `T` that satisfies `(1 << T) > N`.
```
Cyan4973, line 1087: I believe it's actually `(1 << T) >= N` (note the `>=`)

elasota: I believe it is correct as `>` because of the power-of-2 case and N being inclusive.  e.g. suppose `N` is 256, in that case it must be able to read the binary number 100000000 which is 9 bits.  If it were `>=` then it would resolve as T=8 which is wrong.

Cyan4973: Yes you are right

It depends on how is `N` defined.
If `N` is the alphabet size, then a size of `256` implies that the highest possible value is `255` since the alphabet starts at `0`. In which case, 8-bits are enough.
But if `N` is the highest possible value, then indeed, 9 bits are required to encode `256`.

In the example above, `N` takes the place of `157`, which is defined as the highest possible value inclusive.
So it's a `<`.

Cyan4973 (approved):

PR #3807, opened 2023-10-31
[cmake] Require CMake version 3.5 or newer

gruenich:
More recent versions of CMake emit the following warning:
```
CMake Deprecation Warning at cmake/CMakeLists.txt:10 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -7,7 +7,7 @@
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
 
-cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 # As of 2018-12-26 ZSTD has been validated to build with cmake version 3.13.2 new policies.
 # Set and use the newest cmake policies that are validated to work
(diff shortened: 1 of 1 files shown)


Cyan4973:
I'm fine with the prospect of increasing the minimum `cmake` version to `3.5`, as it's a version supported by Ubuntu Xenial, which is probably the oldest Ubuntu version still worth supporting today.

It's probably worth going through the `cmake` scripts to see if there are more changes warranted.
Particularly, I wouldn't be surprised if some code paths meant to deal with older versions of `cmake` could be simplified.

gruenich:
I only found one code path for CMake < 3.0. The next one would be for CMake < 3.7, I did not touch this.

Cyan4973:
It seems there is another one on Line 16 ?

gruenich:
Oh sorry, I missed the second case. Is now part of the second commit.

Cyan4973 (approved):

gruenich:
Any reason holding back merging of this pull request?

Cyan4973:
Nope, that's fine,
there was enough time given for additional reviews.

PR #3811, opened 2023-11-07
Export a `zstd::libzstd` CMake target if only static or dynamic linkage is specified.

teo-tsirpanis:
With this PR, zstd can be referenced from CMake the same way, regardless of whether it was compiled with static or dynamic linkage.


Cyan4973:
Is there a test checking this property ?

teo-tsirpanis:
I have tested it locally with success, by patching the vcpkg port for zstd.

Cyan4973:
We probably want to add something in CI,
otherwise this property might be lost in a future update.

Cyan4973 (approved):

teo-tsirpanis:
PR updated to use the same approach as lz4/lz4#1372. By defining the `libzstd` target inside the main CMake project, exporting it gets validated by CMake, making it less error prone.

Cyan4973:
`freebsd-12-4` is no longer available on `Cirrus-ci`.
Not sure why it wants to run this VM version. We have already migrated this test to `FreeBSD-14-0` : 
https://github.com/facebook/zstd/blob/dev/.cirrus.yml .
Maybe the PR just needs to be rebased.

Cyan4973 (approved):

PR #3813, opened 2023-11-09
Clarify that a stream containing too many Huffman weights is invalid

elasota:
The Huffman weight decoding section specifies how the maximum decoded size is to be calculated but doesn't specify what should happen if that size is exceeded (i.e. if this is a failure condition, or if surplus data should be ignored, or something else).

This clarifies that the expected behavior is to reject the data, matching the behavior of `FSE_decompress_wksp_bmi2`.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -1353,6 +1353,9 @@ If updating state after decoding a symbol would require more bits than
 remain in the stream, it is assumed that extra bits are 0.  Then,
 symbols for each of the final states are decoded and the process is complete.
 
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

elasota:
To be clear, part of the reason for this is to implicitly clarify that a Huffman table where all entries contain the same weight (which would produce an FSE bitstream containing infinitely many 0-sized symbols) is invalid and a compliant encoder must use a different block type instead in such a situation.

(Such a block will always be larger than a raw literals block because all Huffman codes will be 8 bits so there is no reason for an encoder to produce such a block.)

PR #3816, opened 2023-11-09
Fix state table formatting

elasota:
Markdown only allows one header row in a table, this moves the errant row into the table so it is formatted correctly.


Cyan4973:
I can't seem to remember what `state value` is supposed to mean / explain here...

elasota:
It looks like it's supposed to be the table positions of a value of probability 5 starting from position 1.  Should it be renamed "table position" or something like that?

Cyan4973 (approved):

PR #3817, opened 2023-11-13
Clarify that probability tables must not contain non-zero probabilities for invalid values

elasota:
The current specification indicates the valid range of values decoded from the FSE entropy coder, but doesn't specify what happens if the probability table declares a non-zero probability for an invalid value but never produces that value.

This adds a specification that a probability table must not contain out-of-range values, even if they are unused, reflecting the behavior of `ZSTD_buildSeqTable`.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -1124,6 +1124,9 @@ When last symbol reaches cumulated total of `1 << Accuracy_Log`,
 decoding is complete.
 If the last symbol makes cumulated total go above `1 << Accuracy_Log`,
 distribution is considered corrupted.
+If this process results in a non-zero probability for a value outside of the
+valid range of values that the FSE table is defined for, even if that value is
+not used, then the data is considered corrupted.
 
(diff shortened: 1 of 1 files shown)


terrelln:
I think this is a reasonable clarification, CC @Cyan4973

This certainly makes the lives of the decoder easier, because if they can validate that all symbols they decode are guaranteed to be in the valid range due to table construction, they can omit bounds checks. So we definitely don't want to make this a valid frame.

elasota:
> if they can validate that all symbols they decode are guaranteed to be in the valid range due to table construction, they can omit bounds checks.

It also reduces the required temporary storage to construct the table since otherwise the table may contain as many as `1<<accuracyLog` unique values (and due to the behavior of less-than-one probabilities, there has to be enough temporary storage for all of the non-zero probability values before the table is constructed).

Specifying this means a decoder only has to track as many unique probabilities as the allowed range.

Cyan4973 (approved):
This clarification looks good to me

Issue #3819, opened 2023-11-13
Disable auto vectorization of xxhash64, when AVX512 is present. 


TocarIP:
AVX512 adds support for `VPMULLQ` instructions, which makes is possible to vectorize `XXH64_update`.  Here is a minimal reproducer where clang already vectrozies it with avx512 enabled: https://godbolt.org/z/cnM3986dx . Unfortunately some architectures (e.g skylake) have a very significant latency (uops.info report 15+ cycles), so this makes `xxhash` ~2x slower. Trivial benchmark shows `BM_hash/16M 1.29ms ± 3%` vs `BM_hash/16M 2.74ms ± 4%`. We already disable vectorization for xxhash32 if sse4 is present, so we should probably do the same for xxhash64 and avx512.

Cyan4973:
> We already disable vectorization for xxhash32 if sse4 is present, so we should probably do the same for xxhash64 and avx512.

Indeed,
this seems like the same issue,
and we should probably use a similar solution.

TocarIP:
Someone asked me about AMD offline so might as well post here. On genoa (Zen4) AVX512 is actually ~30% faster  ( 1135871 ns/op vs 885935 ns/op).

Cyan4973:
OK, so it makes the situation a bit less clear,
since `avx512` is now sometimes beneficial, sometimes detrimental.
It probably opens the door towards offering a user choice on this topic.

However, it doesn't remove the question of "what's a good default",
and if I read the situation correctly, it seems that disabling `avx512` for `xxh64` remains a reasonable default at this point in time.

TocarIP:
Regression on Intel is bigger than gain on AMD, so disabled by default makes sense. Using the same `XXH_ENABLE_AUTOVECTORIZE` to allow opt-in is probably the best.

Cyan4973:
https://github.com/Cyan4973/xxHash/pull/924

PR #3820, opened 2023-11-13
update xxhash library to v0.8.2

Cyan4973:
Complete change list : https://github.com/Cyan4973/xxHash/releases/tag/v0.8.2

This can also be seen as a preparation task before taking care of #3819


Cyan4973:
What does mean:

```
lib/common/xxhash.c: License not found!
lib/common/xxhash.h: License not found!
Fail!
```

The license text is explicitly given in these source files !

Issue #3823, opened 2023-11-17
Spec cleanup: Should fixup behavior when repeat1-1==0 be specified or changed to an error?


elasota:
Doing some more edge case checks on the specification...

In `ZSTD_decodeSequence` there is this code:
```
size_t temp = (offset==3) ? seqState->prevOffset[0] - 1 : seqState->prevOffset[offset];
temp += !temp;   /* 0 is not valid; input is corrupted; force offset to 1 */
```

In the case where `literals_length==0` and `offset_value==3` and `Repeated_Offset1==1` this will insert `1` at the start of the repeated offset list and move the rest down.

Right now the comment says it is invalid, but it is silently processed as if it is valid.  So, should this behavior be put in the spec?  If not, should the code be rejecting this data?

Cyan4973:
This is invalid.
If the code reaches that part, it means data is corrupted, because it's trying to generate an `offset = 0`.

This is considered a potential attack vector, that's why the offset is converted into a `1`: it ensures that the underlying memory content of the destination buffer is overwritten, hence it prevents scenario where such content could be accessed afterwards.

Detecting this issue is good ground to generate an error.
Problem is, inserting an error detection signal at this place in the code results (or at least used to result) in substantial performance drawbacks, because it's an extremely hot section of the decoding loop.
Therefore, we rather rely on the checksum to tell us so (assuming the issue is accidental, following a transmission or storage error).

terrelln:
This is one of the cases where we don't guarantee that we reject all possible invalid frames. Since we guarantee safety on invalid frames, we must not actually copy from offset 0. In this case we decide to copy from offset 1. So given this invalid frame, for performance reasons, we decide to ignore the corruption.

Other implementations are allowed to reject this frame.

Cyan4973:
spec updated

PR #3824, opened 2023-11-19
Specify offset 0 as invalid and specify required fixup behavior

elasota:
Addresses #3823

Specifies that this case is invalid, but also specifies the fixup behavior.

There have historically been some security issues (not in this project) caused by implementation divergence so I think it is best to specify it, as it is the only case I can find where Zstd doesn't error on invalid data.

This would limit compliant implementations to 2 allowed behaviors: Evaluate as 1, or raise an error.

### DIFF
--- a/doc/decompressor_accepted_invalid_data.md
+++ b/doc/decompressor_accepted_invalid_data.md
@@ -1,0 +1,14 @@
+Decompressor Accepted Invalid Data
+==================================
+
+This document describes the behavior of the reference decompressor in cases
+where it accepts an invalid frame instead of reporting an error.
+
+Zero offsets converted to 1
+---------------------------
+If a sequence is decoded with `literals_length = 0` and `offset_value = 3`
+while `Repeated_Offset_1 = 1`, the computed offset will be `0`, which is
+invalid.
+
+The reference decompressor will process this case as if the computed
+offset was `1`, including inserting `1` into the repeated offset list.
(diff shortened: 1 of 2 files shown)


terrelln:
IMO this doesn't belong in the Zstandard format specification, since it is specifying what to do on an invalid frame. All the format description cares about is specifying what a valid frame is, and how to decode it.

We have a second document [decoder_errata.md](https://github.com/facebook/zstd/blob/dev/doc/decompressor_errata.md) which captures cases where the standard decoder accidentally rejected a valid frame.

We could add another document under `doc/` to specify how exactly the standard decoder handles particular types of invalid frames. Particularly cases where the decoder doesn't reject invalid frames, and what it does instead. I'm not really sure what to call it, maybe `decoder_invalid_frame_behavior.md`? But if you have a better name use that instead.

Cyan4973 (changes requested):
The first part of the sentence looks good to me: 
> In the final case, if `Repeated_Offset1 - 1_byte` evaluates to 0, then the
data is considered corrupted. 

The second part though mandates a decoder behavior to deal with such a case, which I believe should not be part of a format specification. I could imagine multiple different ways to deal with the same situation, and each one may or may not make more sense depending on underlying architecture or system. We want to let opened this flexibility to implementers.

Cyan4973 (approved):

PR #3839, opened 2023-11-27
Move offload API params into ZSTD_CCtx_params

embg:
Currently, params for the External Sequence Producer (ESP) API are stored directly in the compression context. This PR moves those params into the  `ZSTD_CCtx_params` struct. I recommend reviewers to start by looking at the changes in `zstd_compress_internal.h`, since everything else flows from there.

The immediate motivation for this refactor is to lay the groundwork for adding a `ZSTD_CCtxParams_registerSequenceProducer()` function to the public API. Such a function would make `ZSTD_estimateCCtxSize_usingCCtxParams()` compatible with ESPs, which is a blocker for supporting compression offload in the Linux kernel.

A side benefit of this refactor is that we can piggy-back on the existing infrastructure around `ZSTD_CCtx_params`. For example, we no longer have to explicitly clear the ESP params in `ZSTD_reset_CCtx()`. This reduces the maintenance cost of the ESP API.

As part of this refactor, I included a small change to `zstd.h` (see https://github.com/facebook/zstd/pull/3839/commits/809c7eb6bff1934745b425437d2116d9c0dbe0df). Theoretically, this could be a breaking change for external code which uses the typedef `ZSTD_sequenceProducer_F` (which is part of the unstable API). I am confident that no external code depends on this typedef -- it is simply a helper for another declaration. External code should only depend on that declaration, not on the helper typedef.

To be extra safe, I checked [Intel's plugin library](https://github.com/intel/QAT-ZSTD-Plugin/tree/main) and confirmed that it does not use the `ZSTD_sequenceProducer_F` typedef directly.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -1365,7 +1365,6 @@ size_t ZSTD_CCtx_reset(ZSTD_CCtx* cctx, ZSTD_ResetDirective reset)
         RETURN_ERROR_IF(cctx->streamStage != zcss_init, stage_wrong,
                         "Reset parameters is only possible during init stage.");
         ZSTD_clearAllDicts(cctx);
-        ZSTD_memset(&cctx->externalMatchCtx, 0, sizeof(cctx->externalMatchCtx));
         return ZSTD_CCtxParams_reset(&cctx->requestedParams);
     }
     return 0;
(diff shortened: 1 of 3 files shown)


lib/compress/zstd_compress.c
```diff
@@ -1365,7 +1365,6 @@ size_t ZSTD_CCtx_reset(ZSTD_CCtx* cctx, ZSTD_ResetDirective reset)
         RETURN_ERROR_IF(cctx->streamStage != zcss_init, stage_wrong,
                         "Reset parameters is only possible during init stage.");
         ZSTD_clearAllDicts(cctx);
-        ZSTD_memset(&cctx->externalMatchCtx, 0, sizeof(cctx->externalMatchCtx));
```
Cyan4973, line 1368: 👍

Cyan4973 (approved):

terrelln (approved):

PR #3840, opened 2023-11-28
lib/decompress: check for reserved bit corruption in zstd

### LINKED ISSUE #3821: No check if Reserved of Symbol_Compression_Modes is 0
**Describe the bug**

According to the https://datatracker.ietf.org/doc/html/rfc8878#name-symbol_compression_modes description, Reserved should have to be 0, in which case the following file should not be able to be extracted

[test-diverge-855c1b40.zst.zip](https://github.com/facebook/zstd/files/13364216/test-diverge-855c1b40.zst.zip)

Using educational_decoder gives the expected error, but using the zstd binary does not.

use educational_decoder

```bash
./harness test-diverge-855c1b40.zst test-diverge-855c1b40
Error: Corruption detected while decompressing
```

use zstd

```bash
zstd -d test-diverge-855c1b40.zst
test-diverge-855c1b40.zst: 102400 bytes
```

**To Reproduce**
1. Downloads data '...'
2. Run '...' with flags '...'
3. Scroll up on the log to '....'

**Expected behavior**

**Screenshots and charts**

 - OS: [e.g. Mac] Mac
 - Version [e.g. 22] *** Zstandard CLI (64-bit) v1.5.5, by Yann Collet ***
 - Compiler [e.g. gcc]
 - Flags [e.g. O2]
 - Other relevant hardware specs [e.g. Dual-core]
 - Build system [e.g. Makefile]

**Additional context**

aimuz:
The patch adds a validation to ensure that the last field, which is
reserved, must be all-zeroes in ZSTD_decodeSeqHeaders. This prevents
potential corruption from going undetected.

Fixes an issue where corrupted input could lead to undefined behavior
due to improper validation of reserved bits.

Fixes #3821 

Signed-off-by: aimuz <mr.imuz@gmail.com>

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -607,6 +607,7 @@ size_t ZSTD_decodeSeqHeaders(ZSTD_DCtx* dctx, int* nbSeqPtr,
 
     /* FSE table descriptors */
     RETURN_ERROR_IF(ip+1 > iend, srcSize_wrong, ""); /* minimum possible size: 1 byte for symbol encoding types */
+    RETURN_ERROR_IF(*ip & 3, corruption_detected, ""); /* The last field, Reserved, must be all-zeroes. */
     {   symbolEncodingType_e const LLtype = (symbolEncodingType_e)(*ip >> 6);
         symbolEncodingType_e const OFtype = (symbolEncodingType_e)((*ip >> 4) & 3);
         symbolEncodingType_e const MLtype = (symbolEncodingType_e)((*ip >> 2) & 3);


Cyan4973:
The specification is pretty clear: these 2 bits should be all zeroes (https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#sequences-section).

Now the decoder tends to be more permissive, meaning in this case that it will ignore these 2 bits, rather than actively checking that they are zeroes. We reserve the right to do so whenever detection or action on some minor details which do not impact the rest of the bitstream may result in too large a complexity or performance cost.

But that's not the case here. Detection is pretty straightforward, and I wouldn't expect performance to be impacted in any way.

Our only risk remaining is to break some existing custom encoder that would not respect the specification regarding these 2 bits.
That being said, since the specification is very clear on the topic, and explicitly states that these 2 bits should be zeroes, and since the reference encoder (which the majority of wrappers employ) also respects it, I believe this risk is very low and would have a fairly small blast impact if any.

So I think this update is good to go.

Cyan4973 (approved):

aimuz:
🎉

PR #3845, opened 2023-12-13
Update FreeBSD CI: drop 12.4 (nearly EOL)

emaste:
12.4 is EOL as of the end of December 2023, and pkg installation will start failing some time after that so remove those jobs now.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,7 +3,6 @@ task:
   freebsd_instance:
     matrix:
       image_family: freebsd-13-2
-      image_family: freebsd-12-4
   install_script: pkg install -y gmake coreutils
   script: |
     MOREFLAGS="-Werror" gmake -j all


Cyan4973:
I'm favorable to the change, 
though it seems that the update surfaces new issues that need fixing too (we require CI 100% correct at all time for merge).

emaste:
Yes, it appears that this uncovers an issue with FreeBSD 14 -- let me force push a new change that only removes the soon-to-be-EOL 12.4, and follow up later on 14.0.

emaste:
I'm not sure what is failing in FreeBSD 14 - the log ends with:

```
Byte comparison: PASSED
Running roundtrip.c: PASSED
(Skipping Emscripten test)
gmake[1]: Leaving directory '/tmp/cirrus-ci-build/tests'

Exit status: 2
```


Will look locally

Cyan4973:
> I'm not sure what is failing in FreeBSD 14 - the log ends with:

If I do remember correctly, it's in the `zlib` wrapper,
something like a warning get triggered there.

It's possibly a consequence of a compiler update rather than FreeBSD proper,
in which case it might be reproducible on some other posix platform,
but who knows, it could also be related to the header version shipped with the OS.

Cyan4973 (approved):

emaste:
Oh - GitHub shows a truncated log; Cirrus-CI has the full log. The issue is many cases of `error: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Werror,-Wdeprecated-non-prototype]`

It is indeed in the zlib wrapper, which has a bunch of K&R function definitions (!)

I suppose this is just the first time this is being built with Clang 16?

zlib 1.3 includes this bullet:
- Building using K&R (pre-ANSI) function definitions is no longer supported.

I'm not sure how zlibwrapper is managed in zstd; is it brought in as vendor code? Would you take a patch to modify it?

emaste:
Trying https://github.com/facebook/zstd/commit/fdac0c11927235b82849a775580a6e35cc505249 now.

emaste:
There were a few more cases but https://github.com/emaste/zstd/commit/71bd9a703636433ce08d92718f91e3bcd0fe2383 was successful on FreeBSD 14. I can post it and a change to add FreeBSD 14 CI tomorrow.

PR #3847, opened 2023-12-16
Fix a nullptr dereference in ZSTD_createCDict_advanced2()

michoecho:
If the relevant allocation returns NULL, ZSTD_createCDict_advanced_internal() will return NULL. But ZSTD_createCDict_advanced2() doesn't check for this and attempts to use the returned pointer anyway, which leads to a segfault.


Cyan4973 (approved):

Cyan4973:
Great find @michoecho !

michoecho:
@Cyan4973 Pinging for merge.

PR #3850, opened 2023-12-18
cli: better errors on argument parsing

KapJI:
Show which argument caused the error.

`usage_advanced` and `badusage` are the only 2 functions which don't follow camel case, rename them.


Bravonome2022:
Api.31

Cyan4973 (approved):
I like it,
this is a welcome quality-of-life improvement.

PR #3854, opened 2023-12-28
Make offload API compatible with static CCtx

embg:
Currently our CCtx size estimation functions don't account for memory used by the block-level external sequence producer API, so it is impossible to use that API together with `ZSTD_initStaticCCtx()` (which requires a correctly-sized buffer). This is a problem in particular for the Linux kernel, which uses `ZSTD_initStaticCCtx()` and also wants to use the offload API.

This PR adds a function `ZSTD_CCtxParams_registerSequenceProducer()` to the public API, which makes it possible to accurately estimate the size of a CCtx for the offload scenario. Simply call the newly-added function on a `ZSTD_CCtx_params` object and then compute a size estimate with `ZSTD_estimateCCtxSize_usingCCtxParams()`.

I added a unit test to demonstrate that compression offload works properly with a static CCtx.

Note: this is a follow-up to https://github.com/facebook/zstd/pull/3839

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -7088,6 +7088,7 @@ void ZSTD_registerSequenceProducer(
     void* extSeqProdState,
     ZSTD_sequenceProducer_F extSeqProdFunc
 ) {
+    assert(zc != NULL);
     ZSTD_CCtxParams_registerSequenceProducer(
         &zc->requestedParams, extSeqProdState, extSeqProdFunc
     );
@@ -7098,6 +7099,7 @@ void ZSTD_CCtxParams_registerSequenceProducer(
   void* extSeqProdState,
   ZSTD_sequenceProducer_F extSeqProdFunc
 ) {
+    assert(params != NULL);
     if (extSeqProdFunc != NULL) {
         params->extSeqProdFunc = extSeqProdFunc;
         params->extSeqProdState = extSeqProdState;
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -7084,14 +7084,25 @@ ZSTD_parameters ZSTD_getParams(int compressionLevel, unsigned long long srcSizeH
 }
 
 void ZSTD_registerSequenceProducer(
-    ZSTD_CCtx* zc, void* extSeqProdState,
+    ZSTD_CCtx* zc,
+    void* extSeqProdState,
     ZSTD_sequenceProducer_F extSeqProdFunc
 ) {
+    ZSTD_CCtxParams_registerSequenceProducer(
+        &zc->requestedParams, extSeqProdState, extSeqProdFunc
+    );
+}
+
+void ZSTD_CCtxParams_registerSequenceProducer(
+  ZSTD_CCtx_params* params,
+  void* extSeqProdState,
+  ZSTD_sequenceProducer_F extSeqProdFunc
+) {
     if (extSeqProdFunc != NULL) {
-        zc->requestedParams.extSeqProdFunc = extSeqProdFunc;
-        zc->requestedParams.extSeqProdState = extSeqProdState;
+        params->extSeqProdFunc = extSeqProdFunc;
+        params->extSeqProdState = extSeqProdState;
     } else {
-        zc->requestedParams.extSeqProdFunc = NULL;
-        zc->requestedParams.extSeqProdState = NULL;
+        params->extSeqProdFunc = NULL;
+        params->extSeqProdState = NULL;
     }
 }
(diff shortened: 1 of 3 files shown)


lib/zstd.h
```diff
@@ -2824,6 +2821,21 @@ ZSTD_registerSequenceProducer(
...
+/*! ZSTD_CCtxParams_registerSequenceProducer() :
+ * Same as ZSTD_registerSequenceProducer(), but operates on ZSTD_CCtx_params.
+ * This is used for accurate size estimation with ZSTD_estimateCCtxSize_usingCCtxParams(),
+ * which is needed when creating a ZSTD_CCtx with ZSTD_initStaticCCtx().
+ *
+ * If you are using the external sequence producer API in a scenario where ZSTD_initStaticCCtx()
+ * is required, then this function is for you. Otherwise, you probably don't need it.
+ *
+ * See tests/zstreamtest.c for example usage. */
+void ZSTD_CCtxParams_registerSequenceProducer(
```
Cyan4973: missing the `ZSTDLIB_STATIC_API` prefix.

tests/zstreamtest.c
```diff
@@ -2356,6 +2356,58 @@ static int basicUnitTests(U32 seed, double compressibility, int bigTests)
...
+        size_t const checkBufSize = CNBufferSize;
+        BYTE* const checkBuf = (BYTE*)malloc(checkBufSize);
+        ZSTD_CCtx_params* params = ZSTD_createCCtxParams();
+        ZSTD_CCtx* staticCCtx;
+        void* cctxBuf;
+        EMF_testCase seqProdState;
+
+        CHECK_Z(ZSTD_CCtxParams_setParameter(params, ZSTD_c_validateSequences, 1));
+        CHECK_Z(ZSTD_CCtxParams_setParameter(params, ZSTD_c_enableSeqProducerFallback, 0));
+        ZSTD_CCtxParams_registerSequenceProducer(params, &seqProdState, zstreamSequenceProducer);
```
Cyan4973, line 2372: While other parameter settings functions, such as `ZSTD_CCtxParams_setParameter()`, can fail,
`ZSTD_CCtxParams_registerSequenceProducer()` specifically cannot fail.
There is probably a topic around interface consistency here.

embg: Note that if we allow `ZSTD_CCtxParams_registerSequenceProducer()` to fail, then we also have to allow `ZSTD_registerSequenceProducer()` to fail.

lib/compress/zstd_compress.c
```diff
@@ -7084,14 +7084,25 @@ ZSTD_parameters ZSTD_getParams(int compressionLevel, unsigned long long srcSizeH
...
+        &zc->requestedParams, extSeqProdState, extSeqProdFunc
+    );
+}
+
+void ZSTD_CCtxParams_registerSequenceProducer(
+  ZSTD_CCtx_params* params,
+  void* extSeqProdState,
+  ZSTD_sequenceProducer_F extSeqProdFunc
 ) {
     if (extSeqProdFunc != NULL) {
```
Cyan4973, line 7103: If this function cannot fail,
at least add some debug input sanitization checks,
such as `assert(params != NULL);`.

lib/zstd.h
```diff
@@ -2824,6 +2821,22 @@ ZSTD_registerSequenceProducer(
...
+/*! ZSTD_CCtxParams_registerSequenceProducer() :
+ * Same as ZSTD_registerSequenceProducer(), but operates on ZSTD_CCtx_params.
+ * This is used for accurate size estimation with ZSTD_estimateCCtxSize_usingCCtxParams(),
+ * which is needed when creating a ZSTD_CCtx with ZSTD_initStaticCCtx().
+ *
+ * If you are using the external sequence producer API in a scenario where ZSTD_initStaticCCtx()
+ * is required, then this function is for you. Otherwise, you probably don't need it.
+ *
+ * See tests/zstreamtest.c for example usage. */
+ZSTDLIB_STATIC_API void
```
Cyan4973, line 2833: If this function cannot fail,
then there are some conditions for this function to be necessarily successful.
Publish them explicitly.

tests/zstreamtest.c
Cyan4973: Yes, I guess they both should follow the same convention.

Cyan4973 (approved):

lib/zstd.h
embg: Discussed offline -- only condition is that the `ZSTD_CCtx` or `ZSTD_CCtx_params` pointer is valid -- we don't need to explicitly document this -- can just use asserts.

PR #3858, opened 2024-01-02
CI: meson: use builtin handling for MSVC

eli-schwartz:
This avoids downloading -- and periodically bumping the checksum for -- a third-party action that isn't strictly required, and thus helps keep down dependencies and reduce update churn.

@Cyan4973 suggested this at https://github.com/facebook/zstd/commit/937e9d3b6257304e1533e9e6104ecff8e8dd5a30#r136033915

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -63,7 +63,7 @@ jobs:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # tag=v3
     - name: build zstd bin against a dynamic lib (debuglevel for more dependencies)
       run: |
-        make -C lib lib-mt-release 
+        make -C lib lib-mt-release
         DEBUGLEVEL=2 make -C programs zstd-dll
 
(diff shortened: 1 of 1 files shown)


eli-schwartz:
The ilammy/msvc-dev-cmd action is pretty useful in the generic case. Microsoft's compilers require both being added to PATH, as well as various inscrutable variables to be set -- the only canonical way to set it up is to run their batch script, which is basically full of private implementation details. This action sets that up for you and exports it to future steps in your action.

Meson can internally run this batch script for you, scrape its data, and provide it for all uses of meson itself, whether for setup, compilation, testing, or installing, as needed. This requires launching builds through the `meson compile` wrapper (unneeded on non-Windows platforms). It means that you shouldn't need ilammy/msvc-dev-cmd when running meson tests.

Since meson is already a dependency either way, this allows dropping down from 2 dependencies to 1 dependency.

Cyan4973 (approved):

Cyan4973:
Thanks @eli-schwartz !
This looks good to me !

PR #3860, opened 2024-01-13
Fix building xxhash on AIX 5.1

likema:
When building `zstd` by **GCC 4.5.4** on **AIX 5.1** , GCC could not found `stdint.h`

Similar to #2281


Cyan4973:
I wonder if there would be a way to test `AIX` in CI ?

Cyan4973 (approved):

PR #3861, opened 2024-01-13
Fix building on Solaris 10 x86_64 GCC 4.9.2

likema:
Prevent building x86_64 ASM from Solaris x86_64


build/cmake/lib/CMakeLists.txt
```diff
@@ -42,7 +42,10 @@ if (MSVC)
     file(GLOB DecompressSources ${LIBRARY_DIR}/decompress/*.c)
     add_compile_options(-DZSTD_DISABLE_ASM)
 else ()
-    file(GLOB DecompressSources ${LIBRARY_DIR}/decompress/*.c ${LIBRARY_DIR}/decompress/*.S)
+    file(GLOB DecompressSources ${LIBRARY_DIR}/decompress/*.c)
+    if (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" AND NOT CMAKE_SYSTEM_NAME STREQUAL "SunOS")
+        file(GLOB DecompressASMSources ${LIBRARY_DIR}/decompress/*.S)
+    endif ()
```
Cyan4973: Should there be some kind of :
```
else ()
add_compile_options(-DZSTD_DISABLE_ASM)
```
when the `*.S` assembly files are not part of the source list ?

Cyan4973:
This PR conflicts with an old patch that was recently merged.
It should be trivially fixable though.

likema:
> This PR conflicts with an old patch that was recently merged. It should be trivially fixable though.

Fixed

Cyan4973 (approved):

Cyan4973:
I finally found the time to create a Solaris test for Github Actions.
It can be seen here : https://github.com/facebook/zstd/actions/runs/7690458663/job/20954282520?pr=3885

The test includes running `cmake`, both `build` and `install` stages.
I was expecting it to fail, considering the description of this PR, which is accepted but not merged, hence not yet applied to #3885.
But apparently, it's not necessary : the test just runs completely fine without it.

Among the potential differences that could explain this surprising outcome: 
this PR mentions an issue with Solaris 10 and `gcc` 4.9.2,
while #3885 runs a Solaris VM version 11.4, shipping `gcc` 5.5.0.
So maybe something changed between versions 10 and 11 (`gcc` version notably).

Conversely, it also means that this patch might be a bit too strong: 
if I read it correctly, it removes assembly support for _any_ version of `SunOS`.
But, since `11.4` works fine, it could miss some performance due to the removal of this support.

Therefore, it seems a more adjusted solution would be to limit removal of asm support for Solaris version 10 and below.

likema:
You are right.

Checking if x86_64 ASM source can be compiled in `cmake` may be a better solution. 

I will try to do it later

Cyan4973 (changes requested):
Find a different way for this constraint (no-asm) to be more limited,
since Solaris 11.4 (at least) doesn't seem to need it.

PR #3862, opened 2024-01-13
Fix building on HP-UX 11.11 PA-RISC

likema:
By CMake 3.8.2 and GCC 4.7.1


Cyan4973:
wow ! how are you able to test these less-than-common systems @likema ?

likema:
Our company has some HP-UX 11.11 PA-RISC machines.

build/cmake/CMakeLists.txt
```diff
@@ -115,9 +115,20 @@ endif ()
 # External dependencies
 #-----------------------------------------------------------------------------
 if (ZSTD_MULTITHREAD_SUPPORT AND UNIX)
-    set(THREADS_PREFER_PTHREAD_FLAG ON)
-    find_package(Threads REQUIRED)
-    if(CMAKE_USE_PTHREADS_INIT)
+    if (CMAKE_SYSTEM_NAME MATCHES "HP-UX")
+        find_package(Threads)
+        if (NOT Threads_FOUND)
```
Cyan4973, line 120: If I understand correctly, this patch covers an exception scenario, where :
- OS is `HP-UX`
- `find_package(Threads)` _may_ be successful (?), but it may also fail
- when it fails, the script enters this branch
- The actions in this branch essentially force using `pthread`, even though it was not found by `find_package(Threads)`, by initializing the relevant flags

Is that a correct interpretation ?

likema: Yes.

* On HP-UX 11.31 IA64  , CMake 3.8.2 with GCC 4.7.1 `find_package(Threads)` successfully.
* On HP-UX 11.11 PA-RISC,   CMake 3.8.2 with GCC 4.7.1 `find_package(Threads)` error.

However, it is hard to detect whether the architecture is PA-RISC by CMake:

```bash
/bin/uname -a # HP-UX 11.31 IA64
HP-UX hprx2660 B.11.31 U ia64 3240820735 unlimited-user license

/bin/uname -a # HP-UX 11.11 PA-RISC

HP-UX hpux11iv B.11.11 U 9000/800 360203311 unlimited-user license
```

I'm not sure whether `9000/800` denotes PA-RISC.

Cyan4973: OK, thanks, that's clearer.

It's probably worth explaining in comments,
because this distinction between "HP-UX" variants `PA-RISC` vs `IA64` and their consequences is not obvious from the code.

Cyan4973:
Since the code path employed by other systems (non "HP-UX") seems unaltered,
I think this change is safe,

Still, I believe there can be an opened topic about readability,
which matters for maintenance and future evolution of the script.

This patch manages to keep all the alterations in one place, so that's already quite good.

But I believe an argument could be made that reading this segment of code is now less obvious,
just by the nature of inserting an `if / then / else / end` branch in the middle of the "common" scenario,
all this for the benefit of a relatively rare scenario.

I was wondering if there could be a way to express the same change
but in a way which weights less on the readability of the more "common" code path.

I'm not prescriptive on a solution here, since I'm less familiar with `cmake`.
So it's more an opened discussion.

Cyan4973 (approved):
OK, I'll take it and likely will update it, in order to move some logic away from the main code path. This is a very specialized corner case, we don't want it to make the mental parsing of the main code path more difficult.

PR #3878, opened 2024-01-21
tcc on windows dosen't support multi thread & intrinsics?

kbkpbot:
As tcc on windows dosen't support multi thread & intrinsics, remove them from the single_file_libs.


Cyan4973:
This patch looks fine.
Note though that the CLA is compulsory before moving to merge.

PR #3884, opened 2024-01-27
disable Intel CET Compatibility tests

Cyan4973:
The binary blob that must be downloaded from intel.com is no longer available


felixhandte:
The current version is available here: https://downloadmirror.intel.com/813591/sde-external-9.33.0-2024-01-07-lin.tar.xz

I'll try submitting a patch that updates the link to that, unless you have any objection.

Cyan4973:
> I'll try submitting a patch that updates the link to that, unless you have any objection.

Sounds good to me

PR #3885, opened 2024-01-29
add a Solaris VM test

Cyan4973:
to Github Actions CI

### DIFF
--- a/.github/workflows/solaris.yml
+++ b/.github/workflows/solaris.yml
@@ -48,4 +48,8 @@ jobs:
           gcc -v
           CC=gcc CFLAGS="-Werror -O1" gmake -j check
           pkg install cmake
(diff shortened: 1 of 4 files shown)


Cyan4973:
One issue with the `cmake` recipe is that it requires the `CXX` compiler to compile `zstd`.
This is incorrect. `zstd` itself doesn't require `CXX`, only `C`. This is a problem for systems which can provide a `C` compiler but not a `C++` one, since the `cmake` recipe will fail them.
It also seems this over-reach has been used and depended upon in some parts of the complex multi-files `cmake` recipe. Consequently, removing `CXX` from the `zstd` project creates errors in some environments,
`cmake` + `visual` being one of them.
I guess fixing this is a preliminary step in order to allow `Solaris VM` as a test.

PR #3886, opened 2024-01-29
add sparc64 compilation test

Cyan4973:


Cyan4973:
Ignore the "error" reported "Build Parsing Results".
it's a transient issue from CirrusCI, which has correctly completed its tests since.
Beside, this PR only touches Github Actions.
The only test that matters is this one:
https://github.com/facebook/zstd/actions/runs/7691177864/job/20956093323?pr=3886

PR #3893, opened 2024-01-31
Update Dependency in Intel CET Test; Re-Enable Test

felixhandte:
Motivated by #3884.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -577,6 +577,25 @@ jobs:
         make -C programs zstd-pgo
         ./programs/zstd -b
 
+  intel-cet-compatibility:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
While the change itself looks good to me,
I was also expecting to see the test in action in the Github Actions list,
and, for some reason, I don't see it ?
Not sure why, maybe I just missed it,
but it would be a good proof that everything is working correctly.,

Cyan4973:
Good, the Intel CET test is there in CI now :
https://github.com/facebook/zstd/actions/runs/7745901014/job/21122920425?pr=3893
and it works fine, so that covers it.

PS: no idea why you had to force-push your change in order for this test to be triggered ...

felixhandte:
I mis-indented it the first time. :unamused:

Issue #3793, opened 2023-10-16
higher zstd compression level resulting in larger compressed data


revintec:
using zstd 1.5.5, latest version as of writing
prepare an int array(each int occupies 4 bytes, little endian)
`[0,30,60,90,...]` 65536 ints, 65536*4 bytes
then compress it using various compression levels(simple compression, no dict):
```
2023-10-17 02:14:22.792 TRACE original size: 262144
2023-10-17 02:14:22.851 TRACE level 0 204162
2023-10-17 02:14:22.852 TRACE level 1 204103
2023-10-17 02:14:22.852 TRACE level 2 204118
2023-10-17 02:14:22.853 TRACE level 3 204162
2023-10-17 02:14:22.854 TRACE level 4 204136
2023-10-17 02:14:22.856 TRACE level 5 204147
2023-10-17 02:14:22.858 TRACE level 6 204141
2023-10-17 02:14:22.860 TRACE level 7 204161
2023-10-17 02:14:22.862 TRACE level 8 204161
2023-10-17 02:14:22.863 TRACE level 9 204161
2023-10-17 02:14:22.865 TRACE level 10 204161
2023-10-17 02:14:22.868 TRACE level 11 204165
2023-10-17 02:14:22.871 TRACE level 12 204161
2023-10-17 02:14:22.877 TRACE level 13 204143
2023-10-17 02:14:22.893 TRACE level 14 83240
2023-10-17 02:14:22.907 TRACE level 15 83240
2023-10-17 02:14:22.923 TRACE level 16 83242
2023-10-17 02:14:22.940 TRACE level 17 83242
2023-10-17 02:14:22.958 TRACE level 18 142849
2023-10-17 02:14:22.976 TRACE level 19 142849
2023-10-17 02:14:22.998 TRACE level 20 142849
2023-10-17 02:14:23.017 TRACE level 21 142849
2023-10-17 02:14:23.035 TRACE level 22 142849
```
as seen from the above output, higher compression level(18) starts resulting in larger compressed data
1. is that in line with exceptions? I thought higher compression level should resulting in smaller compressed data, this one is over 70% larger. how can I produce the smallest output data(ignoring compress time and/or memory consumption)?
-- a search using `compression level size` in issues results in no relative information in the first page, nor relative result in google :( sorry if this has already been brought up

and there's a related questions I'm putting into a same issue(forgive me :)

[...]

Issue #3014, opened 2022-01-20
Poor compression of binary numeric data with surprising results


RedBeard0531:
**Describe the bug**

I wrote a simple test to produce a file with 1 million 4-byte incrementing ints, and tried to compress with various settings. I tried big and little endian, and I tried starting at both 0 and 0x1234567 to fill all bytes. I'm sure there are a lot more interesting cases, but this seemed like a good starting point and revealed a lot already. Zstd generally did fairly poorly, but there were some interesting results that give me a glimmer of hope that there is room for significant improvement.

Interesting findings:
1. Compression levels 1 through 17 behave roughly the same regardless of input, reducing to roughly 67% of initial size.
2. Something changes at level 18 and for most inputs it gives significantly better results
3. `--format=xz -1` still compresses this data much better and faster than `--format=zstd -18`
4. Zstd is very sensitive to endianness and base value, while xz barely changes (I wrote this before noticing that I wasn't passing `-1` when testing xz as I intended to. When I added it and reran the testing, only one realy changed size, and it got smaller...)

Results:

|endian|base|zstd -18|xz -1|notes|
|-|-|:-|:-|-|
|Big|0|10.46% in 0.791s|6.47% in 0.272s|best result for zstd|
|Big|0x1234567|26.84% in 0.578s|6.38% in 0.263s| |
|Little|0|67.21% in 0.532s|2.52% in 0.243s|Worst result for zstd, -18 is same compression as -1. <br> Best result for xz, although default level compresses to 6.30%|
|Little|0x12345678| 25.43% in 0.568ms| 6.37% in 0.257|With a non-zero base, endianness doesn't matter|

**To Reproduce**

I used this python file to generate the test data, toggling the commented out lines.

```python
from array import array

BASE = 0
#BASE = 0x12345678 # uncomment to test non-zero high bytes

a = array('i')
for i in range(1000000):
    a.append(BASE + i)

#a.byteswap() # uncomment to try bigendian

with open('test.bin', 'wb') as f:
    a.tofile(f)
```

I ran `python ints.py && time zstd --format=zstd -18 test.bin -o /dev/null; time zstd --format=xz -1 test.bin -o /dev/null` to see the compression ratios and times for zstd and xz. Other formats and levels were also tested, but that is the command I used to produce the table above.

**Expected behavior**

I know this isn't necessarily the target use case for zstd, but it still seemed to do much worse than I expected. In particular:

[...]

PR #3895, opened 2024-02-04
Improve compression of Arrays of Integers (High compression mode)

Cyan4973:
This is an answer to issues similar to #3793 and #3014,
which report a poor compression ratio at high compression level
on synthetic data, typically arrays of 32-bit integers with very little consecutive variation (like an increasing series).
While such synthetic scenario tends to be "too regular", and therefore not representative of less regular "real" data,
it's nonetheless frequently used as a test scenario, because it is easy to produce.

The High compression mode has indeed a weakness in such a scenario,
and this patch tries to address it.
A first patch was able to improve these synthetic scenarios, increasing compression ratio by a significant amount. But unfortunately, it would then lose some compression ratio for other test data.
This patch is a variant, which manages to improve compression ratio in the case of 32-bit integer arrays, without losing anything in other cases.

This is illustrated in below test cases : 
| file | raw | `dev` | this `PR` | Improvement |
| --- | ---:| ---:| ---:| --- |
| inc_1M | 4000000 | 2687382 | 831424 | x3.23 | 
| issue3793 | 262144 | 142849 | 82866 | x1.72 |
| `.debug_str_offsets` | 296304 | 179814 | 131410 | +36.8% |
| canterbury/kennedy.xls | 1029744 | 69724 | 64825 | +7.5% |
| calgary/geo | 102400 | 64712 | 63051 | +2.6% |
| silesia/x-ray | 8474240 | 5155749 | 5129823 | +0.5% | 
| silesia/samba | 21606400 |  3907283 | 3899439 | +0.2% |
| silesia.tar | 211957760 | 52988966 | 52877671 | +0.2%  |

As can be seen, compression benefits are very sensible on the synthetic examples. More realistic samples (such as standard compression corpuses) get lower yet noticeable gains, notably `.debug_str_offsets`, `kennedy.xls`, `geo` and `x-ray`. Several other files get smaller gains (+0.1% range), but more importantly, no file in any compression corpus is negatively impacted by this patch (beyond a few bytes here and there just as a consequence of random luck). 
Overall, the resulting outcome of this patch on `silesia` at level 19 is > 100 KB savings. 

As a side note, I was hoping for an even better outcome because the gains from an earlier patch for 32-bit integer arrays were larger. But this previous patch was also negatively impacting compression ratio of other (non 32-bit integers) files. Therefore I believe this positive-only patch is a more acceptable trade-off for merge.

I might try later to reproduce the stronger results of the first patch (on 32-bit integer arrays), with the hope to find a solution which doesn't negatively impact other data. But this may require a complete overhaul of the optimal parser, and in the meantime, we can  at least use this patch.

### DIFF
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -1361,7 +1361,7 @@ _shortestPath:   /* cur, last_pos, best_mlen, best_off have to be set */
 
             DEBUGLOG(6, "start reverse traversal (last_pos:%u, cur:%u)",
                         last_pos, cur); (void)last_pos;
-            assert(storeEnd < ZSTD_OPT_NUM);
+            assert(storeEnd < ZSTD_OPT_SIZE);
             DEBUGLOG(6, "last stretch copied into pos=%u (llen=%u,mlen=%u,ofc=%u)",
                         storeEnd, lastStretch.litlen, lastStretch.mlen, lastStretch.off);
             if (lastStretch.litlen > 0) {
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -1138,7 +1138,7 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
              */
             opt[0].price = LL_PRICE(litlen);
             ZSTD_STATIC_ASSERT(sizeof(opt[0].rep[0]) == sizeof(rep[0]));
-            memcpy(&opt[0].rep, rep, sizeof(opt[0].rep));
+            ZSTD_memcpy(&opt[0].rep, rep, sizeof(opt[0].rep));
 
             /* large match -> immediate encoding */
             {   U32 const maxML = matches[nbMatches-1].len;
--- a/lib/compress/zstd_opt.c
+++ b/lib/compress/zstd_opt.c
@@ -566,7 +566,7 @@ void ZSTD_updateTree_internal(
     const BYTE* const base = ms->window.base;
     U32 const target = (U32)(ip - base);
     U32 idx = ms->nextToUpdate;
-    DEBUGLOG(6, "ZSTD_updateTree_internal, from %u to %u  (dictMode:%u)",
+    DEBUGLOG(7, "ZSTD_updateTree_internal, from %u to %u  (dictMode:%u)",
                 idx, target, dictMode);
 
     while(idx < target) {
@@ -1069,6 +1069,10 @@ listStats(const U32* table, int lastEltID)
 
 #endif
 
+#define LIT_PRICE(_p) (int)ZSTD_rawLiteralsCost(_p, 1, optStatePtr, optLevel)
+#define LL_PRICE(_l) (int)ZSTD_litLengthPrice(_l, optStatePtr, optLevel)
+#define LL_INCPRICE(_l) (LL_PRICE(_l) - LL_PRICE(_l-1))
+
 FORCE_INLINE_TEMPLATE
 ZSTD_ALLOW_POINTER_OVERFLOW_ATTR
 size_t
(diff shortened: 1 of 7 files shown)


Cyan4973:
So there is still a reported problem ("use-after-poison") in one of the tests (`asan-ubsan-regression`), 
but, while I can reproduce the error locally on a linux workstation,
I'm currently unable to extract any trace from it,
which is hampering deeper investigation.

_edit_ : now fixed, but could probably have been easier.

lib/compress/zstd_opt.c
```diff
@@ -1143,38 +1148,41 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
                             nbMatches, maxML, maxOffBase, (U32)(ip-prefixStart));
 
                 if (maxML > sufficient_len) {
-                    lastSequence.litlen = litlen;
-                    lastSequence.mlen = maxML;
-                    lastSequence.off = maxOffBase;
-                    DEBUGLOG(6, "large match (%u>%u), immediate encoding",
+                    lastStretch.litlen = 0;
```
terrelln, line 1158: I don't understand this change.

Edit: After reading the code I understand. It would've been helpful to see a description of the change that this diff made in the summary, because it was non-obvious until I finished reading the code.

lib/compress/zstd_opt.c
```diff
@@ -1271,63 +1297,104 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
...
+        /* let's write the shortest path solution
+         * solution is stored in @opt,
+         * in reverse order,
+         * starting from @storeEnd (==cur+1)
+         * (effectively partially overwriting @opt).
+         * Content is changed too:
+         * - So far, @opt stored stretches, aka a match followed by literals
+         * - Now, it will store sequences, aka literals followed by a match
+         */
+        {   U32 const storeEnd = cur + 2;
```
terrelln, line 1371: Why +2? Also update the docs which say that `storeEnd == cur+1`.

lib/compress/zstd_opt.c
```diff
@@ -1185,40 +1193,60 @@ ZSTD_compressBlock_opt_generic(ZSTD_matchState_t* ms,
...
+                            DEBUGLOG(7, "==> match+1lit is cheaper (%.2f < %.2f) (hist:%u,%u,%u) !",
+                                        ZSTD_fCost(with1literal), ZSTD_fCost(withMoreLiterals),
+                                        newReps.rep[0], newReps.rep[1], newReps.rep[2] );
+                            opt[cur+1] = prevMatch;  /* mlen & offbase */
+                            ZSTD_memcpy(opt[cur+1].rep, &newReps, sizeof(repcodes_t));
+                            opt[cur+1].litlen = 1;
+                            opt[cur+1].price = with1literal;
+                            if (last_pos < cur+1) last_pos = cur+1;
+                        }
+                    }
```
terrelln, line 1240: Besides the change to `litlength`, which is to facilitate this change, this section is the major functional change of this diff, correct?

terrelln (approved):

lib/compress/zstd_opt.c
Cyan4973: > Besides the change to `litlength`, which is to facilitate this change, this section is the major functional change of this diff, correct?

Yes

lib/compress/zstd_opt.c
Cyan4973: Note below, if `lastStretch.litlen > 0`, i.e. there are some literals _after_ the last match,
then we start to write into `opt[storeStart]` with `storeStart == storeEnd-1`,
therefore overwriting the content of `opt[storeStart]`. And if `storeEnd==cur+1`, then `storeStart` would be `==cur`.
However, `opt[cur]` is where the next _stretch_ is stored, and we haven't read it yet, since we are still dealing with `lastStretch`. So we would be overwriting a component of our next sequence.

A `+2` avoids this overwriting.
This is fine because `ZSTD_OPT_SIZE == ZSTD_OPT_NUM + 2`, so we have enough margin, even when the array is entirely filled. This is checked in the following `assert` : `assert(storeEnd < ZSTD_OPT_SIZE);`.

ribasushi:
Hi @Cyan4973, excellent work! Could you please clarify which "High compression mode" levels benefit from this work? Only `19`? `17+`? `15+`? 

I tried to figure this out from the changelog/code diff, but either I am missing it or it is just unclear...

Cyan4973:
It think it's active for strategies `bt_ultra` and `bt_ultra2`.
Depending on the size of the input, these strategies are triggered at different compression levels.
But assuming the file is "large", this should be levels 18 and above.

PR #3903, opened 2024-02-11
Reduce scope of variables

gruenich:
This improves readability, keeps variables local, and prevents the unintended use (e.g. typo) later on.
Found by Cppcheck (variableScope)

### DIFF
--- a/lib/compress/huf_compress.c
+++ b/lib/compress/huf_compress.c
@@ -1062,12 +1062,12 @@ HUF_compress1X_usingCTable_internal_body(void* dst, size_t dstSize,
     const BYTE* ip = (const BYTE*) src;
     BYTE* const ostart = (BYTE*)dst;
     BYTE* const oend = ostart + dstSize;
-    BYTE* op = ostart;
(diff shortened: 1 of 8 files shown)


gruenich:
This pull request does not make sense with Zstd relying on C89/C90. It should be closed, if #3904 rejected.

Cyan4973:
The general idea of this PR is good.
Reducing variable's scope tends to improve readability.

Some of the proposed changes are not compatible with the `C90` standard,
and some could be refactored to become compatible with it.

It's not a problem to open new blocks just for the purpose of reducing the scope of one or several variables. I know it's unusual, and some people do not like opening additional blocks, but the value is in ending the variable's scope, and to that end, such blocks are useful.

gruenich:
Thanks Yann for you kind and encouraging words. I updated this pull request.

Cyan4973 (approved):
Pretty straightforward.
All these little changes do not impact the intention of the code, but make it a bit clearer to read, hence to maintain. 
It's all good to me.

PR #3910, opened 2024-02-14
Starting state: Sequences

roynalnaruto:
While initialising the FSE state in the Sequences section, we follow the interleaves order:
```
[..., offset, match_length, literal_length, literal_length, match_length, offset, ...]
```

This suggests there might have been a mismatch in the document?

Please close PR if I misunderstood.

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -819,8 +819,8 @@ each using the required number of bits in their respective _accuracy_,
 decoded previously from their normalized distribution.
 
 It starts by `Literals_Length_State`,
-followed by `Offset_State`,
-and finally `Match_Length_State`.
+followed by `Match_Length_State`,
+and finally `Offset_State`.
 
 Reminder : always keep in mind that all values are read _backward_,
 so the 'start' of the bitstream is at the highest position in memory,


Cyan4973 (changes requested):
The specification is correct : https://github.com/facebook/zstd/blob/v1.5.5/lib/decompress/zstd_decompress_block.c#L1378

roynalnaruto:
If the specs are correct, then we can close this PR. Thank you :)

roynalnaruto:
@Cyan4973 however there is another section below that states:
```
If it is not the last sequence in the block, the next operation is to update states. Using the rules pre-calculated in the decoding tables, Literals_Length_State is updated, followed by Match_Length_State, and then Offset_State. See the [FSE section](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#fse) for details on how to update states from the bitstream.
```

Does it mean the order is incorrect here, and should instead be:
```
... Using the rules pre-calculated in the decoding tables, Literals_Length_State is updated, followed by Offset_State, and then Match_Length_State ...
```

Cyan4973:
No, the order is correct too : 
https://github.com/facebook/zstd/blob/v1.5.5/lib/decompress/zstd_decompress_block.c#L1291

There are a few things which can be confused: 
- The spec must be understood from a decompressor point of view, with the understanding that the encoder is doing the reverse operation, hence in reverse order.
- Reading the initial FSE state values is a separated operation. It's not part of the main decoding loop, and therefore isn't constrained by the ordering inside the loop. Each operation ends up having its own ordering.

roynalnaruto:
So if I understand correctly:
- Initialising the FSE state (from a decoder's perspective): `LL state -> Offset state -> ML state`
- Further decoding loop: `numbits Offset -> numbits ML -> numbits LL -> update LL state -> update ML state -> update Offset state`

Cyan4973:
Yes

roynalnaruto:
Closing as there is no issue with the original documentation

PR #3915, opened 2024-02-24
Improve speed of ZSTD_c_targetCBlockSize

Cyan4973:
This PR improves the speed of parameter `ZSTD_c_targetCBlockSize`, which is employed to generate smaller compressed blocks. This parameter was seldom used up to now, and as a consequence, its performance impact was not properly investigated. It proved to be major, as following measurements show.

Some benchmarks for reference, on a local M1 Pro laptop : 
```
/* normal compression */
./zstd -b8 ~/dev/bench/enwik7
 8#enwik7            :  10000000 ->   3184407 (x3.140),   62.8 MB/s, 1267.7 MB/s
 
/* before this PR */
./zstd -b8 ~/dev/bench/enwik7 --target-compressed-block-size=4KB
 8#enwik7            :  10000000 ->   3194310 (x3.131),   3.75 MB/s, 1080.1 MB/s

/* after this PR */
./zstd -b8 ~/dev/bench/enwik7 --target-compressed-block-size=4KB
 8#enwik7            :  10000000 ->   3194256 (x3.131),   62.1 MB/s, 1075.1 MB/s
```

The performance cost of this parameter has been reduced to ~10% of level 1, quickly becoming negligible as a portion of total compression time as level increases (~1% at level 8).

The drawback of this new method is that compressed blocks have a more variable size: they still "target" the requested compressed block size on average, but the variation around the average is higher.
Note that this contract has always been "approximate": even the previous "slow" method would not produce blocks with the requested size, just "close enough". Here, we take full advantage of the nature of this contract for speed benefits.

It is expected that, considering the context in which this parameter is employed, where the goal is to reduce latency to first produced byte, the impact of this variance should not be significant. But the speed improvement, on the other hand, is major, and unlocks usage of this parameter.

If more regularity or guarantees on the compressed size become required or desirable in the future, more refinements will be possible in future updates.

### DIFF
--- a/lib/compress/zstd_compress_superblock.c
+++ b/lib/compress/zstd_compress_superblock.c
@@ -136,6 +136,7 @@ ZSTD_seqDecompressedSize(seqStore_t const* seqStore,
     size_t matchLengthSum = 0;
     size_t litLengthSum = 0;
     (void)(litLengthSum); /* suppress unused variable warning on some environments */
+    DEBUGLOG(6, "ZSTD_seqDecompressedSize (%u sequences from %p) (last==%i)", (unsigned)nbSeq, sp, lastSequence);
     while (sp < send) {
         ZSTD_sequenceLength const seqLen = ZSTD_getSequenceLength(seqStore, sp);
         litLengthSum += seqLen.litLength;
@@ -279,7 +280,8 @@ static size_t ZSTD_compressSubBlock(const ZSTD_entropyCTables_t* entropy,
                 litSize, nbSeq, writeLitEntropy, writeSeqEntropy, lastBlock);
     {   size_t cLitSize = ZSTD_compressSubBlock_literal((const HUF_CElt*)entropy->huf.CTable,
                                                         &entropyMetadata->hufMetadata, literals, litSize,
-                                                        op, oend-op, bmi2, writeLitEntropy, litEntropyWritten);
+                                                        op, (size_t)(oend-op),
+                                                        bmi2, writeLitEntropy, litEntropyWritten);
         FORWARD_IF_ERROR(cLitSize, "ZSTD_compressSubBlock_literal failed");
         if (cLitSize == 0) return 0;
         op += cLitSize;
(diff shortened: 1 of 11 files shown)


Cyan4973:
Merging early, to increase time spent in oss-fuzz, since this parameter is part of the fuzzer suite

lib/compress/zstd_compress_superblock.c
```diff
@@ -450,96 +465,113 @@ static size_t ZSTD_compressSubBlock_multi(const seqStore_t* seqStorePtr,
...
-        cBlockSizeEstimate = ZSTD_estimateSubBlockSize(lp, litSize, ofCodePtr, llCodePtr, mlCodePtr, seqCount,
-                                                       &nextCBlock->entropy, entropyMetadata,
-                                                       workspace, wkspSize, writeLitEntropy, writeSeqEntropy);
-        if (cBlockSizeEstimate > targetCBlockSize || lastSequence) {
+        /* Note: this is very approximative. Obviously, some sub-blocks will be larger and others smaller.
+         * But the contract of this feature has always been approximative, so for now we'll leverage it for speed.
+         * It can be refined later, for closer-to-target compressed block size, if it ever matters. */
+    }
+
+    /* write sub-blocks */
```
terrelln, line 503: One way to guarantee the contract that we don't violate compress bound, is to ensure that the total sequence length in every block is >= `targetCBlockSize`, and increase the number of sequences until that condition holds.

We are already iterating over all the sequences in the sub-block in `countLiterals()`, so it shouldn't cost more to enforce a minimum.

If we don't do this, I'm worried that we will hit an edge case where we violate `ZSTD_compressBound()`. I am not confident that this case is impossible.

Cyan4973: In this case, the total number of sub-blocks produced is bounded.
So it cannot degenerate in a situation where it produces a lot of sub-blocks, therefore a lot of block headers, violating `ZSTD_compressBound()`.

PR #3917, opened 2024-02-26
More regular block sizes with `targetCBlockSize`

Cyan4973:
This is a follow up from #3915,
which fixed the performance issue of `targetCBlockSize` by setting up a fast heuristic to determine individual block boundaries. This proved way faster, but also more approximative.

This patch is a follow-up refinement, which fixes most of the variability issues observed in #3915.

In order to observe the impact, let's record the block sizes produced by `zstd -8 --target-compressed-block-size=4KB` on a range of files, as observed by the decoder : 

| enwik5 | block1 | block2 | block3 | block4 | block5 | block6 |  block7 |  block8 | 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | 
| `dev` | 5474 | 4096 | 3605 | 3740 | 3701 | 4202 | 3528 | 3384 |
| `PR`  | 3894 | 3775 | 3832 | 3866 | 3901 | 3982 | 4105 | 3842 | 

| calgary/bib | block1 | block2 | block3 | block4 | block5 | block6 |  block7 |  block8 | 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | 
| `dev` | 4888 | 3896 | 3668 | 3611 | 3627 | 3524 | 3627 | 3473 |
| `PR`  | 3596 | 3642 | 3680 | 3733 | 3774 | 3762 | 3820 | 3833 |

| calgary/geo | block1 | block2 | block3 | block4 | block5 | block6 |  block7 |  block8 | 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | 
| `dev` |12756 | 8245 | 4849 | 4099 | 3352 | 2924 | 4027 | 3522 |
| `PR`  | 3834 | 3874 | 3960 | 3965 | 4020 | 3930 | 4010 | 3965 |

| calgary/obj2 | block1 | block2 | block3 | block4 | block5 | block6 |  block7 |  block8 | 
| --- | --- | --- | --- | --- | --- | --- | --- | --- | 
| `dev` | 5200 | 5382 | 5646 | 4835 | 3722 | 3155 | 3077 | 3390 |
| `PR`  | 3843 | 3773 | 3522 | 3749 | 3789 | 3868 | 3916 | 3883 |

As can be observed, the block sizes are more regular with this new patch.
Especially, the first block tends to be larger in previous `dev` strategy, as a consequence of being generally less compressible. As one of the goals of `targetCBlockSize` is to improve latency to first produced byte, this update plays well in this direction.

The cost is a higher runtime for size evaluation,
but fortunately, it's still very acceptable: 
previous `dev` strategy was using ~10% of `-1` compression time,
this new one uses ~15% of `-1` compression time.
In both case, the relative cost of this strategy goes down as the compression level increases, quickly becoming irrelevant:
this PR has an impact of ~2% compression speed at level 8.
The compression ratio remains virtually unchanged compared to #3915.

### DIFF
--- a/lib/compress/zstd_compress_superblock.c
+++ b/lib/compress/zstd_compress_superblock.c
@@ -443,21 +443,29 @@ static size_t sizeBlockSequences(const seqDef* sp, size_t nbSeqs,
                 size_t targetBudget, size_t avgLitCost, size_t avgSeqCost,
                 int firstSubBlock)
 {
-    size_t n, budget = 0;
+    size_t n, budget = 0, inSize=0;
     /* entropy headers */
-    if (firstSubBlock) {
-        budget += 120 * BYTESCALE; /* generous estimate */
-    }
+    size_t const headerSize = (size_t)firstSubBlock * 120 * BYTESCALE; /* generous estimate */
+    assert(firstSubBlock==0 || firstSubBlock==1);
+    budget += headerSize;
+
     /* first sequence => at least one sequence*/
     budget += sp[0].litLength * avgLitCost + avgSeqCost;
     if (budget > targetBudget) return 1;
+    inSize = sp[0].litLength + (sp[0].mlBase+MINMATCH);
 
     /* loop over sequences */
     for (n=1; n<nbSeqs; n++) {
         size_t currentCost = sp[n].litLength * avgLitCost + avgSeqCost;
-        if (budget + currentCost > targetBudget) break;
         budget += currentCost;
+        inSize += sp[n].litLength + (sp[n].mlBase+MINMATCH);
+        /* stop when sub-block budget is reached */
+        if ( (budget > targetBudget)
+            /* though continue to expand until the sub-block is deemed compressible */
+          && (budget < inSize * BYTESCALE) )
+            break;
     }
+
     return n;
 }
 
--- a/lib/compress/zstd_compress_superblock.c
+++ b/lib/compress/zstd_compress_superblock.c
@@ -449,7 +449,7 @@ static size_t sizeBlockSequences(const seqDef* sp, size_t nbSeqs,
         budget += 120 * BYTESCALE; /* generous estimate */
     }
     /* first sequence => at least one sequence*/
-    budget += sp[0].litLength * avgLitCost + avgSeqCost;;
+    budget += sp[0].litLength * avgLitCost + avgSeqCost;
     if (budget > targetBudget) return 1;
 
     /* loop over sequences */
--- a/lib/compress/zstd_compress_superblock.c
+++ b/lib/compress/zstd_compress_superblock.c
@@ -521,6 +521,9 @@ static size_t ZSTD_compressSubBlock_multi(const seqStore_t* seqStorePtr,
         DEBUGLOG(5, "estimated fullblock size=%u bytes ; avgLitCost=%.2f ; avgSeqCost=%.2f ; targetCBlockSize=%u, nbSubBlocks=%u ; avgBlockBudget=%.0f bytes",
                     (unsigned)ebs.estBlockSize, (double)avgLitCost/BYTESCALE, (double)avgSeqCost/BYTESCALE,
                     (unsigned)targetCBlockSize, (unsigned)nbSubBlocks, (double)avgBlockBudget/BYTESCALE);
+        /* simplification: if estimates states that the full superblock doesn't compress, just bail out immediately
+         * this will result in the production of a single uncompressed block covering @srcSize.*/
+        if (ebs.estBlockSize > srcSize) return 0;
 
         /* compress and write sub-blocks */
         for (n=0; n+1 < nbSubBlocks; n++) {
@@ -568,7 +571,7 @@ static size_t ZSTD_compressSubBlock_multi(const seqStore_t* seqStorePtr,
                     sp += seqCount;
                     blockBudgetSupp = 0;
             }   }
-            /* otherwise : do not compress yet, coalesce current block with next one */
+            /* otherwise : do not compress yet, coalesce current sub-block with following one */
         }
     } /* if (nbSeqs > 0) */
 
--- a/lib/compress/zstd_compress_superblock.c
+++ b/lib/compress/zstd_compress_superblock.c
@@ -390,7 +390,11 @@ static size_t ZSTD_estimateSubBlockSize_sequences(const BYTE* ofCodeTable,
     return cSeqSizeEstimate + sequencesSectionHeaderSize;
 }
 
-static size_t ZSTD_estimateSubBlockSize(const BYTE* literals, size_t litSize,
+typedef struct {
+    size_t estLitSize;
+    size_t estBlockSize;
+} EstimatedBlockSize;
+static EstimatedBlockSize ZSTD_estimateSubBlockSize(const BYTE* literals, size_t litSize,
                                         const BYTE* ofCodeTable,
                                         const BYTE* llCodeTable,
                                         const BYTE* mlCodeTable,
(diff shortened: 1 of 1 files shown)


lib/compress/zstd_compress_superblock.c
```diff
@@ -427,17 +433,43 @@ static size_t countLiterals(seqStore_t const* seqStore, const seqDef* sp, size_t
...
+                size_t targetBudget, size_t avgLitCost, size_t avgSeqCost,
+                int firstSubBlock)
+{
+    size_t n, budget = 0;
+    /* entropy headers */
+    if (firstSubBlock) {
+        budget += 120 * BYTESCALE; /* generous estimate */
+    }
+    /* first sequence => at least one sequence*/
+    budget += sp[0].litLength * avgLitCost + avgSeqCost;;
```
terrelln:
```suggestion
    budget += sp[0].litLength * avgLitCost + avgSeqCost;
```

lib/compress/zstd_compress_superblock.c
```diff
@@ -427,17 +433,43 @@ static size_t countLiterals(seqStore_t const* seqStore, const seqDef* sp, size_t
...
-    DEBUGLOG(5, "countLiterals for %zu sequences from %p => %zu bytes", seqCount, (const void*)sp, total);
+    DEBUGLOG(6, "countLiterals for %zu sequences from %p => %zu bytes", seqCount, (const void*)sp, total);
     return total;
 }
 
+#define BYTESCALE 256
+
+static size_t sizeBlockSequences(const seqDef* sp, size_t nbSeqs,
+                size_t targetBudget, size_t avgLitCost, size_t avgSeqCost,
+                int firstSubBlock)
```
terrelln, line 444: It would be pretty easy to modify this function to count the total uncompressed bytes in the sequences, and not cut a block as long as the uncompressed bytes is less than `targetCBlockSize`. That would be a simple way to guarantee that we don't violate compress bound.

I think it would also help keep block sizes even, since if the uncompressed size is less than `targetCBlockSize`, then the compressed size will be as well.

Given PR #1947, it likely isn't 100% necessary, but I think it is a nice sanity check, and should help improve compression ratio in edge cases.

lib/compress/zstd_compress_superblock.c
```diff
@@ -465,90 +496,125 @@ static size_t ZSTD_compressSubBlock_multi(const seqStore_t* seqStorePtr,
...
-        if (nbSeqs > nbSubBlocks) {
-            nbSeqsPerBlock = nbSeqs / nbSubBlocks;
-        } else {
-            nbSeqsPerBlock = 1;
-            nbSubBlocks = nbSeqs;
+        size_t const avgLitCost = nbLiterals ? (ebs.estLitSize * BYTESCALE) / nbLiterals : BYTESCALE;
+        size_t const avgSeqCost = ((ebs.estBlockSize - ebs.estLitSize) * BYTESCALE) / nbSeqs;
+        size_t nbSubBlocks = (ebs.estBlockSize + (targetCBlockSize-1)) / targetCBlockSize;
+        size_t n, avgBlockBudget, blockBudgetSupp=0;
+        if (nbSubBlocks<1) nbSubBlocks=1;
```
terrelln:
```suggestion
        const size_t nbSubBlocks = MAX((ebs.estBlockSize + (targetCBlockSize-1)) / targetCBlockSize, 1);
        size_t n, avgBlockBudget, blockBudgetSupp=0;
```

lib/compress/zstd_compress_superblock.c
```diff
@@ -465,90 +496,125 @@ static size_t ZSTD_compressSubBlock_multi(const seqStore_t* seqStorePtr,
...
+        avgBlockBudget = (ebs.estBlockSize * BYTESCALE) / nbSubBlocks;
+        DEBUGLOG(5, "estimated fullblock size=%u bytes ; avgLitCost=%.2f ; avgSeqCost=%.2f ; targetCBlockSize=%u, nbSubBlocks=%u ; avgBlockBudget=%.0f bytes",
+                    (unsigned)ebs.estBlockSize, (double)avgLitCost/BYTESCALE, (double)avgSeqCost/BYTESCALE,
+                    (unsigned)targetCBlockSize, (unsigned)nbSubBlocks, (double)avgBlockBudget/BYTESCALE);
+
+    /* compress and write sub-blocks */
```
terrelln: nit: This indentation is wrong

lib/compress/zstd_compress_superblock.c
```diff
@@ -465,90 +496,125 @@ static size_t ZSTD_compressSubBlock_multi(const seqStore_t* seqStorePtr,
...
+                                                llCodePtr, mlCodePtr, ofCodePtr,
+                                                cctxParams,
+                                                op, (size_t)(oend-op),
+                                                bmi2, writeLitEntropy, writeSeqEntropy,
+                                                &litEntropyWritten, &seqEntropyWritten,
+                                                0);
+                FORWARD_IF_ERROR(cSize, "ZSTD_compressSubBlock failed");
+
+                /* check compressibility, update state components */
+                if (cSize > 0 && cSize < decompressedSize) {
```
terrelln, line 564: Now that we've removed the restriction that blocks must not be expanded, I think we could remove this `cSize < decompressedSize` check, and just check that it isn't beyond the maximum block size.

I don't think we should do it in this PR, because we want to cherry pick it into a release, but we should consider it.

terrelln (changes requested):
I think that the `dstSize_tooSmall` case is handled well now. Looks like I fixed it in https://github.com/facebook/zstd/pull/1947.

I think we should update `sizeBlockSequences()` to consider the uncompressed size, then it looks good to me.

lib/compress/zstd_compress_superblock.c
Cyan4973: > That would be a simple way to guarantee that we don't violate compress bound.

If `compressBound()` violation is the concern here,
let's underline that the total nb of blocks produced is limited, bounded by `nbSubBlocks` : 
this code never generates more than `nbSubBlocks` blocks.

Since `nbSubBlocks = estimateCSize / targetCBlockSize`,
and `targetCBlockSize` is bounded by a minimum of `CBLOCK_TARGET_SIZE_MIN` (`1300` bytes currently),
`nbSubBlocks` cannot define a nb of blocks > 1 block per 1 KB of input data
hence it cannot defeat `compressBound()`,

unless the source block itself is < 1 KB as a consequence of user-controlled `flush()`,
in which case the `compressBound()` guarantee is out of scope, and there will be no split.

Cyan4973: Another way to look at it is that `compressBound()` is only a concern when blocks are not compressed.
In this case, note that this is controlled just below,
and if a sub-block is not compressed, it's not generated immediately => it's actually coalesced with the next one (effectively creating a larger sub-block). This process continues until the coalesced sub-block becomes compressible, or until reaching the end of the super-block. In which case, a single (final) uncompressed sub-block is produced per superblock.
So it can never degenerates into producing many small uncompressed sub-blocks.

terrelln: I still think that it is worth controlling the case where `uncompressedSize < targetCBlockSize`.

I know that compress bound really shouldn't be violated. First, by the fact that `nbSubBlocks` is supposed to not be too big. Second by the final check at the end that will emit an uncompressed block if `dstSize_tooSmall` or we don't shrink by enough.

However, I think this is a simple check that we can add that will stop us from emitting too-small blocks in some edge cases, and will help block consistency. Because if `uncompressedSize < targetCBlockSize`, we don't want to emit a block anyway.

Cyan4973: >  if uncompressedSize < targetCBlockSize, we don't want to emit a block anyway.

Isn't that already taken care of by the following test : 
`if (cSize > 0 && cSize < decompressedSize) {`

_edit_: 
another way to think about it is: could it be a faster way to reach the same conclusion ? In effect an optimization ?

This is interesting, but I'm concerned that I currently have a fairly straightforward way to guarantee reaching the last sub-block, and if intermediate sub-block may have unspecified large size, it may mess up the logic.
So this would have to be controlled, to ensure it doesn't introduce new side effects.

Cyan4973:
I've updated `sizeBlockSequences()` so that it also tracks uncompressed bytes, and only emits a block boundary if it believes that the block is compressible.

I still don't see how it could help with regards to `compressBound()` limit violation or with the risk of "emitting too-small blocks in some edge cases". 
But I believe it can improve speed in situations where data is almost-but-not-completely incompressible (if it's clearly incompressible, we bail out sooner, without even invoking `sizeBlockSequences()`), because it avoids having to compress the data, then realize it's not compressible, and then do another round: instead, we go straight to the enlarge block, or to the end of the superblock.
So I see that more as a speed optimization refinement.

I had concerns about making sure that the algorithm would not get lost by not seeing the expected nb of sub-blocks, but it turns out it's fine, there is already a check that jumps directly to the last block when all sequences are consumed, and it naturally fits there.

terrelln (approved):

PR #3931, opened 2024-03-07
meson: don't add `-pthread` to static linking flags on Windows

bgilbert:
Meson always returns `-pthread` in `dependency('threads')` on non-MSVC compilers.  On Windows we use Windows threading primitives, so we don't need this.  Avoid adding `-pthread` to libzstd's link flags, either as a Meson subproject or via pkg-config `Libs.private`, so the application doesn't inadvertently depend on winpthreads.

Add a Meson MinGW cross-compile CI test that checks for this.  It turns out that `pzstd` fails to build in that environment, so have the test skip building `contrib` for now.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -237,6 +237,50 @@ jobs:
           meson test -C builddir/ --print-errorlogs
           meson install -C builddir --destdir staging/
 
+  meson-mingw-cross-compilation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # tag=v4.1.1
+      - name: Install packages
+        run: |
+          sudo apt-get -qqq update
+          sudo apt-get -y install build-essential python3-pip ninja-build {gcc,g++}-mingw-w64-x86-64
+          pip install --pre meson
+      - name: Build with Meson
+        run: |
+          cat > cross.ini <<EOF
+          [binaries]
+          ar = 'x86_64-w64-mingw32-ar'
+          c = 'x86_64-w64-mingw32-gcc'
+          cpp = 'x86_64-w64-mingw32-g++'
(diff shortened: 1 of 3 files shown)


bgilbert:
Is there anything I can do to help this move forward?

Cyan4973:
It's mostly a matter of finding someone for the review.

In my understanding, this PR is helpful when using `meson` to build `zstd` on Windows.

It seems this is a combination we don't have in CI, hence we can't "witness" the impact of this change.
An opened question at this stage is if it would be worthwhile to add (and if yes: how, and what's the runtime cost).

bgilbert:
It's helpful for building zstd for Windows with MinGW, possibly via cross-compilation.  Many application stacks will already use winpthreads, either because something in the stack uses pthreads explicitly, or because something uses C/C++ thread intrinsics and MinGW-w64 is configured to implement those via pthreads.  In that case, zstd's current behavior amounts to harmless overlinking.  My stack uses MinGW-w64 configured to implement intrinsics via Win32 primitives (a less common configuration), and does not otherwise use or ship winpthreads, so the current behavior produces an unsatisfied dependency.

If desired, I can see about adding a short test job for MinGW + Meson, or provide manual build instructions for verification.

Cyan4973:
That would be a good idea.
It would help preserve this property across future modifications,
otherwise another contributor might come and change a detail which breaks this property, and we wouldn't notice.

bgilbert:
Added a CI job.

It turns out that the DLL doesn't end up linked against winpthreads, so the problem is limited to the linker flags we tell applications to use when linking with our static library.  I've updated the PR description and commit message accordingly.

Cyan4973 (approved):
This looks good to me!

Cyan4973:
I assume the last remaining CI error (`cygwin`) is unrelated

bgilbert:
Yes, looks like the Cygwin failure is a transient network issue.

bgilbert:
Is this ready to go in?

Cyan4973:
> Is this ready to go in?

yep

PR #3937, opened 2024-03-08
new method to deal with offset==0 erroneous edge case

Cyan4973:
It's possible to send an `offset==0` in some fairly specific edge case.
Such situation is erroneous, meaning the data is considered corrupted.

It could also be an attempt at crafting an attack vector, trying to produce a match of distance `0`, thus copying the output buffer with itself, effectively keeping the memory in the output buffer uninitialized.
To negate the attack vector possibility, such an offset is converted to `1` by the decoder, ensuring the output buffer is always overwritten, thus denying the possibility to generate uninitialized memory.

For the rare cases where this scenario would be triggered by some storage or transmission error, the decoder relies on the checksum to tell that data was corrupted. But if there is no checksum, it silently succeeds.

It would be better for such case to be unambiguously detected as an error.
The problem was that, at the time this issue was understood and protected against, introducing a new branch specifically to deal with it would measurably impact decompression speed for all use cases, because it's part of the hottest loop in the decoder, and this felt like a bad trade-off.

In this new method, when an `offset==0` is detected, it's converted into `(size_t)(-1)`, instead of 1.
The logic is that `(size_t)(-1)` is effectively an extremely large positive number, which should not pass the offset distance test at next stage (`execSequence()`). 

This is possible now because the offset is always checked (as it should), using a formula which is not vulnerable to arithmetic overflow:
```
RETURN_ERROR_IF(sequence.offset > (size_t)(oLitEnd - virtualStart),
```
It might be that when this issue was first raised, the `offset` check may have been different, and may have been a more risky proposal due to arithmetic overflow.
Now, given the current check, relying on this detection mechanism seems safe, and since it's already there, it doesn't impact performance.

The benefit is that the `offset==0` is now always detected as corrupted data, as opposed to relying on the checksum to detect the error.


Cyan4973:
Tried to create an example of such a case, at least for test, and then for illustration,
but the sequence compressor doesn't accept repcodes as input,
it instead deduces them from the offset history,
and passing `0` as an offset doesn't work (which feels logical)...

Cyan4973:
Finally created a golden sample `off0.bin.zst` by locally modifying the library to bypass all checks, thus forcing the creation of a `zstd` compressed frame featuring a `repcode==3` which will produce an `offset==0` scenario.
Checked that this frame is correctly detected as corrupted after this patch, 
and is decoded without producing an error by older versions (replacing the offset by a `1`), due to the absence of a checksum.

The golden sample is now added to the list of samples for which the decoder must fail during tests.

terrelln (approved):

embg:
@Cyan4973 FYI, I was trying to figure out why before this PR, there is no `golden-decompression-errors` folder, despite references to that folder in various tests. It looks like in this [old commit](https://github.com/facebook/zstd/commit/b46236278a0adea097ce7792824f93a678a74069) you forgot to `git add` the file `zeroSeq_extraneous.zst`.

That file is referenced in [tests/playTests.sh](https://github.com/facebook/zstd/blob/dev/tests/playTests.sh#L452), but it does not exist currently in `dev`, nor does the folder `golden-decompression-errors`.

This PR is fine, because the file you added will be picked up by [this test](https://github.com/facebook/zstd/blob/dev/tests/cli-tests/decompression/detectErrors.sh). So I'm not requesting changes on this PR, just flagging that the [old commit](https://github.com/facebook/zstd/commit/b46236278a0adea097ce7792824f93a678a74069) has an issue :).

Cyan4973:
> old commit

Interesting,
this file is present on my local git repository.

But indeed it's not on origin. Something must filter it out.
I'll take care of it.

Cyan4973:
https://github.com/facebook/zstd/pull/3954

PR #3943, opened 2024-03-11
Use ZSTD_LEGACY_SUPPORT=5 in "make test"

embg:
Currently `make test` builds without legacy support, while the library builds (by default) with `ZSTD_LEGACY_SUPPORT=5`. This means that the library we are building for tests is slightly different than the library most users are calling into. This PR moves our test build to `ZSTD_LEGACY_SUPPORT=5`.

After this build change, CI jobs found [two](https://github.com/facebook/zstd/blob/372fddf4e6a6db6776b745f31c02a7c8c8dfc83f/lib/legacy/zstd_v02.c#L3444) [violations](https://github.com/facebook/zstd/blob/372fddf4e6a6db6776b745f31c02a7c8c8dfc83f/lib/legacy/zstd_v03.c#L3084) of the static aliasing rule in the legacy codebase. I fixed those bugs and included the fixes in this PR.

There is one failing CI job on my [draft PR](https://github.com/embg/zstd/pull/65), but I believe this is unrelated to my changes, as I am seeing a similar failure on a [dummy PR](https://github.com/embg/zstd/pull/66).

### DIFF
--- a/lib/legacy/zstd_v02.c
+++ b/lib/legacy/zstd_v02.c
@@ -862,7 +862,7 @@ extern "C" {
 *  Streaming functions
 ***************************************/
 
-typedef struct ZSTD_DCtx_s ZSTD_DCtx;
+typedef struct ZSTDv02_Dctx_s ZSTD_DCtx;
 
 /*
   Use above functions alternatively.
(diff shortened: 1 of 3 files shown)


Cyan4973 (approved):
Looks good to me !

PR #3946, opened 2024-03-11
cmake refactor: move HP-UX specific logic into its own function

Cyan4973:
reduce visual clutter from corner case configuration.


Cyan4973:
Note:
only the `cmake` script is modified,

the sanitizer test failures are unrelated (they are CI infrastructure errors).

terrelln (approved):

Issue #3923, opened 2024-03-02
Add common file types that are compressed to ' --exclude-compressed'


dcog989:
https://github.com/facebook/zstd/issues/1579 produced the switch '--exclude-compressed' but that only excludes some archive file types (https://github.com/facebook/zstd/blob/a58b48ef0e543980888a4d9d16c9072ff22135ca/programs/fileio.c#L1897).

Request that common file types that are already compressed are added to the above list. 

A starter list (note: also includes archive formats so some duplication with current code): https://github.com/dcog989/Common-file-types-that-are-compressed/blob/main/compressed-already.yaml

Cyan4973:
feature completed by @daniellerozenblit at #3951

dcog989:
Excellent! That should bring a nice bump in compression times for mixed data sets. Thanks @daniellerozenblit and @Cyan4973.

Yann - any ETA for next release?

Cyan4973:
"soon"

PR #3951, opened 2024-03-12
Add common compressed file types to --exclude-compressed

daniellerozenblit:
This pull request addresses issue #3923, which requests additional common compressed file types to be included in `--exclude-compressed`.


Cyan4973 (approved):
Perfect !

PR #3952, opened 2024-03-12
Stop Hardcoding the POSIX Version on BSDs

### LINKED ISSUE #3748: zstd wrongly uses utime instead of utimensat on BSDs
**Describe the bug**
zstd and unzstd attempts to set the output file's atime and mtime to match the input file's.  This behavior cannot be disabled.  On platforms that support it, unzstd uses `utimensat`.  Otherwise, it uses `utime`.  Unfortunately, `utime` only sets the timestamp's seconds field, not its nanoseconds field.  That matters, because zstd wrongly determines which platforms support `utimensat`.  For the BSDs, it simply hardcodes `PLATFORM_POSIX_VERSION` to a number that's too old (see https://github.com/facebook/zstd/blob/396ef5b434e5e7f15773a7495f374a99a6377778/programs/platform.h#L77).

Tragically, this means that the output file's mtime is almost always a fraction of a second _older_ than the input file's.  So, for example, you can't create a Makefile rule like the below.  make will always execute the command.
```make
foo: foo.zst
        unzstd -fo foo foo.zst
```

**To Reproduce**
```
> touch /tmp/foo
> zstd /tmp/foo
> rm /tmp/foo
> touch /tmp/foo.zst
> unzstd /tmp/foo.zst
> stat -f "%N %Fm" /tmp/foo /tmp/foo.zst
/tmp/foo 1694052370.000000000
/tmp/foo.zst 1694052370.336394000
```
Notice that foo has an _older_ timestamp than foo.zst.  It should be equal or greater.

**Expected behavior**
The output file's timestamps should always be greater than or equal to the input file's.

 - OS: FreeBSD
 - Version: 14.0-CURRENT
 - Compiler: Clang
 - Flags: The defaults. Probably -O2
 - Other relevant hardware specs: None
 - Build system: FreeBSD's builtin build system

felixhandte:
BSDs should all have a `unistd.h` header.

Fixes #3748.

### DIFF
--- a/programs/platform.h
+++ b/programs/platform.h
@@ -74,8 +74,7 @@ extern "C" {
 ***************************************************************/
 #ifndef PLATFORM_POSIX_VERSION
 
-#  if (defined(__APPLE__) && defined(__MACH__)) || defined(__SVR4) || defined(_AIX) || defined(__hpux) /* POSIX.1-2001 (SUSv3) conformant */ \
-     || defined(__DragonFly__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)  /* BSD distros */
+#  if (defined(__APPLE__) && defined(__MACH__)) || defined(__SVR4) || defined(_AIX) || defined(__hpux) /* POSIX.1-2001 (SUSv3) conformant */
      /* exception rule : force posix version to 200112L,
       * note: it's better to use unistd.h's _POSIX_VERSION whenever possible */
 #    define PLATFORM_POSIX_VERSION 200112L


Cyan4973 (approved):

asomers:
Thanks!  Have you checked whether zstd's build system actually uses utimensat now?

felixhandte:
@asomers, I'll admit I haven't. If you have easy access to a BSD machine, which I don't, would you mind checking?

asomers:
I just checked, and it doesn't work :( .  It looks like the easiest thing to do that _would_ work would be simply to hardcode it like this:
```
#  elif defined(__FreeBSD__)
#    define PLATFORM_POSIX_VERSION 200809L
```
I'm doing a full OS build right now and I'll run all of the zstd-related tests.  I haven't checked any other BSDs.

bsdimp:
> I just checked, and it doesn't work :( . It looks like the easiest thing to do that _would_ work would be simply to hardcode it like this:
> 
> ```
> #  elif defined(__FreeBSD__)
> #    define PLATFORM_POSIX_VERSION 200809L
> ```
> 
> I'm doing a full OS build right now and I'll run all of the zstd-related tests. I haven't checked any other BSDs.

I think we would have to have an #elif here like asomers suggests to ensure that we don't fall into defining PLATFORM_POSIX_VERSION to be 0.  Alternatively, you could add the BSDs all to the #if that follows, I think.   Its value is later used to define. FreeBSD's _POSIX_VERSION, though is defined like this:
```
sys/sys/unistd.h:#define	_POSIX_VERSION		200112L
```
so it wouldn't help (we need to fix that since we support at least 200809 and its technical corrections), so you'd still want to do Allan's changes.

bsdimp:
actually, you should just move all the BSD defined() lines down to the next #elif. Then if they support it it will work. FreeBSD should then fix its bug so we start to use it and there's no an extra grotty special case we'll have to change down later.

felixhandte:
@asomers ok it looks like OpenBSD targets POSIX 2008 ([source](https://github.com/openbsd/src/blob/7639492a55b340b1e1529491af78af7a5d3c8ea0/sys/sys/unistd.h#L46)), but FreeBSD and NetBSD only target POSIX 2001 ([source](https://github.com/freebsd/freebsd-src/blob/937a0055858a098027f464abf0b2b1ec5d36748f/sys/sys/unistd.h#L96), [source](https://github.com/NetBSD/src/blob/9b239cc71c861335c51378881b94747bfb1fb4e6/sys/sys/unistd.h#L62)).

felixhandte:
@bsdimp,

Fixing FreeBSD's advertised POSIX compliance would definitely be the best solution.

I am wary of assuming the availability of symbols beyond what POSIX guarantees. (Actually, even assuming that platforms expose the symbols they claim to has proven to be dangerous: we've seen platforms that claim a level of POSIX compliance but then are missing symbols from it.) Even if FreeBSD now supports 2008 fully even though it only advertises 2001, it would only be safe to add a naive exemption if *every* FreeBSD version supported `utimensat()`, which is likely not the case. If there's some version test I can run along the lines of

```
#if defined(__FreeBSD__) && defined(_SOME_FREEBSD_VERSION_MACRO) && _SOME_FREEBSD_VERSION_MACRO >= some_version
#define PLATFORM_POSIX_VERSION 200809L
#elif
...
```

then I would be happy to add this kind of exception. But otherwise, this will break builds on (very) old versions of FreeBSD. Which is not out of the question of someone trying to build on. You would not believe some of the extremely niche and/or old platforms/versions that people try to build Zstd on.

See also the similar case of #2986.

felixhandte:
Actually I do see there's a `__FreeBSD_version` macro. And https://docs.freebsd.org/en/books/porters-handbook/versions/ says `utimensat()` was added in 1100056. So I'll use that.

PR #3954, opened 2024-03-12
Add the zeroSeq sample

Cyan4973:
that should have been part of #3674.

Also: ensure that `*.zst` files in `golden-decompress-errors/` are not filtered out.


embg (approved):
Looks good! Can you make the same gitignore change for `golden-decompression` as well?

terrelln (approved):

terrelln (changes requested):
Looks like the test is trying to decompress `.gitignore`. We should probably update the tests to only attempt to decompress files that end with `.zst`

Cyan4973:
fixed

terrelln (approved):

PR #3957, opened 2024-03-12
[cpu] Backport fix for rbx clobbering on Windows with Clang

terrelln:
Backport folly fix for rbx clobbering: https://github.com/facebook/folly/commit/f22f88b8b9d70160388f0f149bc9abaeb82c250b

We have a CI job that tests this code path: [cmake-visual-2022 (Visual Studio 17 2022, -T ClangCL)](https://github.com/facebook/zstd/actions/runs/8254641561/job/22579288253?pr=3957)

This supercedes PR #3646.

### DIFF
--- a/lib/common/cpu.h
+++ b/lib/common/cpu.h
@@ -35,6 +35,7 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
     U32 f7b = 0;
     U32 f7c = 0;
 #if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))
+#if !defined(__clang__)
     int reg[4];
     __cpuid((int*)reg, 0);
     {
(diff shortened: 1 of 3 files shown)


lib/common/cpu.h
```diff
@@ -50,6 +51,41 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
             f7c = (U32)reg[2];
         }
     }
+#else
+    #error NOCOMMIT: Test to see if we have CI coverage for Windows + clang
```
terrelln: I will delete this line before merging.

This is here to test to see if we have CI coverage for Windows + clang. If we don't, I will attempt to add CI coverage, so we can ensure that this works.

I have reasonably high confidence that it will work, given that it is copied from folly. But it is possible that it doesn't even compile because of a typo.

terrelln: We do have Visual Studios ClangCL coverage: https://github.com/facebook/zstd/actions/runs/8254616105/job/22579195705?pr=3957

Cyan4973 (approved):

terrelln:
We already had CMake build tests on Windows. I've just attempted to also run tests with those builds, so lets see if that works.

PR #3959, opened 2024-03-12
Exercise ZSTD_findDecompressedSize() in the simple decompression fuzzer

embg:
If `ZSTD_decompressDCtx()` succeeds on a buffer, then `ZSTD_findDecompressedSize()` should return either `ZSTD_CONTENTSIZE_UNKNOWN` or a matching decompressed size. This PR adds that invariant to the contract fuzzed by `tests/fuzz/simple_decompress.c`.

Adding that invariant to the fuzzer uncovered a bug. On legacy frames, the decoder doesn't reject on frame headers with incorrect sizes, which violates the above invariant. I fixed that bug and added a unit test for one of the bad inputs discovered by the fuzzer.

Without bugfix (fuzzer):
```
simple_decompress.c: 45: Assertion: `(expectedSize == (0ULL - 1) || expectedSize == dSize)' failed. 
==866880== ERROR: libFuzzer: deadly signal
```

Without bugfix (unit test):
```
test 82 : Decoder should reject invalid frame header on legacy frames: Error => must reject when legacy frame header is invalid (seed 3173672436, test nb 83, line 2416)
```

With bugfix (fuzzer): doesn't crash after 5min

With bugfix (unit test):
```
test 82 : Decoder should reject invalid frame header on legacy frames: OK
```

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -1093,6 +1093,15 @@ size_t ZSTD_decompressMultiFrame(ZSTD_DCtx* dctx,
             decodedSize = ZSTD_decompressLegacy(dst, dstCapacity, src, frameSize, dict, dictSize);
             if (ZSTD_isError(decodedSize)) return decodedSize;
 
+            {
+                unsigned long long const expectedSize = ZSTD_getFrameContentSize(src, srcSize);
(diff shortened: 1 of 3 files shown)


Cyan4973 (approved):
Great catch @embg !

PR #3960, opened 2024-03-12
Use `utimensat()` on FreeBSD

felixhandte:
FreeBSD only claims to support POSIX 2001 [0]. But they do in fact support `utimensat()`. This adds a specific check to opt them in to using it. This value was selected by consulting [1].

See discussion on #3952.

Further addresses #3748.

[0] https://github.com/freebsd/freebsd-src/blob/937a0055858a098027f464abf0b2b1ec5d36748f/sys/sys/unistd.h#L96

[1] https://docs.freebsd.org/en/books/porters-handbook/versions/

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -23,16 +23,27 @@ extern "C" {
 #include <errno.h>
 #include <assert.h>
 
+#if defined(__FreeBSD__)
+#include <sys/param.h> /* __FreeBSD_version */
+#endif /* #ifdef __FreeBSD__ */
+
 #if defined(_WIN32)
 #  include <sys/utime.h>  /* utime */
 #  include <io.h>         /* _chmod */
+#  define ZSTD_USE_UTIMENSAT 0
 #else
 #  include <unistd.h>     /* chown, stat */
-#  if PLATFORM_POSIX_VERSION < 200809L || !defined(st_mtime)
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
I wonder if it would be useful to put this explanation (and documentation link) in code comment too.

programs/util.c
```diff
@@ -259,7 +259,8 @@ int UTIL_utime(const char* filename, const stat_t *statbuf)
      * that struct stat has a struct timespec st_mtim member. We need this
      * check because there are some platforms that claim to be POSIX 2008
      * compliant but which do not have st_mtim... */
-#if (PLATFORM_POSIX_VERSION >= 200809L) && defined(st_mtime)
+#if (((PLATFORM_POSIX_VERSION >= 200809L) && defined(st_mtime)) \
+    || (defined(__FreeBSD__) && __FreeBSD_version >= 1100056))
```
terrelln: Looks like you need to move the `<sys/param.h>` include on line 1550 up to the top of the file, so you have access to the `__FreeBSD_version` define.

terrelln (changes requested):

terrelln (approved):
Nice! LGTM assuming that last FreeBSD test passes

PR #3962, opened 2024-03-13
reduce the amount of #include in cover.h

Cyan4973:
Initially, I was hunting the presence of direct invocations of `malloc()`,
in order to change them into invocations of the user-definable custom allocator,
and discovered that `cover.c` has many such instances.

Unfortunately, this goal will require major code modifications, and this is too much effort or complexity at this point before a release.

But a side effect of this investigation worth merging is a significant reduction in the amount of `#include` within `"cover.h"`, which are then transitively included into any other unit including `cover.h`. It previously included standard headers such as `<stdlib.h>`, `<stdio.h>` or `<time.h>` that we typically _don't_ want to invoke directly nor even just see in any other parts of `libzstd`. The initial list was likely made for the implementation `cover.c`, and is way overkill for the declaration.

### DIFF
--- a/lib/compress/fse_compress.c
+++ b/lib/compress/fse_compress.c
@@ -25,7 +25,7 @@
 #include "../common/error_private.h"
 #define ZSTD_DEPS_NEED_MALLOC
 #define ZSTD_DEPS_NEED_MATH64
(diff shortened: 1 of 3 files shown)


terrelln (approved):
Nice cleanup

PR #3964, opened 2024-03-13
Promote `ZSTD_c_targetCBlockSize` Parameter to Stable API

felixhandte:
This feature has demonstrated itself to be useful in web compression and we want to encourage other folks to use it. But we currently make it difficult to do so since it's locked away in the experimental API.

The API itself is really straightforward and I think it's fine to commit to maintaining support / compatibility for this API even if in the future the underlying implementation may continue to evolve.

Note that this commit changes its enum name and also its numeric value. Users who respected the instructions of using the experimental API should be fine with both of these changes since they should only have referred to it by the.

Conceivably someone could have done bad feature detection of this capability by doing `#ifdef ZSTD_c_targetCBlockSize` which will now return false since it's no longer a macro... but I think that's an acceptable hypothetical breakage.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -390,6 +390,16 @@ typedef enum {
                               * The higher the value of selected strategy, the more complex it is,
                               * resulting in stronger and slower compression.
                               * Special: value 0 means "use default strategy". */
+    ZSTD_c_targetCBlockSize=130, /* Tries to fit compressed block size to be
+                                  * around targetCBlockSize. No target when
+                                  * targetCBlockSize == 0. There is no guarantee
+                                  * on compressed block size (default:0).
+                                  * Since the decoder has to buffer a complete
+                                  * block to begin decoding it, in low band-
+                                  * width streaming environments this may
+                                  * improve end-to-end latency. Bound by
+                                  * ZSTD_TARGETCBLOCKSIZE_MIN and
+                                  * ZSTD_TARGETCBLOCKSIZE_MAX. */
     /* LDM mode parameters */
     ZSTD_c_enableLongDistanceMatching=160, /* Enable long distance matching.
                                      * This parameter is designed to improve compression ratio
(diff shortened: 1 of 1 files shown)


terrelln:
One question I was just thinking about when writing the [memory constrained wiki](https://github.com/facebook/zstd/wiki/Using-Zstandard-in-a-memory-constrained-environment): How does `ZSTD_c_targetBlockSize` interact with `ZSTD_c_maxBlockSize`?

I would expect that they would work together, and enforce that the block is not larger than `ZSTD_c_maxBlockSize`, since the max block size is enforced by passing in less data to match finder, and the target block size works on the sequences. But I just want to double check that that is correct.

Cyan4973:
`ZSTD_c_maxBlockSize` cuts _before_ compression,
so the value is compared to decompressed size, and it's a hard maximum,

`ZSTD_c_targetCBlockSize` cuts _after_ compression,
so the value is compared to compressed size, and it's a loose objective, more like an average to converge towards.

Since we don't fuse compressed blocks at this point,
any block produced after `ZSTD_c_maxBlockSize` results in at least one compressed block.

On the other hand, a block cut before compression can still result in multiple compressed blocks due to `ZSTD_c_targetCBlockSize`.

terrelln:
> Since we don't fuse compressed blocks at this point, any block produced after ZSTD_c_maxBlockSize results in at least one compressed block.

It might be good to specify somewhere that this isn't just an accident of implementation, but that `ZSTD_c_targetBlockSize` must respect `ZSTD_c_maxBlockSize`. This is necessary because the decoder may also be using `ZSTD_d_maxBlockSize`.

Cyan4973:
> It might be good to specify somewhere that this isn't just an accident of implementation, but that `ZSTD_c_targetBlockSize` must respect `ZSTD_c_maxBlockSize`. This is necessary because the decoder may also be using `ZSTD_d_maxBlockSize`.

These are orthogonal parameter.

If `ZSTD_c_targetCBlockSize` is larger than `ZSTD_c_maxBlockSize`, then it will simply not be effective, since each block is already by construction smaller than the target size. It's not a bug, it's just an application of their rules.

`ZSTD_c_targetCBlockSize` also cannot be used to satisfy `ZSTD_d_maxBlockSize`.
To begin with, `ZSTD_c_targetCBlockSize` is approximative, and offers no maximum guarantee, required by `ZSTD_d_maxBlockSize`. But more importantly, `ZSTD_c_targetCBlockSize` is about the compressed size, while `ZSTD_d_maxBlockSize`, like `ZSTD_c_maxBlockSize`, are about the decompressed sizes. Entirely different metrics.

PR #3965, opened 2024-03-14
[cmake] Always create libzstd target

### LINKED ISSUE #3859: Export "selected" CMake target for zstd
Can we have a consistent CMake library target exported?

For example with the curl project, if the standard CMake global variable `BUILD_SHARED_LIBS` is set, `LIB_SELECTED` will match accordingly: https://github.com/curl/curl/blob/master/CMakeLists.txt#L155-L159

```
# lib flavour selected for example and test programs.
if(BUILD_SHARED_LIBS)
  set(LIB_SELECTED ${LIB_SHARED})
else()
  set(LIB_SELECTED ${LIB_STATIC})
endif()
```

This is then used as an exported target (in addition to the fixed static/shared targets):
```
add_library(${LIB_NAME} ALIAS ${LIB_SELECTED})
add_library(${PROJECT_NAME}::${LIB_NAME} ALIAS ${LIB_SELECTED})
```

This makes pulling in curl quite clean. In comparison, I have to do the following when using `fetchcontent` with zstd:
```
set(ZSTD_LIBRARY libzstd_static)
if(BUILD_SHARED_LIBS)
  set(ZSTD_LIBRARY libzstd_shared)
endif()
```

terrelln:
If both `ZSTD_BUILD_SHARED` and `ZSTD_BUILD_STATIC` are set, then cmake exports the libraries `libzstd_shared` and `libzstd_static` only. It does not export `libzstd`, which is only exported when exactly one of `ZSTD_BUILD_SHARED` and `ZSTD_BUILD_STATIC` is set. This PR exports `libzstd` in that case, based on the value of the standard CMake variable [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html). This ensures that `libzstd` can always be used to refer to the exported zstd library, since the build errors if neither `ZSTD_BUILD_SHARED` nor `ZSTD_BUILD_STATIC` are set.

I tested all the possible combinations of `ZSTD_BUILD_SHARED`, `ZSTD_BUILD_STATIC`, and `BUILD_SHARED_LIBS` and they always worked as expected:
* If only exactly one of `ZSTD_BUILD_SHARED` and `ZSTD_BUILD_STATIC` is set, that is used as `libzstd`.
* Otherwise, libzstd is set based on `BUILD_SHARED_LIBS`.

Fixes #3859.

### DIFF
--- a/build/cmake/lib/CMakeLists.txt
+++ b/build/cmake/lib/CMakeLists.txt
@@ -159,6 +159,20 @@ if (ZSTD_BUILD_STATIC AND NOT ZSTD_BUILD_SHARED)
     target_link_libraries(libzstd INTERFACE libzstd_static)
     list(APPEND library_targets libzstd)
 endif ()
+if (ZSTD_BUILD_SHARED AND ZSTD_BUILD_STATIC)
+    # If both ZSTD_BUILD_SHARED and ZSTD_BUILD_STATIC are set, which is the
+    # default, fallback to using BUILD_SHARED_LIBS to determine whether to
+    # set libzstd to static or shared.
+    if (BUILD_SHARED_LIBS)
+        add_library(libzstd INTERFACE)
+        target_link_libraries(libzstd INTERFACE libzstd_shared)
+        list(APPEND library_targets libzstd)
+    else ()
+        add_library(libzstd INTERFACE)
+        target_link_libraries(libzstd INTERFACE libzstd_static)
+        list(APPEND library_targets libzstd)
+    endif ()
+endif ()
 
 # Add specific compile definitions for MSVC project
 if (MSVC)


Cyan4973 (approved):

teo-tsirpanis:
Sorry, just saw it. The problem is that we have two variables in `BUILD_SHARED_LIBS` and `ZSTD_BUILD_SHARED` that have a very similar meaning yet different semantics.

Ideally `ZSTD_BUILD_SHARED` and `ZSTD_BUILD_STATIC` should be deprecated and replaced by `BUILD_SHARED_LIBS` and the still non-standard `BUILD_STATIC_LIBS`, just like lz4. Even more ideally, we should be building only one flavor of zstd at a time, controlled by the value of `BUILD_SHARED_LIBS`.

terrelln:
@teo-tsirpanis I don't want to remove `ZSTD_BUILD_SHARED`, and `ZSTD_BUILD_STATIC`, since that will break users. I'm okay with breaking users on a major library version bump, but then I would want to do a full overhaul of the CMake build system, since it is mostly from contributors and there are quite a few things to be cleaned up.

This PR achieves the goal of allowing exporting one target `libzstd`, independent of the settings of `ZSTD_BUILD_SHARED`, `ZSTD_BUILD_STATIC`, and `BUILD_SHARED_LIBS`. Additionally, if you set `BUILD_SHARED_LIBS`, and don't modify the values for `ZSTD_BUILD_SHARED`, and `ZSTD_BUILD_STATIC`, you'll get the right behavior (but you will build both libraries). And visa-versa if you set `ZSTD_BUILD_SHARED`, and `ZSTD_BUILD_STATIC` explicitly, you won't be impacted by `BUILD_SHARED_LIBS`.

This seems like a reasonable middle ground, that produces sane behavior. But I'm by no means a CMake expert, so please let me know what you think, given the constraints.

teo-tsirpanis:
My concern is that if you configure with `-DZSTD_BUILD_SHARED=OFF -DZSTD_BUILD_STATIC=ON -DBUILD_SHARED_LIBS=ON`, it will fail, which might or might not be important enough to care. As I am imagining it, `ZSTD_BUILD_***` would not break, but specifying them will emit a deprecation warning, suggesting the user to switch to `BUILD_***_LIBS`.

terrelln:
I won't fail, it will produce a static library, because the `ZSTD_` variables take precedence over `BUILD_SHARED_LIBS`. I think this is an acceptable tradeoff, because these are the more specific variables to zstd.

But, I will put up a PR to document that behavior.

PR #3971, opened 2024-03-15
Fix bug with streaming decompression of magicless format

embg:
This is the first of two PRs related to fuzzing the magicless format. I'm splitting up the changes into two PRs to make review easier.

Under certain conditions, `ZSTD_decompressStream()` uses single-shot decompression to improve performance. The implementation of this optimization is currently [incorrect](https://github.com/facebook/zstd/blob/0fcdc62c2d13f4d978ecba703c7d381ce97c97f1/lib/decompress/zstd_decompress.c#L2181) for magicless format as it relies on `ZSTD_findFrameCompressedSize()`, which only supports the standard format.

The obvious fix is to just disable the optimization for magicless format, as we don't care about streaming decompression performance on magicless frames. Unfortunately, for annoying reasons, this greatly complicates the fuzzing situation.

So instead of disabling the optimization, this PR makes it properly support magicless frames by adding a private function `ZSTD_findFrameCompressedSize_advanced()`. The new function fixes the bug by taking the frame format as a parameter.

Test coverage check:
```
[embg@server ~/repos/zstd/tests (compressed_size_advanced)]$ git diff
diff --git a/lib/decompress/zstd_decompress.c b/lib/decompress/zstd_decompress.c
index b0a00396..e8b3ed5f 100644
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -2183,6 +2183,9 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
                 && zds->fParams.frameType != ZSTD_skippableFrame
                 && (U64)(size_t)(oend-op) >= zds->fParams.frameContentSize) {
                 size_t const cSize = ZSTD_findFrameCompressedSize_advanced(istart, (size_t)(iend-istart), zds->format);
+                if (zds->format == ZSTD_f_zstd1_magicless) {
+                    abort();
+                }
                 if (cSize <= (size_t)(iend-istart)) {
                     /* shortcut : using single-pass mode */
                     size_t const decompressedSize = ZSTD_decompress_usingDDict(zds, op, (size_t)(oend-op), istart, cSize, ZSTD_getDDict(zds));

[embg@server ~/repos/zstd/tests (compressed_size_advanced)]$ make -j zstreamtest
[embg@server ~/repos/zstd/tests (compressed_size_advanced)]$ ./zstreamtest -v
test 83 : Test single-shot fallback for magicless mode: Aborted (core dumped)
```

Note: after v1.5.6, I will come back and refactor `ZSTD_findFrameSizeInfo()` to avoid duplication of logic with `ZSTD_getFrameHeader_advanced()`. Will file an issue for this when I submit my second magicless fuzzing PR.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -729,17 +729,17 @@ static ZSTD_frameSizeInfo ZSTD_errorFrameSizeInfo(size_t ret)
     return frameSizeInfo;
 }
 
-static ZSTD_frameSizeInfo ZSTD_findFrameSizeInfo(const void* src, size_t srcSize)
+static ZSTD_frameSizeInfo ZSTD_findFrameSizeInfo(const void* src, size_t srcSize, ZSTD_format_e format)
 {
     ZSTD_frameSizeInfo frameSizeInfo;
     ZSTD_memset(&frameSizeInfo, 0, sizeof(ZSTD_frameSizeInfo));
 
 #if defined(ZSTD_LEGACY_SUPPORT) && (ZSTD_LEGACY_SUPPORT >= 1)
-    if (ZSTD_isLegacy(src, srcSize))
+    if (format == ZSTD_f_zstd1 && ZSTD_isLegacy(src, srcSize))
         return ZSTD_findFrameSizeInfoLegacy(src, srcSize);
 #endif
 
-    if ((srcSize >= ZSTD_SKIPPABLEHEADERSIZE)
+    if (format == ZSTD_f_zstd1 && (srcSize >= ZSTD_SKIPPABLEHEADERSIZE)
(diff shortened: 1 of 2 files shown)


lib/decompress/zstd_decompress_internal.h
```diff
@@ -236,5 +236,14 @@ size_t ZSTD_loadDEntropy(ZSTD_entropyDTables_t* entropy,
  *  This function cannot fail. */
 void ZSTD_checkContinuity(ZSTD_DCtx* dctx, const void* dst, size_t dstSize);
 
+/*! ZSTD_findFrameCompressedSize_advanced() :
```
Cyan4973: Why does it need to be published ?
Could it remain a `static` unit-internal function ?

terrelln (approved):

PR #3976, opened 2024-03-18
Fuzzing and bugfixes for magicless-format decoding

embg:
### Overview
Follow-up to https://github.com/facebook/zstd/pull/3971.

This PR adds fuzzer coverage for magicless-format decoding. The fuzzers exposed several bugs (instances where the decoder rejects valid frames), which are also fixed in this PR. I constructed an [example](https://gist.github.com/embg/9940726094f4cf2cef162cffe9319232) which proves that the reference encoder can sometimes produce frames that hit these bugs.

*Note that these bugs only affect users of the magicless format. If your code doesn't set parameters `ZSTD_c_format` and `ZSTD_d_format`, you are not affected.*

### Bugfixes
See changes in `zstd_decompress.c`. The bugs fall into three categories:
* Frames that happen to start with legacy magic are rejected
* Frames that happen to start with skippable magic are rejected
* The bug fixed in https://github.com/facebook/zstd/pull/3971, which is entirely separate

### Fuzzing setup
At a high level, the fuzzing setup is as follows:
* Fuzzer picks a random buffer, which we interpret as magicless compressed data
* From the magicless compressed data, we create an equivalent standard-format compressed data by copying to a new buffer and prepending `ZSTD_MAGICNUMBER`.
* We decompress using both the standard-format decoder and the magicless-format decoder.
* We enforce that if the standard decoder accepts, the magicless decoder must also accept.

The idea is that if a frame is valid, then the standard decoder should accept it, in which case we will enforce that the magicless decoder should also accept it.

### Fuzzing limitations
We exercise both one-shot decoding and streaming decoding. One limitation is that we only do a single invocation of `ZSTD_decompressStream()`, which tests the most common *streaming* code path for magicless format: input fits in a single `ZSTD_inBuffer`, output fits in a single `ZSTD_outBuffer`. This is the common case because magicless format is all about tiny data.

A future improvement here could be to test other calling patterns of `ZSTD_decompressStream()`. I spent a bit of time trying to add this, found that it would require non-trivial effort, and decided to leave it out of scope.

Another limitation: the fuzzer can only test exactly one frame at a time. We don't test zero frames or 2+ frames. This would require a completely different approach, and isn't the common case for users of the magicless format.

### Fuzzer coverage checks
I made sure that each of these changes in `zstd_decompress.c` individually cause the fuzzer to crash:
```
-        if (dctx->format == ZSTD_f_zstd1 && ZSTD_isLegacy(src, srcSize)) {
+        if (/*dctx->format == ZSTD_f_zstd1 &&*/ ZSTD_isLegacy(src, srcSize)) {
```
```
-        if (dctx->format == ZSTD_f_zstd1 && srcSize >= 4) {
+        if (/*dctx->format == ZSTD_f_zstd1 &&*/ srcSize >= 4) {
```
```
-                size_t const cSize = ZSTD_findFrameCompressedSize_advanced(istart, (size_t)(iend-istart), zds->format);
+                size_t const cSize = ZSTD_findFrameCompressedSize(istart, (size_t)(iend-istart));
```
```
-            if (zds->format == ZSTD_f_zstd1
-                && (MEM_readLE32(zds->headerBuffer) & ZSTD_MAGIC_SKIPPABLE_MASK) == ZSTD_MAGIC_SKIPPABLE_START) {  /* skippable frame */
+            if ((MEM_readLE32(zds->headerBuffer) & ZSTD_MAGIC_SKIPPABLE_MASK) == ZSTD_MAGIC_SKIPPABLE_START) {  /* skippable frame */
```

### DIFF
--- a/doc/decompressor_errata.md
+++ b/doc/decompressor_errata.md
@@ -125,3 +125,24 @@ The total `Block_Content` is `5` bytes, and `Last_Table_Offset` is `2`.
 See the compressor workaround code:
 
 https://github.com/facebook/zstd/blob/8814aa5bfa74f05a86e55e9d508da177a893ceeb/lib/compress/zstd_compress.c#L2667-L2682
+
+Magicless format
+----------------------
+
+**Last affected version**: v1.5.5
+
+**Affected decompressor component(s)**: Library & CLI
+
+**Produced by the reference compressor**: Yes (example: https://gist.github.com/embg/9940726094f4cf2cef162cffe9319232)
+
+**Example Frame**: `27 b5 2f fd 00 03 19 00 00 66 6f 6f 3f ba c4 59`
+
+v1.5.6 fixes several bugs in which the magicless-format decoder rejects valid frames.
+These include but are not limited to:
+* Valid frames that happen to begin with a legacy magic number (little-endian)
+* Valid frames that happen to begin with a skippable magic number (little-endian)
+
+If you are affected by this issue and cannot update to v1.5.6 or later, there is a
+workaround to recover affected data. Simply prepend the ZSTD magic number
+`0xFD2FB528` (little-endian) to your data and decompress using the standard-format
+decoder.
--- a/doc/decompressor_errata.md
+++ b/doc/decompressor_errata.md
@@ -131,7 +131,7 @@ Magicless format
 
 **Last affected version**: v1.5.5
 
-**Affected decompressor component(s)**: Library & CLI
+**Affected decompressor component(s)**: Library
 
 **Produced by the reference compressor**: Yes (example: https://gist.github.com/embg/9940726094f4cf2cef162cffe9319232)
 
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -124,7 +124,8 @@ FUZZ_TARGETS :=       \
 	sequence_compression_api \
 	seekable_roundtrip \
 	huf_round_trip \
-	huf_decompress
+	huf_decompress \
+	decompress_cross_format
 
 all: libregression.a $(FUZZ_TARGETS)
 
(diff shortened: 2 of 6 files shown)


tests/fuzz/Makefile
```diff
@@ -124,7 +124,8 @@ FUZZ_TARGETS :=       \
 	sequence_compression_api \
 	seekable_roundtrip \
 	huf_round_trip \
-	huf_decompress
+	huf_decompress \
+	frame_header_cross_format
```
terrelln: This is out of date, likely what is causing the build issues

embg: Thanks, good catch!

embg:
The CI jobs are failing because we added a new fuzz target, but that fuzz target (because it's new) doesn't have a corpus .zip file yet:
```
     0K . ............ ..... ................ .......... .make[1]: *** [Makefile:251: corpora/decompress_cross_format_seed_corpus.zip] Error 8
```
Solution is to merge the PR, let OSS fuzz build the corpus, then jobs will be green again. This is annoying, but we add new targets so infrequently that it's not a big deal.

terrelln (commented):
Looks good once you fix the tests

doc/decompressor_errata.md
```diff
@@ -125,3 +125,24 @@ The total `Block_Content` is `5` bytes, and `Last_Table_Offset` is `2`.
 See the compressor workaround code:
 
 https://github.com/facebook/zstd/blob/8814aa5bfa74f05a86e55e9d508da177a893ceeb/lib/compress/zstd_compress.c#L2667-L2682
+
+Magicless format
+----------------------
+
+**Last affected version**: v1.5.5
+
+**Affected decompressor component(s)**: Library & CLI
```
Cyan4973: Not the CLI,
`zstd` CLI is unable to process magicless frames

doc/decompressor_errata.md
```diff
@@ -125,3 +125,24 @@ The total `Block_Content` is `5` bytes, and `Last_Table_Offset` is `2`.
...
+**Produced by the reference compressor**: Yes (example: https://gist.github.com/embg/9940726094f4cf2cef162cffe9319232)
+
+**Example Frame**: `27 b5 2f fd 00 03 19 00 00 66 6f 6f 3f ba c4 59`
+
+v1.5.6 fixes several bugs in which the magicless-format decoder rejects valid frames.
+These include but are not limited to:
+* Valid frames that happen to begin with a legacy magic number (little-endian)
+* Valid frames that happen to begin with a skippable magic number (little-endian)
+
+If you are affected by this issue and cannot update to v1.5.6 or later, there is a
```
Cyan4973, line 145: 👍

Cyan4973 (approved):
I only have one minor correction for the `decompressor_errata` document,
apart from that it's pretty good !

PR #3977, opened 2024-03-18
Doc update

Cyan4973:
updated documentation on corner cases:
- behavior of CCtx and DCtx after an error
- behavior of `targetCBlockSize` prior to v1.5.6


Cyan4973:
weird, why are there errors triggered in "CLI tests" for a documentation-only change ?

PR #3978, opened 2024-03-18
Fix bugs in simple decompression fuzzer

yoniko:
Fixes 2 issue in `simple_decompress.c`:

1. Wrong type used for storing the results of `ZSTD_findDecompressedSize` resulting in never matching to `ZSTD_CONTENTSIZE_ERROR` or `ZSTD_CONTENTSIZE_UNKNOWN`.
2. Experimental API is used (`ZSTD_findDecompressedSize`) without defining `ZSTD_STATIC_LINKING_ONLY`.

PR #3979, opened 2024-03-18
Fail on errors when building fuzzers

yoniko:
Fails on errors when building fuzzers with `fuzz.py` (adds `Werror`).
Currently allows `declaration-after-statement`, `c++-compat` and `deprecated` as they are abundant in code (some fixes to declaration-after-statement are presented in this commit).

Would help avoid issues like those solved in #3978.

### DIFF
--- a/tests/fuzz/fuzz_data_producer.c
+++ b/tests/fuzz/fuzz_data_producer.c
@@ -28,11 +28,11 @@ void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer) { free(producer); }
 
 uint32_t FUZZ_dataProducer_uint32Range(FUZZ_dataProducer_t *producer, uint32_t min,
                                   uint32_t max) {
-    FUZZ_ASSERT(min <= max);
-
     uint32_t range = max - min;
     uint32_t rolling = range;
     uint32_t result = 0;
+
+    FUZZ_ASSERT(min <= max);
 
     while (rolling > 0 && producer->size > 0) {
       uint8_t next = *(producer->data + producer->size - 1);
@@ -79,11 +79,11 @@ int FUZZ_dataProducer_empty(FUZZ_dataProducer_t *producer) {
 
 size_t FUZZ_dataProducer_contract(FUZZ_dataProducer_t *producer, size_t newSize)
 {
-    newSize = newSize > producer->size ? producer->size : newSize;
+    const size_t effectiveNewSize = newSize > producer->size ? producer->size : newSize;
 
-    size_t remaining = producer->size - newSize;
+    size_t remaining = producer->size - effectiveNewSize;
     producer->data = producer->data + remaining;
-    producer->size = newSize;
+    producer->size = effectiveNewSize;
     return remaining;
 }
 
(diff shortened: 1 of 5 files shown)


yoniko:
UPDATE:
All checks are green, this should be ready for review

Cyan4973 (approved):

embg (approved):
This is a great improvement. Thank you for improving the fuzzer build to save us time root causing issues that could have been caught at compile time.

My only comment: I don't agree with the random `declaration-after-statement` fixes since they don't add value and potentially can break coverage. They don't add value since there is no reason that our fuzzer code should comply with C89 rules.

I did review the changes carefully and they all look correct to me, so we could merge them, but we could also... not merge them. In general I don't support touching the fuzzer code for no reason since if we do break coverage, realistically we won't find out for a long time (if ever).

I'm very concerned, in general, about the risk of losing coverage by touching old fuzzer code. We don't currently have a way to test if a change reduces coverage.

At the end of the day this is a matter of style, if you do feel that the changes add value, I will not block you from merging them. Just sharing my opinion :)

tests/fuzz/fuzz_data_producer.c
```diff
@@ -28,12 +28,12 @@ void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer) { free(producer); }
 
 uint32_t FUZZ_dataProducer_uint32Range(FUZZ_dataProducer_t *producer, uint32_t min,
                                   uint32_t max) {
-    FUZZ_ASSERT(min <= max);
```
embg, line 31: This change actually reduces readability of the code. It is more readable to assert `min <= max` before using that property for `range = max - min`.

The only benefit of this change is that we get closer to complying with C89 rules... but why?

yoniko:
> This is a great improvement. Thank you for improving the fuzzer build to save us time root causing issues that could have been caught at compile time.
> 
> My only comment: I don't agree with the random `declaration-after-statement` fixes since they don't add value and potentially can break coverage. They don't add value since there is no reason that our fuzzer code should comply with C89 rules.
> 
> I did review the changes carefully and they all look correct to me, so we could merge them, but we could also... not merge them. In general I don't support touching the fuzzer code for no reason since if we do break coverage, realistically we won't find out for a long time (if ever).
> 
> I'm very concerned, in general, about the risk of losing coverage by touching old fuzzer code. We don't currently have a way to test if a change reduces coverage.
> 
> At the end of the day this is a matter of style, if you do feel that the changes add value, I will not block you from merging them. Just sharing my opinion :)

We've aligned on this offline, to summarize: 

The "random" fixes aren't no random, they fix one of the fuzzers (simple_decomress) to build with no warnings even without the `Werror-no` flags. I've introduced them before it was clear the problem is more widespread and `Werror-no` flags are needed.

Having said that, I agree with the comment and don't think we should uphold the fuzzers to irrelevant part of the C89 standard. Given the changes are already there and have been tested and gone through CR I lean toward just merging them but not making any additional effort to resolve `declaration-after-statement`.

Cyan4973:
The requirement to be `C90` compliant is for the `libzstd` library and the `zstd` CLI.

The test tools don't have to follow the same standard as strictly.

PR #3981, opened 2024-03-19
Fix & fuzz ZSTD_generateSequences

terrelln:
This function was seriously flawed:
* It didn't do output bounds checks
* It produced invalid sequences when an uncompressed or RLE block was emitted
* It produced invalid sequences when the block splitter was enabled
* It produced invalid sequences when ZSTD_c_targetCBlockSize was enabled

I've attempted to fix these issues, but this function is just a bad idea, so I've marked it as deprecated and unsafe. We should replace it with `ZSTD_extractSequences()` which operates on a compressed frame.

This PR only changes logic that is called when `zc->seqCollector.collectSequences == 1`, which is only true for `ZSTD_generateSequences()`, so it should be safe to land even right before a release, since this function was already seriously broken.

### DIFF
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -3361,29 +3361,38 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
     return ZSTDbss_compress;
 }
 
-static void ZSTD_copyBlockSequences(ZSTD_CCtx* zc)
+static size_t ZSTD_copyBlockSequences(SeqCollector* seqCollector, const seqStore_t* seqStore, const U32 prevRepcodes[ZSTD_REP_NUM])
 {
-    const seqStore_t* seqStore = ZSTD_getSeqStore(zc);
-    const seqDef* seqStoreSeqs = seqStore->sequencesStart;
-    size_t seqStoreSeqSize = seqStore->sequences - seqStoreSeqs;
-    size_t seqStoreLiteralsSize = (size_t)(seqStore->lit - seqStore->litStart);
-    size_t literalsRead = 0;
-    size_t lastLLSize;
+    const seqDef* inSeqs = seqStore->sequencesStart;
+    const size_t nbInSequences = seqStore->sequences - inSeqs;
+    const size_t nbInLiterals = (size_t)(seqStore->lit - seqStore->litStart);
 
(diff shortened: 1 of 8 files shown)


terrelln:
I need `ZSTD_extractSequences()` for D53496856 anyways, so I will look into what implementing it would look like.

That would give us a candidate for a replacement, so we can delete this function quickly, which will simplify the compression code.

embg:
Would be a good idea to update CHANGELOG as well since the v1.5.6 changes have already been added.

Cyan4973:
> this function is just a bad idea, so I've marked it as deprecated and unsafe. We should replace it with ZSTD_extractSequences() which operates on a compressed frame.

I agree.
`ZSTD_extractSequences()` can actually be useful.

Note that it will also force the topic of representing special blocks, which the current model, focused on Sequences, is unable to do. But that's actually a good question to solve, and the learning might later lead to an evolution of `ZSTD_compressSequences()`.

lib/compress/zstd_compress.c
```diff
@@ -3361,29 +3361,38 @@ static size_t ZSTD_buildSeqStore(ZSTD_CCtx* zc, const void* src, size_t srcSize)
...
+    /* Bounds check that we have enough space for every input sequence
+     * and the block delimiter
+     */
+    assert(seqCollector->seqIndex <= seqCollector->maxSequences);
+    RETURN_ERROR_IF(
+        nbOutSequences > (size_t)(seqCollector->maxSequences - seqCollector->seqIndex),
+        dstSize_tooSmall,
+        "Not enough space to copy sequences");
+
+    memcpy(&repcodes, prevRepcodes, sizeof(repcodes));
```
Cyan4973: `memcpy` -> `ZSTD_memcpy`

terrelln:
> Note that it will also force the topic of representing special blocks, which the current model, focused on Sequences, is unable to do. But that's actually a good question to solve, and the learning might later lead to an evolution of ZSTD_compressSequences().

The simplest thing would be to emit raw blocks as just a block delimiter with the last-literals size = block size, and RLE blocks as a single offset=1, ll=1, ml=block_size - 1 sequence (or a raw block if block_size <= 3).

That would preserve the current API, but there may be a better way if we want to change things up.

Cyan4973 (approved):

VVD:
How to replace `ZSTD_generateSequences` with `ZSTD_extractSequences`?
Can't find `ZSTD_extractSequences` in `zstd.h`.
This change break build of consumers: https://github.com/LWJGL/lwjgl3/issues/968

PR #3983, opened 2024-03-21
Fix -Werror=pointer-arith in fuzzers

embg:
Should fix broken CI jobs.


yoniko (approved):
Looks good, thank you for fixing.
Any idea why it didn't trigger an error in GH CI?

Cyan4973 (approved):

embg:
> Looks good, thank you for fixing. Any idea why it didn't trigger an error in GH CI?

@yoniko The failure is in a file added by my magicless fuzzing PR. I ran CI on that PR before you merged the -Werror change (and vice versa for your PR), so we didn't see red signal on dev until after both PRs merged. I'm 99% sure that's what happened.

PR #3994, opened 2024-03-27
Use md5sum rather than gmd5sum for FreeBSD

sunpoet:
FreeBSD has md5sum in the base system. It also avoids the [hack for gmd5sum](https://github.com/freebsd/freebsd-ports/blob/main/archivers/zstd/Makefile#L22).

Reference:	https://man.freebsd.org/cgi/man.cgi?query=md5sum

### DIFF
--- a/lib/libzstd.mk
+++ b/lib/libzstd.mk
@@ -213,8 +213,6 @@ ifeq ($(UNAME), Darwin)
   ifeq ($(shell md5 < /dev/null > /dev/null; echo $$?), 0)
     HASH ?= md5
   endif
-else ifeq ($(UNAME), FreeBSD)
-  HASH ?= gmd5sum
 else ifeq ($(UNAME), NetBSD)
   HASH ?= md5 -n
 else ifeq ($(UNAME), OpenBSD)
--- a/tests/cli-tests/common/platform.sh
+++ b/tests/cli-tests/common/platform.sh
@@ -18,7 +18,6 @@ esac
 
 case "$UNAME" in
   Darwin) MD5SUM="md5 -r" ;;
-  FreeBSD) MD5SUM="gmd5sum" ;;
   NetBSD) MD5SUM="md5 -n" ;;
   OpenBSD) MD5SUM="md5" ;;
   *) MD5SUM="md5sum" ;;
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -123,7 +123,6 @@ esac
 
 case "$UNAME" in
   Darwin) MD5SUM="md5 -r" ;;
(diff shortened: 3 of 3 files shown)


Cyan4973:
I wonder why there was a contribution to employ `gmd5sum` on FreeBSD, at a time when `md5sum` was our default ?
It implied that `md5sum` was not a good default for `FreeBSD` at the time ?
(Note: this contribution was made by @DimitryAndric in #475. Feel free to weight in)

Also, looking around on this topic, I see this sentence : 
> Use md5 on FreeBSD. The standard command for calculating MD5 checksums on FreeBSD systems is md5, not md5sum or gmd5sum. 

And now I'm even more confused.

DimitryAndric:
I think #475 was because we didn't have a GNU compatible `md5sum` command in base at the time, but only a slightly incompatible `md5` command (it produces md5 hashes obviously, but its output is different than GNU md5sum's).

Now all supported FreeBSD versions (I think from 13.x onwards) do have a `md5sum` command that behaves similarly to the GNU version.

dearblue:
For reference.
The `md5sum` command was introduced with the `md5 -r` switch on FreeBSD in 13.1 (2022-05-16) or 14.0 (2023-11-20).

  - <https://www.freebsd.org/releases/13.1R/relnotes/#userland-programs>
  - <https://www.freebsd.org/releases/14.0R/relnotes/#mfc-highlights>
  - https://github.com/freebsd/freebsd-src/commit/086feed850c31c278f25c958b97992d024139896

In addition, currently supported FreeBSD versions are 14.0, 13.3 and 13.2 (end of life is 2024-06-30).

Finally, a small difference between `md5sum` and `md5 -r` is quoted from the manual.

<details>
<summary>difference between `md5sum` and `md5 -r`</summary>

from: <https://man.freebsd.org/cgi/man.cgi?query=md5sum&manpath=FreeBSD+14.0-RELEASE+and+Ports>

> ```
> Calculate the checksum of multiple files reversing the output:
>
>       $ md5 -r /boot/loader.conf /etc/rc.conf
>       ada5f60f23af88ff95b8091d6d67bef6 /boot/loader.conf
>       d80bf36c332dc0fdc479366ec3fa44cd /etc/rc.conf
>
> This is almost but not quite identical to the output from GNU mode:
>
>       $ md5sum /boot/loader.conf /etc/rc.conf
>       ada5f60f23af88ff95b8091d6d67bef6  /boot/loader.conf
>       d80bf36c332dc0fdc479366ec3fa44cd  /etc/rc.conf
>
> Note the two spaces between hash and file name.
> ```

</details>

Cyan4973 (approved):

PR #3998, opened 2024-03-28
Fix building on windows-x86 if clang already includes D101338

georgthegreat:
[D101338](https://reviews.llvm.org/D101338) landed in 2021, so clang16 should have it


terrelln:
@georgthegreat I just saw your comment on https://github.com/facebook/zstd/commit/94c102038b81ed89e3b013cb1977496612609f85, thanks for reporting the issue!

That code definitely needs to be updated to support 32-bits. And it looks like we are missing a 32-bit clang-cl CI job. I'll open an issue.

terrelln (approved):

terrelln:
I've filed https://github.com/facebook/zstd/issues/4010

PR #4001, opened 2024-03-29
Optimized The HUF_initFastDStream Function, improving Bit manipulation For Faster decompression

purple-claw:
static size_t HUF_initFastDStream(BYTE const* ip) {
    BYTE const lastByte = ip[7];
    size_t const bitsConsumed = lastByte ? 8 - ZSTD_highbit32(lastByte) : 0;
    size_t const value = MEM_readLEST(ip) | 1;
    assert(bitsConsumed <= 8);
    assert(sizeof(size_t) == 8);
    
    return (value << bitsConsumed) | (ip[bitsConsumed >> 3] >> (8 - bitsConsumed & 7));
}
This improves the bit manipulation by combining the shift and extraction operations into a single line, reducing the number of operations and potentially improving performance

After shifting value by bitsConsumed bits to the left, it performs a bitwise OR operation with the extracted bits from ip. It extracts the bits from ip by right-shifting ip[bitsConsumed >> 3] by (8 - bitsConsumed & 7) bits. This effectively combines the shifted value with the extracted bits from ip.

Also added some tests to showcase the difference in Performance


Cyan4973:
A lot of `.md` have been (accidentally?) modified in this commit.
Could this be avoided ?

purple-claw:
Sure I will go through that

PR #4002, opened 2024-03-30
Improved the Bit manipulation of HUF_initFastDStream, Func

purple-claw:
return (value << bitsConsumed) | (ip[bitsConsumed >> 3] >> (8 - bits Consumed & 7));

 Effeciently performing a bitwise operation to combine two values, value and a portion of the ip array. It shifts value left by bitsConsumed bits, then ORs that with the right-shifted portion of ip.


terrelln:
This breaks decompression. I don't understand the goal of this PR. Can you explain what you are trying to accomplish?

PR #4012, opened 2024-04-02
Remove specification text stating that probability overflow is invalid

elasota:
Due to the FSE probability encoding scheme, the largest possible value that can be decoded is the one that will make the cumulative probability equal the maximum, terminating the loop, so the situation that this text describes isn't possible.


Cyan4973 (approved):
Good point!

Issue #12, opened 2015-01-30
lib/zstd_static and programs/fileio.h have different line ending


t-mat:
It seems that these files in all branches have CRLF line endings instead of LF.

Cyan4973:
Mmmh, this is possible, since the development is done in both Windows and Linux/BSD configurations.
However, that's exactly the reason why there is a .gitattributes file : it's supposed to check for line ending on all files, and even correct them on the fly.

Now, I wonder if a .gitattributes scope is limited to its current directory. Maybe sub-directories remain unaffected...

t-mat:
As far as I know, usual glob filenames in `.gitattributes` are recognized as file mask for all subdirectories as you expected.

Perhaps, "Refreshing a repository after changing line endings" section in GitHub's guide "[Dealing with line endings](https://help.github.com/articles/dealing-with-line-endings/)" helps you. Since these files have not changed from initial commit, just "refreshing" may resolve this problem.

Cyan4973:
OK, thanks
The procedure is clear to follow.
Line endings are now fixed in both branches.

Issue #11, opened 2015-01-30
zstd crashes on decoding invalid archives


magv:
Hi. It seems that zstd will read illegal pointers and crash when presented with mangled archives. Here's one such example file (GitHub doesn't allow binary attachments, so I'm providing a hex dump):

    0000000    fd  2f  b5  1c  00  00  1c  40  00  12  31  32  31  31  31  31
    0000020    31  31  31  31  32  32  32  32  32  32  32  0a  10  98  00  ff
    0000040    7f  00  84  c0  00  00

Here's what gdb has to say about this problem:

    (gdb) run -d <example.zst >example
    Starting program: zstd -d <example.zst >example

    Program received signal SIGSEGV, Segmentation fault.
    0x0000000000410965 in ZSTD_decompressBlock (srcSize=28, src=0x801011000, maxDstSize=524288, dst=0x801032000, ctx=0x801006000) at lib/zstd.c:1533
    (gdb) bt
    #0  0x0000000000410965 in ZSTD_decompressBlock (srcSize=28, src=0x801011000, maxDstSize=524288, dst=0x801032000, ctx=0x801006000) at lib/zstd.c:1533
    #1  ZSTD_decompressContinue (dctx=0x801006000, dst=0x801032000, maxDstSize=524288, src=0x801011000, srcSize=31) at lib/zstd.c:1680
    #2  0x0000000000408681 in FIO_decompressFilename (output_filename=0x410f65 "-",input_filename=0x410f65 "-") at programs/fileio.c:363
    #3  0x0000000000401a4d in main (argc=2, argv=0x7fffffffd9d0) at programs/zstdcli.c:314

This is with zstd as of commit 00f9507; the crash is located [over here](//github.com/Cyan4973/zstd/blob/00f9507d26ededd25f8fc023c9f58113a609a27b/lib/zstd.c#L1533). The problem is that `ZSTD_decompressBlock` does not validate how big `matchLength` can get; in this case it is equal to 8650883, while the `maxDstSize` is only 524288 bytes, which results in an attempt to copy past the end of the output buffer.

Cyan4973:
zStd is not yet protected against invalid input.
This is an objective, obviously.

Part of the work is done, since FSE seems to properly handle invalid input.
Now, zStd proper must receive the same attention.

geeknik:
I found a variation of this crash that is most likely exploitable by people with malicious intent. Let me know where I can send the details.

Cyan4973:
[...]

Issue #10, opened 2015-01-28
Types of objects it can compress


MAK11:
Hi,
is this an alternative to Gzip? 
Also, what kind of files/objects can it compress - just base HTML or also JS, CSS, Images etc? 
Lastly, can it be deployed to any OS - Apache/IIS?
Also, if I need to 'activate' this on an existing website, what steps do I need to follow?

Cyan4973:
Hello

Keep in mind that this is a first release, it will require some tuning and testing to get all the intended properties.
That being said, there is no "limit" to the kind of files/objects that can be compressed, nor platform targeted. The algorithm is lossless and generic, like gzip. The code is supposed to be portable. Hopefully, a few wrapper will also help to make it available for a few more languages beyond C.

Note that zstd is still considered in "beta" phase, and will require some time and feedback to reach final "production" status.

Regards

MAK11:
Thanks for the reply @Cyan4973

PR #4016, opened 2024-04-03
ZSTD_decompressStream checks its state parameter is not NULL

GitMensch:
preventing segmentation faults seen with compressed perf recordings

`perf record --aio -z -- true && perf annotate`

Stacktrace from Debian, nearly identical in RHEL.

~~~
Program received signal SIGSEGV, Segmentation fault.
bZSTD_decompressStream (zds=0x0, output=output@entry=0x7fffffffc060, input=input@entry=0x7fffffffc040) at decompress/zstd_decompress.c:1657
1657    decompress/zstd_decompress.c: No such file or directory.
(gdb) bt
#0  ZSTD_decompressStream (zds=0x0, output=output@entry=0x7fffffffc060, input=input@entry=0x7fffffffc040) at decompress/zstd_decompress.c:1657
#1  0x00005555558dd2dc in zstd_decompress_stream (data=data@entry=0x555555de4898, src=src@entry=0x7ffff7ffb248, src_size=src_size@entry=389, dst=0x7ffff665a028,
    dst_size=dst_size@entry=528384) at util/zstd.c:100
#2  0x000055555583ea8d in perf_session__process_compressed_event (session=0x555555dddc10, event=0x7ffff7ffb240, file_offset=<optimized out>) at util/session.c:74
#3  0x00005555558409c1 in perf_session__process_user_event (session=session@entry=0x555555dddc10, event=event@entry=0x7ffff7ffb240, file_offset=file_offset@entry=576)
    at util/session.c:1664
#4  0x000055555584242b in perf_session__process_event (file_offset=576, event=0x7ffff7ffb240, session=0x555555dddc10) at util/session.c:1797
#5  0x0000555555843afa in reader__process_events (prog=0x7fffffffc720, session=0x555555dddc10, rd=<synthetic pointer>) at util/session.c:2251
#6  __perf_session__process_events (session=0x555555dddc10) at util/session.c:2309
#7  perf_session__process_events (session=session@entry=0x555555dddc10) at util/session.c:2342
#8  0x000055555574991d in __cmd_annotate (ann=0x7fffffffc880) at builtin-annotate.c:402
#9  cmd_annotate (argc=<optimized out>, argv=0x7fffffffdc40) at builtin-annotate.c:647
#10 0x00005555557ddb43 in run_builtin (p=p@entry=0x555555bb5168 <commands+360>, argc=argc@entry=1, argv=argv@entry=0x7fffffffdc40) at perf.c:313
#11 0x0000555555747c08 in handle_internal_command (argv=0x7fffffffdc40, argc=1) at perf.c:365
#12 run_argv (argv=<synthetic pointer>, argcp=<synthetic pointer>) at perf.c:409
#13 main (argc=1, argv=0x7fffffffdc40) at perf.c:539
~~~

_Likely_ perf is to be blamed in the first place, but as there are already checks for other issues it seems to be useful to cater for that.

### DIFF
--- a/lib/decompress/zstd_decompress.c
+++ b/lib/decompress/zstd_decompress.c
@@ -2095,6 +2095,10 @@ size_t ZSTD_decompressStream(ZSTD_DStream* zds, ZSTD_outBuffer* output, ZSTD_inB
 
     DEBUGLOG(5, "ZSTD_decompressStream");
     RETURN_ERROR_IF(
+        zds == NULL,
+        stage_wrong,
+        "forbidden. DStream is NULL");
+    RETURN_ERROR_IF(
         input->pos > input->size,
         srcSize_wrong,
         "forbidden. in: pos: %u   vs size: %u",


GitMensch:
CLA signed

Cyan4973:
I would change the title, because it is misleading.
What you mean is that `ZSTD_decompressStream()` should check that the state pointer is not `NULL`.

Indeed, such a scenario is a major breach of contract, it points to a serious error on the user side.

So far, this scenario has been classified as a "breach of contract", meaning we just document the (obvious) convention that `ZSTD_DStream*` must be valid, but if it's not, it's UB (Undefined Behavior). For perspective, same thing happens if the pointer is not `NULL` but completely bogus (dangling).

We could nonetheless debate about extending the contract to make `ZSTD_decompressStream()` able to detect a `NULL` state and cleanly return an error in this case. For this case, the error should be clearly documented (I suspect no corresponding error code exist yet, and `stage_wrong` is a too rough approximation of the problem, that can be misleading for debugging).

What I'm concerned though is about consistency: if we do that for this symbol, we would have to do that for every other symbol which manipulates a state as head parameter, for consistency sake. And there are really a lot of them.

Cyan4973:
We'll preserve the current contract,
which requires that the state pointer must be valid.
Therefore, this should be an `assert()` and not a test.

Cyan4973:
replaces by #4249

PR #4019, opened 2024-04-05
fix missing include folder for resource compiler

b7f7:
the cmake build with VS studio 2022 on windows does fail to build because rc does not have the include directory set.
include is now set for rc and build works


Cyan4973 (approved):
LGTM

PR #4020, opened 2024-04-06
Provide variant pkg-config file for multi-threaded static lib

Ansuel:
Multi-threaded static library require -pthread to correctly link and works.
The pkg-config we provide tho only works with dynamic multi-threaded library
and won't provide the correct libs and cflags values if lib-mt is used.

To handle this, introduce an env variable MT to permit advanced user to
install and generate a correct pkg-config file for lib-mt or detect if
lib-mt target is called.

With MT env set on calling make install-pc, libzstd.pc.in is a
pkg-config file for a multi-threaded static library.

On calling make lib-mt, a libzstd.pc is generated for a multi-threaded
static library as it's what asked by the user by forcing it.

libzstd.pc is changed to PHONY to force regeneration of it on calling
lib targets or install-pc to handle case where the same directory is
used for mixed compilation.

This was notice while migrating from meson to make build system where
meson generates a correct .pc file while make doesn't.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -63,6 +63,8 @@ CPPFLAGS_DYNLIB  += -DZSTD_MULTITHREAD # dynamic library build defaults to multi
 LDFLAGS_DYNLIB   += -pthread
 CPPFLAGS_STATICLIB +=                  # static library build defaults to single-threaded
 
+# pkg-config Libs.private points to LDFLAGS_DYNLIB
+PCLIB := $(LDFLAGS_DYNLIB)
 
 ifeq ($(findstring GCC,$(CCVER)),GCC)
 decompress/zstd_decompress_block.o : CFLAGS+=-fno-tree-vectorize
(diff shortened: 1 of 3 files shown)


lib/Makefile
```diff
@@ -319,6 +319,17 @@ libzstd.pc: libzstd.pc.in
 	        -e 's|@LIBS_PRIVATE@|$(LDFLAGS_DYNLIB)|' \
 	        $< >$@
 
+mt-libzstd.pc: mt-libzstd.pc.in
```
Cyan4973: Both recipes look identical. 
So they could be combined.

Cyan4973:
I understand the need,
but I suspect the same capability could be created without adding a new `make` target and a new `sed` intermediate.
I would likely prefer this capability to be controlled via a build parameter.

Ansuel:
Do you mean a special env on calling make install-pc ?

Also IMHO adding extra sed and token in the .in file might result in the file more bloated than keeping the thing separated but I understand it since I also hate duplicating things... It's just that makefile makes thing hard to handle with file stamp.

Cyan4973:
> Do you mean a special env on calling make install-pc ?

Yes.

It seems it would make sense to use the same variable both to create a MT static library, which would now depend on `-pthread`, and then to update the `libzstd.pc` with the correct link flags.

Ansuel:
For static multi thread we have make lib-mt... Maybe something like MT=1
make install-pc ?

Il Sab 6 Apr 2024, 23:35 Yann Collet ***@***.***> ha scritto:

Ansuel:
@Cyan4973 I pushed a more advanced version of this...  What do you think?

I also think there was a very latent bug in how the .pc file was handled on dirty build. (since lib generates the .pc, any call to install-pc would ignore any custom values of the PREFIX env) should be fixed by forcing .pc regeneration every time.

Cyan4973:
Yes, that looks much better.

Give me some time to properly review the build/install flow, I should be able to comment this week.

Ansuel:
@Cyan4973 no problem and thanks for reviewing this. We recently switched from xz to zstd on openwrt and this is just an effort to upstream our downstream patch (we use multi-threaded static library on our build systems)

Cyan4973 (approved):
This looks good to me!

PR #4025, opened 2024-04-12
Fixed all memory leaks and almost all undefined behaviour

413x1nkp:
## Misc
- Initialized as many variables as possible to avoid any undefined behaviour.
- Removed `fclose(f)` line at `tests/paramgrill.c` in `createBuffers` function, since if `f` couldn't initialize properly, there's no need to close it either.
- Added a fallback at `tests/regression/result.c` in `result_get_error_string(result_t result)` function, just in case.

### Fixed memory leaks/possible memory leaks at:
- `doc/educational_decoder/harness.c` during `read_file` function.
- `tests/bigdict.c` now goes to cleanup if condition `(!buffer || !out || !roundtrip || !cctx || !dctx)` is met, instead of simply returning 1.
- `zlibWrapper/gzwrite.c` now frees state.state before returning from the function.

### DIFF
--- a/doc/educational_decoder/harness.c
+++ b/doc/educational_decoder/harness.c
@@ -50,6 +50,7 @@ static buffer_s read_file(const char *path)
 
     fclose(f);
     buffer_s const b = { ptr, size };
+    free(ptr);
     return b;
 }
 
--- a/doc/educational_decoder/harness.c
+++ b/doc/educational_decoder/harness.c
@@ -50,7 +50,6 @@ static buffer_s read_file(const char *path)
 
     fclose(f);
     buffer_s const b = { ptr, size };
-    free(ptr);
     return b;
 }
 
--- a/tests/bigdict.c
+++ b/tests/bigdict.c
@@ -70,12 +70,14 @@ int main(int argc, const char** argv)
     char* buffer = (char*)malloc(bufferSize);
     void* out = malloc(outSize);
     void* roundtrip = malloc(dataSize);
+    int _exit_code = 0;
     (void)argc;
     (void)argv;
 
     if (!buffer || !out || !roundtrip || !cctx || !dctx) {
         fprintf(stderr, "Allocation failure\n");
-        return 1;
+        _exit_code = 1;
+        goto cleanup;
     }
 
     if (ZSTD_isError(ZSTD_CCtx_setParameter(cctx, ZSTD_c_windowLog, 31)))
@@ -119,10 +121,13 @@ int main(int argc, const char** argv)
 
     fprintf(stderr, "Success!\n");
 
+    goto cleanup;
+
+cleanup:
     free(roundtrip);
     free(out);
     free(buffer);
-    ZSTD_freeDCtx(dctx);
     ZSTD_freeCCtx(cctx);
-    return 0;
+    ZSTD_freeDCtx(dctx);
+    return _exit_code;
 }
--- a/tests/paramgrill.c
+++ b/tests/paramgrill.c
@@ -1273,7 +1273,6 @@ static int createBuffers(buffers_t* buff, const char* const * const fileNamesTab
         f = fopen(fileNamesTable[n], "rb");
         if (f==NULL) {
             DISPLAY("impossible to open file %s\n", fileNamesTable[n]);
-            fclose(f);
             ret = 10;
             goto _cleanUp;
         }
--- a/zlibWrapper/examples/minigzip.c
+++ b/zlibWrapper/examples/minigzip.c
@@ -287,7 +287,7 @@ int gzclose _Z_OF((gzFile));
 
 int gzclose(gzFile gz) {
     z_stream *strm;
-    unsigned char out[BUFLEN] = 0;
+    unsigned char out[BUFLEN] = { 0 };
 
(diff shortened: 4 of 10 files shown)


doc/educational_decoder/harness.c
```diff
@@ -50,6 +50,7 @@ static buffer_s read_file(const char *path)
 
     fclose(f);
     buffer_s const b = { ptr, size };
+    free(ptr);
```
Cyan4973, line 53: :+1:

tests/bigdict.c
```diff
@@ -70,12 +70,14 @@ int main(int argc, const char** argv)
...
     void* roundtrip = malloc(dataSize);
+    int _exit_code = 0;
     (void)argc;
     (void)argv;
 
     if (!buffer || !out || !roundtrip || !cctx || !dctx) {
         fprintf(stderr, "Allocation failure\n");
-        return 1;
+        _exit_code = 1;
+        goto cleanup;
```
Cyan4973, line 80: :+1:

tests/paramgrill.c
```diff
@@ -1273,7 +1273,6 @@ static int createBuffers(buffers_t* buff, const char* const * const fileNamesTab
         f = fopen(fileNamesTable[n], "rb");
         if (f==NULL) {
             DISPLAY("impossible to open file %s\n", fileNamesTable[n]);
-            fclose(f);
```
Cyan4973, line 1276: :+1:

Cyan4973:
The first list of changes, detailed in the summary and present in the first commit, looks mostly good to me (with the exception of `gzwrite` change, which is likely wrong).

But then, there is a second list of changes, named "fixed ISO C incompatibility" in commits 2 and 3, which are a bit more concerning and would deserve some scrutiny. And strangely, while they appear in the commit timeline, I don't see them in the GitHub PR comparison view. 
It's the first time I observe such discrepancy, and it means that, I'm not even sure what's the exact content of this PR...

zlibWrapper/gzwrite.c
```diff
@@ -64,6 +64,8 @@ local int gz_init(gz_statep state) {
         strm->next_out = state.state->out;
         state.state->x.next = strm->next_out;
     }
+
+    free(state.state);
```
Cyan4973: That one is weird.
I'm not even sure what's going on in this code.

One of the first actions in this function is : `state.state->in = (unsigned char*)malloc(state.state->want << 1);`,
which presumes that `state.state` is already allocated (should probably be asserted), _before_ entering the function,
which means that, something else has allocated `state.state`, and therefore something else is in charge of freeing it.
I don't see how it could be good to `free` it here.

Cyan4973 (changes requested):

zlibWrapper/gzwrite.c
413x1nkp: Apologies, You're right. That was an oversight from me.
I'll fix this tomorrow morning... Again, sorry for the inconvenience

413x1nkp:
> The first list of changes, detailed in the summary and present in the first commit, looks mostly good to me (with the exception of `gzwrite` change, which is likely wrong).
> 
> But then, there is a second list of changes, named "fixed ISO C incompatibility" in commits 2 and 3, which are a bit more concerning and would deserve some scrutiny. And strangely, while they appear in the commit timeline, I don't see them in the GitHub PR comparison view. It's the first time I observe such discrepancy, and it means that, I'm not even sure what's the exact content of this PR...

Yeah, about those,
Basically, i decided to initialize structs by `= {}`, while i completely forgot that this is not allowed on C version that one of the checks runs on

doc/educational_decoder/harness.c
Cyan4973: This change is incorrect :
the whole point of this function is to return a populated buffer, passed as a member of the `buffer_s` structure, effectively transferring ownership to the caller of the function (which will have to free it later, using the provided `freeBuffer()` function).

Maybe this could be documented if it's not clear enough...

Cyan4973 (changes requested):

zlibWrapper/examples/minigzip.c
```diff
@@ -234,7 +234,7 @@ int gzwrite _Z_OF((gzFile, const void *, unsigned));
 
 int gzwrite(gzFile gz, const void *buf, unsigned len) {
     z_stream *strm;
-    unsigned char out[BUFLEN];
+    unsigned char out[BUFLEN] = 0;
```
Cyan4973: Did you meant `{ 0 }` ?

413x1nkp: God damn, I'm absent-minded...
Sorry I made so much mistakes in something that's supposed to fix stuff...

Will fix this though, thanks for noticing

Cyan4973 (approved):
lgtm

Cyan4973:
There is another bug in this PR,
in `result.c`, line 28:
```
return "unknown error - " + result_get_error(result);
```
This formulation is erroneous:
it's not python, it will not concatenate 2 strings to create a new one.
Instead, it results in adding some integer to a pointer, 
producing another pointer somewhere in memory, potentially out of bound.

PR #4031, opened 2024-04-23
update documentation of ZSTD_decompressStream()

### LINKED ISSUE #4030: Improve misleading wording in the streaming decompression howto
**Describe the bug**

The wording [here](https://github.com/facebook/zstd/blob/ff7a151f2e6c009b657d9f798c2d9962b0e3feb5/lib/zstd.h#L870-L871) is slightly misleading. You should not call again `ZSTD_decompressStream()` if  `output.pos == output.size` **and** the function returned `0`.

**Expected behavior**

Perhaps this could be changed to:

```
*  But if `output.pos == output.size` and the function returns greater than 0 there might be some data left within internal buffers.,
```

(I would have gladly PR'd but on a matter of principle I don't sign CLAs).

Cyan4973:
wording slightly more precise, by recommending to check the return value. 
suggested by @dbuenzli .
fix #4030

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -855,7 +855,7 @@ ZSTDLIB_API size_t ZSTD_endStream(ZSTD_CStream* zcs, ZSTD_outBuffer* output);
 *
 *  A ZSTD_DStream object is required to track streaming operations.
 *  Use ZSTD_createDStream() and ZSTD_freeDStream() to create/release resources.
-*  ZSTD_DStream objects can be reused multiple times.
+*  ZSTD_DStream objects can be re-employed multiple times.
 *
 *  Use ZSTD_initDStream() to start a new decompression operation.
 * @return : recommended first input size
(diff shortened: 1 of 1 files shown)


dbuenzli:
Thanks, that's much better. 

I wonder though if something could still not be said more explictly on how to make sure that given the end of your input stream (you no longer have data to push): 

1. You didn't have a malformed or truncated input stream.
2. You have everything output.


I believe this is the condition:
```
input.pos == input.size
output.pos <= output.size
@return = 0
```

Cyan4973:
> Thanks, that's much better.
> 
> I wonder though if something could still not be said more explictly on how to make sure that given the end of your input stream (you no longer have data to push):
> 
> 1. You didn't have a malformed or truncated input stream.
> 2. You have everything output.
> 
> I believe this is the condition:
> 
> ```
> input.pos == input.size
> output.pos <= output.size
> @return = 0
> ```

The official way to check for the end of a frame is `@return == 0`.

Now, an input _may_ consist of _multiple_ frames concatenated, so the **end of input** is both `input.pos == input.size` and `@return = 0`.

`output.pos <= output.size` is redundant. It's always true. If not, there's a bug.

terrelln (approved):

PR #4046, opened 2024-05-14
Improve support for IAR compiler with attributes and intrinsics

josepho0918:
This patch improves development experience and performance for projects using the IAR toolchain.

- Utilize IAR compiler attributes
- Optimized with IAR intrinsics

### DIFF
--- a/lib/common/entropy_common.c
+++ b/lib/common/entropy_common.c
@@ -38,11 +38,7 @@ const char* HUF_getErrorName(size_t code) { return ERR_getErrorName(code); }
 /*-**************************************************************
 *  FSE NCount encoding-decoding
 ****************************************************************/
-#if defined(__IAR_SYSTEMS_ICC__)
-MEM_STATIC
-#else
 FORCE_INLINE_TEMPLATE
-#endif
 size_t FSE_readNCount_body(short* normalizedCounter, unsigned* maxSVPtr, unsigned* tableLogPtr,
                            const void* headerBuffer, size_t hbSize)
 {
--- a/lib/common/entropy_common.c
+++ b/lib/common/entropy_common.c
@@ -38,7 +38,11 @@ const char* HUF_getErrorName(size_t code) { return ERR_getErrorName(code); }
 /*-**************************************************************
 *  FSE NCount encoding-decoding
 ****************************************************************/
+#if defined(__IAR_SYSTEMS_ICC__)
+MEM_STATIC
+#else
 FORCE_INLINE_TEMPLATE
+#endif
 size_t FSE_readNCount_body(short* normalizedCounter, unsigned* maxSVPtr, unsigned* tableLogPtr,
                            const void* headerBuffer, size_t hbSize)
 {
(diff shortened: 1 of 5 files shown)


lib/common/entropy_common.c
```diff
@@ -38,7 +38,11 @@ const char* HUF_getErrorName(size_t code) { return ERR_getErrorName(code); }
 /*-**************************************************************
 *  FSE NCount encoding-decoding
 ****************************************************************/
+#if defined(__IAR_SYSTEMS_ICC__)
+MEM_STATIC
```
Cyan4973: Is there a problem with `FORCE_INLINE_TEMPLATE` ?
Why is it a problem here specifically ?

Cyan4973:
Sometimes I see `__ICCARM__` being tested, and sometimes `__IAR_SYSTEMS_ICC__`.
Would you mind explaining the difference, and why sometimes one is better than the other ?

lib/common/entropy_common.c
josepho0918: > Is there a problem with `FORCE_INLINE_TEMPLATE` ? Why is it a problem here specifically ?

The IAR compiler complains that the `FSE_readNCount_body` function is too complex to be inlined.

josepho0918:
> Sometimes I see `__ICCARM__` being tested, and sometimes `__IAR_SYSTEMS_ICC__`. Would you mind explaining the difference, and why sometimes one is better than the other ?

`__IAR_SYSTEMS_ICC__` identifies the IAR compiler platform, while `__ICCARM__` specifically designates the IAR compiler for Arm architecture. Since there're IAR versions for various architectures, such as RISC-V, `__ICCARM__` is utilized for IAR intrinsics.

lib/common/entropy_common.c
Cyan4973: The function is indeed pretty big.

It seems the point of `FORCE_INLINE_TEMPLATE` is to pass down `BMI2_TARGET_ATTRIBUTE` compilation directive to an instance of the function, which is invoked when the appropriate target (`AVX2` + `BMI2` architecture) is detected at runtime.

Of course, this capability is irrelevant for `ARM` architectures.

So yes, technically this modification works. It's just a pity that it translates into one exception which will have to be maintained for one compiler.

Cyan4973:
Quick question : 
does it sound plausible to feature an `__IAR_SYSTEMS_ICC__` compiler test during CI tests on Github Actions ?

If not, this is a "blind fix", and any future update could break compatibility is subtle ways, which would remain invisible.

josepho0918:
> Quick question : does it sound plausible to feature an `__IAR_SYSTEMS_ICC__` compiler test during CI tests on Github Actions ?
> 
> If not, this is a "blind fix", and any future update could break compatibility is subtle ways, which would remain invisible.

It will be fantastic to have CI tests for IAR. However, it's a paid tool.

lib/common/entropy_common.c
josepho0918: Would it be better to use `HINT_INLINE` instead of `FORCE_INLINE_TEMPLATE` for `FSE_readNCount_body`?

Cyan4973 (approved):

PR #4048, opened 2024-05-15
zstd.1: remove spurious newlines

concussious:
Three IP macros define the indentation of paragraphs that don't exist, creating several spurious newlines.

Slightly related, would a rewrite of this document in semantic markup (mdoc(7) instead of man(7)) be welcome? mdoc(7) is 28 years old and works with every manual presentation stack I've seen, but brings a lot of features (including mature, stable semantic search!) and massively increased legibility to the table, for zero extra dependencies and marginal file size difference.

If it's welcome, I can work on that in my free time over the next few months.

### DIFF
--- a/programs/zstd.1
+++ b/programs/zstd.1
@@ -37,8 +37,6 @@ When compressing a single file, \fBzstd\fR displays progress notifications and r
 .IP "\(bu" 4
 \fBzstd\fR does not store the input\'s filename or attributes, only its contents\.
 .
-.IP "" 0
-.
 .P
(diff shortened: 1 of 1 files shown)


Cyan4973:
`zstd.1` is automatically generated from `zstd.1.md`.
We don't accept manual changes to this file, since they will be erased by next automated update.

See `make -C programs man`.

If the current process generates unwanted artifacts, what could be interesting is an updated `man` page generation process.

concussious:
Copy. I'll poke around with that but it is a little out of my wheelhouse.

Also, like all human beings, I use your work everyday, I think that's really cool - thanks.

Issue #4026, opened 2024-04-16
Missing check on failed allocation leads to NULL-ptr dereference


benhawkes:
**Describe the bug**

In the ZBUFFv06_createDCtx function in lib/legacy/zstd_v06.c, there is an unchecked allocation failure that can result in a NULL-ptr dereference when calling ZSTD_decompressStream in low-memory conditions. Specifically the following line in ZBUFFv06_createDCtx is unchecked:

`zbd->zd = ZSTDv06_createDCtx();`

If this allocation fails, a NULL-ptr deference in ZSTDv06_decompressBegin (via ZSTDv06_decompressBegin_usingDict) can be observed.

It looks like there are similar cases in lib/legacy/zstd_v05.c and lib/legacy/zstd_v04.c

Adenilson:
Seems like an one-liner fix, just added a patch for it.

PR #4050, opened 2024-05-17
[fix] Add check on failed allocation in legacy/zstd_v06

Adenilson:
As reported by Ben Hawkes in #4026, a failure to allocate a zstd context would lead to a dereference of a NULL pointer due to a missing check on the returned result of ZSTDv06_createDCtx().

This patch fix the issue by adding a check for valid returned pointer.

### DIFF
--- a/lib/legacy/zstd_v06.c
+++ b/lib/legacy/zstd_v06.c
@@ -3919,6 +3919,10 @@ ZBUFFv06_DCtx* ZBUFFv06_createDCtx(void)
     if (zbd==NULL) return NULL;
     memset(zbd, 0, sizeof(*zbd));
     zbd->zd = ZSTDv06_createDCtx();
+    if (zbd->zd==NULL) {
+        ZBUFFv06_freeDCtx(zbd); /* avoid leaking the context */
+        return NULL;
+    }
     zbd->stage = ZBUFFds_init;
     return zbd;
 }


Adenilson:
I'm trying to follow the same coding style used in:
https://github.com/facebook/zstd/blob/dev/lib/legacy/zstd_v06.c#L3919

lib/legacy/zstd_v06.c
```diff
@@ -3919,6 +3919,7 @@ ZBUFFv06_DCtx* ZBUFFv06_createDCtx(void)
     if (zbd==NULL) return NULL;
     memset(zbd, 0, sizeof(*zbd));
     zbd->zd = ZSTDv06_createDCtx();
+    if (zbd->zd == NULL) return NULL;
```
Adenilson: Remove empty spaces around ==.

Adenilson: Done.

lib/legacy/zstd_v06.c
```diff
@@ -3919,6 +3919,7 @@ ZBUFFv06_DCtx* ZBUFFv06_createDCtx(void)
     if (zbd==NULL) return NULL;
     memset(zbd, 0, sizeof(*zbd));
     zbd->zd = ZSTDv06_createDCtx();
+    if (zbd->zd==NULL) return NULL;
```
Cyan4973: Returning immediately here would result in a memory leak, because `zbd` has been allocated at that point.
Free `zbd`, using `ZBUFFv06_freeDCtx()`, before returning.

Cyan4973 (changes requested):

lib/legacy/zstd_v06.c
Adenilson: Oh, indeed. Thanks a lot for the review, I will fix that.

Adenilson: Done.

Cyan4973 (approved):

adenilsoncavalcanti:
@Cyan4973 anything else you like to see changed in the patch?
:-)

Cyan4973:
Nope, it's good!

PR #4065, opened 2024-06-03
Drop support for FreeBSD 13.2 CI

embg:
It seems like Cirrus CI doesn't support FreeBSD 13.2 right now. This is causing our CI jobs to fail. This PR drops 13.2 support and keeps support for 14.0, which works fine in CI.

Example failing job: https://github.com/facebook/zstd/runs/25742617625

Example error message:
```
"message": "The resource 'projects/freebsd-org-cloud-dev/global/images/family/freebsd-13-2' was not found",
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,7 +3,6 @@ task:
   freebsd_instance:
     matrix:
       image_family: freebsd-14-0
-      image_family: freebsd-13-2
   install_script: pkg install -y gmake coreutils
   script: |
     MOREFLAGS="-Werror" gmake -j all


Cyan4973:
Thanks for trying to keep the CI clean @embg !

I've been looking into this issue,
especially as we recently dropped `freebsd-12` support because Cirrus-CI deprecated it.

But, according to Cirrus-CI documentation, `freebsd-13.2` is still a valid target:
https://cirrus-ci.org/guide/FreeBSD/

I haven't found any announcement that states that `freebsd-13.2` is deprecated or will soon be.

Unfortunately, such kind of issues happen from time to time, CI tests are not as stable as they should. When it does, we have little other solution than to wait, hoping it's a transient issue that's getting fixed.

That being said, we are not _that_ dependent on `FreeBSD-13.2` specifically. What we care about is to have _at least_ one `FreeBSD` target tested. And we still have `FreeBSD-14`.

Consequently, we could give Cirrus CI a bit more time to get it fixed, say, a few more days. And if it isn't well, let's drop it, it's not an "important" test (given we also have another FreeBSD test working fine).

Cyan4973:
It's been a few days and FreeBSD 13.2 is still not fixed,
let's drop it.

Cyan4973 (approved):

PR #4067, opened 2024-06-05
Fix $filter operants and Msys/Cygwin

QBos07:
- switched the patter and input of $filter into the right places
- added pattern wildcard to MSYS_NT & CYGWIN_NT as they change with windows versions
- correctly identify MSYS2, even in an env like MINGW64

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -266,7 +266,7 @@ clean:
 #-----------------------------------------------------------------------------
 # make install is validated only for below listed environments
 #-----------------------------------------------------------------------------
-ifneq (,$(filter $(UNAME),Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS Haiku AIX MSYS_NT CYGWIN_NT))
+ifneq (,$(filter Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS Haiku AIX MSYS_NT% CYGWIN_NT%,$(UNAME)))
 
 lib: libzstd.pc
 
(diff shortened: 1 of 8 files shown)


QBos07:
CI failure not because of changes

Cyan4973:
Are there scenarios that would fail before this PR ?
If yes, does it feel possible to add a corresponding test ?

QBos07:
Yes, Msys and Cygwin didn't work before.
You could add tests but in practice this should be a one off with little impact if it doesn't work

Cyan4973:
I'm mostly concerned that, if `Msys` and `Cygwin` were failing before,
it remained invisible in our CI, so we might have a blind spot, and this may be worth fixing.

Also, without a test, it could be that the delivered property (works with `Msys` and `Cygwin`) will be broken by accident in some future PR, and we would not notice it.

Looking at our current CI tests,
it seems we already have a bunch of `mingw` tests running,
we even have one [`cygwin`](https://github.com/facebook/zstd/actions/runs/9388161999/job/25852811648?pr=4067) test,
but they apparently all completed fine, 
therefore, were these tests wrong, or incomplete?

QBos07:
The toolchain for msys already uses the workaround and the most impacted part wich is installing is not tested

Cyan4973 (approved):

lib/Makefile
```diff
@@ -305,13 +305,21 @@ else
   PCLIB := $(LDFLAGS_DYNLIB)
 endif
 
-ifneq (,$(filter $(UNAME),FreeBSD NetBSD DragonFly))
+
+ifneq ($(MT),)
```
Cyan4973, line 309: That section is a bit unclear to me.
Why was it changed ?
It seems it just duplicates the section above.

PR #4073, opened 2024-06-12
added a cygwin install test to check #4067

Cyan4973:


Cyan4973:
And as expected :
```
make: *** No rule to make target 'install'.  Stop.
```

Cyan4973:
And it works as expected after applying #4067.

So I guess we have our new test for #4067.

Issue #4045, opened 2024-05-14
ZDICT_trainFromBuffer_cover is not thread safe


as-com:
**Describe the bug**

ZDICT_trainFromBuffer_cover is not thread safe due to the use of global state:

https://github.com/facebook/zstd/blob/78955f5f9ddb3601566884ce217d91c132f5edc1/lib/dictBuilder/cover.c#L235-L236

Calling this function from multiple threads may result in segfault/access violation, use-after-free and other badness.

**To Reproduce**
Call ZDICT_trainFromBuffer_cover from multiple threads

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 11 23H2
 - Compiler: MSVC, Visual Studio 2022
 - Flags: default
 - Other relevant hardware specs: N/A
 - Build system: whatever zstd-sys/cargo uses

Adenilson:
From what I could understand from the code, the issue derives from having to use a global context to pass extra information (i.e. a context) while performing the call to qsort().

Perhaps the reentrant qsort_r() could be a good fit for this problem?

From the documentation (https://man7.org/linux/man-pages/man3/qsort_r.3.html), quote: "The qsort_r() function is identical to qsort() except that the comparison function compar takes a third argument.  A pointer is passed to the comparison function via arg.  In this way, the comparison function does not need to use global variables to pass through arbitrary arguments, and is therefore reentrant and safe to use in threads.".

It is available with small variations (i.e.  qsort_s()) for both Windows (https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort-s?view=msvc-170) and BSD (https://man.freebsd.org/cgi/man.cgi?query=qsort_r&sektion=3), therefore should be portable.

I would like to help on this one if it is available.
:-)

adenilsoncavalcanti:
An update: I have a first draft patch, it seems to work fine on OSX@clang and Linux@gcc.

I will need a bit of time to get a Windows machine (and apparently a BSD machine too...), so I hope this is not a high priority bug.

I got a few questions, please see below:

@as-com: as you mentioned zstd-sys/cargo, I would assume that you are using zstd from Rust? Do you think it is possible to share a simple test case exposing the bug?

@Cyan4973: concerning testing... what would be the recommended test suite (and how to build/run/etc) for changes in the cover.c file?

[...]

PR #4078, opened 2024-06-19
[zstd][dict] Ensure that dictionary training functions are fully reen…

Adenilson:
…trant

The two main functions used for dictionary training using the COVER algorithm require initialization of a COVER_ctx_t where a call to qsort() is performed.

The issue is that the standard C99 qsort() function doesn't offer a way to pass an extra parameter for the comparison function callback (e.g. a pointer to a context) and currently zstd relies on a *global* static variable to hold a pointer to a context needed to perform the sort operation.

If a zstd library user invokes either ZDICT_trainFromBuffer_cover or ZDICT_optimizeTrainFromBuffer_cover from multiple threads, the global context may be overwritten before/during the call/execution to qsort() in the initialization of the COVER_ctx_t, thus yielding to crashes and other bad things (Tm) as reported on issue #4045.

Enters qsort_r(): it was designed to address precisely this situation, to quote from the documention [1]: "the comparison function does not need to use global variables to pass through arbitrary arguments, and is therefore reentrant and safe to use in threads."

It is available with small variations for multiple OSes (GNU, BSD[2], Windows[3]), and the ISO C11 [4] standard features on annex B-21 qsort_s() as part of the <stdlib.h>. Let's hope that compilers eventually catch up with it.

For now, we have to handle the small variations in function parameters for each platform.

The current fix solves the problem by allowing each executing thread pass its own COVER_ctx_t instance to qsort_r(), removing the use of a global pointer and allowing the code to be reentrant.

[1] https://man7.org/linux/man-pages/man3/qsort_r.3.html
[2] https://man.freebsd.org/cgi/man.cgi?query=qsort_r
[3] https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort-s?view=msvc-170
[4] https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -238,6 +238,11 @@ typedef struct {
   unsigned d;
 } COVER_ctx_t;
 
+#if !defined(_GNU_SOURCE) && !defined(__APPLE__) && !defined(_MSC_VER)
+/* C90 only offers qsort() that needs a global context. */
+static COVER_ctx_t *g_coverCtx = NULL;
+#endif
+
 /*-*************************************
 *  Helper functions
 ***************************************/
@@ -284,8 +289,10 @@ static int COVER_cmp8(COVER_ctx_t *ctx, const void *lp, const void *rp) {
 static int WIN_CDECL COVER_strict_cmp(void* g_coverCtx, const void* lp, const void* rp) {
 #elif defined(__APPLE__)
 static int WIN_CDECL COVER_strict_cmp(void *g_coverCtx, const void *lp, const void *rp) {
-#elif defined(_GNU_SOURCE) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+#elif defined(_GNU_SOURCE)
 static int WIN_CDECL COVER_strict_cmp(const void *lp, const void *rp, void *g_coverCtx) {
+#else /* C90 fallback.*/
+static int WIN_CDECL COVER_strict_cmp(const void *lp, const void *rp) {
 #endif
   int result = COVER_cmp((COVER_ctx_t*)g_coverCtx, lp, rp);
   if (result == 0) {
@@ -300,8 +307,10 @@ static int WIN_CDECL COVER_strict_cmp(const void *lp, const void *rp, void *g_co
 static int WIN_CDECL COVER_strict_cmp8(void* g_coverCtx, const void* lp, const void* rp) {
 #elif defined(__APPLE__)
 static int WIN_CDECL COVER_strict_cmp8(void *g_coverCtx, const void *lp, const void *rp) {
-#elif defined(_GNU_SOURCE) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+#elif defined(_GNU_SOURCE)
 static int WIN_CDECL COVER_strict_cmp8(const void *lp, const void *rp, void *g_coverCtx) {
+#else /* C90 fallback.*/
+static int WIN_CDECL COVER_strict_cmp8(const void *lp, const void *rp) {
 #endif
   int result = COVER_cmp8((COVER_ctx_t*)g_coverCtx, lp, rp);
   if (result == 0) {
@@ -637,7 +646,7 @@ static size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
     qsort_r(ctx->suffix, ctx->suffixSize, sizeof(U32),
             ctx,
             (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp));
-#elif defined(_GNU_SOURCE) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+#elif defined(_GNU_SOURCE)
     qsort_r(ctx->suffix, ctx->suffixSize, sizeof(U32),
             (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp),
             ctx);
@@ -645,8 +654,15 @@ static size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
     qsort_s(ctx->suffix, ctx->suffixSize, sizeof(U32),
             (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp),
             ctx);
+#elif defined(__OpenBSD__)
+    mergesort(ctx->suffix, ctx->suffixSize, sizeof(U32),
+          (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp));
+#else /* C90 fallback.*/
+    g_coverCtx = ctx;
+    /* TODO(cavalcanti): implement a reentrant qsort() when is not available. */
+    qsort(ctx->suffix, ctx->suffixSize, sizeof(U32),
+          (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp));
 #endif
-
   }
   DISPLAYLEVEL(2, "Computing frequencies\n");
   /* For each dmer group (group of positions with the same first d bytes):
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -23,7 +23,9 @@
 ***************************************/
 /* qsort_r is an extension. */
 #if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || defined(__CYGWIN__) || defined(__MSYS__)
+#if !defined(_GNU_SOURCE)
 #define _GNU_SOURCE
+#endif
 #endif
 
 #include <stdio.h>  /* fprintf */
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -22,9 +22,10 @@
 *  Dependencies
 ***************************************/
 /* qsort_r is an extension. */
-#if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__)
+#if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || defined(__CYGWIN__) || defined(__MSYS__)
 #define _GNU_SOURCE
 #endif
+
 #include <stdio.h>  /* fprintf */
 #include <stdlib.h> /* malloc, free, qsort_r */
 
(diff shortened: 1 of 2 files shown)


dearblue:
I was curious, so I did a little research, focusing on FreeBSD.

The version of FreeBSD that adds support for the POSIX standard `qsort_r()` is 14.0, released on 2023-11-20. It is described as using "C11 Generic selection" to maintain past compatibility.
https://github.com/freebsd/freebsd-src/commit/af3c78886fd8d4ca5eebdbe581a459a6f6d29d6a
:crying_cat_face: I could not find where `qsort_r()` is defined by the POSIX standard.

Up to the FreeBSD 13 series `qsort_r()` is only available in the first specification introduced.
https://man.freebsd.org/cgi/man.cgi?query=qsort_r&manpath=FreeBSD+13.3-RELEASE
The End-of-life for the FreeBSD 13 series is 2026-04-30.
https://www.freebsd.org/security/#sup

Notably, FreeBSD was the first to introduce `qsort_r()`, followed by GNU with an incompatible implementation of the same name with different arguments.
Apple brought `qsort_r()` from FreeBSD.
https://stackoverflow.com/a/39561369

Also, I have not actually checked, but it is possible that NetBSD and OpenBSD do not have both `qsort_r()` and `qsort_s()`.
At least not in the latest man page, and I could not find definitions in `src/include/stdlib.h`.
https://man.netbsd.org/qsort_r https://man.netbsd.org/qsort_s
https://man.openbsd.org/qsort_r https://man.openbsd.org/qsort_s

illumos seems to have a GNU-compatible `qsort_r()`, but does not seem to have `qsort_s()`.
https://github.com/illumos/illumos-gate/commit/44431c82ebd7ee1d7c240683235e728d70d96cf2

To make it portable without confusion and usable even in C99 (C89?), I think one option would be to implement it independently in zstd.

adenilsoncavalcanti:
@dearblue thanks a lot for the comments, I really appreciate it.

I've validated the current patch with a VM running FreeBSD 14.1, since I believe new software (e.g. upcoming zstd) should be validated in new OSes first.

I incorrectly assumed that all *BSD versions shared the same libc (note: I'm not a *BSD user or developer), but what you are reporting is that NetBSD + OpenBSD can have a diverging libc when compared to FreeBSD.

While a a qsort() function can be easily copied & pasted to the current patch, validating it is a bit more tricky: I'm aware that some OS vendors (e.g. Apple) *do* ship **heavily** optimized system libraries in their products (e.g. iOS, OSX, etc).

A replacement qsort() function would have to demonstrate similar or superior performance than what zstd currently uses (i.e. the default system libc on each OS/compiler) to avoid performance regressions.

That validation effort would further delay landing this fix to a real SEGFAULT in zstd thanks to what appears to be a *BSD idiosincrasy.

On top of that, I couldn't detect qsort() is a hotspot in a regular 'perf report'.

If it was a real hotspot in zstd it would be worthwhile the effort in optimizing it further (e.g. using SIMD techniques exploiting AVX-512 as deployed on https://github.com/intel/x86-simd-sort).

The current patch works fine on Linux, OSX and Windows (probably Android, I will test it soon), it seems that *BSD is the one giving us trouble at the moment.

Question: since you mentioned that 'Apple brought qsort_r() from FreeBSD'. If I treat *BSD as __APPLE__ would that work on the current stable versions of *BSD?

Cyan4973:
Note that strict `C90` compatibility is a must have for `libzstd`.

While it's allowed to use posix-specific symbols, it also means that some backup strategy must exist and be triggered automatically for other `C90` platforms without these facilities. We don't even need exact feature parity, it just needs to compile and run properly.

Adenilson:
@Cyan4973 to clarify, you suggest that if we can't get something working on all versions of *BSD, we should provide some kind of 'fallback' (e.g. a version of qsort_r()) to allow it to build cleanly. Is that correct?

lib/dictBuilder/cover.c
```diff
@@ -276,11 +279,15 @@ static int COVER_cmp8(COVER_ctx_t *ctx, const void *lp, const void *rp) {
...
 /**
  * Same as COVER_cmp() except ties are broken by pointer value
- * NOTE: g_coverCtx must be set to call this function.  A global is required because
- * qsort doesn't take an opaque pointer.
  */
-static int WIN_CDECL COVER_strict_cmp(const void *lp, const void *rp) {
-  int result = COVER_cmp(g_coverCtx, lp, rp);
+#if defined(_WIN32) && defined(_MSC_VER)
+static int WIN_CDECL COVER_strict_cmp(void* g_coverCtx, const void* lp, const void* rp) {
+#elif defined(__APPLE__)
```
Adenilson, line 292: I got test if treating *BSD as APPLE could do the trick...

lib/dictBuilder/cover.c
```diff
@@ -276,11 +279,15 @@ static int COVER_cmp8(COVER_ctx_t *ctx, const void *lp, const void *rp) {
...
+static int WIN_CDECL COVER_strict_cmp(void *g_coverCtx, const void *lp, const void *rp) {
+#elif defined(_GNU_SOURCE) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
```
Adenilson: Apparently this won't work on Net/OpenBSDs and older FreeBSD.

lib/dictBuilder/cover.c
```diff
@@ -620,17 +633,20 @@ static size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
...
+            ctx,
+            (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp));
+#elif defined(_GNU_SOURCE) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+    qsort_r(ctx->suffix, ctx->suffixSize, sizeof(U32),
+            (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp),
+            ctx);
+#elif defined(_WIN32) && defined(_MSC_VER)
+    qsort_s(ctx->suffix, ctx->suffixSize, sizeof(U32),
+            (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp),
+            ctx);
```
Adenilson, line 658: And we will have to add some kind of fallback for C90.

Cyan4973:
> @Cyan4973 to clarify, you suggest that if we can't get something working on all versions of *BSD, we should provide some kind of 'fallback' (e.g. a version of qsort_r()) to allow it to build cleanly. Is that correct?

Yes,
even losing the re-entrant property would be acceptable,
as long as it can run fine on a barebone C90-compliant system.

lib/dictBuilder/cover.c
```diff
@@ -21,8 +21,14 @@
 /*-*************************************
 *  Dependencies
 ***************************************/
+/* qsort_r is an extension. */
+#if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || defined(__CYGWIN__) || defined(__MSYS__)
+#define _GNU_SOURCE
```
Adenilson, line 27: Only define it *if* not already defined.

dearblue:
> Question: since you mentioned that 'Apple brought qsort_r() from FreeBSD'. If I treat *BSD as **APPLE** would that work on the current stable versions of *BSD?

No. I actually checked, and `qsort_r()` and `qsort_s()` are indeed not present, even on the latest NetBSD and OpenBSD.

lib/dictBuilder/cover.c
Adenilson: Done.

lib/dictBuilder/cover.c
Adenilson: It won't. Newer FreeBSD has the same syntax as GNU, but older has the same syntax as APPLE.

I will let go of FreeBSD for now and revisit the issue in a follow up patch.

lib/dictBuilder/cover.c
Adenilson: Done. The fallback for no is the same code as is being shipped by zstd.

In a follow up patch we can try to provide a reentrant qsort_r() for platforms that don't have it.

Adenilson:
@Cyan4973 apparently all bots in the CI are green with the latest version of the patchset.

I got two questions:
a) Do you wish to change anything on it?

b) Should I squash all the 5 commits in a single one?

lib/dictBuilder/cover.c
```diff
@@ -276,11 +286,17 @@ static int COVER_cmp8(COVER_ctx_t *ctx, const void *lp, const void *rp) {
 
 /**
  * Same as COVER_cmp() except ties are broken by pointer value
- * NOTE: g_coverCtx must be set to call this function.  A global is required because
- * qsort doesn't take an opaque pointer.
  */
+#if defined(_WIN32) && defined(_MSC_VER)
```
Cyan4973, line 290: The way `COVER_strict_cmp()` (and below `COVER_strict_cmp8()`) are implemented "feels" unusual : they share the same (short) body but feature different signatures depending on target system, controlled through preprocessor branching.
 
Given that the function signature is a mandated parameter of the `qsort_r()` function, and `qsort_r()` (of course!) varies depending on target system, I don't have a better solution to propose. Rather, I just forward some minor comments : 
    - `WIN_CDECL` is used even in non-Windows context. Granted, it's reduced to nothing in non-Windows environments, but seeing it listed for systems specifically tested as non Windows feels irrelevant
    - `_MSC_VER` and `__APPLE__` environments seem to share the same signature, so _maybe_ they could be merged (?)

Ultimately, this is a comment focused on "readability". The functionality itself looks good.

lib/dictBuilder/cover.c
```diff
@@ -620,14 +644,24 @@ static size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
     for (i = 0; i < ctx->suffixSize; ++i) {
       ctx->suffix[i] = i;
     }
-    /* qsort doesn't take an opaque pointer, so pass as a global.
-     * On OpenBSD qsort() is not guaranteed to be stable, their mergesort() is.
-     */
-    g_coverCtx = ctx;
-#if defined(__OpenBSD__)
+#if defined(__APPLE__)
```
Cyan4973, line 647: `qsort_r()` is invoked directly from within in `COVER_ctx_init()`, resulting in some `#if ... #elif ... #else ... ` preprocessor logic in the function's body, because `qsort_r()` is different depending on target system.
In a bid to lower preprocessor logic within `COVER_ctx_init()`, I was wondering if this design could be simplified by an invocation to a single function signature, and this function would be implemented differently depending on target system. 
Such a design requires a common signature across all systems, so there might be some limits there.

Cyan4973 (commented):
Thanks @Adenilson !
This is great work, and I think it works as intended.

I've focused my comments on source code maintainability, given that the functionality itself is already in a good place.

Cyan4973:
> Should I squash all the 5 commits in a single one?

Yes,
when some intermediate steps tend to break CI, it's better to remove them from the commit history, otherwise it will make future debugging `bisect` operations more problematic to manage.

Adenilson:
Github was refusing to accept a push of my local + rebased changes to remote, so I was forced to delete the remote branch to be able to push and as a result the merge request was purged.

The new one is at:
https://github.com/facebook/zstd/pull/4086

PR #4079, opened 2024-06-20
Throw error if Huffman weight initial states are truncated

### LINKED ISSUE #4037: Question about ZSTD protocole
Hi,
My question about ZSTD protocol is:
Can it possible on literal part FSE table, which contains 2 probabilities and only one weight (even but without odd) extracted from table?
the file I sent (after mv it to ForCheck.zst ) decompressed on tool version zstd-1.5.5 but not in former version zstd-1.5.2
[ForCheck.zst.txt](https://github.com/facebook/zstd/files/15159891/ForCheck.zst.txt)
thanks,
Shuli,

### LINKED ISSUE #4071: Possibly missing check for truncated initial states in Huffman weight block
Found another omission in the spec while trying to figure out if it was possible to encode only one Huffman weight in FSE mode, it appears that the answer is "no" but Zstd's behavior in such a situation is not well-defined.

`FSE_decompress_usingDTable_generic` will load 2 initial state values from the input stream, but doesn't do any bitstream overflow check.  With the current code paths, this results in the state being decoded as zero if it's fully or partially truncated, although the comments in `BIT_lookBits` state that the results of bit overreads are undefined (and the code path that is currently disabled by a preprocessor defs in `BIT_lookBits` will produce a different result).

This doesn't necessarily trigger an error due to the Huffman tree being malformed either, if the initial states decode to 0 and 1 (in either order) then it will specify a valid Huffman tree.

I'll try to get a sample that reproduces this behavior.

elasota:
Fixes #4071 and effectively fixes #4037 by considering this behavior to be invalid


doc/decompressor_permissive.md
```diff
@@ -58,3 +58,24 @@ This behavior has no consequence for the rest of the frame decoding process.
 
 In newer versions, the 2 reserved bits are actively checked for value zero,
 and the decoder reports a corruption error if they are not.
+
+
+
+Truncated Huffman states
```
Cyan4973: minor nit:
this file is supposed to be sorted in recency order, so this new paragraph would appear above existing "Non-zeroes reserved bits" paragraph.

doc/zstd_compression_format.md
```diff
@@ -1353,7 +1353,7 @@ and the second (`State2`) encodes the odd indexed symbols.
 decoding a single symbol and updating their state.
 For more details on these FSE operations, see the [FSE section](#fse).
 
-The number of symbols to decode is determined
+The number of symbols to decode is determined
```
Cyan4973: minor nit: extraneous space at the end

lib/common/fse_decompress.c
```diff
@@ -184,16 +184,22 @@ FORCE_INLINE_TEMPLATE size_t FSE_decompress_usingDTable_generic(
     FSE_DState_t state1;
     FSE_DState_t state2;
 
+    BIT_DStream_status reloadStat;
```
Cyan4973: If it was a hot portion of the code, I would be concerned by defining a variable outside of the loop, as it makes compiler optimizations a bit more difficult.
But since this code is only used to decompress Huffman weights, I presume any speed different is unlikely to be noticeable.

Cyan4973 (commented):
Thanks @elasota , 
that's pretty good work,
it adds clarity to a very deep erroneous scenario.

To be clear, the reference decoder doesn't guarantee that it will detect and report every case of invalid frame. It only guarantees that it will properly decode any *valid* frame, and for all others it will operate safely, meaning without allowing any kind of UB / segfaults issues.
That being said, whenever it's possible to detect and report such cases in a way that doesn't cost much (especially doesn't cost cpu runtime), then we should absolutely do it, and this update seems to fit this bill.

I just have a couple of nits, and it will good to go.

lib/common/fse_decompress.c
elasota: From an optimization standpoint, it was more about avoiding a redundant extra call to `BIT_reloadDStream` (since I don't think the compiler would be able to reliably tell that calling it multiple times consecutively would be a no-op) but I guess I'll change it back.

elasota:
I think this is good to have for safety reasons anyway since `BIT_readBits` is undefined for over-consuming bits, and I would interpret undefined as "not guaranteed to return an in-range value," which is unsafe in this situation.  I'm not sure what situation it would ever do such a thing, but this will take that off the table.

elasota:
Poking CI to re-run the check that failed due to server error.

Cyan4973 (approved):

PR #4080, opened 2024-06-20
[contrib/pzstd] Prevent hangs when there are errors

yotann:
When two threads are using a WorkQueue and the reader thread exits due to an error, it must call WorkQueue::finish() to wake up the writer thread. Otherwise, if the queue is full and the writer thread is waiting for a free slot, it could hang forever.

This can happen in pratice when decompressing a large, corrupted file that does not contain pzstd skippable frames.

### DIFF
--- a/contrib/pzstd/Pzstd.cpp
+++ b/contrib/pzstd/Pzstd.cpp
@@ -269,7 +269,10 @@ static void compress(
     std::shared_ptr<BufferWorkQueue> out,
     size_t maxInputSize) {
   auto& errorHolder = state.errorHolder;
-  auto guard = makeScopeGuard([&] { out->finish(); });
+  auto guard = makeScopeGuard([&] {
+    in->finish();
+    out->finish();
+  });
   // Initialize the CCtx
   auto ctx = state.cStreamPool->get();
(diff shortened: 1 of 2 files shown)


embg:
Why is `in->finish()` here not sufficient?

https://github.com/facebook/zstd/blob/5dc9c3cd5b694d80a57fb6eb080e535ecda1fe2c/contrib/pzstd/Pzstd.cpp#L405

I do agree that the changes in `writeFile()` are necessary.

embg:
OK, I see what's going on now. In both `asyncCompressChunks()` and `asyncDecompressFrames()`, the scope guard is in the same scope as `readData()`. So the scope guard destructor won't run if `readData()` is stuck, which can cause a deadlock.

The only remaining question I have is whether it's safe to `finish()` a WorkQueue twice. I see you added a code comment addressing this question. Will take a look on Monday.

embg:
OK, I don't see any reason why we can't call `finish()` twice on the same `WorkQueue`. Locking twice is safe, setting `done_` twice is safe, and `notify_all()` doesn't have any pre-conditions according to [C++ docs](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all).

LGTM.

terrelln:
Thanks for the PR @yotann & thanks for reviewing @embg!

PR #4086, opened 2024-07-01
[zstd][dict] Ensure that dictionary training functions are fully reen…

Adenilson:
…trant

The two main functions used for dictionary training using the COVER algorithm require initialization of a COVER_ctx_t where a call to qsort() is performed.

The issue is that the standard C99 qsort() function doesn't offer a way to pass an extra parameter for the comparison function callback (e.g. a pointer to a context) and currently zstd relies on a *global* static variable to hold a pointer to a context needed to perform the sort operation.

If a zstd library user invokes either ZDICT_trainFromBuffer_cover or ZDICT_optimizeTrainFromBuffer_cover from multiple threads, the global context may be overwritten before/during the call/execution to qsort() in the initialization of the COVER_ctx_t, thus yielding to crashes and other bad things (Tm) as reported on issue #4045.

Enters qsort_r(): it was designed to address precisely this situation, to quote from the documention [1]: "the comparison function does not need to use global variables to pass through arbitrary arguments, and is therefore reentrant and safe to use in threads."

It is available with small variations for multiple OSes (GNU, BSD[2], Windows[3]), and the ISO C11 [4] standard features on annex B-21 qsort_s() as part of the <stdlib.h>. Let's hope that compilers eventually catch up with it.

For now, we have to handle the small variations in function parameters for each platform.

The current fix solves the problem by allowing each executing thread pass its own COVER_ctx_t instance to qsort_r(), removing the use of a global pointer and allowing the code to be reentrant.

Unfortunately for *BSD, we cannot leverage qsort_r() given that its API has changed for newer versions of FreeBSD (14.0) and the other BSD variants (e.g. NetBSD, OpenBSD) don't implement it.

[1] https://man7.org/linux/man-pages/man3/qsort_r.3.html
[2] https://man.freebsd.org/cgi/man.cgi?query=qsort_r
[3] https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort-s?view=msvc-170
[4] https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -21,8 +21,17 @@
 /*-*************************************
 *  Dependencies
 ***************************************/
+/* qsort_r is an extension. */
+#if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || \
+    defined(__CYGWIN__) || defined(__MSYS__)
+#if !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
+#endif
+
 #include <stdio.h>  /* fprintf */
-#include <stdlib.h> /* malloc, free, qsort */
+#include <stdlib.h> /* malloc, free, qsort_r */
+
 #include <string.h> /* memset */
 #include <time.h>   /* clock */
 
@@ -232,8 +241,10 @@ typedef struct {
   unsigned d;
 } COVER_ctx_t;
 
-/* We need a global context for qsort... */
+#if !defined(_GNU_SOURCE) && !defined(__APPLE__) && !defined(_MSC_VER)
+/* C90 only offers qsort() that needs a global context. */
 static COVER_ctx_t *g_coverCtx = NULL;
+#endif
 
 /*-*************************************
 *  Helper functions
@@ -276,11 +287,15 @@ static int COVER_cmp8(COVER_ctx_t *ctx, const void *lp, const void *rp) {
 
 /**
  * Same as COVER_cmp() except ties are broken by pointer value
- * NOTE: g_coverCtx must be set to call this function.  A global is required because
- * qsort doesn't take an opaque pointer.
  */
-static int WIN_CDECL COVER_strict_cmp(const void *lp, const void *rp) {
-  int result = COVER_cmp(g_coverCtx, lp, rp);
+#if (defined(_WIN32) && defined(_MSC_VER)) || defined(__APPLE__)
+static int WIN_CDECL COVER_strict_cmp(void* g_coverCtx, const void* lp, const void* rp) {
+#elif defined(_GNU_SOURCE)
+static int COVER_strict_cmp(const void *lp, const void *rp, void *g_coverCtx) {
+#else /* C90 fallback.*/
+static int COVER_strict_cmp(const void *lp, const void *rp) {
+#endif
+  int result = COVER_cmp((COVER_ctx_t*)g_coverCtx, lp, rp);
   if (result == 0) {
     result = lp < rp ? -1 : 1;
   }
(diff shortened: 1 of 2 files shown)


Adenilson:
This is the latest version for #4078.

Adenilson:
Main changes are:
a) Reuse the same code for both WIN + APPLE when possible.
b) Only use WIN_CDECL where needed.
c) Rebased all changes in a single patch.
d) Better comments on zstd_deps.h.

lib/dictBuilder/cover.c
```diff
@@ -620,14 +641,24 @@ static size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
     for (i = 0; i < ctx->suffixSize; ++i) {
       ctx->suffix[i] = i;
     }
-    /* qsort doesn't take an opaque pointer, so pass as a global.
-     * On OpenBSD qsort() is not guaranteed to be stable, their mergesort() is.
-     */
-    g_coverCtx = ctx;
-#if defined(__OpenBSD__)
+#if defined(__APPLE__)
```
Adenilson: I wonder if it would make sense to move this block to its own new function?

Cyan4973: Yeah, it probably would help readability.

Adenilson: Done.

lib/dictBuilder/cover.c
```diff
@@ -620,14 +662,15 @@ static size_t COVER_ctx_init(COVER_ctx_t *ctx, const void *samplesBuffer,
...
     }
-    /* qsort doesn't take an opaque pointer, so pass as a global.
-     * On OpenBSD qsort() is not guaranteed to be stable, their mergesort() is.
-     */
+#if defined(_GNU_SOURCE) || defined(__APPLE__) || defined(_WIN32)
+    unify_qsort_r_interface(ctx);
+#elif defined(__OpenBSD__)
     g_coverCtx = ctx;
-#if defined(__OpenBSD__)
     mergesort(ctx->suffix, ctx->suffixSize, sizeof(U32),
```
Adenilson: Do we actually have real data showing the benefits for this special case?

Cyan4973: This change was implemented in #1100 following a bug report in #1088 .
Possibly an issue with reproducibility in some test scenarios. I don't remember all the details, maybe @terrelln can chime in to explain what sort stability is and why it matters.

Adenilson:
Let's see how the CI bots digest the latest patch version. :-)

lib/dictBuilder/cover.c
```diff
@@ -21,8 +21,17 @@
 /*-*************************************
 *  Dependencies
 ***************************************/
+/* qsort_r is an extension. */
+#if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || \
+    defined(__CYGWIN__) || defined(__MSYS__)
```
Adenilson, line 26: Mingw defines _WIN32 and __MINGW32/64__ but doesn't appear to have qsort_r(), i.e.
https://stackoverflow.com/questions/18796877/in-what-header-is-qsort-s-located-in-minggw-gcc-4-8-for-windows

lib/dictBuilder/cover.c
```diff
@@ -289,14 +304,43 @@ static int WIN_CDECL COVER_strict_cmp(const void *lp, const void *rp) {
...
   return result;
 }
 
+/**
+ * Abstract away divergence of qsort_r() parameters.
+ * Hopefully when C11 become the norm, we will be able
+ * to clean it up.
+ */
+#if defined(_GNU_SOURCE) || defined(__APPLE__) || (defined(_WIN32) && defined(_MSC_VER))
+static void unify_qsort_r_interface(COVER_ctx_t *ctx) {
```
Adenilson: Sounds verbose. Maybe 'qsort_r_proxy' would be better...

Cyan4973 (approved):

Adenilson:
@Cyan4973 thanks a lot for the review and comments, I really appreciate it.

I was pleasantly surprised  how the CI works (i.e. it really helps a lot to catch issues early on) and how good is the zstd code base (i.e. it is well written and things actually make sense in general).

I will start working on the next patch.
:-)

lib/dictBuilder/cover.c
Cyan4973: Another strategy could be to call it `stableSort()` or even `sort()` and bundle into it the `OpenBSD` and `C90` fallbacks.

Adenilson: I really like the idea, let me try to implement it!

Adenilson: Done!

Cyan4973 (approved):

Issue #4103, opened 2024-07-28
Building libzstd.a using NDK is broken


YCJDWCJI:
**Describe the bug:**

Since commit [345bcb5], building with NDK is broken, error:

```Text
dictBuilder/cover.c:332:5: error: call to undeclared function 'qsort_r'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

[345bcb5]: https://github.com/facebook/zstd/blob/345bcb5ff7001fbe2dfd59974808170c9e0f4d5d/lib/dictBuilder/cover.c

**To reproduce:**

1. Configure [NDK toolchain].
2. `git clone https://github.com/facebook/zstd`.
3. `cd zstd/lib`. 
4. `make libzstd.a`.
5. See error.

[NDK toolchain]: https://developer.android.com/ndk/guides/other_build_systems

**Expected behavior:**

The build process should complete successfully and produce the **libzstd.a** file.

**Desktop:**

 - OS: [Arch Linux] [Testing]
 - Compiler: aarch64-linux-android31-clang
 - Build system: makefile

[Arch Linux]: https://archlinux.org
[Testing]: https://wiki.archlinux.org/title/Official_repositories#Testing_repositories

**Temporary Solution:**

Using **C90 fallback** under the `_GNU_SOURCE` macro.

Cyan4973:
cc @Adenilson

Adenilson:
I will proceed to repro and fix this one.

Adenilson:
I managed to repro the bug report by installing the NDK r27 (https://developer.android.com/ndk/downloads) and trying to build by exporting CC and CXX variables to point to the aarch64 cross compiler.

The output was:

//-------------------------------------------------------------------------------------------------------------------------------------------
CC obj/conf_e35c1c02789d063e7f14d8327851243f/dynamic/divsufsort.o
/home/adenilson/zstd_fork/lib//dictBuilder/cover.c:332:5: error: call to undeclared function 'qsort_r'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  332 |     qsort_r(ctx->suffix, ctx->suffixSize, sizeof(U32),
      |     ^
/home/adenilson/zstd_fork/lib//dictBuilder/cover.c:332:5: note: did you mean 'qsort'?
[...]

PR #4107, opened 2024-07-30
[zstd][android] Fix build with NDK r27

Adenilson:
The NDK cross compiler declares the target as __linux (which is not technically incorrect), which triggers the enablement of _GNU_SOURCE in the newly added code that requires the presence of qsort_r() used in the COVER dictionary code.

Even though the NDK uses llvm/libc, it doesn't declare qsort_r() in the stdlib.h header.

The build fix is to only activate the _GNU_SOURCE macro if the OS is *not* Android, as then we will fallback to the C90 compliant code.

This patch should solve the reported issue number #4103.

### DIFF
--- a/lib/common/zstd_deps.h
+++ b/lib/common/zstd_deps.h
@@ -31,7 +31,7 @@
  */
 #if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || \
     defined(__CYGWIN__) || defined(__MSYS__)
-#if !defined(_GNU_SOURCE)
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Adenilson:
@Cyan4973 I've reached out to the Android folks at google and they started the work to actually fix the NDK/AOSP in upcoming future versions:
https://android-review.googlesource.com/c/platform/bionic/+/3199671

The function signature should be the same as on GNU/Linux systems.

Adenilson:
So hopefully in the future when API level 36 is released to the wild, we could actually enable the qsort_r() fix for that API level and fallback to the C90 code for older versions.

Cyan4973:
Looks good @Adenilson !

PR #4109, opened 2024-07-30
added android-ndk-build

Cyan4973:
Adding an Android NDK build test.

It fails, as expected, due to lack of `qsort_r()` symbol.

The issue should be fixed after rebasing on top of @Adenilson 's #4107


Adenilson:
That is pretty awesome!

Cyan4973:
And as expected, the new `NDK` test works fine after integrating #4107.
So we have a good new test, able to catch similar `Android NDK` compilation issues in the future.

Issue #4069, opened 2024-06-06
Enable weak symbol support for Risc-V?


guoyuqi020:
I am using zstd on Risc-V machines. I noticed that zstd_trace is disabled on Risc-V. After some investigating, I found this was because the zstd's weak symbol support was conservatively disabled on Risc-V.

However, as far as I know, Risc-V supports weak symbols, and the `weak` attribute can be used on Risc-V’s GCC.

So, will you consider enabling weak symbol support for Risc-V? 

I guess adding a `defined (__riscv)` in the following code will make it work.

```c
// lib/common/zstd_trace.h

/* weak symbol support
 * For now, enable conservatively:
 * - Only GNUC
 * - Only ELF
 * - Only x86-64, i386 and aarch64
 * Also, explicitly disable on platforms known not to work so they aren't
 * forgotten in the future.
 */
#if !defined(ZSTD_HAVE_WEAK_SYMBOLS) && \
    defined(__GNUC__) && defined(__ELF__) && \
    (defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86) || defined(__aarch64__)) && \
    !defined(__APPLE__) && !defined(_WIN32) && !defined(__MINGW32__) && \
    !defined(__CYGWIN__) && !defined(_AIX)
#  define ZSTD_HAVE_WEAK_SYMBOLS 1
#else
#  define ZSTD_HAVE_WEAK_SYMBOLS 0
#endif
#if ZSTD_HAVE_WEAK_SYMBOLS
#  define ZSTD_WEAK_ATTR __attribute__((__weak__))
#else
#  define ZSTD_WEAK_ATTR
#endif
```

adenilsoncavalcanti:
I can help with this one, I got access to RVV 1.0 compliant RISC-V boards to test.

guoyuqi020:
Seems this has been solved. Now I close this issue.

PR #4114, opened 2024-08-07
[riscv] Enable support for weak symbols

Adenilson:
Both gcc and clang support weak symbols on RISC-V, therefore let's enable it.

This should fix issue #4069.

### DIFF
--- a/lib/common/zstd_trace.h
+++ b/lib/common/zstd_trace.h
@@ -21,13 +21,14 @@ extern "C" {
  * For now, enable conservatively:
  * - Only GNUC
  * - Only ELF
- * - Only x86-64, i386 and aarch64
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973 (approved):
I presume this change is automatically tested as part of : 
https://github.com/facebook/zstd/actions/runs/10275661201/job/28434730175?pr=4114

In which case, it looks all good to me

Adenilson:
@Cyan4973 I also compiled/run zstd in a real RISC-V board (built with both gcc 13.2.1 and clang-18) to ensure that it should be all fine.

That being said, I think it is pretty cool that zstd already has a RISC-V test/bot.

Issue #4112, opened 2024-08-06
There is a memory leak defect at line 3466 in the file /zstd/lib/compress/zstd_compress.c.


LuMingYinDetect:
A pointer variable named dst is defined at line 3461 in the file /zstd/lib/compress/zstd_compress.c and a new dynamic memory area is allocated to it using the function ZSTD_customMalloc. If the RETURN_ERROR_IF macro condition at line 3466 is met, the function will return at this line, and the release operation of the dynamic memory area pointed to by dst at line 3484 will not be executed, resulting in a memory leak, as shown in the figure below:
https://github.com/LuMingYinDetect/zstd_defects/blob/main/zstd_1.png

Adenilson:
Perhaps something like this could help:
https://gist.github.com/Adenilson/7d822b4223ca11753a1c632cd7dee9bc

terrelln:
Fixed by https://github.com/facebook/zstd/pull/4115

PR #4115, opened 2024-08-07
[zstd][leak] Avoid memory leak on early return of ZSTD_generateSequence

Adenilson:
Sanity checks on a few of the context parameters (i.e. workers and block size) may prompt an early return on ZSTD_generateSequences.

Allocating the destination buffer past those return points avoids a potential memory leak.

This patch should fix issue #4112.


Cyan4973 (approved):

terrelln:
Fixes #https://github.com/facebook/zstd/issues/4112

PR #4116, opened 2024-08-07
Disallow 32-bit mode in clang section

medns:
Fix register %rbx is only available in 64-bit mode

### DIFF
--- a/lib/common/cpu.h
+++ b/lib/common/cpu.h
@@ -34,8 +34,8 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
     U32 f1d = 0;
     U32 f7b = 0;
     U32 f7c = 0;
-#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))
-#if !defined(__clang__) || __clang_major__ >= 16
+#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))                 \
+    && (!defined(__clang__) || __clang_major__ >= 16)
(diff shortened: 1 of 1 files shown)


lib/common/cpu.h
```diff
@@ -34,8 +34,8 @@ MEM_STATIC ZSTD_cpuid_t ZSTD_cpuid(void) {
     U32 f1d = 0;
     U32 f7b = 0;
     U32 f7c = 0;
-#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))
-#if !defined(__clang__) || __clang_major__ >= 16
+#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))                 \
+    && (!defined(__clang__) || __clang_major__ >= 16)
```
terrelln, line 38: As-is this leaves the CPUID for MSVC 32-bit builds unset. Instead, based on https://reviews.llvm.org/D101338, it looks like the clang bug is only present in 64-bit builds. So we could do

```
#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))
#if !defined(_M_X64) || !defined(__clang__) || __clang_major__ >= 16
...
#else
...
#endif
```

terrelln (changes requested):
Thanks for flagging the issue!

lib/common/cpu.h
medns: Ok, I will change it

medns:
I will close it, new PR in #4117 
@terrelln

PR #4118, opened 2024-08-08
Disallow 32-bit mode in clang section

medns:
Fix register %rbx is only available in 64-bit mode

The previous PR is in #4116


medns:
Waiting for @terrelln review ...

terrelln:
Thanks for the PR @medns, and sorry for the delay! This looks good to me!

PR #4119, opened 2024-08-08
Fix zstd-pgo run error

xionghul:
The -Werror=missing-profile caused thread/zlib/lzma/lz4 detection failure when build with profile-use, thus caused ZSTD_MULTITHREAD etc. is not defined for profile-use, then there will be many profile mismatch information in output and the final binary reports run error sometimes as below:

Error : ZSTD_CCtx_setParameter(ctx, ZSTD_c_nbWorkers, adv->nbWorkers) failed : Unsupported parameter


xionghul:
@facebook-github-bot signed CLA already, why still fail?

Cyan4973 (approved):

PR #4125, opened 2024-08-13
It is recommended to use emplace_back instead of push_back.

monster29000:


terrelln:
`emplace_back()` is only needed when you are constructing the object in place. It doesn't seem like any of these places are doing that. If you can show me a perf difference, we can reconsider, but I don't expect to see one.

PR #4134, opened 2024-08-31
Fixed warning when compiling pzstd with CPPFLAGS=-Wunused-result and …

### LINKED ISSUE #4133: Build Warning with -Wunused-return on gcc
**Describe the bug**
When building with `-std=c++17` and `-Wunused-return`, gcc issues a warning.
I noticed this when building the Arch Linux build, but I reproduced it when building the `dev` branch as well.

I read the note about static analysis in the contribution guidelines. Using something like `std::ignore` would clarify that the return value here is irrelevant.

**To Reproduce**
Steps to reproduce the behavior (with `make`):
1. `cd contrib/pzstd`
2. Comment out setting `PZSTD_CXX_STD` in `Makefile`
3. `export CPPFLAGS="-Wunused-result"`
4. `export CXXFLAGS="-std=c++17"`
5. `make`

**Expected behavior**
There should be no compiler warnings when building.

**Screenshots and charts**
2024-08-31T09:32:41,508334791-05:00_blured

 - OS: Arch Linux with kernel  6.10.3-zen1-2-zen
 - Version: 1.5.6 and dev branch
 - Compiler: gcc 14.2.1
 - Build system: CMake

**Additional context**

inventor500:
This should close #4133.
This problem is not particularly important, since this was a build warning and does not seem to effect anything, but it does remove a warning when compiling on GCC.

### DIFF
--- a/contrib/pzstd/Options.cpp
+++ b/contrib/pzstd/Options.cpp
@@ -322,7 +322,7 @@ Options::Status Options::parse(int argc, const char **argv) {
   g_utilDisplayLevel = verbosity;
   // Remove local input files that are symbolic links
   if (!followLinks) {
-	  std::ignore = std::remove_if(localInputFiles.begin(), localInputFiles.end(),
+	  localInputFiles.erase(std::remove_if(localInputFiles.begin(), localInputFiles.end(),
                      [&](const char *path) {
                         bool isLink = UTIL_isLink(path);
                         if (isLink && verbosity >= 2) {
@@ -332,7 +332,7 @@ Options::Status Options::parse(int argc, const char **argv) {
                                     path);
                         }
                         return isLink;
-                    });
+					 }), localInputFiles.end());
   }
 
   // Translate input files/directories into files to (de)compress
--- a/contrib/pzstd/Options.cpp
+++ b/contrib/pzstd/Options.cpp
@@ -322,7 +322,7 @@ Options::Status Options::parse(int argc, const char **argv) {
   g_utilDisplayLevel = verbosity;
   // Remove local input files that are symbolic links
   if (!followLinks) {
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

neheb:
I don't think this is right. The code seems to be wanting to do https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

Meaning

```
--- a/contrib/pzstd/Options.cpp
+++ b/contrib/pzstd/Options.cpp
@@ -322,7 +322,7 @@ Options::Status Options::parse(int argc, const char **argv) {
   g_utilDisplayLevel = verbosity;
   // Remove local input files that are symbolic links
   if (!followLinks) {
-      std::remove_if(localInputFiles.begin(), localInputFiles.end(),
+      localInputFiles.erase(std::remove_if(localInputFiles.begin(), localInputFiles.end(),
                      [&](const char *path) {
                         bool isLink = UTIL_isLink(path);
                         if (isLink && verbosity >= 2) {
@@ -332,7 +332,7 @@ Options::Status Options::parse(int argc, const char **argv) {
                                     path);
                         }
                         return isLink;
-                    });
+                    }), localInputFiles.end());
   }
 
   // Translate input files/directories into files to (de)compress
```

is the proper solution.

It doesn't affect anything since this is contrib and not really used.

edit: as the wiki page states, this should be using std::erase_if and not std::remove_if. But that needs C++20.

terrelln (changes requested):
Thanks for the PR! Please make the change that @neheb is suggesting.

terrelln (approved):

terrelln:
Thanks for the bug fix @inventor500 & @neheb!

PR #4136, opened 2024-09-03
Block splitter

Cyan4973:
Instead of ingesting full blocks only (128 KB), 
make an a-priori analysis of the data, 
and infer a position to split a block at a more appropriate boundary.

This can notably happen in an archive scenario, at the boundary between 2 files of different nature within the archive.

This leads to some non-trivial compression gains, for a correspondingly acceptable speed cost.
The benefit is higher when there isn't already a post-splitter (like in higher `btopt` levels and above (16+)), 
but even when a post-splitter is active, there is _still_ some small compression ratio benefit, making this strategy desirable even for higher compression modes.

However, this input analysis is not free. 
The initial variant is currently reserved for higher compression strategies (>= `btopt`) where it combines well with the post-splitter operating on known sequences.
A second, faster variant, is applied to middle strategies `btlazy2` and `lazy2`, where it improves compressed size by ~300 KB. It achieves its faster speed by sampling the input, instead of scanning it sequentially.

For faster modes, the analysis is skipped, and replaced by a static split size (no longer limited to 128 KB only). Through tests, it appears that a static 92 KB block size brings higher compression ratio, at a small-to-negligible compression speed loss (mostly due to increased nb of blocks, hence of block headers).

Here are some benchmarks, focusing on compression savings:

`silesia.tar` :
| level | `dev` | this `PR` | savings | note |
| 1 | 73422432 | 73324476 | -97956 | static block size
| 7 | 60546773 | 60488311 | -58462 | static block size (last)
| 8 | 60015264 | 59700036 | -315228 | dynamic sampling (first)
| 15 | 57176243 | 56867697 | -308546 | dynamic sampling (last)
| 16 | 55338671 | 55256977 | -81694 | full splitting analysis (pre+post)
| 19 | 52887208 | 52842629 |  -44579 | idem
| 22 | 52322760 | 52284324 | -38436 | idem

`calgary.tar` : 
| level | `dev` | this `PR` | savings | note |
| 1 | 1143607 | 1129229 | -14378 | static block size
| 7 | 948081 | 946941 | -1140 | static block size (last)
| 8 | 940280 | 932212 | -8068 | dynamic sampling (first)
| 15 | 921606 | 911627 | -9979 | dynamic sampling (last)
| 16 | 882353 | 881217 | -1136 | full splitting analysis (pre+post)
| 19 | 861336 | 859737|  -1599 | idem
| 22 | 861200 | 859578| -1622 | idem
 

Follow up : 
- Consider multiple variants, with incremental speed / accuracy trade-offs
- Make the choice of `splitting` strategy selectable via compression parameter


Cyan4973:
mmmh,
this could be a problem:
```
test 45 : in-place decompression : Error : test (margin <= ZSTD_DECOMPRESSION_MARGIN(CNBuffSize, ZSTD_BLOCKSIZE_MAX)) failed
```

I suspect the problem is that the macro `ZSTD_DECOMPRESSION_MARGIN()` may make an assumption about block sizes being always full (128 KB), and derive an assumption about maximum expansion, which is no longer respected when block sizes are 92 KB by default (lower compression levels). In contrast, `ZSTD_decompressionMargin()` scans the compressed data, hence it probably ends up with a different result, and now both results differ.

I'm not completely sure what's the wanted behavior here...

_edit_: confirmed that, when I change the default block size to anything other than 128 KB, it breaks this test.
On the other hand, `ZSTD_DECOMPRESSION_MARGIN()` macro requires a parameter `blockSize`, and the test passes `ZSTD_BLOCKSIZE_MAX`, aka 128 KB, so it's pretty clear what it's expecting.
So the question is: how could this test be passed "reasonably", i.e. without inner knowledge of how the internal block splitting decision is done ?

Cyan4973:
Who knew adding a single source file (`zstd_preSplit.c`) would be such a big problem...

currently blocked trying to get the single-file library builder to work, 
it doesn't include the new file, resulting in a link stage error.

and then each and every build system also requires updating its own list of files in its own format and location.

Cyan4973:
Weird stuff :
```
error: undefined reference to '__mulodi4'
```

It only happens during compilation of the [`clang-asan-ubsan-fuzz32`](https://github.com/facebook/zstd/actions/runs/11394316878/job/31704403992?pr=4136#logs) test, aka `undefined` sanitizer enabled for 32-bit compilation on `clang` (same test with `gcc` compiles fine).

The failure seems to correspond to where `*` multiplication operations on `S64` aka `long long` variables happen.

And of course, it happens all the time on github CI, but not on any other system I can test the same code and build rule with.

Cyan4973:
All tests passed, ready for review

.github/workflows/dev-long-tests.yml
```diff
@@ -156,15 +156,30 @@ jobs:
...
-    - name: clang + ASan + UBSan + Fuzz Test 32bit
+    - name: clang + ASan + Fuzz Test 32bit
       run: |
         sudo apt-get -qqq update
         make libc6install
-        CC=clang CFLAGS="-O3 -m32" FUZZER_FLAGS="--long-tests" make uasan-fuzztest
+        CC=clang CFLAGS="-O3 -m32" FUZZER_FLAGS="--long-tests" make asan-fuzztest
+
+# The following test seems to have issues on github CI specifically,
+# it does not provide the `__mulodi4` instruction emulation
```
terrelln, line 170: Does it have issues before this PR, or only after? If it is the latter, this might also have issues in the kernel, and we would need to do something similar to what we do with `ZSTD_div64()`

https://github.com/facebook/zstd/blob/b880f20d52a925ebee373b5050c206ba325d935d/lib/common/zstd_deps.h#L82

lib/common/compiler.h
```diff
@@ -278,6 +278,8 @@
 *  Alignment check
 *****************************************************************/
 
+#define ZSTD_IS_POWER_2(a) (((a) & ((a)-1)) == 0)
```
terrelln: nit: We don't need this for compile time constants like array bounds, could we make this a function `ZSTD_isPow2()`?

lib/compress/zstd_compress.c
```diff
@@ -4484,6 +4489,27 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms,
...
+    /* dynamic splitting has a cpu cost for analysis,
+     * due to that cost it's only used for btlazy2+ strategies */
+    if (strat >= ZSTD_btlazy2)
+        return ZSTD_splitBlock_4k(src, srcSize, blockSizeMax, cctx->tmpWorkspace, cctx->tmpWkspSize);
+    /* blind split strategy
+     * no cpu cost, but can over-split homegeneous data.
+     * heuristic, tested as being "generally better".
+     * do not split incompressible data though: respect the 3 bytes per block overhead limit.
+     */
+    return savings ? 92 KB : 128 KB;
```
terrelln: This looks like it is splitting if the savings is non-zero. So it will split if the savings is negative. Is that correct? If so why do we want to do that?

lib/compress/zstd_compress.c
```diff
@@ -4496,29 +4522,31 @@ static size_t ZSTD_compress_frameChunk(ZSTD_CCtx* cctx,
...
-    DEBUGLOG(4, "ZSTD_compress_frameChunk (blockSize=%u)", (unsigned)blockSize);
+    DEBUGLOG(4, "ZSTD_compress_frameChunk (blockSizeMax=%u)", (unsigned)blockSizeMax);
     if (cctx->appliedParams.fParams.checksumFlag && srcSize)
         XXH64_update(&cctx->xxhState, src, srcSize);
 
     while (remaining) {
         ZSTD_matchState_t* const ms = &cctx->blockState.matchState;
-        U32 const lastBlock = lastFrameChunk & (blockSize >= remaining);
+        U32 const lastBlock = lastFrameChunk & (blockSizeMax >= remaining);
+        size_t const blockSize = ZSTD_optimalBlockSize(cctx, ip, remaining, blockSizeMax, cctx->appliedParams.cParams.strategy, savings);
```
terrelln, line 4548: This seems to rely on a non-obvious relationship for correctness.

`ZSTD_optimalBlockSplit()` will not split when `remaining <= blockSizeMax`. But if it ever decides to split when that condition holds, then `lastBlock` will no longer be correct.

Can we fix this so that `lastBlock` is correct regardless of what `ZSTD_optimalBlockSize()` decides to do?

lib/compress/zstd_compress.c
```diff
@@ -4484,6 +4489,27 @@ static void ZSTD_overflowCorrectIfNeeded(ZSTD_matchState_t* ms,
...
 }
 
+#include "zstd_preSplit.h"
+
+static size_t ZSTD_optimalBlockSize(ZSTD_CCtx* cctx, const void* src, size_t srcSize, size_t blockSizeMax, ZSTD_strategy strat, S64 savings)
+{
+    /* note: conservatively only split full blocks (128 KB) currently,
+     * and even then only if there is more than 128 KB input remaining.
+     */
+    if (srcSize <= 128 KB || blockSizeMax < 128 KB)
```
terrelln: Does this mean that the block splitter won't work in streaming mode? What if the user passes in exactly 128KB at a time?

lib/compress/zstd_compress.c
```diff
@@ -4556,7 +4584,7 @@ static size_t ZSTD_compress_frameChunk(ZSTD_CCtx* cctx,
                 }
             }  /* if (ZSTD_useTargetCBlockSize(&cctx->appliedParams))*/
 
-
+            if (cSize < blockSize) savings += (blockSize - cSize);
```
terrelln: Do we also need to subtract savings if we grow the block?

If we aren't precisely tracking the `savings` I can see cases where an adversarial pattern could cause compression to fail.

lib/compress/zstd_cwksp.h
```diff
@@ -520,6 +525,20 @@ MEM_STATIC void* ZSTD_cwksp_reserve_object(ZSTD_cwksp* ws, size_t bytes)
 
     return alloc;
 }
+/**
+ * with alignment control
+ * Note : should happen only once, at workspace first initialization
+ */
+MEM_STATIC void* ZSTD_cwksp_reserve_object_aligned(ZSTD_cwksp* ws, size_t byteSize, size_t alignment)
```
terrelln, line 537: Why do we need alignment smaller than `64`? It seems it might be simpler to just keep 64-byte alignment on all allocations.

terrelln (changes requested):
Reviewed everything except `zstd_preSplit.[hc]`

lib/compress/zstd_preSplit.c
```diff
@@ -0,0 +1,157 @@
...
+static unsigned hash2(const void *p)
+{
+    return (U32)(MEM_read16(p)) * KNUTH >> (32 - HASHLOG);
+}
+
+
+typedef struct {
+  int events[HASHTABLESIZE];
+  S64 nbEvents;
+} FingerPrint;
```
terrelln: nit: Just one word

```suggestion
} Fingerprint;
```

lib/compress/zstd_preSplit.c
```diff
@@ -0,0 +1,157 @@
...
+    for (n = 0; n < HASHTABLESIZE; n++) {
+        assert(acc->events[n] >= slice->events[n]);
+        acc->events[n] -= slice->events[n];
+    }
+    acc->nbEvents -= slice->nbEvents;
+}
+
+#define CHUNKSIZE (8 << 10)
+/* Note: technically, we use CHUNKSIZE, so that's 8 KB */
+size_t ZSTD_splitBlock_4k(const void* src, size_t srcSize,
```
terrelln: nit: Can we pick a better name, since as noted we're splitting by 8 KB currently? Maybe `ZSTD_splitBlock_byFixedChunks()`

lib/compress/zstd_preSplit.c
```diff
@@ -0,0 +1,157 @@
...
+    assert(blockSizeMax == (128 << 10));
+    assert(workspace != NULL);
+    assert((size_t)workspace % ZSTD_ALIGNOF(FPStats) == 0);
+    ZSTD_STATIC_ASSERT(ZSTD_SLIPBLOCK_WORKSPACESIZE == sizeof(FPStats));
+    assert(wkspSize >= sizeof(FPStats)); (void)wkspSize;
+
+    initStats(fpstats);
+    recordFingerprint(&fpstats->pastEvents, p, CHUNKSIZE);
+    for (pos = CHUNKSIZE; pos < blockSizeMax; pos += CHUNKSIZE) {
+        assert(pos <= blockSizeMax - CHUNKSIZE);
```
terrelln: I know this holds because `blockSizeMax == (128 << 10)`, but can we fix the looping logic to guarantee this instead?

It would be cleaner to have this function support any `blockSizeMax`. Given that the checks in outer loop isn't super performance critical, I'd prefer clarity and explicit checks rather than assumptions.

lib/compress/zstd_preSplit.c
```diff
@@ -0,0 +1,157 @@
...
+        recordFingerprint(&fpstats->newEvents, p + pos, CHUNKSIZE);
+        if (compareFingerprints(&fpstats->pastEvents, &fpstats->newEvents, penalty)) {
+            return pos;
+        } else {
+            mergeEvents(&fpstats->pastEvents, &fpstats->newEvents);
+            ZSTD_memset(&fpstats->newEvents, 0, sizeof(fpstats->newEvents));
+            penalty = penalty - 1 + (penalty == 0);
```
terrelln: nit: Lets rewrite for clarity
```suggestion
            if (penalty > 0) {
                --penalty;
            }
```

.github/workflows/dev-long-tests.yml
terrelln: It looks like this might be a compiler bug.

Cyan4973: Yes, the issue doesn't reproduce on my local linux workstation.
I suspect a bug in the version of `ubsan` for `clang` currently deployed within github ci.

It's likely that this issue will be fixed at some unspecified time in the future.
We just can't wait for this issue to be fixed, the CI needs to continue running.
We'll re-enable this test when it can work properly on Github CI again.

Also, dropping temporarily the `ubsan` test for `clang` for 32-bit `x86` targets seems like a minor inconvenience, given that: 
- we mostly care about 64-bit `x64`
- `ubsan` 32-bit is still running in CI with `gcc`
- we keep the more important `asan` test for 32-bit on `clang`

lib/compress/zstd_cwksp.h
```diff
@@ -272,7 +276,7 @@ MEM_STATIC size_t ZSTD_cwksp_bytes_to_align_ptr(void* ptr, const size_t alignByt
  * which we can allocate from the end of the workspace.
  */
 MEM_STATIC void*  ZSTD_cwksp_initialAllocStart(ZSTD_cwksp* ws) {
-    return (void*)((size_t)ws->workspaceEnd & ~(ZSTD_CWKSP_ALIGNMENT_BYTES-1));
+    return (void*)((size_t)ws->workspaceEnd & (size_t)~(ZSTD_CWKSP_ALIGNMENT_BYTES-1));
```
felixhandte: I don't think this cast is actually correct. You need to cast inside the not expression if your goal is to ensure that the not doesn't happen to a narrower type that then gets zero-extended during promotion to the `size_t`.

Also should we be casting to `intptr_t`, not `size_t`?

lib/compress/zstd_compress.c
Cyan4973: Nope, `savings` will only be `>0` if there were some observed savings (aka, smaller `cSize` than `srcSize`) from previous blocks.

It also means that, for the first block, it will never split.

lib/compress/zstd_compress.c
Cyan4973: Indeed, if the user passes exactly 128 KB at a time, block splitting will not be triggered.

lib/compress/zstd_compress.c
```diff
@@ -137,11 +137,12 @@ ZSTD_CCtx* ZSTD_initStaticCCtx(void* workspace, size_t workspaceSize)
     ZSTD_cwksp_move(&cctx->workspace, &ws);
     cctx->staticSize = workspaceSize;
 
-    /* statically sized space. entropyWorkspace never moves (but prev/next block swap places) */
+    /* statically sized space. tmpWorkspace never moves (but prev/next block swap places) */
     if (!ZSTD_cwksp_check_available(&cctx->workspace, ENTROPY_WORKSPACE_SIZE + 2 * sizeof(ZSTD_compressedBlockState_t))) return NULL;
```
felixhandte: Don't we basically have to replace every usage of `ENTROPY_WORKSPACE_SIZE` with `TMP_WORKSPACE_SIZE`? Including here?

lib/compress/zstd_cwksp.h
Cyan4973: I guess it could also be done this way.
As align-8 is sufficient, align-64 felt overkill.

lib/compress/zstd_preSplit.c
Cyan4973: Yes, this name is fixed in the next patch.

lib/compress/zstd_cwksp.h
felixhandte: Agreed. Even if this is a desirable change (why?), this seems like it could be done in a different PR.

lib/common/compiler.h
Cyan4973: done

lib/compress/zstd_compress.c
Cyan4973: good catch !
fixed by checking `lastBlock` after determining `blockSize`

lib/compress/zstd_compress.c
Cyan4973: OK, we can add some malus when a block is not compressed.
This will be a conservative over-estimate, but that's not important, we just want to make sure there is no possible scenario that could expand data by more than 3-bytes per full block.

lib/compress/zstd_preSplit.c
Cyan4973: I went ahead and merged the next patch (sample5) in this feature branch,
so that it also updates the naming convention in the process.

lib/compress/zstd_preSplit.c
Cyan4973: done

lib/compress/zstd_cwksp.h
Cyan4973: > Also should we be casting to `intptr_t`, not `size_t`?

Unfortunately, `intptr_t` is not guaranteed to exist, while `size_t` is.

Also, since this is just for alignment checks, we just care about the last bits (realistically, the last 6 bits maximum for cache-line alignment). So we could even use a smaller type. But `size_t` is good enough already.

Cyan4973: rewrote `ZSTD_cwksp_initialAllocStart()` for clarity

lib/compress/zstd_compress.c
Cyan4973: Not every instance, since in some cases it's really "just" about entropy,
but quite possibly here.
This might be a left over.

_edit_ :
yes, it was a left over, to be fixed. 
Great catch @felixhandte !

lib/compress/zstd_cwksp.h
Cyan4973: The new `TMP_WORKSPACE` requires an 8-bytes alignment.
The original `ZSTD_cwksp_reserve_object()` that was used here was aligned on `sizeof(void*)`.
It was breaking tests on 32-bit systems.

lib/compress/zstd_preSplit.c
Cyan4973: done

felixhandte (commented):
Yeah, I think you should be able to avoid all of these cwksp changes if you just switch the entropy workspace allocation from being in the objects group (and allocation phase) to being in the aligned group.

This should just mean moving the allocation down into `ZSTD_reset_matchState()` [where we do the other aligned allocations](https://github.com/facebook/zstd/blob/b880f20d52a925ebee373b5050c206ba325d935d/lib/compress/zstd_compress.c#L2053) as well as changing:

https://github.com/facebook/zstd/blob/b880f20d52a925ebee373b5050c206ba325d935d/lib/compress/zstd_compress.c#L1709

to use `ZSTD_cwksp_aligned_alloc_size()`.

The `ZSTD_SLIPBLOCK_WORKSPACESIZE` is 8208, which is not an even divisor of 64. So it does mean we'll grow the workspace to be a multiple of 64 and throw away 48 bytes. We could look at improving this in a follow-up PR.

Cyan4973:
Addressed the last comment by removing the need for 64-bit members in the workspace structure, so that it doesn't need 8-bytes alignment. Now, `TMP_WORKSPACE` uses the same `reserve_object()` method and alignment rule (`sizeof(void*)`) as was previously used by `ENTROPY_WORKSPACE`, and is therefore just a size change.

Note that signed 64-bit multiplications are still needed for the distance function, so this doesn't address the problem of a bug in the Github CI `ubsan` version for `clang` in 32-bit compilation mode, which remains disabled for the time being.

Cyan4973:
All comments addressed

lib/compress/zstd_compress.c
terrelln: I don't think that is a good idea. In my model of the streaming API, how you chunk your data doesn't impact the compression, with the exception of the final block. This seems like it could introduce confusing behavior. Can we get rid of this restriction?

lib/compress/zstd_compress.c
terrelln: But this currently will also split if the savings is negative, which doesn't seem like the desired behavior. Is this what you intended?

```suggestion
    return savings > 0 ? 92 KB : 128 KB;
```

terrelln: Additionally, we need to compute how much overhead this can possibly add, and ensure that this block split can never grow us beyond the compress bound.

E.g.

```suggestion
    return savings >= 3 ? 92 KB : 128 KB;
```

lib/compress/zstd_compress.c
terrelln: > OK, we can add some malus when a block is not compressed.

I don't like this solution because it is a patch that hides the problem. But I don't see how it guarantees that it can never occur. I'd prefer to precisely calculate the `savings` so we can know exactly if we are allowed to split or not.

terrelln (changes requested):
The logic in `ZSTD_compress_frameChunk()` doesn't look like it **guarantees** that it respects `ZSTD_compressBound()`. It looks like it is very likely to respect the bound. But I'm positive that the fuzzers will eventually find a case where it doesn't.

lib/compress/zstd_compress.c
Cyan4973: OK, I'll look into it.

Cyan4973: All full blocks are now sent to the splitter function, irrespective of the presence of additional data beyond the full block.

lib/compress/zstd_compress.c
Cyan4973: I've added a paragraph which explains why a 1-byte malus to `savings` for each incompressible block is enough to guarantee that the 3-bytes per 128 KB expansion limit cannot be breached.

lib/compress/zstd_compress.c

Cyan4973: > But this currently will also split if the savings is negative, which doesn't seem like the desired behavior. Is this what you intended?

That's a very good point.
In my mind, `savings` was boxed to remain positive,
which it is as long as it remains within `compress_frameChunk()`.

But as soon as we leave that function and return (streaming scenario),
`savings` is regenerated from actual bytes consumes and written,
which can be negative if data was so far incompressible.

This is fixed in newer version, thanks to the `>3` test.

Great catch @terrelln !

Cyan4973:
I updated the blind-split strategy with the recommended updates from the code review.
I'm no longer concerned by a risk of over-expanding incompressible data at this point.

That being said, the blind-split strategy was proposed initially as a way to leverage "something" from the capability to ingest partial blocks for lower levels, for which the initial split heuristic was not fast enough. 

With now an execution plan which seems able to offer block splitting all the way down to level 1, it makes the blind-split strategy less important, now essentially reserved to negative compression levels.

Issue is, for negative compression levels, the blind-split strategy isn't effective. This is likely because negative compression levels drop the huffman compression of literals, which is where it had the most impact.

So, if it's not useful anywhere, it's debatable if this split strategy deserves to stay in the code.

An alternative could be to just drop it.

Cyan4973:
I updated the policy around `savings` over-split protection variable
to employ it for all splitting strategies,
not just the last "blind-split" strategy.

This is in anticipation of some adversarial input that would take advantage of the sampling pattern of faster split detectors,
thus making it split blocks which are effectively incompressible, thus creating additional block headers.

Note that, even without this protection, the `ZSTD_COMPRESSBOUND()` limit was never at risk:
this limit is based on a worst-case scenario of 1KB per block,
so even if the splitter is made to split incompressible data via adversarial inputs,
it will never get close to that limit (current minimum block size after splitting is 8 KB).

The problem is more about `ZSTD_DECOMPRESSION_MARGIN()`,
which uses its own worst size estimation, separate from `ZSTD_COMPRESSBOUND()`,
which is much more stringent (effectively, expansion there is limited to 3 bytes per full 128 KB block).
`ZSTD_DECOMPRESSION_MARGIN()` though is a very niche use case (overlapping input and output).

lib/compress/zstd_compress.c
```diff
@@ -4556,6 +4591,21 @@ static size_t ZSTD_compress_frameChunk(ZSTD_CCtx* cctx,
...
+             * and the splitter never creates blocks that small (current lower limit is 8 KB),
+             * there is already no risk to expand beyond ZSTD_COMPRESSBOUND() limit.
+             * But if the goal is to not expand by more than 3-bytes per 128 KB full block,
+             * then yes, it becomes possible to make the block splitter oversplit incompressible data.
+             * Using @savings, we enforce an even more conservative condition,
+             * requiring the presence of enough savings (at least 3 bytes) to authorize splitting,
+             * otherwise only full blocks are used.
+             * But being conservative is fine,
+             * since splitting barely compressible blocks is not fruitful anyway */
+            savings += (S64)blockSize - (S64)cSize;
```
terrelln, line 4608: Thanks for fixing this! IMO this method is also more clear, because it is clear that it is tracking exactly the savings from compression.

terrelln (approved):

PR #4153, opened 2024-09-30
meson: Do not export private headers in libzstd_dep to avoid name clash

artem:
This way libzstd_dep does not override, for instance, <xxhash.h>

### DIFF
--- a/build/meson/lib/meson.build
+++ b/build/meson/lib/meson.build
@@ -124,7 +124,7 @@ libzstd = library('zstd',
   version: zstd_libversion)
 
 libzstd_dep = declare_dependency(link_with: libzstd,
-  include_directories: libzstd_includes)
+  include_directories: join_paths(zstd_rootdir,'lib')) # Do not expose private headers
 
 # we link to both:
 # - the shared library (for public symbols)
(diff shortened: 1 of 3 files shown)


artem:
Signed CLA.

perhaps cc @eli-schwartz for review?

build/meson/lib/meson.build
```diff
@@ -124,7 +124,7 @@ libzstd = library('zstd',
   version: zstd_libversion)
 
 libzstd_dep = declare_dependency(link_with: libzstd,
-  include_directories: libzstd_includes)
+  include_directories: join_paths(zstd_rootdir,'lib')) # Do not expose private headers
```
eli-schwartz, line 127: This makes sense to me. It corresponds to the only 3 headers that get installed, and libzstd_dep is supposed to be equivalent to the installed zstd library.

It's still possible for people to do something filthy like `#include <common/xxhash.h>` but if they do then they broke it themselves, so they get to keep both pieces.

artem:
The second commit fixes tests and contrib that I broke =)

Cyan4973 (approved):

PR #4165, opened 2024-10-08
Improve compression speed on small blocks

Cyan4973:
This is a follow up from #4144 , from @TocarIP .

#4144 improves compression speed on small blocks, by forcing usage of a `cmov` via assembly code (`x64` only).
It works great on small blocks, but since the strategy is applied unconditionally, it's also active for long inputs. In which case, it introduces a speed regression (- 3-6%). 

This PR is a follow up, which attempts to bring the best of both worlds:
instantiate 2 variants, one using `cmov` and the other using a branch,
and select the appropriate variant at runtime.

This PR implements this strategy for the `fast` strategy.

It works fine, on `gcc`.

Initially `clang` didn't like it unfortunately.

Scenario: `silesia.tar`, cut into blocks of 32 KB, i7-9700k (no turbo), Ubuntu 24.04:

| compiler |  `dev` | #4144 | this PR |
| --- | --- | --- | --- |
| gcc-9 | 280 MB/s | 342 MB/s | 343 MB/s |
| clang-16 | 281 MB/s | 342 MB/s | **271 MB/s** |

After investigation, it appears that once the initial formulation `if (test1 && test2) { ... }` is changed into `if (inlined_test(...) { ... }` with `inlined_test() { return test1 && test2; }`, then the order of tests (`test1` _then_ `test2`) is no longer preserved. But, because of inlining, they still generate individual branches ! So `clang` was re-ordering the branches, making the range check branch first, thus negating the benefits of #4144.

One solution, suggested by @terrelln, is to use a memory barrier to force execution in the written order.
This suggestion has worked fine : 

| compiler |  `dev` | #4144 | this PR |
| --- | --- | --- | --- |
| gcc-9 | 280 MB/s | 342 MB/s | 342 MB/s |
| clang-16 | 281 MB/s | 342 MB/s | 342 MB/s |

On the other hand, with this PR, performance on longer inputs is no longer negatively impacted: 

Scenario: `silesia.tar`, entire file, i7-9700k (no turbo), Ubuntu 24.04:

| compiler |  `dev` | #4144 | this PR |
| --- | --- | --- | --- |
| gcc-9 | 386 MB/s | 368 MB/s | 384 MB/s |
| clang-16 | 379 MB/s | 368 MB/s | 386 MB/s |

_edit_ : 
As a good surprise, the current code seems to work fine as is on Apple's M1, without a need for some inline assembly.
It's unclear though if it does generalize well over all `aarch64` architectures and compilers.

Scenario: `silesia.tar`, cut into blocks of 32 KB, M1 Pro, Ubuntu 24.04:

| cpu |  `dev` | #4144 | this PR |
| --- | --- | --- | --- |
| M1 Pro | 410 MB/s | 408 MB/s | 535 MB/s |

_note 2_:
The negative impact of the `cmov` patch on compression speed for long inputs 
is much less pronounced at level 3 than it was at level 1, 
somewhere in the -1/-3% range.
The speed benefits on small data though are well present (> +10% at 32 KB blocks).

Scenario: `silesia.tar`, cut into blocks of 32 KB, level `3`:

| cpu |  `dev` | #4144 | this PR |
| --- | --- | --- | --- |
| i7-9700k | 283 MB/s | 311 MB/s | 317 MB/s | 
| M1 Pro | 336 MB/s | 408 MB/s | 408 MB/s |

Consequently, I believe it's not worth the time and binary size to generate some special `dfast` variant for "long inputs".

Potential follow up:
the current patch applies the new strategy to the `noDict` variants of `fast` and `dfast` _only_.
It seems interesting to also test and eventually adopt this strategy for dictionary compression too.

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -201,7 +201,7 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
              * However expression below complies into conditional move. Since
              * match is unlikely and we only *branch* on idxl0 > prefixLowestIndex
              * if there is a match, all branches become predictable. */
-            matchl0_safe = ZSTD_selectAddr(prefixLowestIndex, idxl0, &dummy[0], matchl0);
+            matchl0_safe = ZSTD_selectAddr(idxl0, prefixLowestIndex, matchl0, &dummy[0]);
 
             /* check prefix long match */
             if (MEM_read64(matchl0_safe) == MEM_read64(ip) && matchl0_safe == matchl0) {
@@ -215,7 +215,7 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
             matchl1 = base + idxl1;
 
             /* Same optimization as matchl0 above */
-            matchs0_safe = ZSTD_selectAddr(prefixLowestIndex, idxs0, &dummy[0], matchs0);
+            matchs0_safe = ZSTD_selectAddr(idxs0, prefixLowestIndex, matchs0, &dummy[0]);
 
             /* check prefix short match */
             if(MEM_read32(matchs0_safe) == MEM_read32(ip) && matchs0_safe == matchs0) {
@@ -662,7 +662,7 @@ size_t ZSTD_compressBlock_doubleFast_extDict_generic(
         size_t mLength;
         hashSmall[hSmall] = hashLong[hLong] = curr;   /* update hash table */
 
-        if (((ZSTD_index_overlap_check(prefixStartIndex, repIndex)) 
+        if (((ZSTD_index_overlap_check(prefixStartIndex, repIndex))
             & (offset_1 <= curr+1 - dictStartIndex)) /* note: we are searching at curr+1 */
           && (MEM_read32(repMatch) == MEM_read32(ip+1)) ) {
             const BYTE* repMatchEnd = repIndex < prefixStartIndex ? dictEnd : iend;
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -140,11 +140,17 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
     U32 idxl1; /* the long match index for ip1 */
 
     const BYTE* matchl0; /* the long match for ip */
+    const BYTE* matchl0_safe; /* matchl0 or safe address */
     const BYTE* matchs0; /* the short match for ip */
     const BYTE* matchl1; /* the long match for ip1 */
+    const BYTE* matchs0_safe; /* matchs0 or safe address */
 
     const BYTE* ip = istart; /* the current position */
     const BYTE* ip1; /* the next position */
+    /* Array of ~random data, should have low probability of matching data
+     * we load from here instead of from tables, if matchl0/matchl1 are
+     * invalid indices. Used to avoid unpredictable branches. */
+    const BYTE dummy[] = {0x12,0x34,0x56,0x78,0x9a,0xbc,0xde,0xf0,0xe2,0xb4};
 
     DEBUGLOG(5, "ZSTD_compressBlock_doubleFast_noDict_generic");
 
@@ -191,24 +197,29 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
 
             hl1 = ZSTD_hashPtr(ip1, hBitsL, 8);
 
-            if (idxl0 > prefixLowestIndex) {
-                /* check prefix long match */
-                if (MEM_read64(matchl0) == MEM_read64(ip)) {
-                    mLength = ZSTD_count(ip+8, matchl0+8, iend) + 8;
-                    offset = (U32)(ip-matchl0);
-                    while (((ip>anchor) & (matchl0>prefixLowest)) && (ip[-1] == matchl0[-1])) { ip--; matchl0--; mLength++; } /* catch up */
-                    goto _match_found;
-                }
+            /* idxl0 > prefixLowestIndex is a (somewhat) unpredictable branch.
+             * However expression below complies into conditional move. Since
+             * match is unlikely and we only *branch* on idxl0 > prefixLowestIndex
+             * if there is a match, all branches become predictable. */
+            matchl0_safe = ZSTD_selectAddr(prefixLowestIndex, idxl0, &dummy[0], matchl0);
+
+            /* check prefix long match */
+            if (MEM_read64(matchl0_safe) == MEM_read64(ip) && matchl0_safe == matchl0) {
+                mLength = ZSTD_count(ip+8, matchl0+8, iend) + 8;
+                offset = (U32)(ip-matchl0);
+                while (((ip>anchor) & (matchl0>prefixLowest)) && (ip[-1] == matchl0[-1])) { ip--; matchl0--; mLength++; } /* catch up */
+                goto _match_found;
             }
 
             idxl1 = hashLong[hl1];
             matchl1 = base + idxl1;
 
-            if (idxs0 > prefixLowestIndex) {
-                /* check prefix short match */
-                if (MEM_read32(matchs0) == MEM_read32(ip)) {
-                    goto _search_next_long;
-                }
+            /* Same optimization as matchl0 above */
+            matchs0_safe = ZSTD_selectAddr(prefixLowestIndex, idxs0, &dummy[0], matchs0);
+
+            /* check prefix short match */
+            if(MEM_read32(matchs0_safe) == MEM_read32(ip) && matchs0_safe == matchs0) {
+                  goto _search_next_long;
             }
 
             if (ip1 >= nextStep) {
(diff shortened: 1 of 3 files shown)


felixhandte (commented):
I've been banging my head against this optimization problem as well. I haven't gotten something globally better, but I have found one trick you may want to include:

I was looking at the disassembly of my version of this in zstd_dfast and I saw that it was using a separate register for `matchAddress` than for the `match` pointer in the calling frame. It does this, I guess, because the `matchAddress` gets potentially mutated by the `cmov`, but the calling frame's `match` isn't supposed to be. And the unmodified value of `match` is used subsequently, if a match is found.

But because `match` is always *only used subsequently* if a match is found, and because the `cmov` *only changes the value* of `matchAddress` if the match is invalid, it's actually fine for these to be the same register, and for us to clobber the caller's `match` if the match fails.

We can tell the compiler this is ok by rewriting these functions to take the `matchAddress` by reference rather than by value, and by writing back the result of the `cmov` into the pointed-to `match` variable.

E.g.,

```
MEM_STATIC int
ZSTD_findMatch_cmov4(const BYTE* currentPtr, const BYTE** matchAddressPtr, U32 matchIdx, U32 lowLimit)
{
    const BYTE* matchAddress = *matchAddressPtr;  // <----
    char valid;
    __asm__ (
        "cmp %2, %3\n"
        "setb %1\n"
        "cmovae %4, %0"
        : "+r"(matchAddress), "=r"(valid)
        : "r"(matchIdx), "r"(lowLimit), "r"(currentPtr)
    );
    *matchAddressPtr = matchAddress;  // <----
    return MEM_read32(currentPtr) == MEM_read32(matchAddress) && valid;
}
```

called like

```
FORCE_INLINE_TEMPLATE
ZSTD_ALLOW_POINTER_OVERFLOW_ATTR
size_t ZSTD_compressBlock_doubleFast_noDict_generic(
        ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
        void const* src, size_t srcSize, U32 const mls /* template */)
{
    ...
    if (ZSTD_findMatch_cmov4(ip, &matchs0, idxs0, prefixLowestIndex)) {
        goto _search_next_long;
    }
    ...
}
```

This avoids a needless spill & reload that otherwise gets triggered by using an additional register.

Cyan4973:
> I've been banging my head against this optimization problem as well. I haven't gotten something globally better, but I have found one trick you may want to include:
> 
> I was looking at the disassembly of my version of this in zstd_dfast and I saw that it was using a separate register for `matchAddress` than for the `match` pointer in the calling frame. It does this, I guess, because the `matchAddress` gets potentially mutated by the `cmov`, but the calling frame's `match` isn't supposed to be. And the unmodified value of `match` is used subsequently, if a match is found.
> 
> But because `match` is always _only used subsequently_ if a match is found, and because the `cmov` _only changes the value_ of `matchAddress` if the match is invalid, it's actually fine for these to be the same register, and for us to clobber the caller's `match` if the match fails.
> 
> We can tell the compiler this is ok by rewriting these functions to take the `matchAddress` by reference rather than by value, and by writing back the result of the `cmov` into the pointed-to `match` variable.
> 
> E.g.,
> 
> ```
> MEM_STATIC int
> ZSTD_findMatch_cmov4(const BYTE* currentPtr, const BYTE** matchAddressPtr, U32 matchIdx, U32 lowLimit)
> {
>     const BYTE* matchAddress = *matchAddressPtr;  // <----
>     char valid;
>     __asm__ (
>         "cmp %2, %3\n"
>         "setb %1\n"
>         "cmovae %4, %0"
>         : "+r"(matchAddress), "=r"(valid)
>         : "r"(matchIdx), "r"(lowLimit), "r"(currentPtr)
>     );
>     *matchAddressPtr = matchAddress;  // <----
>     return MEM_read32(currentPtr) == MEM_read32(matchAddress) && valid;
> }
> ```
> 
> called like
> 
> ```
> FORCE_INLINE_TEMPLATE
> ZSTD_ALLOW_POINTER_OVERFLOW_ATTR
> size_t ZSTD_compressBlock_doubleFast_noDict_generic(
>         ZSTD_matchState_t* ms, seqStore_t* seqStore, U32 rep[ZSTD_REP_NUM],
>         void const* src, size_t srcSize, U32 const mls /* template */)
> {
>     ...
>     if (ZSTD_findMatch_cmov4(ip, &matchs0, idxs0, prefixLowestIndex)) {
>         goto _search_next_long;
>     }
>     ...
> }
> ```
> 
> This avoids a needless spill & reload that otherwise gets triggered by using an additional register.


I tried the proposed variant, but couldn't observe any performance benefit, tested across many compiler versions.

Cyan4973:
I tried to find some "more clever" heuristic to switch between `cmov` and branch (`fast` strategy only).

Typically, it seems that finding a candidate in range is more difficult at the beginning of each file.
Consequently, it would be interesting to start with the `cmov` strategy, and then switch to a branch once there's enough  input to search into.

It doesn't work. Switching between these 2 types of tests is worse in benchmark than sticking to anyone of them.
The reason is not entirely clear at this stage, but the suspicion is that switching `fast` function instance mid-flight is less good for the instruction cache and for branch prediction.

Consequently, I'm keeping the current heuristic, which selects the branch test when `wlog >= 19`, and switch to `cmov` for any `wlog` value lower than that.

lib/compress/zstd_double_fast.c
```diff
@@ -191,24 +197,29 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
...
-                    mLength = ZSTD_count(ip+8, matchl0+8, iend) + 8;
-                    offset = (U32)(ip-matchl0);
-                    while (((ip>anchor) & (matchl0>prefixLowest)) && (ip[-1] == matchl0[-1])) { ip--; matchl0--; mLength++; } /* catch up */
-                    goto _match_found;
-                }
+            /* idxl0 > prefixLowestIndex is a (somewhat) unpredictable branch.
+             * However expression below complies into conditional move. Since
+             * match is unlikely and we only *branch* on idxl0 > prefixLowestIndex
+             * if there is a match, all branches become predictable. */
+            matchl0_safe = ZSTD_selectAddr(prefixLowestIndex, idxl0, &dummy[0], matchl0);
```
terrelln: All these arguments are reversed. It works, but is weird.

If the ordering is required for performance, we should rename these arguments.

```suggestion
            matchl0_safe = ZSTD_selectAddr(idxl0, prefixLowestIndex, matchl0, &dummy[0]);
```

lib/compress/zstd_double_fast.c
```diff
@@ -191,24 +197,29 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
...
             idxl1 = hashLong[hl1];
             matchl1 = base + idxl1;
 
-            if (idxs0 > prefixLowestIndex) {
-                /* check prefix short match */
-                if (MEM_read32(matchs0) == MEM_read32(ip)) {
-                    goto _search_next_long;
-                }
+            /* Same optimization as matchl0 above */
+            matchs0_safe = ZSTD_selectAddr(prefixLowestIndex, idxs0, &dummy[0], matchs0);
```
terrelln:
```suggestion
            matchs0_safe = ZSTD_selectAddr(idxs0, prefixLowestIndex, matchs0, &dummy[0]);
```

lib/compress/zstd_fast.c
```diff
@@ -97,6 +97,50 @@ void ZSTD_fillHashTable(ZSTD_matchState_t* ms,
...
 
+typedef int (*ZSTD_match4Found) (const BYTE* currentPtr, const BYTE* matchAddress, U32 matchIdx, U32 idxLowLimit);
+
+static int
+ZSTD_match4Found_cmov(const BYTE* currentPtr, const BYTE* matchAddress, U32 matchIdx, U32 idxLowLimit)
+{
+    /* Array of ~random data, should have low probability of matching data.
+     * Load from here if the index is invalid.
+     * Used to avoid unpredictable branches. */
+    static const BYTE dummy[] = {0x12,0x34,0x56,0x78 };
```
terrelln:
```suggestion
    static const BYTE dummy[] = {0x12,0x34,0x56,0x78};
```

lib/compress/zstd_fast.c
```diff
@@ -190,6 +232,7 @@ size_t ZSTD_compressBlock_fast_noDict_generic(
     size_t step;
     const BYTE* nextStep;
     const size_t kStepIncr = (1 << (kSearchStrength - 1));
+    const ZSTD_match4Found findMatch = useCmov ? ZSTD_match4Found_cmov : ZSTD_match4Found_branch;
```
terrelln: nit: Rename to `foundMatch()`, because it isn't actively searching for a match.
```suggestion
    const ZSTD_match4Found foundMatch = useCmov ? ZSTD_match4Found_cmov : ZSTD_match4Found_branch;
```

terrelln (approved):

Cyan4973:
All suggestions integrated

PR #4167, opened 2024-10-11
faster 32-bit CI tests

Cyan4973:
this is the longest CI test, reaching ~40mn on last PR, so we are waiting CI completion on this one currently.

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -50,7 +50,7 @@ jobs:
       run: |
         sudo apt-get -qqq update
         make libc6install
-        CFLAGS="-m32" make test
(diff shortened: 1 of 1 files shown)


Cyan4973:
It reduced the target test to 20mn.
Still long, but better.

Our new longest test is now `clang-msan-testzstd` (29mn)

Cyan4973:
`make-test-32bit` is now down to 12 minutes, no longer an issue.

Good, let's now have a look at the next longest one, `clang-msan-testzstd`

_edit_: this test is run entirely within `playTests.sh`, which is a single `sh` script, serially executed by nature.
Parallel execution would require breaking this test into multiple smaller parts, that could be started in parallel.
This will have to be a separate PR.

PR #4171, opened 2024-10-16
Improve compression ratio of levels 3 & 4

Cyan4973:
Following #4170 , there were a few ideas I wanted to test for the general compression performance of level 3, aka `dfast` strategy.

The one proposed in this PR is consistently positive.
It feels retrospectively obvious: select the supplemental "long" match candidate only if it's effectively longer.
On `silesia.tar` corpus, this modification saves ~75 KB at level 3. 

And thankfully, the measured speed cost is negligible, below noise level (between 0 and -1% depending on measurements).

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -252,19 +252,23 @@ _cleanup:
 
 _search_next_long:
 
-        /* check prefix long +1 match */
+        /* short match found: let's check for a longer one */
+        mLength = ZSTD_count(ip+4, matchs0+4, iend) + 4;
+
+        /* check long match at +1 position */
         if (idxl1 > prefixLowestIndex) {
             if (MEM_read64(matchl1) == MEM_read64(ip1)) {
-                ip = ip1;
-                mLength = ZSTD_count(ip+8, matchl1+8, iend) + 8;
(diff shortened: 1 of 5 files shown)


lib/compress/zstd_double_fast.c
```diff
@@ -253,19 +253,23 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
...
-                while (((ip>anchor) & (matchl1>prefixLowest)) && (ip[-1] == matchl1[-1])) { ip--; matchl1--; mLength++; } /* catch up */
-                goto _match_found;
-            }
+                size_t const llen = ZSTD_count(ip1+8, matchl1+8, iend) + 8;
+                if (llen > mLength) {
+                    ip = ip1;
+                    mLength = llen;
+                    offset = (U32)(ip-matchl1);
+                    while (((ip>anchor) & (matchl1>prefixLowest)) && (ip[-1] == matchl1[-1])) { ip--; matchl1--; mLength++; } /* catch up */
+                    goto _match_found;
```
terrelln: Could this code be merged with the code in lines 273-275?

We could set `matchs0 = matchl1`, and then just fallthrough to the code on line 273. Though we might want to rename some variables.

terrelln (approved):

PR #4180, opened 2024-10-28
Block splitter control parameter

Cyan4973:
Make it possible to explicit select a block splitter level, via a new `CCtx` parameter `ZSTD_c_blockSplitter_level`.

This capability is then exploited in a test, ensuring that incompressible data is not overly split, even in presence of an adversarial input (with full knowledge of the sampling pattern).

Note: possible follow-up:
This PR just adds a new parameter, to control the behavior of the new block splitter.
It doesn't modify the existing parameters.

But as a consequence, there are now 2 parameters for block splitters,
one (legacy) that is controlling the post block-splitter (after sequences are determined)
and a new one that is controlling the new pre block-splitter (before sequences are produced).
Already, it's debatable if it's useful for a user to be exposed to these concepts.
More importantly, the distinction between pre and post block splitter is not clear from the current parameters' names (`ZSTD_c_useBlockSplitter` vs `ZSTD_c_blockSplitter_level`).

So it opens the question of a refactoring of these parameters.
For example, maybe both parameters could be fused into a single one, the new `ZSTD_c_blockSplitter_level`, that would be charged to enable both when level is high enough.
Or maybe there is still value in keeping both these parameters separated, for example for an optimizer tool which could more naturally influence both code paths and maybe find a better combination for some specific use case. In which case, it's probably still useful to debate about meaningful parameter names.

### DIFF
--- a/lib/zstd.h
+++ b/lib/zstd.h
@@ -496,7 +496,7 @@ typedef enum {
      * ZSTD_c_prefetchCDictTables
      * ZSTD_c_enableSeqProducerFallback
      * ZSTD_c_maxBlockSize
-     * ZSTD_c_blockSplitter_level
+     * ZSTD_c_blockSplitterLevel
      * Because they are not stable, it's necessary to define ZSTD_STATIC_LINKING_ONLY to access them.
      * note : never ever use experimentalParam? names directly;
      *        also, the enums values themselves are unstable and can still change.
@@ -2265,7 +2265,7 @@ ZSTDLIB_STATIC_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const vo
  */
 #define ZSTD_c_searchForExternalRepcodes ZSTD_c_experimentalParam19
 
-/* ZSTD_c_blockSplitter_level
+/* ZSTD_c_blockSplitterLevel
  * note: this parameter only influences the first splitter stage,
  *       which is active before producing the sequences.
  *       ZSTD_c_useBlockSplitter influence the next splitter stage,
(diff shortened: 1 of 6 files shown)


lib/zstd.h
```diff
@@ -2264,6 +2265,24 @@ ZSTDLIB_STATIC_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const vo
...
+ * Allowed values are between 0 and 6.
+ * 0 means "auto", which will select a value depending on current ZSTD_c_strategy.
+ * 1 means no splitting.
+ * Then, values from 2 to 6 are sorted in increasing cpu load order.
+ *
+ * Note that currently the first block is never split,
+ * to ensure expansion guarantees in presence of incompressible data.
+ */
+#define ZSTD_BLOCKSPLITTER_LEVEL_MAX 6
+#define ZSTD_c_blockSplitter_level ZSTD_c_experimentalParam20
```
terrelln: nit: Lets call this `ZSTD_c_blockSplitterLevel` to match the naming convention of the rest of the parameters.

terrelln (approved):

Cyan4973:
I think I'll keep both parameters,
because both splitting methods (before and after sequences) may evolve independently,
and the way they combine or compete could change over time.

But it's necessary to change the names, so that it's less misleading.

In particular, `ZSTD_c_useBlockSplitter` implies a full on/off control over anything block-splitter related,
but that's not what this parameter is doing: it only controls the second splitter (which used to be the only splitter), which is triggered after sequences determination. So the parameter name should reflect that scope.

Current name in mind: `ZSTD_c_sequenceSplitter`, which emphasizes the fact that it's related to sequences. Importantly, it makes it clear that `ZSTD_c_blockSplitterLevel ` and `ZSTD_c_sequenceSplitter` are 2 separate decisions.
However, the name could also be understood as splitting sequences, 
as opposed to splitting blocks according to sequences already found.

Cyan4973:
Finally settled on `ZSTD_c_splitAfterSequences`.

Issue #4183, opened 2024-10-31
Question and suggestion - behavior with small buffers and errors


Eugeny1:
I am a starter with zstd (using simple ZSTD_compress and ZSTD_decompress) and have three things to say after some research and having specific task.

Most probably I can find the answers looking into the code, but I'd like to state the issues I have at least as a first step.

1. I need to compress the data and see if compressed size is less than uncompressed (to decide what data to go with). I'd allocate new buffer of the same size of source data and try to compress there. I would expect if compressed data is bigger, I'd get an out of buffer error. However, the following things make me wondering:
  - I read that "bigger buffer makes compression faster". Because logically extrapolating this statement the smaller buffer makes compression slower, and what happens when buffer size is so small it is about to get out of memory (compression stalls)? While my question may sound funny and stupid, I believe it must be addressed.
  - I can't find clear definition of errors returned. Apart from all the errors, I'd like to check for exactly this out of memory error, but struggling to figure out how to do it.

2. In general, looking at the documentation here [herel](https://facebook.github.io/zstd/zstd_manual.html) I was unable to quickly find how to handle the errors except using

    unsigned    ZSTD_isError(size_t code);          /*!< tells if a `size_t` function result is an error code */
    const char* ZSTD_getErrorName(size_t code);     /*!< provides readable string from an error code */

with first function saying that there's an error, and second returning the string. But what is exact error? [This](https://github.com/facebook/zstd/blob/dev/lib/zstd_errors.h) lists errors, am I expected to compare against those identifiers using ZSTD_getErrorCode()? The documentation I linked to earlier does not mention this function! Also what is the error enum for out of destination buffer issues - ZSTD_error_noForwardProgress_destFull? What does this name mean?

3. Small, but I found it kind of bothering. When looking at the error enum identifiers, I see some of them label destination as "dest" and some as "dst". That would be very convenient if uniform approach would be selected, so that searching for the keyword shows all the instances related to destination.

Thanks.

Cyan4973:
> I'd allocate new buffer of the same size of source data and try to compress there. I would expect if compressed data is bigger, I'd get an out of buffer error. 

Yes, this design works, and is in active usage.
A prime example would be compressed filesystems, like BtrFS or ZFS.

> Because logically extrapolating this statement the smaller buffer makes compression slower, 

Yes, but don't overthink it. The difference is small, and only perceptible at fastest compression levels.

> what happens when buffer size is so small it is about to get out of memory (compression stalls)?

[...]

PR #4184, opened 2024-10-31
elevated ZSTD_getErrorCode() to stable status

Cyan4973:
answering #4183.

Also:
- updated manual, so that `ZSTD_getErrorCode()` is clearly mentioned in the paragraph dealing with handling of errors
- updated version number, since this is technically a change of officially stable API


Cyan4973:
mmmh,
including `zstd_errors.h` from `zstd.h` 
seems in conflict with our single file library build system...

_edit_ : nope, hopefully, it's just the test of the single file library which is failing.

But it does illustrate the fact that, since `zstd.h` now includes `zstd_errors.h` after this patch,
both files, `zstd.h` and `zstd_errors.h`, are now required,
which could be surprising if the build has been created with the assumption that only `zstd.h` is enough.

PR #4191, opened 2024-11-08
cmake: Fix #4038 MacOS target older OS and SDK versions

### LINKED ISSUE #4038: Building on MacOS 13 and targeting MacOS 11 and SDK 11.3 (or any other MacOS version) does not work
**Describe the bug**
As of v1.5.6, the minos and sdk settings are no longer respected. This works on version 1.5.5.

(It is possible this is an interaction with MacPorts where I am building the library.

**To Reproduce**
1. Install macports
2. `echo 'macosx_deployment_target 11.0' | tee -a ${PREFIX}/etc/macports/macports.conf`
  `echo 'macosx_sdk_version 11.3' | tee -a ${PREFIX}/etc/macports/macports.conf`
4. `sudo port install -sv zstd`
5. look in macports `lib` directory and run `otool -l libzstd.1.dylib`
6. see output includes `minos` and `sdk` of the current OS, rather than `minos 11.0` and `sdk 11.3` 

**Expected behavior**
Should correctly set the OS and SDK versions in the dylib.

**Screenshots and charts**

 - OS: Mac
 - Version 14.4.1
 - Compiler clang
 - Flags not sure
 - Other relevant hardware specs not relevant
 - Build system CMAKE

**Additional context**
Downstream bug reports:
https://trac.macports.org/ticket/69944#comment:1
https://gitlab.gnome.org/GNOME/gimp/-/issues/11483

lukaso:
This fix ensures that when `MACOSX_DEPLOYMENT_TARGET` and `SDKROOT` are set, they are respected when building the libraries and executables.

How to test (on a MacOS machine):

You will have to download an older SDK version. In this case I'm using 11.3.

```
export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk
export MACOSX_DEPLOYMENT_TARGET=11.0
make install
otool -l lib/libzstd.1.5.7.dylib
```

The output will include:

```
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 11.0
      sdk 11.3
```

Without this fix, the output will looks something like:

```
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 13.0
      sdk 13.0
```

Depending on the version of MacOS installed and the SDK installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,9 +200,9 @@ travis-install:
 .PHONY: clangbuild-darwin-fat
 clangbuild-darwin-fat: clean
 	clang -v
-	CXX=clang++ CC=clang CFLAGS="-Werror -Wconversion -Wno-sign-conversion -Wdocumentation -arch arm64" $(MAKE) zstd-release
+	CXX=clang++ CC=clang CFLAGS+="-Werror -Wconversion -Wno-sign-conversion -Wdocumentation -arch arm64" $(MAKE) zstd-release
 	mv programs/zstd programs/zstd_arm64
-	CXX=clang++ CC=clang CFLAGS="-Werror -Wconversion -Wno-sign-conversion -Wdocumentation -arch x86_64" $(MAKE) zstd-release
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):
Thanks @lukaso , this looks good to me

PR #4199, opened 2024-11-20
Fix printing of filesize if >4GB

jimis:


Cyan4973:
Thanks for the proposed patch.
It looks good to me,

note that CLA compliance is a requirement before the merge.

jimis:
Alright, but it might take a bit since I have to go through my employment contract first. I thought such a small fix was non-copyrightable so the CLA would not be needed.

Cyan4973:
The CLA rules are quite stringent. However, if the process becomes too complex for you, one possible solution could be to duplicate your patch and merge it under another CLA-compliant identity.

jimis:
Thank you for your patience @Cyan4973, I just signed the CLA.

Cyan4973 (approved):

PR #4201, opened 2024-11-24
prevent possible segfault when creating seek table

### LINKED ISSUE #4200: ZSTD_seekTable_create_fromSeekable can cause segmentation faults
**Describe the bug**
`ZSTD_seekTable_create_fromSeekable`  causes a segmentation fault when called with a `ZSTD_seekable` that is not initialized. 

**To Reproduce**
1. `ZSTD_seekable* const stream = ZSTD_seekable_create();`
2. `ZSTD_seekTable* const zst = ZSTD_seekTable_create_fromSeekable(stream);`

**Expected behavior**
`ZSTD_seekTable_create_fromSeekable` returns a NULL pointer.

rorosen:
Add a check in `ZSTD_seekTable_create_fromSeekable` whether the seek table of a `ZSTD_seekable` is initialized before creating a new seek table from it. Formerly, the function created a segfault when the seek table of `ZSTD_seekable`was not initialized.

Fixes https://github.com/facebook/zstd/issues/4200

### DIFF
--- a/contrib/seekable_format/zstdseek_decompress.c
+++ b/contrib/seekable_format/zstdseek_decompress.c
@@ -252,6 +252,8 @@ size_t ZSTD_seekable_free(ZSTD_seekable* zs)
 
 ZSTD_seekTable* ZSTD_seekTable_create_fromSeekable(const ZSTD_seekable* zs)
 {
+    assert(zs != NULL);
+    if (zs->seekTable.entries == NULL) return NULL;
     ZSTD_seekTable* const st = (ZSTD_seekTable*)malloc(sizeof(ZSTD_seekTable));
     if (st==NULL) return NULL;
 


contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -252,6 +252,7 @@ size_t ZSTD_seekable_free(ZSTD_seekable* zs)
 
 ZSTD_seekTable* ZSTD_seekTable_create_fromSeekable(const ZSTD_seekable* zs)
 {
+    if (zs->seekTable.entries == NULL) return NULL;
```
Cyan4973, line 256: Since we are dereferencing `zs` here, I would prefer an `assert(zs != NULL);` just before this statement.

Arguably, it should have been added a few lines below too, but wasn't.

Cyan4973 (approved):
Indeed, given that `ZSTD_seekTable` is essentially a limited extract from `ZSTD_seekable`, it involves copying some content from one side to the other.
And it's a problem if the source doesn't exist (is not initialized).
This simple check is a good fix.

contrib/seekable_format/zstdseek_decompress.c
rorosen: Thanks for your review. I added the assert

PR #4204, opened 2024-11-26
Add ARM's assembly code for huffman decoding

Nicoshev:
We would like to have a dedicated ARM's asm function for decoding huffman


JunHe77:
Would be good to add PAC and BTI support,  #3961 could be referred for adding BTI description in section

terrelln:
The latest `clang-trunk` and `gcc-14` both match or beat the assembly performance for both 4x1 and 4x2 using the "fast C" decoding loops.

Given that newer compilers perform just as well or better, and the maintenance burden of adding more assembly, we will not be merging this upstream. Adding non-x86-64 assembly makes it very hard to fuzz test the code. And down the line I would love to also delete the x86-64 assembly. Maintaining assembly in Zstd has ended up being a huge PITA, due to the huge number of platforms & compilers we support.

Thanks for the PR @Nicoshev!

PR #4212, opened 2024-12-10
tests: fix non-regular file test on GNU/Hurd

diegonc:
Since commit b21b03ca6 [1] the behaviour of writes to /dev/zero has been fixed and now the non-regular file removal test no longer needs to be done on /dev/random (which no longer works as random is not writable now).

[1] https://git.savannah.gnu.org/cgit/hurd/hurd.git/commit/?id=b21b03ca624b89caeedfe58430cea4b40317d39f

### DIFF
--- a/tests/playTests.sh
+++ b/tests/playTests.sh
@@ -109,10 +109,7 @@ isTerminal=${isTerminal:-$detectedTerminal}
 
 isWindows=false
 INTOVOID="/dev/null"
-case "$UNAME" in
-  GNU) DEVDEVICE="/dev/random" ;;
-  *) DEVDEVICE="/dev/zero" ;;
-esac
+DEVDEVICE="/dev/zero"
 case "$OS" in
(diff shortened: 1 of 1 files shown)


Cyan4973:
which indirectly makes me wonder: 
do we have a GNU/Hurd test on Github Actions ?
That sounds unlikely.
And in which case, should we add one ? How complex is it to setup ? and then to maintain ?

Cyan4973:
> which indirectly makes me wonder: do we have a GNU/Hurd test on Github Actions ? That sounds unlikely. And in which case, should we add one ? How complex is it to setup ? and then to maintain ?

I had a quick look at it,
and, unless someone already did it and it's "just" a matter of copy paste,
it sounds like a nightmarish proposition to setup and maintain.
So, I guess we'll pass.

Cyan4973 (approved):

PR #4217, opened 2024-12-17
ZSTD_compressSequencesAndLiterals

Cyan4973:
This PR introduces a variant of `ZSTD_compressSequences()`,
named `ZSTD_compressSequencesAndLiterals()`,
which accepts a buffer of literals instead of the source buffer.

This is meant to match specific limitations present in some transcoder operations,
in which it's easy to present Sequences and Literals coming from an external generator,
but it would be more complex or costly to restore the corresponding source.
As an example, one such use case happens within the Linux kernel,
where the source comes as scattered pages which do not form a continuous buffer.

This variant is a bit faster on its own than regular `ZSTD_compressSequences()`, by almost ~+20%,
as measured below using `fullbench` on a `i7-9700k`: 

| source | parser | `compressSequences` | `compressSequencesAndLiterals` | difference |
| calgary.tar | level 19 | 805 MB/s | 960 MB/s | +19% |
| enwik6 | level 4 | 761 MBs | 917 MB/s | +20 % |

The difference is pretty consistent, driven primarily by a reduction in overhead during the sequence transcoding operation.

In addition, the goal is to cumulate the speed benefits of this variant with potential other benefits on the caller side, by no longer requiring presentation of the original data.

That being said, `ZSTD_compressSequencesAndLiterals()` also features some limitations:
- Only supports Explicit Block Delimiter mode. While this could be extended later on, it's unclear if it would be useful.
- Does not support Frame Checksum, since it doesn't see the original data. Requesting this checksum will result in an error.
- Does not write the content size within the frame header
- Cannot generate uncompressed blocks, since it doesn't have the original data. While this could be partially mitigated in some favorable cases, at the cost of complexity, there are many more complex scenarios for which it cannot be _completely_ eliminated. Thus, if the compression process tries to generate an uncompressed block, it results in a compression error.

For this reason, it's best to keep this method for compression of "small" data (typically <= 128 KB), where data is typically stored uncompressed when it doesn't compress well enough. This setup corresponds to typical kernel use cases like `zram`, `zswap` or `BtrFS`.


Cyan4973:
Updated implementation, with slightly different prototype (does no longer request `srcSize`) and improved compression speed.

lib/zstd.h
```diff
@@ -1633,27 +1635,57 @@ ZSTDLIB_STATIC_API size_t ZSTD_mergeBlockDelimiters(ZSTD_Sequence* sequences, si
...
+ * This can be useful if the process generating the sequences also happens to generate the buffer of literals,
+ * thus skipping an extraction + caching stage.
+ * It's a speed optimization, useful when the right conditions are met,
+ * but it also features the following limitations:
+ * - Only supports explicit delimiter mode
+ * - Not compatible with frame checksum, which must disabled
+ * - Does not write the content size in frame header
+ * - If any block is incompressible, will fail and return an error
+ * - @litSize must be == sum of all @.litLength fields in @inSeqs. Any discrepancy will generate an error.
+ * - the buffer @literals must be larger than @litSize by at least 8 bytes.
```
terrelln: Can we take another parameter `litCapacity` here? This seems like an easy constraint to break, and it would be good to force users to pass in a value here and check it directly.

tests/fuzz/sequence_compression_api.c
```diff
@@ -232,7 +232,7 @@ static size_t roundTripTest(void* result, size_t resultCapacity,
                             const void* src, size_t srcSize,
```
terrelln, line 294: We should test `ZSTD_compressSequencesAndLiterals()` here as well. We could just test both `ZSTD_compressSequences()` and `ZSTD_compressSequencesAndLiterals()`

terrelln (commented):
Reviewed everything except the actual implementation in `zstd_compress.c` and `zstd_compress_internal.h`.

lib/compress/zstd_compress.c
```diff
@@ -6918,11 +6980,13 @@ ZSTD_compressSequences_internal(ZSTD_CCtx* cctx,
         }
 
         RETURN_ERROR_IF(dstCapacity < ZSTD_blockHeaderSize, dstSize_tooSmall, "not enough dstCapacity to write a new compressed block");
-        compressedSeqsSize = ZSTD_entropyCompressSeqStore(&cctx->seqStore,
-                                &cctx->blockState.prevCBlock->entropy, &cctx->blockState.nextCBlock->entropy,
-                                &cctx->appliedParams,
+        compressedSeqsSize = ZSTD_entropyCompressSeqStore_wExtLitBuffer(
```
terrelln: It looks like this is the same as `ZSTD_entropyCompressSeqStore()`, can we call that instead? Or am I missing something?

lib/compress/zstd_compress.c
```diff
@@ -7022,6 +7092,246 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* cctx,
     return cSize;
 }
 
+/*
+ * Note: Sequence validation functionality has been disabled (removed).
+ * This is helpful to find back simplicity, leading to performance.
+ * It may be re-inserted later.
+ */
```
terrelln: Do we error if sequence validation was requested?

lib/compress/zstd_compress.c
```diff
@@ -7022,6 +7092,246 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* cctx,
...
+    assert(cctx->appliedParams.searchForExternalRepcodes != ZSTD_ps_auto);
+
+    DEBUGLOG(4, "ZSTD_compressSequencesAndLiterals_internal: nbSeqs=%zu, litSize=%zu", nbSequences, litSize);
+    RETURN_ERROR_IF(nbSequences == 0, externalSequences_invalid, "Requires at least 1 end-of-block");
+
+    /* Special case: empty frame */
+    if ((nbSequences == 1) && (inSeqs[0].litLength == 0)) {
+        U32 const cBlockHeader24 = 1 /* last block */ + (((U32)bt_raw)<<1);
+        RETURN_ERROR_IF(dstCapacity<4, dstSize_tooSmall, "No room for empty frame block header");
+        MEM_writeLE32(op, cBlockHeader24);
```
terrelln: We have `MEM_writeLE24` right? Can we use it here to support `dstCapacity == 3`?

lib/compress/zstd_compress.c
```diff
@@ -7022,6 +7092,246 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* cctx,
...
+        nbSequences -= nbBlockSeqs;
+
+        /* Note: when blockSize is very small, other variant send it uncompressed.
+         * Here, we still send the sequences, because we don't have the original source to send it uncompressed.
+         * One could imagine it possible to reproduce the source from the sequences,
+         * but that's complex and costly memory intensive, which goes against the objectives of this variant. */
+
+        RETURN_ERROR_IF(dstCapacity < ZSTD_blockHeaderSize, dstSize_tooSmall, "not enough dstCapacity to write a new compressed block");
+
+        compressedSeqsSize = ZSTD_entropyCompressSeqStore_internal(
```
terrelln, line 7245: Do we need to call `ZSTD_entropyCompressSeqStore_internal()` or can we call `ZSTD_entropyCompressSeqStore_wExtLitBuffer()`?

lib/compress/zstd_compress.c
Cyan4973: `ZSTD_entropyCompressSeqStore()` is invoking `ZSTD_entropyCompressSeqStore_wExtLitBuffer()`,
so they both converge to the same implementation.

The difference is that `ZSTD_entropyCompressSeqStore()` forces the `literals` buffer to be the one reserved into `cctx`, while `_wExtLitBuffer()` let the caller select where the buffer of `literals` is.

lib/compress/zstd_compress.c
Cyan4973: nope, at this point, it's just ignored.
It could be a good idea to add such check and error out clearly.

lib/zstd.h
Cyan4973: parameter `litCapacity` added

lib/compress/zstd_compress.c
Cyan4973: check added

lib/compress/zstd_compress.c

lib/compress/zstd_compress.c
Cyan4973: The initial code was invoking `ZSTD_entropyCompressSeqStore_wExtLitBuffer()`.
But because `ZSTD_entropyCompressSeqStore_wExtLitBuffer()` is essentially `ZSTD_entropyCompressSeqStore()`, just with a selectable `litBuffer`,
it adds a few constraints, which favor the creation of uncompressed blocks.
In the normal case, it's fine (an uncompressed block is preferable to a compressed one with only 1 byte of saving),
but in the case of `ZSTD_compressSequencesAndLiterals()`, it is not good, because this variant is unable to generate an uncompressed block (since it doesn't see the original data), and the only possible outcome is to generate an error.
So we prefer skipping these additional constraints, by invoking the next stage directly, `ZSTD_entropyCompressSeqStore_internal()`.

lib/compress/zstd_compress.c
terrelln: @Cyan4973 I understand that, but in this case the literals are in the seqStore, right?

There are so many parameters that it is hard to read, but it looks like this could just invoke `ZSTD_entropyCompressSeqStore()` instead, because the literals are `cctx->seqStore.litStart, (size_t)(cctx->seqStore.lit - cctx->seqStore.litStart)`.

terrelln (commented):
Looks good once we add the fuzzer

lib/compress/zstd_compress.c
Cyan4973: Ah, you meant this specific instance.
And yes, you are right, it could be replaced by `ZSTD_entropyCompressSeqStore()`.

This modification was made early during development, when I was trying to have a common pipeline for the 2 variants. This is a left over from this period. Since I gave up this approach, we now have 2 separate pipelines, so it's possible to restore the call to `ZSTD_entropyCompressSeqStore()`, especially if it helps readability.

tests/fuzz/sequence_compression_api.c
Cyan4973: Fuzzer test for `ZSTD_compressSequencesAndLiterals()` added,
it piggy - backs on `ZSTD_compressSequences()`,
and gets triggered automatically whenever a compatible set of parameters is detected.

Cyan4973:
All points addressed

Issue #4187, opened 2024-11-04
zstd does #include within extern "C" scopes.  This breaks trying to build a clang module.


alecazam:
**Describe the bug**
#include should never be inside of scopes or namepsaces.  But zstd seems to extern "C" and then include various headers within them.  This then causes clang module building to fail.   I will trying moving all of these outside of these scopes, but in general this is bad practice for code used in so many applications..

import of C++ module 'std_stdint_h' appears within extern "C" language linkage specification

```
#if  !defined (__VMS) && (defined (__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */) )
#  if defined(_AIX)
#    include <inttypes.h>
#  else
#    include <stdint.h> /* intptr_t */ <-
#  endif

Another example from the concatenated zstddeclib.cpp

#if defined (__cplusplus)
extern "C" {
#endif

/*-****************************************
*  Dependencies
******************************************/
#include <stddef.h>  /* size_t, ptrdiff_t */

```

**To Reproduce**
1. turn on -fmodules and -fcxx-modules 
2. setup a module.modulemap with zstd in it
3. Watch compile fail trying to build zstd

**Expected behavior**
All headers should just be included first and outside of any scoping.  Then any concatenation should just work.  extern "C" should only wrap function calls and structs for naming reasons.

**Screenshots and charts**
None

**Desktop (please complete the following information):**
 - OS: macOS 
 - Version 15.0
 - Compiler clang
 - Flags any
 - Other relevant hardware specs 8/4 cores
 - Build system Xcode

**Additional context**
None

alecazam:
Alternatively, just have one file with all  #include and then prepend that at the top of the concatenated files.

Cyan4973:
> example from the concatenated zstddeclib.cpp

Is the issue specific to the amalgamated single-file library ?

alecazam:
Yes, from what I see one concatenated header will start an extern "C" scope, and then do #include of other zstd files.  So end up with #include nested inside of these scopes.   I've commented out the includes here.

This is from the top of zstd.h.   I've just commented out the includes.  But zstd.h, zstddeclib.cpp, and zstd.cpp all have similar patterns.

```
#if defined (__cplusplus)
extern "C" {
#endif

#ifndef ZSTD_H_235446
#define ZSTD_H_235446

/* ======   Dependency   ======*/
//#include <limits.h>   /* INT_MAX */
[...]

PR #4218, opened 2024-12-18
Move #includes out of `extern "C"` blocks

Victor-C-Zhang:
Do some include shuffling for `**.h` files within lib, programs, tests, and zlibWrapper. `lib/legacy` and `lib/deprecated` are untouched.
`#include`s within `extern "C"` blocks in .cpp files are untouched.

todo: shuffling for `xxhash.h`

I have no way of exhaustively testing this... CI to the rescue?

### DIFF
--- a/contrib/pzstd/test/PzstdTest.cpp
+++ b/contrib/pzstd/test/PzstdTest.cpp
@@ -7,9 +7,7 @@
  * in the COPYING file in the root directory of this source tree).
  */
 #include "Pzstd.h"
-extern "C" {
 #include "datagen.h"
-}
 #include "test/RoundTrip.h"
 #include "utils/ScopeGuard.h"
 
--- a/contrib/pzstd/test/RoundTripTest.cpp
+++ b/contrib/pzstd/test/RoundTripTest.cpp
@@ -6,9 +6,7 @@
  * LICENSE file in the root directory of this source tree) and the GPLv2 (found
  * in the COPYING file in the root directory of this source tree).
  */
-extern "C" {
 #include "datagen.h"
-}
 #include "Options.h"
 #include "test/RoundTrip.h"
 #include "utils/ScopeGuard.h"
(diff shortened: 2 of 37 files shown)


felixhandte:
You could build the library before and after and see that the symbol list given by `nm` doesn't change? (I.e., no symbols become mangled.)

Victor-C-Zhang:
@Cyan4973 what should we do about copied deps like `fse` and `xxhash`? Does this warrant a PR in their respective repos too?

Cyan4973:
How did you check that this PR fixed #4187 ?

Victor-C-Zhang:
> How did you check that this PR fixed #4187 ?

I haven't. I will try to repro #4187 with the limited information provided and do a before-and-after. From reading forum posts of people with similar issues, this may be an issue specific to an old version of apple clang from an old XCode, but I'll need to double-check to make sure...

Victor-C-Zhang:
Update: I was unable to repro the build failure.
Here's what I did:

1. Generate zstd.c from `create_single_file_library.sh`
2. Copy `zstd.c` and `lib/zstd.h`, `lib/zdict.h`, `lib/zstd_error.h`, `lib/module.modulemap` to a fresh directory (outside zstd repo) `~/module_test`
3. In the directory `~/module_test` use the following `main.cpp`:
```
#include "zzz.h"
#include <iostream>
#include <string>

int main() {
  std::cout << "Hello, world!" << std::endl;
  std::string input = "abcdabcdabcdabcdabcdabcdabcdabcd";
  std::string output('\0', input.size() * 100);
  auto res = ZSTD_compress(output.data(), output.size(),
                           input.data(), input.size(), 3);
  std::cout << res << std::endl;

  return 0;
}
```
4. For safety, rename the module `libzstd` -> `zzz`, and rename `zstd.h` -> `zzz.h`, `zstd.c` -> `zzz.c`. Also do this in the module.modulemap.
5. Build with module
```
clang -shared -o libzzz.dylib zzz.c
clang++ -std=c++20 main.cpp -fmodules -fcxx-modules -L. -lzzz
```
6. It runs:
```
% ./a.out
Hello, world!
18446744073709551546
```

Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin24.1.0
Thread model: posix

Cyan4973 (approved):

PR #4220, opened 2024-12-19
[fix] Added UNAME_TARGET_SYSTEM to make shared lib compile flag selection configurable

### LINKED ISSUE #4219: Shared library flags depend on `uname`, undesired for cross compilation
**Describe the bug**
The shared library build options depend on the system's `uname`. This especially causes builds to fail when cross compiling from macOS because of the check here: https://github.com/facebook/zstd/blob/0218c8de0fa77bbd87e75f2ea70ba00b93460e15/lib/Makefile#L76

**To Reproduce**
- Plattform: macOS
- Install arm-none-eabi-gcc: `brew install --cask gcc-arm-embedded`
- Use the following to build
```bash
#!/bin/bash

set -e

export ZSTD_LIB_COMPRESSION=0
export ZSTD_LIB_DECOMPRESSION=1
export ZSTD_LIB_DICTBUILDER=0
export ZSTD_LIB_DEPRECATED=0
export ZSTD_LEGACY_SUPPORT=0

export CC=$(which arm-none-eabi-gcc)
export CFLAGS="-march=armv6k+fp -mtune=mpcore -mfloat-abi=hard -mtp=soft -O3 -fomit-frame-pointer"

make lib-nomt
```

**Expected behavior**
Assignment of the shared lib flags should not depend on uname and should be configurable.

**Desktop:**
 - OS: macOS 15.0.1 (24A348)
 - Compiler: arm-none-eabi-gcc 14.2.1 20241119 (Arm GNU Toolchain 14.2.Rel1 (Build arm-14.52))
 - Flags: see above
 - Build system: Makefile

RicBent:
Fixes #4219 by adding `UNAME_TARGET_SYSTEM` which, when not set, defaults to `UNAME`.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -110,9 +110,6 @@ ifndef BUILD_DIR
 # determine BUILD_DIR from compilation flags
 
 libzstd.a:
-	@echo uname_target $(UNAME_TARGET)
-	@echo target_system $(TARGET_SYSTEM)
-	@echo uname $(uname)
 	$(SET_CACHE_DIRECTORY)
 
 else
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -73,13 +73,15 @@ endif
 
 # macOS linker doesn't support -soname, and use different extension
 # see : https://developer.apple.com/library/mac/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html
-ifeq ($(UNAME), Darwin)
+UNAME_TARGET_SYSTEM ?= $(UNAME)
+
+ifeq ($(UNAME_TARGET_SYSTEM), Darwin)
   SHARED_EXT = dylib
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
This is a good stop-gap, it allows a user to manually take control of the situation, without impacting the current default behavior.

I presume that a more robust solution would likely employ the output of the compiler to determine what's the target, but that's more work and more risk, which can be attempted later.

As a side note, `UNAME` is a variable, so it can already be taken over by the user,
but I guess it's also used in several other places, and you only want to impact this one place.

RicBent:
Exactly, the `UNAME` variable is also used to select different tools depending on what OS you are running on, like here: https://github.com/facebook/zstd/blob/0218c8de0fa77bbd87e75f2ea70ba00b93460e15/lib/libzstd.mk#L212

terrelln:
Thanks for the PR @RicBent!

PR #4222, opened 2024-12-20
[cmake] Fix -z noexecstack portability

terrelln:
Summary:
Issue reported by @ryandesign and @MarcusCalhoun-Lopez. Thanks to @ryandesign for investigating this issue in-depth & making fixing it easy.

CMake doesn't support spaces in flags. This caused older versions of gcc to ignore the unknown flag "-z noexecstack" on MacOS since it was interpreted as a single flag, not two separate flags. Then, during compilation it was treated as "-z" "noexecstack", which was correctly forwarded to the linker. But the MacOS linker does not support `-z noexecstack` so compilation failed.

The fix is to use `-Wl,-z,noexecstack`. This is never misinterpreted by a compiler. However, not all compilers support this syntax to forward flags to the linker. To fix this issue, we check if all the relevant `noexecstack` flags have been successfully set, and if they haven't we disable assembly.

See also PR #4056 and PR #4061. I decided to go a different route because this is simpler. It might not successfully set these flags on some compilers, but in that case it also disables assembly, so they aren't required.

Test Plan:
```
mkdir build-cmake
cmake ../build/cmake/CMakeLists.txt
make -j
```

See that the linker flag is successfully detected & that assembly is enabled.

Run the same commands on MacOS which doesn't support `-Wl,-z,noexecstack` and see that everything compiles and that `LD_FLAG_WL_Z_NOEXECSTACK` and `ZSTD_HAS_NOEXECSTACK` are both false.


Cyan4973:
Is there a test we could create in CI that would trigger the problem (without this PR) ?

terrelln:
I'm not sure. My understanding is that it requires quite an old version of gcc on MacOS to test. I'm not sure we'll be able to get a gcc old enough in our Mac test environment.

Cyan4973 (approved):

Andarwinux:
This breaks Windows builds, ASM is silently disabled

PR #4229, opened 2025-01-02
Support for libc variants without fseeko/ftello

Victor-C-Zhang:
Some older Android libc implementations don't support `fseeko` or `ftello`. This commit adds a new compile-time macro `LIBC_NO_FSEEKO` as well as a usage in CMake for Android APIs 23 and below. Note this change is NOT intended to add support for zstd within old AOSP distributions. Google has long since stopped updating old branches. If you are an app developer working on API 23 or lower, this will allow you to package zstd in your app yourself.

Tested by [adding the project in Android Studio](https://developer.android.com/studio/projects/gradle-external-native-builds) and building the enclosing app on SDK versions 23 and 34 (current).
Used defaultConfig:
```
externalNativeBuild {
            cmake {
                cFlags += listOf("-DZSTD_BUILD_PROGRAMS", "-DZSTD_BUILD_STATIC")
                cppFlags += ""
                targets += listOf("zstd", "libzstd_static")
            }
        }
```


contrib/seekable_format/zstdseek_decompress.c
```diff
@@ -77,7 +77,10 @@
 /* ************************************************************
 * Avoid fseek()'s 2GiB barrier with MSVC, macOS, *BSD, MinGW
 ***************************************************************/
-#if defined(_MSC_VER) && _MSC_VER >= 1400
+#if defined(LIBC_NO_FSEEKO)
+/* Some older libc implementations don't include these functions (e.g. Bionic < 24) */
+#define LONG_SEEK fseek
```
Cyan4973: minor nit: indentation: not aligned with below `#define`

Cyan4973 (approved):
Globally fine,
just try to follow the same `#define` indentation as other `#elif` sections.

Issue #4099, opened 2024-07-19
[reproducibility] Zstd varies compressed output based on the presence of SSE / NEON in some cases


terrelln:
https://github.com/facebook/zstd/blob/0ff651dd876823b99fa5c5f53292be28381aee9b/lib/compress/zstd_compress.c#L238-L249

We haven't historically strongly guarantee reproducibility across compilations on different systems. However, we've been moving in this direction. We should consider removing this source of difference, either by default or by opting into reproducible mode via a flag. I'm leaning towards by default, because I think this is one of the few places left where we have differences in compressed output based on the platform.

Cyan4973:
Agreed

felixhandte:
IIUC, the diff in possible output is specific to the range `14 < windowLog <= 17`. And you're proposing to close that gap?

lulcat:
Actually, I am a bit miffed at the moment.. I am getting _different output_ using the examples bundled in examples/

I tried vanilla on a laptop (same OS though), and a desktop and I am seeing this acros the boaard. Also in some bindings. I am rather shocked/confused whatever... How to nail this down? Is it my computers? is it arch like distros? Is it the examples?
using the threads_pool example say... lvl 3, 
// a find example for files in curdir only. **/* **(-D.) depending on shells and so on if desired.

find . -maxdepth 1 -type f -exec zstd -3  {} \; 
ls *.zst | wc -l
74

// I save to .pst just to not overwrite here
find . -maxdepth 1 -type f -exec  ./streaming_compression_threads_pool {} \;
 for f in *.zst ; do diff $f ${f%.zst}.pst;done

Binary files a.zst and a.pst differ
Binary files b.zst and b.pst differ
Binary files c.zst and c.pst differ
Binary files struct_return.zst and struct_return.pst differ
Binary files class_array.zst and class_array.pst differ
Binary files hello.zst and hello.pst differ
Binary files poolz.zst and poolz.pst differ
Binary files usrbins.zst and usrbins.pst differ

... what gives?

I just am wondering as of course ideally I would want the shared binaries people have on linux systems to create same determinstic bytes in a file as long as same ocmpression level is set (I am disregarding any dicts here).

I am ok with this determinism to be platform specific, unlike the original topic, (so sorry about the slight digression); however, on the same platform, and same compr level I would expect libs, examples and upstrema binaries to all produce same output. 

Something is off here and I need to figure out if it's this 'distro' (currently arch linux) or a hardware issue.

[...]

PR #4230, opened 2025-01-03
Do not vary row matchfinder selection based on availability of SSE2/Neon

Victor-C-Zhang:
Move towards a stronger guarantee of reproducibility by removing this small difference for machines without SSE2/Neon. The SIMD behavior is now the default for all platforms.

#4099


Cyan4973 (approved):
cc @terrelln 
If I remember correctly, this is indeed the wanted behavior

PR #4231, opened 2025-01-05
Update MANDIR for FreeBSD

sunpoet:
share/man became a valid path for manpage since Jan 2020. And we converted the whole ports tree to share/man around last March.


Cyan4973 (approved):

sunpoet:
Thanks!

PR #4232, opened 2025-01-08
Improve speed of ZSTD_compressSequencesAndLiterals() using AVX2

Cyan4973:
This PR improves the speed of `ZSTD_compressSequencesAndLiterals()`, especially when compiled with `AVX2` support.

For illustration, here are some benchmark numbers, on a i7-9700k @3.6 Ghz (turbo off),
using `enwik5` (100KB) and the sequences produced by level 5,
which is an unfavorable scenario because it produces many smaller sequences.

| dataset | compiler | `ZSTD_compressSequences()` | `ZSTD_compressSequencesAndLiterals()` on `dev` | `ZSTD_compressSequencesAndLiterals()` on this PR, scalar mode | `ZSTD_compressSequencesAndLiterals()` with `AVX2` enabled |
| --- | --- | --- | --- | --- | --- |
| `enwik5`, level 5 | `gcc` `v13.3` | 528 MB/s | 598 MB/s | 630 MB/s | 670 MB/s |
| `enwik5`, level 5 | `clang` `v18.1` | 512 MB/s | 584 MB/s | 601 MB/s | 665 MB/s |

The improvements to the scalar code path are small but generic.

The `AVX2` code path improves even more, but obviously requires the corresponding vector support, which is not guaranteed or may come with strings attached, for example within kernel mode.

The vector code path could be even faster, and is mostly hampered by the need to check for exceptional cases.
There might be ways to improve performance even more by streamlining the path in favorable scenarios.
But note that all this work does is to reduce the overhead of translating the Sequence public format into the internal one,
so there's a limit to how much overhead can be removed, and we are already getting pretty close to this limit after these recent optimizations.

Also:
improved benchmark framework, by adding an (optional) validation function.


lib/compress/zstd_compress.c
```diff
@@ -7103,15 +7103,214 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* cctx,
     return cSize;
 }
 
+
+#if defined(__AVX2__)
```
terrelln, line 7107: We should add a constant `ZSTD_ARCH_X86_AVX2` to `compiler.h` here, and make sure we respect the `ZSTD_NO_INTRINSICS` macro.

https://github.com/facebook/zstd/blob/a610550e2c05cd08842e173bbeb830f596fdfaeb/lib/common/compiler.h#L225-L239

lib/compress/zstd_compress.c
```diff
@@ -7129,21 +7328,40 @@ size_t ZSTD_convertBlockSequences(ZSTD_CCtx* cctx,
...
+        for (seqNb = 0; seqNb < nbSequences - 1 ; seqNb++) {
+            U32 const litLength = inSeqs[seqNb].litLength;
+            U32 const matchLength = inSeqs[seqNb].matchLength;
+            U32 offBase;
 
-        DEBUGLOG(6, "Storing sequence: (of: %u, ml: %u, ll: %u)", offBase, matchLength, litLength);
-        ZSTD_storeSeqOnly(&cctx->seqStore, litLength, offBase, matchLength);
+            if (!repcodeResolution) {
+                offBase = OFFSET_TO_OFFBASE(inSeqs[seqNb].offset);
+            } else {
```
terrelln: This is impossible

lib/compress/zstd_compress.c
```diff
@@ -7172,20 +7390,78 @@ size_t ZSTD_convertBlockSequences(ZSTD_CCtx* cctx,
     return 0;
 }
 
-typedef struct {
-    size_t nbSequences;
-    size_t blockSize;
-    size_t litSize;
-} BlockSummary;
+#if defined(__AVX2__)
```
terrelln: Use `ZSTD_ARCH_X86_AVX2`

lib/compress/zstd_compress.c
```diff
@@ -7172,20 +7390,78 @@ size_t ZSTD_convertBlockSequences(ZSTD_CCtx* cctx,
...
+
+/* C90-compatible alignment macro (GCC/Clang). Adjust for other compilers if needed. */
+#if defined(__GNUC__)
+#  define ALIGNED32 __attribute__((aligned(32)))
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* C11 */
+#  define ALIGNED32 alignas(32)
+#else
+   /* this compiler will require its own alignment instruction */
+#  define ALIGNED32
+#endif
```
terrelln: I know there are other places we align. Should we unify this into a common macro in `compilers.h`?

lib/compress/zstd_compress.c
```diff
@@ -7129,21 +7328,40 @@ size_t ZSTD_convertBlockSequences(ZSTD_CCtx* cctx,
...
+            assert(cctx->seqStore.longLengthType == ZSTD_llt_none);
+            if (longl <= nbSequences-1) {
+                DEBUGLOG(5, "long match length detected at pos %zu", longl-1);
+                cctx->seqStore.longLengthType = ZSTD_llt_matchLength;
+                cctx->seqStore.longLengthPos = (U32)(longl-1);
+            } else {
+                DEBUGLOG(5, "long literals length detected at pos %zu", longl-nbSequences);
+                assert(longl <= 2* (nbSequences-1));
+                cctx->seqStore.longLengthType = ZSTD_llt_literalLength;
+                cctx->seqStore.longLengthPos = (U32)(longl-nbSequences);
```
terrelln: nit: This is worth a comment, because there seems to be +1 and -1 cancelling out here, which is confusing.

lib/compress/zstd_compress.c
```diff
@@ -7103,15 +7103,214 @@ size_t ZSTD_compressSequences(ZSTD_CCtx* cctx,
...
+    ZSTD_STATIC_ASSERT(offsetof(ZSTD_Sequence, offset) == 0);
+    ZSTD_STATIC_ASSERT(offsetof(ZSTD_Sequence, litLength) == 4);
+    ZSTD_STATIC_ASSERT(offsetof(ZSTD_Sequence, matchLength) == 8);
+    ZSTD_STATIC_ASSERT(sizeof(SeqDef) == 8);
+    ZSTD_STATIC_ASSERT(offsetof(SeqDef, offBase) == 0);
+    ZSTD_STATIC_ASSERT(offsetof(SeqDef, litLength) == 4);
+    ZSTD_STATIC_ASSERT(offsetof(SeqDef, mlBase) == 6);
+
+    /* Process 2 sequences per loop iteration */
+    for (; i + 1 < nbSequences; i += 2) {
```
terrelln, line 7191: Not sure how much time you want to spend on this, but it would likely be a bit faster to unroll it once more to process 4 sequences per loop.

If you process 4 sequences per loop you can do half as many cross-lane shuffles. You would keep the first 2 sequences as-is, and then put the second 2 sequences in the top half of each lane. Then use `_mm256_blend_epi32()` to blend them together, and you only need the single `_mm256_permute4x64_epi64()` for 4 sequences.

terrelln (approved):

lib/compress/zstd_compress.c
Cyan4973: restored full equation, for clarity

lib/compress/zstd_compress.c
Cyan4973: This is better kept as a potential optimization that for a future iteration.
At this stage, we don't know yet if `AVX2` will be useful in kernel mode.

Cyan4973:
Lots of CI issues suddenly, 
though they seem uncorrelated to this PR.

Noticed so far:
- `liblzma` and `lzma.h` no longer available
- `clang-14` no longer available
- `-lgcc` missing at linking stage (??)

If I were to guess, the CI VM probably just got silently updated,
but I sure am surprised by the nb of tests it broke in the process.

Cyan4973:
All comments addressed

PR #4234, opened 2025-01-11
[MSVC] Wrong define checked when building for 64bit with Visual Studio

MessyHack:
Should check defined(_M_X64) not defined(_M_X86) when building with MSVC.

_M_X86 is only defined under MSVC 32Bit
_M_X64 is only defined under MSVC 64Bit

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -309,7 +309,7 @@ FORCE_INLINE_TEMPLATE size_t BIT_getMiddleBits(BitContainerType bitContainer, U3
      * such cpus old (pre-Haswell, 2013) and their performance is not of that
      * importance.
      */
-#if defined(__x86_64__) || defined(_M_X86)
+#if defined(__x86_64__) || defined(_M_X64)
     return (bitContainer >> (start & regMask)) & ((((U64)1) << nbBits) - 1);
 #else
     return (bitContainer >> (start & regMask)) & BIT_mask[nbBits];


MessyHack:
signed CLA

Cyan4973 (approved):
Reasoning looks good,

I'm curious if it impacts performance in any way on Visual x64.

MessyHack:
> Reasoning looks good,
> 
> I'm curious if it impacts performance in any way on Visual x64.

There's a slight perf gain when building with MSVC and targeting AVX2. (Using current VS 2022 preview)
I should double check the SSE behavior

When targeting AVX2, MSVC will use bzhi (bmi2 instruction).  This only happens when targeting AVX2.

MessyHack:
> > Reasoning looks good,
> > I'm curious if it impacts performance in any way on Visual x64.
> 
> There's a slight perf gain when building with MSVC and targeting AVX2. (Using current VS 2022 preview) I should double check the SSE behavior
> 
> When targeting AVX2, MSVC will use bzhi (bmi2 instruction). This only happens when targeting AVX2.

SSE seems ok; but not really that different.

(before change; SSE)
 3#Lorem ipsum       :  10000000 ->   2983345 (x3.352),  337.1 MB/s  1181.9 MB/s
vs
(after change; SSE
 3#Lorem ipsum       :  10000000 ->   2983345 (x3.352),  339.4 MB/s  1192.0 MB/s

Cyan4973:
> When targeting AVX2, MSVC will use bzhi (bmi2 instruction). This only happens when targeting AVX2.

I thinks that's the main goal.
We observed that `bmi2` gives a small but solid performance gain.
Enabling that should be beneficial.

PR #4238, opened 2025-01-13
fix out-of-bounds array index access

szsam:


Cyan4973:
That's a good starting point !

Looking a bit more into it, if execution reaches this code path with `idx == inSeqsSize`,
it is probably better to return an error, because it means that input is invalid.
In contrast, current proposed fix just skips the `if () {...}` code block, and still returns a success, as if everything was fine.

Cyan4973 (approved):

PR #4243, opened 2025-01-15
fix ci tests

Cyan4973:
Some Github CI tests have started failing recently, as a consequence of updating the underlying ubuntu image.
Try to fix the created issues one by one in this PR, to end up on an all-green signal.

- `clang-pgo` : this test was tied to `clang-14`, which is no longer supported in latest ubuntu image. Use the "generic" `clang` instead, requiring installation of the companion profiler tool.
- `meson-linux`, `cmake-build-and-check`, `make-external-compressor` : require explicit installation of `liblzma-dev` package (no longer provided by default in the image)

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install gcc-powerpc-linux-gnu gcc-arm-linux-gnueabi gcc-aarch64-linux-gnu libc6-dev-ppc64-powerpc-cross libcurl4-gnutls-dev lib64gcc-11-dev-powerpc-cross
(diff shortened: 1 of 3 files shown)


Cyan4973:
The `zlib-wrapper` error can also be reproduced locally, so it's not related to Github Actions VMs.
It's curious it started failing at the same time.
Maybe the library version changed, and modified a behavior ?

No clear explanation for the time being: the file `example.c` fails [here](https://github.com/facebook/zstd/blob/dev/zlibWrapper/examples/example.c#L445) .
It's a direct correspondence from a `zlib` test file.

_edit_ : indeed, in more recent versions of `zlib`, the test changed, and now expects a different error code (`Z_STREAM_END` instead of `Z_DATA_ERROR`). to be fixed.

Cyan4973:
All tests fixed, CI signal green.

I'm going to merge this, so that we can rebase ongoing PRs on top of it, and observe a clean green signal before merge.

PR #4248, opened 2025-01-18
Use _bzhi_u32 for 32-bit builds when building with STATIC_BMI2

pps83:
`_bzhi_u64` is available only for 64-bit builds, while `BIT_getLowerBits` expects `nbBits` to be less than `BIT_MASK_SIZE` (`BIT_MASK_SIZE` is 32)

### DIFF
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -162,11 +162,11 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
 FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)
 {
 #if defined(STATIC_BMI2) && STATIC_BMI2 == 1 && !defined(ZSTD_NO_INTRINSICS)
-#if defined(__x86_64__) || defined(_M_X64)
-    return _bzhi_u64(bitContainer, nbBits);
-#else
+#  if defined(__x86_64__) || defined(_M_X64)
+    return sizeof(size_t) == sizeof(U64) ? _bzhi_u64(bitContainer, nbBits) : _bzhi_u32((U32)bitContainer, nbBits);
+#  else
     return _bzhi_u32(bitContainer, nbBits);
-#endif
+#  endif
 #else
     assert(nbBits < BIT_MASK_SIZE);
     return bitContainer & BIT_mask[nbBits];
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -162,7 +162,11 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
 FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)
 {
 #if defined(STATIC_BMI2) && STATIC_BMI2 == 1 && !defined(ZSTD_NO_INTRINSICS)
-    return  _bzhi_u64(bitContainer, nbBits);
+#if defined(__x86_64__) || defined(_M_X64)
+    return _bzhi_u64(bitContainer, nbBits);
+#else
+    return _bzhi_u32(bitContainer, nbBits);
+#endif
 #else
     assert(nbBits < BIT_MASK_SIZE);
     return bitContainer & BIT_mask[nbBits];
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -165,7 +165,7 @@ FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(BitContainerType bitContainer, U32
 #  if (defined(__x86_64__) || defined(_M_X64)) && !defined(__ILP32__)
     return _bzhi_u64(bitContainer, nbBits);
 #  else
-    DEBUG_STATIC_ASSERT(sizeof(size_t) == sizeof(U32));
+    DEBUG_STATIC_ASSERT(sizeof(bitContainer) == sizeof(U32));
     return _bzhi_u32(bitContainer, nbBits);
 #  endif
 #else
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -159,7 +159,7 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
     return 0;
 }
 
-FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)
+FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(BitContainerType bitContainer, U32 const nbBits)
 {
 #if defined(STATIC_BMI2) && STATIC_BMI2 == 1 && !defined(ZSTD_NO_INTRINSICS)
 #  if (defined(__x86_64__) || defined(_M_X64)) && !defined(__ILP32__)
--- a/lib/common/bitstream.h
+++ b/lib/common/bitstream.h
@@ -162,9 +162,10 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
 FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)
 {
 #if defined(STATIC_BMI2) && STATIC_BMI2 == 1 && !defined(ZSTD_NO_INTRINSICS)
-#  if defined(__x86_64__) || defined(_M_X64)
-    return sizeof(size_t) == sizeof(U64) ? _bzhi_u64(bitContainer, nbBits) : _bzhi_u32((U32)bitContainer, nbBits);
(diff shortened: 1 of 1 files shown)


lib/common/bitstream.h
```diff
@@ -162,7 +162,11 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
 FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)
 {
 #if defined(STATIC_BMI2) && STATIC_BMI2 == 1 && !defined(ZSTD_NO_INTRINSICS)
-    return  _bzhi_u64(bitContainer, nbBits);
+#if defined(__x86_64__) || defined(_M_X64)
```
Cyan4973: minor: indentation:
we would use : 
```
# if ...
# else
# endif
```
(notice the space between `#` and the keyword)
to distinguish the indentation level,
thus making it visually clear that it's inside the upper `#if ...`.

Cyan4973: What matters is that the size of `bitContainer`, aka the size of `size_t`, is 64-bit or 32-bit.

Alas, `sizeof(size_t)` doesn't work at preprocessor time. Hence the work around to employ build macros, such as `__x86_64__`.

The size of `size_t` may seem implied by `defined(__x86_64__) || defined(_M_X64)`,
but there is actually one exception: the `x32` mode.

The `x32` mode uses the x64 64-bit instruction, but only employs 32-bit memory addresses. As a consequence, `size_t` is also 32-bit. Using `_bzhi_u64()` on a 32-bit variable would be incorrect.
This exception can only happen on Linux.
Here, the `x32` mode can be detected because it defines both `__x86_64__` and `__ILP32__`.

Therefore, `defined(__x86_64__) && !defined(__ILP32__)` is a good way to ensure that the compiler is operating in 64-bit mode, with 64-bit `size_t` type.

Cyan4973 (changes requested):

lib/common/bitstream.h
pps83: What about the non-intrinsic case? Seems like it's always capped to [32bit sizes](https://github.com/facebook/zstd/pull/4248/files#diff-7baf72b27df2e0e6193e491e0815741ae9b879e3036400fe57dbb0765ce5d6a7R182) anyways.

lib/common/bitstream.h
```diff
@@ -162,7 +162,12 @@ MEM_STATIC size_t BIT_initCStream(BIT_CStream_t* bitC,
 FORCE_INLINE_TEMPLATE size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)
```
Cyan4973: minor:

looking at the rest of the file,
it seems `bitContainer`'s type should rather be `BitContainerType`.

It doesn't change anything, since it's `typedef` to `size_t`, 
but it's just proper,
and could allow, in some future, to switch `BitContainerType` to employ more efficient `64-bit` registers even in `x32` mode.

lib/common/bitstream.h
Cyan4973: That's a good point.

The way I see it, it would make more sense for `BIT_mask[]` to be an array of `size_t`. And to make it a truly generic solution, it should extend to `63` bits.

That being said, `BIT_getLowerBits()` is in fact not generic.
It's limited to 31 bits.
That's because the only place where it's invoked is `BIT_addBits()` which is, itself, limited to adding 31 bits max (note: and effectively lower than that, aka 25 bits max in 32-bit mode, but that's more complex to explain these details).

Would it make sense to fix that situation and make `BIT_getLowerBits()` more generic, effectively extending `BIT_addBits()` capability ?

Well, maybe, but that's not sure.
One would have to look at the whole program to understand the situation and decide on a good trade off, and it's a tall order at this point.

One advantage I can see of having `BIT_mask[]` defined as an array of 32-bit `unsigned` is that it's a smaller array (32 * 4 = 128 bytes). If it was an array of 64-bit `size_t` extending up to 63 bits, it would end up being even larger, at 512 bytes. Increased table size impacts binary size, and pollute cache.

One downside of having a mask as a 32-bit value is that it probably has to be extended to 64-bit for the `&` AND operation. This is probably a very cheap operation, but is it nonetheless perceptible ? I don't know.
_edit_ : after looking it up, zeroing the upper part of a 64-bit register is essentially a free operation.

Both impacts (increased array => cache pollution, register extension to 64-bit) seem to counter balance, and may be the increased array size is the worse one, or even if they are just equal, it's still better in general to go for the smaller array. And maybe this is why today the code is structured this way.

Cyan4973:
Here also, if there were some benchmark numbers to illustrate the potential performance change introduced by this patch, this would be great for future reference.
My understanding is that the change is only expected to impact 32-bit mode.
I'm not expecting _huge_ changes to be fair, but even just ensuring that performance is not negatively impacted would be a good start.

Cyan4973 (approved):
This looks good to me.

Try to benchmark the change for documentation purpose.

Also, the discussion opened a Pandora Box of additional questions, sometimes quite subtle, so I'll take the follow ups from there.

pps83:
> Try to benchmark the change for documentation purpose.

I do 64 build builds, this change won't affect those. But I also do 32-bit builds for correctness check and got an error when building for 32bit with `-DSTATIC_BMI2`

Cyan4973:
And i presume this error gets fixed after this PR ?

If that's the case, it might be worth extending the CI test suite to cover this scenario, which has been missing so far.

Cyan4973:
Something I'm currently wondering :
in order to make the compilation fail, one must set explicitly `-DSTATIC_BMI2=1`.

I was wondering: why is it not automatically set, on detection of `AVX2` typically ?

Well, looking at the code where this variable is automatically set, it indeed requires the compilation mode to be `__x86_64__` . 
But why ? Shouldn't just the detection of `AVX2` or even `BMI2` be enough to set `STATIC_BMI2` to `1` ?

And now I'm wondering: is it only set in 64-bit mode because otherwise compilation fails, precisely because this fix was missing ?

In which case, it would be a good reason to change the way `STATIC_BMI2` is set after this patch is merged.

pps83:
> Something I'm currently wondering : in order to make the compilation fail, one must set explicitly `-DSTATIC_BMI2=1`.
> 
> I was wondering: why is it not automatically set, on detection of `AVX2` typically ?

I set -DSTATIC_BMI2=1 for compilation.
There is a weird issue with zstd. It has `compiler.h` and `portability_macros.h` (compiler includes portability_macros). Then, entire STATIC_BMI2 block inside compiler.h is weird:

```
/*Like DYNAMIC_BMI2 but for compile time determination of BMI2 support*/
#ifndef STATIC_BMI2
#  if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_I86))
#    ifdef __AVX2__  //MSVC does not have a BMI2 specific flag, but every CPU that supports AVX2 also supports BMI2
#       define STATIC_BMI2 1
#    endif
#  elif defined(__BMI2__) && defined(__x86_64__) && defined(__GNUC__)
#    define STATIC_BMI2 1
#  endif
#endif

#ifndef STATIC_BMI2
    #define STATIC_BMI2 0
#endif
```

it starts with `#ifndef STATIC_BMI2` meaning that it could be declared from outside (that's what I did). Then, STATIC_BMI2 gets declared for msvc for 32 and 64 bit builds. **but** only for 64bit builds with gcc. That's weird. Then, if you search entire code base, `__BMI2__` is still used in the code in a couple of places, while `STATIC_BMI2` should have been used instead imo.

`DYNAMIC_BMI2` clearly should have been declared rigth there, but it's inside portability_macros.h which gets processed before compiler.h. Also, DYNAMIC_BMI2 should be 1 only when `STATIC_BMI2` is 0, but STATIC_BMI2 gets defined before it. Quite convoluted mess imo.

Cyan4973:
I could understand why `DYNAMIC_BMI2` ends up into `portability_macros.h`:
because it's a runtime capability which is target specific,
whereas `STATIC_BMI2` is a fixed decision selected for the entire compilation session.

But I do not understand why it comes first in the dependency chain.

It looks to me that `portability_macros.h` should depend on `compiler.h`, and not the other way round.

This would solve the issue that `DYNAMIC_BMI2` should only be enabled when `STATIC_BMI2` isn't.

This code likely requires some refactor.

Cyan4973:
I now see this comment in `portability_macros.h`

```
This header is shared between C and ASM code, so it MUST only contain macro definitions. It MUST not contain any C code.
```

That would explain why it doesn't include `compiler.h`,
because `compiler.h` does include some prototype definitions, not just macros.

But to be frank, it looks to me that it's `compiler.h` which has been over-extended and does way too much.

Anyway, a refactor exercise isn't going to be trivial, and requires some attention to be done properly.

pps83:
IMO, the should be merged into a single file to avoid the mess (all macro from compiler.h should be moved to portability_macros.h)

PR #4250, opened 2025-01-19
added a CI test for x86 32-bit + avx2 combination

Cyan4973:
This combination is expected to be quite rare, but nonetheless possible.

This test is initially expected to fail, before integration of #4248 fix

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -192,7 +192,7 @@ jobs:
       run: |
         sudo apt-get -qqq update
         make libc6install
-        CFLAGS="-Werror -m32 -mavx2" make -j all32
+        CPPFLAGS="-DSTATIC_BMI2=1" CFLAGS="-Werror -m32 -mavx2" make -j all32
 
 
   gcc-8-make:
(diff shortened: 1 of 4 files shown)


Cyan4973:
As anticipated, the new test [just failed](https://github.com/facebook/zstd/actions/runs/12850316882/job/35829864803?pr=4250),
but not at the expected place,
so there is more investigation to do.

_edit_ : and indeed, there is another AVX2 compilation issue to fix too.

Cyan4973:
OK, so now that the "generic" `AVX2` compilation test has been fixed,
the more specific `-m32 -mavx2` test works too.
Which is not the expectation: rather it was expected that it would fail without #4248.

This suggests that the error condition reported [here](https://github.com/facebook/zstd/pull/4248#issuecomment-2600524924) is not correctly reproduced in [this test](https://github.com/facebook/zstd/actions/runs/12850452058/job/35830134496?pr=4250).

_edit_ : got it, in order to make it fail, the build variable `STATIC_BMI2=1` must be manually set at compilation time.

Cyan4973:
And as expected, it all works after the #4248 fix

PR #4257, opened 2025-01-21
Use _M_X64 only without mixing with _M_AMD64

pps83:


Cyan4973:
To the best of my knowledge, `_M_AMD64` and `_M_X64` are equivalent.
Yet, you seem to prefer `_M_X64`, and suggest consolidating around this preprocessor macro name only.
Any specific reason for this choice ?

pps83:
Yes, these are identical in ms compiler. It's better to use one. In my experience `_M_X64` is more widely used by projects, plus `_M_X64` looks somewhat closer to `__x86_64__`

pps83:
### _M_X64:
 image

<br/><br/><br/>
### _M_AMD64:
image

Cyan4973 (approved):

PR #4259, opened 2025-01-21
Add support for Apple framework builds

Treata11:
Greetings,  

In order to enable iOS-derived compilations, I firstly had to bump the minimum CMake version to 3.14 and I also introduced a new flag, `ZSTD_FRAMEWORK`, which is set to FALSE by default.
I introduced the required properties for framework generation as described in [CMake docs](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK)

And building is successful using the following commands:
```bash 
cmake -S. -B build- -DZSTD_FRAMEWORK=ON -DCMAKE_SYSTEM_NAME=iOS

sudo cmake --build build- --target install --config Release
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@
 *.so
 *.so.*
 *.dylib
+*.framework
+*.xcframework
 
 # Executables
 /zstd
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -7,7 +7,7 @@
 # in the COPYING file in the root directory of this source tree).
 # ################################################################
 
-cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
 # As of 2018-12-26 ZSTD has been validated to build with cmake version 3.13.2 new policies.
 # Set and use the newest cmake policies that are validated to work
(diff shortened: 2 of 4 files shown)


Cyan4973:
Why is the `3.14` version required ?

Note that increasing the minimum `cmake` version is a non-trivial operation,
because it implies dropping support for some platforms which do not offer the newly enforced version.

In this case, dropping support for `3.10` means dropping Ubuntu 18 `bionic`, which  is notable.
Other examples are: dropping Fedora 28 & 29, OpenSuse 15 or Alpine LInux 3.8 & 3.9 for example.

Ubuntu 18 `bionic` is still within its extended support range, so it's not great,
but it can be compensated by bringing something of value in exchange,
in this case "iOS-derived compilations".

Now, assuming we nonetheless want to increase the minimum version,
the rest of the `cmake `script may have to be adjusted due to the change of version.

In this case, the impact looks limited:
mostly, we have a `ZSTD_CMAKE_POLICY_VERSION` set to `3.13`,
which feels somewhat inconsistent with a minimum version set to `3.14`.
It probably needs to be updated.

Note that, in contrast with the minimum version,
the policy version can be updated to some much newer version,
as long as a test proves that the script works fine under the tested version.

Treata11:
@Cyan4973, thank you for the clarification. 
Framework builds were added in CMake v3.14 and I thought that it's probably sufficiently old (Oct 2019)...
I can revert it back to v3.10 and the framework properties that I added in this PR would work just fine with [iOS-Cmake](https://github.com/leetal/ios-cmake) third party toolchain and framework builds would be possible with slight changes in the configuration command:

```bash
cmake -B build -G Xcode -DCMAKE_TOOLCHAIN_FILE=<path to ios.toolchain.cmake> -DPLATFORM=OS64 -DZSTD_FRAMEWORK=ON
```

Cyan4973:
Generally, our support policy is closer to 10 years,
thought there can be exceptions at times.

I like the second solution you propose.
It probably deserves to be documented, 
typically within https://github.com/facebook/zstd/blob/dev/build/cmake/README.md .

Treata11:
All done!

Cyan4973:
Would it be possible to have a test in CI that checks this functionality ?
Otherwise, the risk is that this functionality might be broken by some future update, and we wouldn't notice it.

Treata11:
> Would it be possible to have a test in CI that checks this functionality ? Otherwise, the risk is that this functionality might be broken by some future update, and we wouldn't notice it.

It's possible to create a build job with `ZSTD_FRAMEWORK` flag, which cannot run the tests, but it can confirm it builds and validate the install.
I'll save this for another PR, if agreed.

Cyan4973:
>  it can confirm it builds and validate the install.

Yes, that will be good enough

Cyan4973 (approved):

PR #4260, opened 2025-01-21
[linux] Expose ZSTD_compressSequences*() in the kernel

gcabiddu:
Make the functions `ZSTD_compressSequences()`, `ZSTD_compressSequencesAndLiterals()` and `ZSTD_CCtx_setParameter()` available in kernel space.
These will be used in future by the Intel QAT driver.

Note that:
* `ZSTD_compressSequencesAndLiterals()` is sufficient for QAT. We could just expose that function and not `ZSTD_compressSequences()`.
* In order for the kernel to compile in contrib/linux-kernel/mem.h, `#include <asm/unaligned.h>` should be replaced with `#include <linux/unaligned.h>`. I haven't made that change yet as it was causing some regressions in the tests.


Cyan4973:
Looks good to me,
cc @terrelln since it's kernel mode specifically

Cyan4973 (approved):

contrib/linux-kernel/linux_zstd.h
```diff
@@ -136,10 +136,20 @@ typedef ZSTD_parameters zstd_parameters;
...
 
+/**
+ * zstd_cctx_set_param() - sets a compression parameter
+ * @cctx:         The context. Must have been initialized with zstd_init_cctx().
+ * @param:        The parameter to set.
+ * @value:        The value to set the parameter to.
+ *
+ * Return:        Zero or an error, which can be checked using zstd_is_error().
+ */
+size_t zstd_cctx_set_param(zstd_cctx *cctx, ZSTD_cParameter param, int value);
```
terrelln: minor: We could also add a typedef for `zstd_cparameter` to give it a kernel-style name.

terrelln:
> In order for the kernel to compile in contrib/linux-kernel/mem.h, #include <asm/unaligned.h> should be replaced with #include <linux/unaligned.h>. I haven't made that change yet as it was causing some regressions in the tests.

You should be able to fix the tests by moving this file:

https://github.com/facebook/zstd/blob/dev/contrib/linux-kernel/test/include/asm/unaligned.h

Into this directory:

https://github.com/facebook/zstd/tree/dev/contrib/linux-kernel/test/include/linux

This include directory fakes the Linux headers that Zstd depends on.

terrelln (approved):

contrib/linux-kernel/linux_zstd.h
gcabiddu: I added a typedef for `ZSTD_cParameter` but I didn't add the documentation as the elements of the enum follow the ZSTD coding standard. Shall I redefine all parameters, or at least the one I need, using the kernel style?

In my driver I will have something like:
```
zstd_cctx_set_param(ctx, ZSTD_c_blockDelimiters, ZSTD_sf_explicitBlockDelimiters);
```

contrib/linux-kernel/linux_zstd.h
```diff
@@ -480,4 +490,53 @@ typedef ZSTD_FrameHeader zstd_frame_header;
...
+ * @dst_capacity: The size of the destination buffer.
+ * @in_seqs: The array of zstd_sequence to compress.
+ * @in_seqs_size: The number of sequences in in_seqs.
+ * @src: The input data to compress.
+ * @src_size: The size of the data to compress.
+ *
+ * Return: The compressed size or an error, which can be checked using
+ * 	   zstd_is_error().
+ */
+size_t zstd_compress_sequences(zstd_cctx *cctx, void* dst, size_t dst_capacity,
```
gcabiddu: Shall we keep this even if it might end up not being used in the kernel?

gcabiddu:
> > In order for the kernel to compile in contrib/linux-kernel/mem.h, #include <asm/unaligned.h> should be replaced with #include <linux/unaligned.h>. I haven't made that change yet as it was causing some regressions in the tests.
> 
> You should be able to fix the tests by moving this file:
> 
> https://github.com/facebook/zstd/blob/dev/contrib/linux-kernel/test/include/asm/unaligned.h
> 
> Into this directory:
> 
> https://github.com/facebook/zstd/tree/dev/contrib/linux-kernel/test/include/linux
> 
> This include directory fakes the Linux headers that Zstd depends on.

Thanks @terrelln - I sorted it by moving that include as you suggested.

contrib/linux-kernel/zstd_compress_module.c
```diff
@@ -216,5 +222,25 @@ void zstd_register_sequence_producer(
 }
 EXPORT_SYMBOL(zstd_register_sequence_producer);
 
+size_t zstd_compress_sequences(zstd_cctx *cctx, void* dst, size_t dst_capacity,
+			       const zstd_sequence *in_seqs, size_t in_seqs_size,
+			       const void* src, size_t src_size)
+{
+	return ZSTD_compressSequences(cctx, dst, dst_capacity, in_seqs,
```
pps83: spaces instead of tabs and everywhere else

gcabiddu: This code will be ported to the kernel automatically. The kernel coding stiles uses tabs instead of spaces:
https://www.kernel.org/doc/html/latest/process/coding-style.html#indentation

pps83: oh, sorry, didn't notice it was in contrib.

contrib/linux-kernel/linux_zstd.h
terrelln: If we don't need it, and don't expect to need it, we can remove it. If we need to add it later we can.

contrib/linux-kernel/linux_zstd.h
terrelln: I could go either way. I don't think we absolutely need to, given that we don't have many users of this API.

contrib/linux-kernel/linux_zstd.h
gcabiddu: Thanks @terrelln. I updated the PR.

Who usually marks the comments as resolved? The submitter or the maintainer? Thanks.

terrelln (approved):

PR #4261, opened 2025-01-22
Add Controls to Constrain Window Size to Protocol Limits

felixhandte:
This PR adds convenience options that ensure Zstd produces compressed outputs compatible with protocols that place limits on the window size. Currently, it supports two protocols:

- The `zstd` HTTP Content-Encoding ([RFC 8878](https://www.rfc-editor.org/rfc/rfc8878.html#name-content-encoding) and [RFC 9659](https://www.rfc-editor.org/rfc/rfc9659.html#name-window-size)).
- The `dcz` HTTP Content-Encoding ([`draft-ietf-httpbis-compression-dictionary-19`](https://www.ietf.org/archive/id/draft-ietf-httpbis-compression-dictionary-19.html#name-dictionary-compressed-zstan)).

This option is exposed as a new CCtx param `ZSTD_c_constrainWindowForProtocol` and a corresponding CLI argument `--constrain-window`.

To-Do:

- [x] Add CCtx Param.
- [x] Add CLI Flags.
- [x] Add tests.


felixhandte:
cc @pmeenan, @nidhijaju. We'd talked about this being desirable for adoption of `dcz`. Take a look and make sure it meets your needs? It should go out in the next Zstd release which is coming very soon.

lib/zstd.h
```diff
@@ -2355,6 +2357,70 @@ ZSTDLIB_STATIC_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const vo
...
+  ZSTD_ConstrainWindow_disable = ZSTD_ps_disable,
+
+  /* Constrains the window size to comply with the limits imposed on the
+   * `zstd` Content-Encoding, specified by RFCs 8878 and 9659. I.e., liimits
+   * the window to <= 8MB. */
+  ZSTD_ConstrainWindow_HTTP_Zstd = 3,
+
+  /* Constrains the window size to comply with the limits imposed on the
+   * `dcz` Content-Encoding, as specified by the Compression Dictionary
+   * Transport protocol.
```
pmeenan, line 2380: Worth mentioning what the actual limit is?

pmeenan (commented):
LGTM, just the one nit.

We also talked about an option to add the dictionary hash in the skippable frame automatically. I'm assuming that will come separately, maybe after the RFC publishes.

FWIW, the draft was approved for publication and is just held up by the shared brotli normative reference. The shared brotli draft was also just approved for publication but has a normative reference to highway hash which will probably drag out the actual publication a bit.

lib/zstd.h
```diff
@@ -2355,6 +2357,70 @@ ZSTDLIB_STATIC_API size_t ZSTD_CCtx_refPrefix_advanced(ZSTD_CCtx* cctx, const vo
 #define ZSTD_c_repcodeResolution ZSTD_c_experimentalParam19
 #define ZSTD_c_searchForExternalRepcodes ZSTD_c_experimentalParam19 /* older name */
 
+/**
+ * Used in conjunction with the `ZSTD_c_constrainWindowForProtocol` option
+ * described below.
+ */
+typedef enum {
+  /* Equivalent to disabling. */
+  ZSTD_ConstrainWindow_auto = ZSTD_ps_auto,
```
nidhijaju, line 2366: Just curious, what is the difference between "auto" and "disable", they seem basically the same? Do you see a situation where they aren't the same in the future?

The command line option also only has "none".

nidhijaju (commented):
Looks good to me, thanks for putting this together!

lib/zstd.h
felixhandte: They're just here by analogy to the [`ZSTD_paramSwitch_e`](https://github.com/facebook/zstd/blob/v1.5.6/lib/zstd.h#L1414-L1422) enum, which is used for most cctx param options. They behave identically.

lib/compress/zstd_compress.c
```diff
@@ -1462,6 +1480,35 @@ static U32 ZSTD_dictAndWindowLog(U32 windowLog, U64 srcSize, U64 dictSize)
...
+        return 23; /* 8 MB */
+    case ZSTD_ConstrainWindow_HTTP_DCZ: {
+        U32 wlog;
+        if (dictSize < 8 MB) {
+            wlog = 23; /* 8 MB */
+        } else if (dictSize >= 128 MB) {
+            wlog = 27; /* 128 MB */
+        } else {
+            dictSize += dictSize >> 2; /* dictSize *= 1.25 */
+            wlog = ZSTD_highbit32((U32)dictSize);
```
Cyan4973, line 1503: We will need something more accurate than a power of 2.

The Zstandard specification makes it possible to specify a window size which is a power of 2 + n * 1/8th of the power of 2. This is the reason why the `*1.25` rule was selected, because it guarantees that such a window size, both larger the `dictSize`, but not by too much, necessarily exists.

If we stick to pure power of 2, such a window size may not exist. 
For example, 12 MB window size will become 15 MB max, which will be round down to 8 MB.
As a consequence, the dictionary will only remain valid up to 8 MB.
In contrast, using the 1/8th increment, it would be possible to match the 15 MB maximum window size.

One of the issues though is that the current code base is unprepared to handle window sizes which are not power of 2. Even if it can, this information cannot be returned as `wlog`.
So we probably need some in-depth tech talk on this topic.

Cyan4973 (changes requested):

PR #4264, opened 2025-01-23
Move STATIC_BMI2 define to portability_macros.h

pps83:
This change makes sure that STATIC_BMI and DYNAMIC_BMI are defined close to each other, and not in other headers.
Also, it reorders `__BMI2__` check:
 + if `__BMI2__` defined, then set STATIC_BMI2 for all compilers
 + use `defined(_MSC_VER) && defined(__AVX2__)` as fallback for ms compiler

### DIFF
--- a/lib/common/compiler.h
+++ b/lib/common/compiler.h
@@ -207,21 +207,6 @@
 #  pragma warning(disable : 4324)        /* disable: C4324: padded structure */
 #endif
 
-/* Like DYNAMIC_BMI2 but for compile time determination of BMI2 support */
-#ifndef STATIC_BMI2
-#  if defined(_MSC_VER)
-#    ifdef __AVX2__  /* MSVC does not have a BMI2 specific flag, but every CPU that supports AVX2 also supports BMI2 */
-#       define STATIC_BMI2 1
-#    endif
-#  elif defined(__BMI2__)
-#    define STATIC_BMI2 1
-#  endif
-#endif
-
-#ifndef STATIC_BMI2
-#  define STATIC_BMI2 0
-#endif
-
 /* compile time determination of SIMD support */
 #if !defined(ZSTD_NO_INTRINSICS)
 #  if defined(__AVX2__)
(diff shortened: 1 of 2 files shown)


pps83:
Note, portability_macros.h is included only in 3 places:
 1. compiler.h
 2. zstd_cwksp.h
 3. huf_decompress_amd64.S

compiler.h does include portability_macros.h.
zstd_cwksp.h includes compiler.h right after portability_macros.h
huf_decompress_amd64.S is asm file, doesn't check for STATIC_BMI

Cyan4973:
I'm supportive of this change.

I'm also wondering if `DYNAMIC_BMI2` is impacted by this welcome ordering change.
I would have expected `DYNAMIC_BMI2` to depend on the state of `STATIC_BMI2`.

Cyan4973:
Also, cc @terrelln , since it might be relevant to kernel code and assembly units

pps83:
> I would have expected `DYNAMIC_BMI2` to depend on the state of `STATIC_BMI2`.

yep, but then I get more and more changes on top :)
in short, my understanding: DYNAMIC_BMI2 should only be set to 1 if STATIC_BMI2 is 0. But then, looking into the code, it seems it would affect it, as the checks do not seem to be consistent. And in some cases, code is quite questionable.

pps83:
@terrelln ping

pps83:
> yep, but then I get more and more changes on top :) in short, my understanding: DYNAMIC_BMI2 should only be set to 1 if STATIC_BMI2 is 0. But then, looking into the code, it seems it would affect it, as the checks do not seem to be consistent. And in some cases, code is quite questionable.

After the change, step by step changes:
 * there are only a few places where `__BMI2__` is used, it should be replaced with `STATIC_BMI2` (including [this one](https://github.com/facebook/zstd/pull/4264#issuecomment-2611138276))
 * there is a place with the check `if (defined(__BMI__) || defined(__BMI2__)) && defined(__GNUC__)`. Perhaps, checking for `__BMI__` is enough: if `__BMI2__` is enabled, that means that `__BMI__` is also enabled.

Cyan4973 (approved):
Looks good to me

PR #4273, opened 2025-01-29
updated xxhash to latest `dev` branch commit

Cyan4973:
this version supports redirection of `XXH_memcpy` and `XXH_memset`.
This capability is used during local zstd adaptation to redirect to `zstd_deps.h`'s `ZSTD_memcpy` and `ZSTD_memset`.

### DIFF
--- a/lib/common/xxhash.h
+++ b/lib/common/xxhash.h
@@ -19,6 +19,10 @@
 # define XXH_NAMESPACE ZSTD_
 #endif
 
+#include "zstd_deps.h"  /* ZSTD_memcpy, ZSTD_memset */
+#define XXH_memcpy ZSTD_memcpy
+#define XXH_memset ZSTD_memset
+
 /*!
  * @mainpage xxHash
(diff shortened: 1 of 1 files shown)


Cyan4973:
The visual studio compilation error in `xxhash.h` is kinda new.

More worrying, I can't make sense of the error message :
```
D:\a\zstd\zstd\lib\common\xxhash.h(3182): error C4702: unreachable code [D:\a\zstd\zstd\build\VS2010\fuzzer\fuzzer.vcxproj]
```

When looking at `xxhash.h(3182)`, it does not feel correct.
This code is obviously reachable and reached, it's very simple code, nothing weird happening there.

One possibility is that the line number `3182` might be completely wrong,
but then it doesn't help identifying what could be the problem.

No other compiler triggers this warning.
I also note that the issue only happens in combination with `fuzzer.c`, compilation of `xxhash` is fine within `libzstd` or any other program.

PR #4275, opened 2025-01-30
Support Fractional-Power-of-Two Window Sizes in Compression

felixhandte:
This PR adds support for configuring and performing compression with non-power-of-two window sizes.

The frame format and decoder have always supported more window sizes than just powers of two. The frame format defines a 3 bit window mantissa field in addition to the 4 bit window log descriptor. However, the compressor has only ever supported power-of-two windows. This PR changes that. In order to do that, it makes a number of changes:

- It separates the public and private CParams struct definitions.
- It adds a `windowFrac` field to the private CParams (but not to the public one, because we are scared of changing that struct definition.
- It adds a new CCtx Param `ZSTD_c_windowFrac` which allows users to set this param.

To-Do:

- [x] Testing.
- [x] CLI Support.


lib/compress/clevels.h
```diff
@@ -22,110 +22,110 @@
 __attribute__((__unused__))
 #endif
 
-static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEVEL+1] = {
+static const ZSTD_CParams ZSTD_defaultCParameters[4][ZSTD_MAX_CLEVEL+1] = {
```
Cyan4973, line 25: Another way to pass this new parameter is to store it directly in the existing `ZSTD_CCtx_params` structure, which is internal only, can be updated anytime, 
and is already used by many other advanced parameters,
such as for example the recently added `ZSTD_c_blockSplitterLevel`.

That's my recommendation to introduce any parameter that goes beyond `ZSTD_compressionParameters` definition.

Cyan4973 (changes requested):
I'm not convinced by the strategy of creating a new intermediate internal structure to pass the new parameter.
We have an established pattern to add new parameters via `ZSTD_CCtx_params`.
I prefer we stick to it and don't introduce a new pattern specific to this new parameter.

.github/workflows/dev-long-tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     steps:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # tag=v4.1.1
     - name: make test
-      run: make test
+      run: MOREFLAGS="-DZSTD_WINDOW_ALLOW_PICKING_FRACTIONAL_SIZES=1" make test
```
Cyan4973, line 31: I initially thought this compilation variable was enabling or disabling some tests specific to fractional sizes.
But looking at the code below, it rather seems to impact the behavior of the new `windowLogAndFrac` functions.
In which case, I wonder about the usefulness of this compilation variable.
Why would it be useful to ignore the fractional parameter whenever a compile-time variable is not set ?

felixhandte:
I agree that it would have been more expedient to just put it in the `CCtxParams`, but it seems bad to separate these highly interdependent variables.

In this future where the Zstd compression side supports fractional window sizes, it is a **bug** if you look at just the `windowLog` and don't look at the `windowFrac` at the same time. I worry that if they're not *right* next to each other, it is both mechanically harder to make both available to all the places that interact with window sizing, it is also easier to make mistakes as a developer and look only at the `windowLog`. (Similarly, if you look closely at the code for setting these values, there are nuances that require that they interact.)

And, as we discussed, the match-finders need access to this parameter, so it needs to be available from the `MatchState`, which currently only has the `CParams`, and not the `CCtxParams`.

I considered an alternate approach where, rather than introduce a new variable, we change the internal `windowLog` to be a "`windowDesc`" which carries both components, as, e.g., `windowLog << 3 | windowFrac`. But then even if you didn't have to change the struct shape, you'd still have to translate the values of the internal cParams with the public ones, and it's actually harder to validate that you're doing that correctly if they are the same struct.

.github/workflows/dev-long-tests.yml
felixhandte: The `windowFrac` parameter isn't ignored. Rather, this is the macro that controls whether Zstd allows *itself* to select fractional window sizes.

There's several different features that this PR implements:

1. Correctly transport and handle fractional window sizes in `lib/compress`.
2. Allow explicit configuration of fractional window sizes.
3. Automatically select fractional window sizes when Zstd picks the window size.

This PR implements all three of those things. We need #s 1 and 2 for #4261. As discussed, we don't want to ship the 3rd in this release, since we're so close to it and it's a significant change in behavior that in theory has user visible effects. So that third part could have just been in a separate PR. *Except*, having that support is really the best way to test and exercise the implementation. I can add explicit tests, and I have, but turning it on for *all* of the tests is much better. Hence the decision to include that functionality in this PR, but put it behind a macro that is off by default: users will get a zstd that supports, but does not on its own select, fractional window sizes, but we can land tests that broadly exercise this new feature.

And then immediately after the release, we can land a PR that removes this macro guard and turns support on by default.

tests/fuzzer.c
```diff
@@ -3670,7 +3781,7 @@ static int basicUnitTests(U32 const seed, double compressibility)
     {   ZSTD_CCtx* const cctx = ZSTD_createCCtx();
         ZSTD_DCtx* const dctx = ZSTD_createDCtx();
         static const size_t dictSize = 65 KB;
-        static const size_t blockSize = 100 KB;   /* won't cause pb with small dict size */
+        static const size_t blockSize = 71 KB;   /* won't cause pb with small dict size */
```
Cyan4973, line 3784: Interesting,
why changing `blockSize` in this test ?
Is this related to non-power-of-2 window sizes ?
In which case, what's better tested with 71 KB than with previous 100 KB ?

lib/compress/zstd_compress_internal.h
```diff
@@ -191,6 +191,29 @@ size_t ZSTD_buildBlockEntropyStats(
...
+    unsigned chainLog;        /**< fully searched segment : larger == more compression, slower, more memory (useless for fast) */
+    unsigned hashLog;         /**< dispatch table : larger == faster, more memory */
+    unsigned searchLog;       /**< nb of searches : larger == more compression, slower */
+    unsigned minMatch;        /**< match length searched : larger == faster decompression, sometimes less compression */
+    unsigned targetLength;    /**< acceptable match size for optimal parser (only) : larger == more compression, slower */
+    ZSTD_strategy strategy;   /**< see ZSTD_strategy definition above */
+} ZSTD_CParams;
+
+/**
+ * Internal equivalent of public ZSTD_Params struct, wrapping the internal
```
Cyan4973, line 209: nit: public `ZSTD_parameters` struct

lib/compress/zstd_compress_internal.h
```diff
@@ -1538,12 +1652,32 @@ BlockSummary ZSTD_get1BlockSummary(const ZSTD_Sequence* seqs, size_t nbSeqs);
...
  * ============================================================== */
 
+/* ZSTD_getCParamsFromPublicCParams(), ZSTD_getPublicCParamsFromCParams() :
+ * Translates between the public and internal structs.
+ *
+ * Note: as these structs diverge, this may increasingly become a lossy
+ * translation. The only long-term justified use of these should be at the
+ * User API.
+ */
+ZSTD_CParams ZSTD_getCParamsFromPublicCParams(ZSTD_compressionParameters cParams);
```
Cyan4973, line 1662: #### Using Published Names for Clarity

When referring to specific parameters or functions, it feels more consistent to use their published names to reduce confusion. In the case of `PublicCParams`, while its meaning can be inferred from context, using the public name directly feels preferable.

Consider renaming the following functions to use more consistent names:

* `ZSTD_getCParamsFromPublicCParams()` => `ZSTD_getCParamsFromCompressionParameters()`
* `ZSTD_getPublicCParamsFromCParams()` => `ZSTD_getCompressionParametersFromCParams()`
* `ZSTD_getParamsFromPublicParams()` => `ZSTD_getParamsFromParameters()`
* etc.

By using published names consistently, code becomes more readable for future contributors.

Cyan4973: also: minor: `get` feels very generic.
`convert` might be more accurate to describe this operation.

Cyan4973:
Just some minor naming suggestions.

`zstd_compress.c` is relatively straightforward to review, compared to #4305, because the code structure remains the same. So no comment there.

tests/fuzzer.c
felixhandte: Yeah, the commit message for the commit that makes this change describes the motivation:

> Using a blocksize of 100KB with a 65KB dictionary just *happens* to work with
the current window-sizing logic, because it has to round all the way up to
the next power of two. But it doesn't *have* to work.
>
> So this pre-emptively fixes up this test to use a size that will work after
switching to tighter window sizing, but still uses a block size that is
bigger than the dict size in case that matters for some reason?

As discussed elsewhere, this PR doesn't change the default window sizing logic: it still rounds up to the next power of two by default (this is what is controlled by the `ZSTD_WINDOW_ALLOW_PICKING_FRACTIONAL_SIZES` macro, which defaults to false). Which is to say, this PR doesn't break this test. But if you set `-DZSTD_WINDOW_ALLOW_PICKING_FRACTIONAL_SIZES=1`, this test as it exists will break, because Zstd will pick a 72KB window size, and blocks can't be larger than the window size, and this test tries to invoke the block-level compression API with a larger input.

So the change here preserves the behavior of the test of checking a block larger than the dictionary (although I don't think that's actually an important property of how this test is set up), while complying with the new, smaller selected window size.

PR #4276, opened 2025-01-30
fix speed of --patch-from mode at high compression levels

Cyan4973:
When employing `--patch-from` in combination with very high compression levels,
compression speed plummets significantly, becoming extremely slow.

This issue is mitigated by this patch,
which sensibly reduces compression time for this scenario.

Benchmark: patching `linux-6.7.tar` (~1.4GB) from `linux-6.6.tar`
on a M1 Pro laptop (employing 2 threads by default)

| level | `dev` | this PR |
| --- | ---:| ---:|
| 18 | 250 s | 80 s|
| 19 | 350 s | 118 s |
| 20 | 420 s | 145 s |
| 21 | 510 s | 200 s |
| 22 | 635 s | 320 s |

The impact is even more pronounced with a larger amount of threads.
Benchmark on a core i7-9700k with 6 threads : 
| level | `dev` | this PR |
| --- | ---:| ---:|
| 18 | 208 s | 34 s|
| 19 | 264 s | 48 s |
| 20 | 300 s | 63 s |
| 21 | 352 s | 88 s |
| 22 | 472 s | 160 s |

Also: improved memory usage of `--patch-from`, by skipping an internal `CDict` creation that was duplicating reference content in memory.

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -12,8 +12,8 @@
  * that shall **only** be used by modules within lib/compress.
  */
 
-#ifndef ZSTD_COMPRESS_H
-#define ZSTD_COMPRESS_H
+#ifndef ZSTD_COMPRESS_INTERNAL_H
+#define ZSTD_COMPRESS_INTERNAL_H
 
 /*-*************************************
 *  Dependencies
@@ -21,10 +21,10 @@
 #include "../common/zstd_internal.h"
 #include "zstd_cwksp.h"
 #ifdef ZSTD_MULTITHREAD
-#  include "zstdmt_compress.h"
+#  include "zstdmt_compress.h" /* ZSTDMT_CCtx */
 #endif
 #include "../common/bits.h" /* ZSTD_highbit32, ZSTD_NbCommonBytes */
-#include "zstd_preSplit.h" /* ZSTD_SLIPBLOCK_WORKSPACESIZE */
+#include "zstd_preSplit.h"  /* ZSTD_SLIPBLOCK_WORKSPACESIZE */
 
 /*-*************************************
 *  Constants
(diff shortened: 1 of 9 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -1334,9 +1335,32 @@ size_t ZSTDMT_initCStream_internal(
...
+    mtctx->cdictLocal = NULL;
+    mtctx->cdict = NULL;
+    if (dict) {
+        if (dictContentType == ZSTD_dct_rawContent) {
+            mtctx->inBuff.prefix.start = (const BYTE*)dict;
+            mtctx->inBuff.prefix.size = dictSize;
+        } else {
+            /* note : a loadPrefix becomes an internal CDict */
+            mtctx->cdictLocal = ZSTD_createCDict_advanced(dict, dictSize,
+                                                        dictLoadMethod, dictContentType,
```
terrelln: We don't actually care about the `dictLoadMethod` here, right? If it was `ZSTD_dlm_byCopy`, the outer `ZSTD_CCtx` must have already token ownership of the buffer. So we can use `byRef` here.
```suggestion
                                                        ZSTD_dlm_byRef, dictContentType,
```

lib/compress/zstdmt_compress.c
```diff
@@ -1232,7 +1248,7 @@ static size_t ZSTDMT_computeOverlapSize(const ZSTD_CCtx_params* params)
 
 size_t ZSTDMT_initCStream_internal(
         ZSTDMT_CCtx* mtctx,
-        const void* dict, size_t dictSize, ZSTD_dictContentType_e dictContentType,
+        const void* dict, size_t dictSize, ZSTD_dictContentType_e dictContentType, ZSTD_dictLoadMethod_e dictLoadMethod,
```
terrelln: Since we can always use `byRef`, we don't need to ass a `dictLoadMethod` in here.

lib/compress/zstd_compress_internal.h
```diff
@@ -46,11 +46,12 @@
 typedef enum { ZSTDcs_created=0, ZSTDcs_init, ZSTDcs_ongoing, ZSTDcs_ending } ZSTD_compressionStage_e;
 typedef enum { zcss_init=0, zcss_load, zcss_flush } ZSTD_cStreamStage;
 
-typedef struct ZSTD_prefixDict_s {
+typedef struct {
     const void* dict;
     size_t dictSize;
     ZSTD_dictContentType_e dictContentType;
-} ZSTD_prefixDict;
+    ZSTD_dictLoadMethod_e loadMethod;
```
terrelln: Since we can always use `byRef` in ZSTDMT, we don't need to store it here.

terrelln (approved):

PR #4279, opened 2025-02-01
Hc patchfrom part2

Cyan4973:
just checking CI

### DIFF
--- a/.github/workflows/dev-long-tests.yml
+++ b/.github/workflows/dev-long-tests.yml
@@ -67,11 +67,11 @@ jobs:
     - name: thread sanitizer zstreamtest
       run: CC=clang ZSTREAM_TESTTIME=-T3mn make tsan-test-zstream
 
-  ubsan-zstreamtest:
+  uasan-zstreamtest:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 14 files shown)


Cyan4973:
The new MT strategy in this PR favors keeping the jobs busy,
while the previous approach favored keeping I/O maximally completed.

As a consequence, this has an impact on the state of internal buffers at wake-up events,
and the `--adapt` mode, which was depending on them, doesn't work anymore.

This leaves a few option:
- Do not change the MT mode
- Do no longer support `--adapt`.
- Refactor the MT mode to make it compatible with `--adapt`
- Refactor `--adapt` to no longer depend on internal buffer state.

The last 2 items obviously cost a lot more time than the first 2 ones,
and this unfortunately may be the most critical ingredient in this equation.

Cyan4973:
Changed for a combined strategy of waking up on flush event + waking up on job slot availability.
This requires sharing a condition variable, which is more complex to setup.

Cyan4973:
This PR won't be merged. It's too complex and not stable enough.

I'll see if some of the simpler ideas can provide benefits and create a separate PR with them if that's the case.

PR #4305, opened 2025-02-20
Put `windowFrac` in the CCtxParams, not the CParams

felixhandte:
This PR is a set of commits stacked on top of #4275.

It refactors much of that PR, by moving where we store the `windowFrac` parameter from the CParams to instead storing it in the CCtxParams. This resolve the unpleasant part of that PR, which required splitting the CParams (and Params) structs into public and private variants.

This version has its own downsides, which requires that we convert many of the places that we handle and pass CParams around to instead pass around CCtxParams. That includes switching the MatchState to point to a CCtxParams rather than hold a CParams. This also requires that the CDict hold a CCtxParams that its MatchState can point to.

A particularly annoying result of this refactor is that where previously there was often a straightforward path of resolution from compression level -> CParams -> CCtxParams, the CCtxParams now have to be used in two different ways. We have to use the CCtxParams where previously we would transport CParams. But in those places, the CCtxParams are incomplete and are *different* from the real CCtxParams (because only the CParams part is set), and have to be translated/resolved into the real params before they are actually used. But because they're the same type, this distinction is not easy to discover, and it's not clear which kind of CCtxParams a particular instance is at any given point.

One nice corollary though is that this change does allow us in some internal APIs to remove other parameters that are now redundant now that the function has access to the whole CCtxParams.

This PR is a pure refactor and intends to introduce no functional changes.

I continue to believe we should abandon this PR and merge #4275 as-is.


lib/compress/zstd_compress_internal.h
```diff
@@ -1077,6 +1081,88 @@ MEM_STATIC ZSTD_dictMode_e ZSTD_matchState_dictMode(const ZSTD_MatchState_t *ms)
             ZSTD_noDict;
 }
 
+/**
+ * Fractional window sizes can always be picked by the user explicitly
+ * setting ZSTD_c_windowFrac. This macro controls whether, when Zstd is
+ * picking a window size itself, it is allowed to pick a non-power-of-two
```
Cyan4973, line 1087: 👍

lib/compress/zstd_compress_internal.h
```diff
@@ -1077,6 +1081,88 @@ MEM_STATIC ZSTD_dictMode_e ZSTD_matchState_dictMode(const ZSTD_MatchState_t *ms)
...
+#endif
+
+/**
+ * Return the window size described by the windowLog and windowFrac in the
+ * provided CParams.
+ */
+MEM_STATIC U32 ZSTD_windowSize(const ZSTD_CCtx_params* params) {
+    return (U32)(((8ull + params->windowFrac) << params->cParams.windowLog) >> 3);
+}
+MEM_STATIC U32 ZSTD_windowSizeLDM(const ldmParams_t* ldmParams) {
```
Cyan4973, line 1103: minor: it seems this function is only useful within `zstd_ldm.c`.
In which case, is there any benefit in sharing it more widely within `zstd_compress_internal.h` ?

lib/compress/zstd_compress_internal.h
```diff
@@ -1077,6 +1081,88 @@ MEM_STATIC ZSTD_dictMode_e ZSTD_matchState_dictMode(const ZSTD_MatchState_t *ms)
...
+        return 0;
+    }
+    return 1;
+}
+
+/**
+ * Calculates the minimum legal window log and fraction that contain the
+ * provided source size.
+ */
+MEM_STATIC void ZSTD_setMinimalWindowLogAndFrac(ZSTD_CCtx_params* params, const U32 srcSize, const U32 minWindowLog) {
```
Cyan4973, line 1135: Same question here,
`ZSTD_setMinimalWindowLogAndFrac()` seems only useful within `zstd_compress.c`.
What benefit is there to share it more widely within `zstd_compress_internal.h` ?

Note that `ZSTD_windowLogAndFracAreMinimal()` above seems to be a dependency of this function.

lib/compress/zstd_compress_internal.h
```diff
@@ -1100,10 +1186,12 @@ MEM_STATIC U32 ZSTD_window_canOverflowCorrect(ZSTD_window_t const window,
                                               U32 loadedDictEnd,
                                               void const* src)
 {
+    /* overflow correction only handles power-of-two index moves. */
+    U32 const roundedMaxDist = 1u << (ZSTD_highbit32(maxDist - 1) + 1);
```
Cyan4973, line 1190: question: is this change fixing some kind of potential bug ?

lib/compress/zstd_compress_internal.h
```diff
@@ -1562,6 +1656,11 @@ void ZSTD_resetSeqStore(SeqStore_t* ssPtr);
  *  as the name implies */
 ZSTD_compressionParameters ZSTD_getCParamsFromCDict(const ZSTD_CDict* cdict);
 
+/*! ZSTD_checkCCtxCParams_internal() :
+ *  Checks the CParams in the CCtxParams (including related parameters not
+ * *actually* stored in the CParams struct). */
```
Cyan4973, line 1661: maybe name them ?
It seems to be limited to `params->windowFrac` at this point.

lib/compress/zstd_ldm.c
```diff
@@ -133,9 +133,10 @@ static size_t ZSTD_ldm_gear_feed(ldmRollingHashState_t* state,
 }
 
 void ZSTD_ldm_adjustParameters(ldmParams_t* params,
-                        const ZSTD_compressionParameters* cParams)
+                               const ZSTD_CCtx_params* cctxParams)
 {
-    params->windowLog = cParams->windowLog;
+    params->windowLog = cctxParams->cParams.windowLog;
```
Cyan4973, line 138: minor:
you could use the same trick you employed in other parsers,
by defining `const ZSTD_compressionParameters* const cParams = &cctxParams->cParams;` at the top of the function, and then most of the rest of the function would remain identical.

programs/zstd.1.md
```diff
@@ -387,6 +387,18 @@ The list of available _options_:
     Note: If `windowLog` is set to larger than 27, `--long=windowLog` or
     `--memory=windowSize` needs to be passed to the decompressor.
 
+- `windowFrac`=_wfrac_, `wfrac`=_wfrac_:
```
Cyan4973, line 390: 👍

tests/fuzz/zstd_helpers.c
```diff
@@ -102,6 +102,7 @@ void FUZZ_setRandomParameters(ZSTD_CCtx *cctx, size_t srcSize, FUZZ_dataProducer
 {
     ZSTD_compressionParameters cParams = FUZZ_randomCParams(srcSize, producer);
     set(cctx, ZSTD_c_windowLog, cParams.windowLog);
+    setRand(cctx, ZSTD_c_windowFrac, 0, 7, producer);
```
Cyan4973, line 105: 👍 
Great way to naturally bring `windowFrac` within the fuzzer framework.

lib/compress/zstd_compress.c
```diff
@@ -87,11 +87,10 @@ struct ZSTD_CDict_s {
...
     U32 dictID;
-    int compressionLevel; /* 0 indicates that advanced API was used to select CDict params */
-    ZSTD_ParamSwitch_e useRowMatchFinder; /* Indicates whether the CDict was created with params that would use
-                                           * row-based matchfinder. Unless the cdict is reloaded, we will use
-                                           * the same greedy/lazy matchfinder at compression time.
-                                           */
+    ZSTD_CCtx_params params; /* Storage for matchState.cctxParams pointer.
+                              * Not all member fields are used. Used fields:
+                              * cParams, compressionLevel, useRowMatchFinder. */
+    /* A compression level of 0 indicates that advanced API was used to select CDict params */
```
Cyan4973, line 93: 👍

lib/compress/zstd_compress.c
```diff
@@ -311,7 +311,7 @@ static ZSTD_CCtx_params ZSTD_makeCCtxParamsFromCParams(
     /* Adjust advanced params according to cParams */
     cctxParams.ldmParams.enableLdm = ZSTD_resolveEnableLdm(cctxParams.ldmParams.enableLdm, &cParams);
     if (cctxParams.ldmParams.enableLdm == ZSTD_ps_enable) {
-        ZSTD_ldm_adjustParameters(&cctxParams.ldmParams, &cParams);
+        ZSTD_ldm_adjustParameters(&cctxParams.ldmParams, &cctxParams);
```
Cyan4973, line 314: given that `cctxParams.ldmParams` is within `cctxParams`,
one has to wonder why it is initialised separately.
It looks to me that this is just a part of adjusting `cctxParams`, and should be bundled there,
thus reducing risk of forgetting this action in other places of the code.

lib/compress/zstd_compress.c
```diff
@@ -365,33 +365,46 @@ size_t ZSTD_CCtxParams_init(ZSTD_CCtx_params* cctxParams, int compressionLevel)
...
     DEBUGLOG(4, "ZSTD_CCtxParams_init_internal: useRowMatchFinder=%d, useBlockSplitter=%d ldm=%d",
                 cctxParams->useRowMatchFinder, cctxParams->postBlockSplitter, cctxParams->ldmParams.enableLdm);
 }
 
+/**
+ * Initializes `cctxParams` from `params` and `compressionLevel`.
+ * @param compressionLevel If params are derived from a compression level then that compression level, otherwise ZSTD_NO_CLEVEL.
+ */
+static void
+ZSTD_CCtxParams_init_internal(ZSTD_CCtx_params* cctxParams,
```
Cyan4973, line 396: minor: API consistency:
above function is named `ZSTD_CCtxParams_init_fromCCtxParams()`,
so maybe the function below would benefit a rename along the same lines.

lib/compress/zstd_compress.c
```diff
@@ -1469,56 +1520,57 @@ static U32 ZSTD_dictAndWindowLog(U32 windowLog, U64 srcSize, U64 dictSize)
  * `mode` is the mode for parameter adjustment. See docs for `ZSTD_CParamMode_e`.
  *  note : `srcSize==0` means 0!
  *  condition : cPar is presumed validated (can be checked using ZSTD_checkCParams()). */
-static ZSTD_compressionParameters
-ZSTD_adjustCParams_internal(ZSTD_compressionParameters cPar,
+static void
+ZSTD_adjustCParams_internal(ZSTD_CCtx_params* params,
```
Cyan4973, line 1524: Could benefit from a rename: `ZSTD_adjustCCtxParams_internal()`.
In turn, the new name would help in extending the scope of this function beyond the original `ZSTD_compressionParameters` fields.

lib/compress/zstd_compress.c
```diff
@@ -1594,60 +1669,88 @@ ZSTD_adjustCParams_internal(ZSTD_compressionParameters cPar,
...
+        size_t dictSize,
+        ZSTD_CParamMode_e mode);
+static ZSTD_parameters ZSTD_getParams_internal(
+        int compressionLevel,
+        unsigned long long srcSizeHint,
+        size_t dictSize, ZSTD_CParamMode_e
+        mode);
+
+/* fills unset fields in `dst` with values from `defaults` */
+static void ZSTD_fillCParams(
```
Cyan4973, line 1719: The naming feels weird.
What is being filled is a `ZSTD_CCtx_params*` object, not `CParams`,
albeit admittedly on a reduced internal scope which is somewhat associated with `CParams`,
but is that actually a good idea to keep the same scope as it was before ?

It looks to me that this operation is about setting parameters that are not yet set,
using a list provided in `defaults`.
Then a few more parameters must still be set via functions.

But is it meaningful to separate these 2 operations, or should they all be combined in the same one ?

lib/compress/zstd_compress.c
```diff
@@ -1753,23 +1853,22 @@ static size_t ZSTD_estimateCCtxSize_usingCCtxParams_internal(
...
     /* estimateCCtxSize is for one-shot compression. So no buffers should
      * be needed. However, we still allocate two 0-sized buffers, which can
      * take space under ASAN. */
     return ZSTD_estimateCCtxSize_usingCCtxParams_internal(
-        &cParams, &params->ldmParams, 1, useRowMatchFinder, 0, 0, ZSTD_CONTENTSIZE_UNKNOWN, ZSTD_hasExtSeqProd(params), params->maxBlockSize);
+        &par, 1, 0, 0, ZSTD_CONTENTSIZE_UNKNOWN);
 }
 
-size_t ZSTD_estimateCCtxSize_usingCParams(ZSTD_compressionParameters cParams)
+static size_t ZSTD_estimateCCtxSize_usingCCtxParamsCParams(const ZSTD_CCtx_params* params)
```
Cyan4973, line 1868: well, `ZSTD_estimateCCtxSize_usingCCtxParams()` feels enough.

Plus, it opens the opportunity to decide `CCtx` size based on other parameters present in `CCtx_params`.
For example, a pretty important one would be `hasExternalSequenceProducer`.

Cyan4973:
While reviewing the `zstd_compress.c` file, I indeed noticed a recurring pattern where the function parameter `CParams` is being replaced with `CCtxParams*`, while maintaining the same scope for the function. This approach has the advantage of minimizing changes to the source code structure.

However, this substitution introduces some scope-related issues. Specifically, `CCtxParams*` is now used to convey only a subset of its content or is updated partially, and in some cases, it is used as a reference to modify other parts of itself. These issues necessitate a reevaluation of the functions' scope, which previously focused solely on `CParams`. In the context of `CCtxParams*` being passed around, such a restricted scope may no longer be relevant.

Unfortunately, addressing these issues requires more than a simple mechanical update from one type to another. 
I now wonder: it may have been beneficial to first transition from `CParams*` to `CCtxParams*` internally before introducing `windowFrac`. This intermediate step would have simplified the subsequent addition of `windowFrac`, reducing the complexity of each step and making them more manageable.

PR #4320, opened 2025-02-26
build_package.bat: fix path to zstd_errors.h, avoid silently ignore of the errors if build failed

### LINKED ISSUE #4318: Release v.1.5.7: prebuilt win-artifacts `zstd-v1.5.7-win*.zip` miss header `include/zstd_errors.h`
**Describe the bug**
Both archives `zstd-v1.5.7-win*.zip` of [release v1.5.7](https://github.com/facebook/zstd/releases/tag/v1.5.7) seem to miss the header file `include/zstd_errors.h` (see line `include/zstd.h:18`).

**To Reproduce**
1. Downloads `zstd-v1.5.7-win*.zip` from [release v1.5.7](https://github.com/facebook/zstd/releases/tag/v1.5.7).
2. Try to build any project included `include/zstd.h`
3. See error:
```
include/zstd.h:18:10: fatal error: zstd_errors.h: No such file or directory
   18 | #include "zstd_errors.h" /* list of errors */
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

**Expected behavior**
The header is included so it'd built successfully.

 - OS: Windows
 - Version: Any
 - Compiler and toolchain: Any

**Additional context**
As workaround one could surely do something like that:
```bash
wget 'https://raw.githubusercontent.com/facebook/zstd/refs/tags/v1.5.7/lib/zstd_errors.h'
```
and everything would work hereafter, but for the sake of completeness it would be nice to have all headers in place
(or the content of header included in file `include/zstd.h` in place of `#include "zstd_errors.h"`).

sebres:
This PR fixes:
- build_package.bat: fix path to `zstd_errors.h` (it is in `lib` not in `lib/common`)
  also closes #4318
- build_package.bat: don't swallow the error(s) by copy, exit with error if it failed somewhere
- GHA/windows-artifacts.yml: don't ignore the error if build-package batch failed

### DIFF
--- a/.github/workflows/windows-artifacts.yml
+++ b/.github/workflows/windows-artifacts.yml
@@ -46,7 +46,7 @@ jobs:
 
     - name: Create artifacts
       run: |
-        ./lib/dll/example/build_package.bat
+        ./lib/dll/example/build_package.bat || exit 1
         mv bin/ zstd-${{ github.ref_name }}-${{matrix.ziparch}}/
         7z a -tzip -mx9 zstd-${{ github.ref_name }}-${{matrix.ziparch}}.zip zstd-${{ github.ref_name }}-${{matrix.ziparch}}/
         cd ..
--- a/lib/dll/example/build_package.bat
+++ b/lib/dll/example/build_package.bat
@@ -15,6 +15,6 @@ COPY lib\dll\example\Makefile bin\example\
 COPY lib\dll\example\fullbench-dll.* bin\example\
 COPY lib\dll\example\README.md bin\
 COPY lib\zstd.h bin\include\
-COPY lib\common\zstd_errors.h bin\include\
+COPY lib\zstd_errors.h bin\include\
 COPY lib\dictBuilder\zdict.h bin\include\
 COPY programs\zstd.exe bin\zstd.exe
--- a/lib/dll/example/build_package.bat
+++ b/lib/dll/example/build_package.bat
@@ -1,20 +1,30 @@
 @ECHO OFF
 MKDIR bin\dll bin\static bin\example bin\include
-COPY tests\fullbench.c bin\example\
-COPY programs\datagen.c bin\example\
-COPY programs\datagen.h bin\example\
-COPY programs\util.h bin\example\
-COPY programs\platform.h bin\example\
-COPY lib\common\mem.h bin\example\
-COPY lib\common\zstd_internal.h bin\example\
-COPY lib\common\error_private.h bin\example\
-COPY lib\common\xxhash.h bin\example\
-COPY lib\libzstd.a bin\static\libzstd_static.lib
-COPY lib\dll\libzstd.* bin\dll\
-COPY lib\dll\example\Makefile bin\example\
-COPY lib\dll\example\fullbench-dll.* bin\example\
-COPY lib\dll\example\README.md bin\
-COPY lib\zstd.h bin\include\
-COPY lib\zstd_errors.h bin\include\
-COPY lib\dictBuilder\zdict.h bin\include\
-COPY programs\zstd.exe bin\zstd.exe
+SET CpyError=
+COPY tests\fullbench.c bin\example\ || (SET CpyError=%CpyError% tests\fullbench.c)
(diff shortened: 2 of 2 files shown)


.github/workflows/windows-artifacts.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     - name: Create artifacts
       run: |
-        ./lib/dll/example/build_package.bat
+        ./lib/dll/example/build_package.bat || exit 1
```
Cyan4973, line 49: 👍

lib/dll/example/build_package.bat
```diff
@@ -1,20 +1,30 @@
...
+COPY lib\common\xxhash.h bin\example\ || (SET CpyError=%CpyError% lib\common\xxhash.h)
+COPY lib\libzstd.a bin\static\libzstd_static.lib || (SET CpyError=%CpyError% lib\libzstd.a)
+COPY lib\dll\libzstd.* bin\dll\ || (SET CpyError=%CpyError% lib\dll\libzstd.*)
+COPY lib\dll\example\Makefile bin\example\ || (SET CpyError=%CpyError% lib\dll\example\Makefile)
+COPY lib\dll\example\fullbench-dll.* bin\example\ || (SET CpyError=%CpyError% lib\dll\example\fullbench)
+COPY lib\dll\example\README.md bin\ || (SET CpyError=%CpyError% lib\dll\example\README.md)
+COPY lib\zstd.h bin\include\ || (SET CpyError=%CpyError% lib\zstd.h)
+COPY lib\zstd_errors.h bin\include\ || (SET CpyError=%CpyError% lib\zstd_errors.h)
+COPY lib\dictBuilder\zdict.h bin\include\ || (SET CpyError=%CpyError% lib\dictBuilder\zdict.h)
+COPY programs\zstd.exe bin\zstd.exe || (SET CpyError=%CpyError% programs\zstd.exe)
```
Cyan4973, line 21: I understand the objective, which is to detect and qualify an error, which is good.
But it comes at a serious repetition cost, which is detrimental to maintenance and evolutions.

Could there be a way to reduce repetition ?
For example a `for ... do ( ... )` iterating over a list of files and destinations ?

sebres: In win-batch? no idea... I am not windows friend (if we could switch to some proper shell, like bash/tclsh/whatever, then sure).
But how often do you change it? copy & paste some line, change file name, that's it.

Cyan4973: That's fine,
I'll figure out something later to deal with repetition

Cyan4973 (approved):

lib/dll/example/build_package.bat
sebres: I guess, one cannot do it simple in windows batch (at least iterator for 2 tokens over set with wildcards like `lib\dll\example\fullbench-dll.*` inside), to reach the file expansion of it at copy command...
Even using nested `for`, I have no idea how, sorry.
Maybe powershell or bash instead?

Cyan4973: I'll try a function, see if it works, 
nothing urgent, it can still get merged and used this way

sebres:
Just for the sake of completeness, shall not one rebuild (or maybe add the header manually to) both artifacts `zstd-v1.5.7-win*.zip` in the release?

Cyan4973:
sure !

Cyan4973:
The packages are updated.

As a data point, I also tried the new script with an intentional error in the list of files.
The error is reported in the log, but the task still completes "successfully", i.e. all green signal.

https://github.com/Cyan4973/zstd/actions/runs/13574068432/job/37945924130

I'll look into it.

sebres:
> but the task still completes "successfully", i.e. all green signal.

Indeed... 

The command processor works as expected in shell... tried it in my CI:
```
$ cd /tmp; echo 'exit /b 0' > ./test.bat; { ./test.bat && echo 'where is my error'; } || echo 'error occurred'

T:\Temp>exit /b 1
error occurred
```
May be the bug is in line 28 (`%errorlevel%` got overwritten by last success), previously I implemented it with `goto :error`, so it was OK, but now it's surely wrong. 
Patch follows.

PR #4321, opened 2025-02-27
Automate Windows Artifact generation and publication

Cyan4973:
which generates the directory for the Windows Release packages.

- Rewrite the script to use a subroutine, reducing the size of each line
- Ensure that the script fails (i.e. generates a red signal in CI) when one of the expected input (file) isn't present as expected. This will fix silent issues like the recent https://github.com/facebook/zstd/issues/4318 reported by @sebres
- update `zlib` dependency to `v1.3.1`

### DIFF
--- a/lib/dll/example/build_package.bat
+++ b/lib/dll/example/build_package.bat
@@ -21,12 +21,16 @@ call :copyFile "lib\dll\example\fullbench-dll.*" "bin\example\"
 call :copyFile "lib\dll\example\README.md" "bin\"
 call :copyFile "lib\zstd.h" "bin\include\"
 call :copyFile "lib\zstd_errors.h" "bin\include\"
-call :copyFile "lib\dictBuilder\zdict.h" "bin\include\"
+call :copyFile "lib\zdict.h" "bin\include\"
 call :copyFile "programs\zstd.exe" "bin\zstd.exe"
 
 endlocal
-exit /b
+exit /b 0
 
 :copyFile
-copy "%~1" "%~2" || (echo Failure processing "%~1" & exit /b %errorlevel%)
-goto :eof
+copy "%~1" "%~2"
+if errorlevel 1 (
+    echo Failed to copy "%~1"
+    exit 1
+)
+exit /b
--- a/lib/dll/example/build_package.bat
+++ b/lib/dll/example/build_package.bat
@@ -1,31 +1,32 @@
-@ECHO OFF
-MKDIR bin\dll bin\static bin\example bin\include
-SET CpyError=
-COPY tests\fullbench.c bin\example\ || (SET CpyError=%CpyError% tests\fullbench.c)
-COPY programs\datagen.c bin\example\ || (SET CpyError=%CpyError% programs\datagen.c)
-COPY programs\datagen.h bin\example\ || (SET CpyError=%CpyError% programs\datagen.h)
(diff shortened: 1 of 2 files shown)


Cyan4973:
And of course, there is a new error with `meson-windows` that starts to appears in this PR. great.

_edit_ : fixed by rebasing on top of #4322

lib/dll/example/build_package.bat
```diff
@@ -1,30 +1,32 @@
...
+call :copyFile "lib\dll\example\fullbench-dll.*" "bin\example\"
+call :copyFile "lib\dll\example\README.md" "bin\"
+call :copyFile "lib\zstd.h" "bin\include\"
+call :copyFile "lib\zstd_errors.h" "bin\include\"
+call :copyFile "lib\dictBuilder\zdict.h" "bin\include\"
+call :copyFile "programs\zstd.exe" "bin\zstd.exe"
 
-:EOF
+endlocal
+exit /b
```
sebres: I don't think it is OK - this line would exit without error in any case.

Maybe something like that instead:
```batch
@echo off
setlocal

rem Create required directories.
mkdir bin\dll bin\static bin\example bin\include

rem Copy files using a subroutine. Exits immediately on failure.
set errCounts=0
call :copyFile "tests\fullbench.c" "bin\example\"
call :copyFile "programs\datagen.c" "bin\example\"
call :copyFile "programs\datagen.h" "bin\example\"
call :copyFile "programs\util.h" "bin\example\"
call :copyFile "programs\platform.h" "bin\example\"
call :copyFile "lib\common\mem.h" "bin\example\"
call :copyFile "lib\common\zstd_internal.h" "bin\example\"
call :copyFile "lib\common\error_private.h" "bin\example\"
call :copyFile "lib\common\xxhash.h" "bin\example\"
call :copyFile "lib\libzstd.a" "bin\static\libzstd_static.lib"
call :copyFile "lib\dll\libzstd.*" "bin\dll\"
call :copyFile "lib\dll\example\Makefile" "bin\example\"
call :copyFile "lib\dll\example\fullbench-dll.*" "bin\example\"
call :copyFile "lib\dll\example\README.md" "bin\"
call :copyFile "lib\zstd.h" "bin\include\"
call :copyFile "lib\zstd_errors.h" "bin\include\"
call :copyFile "lib\dictBuilder\zdict.h" "bin\include\"
call :copyFile "programs\zstd.exe" "bin\zstd.exe"

if "%errCounts%" == "0" goto :eof
echo Filed to process %errCounts% entries.
endlocal
exit /b 1

:copyFile
copy "%~1" "%~2" || (set /A errCounts=errCounts+1 && echo ^    failed processing "%~1" & exit /b %errorlevel%)
goto :eof
```

Cyan4973:
The script has been updated, so that it fails explicitly when one copy operation fails.
This behavior has been tested successfully in [this branch](https://github.com/Cyan4973/zstd/actions/runs/13577070391/job/37955642038).

Btw, there _was_ an error in the list, with `zdict.h` being incorrectly set. Fixed it in this PR too.

terrelln (approved):

Cyan4973:
After this PR is merged,
new Windows artifact should be generated and published automatically when a new release is created.

Also used this opportunity to update the Windows artifacts on the `v1.5.7` page, using this new script.
They feature all 3 `*.h` include files,
and the top directory features the expected name "zstd-v1.5.7-win64".

PR #4329, opened 2025-03-04
New commands --split and --jobsize

Cyan4973:
This is just a consolidation.

Technically, these capabilities were already provided by command `-B#`,
but it's not a very clear command, it's overloaded and mean different things in different contexts,
and it evokes "Block", which is confusing because "Block" is an entirely different concept in Zstandard.

So change for new command names, `--split=#` and `--jobsize=#`, which are more descriptive of their action.

Previous command `-B#` is still supported, to remain compatible with existing scripts,
but it's no longer documented, in order to discourage its usage.


terrelln (approved):
Nice cleanup!

PR #4331, opened 2025-03-06
[test] Add tests for determinism

### LINKED ISSUE #4330: [reproducibility] Add tests that validate reproducibility across hardware platforms
We have some determinism fuzzers that check for deterministic output within a single platform:

https://github.com/facebook/zstd/blob/99cf130cfcdce3119fb39809c9935a686be960e4/tests/fuzz/simple_round_trip.c#L68-L77

https://github.com/facebook/zstd/blob/99cf130cfcdce3119fb39809c9935a686be960e4/tests/fuzz/dictionary_round_trip.c#L47-L58

But we don't have any tests that check for determinism across hardware platforms. Lets check in some simple tests under `tests/cli-tests/` that validate we produce a pre-determined set of compressed checksums. These tests run in all of our different hardware platforms in CI.

terrelln:
* Add a function `ZSTD_isDeterministicBuild()` to `zstd.h`, and print a message for non-deterministic builds in `zstd -vv --version`. When it returns true, the build isn't using any compilation flags that may impact the compressed output.
* Run determinism tests in cli-tests when we detect that the build supports determinism.
* Run CLI tests in `dev-short-tests` when running `make check` on the different platforms.

Fixes #4330

To regenerate the checksums run this command, which completes in under two minutes.

```
make -C tests update-cli-tests
```


tests/cli-tests/determinism/multi.sh
```diff
@@ -0,0 +1,22 @@
...
+set -e
+
+datagen -g0 > file0
+datagen -g1 > file1
+datagen -g1000 > file1000
+datagen -g100000 > file100000
+
+for level in $(seq 1 19); do
+    echo "$level"
+    zstd -f -$level file0 file1 file1000 file100000
```
Cyan4973: Why is the same test repeated 4 times ?

terrelln: Different ordering of the files

Cyan4973: OK, so this is checking the absence of impact on the compressed output when compressing multiple payloads using the same context.

terrelln: Yeah, I'll add a comment here. It is a very basic test.

terrelln:
I've given up on getting the CLI tests to run in MinGW for now. It is likely doable, but its a bit more effort that I'm willing to spend right now. But we're still running in all the QEMU tests.

Cyan4973 (approved):
The `QEMU` tests look the more important ones for this new test.

PR #4334, opened 2025-03-10
fix #4332: setting ZSTD_NBTHREADS=0 via environment variable

### LINKED ISSUE #4332: ZSTD_NBTHREADS=0 from environment not multithreaded since v1.5.7
**Describe the bug**
Since v1.5.7 setting `ZSTD_NBTHREADS=0` from the environment no longer allows for automatically setting the number of threads.

This was introduced in https://github.com/facebook/zstd/pull/4211/
- https://github.com/facebook/zstd/pull/4211/files#diff-882b9498850a4b4291d6f92d586c46a2ca296d33688a3a1837068cc29060fa5eR857-L858
- https://github.com/facebook/zstd/pull/4211/files#diff-882b9498850a4b4291d6f92d586c46a2ca296d33688a3a1837068cc29060fa5eR1299-R1317

I attached a fairly trivial patch that moves the block for core detection.
It is tested and works, but I have not looked too thoroughly at it yet so it might be done better.

[patch.txt](https://github.com/user-attachments/files/19142860/patch.txt)

**To Reproduce**
```shell
ZSTD_NBTHREADS=0 zstd -vv <in> -o out.zst
*** Zstandard CLI (64-bit) v1.5.7, by Yann Collet ***
Compressing with 0 worker threads 
```

**Expected behavior**
Using ZSTD_NBTHREADS=0 from the environment should set the number of threads based on the number of logical or physical cores as before.

Before v1.5.7 or with patch applied:
```shell
ZSTD_NBTHREADS=0 ./zstd -vv <in> -o out.zst
Note: 12 physical core(s) detected 
Compressing with 12 worker threads 
```

Cyan4973:
Since `v1.5.7`, setting `ZSTD_NBTHREADS=0` via environment variable doesn't work, due to a logic mistake in the initialization routine.
Reported in #4332, by [JcKlomp](https://github.com/JcKlomp).

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -44,7 +44,7 @@
 #endif
 
 #ifndef ZSTDCLI_NBTHREADS_DEFAULT
-#define ZSTDCLI_NBTHREADS_DEFAULT MAX(1, MIN(4, UTIL_countLogicalCores() / 4))
+#define ZSTDCLI_NBTHREADS_DEFAULT (unsigned)(MAX(1, MIN(4, UTIL_countLogicalCores() / 4)))
 #endif
 
 
@@ -94,6 +94,7 @@ static U32 g_ldmBucketSizeLog = LDM_PARAM_DEFAULT;
 
 
 #define DEFAULT_ACCEL 1
+#define NBWORKERS_AUTOCPU 0
 
 typedef enum { cover, fastCover, legacy } dictType;
 
@@ -745,7 +746,7 @@ static void printActualCParams(const char* filename, const char* dictFileName, i
 
 /* Environment variables for parameter setting */
 #define ENV_CLEVEL "ZSTD_CLEVEL"
-#define ENV_NBTHREADS "ZSTD_NBTHREADS"    /* takes lower precedence than directly specifying -T# in the CLI */
+#define ENV_NBWORKERS "ZSTD_NBTHREADS"    /* takes lower precedence than directly specifying -T# in the CLI */
 
 /* pick up environment variable */
 static int init_cLevel(void) {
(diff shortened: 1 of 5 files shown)


tests/cli-tests/compression/multi-threaded.sh
```diff
@@ -10,7 +10,3 @@ zstd -T0 -f file -q                     ; zstd -t file.zst
 zstd -T0 --auto-threads=logical -f file -q ; zstd -t file.zst
 zstd -T0 --auto-threads=physical -f file -q ; zstd -t file.zst
 zstd -T0 --jobsize=1M -f file -q        ; zstd -t file.zst
```
terrelln, line 12: Should we add a test here that tests the environment variable works?

If we print the number of threads under some verbosity we could do something like:

```
ZSTD_NBTHREADS=0 zstd -f file -vv 2>&1 | grep "physical core(s) detected"
zstd -tq file.zst
ZSTD_NBTHREADS=4 zstd -f file -vv 2>&1 | grep "4 threads"
zstd -tq file.zst
```

programs/zstdcli.c
```diff
@@ -1332,27 +1335,16 @@ int main(int argCount, const char* argv[])
     DISPLAYLEVEL(3, WELCOME_MESSAGE);
 
 #ifdef ZSTD_MULTITHREAD
-    if ((operation==zom_decompress) && (nbWorkers > 1)) {
-        DISPLAYLEVEL(2, "Warning : decompression does not support multi-threading\n");
-    }
```
terrelln, line 1337: This was added due to Issue #2918. This may be a useful warning to keep.

terrelln (approved):

tests/cli-tests/compression/multi-threaded.sh
Cyan4973: Yes, definitely,
I was considering to do it in another PR, in order to observe the test failure under current conditions, i.e. without the patch.
But I could also just add it to this PR as a follow up commit.

programs/zstdcli.c
Cyan4973: Issue is, with current code,
there is nothing to tell that `nbWorkers` wasn't set.
Whether it's set via the environment variable, or via the core count detection, it's necessarily set to "some value".
There is no difference with a user setting this value manually via command line.

Therefore, on any modern system with >= 8 logical cores, `nbWorkers >= 2`, and consequently _always_ triggers this warning.

This could be solved by adding yet another variable, which would track if `nbWorkers` was set explicitly, or if it happens to be set to a value `>1` just as a consequence of automatic core count detection. We would also have to decide if setting `nbWorkers` via environment variable count or not to trigger this warning.

So it's a bit more work, a bit more questions, and specific to the topic of displaying a warning during decompression mode. It felt it could deserve its own PR for proper focus.

Cyan4973:
test added in this PR,
checked manually on `dev` branch (where it produces an error signal, as expected).

Cyan4973:
In the process, fixed a latent limitation of the test shell script `playTests.sh`,
where was unable to pass environment variables (such as `ZSTD_NBTHREADS`) to `zstd`
specifically in FreeBSD environment (and possibly other *BSD).

PR #4336, opened 2025-03-11
fix #4312 - musl compilation compatibility

Cyan4973:
#4312 tried to implement a manual workaround for `musl` compilation issues.
It's manual because `musl` does not provide a version number, making it difficult to automate decisions.

Unfortunately, the fix was incomplete, and the test did not catch it because it only detected a "warning" and had no runtime error.
Fixed both the test, which will now fail both at run time and compile time under similar conditions, 
and the code, which now passes the more stringent test.

Used the opportunity to restructure the code, so that the dispatch strategy among the different variants of `qsort*()` is more maintainable and extendable.
Employed the new framework to add support for `C11` Annex K `qsort_s()` re-entrant variant.

also: 
Reduced the duration of long CLI tests, particularly dictionary building tests, by utilizing multithreading with the `-T0` flag. This optimization significantly decreases the execution time of the longest CI jobs, enhancing overall CI efficiency and throughput.

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -364,14 +364,6 @@ static void stableSort(COVER_ctx_t *ctx)
     qsort_s(ctx->suffix, ctx->suffixSize, sizeof(U32),
             (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp),
             ctx);
-#elif defined(__OpenBSD__)
-    /* On OpenBSD, qsort() is not guaranteed to be stable, their mergesort() is.
-     * Note(@cyan): qsort() is never guaranteed to be stable,
-     *              so why would this property only matter for OpenBSD ?
-     */
-    g_coverCtx = ctx;
-    mergesort(ctx->suffix, ctx->suffixSize, sizeof(U32),
-          (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp));
 #else /* C90 fallback.*/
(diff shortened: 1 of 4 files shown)


Cyan4973:
And indeed, it helps a bit:
the longest test, `clang-msan-testzstd` went down, from 28mn to 19mn.
So this will help improving CI velocity.

lib/dictBuilder/cover.c
```diff
@@ -317,20 +345,30 @@ static int COVER_strict_cmp8(const void *lp, const void *rp) {
...
+            ctx);
+#elif (ZDICT_QSORT == ZDICT_QSORT_C11)
     qsort_s(ctx->suffix, ctx->suffixSize, sizeof(U32),
             (ctx->d <= 8 ? &COVER_strict_cmp8 : &COVER_strict_cmp),
             ctx);
 #elif defined(__OpenBSD__)
+    /* On OpenBSD, qsort() is not guaranteed to be stable, their mergesort() is.
+     * Note(@cyan): qsort() is never guaranteed to be stable,
+     *              so why would this property only matter for OpenBSD ?
+     */
```
terrelln: I think we can drop this. This was added before we switched to use `COVER_strict_cmp*` which enforces a strict ordering on the array.

terrelln (approved):

iucoen:
g_coverCtx can be declared as thread local so this code will at least be thread-safe on older versions of musl

Cyan4973:
In `C11` yes, but not in `C90`.

PR #4341, opened 2025-03-20
lib/zstd.h: don't put pragma after static

clan:
otherwise will cause dev-python/zstandard build failed when compiling with clang as reported at https://bugs.gentoo.org/950259

the root cause is pycparser, which is unfixed since reported 2.5 years ago, :(


Cyan4973 (approved):

thesamesam:
Thank you @Cyan4973, it's appreciated.

PR #4343, opened 2025-03-21
add library.json file to support esp32 target via Platformio

mhaberler:
usage example: https://github.com/mhaberler/m5stick-mqtt/blob/qrcode/platformio.ini

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,50 @@
+{
+    "name": "zstd",
+    "version": "1.5.5",
+    "description": "Zstandard - Fast real-time compression algorithm",
+    "keywords": "compression, zstd, zstandard, decompression, lossless",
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thank you for contributing, @mhaberler. Your solution likely works as intended. However, the upstream repository isn’t the place to support every package manager. We aim to keep the repository’s root directory free of special-case files and only add features we can reliably maintain over time.

In this case, a dedicated fork for PlatformIO would be ideal. That fork would include and maintain the additional file, updating it whenever a new release occurs. (For instance, the “version” field is still set to `v1.5.5`.)

mhaberler:
fair enough. 

I will keep it in my fork in the platformio branch and rebase as needed.

PR #4353, opened 2025-03-28
Check regular file for sparse support after opening

afq984:
The doc for `--sparse` says "default:  enabled when output is into a file", 
however it only works when the output file already exists.

A regular file may be created by the open call.
Checking after opening allows sparseFileSupport even if dstFileName does not already exist.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -630,6 +630,7 @@ FIO_openDstFile(FIO_ctx_t* fCtx, FIO_prefs_t* const prefs,
     }
 
     {
+        int isDstRegFile;
 #if defined(_WIN32)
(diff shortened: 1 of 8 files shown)


Cyan4973:
Great fix,

it also shows that we lost this capability at some point
due to an inability to test it in CI.

We should consider writing a test specifically validating sparse file support.

Cyan4973 (approved):

afq984:
@Cyan4973  Shall I add the test in this PR or is it OK as a follow up?

Cyan4973:
> @Cyan4973 Shall I add the test in this PR or is it OK as a follow up?

It is fine as a follow up

PR #4367, opened 2025-04-17
Add unwind information in huf_decompress_amd64.S

al13n321:
These two asm functions clobber `rbp` register (aka omit frame pointer), but didn't provide stack unwinding information. So stack traces in profilers and debuggers couldn't go past these functions. This PR adds the unwind information, in a straightforward way.

I tested it with one profiler (ClickHouse's built-in query profiler, uses libunwind) and one debugger (https://github.com/al13n321/nnd), both managed to unwind through this function with this fix.

### DIFF
--- a/lib/decompress/huf_decompress_amd64.S
+++ b/lib/decompress/huf_decompress_amd64.S
@@ -117,22 +117,55 @@ ZSTD_HIDE_ASM_FUNCTION(_HUF_decompress4X1_usingDTable_internal_fast_asm_loop)
 _HUF_decompress4X1_usingDTable_internal_fast_asm_loop:
 HUF_decompress4X1_usingDTable_internal_fast_asm_loop:
     ZSTD_CET_ENDBRANCH
+    .cfi_startproc
+    .cfi_def_cfa_offset 8
+    .cfi_offset %rip, -8
     /* Save all registers - even if they are callee saved for simplicity. */
     push %rax
+    .cfi_def_cfa_offset 16
+    .cfi_offset rax, -16
     push %rbx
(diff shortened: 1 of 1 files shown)


al13n321:
Example of profiler output before the fix (truncated stack traces in the left half):
before

After the fix (no-longer-truncated zstd stack traces in the right half):
after

al13n321:
Btw, I'm confused: you guys at Meta use profilers (I used to use strobelight a lot), and zstd decompression probably takes a good fraction of cpu time in a lot of Meta software. Surely many people would notice if stack unwinding doesn't work well on zstd decompression? Why wasn't this already fixed ages ago? I'm probably missing something.

Cyan4973:
cc @terrelln

embg:
I am a big fan of this change. @MatzeB and I were just chatting about this issue a few weeks ago. This would be a great improvement IMO.

MatzeB:
I mentioned a while ago that we need frame pointers (aka rbp) to be setup for the profilers in our company. For all I know our profiling tools don't interpret unwind tables for performance reasons, so this change in itself won't help our setup. Of course adding proper unwind table information is a great idea regardless.

Cyan4973:
Sounds good @embg !
I wanted @terrelln to have a look at it, because it changes the assembly source code.

Also there were a few test failures blocking this merge,
but one of them is unclear (it might just be a CI issue),
and the other one seems to happen in a different file, which has likely been fixed since.
I suspect a rebase of this PR on latest `dev` branch would be enough to clean these test errors.

al13n321:
Rebased.

Cyan4973 (approved):

PR #4371, opened 2025-04-18
Fix duplicate LC_RPATH error on MacOS

### LINKED ISSUE #4369: Duplicate LC_RPATH error on MacOS>=15.4
**Describe the bug**
When providing a rpath parameter during the build, the library cannot be loaded on MacOS>= 15.4.

**To Reproduce**
1. Build libzstd and provide some rpath parameters in LDFLAGS
2. Try to use the binary on MacOS version 15.4
3. The library fails to load with this error about duplicate LC_RPATH:

```
dyld[29361]: Library not loaded: @loader_path/../lib/libzstd.1.dylib
      Referenced from: <7131C877-3CF0-33AC-AA05-257BA4FDD770> /Users/foobar/...
      Reason: tried: '/Users/foobar/..../lib/libzstd.1.dylib' (duplicate LC_RPATH '/usr/mypath.../lib')
```

**Expected behavior**

The library should be loaded.

**Screenshots and charts**

 - OS: MacOS version >=15.4 (it was working before the update, but the behavior of the dynamic loader changed in the 15.4 update)
 - Version: tested with release branch, but basically impacts all versions which were shipped on MacOS previously.
 - Compiler: GCC
 - Flags: Wl,-rpath,/usr/foobar/lib
 - Other relevant hardware specs [e.g. Dual-core]
 - Build system: Makefile

**Additional context**
I'm working on a fix. There are other projects impacted like e.g. https://github.com/conda/conda-build/issues/5671 
This regression is caused by this [code](https://github.com/apple-oss-distributions/dyld/blob/fba6732ffad1c798f31b2129eee4e073e6ae07dc/mach_o/Policy.cpp#L354) in dyld which checks whether the sdk version is newer than 2024.

ecordonnier-sc:
After the update to MacOS 15.4, the dynamic loader dyld treats duplicated LC_RPATH as an [error](https://github.com/apple-oss-distributions/dyld/blob/fba6732ffad1c798f31b2129eee4e073e6ae07dc/mach_o/Policy.cpp#L354). The `FLAGS` variable already contains `LDFLAGS`, thus using both `FLAGS` and `LDFLAGS` duplicates all `LDFLAGS`, including `-Wl,rpath` parameters.

The duplicate LC_RPATH causes this kind of errors:

```
dyld[29361]: Library not loaded: @loader_path/../lib/libzstd.1.dylib
      Referenced from: <7131C877-3CF0-33AC-AA05-257BA4FDD770> /Users/foobar/...
      Reason: tried: '/Users/foobar/..../lib/libzstd.1.dylib' (duplicate LC_RPATH '/usr/mypath.../lib')
```

Closes https://github.com/facebook/zstd/issues/4369

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -166,7 +166,7 @@ $(ZSTD_DYNLIB): $(ZSTD_DYNLIB_OBJ)
 	$(if $(filter -DZSTD_MULTITHREAD,$(CPPFLAGS)),\
     @echo compiling multi-threaded dynamic library $(LIBVER),\
     @echo compiling single-threaded dynamic library $(LIBVER))
-	$(CC) $(FLAGS) $^ $(LDFLAGS) $(SONAME_FLAGS) -o $@
+	$(CC) $(FLAGS) $^ $(SONAME_FLAGS) -o $@
 	@echo creating versioned links
 	ln -sf $@ libzstd.$(SHARED_EXT_MAJOR)
 	ln -sf $@ libzstd.$(SHARED_EXT)
@@ -255,7 +255,7 @@ libzstd-nomt: $(ZSTD_NOMT_FILES)
         echo "Error: Found zstdmt in list."; \
         exit 1; \
     fi
-	$(CC) $(FLAGS) $^ $(LDFLAGS) $(SONAME_FLAGS) -o $@
+	$(CC) $(FLAGS) $^ $(SONAME_FLAGS) -o $@
 
 .PHONY: clean
 clean:


ecordonnier-sc:
@Cyan4973 can you please take a look?

Cyan4973 (approved):
The general idea is sound:
since `LDFLAGS` is already included into `FLAGS`, there is no need to repeat it.

A minor difference is that `LDFLAGS` appears _before_ `$^`, aka the list of units, instead of _after_.
It probably doesn't matter. It might make a difference for `LDLIBS`, but not `LDFLAGS`.

So, the last remaining issue is the CLA, and I see it has also been fixed last week.
So it is good to go now.

PR #4377, opened 2025-04-28
Add License variable to pkg-config file

iwamatsu:
The pkg-config file has License variable that allows you to set the license for the software. 
This sets 'BSD-3-Clause OR GPL-2.0-only' to License.

Ref: https://github.com/pkgconf/pkgconf/blob/master/man/pc.5#L116


Cyan4973 (approved):

Cyan4973:
Note: the test error observed on this PR are not related to the PR itself
and have been fixed in `dev` branch.
They will disappear once this branch is rebased on top of latest `dev`.

PR #4378, opened 2025-04-29
fix minor warning in zstreamtest

Cyan4973:
```
zstreamtest.c:1899:32: error: initializer-string for array of 'char' truncates NUL terminator but destination lacks 'nonstring' attribute (7 chars into 6 available) [-Werror=unterminated-string-initialization]
 1899 |           char testBuffer[6] = "\xAA\xAA\xAA\xAA\xAA\xAA";
      |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

It's indeed a (minor) mistake. Not sure why it only flags now, maybe only latest version of `gcc` are able to detect it.


felixhandte:
Yeah, it's new in GCC 15: https://lwn.net/SubscriberLink/1018486/1dcd29863655cb25/

Issue #4368, opened 2025-04-17
Possible division by zero in zstdcli_trace.c


Aruise:
Hi,

In function [TRACE_log](https://github.com/facebook/zstd/blob/d654fca78690fa15cceb8058ac47454d914a0e63/programs/zstdcli_trace.c#L80) 'duration' is received from function [ZSTD_trace_compress_end](https://github.com/facebook/zstd/blob/d654fca78690fa15cceb8058ac47454d914a0e63/programs/zstdcli_trace.c#L135), which states that it could be zero

Which leads to a possible division by zero bug
```
void ZSTD_trace_compress_end(ZSTD_TraceCtx ctx, ZSTD_Trace const* trace)
{
    PTime const beginNanos = (PTime)ctx;
    PTime const endNanos = UTIL_clockSpanNano(g_enableTime);
    PTime const durationNanos = endNanos > beginNanos ? endNanos - beginNanos : 0;
    assert(g_traceFile != NULL);
    assert(trace->version == ZSTD_VERSION_NUMBER); /* CLI version must match. */
    TRACE_log("compress", durationNanos, trace);
}
```

```PTime const durationNanos = endNanos > beginNanos ? endNanos - beginNanos : 0;``` states that 'durationNanos' will be either 'endNanos - beginNanos', if 'endNanos' is bigger, or zero otherwise

Next this function calls ```TRACE_log("compress", durationNanos, trace);```

In TRACE_log, further, it passed to a equation, where possibly division by zero happens

```
static void TRACE_log(char const* method, PTime duration, ZSTD_Trace const* trace)
{
    int level = 0;
    int workers = 0;
    double const ratio = (double)trace->uncompressedSize / (double)trace->compressedSize;
    double const speed = ((double)trace->uncompressedSize * 1000) / (double)duration;
    if (trace->params) {
...
```
```double const speed = ((double)trace->uncompressedSize * 1000) / (double)duration;``` here, if 'durationNanos' was zero, will be a division by zero

Same with [ZSTD_trace_decompress_end](https://github.com/facebook/zstd/blob/d654fca78690fa15cceb8058ac47454d914a0e63/programs/zstdcli_trace.c#L153)

terrelln:
Thanks for the report @Aruise! Looks like a bug. It will only be triggered if the user passes `--trace` on the command line, which I don't expect many people are doing, so the impact should be quite limited.

Mrigankkh:
[...]

PR #4379, opened 2025-05-01
Bug Fix: Fixed potential Divide by Zero in zstdcli_trace.c

Mrigankkh:
Bug Fix for Issue #4368 
 

In the TRACE_log function in zstdcli_trace.c, in case duration is 0, speed is set to 0, circumventing the potential divide by 0 error.

### DIFF
--- a/programs/zstdcli_trace.c
+++ b/programs/zstdcli_trace.c
@@ -82,7 +82,9 @@ static void TRACE_log(char const* method, PTime duration, ZSTD_Trace const* trac
     int level = 0;
     int workers = 0;
     double const ratio = (double)trace->uncompressedSize / (double)trace->compressedSize;
-    double const speed = ((double)trace->uncompressedSize * 1000) / (double)duration;
+    const double speed = duration != 0
+        ? ((double)trace->uncompressedSize * 1000) / (double)duration
+        : 0.0;
     if (trace->params) {
         ZSTD_CCtxParams_getParameter(trace->params, ZSTD_c_compressionLevel, &level);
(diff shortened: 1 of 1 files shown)


programs/zstdcli_trace.c
```diff
@@ -82,7 +82,9 @@ static void TRACE_log(char const* method, PTime duration, ZSTD_Trace const* trac
     int level = 0;
     int workers = 0;
     double const ratio = (double)trace->uncompressedSize / (double)trace->compressedSize;
-    double const speed = ((double)trace->uncompressedSize * 1000) / (double)duration;
+    const double speed = duration != 0
+        ? ((double)trace->uncompressedSize * 1000) / (double)duration
+        : 0.0;
```
sebres, line 87: Zero speed? Shall not it be INFINITY either?..

However, since duration is a precise time  counter (and it's nanosecond precise), I don't think the duration as diff between 2 measurements could be ever 0, because would mean few CPU instructions.
But to be sure (and safe in the future), I'd suggest simply something under its min-value (e. g. small fraction like `.1` or `1e-4`):
```diff
- double const speed = ((double)trace->uncompressedSize * 1000) / (double)duration;
+ double const speed = ((double)trace->uncompressedSize * 1000) / (duration ? (double)duration : .1);
```

Cyan4973: I like this suggestion

sebres:
Just for the record - fixed in ae64545c6b6257c226dca6240be7b2d345d9bee5 (#4464).

PR #4382, opened 2025-05-07
Improve seekable_format/examples/parallel_compression

### LINKED ISSUE #3980: seekable_format/examples/parallel_compression.c is not parallel
**Describe the bug**
The parallel compression example does not seem to be working in parallel.
It runs on a single thread only, regardless of how many threads are requested.

**To Reproduce**
1. Build the example parallel_compression
2. Run:
`./parallel_compression data.txt  $((2*1024*1024)) 4`
data.txt can be any sufficiently large file so that compression takes a handful of seconds.
In my case, this is ASCII data, 2.8 GB.
3. While the compression is running, look at e.g. top
4. Just a single thread is used. This is usually slower than the single threaded seekable_compression,

**Expected behavior**
Parallel execution should use multiple cores and be faster than single threaded.
The full `zstd` executable works as expected; given the -T option it is faster than without and saturates all cores.

Also, I'd love for --seekable to become a standard option for `zstd`, but that should probably be a separate issue.

**Desktop:**
 - OS: Linux
 - Version Ubuntu 22.04
 - Compiler: gcc 11.4
 - Flags: -O3
 - System: Quad-core Intel Core i7-1165G7
 - Build system Makefile
- Built from git version:
commit 559762da12f54712d44f619098aa4a7e7bc5727b, Mar 14 2024

vasi:
Change the design of parallel_compression, to allow it to actually work in parallel without blowing up memory usage. Fixes #3980.

In the new approach, we don't wait for all jobs to be accepted by the queue before writing any results. Instead, as each job finishes, it adds its result to an ordered-linked-list, and then writes (and frees) as many results as are available. We use a mutex for synchronization, so this should no longer be racy. This approach also conveniently allows streaming input, since we no longer need to know the number of frames ahead of time.

An added test enforces that parallel_compression doesn't use excessive memory.

This doesn't change anything about the main zstd program, so we should leave #3980 open, or open a new issue about zstd CLI seekability.

A variety of other fixes were necessary to get test passing, or were obvious cleanups once this was complete. Each is a separate commit.

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -170,7 +170,6 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # tag=v4.2.2
     - name: LZ4, thread pool, and libs build testslib wrapper test
       run: |
-        make lz4install
         make -C tests test-lz4
         make check < /dev/null | tee   # mess with lz4 console detection
         make clean
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,7 @@ apt-add-repo:
 	sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
 	sudo apt-get update -y -qq
 
-.PHONY: ppcinstall arminstall valgrindinstall libc6install gcc6install gcc7install gcc8install gpp6install clang38install lz4install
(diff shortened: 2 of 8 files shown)


Cyan4973 (approved):
Thanks @vasi !
Looks good to me !
It's also pretty clean, very readable, style is consistent, 
and the modification comes with a test run in CI.
I like it.

PR #4384, opened 2025-05-09
update `--rm` cmd help info

xiaoge1001:
Starting from cee6bec9fa6aa249f2df9f84165b682eb793eab, --rm is ignored when the output is `stdout`.

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -147,7 +147,7 @@ static void usage(FILE* f, const char* programName)
     DISPLAY_F(f, "Options:\n");
     DISPLAY_F(f, "  -o OUTPUT                     Write output to a single file, OUTPUT.\n");
     DISPLAY_F(f, "  -k, --keep                    Preserve INPUT file(s). [Default] \n");
-    DISPLAY_F(f, "  --rm                          Remove INPUT file(s) after successful (de)compression.\n");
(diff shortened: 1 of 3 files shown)


programs/zstdcli.c
```diff
@@ -147,7 +147,7 @@ static void usage(FILE* f, const char* programName)
     DISPLAY_F(f, "Options:\n");
     DISPLAY_F(f, "  -o OUTPUT                     Write output to a single file, OUTPUT.\n");
     DISPLAY_F(f, "  -k, --keep                    Preserve INPUT file(s). [Default] \n");
-    DISPLAY_F(f, "  --rm                          Remove INPUT file(s) after successful (de)compression.\n");
+    DISPLAY_F(f, "  --rm                          Remove INPUT file(s) after successful (de)compression. Ignoring `--rm` command when the output is `stdout`.\n");
```
Cyan4973: For `zstd -h`, we try to keep displayed lines short, for a good terminal experience.
So I would try to shorten this one if possible, for example: 
`Remove INPUT file(s) after successful (de)compression to file.`

Cyan4973 (changes requested):

Cyan4973 (approved):

programs/zstdcli.c
xiaoge1001: Thanks, I have already made the necessary modifications.

PR #4392, opened 2025-05-25
Fix potential memory leak in function `benchMem`

mugitya03:
`speedPerRound` is allocated at the start of function `benchMem` to collect per-round speeds, but is never freed, causing a potential memory leak.
```c
double *const speedPerRound = (double *)malloc(nbRounds * sizeof(double));
``` 

We add `free(speedPerRound);` after aggregating the data to release the heap allocation and prevent memory buildup.


Cyan4973 (approved):
Nice catch !

PR #4394, opened 2025-05-26
Remove redundant setting of allJobsCompleted to 1

SiliconA-Z:
This will do it automatically.

### DIFF
--- a/lib/compress/zstdmt_compress.c
+++ b/lib/compress/zstdmt_compress.c
@@ -1268,8 +1268,7 @@ size_t ZSTDMT_initCStream_internal(
 
     if (mtctx->allJobsCompleted == 0) {   /* previous compression not correctly finished */
         ZSTDMT_waitForAllJobsCompleted(mtctx);
-        ZSTDMT_releaseAllJobResources(mtctx);
-        mtctx->allJobsCompleted = 1;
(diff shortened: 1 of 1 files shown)


lib/compress/zstdmt_compress.c
```diff
@@ -1268,8 +1268,7 @@ size_t ZSTDMT_initCStream_internal(
 
     if (mtctx->allJobsCompleted == 0) {   /* previous compression not correctly finished */
         ZSTDMT_waitForAllJobsCompleted(mtctx);
-        ZSTDMT_releaseAllJobResources(mtctx);
-        mtctx->allJobsCompleted = 1;
+        ZSTDMT_releaseAllJobResources(mtctx); // Will set allJobsCompleted to 1
```
Cyan4973: Looks good.

One minor issue:
this project requires C90 compatibility, at least for the core library,
which means `/* .... */` comment style is the only possible one.

Cyan4973 (changes requested):

lib/compress/zstdmt_compress.c
SiliconA-Z: Done!

Cyan4973 (approved):

PR #4397, opened 2025-05-31
Fix several locations with potential memory leak

xiaoge1001:
Fix several locations with potential memory leak


Cyan4973 (approved):

Cyan4973:
Seems like we have a conflict after merging another patch modifying the same file.

xiaoge1001:
> Seems like we have a conflict after merging another patch modifying the same file.

Conflicts with the following PR:
https://github.com/facebook/zstd/pull/4401

It also solved the memory leak problem. I will cancel the changes to contrib/externalSequenceProducer/main.c.

PR #4401, opened 2025-06-03
Release resources before returning

mugitya03:
In main, resources were freed on the success path but not in the error path. This change ensures all allocated resources are released before returning.


Cyan4973 (approved):
I understand that it's considered "proper" to clean all resources on exiting the function.
To be fair, since it's `main()`, it's leaving program execution immediately on `return()`. Hence all resources are going to be released anyway. So the benefit seems moot.

That being said, the complexity introduced is also quite low, so it's a reasonable trade off.

PR #4402, opened 2025-06-04
Release resources in error paths via cleanup

mugitya03:
Replace direct returns in error-handling branches with a unified cleanup block that frees allocated resources before returning, improving code quality and robustness.


Cyan4973 (approved):

Cyan4973:
Unfortunately, merge conflicts have emerged since the review. 
Would you be able to fix them ?

tests/bigdict.c
```diff
@@ -0,0 +1,132 @@
+/*
```
Cyan4973: This file simply shouldn't exist anymore.
It's a duplicate.

Cyan4973 (changes requested):

tests/bigdict.c
mugitya03: Yes. I just deleted it.

mugitya03: I tried to squash these three commits locally, but it seems that squashing a merge commit is not straightforward.

Cyan4973: > I tried to squash these three commits locally, but it seems that squashing a merge commit is not straightforward.

This can be done at merge time, via the github UI.

tests/largeDictionary.c
```diff
@@ -70,61 +70,39 @@ int main(int argc, const char** argv)
     char* buffer = (char*)malloc(bufferSize);
     void* out = malloc(outSize);
     void* roundtrip = malloc(dataSize);
+    int _exit_code = 1;
     (void)argc;
     (void)argv;
     int _exit_code = 0;
```
Cyan4973:
```
int _exit_code = 1;
(...)
int _exit_code = 0;
```

Declaring the same variable twice ?
This feels like it shouldn't work ?
I'm surprised it even compiles.

Even if it does, it looks to me that this is not the wanted behavior: due to overloading, `_exit_code` value would always be `0`.

mugitya03: Oh, thanks for pointing that out. This was duplicated during the merge—I’ll remove the second declaration.

Cyan4973: That being said, I still don't understand why CI did not turn red.
I would have expected this unit to fail compilation. But CI is green.
Which leaves use with a few options:
- It can actually compile and is legal (which I doubted, so that would be new learning for me)
- It fails, but the CI does not detect it and turns green anyway (bad)
- The file is never compiled during CI (bad, should be fixed)

Cyan4973: tested locally:
```
largeDictionary.c:76:9: error: redefinition of '_exit_code'
   76 |     int _exit_code = 0;
      |         ^
largeDictionary.c:73:9: note: previous definition is here
   73 |     int _exit_code = 1;
      |         ^
largeDictionary.c:76:9: warning: mixing declarations and code is incompatible with standards before C99 [-Wdeclaration-after-statement]
   76 |     int _exit_code = 0;
      |         ^
```

Compilation fails as expected.
So we are left with the other 2 options.

Cyan4973: OK, I think it's a case of test not run in CI.
It seems to be a long test btw, maybe that's why it was not set in CI.

I think that a test that is not run in CI, and is just expected to be run manually from time to time,
loses a lot of its value.
If it's too long to run in CI, maybe it's worth making it shorter, or at least allow a shorter version to be run in CI.

This is somewhat separate from your PR, so you are not expected to act on this.
But it's also somewhat related too, since your changes in this PR are not captured by CI for the time being (hence errors remain invisible).

Cyan4973:
In #4416, I'm adding a test to CI, that should run this unit.

It is expected to fail, due to an existing warning.

It will be rebased on top of this PR once it's merged,
thus ensuring that this unit is tested in CI from now on.

Cyan4973 (approved):

PR #4403, opened 2025-06-05
fuzz: Fix FUZZ_malloc_rand() to return non-NULL for zero-size allocations

dloidolt:
This PR fixes a bug in the `FUZZ_malloc_rand` helper function introduced in fcaf06ddb489f683.

The FUZZ_malloc_rand() function was incorrectly always returning NULL for zero-size allocations. The random offset generated by FUZZ_dataProducer_int32Range() was not being added to the pointer variable, causing the function to always return (void *)0.

### DIFF
--- a/tests/fuzz/fuzz_helpers.c
+++ b/tests/fuzz/fuzz_helpers.c
@@ -31,12 +31,11 @@ void* FUZZ_malloc_rand(size_t size, FUZZ_dataProducer_t *producer)
         return mem;
     } else {
         uintptr_t ptr = 0;
-        /* Add +- 1M 50% of the time */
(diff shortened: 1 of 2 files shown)


Cyan4973:
I would really like to understand the perspective of @daniellerozenblit and @terrelln on this test,
because I'm not sure to understand what's the point of returning a junk pointer, 
as opposed to just returning a `NULL` one.

terrelln (approved):
Thanks for the fix!

terrelln:
@Cyan4973 It was to catch UB like this: https://github.com/facebook/zstd/issues/3507

There is more context linked in the PR description: https://github.com/facebook/zstd/pull/3555

PR #4411, opened 2025-06-16
AArch64: Add SVE2 implementation of histogram computation

arpadpanyik-arm:
The existing scalar implementation uses a 4-way pipelined histogram calculation which is very efficient on current CPUs. However, this can be further accelerated using the SVE2 [HISTSEG](https://developer.arm.com/documentation/ddi0602/2025-03/SVE-Instructions/HISTSEG--Count-matching-elements-in-vector-segments-?lang=en) instruction. Details are in the commit messages.

We also added some unit tests to the `fuzzer.c` to cover the existing and new SVE2 histogram functions, along with new SVE2-enabled QEMU runs in the CI (had to increase the timeout of the `test-cli-tests` from 200s to 800s).

Compression uplifts on a Neoverse V2 system, using Zstd-1.5.8 (e26dde3d) as a baseline, compiled with
`-O3 -march=armv8.2-a+sve2`:

```
                Clang-20    GCC-14
1#silesia.tar:   +6.173%   +5.987%
2#silesia.tar:   +5.200%   +5.011%
3#silesia.tar:   +4.332%   +5.031%
4#silesia.tar:   +2.789%   +3.064%
5#silesia.tar:   +2.028%   +1.838%
6#silesia.tar:   +1.562%   +1.340%
7#silesia.tar:   +1.160%   +0.959%
```
 The uplifts are much smaller at higher compression levels!

### DIFF
--- a/.github/workflows/dev-short-tests.yml
+++ b/.github/workflows/dev-short-tests.yml
@@ -435,6 +435,8 @@ jobs:
         make clean
         LDFLAGS="-static" CC=$XCC QEMU_SYS=$XEMU make -j check
         LDFLAGS="-static" CC=$XCC QEMU_SYS=$XEMU make -j -C tests test-cli-tests
+        CFLAGS="-march=armv8.2-a+sve2" LDFLAGS="-static" CC=$XCC QEMU_SYS=$XEMU make -j check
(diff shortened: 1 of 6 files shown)


Cyan4973 (approved):
It's a well located change, easy to follow, complete with published benchmark and a test set.

PR #4415, opened 2025-06-21
meson: drop unused variable

bgilbert:


Cyan4973:
Interesting.

`buildtype` is declared at line 20: 
```meson
'buildtype=release',
```

`buildtype` seems to be a standard `meson` variable, with substantial impacts on the compiler build flags.
So I'm not sure we could state that it's "unused".

To be fair, I don't understand what line 68 is supposed to achieve.

cc @eli-schwartz who knows this way better than I do.

Cyan4973:
OK, so here is my current understanding:

In `meson`, `buildtype` is a core _build option_.
At line 20, it receives a default value `release`.

But a _build option_ is **not** a _variable_.
It can't be used for a branch decision or assignment later in the script.
Therefore, at line 68, a _variable_ `buildtype` is created _from_ the _build option_ `buildtype`.
The fact that they have the same name is confusing here, especially for anyone not familiar with `meson` type system.
Anyway, even though they have the same name, they belong to separate naming arenas, and are different objects.

Line 68 creates a _variable_ `buildtype` in order to be employed later in the script.
But it's never used afterwards.
Hence the suggestion to remove it in this PR.
Even if the _variable_ is not created (i.e. "dropped"), it doesn't impact the _build option_, which already exists in its own separate plane.

Hence it makes sense to propose to _not_ create a variable that is not being used afterwards.
However, as mentioned, I'm not familiar enough with the `meson` ecosystem to know what are usual expectations. I could imagine for example that it's more or less standard practice to set this variable "just in case", and only think later if it's actually useful.

bgilbert:
Yes, your analysis is correct.  It isn't standard practice to set the variable "just in case"; it appears the commit that added the variable had intended to use it but ended up going in a different direction.

(We carry a copy of this file downstream in WrapDB and this came up during a repo-wide cleanup of the `buildtype` _option_.)

Cyan4973 (approved):

eli-schwartz:
Yup, this PR looks good to me. Meson internally uses the builtin option to make decisions, but build scripts can also choose to lookup the value and do something with it, like follow if/else branches based on comparing its string value. If it's never being used, it's a dead store.

Looking up builtin options is *usually* not needed at all, as they get used internally, in contrast to project options that *never* do anything unless you look them up in your build file and manually do something with them. Of course, one may have specific needs, so you are allowed to look up any builtin option too, but it's hardly a standard convention.


tl;dr dead store elimination good

PR #4416, opened 2025-06-21
added test-largeDictionary to dev-long CI script

Cyan4973:
note: expected to fail, due to a minor C90 warning.

Will be updated and fixed after merging of #4402.


Cyan4973:
After rebasing on top of #4402, the new test works, as intended

PR #4418, opened 2025-06-24
AArch64: Improve ZSTD_decodeSequence performance

arpadpanyik-arm:
LLVM's alias-analysis sometimes fails to see that a static-array member of a struct cannot alias other members. This patch:
- Reduces array accesses via struct indirection to aid load/store alias analysis under Clang.
- Converts dynamic array indexing into conditional-move arithmetic, eliminating branches and extra loads/stores for out-of-order CPUs.
- Reloads the bitstream only when match-length bits are consumed (assuming each reload only needs to happen once per match-length read), improving branch-prediction rates.
- Removes the `UNLIKELY()` hint, which recent compilers already handle well without cost.

Decompression uplifts on a Neoverse V2 system, using Zstd-1.5.8 compiled with `-O3 -march=armv8.2-a+sve2`:
```
                 Clang-19  Clang-20   Clang-*    GCC-13    GCC-14    GCC-15
 1#silesia.tar:  +11.556%  +16.203%   +0.240%   +0.609%   +2.216%   +7.891%
 2#silesia.tar:  +15.493%  +21.140%   -0.041%   +0.673%   +2.850%   +9.926%
 3#silesia.tar:  +16.887%  +22.570%   -0.183%   +1.029%   +3.056%  +10.660%
 4#silesia.tar:  +17.785%  +23.315%   -0.262%   +1.377%   +3.343%  +11.187%
 5#silesia.tar:  +18.125%  +24.175%   -0.466%   +1.213%   +3.350%  +11.228%
 6#silesia.tar:  +17.607%  +23.339%   -0.591%   +1.194%   +3.175%  +10.851%
 7#silesia.tar:  +17.463%  +22.837%   -0.486%   +1.372%   +3.292%  +10.868%
```
The same benchmarks on a Cortex-A720 system:
```
                 Clang-19  Clang-20   Clang-*    GCC-13    GCC-14    GCC-15
 1#silesia.tar:  +11.753%  +15.278%   +3.860%   +0.461%   +3.593%   +7.145%
 2#silesia.tar:  +14.838%  +17.210%   +4.785%   +0.809%   +4.489%   +8.910%
 3#silesia.tar:  +15.713%  +16.355%   +4.517%   +1.200%   +4.716%   +9.361%
 4#silesia.tar:  +15.733%  +14.869%   +4.065%   +1.628%   +4.638%   +9.381%
 5#silesia.tar:  +16.406%  +16.097%   +4.547%   +1.422%   +4.853%   +9.693%
 6#silesia.tar:  +16.195%  +15.881%   +4.460%   +1.279%   +4.531%   +9.260%
 7#silesia.tar:  +15.597%  +14.336%   +3.907%   +1.564%   +4.259%   +9.057%

* Requires Clang-21 support from LLVM commit hash
  `a53003fe23cb6c871e72d70ff2d3a075a7490da2`
  (Clang-21 hasn’t been released as of this writing)
```

This change could be advantageous or detrimental on other platforms with some compilers (±7%), so we only enabled it for AArch64 only.

Co-authored by:
 David Sherwood, David.Sherwood@arm.com
 Ola Liljedahl, Ola.Liljedahl@arm.com

### DIFF
--- a/lib/decompress/zstd_decompress_block.c
+++ b/lib/decompress/zstd_decompress_block.c
@@ -1236,6 +1236,10 @@ FORCE_INLINE_TEMPLATE seq_t
 ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets, const int isLastSeq)
 {
     seq_t seq;
+#if defined(__aarch64__)
+    size_t prevOffset0 = seqState->prevOffset[0];
+    size_t prevOffset1 = seqState->prevOffset[1];
+    size_t prevOffset2 = seqState->prevOffset[2];
     /*
      * ZSTD_seqSymbol is a 64 bits wide structure.
      * It can be loaded in one operation
@@ -1244,7 +1248,7 @@ ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets, c
      * operations that cause performance drop. This can be avoided by using this
      * ZSTD_memcpy hack.
      */
-#if defined(__aarch64__) && (defined(__GNUC__) && !defined(__clang__))
+#  if defined(__GNUC__) && !defined(__clang__)
     ZSTD_seqSymbol llDInfoS, mlDInfoS, ofDInfoS;
     ZSTD_seqSymbol* const llDInfo = &llDInfoS;
     ZSTD_seqSymbol* const mlDInfo = &mlDInfoS;
@@ -1252,11 +1256,11 @@ ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets, c
     ZSTD_memcpy(llDInfo, seqState->stateLL.table + seqState->stateLL.state, sizeof(ZSTD_seqSymbol));
     ZSTD_memcpy(mlDInfo, seqState->stateML.table + seqState->stateML.state, sizeof(ZSTD_seqSymbol));
     ZSTD_memcpy(ofDInfo, seqState->stateOffb.table + seqState->stateOffb.state, sizeof(ZSTD_seqSymbol));
-#else
+#  else
     const ZSTD_seqSymbol* const llDInfo = seqState->stateLL.table + seqState->stateLL.state;
     const ZSTD_seqSymbol* const mlDInfo = seqState->stateML.table + seqState->stateML.state;
     const ZSTD_seqSymbol* const ofDInfo = seqState->stateOffb.table + seqState->stateOffb.state;
-#endif
+#  endif
     seq.matchLength = mlDInfo->baseValue;
     seq.litLength = llDInfo->baseValue;
     {   U32 const ofBase = ofDInfo->baseValue;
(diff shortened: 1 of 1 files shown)


Cyan4973:
These are impressive speed gains,

it's also interesting to note that the measured benefits seem only significant for latest versions of `clang` and `gcc`.
Since we only see the relative gains, not absolute numbers, it's unclear if this PR fixes a performance degradation that appeared in these versions, or provide much better performance when combined with recent compilers.

Cyan4973:
As a complementary datapoint,
using Apple clang version 17.0.0 (clang-1700.0.13.5) compiler,
I observe ~+8/9% decompression speed improvement on a M1 Pro,
which is still pretty significant.

arpadpanyik-arm:
We mostly tested on recent compilers: GCC-13..15, Clang-19..21. Here are some results relative to Clang-19 as baseline for the first 7 compression levels on `silesia.tar`:

Neoverse V2 before:
|Clang-19 |Clang-20 |Clang-21 | GCC-13  | GCC-14  | GCC-15  |
|---------|---------|---------|---------|---------|---------|
| 100.000% | 103.098% | 119.322% | 101.879% | 121.210% | 106.283% |
| 100.000% | 101.894% | 123.367% | 99.660%  | 124.836% | 104.561% |
| 100.000% | 101.394% | 124.384% | 98.772%  | 126.234% | 104.129% |
| 100.000% | 101.188% | 124.886% | 98.466%  | 126.766% | 103.661% |
| 100.000% | 100.995% | 125.810% | 98.107%  | 127.285% | 103.580% |
| 100.000% | 101.242% | 125.456% | 98.448%  | 127.166% | 104.153% |
| 100.000% | 101.238% | 124.760% | 98.598%  | 126.777% | 104.099% |

Neoverse V2 after:
|Clang-19 |Clang-20 |Clang-21 | GCC-13  | GCC-14  | GCC-15  |
|---------|---------|---------|---------|---------|---------|
| 111.556% | 119.804% | 119.610% | 102.500% | 123.866% | 114.670% |
| 115.493% | 123.434% | 123.316% | 100.331% | 128.429% | 114.940% |
| 116.887% | 124.278% | 124.156% | 99.789%  | 130.083% | 115.230% |
| 117.785% | 124.780% | 124.558% | 99.822%  | 131.022% | 115.258% |
| 118.125% | 125.410% | 125.224% | 99.298%  | 131.541% | 115.211% |
| 117.607% | 124.872% | 124.714% | 99.624%  | 131.178% | 115.455% |
| 117.463% | 124.358% | 124.153% | 99.950%  | 130.925% | 115.413% |

Cortex-A720 before:
|Clang-19 |Clang-20 |Clang-21 | GCC-13  | GCC-14  | GCC-15  |
|---------|---------|---------|---------|---------|---------|
| 100.000% | 102.657% | 114.968% | 105.732% | 119.934% | 107.245% |
| 100.000% | 102.134% | 115.215% | 104.371% | 119.997% | 104.531% |
| 100.000% | 102.300% | 114.657% | 104.751% | 119.145% | 104.198% |
| 100.000% | 102.895% | 114.136% | 105.519% | 117.830% | 103.828% |
| 100.000% | 102.496% | 114.464% | 104.900% | 118.929% | 103.712% |
| 100.000% | 102.602% | 114.481% | 105.119% | 119.230% | 104.292% |
| 100.000% | 103.061% | 113.965% | 105.895% | 118.203% | 104.326% |

Cortex-A720 after:
|Clang-19 |Clang-20 |Clang-21 | GCC-13  | GCC-14  | GCC-15  |
|---------|---------|---------|---------|---------|---------|
| 111.753% | 118.341% | 119.406% | 106.220% | 124.243% | 114.908% |
| 114.838% | 119.712% | 120.728% | 105.216% | 125.385% | 113.845% |
| 115.713% | 119.032% | 119.836% | 106.008% | 124.764% | 113.953% |
| 115.733% | 118.195% | 118.777% | 107.237% | 123.295% | 113.568% |
| 116.406% | 118.996% | 119.669% | 106.393% | 124.702% | 113.764% |
| 116.195% | 118.897% | 119.588% | 106.463% | 124.633% | 113.950% |
| 115.597% | 117.836% | 118.418% | 107.552% | 123.238% | 113.776% |

Vanilla GCC-14 is the fastest, the in-development Clang-21 closed the gap mostly. With the patch applied GCC-14 is still the fastest but Clang-20 became the closest competitor.

Cyan4973 (approved):
There might be ways to reduce code duplication in order to improve code maintenance,
but I'm concerned such an effort would be challenging to synchronize on github.
So let's take this optimization as is, we'll figure out a way to reduce its maintenance complexity in a future effort.

lib/decompress/zstd_decompress_block.c
```diff
@@ -1244,19 +1248,127 @@ ZSTD_decodeSequence(seqState_t* seqState, const ZSTD_longOffset_e longOffsets, c
...
+                        prevOffset1 = prevOffset0;
+                        prevOffset0 = offset = temp;
+            }   }   }
+            seq.offset = offset;
+        }
+
+        if (mlBits > 0) {
+            seq.matchLength += BIT_readBitsFast(&seqState->DStream, mlBits/*>0*/);
+
+            if (MEM_32bits() && (mlBits+llBits >= STREAM_ACCUMULATOR_MIN_32-LONG_OFFSETS_MAX_EXTRA_BITS_32))
```
danilak-G, line 1338: Is this one the equivalent code? @Cyan4973 

In x86 version I see there is no scope for `if (mlBits > 0)`

```
        if (mlBits > 0)
            seq.matchLength += BIT_readBitsFast(&seqState->DStream, mlBits/*>0*/);

        if (MEM_32bits() && (mlBits+llBits >= STREAM_ACCUMULATOR_MIN_32-LONG_OFFSETS_MAX_EXTRA_BITS_32))
            BIT_reloadDStream(&seqState->DStream);
        if (MEM_64bits() && UNLIKELY(totalBits >= STREAM_ACCUMULATOR_MIN_64-(LLFSELog+MLFSELog+OffFSELog)))
            BIT_reloadDStream(&seqState->DStream);
 ```

Comment (Cyan4973) [reply lib/decompress/zstd_decompress_block.c:1338]: Indeed, it's not "strictly equivalent".

For the first `if (MEM_32bits(...` branch, I believe it works, accidentally.
One can see that, if `mlBits == 0`,
then we will not reach `mlBits+llBits >= STREAM_ACCUMULATOR_MIN_32-LONG_OFFSETS_MAX_EXTRA_BITS_32`, because `llBits` is capped `<= 16`.
This works because `STREAM_ACCUMULATOR_MIN_32 == 25` and `LONG_OFFSETS_MAX_EXTRA_BITS_32 == 5`.

Unfortunately, for the second `if (MEM_64bits(...` branch, it doesn't work so well.
`mlBits==0`, but we could still max out at `llBits==16` + `ofBits==30`, for a total of 46.
It's of course unlikely, but not impossible,
However, the limit is supposed to be `(STREAM_ACCUMULATOR_MIN_64-(LLFSELog+MLFSELog+OffFSELog))`.
with `STREAM_ACCUMULATOR_MIN_64==57` and `(LLFSELog+MLFSELog+OffFSELog)==9+9+8==26`,
hence a safe maximum of 41.

Now, in practice, there is some additional margin, because it implies that all FSE symbols are maxed out too, which is less likely when the nb of symbols is low, which it probably is if `llBits==16`. But all this is just "probably". Not guaranteed. A specific input might be able to trigger this issue.

It also shows that it's difficult to reach these conditions, because it implies a huge amount of data, in order to reach very large offset and length values, in order to produce many bits in a single sequence. Most automated tests focus on smaller dataset, for a question of iteration speed and also memory limits.

Anyway, I think it would deserve a fix.

Even the first `if()` branch, which seems safe, would deserve a `STATIC_ASSERT()` test if it stays within the  `if (mlBits > 0) { ... }` code block, to ensure that the constants have the required values to guarantee success, in case someone changes them in some future.

Comment (danilak-G) [reply lib/decompress/zstd_decompress_block.c:1338]: Thanks for a such fast response

If I set in the x86 version

```diff
if (MEM_64bits() && UNLIKELY(totalBits >= STREAM_ACCUMULATOR_MIN_64-(LLFSELog+MLFSELog+OffFSELog)))
+{
+           if (mlBits == 0) abort();
            BIT_reloadDStream(&seqState->DStream);
+}
```

Some tests from fuzzer.c tests fail with abort

Cyan4973: So that's good news, it seems we have a test that would have detected the issue.
It just (probably) wasn't run on this code path.

danilak-G: Or it was not enough to detect the corruption, i.e. worked for other reasons

Anyway, thanks for a blazingly fast response. Credit to 1 internal team at Google which caught it during the qual release and brought it to our attention

arpadpanyik-arm: Thanks for pointing out that this could lead to errors in extreme cases!
If we restore the bitstream reload to its original position with an `UNLIKELY` hint (useful for most compilers here):
```cpp
        if (mlBits > 0)
            seq.matchLength += BIT_readBitsFast(&seqState->DStream, mlBits/*>0*/);

        if (UNLIKELY(totalBits >= STREAM_ACCUMULATOR_MIN_64-(LLFSELog+MLFSELog+OffFSELog)))
            BIT_reloadDStream(&seqState->DStream);
```
the performance gains mostly disappear on GCC (this patch was specifically created for LLVM), while Clang remains largely unaffected, Clang-19 can even improve!

Measurements on **Neoverse-V2** show the following results:
```
                 Clang-19  Clang-20   Clang-21   GCC-13    GCC-14    GCC-15
 1#silesia.tar:   +2.950%   +0.000%   -0.229%   -0.896%   -2.331%   -6.305%
 2#silesia.tar:   +3.752%   +0.041%   -0.452%   -1.052%   -2.615%   -7.742%
 3#silesia.tar:   +3.872%   +0.000%   -0.457%   -1.102%   -2.768%   -8.321%
 4#silesia.tar:   +3.734%   -0.049%   -0.453%   -1.336%   -2.964%   -8.686%
 5#silesia.tar:   +4.158%   -0.049%   -0.497%   -1.292%   -2.945%   -8.668%
 6#silesia.tar:   +3.965%   -0.053%   -0.454%   -1.251%   -2.930%   -8.451%
 7#silesia.tar:   +3.647%   -0.138%   -0.414%   -1.300%   -2.989%   -8.452%
```

Shall I create a follow-up patch?

Cyan4973: Yeah, we need to fix that.
But we also probably need to expand our CI test suite a bit, adding a test that would have been able to catch the issue.

Cyan4973: An earlier [message](https://github.com/facebook/zstd/pull/4418#discussion_r2387115338) from @danilak-G suggested that it would be relatively easy to reproduce this bug during testing using the `fuzzer`, making it straightforward to detect.

However, the situation is more complicated than it appears.

The `assert()` only demonstrates that the "dangerous" branch can be executed even when `mlBits == 0`. The test itself merely checks that decoding will always be successful. If the test fails, it indicates that there might be a risk—something could potentially go wrong. In reality, several conditions must occur simultaneously to actually trigger a decoding error, so successful decoding remains quite likely.

Hence, so far, I have not been able to produce a decoding issue using the `fuzzer`.

PR #4419, opened 2025-06-24
Check for job before releasing resources

SiliconA-Z:
ZSTDMT_freeCCtx calls ZSTDMT_releaseAllJobResources, but ZSTDMT_releaseAllJobResources may be called when ZSTDMT_freeCCtx is called when initialization fails, resulting in a NULL pointer dereference.


Cyan4973 (approved):

Cyan4973:
Thanks, good catch @AZero13 !

PR #4425, opened 2025-07-03
Add CI setup for packaging Windows on ARM artifacts

MugundanMCW:
Description:
- The PR introduces CI setup for building and packaging zstd artifacts for Windows on ARM devices
- The CI setup depends on CMake and Visual Studio 2022 for building zstd from source.

### DIFF
--- a/.github/workflows/windows-artifacts.yml
+++ b/.github/workflows/windows-artifacts.yml
@@ -13,69 +13,107 @@ jobs:
   windows-artifacts:
     permissions:
       contents: write # to fetch code and upload artifacts
-    # see https://ariya.io/2020/07/on-github-actions-with-msys2
-    runs-on: windows-latest
-    # see https://github.com/msys2/setup-msys2
+    # For msys2, see https://ariya.io/2020/07/on-github-actions-with-msys2
+    runs-on: ${{ matrix.build_type == 'cmake' && 'windows-11-arm' || 'windows-latest' }}
     strategy:
+      # For msys2, see https://github.com/msys2/setup-msys2
       matrix:
         include:
-          - { msystem: mingw64, env: x86_64, ziparch: win64 }
-          - { msystem: mingw32, env: i686, ziparch: win32 }
+          - { msystem: mingw64, env: x86_64, ziparch: win64, build_type: msys2 }
+          - { msystem: mingw32, env: i686, ziparch: win32, build_type: msys2 }
+          - { msystem: null, env: arm64, ziparch: win-arm64, build_type: cmake }
+    
     defaults:
       run:
-        shell: msys2 {0}
+        shell: ${{ matrix.build_type == 'cmake' && 'pwsh' || 'msys2 {0}' }}
+    
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # tag=v4.2.2
-    - uses: msys2/setup-msys2@40677d36a502eb2cf0fb808cc9dec31bf6152638 # tag=v2.28.0
-      with:
-        msystem: ${{ matrix.msystem }}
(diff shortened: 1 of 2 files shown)


Cyan4973:
A few points :
- CLA signing is a mandatory part of this process. At some point it will have to be fixed for this contribution to be merged. 
- The modifications of the Github Action are significant, which makes it difficult to validate through review only.  Ideally, we would also need this modification to be tested in CI. It's unusually more difficult in this case, because the modified action is disabled under most circumstances. But we need to find a solution.

One of the options would be to name the feature branch `test_artifacts`, which should trigger the modified action.

.github/workflows/windows-artifacts.yml
```diff
@@ -13,69 +13,107 @@ jobs:
...
-    # see https://github.com/msys2/setup-msys2
+    # For msys2, see https://ariya.io/2020/07/on-github-actions-with-msys2
+    runs-on: ${{ matrix.build_type == 'cmake' && 'windows-11-arm' || 'windows-latest' }}
     strategy:
+      # For msys2, see https://github.com/msys2/setup-msys2
       matrix:
         include:
-          - { msystem: mingw64, env: x86_64, ziparch: win64 }
-          - { msystem: mingw32, env: i686, ziparch: win32 }
+          - { msystem: mingw64, env: x86_64, ziparch: win64, build_type: msys2 }
```
Cyan4973: In this naming scheme, `build_type` can be either `msys2` or `cmake`.
This feels a bit strange, as `msys2` is an environment, not a build solution nor type.

Also, pay attention that, within `cmake` context, `build type` traditionally refers to something specific, like `opt` or `debug` build types.

For clarity of maintenance, I would invite to reconsider this naming scheme.
You probably meant something like `shell` == `msys2` || `pwsh`, or `build` == `make` || `cmake`.

.github/workflows/windows-artifacts.yml
```diff
@@ -13,69 +13,107 @@ jobs:
...
+          - { msystem: mingw32, env: i686, ziparch: win32, build_type: msys2 }
+          - { msystem: null, env: arm64, ziparch: win-arm64, build_type: cmake }
+    
     defaults:
       run:
-        shell: msys2 {0}
+        shell: ${{ matrix.build_type == 'cmake' && 'pwsh' || 'msys2 {0}' }}
```
Cyan4973: There are more idiomatic ways to state the same thing, with less long-context to maintain, 
for example:

```YAML
- name: Build (MSYS2)
  if: matrix.build_type == 'msys2'
  shell: msys2 {0}
  run: |
    # msys2 commands here

- name: Build (CMake/ARM64)
  if: matrix.build_type == 'cmake'
  shell: pwsh
  run: |
    # PowerShell commands here
 ```

I have not yet understood why Powershell `pwsh` is necessary here, since `msys2` also provides `cmake`.
 
On the other hand, if both code paths feature a lot of differences, maybe it would be better to just separate them completely into different actions.

.github/workflows/windows-artifacts.yml
MugundanMCW: @Cyan4973 Thanks for the suggestion, I have replaced built type with shell for better naming scheme and maintainability.

.github/workflows/windows-artifacts.yml
MugundanMCW: @Cyan4973 Yes, MSYS2 also provides CMake, but currently there's no native MSYS2 environment available for Windows on ARM64. As a result, running CMake within the MSYS2 environment on Windows ARM64 often involves emulation.

thiru-mcw:
Closing this PR as #4465 is created with the same changes.

Thanks.

PR #4435, opened 2025-07-13
add riscv  ci

zijianli1234:
This PR adds RISCV RVV CI support, including tests for vlen=128, 256, 512, and includes detection of compiler versions for RVV extensions. Some compilation warnings have also been fixed. Additionally, on the Spacemit® X60 @1.6 GHz platform, when the RISC-V Vector (V) extension is enabled along with O3 optimization,unaligned memory access errors may occur if `-DMEM_FORCE_MEMORY_ACCESS=0` is not specified. Therefore, this compile-time flag has been added to the CI configuration.This PR is largely identical to a previous one(#4422 ). The reason for closing and resubmitting is that the previous branch was out of sync due to an outdated version. After pulling the latest changes, the PR is re-submitted.


Cyan4973 (approved):
Looks good to me !

PR #4441, opened 2025-07-18
lib/Makefile: fix MinGW installation issues

Keno:
1. Install the .dll into bin/ not lib/
2. Install the .dll.a import library into lib/
3. Do not attempt to create symlinks for so-versioned binaries
4. Incorporate the LIBVER_MAJOR per windows convention.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -134,7 +134,7 @@ endif
 
 ifneq (,$(filter Windows%,$(TARGET_SYSTEM)))
 
-LIBZSTD = dll/libzstd.dll
+LIBZSTD = dll/libzstd-$(LIBVER_MAJOR).dll
 $(LIBZSTD): $(ZSTD_FILES)
 	@echo compiling dynamic library $(LIBVER)
 	$(CC) $(FLAGS) -DZSTD_DLL_EXPORT=1 -Wl,--out-implib,dll/libzstd.dll.a -shared $^ -o $@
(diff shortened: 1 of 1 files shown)


lib/Makefile
```diff
@@ -284,6 +284,12 @@ exec_prefix ?= $(PREFIX)
 EXEC_PREFIX ?= $(exec_prefix)
 libdir      ?= $(EXEC_PREFIX)/lib
 LIBDIR      ?= $(libdir)
+ifeq (,$(filter Windows%,$(TARGET_SYSTEM)))
+shlibdir    ?= $(EXEC_PREFIX)/bin
+else
+shlibdir    ?= $(EXEC_PREFIX)/bin
```
Cyan4973, line 290: I don't get it, 
do they both receive the same value ?

Keno: Sorry, typo in refactoring for preparing the PR. The non-windows one is supposed to be lib of course.

Keno: fixed

Cyan4973:
The title states : "Correct a few Windows build issues"

Do you actually mean:
"Fix a few mingw/msys installation issues ?"

Keno:
Yes, they're installation issues. I think it would be accurate to say that they are mingw issues in that mingw is the only Windows toolchain that lib/Makefile supports in practice [1]. I used Windows in the title since that's what the Makefile actually checks for, I don't have any problem with mingw instead though. I don't think any of the issues are msys specific.

[1] With respect to the soname change, I believe the `-somajor-sominor.dll` convention was originally introduced by cygwin in 2000 (https://cygwin.com/pipermail/cygwin/2000-August/040045.html is the earliest discussion I could find), for which it was then added to libtool in 2001. Subsequent toolchains then copied the convention when they added themselves to libtool, making it the de facto standard for everything that used libtool to target windows, so not technically specific to either mingw nor msys, but probably most closely associated with mingw in practice. It's also been adopted as the major convention by non-autotools toolchains. For example, the meson (https://github.com/mesonbuild/meson/blob/ab261f9951f235c776b8664fbdbf0365a8cf535d/mesonbuild/build.py#L2364) build for libzstd will produce `libzstd-1.dll`, not `libzstd.dll`. That said, it is true that windows naming conventions are a bit all over the place here. My primary objective was making the makefil build match meson, since we use both downstream, depending on the environment.

cclauss:
Please rebase.
> This branch has conflicts that must be resolved

PR #4454, opened 2025-08-05
fix: off-by-one error in `convertPathnameToDirName`

### LINKED ISSUE #4380: `convertPathnameToDirName` fails to strip trailing slashes due to off‑by‑one index
**Describe the bug**
In `convertPathnameToDirName()`, the code attempts to remove trailing `/` characters with:
```c
len = strlen(pathname);
while (pathname[len] == PATH_SEP) {
    pathname[len] = '\0';
    len--;
}
``` 
However, `strlen(pathname)` is the index of the terminating NUL.  As a result `pathname[len]` is always `'\0'`, never `'/'`, so the loop never runs and trailing slashes remain.

**To Reproduce**

1. Call `convertPathnameToDirName("foo/bar/")`.
2. Expect `"foo/bar"`, but function returns `"foo/bar/"`.

Pineapple-Soup:
## Fixes #4380 

- Use `pathname[len-1]` instead of `pathname[len]` to compare/modify last character instead of null terminator
- Move `len == 0` check to inside loop to avoid OOB

### DIFF
--- a/programs/util.c
+++ b/programs/util.c
@@ -1148,11 +1148,11 @@ static void convertPathnameToDirName(char *pathname)
     /* remove trailing '/' chars */
     len = strlen(pathname);
     assert(len > 0);
-    while (pathname[len] == PATH_SEP) {
-        pathname[len] = '\0';
+    while (pathname[len-1] == PATH_SEP) {
+        pathname[len-1] = '\0';
         len--;
+        if (len == 0) return;
     }
-    if (len == 0) return;
 
     /* if input is a single file, return '.' instead. i.e.
      * "xyz/abc/file.txt" => "xyz/abc"


programs/util.c
```diff
@@ -1148,11 +1148,11 @@ static void convertPathnameToDirName(char *pathname)
     /* remove trailing '/' chars */
```
Cyan4973, line 1148: To be fair, I'm not sure if this part of the code makes sense.

The goal of the method `convertPathnameToDirName()` is to convert a full path of a filename into a just its directory part.

But if `pathname` ends with a `/` character, then it already represents a directory name.
In which case, there should be no need to change anything.

In contrast, if we remove the last `/`, the next stage would strip the last part of the path, and we would end up with the upper directory. And I suspect that's not the wanted behavior.

At this point, my suspicion is that this paragraph is useless, and should just be removed. "Fixing" it could actually introduce problems.

Note: a good follow up could be to improve the tests, by finding scenarios that introduce problems if this function is badly defined. Since our CI signal is all-greens, that means it does not capture the issue.

Cyan4973 (changes requested):

Ari4ka (commented):
__

PR #4461, opened 2025-08-17
Update CMakeLists.txt

qr243vbi:


Cyan4973:
The PR doesn't clearly explain what specific issue is supposed to be solved. 

- What exact scenario is failing currently?
- Could you provide a minimal reproduction case?

Also:
- `cmake` tests are currently failing. It should be possible for you to run the same tests locally, to observe the behavior.
- (Once ready) CLA is a required part of the process to authorize a merge

PR #4463, opened 2025-08-19
Check for build environment instead of just _GNU_SOURCE

brad0:
Fixes the build on OpenBSD and NetBSD. It is too easy for _GNU_SOURCE
to be defined even on non-Linux systems. Found via py-zstandard with
the embedded copy of zstandard and Python defines _GNU_SOURCE.

Also simplify the Linux checking, there is no need to check the rest
of the symbol names.

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -21,10 +21,12 @@
 /*-*************************************
 *  Dependencies
 ***************************************/
-/* qsort_r is an extension. */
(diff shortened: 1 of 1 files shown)


Cyan4973:
The proposed change seems to be problematic for Android environment.

brad0:
> The proposed change seems to be problematic for Android environment.

Yes, I should have noticed that the first time around.

Cyan4973 (approved):

PR #4468, opened 2025-08-25
Enable qsort_r() on more OS's

brad0:
qsort_r() is now part of POSIX.1-2024.

Enable qsort_r() on Android API level 36 AKA Android 16, Hurd (glibc),
Haiku, modern FreeBSD 14 versions or newer and modern NetBSD -current
or 11 and newer.

### DIFF
--- a/lib/dictBuilder/cover.c
+++ b/lib/dictBuilder/cover.c
@@ -21,11 +21,8 @@
 /*-*************************************
 *  Dependencies
 ***************************************/
-/* qsort_r is an extension.
- *
- * Android NDK does not ship qsort_r().
- */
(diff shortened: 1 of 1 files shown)


Cyan4973:
> qsort_r() is now part of POSIX.1-2024.

That's great news !

The updated detection method relies on specific OS versioning logic.
It works fine for the intended systems,
but it feels like, in the best of cases, it should be able to be more generic by employing posix macro variables,
thus making it compatible with a bunch of present and future posix OSes,
as opposed to special casing them one by one.

I would expect variables like `_POSIX_VERSION` and `_POSIX_C_SOURCE` to be used for this detection.

PR #4474, opened 2025-09-02
Update manual about the default value of `-T#`/`--threads=#`

jlokier:
The section about `ZSTD_NBTHREADS` already explains the default number of threads, since it changed from 1 (commit 17beeb5).  But the option description for `-T#`/`--threads=#` incorrectly said the default was still 1.

I noticed this when I found compression slower with `-T1` than without it.


Cyan4973 (approved):

Cyan4973:
Thanks @jlokier , that's indeed a good point !

PR #4478, opened 2025-09-05
add RVV optimization for ZSTD_row_getMatchMask

w1m024:
This pull request introduces a RISC-V Vector (RVV) specific optimization for the `ZSTD_row_getMatchMask` function, replacing the generic SWAR implementation on RV64 platforms with V-extension support. The goal is to leverage RVV's parallel computation capabilities to improve performance on the RISC-V architecture. 

### Microbenchmark Results
A microbenchmark isolating the `ZSTD_row_getMatchMask` function shows a significant speedup compared to the SWAR fallback.

| rowEntries | Speedup   |
| ---------- | --------- |
| 16 bytes   | **3.45x** |
| 32 bytes   | **4.24x** |
| 64 bytes   | **4.73x** |

### Fullbench

The overall impact on the `fullbench` is modest. However, the new implementation shows a consistent small improvement and, most importantly, **no performance regression**.

Ten consecutive runs were performed for both the original version and the RVV-optimized version. The RVV version demonstrates a **~0.3%** average performance increase for this specific function under these test conditions.

| Version    | Average Speed        | Fastest Run          | Slowest Run      |
| ---------- | -------------------- | -------------------- | ---------------- |
| **Origin** | 1,081,809.8 MB/s     | 1,091,072.1 MB/s     | **1,068,435.0 MB/s** |
| **RVV**    | **1,085,074.8 MB/s** | **1,096,995.4 MB/s** | 1,063,997.2 MB/s |

## Validation
- [x] All quick check pass (`make check`).  
- [x] All longer check pass（`make test`）
- [x] Static analysis reports no new issues (`make staticAnalyze`).

### DIFF
--- a/.github/workflows/android-ndk-build.yml
+++ b/.github/workflows/android-ndk-build.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # tag=v5.0.0
 
     - name: Set up JDK 17
-      uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
+      uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5.0.0
       with:
         java-version: '17'
         distribution: 'temurin'


camel-cdr:
The code seems incorect, as you only load 16 8-bit elements in every code path.

You also dynamically select LMUL based on vl, but don't do it relative to VLEN.

Edit: the function should be very simple, just vle8+vmseq and then a e64 vmv.x.v to conver the mask to GPR. Then you can specialize LMUL delection based on vlenb and rowEntires.

You should use the SSE implementation as a reference, the NEON one is weird because NEON doesn't have movemask.

Cyan4973:
I presume this PR was closed due to correctness concerns ?

Note that I liked that it was trying to simplify the `RVV` code path,
in which case, even with identical performance, it's still a win.

But obviously it has to be correct.
On this front, @t-mat created the `sanity_test` validation tool, in `tests/`.
I would have expected this tool to be useful to check that a new formulation is correct.

But if the new formulation was incorrect and the tool did not discover it, then we may need to update this tool.

w1m024:
> The code seems incorect, as you only load 16 8-bit elements in every code path.
> 
> You also dynamically select LMUL based on vl, but don't do it relative to VLEN.
> 
> Edit: the function should be very simple, just vle8+vmseq and then a e64 vmv.x.v to conver the mask to GPR. Then you can specialize LMUL delection based on vlenb and rowEntires.
> 
> You should use the SSE implementation as a reference, the NEON one is weird because NEON doesn't have movemask.

Thank you  for your detailed feedback and suggestions.

Your analysis of the relationship between `vl `and `VLEN `was spot on and helped me understand the root cause of the portability issue. Your suggestion to use the SSE implementation as a reference was the key hint that pointed me in the right direction.

I will refactor the code based on your advice and submit it in a new PR.

w1m024:
> I presume this PR was closed due to correctness concerns ?
> 
> Note that I liked that it was trying to simplify the `RVV` code path, in which case, even with identical performance, it's still a win.
> 
> But obviously it has to be correct. On this front, @t-mat created the `sanity_test` validation tool, in `tests/`. I would have expected this tool to be useful to check that a new formulation is correct.
> 
> But if the new formulation was incorrect and the tool did not discover it, then we may need to update this tool.

Thank you for your encouragement and for pointing out the importance of `sanity_test.`

Regarding the testing issue you mentioned, I'd like to clarify. This PR did indeed pass `make test` before I submitted it. As @camel-cdr  pointed out, my initial code had a serious portability issue. It likely passed the test by coincidence: the VLEN of my hardware happened to cause the correct return value of `vsetvl `to be the same as my hardcoded `vl `value.

This means that while the code worked on my specific machine, it would fail on hardware with a different VLEN. Therefore, `sanity_test `couldn't expose this latent, hardware-dependent bug in my environment.

This PR was closed due to a mistake I made with Git, as I'm still learning the ropes. I'm currently working on the correct implementation, which references the SSE version to solve this problem at its root. I will open a new PR with the fix soon, after I've fully verified it with `sanity_test`.

PR #4507, opened 2025-10-06
Advertise OpenZL in README

felixhandte:
I recall we had discussed cross-promoting OpenZL in this way, but I don't recall where we landed. What do you think of a mention like this?


Ari4ka (commented):
****

PR #4525, opened 2025-11-03
Add RISC-V 64-bit architecture detection

Polaris-911:
- Add compile-time check for RISC-V 64-bit via __riscv and __riscv_xlen == 64
- Align with existing 64-bit architecture detection logic (x86_64/ARM64 etc.)
- Ensure __64BIT__ is defined for RISC-V 64-bit builds


Polaris-911:
Hi @Cyan4973 , thanks again for your time! Would you also be able to take a look at this PR when convenient?

Cyan4973 (approved):
Looks good to me !

Issue #3892, opened 2024-01-31
Compiler warnings present when integrated with Swift Package Manager


anlaital-oura:
**Describe the bug**

There are bunch of warnings from `zstd.h` in the following form when compiling:
> Definition of configuration mac 'ZSTDLIB_VISIBLE' has no effect on the import of 'libzstd'; pass '-DZSTDLIB_VISIBLE=...' on the command line to configure the module

It is possible to have these preprocessor definitions in `cSettings` in the `Package.swift` integration site, but that should not be necessary in my opinion and the project should build without warnings.

**To Reproduce**

1. Use library with Swift Package Manager
2. Compile with `swift build`

**Expected behavior**

No warnings should be given.

**Screenshots and charts**

![image](https://github.com/facebook/zstd/assets/133648611/6fae113c-8381-4fd1-aa3c-8976c8364f92)

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 14.2.1
 - Compiler swiftc
 - Flags n/a
 - Other relevant hardware specs n/a
 - Build system SPM

PR #4531, opened 2025-11-25
modulemap: remove `config_macros`

### LINKED ISSUE #3328: SwiftPM package always produces warnings when `import libzstd`
**Describe the bug**
Import libzstd into Swift file causes compilation warnings.

**To Reproduce**
1. Clone repro repository `git clone https://github.com/yyaroshevich/zstd-warning-swift.git`
2. Compile project `(cd zstd-warning-swift; xcodebuild -project ./ZStdWarning.xcodeproj)`
3. Observe compilation is failed due to warnings (warnings as errors are enabled)

**Expected behavior**
No warnings produced when import `libzstd`.

**Screenshots and charts**
image

 - OS: Mac OS X 13.0.1
 - Version libzstd: dev branch & 1.5.2
 - Compiler: Swift 5.7, XCode 14.1

lukaskollmer:
fixes #3328 and #3892

issue at hand: the module.modulemap file was changed in #2953 and #3363 to contain a `config_macros` block with a list of most of the macros that can be used to configure various aspects of how Zstd builds and behaves.

however, clang's definition of a "config macro" is different from what one might instinctively assume a config macro to be: it only covers those macros that are exclusively passed via the command line (e.g. as `-DMACRO_NAME`), and does not allow the macro be `#define`-d within the codebase, even if it wasn't defined via the CLI.

zstd.h (and some other headers) currently contain blocks that conditionally `#define` some of the macros used to configure zstd, setting them to their default values if they are undefined (ie, if they haven't explicitly been set somewhere prior in the build process).
this causes clang to emit a warning for each of these macros, unless the program using zstd explicitly specifies their values via the command line (which a program that wants to simply use the default value obviously would not do)

you can see e.g. [here](https://reviews.llvm.org/D41239) for an example of clang's / LLVM's usage & meaning of the term "config macro": they're effectively simple valueless macros that derive their meaning exclusively from the fact whether they are defined or whether they are not defined.

### DIFF
--- a/lib/module.modulemap
+++ b/lib/module.modulemap
@@ -1,27 +1,6 @@
 module libzstd [extern_c] {
     header "zstd.h"
     export *
-    config_macros [exhaustive] \
-        /* zstd.h */ \
-        ZSTD_STATIC_LINKING_ONLY, \
-        ZSTDLIB_VISIBILITY, \
-        ZSTDLIB_VISIBLE, \
-        ZSTDLIB_HIDDEN, \
-        ZSTD_DLL_EXPORT, \
-        ZSTDLIB_STATIC_API, \
-        ZSTD_DISABLE_DEPRECATE_WARNINGS, \
-        ZSTD_CLEVEL_DEFAULT, \
-        /* zdict.h */ \
-        ZDICT_STATIC_LINKING_ONLY, \
-        ZDICTLIB_VISIBLE, \
-        ZDICTLIB_HIDDEN, \
-        ZDICTLIB_VISIBILITY, \
-        ZDICTLIB_STATIC_API, \
-        ZDICT_DISABLE_DEPRECATE_WARNINGS, \
-        /* zstd_errors.h */ \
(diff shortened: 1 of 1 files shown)


Cyan4973:
OK, so if I understand correctly, `config_macros` should only be for macros which are command-line only, and actually command-line required ? Otherwise, `clang` warn for any macro value not defined on the command line ?

lukaskollmer:
correct; technically clang is warning about the fact that in addition to the command line definition there also is a `#define` for the macro, though of course in a sense that's the same thing as what you described.

i think eg `ZSTD_MULTITHREAD` would qualify as a config macro but i haven't tested that

Cyan4973 (approved):

lukaskollmer:
@Cyan4973 thanks for merging this! do you have a rough estimate when the next release will be tagged?

Cyan4973:
There's no ETA at this point.
We will likely discuss this next year.

PR #4533, opened 2025-11-26
Improve compression performance by conditionally skipping lazy evaluation

BiplabRaut:
- Optional ZSTD_COMPRESS_FAST flag provided to speed up compression by restricting lazy evaluation to short matches.
  Applied to levels 6 and 7. This helps improve lazy compression strategy. Optimization disabled by default.

- This code change improves compression performance with little to no tradeoff for compression ratio and decompression   performance. It is useful for applications that have high compression speed requirements.

- Skip determinism test when this optimization option is enabled. Since the compression ratio varies, so it can't satisfy the determinism tests.

- **Performance Highlights** :-
  Benchmark result (single-threaded execution) graphs are shared below for the datasets: Silesia, Calgary, Canterbury and FreeBSD. 
  Other datasets like geekBench and Enwik showed similar trends.
  Test Setup: AMD Genoa (Zen4) CPU, SMT OFF, GCC 14.2.1
  
silesia_comparison

calgary_comparison

canterbury_comparison

freeBSD_comparison

### DIFF
--- a/lib/README.md
+++ b/lib/README.md
@@ -212,6 +212,11 @@ The file structure is designed to make this selection manually achievable for an
   For this scenario, it can be set as `ZDICT_QSORT=ZDICT_QSORT_C90`.
   Other selectable suffixes are `_GNU`, `_APPLE`, `_MSVC` and `_C11`.
 
+- The build macro `ZSTD_LAZY_SKIP_LONG_MATCHES` can be defined for fast lazy 
+  evaluation from compression level 6 to 7. This improves compression performance with
+  little to no tradeoff for compressibility and decompression performance.
+  This option is designed for applications with demanding compression speed requirements.
+
 #### Windows : using MinGW+MSYS to create DLL
 
 DLL can be created using MinGW+MSYS with the `make libzstd` command.
--- a/lib/compress/clevels.h
+++ b/lib/compress/clevels.h
@@ -129,6 +129,32 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
 },
 };
 
+#ifdef ZSTD_LAZY_SKIP_LONG_MATCHES
+static const ZSTD_compressionParameters ZSTD_CParametersFastCompress[4][2] = {
+{   /* "default" - for any srcSize > 256 KB */
+    /* W,  C,  H,  S,  L, TL, strat */
+    /* Larger windows and settings from next levels to improve ratio.
+     * Accounts for ratio drop from selective lazy evaluation */
+    { 23, 19, 20,  4,  5,  8, ZSTD_lazy    },  /* level  6 */
+    { 23, 19, 20,  4,  5, 16, ZSTD_lazy    },  /* level  7 */
+},
+{   /* for srcSize <= 256 KB */
+    /* W,  C,  H,  S,  L,  T, strat */
+    { 20, 18, 19,  4,  4,  4, ZSTD_lazy    },  /* level  6 */
+    { 20, 18, 19,  4,  5,  8, ZSTD_lazy    },  /* level  7 */
+},
+{   /* for srcSize <= 128 KB */
+    /* W,  C,  H,  S,  L,  T, strat */
+    { 19, 16, 17,  3,  4,  8, ZSTD_lazy   },  /* level  6 */
+    { 19, 16, 17,  4,  5,  16, ZSTD_lazy  },  /* level  7 */
+},
+{   /* for srcSize <= 16 KB */
+    /* W,  C,  H,  S,  L,  T, strat */
+    { 14, 14, 14,  4,  4,  8, ZSTD_lazy2   },  /* level  6 */
+    { 14, 14, 14,  6,  4,  8, ZSTD_lazy2   },  /* level  7 */
+},
+};
+#endif
 
 
 #endif  /* ZSTD_CLEVELS_H */
--- a/lib/compress/zstd_compress.c
+++ b/lib/compress/zstd_compress.c
@@ -5260,6 +5260,13 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
                                     ZSTD_buffered_policy_e zbuff)
 {
     size_t const dictContentSize = cdict ? cdict->dictContentSize : dictSize;
+#ifdef ZSTD_LAZY_SKIP_LONG_MATCHES
+    int compressionLevel = (params->compressionLevel == 0) ? cctx->requestedParams.compressionLevel 
+                                                       : params->compressionLevel;
+    compressionLevel = MIN(MAX(0, compressionLevel), ZSTD_maxCLevel());
+    assert(compressionLevel >= 0 && compressionLevel <= ZSTD_MAX_CLEVEL);
+    cctx->seqStore.lazyLimit = ZSTD_compressFastLazyLimit[compressionLevel];
+#endif
 #if ZSTD_TRACE
     cctx->traceCtx = (ZSTD_trace_compress_begin != NULL) ? ZSTD_trace_compress_begin(cctx) : 0;
 #endif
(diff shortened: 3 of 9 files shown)


lib/compress/clevels.h
```diff
@@ -31,8 +31,15 @@ static const ZSTD_compressionParameters ZSTD_defaultCParameters[4][ZSTD_MAX_CLEV
     { 21, 16, 17,  1,  5,  0, ZSTD_dfast   },  /* level  3 */
     { 21, 18, 18,  1,  5,  0, ZSTD_dfast   },  /* level  4 */
     { 21, 18, 19,  3,  5,  2, ZSTD_greedy  },  /* level  5 */
+#ifdef ZSTD_ENABLE_COMPRESS_FAST
```
Cyan4973: This is too disruptive, and will make maintenance of these tables more complex for future patches.

lib/README.md
```diff
@@ -212,6 +212,11 @@ The file structure is designed to make this selection manually achievable for an
   For this scenario, it can be set as `ZDICT_QSORT=ZDICT_QSORT_C90`.
   Other selectable suffixes are `_GNU`, `_APPLE`, `_MSVC` and `_C11`.
 
+- The build macro `ZSTD_COMPRESS_FAST` can be defined for fast lazy evaluation
```
Cyan4973: The macro name is not specific enough. The name implies that this is about the "fast mode", while the changes are actually impacting the `ZSTD_lazy` strategy. Consider a more accurate name, or better, if we can find just a better trade off, maybe there would be no need for another build macro.

lib/compress/zstd_compress.c
```diff
@@ -5260,12 +5260,22 @@ static size_t ZSTD_compressBegin_internal(ZSTD_CCtx* cctx,
...
+    int compressionLevel = 0;
+#endif
 #if ZSTD_TRACE
     cctx->traceCtx = (ZSTD_trace_compress_begin != NULL) ? ZSTD_trace_compress_begin(cctx) : 0;
 #endif
     DEBUGLOG(4, "ZSTD_compressBegin_internal: wlog=%u", params->cParams.windowLog);
     /* params are supposed to be fully validated at this point */
     assert(!ZSTD_isError(ZSTD_checkCParams(params->cParams)));
+#ifdef ZSTD_ENABLE_COMPRESS_FAST
+    compressionLevel = (params->compressionLevel == 0) ? cctx->requestedParams.compressionLevel
```
Cyan4973: create a block which both starts and end `compressionLevel` lifetime.

lib/compress/zstd_compress_internal.h
```diff
@@ -39,6 +39,38 @@
                                        The benefit is that ZSTD_DUBT_UNSORTED_MARK cannot be mishandled after table reuse with a different strategy.
                                        This constant is required by ZSTD_compressBlock_btlazy2() and ZSTD_reduceTable_internal() */
 
+#ifdef ZSTD_ENABLE_COMPRESS_FAST
+/* Lazy evaluation is performed only if the first match length 
+ * is <= ZSTD_compressFastLazyLimit */
+#define ZSTD_COMPRESS_FAST_BASE_LAZY_LIMIT 5
```
Cyan4973, line 45: That seems way too low.

In fact, in many cases, the search is targeting 5-byte matches, so all matches would trigger this condition, effectively changing `_lazy` into `_greedy`.

lib/compress/zstd_compress_internal.h
```diff
@@ -105,6 +137,9 @@ typedef struct {
     BYTE*  ofCode;
     size_t maxNbSeq;
     size_t maxNbLit;
+#ifdef ZSTD_ENABLE_COMPRESS_FAST
+    size_t lazyLimit;      /* Match length limit to allow lazy evaluation */
```
Cyan4973, line 141: This is unwelcome, I would prefer we avoid adding a conditional variable here.

lib/compress/zstd_lazy.c
```diff
@@ -1590,6 +1590,9 @@ size_t ZSTD_compressBlock_lazy_generic(
                                      prefixLowestIndex - (U32)(dictEnd - dictBase) :
                                      0;
     const U32 dictAndPrefixLength = (U32)((ip - prefixLowest) + (dictEnd - dictLowest));
+#ifdef ZSTD_ENABLE_COMPRESS_FAST
```
Cyan4973: too many such `#ifdef` blocks across the code base

lib/README.md
Cyan4973: I also see that there is a confusion between a C macro and a Makefile macro.
They are not the same.

tests/cli-tests/common/platform.sh
```diff
@@ -40,7 +40,7 @@ else
     hasMT="true"
 fi
 
-if zstd -vv --version | grep -q 'non-deterministic'; then
+if zstd -vv --version | grep -q 'non-deterministic' || [ "$ZSTD_COMPRESS_FAST" = "1" ]; then
```
Cyan4973: Removing determinism is pretty wild, though I guess it might be acceptable if this setting must be explicitly opt-in.

Cyan4973 (changes requested):
The core idea seems sound: do not search for a better match when you already have a "good enough" one.
This will save some work, and consequently reduce the amount of cpu spent in search.
It will also reduce opportunities to find better matches, and therefore negatively impact compression ratio.

So it's not a clear win situation, and more a trade off.
Therefore, what matters is the relative sizes of these trade off: how much is won, how much is lost.

And when looking at the numbers provided here, it does not look great.
The announced compression speed gains are significant, in the double digit range.
But we also take reduction in compression ratio pretty seriously. Something < 0.1% could be considered roughly negligible. But anything closer to ~1% is significant. And there are a few cases where compression ratio is negatively impacted by a significant amount.

I guess we could have a discussion here, about the pros and cons of the trade off, and maybe find another compromise point.

The second part though is about code complexity.
The core of the concept is not difficult, it adds a branch within the search loop, in `ZSTD_compressBlock_lazy_generic()`.

But unfortunately, the current implementation also adds a ton of other changes in other parts of the source code.
This is more difficult to maintain.
And considering the impact of the change, which merely offers a different position in the trade-off curve, I feel this is not a reasonable proposition. We need to keep things simpler if we want the source code to remain maintainable in the future.

As it stands, this PR is only suitable for a dedicated fork, where an independant maintenance policy can be set.

lib/README.md
BiplabRaut: Sure, we will address this

BiplabRaut:
> The core idea seems sound: do not search for a better match when you already have a "good enough" one. This will save some work, and consequently reduce the amount of cpu spent in search. It will also reduce opportunities to find better matches, and therefore negatively impact compression ratio.
> 
> So it's not a clear win situation, and more a trade off. Therefore, what matters is the relative sizes of these trade off: how much is won, how much is lost.
> 
> And when looking at the numbers provided here, it does not look great. The announced compression speed gains are significant, in the double digit range. But we also take reduction in compression ratio pretty seriously. Something < 0.1% could be considered roughly negligible. But anything closer to ~1% is significant. And there are a few cases where compression ratio is negatively impacted by a significant amount.
> 
> I guess we could have a discussion here, about the pros and cons of the trade off, and maybe find another compromise point.
> 
> The second part though is about code complexity. The core of the concept is not difficult, it adds a branch within the search loop, in `ZSTD_compressBlock_lazy_generic()`.
> 
> But unfortunately, the current implementation also adds a ton of other changes in other parts of the source code. This is more difficult to maintain. And considering the impact of the change, which merely offers a different position in the trade-off curve, I feel this is not a reasonable proposition. We need to keep things simpler if we want the source code to remain maintainable in the future.
> 
> As it stands, this PR is only suitable for a dedicated fork, where an independent maintenance policy can be set.

Thank you Yann for your detailed review comments.

Broadly, there are two main points:-
1) compression ratio tradeoffs :
   The compression ratio compromise is intentional and has been guided by a couple of our customer requests who are ok to such an aggressive tradeoff. For the same reason, this code can not be in default path so we introduced a separate macro.
2) maintainable simpler code changes
   I will get back on this on what we can do to address your points

Thanks again.

Issue #4547, opened 2025-12-10
Data race on `ctx->threadLimit` between `POOL_thread()` and `POOL_create_advanced()`


guoyuqi020:
It seems that the `ctx->threadLimit` variable is updated in `POOL_create_advanced()` without being protected by the `ctx->queueMutex`. This can lead to a data race. Will this be a problem?

https://github.com/facebook/zstd/blob/bfd8ad886f2948a8a0159a28e0aa6153611cd1c1/lib/common/pool.c#L67-L90

https://github.com/facebook/zstd/blob/bfd8ad886f2948a8a0159a28e0aa6153611cd1c1/lib/common/pool.c#L143-L159

sebres:
There is basically no RC at the moment, since the `ctx` (created here) isn't used somewhere else yet, excepting the workers in pool, where the first of them would hold the lock in:
https://github.com/facebook/zstd/blob/ae9f20ca2716f2605822ca375995b7d876389b64/lib/common/pool.c#L72
and then enter the conditional wait, because of `ctx->queueEmpty`, no matter what `ctx->threadLimit` is, because by `||` (or) the value of `(ctx->numThreadsBusy >= ctx->threadLimit)` is irrelevant:
https://github.com/facebook/zstd/blob/ae9f20ca2716f2605822ca375995b7d876389b64/lib/common/pool.c#L74-L75

But to avoid possible RC in the future (if the logic would be not enough anymore) one could indeed rewrite it a bit.
The naive attempt to "fix" it with lock around set of #4558 is in my opinion not correct (even wise versa, can cause a sporadic deadlock).
I'll provide a new PR for that.

PR #4558, opened 2025-12-23
Fix data race in POOL_create_advanced by protecting threadLimit

richardsonnick:
Fixes a potential data race in `POOL_create_advanced` where `ctx->threadLimit` was being updated without holding the `queueMutex`. Worker threads could access `threadLimit` before it was fully initialized, leading to undefined behavior.

The update to `ctx->threadLimit` is now protected by the `ctx->queueMutex`  to ensure thread safety during initialization.

`ctx->threadCapacity` remains unprotected, as it is only modified by the parent thread during setup.

Fix for #4547

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -146,15 +146,20 @@ POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize,
     /* Check for errors */
     if (!ctx->threads || !ctx->queue) { POOL_free(ctx); return NULL; }
     /* Initialize the threads */
-    {   size_t i;
+    {   
+        size_t i;
         for (i = 0; i < numThreads; ++i) {
             if (ZSTD_pthread_create(&ctx->threads[i], NULL, &POOL_thread, ctx)) {
                 ctx->threadCapacity = i;
                 POOL_free(ctx);
                 return NULL;
-        }   }
+            }   
+        }
         ctx->threadCapacity = numThreads;
+
+        ZSTD_pthread_mutex_lock(&ctx->queueMutex);
         ctx->threadLimit = numThreads;
+        ZSTD_pthread_mutex_unlock(&ctx->queueMutex);
     }
     return ctx;
 }


lib/common/pool.c
```diff
@@ -146,15 +146,20 @@ POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize,
...
             if (ZSTD_pthread_create(&ctx->threads[i], NULL, &POOL_thread, ctx)) {
                 ctx->threadCapacity = i;
                 POOL_free(ctx);
                 return NULL;
-        }   }
+            }   
+        }
         ctx->threadCapacity = numThreads;
+
+        ZSTD_pthread_mutex_lock(&ctx->queueMutex);
```
sebres, line 160: Holding the lock around the set of threadLimit wouldn't prevent possible RC at all, because the threads going created before the main thread would enter the lock here...
Let alone such set is nearby to atomic operation, so it is strange to try to protect it (alone) with a mutex.
Not to mention that there is basically no RC, since the ctx (created here) isn't used somewhere else yet, excepting the workers in pool, where the first of them would hold the lock in:
https://github.com/facebook/zstd/blob/ae9f20ca2716f2605822ca375995b7d876389b64/lib/common/pool.c#L72
and then enter the conditional wait, because of ctx->queueEmpty, no matter what ctx->threadLimit is, because by || (or) the value of (ctx->numThreadsBusy >= ctx->threadLimit) is irrelevant:
https://github.com/facebook/zstd/blob/ae9f20ca2716f2605822ca375995b7d876389b64/lib/common/pool.c#L74-L75
So this lock would not necessarily change something, still worse it may cause a typical deadlock, if the first worker thread would manage to enter the lock, and the main also trying to do that, so nobody would be able to put something to the queue and notify conditional wait to wake the worker up (which could also unlock the mutex for this thread).

terrelln: I don't think this could cause a deadlock. But, I agree that it is unnecessary.

terrelln (commented):
Thanks for the PR @richardsonnick! I'm going to close because PR #4601 supercedes.

lib/common/pool.c
sebres: > I don't think this could cause a deadlock

But it can, take a closer look at that scenario:
   1. pool master thread creates a 1<sup>st</sup>-worker, which enters the `POOL_thread` and then in line 72 it locks the mutex.
   2. pool master thread creates a 2<sup>nd</sup>- ... N<sup>th</sup>-workers, and the longer it takes (and the more context switches happens) the more likely the 1<sup>st</sup>-worker manages to enter the aforementioned lock.
   3. before leave the `POOL_create_advanced` the pool master thread also trying to lock the mutex, but it is locked by 1<sup>st</sup>-worker (which is waiting for the queue), so the master never leave `POOL_create_advanced` and `ctx` remains unknown for whole program, and nobody can add something to the queue to wake up the 1<sup>st</sup>-worker. It is typical deadlock.

Just for the record.

PR #4559, opened 2025-12-23
fix potential invalid mmap call and uninitialized file size

richardsonnick:
Fixes a warning from Clang 21 (unix.StdCLibraryFunctions) where mmap() could be called with a length of 0 if the dictionary file is empty. Mapping a zero-length range is undefined behavior under POSIX.

- Added guard for zero-length dictionary files.

### DIFF
--- a/programs/fileio.c
+++ b/programs/fileio.c
@@ -1040,6 +1040,7 @@ static size_t FIO_setDictBufferMMap(FIO_Dict_t* dict, const char* fileName, FIO_
         EXM_THROW(33, "Couldn't open dictionary %s: %s", fileName, strerror(errno));
     }
 
+    // filesize could be garbage if UTIL_getFileSizeStat fails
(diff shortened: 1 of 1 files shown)


programs/fileio.c
```diff
@@ -1049,6 +1050,11 @@ static size_t FIO_setDictBufferMMap(FIO_Dict_t* dict, const char* fileName, FIO_
         }
     }
 
+    if (fileSize == 0) return 0;
+    if (fileSize == (unsigned long long)-1) {
+        EXM_THROW(35, "Could not determine size of dictionary %s", fileName);
+    }
```
terrelln: Can you move this check up to just after line 1044, because the flow should never reach here, due to the check on 1046.

And please also check against `UTIL_FILESIZE_UNKNOWN` rather than -1.

terrelln (changes requested):
One minor change, otherwise looks good. Thanks for the PR!

PR #4560, opened 2025-12-23
correct mmap error check

richardsonnick:
According to the [mmap docs](https://man7.org/linux/man-pages/man2/mmap.2.html)  `MAP_FAILED` is returned upon error:

> On success, mmap() returns a pointer to the mapped area.  On
       error, the value MAP_FAILED (that is, (void *) -1) is returned,
       and [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is set to indicate the error.

Update mmap error validation to check for `MAP_FAILED` instead of NULL. POSIX specifies that mmap returns `MAP_FAILED` (-1) on failure.

Reported-by: Clang 21 Static Analyzer


terrelln (approved):
Thanks for the fix @richardsonnick!

PR #4568, opened 2026-01-16
Fix compiler warnings by explicit casting in divsufsort.c

rzikm:
On newer MSVC compilers, compilation of `divsufsort.c` causes `C4242` warnings.

```
divsufsort.c(438): error C4242: 'function': conversion from '__int64' to 'int', possible loss of data
```

This and C4244 are similar warnings which should not be generally suppressed.

This PR fixes those warnings by explicitly casting `ptrdiff_t` to `int` via a helper macro.

The alternative is to change certain variable types, but the current approach seemed like less risky change.

### DIFF
--- a/lib/dictBuilder/divsufsort.c
+++ b/lib/dictBuilder/divsufsort.c
@@ -25,22 +25,21 @@
  */
 
 /*- Compiler specifics -*/
-#ifdef __clang__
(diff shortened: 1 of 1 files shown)


terrelln (approved):

terrelln:
Thanks for the PR @rzikm!

PR #4578, opened 2026-01-30
Add support for zlib format

afbjorklund:
This is the compression frame and a checksum, without the gzip
header. It is used by for instance git objects, and is in zlib.

It is similar to the lzma frames being included in the xz support.
There are no real .zlib files, but it made testing more consistent.

-----

I love how `zstdcat` has transparent support also for legacy gz and xz files. And that it supports lzma (and lz4)

I was playing around with `.git/objects`, and wished that it would support those as well. So I added it, to zstd...

It is just a few magic bytes and a parameter, the support was already included in the zlib library since before.

Added a .zlib extension to match the others, but it is not _really_ a stand-alone file format (use .gz for that)

-----

PS. Yes, I know I could have used `git cat-file` as well.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -74,10 +74,10 @@ else
 endif
 
 # zlib detection
-NO_ZLIB_MSG := ==> no zlib, building zstd without .gz support
+NO_ZLIB_MSG := ==> no zlib, building zstd without .gz/.zlib support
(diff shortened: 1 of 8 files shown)


terrelln:
This seems like a reasonable feature to me, we will definitely consider this for our next release.

Issue #4581, opened 2026-02-02
Swift: unable to use any of the advanced/static-linking-only APIs


lukaskollmer:
**Describe the bug**
The Swift interface exposed via the `libzstd` target defined in Package.swift currently does not include any of the static-linking-only functions, and there is no way for a swift app/package depending on zstd to enable these APIs, even if zstd ends up getting statically linked.

**To Reproduce**
Add zstd as a dependency to a swift package, and try to use one of the static-only APIs (eg `ZSTD_isFrame`)

**Expected behavior**
Clearly it should be possible to always use the full API, regardless of which language it is being bridged into (provided you're statically linking the library).

**Possible Solutions**
There are multiple ways this could be resolved:
- we could introduce a shim header that sets the `ZSTD_STATIC_LINKING_ONLY` define and then includes zstd.h, and change the modulemap to include the shim header instead of directly including zstd.h
    - this would resolve the issue, but unconditionally requiring static linking via the modulemap seems like a bad idea, since it'd also affect other (non-swift-package-related) situations where the modulemap is used
- we could introduce a second, separate modulemap that is used only by the SPM package (i.e., the Package.swift file)
    - this approach would ensure that any existing code / setups that use the modulemap will continue to work unchanged

(Note: simply adding `.define("ZSTD_STATIC_LINKING_ONLY")` to the `cSettings` array in Package.swift does not work, since flags defined in this way only affect the compilation of the code, but do not affect the clang importer.)

**Additional context**
Tbh i have no idea if any / which other things (besides Swift's clang importer) use the modulemap file, and it might actually be safe to simply modify it in place, but i'm not comfortable doing that and think it'd be a bad idea, so adding the secondary, Swift-only one seems like the best approach.

Issue #3085, opened 2022-03-04
Alternative way to include unstable declarations


yfeldblum:
The current way of getting unstable declarations available only in statically-linked code is to set a macro and then include zstd. But this is clumsy for users in larger projects. For example, including one header that includes the zstd header, then including another header that includes the zstd header with unstable declarations, won't work. Effectively, including zstd unstable declarations from a library header is broken. As a workaround, it is often the case that libraries could wrap up the zstd unstable declarations in their own wrappers and only include zstd from a source file owned by the library. But this would not be acceptable for header-only C++ libraries.

An alternative would be to have a separate header for these declarations. Rather than setting a macro and then including zstd, users would just include zstd-static.

This could be done compatibly, where users could do either one at will. Effectively:

```
// zstd.h
#pragma once
#if ZSTD_STATIC_LINKING_ONLY
#include <zstd/unstable.h>
#endif
// ...
```

This would not break existing code and the current pattern of setting the macro before including zstd could still be encouraged. But library headers could get an alternative which is not broken.

Cyan4973:
> For example, including one header that includes the zstd header, then including another header that includes the zstd header with unstable declarations, won't work.

Could you describe what is the issue in above scenario ?
The current design of `zstd.h` is supposed to allow such pattern.

terrelln:
> Effectively, including zstd unstable declarations from a library header is broken.

`zstd.h` handles multiple includes correctly.

```c
#include <zstd.h>
#define ZSTD_STATIC_LINKING_ONLY
#include <zstd.h>
```

This works as expected (you'll get the static symbols on the 2nd include).

However, I agree that it is non-standard, and that it can cause problems with modules. And it would be nice to switch to a `zstd_unstable.h` header file.

yfeldblum:
Okay, this works as expected when directly including `zstd.h` because `zstd.h` internally has the required trickery.

However, when a library wraps `zstd.h` and does not match the required trickery, the library authors and authors of libraries and applications using the wrapping library may run into surprising behavior.

Beyond that, I agree that this interface is unconventional - even taking into account that glibc uses something like it. But, from one perspective, people are machines of convention. Unconventional interfaces require extra documentation, extra learning, and extra trial and error.

terrelln:
Bumping this issue. We should fix this in the next release.

PR #4582, opened 2026-02-02
SPM: ZSTDLIB_STATIC_API

lukaskollmer:
see #4581 for details

context / thoughts:
- SPM already very aggressively favors static linking over dynamic linking, so explicitly defining the library type as static should, effectively have no impact on this
    - not defining it explicitly allows SPM to choose between making the package product a static or dynamic library, which in effect means that it will be a static library, unless it contains resources, which is not the case here
    - if, at some point in the future, we'd want to offer dynamic linking for the Swift package, this would be possible: the convention would be to define a `libzstdDynamic` package product, whose kind is explicitly set to dynamic
- this PR introduces a second modulemap file
    - looking at the commit history, it seems that the current modulemap file exists solely to support the Package.swift file
    - but: i'm not 100% sure that there isn't some other toolchain/setup somewhere that also depends on the modulemap, so IMO defining a fully separate modulemap for these changes is better than modifying the existing one
    - since the new modulemap is in a dedicated `spm` directory, it will only be seen if clang is actually pointed at the `spm` directory (as is now happening in the Package.swift file)
    - any existing code should continue to work as-is, and not see the new modulemap at all (since there will be no existing code that `#include`s the spm directory

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,6 +12,7 @@ let package = Package(
         // Products define the executables and libraries a package produces, and make them visible to other packages.
         .library(
             name: "libzstd",
+            type: .static,
             targets: [ "libzstd" ])
     ],
     dependencies: [
@@ -25,7 +26,7 @@ let package = Package(
             name: "libzstd",
             path: "lib",
             sources: [ "common", "compress", "decompress", "dictBuilder" ],
-            publicHeadersPath: ".",
+            publicHeadersPath: "spm",
             cSettings: [
                 .headerSearchPath(".")
             ])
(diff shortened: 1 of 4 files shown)


terrelln:
@lukaskollmer I don't love having a separate shim for Swift in the repo here, with `#include "../zstd.h"`, it feels hacky. If it is possible to make this change by modifying the existing module map, I think we could consider just doing that.

I think it should be fine to have the module always define `ZSTD_STATIC_LINKING_ONLY`. IMO, we should move away from this and into two separate headers, because it doesn't play nicely with modules. But that is a change for another time. I've bumped the issue related to that https://github.com/facebook/zstd/issues/3085, and tagged as release blocking, so we can consider it for our next release.

PR #4583, opened 2026-02-02
MacOS release builds

### LINKED ISSUE #3818: Provide Linux & Darwin (macOS) builds via GitHub Releases
**Is your feature request related to a problem? Please describe.**

In short:

Downloading 500mb+ of brew to install something to unzip a 2mb file is, of course, silly.

In long:

I run https://webinstall.dev and our _thing_ is basically "instant installs from official builds, no dependencies, no sudo, no package manager, no change of accidentally borking your system".

I noticed that a few of the other packages we provide github installers for now provide `.tar.zstd` as an option and so I'd like to support that by providing `unzstd` like we do with `xz` - but I don't see any _official_ builds other than Windows (here in GH Releases).

**Describe the solution you'd like**

Release amd64, aarch64 (and preferably armv6) builds for Linux and macOS via <https://github.com/facebook/zstd/releases>

**Describe alternatives you've considered**

In the short term we won't offer `zstd` as an installable thing because we have no official source for it.

In the long term, although we don't host 3rd-party builds, if this is "done" I could fork it and host assets at a 3rd party GitHub repo and note that `zstd` is an exception to our "official builds" policy.

**Additional context**

That is all. Have a great day.

yawn:
Attempt to fix #3818 partially.

This is currently just a 1:1 copy of the windows release without any attempt to DRY up pinned `checkout` version (or other dependencies).

When triggered appropriately this will build `zstd` using the same structure as the windows equivalents. Both work when I tested it manually on 26.2 on an M2.

Missing is code signing so currently users need to execute `xattr -cr` on the binaries.

Any preferences here on how to approach this, e.g.

1. Document `xattr -cr`?
2. Ad-hoc signing with `codesign`?
3. `codesign` backed by a real developer id?

Open issues:

- [ ] Decide on code signing strategy (see above)

### DIFF
--- a/.github/workflows/macos-artifacts.yml
+++ b/.github/workflows/macos-artifacts.yml
@@ -1,0 +1,71 @@
+name: macos-artifacts
+
+on:
+  push:
+    branches: [test_artifacts, macos_artifacts, release]
+  release:
+    types:
+      - published
+
+permissions: read-all
+
+jobs:
+  macos-artifacts:
+    permissions:
+      contents: write # to fetch code and upload artifacts
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - { os: macos-15, ziparch: macos-arm64 }
+          - { os: macos-15, ziparch: macos-x86_64 }
+
+    steps:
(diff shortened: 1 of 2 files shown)


yawn:
Let me know if I should followup with DRY-ing up the workflows and potentially Linux here as well or if this is entirely out of scope for PR's.

PR #4584, opened 2026-02-04
Build: Add `-mno-strict-align` flag for RISC-V architecture

Polaris-911:
## Description
This PR updates the root `Makefile` to automatically detect RISC-V architectures (via `uname -m`) and append the `-mno-strict-align` flag to `MOREFLAGS`.

On many RISC-V implementations standard builds can suffer from significant performance degradation. Adding `-mno-strict-align` instructs the compiler to generate code that handles unaligned accesses via byte-wise instructions or other safe methods, thereby avoiding expensive traps and improving runtime performance.

## Changes
- Detects `riscv%` in `UNAME_M`.
- Appends `-mno-strict-align` to `MOREFLAGS`.
- Exports `MOREFLAGS` to ensure the flag is correctly propagated to sub-makefiles (lib, programs, etc.).

## Validation & Benchmarks
Tested on a RISC-V device (Sophon SG2044 / 64-core RISC-V).

**Benchmark Environment:**
- **Hardware:** Sophon SG2044
- **OS:** Linux (RISC-V 64-bit)
- **Compiler:** GCC 12.3.1

**Results:**
Comparison between the original build and the build with `-mno-strict-align`.

| Metric | Original | With `-mno-strict-align` | Improvement rate |
| :--- | :--- | :--- | :--- |
| **Compression Speed** | 64.0 MB/s | 75.2 MB/s | **+17.50%** |
| **Decompression Speed** | 102.9 MB/s | 165.1 MB/s | **+60.45%** |

*(Note: Data collected using `zstd -b` benchmark)*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,13 @@ else
   EXT =
 endif
 
+# Detect if the architecture is RISC-V (native build)
+UNAME_M ?= $(shell uname -m)
+ifneq (,$(filter riscv%,$(UNAME_M)))
+    MOREFLAGS += -mno-strict-align
+    export MOREFLAGS
+endif
+
 ## default: Build lib-release and zstd-release
 .PHONY: default
 default: lib-release zstd-release


terrelln:
@Polaris-911 all of our hot unaligned accesses are through these functions:

https://github.com/facebook/zstd/blob/dev/lib/common/mem.h#L162-L191

By default we use `memcpy()`, but can opt into other styles of access. I suspect a better approach here may be to opt into `MEM_FORCE_MEMORY_ACCESS` values `1` or `2`. Or perhaps adding another variant for RISC-V.

Polaris-911:
@terrelln Thanks for the suggestion. I opened a follow-up PR that updates `mem.h` to use `MEM_FORCE_MEMORY_ACCESS=2` for RISC-V when `__riscv` and `__riscv_zicclsm` are defined. Could you please take a look when you have a moment?
https://github.com/facebook/zstd/pull/4596

PR #4601, opened 2026-03-01
avoid potential RC on ctx->threadLimit (gh-4547), code review

### LINKED ISSUE #4547: Data race on `ctx->threadLimit` between `POOL_thread()` and `POOL_create_advanced()`

sebres:
As already said in #4547, there is no RC currently, but for consistence reasons and to avoid potential RC in the future by possible changes...
Possible alternative could be to lock the fresh created mutex before first worker thread gets created (and unlock it before return).
Closes #4547, replaces #4558

### DIFF
--- a/lib/common/pool.c
+++ b/lib/common/pool.c
@@ -142,19 +142,17 @@ POOL_ctx* POOL_create_advanced(size_t numThreads, size_t queueSize,
     /* Allocate space for the thread handles */
     ctx->threads = (ZSTD_pthread_t*)ZSTD_customCalloc(numThreads * sizeof(ZSTD_pthread_t), customMem);
     ctx->threadCapacity = 0;
+    ctx->threadLimit = numThreads;
     ctx->customMem = customMem;
     /* Check for errors */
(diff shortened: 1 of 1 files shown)


lib/common/pool.c
```diff
@@ -220,23 +218,22 @@ static int POOL_resize_internal(POOL_ctx* ctx, size_t numThreads)
         return 0;
     }
     /* numThreads > threadCapacity */
+    ctx->threadLimit = numThreads;
```
terrelln, line 221: This is a minor change in behavior when an operation fails below, but it seems fine & slightly preferable.

terrelln (approved):

PR #4612, opened 2026-03-06
compress: vectorize ZSTD_count() with SSE2

rmilkowski:
## Summary

Add an SSE2 fast path to `ZSTD_count()` on x86/x86_64.

After the initial machine-word compare, the new path compares 16 bytes at a time with `_mm_cmpeq_epi8()` / `_mm_movemask_epi8()` and uses `ZSTD_countTrailingZeros32()` to jump to the first mismatch. Non-x86 targets and builds without SSE2 keep the existing scalar path.

## Why

`ZSTD_count()` is a hot helper in match finding. When matches extend past the first word, counting them 16 bytes at a time reduces scalar work on match-heavy inputs.

## Benchmark

Host: Intel Xeon Gold 6254, Linux x86_64

Method:
- levels `1..22`
- `3` runs per file / level
- page cache warmed before each run

Overall:
- mean compression delta: `+16.1%`
- median compression delta: `+8.3%`
- suite CPU time: `-1.8%`

Largest gains:
- `pattern_64kb_gap_0kb_128mb.bin`: `+98.3%` at level `22`
- `pattern_64kb_gap_0kb_128mb.bin`: `+86.0%` at level `10`
- `pattern_64kb_gap_0kb_128mb.bin`: `+75.3%` at level `9`
- `runlength_64kb_pad64_32mb.bin`: `+54.7%` at level `10`

Largest regressions observed:
- `pattern_1kb_gap_1023kb_128mb.bin`: `-6.7%` at level `8`
- `high_entropy_190sym_64mb.bin`: `-4.6%` at level `5`
- `pattern_1kb_gap_1023kb_128mb.bin`: `-2.7%` at level `1`

Overall the tradeoff looks favorable: large wins on repetitive / match-heavy inputs, with limited downside on sparse or high-entropy inputs.

## Testing

- `playTests.sh`
- CLI tests
- `test-fullbench`
- `test-fuzzer`
- `test-zstream`
- `test-invalidDictionaries`
- `test-legacy`
- `test-decodecorpus`
- `test-pool`
- `test-longmatch`
- `CFLAGS='-Werror -O2' make -j1 lib zstd`
- `make c89build`
- `make gnu90build`
- `make c99build`
- `make cxxtest`
- valgrind checks
- CMake configure / build

## Notes

- No API change.
- No format change.

### DIFF
--- a/lib/compress/zstd_compress_internal.h
+++ b/lib/compress/zstd_compress_internal.h
@@ -866,8 +866,8 @@ MEM_STATIC size_t ZSTD_count(const BYTE* pIn, const BYTE* pMatch, const BYTE* co
 #if defined(ZSTD_ARCH_X86_SSE2)
         if ((size_t)(pInLimit - pIn) >= 32) {
             while (pIn < pInLimit16) {
-                __m128i const matchVec = _mm_loadu_si128((const __m128i*)pMatch);
-                __m128i const inVec = _mm_loadu_si128((const __m128i*)pIn);
+                __m128i const matchVec = _mm_loadu_si128((const __m128i*)(const void*)pMatch);
+                __m128i const inVec = _mm_loadu_si128((const __m128i*)(const void*)pIn);
                 U32 const matchMask = (U32)_mm_movemask_epi8(_mm_cmpeq_epi8(matchVec, inVec));
                 if (matchMask != 0xFFFF) {
                     U32 const diffMask = ~matchMask & 0xFFFF;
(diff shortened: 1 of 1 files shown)


terrelln:
Thanks for the PR @rmilkowski! I've implemented this same thing several times, and I've always found that on real data it hasn't shown any significant gains. You should benchmark on each file in the [Silesia corpus](https://sun.aei.polsl.pl/~sdeor/index.php?page=silesia) to start.

Let me get back to you with more details about the benchmarks I've run to test this.

rmilkowski:
Thanks. I reran the benchmarks on the full Silesia corpus, across all files and all compression levels.

I agree that on this kind of real data there isn’t a noticeable improvement from this change. While it can show gains on generated / synthetic inputs, that benefit does not seem to translate into a meaningful real-data win here.

PR #4613, opened 2026-03-07
Fix potential NULL pointer dereference in ZSTD_customCalloc when cust…

nidu-ninja:
…om allocator fails

The fix introduces a NULL check immediately after the custom allocation call. If allocation fails, the function now safely returns NULL instead of dereferencing the pointer.

This improves the robustness of the memory allocation path and prevents possible crashes when using custom memory allocators.


terrelln (approved):
Thanks for the fix @nidu-ninja!

Cyan4973 (approved):

PR #4618, opened 2026-03-09
dibio: enforce minimum sample size for dictionary training

BhavyaBibra:
## Summary

Enforce a minimum sample size of 8 bytes when gathering training data for dictionary building. Files smaller than this threshold provide negligible training value and may cause edge-case issues in the dictionary builder.

This resolves two TODO comments in [dibio.c](cci:7://file:///Users/bhavyabibra/Documents/AG/zstd-Meta/zstd/programs/dibio.c:0:0-0:0):
- Line 282: `/* TODO: is there a minimum sample size? What if the file is 1-byte? */`
- Line 297: `/* TODO: is there a minimum sample size? Can we have a 1-byte sample? */`

### [programs/dibio.c](cci:7://file:///Users/bhavyabibra/Documents/AG/zstd-Meta/zstd/programs/dibio.c:0:0-0:0)

**In [DiB_fileStats()](cci:1://file:///Users/bhavyabibra/Documents/AG/zstd-Meta/zstd/programs/dibio.c:280:0-356:1) — file stats gathering (line ~282):**
- Files with size > 0 but < 8 bytes are now skipped with a `DISPLAYLEVEL(3)` warning
- Prevents tiny files from inflating `nbSamples` count without meaningful data

**In [DiB_fileStats()](cci:1://file:///Users/bhavyabibra/Documents/AG/zstd-Meta/zstd/programs/dibio.c:280:0-356:1) — chunked sample path (line ~297):**
- Same 8-byte minimum applied when files are split into sample chunks
- Skips with warning instead of hard error (backwards-compatible)

## Design Decisions

- **Threshold of 8 bytes** was chosen as a conservative minimum — large enough to filter noise, small enough to avoid rejecting legitimate edge cases
- **Warning, not error** — files are skipped with a log message at `displayLevel >= 3`, not a fatal error. This preserves backwards compatibility for scripts that may pass heterogeneous file lists
- **No new CLI flag** — the threshold is not configurable. If maintainers prefer a configurable option, this can be extended

## Testing

```bash
make -C programs zstd  # ✅ builds successfully with changes

### DIFF
--- a/programs/dibio.c
+++ b/programs/dibio.c
@@ -8,100 +8,112 @@
  * You may select, at your option, one of the above-listed licenses.
  */
 
-
-
 /* **************************************
-*  Compiler Warnings
-****************************************/
+ *  Compiler Warnings
+ ****************************************/
 #ifdef _MSC_VER
-#  pragma warning(disable : 4127)    /* disable: C4127: conditional expression is constant */
+#pragma warning(                                                               \
+    disable : 4127) /* disable: C4127: conditional expression is constant */
 #endif
 
-
 /*-*************************************
-*  Includes
-***************************************/
-#include "platform.h"       /* Large Files support */
-#include "util.h"           /* UTIL_getFileSize, UTIL_getTotalFileSize */
-#include <stdlib.h>         /* malloc, free */
-#include <string.h>         /* memset */
-#include <stdio.h>          /* fprintf, fopen, ftello64 */
-#include <errno.h>          /* errno */
+ *  Includes
+ ***************************************/
+#include "platform.h" /* Large Files support */
+#include "util.h"     /* UTIL_getFileSize, UTIL_getTotalFileSize */
+#include <errno.h>    /* errno */
+#include <stdio.h>    /* fprintf, fopen, ftello64 */
+#include <stdlib.h>   /* malloc, free */
+#include <string.h>   /* memset */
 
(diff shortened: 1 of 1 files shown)


Cyan4973 (changes requested):
While the core idea seems sound—enforcing a minimum sample size for dictionary training is a good safeguard, and 8 bytes is indeed extremely small. I also wouldn’t be surprised if a similar constraint is already enforced elsewhere in the pipeline.

However, a tactically detrimental aspect of this PR is that it bundles an unrelated change in source-code indentation style. This causes a large number of purely mechanical diffs across multiple files, which makes it difficult (if not impossible) to determine what functional behavior is actually changing and what the real impact is.

Given the large attack surface exposed through Zstandard, we cannot accept external contributions unless the changes are fully understood.

Please refactor this PR to include only the minimal, essential functional changes, so the review can be completed safely and with high confidence.

BhavyaBibra:
Thank you for the review and for pointing that out! I apologize for the noise—my editor's auto-formatter was applying unintended whitespace changes on save.

I've updated the PR and force-pushed a clean version. The diff now contains strictly the minimal, essential functional changes to enforce the 8-byte minimum sample size constraint without any unrelated formatting changes.

If you compare the changes against the dev branch (e.g., git diff origin/dev..HEAD programs/dibio.c), you'll see it is now completely isolated to just those specific logical additions.

Let me know if this looks good to move forward with the review!

Issue #4455, opened 2025-08-06
Promote `-c` option to standard option displayed by `-h`


tats-u:
**Is your feature request related to a problem? Please describe.**

Other similar tools (gzip & bzip2 & lz4 & xz) display `-c` option in the standard help by `-h`, but zstd does not.

```
$ gzip -h
Usage: gzip [OPTION]... [FILE]...
Compress or uncompress FILEs (by default, compress FILES in-place).

Mandatory arguments to long options are mandatory for short options too.

  -c, --stdout      write on standard output, keep original files unchanged
  -d, --decompress  decompress
  -f, --force       force overwrite of output file and compress links
  -h, --help        give this help
  -k, --keep        keep (don't delete) input files
  -l, --list        list compressed file contents
  -L, --license     display software license
  -n, --no-name     do not save or restore the original name and timestamp
  -N, --name        save or restore the original name and timestamp
  -q, --quiet       suppress all warnings
  -r, --recursive   operate recursively on directories
      --rsyncable   make rsync-friendly archive
  -S, --suffix=SUF  use suffix SUF on compressed files
      --synchronous synchronous output (safer if system crashes, but slower)
  -t, --test        test compressed file integrity
  -v, --verbose     verbose mode
  -V, --version     display version number
  -1, --fast        compress faster
  -9, --best        compress better

With no FILE, or when FILE is -, read standard input.

Report bugs to <bug-gzip@gnu.org>.

$ bzip2 -h
bzip2, a block-sorting file compressor.  Version 1.0.8, 13-Jul-2019.

   usage: bzip2 [flags and input files in any order]

   -h --help           print this message
   -d --decompress     force decompression
   -z --compress       force compression
   -k --keep           keep (don't delete) input files
   -f --force          overwrite existing output files
   -t --test           test compressed file integrity
   -c --stdout         output to standard out
   -q --quiet          suppress noncritical error messages
   -v --verbose        be verbose (a 2nd -v gives more)
   -L --license        display software version & license
   -V --version        display software version & license
   -s --small          use less memory (at most 2500k)
   -1 .. -9            set block size to 100k .. 900k
   --fast              alias for -1
   --best              alias for -9

   If invoked as `bzip2', default action is to compress.
              as `bunzip2',  default action is to decompress.
              as `bzcat', default action is to decompress to stdout.

   If no file names are given, bzip2 compresses or decompresses
   from standard input to standard output.  You can combine
[...]

PR #4619, opened 2026-03-09
Promote -c flag

jlee303:
Promote -c to be standard option displayed by -h #4455 
Thanks for the suggestion @tats-u !

### DIFF
--- a/programs/zstdcli.c
+++ b/programs/zstdcli.c
@@ -168,8 +168,7 @@ static void usage(FILE* f, const char* programName)
     DISPLAY_F(f, "  -f, --force                   Disable input and output checks. Allows overwriting existing files,\n");
     DISPLAY_F(f, "                                receiving input from the console, printing output to STDOUT, and\n");
     DISPLAY_F(f, "                                operating on links, block devices, etc. Unrecognized formats will be\n");
-    DISPLAY_F(f, "                                passed-through through as-is.\n");
-    DISPLAY_F(f, "  -q, --quiet                   Suppress warnings; pass twice to suppress errors.\n\n");
+    DISPLAY_F(f, "                                passed-through through as-is.\n\n");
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
While moving `-c/--stdout` in a more visible place is straightforward and correctly done in this PR,

the situation is less clear for `-q/--quiet`.
I think it makes more sense to keep `-q` alongside `-v`, as they both impact the same thing (i.e. the amount of verbosity). Other `-h` help screens agree with this principle, with the exception of those which follow the alphabetical order strictly.

So we could also bring `-v` along with `-q`? But now it introduces another question: which options deserves to be tagged "important", and be promoted to first help screen (as opposed to the more advanced and complete `-H`). We can't just promote everything blindly, because if everything is important, nothing is.

In order to keep this PR short, I would separate the 2 issues, so that `-c/--stdout` can be validated and merged without delay nor uncertainty related to other less obvious topics.

jlee303:
> While moving `-c/--stdout` in a more visible place is straightforward and correctly done in this PR,
> 
> the situation is less clear for `-q/--quiet`. I think it makes more sense to keep `-q` alongside `-v`, as they both impact the same thing (i.e. the amount of verbosity). Other `-h` help screens agree with this principle, with the exception of those which follow the alphabetical order strictly.
> 
> So we could also bring `-v` along with `-q`? But now it introduces another question: which options deserves to be tagged "important", and be promoted to first help screen (as opposed to the more advanced and complete `-H`). We can't just promote everything blindly, because if everything is important, nothing is.
> 
> In order to keep this PR short, I would separate the 2 issues, so that `-c/--stdout` can be validated and merged without delay nor uncertainty related to other less obvious topics.

Makes sense! I've updated pr and title to only promote -c. Let me know if there's anything else :)

Cyan4973 (approved):

Issue #4492, opened 2025-09-22
Expose definable definitions to make in Makefile.


tansy:
Expose definable definitions to make in Makefile.

There are some definitions* in Makefile that seem to be exposed to external definition but there is no way to pass them to Makefile.
Can you add some variable like \`USER_CFLAGS' so user could pass additional options to Makefile and not have to manually change command line or edit the file.
And describe example in the Readme.

Or, if it's already possible describe it in Readme.

* like ZSTDCLI_CLEVEL_DEFAULT, ZSTDCLI_NBTHREADS_DEFAULT, ZSTD_NOBENCH, etc.

----

Also a request to enclose DISPLAY_LEVEL_DEFAULT in \`#ifdef ... #endif', so it too could be passed through makefile.
```
+#ifndef DISPLAY_LEVEL_DEFAULT
 #define DISPLAY_LEVEL_DEFAULT 2
+#endif
```
Here's a [diff](https://dpaste.com/9AGC8ZEQA.txt) for convenience. Just add a new line at the end.

Cyan4973:
These are not `Makefile` variables, but C Preprocessor variables.

The way to [pass such variables to the compiler](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html?utm_source=chatgpt.com) via `make` is presumed standard, and looks like this:
```
CPPFLAGS="-DZSTDCLI_CLEVEL_DEFAULT=19 -DZSTDCLI_NBTHREADS_DEFAULT=16" make
```

It would be possible to add a paragraph in `programs/README.md` to explain that process and list preprocessor variables.

tansy:
> These are not `Makefile` variables, but C Preprocessor variables.

I know but that's not what I meant. Let me explain it - if a user passes CFLAGS to make it will replace the existing one.

```
$ echo "int main(){ return 0; }" > prog.c
$ cat << EOF > Makefile
CFLAGS += -O2 -Wall -Iinclude -DDEFINE1
LD=$(CC)

.c.o:
        $(CC) $(CFLAGS) -c -o $@ $<

prog: prog.o
        $(LD) -o $@ $+
EOF

$ make CFLAGS='-DDISPLAY_LEVEL_DEFAULT=1' -n
cc -DDISPLAY_LEVEL_DEFAULT=1 -c -o prog.o prog.c
cc -o prog prog.o
```

What I would expect to see is
```
[...]

PR #4620, opened 2026-03-10
Make DISPLAY_LEVEL_DEFAULT configurable at compile time

jlee303:
Enables DISPLAY_LEVEL_DEFAULT to be configured at compile time by wrapping it in #ifndef guards and adding the ZSTD_ prefix to avoid namespace pollution. Allows for quieter output without need to pass in -q every invocation.

To verify change, compare difference in verbosity with `make` vs `CPPFLAGS='-DZSTD_DISPLAY_LEVEL_DEFAULT=3' make`

Thanks @tansy for suggestion! #4492


contrib/largeNbDicts/largeNbDicts.c
```diff
@@ -45,7 +45,9 @@
 #define RUN_TIME_DEFAULT_MS    1000
 #define BENCH_TIME_DEFAULT_MS (BENCH_TIME_DEFAULT_S * RUN_TIME_DEFAULT_MS)
 
-#define DISPLAY_LEVEL_DEFAULT 3
+#ifndef ZSTD_DISPLAY_LEVEL_DEFAULT
+#define ZSTD_DISPLAY_LEVEL_DEFAULT 3
```
Cyan4973: minor: indent:
in the `zstd` code base, within an `#ifdef` section, we typically write 
```
# define ZSTD_DISPLAY_LEVEL_DEFAULT N
```
notice the space between `#` and `define`.

jlee303: Updated! Let me know if there's anything else :)

Cyan4973 (approved):

PR #4637, opened 2026-04-06
harden filename construction against integer overflow and OOB access

metsw24-max:
This patch hardens filename construction logic in `fileio.c` against integer overflow, out-of-bounds access, and unsafe string operations.

Added precomputed length handling and overflow-checked capacity calculation

Fixed separator logic when output directory is empty to avoid out-of-bounds indexing

Added explicit destination-capacity overflow guards in compressed-name generation

Replaced unsafe suffix write with length-aware copy


metsw24-max:
any update on this pr

PR #4638, opened 2026-04-06
harden seek table parsing against overflow and oversized frame metadata

metsw24-max:
This hardens seek table parsing in `zstdseek_decompress.c` by adding strict bounds checks for untrusted metadata before size arithmetic and memory allocation

Added a hard upper bound check for numFrames from seek table metadata

Added checked arithmetic for table size and frame size computations before multiply/add operations

Added allocation-size overflow guards when allocating seek entries in load path and clone path

These checks reduce memory safety risk by ensuring untrusted metadata cannot trigger wrapped size calculations or undersized allocations.


metsw24-max:
gentle ping

PR #4640, opened 2026-04-08
harden output file handling against race conditions and symlink attacks

metsw24-max:
This change hardens the CLI output file creation path by eliminating race-prone patterns and preventing symlink-based overwrite attacks.

The update replaces the existing unlink-and-recreate workflow with atomic file creation semantics and introduces additional safety checks for output file handling

Replace non-atomic `unlink() + open()` pattern with atomic open using `O_EXCL` for non-overwrite mode

Prevent symlink traversal during output file creation using `O_NOFOLLOW`

Add `O_CLOEXEC` to reduce file descriptor leakage across exec boundaries

Introduce explicit handling of `EEXIST` and `ELOOP` for clearer and safer error reporting


metsw24-max:
Pushed a fix for the FreeBSD job. The open flags were being built with a couple of statements between the two declarations, which trips -Wdeclaration-after-statement under -Werror on that build. Folded them back into the initialiser so it stays C89-clean, and marked the new cli-test executable so the runner picks it up.

metsw24-max:
any update?

PR #4668, opened 2026-05-14
riscv: add prefetch optimization for doubleFast compressor

Felix-Gong:
## Summary
- Add RISC-V 64-bit prefetch optimization to zstd doubleFast block compressor
- Extend existing aarch64 `PREFETCH_L1(ip+256)` to RISC-V via macro guard
- 2 lines changed in `zstd_double_fast.c`

## Changes
Extended the architecture guard from `#if defined(__aarch64__)` to `#if defined(__aarch64__) || (defined(__riscv) && (__riscv_xlen == 64))` in two locations:
1. `ZSTD_compressBlock_doubleFast_noDict_generic` (line 236)
2. `ZSTD_compressBlock_doubleFast_dictMatchState_generic` (line 451)

## Performance Results (RISC-V 64-bit, RV64GC)

| Data Type | Compress | Decompress |
|-----------|----------|------------|
| Random (100MB) | +9.0% (337.3 → 367.6 MB/s) | +8.0% (2763.0 → 2983.9 MB/s) |
| Zero (100MB) | -0.3% (within noise) | -0.05% (within noise) |

## Testing
- Platform: RISC-V 64-bit server
- Compiler: GCC 15.1
- Test data: Random and zero-filled data (100MB each)
- Benchmark: `zstd -b`

## Notes
- The prefetch targets sequential data access patterns in the compression loop
- No additional computation added to the hot loop
- Follows the same pattern as the existing aarch64 optimization

### DIFF
--- a/lib/compress/zstd_double_fast.c
+++ b/lib/compress/zstd_double_fast.c
@@ -233,7 +233,7 @@ size_t ZSTD_compressBlock_doubleFast_noDict_generic(
             hl0 = hl1;
             idxl0 = idxl1;
             matchl0 = matchl1;
-    #if defined(__aarch64__)
+    #if defined(__aarch64__) || (defined(__riscv) && (__riscv_xlen == 64))
             PREFETCH_L1(ip+256);
     #endif
         } while (ip1 <= ilimit);
@@ -448,7 +448,7 @@ size_t ZSTD_compressBlock_doubleFast_dictMatchState_generic(
         }   }
 
         ip += ((ip-anchor) >> kSearchStrength) + 1;
-#if defined(__aarch64__)
+#if defined(__aarch64__) || (defined(__riscv) && (__riscv_xlen == 64))
         PREFETCH_L1(ip+256);
 #endif
         continue;


luhenry:
@gctony hey :wave: would you know who would be a good person to review this? Thank you!

Cyan4973:
This is a fairly simple and risk-free modification.
It should not impact any other architecture beyond `risc-v`.
And even on `risc-v`, it only impacts speed, not correctness.

This repository is only equipped to validate correctness, via `qemu`.
For speed, we are entirely reliant on external measurements, 
keeping in mind that results can vary a lot depending on exact silicon.

In this case, my only gripe is that speed has been measured on 2 opposite corner cases,
impossible to compress, or a bunch of zeroes,
none of which being representative of any real-world scenario.

For something a bit more representative,
I would recommend using the `lorem ipsum` generator,
which is provided as a stand-alone tool `tests/datagen`,
or the synthetic generator directly integrated within `zstd -b`.
Both will provide "reasonably compressible" data, without falling into either extreme.

Also: this should really only impact compression speed, given the modification only mutates the source file for compression levels 3 & 4. Speed impact on the decompression side should be regarded as fishy, or at least unintentional.

Issue #4667, opened 2026-05-13
ZSTD SW fails to decompress a valid compressed file


qahsan99:
**Describe the bug**

**To Reproduce**
1. Downloads data '...'
2. Run '...' with flags '...'
3. Scroll up on the log to '....'

**Expected behavior**

**Screenshots and charts**
If applicable, add screenshots and charts to help explain your problem.

**Desktop (please complete the following information):**
 - OS: [e.g. Mac]
 - Compiler [e.g. gcc]
 - Flags [e.g. O2]
 - Other relevant hardware specs [e.g. Dual-core]
 - Build system [e.g. Makefile]

**Additional context**

terrelln:
@qahsan99 The body of this issue seems to not be filled out. Until it is this is not actionable.

qahsan99:
The File is poorly compressed 

-     Cleartext size is 7245 bytes
-     It has 1 token of length 10
-     The rest are literals
-     As a result the compressed file size (7262) > cleartext size (7245)
-     The SW v1.5.4 can decompress it
-     But the latest  SW v1.5.7 cannot decompress it. It gives the Error  "Decoding error (36) : Src size is incorrect"
- 

[Files.tar.gz](https://github.com/user-attachments/files/27735439/Files.tar.gz)

qahsan99:
Hi Nick, I have given all the info, please let me know if you need anything else

qahsan99:
From RFC9978

[3.1.1.2.4. ](https://datatracker.ietf.org/doc/html/rfc8878#section-3.1.1.2.4)[Block_Content and Block_Maximum_Size](https://datatracker.ietf.org/doc/html/rfc8878#name-block_content-and-block_max)
The size of Block_Content is limited by Block_Maximum_Size, which is the smallest of:

Window_Size
128 KB
Block_Maximum_Size is constant for a given frame. This maximum is applicable to both the decompressed size and the compressed size of any block in the frame.

The reasoning for this limit is that a decoder can read this information at the beginning of a frame and use it to allocate buffers. The guarantees on the size of blocks ensure that the buffers will be large enough for any following block of the valid frame.

**If the compressed block is larger than the uncompressed one, sending the uncompressed block (i.e., a Raw_Block) is recommended instead.**

terrelln:
@qahsan99 do you know how the compressed file was produced?

terrelln:
[...]

PR #4669, opened 2026-05-14
[doc] Clarify requirements around Block_Maximum_Size in spec

terrelln:
Clarify that it is legal to send compressed blocks that are larger than the uncompressed size so long as their `Block_Content` is no larger than `Block_Maximum_Size`.

See https://github.com/facebook/zstd/issues/4667

### DIFF
--- a/doc/zstd_compression_format.md
+++ b/doc/zstd_compression_format.md
@@ -16,7 +16,7 @@ Distribution of this document is unlimited.
 
 ### Version
 
-0.4.4 (2025-03-22)
+0.4.5 (2026-05-14)
 
 
 Introduction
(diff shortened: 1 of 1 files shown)


doc/zstd_compression_format.md
```diff
@@ -401,6 +401,11 @@ at the beginning of a frame and use it to allocate buffers.
 The guarantees on the size of blocks ensure that
 the buffers will be large enough for any following block of the valid frame.
 
+If the compressed block is larger than the uncompressed one, but the
+`Block_Content` is no larger than `Block_Maximum_Size`, it is legal to
+send the compressed block. However, sending the uncompressed block
+(i.e., a `Raw_Block`) is recommended instead.
```
Cyan4973: I would phase it this way:

```
If a compressed block is larger than its uncompressed content, 
it is recommended to send it uncompressed (i.e., a `Raw_Block`).
However, as long as `Block_Content` is no larger than `Block_Maximum_Size`, 
it is legal to send such a compressed block, 
even if it's larger than its uncompressed content.

If `Block_Content` is larger than `Block_Maximum_Size`, then this is out of spec,
and a compliant decoder is allowed to reject such block.
```

terrelln: I added the first paragraph, and folded the second into some other changes to the beginning of the section.

Cyan4973 (approved):

